PR #726, opened 2019-10-19
Optimize/delete redundant cypress tests

sandrina-p:
Our Cypress tests limit per 2 weeks is 500 tests. So I reviewed all tests and merged those that are redundant/too small. Reduced the number from 32 to 17. So we can run 2x more tests :D


taoeffect (approved):
Nice 👍

PR #727, opened 2019-10-20
New And Improved i18n tag! Supports all HTML and components!

taoeffect:
**IMPORTANT: PLEASE READ:**

From now on:

**DO NOT** use the `html` attribute. Instead, use `:args` to include HTML opening/closing tags, and add `compile` attribute. Anywhere you see it used, boyscout in your PRs conversion to child text like so:

```pug
// INSTEAD OF:

i18n(
  tag='p'
  html='My <b>great</b> text <router-link to="page">Foo</router-link>'
)

// REWRITE AS:

i18n(
  tag='p'
  :args='{ b1: "<b>", b2: "</b>", r1: "<router-link to=\'page\'>", r2: "</router-link>" }'
  compile
) My {b1}great{b2} text {r1}Foo{r2}
```

This ensures translators do not have to deal with long strings containing confusing HTML, and speeds up rendering by calling `Vue.compile` only when necessary.


frontend/views/utils/translations.js
```diff
@@ -28,3 +26,47 @@ export default function L (
...
+    if (!translation) {
+      console.warn('The following i18n text was not translated correctly:', text)
+      return h(context.props.tag, context.data, text)
+    }
+    if (context.props.html) {
+      console.warn("i18n: using 'html' as string is deprecated! Instead: set 'compile' attribute and place HTML opening/closing tags in :args variables, then place text as child of i18n!", { text })
+      context.props.compile = true
+    }
+    if (context.props.compile) {
+      var result = Vue.compile('<wrap>' + translation + '</wrap>')
```
sandrina-p, line 54: what's this `wrap` element for?

taoeffect: This is because Vue.js requires a top-level tag when compiling templates

sandrina-p (approved):

Issue #688, opened 2019-09-27
Error, success and app error banners [Discussion]


mmbotelho:
# Problem

At the moment, the app does not communicate any errors or success messages with the users.
_Note: This issue does not address notifications. For notifications, see #663_

# Solution
Create a component to be used to inform users of the state of the app.

### What types of messages are we displaying?
There are 3 types of messages we want to display to the user:
1. **Form error** - these errors are displayed whenever the user makes a mistake when filling out a text / number / textarea / password input. These errors have already been designed and can be found in our [design system](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=48%3A4) and for this reason, I won't write a lot about these errors.
2. **User action feedback** - these banners let the user know that a certain action was completed successfully OR that an error occurred, and what they can do about it. These errors are more general than form errors, because they concern the action as a whole (not an error in one specific input) or errors with other elements that are not inputs (e.g.: avatar upload);
2. **App Error** - informs the user that there's something going on with the app itself, or with something that is not related with a user action. 

### User action feedback (success and error banners)
<img width="1103" alt="Captura de ecrã 2019-09-27, às 17 57 36" src="https://user-images.githubusercontent.com/22004566/65783601-6a62bd80-e150-11e9-8996-1b9bdac3f50c.png">
<img width="665" alt="Captura de ecrã 2019-09-27, às 17 57 59" src="https://user-images.githubusercontent.com/22004566/65783603-6a62bd80-e150-11e9-8b70-b40bae932ceb.png">

- The user can dismiss them, by clicking on the X button;
- Navigating to a different page automatically removes them;
- The errors are displayed in-context, immediately below the component where the error / success occured;
- There cannot be 2 banners for the same context - however, there can be more than one banner in a page (as long as they are not in the same group of information);

----

### App Error 

<img width="634" alt="Captura de ecrã 2019-09-27, às 17 58 09" src="https://user-images.githubusercontent.com/22004566/65783644-7cdcf700-e150-11e9-9c96-2b6e5abd1f89.png">

- The user cannot dismiss them;
- They remain visible until the problem is solved;
- These errors are global, and remain visible even when the user changes pages / opens modals, etc.
- Only one of these can be displayed at a time - if there's the necessity for the app to output more than one mistake, this just means that we are not creating the right errors;
- Because this error is global and needs to be displayed in all pages, it is fixed at the top of the page, and does not overlap content.

[...]

PR #730, opened 2019-10-20
Fix/create card banner fix dropdown styles and router issue

### LINKED ISSUE #696: Improve router usage after a group is created.
### Problem

Context: https://github.com/okTurtles/group-income-simple/pull/679/files#r329122719

There was a weird rare false positive in a test failure in this step. Let me explain:

image

This view is visible for 1ms. It happens immediately after a user with a group logs in. How is that possible? Well...
Let's see the current conditions for each view:
- sidebar: isloggedIn and hasGroups
- welcomeGroup: isloggedIn

So, both conditions are indeed true.

If we look at the code, after the login happens, there's a tiny moment where we verify it the user has a group and then ask $router to redirect to `/dashboard`.
```js
// Login.vue#L91
await sbp('state/vuex/dispatch', 'login', { username, identityContractID })
if (this.$store.state.currentGroupId) {
  this.$router.push({ path: '/dashboard' })
}
```
That's the exactly moment where this interface (the screenshot) takes place in time.

For that reason, the condition to find "welcomeHomeLoggedIn" (L33) is positive. However, in the next millisecond, when we try to get the element again and assert the text, the element isn't there anymore because the $router has changed to `/dashboard`.

Based on this, it's a tricky task to get the correct element on the page in this moment. So I decided to change the assertions: if the login was successful, it means we won't display the login button anymore. That's it.

I think the ideal solution would be to move the welcome view (create/join group) to another router (ex: `/welcome`).  That would prevent this bug and we would still have better login assertions.

### Solution

Improve how we are using the router `/welcome` and how we present that to the users after a group is created.

cc @pieer and @taoeffect

sandrina-p:
- Create card banner component (cc @mmbotelho )
- Fix dropdown styles  (cc @mmbotelho )
- Add reduced-motion feature (enabled by default on Cypress) and try to fix #696

### DIFF
--- a/frontend/views/containers/settings/Appearence.vue
+++ b/frontend/views/containers/settings/Appearence.vue
@@ -11,7 +11,7 @@
       label
         input(type='checkbox' :checked='$store.state.reducedMotion' @change='handleCheckbox')
         i18n Reduced motion
-        .help When enabled the amount of animations you see around are reduced.
+        i18n.help When enabled the amount of animations you see around are reduced.
 
     section.card
       h2.settings-subtitle Text settings
--- a/frontend/views/containers/settings/Appearence.vue
+++ b/frontend/views/containers/settings/Appearence.vue
@@ -6,6 +6,14 @@
       SelectorTheme
 
     section.card
+      h2.settings-subtitle Reduced Motion
+
+      label
+        input(type='checkbox' :checked='$store.state.reducedMotion' @change='handleCheckbox')
+        i18n Reduced motion
+        .help When enabled the amount of animations you see around are reduced.
+
+    section.card
       h2.settings-subtitle Text settings
 
       //- TODO in separate tickets
@@ -17,15 +25,24 @@
 </template>
 
 <script>
+import { mapMutations } from 'vuex'
 import SelectorTheme from '@components/Settings/Theme.vue'
 // import SelectorFontSize from '@components/Settings/FontSize.vue'
 
 export default {
   name: 'SettingsAppearence',
-
   components: {
     SelectorTheme
-  //   SelectorFontSize
+    //   SelectorFontSize
+  },
+  methods: {
+    ...mapMutations([
+      'setReducedMotion'
+    ]),
+    handleCheckbox (e) {
+      console.log(e.target.checked)
+      this.setReducedMotion(e.target.checked)
+    }
   }
 }
 </script>
--- a/frontend/views/containers/settings/Appearence.vue
+++ b/frontend/views/containers/settings/Appearence.vue
@@ -40,7 +40,6 @@ export default {
       'setReducedMotion'
     ]),
     handleCheckbox (e) {
-      console.log(e.target.checked)
       this.setReducedMotion(e.target.checked)
     }
   }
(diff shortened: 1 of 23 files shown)


taoeffect (commented):
This looks good! Only concern I have at the moment is whether "Banner" is a good name for the new component you created, since @mmbotelho has designed something called "Banners" and already is using them for a very different purpose here: #688.

To me the "Banner" here looks more like a card of some sort:

Screen Shot 2019-10-20 at 3 03 46 PM

Maybe a different name can be picked to save "Banner" for the things in #688?

mmbotelho:
We can call them _Callout Cards_ instead, as in [Shopify's Polaris](https://polaris.shopify.com/components/structure/callout-card#navigation).

@sandrina-p @taoeffect

mmbotelho (changes requested):
- The dropdowns have the wrong font - should be body text;
- The dropdown list-item should be 32px high (the implemented version has 40-something px);
- This spacing:
Captura de ecrã 2019-10-20, às 15 22 30


Other than that I couldn't find anything else wrong. I am not happy at all with how it looks on mobile, but that's work for me, don't worry about it 😄

mmbotelho (dismissed):
Great job! 🎉

frontend/views/containers/settings/Appearence.vue
```diff
@@ -5,6 +5,14 @@
 
       SelectorTheme
 
+    section.card
+      h2.settings-subtitle Reduced Motion
+
+      label
+        input(type='checkbox' :checked='$store.state.reducedMotion' @change='handleCheckbox')
+        i18n Reduced motion
+        .help When enabled the amount of animations you see around are reduced.
```
taoeffect: missing i18n here and on line 9

taoeffect (changes requested):
Minor i18n thing in appearance, once that's fixed it's GTD!

frontend/views/containers/settings/Appearence.vue
```diff
@@ -5,6 +5,14 @@
 
       SelectorTheme
 
+    section.card
+      h2.settings-subtitle Reduced Motion
```
taoeffect, line 9: "and on line 9" :)

taoeffect: Noticing, on lines 4, 17, and 23 - 24 too

test/cypress/support/commands.js
```diff
@@ -40,16 +40,15 @@ Cypress.Commands.add('giLogin', (userName, password = '123456789') => {
...
-  cy.getByDT('app').then(($app) => {
-    if ($app.find('[data-test="userProfile"]').length) {
-      cy.getByDT('settingsBtn').click()
-      cy.getByDT('link-logout').click()
-      cy.getByDT('closeModal').should('not.exist')
-    } else {
-      cy.getByDT('logout').click()
-    }
-  })
+Cypress.Commands.add('giLogout', ({ hasNoGroup = false } = {}) => {
```
taoeffect, line 43: this doesn't seem like valid javascript?

taoeffect:

PR #734, opened 2019-10-21
Task/709 iterate add members proposal

### LINKED ISSUE #709: Iterate "Add members" UX
### Problem

When a user clicks on the "Add more" link, the modal is pushed up (as expected), and the content moves. This makes this interaction a bit awkward.

### Solution

As discussed offline in the Hackathon, we would like to have a similar behaviour to what we had previously, where you can use 1 input to add all members.

_Note: This is a low priority task and can be done after the prototype is live_

sandrina-p:
Closes #709

Something wrong is happening on the invitation process... Let's hope the last PR from @taoeffect fixes that. If it does, all tests should work.

This screenshots happens when inviting 2 registered users to a group with 1 member:

image


mmbotelho (changes requested):
Captura de ecrã 2019-10-22, às 09 44 19

This was the only thing I found :) great job! 🎉

sandrina-p:
Fixed @mmbotelho, I also updated the component correctly on Design System

frontend/model/contracts/group.js
```diff
@@ -23,10 +23,11 @@ export const PAYMENT_TYPE_BITCOIN = 'bitcoin'
 export const PAYMENT_TYPE_PAYPAL = 'paypal'
 export const paymentType = unionOf(...[PAYMENT_TYPE_MANUAL, PAYMENT_TYPE_BITCOIN, PAYMENT_TYPE_PAYPAL].map(k => literalOf(k)))
 
-export function generateInvites (numInvites: number) {
+export function generateInvites (numInvites: number = 1, inviteeName) {
   return {
     inviteSecret: `${parseInt(Math.random() * 10000)}`, // TODO: this
-    numInvites
+    numInvites,
+    inviteeName
```
sandrina-p: Note: We need the inviteeName to show on links management component with real data for @SebinSong.

frontend/views/components/Modal/Modal.vue
```diff
@@ -48,7 +48,7 @@ export default {
       } else {
         // When the route change we compare to see if the modal changed
         // If so, we send the event to close the modal
-        if (from.query.modal) {
+        if (this.content && this.content.length) {
```
taoeffect: I'm not sure I understand this line, can you explain? What was the reason for this change? I remember adding the check `from.query.modal` to prevent a bug before.

Unrelated note: checking `this.content.length` is unnecessary since `this.content` is either a string or null, and both `''` and `null` are falsy

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -111,67 +99,51 @@ export default {
...
-        const expiresDateMs = Date.now() + this.groupSettings.proposals[PROPOSAL_INVITE_MEMBER].expires_ms
-        for (const index in this.ephemeral.invitesToSend) {
-          const member = this.ephemeral.invitesToSend[index]
-          const mailbox = member.attributes.mailbox
-          const memberName = member.attributes.name
+        let identityContract
+        try {
+          const contractID = await sbp('namespace/lookup', inviteeName)
+          identityContract = await sbp('state/latestContractState', contractID)
+        } catch (e) {
```
sandrina-p: TIL: When using async/await inside a loop, use _for of / for in_. Array.forEach doesn't support async await correctly out of the box. cc @pieer @SebinSong

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -180,43 +152,53 @@ export default {
...
+          } else {
+            invitationsSent.push(inviteeName)
+            console.warn(`The invitee ${inviteeName} isn't registered. TODO #614`)
+          }
+
+          this.ephemeral.formFeedbackMsg = {
+            text: L('Invites to {invitationsSent} sent successfully!', {
+              invitationsSent: invitationsSent.join(', ')
+            }),
+            class: 'has-text-success'
```
taoeffect, line 193: Can you reset the form upon success so that all the invite names are removed?

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -111,67 +99,51 @@ export default {
...
+        let identityContract
+        try {
+          const contractID = await sbp('namespace/lookup', inviteeName)
+          identityContract = await sbp('state/latestContractState', contractID)
+        } catch (e) {
+          // The invitee isn't registered... but it's okay at #614
+        }
 
-          if (this.groupShouldPropose) {
+        if (this.groupShouldPropose) {
```
taoeffect, line 138: Pressing "enter" key in field does nothing when making a proposal. It neither adds a new input field, nor clicks "Next".

It needs to do one or the other (and consistently with when there is no proposal)

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -111,67 +99,51 @@ export default {
...
+          try {
```
taoeffect, line 139: Boy scouting warning on proposal:

```
i18n: using 'html' as string is deprecated! Instead: set 'compile' attribute and place HTML opening/closing tags in :args variables, then place text as child of i18n! 
Object { text: "Members of your group will now be asked to vote.</br>You need <strong>{value} yes votes</strong> for  your proposal to be accepted." }
```

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -111,67 +99,51 @@ export default {
...
             const proposal = await sbp('gi.contracts/group/proposal/create',
               {
                 proposalType: PROPOSAL_INVITE_MEMBER,
                 proposalData: {
-                  member: memberName,
+                  member: inviteeName,
```
taoeffect, line 144: (Putting this comment here because not sure where else to put it).

When the member who was invited through a proposal visits the link to join, there are errors and warnings in the console that need to be fixed:

```
[sbp] state/vuex/state 
Array []
main.js:44:14
latestContractState: ignoring mutation 21XWnNSWL9Y142KLMCsAJbPoTjo1wBrQXRN7Mi5dUoB1xs1QC6|gi.contracts/group/proposal/process because of TypeError state.js:69:18
[sbp] gi.contracts/group/proposalVote/process 
Array [ {…}, {…} ]
main.js:44:14
proposalVote: no proposal for 21XWnNSWL9Y142KLMCsAJbPoTjo1wBrQXRN7Mi5dUoB1xs1QC6! 
Object { data: {…}, meta: {…}, hash: "21XWnNLyrqJoU8viyDrdzMRgmnZiJm1V37So5EsHooLCmoBGYQ" }
group.js:183:18
latestContractState: ignoring mutation 21XWnNLyrqJoU8viyDrdzMRgmnZiJm1V37So5EsHooLCmoBGYQ|gi.contracts/group/proposalVote/process because of GIErrorIgnoreAndBanIfGroup state.js:69:18
[sbp] gi.contracts/group/proposalVote/process 
Array [ {…}, {…} ]
main.js:44:14
proposalVote: no proposal for 21XWnNSWL9Y142KLMCsAJbPoTjo1wBrQXRN7Mi5dUoB1xs1QC6! 
Object { data: {…}, meta: {…}, hash: "21XWnNLHmwSWijBacCSGy1Tff5KXRDP2AKyn54i1ESzsnYXwgh" }
group.js:183:18
latestContractState: ignoring mutation 21XWnNLHmwSWijBacCSGy1Tff5KXRDP2AKyn54i1ESzsnYXwgh|gi.contracts/group/proposalVote/process because of GIErrorIgnoreAndBanIfGroup state.js:69:18
```

This happens on the "You've been invited to join a group!" screen

taoeffect (changes requested):
While you're working on the review feedback, I'll work on converting `'namespace/lookup'` to return `null` when there's no user so that you don't need to use an exception handler to check if a user doesn't exist

mmbotelho (dismissed):
Perfect @sandrina-p ! Thank you! 😎

frontend/views/containers/proposals/AddMembers.vue
sandrina-p: @taoeffect what action do you prefer to happen? add field or next?

frontend/views/components/Modal/Modal.vue
sandrina-p: What bug did you prevent to happen?
Here, we are forcing the modal to close if there isn't a modal content to show to prevent this bug where you click to close a modal at the same time you are asking to redirect a page (logout): So, sometimes the logout and redirect would happen before the modal closes, and in that moment `from.query.modal` doesn't exist anymore. That's why the modal didn't close.

image

taoeffect:
@sandrina-p #738 ready for review for this PR!

frontend/views/containers/proposals/AddMembers.vue
sandrina-p: Done and added some examples to `translations.js file`.

I believe you can review the PR again

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -25,7 +25,10 @@
 
       .c-confirmation(v-if='isConfirmation' key='confirmation')
         svg-proposal.c-svg
-        i18n(html='Members of your group will now be asked to vote.</br>You need <strong>{value} yes votes</strong> for  your proposal to be accepted.' :args='{value: rule.value}')
+        i18n(
+          compile
+          :args='{ ...LTags("strong", "br"), value: rule.value }'
```
taoeffect: "br" is added by default :)

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -25,7 +25,10 @@
...
+        ) Members of your group will now be asked to vote. {br_} You need {strong_}{value} yes votes{_strong} for your proposal to be accepted.
```
taoeffect: No space before {br_}

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -111,71 +99,49 @@ export default {
...
       this.ephemeral.invitesCount += 1
+      Vue.nextTick(() => {
+        const inviteeSlots = this.$refs.fieldset.getElementsByTagName('label')
+        const newInviteeSlot = inviteeSlots[inviteeSlots.length - 1]
+        newInviteeSlot && newInviteeSlot.focus()
+      })
+    },
+    resetInvitations () {
+      this.ephemeral.invitesCount = 1
+      this.form.invitees[0] = ''
```
taoeffect, line 124: Did you try just calling `this.$v.$reset()` instead of these two lines?

frontend/views/utils/translations.js
```diff
@@ -17,6 +17,24 @@ i18next.use(XHR).init({
...
+String with variables:
+  i18n(
+    :args='{ name: ourUsername }'
+  ) Hello {name}!
+
+String with markup inside:
+  i18n(
+    compile
+    :args='{ ...LTags("strong"), name: ourUsername }'
+  ) Hello {strong_}{name}{_strong}!
```
taoeffect: Maybe a different example for this one, because if {name} is a variable too, you should instead just do:

```
  i18n(
    :args='{ name: `<strong>${ourUsername}</strong>`` }'
  ) Hello {name}!
```

Note also `compile` should only be added for strings that use Vue.js components like `<router-link>`, otherwise it shouldn't be added. Regular HTML works just fine without it.

In this case the example can be:

```
  i18n(
    :args='LTags("strong")'
  ) Hello and {strong_}welcome{_strong}!
```

frontend/views/containers/proposals/AddMembers.vue
sandrina-p: invitesCount controls the number of inputs to show (1). We are not using vuelidate here because we don't need validations in this form. Besides that, it's a dynamic form, so I decided to keep it simple.

frontend/views/utils/translations.js
sandrina-p: done, added more examples!

frontend/views/containers/proposals/AddMembers.vue
taoeffect: add a field

taoeffect: if there is already an empty field, it should do nothing, so that it doesn't keep adding field ad infinitum

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -25,7 +25,10 @@
 
       .c-confirmation(v-if='isConfirmation' key='confirmation')
         svg-proposal.c-svg
-        i18n(html='Members of your group will now be asked to vote.</br>You need <strong>{value} yes votes</strong> for  your proposal to be accepted.' :args='{value: rule.value}')
+        i18n(
+          compile
+          :args='{ ...LTags("strong"), value: rule.value }'
+        ) Members of your group will now be asked to vote.{br_} You need {strong_}{value} yes votes{_strong} for your proposal to be accepted.
```
taoeffect: - `compile` isn't needed here because this doesn't contain any Vue.js components
- Can you rename `value` to to `numVotes` to give translators more context?

frontend/views/pages/DesignSystem.vue
```diff
@@ -691,7 +691,7 @@ page(
           td
             label.field
               .label New Amount
-              .input-combo(:class='{ error: ephemeral.forms.hasError }')
+              .inpusst-combo(:class='{ error: ephemeral.forms.hasError }')
```
taoeffect: `inpusst`...?

frontend/views/utils/translations.js
```diff
@@ -1,7 +1,7 @@
 import i18next from 'i18next'
 import XHR from 'i18next-xhr-backend'
 import Vue from 'vue'
-import template from '~/frontend/utils/stringTemplate.js'
+import template from '@utils/stringTemplate.js'
```
taoeffect: Plz revert, this will cause an error when my changes get merged

taoeffect (approved):
FANTASTIC WORK!!

PR #738, opened 2019-10-22
'namespace/lookup' now returns null on 404 instead of throwing

taoeffect:


frontend/views/components/CreateGroupSteps/GroupInvitees.vue
```diff
@@ -45,6 +45,7 @@ div(
 </template>
 
 <script>
+// TODO: delete both this file and Invite.vue
```
sandrina-p, line 48: This file isn't being used at all, can be deleted. Invites happen at `AddMembers.vue`

sandrina-p (approved):

taoeffect:

frontend/views/components/CreateGroupSteps/GroupInvitees.vue
taoeffect: Makes sense to delete it in your PR :) 👍

Issue #686, opened 2019-09-27
Fix all uncaught exceptions and console errors


taoeffect:
### Problem

- There is currently a mysterious `uncaught exception: Object` error in the console when the Settings are first opened.
- After #680 is merged, there will be numerous `Object { _name: "NavigationDuplicated", name: "NavigationDuplicated", message: "Navigating to current location (\"/\") is not allowed", stack: "", … }` errors in the console (discussion: https://github.com/vuejs/vue-router/issues/2881)

### Solution

Find out the reason for these exceptions, fix them, and also try to close #684 at the same time as this issue.

SebinSong:
@taoeffect Could you elaborate on 'when the Settings are first opened'? :) I would love to try reproducing the same situation you got this error from.

In my case, the same error occurred when I tried to close any modal and it looks like the reason is that the modal tries to navigate to the same route while trying to update the query string ([here](https://github.com/okTurtles/group-income-simple/blob/master/frontend/views/components/Modal/Modal.vue#L77))
 (I remember you and @pieer also discussing it on the slack channel :) ).

One possible solution suggested by a vue core team member in [here](https://github.com/vuejs/vue-router/issues/2872#issuecomment-519073998) is simply not logging (or ignoring) the thrown error.

like this: 
this.$router.push({ query: null }).catch(err => {}) // eslint-disable-line handle-callback-err

what would you think about this?

taoeffect:
> @taoeffect Could you elaborate on 'when the Settings are first opened'? :) I would love to try reproducing the same situation you got this error from.

Darn, I tried just now to replicate this again in Firefox Developer Edition and wasn't able to. :-\

Hmm. So maybe it happens randomly. The good news is that properly closing this issue should catch it next time, especially if you make sure to `console.error(err.stack)` to show where the error came from.

> In my case, the same error occurred when I tried to close any modal and it looks like the reason is that the modal tries to navigate to the same route while trying to update the query string (here)
(I remember you and @pieer also discussing it on the slack channel :) ).

Yeah, that's the other error mentioned in the second bullet point.

> One possible solution suggested by a vue core team member in here is simply not logging (or ignoring) the thrown error.

I think that's only OK if you make sure to log all errors that are *not* `NavigationDuplicated` errors.

I.e. defining a special logger function like so:

```js
function logExceptNavigationDuplicated (err) {
  err.name !== 'NavigationDuplicated' && console.error(err)
}
// then elsewhere
[...]

PR #739, opened 2019-10-22
Fix NavigationDuplicated Error upon logout

SebinSong:
related to #686 

- problem
Upon logout, `NavigationDuplicated` error occurs because it tries to navigate to the same route.
- solution
Make it navigate to the route only when the current route is not the same as the destination.


taoeffect (approved):
Great job!

PR #740, opened 2019-10-22
fix Join.vue 'gi.contracts/group/proposal' error + more payments plumming

taoeffect:


frontend/model/contracts/group.js
```diff
@@ -132,31 +151,25 @@ DefineContract({
...
-        const groupID = sbp('state/vuex/state').currentGroupId
-        const groupState = sbp('state/vuex/state')[groupID]
-        for (const hash in groupState.proposals) {
-          const prop = groupState.proposals[hash]
+        for (const hash in state.proposals) {
+          const prop = state.proposals[hash]
           if (prop.status === STATUS_OPEN &&
             prop.data.proposalType === data.proposalType &&
-            prop.meta.username === ourUsername &&
+            prop.meta.username === meta.username &&
```
sandrina-p, line 166: Why this validation is done during process instead of validate as before?

frontend/model/state.js
```diff
@@ -220,9 +246,9 @@ const getters = {
       .filter(contractID => contracts[contractID].type === 'group' && state[contractID].settings)
       .map(contractID => ({ groupName: state[contractID].settings.groupName, contractID }))
   },
-  memberProfile (state, getters) {
-    return (username, groupId) => {
-      var profile = state[groupId || state.currentGroupId].profiles[username]
+  memberProfile (state) {
+    return (username) => {
+      var profile = state[state.currentGroupId].profiles[username]
```
sandrina-p, line 251: why can't we pass groupId here?

frontend/model/contracts/group.js
taoeffect: Because we cannot access state that exists outside of the contract (submodule) state, and `sbp('state/vuex/getters').ourUsername` does that

sandrina-p (approved):

frontend/model/state.js
taoeffect: because right now getters cannot be access from the contract safely, and i'm working on making it possible to do that... i'll explain in more detail offline

taoeffect:

Issue #664, opened 2019-09-13
Implement invite management section (group settings page)


mmbotelho:
### Problem
At the moment, there is no way to manage the invite links that are generated in a group.

### Solution
To implement the solution discussed on issue #612. You can see the Figma mockups here:
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=3824%3A0

sandrina-p:
Fixed by #723

sandrina-p:
Dummy layout was made at #723 but the fetching the real data is still missing. @SebinSong will be working on it

sandrina-p:
Functionality missing to close this issue:
- "See original proposal" action
- "Revoke link" action

-- EDIT 22 May --

"See original proposal" action will be left out to another issue #925. That way this issue (the core invite management functionality) can be closed.

mmbotelho:
ooops closed this by mistake

PR #742, opened 2019-10-22
Use the real invite data in inviteLinks.vue

SebinSong:
working on #664 

Take out the dummy invite List from `InviteLinks.vue` and hook it up with the real invite data from the store.


frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -67,7 +64,7 @@ page-section.c-section(:title='L("Invite links")')
                 )
                   i18n See original proposal
```
sandrina-p, line 69: Leave a TODO here so we don't forget to make this menu actually work. It's something you can also do in a next PR :D

@mmbotelho what is the exact view I should see when clicking at "See original proposal"?

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -172,9 +126,18 @@ export default {
...
+    invitesToShow () {
+      const { invites } = this.currentGroupState
+      const invitesList = Object.entries(invites).map(([inviteSecret, invite]) => ({
+        inviteSecret,
+        inviteLink: buildInvitationUrl(this.currentGroupId, inviteSecret),
+        ...invite
+      }))
+
+      return this.ephemeral.selectbox.selectedOption === 'Active' ? invitesList.filter(invite => invite.status === 'valid')
+        : this.ephemeral.selectbox.selectedOption === 'All' && invitesList
```
sandrina-p: To avoid this complex ternary, you can use a different approach with objects. Besides being faster, I find it easier to understand as well.

```js
// each key is a function, so the code inside is only executed if called.
const options = {
 Active: () => invitesList.filter(invite => invite.status === 'valid'),
 All: () => invitesList
}

return option[this.ephemeral.selectbox.selectedOption]()
```

sandrina-p (changes requested):

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -48,11 +48,8 @@ page-section.c-section(:title='L("Invite links")')
           )
             i.icon-ellipsis-v
         td.c-state
-          i18n.c-state-description {{ item.state.description }}
-          i18n.c-state-expire(
-            v-if='item.state.expireInfo'
-            :class='item.state.isExpired ? "expired" : ""'
-          ) {{ item.state.expireInfo }}
+          i18n.c-state-description {{ item.status }}
```
sandrina-p: For now this PR seems good, but later we'll need to also add the "Anyone" link. The logic for it is being implemented at #747. Once it's merged you can create another PR and address that.

SebinSong: Ok. if #747 is merged earlier than this PR, I will implement it and let you know for a review!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: I agree it looks much neater that way. fixed! 👍

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: I will work on it in next PR definitely.

frontend/views/components/GroupSettings/InviteLinks.vue
sandrina-p: Another thing I noticed: Here you are using the status directly as a copy. We can't do that because of translations.  So you need to map each status to its correct translation. You can use the object trick again, but this time, each key is a string instead of an object. You can see an example at `ProposalTemplate.vue`, method `submitTextNonProposal()`

SebinSong: Thanks! :) I just checked your code and am liking how you use objects for switching among multiple options. I will push a commit maybe sometime later today. 👍

SebinSong: @sandrina-p Just pushed a commit for it!

sandrina-p (dismissed):

frontend/views/components/GroupSettings/InviteLinks.vue
mmbotelho: @sandrina-p it should take the user to the "Proposals" modal and show the proposal where the link was originally generated

mmbotelho:
Do you need me to review this one @SebinSong ? I believe this is more of a technical PR that doesn't have any changes to the UI, right?

SebinSong:
@mmbotelho There is no new UI part implemented on this PR, Yes. :)

frontend/model/contracts/group.js
```diff
@@ -229,14 +230,16 @@ DefineContract({
...
+        invitee: string
       }),
       process (state, { data, meta }) {
         Vue.set(state.invites, data.inviteSecret, {
           generated: data.numInvites,
           creator: meta.username,
           responses: {},
-          status: 'valid'
+          status: 'valid',
+          invitee: data.invitee
```
taoeffect: We already have `creator` generated from `meta.username`, so I think you can delete `invitee` and use `creator` instead

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -24,13 +24,13 @@ page-section.c-section(:title='L("Invite links")')
...
-        :key='index'
+        v-for='(item, index) in invitesToShow'
+        :key='item.inviteSecret'
       )
         td.c-name
-          | {{ item.name }}
+          | {{ item.invitee }}
           tooltip.c-name-tooltip(
-            v-if='item.isAnyoneLink'
+            v-if='false'
```
taoeffect: Why is there a `v-if` if it's always false?

frontend/model/contracts/group.js
```diff
@@ -229,14 +230,16 @@ DefineContract({
...
+        numInvites: number,
+        invitee: string
       }),
       process (state, { data, meta }) {
         Vue.set(state.invites, data.inviteSecret, {
           generated: data.numInvites,
           creator: meta.username,
           responses: {},
-          status: 'valid'
+          status: 'valid',
```
taoeffect: These values should come from constants defined somewhere. So add a TODO here, like `// TODO: define all possible values as constants`

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +110,46 @@ export default {
...
           title: this.L('Your invite'),
           url: inviteLink
         })
       }
+    },
+    translateStatus (status) {
+      return ({
+        'used': L('used'),
+        'valid': L('valid')
+      })[status]
```
taoeffect: `({` => `{`
`})` => `}`

(Here the `(` and `)` surrounding the object braces are unnecessary)

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +110,46 @@ export default {
...
+        inviteSecret,
+        inviteLink: buildInvitationUrl(this.currentGroupId, inviteSecret),
+        status: this.translateStatus(invite.status)
+      }))
+      const options = {
+        Active: () => invitesList.filter(invite => invite.status === 'valid'),
+        All: () => invitesList
+      }
+
+      return options[this.ephemeral.selectbox.selectedOption]()
```
taoeffect, line 205: This is really well done 👍

taoeffect (changes requested):
Very nicely done! Noticed only minor issues + have some questions

frontend/model/contracts/group.js
sandrina-p: @taoeffect invitee is the person invited, not the creator.

frontend/views/components/GroupSettings/InviteLinks.vue
sandrina-p: For now it's false because it's missing the implementation of "Anyone link", that will be done in a next PR after #747 is merged

frontend/model/contracts/group.js
SebinSong: Fixed!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: fixed!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: thanks.

frontend/model/contracts/group.js
taoeffect: Ah, true. Hmm, I think let's wait for @sandrina-p's PR to be merged (#747), since it's almost done and also changes a lot of this stuff.

frontend/views/components/GroupSettings/InviteLinks.vue
taoeffect: Gotcha! OK, let's wait for #747 to be merged then, especially because of the other changes to group.js

frontend/model/contracts/group.js
SebinSong: Ok. once her PR is merged, I will come back and work on this part.

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +114,94 @@ export default {
...
+      remainder = remainder % MIL_HR
+      const minutes = Math.ceil(remainder / MIL_MIN)
+      const options = [
+        undefined,
+        L('{minutes}m left', { minutes }),
+        L('{hours}h {minutes}m left', { hours, minutes }),
+        L('{days}d {hours}h {minutes}m left', { days, hours, minutes })
+      ]
+
+      return options[!!days + !!hours + !!minutes]
```
taoeffect: @SebinSong this line is pure genius, did you come up with this yourself?

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +114,94 @@ export default {
...
+    },
+    mapInvite ({
+      creator,
+      inviteSecret,
+      responses,
+      quantity,
+      expires: expiryTime
+    }) {
+      const isAnyoneLink = creator === 'INVITE_INITIAL_CREATOR'
```
taoeffect: Use the `INVITE_INITIAL_CREATOR` constant by importing it from `group.js` so that:

1. We know where this concept is coming from and what code it relates to
2. If ever for some reason the value changes our linters will catch it

taoeffect (changes requested):
Sebin this is great! I only noticed very minor problems:

#### Problem 1 - `INVITE_INITIAL_CREATOR` needs to be a const

See [inline comment in review](https://github.com/okTurtles/group-income-simple/pull/742/files#r344991683).

#### Problem 2 - Created for column blank when invite not accepted yet

When a proposal has been accepted to invite a person, and the link generated, but not yet used, the "Created for" column is blank:

Screen Shot 2019-11-11 at 6 48 32 PM

#### Optional - improve drop-down menu look

This isn't something that needs to be done for this PR (it can be done later): the drop-down doesn't have the same look and field as other menu drop-downs:

Screen Shot 2019-11-11 at 6 45 59 PM

And there is a somewhat weird visual glitch on Firefox when the menu is opened and closed:

Screen Shot 2019-11-11 at 6 46 05 PM

However it is up to @mmbotelho to design what this drop-down needs to look like, and decide whether we should use the same style drop-down as the one in the other menu:

Screen Shot 2019-11-11 at 6 46 10 PM

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +114,94 @@ export default {
...
+    inviteStatusDescription ({
+      isAnyoneLink,
+      isInviteExpired,
+      isAllInviteUsed,
+      quantity,
+      numberOfResponses
+    }) {
+      if (isAnyoneLink) return L('{numberOfResponses}/{quantity} used', { numberOfResponses, quantity })
+      else if (isAllInviteUsed) return L('Used')
+      else return isInviteExpired ? L('Not used') : L('Not used yet')
```
taoeffect, line 143: No clever way of avoiding if/else blocks for this one? :)

(You don't need to change it if you don't feel like it, just noticing how this function stands out from the other two)

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Those points make sense. Fixed!

SebinSong:
As to **Problem 2**,
Could anyone help me with how to reproduce **When a proposal has been accepted to invite a person, and the link generated, but not yet used**? 
(FYI, What I did is I just run `group-proposals.specs.js` cypress test and worked based on the invites created during the test)

taoeffect:
> As to Problem 2,
Could anyone help me with how to reproduce When a proposal has been accepted to invite a person, and the link generated, but not yet used?

The way I did it was:

1. Create 4 container tabs in Firefox Dev Edition for users A, B, C and D
2. Create group with User A and invite users B and C
3. Join B and C into group
4. User creates invite/proposal to add user D to group
5. Vote to pass the proposal, now user A will have the proposal-generated invite link

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: It was inspired by @sandrina-p 's comment [here](https://github.com/okTurtles/group-income-simple/pull/742#discussion_r340702726)

SebinSong:
@taoeffect Thanks!! 👍 
@sandrina-p I noticed something while trying to fix the problem 2. 
inviteCreator

This is an invite link for user4, which is owned by(or was created by) user1. But the creator here is equal to 'user4' as you can see. isn't it meant to be 'user1' instead in this case? and also the 'invitee' property(which is missing here) has to be 'user4'?

I'm sorry for a verbose question. 
In short, shouldn't the object look like this in this case?

`{
  creator: 'user1',
  ..... ,
  invitee: 'user4'
}`

(I'm still not so sure if I'm giving enough info about the context here. my apology if so)

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +114,94 @@ export default {
...
+      expires: expiryTime
+    }) {
+      const isAnyoneLink = creator === 'INVITE_INITIAL_CREATOR'
+      const isInviteExpired = expiryTime <= 0
+      const numberOfResponses = Object.keys(responses).length
+      const isAllInviteUsed = numberOfResponses === quantity
+
+      return {
+        isAnyoneLink,
+        invitee: isAnyoneLink ? 'Anyone' : Object.keys(responses)[0],
```
SebinSong: @sandrina-p  `Object.keys(responses)[0]` here is because I couldn't find `invitee` property from the raw invite data fetched from the store.

taoeffect:
@SebinSong to answer your question, have a look in group.js. There you'll find the function `createInvite()`. Do a project-wide search for calls to that function. Then you should find this line:

```js
          payload.passPayload = createInvite({ creator: this.proposal.data.proposalData.member })
        }
```

Feel free to fix this bug if you'd like, or leave a TODO in a comment / open an issue. If anything is unclear, let me know!

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -48,11 +48,11 @@ page-section.c-section(:title='L("Invite links")')
...
             i.icon-ellipsis-v
         td.c-state
-          i18n.c-state-description {{ item.state.description }}
+          i18n.c-state-description {{ item.status.description }}
           i18n.c-state-expire(
-            v-if='item.state.expireInfo'
-            :class='item.state.isExpired ? "expired" : ""'
-          ) {{ item.state.expireInfo }}
+            v-if='item.status.expiryInfo'
+            :class='{ "expired": item.status.isExpired }'
```
sandrina-p: CSS classes that reflect a state should start with is `isExpired`

frontend/views/components/GroupSettings/InviteLinks.vue
sandrina-p: @taoeffect was talking about the fact you sum boolean to access an array. That's not only genius, it's also madness x)

frontend/views/components/GroupSettings/InviteLinks.vue
sandrina-p: Oh, it's a bug! Checked and it's a quick fix. `ProposalVoteOptions.vue`, when calling `createInvite()` at line ~105.

```diff
- payload.passPayload = createInvite({ creator: this.proposal.data.proposalData.member })
+ payload.passPayload = createInvite({ invitee: this.proposal.data.proposalData.member, creator: this.proposal.meta.username })
```

sandrina-p: Btw, it's missing L around Anyone.

sandrina-p (approved):

sandrina-p (changes requested):
check my last comments

mmbotelho (changes requested):
I decided to review this PR anyway, an extra set of 👀 is always helpful. Here are my comments:

- On desktop, the 3 dots button is not visible for the "Anyone" link, because there is no action the user can take. On mobile, however, it needs to be visible so users can copy or share the link:
Captura de ecrã 2019-11-13, às 14 16 33
Captura de ecrã 2019-11-13, às 14 16 54
Captura de ecrã 2019-11-13, às 14 19 32


- There seems to be a problem with scroll when testing on Chrome (desktop) with a small viewport. I can't scroll on the Group Settings page;

- This tooltip should say "This invite link is only available during the onboarding period.":
Captura de ecrã 2019-11-13, às 14 18 12

- I can't test most things on mobile because I can't scroll 😂 let me know when this is fixed so I can take a better look at it;

- I also noticed the empty row on the "Created for" column (as @taoeffect already mentioned).

- I tagged you on Figma regarding how this should behave when a link is revoked, can you please take a look at it? Feel free to leave that for a future PR if you want;

- When I tried to use a generated link from the Invite links table, I got this error:
Captura de ecrã 2019-11-13, às 14 30 59


I'll be waiting for the scroll to be fixed on mobile so I can test that as well. Thank you! 😄

mmbotelho:
Another thing I noticed but wasn't sure it was related to your PR - the horizontal position of the main white card and the page titles in desktop changes when I switch between pages (Pay Group and Group Settings for example). I tested on master and it doesn't happen there.

SebinSong:
@mmbotelho I will check/work on them soon!
PS: Do you mean chrome dev-tool's mobile mode by 'on mobile' on your comments? I had the same issue while working on some other previous PRs but just sort of worked around it, a**ssuming other devs had disabled it for some reasonable reasons**. So I'm afraid it's not something I have to fix on this PR because It's not a problem generated here..! (But I will definitely look into it and see if there's anything I can do about it! :) )

Thank you.

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: @sandrina-p Thanks again for the inspiration! 👍 (I keep saying this but I'm learning a lot from the project)

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Thanks! It's fixed that way.

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: fixed!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Let me keep it like this for now pls. I will continue to try to come up with a better-looking code than this though! :)

SebinSong:
@taoeffect  As to the strange shape of the dropdown select menu you pointed out on **optional**, It was intentionally made to look like that as the result of the discussion @mmbotelho and @pieer had during the hackathon because they decided that the button looks better that way(when the dropdown stays open) than the one with all four corners being round.

from 
fourCornerRounded

 to -->

changedOne


and the button comes back to it's original four-corners-round shape once it loses focus. :)

SebinSong:
In the new commit for your change requests,

- @taoeffect 's CR 
**Problem 1:** fixed
**Problem 2:** fixed
**Optional:** pls refer to [my previous comment](https://github.com/okTurtles/group-income-simple/pull/742#issuecomment-554172283)

- @mmbotelho 's CR
**No action for anyone link on desktop:** fixed
**a typo in tooltip:** fixed
**empty row on 'created for' column:** fixed
**An expired link issue:** I tried the links generated both during cypress test and while doing greg's scenario [here](https://github.com/okTurtles/group-income-simple/pull/742#issuecomment-552748519) but haven't been able to run into an expired link.
**Implementing 'revoke link' functionality:** the internal logic for this hasn't been implemented yet as far as I know, so I will leave it for an upcoming PR.
**Scroll problem:** This is not an issue introduced by this PR(Like, I never disabled 'scroll' or 'touch' events here and personally [had been having the same problem before hackathon](https://github.com/okTurtles/group-income-simple/pull/742#issuecomment-554156330)). So pls discuss this one with other front-end devs. At least I wanted to check whether the scroll does work on actual mobile device or not, but couldn't really do it because of [this issue](https://github.com/okTurtles/group-income-simple/issues/780).

- @sandrina-p 's CR
all fixed.

Thanks.

mmbotelho (dismissed):
Perfect! Thank you @SebinSong 🎉

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -48,13 +48,13 @@ page-section.c-section(:title='L("Invite links")')
...
         td.c-state
-          i18n.c-state-description {{ item.state.description }}
+          i18n.c-state-description {{ item.status.description }}
           i18n.c-state-expire(
-            v-if='item.state.expireInfo'
-            :class='item.state.isExpired ? "expired" : ""'
-          ) {{ item.state.expireInfo }}
+            v-if='item.status.expiryInfo'
+            :class='{ "is-expired": item.status.isExpired }'
+          ) {{ item.status.expiryInfo }}
```
taoeffect, line 55: Remember, i18n is just like the L function, it only ever takes string literals. Here, having `{{ item.status.expiryInfo }}` inside is the same as doing `L(item.status.expiryInfo)`, and neither will generate translatable strings.

So if you're using a variable that already has a translated string, you should not insert it using `i18n`, but instead you are free to include it without it. Since i18n outputs a `span` tag by default, here you can have `span.c-state-expire` instead, unless you want it to be a div (or something else)

Please make sure to double-check all other i18n usage, like on line 51, etc.

frontend/views/components/GroupSettings/InviteLinks.vue
taoeffect: @SebinSong Alas, I think there is still some math issue here, because if there is 1 day, 0 hours, and 30 minutes left, it will display `0h 30m left` (which is incorrect) instead of either `1d 0h 30m left` or `1d 30m left`

taoeffect (changes requested):
Great job @SebinSong! I think it's almost ready to merge! I found two issues, one related to i18n usage, and a bug in `readableExpiryInfo` that I didn't notice before, please have a look!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Yeah, I missed those parts. Fixed! :)

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: You are right. Fixed! :)

SebinSong:
Will work on the cypress error soon! (It passes on my laptop, FYI)

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +115,95 @@ export default {
...
+      const hours = Math.floor(remainder / MIL_HR)
+      remainder = remainder % MIL_HR
+      const minutes = Math.ceil(remainder / MIL_MIN)
+      const args = {
+        days: days ? `${days}d ` : '',
+        hours: hours ? `${hours}h ` : '',
+        minutes: minutes ? `${minutes}m ` : ''
+      }
+
+      return L('{days}{hours}{minutes}left', args)
```
taoeffect: OK, so we are almost perfect here. :-)

Remember that in another language days, hours, and minutes likely begin with symbols other than `d`, `h`, and `m`, and that translators might want to translate that too.

So this is almost perfect, but we still need to deal with that.

I see two approaches:

### One: Have everything, including "Expired" here

```js
  if (days) return L('{days}d {hours}h {minutes}m left', { days, hours, minutes })
  if (hours) return L('{hours}h {minutes}m left', { hours, minutes })
  if (minutes) return L('{minutes}m left', { minutes })
  return L('Expired')
```

In terms of raw speed, this approach should be just as fast as what you have above with the `?:` operator, because that operator creates [a branch](https://en.wikipedia.org/wiki/Branch_(computer_science)), just like an `if` statement.

For info on why branching can slow down execution, you might want to watch some videos on ["CPU pipelining"](https://www.youtube.com/watch?v=doJpguZFTe0) and "branch prediction".

But what I'm saying is that it seems like in this case there's no way to avoid it — AND it's not really an issue in terms of performance. It only becomes an issue when:

- Giant if/else blocks are used where they're unnecessary, making code ugly and hard to read
- Giant if/else blocks are located in a performance "hot zone", i.e. in code that gets executed constantly, such as in the main logic of a video game's run-loop

My personal vote is to go with this approach shown above, since it includes the "Expired" string within this function, and because it generates strings that give more context to translators.

The other approach is a modified version of what you have above.

If you choose this route, then remember to remove `L('Expired')` below in the `mapInvite` function.

### Two: Modified version of above

```js
const args = {
  days: days ? L('{numDays}d', { numDays: days }) : '',
  hours: hours ? L('{numHours}h', { numHours: hours }) : '',
  minutes: minutes ? L('{numMinutes}m', { numMinutes: minutes }) : '',
}
// HTML converts multiple spaces into a single space (unless &nbsp; is used)
return L('{days} {hours} {minutes} left', args)
```

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -109,72 +115,95 @@ export default {
...
   },
   computed: {
-    activeList () {
-      return this.ephemeral.selectbox.selectedOption === 'Active' ? this.ephemeral.dummyInviteList.filter(item => !item.state.isExpired)
-        : this.ephemeral.selectbox.selectedOption === 'All' && this.ephemeral.dummyInviteList
+    ...mapGetters(['currentGroupState']),
+    ...mapState(['currentGroupId']),
+    invitesToShow () {
+      const { invites } = this.currentGroupState
+      console.log('raw invites: ', invites)
```
taoeffect: Debug logging should be removed once it's no longer necessary

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Thanks for the mini compsci session here! 👍

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -24,16 +24,16 @@ page-section.c-section(:title='L("Invite links")')
...
       )
         td.c-name
-          | {{ item.name }}
+          | {{ item.invitee }}
           tooltip.c-name-tooltip(
             v-if='item.isAnyoneLink'
             direction='top'
-            :isTextCenter='true'
-            :text='L("This invite link was only available during the onboarding period.")'
+            isTextCenter
```
taoeffect: I'm noticing now that this is resulting in the following error in the console:

Screen Shot 2019-11-22 at 7 59 06 PM

This happened after I created a proposal to invite someone, and voted to generate the invite link, and then visited the Group Settings page.

And a totally separate issue is that the invite link that's generated from the proposal is being shown to everyone. I think it's supposed to only be shown to the user who created it? At least that's what I recall the designs saying, and the text says:

> Here's a list of all invite links you own

taoeffect (changes requested):
See [this comment](https://github.com/okTurtles/group-income-simple/pull/742/files#r349857192) for an error in the console related to the Tooltip and how to reproduce it.

Also, I noticed that links generated from proposals are being shown to everyone instead of being shown just to the creator.

SebinSong:
will work on them soon/!

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: fixed!

SebinSong:
@taoeffect I think this PR (finally...) is ready to be merged as long as you don't have any more change requests. sorry, it took me a long while to come back to this PR, :) (my life here has been very busy for the past few weeks!)

taoeffect (approved):
Great job @SebinSong!!

sandrina-p (approved):

PR #744, opened 2019-10-22
basic PayGroup payment sending + state/groupContractSafeGetters + more

### LINKED ISSUE #608: Automatically resend messages in case previousHEAD and latestHash out of sync
### Problem

In a particularly busy group, there is a very tiny (but still non-zero) chance that in the time between message construction (via `/create`) and sending, someone else will have sent a message and have made it so that `previousHEAD != latestHash`.

### Solution

In such a situation, the simple thing is to simply attempt to re-create and re-send the message. However, how many times will this process be repeated before a more serious error is thrown? (And how will that be handled?)

Related:

- #356

taoeffect:
+ can send payments in PayGroup
+ 'state/groupContractSafeGetters' makes it safe to use getters in group contract
+ handled a bunch of random previousHEAD related sync errors
+ Closes #608 - will retry publishLogEntry on autoban
+ groupMembers is now groupProfiles, and globalProfile is separated out to its own getter
+ fix error with L function not localizing string by no longer importing i18next

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -30,6 +30,7 @@ var store // this is set and made the default export at the bottom of the file.
 // we have it declared here to make it accessible in mutations
 // 'state' is the Vuex state object, and it can only store JSON-like data
 var dropAllMessagesUntilRefresh = false
+var attemptToReprocessMessageID
 const contractIsSyncing: {[string]: boolean} = {}
 
 const initialState = {
@@ -38,7 +39,6 @@ const initialState = {
   pending: [], // contractIDs we've just published but haven't received back yet
   loggedIn: false, // false | { username: string, identityContractID: string }
   theme: 'blue',
-  paymentsByMonth: {},
   reducedMotion: false,
   fontSize: 1
 }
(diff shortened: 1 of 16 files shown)


frontend/model/state.js
```diff
@@ -467,7 +483,19 @@ const actions = {
         dropAllMessagesUntilRefresh = true
       }
       if (banUser) {
-        await handleEvent.autoBanSenderOfMessage(message, e)
+        if (contractIsSyncing[contractID]) {
+          console.warn(`handleEvent: skipping autoBanSenderOfMessage since we're syncing ${contractID}`)
+        } else {
+          // NOTE: this random delay is here for multiple reasons:
+          const randomDelay = _.randomIntFromRange(0, 1500)
```
sandrina-p, line 490: this random could return 1, and that's not a big delay... maybe (500, 1500) is safer random number

sandrina-p (approved):

frontend/model/state.js
taoeffect: it's OK if there's no delay for one user as long as the delays are different between the users. It will also try again if there's a conflict

Issue #661, opened 2019-09-09
Figure out (and fix) the removal of the mysterious 'v'


taoeffect:
### Problem

In PR #655, at [my suggestion and request](https://github.com/okTurtles/group-income-simple/pull/655/files#r321982339), a mysterious `v` variable was removed from several files.

There were several problems with this variable:

1. It was declared outside of `ephemeral`, which violated [our data model style/rules](https://github.com/okTurtles/group-income-simple/blob/master/docs/Style-Guide.md#group-income-data-model-rules)
2. It was an example of "magic" code that had no clear usage or purpose in the file it was declared
3. It created confusion over a similarly named variable `$v` that is injected into components by [vuelidate](https://vuelidate.netlify.com)

However, during [my review](https://github.com/okTurtles/group-income-simple/pull/659/files#r322378946) of PR #659 I noticed that in fact this variable (or a variable like it) did appear to be used in some other files.

### Solution

So, what's going on?! Your job is to solve the mystery and refactor a whole bunch of files to make this variable no longer "magic", or perhaps get rid of it completely without breaking things (by using `$v` only, if possible, or some other technique).

Whatever you do, always make sure our [data model rules](https://github.com/okTurtles/group-income-simple/blob/master/docs/Style-Guide.md#group-income-data-model-rules) are followed.

EDIT: it looks like @sandrina-p might [have figured it out](https://github.com/okTurtles/group-income-simple/pull/679#discussion_r331058148). So maybe for this issue let's see if there's a way to make the code simpler/clearer. At the very least, we shouldn't have a variable called `v`, being used in a very similar way to `$v`, as that can lead people (like me!) to think there might be a typo and a bug. So a clearer name for the variable would be a minimum for closing this issue, but even better would be if there were some way to do what `v` is doing but in an even simpler way, without passing down props and emitting back up events.

Issue #354, opened 2018-02-14
Implement snapshotting / checkpointing


taoeffect:
### Problem

A pain on resources for new users to have to sync whole chain.

### Solution

Create a new contract with the initial state based on the end-state of the previous one, and tell the server to delete all prior messages (once all members have migrated).

Related: #691 - Light client

### Edit: promising solution / protocol

In tandem with consensus (#813), this protocol describes how contract snapshots can be saved and voted upon: https://gitlab.okturtles.org/okturtles/group-income-simple/-/wikis/State-Snapshots-and-Consensus

PR #747, opened 2019-10-23
Group Invitation Page working to everyone

### LINKED ISSUE #609: Redesign and rewrite Invite.vue to generate invite links
### Problem

We no longer invite people by username. Instead:

- If less than 3 members exist, invite links are created (that can each specify how many members they allow to join)
- If at least 3 members exist, a proposal is created, containing the names of the people to be invited (not usernames, just their full names), and if it passes an invite link for those people is created

### Solution

- Completely rewrite/redesign all related views. Delete anything no longer needed.
- Create a page to manage and view all invite links that have been created (see `currentGroupState.invites` after the proposals PR is merged) (This is now #612)
- Create a single proposal when proposing to invite multiple members, instead of multiple proposals

### LINKED ISSUE #614: Implement Signup-And-Join flow when given a link and not logged in
### Problem

The types of links generated by #611 , e.g. (copying from `Invite.vue`):

```js
                message: `Hi ${memberName},

                Here's your special invite link:

                ${process.env.FRONTEND_URL}/app/join?groupId=${this.$store.state.currentGroupId}&secret=${invite.inviteSecret}`
```

Should be usable even when no user has been created.

### Solution

Create and implement the Signup and Join flow that @mmbotelho designed and should be instantiated when visiting a link like that while not logged in. [Here is the Figma page for this.](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=875%3A2)

### LINKED ISSUE #701: Create a welcome page for new members of a group
### Problem
At the moment, we still don't have a substitute for the welcome page - the first page new members see when they join a new group. The current version looks like this:

Captura de ecrã 2019-10-16, às 10 04 07

### Solution
Update this page.

### LINKED ISSUE #741: Update proposals widget UI
### Problem
The title + description + proposal order we had on the proposals dashboard widget did not make a lot of sense.

### Solution
Move the description of the proposals below the actual proposals. I also took this chance to improve the spacings and the mobile UI for this widget (both for the dashboard and the Proposals page). I also added a "Read more" link, so when users add a very long description, it does not "break" the interface.

**Figma:**
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=3043%3A524

sandrina-p:
Closes #701, closes #614, closes #609 and closes #741 (_Pardon moi_ for changing so many files. It was during our Hackathon, after all 🙈)

image

image

- Invite members now only works through an invitation link. Messages to the inbox were removed.
- Proposals should keep working as before, but now with real invitation link!

Missing:
- [x] Polish cypress tests
- [x] #741 - Read more link
- [x] DRY and typos around 

Everyone, please test this well around.
- @mmbotelho feel free to point out any design tweak related to this. Note: we are reusing the same component to login/signup at both scenarios: alone or by invitation. In the designs, I noticed some inconsistencies in the copy and spacing. Please review those and tell me if any tweak is needed. 
- @taoeffect take a look at the group contract, I had to adapt some things related to the invitations.


frontend/model/contracts/group.js
```diff
@@ -239,15 +266,15 @@ DefineContract({
...
-          generated: data.numInvites,
-          creator: meta.username,
-          responses: {},
-          status: 'valid'
-        })
+        Vue.set(state.invites, data.inviteSecret, initInvite({
+          inviteSecret: data.inviteSecret,
+          numInvites: data.numInvites,
+          creator: meta.username
+        }))
```
taoeffect: missing `responses`, and `status`

frontend/views/containers/modals/SignUp.vue
```diff
@@ -3,190 +3,42 @@
...
 import ModalTemplate from '@components/Modal/ModalTemplate.vue'
-import FormPassword from '@components/Forms/Password.vue'
-import L from '@view-utils/translations.js'
+import FormSignup from '@containers/forms/FormSignup.vue'
 
-// TODO: fix all this
 export default {
-  name: 'SignUpModal',
-  mixins: [validationMixin],
+  name: 'Signup',
```
taoeffect: Please keep it `SignUp` or rename the file to `Signup.vue`. I personally think it's fine as `SignUp`

frontend/model/contracts/group.js
sandrina-p: No need for that. When creating an invitation, it always has the same responses ans status values, so `initInvite` already does that for us ;)

frontend/views/containers/modals/SignUp.vue
sandrina-p: I changed to signup because we say login instead of logIn. So it was to keep consistency. I'll rename the file to `Signup.vue`

frontend/views/containers/modals/SignUp.vue
```diff
@@ -3,190 +3,42 @@
...
-          }
-          this.$refs.modal.close()
-        } catch (ex) {
-          console.error('SignUp.vue submit() error:', ex)
-          sbp('state/vuex/dispatch', 'logout')
-          this.ephemeral.errorMsg = ex.toString()
-        }
+    submit () {
+      if (this.$route.query.next) {
+        // TODO: get rid of this timeout and fix/update tests accordingly
```
sandrina-p: @taoeffect do you know why exactly we had to add this timeout here? I believe this was an issue from nightmare times... not cypress :D

mmbotelho (changes requested):
- When I'm on the "Why are you proposing these changes?" step, with the textarea focused, pressing tab goes to the cancel button but changes focus to the close button immediately (without me pressing tab again);
- Something weird happened when testing this... I created 3 different users. Created the first, created the second on an icognito window, logged out, created the third user. When trying to login with the second user, I was directed to the "Create / Join group" page instead of the Dashboard 😕 ;
- Did you implement the mobile "read more" yet?  If yes, we probably need to reduce the number of characters;
- Spacings are not according the latest version of the widget, both on desktop and mobile. Let me know if you want me to specify which ones!

Captura de ecrã 2019-11-01, às 13 34 12

sandrina-p:
@mmbotelho 

> When I'm on the "Why are you proposing these changes?" step, with the textarea focused, pressing tab goes to the cancel button but changes focus to the close button immediately (without me pressing tab again);

That's an issue with the Modal component itself, it's happening in all the modals... I'll take a look and try to solve it in another PR.

Something weird happened when testing this... I created 3 different users. Created the first, created the second on an icognito window, logged out, created the third user. When trying to login with the second user, I was directed to the "Create / Join group" page instead of the Dashboard 😕 

> What's wrong with that? When creating a user, you should see the create / join page. You only see the dashboard if it has a group. And based on what you described, the user hasn't created a group yet.

> Did you implement the mobile "read more" yet? If yes, we probably need to reduce the number of characters;

I did the same logic for both. Now, I distinguish it: Characters Limit: mobile 90, desktop 210. Note that if you resize the page, you need to refresh it after to see the limit working properly.

> Spacings are not according to the latest version of the widget, both on desktop and mobile. Let me know if you want me to specify which ones!

Upps... please check again!

frontend/views/containers/proposals/ProposalBox.vue
```diff
@@ -42,17 +53,16 @@ export default {
...
-        return L('{who} proposed', { who })
+        return L('{who}{_strong} proposed:', { who, ...this.LTags('strong') })
       }
 
-      return username === currentUsername
-        ? L('You are proposing')
-        : L('{who} is proposing', { who })
+      return username === this.ourUsername
+        ? L('You are proposing:')
+        : L('{strong_}{who}{_strong} is proposing:', { who, ...this.LTags('strong') })
```
taoeffect: See same [feedback to @pieer](https://github.com/okTurtles/group-income-simple/pull/724#discussion_r338022233), copied here for convenience:

The pattern: `{tag_}{variable}{_tag}` should be rewritten using just `{variable}` to make life less confusing for the translators. Tags are useful when the thing between them is a word to be translated, i.e. `{tag_}english word or fragment that is not a variable{_tag}`.

So here you can do:

```
        i18n(
          v-if='needsIncome'
          tag='p'
          data-test='headerNeed'
          :args='{ amount: `<span class="has-text-bold has-text-0">${upTo}</span>` }'
        ) You need {amount}
```

The same goes for the two `i18n`s below and any other ones in this PR that follow the pattern `{tag_}{variable}{_tag}`

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,26 +6,29 @@ modal-template(ref='modal')
...
     i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 60 free passes!
-    a.link.has-icon.c-link(:href='link' target='_blank')
-      i.icon-link
-      | {{link}}
-    i18n.has-text-1(tag='p') This invite link expires on the 4th of February.
+    .c-link
+      a.link.has-icon(:href='link' target='_blank' data-test='invitationLink')
+        i.icon-link
+        | {{link}}
+    i18n.has-text-1(tag='p') This invite link expires on the [4th of February].
```
taoeffect: `[4th of February]` should be a variable, i.e. `{ expiryDate }` (you can still use ``[4th of February]`` as the temporary value, but the variable needs to be there)

sandrina-p: `expiryDate` was not implemented yet (see group.js TODO)... that's why I left a dummy value.

mmbotelho (changes requested):
Two minor style problems - everything else is working ewxactly as expected!

Captura de ecrã 2019-11-03, às 12 07 00
Captura de ecrã 2019-11-03, às 12 08 27

mmbotelho:
> What's wrong with that? When creating a user, you should see the create / join page. You only see the dashboard if it has a group. And based on what you described, the user hasn't created a group yet.

The problem was that the user was already in the group - they were even displayed in the sidebar of other members. I was not able to reproduce this error yet. I will let you know if I do! @sandrina-p

frontend/views/containers/modals/InviteByLink.vue
taoeffect: Yes, but you can still put the variable there for translators. The value can be fixed later

mmbotelho (dismissed):
Approved!!!!!!!!!! 🎉 

Thank you for making the changes I requested :) I noticed that on the description of this PR you say that the read more is not done yet, but it is, right?

Well it makes me very happy that this PR will close so many different issues 😃

frontend/views/containers/proposals/ProposalBox.vue
sandrina-p: So.. I didn't understand exactly when we should use `{tag_}` over `<strong>` in `args`. Can you clarify that?

frontend/views/containers/modals/InviteByLink.vue
sandrina-p: Done.

sandrina-p:
> I noticed that on the description of this PR you say that the read more is not done yet, but it is, right?

@mmbotelho MR description updated ;)

frontend/views/containers/proposals/ProposalBox.vue
taoeffect: Let's make sure we're talking about the same thing, because the L function doesn't have anything called `args`, while `i18n` does.

In i18n, variables are passed through `:args=`, and in the `L` function, they are passed as the second parameter.

String literals (the stuff that's given to translators), must never contain HTML.

Here we have a variable that translators can move around, `{who}`. It doesn't need the variables `{strong_}` and `{_string}` around it, because that's redundant. IF instead we had something like `{who} some text` that needed to be made bold as a whole unit, THEN we can introduce `{strong_}` and `{_strong}` around it, like so: `{strong_}{who} some text{_strong}`.

But in the case of the variable itself needing to be bolded, there is absolutely no reason to bother the translators by introducing more variables than there needs to be. That is the case here. We only need `{who}` in here. The fact that the value is surrounded by HTML `<strong>` doesn't need to be evident in the string-to-be-translated. So here it should be removed, and placed into the value-part. In this case:

```
{ who: `<strong>${who}</strong>` }
```

frontend/assets/style/components/_buttons.scss
```diff
@@ -212,7 +212,7 @@ button.link {
   color: $primary_0;
   font-weight: 600;
   display: inline-block;
-  line-height: inherit;
+  line-height: 1.4; // better underline visually
```
taoeffect, line 215: @pieer can you verify this change? Does it make sense to not use a variable here?

frontend/assets/style/components/_links.scss
```diff
@@ -10,6 +10,7 @@
   padding-left: 0;
   padding-right: 0;
   border-width: 0 0 1px 0;
+  line-height: 1.4; // better underline visually
```
taoeffect, line 13: @pieer same question here [as above](https://github.com/okTurtles/group-income-simple/pull/747/files#r343955892)

frontend/assets/style/components/_buttons.scss
pieer: Inherit is not a variable. All good :)

frontend/assets/style/components/_links.scss
pieer: same same

frontend/assets/style/components/_buttons.scss
taoeffect: OK, not sure I understand since `inherit` is being replaced with `1.4`, but assume it's OK to you

frontend/views/containers/modals/CreateGroup.vue
```diff
@@ -121,30 +122,34 @@ export default {
...
-              proposals[PROPOSAL_INVITE_MEMBER].defaults,
-              { ruleSettings: { [RULE_THRESHOLD]: { threshold: this.form.memberApprovalThreshold } } }
-            ),
-            [PROPOSAL_REMOVE_MEMBER]: merge({},
-              proposals[PROPOSAL_REMOVE_MEMBER].defaults,
-              { ruleSettings: { [RULE_THRESHOLD]: { threshold: this.form.memberRemovalThreshold } } }
-            ),
-            [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposals[PROPOSAL_PROPOSAL_SETTING_CHANGE].defaults,
-            [PROPOSAL_GENERIC]: proposals[PROPOSAL_GENERIC].defaults
+          initialInvite,
```
taoeffect: 1. To save a line, place `generateInvites(60)` here instead of on line 125, since `initialInvite` variable is only ever used once
2. Instead of creating `initialInvite`, which complicates the `group` datastructure for the sake of a single invite, use the existing `invites` key and add a single invite (see related comments in `group.js`)

frontend/model/contracts/group.js
```diff
@@ -14,14 +14,25 @@ import { merge, deepEqualJSONType } from '~/frontend/utils/giLodash.js'
...
   return {
     inviteSecret: `${parseInt(Math.random() * 10000)}`, // TODO: this
-    numInvites
+    numInvites,
+    invitee
     // expires: // TODO: this
   }
 }
 
+function initInvite ({ inviteSecret, numInvites, creator }) {
```
taoeffect: Having two functions that are doing about the same thing and duplicating work should be avoided if possible. We can simplify this by merging these two functions in a single function, and give it a new name: `createInvite`:

```js
// TODO: add flowtype annotations to parameters
function createInvite ({ numInvites, creator, invitee, expires }) {
  return {
    inviteSecret: `${parseInt(Math.random() * 10000)}`, // TODO: this
    numInvites,
    creator,
    invitee,
    expires,
    responses: {}, // { bob: true } list of usernames that answered
    status: 'valid' // TODO: this must come from a constant defined somewhere
  }
}
```

Then we can use that instead of `generateInvites`, and remove all usage of `initInvite` since it's no longer necessary

frontend/model/contracts/group.js
```diff
@@ -41,27 +52,45 @@ DefineContract({
...
-        [PROPOSAL_INVITE_MEMBER]: proposalSettingsType,
-        [PROPOSAL_REMOVE_MEMBER]: proposalSettingsType,
-        [PROPOSAL_GROUP_SETTING_CHANGE]: proposalSettingsType,
-        [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposalSettingsType,
-        [PROPOSAL_GENERIC]: proposalSettingsType
+      initialInvite: objectOf({
+        inviteSecret: string,
+        numInvites: number,
+        invitee: optional(string)
+      }),
```
taoeffect: - This can then be replaced with a validation for `invites` (a `mapOf(string, inviteType)`, where `const inviteType = ...` is defined above `DefineContract` as an `objectOf()` that validates the object defined by `createInvite()`)
- Then on line 54, replace `objectOf` with `objectMaybeOf`, so that we can in the future init the group contract with more values, to made upgrades easier and to partially address #354 - snapshots

frontend/model/contracts/group.js
```diff
@@ -41,27 +52,45 @@ DefineContract({
...
+          groupCreator: meta.username,
+          groupPicture: settings.groupPicture || '/assets/images/default-avatar.png' // TODO - better avatar
+        },
+        invites: {
+          [initialInvite.inviteSecret]: initInvite({
+            inviteSecret: initialInvite.inviteSecret,
+            numInvites: initialInvite.numInvites,
+            creator: 'GROUP_WELCOME' // To be used as initial invitation to first X members.
+          })
+        },
```
taoeffect, line 95: Now here we're able to save a whole bunch of lines by using `merge` directly on the data:

```js
      const initialState = merge({
        payments: {},
        invites: {},
        proposals: {}, // hashes => {} TODO: this, see related TODOs in GroupProposal
        settings: {
          groupCreator: meta.username
        },
        profiles: {
          [meta.username]: initGroupProfile(meta.identityContractID)
        },
        userPaymentsByMonth: {
          [currentMonthTimestamp()]: initMonthlyPayments()
        }
      }, data)
```

I also think it's important to set the default group avatar not here but someplace else, since we don't know what sorts of URL types we'll be using here.

frontend/model/contracts/group.js
```diff
@@ -229,15 +258,15 @@ DefineContract({
     'gi.contracts/group/invite': {
       validate: objectOf({
         inviteSecret: string, // NOTE: simulate the OP_KEY_* stuff for now
-        numInvites: number
+        numInvites: number,
+        invitee: string
```
taoeffect: Here instead of `objectOf` we can now use our `inviteType` to validate the data

frontend/model/contracts/group.js
```diff
@@ -229,15 +258,15 @@ DefineContract({
...
-          generated: data.numInvites,
-          creator: meta.username,
-          responses: {},
-          status: 'valid'
-        })
+        Vue.set(state.invites, data.inviteSecret, initInvite({
+          inviteSecret: data.inviteSecret,
+          numInvites: data.numInvites,
+          creator: meta.username
+        }))
```
taoeffect: And since `createInvite` does everything we need, we can save many more lines:

```js
Vue.set(state.invites, data.inviteSecret, data)
```

frontend/views/containers/forms/FormSignup.vue
```diff
@@ -1,75 +1,64 @@
...
+      data-test='signEmail'
+      v-error:email='{ tag: "p", attrs: { "data-test": "badEmail" } }'
+    )
 
-      p.error(v-if='ephemeral.errorMsg') {{ ephemeral.errorMsg }}
+  form-password(
+    :label='L("Password")'
+    :value='form'
+    :v='$v.form'
+    @input='(newPassword) => {password = newPassword}'
```
taoeffect, line 36: This `:v=` and `@input='(newPassword) => {password = newPassword}'` such a confusing way of doing things. I'm hoping an issue can be created, and/or a TODO written here, to find a clearer way of doing this

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -125,80 +120,44 @@ export default {
...
+          invitationsSent.push(inviteeName)
+          this.ephemeral.formSuccessMsg = L('Invites to {invitationsSent} sent successfully!', {
+            invitationsSent: invitationsSent.join(', ')
+          })
+        } catch (e) {
+          console.error(`Invite to ${inviteeName} failed to be sent!`, e)
+          invitationsFailed.push(inviteeName)
+          this.ephemeral.formErrorMsg = L('Invites to {invitationsFailed} failed!', {
+            invitationsFailed: invitationsFailed.join(', ')
+          })
```
taoeffect: A `try/catch` inside of a `for` should be avoided for various reasons, including performance reasons, but also because it can lead to poor error handling. For example, here if there was an error it will still continue looping, and it shouldn't

frontend/views/containers/proposals/ProposalBox.vue
```diff
@@ -65,9 +77,28 @@ export default {
...
       const reason = this.proposal.data.proposalData.reason
-      return reason ? `"${reason}"` : undefined
+      const maxlength = this.config.reasonMaxLength
+      if (this.ephemeral.isReasonHidden && this.shouldTruncateReason) {
+        // Prevent "..." to be added after an empty space. ex: "they would ..." -> "they would..."
+        const charToTruncate = reason.charAt(maxlength - 1) === ' ' ? maxlength - 1 : maxlength
+        return `"${reason.substr(0, charToTruncate)}..."`
+      }
+
+      this.ephemeral.isReasonHidden = false // eslint-disable-line vue/no-side-effects-in-computed-properties
```
taoeffect: This is a good vue linter rule. There should be no side-effects in computed properties. So this line should be removed

frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -8,11 +8,10 @@
           :class='{ "has-text-danger": proposal.status === statuses.STATUS_FAILED, "has-text-success": proposal.status === statuses.STATUS_PASSED }'
           data-test='statusDescription'
           ) {{statusDescription}}
-      .ctas
-        proposal-vote-options(
-          v-if='proposal.status === statuses.STATUS_OPEN'
-          :proposalHash='proposalHash'
-        )
+      proposal-vote-options(
+        v-if='proposal.status === statuses.STATUS_OPEN'
```
taoeffect, line 12: Similarly, all invites should use constants like this to define and check their status, instead of using string literals like `'valid'`

frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -62,6 +62,11 @@ export default {
     proposalType () {
       return this.proposal.data.proposalType
     },
+    isOurProposal () {
+      const { identityContractID } = this.proposal.meta
+      const username = this.$store.state[identityContractID].attributes.name
+      return username === this.ourUsername
```
taoeffect: It's unclear whether we will (in the very distant future) consider usernames immutable or identityContractIDs immutable.

But whatever the case may be, here it would be cleaner & faster to do either:

```js
return this.proposal.meta.username === this.ourUsername
```

Or:

```js
return this.proposal.meta.identityContractID === this.$store.state.loggedIn.identityContractID
```

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,26 +6,40 @@ modal-template(ref='modal')
...
+  computed: {
+    ...mapGetters([
+      'currentGroupState'
+    ]),
+    invitationWelcome () {
+      const invites = this.currentGroupState.invites
+      return Object.keys(invites).find(invite => invites[invite].creator === 'GROUP_WELCOME')
+    },
+    link () {
+      return buildInvitationUrl(this.$store.state.currentGroupId, this.currentGroupState.invites[this.invitationWelcome].inviteSecret)
```
taoeffect: The value of `this.invitationWelcome` is the inviteSecret, so you can rename `invitationWelcome` to `welcomeInviteSecret` and return:

```js
return buildInvitationUrl(this.$store.state.currentGroupId, this.welcomeInviteSecret)
```

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,153 @@
...
+      p.has-text-1 {{ ephemeral.errorMsg }}
+      i18n.c-goHome(tag='button' @click='goHome') Take me home
+
+  // TODO - Remove these. Only used for debug.
+  .buttons.c-debug
+    button.is-small.is-outlined(@click='setStatus("LOADING")') LOADING
+    button.is-small.is-outlined(@click='setStatus("SIGNING")') SIGNING
+    button.is-small.is-outlined(@click='setStatus("WELCOME")') WELCOME
+    button.is-small.is-outlined(@click='setStatus("EXPIRED")') EXPIRED
+    button.is-small.is-outlined(@click='setStatus("INVALID")') INVALID
```
taoeffect: Just a comment, as here's it's not a big deal, so no change is needed. It's generally good practice to avoid using string literals as enums, since it can introduce hard-to-catch bugs if there's a typo in the string literal. For example, `setStatus("INVALD")` might not generate an error, but `ephemeral.pageStatus = status.INVALD` is more likely to, and some smart IDEs can catch such errors as you're editing the file.

The reason it's OK to using string literals in SBP is because a typo in a selector will result in an error when the selector is called.

So an alternative implementation could look like:

```js
computed: {
  status: () => ({
    LOADING: 'LOADING',
    SIGNING: 'SIGNING',
    EXPIRED: 'EXPIRED',
    // etc...
  })
  // ...
```

And then:

```pug
button.is-small.is-outlined(@click='ephemeral.pageStatus = status.LOADING')
```

Alternatively, you can guarantee that all typos will be caught by using a [computed property for the pageStatus](https://vuejs.org/v2/api/#computed), and throwing an exception in the setter:

```js
computed: {
  pageStatus: {
    set (s) {
      if (!['LOADING', 'WELCOME', 'SIGNING', 'LOGGING', 'EXPIRED', 'INVALID'].includes(s)) {
        throw new Error(`bad status: ${s}`)
      }
      this.ephemeral.pageStatus = s
    }
  }
}
```

And this way you don't need the comment on line 83, and you don't need the `setStatus` or `isStatus` functions, since in the markup you can directly compare or set against `pageStatus`:

```pug
button.is-small.is-outlined(@click='pageStatus = "LOADING"')
```

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,153 @@
...
+      svg-broken-link.c-svg
+      i18n.title.is-1(
+        v-if='isStatus("EXPIRED")'
+        tag='h1'
+        :args='{ ...LTags() }'
+      ) Oh no! {br_} Something went wrong.
+      i18n.title.is-1(
+        v-else
+        tag='h1'
+        :args='{ ...LTags() }'
```
taoeffect: An object copy can be avoided by doing:

```
:args='LTags()`
```

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,153 @@
...
+          groupPicture: state.settings.groupPicture,
+          creator,
+          creatorPicture,
+          message
+        }
+        this.setStatus('SIGNING')
+      }
+    } catch (e) {
+      console.error(e)
+      this.ephemeral.errorMsg = `${L('This invite is not valid.')} ${e.toString()}`
```
taoeffect: - I think error messages will be typically attached to `e.message`, but this might be OK too
- Note that "This invite is not valid" is not necessarily the correct message to display. The error could be thrown from any of the calls to `await`. For example, if `namespace/lookup` failed. So probably better to do something like `this.ephemeral.errorMsg = e.message`

frontend/views/pages/Page.vue
```diff
@@ -1,10 +1,11 @@
 <template lang='pug'>
 div(:data-test='pageTestName' :class='$scopedSlots.sidebar ? "p-with-sidebar" : "p-no-sidebar"')
   header.p-header
-    h1.p-title(:data-test='pageTestHeaderName')
+    slot(name='header')
+    h1.p-title(:data-test='pageTestHeaderName' v-if='$slots.title')
       img.c-logo(
         src='/assets/images/group-income-icon-transparent.png'
-        :alt='L("Group Income\'s logo")'
+        alt='GroupIncome\'s logo'
```
taoeffect: Why was this modified? Alt labels need localization too for accessibility

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,153 @@
...
+          await this.accept()
+          return
+        }
       }
+      const state = await sbp('state/latestContractState', this.$route.query.groupId)
+      const invite = state.invites[this.$route.query.secret]
+      if (!invite) {
+        this.ephemeral.errorMsg = L('This invite is not valid.')
+        return this.setStatus('INVALID')
+      } else if (invite && invite.status === 'used') {
```
taoeffect: Here especially we want to be checking the status against a constant that's imported from `group.js` or some new file, like `INVITE_USED`. Besides catching typos, this will make it easier to do project-wide refactoring in case any status types need to be redefined.

taoeffect (changes requested):
Great job @sandrina-p! I reviewed perhaps not everything but on the order of ~80% of the logic. After you push all the changes for the review, let me know, and I'll test it out locally

frontend/assets/style/components/_buttons.scss
sandrina-p: in this case inherit receives the value from the body (1.5). That causes the link border-bottom to be too far from the text. So I manually decreased it a little to look better. Here's an image for you:

image

The first paragraph has `line-height: inherit;` (1.5) and the second `1.4`.

frontend/views/containers/forms/FormSignup.vue
sandrina-p: Github shows this as "new code" but it was only moved from a previous file. I have #661 on my todo-list. Once this is merged is the first thing I'll clarify/solve.

frontend/views/containers/proposals/ProposalBox.vue
sandrina-p: There should be no side-effects if you are not aware of them or don't handle them correctly. In this case, this side-effect is doing nothing wrong... but I'll look again and try to avoid it

frontend/views/containers/proposals/AddMembers.vue
sandrina-p: I did it inside a loop so the catch has access to the `inviteeName`. However, I fixed it so catch breaks the loop. Regarding performance, this is a small loop, I'm pretty sure the performance won't be noticeable, and we would be falling in the _premature optimization_ trap.

frontend/views/containers/proposals/ProposalBox.vue
sandrina-p: Gotcha. So, in summary, use tags when they wrap variables (who) and raw text. Otherwise, write the html directly as part of the variable value, so translators don't need to bother with extra information. I'll add it to our i18n docs.

frontend/model/contracts/group.js
sandrina-p: What someplace else?

taoeffect: Well, at the very least it should be done outside of group.js, which will be frozen, so any modifications to it must be carefully thought out. Nothing can be added to it with the intention of removing it in the future. So "someplace else" can be in the `gi.contracts/group/create` message

taoeffect: i.e. when you're creating a group in CreateGroup.vue, etc

sandrina-p: Exactly my thought, done!

frontend/views/pages/Join.vue
```diff
@@ -24,11 +24,11 @@ div
       p.c-switchEnter(v-if='isStatus("SIGNING")')
         i18n Already have an account?
         | &nbsp;
-        i18n.link(tag='button' @click='setStatus("LOGGING")' data-test='goToLogin') Log in
+        i18n.link(tag='button' @click='pageStatus = "LOGGING"' data-test='goToLogin') Log in
       p.c-switchEnter(v-else)
         i18n Not on Group Income yet?
         | &nbsp;
-        i18n.link(tag='button' @click='setStatus("SIGNING")' data-test='goToSingup') Create an account
+        i18n.link(tag='button' @click='pageStatus = "SIGNING"' data-test='goToSingup') Create an account
```
taoeffect: Typo: `goToSingup` => `goToSignup`

frontend/views/pages/Join.vue
```diff
@@ -130,17 +140,18 @@ export default {
...
       console.error(e)
       this.ephemeral.errorMsg = `${L('Something went wrong.')} ${e.message}`
-      this.setStatus('INVALID')
+      this.pageStatus = 'INVALID'
     }
   },
   methods: {
     isStatus (status) {
-      return this.ephemeral.pageStatus === status
+      console.log('hum', this.pageStatus)
```
taoeffect: Debug logging should be removed if it's not needed

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -29,16 +29,16 @@ export default {
...
       'currentGroupState'
     ]),
-    invitationWelcome () {
+    welcomeInviteSecret () {
       const invites = this.currentGroupState.invites
       return Object.keys(invites).find(invite => invites[invite].creator === 'GROUP_WELCOME')
     },
     link () {
-      return buildInvitationUrl(this.$store.state.currentGroupId, this.currentGroupState.invites[this.invitationWelcome].inviteSecret)
+      return buildInvitationUrl(this.$store.state.currentGroupId, this.currentGroupState.invites[this.welcomeInviteSecret].inviteSecret)
```
taoeffect:
```js
return buildInvitationUrl(this.$store.state.currentGroupId, this.welcomeInviteSecret)
```

frontend/views/containers/modals/CreateGroup.vue
```diff
@@ -121,30 +122,36 @@ export default {
       // create the GroupContract
       try {
         this.ephemeral.errorMsg = null
+        const initialInvite = createInvite(60, 'GROUP_WELCOME')
```
taoeffect: Should use INVITE_INITIAL_CREATOR

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,26 +6,40 @@ modal-template(ref='modal')
...
-  data () {
-    return {
-      link: 'https://groupincome.org/f4k3_H45h_t0_5h4r3'
+  computed: {
+    ...mapGetters([
+      'currentGroupState'
+    ]),
+    welcomeInviteSecret () {
+      const invites = this.currentGroupState.invites
+      return Object.keys(invites).find(invite => invites[invite].creator === 'GROUP_WELCOME')
```
taoeffect: INVITE_INITIAL_CREATOR

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,165 @@
...
+        let creator = null
+        let creatorPicture = null
+        let message = null
 
-      state.members = members
-
-      // Mocked histoy to show on members-circle
-      state.mincomeCurrencySign = '$'
-      state.history = [1.1, 1.3, 0.7, 1.05, 1, 1.3]
+        if (invite.creator === 'GROUP_WELCOME') {
```
taoeffect: INVITE_INITIAL_CREATOR (and anywhere else it's missing)

frontend/views/containers/proposals/ProposalBox.vue
taoeffect: > use tags when they wrap variables (who) and raw text

(or just raw text, yes)

> Otherwise, write the html directly as part of the variable value, so translators don't need to bother with extra information. I'll add it to our i18n docs.

😄 👍

frontend/model/contracts/group.js
```diff
@@ -14,11 +14,31 @@ import { merge, deepEqualJSONType } from '~/frontend/utils/giLodash.js'
...
+  inviteSecret: string,
+  numInvites: number,
+  creator: string,
+  invitee: optional(string),
+  status: string,
+  responses: mapOf(string, string),
+  expires: number
+})
+
+export function createInvite (numInvites: number, creator: string, invitee: string) {
```
taoeffect: I did like how you had it before, passing in an object, it made the rest of the code that uses it clearer. You can keep it this way if you'd like, or if you'd like to revert but are unsure of how to keep the flowtypes, have a look at how the Vuex actions in state.js are defined (e.g. handleEvent). In their function signature it shows how to add types to object parameters in functions

frontend/model/contracts/group.js
```diff
@@ -183,7 +208,7 @@ DefineContract({
       validate: objectOf({
         proposalHash: string,
         vote: string,
-        passPayload: optional(unionOf(object, string)) // TODO: this, somehow we need to send an OP_KEY_ADD GIMessage to add a generated once-only writeonly message public key to the contract, and (encrypted) include the corresponding invite link, also, we need all clients to verify that this message/operation was valid to prevent a hacked client from adding arbitrary OP_KEY_ADD messages, and automatically ban anyone generating such messages
+        passPayload: optional(unionOf(object, string)) // TODO: this, somehow we need to send an OP_KEY_ADD GIMessage to add a numInvites once-only writeonly message public key to the contract, and (encrypted) include the corresponding invite link, also, we need all clients to verify that this message/operation was valid to prevent a hacked client from adding arbitrary OP_KEY_ADD messages, and automatically ban anyone generating such messages
```
taoeffect: This seems like it was an unfortunate casualty of a project-wide search/replace

frontend/views/containers/forms/FormLogin.vue
```diff
@@ -0,0 +1,116 @@
...
+        })
+        this.$emit('submitSucceeded')
+      } catch (error) {
+        this.ephemeral.errorMsg = error.message
+        console.error(error)
+      }
+    },
+    forgotPassword () {
+      // TODO: implement forgot password
+      console.log('Coming soon')
```
taoeffect: You can add an `alert(L('Coming soon'))` here instead, so that the user can see it too

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -119,86 +109,38 @@ export default {
...
-            console.warn(`The invitee ${inviteeName} isn't registered. TODO #614`)
-          }
-
-          this.ephemeral.formFeedbackMsg = {
-            text: L('Invites to {invitationsSent} sent successfully!', {
-              invitationsSent: invitationsSent.join(', ')
-            }),
-            class: 'has-text-success'
-          }
-          this.resetInvitations()
```
taoeffect, line 195: Great job on saving many lines of code! They are in a better place now :)

frontend/views/containers/proposals/ProposalBox.vue
```diff
@@ -42,17 +58,20 @@ export default {
     title () {
       const { identityContractID } = this.proposal.meta
       const username = this.$store.state[identityContractID].attributes.name
-      const currentUsername = this.ourUserIdentityContract.attributes.name
-      const who = username === currentUsername ? L('You') : username
+      const who = username === this.ourUsername ? `<strong>${L('You')}</strong>` : `<strong>${username}</strong>`
```
taoeffect: I know you didn't write this code, but I think for now it would be better if we fetched their username using `this.proposal.meta.username` instead of going this roundabout way (see lines 59-60)

frontend/views/containers/proposals/ProposalBox.vue
```diff
@@ -42,17 +58,20 @@ export default {
...
+          ? L('{you} proposed:', { you: who })
+          : L('{someone} proposed:', { someone: who })
       }
 
-      return username === currentUsername
-        ? L('You are proposing')
-        : L('{who} is proposing', { who })
+      return username === this.ourUsername
+        ? L('{you} are proposing:', { you: who })
+        : L('{someone} is proposing:', { someone: who })
```
taoeffect: A couple issues:

1. The check for `username === this.ourUsername` has already been done, so it need not be done again. You can cache the result in a boolean const variable at the top to only perform the check once, thereby DRY'ing it, which helps improve readability, and also gives a slight boost to performance since `this.ourUsername` is a function that has to perform computation, and this way we only need to do the computation once
2. Regarding i18n stuff, remember that it's dangerous to translate single words. On line 61, the single word "You" is being translated, and then it's being passed as a variable via `{you}` in some cases. This is super confusing, and can lead translators to produce incorrect translations. So we need to avoid translating single words here. So Instead of `{you} proposed:`/`{you} are proposing:`, we should write `{strong_}You{_strong} proposed:`/`{strong_}You{_strong} are proposing`, and instead of `{someone} proposed`, we can write `{username} proposed`, which will give better context to translators about what is going into that variable

frontend/views/containers/proposals/ProposalVoteOptions.vue
```diff
@@ -113,35 +113,6 @@ export default {
...
-                ${groupName} has voted to invite you! Horray!
-                Here's your special invite link:
-                
-                ${buildInvitationUrl(this.$store.state.currentGroupId, payload.passPayload.inviteSecret)}`
-              },
-              identityContract.attributes.mailbox
-            )
-            await sbp('backend/publishLogEntry', inviteToMailbox)
-          }
-        }
```
taoeffect, line 144: 👏

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,165 @@
...
   },
   methods: {
-    accept: async function () {
+    isStatus (status) {
+      console.log('hum', this.pageStatus)
+      return this.pageStatus === status
+    },
+    setStatus (status) {
+      this.ephemeral.pageStatus = status
+    },
```
taoeffect: `setStatus` is no longer being used and can be removed

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,165 @@
...
+
+    group-welcome.c-welcome(v-else-if='isStatus("WELCOME")')
+
+    .c-broken(v-else-if='isStatus("INVALID") || isStatus("EXPIRED")')
+      svg-broken-link.c-svg
+      i18n.title.is-1(
+        v-if='isStatus("EXPIRED")'
+        tag='h1'
+        :args='LTags()'
+      ) Oh no! {br_} Something went wrong.
```
taoeffect: Typically there are no spaces before and after a `<br/>`, it's the difference between:

```
Oh no![ ]
[ ]Something went wrong.
```

And

```
Oh no!
Something went wrong.
```

Are we sure they need to be there? If not, on line 46 is the same issue

frontend/views/pages/Join.vue
```diff
@@ -1,79 +1,165 @@
...
+      p.has-text-1 {{ ephemeral.errorMsg }}
+      i18n.c-goHome(tag='button' @click='goHome') Take me home
+
+  // TODO - Remove these. Only used for debug.
+  .buttons.c-debug
+    button.is-small.is-outlined(@click='pageStatus = "LOADING"') LOADING
+    button.is-small.is-outlined(@click='pageStatus = "SIGNING"') SIGNING
+    button.is-small.is-outlined(@click='pageStatus = "WELCOME"') WELCOME
+    button.is-small.is-outlined(@click='pageStatus = "EXPIRED"') EXPIRED
+    button.is-small.is-outlined(@click='pageStatus = "INVALID"') INVALID
```
taoeffect: Should these be removed now that you're done?

frontend/views/utils/translations.js
```diff
@@ -31,6 +31,19 @@ String with Vue components inside:
...
+
+  i18n(
+    :args='{ count: 5, ...LTags("strong") }'
+  ) Invite {strong}{count} members{strong} to the party!
+
+- Write HTML when it wraps only the variable.
+-- That way translators don't need to worry about extra information.
+  i18n(
+    :args='{ count: "<strong>5</strong>" }'
+  ) Invite {count} members to the party!
```
taoeffect, line 46: Excellent! Thank you for improving our docs! 🎉 🥇 🙌

taoeffect (changes requested):
Review finished! See comments below and above. When these are addressed I'll test it, and if it works I think it should be GTM :)

frontend/model/contracts/group.js
```diff
@@ -41,34 +61,39 @@ DefineContract({
...
-      sharedValues: string,
-      mincomeAmount: number,
-      mincomeCurrency: string,
-      proposals: objectOf({
-        [PROPOSAL_INVITE_MEMBER]: proposalSettingsType,
-        [PROPOSAL_REMOVE_MEMBER]: proposalSettingsType,
-        [PROPOSAL_GROUP_SETTING_CHANGE]: proposalSettingsType,
-        [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposalSettingsType,
-        [PROPOSAL_GENERIC]: proposalSettingsType
+      invites: mapOf(string, inviteType),
```
taoeffect, line 70: Great! And on line 63 you can change `objectOf` => `objectMaybeOf` to help prepare for the future when this contract is frozen

frontend/views/containers/proposals/ProposalBox.vue
sandrina-p: Actually it was me who did this a month ago eheh. fixed

frontend/views/containers/proposals/ProposalBox.vue
sandrina-p: > this.ourUsername is a function that has to perform computation, and this way we only need to do the computation once

Not really. `this.ourUsername` is just a variable at the moment we access it in the component as a getter. It was already computed (function executed) when the component is being mounted and computing the getters, and it does it only once.

But nevertheless, I did what you suggested for both 1 and 2.

frontend/views/pages/Join.vue
sandrina-p: Removed the second space, the first I can't join `!{br_}` because of the following error:
> 41:14  String interpolation operators must not be used

taoeffect (approved):
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

Woohoo!!! Changes look great and just finished testing it locally and it works great too! 🙌 

Great job @sandrina-p!! 😄

PR #765, opened 2019-11-06
Uniformize invitation link display across the app

### LINKED ISSUE #756: Uniformize invitation link display across the app
# Problem

Recently @mmbotelho designed a better invitation link display with "click-to-copy" functionality. That was implemented only in the Invite Links table (#664).

image

We need to also implement it at:
- [Modal "Add members"](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=2335%3A1437) when the group has less than 3 members.
- [Proposal widget](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1579%3A202) when an invite member proposal is accepted.

### Solution
- Implement the component in the missing places.
- Add a usage demo to the Design System

SebinSong:
closes #756 

--- works done on this PR ---
1) Changed `CopyToClipboard.vue` to `LinkToCopy.vue` and makes it contain all the needed UI components inside it(span and button elements. This component previously only encapsulated the 'click to copy' logic, receiving the UI elements as slots).
2) replaced the invitelinks across the app with `LinkToCopy.vue` component.
3) added `LinkToCopy.vue` to `/app/design-system`.
4) Tweaked the cypress test logic that involves the invite link. ( @sandrina-p  Although I tried my best to put myself into your shoes in this part, if the modified code doesn't makes sense to you or hurts your original intention, please leave a comment on it.)


frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+          el.parentNode.removeChild(el)
+        }
+      }
+    }
+  }
+}
+</script>
+
+<style lang='scss' scoped>
+@import '../../assets/style/_variables.scss';
```
taoeffect: Here we use `@import '@assets/style/_variables.scss';` now :)

mmbotelho (changes requested):
Hello Sebin!

- Add members modal approved ✅
- Invite link table (group settings page)
   * The dropdown menu is not working for me on mobile (chrome), clicking the 3 dots on the right-hand side doesn't do anything;
- Invite link on the proposals widget (Dashboard):
  
Captura de ecrã 2019-11-07, às 12 02 59
Captura de ecrã 2019-11-07, às 12 01 54

The link should not be displayed completely. It should be made to fit the remaining available space horizontally (as you can see in the mockup). However, it should have a minimum width of 112px (7rem). If the component is smaller than that it should be displayed on the line below, and take up about 50% of the available space there.

Does this make sense to you? Let me know if you need me to explain it differently or to jump in a call and talk about it :)

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+      const spacing = 16
+      const x = scrollX + left + width / 2 - this.ephemeral.tooltip.width / 2
+      const y = scrollY + top - (this.ephemeral.tooltip.height + spacing)
+
+      this.ephemeral.tooltipPosition = `transform: translate(${x}px, ${y}px)`
+    }
+  },
+  directives: {
+    appendToBody: {
+      // refer to the comment in Tooltip.vue for the further information on this custom directive.
```
taoeffect: If this is identical code to what's in Tooltip.vue, then to DRY it, can you put `appendToBody` in a Vue mixin and import it here and in Tooltip.vue?

And I am wondering also if the same applies to `adjustPosition`?

taoeffect (changes requested):
Minor DRY related concerns

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -248,12 +245,11 @@ export default {
...
-    }
 
-    .c-invite-link-button {
-      margin-left: $spacer-xs;
-      flex-shrink: 0;
-      font-weight: normal;
+      ::v-deep .c-copy-button {
+        flex-shrink: 0;
+        font-weight: normal;
+      }
```
sandrina-p: We'll need this code in other places where this component is used. So, I would just move this directly to the `InviteLinkToCopy` component

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+  :is='tag'
+  @click='copyToClipboard'
+  data-test='invitationLink'
+)
+  input.c-invisible-input(
+    type='text'
+    ref='input'
+    :value='inviteLink'
+  )
+  span.link.has-ellipsis.c-link {{ inviteLink }}
```
sandrina-p: For people using SR (Screen Readers), the link will be announced to them twice (the input and the span). We need to hide one of the elements. In this particular case, to make people's life easier to copy the text (besides the copy button), I would suggest to keep the input (native element with easy copy built-in mechanism) and hide the span. To do this add a `aria-hidden="true"` to the span.

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+  button.is-icon-small.has-background.c-copy-button(:aria-label='L("Copy link")')
+    i.icon-copy.is-regular
+  transition(name='fade')
+    i18n.c-tooltip(
+      v-if='ephemeral.isTooltipActive'
+      :style='ephemeral.tooltipPosition'
+      key='test'
+      tag='p'
+      v-append-to-body=''
+    ) Copied to clipboard!
```
sandrina-p: What is preventing us from using tooltip component here?
Tooltip could be the wrapper of everything in the component and we can manually adjust when it is visible with the prop `shouldShow` (now I would rename it to `isVisible`). You can even add the fade transition to the tooltip component so its consistent across all tooltips.

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+@import '../../assets/style/_variables.scss';
+
+.c-wrapper {
+  position: relative;
+
+  .c-invisible-input {
+    position: absolute;
+    pointer-events: none;
+    opacity: 0;
+    z-index: -100;
```
sandrina-p: I believe `z-index: -100` is not need, because we already use `pointer-events: none`.

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+
+.c-tooltip {
+  position: absolute;
+  top: 0;
+  left: 0%;
```
sandrina-p: when 0, there's no need to add unit. `0` is enough

frontend/views/pages/DesignSystem.vue
```diff
@@ -1163,6 +1191,11 @@ table {
   }
 }
 
+.c-invite-link-demo {
+  display: flex;
+  align-items: center;
+}
```
sandrina-p: A component should work visually by itself, without any additional CSS. Adding styles here usually it's a reflection that something is missing to the original component. In this case, the component should have these properties by default. 

Don't forget to review the other places where InvityLink is used and remove duplicated CSS.

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -42,8 +42,15 @@ export default {
...
 
 .c-link {
+  display: flex;
+  align-items: center;
+  justify-content: center;
   margin-top: $spacer-lg;
   margin-bottom: $spacer-sm;
+
+  ::v-deep .c-copy-button {
+    margin-left: $size-3/2;
```
sandrina-p: We don't use `$size-x` variables for spacing, but for text size.
When implementing a design, we with @mmbotelho use a spacing sytem, 4px, 8px, 16px, 24px, 32px, etc.. If you see something that doesn't match that, use the closest value and double-check with @mmbotelho to see if it's the correct one. I believe that in `InviteLinkToCopy` the space between the link and the button is the same anywhere, so there's no need to override it here.

frontend/views/components/InviteLinkToCopy.vue
```diff
@@ -0,0 +1,134 @@
...
+      :style='ephemeral.tooltipPosition'
+      key='test'
+      tag='p'
+      v-append-to-body=''
+    ) Copied to clipboard!
+</template>
+
+<script>
+export default {
+  name: 'InviteLinkToCopy',
```
sandrina-p: The way you did this component is so agnostic (👏 ) that it doesn't make sense to call it "InviteLinkToCopy". It can take any type of link to copy, so I would just call it `LinkToCopy` and rename the prop `inviteLink` to `link`.

sandrina-p:
> Invite link on the proposals widget (Dashboard):

@mmbotelho what you are asking is kinda tricky.. (I would say impossible as far as I know) because ellipsis only works when they are isolated, in one line as a single block. That's why it works on group settings and in the modal, but not here.
So, my suggestion would be to display the link in a new line, like this:

image

@SebinSong to make this work, you need 2 small but important changes in the code:

- **ProposalBox.vue**
Make ellipsis work correctly when used inside a grid element. [Know more about preventing a grid blowout](https://css-tricks.com/preventing-a-grid-blowout/)
```diff
.c-main {
+  min-width: 0;
}
```

- **ProposalItem.vue**
  - Turn the text "_Please send... to the group:_" into a block element `<p/>` and remove the white space.
  - Remove all the custom styles to `.c-invite-link`. None of it is needed. Make sure to address the flexbox issue I mentioned previously in the CR before doing this.

@pieer if you have any idea how to solve this respecting the original design requirements, please let us know!

SebinSong:
Hi @mmbotelho, As to the three dots button not working, In current implementation, that button is supposed to open the native 'share link' dialog box on mobile view. 
I'm assuming by 'chrome mobile' you meant the chrome dev-tool's mobile device mode? I guess this kind of device-specific behavior can't be simulated by the dev-tool yet. So, I'm afraid we can't really see if it works or not at the moment until we find a way to test localhost on our mobile device. :)

For the other issue related to the invite-link, I will change it in the way @sandrina-p suggested unless you have any disagreement or anything to add to it.

taoeffect:
> So, I'm afraid we can't really see if it works or not at the moment until we find a way to test localhost on our mobile device. :)

Copying from Slack in case it's useful to anyone here: ngrok makes it simple to do this: https://ngrok.com/

mmbotelho:
> I'm assuming by 'chrome mobile' you meant the chrome dev-tool's mobile device mode? I guess this kind of device-specific behavior can't be simulated by the dev-tool yet

I was looking at it on chrome desktop, but with a smaller viewport. We should have a fallback for it though, in case someone is using a desktop window but with a smaller viewport (as I was). Does that make sense @SebinSong ?

I was not aware that what I was asking to do with the share link was difficult, but I totally agree with what @sandrina-p suggested! 

Let me know when you want me to review this again.

SebinSong:
@mmbotelho Yeah, it does. Ok, I will work on them. :) Thanks for clarifying. 👍

frontend/views/components/InviteLinkToCopy.vue
SebinSong: Fixed! :)

frontend/views/components/InviteLinkToCopy.vue
SebinSong: fixed!

frontend/views/components/InviteLinkToCopy.vue
SebinSong: you are right. fixed!

frontend/views/components/InviteLinkToCopy.vue
SebinSong: deleted!

SebinSong:
@mmbotelho Hi Margarida, I got a couple of questions while working on your change requests here.

> We should have a fallback for it though, in case someone is using a desktop window but with a smaller viewport (as I was)

Is this dropdown what you intended as the fallback? it will show on every browsers that don't support [webShare api](https://css-tricks.com/how-to-use-the-web-share-api/)(which includes all desktop browsers when they are in a mobile size screen), which is the thing that triggers the native share dialog box on mobile.


fallbackMenu


If so, what should happen when the 'share link' is clicked?

mmbotelho:
@SebinSong the actions for each table line are:
- See original proposal (for all links except the "Anyone" link);
- Revoke link (except the "Anyone" link);
- Copy link (on desktop, this is done by clicking on the link itself to copy it to the clipboard, on mobile devices, it's one of the options on the dropdown menu);
- Share link (only on mobile devices, because those are the ones that support the webShare API).

So to answer your question, the "share link" option is only available when the wwebshare API is available. If it isn't, when using a desktop browser with a smaller viewport for example, the only option there is to copy the link. 

Does this make sense? I am going to attach some screenshots below that might help:
Captura de ecrã 2019-11-21, às 13 11 04
Captura de ecrã 2019-11-21, às 13 12 31

If this is still not clear we can jump on a quick call (at whatever time works best for you) and I can explain it better! 

Let me know!

SebinSong:
@mmbotelho Sorry, It took me a while to come back to this work and I just noticed that this PR was meant to close issue #756 which is about implementing 'click to copy' functionality across the app. the change request you pointed out about this invite link table doesn't seem to be that related to the issue itself. shouldn't we address this problem as part of the work on issue #664?, not here. :) (Or, maybe by opening a separate issue for it)
Thank you.

frontend/views/components/InviteLinkToCopy.vue
SebinSong: It's redone in the way you suggested here. fixed!

frontend/views/components/InviteLinkToCopy.vue
SebinSong: The issue here(along with another issue [here](https://github.com/okTurtles/group-income-simple/pull/765#discussion_r344121664)) is fixed by using the existing `Tooltip.vue` component. That way we can DRY it without creating extra Vue mixin file. :)

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: Fixed!

frontend/views/pages/DesignSystem.vue
SebinSong: Fixed!

frontend/views/containers/modals/InviteByLink.vue
SebinSong: Fixed!

frontend/views/components/InviteLinkToCopy.vue
SebinSong: I agree. Changed to `LinkToCopy.vue`.

SebinSong:
Hi!
Basically every requests brought up so far have been fixed(As far as I can tell, pls feel free to double-check tho) in the few newest commits, except the one by Margarida [here](https://github.com/okTurtles/group-income-simple/pull/765#issuecomment-557058971) which I believe can(or 'Should') be dealt with as a part of the #664 .
(Which means as long as she agrees on my suggestion [here](https://github.com/okTurtles/group-income-simple/pull/765#issuecomment-562731411) and there's nothing else to be changed, I believe it's ready to be merged)

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,11 +6,8 @@ modal-template(ref='modal')
   .c-container
     i18n.is-title-4(tag='h3') Share this link to grant access to your group.
     i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 60 free passes!
-    .c-link
-      a.link.has-icon(:href='link' target='_blank' data-test='invitationLink')
-        i.icon-link
-        | {{link}}
-    i18n.has-text-1(tag='p' :args='{ expireDate }') This invite link expires on the {expireDate}.
+    link-to-copy.c-link(:link='link')
+    i18n.has-text-1(tag='p') This invite link expires on the 4th of February.
```
taoeffect: Here I would keep the `{expireDate}` argument as it was, since it will only need to be added back.

frontend/views/components/GroupSettings/InviteLinks.vue
```diff
@@ -129,7 +136,7 @@ export default {
           title: L('Your invite'),
           url: inviteLink
         })
-      }
+      } else this.$refs.webShareFallbackBtn[0].handleClick()
```
taoeffect: One issue and one question here:

- Issue: for readability and [safety](https://dwheeler.com/essays/apple-goto-fail.html), it's good practice to use braces with if/else statements, and avoid the syntax that lets you not use them
- Question: why are you using `webShareFallbackBtn[0]` here instead of `webShareFallbackBtn`? I'm not saying it's wrong, it's just wondering why because I'm not familiar with this. If this is indeed how it should be done, then can you add a comment explaining why?

frontend/views/components/Tooltip.vue
```diff
@@ -31,8 +31,8 @@ export default {
   props: {
     text: String,
     // Force to show tooltip manually
-    shouldShow: Boolean,
-    hasTextCenter: Boolean,
+    isVisible: Boolean,
+    isTextCenter: Boolean,
```
taoeffect, line 35: Did you check the entire app to make absolutely sure that all instances of `shouldShow` and `hasTextCenter` usage with `Tooltip` have been replaced?

I noticed that `hasTextCenter` still needs to be replaced with `isTextCenter` in `InviteLinks.vue:35` and `GroupPledgesGraph.vue:36`. So those need to be updated to reflect this rename, and double-check that there are no other missed files.

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,11 +6,8 @@ modal-template(ref='modal')
   .c-container
     i18n.is-title-4(tag='h3') Share this link to grant access to your group.
     i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 60 free passes!
-    .c-link
-      a.link.has-icon(:href='link' target='_blank' data-test='invitationLink')
-        i.icon-link
-        | {{link}}
-    i18n.has-text-1(tag='p' :args='{ expireDate }') This invite link expires on the {expireDate}.
+    link-to-copy.c-link(:link='link')
```
taoeffect, line 9: Clicking this link results in the following console error:

```
[Vue warn]: Invalid prop: type check failed for prop "isVisible". Expected Boolean, got String with value "isVisible".
```

Screen Shot 2020-01-14 at 1 50 43 PM

frontend/views/components/LinkToCopy.vue
```diff
@@ -1,33 +1,46 @@
...
+  @click='copyToClipboard'
+  data-test='invitationLink'
+)
   input.c-invisible-input(
     type='text'
     ref='input'
-    :value='textToCopy'
+    :value='link'
+  )
+  span.link.has-ellipsis.c-link(aria-hidden='true') {{ link }}
```
taoeffect: I can no longer right-click the link to copy it. Can that be added back?

Screen Shot 2020-01-14 at 1 56 12 PM

taoeffect (changes requested):
Great work @SebinSong! I've finished my review, and I was able to find a console error when clicking on the link, a right-click issue, and some misc. stuff. I think this will be GTM once those are addressed, but also waiting to hear from @sandrina-p to double-check if she can find anything else.

frontend/views/components/LinkToCopy.vue
sandrina-p: @SebinSong, the span should keep being an anchor with href. the only difference is we need to use preventDefault when clicking the link.

sandrina-p (commented):
Besides what @taoeffect pointed, everything else seems fine for me. Good job @SebinSong !

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: refs

As you can see from the screenshot attached, There's a difference in how Vue stores the reference to a 'plain html element' and a 'Vue component'. (For some reason, it stores a ref to a vue component in an array). That's the reason why I wrote that line that way.

(using `$refs.webShareFallbackBtn` generates the following error)
erroraboutref

frontend/views/components/LinkToCopy.vue
SebinSong: Fixed it in the way @sandrina-p suggested!

frontend/views/containers/modals/InviteByLink.vue
SebinSong: Fixed!

frontend/views/components/Tooltip.vue
SebinSong: I did the project-wide search for any files using `Tooltip.vue` and those two you mentioned were only ones to be fixed. :)

frontend/views/components/GroupSettings/InviteLinks.vue
SebinSong: the issue has been fixed.

frontend/views/containers/modals/InviteByLink.vue
SebinSong: fixed.

frontend/views/containers/proposals/ProposalBox.vue
```diff
@@ -171,7 +171,7 @@ export default {
   grid-area: main;
   margin-top: 1.5rem;
   word-break: break-word;
-  min-width: 0; // So ellipsis work correctly inside grid.
```
taoeffect, line 174: Unless the comment is saying something incorrect let's keep it around in case it's helpful to others

frontend/views/containers/modals/InviteByLink.vue
```diff
@@ -6,11 +6,8 @@ modal-template(ref='modal')
...
+    i18n.has-text-1(tag='p' :args='{ expireDate }') This invite link expires on { expireDate }.
```
taoeffect: Screen Shot 2020-01-23 at 5 31 14 PM

our string replacement function that i18n uses isn't very smart, and needs no spaces around the variable name

taoeffect (changes requested):
some quick comments, still working through my review

frontend/views/containers/modals/InviteByLink.vue
SebinSong: fixed.

frontend/views/containers/proposals/ProposalBox.vue
SebinSong: it has been added back!

taoeffect (approved):
Great job @SebinSong! 🥇 🎉 😄 

Now just needs approval from @mmbotelho once she checks out your latest changes.

pieer (approved):

mmbotelho (approved):
Approved! Great work @SebinSong 😄

PR #776, opened 2019-11-13
Improve signup and login forms validation handlings

### LINKED ISSUE #661: Figure out (and fix) the removal of the mysterious 'v'

sandrina-p:
- Close #661.
- Make "debounce" smarter on all inputs' validations on signup and login forms.


frontend/views/containers/forms/FormPassword.vue
```diff
@@ -20,10 +22,9 @@ label.field
       i(:class='isLock ? "icon-eye" : "icon-eye-slash"')
 
   i18n.error(
-    tag='p'
-    v-show='v[name].$error'
+    v-show='vForm[name].$error'
```
sandrina-p: @taoeffect I tried to use v-error in this component, but couldn't make it work because the form key is dynamic and v-error always look for `this.$v.form.` which doesn't exist in this component (it's `this.vForm`).

pieer (dismissed):

mmbotelho (changes requested):
Some copy changes, let's make these errors more human:

#### Sign up
- **"Minimum of 7 characters"** change to "Your password must be at least 7 characters long."
- **"cannot contain spaces**" change to "A username cannot contain spaces."
- **"not an e-mail"** change to "Please enter a valid email."
- **"required"** change to "A username is required."
- **"e-mail required"** change to "An email is required."
- **"name is unavailable"** change to "This username is already being used."
- I can create more than one account with the same email. Shouldn't we have a validation here as well? @taoeffect I know this is not in the scope of this PR, so feel free to do it in another one.

#### Log in
- The **"Invalid username or password"** error should be turned into a banner error. I will add this case to Figma;
- **"cannot contain spaces"** change to "A username cannot contain spaces."
- **"Minimum of 7 characters"** change to "Your password must be at least 7 characters long.";

These were all the errors I found. Please let me know if there's some more errors that I missed!

sandrina-p:
@mmbotelho 

> I can create more than one account with the same email. Shouldn't we have a validation here as well? @taoeffect I know this is not in the scope of this PR, so feel free to do it in another one.

We have a TODO in our code for that.

> The "Invalid username or password" error should be turned into a banner error. I will add this case to Figma;

That will be done when working on #688

shared/feedbackMessages.js
```diff
@@ -0,0 +1,16 @@
...
+"required" change to "A username is required."
+"e-mail required" change to "An email is required."
+"name is unavailable" change to "This username is already being used."
+I can create more than one account with the same email. Shouldn't we have a validation here as well? @taoeffect I know this is not in the scope of this PR, so feel free to do it in another one.
+
+Log in
+
+The "Invalid username or password" error should be turned into a banner error. I will add this case to Figma;
+"cannot contain spaces" change to "A username cannot contain spaces."
+"Minimum of 7 characters" change to "Your password must be at least 7 characters long.";
```
taoeffect: This file shouldn't exist

frontend/views/containers/forms/FormPassword.vue
taoeffect: Have you tried passing down `$v` instead of `$v.form`?

taoeffect (changes requested):
This PR is somewhat cosmetic, it improves the situation described in #661 slightly. I left a couple of comments. Note that the tests aren't passing.

It would be great if this could do a more thorough job of closing #661, by investigating whether it's possible to delete some of the current spaghetti trampoline code (i.e. all the event emission code) , and the still-confusing, possibly unnecessary dichotomy between `vForm` and `$v`.

frontend/views/containers/forms/FormPassword.vue
sandrina-p: this wouldn't fix it. the problem is not the prop, is the v-error because it looks for `this.$v.form` when it should look for `this.vForm.form`.

frontend/views/components/CreateGroupSteps/GroupMincome.vue
```diff
@@ -56,16 +56,16 @@ export default {
   },
   methods: {
     update (e) {
-      this.v[e.target.name].$touch()
+      this.vForm[e.target.name].$touch()
```
sandrina-p: @taoeffect 
> This PR is somewhat cosmetic, it improves the situation described in #661 slightly. I left a couple of comments. Note that the tests aren't passing.
> 
> It would be great if this could do a more thorough job of closing #661, by investigating whether it's possible to delete some of the current spaghetti trampoline code (i.e. all the event emission code) , and the still-confusing, possibly unnecessary dichotomy between `vForm` and `$v`.

I understand the issue here, but I'm not a master with vuelidate to come with a better answer. I can try to take a look again later but maybe @pieer has a better idea?

UPDATE: I took another look, tried several things... but at the end, using vForm + emit was the most clean/flexible way to handle all the cases.

UPDATE2: So... after trying and failing over and over again, I decided to just keep the prop name to match exactly the vuelidate key. `$v`. That way v-error can also work on child elements that inherit vuelidate (ex: password).

Another thing I did was to create a debounceValidations mixin so we can start using it around. @greg please take a look! Note: I tried to convert those mixins to a directive, but it seems Vue doesn't support event listeners in directives that have access to _updated_ vnode.

mmbotelho (dismissed):
Perfect! Thank you @sandrina-p 😄

frontend/views/containers/forms/FormLogin.vue
```diff
@@ -8,24 +8,22 @@ form(
...
-  // TODO #661 - improve :v
   form-password(
     :label='L("Password")'
-    :value='form'
-    :v='$v.form'
-    @enter='login'
-    @input='(newPassword) => {password = newPassword}'
+    :$v='$v'
+    @input='debounceField("password")'
+    @blur='updateField("password")'
```
taoeffect: Are you positive you need these? Can you check it see if it's possible to get it working without them?

I noticed that in FormPassword.vue you are now binding the `v-model` to the `.$model`, which might make this no longer necessary?

If you try and are unable to get rid of them, can you write a comment here and/or in FormPassword.vue explaining why it these are required?

sandrina-p: checking...

sandrina-p: Yep, that's right, I added them directly to FormPassword and it works!

frontend/views/containers/forms/FormPassword.vue
```diff
@@ -11,8 +11,8 @@ label.field
       :name='name'
       :data-test='name'
       v-model='$v.form[name].$model'
-      @input='e => $emit("input")'
-      @blur='e => $emit("blur")'
+      @input='debounceField("password")'
+      @blur='updateField("password")'
```
taoeffect, line 15: Same question: did you check whether they need to even be here? :)

If not, they can be removed here, and elsewhere where they aren't needed (in the other files).

And if they are needed, an explanation of why `.$model` is insufficient

EDIT: nevermind, I see, it's to prevent error message from showing immediately

taoeffect (approved):
Approved! 😄

PR #777, opened 2019-11-15
Improve headers scroll behavior and fix related a11y issues

### LINKED ISSUE #626: Minor Design system fixes
### Problem
I found some minor discrepancies between the [design system on Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1%3A7) and what was actually implemented. This list of changes was made after reviewing the `app/design-system` page in `master` (August 19, 2019).

### Solution 

- [x] H1 and H2 should be responsive ([figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=3070%3A106));

----
- [x] Fix small typo, on the subtitle typography atom. "This is an subtitle" should be "This is a subtitle";

----

- [x] Buttons that have icons as a prefix or sufix should not have a margin on the left or a margin on the right side:
Captura de ecrã 2019-08-19, às 16 52 54

----

-  [x] Fix small text movement in some inputs (dropdown, text + dropdown combo, text + button, text with icon prefix), when they change from their normal state to focus (see gif below - it takes a while to show the change):
input-movement

----

- [x] Remove hover CSS changes from the text input with icon prefix:
Captura de ecrã 2019-08-19, às 19 56 11

----

- [x] The subtitle on modals does not have the correct CSS - the subtitle component in the typography section is correct:
Captura de ecrã 2019-08-19, às 20 11 24

----

- [x] At the moment, it's possible to resize textareas horizontally. I understand that this is the default behaviour of textareas. The vertical resize is useful, but the horizontal one can be problematic. Is there a way to limit this movement? Bonus question, for 1000€ - how difficult is it to make a textarea increase its vertical size as the content needs it (aka, dynamically)? [I found this codepen example](https://codepen.io/vsync/pen/frudD)

textarea2

----

- [x] Update selectbox icon (chevron)

### LINKED ISSUE #767: Pin Profile.vue to bottom lower-left
### Problem

When the screen is too short, the navbar hides the user's profile.

### Solution

Pin the Profile.vue box to the bottom, so that it's always visible, while making sure that it still looks good on tablet/mobile. Here's a basic idea of what I mean:

thing

### LINKED ISSUE #768: Update implementation of the tablet + mobile headers
### Problem

I noticed that there was nothing in the Design System (Figma) that specified how the app header should behave on scroll, specifically on tablet and mobile sized screens. The current implementation of this is a bit broken as well - the title of the page flows but the menu and information icons remain in the same place.

### Solution

I have added a new section to the "Headers" artboard, that specifies how this component should behave when the user scrolls. [You can view it here](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=543%3A4369)

Let me know if you have any questions about this!

sandrina-p:
- Closes #767 
- Closes #768
- Closes #626
- All unused breakpoints were removed so we stick with only with the 3 used on our DS: `phone`, `tablet` and `desktop`.
- Refactored `.is-[n]` classes to `.is-title-[n]`
- @mmbotelho anything related to those sidebars or headings, feel free to point out!
- @taoeffect I added a new dependency [`wicg-inert`](https://github.com/WICG/inert), it's a polyfill for `inert`. `inert` is an HTML attribute used to prevent any type of interaction with that element, by using the cursor or the keyboard. At the moment only Chrome supports it. We use it on both sidebars, to hide the content when they are invisible, so when using the keyboard we don't jump to the links inside them. You can read more about `inert HTML attribute` it this article: [[aria-hidden="true"] elements contain focusable descendants](https://web.dev/aria-hidden-focus/) 
- Note: If you wanna test keyboard/SR navigation, I recommend you to use Chrome. Both Safari and Firefox are known for their poor support when it comes to keyboard nav. Be aware with a current issue with modals #778

image

image

image


frontend/views/containers/sidebar/Navigation.vue
```diff
@@ -135,6 +124,10 @@ export default {
     logo () {
       const name = this.colors.theme === 'dark' ? '-white' : ''
       return `/assets/images/logo-transparent${name}.png`
+    },
+    isInert () {
+      // TODO: Update window.innerWidth on resize.
+      return !this.ephemeral.isActive && window.innerWidth < 1200
```
taoeffect: This exists twice, so it needs to be DRY'd somehow, perhaps a mixin or something

frontend/index.html
```diff
@@ -55,7 +55,7 @@ <h1 class="title">Outdated browser</h1>
       <section class="hero is-fullheight">
         <div class="hero-body">
           <div class="container has-text-centered">
-            <h1 class="title">JavaScript disabled!</h1>
+            <h1 class="title is-1">JavaScript disabled!</h1>
```
taoeffect: In some places you add `is-1` and remove `title` to h1 elements. Here you just add `is-1`, while leaving `title`.

- Why do h1 elements need special classes at all?
- Should you remove `title` here like you did elsewhere? If not, should you add it to where you removed it?
- Why was `h1` removed from the `is-1` styling up above?
- Why is it that for `h2` you do not add `is-2` but for `h1` you add `is-1`? This seems very inconsistent and confusing/unnecessarily complicated to me, and there are no rules spelled out about any of this

taoeffect (changes requested):
Nice job! It does seem to close those issues! 👍 See comments. Note, the Avatar error is back in the console. Perhaps merging `master` will fix it, but double-check to make sure.

frontend/views/containers/sidebar/Navigation.vue
sandrina-p: It exists twice but I don't think it's a motive to create an abstraction for something so raw. It's the same as saying "margin-right: 1rem" or `<input type="email />` exists twice, we should create a CSS class / component for it.

Once we solve the "js resize/responsive" thing, we can find a better solution. For now, I would leave it as it is.

ref: [AHA programming](https://kentcdodds.com/blog/aha-programming)

frontend/assets/style/_typography.scss
```diff
@@ -75,43 +75,52 @@ $lineHeight: 3, 2.25, 1.6875, 1.3125, 1;
...
+// and always add the CSS class to the heading. (ex: h1.is-1).
+// What do you prefer?
+
 @include until($tablet) {
-  .is-1,
-  h1 {
+  h1,
+  // h1:not([class^=".is-"]), // Option 3: - This also fixes the bug!
+  // It other words, it means: "apply this to a h1 that doesn't contain .is-[n] class.
+  .is-1 {
```
sandrina-p: cc @pieer and @taoeffect

frontend/index.html
sandrina-p: Okay, I reviewed all the places. `.title` isn't needed at all (removed). Just `.is-[n]` is enough. I updated our DS page to explain a little more about it. 
See my other comment I added in this PR with a tricky situation.

frontend/views/pages/DesignSystem.vue
```diff
@@ -19,22 +19,22 @@ page(
...
       table
         thead
           th code
           th demo
         tr(v-for='(scale, index) in [1, 2, 3, 4]' :key='index')
           td
-            pre h{{scale}} or .title.is-{{scale}}
-          td.title(:class='`is-${scale}`') This is an H{{scale}} title
+            pre h{{scale}} or .is-{{scale}}
+          td(:class='`is-${scale}`') This is an H{{scale}} title
```
sandrina-p: @taoeffect and @pieer 
In our DS, the util classes usually follow this convention `has-[property]-[variant]`. (ex: `has-text-bold` or `has-background-success` or `has-text-1`). The titles are `.is-[n]`. This is like this because of Bulma, but I believe it should be more explicit. `is-title-[n]` or `has-title-[n]`. What do you think?

frontend/assets/style/_typography.scss
```diff
@@ -75,43 +75,52 @@ $lineHeight: 3, 2.25, 1.6875, 1.3125, 1;
...
+
+// Other solution could be to completely remove styles directly applied to h# tag
+// and always add the CSS class to the heading. (ex: h1.is-1).
 @include until($tablet) {
-  .is-1,
-  h1 {
+  h1,
+  // h1:not([class^=".is-"]), // Option 3: - This also fixes the bug!
+  // It other words, it means: "apply this to a h1 that doesn't contain .is-[n] class.
+  .is-1 {
```
sandrina-p: cc @pieer and @taoeffect

My option here would be to completely remove the styles added directly to h# tags and always add `.is-[n]` to a heading. I prefer to avoid coupling styling with semantics to avoid two things: 1. this type of bugs and 2. misusage of heading tags.
The `role="heading"` trick is funny but overkilling. Each time we would need to do it, probably we would forget it until someone notices the bug.
Using `h1:not([class^=".is-"])` fixes it, but it's a reflection of our ~lazy~/simple approach of not styling properly a heading with a `.is-[n]`. Besides that, it increased its CSS specificity and it would make our life harder if we need to override it.

mmbotelho (changes requested):
Captura de ecrã 2019-11-16, às 16 02 41
Captura de ecrã 2019-11-16, às 16 01 55

frontend/views/containers/sidebar/Navigation.vue
taoeffect: The specific concern here is the part where it checks the innerWidth against a constant value. That's not OK, since it means we have to remember to keep the constant value (1200) the same in both locations. Even in our CSS classes we make sure to use variables for such situations to avoid inconsistencies arising.

frontend/views/pages/DesignSystem.vue
pieer: For the record I repeat here: If you prefer to use 'is-title-1' instead of 'is-1', go for it!

frontend/assets/style/_typography.scss
pieer: I agree with you Sandrina

frontend/views/containers/sidebar/Navigation.vue
pieer: I think it's perfectly fine to have this line repeated. But If greg is concerned, about the constant, we can still extract that constant only and replace with a settings variable.

sandrina-p:
@mmbotelho, fixed!

frontend/views/containers/sidebar/Navigation.vue
sandrina-p: I searched for all raw values related to breakpoints and replaced them with constants.

mmbotelho (dismissed):
Perfect @sandrina-p ! Thank you for fixing the minor DS bugs as well! Approved!

frontend/utils/breakpoints.js
```diff
@@ -0,0 +1,3 @@
+// Same as CSS breakpoints at _variables.scss
+export const TABLET = 769
+export const DESKTOP = 1200
```
taoeffect: Would `frontend/views/utils/breakpoints.js` be a better place for this file, since this is only ever used with views?

frontend/assets/style/_typography.scss
```diff
@@ -76,42 +75,35 @@ $lineHeight: 3, 2.25, 1.6875, 1.3125, 1;
...
 
 @include phone {
-  h1 {
+  .is-title-1 {
     font-size: 1.25rem; // 20px
     line-height: 2.125rem; // 34px
   }
 
-  h2 {
+  .is-title-2 {
```
taoeffect, line 95: Do we not need corresponding `.is-title-3` here and above?

frontend/utils/breakpoints.js
sandrina-p: I also thought that, but then, in order for us to do a reusable responsive resize, I believe we would need to use the store or sbp somehow... and that would affect model folder, that's why I kept it here.

frontend/assets/style/_typography.scss
sandrina-p: only title-1 and title-2 are responsive headings. 3, 4 and 5 are the same everywhere.

frontend/utils/breakpoints.js
taoeffect: I have no idea what you mean here:

> but then, in order for us to do a reusable responsive resize, I believe we would need to use the store or sbp somehow...

sandrina-p: Read [this comment](https://github.com/okTurtles/group-income-simple/pull/777/files#diff-c3d9b27ac7e3a8165f584c6325634d49R115-R121). It might help you to understand what I mean

sandrina-p: Update: Direct quote from my changes

```js
  mounted () {
    // TODO - Create a single resize listener to be reused on components
    window.addEventListener('resize', this.config.debounceResize)
  },
  beforeDestroy () {
    window.removeEventListener('resize', this.config.debounceResize)
  },
```

frontend/views/components/Contribution.vue
```diff
@@ -51,7 +51,7 @@ transition(name='replacelist')
       @click='handleEditClick'
       data-test='buttonEditNonMonetaryContribution'
     )
-      i.icon-pencil-alt(aria-hidden='true')
+      i.icon-pencil-alt.is-prefix
```
taoeffect, line 54: You meant to remove `(aria-hidden='true')` here?

frontend/utils/breakpoints.js
taoeffect: I still don't see what this has to do with the store or sbp or what the model folder has to do with it?

frontend/views/components/Contribution.vue
sandrina-p: Yes, by default the icons aren't read by SR, I tested it myself

frontend/utils/breakpoints.js
sandrina-p: For now, nothing. But in the future, we'll need to implement a single resize listener that components would use and the way to do that is by using the store. But if you want I can move it to views/utils and later move it again.

taoeffect: > But if you want I can move it to views/utils and later move it again.

Yeah, that would be helpful I think for now. I'll go ahead and test out this PR right now while waiting for this change -- everything else looks good to me so far!

sandrina-p: done

frontend/views/pages/Page.vue
```diff
@@ -41,14 +45,37 @@ export default {
...
+    isInert () {
+      return !this.ephemeral.isActive && this.ephemeral.isTouch
+    }
+  },
   methods: {
     toggleMenu () {
       this.ephemeral.isActive = !this.ephemeral.isActive
+    },
+    checkIsTouch () {
+      console.log('chingg')
```
taoeffect: Should be removed

sandrina-p: removed

taoeffect (approved):
Approved! Great job @sandrina-p! 💯 

I'll go ahead and merge once Travis passes 😄

Issue #658, opened 2019-09-09
Add a profile picture editor and use it upon profile picture upload


taoeffect:
### Problem

For uploading both the group profile image and the user profile image, there are multiple problems:

- Pictures that are too large, wasting bandwidth (e.g. a 2MB picture can be uploaded, when at most the user will only ever see a tiny version of it)
- Pictures can be oblong in shape, resulting in oval user pictures

### Solution

- Find existing open source solutions to this problem (e.g. [here's an example](https://github.com/beakerbrowser/fritter/blob/master/com/crop-popup.js)) that use Canvas to edit the image
- Limit all profile pictures to some maximum Height x Width (e.g. 512x512)
- Fill in any gaps created by the scaling/resizing of the image
- Make sure the profile picture editor allows the user to shrink/enlarge and then crop the image to the section of the picture they uploaded, and then generate a new picture that will then be uploaded from that. This process is described in more detail below:

The end result is that all profile pictures (both group and user pictures) should be set to some constant values predefined in the app (e.g. let's say 512x512).

The point of the profile picture editor is to take as input [USER PICTURE] and produce as output => [CROPPED AND SCALED USER PICTURE OF SIZE 512x512], that is then the final picture that is actually uploaded to the server.

The profile picture editor needs to allow the user to both scale and move the picture, so that they can pick the section of the photo that they want to be their profile picture. It should also allow them to maybe select a background color to fill in any gaps (if there are any) that are created as a result of moving+scaling the photo in such a way that its edges are visible in the final cropped section.

EDIT: there is a related issue #1132 - which is necessary to close before implementing this issue. It's just about scaling the images - not about the UI for a profile picture editor.

mmbotelho:
I created mockups to solve this issue. Please [take a look at them here](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=6228%3A54859).

sandrina-p:
When doing this, we should also try to solve the issue #771

mmbotelho:
I'm removing myself from this issue, it's ready to be picked up by either @sandrina-p or @pieer !

SebinSong:
Hi @taoeffect,
Since I started working on this issue last week and have made a good progress so far, but as some requirements of this task involves relatively advanced work to be done, thought it would help if we split the work into two chunks and below is my suggestions for it. :)

**Chunk 1 ] Create the editor modal and add zoom in/out functionality.**

[...]

Issue #595, opened 2019-07-23
Save all console messages to a log and create a page in the settings where it can be viewed


taoeffect:
### Problem

Sometimes `console.log/debug/warn` messages put extremely important debugging information into the console that can be critical for recovering from errors. Asking users to open up the dev tools to troubleshoot a problem is a no-go.

### Solution

- We need to make it simple for users to send us the output of the console log, and also simple to adjust what information gets placed there while being careful not to log things like passwords or private keys
- We need to create a page in the UI itself to display this information
- This log needs to be persisted and save information from `.warn` and `.error` messages, because they might log important information necessary for recovery

NOTE: this issue is about the same as #353...

dotmacro:
Note: This issue is related to #353

sandrina-p:
@taoeffect should we do this for the prototype? I think so... maybe I can explore solutions for this.

You say the log should be persisted. Maybe for the prototype, a simple solution is to save it on localStorage, without the need/cost to save it on the server. What do you think?

taoeffect:
> @taoeffect should we do this for the prototype? I think so... maybe I can explore solutions for this.

Yes, this page definitely needs to be created for the prototype. A simple design is needed first from either you or @mmbotelho (or both, i.e. feel free to create one using basic components as you see fit and then make adjusts per @mmbotelho's designs).

It should also be related / tied-to the recovery page (#761) somehow. I.e. links to both of them should be found in the same section of the app, and each should link to the other.

> You say the log should be persisted. Maybe for the prototype, a simple solution is to save it on localStorage, without the need/cost to save it on the server. What do you think?

Yes, localStorage definitely, logs are never sent to the server.

sandrina-p:
Okay, I addressed this issue to me and @mmbotelho. While I wait for her work on this I'll research the technical side of it.

mmbotelho:
@sandrina-p can you help me understand this better? I have a few questions:

1. What do these messages look like? Can you give me some examples?
2. Can you give me some examples of when these might happen? 
3. What can users do with this information? (maybe @taoeffect can answer this one) Open a Github issue? Contact us? Do nothing at all unless we ask them to send them our way?  

Thank you!

sandrina-p:
@mmbotelho

[...]

PR #779, opened 2019-11-18
Implement Feedback Banners

### LINKED ISSUE #688: Error, success and app error banners [Discussion]

### LINKED ISSUE #689: Create self-removing component for scoped banner messages
### Problem

The **Type 2** "Context-Sensitive Scoped Success and Error Banners" (or just, "scoped banners") component described in #688 doesn't exist and needs to be designed and implemented.

Moreover, per the discussion in that issue, it needs to be implemented in a very specific way in order to satisfy our DRY **requirement** (which really needs to be part of our JavaScript Style Guidelines...).

### Solution

As shown in the examples in [this comment](https://github.com/okTurtles/group-income-simple/issues/688#issuecomment-536091742), currently the way these sorts of messages are implemented in our UI is through the use of `v-if` directives and corresponding variables.

This is a very ugly violation of DRY. So the goal for this issue is to see if it's possible (and I believe it should be), to create a component that does not require the use of `v-if` directives and variables.

Quoting from [here](https://github.com/okTurtles/group-income-simple/issues/688#issuecomment-536098776):

> I.e. something like this:
> 
> ```
> insertError(<target ref>, {message: 'error message', icon: 'cross'})
> ```
> And for that to insert a component into `<target ref>`, that itself contains all the logic necessary for automatically removing itself from the parent component

In other words, what is needed to close this issue is:

1. Implementing a component for Type 2 errors (i.e. the markup and CSS)
2. For this component to take props specifying what type of message it's displaying (i.e. an error or a success message, the message itself, and how the message should remove itself from the parent component<sup>[1]</sup>)
2. For a function similar to the `insertError` function above to be implemented (perhaps `insertScopedBanner()`), making it possible to use **one line** to insert this banner inside of the node specified by `<target ref>`. Again: to respect DRY, the parent component must not use `v-if` logic to show/hide this banner

<sup>[1]</sup> Possible values include: automatically removing itself after a given amount of time, or when the X to be clicked, or because another call to the same `<target ref>` has been made with `insertScopedBanner()`, causing the previous banner to be replaced. Barring either of those outcomes, it should always remove itself when the parent component is unloaded (i.e. if we navigate to a different page, or close the containing modal, etc., so that when we return to the parent view, the errors/success messages from before are all cleared)

If any of this is confusing or unclear, please ask questions!

sandrina-p:
Close #688. Closes #689

#### Banner type 2
demo
- Create the banner as `BannerScoped.vue` component.
- Apply that banner on the forms: login/signup, create group, user profile and income details. To avoid a huge PR/CR, I think is better to leave the implementation on other forms to other PR, as an ongoing task.
- Add `BannerScoped` to the Design System.

Note: On user profile, when saving the avatar, I just implemented a temporary banner for it, the final design should be addressed at #658.

#### Banner type 3
image

- Created the banner as component `BannerGeneral.vue`.
- Note: To see this banner you need to manually trigger it at `main.js`. Look for "_Trying to connect..._" and uncomment that function.

#### Misc
- Design tweaks around. User Settings modal and Contributions.
- Split User Profile form into 2 (one for avatar and another for the rest of attributes)
- Add "Feedback Message Style Guide" to our docs. @mmbotelho please take a look at it too!
- Create "TransitionExpand" component. A component that expands smoothly any component, no matter its height. Already applied to BannerScoped and BannerGeneral components and to IncomeDetails 2nd field (when selecting a pledging option).
- Minor tweaks/fixes on the Contributions page.
- Rename `Message` to `BannerSimple`


frontend/views/components/FeedbackBanner.vue
```diff
@@ -0,0 +1,49 @@
+<template lang='pug'>
+  message.c-container(v-if='message' :severity='severity')
```
taoeffect: Maybe here we could have the `v-if` based on whether or not the 'x' has been clicked after the banner has been told to display itself. I'm not sure if that's the right approach or not, but it could be as simple as that.

sandrina-p: hum... I think I didn't understand your suggestion.
The way it's done, the banner shows itself when there's a message passed as prop. when clicking the X, it cleans the message. I think that's a clean approach too.
When using the banner, there's no need for ifs, we just do something like:

```js
// js condition where something went wrong...
this.data.form.submitMessage 'Ups, something went wrong!';
```

```pug
// add to the specific form banner place
feedback-banner(:message="data.form.submitMessage")
```

What do you think? Can we go with this solution for the other cases?

taoeffect: Well, if you do it that way you can again use the `v-if` to check if the message is non-null, and set it to null when the 'x' is clicked.

Using `v-if` to remove the DOM node seems like a good idea. I see no reason to have a DOM node if it's not displaying anything. By removing the DOM node, it also removes all the empty space surrounding the message if any padding or margins are ever used.

taoeffect: However, it should be designed to detect message updates, both to the text and to the message type, so you can switch from error to success.

frontend/views/containers/forms/FormLogin.vue
```diff
@@ -50,7 +51,17 @@ export default {
...
-    FormPassword
+    FormPassword,
+    FeedbackBanner
+  },
+  data () {
+    return {
+      form: {
+        name: null,
+        password: null,
+        submitFeedback: {}
```
taoeffect: This shouldn't be in the `form`, since it's not a form field value. `ephemeral` is probably a better place for it

frontend/views/components/FeedbackBanner.vue
sandrina-p: > Well, if you do it that way you can again use the v-if to check if the message is non-null, and set it to null when the 'x' is clicked.

That's exactly what I've done... I'm not understanding if you agree with the current solution or not 🤔 

> However, it should be designed to detect message updates, both to the text and to the message type, so you can switch from error to success.

It's already designed that way, it accepts 2 props `message` (text) and `severity` (success, danger, etc...). By default severity is 'danger' (error).

taoeffect: > That's exactly what I've done... I'm not understanding if you agree with the current solution or not 🤔

I'm seeing you're doing `this.$emit('update:message', '')`, but I don't see a listener for that. If this is some bizarre Vue.js thing, then a comment needs to be added linking to the text describing how this works.

~~I would have expected `this.message = null`, which is more readable and clearer. Is there a reason you're not doing that?~~ I see, because it's a property as a String type, and because of [this](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier)

> It's already designed that way, it accepts 2 props message (text) and severity (success, danger, etc...). By default severity is 'danger' (error).

Hmm, I will post another comment below about this..

frontend/views/components/FeedbackBanner.vue
```diff
@@ -0,0 +1,49 @@
...
+    message: String,
+    severity: {
+      type: String,
+      default: 'danger',
+      validator: (v) => ['info', 'warning', 'danger'].includes(v)
+    }
+  },
+  methods: {
+    dismiss () {
+      this.$emit('update:message', '')
```
taoeffect: Can you add a comment here linking to https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier ?

frontend/views/components/FeedbackBanner.vue
sandrina-p: > I'm seeing you're doing `this.$emit('update:message', '')`, but I don't see a listener for that. If this is some bizarre Vue.js thing, then a comment needs to be added linking to the text describing how this works.

This is a recent Vue feature (one of my favorites), where we can do “two-way binding” for a prop called [event sync](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier). We are already using this in the project a couple of times.

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -167,10 +163,15 @@ export default {
...
     },
     async submit () {
-      if (this.$v.form.$invalid) {
-        alert(L('Invalid payment info'))
+      const debugThis = 1 + 1 // DELETE THIS BEFORE MERGE!!
+      if (debugThis || this.$v.form.$invalid) {
+        this.$set(this.form, 'submitFeedback', {
+          message: L('Please, review your information and try again.'),
+          severity: 'danger'
+        })
```
taoeffect: Must this use `this.$set`? Is there an issue with `this.form.submitFeedback = `?

frontend/views/components/FeedbackBanner.vue
taoeffect: OK, yes this approach looks promising. Once the other comments are addressed and the build is passing, let me know and I'll give it a test run locally.

Also, see if you can update the other files in the project to use this instead of how we were doing it before. A true test of whether this is a good PR is if it ends up removing lines of code for displaying success/error messages (the more lines of code removed, the better!)

taoeffect: @sandrina-p One way to remove lines of code is to avoid the creation of `submitFeedback` as a data variable on the parent. See if you can instead access the banner through a ref, as mentioned in #689, but with this new approach.

In other words, instead of:

```js
        this.$set(this.form, 'submitFeedback', {
          message: L('Please, review your information and try again.'),
          severity: 'danger'
        })
```

Something like:

```js
this.$refs.submitFeedback.danger(L('Please, review your information and try again.'))
```

And:

```js
this.$refs.submitFeedback.success(L('Profile picture updated!'))
```

sandrina-p: > Also, see if you can update the other files in the project to use this instead of how we were doing it before. 

I didn't do that yet because I wanted to wait for your approval for this POC first (that's why it's a WIP). Based on your comments it seems you approved the POC, so yes, I'll do that next and ask for your review after ;)

taoeffect: > Based on your comments it seems you approved the POC, so yes, I'll do that next and ask for your review after ;)

I think we posted at the same time. Please see the comment above yours :)

sandrina-p: Good idea, I'll try that and let you know if it worked!

frontend/views/containers/contributions/IncomeDetails.vue
taoeffect: Note: can ignore above comment if [this other approach](https://github.com/okTurtles/group-income-simple/pull/779/files#r348781980) works, since we won't need to have `submitFeedback` as a data variable on the parent

frontend/views/containers/forms/FormLogin.vue
taoeffect: Note: can ignore above comment if [this other approach](https://github.com/okTurtles/group-income-simple/pull/779/files#r348781980) works, since we won't need to have `submitFeedback` as a data variable on the parent

docs/Style-Guide.md
```diff
@@ -68,6 +69,33 @@ When developing any new feature, make sure to add or update the respective tests
...
+When the user performs an action (ex: submiting a form), it's expected to show them some type of feedback, when succeedded or when something went wrong. Here's a guide of the current type of messages we show to them:
+
+### Informative / Success messages
+
+> Your changes were saved!  
+
+### Error messages
+
+#### While filling a form...
+The errors are displayed inline, while the user is typing (with a delay) or on input's blur. Here're some examples:
```
taoeffect: Remember to put an empty line after headers

package.json
```diff
@@ -138,6 +138,7 @@
     "cheerio": "1.0.0-rc.3",
     "chokidar": "3.2.3",
     "cypress": "3.5.0",
+    "cypress-file-upload": "3.5.0",
```
sandrina-p, line 141: Note: Doing it in a "native way" is quite a nightmare, everyone was requesting a easier way. I was waiting for this package since I learned Cypress. Finally, it got released 2 weeks ago! [cypress-file-upload](https://www.npmjs.com/package/cypress-file-upload)

cc @taoeffect

docs/Style-Guide.md
```diff
@@ -68,6 +69,33 @@ When developing any new feature, make sure to add or update the respective tests
 
 When writing E2E tests, follow [Cypress Best Practices](https://docs.cypress.io/guides/references/best-practices.html).
 
+
+## Feedback Messages
+
+When the user performs an action (ex: submiting a form), it's expected to show them some type of feedback, when succeedded or when something went wrong. Here's a guide of the current type of messages we show to them:
```
taoeffect: "when succeedded or when something went wrong." => "on success or failure"

docs/Style-Guide.md
```diff
@@ -68,6 +69,33 @@ When developing any new feature, make sure to add or update the respective tests
...
+
+### Informative / Success messages
+
+> Your changes were saved!  
+
+### Error messages
+
+#### While filling a form...
```
taoeffect: I don't find these sections to be very helpful personally. I think it would be more useful to newcomers to simply show them the API you've designed, with a few examples from the actual code, in two sections (one for scoped banner usage, and one for global banner usage).

taoeffect (changes requested):
My review isn't finished, but I've posted 3 comments on the Style-Guide.md changes in case you'd like to work on them while I continue my review

frontend/main.js
```diff
@@ -112,6 +114,11 @@ async function startApp () {
         this.ephemeral.finishedLogin = 'no'
         router.currentRoute.path !== '/' && router.push({ path: '/' }).catch(console.error)
       })
+
+      // TODO - Ready to receive real information
+      setTimeout(() => {
+        this.$refs.bannerGeneral.show('Trying to connect...', 'globe')
+      }, 2500)
```
taoeffect: Why is this uncommented? (Also missing `L`)

frontend/views/components/BannerSimple.vue
```diff
@@ -20,14 +20,15 @@ export default {
...
+    return ['success', 'info', 'warning', 'danger'].indexOf(value) !== -1
   },
   computed: {
     getIcon () {
       return {
-        warning: 'icon-exclamation-triangle',
-        danger: 'icon-times-circle',
-        info: 'icon-info-circle'
+        warning: 'icon-exclamation-triangle c-icon',
+        danger: 'icon-times-circle c-icon',
```
taoeffect, line 28: IMO, there's a slight visual issue with having the X icon being used both on the left and on the right for warning banners. It might make it unclear to users which is the close button.

Perhaps @mmbotelho can offer some advice here? There's an animation that @sandrina-p recorded in the PR description showing what I'm talking about.

frontend/views/components/BannerGeneral.vue
```diff
@@ -0,0 +1,34 @@
...
+  },
+  data: () => ({
+    ephemeral: {
+      message: null,
+      icon: null
+    }
+  }),
+  methods: {
+    // To be used by parent. Example:
+    // this.$refs.bannerGeneral.show('Trying to reconnect...', 'wireless')
```
taoeffect: Comments should show proper `L` function usage too

frontend/views/components/BannerScoped.vue
```diff
@@ -0,0 +1,106 @@
...
+  },
+  data: () => ({
+    ephemeral: {
+      text: null,
+      severity: null
+    }
+  }),
+  methods: {
+    // To be used by parent. Example:
+    // this.$refs.BannerScoped.danger('ups!')
```
taoeffect: Comments should show proper `L` function usage too

frontend/views/components/BannerScoped.vue
```diff
@@ -0,0 +1,106 @@
...
+  name: 'BannerScoped',
+  components: {
+    BannerSimple,
+    TransitionExpand
+  },
+  props: {
+    dataTest: {
+      type: String,
+      default: 'feedbackMsg'
+    }
```
taoeffect, line 30: Perhaps good to pass in all the attributes [using `$attrs`](https://vuejs.org/v2/api/#vm-attrs) instead, so that `data-test` can be used directly on this component just the same way it is used everywhere else, and new developers don't have to wonder, "Why `dataTest` here and `data-test` there?"

taoeffect (commented):
5 more comments, still not done...

frontend/views/components/BannerSimple.vue
taoeffect: But perhaps it's right to have the "X" here when it's used a "banner-simple" on its own (as in Group Rules), since there no additional "X" is added... so it seems almost like a conflict between this file and the other one above

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -184,10 +182,7 @@ export default {
         this.closeModal()
       } catch (e) {
         console.error('setPaymentInfo', e)
-        this.ephemeral.formFeedbackMsg = {
-          text: L(`Failed to update income details. Error: ${e.message}`),
-          class: 'has-text-danger'
-        }
+        this.$refs.formMsg.danger(`${L('Something went wrong, please try again.')} ${e.message}`)
```
taoeffect: Much cleaner 😄

frontend/views/containers/modals/LeaveGroup.vue
```diff
@@ -10,7 +10,7 @@
 
       i18n(tag='p' html='If you leave, you will stop having access to the <b>group chat</b> and <b>contributions</b>. Re-joining the group is possible, but requires other members to <b>vote and reach an agreement</b>.')
 
-      message(severity='danger')
+      banner-simple(severity='danger')
         i18n(tag='p' html=' This action <b>cannot be undone</b>.')
```
taoeffect: Boyscoutting: think you can update these i18n tags to use the [new i18n LTags w/args behavior](https://github.com/okTurtles/group-income-simple/blob/master/frontend/views/utils/translations.js#L20-L27) since `html=` [is deprecated](https://github.com/okTurtles/group-income-simple/blob/master/frontend/views/utils/translations.js#L87-L90)?

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -115,43 +94,46 @@
...
   validations: {
     form: {
-      picture: {},
-      email: { email }
+      email: {
+        [L('An email is required.')]: required,
+        [L('Please enter a valid email.')]: email,
+        [L('This email is already being used.')]: value => {
+          // TODO - verify if e-mail exists
+          return true
```
taoeffect: This will never be done, so this check can be removed

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -170,43 +152,47 @@ export default {
...
+      const fileReceived = fileList[0]
+      let picture
+
+      this.$refs.picture.setFromBlob(fileReceived)
+
+      try {
+        picture = await imageUpload(fileReceived)
+      } catch (error) {
+        console.error(error)
+        this.$refs.pictureMsg.danger(`${L('Failed to upload avatar. Please, try again.')} ${error.message}`)
```
taoeffect: `${L('Failed to upload avatar. Please, try again.')} ${error.message}` => `L('Error uploading avatar: {errorMsg}', { errorMsg: error.message })`

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -170,43 +152,47 @@ export default {
...
+        return false
+      }
+
+      await this.setAttributes({ picture }, 'pictureMsg', L('Picture updated!'))
```
taoeffect: If this fails we probably want to reset the picture in the UI to whatever it was before so that it doesn't look like the upload succeeded.

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -170,43 +152,47 @@ export default {
...
+        this.$refs[refFeedback].success(successMsg)
+      } catch (error) {
+        console.error(error)
+        this.$refs[refFeedback].danger(`${L('Something went wrong, please try again.')} ${error.message}`)
       }
-    },
-    fileChange (fileList) {
-      if (!fileList.length) return
-      this.form.picture = fileList[0]
-      this.ephemeral.newPicture = true
```
taoeffect, line 208: This `newPicture` variable was removed, and it was useful in preventing the same file from being uploaded twice accidentally by a user clicking the upload button twice. Perhaps it's no longer needed? Just double-checking with you to be sure.

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -170,43 +152,47 @@ export default {
...
         await sbp('backend/publishLogEntry', attributes)
-        this.ephemeral.profileSaved = true
-      } catch (ex) {
-        // TODO: handle better
-        console.error(ex)
-        this.ephemeral.errorMsg = L('Failed to Save Profile')
+        this.$refs[refFeedback].success(successMsg)
+      } catch (error) {
+        console.error(error)
+        this.$refs[refFeedback].danger(`${L('Something went wrong, please try again.')} ${error.message}`)
```
taoeffect: `${L('Something went wrong, please try again.')} ${error.message}` => `L('Error saving profile: {errorMsg}', { errorMsg: error.message })`

frontend/views/pages/DesignSystem.vue
```diff
@@ -541,6 +551,25 @@ page(
...
+              | button(
+              |   @click='$refs.banner.danger("Ups!")'
+              | ) Save
+
+          td
+            div
+              banner-scoped(ref='banner')
+              .buttons.is-centered
+                button.is-small.is-outlined(@click='$refs.banner.success("Changes saved!")') Save Success
+                button.is-small.is-outlined(@click='$refs.banner.danger("Something went wrong, please try again.")') Save Danger
```
taoeffect: Nice 👍

taoeffect (changes requested):
Code review finished! Great job @sandrina-p, the new scoped banner code looks excellent, and will save us a lot of time and pain! 🎉 😄

After the issues are addressed I'll give it a test run through. (Please remember to do at least one non-Cypress test run yourself with multiple container tabs!)

frontend/views/components/TransitionExpand.vue
```diff
@@ -0,0 +1,81 @@
+<script>
+// From https://markus.oberlehner.net/blog/transition-to-height-auto-with-vue/
```
taoeffect, line 2: Great find and great job leaving credit! 👍

frontend/views/containers/contributions/IncomeDetails.vue
taoeffect: Though here and in a [few](https://github.com/okTurtles/group-income-simple/pull/779/files#diff-6cb8ae3b98ed4acd79c5dc80d6050ef6R85) [other](https://github.com/okTurtles/group-income-simple/pull/779/files#diff-e453413c9b9dbd7f7486e41030530bafR130) [places](https://github.com/okTurtles/group-income-simple/pull/779/files#diff-459e0a543f1a07208ec740ba89604eb2R165), can you change the way this text is generated to be [more descriptive and using the L-function variables](https://github.com/okTurtles/group-income-simple/pull/779/files#r349846015)?

I.e. here we can do: `${L('Something went wrong, please try again.')} ${e.message}` => `L('Error updating payment info: {message}, { message: e.message })`

This will make it easier for translators to move the error message if necessary.

frontend/assets/style/_layout.scss
```diff
@@ -1,8 +1,10 @@
+$bannerGeneralHeight: 2.3rem;
+
 .l-no-navigation {
   display: flex;
   justify-content: center;
   align-items: flex-start;
-  height: 100%;
+  height: 100vh;
```
pieer, line 5: Probably good idea to check that all the height 100vh work well on mobile because the viewport is jumpy with the search bar. (in some case where there is a lot of space around, it's ok) Same comment apply everywhere where you calcul the space with the vertical height available. If we can't test now, we should add a reminder to double check all those properties

frontend/assets/style/_transitions.scss
```diff
@@ -133,21 +133,6 @@
   }
 }
 
-.expand-enter-active,
```
pieer, line 136: Well spotted

pieer (commented):
What a PR! Nice work @sandrina-p. Good cleaning and good call on creating transition Expend. I couldn't find a good solution for dynamic height using only CSS like @taoeffect requested few months ago. I'm happy that you take that decision. My only concern is mainly around the use of vh that is never really good on mobile ;)

frontend/views/components/BannerSimple.vue
sandrina-p: @mmbotelho can you review this?

frontend/views/components/BannerScoped.vue
sandrina-p: I don't agree with $attrs in this case because the data-test is not appended to the parent level but instead on of the childrens. Nevertheless, people can write `data-test` when using the component because [props interpret any uppercase characters as lowercase ](https://vuejs.org/v2/guide/components-props.html#Prop-Casing-camelCase-vs-kebab-case).

frontend/views/containers/contributions/IncomeDetails.vue
sandrina-p: I did `Something went wrong, please try again.` following @mmbotelho designs. Maybe she can review it. However, I think both messages mean exactly the same thing to the user - _nothing in concrete_. :/ 

I don't see the advantage of placing `e.message` inside the translation because it has nothing to do with the copy itself.

> This will make it easier for translators to move the error message if necessary.

Trying to predict the future...  Anyway, I highly doubt that the code error will be placed somewhere else. 100% of the places I saw this around the web, is the last part of an error message, after the "human message". When we need to do that (if needed), we change the code. Until then I would leave it as it is, simple.

frontend/views/containers/settings/UserProfile.vue
sandrina-p: cc @mmbotelho  she asked this same question a couple of weeks ago.

frontend/views/containers/settings/UserProfile.vue
sandrina-p: Once again, I think @mmbotelho has a word on this :)

frontend/views/containers/settings/UserProfile.vue
sandrina-p: Another approach: change the picture in the UI only if this succeeds.

frontend/views/containers/settings/UserProfile.vue
sandrina-p: > being uploaded twice accidentally by a user clicking the upload button twice. 

How can the user do that twice? 🤔 I think it doesn't make sense anymore.

frontend/views/containers/settings/UserProfile.vue
sandrina-p: cc @mmbotelho

frontend/assets/style/_layout.scss
sandrina-p: Yes, 100vh is a tricky subject, we'll need to review it after we can successfully load GI on a real mobile phone. This won't be forgot for sure.

frontend/views/components/BannerSimple.vue
mmbotelho: I would leave it as it is. It is common across the web, it's ubiquitous and also consistent with other close buttons we have in GI. And besides that, closing the banner is not a critical feature - meaning that it will not prevent or make an action harder to complete.

frontend/views/containers/settings/UserProfile.vue
mmbotelho: I don't understand what you need from me here - can you clarify?

frontend/views/components/BannerScoped.vue
taoeffect: Well in that case can you update the code to write `data-test` consistently everywhere?

frontend/views/containers/contributions/IncomeDetails.vue
taoeffect: > Trying to predict the future...

My job is to predict, and also invent, the future. :)

I think this change needs to be made for the following reasons:

1. If it's not made, we will have inconsistencies in how we translate strings. The code should be consistent. Having a consistent way of translating strings is important, because it helps prevent confusion and makes it easier to onboard new developers.
2. The existing way we have of placing variables into strings is more correct than what's being done here, because it gives our translators more flexibility
3. As far as copy goes, the copy suggested above is better than what was there before. "Something went wrong" is a bad error message, because it gives no useful information to the users, and no useful information to us.

frontend/views/containers/settings/UserProfile.vue
taoeffect: @mmbotelho is welcome to comment, but my [response](https://github.com/okTurtles/group-income-simple/pull/779#discussion_r350262715) remains the same.

frontend/views/components/BannerScoped.vue
sandrina-p: I already did that :)

frontend/assets/style/_layout.scss
taoeffect: So, for a reminder, will a comment do? Or is it worth opening an issue as well?

frontend/views/containers/modals/LeaveGroup.vue
```diff
@@ -10,8 +10,10 @@
 
       i18n(tag='p' html='If you leave, you will stop having access to the <b>group chat</b> and <b>contributions</b>. Re-joining the group is possible, but requires other members to <b>vote and reach an agreement</b>.')
```
taoeffect: Boyscoutting for `<b>` here too? :)

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -2,81 +2,62 @@
...
             v-model='form.bio'
             placeholder='Bio'
             data-test='bio'
           )
 
-        .field
-          i18n.label(tag='label') Email
-
+        label.field
+          i18n.label Email
```
taoeffect, line 48: So, just noticing that the markup here has changed significantly:

- Line 47: a `<div>` has turned into a `<label>`
- Line 48: a `<label>` has turned into a `<span>`

And same elsewhere. You're certain this was intentional?

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -160,53 +138,59 @@ export default {
...
-        console.error(ex)
-        this.ephemeral.errorMsg = L('Failed to Save Profile')
+      }
+
+      try {
+        await this.setAttributes(attrs)
+        this.$refs.formMsg.success(L('Your changes were saved!'))
+      } catch (error) {
+        console.error(error)
+        this.$refs.formMsg.danger(`${L('Something went wrong, please try again.')} ${error.message}`)
```
taoeffect: `${L('Something went wrong, please try again.')} ${error.message}` => `L('Failed to update profile: {error}, { error: error.message })`

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -160,53 +138,59 @@ export default {
...
-        for (const key in this.form) {
-          if (this.form[key] !== this.attributes[key]) {
-            attrs[key] = this.form[key]
-          }
+        await this.setAttributes({ picture })
+        this.$refs.picture.setFromBlob(fileReceived)
+        this.$refs.pictureMsg.success(L('Picture updated!'))
+      } catch (error) {
+        console.error(error)
+        this.$refs.pictureMsg.danger(`${L('Failed to upload picture.')} ${error.message}`)
```
taoeffect: `${L('Failed to upload picture.')} ${error.message}` => `L('Failed to upload picture: {error}', { error: error.message })`

docs/Style-Guide.md
```diff
@@ -68,6 +69,93 @@ When developing any new feature, make sure to add or update the respective tests
...
+<banner-general ref="bannerGeneral"></banner-general>
+```
+
+```js
+// main.js
+import BannerGeneral from './views/components/BannerGeneral.vue'
+
+this.$refs.bannerGeneral.show(L('Trying to reconnect...'), 'wifi')
+```
+
```
taoeffect, line 161: Excellent work! This is very helpful now. :-)

I would make only two more suggestions for now:

1. Add an empty newline after line 73 for consistent headings formatting
2. Add a "NOTE: this section is under construction and will change soon" message somewhere here in the "Type 3" section, since we'll need to add a (most likely an `sbp`-based) API to update the global banner

taoeffect (changes requested):
Review finished! Small things remaining, and I think this should be GTM once they're addressed — *if* my manual testing goes well. 😄

frontend/views/containers/settings/UserProfile.vue
mmbotelho: Can you give me examples of{errorMsg} @taoeffect @sandrina-p ?

Will the user need to know the specific error? Or can we keep it simple?

frontend/views/containers/settings/UserProfile.vue
mmbotelho: Yes, I would say that the picture should only be updated if the upload was successful :)

frontend/views/containers/settings/UserProfile.vue
mmbotelho: Ah I understand now. I have the [same question](https://github.com/okTurtles/group-income-simple/pull/779#discussion_r350760545) - can you please give me examples of what this would be? And again, do the users need to know the exact error or can we keep it simple?

mmbotelho (changes requested):
Captura de ecrã 2019-11-26, às 16 37 18
Captura de ecrã 2019-11-26, às 16 47 14
Captura de ecrã 2019-11-26, às 16 47 36
Captura de ecrã 2019-11-26, às 16 48 08
Captura de ecrã 2019-11-26, às 16 48 54
Captura de ecrã 2019-11-26, às 16 50 16
Captura de ecrã 2019-11-26, às 16 37 18

Others:
- The "trying to reconnect" banner should be visible when the user is on a modal as well;
- The component has the wrong text color - the orange, warning banner, does not have the final text color yet, but all the others do!
- The component is misaligned vertically (icon + text + button) - this will probably require you to do some "nasty tweaks" like pushing things 1px down or up, but it's necessary, or else it looks very janky!

mmbotelho:
As for the actual error copy discussed in this thread by @sandrina-p and @taoeffect : 

I don't think we should show these "robot-like" errors to users (e.g.: "
Failed to upload avatar. Please, try again. Cannot set property 'bar' of undefined") - they have absolutely no advantage in seeing these. I think they should remain in a more "technical" place like the console, for example?
Not all humans speak computer and this seems more like debugging copy than anything else!

mmbotelho:
Found another one @sandrina-p: 
Captura de ecrã 2019-11-26, às 16 59 51

frontend/views/containers/settings/UserProfile.vue
sandrina-p: errorMsg is literally the code error, for example "Error 500 - not found" or "can't find 'expired' of undefined" or any other JS internal error that can happen.

taoeffect: That's good to include in the message. It will help the user help us help them.

So I am for including `{errorMsg}`. If both of you are against it, then please replace `{errorMsg}` with at least: `Please check the {link_}error logs{_link} for more info.` Where `{link_}` is used to go to the debug page mentioned in issue #595 (that doesn't exist yet)

frontend/assets/style/_layout.scss
sandrina-p: In our minds is enough. This will be the first we'll notice when we got GI running on a mobile device

frontend/views/containers/settings/UserProfile.vue
sandrina-p: Yes, this is the most correct and simple way of coding form fields with a11y in mind. Every field should be surrounded by a `label` tag. Inside should have the label itself (ex: Email), the `input` and any other helpful text related to the field, ex: a helper or error message.
By having all this content inside the label, the SC will know that it's all related to the same thing, without the need to use `for` with unique `id`.

sandrina-p:
@greg and @mmbotelho all changes requested made.

@mmbotelho, I didn't plan to modify/improve the Group Settings page, so I ignored the bugs you found there because they were already there before (in master). To keep the CR simpler, I would prefer to address them in another PR, is that okay?

Regarding the copy, I reviewed all error messages using the scoped banner and modified them to the following: 
> _"Failed to [action description], please try again. {codeErrror}"_

@mmbotelho, to be easier for you to review the copy, please [see this commit](https://github.com/okTurtles/group-income-simple/pull/779/commits/2aa547429c70e3732dc27087be24a5b3cb7fd806).

mmbotelho:
Just reviewed it again:
Captura de ecrã 2019-11-29, às 16 06 42
Captura de ecrã 2019-11-29, às 16 08 20

Another one:
- The "trying to reconnect" banner is still not visible on modals

Other than that, everything is perfect!

sandrina-p:
@mmbotelho humm it seems correct here, are u sure u pulled the last version?
image

Note: If you are referring to when the feedback banner appears, I said this would be just a provisory solution until we implement #658, because that banner wouldn't exist then. But since you insisted on it and we don't have plans to implement it soon, I fixed it for you :D
image

- The general banner (trying to reconnect...) is now on top of everything, including modals.
image
Note that the "default" modal on phone/tablet moves a few pixels down to give space to the banner. However, that doesn't happen with fullscreen modals (ex: user settings) because of technical reasons (basically that modal variant isn't a component yet?), so we would need to fix each fullscreen modal manually. Please forgive us on that part and I'll make sure to address that on a next PR (I won't forget it, it's on my TODO list now)

frontend/views/containers/modals/LeaveGroup.vue
```diff
@@ -8,10 +8,15 @@
     )
       template(slot='title') {{ L('Leave a group') }}
 
-      i18n(tag='p' html='If you leave, you will stop having access to the <b>group chat</b> and <b>contributions</b>. Re-joining the group is possible, but requires other members to <b>vote and reach an agreement</b>.')
+      i18n(
+        tag='p'
+        :args='LTags("strong")'
+      ) 'If you leave, you will stop having access to the {strong_}group chat{_strong} and {strong_}contributions{_strong}. Re-joining the group is possible, but requires other members to {strong_}vote and reach an agreement{_strong}.')
```
taoeffect: `'If` => `If`
`{_strong}.')` => `{_strong}.`

frontend/views/containers/settings/UserProfile.vue
```diff
@@ -1,82 +1,63 @@
...
+          input.sr-only(
             type='file'
             name='profilePicture'
             accept='image/*'
             @change='fileChange($event.target.files)'
             placeholder='http://'
             data-test='profilePicture'
           )
+        // TODO #658
+        banner-scoped(ref='pictureMsg' data-test='pictureMsg' class='c-pictureMsg')
```
taoeffect: Usually instead of `class='c-pictureMsg'` the codebase does `banner-scoped.c-pictureMsg`. Is there a reason that was not done here? If so, can it be documented in a comment to explain why?

Also, out of curiosity, which tag is  the `c-pictureMsg` applied to inside of `BannerScoped.vue`?

frontend/views/containers/settings/UserProfile.vue
taoeffect: This appears to have been missed, please double-check that all error messages and calls to `.danger` are consistent and don't use `${` stuff

frontend/views/containers/settings/UserProfile.vue
sandrina-p: changed to `banner-scoped.c-pictureMsg`, it was a typo.

When passing a class on a component, that class is _always_ automatically appended to the parent element of that component, unless explicit said otherwise using `$attrs`.

frontend/views/containers/modals/LeaveGroup.vue
sandrina-p: done

taoeffect (dismissed):
Approved! Just about everything looks good and works well! 🎉 

The only thing I noticed was that the gap between the "Change avatar" link and the box below on mobile seems too small:

Screen Shot 2019-11-29 at 6 50 44 PM

So while waiting for @mmbotelho's approval maybe you can fix that?

mmbotelho (changes requested):
> Note that the "default" modal on phone/tablet moves a few pixels down to give space to the banner. However, that doesn't happen with fullscreen modals (ex: user settings) because of technical reasons (basically that modal variant isn't a component yet?), so we would need to fix each fullscreen modal manually. Please forgive us on that part and I'll make sure to address that on a next PR (I won't forget it, it's on my TODO list now)

Ok, no worries! It cannot stay like this because it covers the X button on desktop, but feel free to fix it in a different PR 😄 

Captura de ecrã 2019-11-30, às 17 30 50

Last one! ☝️

sandrina-p:
@mmbotelho nice catch, fixed. Now it's left aligned no matter the context alignment (right or centered).

image

mmbotelho (dismissed):
Approved! Great job @sandrina-p !

taoeffect (approved):
(Marking an explicit approval so GH lets me merge)

Issue #556, opened 2019-04-19
Implement the "Income Details" modal as a fullscreen modal


mmbotelho:
### Problem

The current "Income Details" modal has a lot of information for a small-sized modal. For more context on the problem and the UI/UX solution, go to issue #536 .

### Solution

- Change the UI to make it more readable and better suited for a full-screen modal;
- The functionality should remain the same as before;
- Users will be able to open this modal from the **Contributions** page and the **Dashboard**.

[Here's link to the Figma mockups](https://www.figma.com/file/ZR0DRPhhUuHdJtf7PbGAM6tW/Group-Income?node-id=1922%3A0).

_**Note**: Because the current version is not 100% functional, I was not able to test all the possible scenarios and see if there was some use case missing. If there are, we can address it future issues._

dotmacro:
For reference, [here's a link to the redesigned Figma mockups](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1922%3A0).

mmbotelho:
Update on this - the modal has been implemented, except for the graphic UI. Once that is done this issue can be closed.

PR #786, opened 2019-11-26
Implement Income Details Graphic

sandrina-p:
Related to #556, implemented the missing graphic related to group pledging status.

Now, that it's connected with real data, there were some scenarios that I want to make sure they are okay visually:

1 - When nobody needs income, the group goal is 0$, so there's no blue bar and the surplus is always 100%:
image

2 - a scenario where all pledge is your full income, nothing weird, but just to check. A reminder: the group goal is the total needed pledge. in this case $100
image

3 - a scenario where surplus + your income fulfill 100% the graphic. nothing wrong, just to check:
image

4 - Now things get interesting: (🐛 bug ahead!) The surplus is bigger than the group goal ($30). Your mincome (orange) conflicts with the surplus. As it is now, the surplus on the graphic has a maximum percentage of 100%, this is, the group goal. Your mincome has the same value - 100% - as in the first scenario. Which one should appear first? Should they appear together? If yes, should we split their space accordingly (in this case 70% to surplus and 30% to income?)
image

Well... it's easier to understand all of this if you try the graphic by yourself. Just create a group with 2 members and that's all you need to see the previous scenarios.

### Misc stuff:
- Add prop `fullscreen` to the modal component and improve its UI (paddings around + missing ESC on the close button). Note: To keep PR small, I only applied this to the Income Details modal. In a next PR it should be applied to all the other fullscreen modals and [address the issue with general banner](https://github.com/okTurtles/group-income-simple/pull/779#issuecomment-559900573)
- Move from `group-proposals.spec.js` the tests related with payment to a new cypress spec `group-payments`. It's a raw copy-paste with a few more assertions related to the Income Details Graphic.


.Gruntfile.babel.js
```diff
@@ -337,7 +337,7 @@ module.exports = (grunt) => {
           match: /\.scss$/,
           recurse: true
         }),
-        eslint({ throwOnError: true, throwOnWarning: true }),
+        eslint({ throwOnError: true, throwOnWarning: true, fix: true }),
```
sandrina-p: NOTE: 👮‍♀ I won't forget to delete this before merging

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -1,79 +1,76 @@
 <template lang='pug'>
-modal-base-template(ref='modal')
-  .wrapper-container
-    .c-content
-      i18n.is-title-2.c-title(tag='h2') Income Details
+modal-base-template(ref='modal' :fullscreen='true')
+  .c-content
+    i18n.is-title-2.c-title(tag='h2') Income Details
```
sandrina-p, line 4: NOTE: I replaced this extra wrapper with the new modal prop fullscreen. The indentation changed! That's why this huge diff..

Just review the `group-pledges-graph` component, that is the real change.

test/cypress/integration/group-payments.spec.js
```diff
@@ -0,0 +1,250 @@
...
+  - needs $190
+  - $40 by pierre
+
+Graphic for user4:
+
+goal: 100 + 190 + userincomeneeded
+totalPledge: grouppledge + userpledge
+needed: total - goal
+surplus: goal - totalPledge
+*/
```
sandrina-p: Just a cheetsheat to better understand the numbers of these test journey.

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
+      const neededPledges = groupGoal - pledgeTotal
+      const surplus = pledgeTotal - othersIncomeNeeded - ourIncomeNeeded
+
+      let ourIncomeToReceive
+      if (neededPledges > 0) {
+        // TODO - get real income to receive based on distribution algorithm
+        const membersNeedingPledges = this.groupMembersCount - othersPledgesCount
+        const avgNeededPledges = neededPledges / membersNeedingPledges
+        const avgMembersNeedingPledges = groupGoal / membersNeedingPledges
+        ourIncomeToReceive = +(ourIncomeNeeded - (ourIncomeNeeded * avgNeededPledges / avgMembersNeedingPledges)).toFixed(0)
```
sandrina-p: Question: @taoeffect can u suggest a clean/easy way of using `IncomeDistribution` here without repeating the code in the getter `groupIncomeDistribution`?

mmbotelho:
> 4 - Now things get interesting: (🐛 bug ahead!) The surplus is bigger than the group goal ($30). Your mincome (orange) conflicts with the surplus. As it is now, the surplus on the graphic has a maximum percentage of 100%, this is, the group goal. Your mincome has the same value - 100% - as in the first scenario. Which one should appear first? Should they appear together? If yes, should we split their space accordingly (in this case 70% to surplus and 30% to income?)

The income to receive should be displayed first - it's more relevant to the user.

sandrina-p:
> The income to receive should be displayed first - it's more relevant to the user.

Okay, it makes sense. What about the next question: "Should they appear together?"

mmbotelho:
@sandrina-p from what I can see, the current design does not handle these edge cases very well... For our sanity, let's just show either one or the other (income to receive has priority over the surplus). 

We can come back later and iterate on the design and implementation 😄what do you think?

sandrina-p:
> For our sanity, let's just show either one or the other (income to receive has priority over the surplus).

So, we should show only income to receive, even if there's space for surplus as well? (scenario 3)

mmbotelho (changes requested):
Looking good @sandrina-p ! Here's some things I found:
Captura de ecrã 2019-11-30, às 18 28 48
Captura de ecrã 2019-11-30, às 18 24 58
Captura de ecrã 2019-11-30, às 18 27 14
Captura de ecrã 2019-11-30, às 18 36 34


-----


Probably not related with this PR: When a user has a $0 income, I get this "false" on the contributions page:
Captura de ecrã 2019-11-30, às 18 32 53

Captura de ecrã 2019-11-30, às 18 35 10

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
-      const userIncomeToReceive = userIncomeNeeded ? (userIncomeNeeded - (userIncomeNeeded * avgNeededPledges / avgMembersNeedingPledges)).toFixed(2) : 0
+      const mincome = this.groupSettings.mincomeAmount
+      const ourPledgeAmount = this.type === 'pledgeAmount' && this.amount >= 0 && this.amount
+      const ourIncomeAmount = this.type === 'incomeAmount' && this.amount < mincome && this.amount
+      const doWeNeedPledge = typeof ourIncomeAmount === 'number'
+      let othersIncomeNeeded = 0
+      let othersPledgesAmount = 0
+      let othersPledgesCount = 0
+
+      Object.keys(this.groupProfiles).forEach(username => {
```
taoeffect: I recommend changing this to a `for/in` loop, as they're faster and to match the style of code in `groupIncomeDistribution`, because this loop will be performing a similar function.

Line 85 will need to be updated (you can just wrap the lines below it in a `if (contractID !== this.ourContractId) { .. }`)

frontend/views/containers/GroupPledgesGraph.vue
taoeffect: I'm taking it easy today, but for when I get back, can you let me know what you mean by "using `IncomeDistribution`"? I don't see that defined anywhere.

sandrina-p: IncomeDistribution, I refer to the following file imported at `state.js`, used to create the getter `groupIncomeDistribution`:

> import incomeDistribution from '~/frontend/utils/distribution/mincome-proportional.js'

taoeffect: OK thanks! I'll have a close look at what you're doing here when after the weekend (or perhaps tomorrow, we'll see)

frontend/model/state.js
```diff
@@ -243,6 +243,9 @@ const getters = {
   ourUsername (state) {
     return state.loggedIn && state.loggedIn.username
   },
+  ourContractId (state) {
```
taoeffect: "ContractId" could refer to any number of contracts, it's important to specify which contractID this is returning; `ourContractId` => `ourIdentityContractId`

frontend/views/components/Tooltip.vue
```diff
@@ -113,7 +113,7 @@ export default {
 <style lang="scss" scoped>
 @import "@assets/style/_variables.scss";
 
-.c-wrapper {
+.c-twrapper {
```
taoeffect, line 116: What is the reason for this change?

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
-      const surplus = totalAmount - groupPledgeGoal
-      const membersNeedingPledges = groupMembersTotal - othersPledges.length
-
-      // Dumb Rule of 3 algorithm to know how much the user will receive (userIncomeToReceive) - just for layout purposes.
-      const avgMembersNeedingPledges = goal / membersNeedingPledges
-      const avgNeededPledges = neededPledges / membersNeedingPledges
-      const userIncomeToReceive = userIncomeNeeded ? (userIncomeNeeded - (userIncomeNeeded * avgNeededPledges / avgMembersNeedingPledges)).toFixed(2) : 0
+      const mincome = this.groupSettings.mincomeAmount
+      const ourPledgeAmount = this.type === 'pledgeAmount' && this.amount >= 0 && this.amount
+      const ourIncomeAmount = this.type === 'incomeAmount' && this.amount < mincome && this.amount
```
taoeffect: I think the checks in the middle (`this.amount >= 0` and `this.amount < mincome`) aren't needed, because vuelidate should perform those checks itself.

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
+      const doWeNeedPledge = typeof ourIncomeAmount === 'number'
```
taoeffect: Think we can avoid performing redundant computation with `const doWeNeedPledge = this.type === 'incomeAmount'`, and then replacing redundant checks like `this.type === 'incomeAmount'` (e.g. the ones on line 76 and 75) with `doWeNeedPledge` and `!doWeNeedPledge` (since they are mutually exclusive)

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
+          othersPledgesAmount += amount
+          othersPledgesCount += 1
+        }
+      })
+
+      const ourIncomeNeeded = doWeNeedPledge ? mincome - ourIncomeAmount : null
+      const pledgeTotal = othersPledgesAmount + ourPledgeAmount
+      const groupGoal = othersIncomeNeeded + ourIncomeNeeded
+      const neededPledges = groupGoal - pledgeTotal
+      const surplus = pledgeTotal - othersIncomeNeeded - ourIncomeNeeded
```
taoeffect:
```js
const neededPledges = Math.max(0, groupGoal - pledgeTotal)
const surplus = Math.max(0, pledgeTotal - othersIncomeNeeded - ourIncomeNeeded)
```

Then can remove the [branches](https://en.wikipedia.org/wiki/Branch_(computer_science)) on lines 121 and 122

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
+
+      let ourIncomeToReceive
```
taoeffect: This can be initialized to `ourIncomeNeeded` so that lines 110-112 can be removed

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,195 @@
...
+      if (neededPledges > 0) {
```
taoeffect: This would then be:

```js
if (doWeNeedPledge && neededPledges > 0) {
```

frontend/views/containers/GroupPledgesGraph.vue
taoeffect: OK, yes, so here's what needs to be done:

1. Up above in the loop starting at line 82, convert that to a for/in loop that looks similar to the loop in `groupIncomeDistribution` in order to fill up an array called `newIncomeDistribution` (just like the `currentIncomeDistribution` inside of `groupIncomeDistribution`)
2. Outside of the loop, perhaps in here, push again to newIncomeDistribution `{ name: username, amount: ourIncomeAmount }`
3. In here, do something like `ourIncomeToReceive = incomeDistribution(newIncomeDistribution, mincome).filter(i => i.to === username).reduce((acc, cur) => cur.amount + acc, 0)` (double-check me there) to generate `ourIncomeToReceive`

taoeffect (changes requested):
See comments + merge conflicts!

frontend/assets/style/_reset.scss
```diff
@@ -34,7 +34,7 @@ h3,
 h4,
 h5,
 h6 {
-  font-weight: bold;
+  font-weight: normal;
```
pieer, line 37: Wow look like a big change for that pull request.. but I guess you've tested everywhere where there is a title that it look like the design, isn't it?

frontend/assets/style/_typography.scss
```diff
@@ -37,8 +37,7 @@ body {
   font-style: normal;
   font-weight: normal;
   font-size: $size-4; // 14px
-  line-height: 1.3125rem; // 21px
-  letter-spacing: 0.1px;
+  line-height: 1.5; // 21/14
```
pieer, line 40: Same here ;)

frontend/assets/style/_typography.scss
```diff
@@ -153,6 +152,14 @@ small {
   .has-background-#{$name} {
     background-color: var(--#{$name}_2);
   }
+
+  .has-background-#{$name}-solid {
```
pieer, line 156: Good call

frontend/views/components/Graphs/PieChart.vue
```diff
@@ -1,53 +1,43 @@
 <template lang='pug'>
 .c-container
-  svg.c-piechart(viewbox='-1 -1 2 2' :style='{ height: size }')
+  svg.c-piechart(viewBox='-1 -1 2 2' aria-hidden='hidden')
```
pieer, line 3: Interesting viewbox. Could you had a comment please?

frontend/views/components/Graphs/PieChart.vue
```diff
@@ -166,4 +150,18 @@ export default {
   text-align: center;
   line-height: 1.2;
 }
+
+// Global Classes for SVG fills
```
pieer: Why did you move it from utilities? I could use the same class for my graph actually

frontend/views/components/Tooltip.vue
pieer: It should not affect the rest because of the scope, I would keep the c-wrapper. If there is a generic css call c-wrapper, that's a mistake

frontend/views/containers/contributions/IncomeDetails.vue
pieer: There is already 2 kind of modals, one full screen and one not. Is this a 3rd kind?

pieer (commented):
Good call job on refactoring group proposal test.
I'm not so sure about the full screen modal that look like overlapping on the other one. But I guess you must have you reason (could you explain quickly?)
I fill that changing the basic font property (not bold) for all titles is surprising at this time of the project, but again I trust your judgement. 
In general, I think that updating something that affect all the website for a pull request with the title implement graphic is not really appropriate. But above all, it's good that you improve the full website when you can... so I let you judge.

frontend/views/containers/contributions/IncomeDetails.vue
sandrina-p: As far as I understand, we don't have a "real" fullscreen modal. To implement a fullscreen modal we have been following different approaches:
1- create a wrapper child of `modal-base-template` and style it, adding the dimensions and the missing `background-color: $general_2`. Examples: this modal, PayGroupHistory, etc...
2 - Style `.modal` "indirectly". I say indirectly because it's not explicit that it's a class inside `modal-base-template`, for example `CreateGroup.vue` or `Join.vue`. There we also manually add the `background-color`. 
3 - Add that same bg to some elements inside the modal, for example, `SettingsWrapper` (the user settings modal), where we add the bg to its sidebar and its main.

And, no matter the approach, all of these modals are incorrectly styled because the paddings and the button style (having ESC) are wrong. 
For those reasons, I decided to explicit add a `fullscreen` prop to `modal-base-template` that deal with that for us. No need to add the bg color. It comes already centered aligned and with the correct paddings and close button styled correctly.

Does this make sense?

frontend/views/components/Graphs/PieChart.vue
sandrina-p: You are right! Moved back to utilities and with better names.

frontend/assets/style/_reset.scss
sandrina-p: Yes, I did exactly that on the last PR where we changed `is-X` to `is-title-X`, so now, *every* title must have the `is-title-x` class, so there's no need to add `font-weight: bold` to heading elements. When we need to use a heading element without looking like a title, we don't need to override its style anymore.

frontend/views/components/Graphs/PieChart.vue
sandrina-p: Yes, sure, there's a comment below with a URL explaining more about how this piechart was made:
// Learn more about SVG & PieCharts
// -> https://hackernoon.com/a-simple-pie-chart-in-svg-dbdd653b6936

frontend/views/components/Tooltip.vue
sandrina-p: I should have explained this... actually renaming `c-wrapper` to `c-twrapper` or `c-gwrapper` (at groupPledgesGraph) has a reson, although I don't like it very much. Maybe @pieer you know a better solution. So, let me explain, by starting to rollback the class name to `c-wrapper` as before:

vue-bug

See those classes in the DOM `c-wrapper`, one for the graphic (blue outline) and other for the tooltip (red outline). For some reason, the `data-v-59ba5d0f` attribute is added also to the tooltip, so the tooltip will have not only its styles (orange styles) but also the styles applied to the graphic wrapper (green styles), breaking the layout.

After some experiments, I realized this only happens when applied to the component top-level element (the wrapper of everything inside that component).

This is really annoying because it made me realize that CSS in Vue isn't truly scoped. So for that reason using `c-wrapper` or `c-container` as the top class of every component can potentially lead to unexpected bugs as this one. So, I believe we should find a little more "unique" names, at least for those top-level elements. That's why I wrote `c-gwrapper` and `c-twrapper`.

frontend/views/containers/GroupPledgesGraph.vue
sandrina-p: Those are needed, otherwise, the graphic would display strange values such as negative values.

frontend/assets/style/_typography.scss
sandrina-p: `line-height` 1.3125rem is exactly the same as 1.5, so it was just a matter of readability.
`letter-spacing` was removed because it isn't needed - I confirmed with @mmbotelho.

mmbotelho:
> So, we should show only income to receive, even if there's space for surplus as well? (scenario 3)

Yes, or else it's very confusing

frontend/views/containers/GroupPledgesGraph.vue
sandrina-p: there's no need to doWeNeedPledge. `neededPledges > 0` is enough

sandrina-p:
@greg, @pieer addressed all the issues you've pointed. Please take a look again.

@mmbotelho same for you, even those not directly related with this PR.

mmbotelho (changes requested):
Only missing this:
Captura de ecrã 2019-12-03, às 14 00 31

sandrina-p:
@mmbotelho done!

mmbotelho (dismissed):
Perfect! Thank you @sandrina-p 😄

frontend/views/components/Tooltip.vue
taoeffect: Wow, @sandrina-p ... is there a way you can add a shortened version of this explanation to the Style Guidelines CSS section? I'm still having some difficulty understanding it, but it's a bit clearer to me now. Maybe link to this comment as well.

Also, yeah, if @pieer you have any ideas on how to avoid this, please do let us know!

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,204 @@
...
-      return 'column c-legend-item'
+    ...mapGetters([
+      'groupSettings',
+      'groupProfiles',
+      'groupMembersCount',
+      'ourIdentityContractId',
+      'ourUsername'
+    ]),
+    doWePledge () {
+      return this.type === 'pledgeAmount'
```
taoeffect: Note: since this function is only called from within `graphData()`, it is better to keep this as a variable in that function. A function call can add significant overhead, and the `graphData()` function has significant computation in it. So anything unnecessary should be removed. So I recommend defining this as a `doWeNeedPledge` const inside `graphData()`

frontend/views/containers/GroupPledgesGraph.vue
taoeffect: `neededPledges > 0` can be true even when `doWeNeedPledge` is false, and when `doWeNeedPledge` is false, we don't need to run this calculation, right?

frontend/views/containers/GroupPledgesGraph.vue
```diff
@@ -1,218 +1,204 @@
...
-      this.isMobile = this.verifyIsMobile()
-    }, 100),
-    verifyIsMobile () {
-      return window.innerWidth < TABLET
+    decimalSlice (amount) {
+      const perc = amount / this.graphData.groupGoal
+      // avoid breaking the graph when perc is bigger than 1,
+      if (perc > 1) return 1
+      if (perc < 0) return 0
+      return perc
```
taoeffect: can save some lines:

```
return Math.max(1, Math.min(0, perc))
```

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -197,7 +194,7 @@ export default {
         [L('This field is required')]: required,
         [L("It seems your income is not lower than the group's mincome.")]: function (value) {
```
taoeffect: Boyscoutting:

`"It seems your income is not lower than the group's mincome."` => `"Income must be lower than group mincome"`

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -197,7 +194,7 @@ export default {
...
           if (this.needsIncome) {
-            return value < this.groupSettings.mincomeAmount - 1
+            return value < this.groupSettings.mincomeAmount
           }
           return true
```
taoeffect: save some lines:

```
return !this.needsIncome || value < this.groupSettings.mincomeAmount
```

.Gruntfile.babel.js
taoeffect: seems about time to remove :)

taoeffect (changes requested):
Some minor suggestions + don't forget to update .Gruntfile.babel.js :)

sandrina-p:
@greg, changes made and branch updated with master

frontend/views/containers/contributions/IncomeDetails.vue
mmbotelho: Can we write a complete sentence so it doesn't sound so robotic? Maybe `Your income must be lower than the group mincome.`

taoeffect: Agreed, that sound even better :)

taoeffect (changes requested):
## Setup

1. Created a group with Person A and Person B
2. Person A - set their monthly income to $600

### Issue 1 - Incorrect center value number and slices

Screen Shot 2019-12-17 at 9 11 37 AM

Here Person B has gone to add their income details, and they're told that the Group Goal is $1400. This is what's shown when neither option is picked, I guess because it's assumed that Person B needs income and has $0 income themselves.

We can see how the Group Goal actually *decreases* when they say their monthly income is $300:

Screen Shot 2019-12-17 at 9 13 17 AM

Instead, when no option is picked, the graph should show different numbers in the center, depending on whether or not there's enough pledges. In this case, there wasn't enough pledges to cover the amount needed, so it should show "$400 needed".

Now let's look at what happens if we have Person B decide to pledge $300:

Screen Shot 2019-12-17 at 9 12 57 AM

Here we see two issues:

1. The pie chart should show enough slices to illustrate the situation: that $300 is being pledged, that $400 was needed, and that $100 is still needed, which means we need to have at least 2 slices in the graph (or 3), but at the moment only one blue slice is shown.
2. The center value of the chart should change, to show how much is still needed, but here it stays fixed, saying that $400 is the "Group Goal". Instead it should change to show that `1000 - (600 + 300) = $100` is still needed to meet the mincome for the group

I stopped playing with the chart at this point, but this is the process I took, to see if the chart fully tells the story of what's going on in the group, depending on which radio button is selected, and what value is entered for either the pledge or the user's income.

### Issue 2 - Travis build still failing

Not sure if the latest failure is an intermittent one or not, but would be great if it can be figured out so that the Travis builds are consistent again.

sandrina-p:
@greg, it's fixed! iI was a typo based on your suggestion, I added `doWePledge` and it should be `!doWePledge` so that’s why the data was wrong. Cypress failed for that same reason (because we have tests to cover those scenarios)

Also, I added `hasData` variable to exclude the user income from the group goal when the user didn't answer the question yet.

taoeffect (changes requested):
Screen Shot 2019-12-17 at 10 07 38 AM

Looks like the white slice is still missing

frontend/views/containers/GroupPledgesGraph.vue
sandrina-p: This is wrong @taoeffect , that's what causing the bug you are facing (missing the white slice)

sandrina-p:
@greg, the white slice is fixed. This [suggestion you made related to `decimalSlice()`](https://github.com/okTurtles/group-income-simple/pull/786#pullrequestreview-330227352) was the cause of the issue.

frontend/views/containers/GroupPledgesGraph.vue
taoeffect: Oops, it should have been:

```js
return Math.min(1, Math.max(0, perc))
```

sandrina-p: It makes sense now (and it works!) changed

taoeffect (approved):
Approved!!! Great job @sandrina-p! The pie chart is so much fun to play with, it seems to work great in my testing!! 😄

PR #791, opened 2019-12-18
Implement proposal "Change mincome"

### LINKED ISSUE #554: Create "change minimum income" proposal
_**Note**: There might be some dependencies with the pull request #343 (Proposals/voting logic)._

## Problem
In the current Master version of the application, there's no way to change the minimum income of the group. I am not sure if this was already implemented on the PR #343.

Captura de ecrã 2019-04-14, às 17 17
Captura de ecrã 2019-04-14, às 17 18 02

## Solution

There are some differences from what was designed at the time of the PR #343:

- The UI is slightly different (updated to fit the new modal design);
- There's one additional step, where the user can add a description to the proposal they are about to create - this will help other members of the group understand why;
- There's a "success" view in the modal, that tells the user what will happen now that a proposal has been created.

[You can see the Figma file here.](https://www.figma.com/file/ZR0DRPhhUuHdJtf7PbGAM6tW/Group-Income?node-id=2314%3A0)

sandrina-p:
Closes #554

image

Note: After talking with @mmbotelho, we decided to replace the `icon-dollar-sign` with `icon-coins`, so it's an icon with monetary purposes no matter the group currency.

### DIFF
--- a/frontend/model/contracts/voting/proposals.js
+++ b/frontend/model/contracts/voting/proposals.js
@@ -78,36 +78,6 @@ const proposals = {
         data: data.passPayload
       })
       sbp('okTurtles.events/emit', PROPOSAL_RESULT, state, VOTE_FOR, data)
-      // TODO: for now, generate the link and send it to the user's inbox
-      //       however, we cannot send GIMessages in any way from here
-      //       because that means each time someone synchronizes this contract
-      //       a new invite would be sent...
-      //       which means the voter who casts the deciding ballot needs to
-      //       include in this message the authorization for the new user to
-      //       join the group.
-      //       we could make it so that all users generate the same authorization
-      //       somehow...
-      //       however if it's an OP_KEY_* message ther can only be one of those!
-      //
-      //       so, the deciding voter is the one that generates the passPayload data for the
-      //       link includes the OP_KEY_* message in their final vote authorizing
-      //       the user.
-      //       additionally, for our testing purposes, we check to see if the
-      //       currently logged in user is the deciding voter, and if so we
-      //       send a message to that user's inbox with the link. The user
-      //       clicks the link and then generates an invite accept or invite decline message.
-      //
-      //       we no longer have a state.invitees, instead we have a state.invites
-      // sbp('okTurtles.events/emit', PROPOSAL_RESULT, state, VOTE_FOR, data)
-      // TODO: generate and save invite link, which is used to generate a group/inviteAccept message
-      //       the invite link contains the secret to a public/private keypair that is generated
-      //       by the final voter, this keypair is a special "write only" keypair that is allowed
-      //       to only send a single kind of message to the group (accepting the invite, deleting
-      //       the writeonly keypair, and registering a new keypair with the group that has
-      //       full member privileges, i.e. their identity contract). The writeonly keypair
-      //       that's registered originally also contains attributes that tell the server
-      //       what kind of message(s) it's allowed to send to the contract, and how many
-      //       of them.
     },
     [VOTE_AGAINST]: voteAgainst
   },
@@ -138,8 +108,15 @@ const proposals = {
         [RULE_DISAGREEMENT]: { threshold: 1 }
       }
     },
-    [VOTE_FOR]: function (state, { proposalHash, passPayload }) {
-      console.error('unimplemented!')
+    [VOTE_FOR]: function (state, data) {
+      const proposal = state.proposals[data.proposalHash]
+      proposal.status = STATUS_PASSED
+      const { setting, proposedValue } = proposal.data.proposalData
+
+      sbp('gi.contracts/group/updateSettings/process', state, {
+        meta: proposal.meta,
+        data: { [setting]: proposedValue }
+      })
     },
     [VOTE_AGAINST]: voteAgainst
   },
--- a/frontend/model/contracts/voting/proposals.js
+++ b/frontend/model/contracts/voting/proposals.js
@@ -78,6 +78,36 @@ const proposals = {
         data: data.passPayload
       })
       sbp('okTurtles.events/emit', PROPOSAL_RESULT, state, VOTE_FOR, data)
+      // TODO: for now, generate the link and send it to the user's inbox
+      //       however, we cannot send GIMessages in any way from here
+      //       because that means each time someone synchronizes this contract
+      //       a new invite would be sent...
+      //       which means the voter who casts the deciding ballot needs to
+      //       include in this message the authorization for the new user to
+      //       join the group.
+      //       we could make it so that all users generate the same authorization
+      //       somehow...
+      //       however if it's an OP_KEY_* message ther can only be one of those!
+      //
+      //       so, the deciding voter is the one that generates the passPayload data for the
+      //       link includes the OP_KEY_* message in their final vote authorizing
+      //       the user.
+      //       additionally, for our testing purposes, we check to see if the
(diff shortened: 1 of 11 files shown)


frontend/model/contracts/voting/proposals.js
```diff
@@ -78,36 +78,6 @@ const proposals = {
...
-      // sbp('okTurtles.events/emit', PROPOSAL_RESULT, state, VOTE_FOR, data)
-      // TODO: generate and save invite link, which is used to generate a group/inviteAccept message
-      //       the invite link contains the secret to a public/private keypair that is generated
-      //       by the final voter, this keypair is a special "write only" keypair that is allowed
-      //       to only send a single kind of message to the group (accepting the invite, deleting
-      //       the writeonly keypair, and registering a new keypair with the group that has
-      //       full member privileges, i.e. their identity contract). The writeonly keypair
-      //       that's registered originally also contains attributes that tell the server
-      //       what kind of message(s) it's allowed to send to the contract, and how many
-      //       of them.
```
sandrina-p: Note: @taoeffect correct me if I'm wrong, but I think this comment is stale because we changed the way we generate invites, that's why I deleted it. If needed, I can add it back/update it with what's missing.

frontend/model/contracts/voting/proposals.js
```diff
@@ -138,8 +108,15 @@ const proposals = {
...
-      console.error('unimplemented!')
+    [VOTE_FOR]: function (state, data) {
+      const proposal = state.proposals[data.proposalHash]
+      proposal.status = STATUS_PASSED
+      const { setting, proposedValue } = proposal.data.proposalData
+
+      sbp('gi.contracts/group/updateSettings/process', state, {
+        meta: proposal.meta,
+        data: { [setting]: proposedValue }
+      })
```
sandrina-p, line 149: @taoeffect I have two questions here:

1. when we do `proposal.status = STATUS_PASSED` we are mutating the state directly. (I copied from INVITE_MEMBER function). Can you remind me why we don't use Vuex mutations to mutate the state?

2. Why do we call sbp `/process` instead of `/create` as we usually do in other places? In this case, why not do the following:

```js
       const updatedSettings = await sbp(
          'gi.contracts/group/updateSettings/create',
          { [setting]:  proposedValue },
          data.proposalData.groupId
        )
        sbp('backend/publishLogEntry', updatedSettings)
```

sandrina-p: Copy/Past @taoeffect's answer from Slack:

> 1. We do use Vuex mutations (as single mutation that does dispatch to sbp). We do it this way because of the SBP integration which gives us greater flexibility
> 2. /create is for creating messages to send. No messages should be sent here. In this case we use /process to DRY.

1. Okay... I think I remember now, the mutation is triggered by the sbp `'backend/publishLogEntry'` that calls the state action `syncContractWithServer`.

2. Why it isn't needed to be sent here?

pieer (dismissed):

frontend/model/contracts/voting/proposals.js
taoeffect: Yeah, this comment is still needed. None of the `OP_KEY_*` stuff has been implemented yet, and so we need this here as a reminder to do that. Any comments related to key stuff should be preserved until after all of that is implemented.

frontend/model/contracts/voting/proposals.js
taoeffect: > Okay... I think I remember now, the mutation is triggered by the sbp 'backend/publishLogEntry' that calls the state action syncContractWithServer.

Not exactly. If we're talking about where the Vuex mutation happens, mentioning `'backend/publishLogEntry'` is to go too far back up the chain of events.

`'backend/publishLogEntry'` is what is used to send a GIMessage, which results in a network event. For conceptual purposes, think of that as a totally separate sequence of events. If we want to discuss where the Vuex mutation is, we have to think not in terms of sending messages, but in terms of receiving them.

The conceptual chain of events for receiving messages begins with `'state/enqueueHandleEvent'`, which is called by our websocket when the client receives a message from the server. From there, the Vuex action `handleEvent` gets called, and from there the sequence of events to the Vuex mutation goes like this: `processMutation` => store.commit(`${contractID}/processMessage` => `processMessage` (**inside the Vuex mutation now!**) => `guardedSBP` => `sbp`

By doing it this way, we decouple our logic from Vuex. We can, if we want to, replace Vuex with any other system. And we can, similarly, turn all of our contract logic into a re-usable library that will work in any project, whether or not they use Vuex, Redux, etc.

> Why it isn't needed to be sent here?

Well think about what is happening. We are receiving and processing message. Why would we need to send any new messages at this point? It makes no sense to send out any other messages.

taoeffect: Oh, oops, there is one more thing, and that is you are right to mention `syncContractWithServer`, because that is the other way messages are received in the app.

There are two ways messages are received:

1. Via the websocket (that flow is described in my previous comment above)
2. Via `syncContractWithServer`, which happens in two cases: upon joining a group, and when we sign back into an existing group after being offline for a while.

Both routes will result in calls to `handleEvent` and the chain of events mentioned above.

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -92,36 +97,54 @@ export default {
   },
```
taoeffect, line 94: Boyscoutting: on line 91, I think (double-check me) you can simply do:

```js
return currencies[this.groupSettings.mincomeCurrency].symbolWithCode
```

And in fact it is more correct too.

taoeffect: Also, another boyscoutting issue: on line 89, we shouldn't be using a string literal `'group-setting-change'`  but the constant `PROPOSAL_GROUP_SETTING_CHANGE`

taoeffect (changes requested):
This is looking pretty good. I tested it and it seems to work well too. I've replied to all your questions and have also made some boyscoutting related requests, please have a look!

frontend/model/contracts/voting/proposals.js
sandrina-p: Hum... but the part talking about "the sending the invite to the inbox" is stale because now members are only invited by link. So... can you clarify what part of the comment should stay/be modifed? :/

frontend/model/contracts/voting/proposals.js
sandrina-p: Thanks for the detailed explanation! Definitely I'm saving this comment on my notes!

mmbotelho (dismissed):

frontend/model/contracts/voting/proposals.js
taoeffect: Just leave the whole thing, it will be clear which parts of it are no longer relevant. Even if some parts aren't relevant, I don't want to lose anything related to what I was thinking at the time.

sandrina-p: added back

mmbotelho (dismissed):

frontend/views/containers/proposals/Mincome.vue
taoeffect: Nice job on using the existing `groupMincomeSymbolWithCode`, but still missing fix for `'group-setting-change'`

sandrina-p: done

taoeffect (approved):
Great job @sandrina-p!

Issue #757, opened 2019-11-04
Improve overall experience of sending / receiving monetary contributions


mmbotelho:
### Problem

As discussed on the 2019 hackathon, Group Income's main function - to send or receive payments to or from members - needs to:
- Allow for Bitcoin payments;
- Display a history of payments made to someone, and their respective status (e.g.: I'm sending a payment via Bitcoin, there were 2 attempts to pay and both failed, the third one was successful);
- Allow users to change the amount of money sent if they wish to;
- Be a great user experience - as @taoeffect said "We need to do one thing and be great at it";

### Solution

We did a workshop together, in Bulgaria, where all team members brainstormed and discussed what an ideal payment system would look like. We took the opportunity to think outside the box, and focus 100% on our user's experience, not paying much attention to technical limitations. These are some of the main ideas that came out of it:
- The payment should be as automatic as possible;
- Members should be able to add more than one payment method;
- Using a traditional bank, Paypal, Venmo and other similar non-crypto systems should be an option;
- Have an option to process all payments automatically;
- Send emails with a summary of that month's transactions to members;
- Notify users by email that it's time to send their payments.

Now, I don't believe we need to implement or discuss all of these possibilities for the MVP. The ones I think are really important are:
- An **email notification** saying that it's time to do the monthly payments. I am aware that we are trying to avoid using email at all costs, but we cannot expect our users to always remember to do their payments or to check our app regularly. If users do not remember to do their payments (and receiving members are shy / afraid to ask because it might be rude), this could render the whole thing useless. I am, of course, open to discuss this. I am not aware of any methods that can work as well as email for this, but if any of you are, please share them!
- Having a way to **do all payments automatically** - this would be opt-in, of course, and only available for Bitcoin (I am not sure if this is possible) and other digital methods. The goal here is for it to be possible to have a subscription-type payment, where users don't even need to go to the app to send their monthly contributions. Psychologically, it can be damaging to be reminded every single month that you are giving money away. If people had to go every month to a website and click a button to pay for their gym subscription, for example, there would be a lot less unused gym subscriptions around. Of course that, in the example I just used, this would be a good thing for everyone, but for Group Income, I believe we should reduce the emotional / psychological burdens to a maximum when it comes to these contributions, and help people focus on the great things that come from sharing their resources with each other (and not the resources themselves);
- Having some more **traditional payment methods** - what services could we use here? 

[...]

PR #793, opened 2019-12-26
Task/563 Implement Contributions Widget

### LINKED ISSUE #563: Implement Contributions Widget
_**Note:** Related to #501 

### Problem

At the moment, it's not easy to understand which contributions have been received / made so far unless the user goes to the "Contributions" page. 

### Solution

Captura de ecrã 2019-04-29, às 14 47 30

Add a "widget" that lets the user:
- Keep track of their monthly contributions (whether they are sending or receiving money);
- Keep their stated pledge/need visible (related to issue #501);
- Keep the action to change their pledge / need at hand, as this information might change often;
- Keep a link to non-monetary contributions (help improve feature discoverability).

For detailed mockups of this widget [go here](https://www.figma.com/file/ZR0DRPhhUuHdJtf7PbGAM6tW/Group-Income?node-id=1579%3A12) (desktop, tablet and mobile). You can also see it in the [context of the Dashboard](https://www.figma.com/file/ZR0DRPhhUuHdJtf7PbGAM6tW/Group-Income?node-id=1253%3A14719).

sandrina-p:
Closes #563 

image

image

- When a group is created, the only widget visible is the one to add the income details. Once that is filled, it's the usual flow - Start Inviting and the other widgets.
- Because #757 is still being discussed, I didn't connect the Payments progress bar to the store. For now, it's fake data - it always says "1 out of 3 ". In another PR, after #757, we can finish it along with PayGroup page's sidebar.


mmbotelho (commented):
In terms of layout, everything seems perfect to me! 

However, the data seems to be behaving in a weird way. In this scenario, one member was pledging and the other needed pledges. However, it says that no one is in need of contributions ATM:
Captura de ecrã 2019-12-27, às 16 08 56
Captura de ecrã 2019-12-27, às 16 09 02

mmbotelho (changes requested):
@sandrina-p I tried everything I could think of to replicate the error I mentioned in [my previous comment](https://github.com/okTurtles/group-income-simple/pull/793#pullrequestreview-336824890) but wasn't able to 😢 If you no else can replicate it as well you can ignore it. Maybe it was some error on my part.

I noticed this small visual bug:
Captura de ecrã 2020-01-07, às 18 14 07

Once that is fixed, I'll aprove the PR!

frontend/model/state.js
```diff
@@ -243,13 +243,84 @@ const getters = {
...
+    return (userContract.attributes && userContract.attributes.displayName) || getters.ourUsername
+  },
   ourIdentityContractId (state) {
     return state.loggedIn && state.loggedIn.identityContractID
   },
   // Logged In user's identity contract
   ourUserIdentityContract (state) {
     return (state.loggedIn && state[state.loggedIn.identityContractID]) || {}
   },
+  ourContributionSummary (state, getters) {
```
pieer, line 261: Good point

frontend/model/state.js
```diff
@@ -243,13 +243,84 @@ const getters = {
   ourUsername (state) {
     return state.loggedIn && state.loggedIn.username
   },
+  ourUserDisplayName (state, getters) {
```
pieer, line 249: Add todo refactor "userDisplayName"  in Profile.vue, Welcome.vue (coming with my commit, but I don't want to mess with this code)

frontend/views/pages/GroupDashboard.vue
```diff
@@ -2,22 +2,28 @@
...
+    start-inviting(v-if='groupMembersCount === 1')
 
-  page-section(title='This months overview')
-    overview
+    contributions-widget(v-if='hasIncomeDetails')
 
-  page-section(title='July Overview')
-    //- group-pledges-graph
-    progress-overview
+  proposals-widget(v-if='hasIncomeDetails || Object.keys(currentGroupState.proposals).length > 0')
```
pieer: Object.keys(currentGroupState.proposals).length > 0
-> should we add this logic into computed property?

frontend/model/state.js
sandrina-p: done

frontend/views/pages/GroupDashboard.vue
sandrina-p: done

sandrina-p:
@mmbotelho the border bug was fixed. I hope this time is fixed, for real, everywhere.

mmbotelho (dismissed):
Approved! Great job @sandrina-p 🎉

frontend/model/contracts/group.js
```diff
@@ -336,6 +336,14 @@ DefineContract({
...
         var groupProfile = state.profiles[meta.username]
         const nonMonetary = groupProfile.nonMonetaryContributions
+
+        // Prevent stale amounts when the user changes their incomeDetailsType
+        if (data.incomeDetailsType === 'pledgeAmount') {
+          Vue.set(groupProfile, 'incomeAmount', null)
+        } else if (data.incomeDetailsType === 'incomeAmount') {
+          Vue.set(groupProfile, 'pledgeAmount', null)
+        }
+
```
taoeffect: What is the reason for this logic? I have a few issues with it:

- It seems unnecessary
- The comment and code assume that we are changing `incomeDetailsType` in this action, which we might not be doing at all

Everything in this file needs to be as clean and minimal as possible, since it will all be "set in stone", and will define a protocol that alternative implementations will have to implement. So anything that isn't necessary should be removed.

frontend/model/state.js
```diff
@@ -243,13 +243,85 @@ const getters = {
...
+        }, [])
+
+        return listWho.length > 0 ? { what: listWhat, who: listWho } : null
+      })(),
+      givesNonMonetary: (() => {
+        const contributions = ourGroupProfile.nonMonetaryContributions
+
+        return contributions.length > 0 ? contributions : null
+      })()
+    }
```
taoeffect, line 325: There are a few issues here:

1. This code belongs in ContributionsWidget.vue, not here. The getters in this file must only be functions that are used throughout the app, and they must be super efficient and compact — i.e. primitive building blocks only. This function is neither highly used throughout the app (it is specific to ContributionsWidget.vue, the use in PaymentsSummary.vue does not need it), nor is it super efficient or compact
2. This function defines and runs code for at least 4 different functions, regardless of whether it's needed or not. It could probably be done more cleanly when moved into ContributionsWidget.vue

sandrina-p: 1. This getter can also be used at `IncomeDetailsGraph.js` and `Contributions.js`. Yes, we only need it 3 times so far, but it's kinda complex and duplicating that code doesn't seem wise.
2. It runs 4 functions, but each one of them has an early return in case it doesn't make sense to perform further logic. I don't see a problem to be as it is for now. When we reuse it at the places mentioned before, if needed, we can evolve it.

sandrina-p: 1. This getter can also be used at IncomeDetailsGraph.js and Contributions.js. Yes, we only need it 3 times so far, but it's kinda complex and duplicating that code doesn't seem wise.
2. It runs 4 functions, but each one of them has an early return in case it doesn't make sense to perform further logic. I don't see a problem to be as it is for now. When we reuse it at the places mentioned before, if needed, we can evolve it.

frontend/model/contracts/group.js
sandrina-p: Removed, it's not necessary indeed. However, I still don't like the idea of leaving "traces" of stale data, there's no reason to keep a value that doesn't make sense anymore.

taoeffect: You might be able to fix that in the UI portion of the app

sandrina-p: > You might be able to fix that in the UI portion of the app

hmm but there's nothing to fix ou the UI side. this is a state optimization, not UI bug.

taoeffect: There is definitely something that can be fixed on the UI side: reset the value. The effect to the user will be the same, while the logic in our contract chains will be kept to a minimum

sandrina-p: I think I'm not understanding... reset the value where exactly? At the moment is all working fine.

taoeffect: In the modal for the income details. If you're worried about a stale value appearing there, that's where you can reset it.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -2,22 +2,28 @@
...
-  //-   group-settings
+    page-section(title='July Overview')
+      //- group-pledges-graph
+      progress-overview
+
+    //- page-section(title='Support History')
+    //-   support-history
+
+    //- page-section(title='Group Settings')
+    //-   group-settings
```
taoeffect: The organization here needs to be improved. See my [two comments](https://github.com/okTurtles/group-income-simple/pull/789/files#diff-7eb644607ce1945a00d85cf4544ada76) to Pierre in `GroupDashboard.vue` about how the logic in this file should be organized. I think we should wait until Pierre's PR is merged before merging this one so that you can resolve the merge conflicts, but you might also want to clean some of this up beforehand too, or at least think about how to go about it.

From a pure programming perspective, it's best to avoid checking the same condition over and over, in this case you have a lot of checks against `hasIncomeDetails`, and ideally there would only ever be one (plus an optional `v-else`).

frontend/model/contracts/group.js
```diff
@@ -336,6 +336,7 @@ DefineContract({
       process (state, { data, meta }) {
         var groupProfile = state.profiles[meta.username]
         const nonMonetary = groupProfile.nonMonetaryContributions
+
```
taoeffect: This change doesn't seem to be needed?

frontend/model/state.js
taoeffect: > This getter can also be used at IncomeDetailsGraph.js and Contributions.js. Yes, we only need it 3 times so far, but it's kinda complex and duplicating that code doesn't seem wise.

I can't find those files in the project... Do you mean `IncomeDetails.vue` and `Contributions.vue`?

I agree that code duplication isn't a good idea. However, if you're going to create code in order to de-duplicate code, then go ahead and do so by modifying those files n this PR.

We still need to address the issue of this function being very large, and running all of the functions regardless of what data is being called.

There has to be a cleaner and more modular way to do this, and a way in which code is run only when it needs to be run. Do you think you can see if you can clean this up a bit while DRYing the other files too? Maybe also finding another file to put this code in (perhaps creating it if necessary)? The thing is, `state.js` is already bigger than it should be. You've got time since Pierre's PR needs to be merged first, and if you'd like I can lend a hand after Pierre's PR is merged.

taoeffect (changes requested):
(This isn't a full review but a partial one.)

frontend/views/pages/GroupDashboard.vue
sandrina-p: The checks for `hasIncomeDetails` are essential. If you see a simpler way to avoid those checks, please let me know

frontend/model/contracts/group.js
sandrina-p: line removed

frontend/model/state.js
sandrina-p: Yes, I did mean .vue files, not .js.

> and a way in which code is run only when it needs to be run.
This code only runs once, when it's checked.

 I'll take a closer look on these as you suggested and let you know when it's ready.

sandrina-p: done @taoeffect. I refactored `Contributions.vue` to use the getter `ourContributionsSummary` and it needed all the 4 functions as I was expecting. So, in summary, this getter is used twice and 100% needed, no code could be removed.

I strongly believe that the right place for `ourContributionsSummary` is as a getter, as it is. The main advantage of using getters instead of a "standard" util function, is that a getter is only run once in the entire app, even if called multiple times (in our case, it's called twice - on the widget and on the contributions page).

> The thing is, state.js is already bigger than it should be.

Totally agree, but this isn't a problem of `ourContributionsSummary`. I mentioned some time ago that `state.js` as it is today, is not scalable. We should take better advantage of Vuex modules, and split the `state.js` into modules, [(its a common pattern for large vuex apps)](https://medium.com/3yourmind/large-scale-vuex-application-structures-651e44863e2f). Well, this is a very valid point that requires further discussion, but I don't see this directly related to this PR. If you want, I can create an issue for us to discuss state there.

frontend/assets/style/components/_links.scss
```diff
@@ -1,5 +1,5 @@
 .link {
-  display: inline-block;
+  display: inline; // so border breaks correctly on multiple lines.
```
taoeffect, line 2: These comments are super helpful 👍

frontend/model/state.js
```diff
@@ -246,13 +246,84 @@ const getters = {
   ourGroupProfile (state, getters) {
     return getters.groupProfile(getters.ourUsername)
   },
+  ourUserDisplayName (state, getters) {
+    // TODO - refactor Profile and Welcome and any other component that needs this
+    const userContract = getters.ourUserIdentityContract || {}
+    return (userContract.attributes && userContract.attributes.displayName) || getters.ourUsername
```
taoeffect, line 252: This is a great example of a getter that definitely belongs in this file, great job! 👍

frontend/views/components/PageSection.vue
```diff
@@ -1,9 +1,9 @@
 <template lang='pug'>
 section.card
   .c-title-wrapper(v-if='$slots.cta')
-    h3.is-title-3(v-if='title') {{ title }}
+    h2.is-title-3(v-if='title') {{ title }}
     slot(name='cta')
-  h3.is-title-3(v-else-if='title') {{ title }}
+  h2.is-title-3(v-else-if='title') {{ title }}
```
taoeffect, line 6: Is this a @mmbotelho sanctioned change? :)

frontend/views/pages/GroupDashboard.vue
```diff
@@ -46,15 +47,23 @@ export default {
...
     ...mapGetters([
+      'ourUsername',
       'currentGroupState', // TODO normalize getters names
       'groupSettings',
       'groupsByName',
       'groupMembersCount',
       'groupProfiles',
-      'ourGroupProfile'
+      'ourGroupProfile',
+      'groupProfile'
```
taoeffect: `groupProfile` is only used once in this file, and it's used to do exactly what `ourGroupProfile` does, so please delete this and replace line 63 with `this.ourGroupProfile`

frontend/model/state.js
```diff
@@ -246,13 +246,84 @@ const getters = {
...
     return state.loggedIn && state.loggedIn.identityContractID
   },
   // Logged In user's identity contract
   ourUserIdentityContract (state) {
     return (state.loggedIn && state[state.loggedIn.identityContractID]) || {}
   },
+  ourContributionSummary (state, getters) {
+    const groupProfiles = getters.groupProfiles
+    const ourUsername = getters.ourUsername
+    const ourGroupProfile = groupProfiles[ourUsername]
```
taoeffect: Use `getters.ourGroupProfile`

frontend/model/state.js
taoeffect: > Well, this is a very valid point that requires further discussion, but I don't see this directly related to this PR. If you want, I can create an issue for us to discuss state there.

OK, that sounds reasonable, please go ahead and create an issue for this, and link to this thread (leaving it unresolved)

frontend/views/components/CalloutCard.vue
```diff
@@ -34,9 +34,9 @@ export default {
   margin-right: $spacer;
   flex-shrink: 0;
 
-  @include desktop {
+  @include tablet {
     width: 6.25rem;
-    height: 6.25rem;
+    height: auto;
```
taoeffect, line 39: @pieer does this change make sense to you?

frontend/model/state.js
```diff
@@ -246,13 +246,84 @@ const getters = {
...
+
+    if (!ourGroupProfile.incomeDetailsType) {
+      return {}
+    }
+
+    const doWeNeedIncome = ourGroupProfile.incomeDetailsType === 'incomeAmount'
+    const distribution = getters.thisMonthsPayments.frozenDistribution || getters.groupIncomeDistribution
+
+    const nonContributionsOf = (username) => groupProfiles[username].nonMonetaryContributions || []
```
taoeffect: These are contributions, not non-contributions, so rename to `nonMonetaryContributionsOf`

taoeffect (changes requested):
My review is almost, but not quite finished, there are still some parts I haven't checked. Leaving comments for what I was able to get to today though!

frontend/views/components/CalloutCard.vue
pieer: I confirm it make sense :) It's a design choice that respect @mmbotelho design

NB: Sandrina created the CalloutCard, that part concern the SVG icon on the left of the callout text, not the graph

frontend/views/components/PageSection.vue
sandrina-p: No, it's just a matter of having the correct heading structure. This component is used as sections of the pages, so the page title is an h1, which means each card is an h2. if it was h3, we would have a heading gap and that shouldn't happen. Here's a screenshot for you to better understand:

image

frontend/model/state.js
sandrina-p: It was a typo, I actually wanted to write `nonMonetaryContributionsOf `! 😶

frontend/model/state.js
sandrina-p: Done. https://github.com/okTurtles/group-income-simple/issues/804

sandrina-p: Done. #804

frontend/views/components/PageSection.vue
taoeffect: So I assume that means h2 and h3 have the same style...?

sandrina-p: the style is applied by the class `.is-title-X`, it has nothing to do with the heading tag

taoeffect: ok, so in other words, does that mean the heading tags should never be used without an `.is-title-X` marker...? Usually the h1, h2, h3 tags determine the style/size by themselves. Why don't they do that in our project, and why do they need a `.is-title-X` for it? (Is this something that can be explained in the Style Guide CSS section? If so can you create an issue for it?)

sandrina-p: That was a change made a few PR ago: https://github.com/okTurtles/group-income-simple/pull/777 Basically we did it to support correctly responsive headers.

> Usually the h1, h2, h3 tags determine the style/size by themselves. 

This is a misconception/bad practice in terms of accessibility. Markup (semantics) should be decoupled from styling. We can style something to look like a heading and not be an actual heading or vice-versa.

>  (Is this something that can be explained in the Style Guide CSS section? If so can you create an issue for it?)

Yes, this is the [best/simpler explanation of why](https://stackoverflow.com/a/19166926/4737729)

taoeffect: OK thanks, I've gone ahead and created issue #805 for the documentation about this

frontend/views/pages/Contributions.vue
```diff
@@ -96,7 +97,7 @@ page(pageTestName='contributionsPage' pageTestHeaderName='contributionsTitle')
           v-if='notContributing'
           tag='p'
           data-test='givingParagraph'
-        ) You can contribute to your group with money or other valuables like teaching skills, sharing your time ot help someone. The sky is the limit!
+        ) You can contribute to your group with money or other valuables like teaching skills, sharing your time to help someone. The sky is the limit!
```
taoeffect, line 100: Nice catch! 👍

frontend/views/pages/Contributions.vue
```diff
@@ -187,90 +188,50 @@ export default {
...
-      'thisMonthsPayments',
       'groupMincomeFormatted',
-      'groupIncomeDistribution',
-      'globalProfile'
+      'globalProfile',
+      'ourContributionSummary'
     ]),
+    ourGroupProfile () {
+      return this.groupProfile(this.ourUsername)
+    },
```
taoeffect: DRY: already defined in `state.js`

frontend/views/pages/GroupDashboard.vue
```diff
@@ -2,15 +2,16 @@
...
   div(v-else)
-
     start-inviting(v-if='groupMembersCount === 1')
 
     contributions-overview(v-if='canDisplayGraph')
 
-    proposals-widget
+    contributions-widget
+
+    proposals-widget(v-if='Object.keys(currentGroupState.proposals).length > 0')
```
taoeffect: DRY: use `hasProposals`

EDIT: re: `master` conflict, this change should be combined with what's in `master`, i.e. the indentation should be decreased (as in Pierre's recent quickfix for #806), but we should also only show it if there are proposals to show

frontend/views/pages/Page.vue
```diff
@@ -23,6 +23,7 @@ div(:data-test='pageTestName' :class='$scopedSlots.sidebar ? "p-with-sidebar" :
     v-if='$scopedSlots.sidebar'
     :class='{ "is-active": ephemeral.isActive }'
   )
+    i18n.sr-only(tag='h2') Page Details
```
taoeffect: Does this need a `.is-title-2` on it? If not, can you add a comment explaining why not? (And maybe add a comment to #805 to use this as an example of when not to use it?)

frontend/views/containers/Welcome.vue
```diff
@@ -1,33 +1,26 @@
...
 import sbp from '~/shared/sbp.js'
-import { mapGetters } from 'vuex'
 import { OPEN_MODAL } from '@utils/events.js'
+import { mapGetters } from 'vuex'
 import CalloutCard from '@components/CalloutCard.vue'
 import SvgHello from '@svgs/hello.svg'
 
 export default {
-  name: 'StartInviting',
+  name: 'AddIncomeDetailsWidget',
```
taoeffect: This should be the same as the filename. So either this needs to be `Welcome` or the file should be renamed

frontend/views/containers/contributions/ContributionsWidget.vue
```diff
@@ -0,0 +1,184 @@
...
+          title: L('You need {br_}{amount}', {
+            ...this.LTags(),
+            amount: this.withCurrency(receivingMonetary.needed)
+          }),
+          status: L('You will receive {amount}.', {
+            amount: this.withCurrency(receivingMonetary.total)
+          })
+        }
+      } else if (givingMonetary) {
+        const copyMonetaryTitle = (amount) => L('You are pledging {br_} {amount}', {
```
taoeffect: Should the space after `{br_}` be there?

taoeffect (changes requested):
Great job @sandrina-p! Minor changes requested. I think this is also ready for @mmbotelho to review to verify everything works per her designs. I still need to do my own tests as well and will do that soon, probably after the next set of changes (or earlier)

test/cypress/integration/group-payments.spec.js
```diff
@@ -131,7 +131,7 @@ describe('Payments', () => {
     // After closing the modal it should dislay how much user need
     cy.getByDT('headerNeed').should('contain', 'You need $100')
     // The user should be inform that even if he can\'t pledge he can still contribute
-    cy.getByDT('givingParagraph').should('contain', 'You can contribute to your group with money or other valuables like teaching skills, sharing your time ot help someone. The sky is the limit!')
+    cy.getByDT('givingParagraph').should('contain', 'You can contribute to your group with money or other valuables like teaching skills, sharing your time to help someone. The sky is the limit!')
```
taoeffect, line 150: ~~Re tests: do you think you can add a couple tests to test that this new widget works?~~

Oh, it's not hooked up to the state yet. But does that mean there's nothing at all that can be tested?

frontend/views/pages/GroupDashboard.vue
pieer: I think merging master might resolve the problem here

taoeffect: Oh yeah, this is already outdated, I'll resolve this thread

frontend/views/pages/Page.vue
pieer: YEs, to stay consistent we need the class to the title now

sandrina-p: No, it doesn't. This is a heading only for screen readers, so they know what's this sidebar about. That's the purpose of `.sr-only`.

frontend/views/containers/Welcome.vue
sandrina-p: File renamed to `AddIncomeDetailsWidget`.

frontend/views/containers/contributions/ContributionsWidget.vue
sandrina-p: ~It's irrelevant... it has no influence on the layout. I would keep the space before and after fo readability.~

Edited: It doesn't make a difference on the layout, but on testing it makes. Ideally, it should have only 1 space between one word and another. So in this case, I added the space before {br_} and removed the one after. That way we can assert with one space only:
- "You are pledging $100" (correct - one space)
- "You are pledging &nbsp;$100" (incorrect - double space)

test/cypress/integration/group-payments.spec.js
sandrina-p: Yes, it is hooked up to the store! The payment progress is the only part not hooked yet. But nevertheless, I just added some tests to Contributions Widget as part of `group-payments.spec.js`!

frontend/views/pages/Contributions.vue
taoeffect: Still here :)

frontend/views/pages/Contributions.vue
```diff
@@ -299,6 +260,9 @@ export default {
     },
     displayName (username) {
       return this.globalProfile(username).displayName || username
+    },
+    wCurrency (amount) {
```
taoeffect: I like how the other file called this function `withCurrency`, can we make this also clearer by naming it `withCurrency`?

frontend/views/pages/Page.vue
```diff
@@ -18,11 +18,11 @@ div(:data-test='pageTestName' :class='$scopedSlots.sidebar ? "p-with-sidebar" :
   main.p-main(:class='mainClass')
     slot
 
-  nav.p-sidebar(
-    :role='L("Secondary")'
+  section.p-sidebar(
     v-if='$scopedSlots.sidebar'
     :class='{ "is-active": ephemeral.isActive }'
   )
+    i18n.sr-only(tag='h2') Page details
```
taoeffect, line 25: Can you update our documentation using this as an example of when to use `h2` without also adding an `.is-title-X` and explaining why?

frontend/views/pages/Page.vue
```diff
@@ -5,7 +5,7 @@ div(:data-test='pageTestName' :class='$scopedSlots.sidebar ? "p-with-sidebar" :
     h1.is-title-2.p-title(:data-test='pageTestHeaderName' v-if='$slots.title')
       img.c-logo(
         src='/assets/images/group-income-icon-transparent.png'
-        :alt='L("Group Income\'s logo")'
```
taoeffect, line 8: Is there a reason this was deleted?

sandrina-p: Yes. This image doesn't add any extra value to the content, it's only decorative. This is a title and having `{{ title }}` is enough. There are no benefits for the SR (screen reader) to read this image to the user, so in this case we just leave an empty alt and the SR will ignore the image.

taoeffect (approved):
Great job @sandrina-p! Excellent work! 😄  🎉

PR #795, opened 2020-01-07
Task/753 containers and components

### LINKED ISSUE #753: Explain the difference between Containers and Components
### Problem

Our [Style Guide](https://github.com/okTurtles/group-income-simple/blob/master/docs/Style-Guide.md) doesn't explain what the difference between the "containers" and "components" folders is.

### Solution

- Update the docs to explain the difference
- Audit our files to make sure all files are in the right folder, and move+refactor as necessary

sandrina-p:
Closes #753 

- Added Folder Structure Style Guide to the docs and reviewed all files at `frontend/views/` to reflect the changes. Deleted unused files/imports.

> ### Naming Conventions
> - The folders are `kebab-case`
> - Vue files are `PascalCase.vue`
> - JS files are `camelCase.js`
> ### Structure
> ```bash
> frontend
> └─ assets # Any static file including Styles.
> └─ controller
> └─ model
> └─ utils 
> └─ views # where Vue files live
>     └─ components # Vue files reused in multiple places.
>     └─ containers # Vue files for a specific part of the app, not reusable.
>     └─ pages # A Vue component connected to a route
>     └─ utils # Utils used only within /frontend/views
>     | main.js # App entry point
> ```

There are still some files I'm not sure about them, **your opinion is welcome**:

- **Module vs Context folder pattern**
 Most `containers` folder are following a "module" pattern, where everything related to X is placed at `containers/X/` (ex: all stuff related to proposals go into `containers/proposals`. However, there are some folders that follow a "context" pattern instead. For example, Modals, we have been placing all modals at `containers/modals` except (wait for it...) proposal modals!
The context pattern is also used for `containers/forms`. What do you think about this? Is it okay for you, do you have any suggestion? I'm okay with it, this is always a work in progress, so I believe with time we can better answer this question.

- **Navigation naming & structure**
We have 2 navigations, those 2 sidebars. What name should we give it? left and right? main/primary and secondary? global and local? Damn, naming things is hard! At the moment it's a mess to distinguish both. I created `containers/navigation` and placed everything related to the _main_ navigation there. Then, for each "local" navigation (the secondary sidebar), I think we can stick with the "module" pattern (everything about x goes into containers/x). For example, the local nav for the dashboard and payments (pay-group) are placed under its `container/x` folder


pieer:
> * **Module vs Context folder pattern**
>   Most `containers` folder are following a "module" pattern, where everything related to X is placed at `containers/X/` (ex: all stuff related to proposals go into `containers/proposals`. However, there are some folders that follow a "context" pattern instead. For example, Modals, we have been placing all modals at `containers/modals` except (wait for it...) proposal modals!
>   The context pattern is also used for `containers/forms`. What do you think about this? Is it okay for you, do you have any suggestion? I'm okay with it, this is always a work in progress, so I believe with time we can better answer this question.

I believe that everything under container should follow the module pattern. The idea behind having for example all the modal in the same folder was to be able to search for modal example.
But this can be done easily doing a search in the editor IF we name all the modal container with ModalSomething.

> * **Navigation naming & structure**
>   We have 2 navigations, those 2 sidebars. What name should we give it? left and right? main/primary and secondary? global and local? Damn, naming things is hard! At the moment it's a mess to distinguish both. I created `containers/navigation` and placed everything related to the _main_ navigation there. Then, for each "local" navigation (the secondary sidebar), I think we can stick with the "module" pattern (everything about x goes into containers/x). For example, the local nav for the dashboard and payments (pay-group) are placed under its `container/x` folder

I agree, let stick to the module pattern

sandrina-p:
@pieer 
> I believe that everything under the container should follow the module pattern. [...] this can be done easily doing a search in the editor IF we name all the modal container with ModalSomething.

I agree. Based on your feedback I deleted `/forms` and `/modals` container folders and moved its files to their respective containers. However, note that the file name is `SomethingModal` instead of `ModalSomething`, so it follows [Vue Style Guide Strongly Recommended Convention](https://vuejs.org/v2/style-guide/#Order-of-words-in-component-names-strongly-recommended). I did the same for Widgets and Forms components (ex: `ProposalWidget` and `LoginForm` respectively)

pieer (dismissed):
Great work!

taoeffect (approved):
@sandrina-p you are on 🔥

PR #801, opened 2020-01-18
799 Cypress bypass ui 

### LINKED ISSUE #799: Cypress - Bypass UI in repetitive flows
### Problem
As we have been building up the tests on Cypress, there are 3 flows that are used a lot of times:
- Sign In
- Log in
- Group Creation 

Each one of the flows is "quick" when done just once. However, when done several times it starts to take too much time and it's tedious to watch. For example, settings up a group of 4 members, takes ~15 seconds. And a group with 20 members with income details takes ~70 seconds.

### Solution

Implement a solution that [bypasses the UI](https://docs.cypress.io/guides/getting-started/testing-your-app.html#Fully-test-the-login-flow-%E2%80%93-but-only-once) on the flows mentioned above.

sandrina-p:
Closes #799 

Create page `/bypass-ui` to perform 3 repetitive actions (flows):
- `signup`
- `login`
- `groupCreation`

Created a file `frontend/actions/[action].js` with the code for each action isolated. This was a simple copy paste with minimal changes. These actions can now be called by both "real UI" and BypassUI pages without duplicating the code and avoid bugs. I also liked these because now the Vue components are way cleaner.

So far, so good. Everything works fine when using the app manually. 

Now, on Cypress side:

Add `bypassUI` to the following cypress commands, to bypass the UI.

```js
cy.giSignup('user1', { bypassUI: true })
cy.giLogin('user1', { bypassUI: true })
cy.giCreateGroup('Turtles', { bypassUI: true })
```

Now, the problems will start to appear. Just one: **Cypress is faster than contracts/state sync**.
A quick dirty solution was to add `cy.wait(500)` to manually wait for it. If it isn't enough, increase the wait time and _it will work_. (see `commands.js` inside function `cyBypassUI()`.)

Now, let's replicate the problem:
- Comment the 2 `cy.wait(500)`.
- Run the project `grunt dev` and wait to start.
- Run cypress `npm run cy:open`.
- Click the test `signup-and-login.js`. the test should fail:

image

That red box is added at `index.html`. Has the `username`, `loggedIn` and `state.contracts` data to help us understand what's going on.

So, why did the test fail? The contracts didn't fully sync to `localforage`. After we login, we create the group immediately by accessing `/app/bypassUI`. It was too quick, the data was lost. So when we access the state, it's empty (username and all contracts lost). That's why the email input is empty and the test fails.

Solution: Reflect on the DOM when the contract is fully synced and get that info by cypress, _before_ accessing bypassUI page. For now, we have a `cy.wait(500)`.

----

After adding the first `cy.wait(500)`, tests started to pass (yey!) However, sometimes I cross another failing test. After login (at `/app`), it doesn't redirect to `/dashboard`. More details on the snapshot:

Screenshot 2020-01-18 at 18 38 10

Quick solution: Add another `cy.wait(500)` after `bypassUI` action was succeeded. The right solution should be similar to what I said before.

## 💡Final Update

Read this comment to understand the final solution: https://github.com/okTurtles/group-income-simple/pull/801#discussion_r370586641


frontend/model/contracts/identity.js
```diff
@@ -8,7 +8,11 @@ DefineContract({
   name: 'gi.contracts/identity',
   contract: {
     validate: objectOf({
-      attributes: object
+      attributes: objectOf({
```
taoeffect: Using `objectOf` fixes and limits the possibilities for how this contract can be initialized for all of time. That is a dangerous thing to do, since it limits what can be done and makes it unnecessarily difficult to modify the software to allow other attributes.

`objectMaybeOf` is likely a better choice here, because it provides a check while not limiting future possibilities.

frontend/views/containers/forms/FormLogin.vue
```diff
@@ -35,14 +35,14 @@ form(
...
 <script>
-import sbp from '~/shared/sbp.js'
 import { validationMixin } from 'vuelidate'
 import { required } from 'vuelidate/lib/validators'
 import { nonWhitespace } from '@views/utils/validators.js'
 import FormPassword from '@containers/forms/FormPassword.vue'
 import BannerScoped from '@components/BannerScoped.vue'
 import L from '@view-utils/translations.js'
 import validationsDebouncedMixins from '@view-utils/validationsDebouncedMixins.js'
+import login from '../../../actions/login.js'
```
taoeffect: Avoid reintroducing cumbersome and ugly `../../`'s to the codebase by using `~/`

frontend/actions/login.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+export default async function login ({
+  username,
+  password // TODO - implement password
+}) {
+  // TODO: Insert cryptography here
+  const identityContractID = await sbp('namespace/lookup', username)
+  console.log('ora bem', identityContractID)
```
taoeffect: I'm sure you're aware this logging needs to be removed, but as part of my duty as reviewer I still need to mention it. :)

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
...
+import { merge } from '@utils/giLodash.js'
+
+// TODO - reuse this to handle other GI specific errors
+class ErrorCaused extends Error {
+  constructor (cause, ...args) {
+    super(...args)
+    this.cause = cause
+    Error.captureStackTrace(this, ErrorCaused)
+  }
+}
```
taoeffect: OK, so how to DRY and handle this new Error class, I recommend the following:

1. Move it into the existing file we have for project errors: `~/frontend/model/errors.js` to keep it grouped with all other project-defined errors
2. Since this intended to be a UI generated runtime ephemeral error — i.e. one that is not generated by processing a contract message — we need to use comment banners to separate it from the other 3 errors in that file (e.g. `// -----------------------`), and also give it a unique name that makes it clear that this error is of a different nature than the other 3 errors in that file. I recommend `GIErrorUIRuntimeError`, which keeps the naming convention of using `GIError` while making it clear that this is a runtime error generated by the UI and not during contract message processing. Add a comment above the class definition explaining all of this.
3. Then, instead of checking `e.cause` or `e.name`, simply use `e.message` -- I will post other comments inline explaining this point. This will make the error handling across the app simpler and cleaner.

In the end, it will look like this:

```js
export class GIErrorUIRuntimeError extends Error {
  constructor (...params) {
    super(...params)
    this.name = this.constructor.name
    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, this.constructor)
    }
  }
}
```

And then in this file we can import it like so:

```js
import { GIErrorUIRuntimeError } from '@model/errors.js'
```

frontend/views/containers/forms/FormLogin.vue
```diff
@@ -68,21 +68,19 @@ export default {
...
         this.$emit('submitSucceeded')
       } catch (e) {
-        console.error(e)
-        this.$refs.formMsg.danger(L('Failed to login, please try again. {codeError}', { codeError: e.message }))
+        if (e.cause === 'INVALID_MATCH') {
+          this.$refs.formMsg.danger(L('Invalid username or password'))
+        } else {
+          console.error(e)
+          this.$refs.formMsg.danger(L('Failed to login. {codeError}', { codeError: e.message }))
+        }
```
taoeffect: These lines would then be replaced with:

```js
console.error('FormLogin.vue login() error:', e)
this.$refs.formMsg.danger(e.message)
```

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
...
+  thresholdMemberRemoval
+}, cb) {
+  let finalPicture = `${window.location.origin}/assets/images/default-group-avatar.png`
+
+  if (picture) {
+    try {
+      finalPicture = await imageUpload(picture)
+    } catch (e) {
+      console.error('Failed to upload the group picture', e)
+      throw new ErrorCaused('PICTURE_UPLOAD_FAILED', e)
```
taoeffect: So here we would write:

```js
throw new GIErrorUIRuntimeError(L('Error uploading group picture: {codeError}', { codeError: e.message }))
```

This exact error string will then be translated properly and displayed correctly regardless of what code is calling this function.

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
...
+    sbp('okTurtles.events/once', hash, (contractID, entry) => {
+      sbp('state/vuex/commit', 'setCurrentGroupId', hash)
+      return cb ? cb() : true
+    })
+    await sbp('backend/publishLogEntry', entry)
+    // add to vuex and monitor this contract for updates
+    await sbp('state/enqueueContractSync', hash)
+  } catch (e) {
+    console.error('Failed to create the group', e)
+    throw Error(e)
```
taoeffect: Similarly, here we'd have:

```js
throw new GIErrorUIRuntimeError(L('Error creating group: {codeError}', { codeError: e.message }))
```

The error handling and throwing for the rest of the actions would need to be similarly modified.

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
+import sbp from '~/shared/sbp.js'
```
taoeffect: Now, these actions are great candidates for SBP selectors.

I recommend using `sbp` to call them instead of calling them directly. You can place them under the domain `gi.actions` -- this domain will be responsible for performing every user action that can be performed in the app. We will need this anyway for making the app voice-controllable. So we'd have the following SBP API:

- `gi.actions/contract/subscribe` <-- takes a string or an array of strings (representing contractIDs) and calls `sbp('state/enqueueContractSync', <hash>)` on them
- `gi.actions/contract/subscribeAndWait` <-- takes a string or an array of strings (representing contractIDs), creates an array of promises that each register an event listener for `CONTRACT_IS_SYNCING` on for their contractID, and resolve when the event listener is called with `isSyncing` = `false` (see `syncContractWithServer` function in `state.js`). Then calls `gi.actions/contract/subscribe` on the string (or array of strings), and then returns `Promise.all(arrayOfPromises)`.
- `gi.actions/group/create`
- `gi.actions/group/createAndSubscribe` <-- calls `gi.actions/group/create` first and then calls `gi.actions/contract/subscribe` on the value returned
- `gi.actions/group/createAndSubscribeWait` <-- same as `gi.actions/group/createAndSubscribe` except calls `gi.actions/contract/subscribeAndWait` instead of `gi.actions/contract/subscribe`
- `gi.actions/group/switch` <-- simply calls vuex commit on `setCurrentGroupId` mutation
- `gi.actions/user/create`
- `gi.actions/user/createAndSubscribe` <-- first calls `gi.actions/user/create`, and then automatically calls `gi.actions/contract/subscribe` on the 2 contractIDs that are returned (the identityContractID and the mailbox for that identityContractID)
- `gi.actions/user/createAndSubscribeWait` <-- same as `gi.actions/user/createAndSubscribe` except calls `gi.actions/contract/subscribeAndWait` instead of `gi.actions/contract/subscribe`
- `gi.actions/user/loginAndSubscribeWait` <-- simply call `sbp('state/vuex/dispatch', 'login', { .. })`, and modify the Vuex action to replace calls to `'state/enqueueContractSync'` with calls to `'gi.actions/contract/subscribeAndWait'`
- `gi.actions/user/logout` <-- For now can simply call `sbp('state/vuex/dispatch', 'logout')`

I then recommend creating these files for organization (note they should be under **frontend/controller/**):

- `frontend/controller/actions/contract.js`
- `frontend/controller/actions/group.js`
- `frontend/controller/actions/user.js`

And you can create a top-level `~/frontend/controller/actions/index.js` that imports all the files in that folder, and then import that in `main.js`.

All of the SBP selectors/actions for group go in `group.js`, and likewise all user related selectors/actions go in `user.js`. Look at the files in `shared/domains` for how to write these files using `export default sbp('sbp/selectors/register', {`

frontend/views/containers/forms/FormSignup.vue
```diff
@@ -76,59 +77,20 @@ export default {
...
+      try {
+        await signup({
+          username: this.form.name,
+          email: this.form.email,
+          password: this.form.password
+        })
+        this.$emit('submitSucceeded')
+      } catch (error) {
+        console.error('Signup.vue submit() error:', error)
+        this.$refs.formMsg.danger(L('Failed to signup. {codeError}', { codeError: error.message }))
```
taoeffect: These lines would then be replaced with:

```js
console.error('FormSignup.vue signup() error:', e)
this.$refs.formMsg.danger(e.message)
```

frontend/views/containers/modals/CreateGroup.vue
```diff
@@ -107,62 +104,26 @@ export default {
...
-        // add to vuex and monitor this contract for updates
-        await sbp('state/enqueueContractSync', hash)
       } catch (e) {
-        console.error('Failed to create the group', e)
-        this.$refs.formMsg.danger(L('Failed to create the group, please try again. {codeError}', { codeError: e.message }))
+        if (e.cause === 'PICTURE_UPLOAD_FAILED') {
+          return this.$refs.formMsg.danger(L('Failed to upload the group picture. {codeError}', { codeError: e.message }))
+        }
+
+        this.$refs.formMsg.danger(L('Failed to create the group. {codeError}', { codeError: e.message }))
```
taoeffect: These lines would then be replaced with:

```js
console.error('CreateGroup.vue submit() error:', e)
this.$refs.formMsg.danger(e.message)
```

frontend/views/pages/BypassUi.vue
```diff
@@ -0,0 +1,150 @@
...
+  },
+  computed: {
+    actions () {
+      return {
+        signup: {
+          actionFn: (params) => {
+            const username = this.$store.getters.ourUsername
+            if (username) {
+              // QUESTION: Should we do this validation inside signup? Same for login
+              throw Error(`You're signed as '${username}'. Logout first and re-run the tests.`)
```
taoeffect: All `Error`s [should be thrown with `new`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw), so all `throw Error` statements should be replaced with `throw new Error`

frontend/views/containers/modals/CreateGroup.vue
```diff
@@ -107,62 +104,26 @@ export default {
...
-              ),
-              [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposals[PROPOSAL_PROPOSAL_SETTING_CHANGE].defaults,
-              [PROPOSAL_GENERIC]: proposals[PROPOSAL_GENERIC].defaults
-            }
-          }
-        })
-        const hash = entry.hash()
-        sbp('okTurtles.events/once', hash, (contractID, entry) => {
-          this.$store.commit('setCurrentGroupId', hash)
+        groupCreation({
```
taoeffect: This is an asynchronous action, so:

```js
const groupID = await sbp('gi.actions/group/create', { ...
```

frontend/actions/signup.js
```diff
@@ -0,0 +1,60 @@
...
+    for (const contract of [user, mailbox]) {
+      await sbp('state/enqueueContractSync', contract.hash())
+    }
+    // TODO: Just add cryptographic magic
+    // login also calls 'state/enqueueContractSync', but not in this case since we
+    // just sync'd it.
+    await sbp('state/vuex/dispatch', 'login', {
+      username,
+      identityContractID: user.hash()
+    })
```
taoeffect: To ensure our actions do *exactly* what their names say they will do (in this case, this selector is called `'gi.actions/user/create'`), we need to move two different things out of here:

1. The loop that calls `'state/enqueueContractSync'` -- we'll want to make a call to `gi.actions/contract/subscribe` in the caller instead (with the two contractIDs that are returned, see comment below). Of course, this should be done in `gi.actions/user/createAndSubscribe` and `gi.actions/user/createAndSubscribeWait`, and then one of those should be called in `FormSignup.vue`
2. The call to the vuex login action -- this also will need to be done in the caller, with an sbp call to the `gi.actions/user/loginAndSubscribeWait ` action

frontend/actions/signup.js
```diff
@@ -0,0 +1,60 @@
...
+
+    return true
+  } catch (error) {
+    sbp('state/vuex/dispatch', 'logout')
+    throw Error(error)
```
taoeffect: Here again we need a different error handler. We don't want to logout here, since this action no longer logs us in. If the proper step is to logout upon error, that should be done in the caller (e.g. `FormSignup.vue`).

frontend/actions/signup.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    // TODO: Just add cryptographic magic
+    // login also calls 'state/enqueueContractSync', but not in this case since we
+    // just sync'd it.
+    await sbp('state/vuex/dispatch', 'login', {
+      username,
+      identityContractID: user.hash()
+    })
+
+    return true
```
taoeffect: As with the create group action, let's have this return the contractID. However, instead of returning `user.hash()`, since there are two contracts that we need to subscribe to, let's return an array instead:

```js
return [ user.hash(), mailbox.hash() ]
```

Then in the caller we can do a loop on this array and call `gi.actions/contract/subscribe` on each one.

Alternatively, the caller can call `gi.actions/user/createAndSubscribe` to have that done for it, followed by `gi.actions/user/login`.

frontend/views/containers/forms/FormSignup.vue
```diff
@@ -76,59 +77,20 @@ export default {
...
-          console.error('Signup.vue submit() error:', e)
-          this.$refs.formMsg.danger(L('Failed to signup, please try again. {codeError}', { codeError: e.message }))
-          sbp('state/vuex/dispatch', 'logout')
-        }
+      try {
+        await signup({
+          username: this.form.name,
+          email: this.form.email,
+          password: this.form.password
+        })
```
taoeffect, line 88: Right after this should then be a call to `await sbp('gi.actions/user/loginAndSubscribeWait', ...`

frontend/actions/login.js
```diff
@@ -0,0 +1,31 @@
...
+  if (!identityContractID) {
+    throw new ErrorCaused('INVALID_MATCH')
+  }
+
+  try {
+    console.debug(`Retrieved identity ${identityContractID}`)
+    await sbp('state/vuex/dispatch', 'login', { username, identityContractID })
+
+    return true
```
taoeffect: Let's return something more useful, like `identityContractID`, which then will match the API of the other actions.

frontend/actions/login.js
```diff
@@ -0,0 +1,31 @@
...
+  } catch (e) {
+    throw Error(e)
```
taoeffect: 1. If we aren't handling the error, and we aren't doing anythign useful after the call to vuex/dispatch, then we do not need the try/catch block at all here, so it can be removed entirely
2. Just for future reference, to re-throw an error you simply do: `throw e`

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
...
+      }
+    })
+    const hash = entry.hash()
+    sbp('okTurtles.events/once', hash, (contractID, entry) => {
+      sbp('state/vuex/commit', 'setCurrentGroupId', hash)
+      return cb ? cb() : true
+    })
+    await sbp('backend/publishLogEntry', entry)
+    // add to vuex and monitor this contract for updates
+    await sbp('state/enqueueContractSync', hash)
```
taoeffect: Now here, instead of `enqueueContractSync` (which will be called via `'gi.actions/contract/subscribe'` in the caller, see comment above), we simply return the contractID upon success:

```js
return hash
```

frontend/actions/groupCreation.js
```diff
@@ -0,0 +1,90 @@
...
+          [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposals[PROPOSAL_PROPOSAL_SETTING_CHANGE].defaults,
+          [PROPOSAL_GENERIC]: proposals[PROPOSAL_GENERIC].defaults
+        }
+      }
+    })
+    const hash = entry.hash()
+    sbp('okTurtles.events/once', hash, (contractID, entry) => {
+      sbp('state/vuex/commit', 'setCurrentGroupId', hash)
+      return cb ? cb() : true
+    })
```
taoeffect: 1. Returning from within this callback does nothing, so no `return` statement is needed here
2. Following our rule that "actions should do what their name says", we want separate out the call to `setCurrentGroupId` so that instead `gi.actions/group/switch` can be called instead from the caller, and we also separate out the call to `enqueueContractSync` and call `gi.actions/contract/subscribe` in the caller
3. This code could be replaced with the following (HOWEVER DO NOT DO THIS, I have an even better idea!)

    ```js
        sbp('okTurtles.events/once', hash, (contractID, entry) => {
          cb && cb(contractID, entry)
        })
    ```

The reason why this `sbp('okTurtles.events/once', hash` code is placed here before `publishLogEntry` and `enqueueContractSync`, is because if we place it after the call to `enqueueContractSync` then it might not get run at all, because the contract might finish syncing (which results in an emit of `hash` from `handleEvent`) before we've set our event handler. We therefore set the event handler *before* calling `await sbp('state/enqueueContractSync'`.

However, since we are refactoring everything into tiny building block actions, we should instead do the following:

1. Do not pass in a callback
2. Do not call `enqueueContractSync` here, instead, do that in the caller

Then, in the caller, we will have code that looks something like the following:

```js
const groupID = await sbp('gi.actions/group/create', { ... })
sbp('okTurtles.events/once', groupID, (contractID, entry) => {
  sbp('gi.actions/group/switch', hash) 
  this.next()
})
await sbp('gi.actions/contract/subscribe', groupID)
// or 
// await sbp('gi.actions/contract/subscribeAndWait', groupID)
```

frontend/main.js
```diff
@@ -90,7 +90,7 @@ async function startApp () {
       return {
         ephemeral: {
           syncs: [],
-          finishedLogin: 'no'
+          finishedLogin: !!this.$store.state.loggedIn
```
taoeffect: This is definitely going to break things since it modifies a string to being a boolean but doesn't modify any of the other code to take that into account.

taoeffect (changes requested):
OK, I think if the suggested changes are implemented the problem should be fixed.

frontend/views/components/TransitionExpand.vue
```diff
@@ -13,6 +13,11 @@ export default {
           element.style.height = 'auto'
         },
         enter (element) {
+          // TODO/QUESTION -How we do access store here?
```
taoeffect: You could use `sbp('state/vuex/state')`

frontend/actions/login.js
sandrina-p: Now, using `GIErrorUIRuntimeError`, I'd suggest replacing the throw for

```js
throw new GIErrorUIRuntimeError(L('Failed to login: {codeError}', { codeError: e.message }))
```

frontend/actions/groupCreation.js
sandrina-p: So we have 2 similar methods: `/subscribe` and `/subscribeAndWait`. Can you give an example where we just need subscribe? Shouldn't we always wait for it?

frontend/actions/signup.js
sandrina-p: Hum... I don't agree. The purpose of all this effort is to encapsulate all the logic related to an action. To be all together. (in this case signup) So, if doing logout is needed in case of error, it should be done here. Otherwise, we would have signup logic repeated in 2 places (formSignup and BypassUI)

frontend/main.js
sandrina-p: updated to return 'yes' or 'no'

frontend/controller/actions/contract.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default sbp('sbp/selectors/register', {
+  'gi.actions/contract/subscribe': async function (contractIDs: string | array) {
+    const listOfIds = typeof contractIDs === 'string' ? [contractIDs] : contractIDs
+
+    for (const id of listOfIds) {
+      await sbp('state/enqueueContractSync', id)
+    }
+  },
+  'gi.actions/contract/subscribeAndWait': async function (contractIDs: string | array) {
```
sandrina-p: @taoeffect, regarding this domain API, you suggested:

> - gi.actions/contract/subscribe <-- takes a string or an array of strings (representing contractIDs) and calls sbp('state/enqueueContractSync', <hash>) on them
> - gi.actions/contract/subscribeAndWait <-- takes a string or an array of strings (representing contractIDs), creates an array of promises that each register an event listener for CONTRACT_IS_SYNCING on for their contractID, and resolve when the event listener is called with isSyncing = false (see syncContractWithServer function in state.js). Then calls gi.actions/contract/subscribe on the string (or array of strings), and then returns Promise.all(arrayOfPromises).

The same for the user and group domains. Build a `create|subscribe` method and 1 similar method prefixed with `andWait`.

My questions are:

1. why not creating only one method that by default waits for CONTRACT_IS_SYNCING? (See another comment related to this below)

2. I tried to do the "andWait" part to set `CONTRACT_IS_SYNCING` but I didn't understand/ don't know how to do it exactly... Even after checking `syncContractWithServer`. Could you be more explicit on that part?

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,89 @@
...
+              proposals[PROPOSAL_REMOVE_MEMBER].defaults,
+              { ruleSettings: { [RULE_THRESHOLD]: { threshold: thresholdMemberRemoval } } }
+            ),
+            [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposals[PROPOSAL_PROPOSAL_SETTING_CHANGE].defaults,
+            [PROPOSAL_GENERIC]: proposals[PROPOSAL_GENERIC].defaults
+          }
+        }
+      })
+      await sbp('backend/publishLogEntry', entry)
+      return entry.hash()
```
sandrina-p: @taoeffect please review this part, is this what you suggested?
I don't like that much this approach because when creating a group, it's part of the "core logic" to switch to it. So, I think that part and the `contract/subscribeAndWait` should be part of it.
I understand why we did it - because of the tricky part with the callback - but still... we should look for a cleaner solution where the logic on the Vue components is as minimal as possible.

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,89 @@
...
+    } catch (e) {
+      throw new GIErrorUIRuntimeError(L('Failed to create the group: {codeError}', { codeError: e.message }))
+    }
+  },
+  // REVIEW these 2...
+  'gi.actions/group/createAndSubscribe': async function () {},
+  'gi.actions/group/createAndSubscribeWait': async function () {},
```
sandrina-p: Same question as above. Why not just creating a method `group/create` that by default it subscribes and waits. And if we really have a valid reason/case for each method, at least we could pass those as options.

```js
await sbp('gi.actions/group/create',
  groupData,
  { subscribe: true, wait: true } // <-- options as 2nd arg
)
```

sandrina-p (commented):
@taoeffect All your suggestions regarding `GIErrorUIRuntimeError` were addressed, as well as other misc stuff.

The only thing missing is the actions. Please check the new discussions for each part that I didn't fully understand. (Tip: read all of them before starting to answer back)

frontend/controller/actions/contract.js
```diff
@@ -0,0 +1,23 @@
+import sbp from '~/shared/sbp.js'
+import { CONTRACT_IS_SYNCING } from '@utils/events.js'
+
+export default sbp('sbp/selectors/register', {
+  'gi.actions/contract/subscribe': async function (contractIDs: string | array) {
```
taoeffect: Hmm, thinking on it more, I think it would be better to rename this to `gi.actions/contract/sync` instead, since this method can be called multiple times, including on contracts we're already subscribed to. Can you do a project-wide search and replace `gi.actions/contract/subscribe` with `gi.actions/contract/sync`?

frontend/controller/actions/contract.js
```diff
@@ -0,0 +1,23 @@
...
+  'gi.actions/contract/subscribeAndWait': async function (contractIDs: string | array) {
+    const listOfIds = typeof contractIDs === 'string' ? [contractIDs] : contractIDs
+
+    // QUESTION: Need help here... I didn't understand exactly the instructions to make
+    // this work when it comes to Promise.all.
+    for (const id of listOfIds) {
+      sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, id, true)
+      await sbp('gi.actions/contract/subscribe', id)
+      sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, id, false)
+    }
```
taoeffect: The issue with the above implementation is that it uses `emit` instead of `on`, and also it needs to be promise-based (see below for what I had in mind). However, looking more closely at how `'okTurtles.eventQueue/queueEvent'` is implemented, and how `'state/enqueueContractSync'` works with `syncContractWithServer`, I'm starting to think that `'gi.actions/contract/sync'` might already wait for all of the contracts to finish syncing before returning. If that is true, then `'gi.actions/contract/syncAndWait'` doesn't need to be implemented at all, because `'gi.actions/contract/sync'` would already wait for all contracts to finish syncing before returning.

I would like you to test with both selectors to find out if that is the case (using console logging to figure it out). Below is the implementation I had in mind for `'gi.actions/contract/syncAndWait'`, give it a try and see if it behaves differently from `'gi.actions/contract/sync'` (note, I haven't tested this code, so there might be bugs):

```js
// NOTE: the way this is implemented the function doesn't need to be marked 'async'
//       but you can still call 'await' on the result (since it's a promise)
function (contractIDs: string | array) {
  const promises = listOfIds.map(id => new Promise ((resolve, reject) => {
    const isDoneSyncing = (contractID, isSyncing) => {
      if (isSyncing === false && contractID === id) {
        sbp('okTurtles.events/off', CONTRACT_IS_SYNCING, isDoneSyncing)
        resolve(id)
      }
    }
    sbp('okTurtles.events/on', CONTRACT_IS_SYNCING, isDoneSyncing)
  }))
  for (const id of listOfIds) {
    sbp('state/enqueueContractSync', id)
  }
  return Promise.all(promises)
}
```

Once again, I'm not sure this is necessary, because I think the way `'gi.actions/contract/sync'` is implemented results in the same thing.

You will need to use console logging to verify whether indeed `'gi.actions/contract/sync'` waits for all contracts passed to it to finish syncing before returning. If so, then the above implementation of `'gi.actions/contract/syncAndWait'` isn't needed.

frontend/controller/actions/contract.js
taoeffect: 1. That's a good suggestion. In fact, I think the way `'gi.actions/contract/sync'` works might already do this! See my comment below.
2. Sure, I've provided the implementation I had in mind below. :)

taoeffect (changes requested):
Partial quick response review!

I'm not done leaving comments, but I've left you about 3 quick responses so to get back to you quickly. After posting this review I will continue where I left off and leave more comments.

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,89 @@
...
+            ),
+            [PROPOSAL_PROPOSAL_SETTING_CHANGE]: proposals[PROPOSAL_PROPOSAL_SETTING_CHANGE].defaults,
+            [PROPOSAL_GENERIC]: proposals[PROPOSAL_GENERIC].defaults
+          }
+        }
+      })
+      await sbp('backend/publishLogEntry', entry)
+      return entry.hash()
+    } catch (e) {
+      throw new GIErrorUIRuntimeError(L('Failed to create the group: {codeError}', { codeError: e.message }))
```
taoeffect, line 89: Right above this line we should also make sure to do a `console.error` log, so that we do not lose information about where in the code this error occurred (it will print out the stack trace).

So right after line 78, add:

```js
console.error('Failed to create group!', e)
```

frontend/controller/actions/group.js
taoeffect: Yes, this is what I had in mind. It's important that actions do only what their names say they will do. This will prevent programmer confusion resulting form unexpected behavior. Since the action is named `create`, it should only create the group, and do nothing else.

If you want to keep the logic out of the Vue components, that is totally fine, but the way to do that is to similarly create a differently named action that does what you want it to do.

So something like:

```js
  'gi.actions/group/createAndSyncAndSwitch': async function (groupData: Object) {
    const groupID = await sbp('gi.actions/group/create', groupData)
    // assuming 'gi.actions/contract/sync' and 'gi.actions/contract/syncAndWait'
    // do the same thing, we can just call 'gi.actions/contract/sync',
    // otherwise we would need to call 'gi.actions/contract/syncAndWait' here
    await sbp('gi.actions/contract/sync', groupID)
    sbp('gi.actions/group/switch', groupID)
  }
```

Note that if you implement `'gi.actions/group/createAndSync'`, then this selector should be implemented using that, so we have something like this instead:

```js
  'gi.actions/group/createAndSync': async function (groupData: Object) {
    const groupID = await sbp('gi.actions/group/create', groupData)
    // assuming 'gi.actions/contract/sync' and 'gi.actions/contract/syncAndWait'
    // do the same thing, we can just call 'gi.actions/contract/sync',
    // otherwise we would need to call 'gi.actions/contract/syncAndWait' here
    return sbp('gi.actions/contract/sync', groupID)
  },
  'gi.actions/group/createAndSyncAndSwitch': async function (groupData: Object) {
    const groupID = await sbp('gi.actions/group/createAndSync', groupData)
    sbp('gi.actions/group/switch', groupID)
  }
```

Note that here we assume that `'gi.actions/contract/sync'` does the same thing as the `'gi.actions/contract/syncAndWait'` implementation I gave in the other comment. You'll need to verify that it does first. If it doesn't, you'll need to replace it here with `'gi.actions/contract/syncAndWait'`.

Finally, note that doing it like this also means that the contract should be fully synced when we return from either `'gi.actions/group/createAndSync'` or `'gi.actions/group/createAndSyncAndSwitch'`, which means that the code you have below in `CreateGroup.vue` should be rewritten, i.e.:

```js
const groupID = await sbp('gi.actions/group/create', { ... })
sbp('okTurtles.events/once', groupID, (contractID, entry) => {
  sbp('gi.actions/group/switch', groupID)
  this.next()
})
await sbp('gi.actions/contract/subscribeAndWait', groupID)

// BECOMES:
const groupID = await sbp('gi.actions/group/createAndSyncAndSwitch', { ... })
this.next()

// OR:
const groupID = await sbp('gi.actions/group/createAndSync', { ... })
sbp('gi.actions/group/switch', groupID)
this.next()

// EDIT: OR, per your other suggestion below:
const groupID = await sbp('gi.actions/group/create', { ... }, { sync: true })
sbp('gi.actions/group/switch', groupID)
this.next()

// Note that `{ sync: true, switch: true }` doesn't work as a set of options
// because you cannot have `{ sync: false, switch: true }`. Or, you can allow
// it, but make sure to throw an exception if `switch: true` is set without `sync: true`
```

frontend/controller/actions/group.js
taoeffect: Yes, that's actually a much cleaner API! (with `subscribe` renamed to `sync` per the above comment)

I like it. 👍 

Only thing is, if, [per the above commentary](https://github.com/okTurtles/group-income-simple/pull/801/files#r368738345), you test and find out that the implementation of `sync` already results in a wait, then you don't need the `wait` parameter here.

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+  },
+  // REVIEW these 2...
+  'gi.actions/group/createAndSubscribe': async function () {},
+  'gi.actions/group/createAndSubscribeWait': async function () {},
+
+  'gi.actions/group/switch': async function (groupId) {
```
taoeffect: Vuex commits are not async (and all async functions should contain an `await` inside of them)

taoeffect (commented):
More commentary. Still not done but getting these out to you ASAP in case you come back online.

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,99 @@
+import sbp from '~/shared/sbp.js'
+import { GIErrorUIRuntimeError } from '@model/errors.js'
+import L from '@view-utils/translations.js'
+
+// import { CONTRACT_IS_SYNCING } from '@utils/events.js'
+
+export default sbp('sbp/selectors/register', {
+  'gi.actions/user/create': async function () {
+    // Is it here we do the signup? Shouldn't this be called user/signup for consistency?
+    // TODO - if yes, rename '/user/signup' to '/user/create'.
```
taoeffect: I think we want to have just `'gi.actions/user/create'`, in order to create a consistent API. In other words, each contract type will have a corresponding `/create` gi.action, and the only thing that will do is create the contracts (and, depending on whether or not we implement the options object, possibly sync them as well via `{ sync: true }`).

We should not do anything else besides that.

Which means:

- To login, we need to move that code into a `/login` action.
- To signup, `'namespace/register'` must be moved into a `/signup` action (which can call the create `/create` action inside of it). But the signup action should not log us in, because signing up a user doesn't imply they are logged in as well.
- We must not call `'gi.actions/user/logout'` in the `/create` action, because it has nothing to do with logging in or logging out

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,99 @@
...
+  },
+  // REVIEW these 2, same question as in group.js
```
taoeffect: Same answer. :)

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+      // TODO: Just add cryptographic magic
+      // login also calls 'state/enqueueContractSync', but not in this case since we
+      // just sync'd it.
+      await sbp('state/vuex/dispatch', 'login', {
+        username,
+        identityContractID: userHash
+      })
+
+      return userHash
```
taoeffect: Remember, we created two contracts, so in order to be able to call `gi.actions/contract/sync` on them, we must return an array here.

```js
return [ userHash, mailbox.hash() ]
```

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,99 @@
...
+    } catch (e) {
+      sbp('gi.actions/user/logout')
```
taoeffect: 1. I'm not sure if this code should be in the `/signup` action... this will need some thought. Maybe add a `// TODO: should this be here?` to it.
2. Per the comment above, make sure not to move it into the `/create` action
3. Don't forget to add here (so we don't lose the stacktrace):
   ```
    console.error(`Error signing up user ${username}:`, e)
   ```

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,99 @@
...
+    if (!identityContractID) {
+      throw new GIErrorUIRuntimeError(L('Invalid username or password'))
+    }
+
+    try {
+      console.debug(`Retrieved identity ${identityContractID}`)
+      await sbp('state/vuex/dispatch', 'login', { username, identityContractID })
+
+      return identityContractID
+    } catch (e) {
```
taoeffect, line 93: Don't forget to add a `console.error` line here to keep the stacktrace (see similar comments for example)

frontend/main.js
```diff
@@ -90,7 +91,7 @@ async function startApp () {
       return {
         ephemeral: {
           syncs: [],
-          finishedLogin: 'no'
+          finishedLogin: this.$store.state.loggedIn ? 'yes' : 'no'
```
taoeffect: I think this is a potentially troublesome change, and would recommend keeping it the way it was...

frontend/views/containers/modals/CreateGroup.vue
```diff
@@ -107,62 +104,27 @@ export default {
...
-        sbp('okTurtles.events/once', hash, (contractID, entry) => {
-          this.$store.commit('setCurrentGroupId', hash)
+        sbp('okTurtles.events/once', groupID, (contractID, entry) => {
+          sbp('gi.actions/group/switch', groupID)
           this.next()
         })
-        await sbp('backend/publishLogEntry', entry)
-        // add to vuex and monitor this contract for updates
-        await sbp('state/enqueueContractSync', hash)
+        await sbp('gi.actions/contract/subscribeAndWait', groupID)
```
taoeffect: See comment above for the change to this

frontend/views/pages/BypassUI.vue
```diff
@@ -0,0 +1,150 @@
...
+      return false
+    }
+
+    const { actionFn } = this.actions[action] || {}
+
+    if (!actionFn) {
+      // Wait for $refs.bannerAction to exist
+      this.$nextTick(() => {
+        this.$refs.bannerAction.danger(`Action ${action} doesn't exist.`)
+      })
```
taoeffect, line 69: If it doesn't exist then we shouldn't proceed below to call it

taoeffect (changes requested):
OK, I *think* (am not sure), but I think my review is finished for now.

frontend/main.js
sandrina-p: Hum... reverted but left a comment to review this later.

frontend/views/pages/BypassUI.vue
sandrina-p: That's why we use $nextTick. To make sure the banner is already mounted.

frontend/controller/actions/contract.js
sandrina-p: If `sync` and `syncAndWait` resulted in the same thing, Cypress tests wouldn't fail... they fail because they move on to the next step before the contracts finish syncing.

frontend/controller/actions/contract.js
sandrina-p: Renamed all `/contract/subscribe` to `/contract/sync`

frontend/controller/actions/contract.js
```diff
@@ -0,0 +1,28 @@
...
+    const promises = listOfIds.map(id => new Promise((resolve, reject) => {
+      const isDoneSyncing = (contractID, isSyncing) => {
+        if (isSyncing === false && contractID === id) {
+          sbp('okTurtles.events/off', CONTRACT_IS_SYNCING, isDoneSyncing)
+          resolve(id)
+        }
+      }
+      sbp('okTurtles.events/on', CONTRACT_IS_SYNCING, isDoneSyncing)
+      sbp('state/enqueueContractSync', id)
+    }))
```
sandrina-p, line 24: @taoeffect I think there was a bug on your Promise solution and I fixed it. The bug fix is we should enqueueContractSync inside the promise, not before. Otherwise, the contract could finish syncing even before creating the promise, and then the promise would never resolve. That bug happened on signup, but on group creation it didn't happen. Why?


Still... this solution didn't fix our core problem here: BypassUI (and any other action) is still returning (resolving) too soon. We must wait for the server to sync and that's not happening dunno why :/

frontend/controller/actions/user.js
sandrina-p: Okay, so what's done so far: 
- `/create` - it creates a user and registers it.
- `/signup` - calls `/create` and signups, it doesn't login. I kept the logout with the TODO as you suggested
- `/login` only logins.
- `/signupAndLogin` - concat `/signup` and `/login`.
- The last 3 methods have a 2nd parameter `sync` that is true by default which will sync the respective contracts (and it should wait too, but that's the only part not working yet...)

sandrina-p (commented):
@taoeffect please check these comments:

frontend/controller/actions/contract.js
taoeffect: @sandrina-p there are two bugs in this code that work together to break things:

1. This function is labeled `async` even though it shouldn't be. Remove the `async` qualifier since that should only ever be used when there is a corresponding `await` (frankly I don't understand why our linter didn't complain)
2. Line 26 below needs to read `return Promise.all(promises)`. Right now a promise is being returned (because of the `async` qualifier), but not the one we want

sandrina-p: hum... but it is an asynchronous function because of Promise. I [followed this tutorial](https://www.taniarascia.com/promise-all-with-async-await/). And if I remove it, then the signup stops working. Maybe it would be better you could checkout this branch and see it locally

taoeffect: > And if I remove it, then the signup stops working. 

You must do both things, remove it and fix line 26, otherwise it won't work. The rule is that `async` functions must have `await`s in them. If they do not, then the function must not have an `async` on it. Return a promise and you can still call await on it. You'll notice everything I'm saying here is validated by that tutorial.

taoeffect: Writing `async` is equivalent to wrapping everything in the function in a promise and returning that promise. So by having `async` here the wrong promise is being returned. It needs to be the one on line 26

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,102 @@
...
+        userID
+      )
+      await sbp('backend/publishLogEntry', attribute)
+      await sbp('namespace/register', username, userID)
+
+      const contracts = [userID, mailboxID]
+
+      if (!sync) { return contracts }
+      await sbp('gi.actions/contract/syncAndWait', contracts)
+      return contracts
```
taoeffect: Minor style, can be written to use a single return:

```js
if (sync) {
  await sbp('gi.actions/contract/syncAndWait', contracts)
}
return contracts
```

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,99 @@
...
+        }
+      })
+      const groupId = entry.hash()
+
+      await sbp('backend/publishLogEntry', entry)
+
+      if (!sync) { return groupId }
+
+      await sbp('gi.actions/contract/syncAndWait', groupId)
+      return groupId
```
taoeffect: Minor style: can be rewritten without redundant return:

```js
if (sync) {
  await sbp('gi.actions/contract/syncAndWait', groupId)
}
return groupId
```

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,99 @@
...
+    } catch (e) {
+      console.error('gi.actions/group/create failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to create the group: {codeError}', { codeError: e.message }))
+    }
+  },
+  'gi.actions/group/createAndSwitch': async function (groupParams, { sync = true } = {}) {
```
taoeffect: This function shouldn't take an options object with `sync`, since in order to switch we must always wait for the sync.

frontend/controller/actions/group.js
```diff
@@ -0,0 +1,99 @@
...
+    const groupID = await sbp('gi.actions/group/create', groupParams)
```
taoeffect: Similarly, here we it would be safer to pass in an options object with `{ sync: true }`, so that we're explicit about it, in case the default ever change

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,102 @@
...
+  } = {}) {
+    try {
+      const [userID, mailboxID] = await sbp('gi.actions/user/create', { username, email, password })
+
+      // set the attribute *after* publishing the identity contract
+      const attribute = await sbp('gi.contracts/identity/setAttributes/create',
+        { mailbox: mailboxID },
+        userID
+      )
+      await sbp('backend/publishLogEntry', attribute)
```
taoeffect: This section (lines 49-54), needs to be moved up into the `/create` action, right after line 35, since this is part of the required setup that links the identity contract to the mailbox.

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,102 @@
...
+  } = {}) {
+    // TODO: Insert cryptography here
+    const userId = await sbp('namespace/lookup', username)
+    if (!userId) {
+      throw new GIErrorUIRuntimeError(L('Invalid username or password'))
+    }
+
+    try {
+      console.debug(`Retrieved identity ${userId}`)
+      await sbp('state/vuex/dispatch', 'login', { username, identityContractID: userId })
```
taoeffect, line 86: Above this line, add:

```
// TODO: move the login vuex action code into this function (see #804)
```

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,102 @@
...
+      return userId
+    } catch (e) {
+      console.error('gi.actions/user/login failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to login: {codeError}', { codeError: e.message }))
+    }
+  },
+  'gi.actions/user/signupAndLogin': async function (signupParams) {
+    const [userId, mailboxId] = await sbp('gi.actions/user/signup', signupParams)
+    await sbp('gi.actions/user/login', signupParams)
+    return [userId, mailboxId]
```
taoeffect: Since neither `userId` nor `mailboxId` are used in this function, these don't need to be destructured and can simply be passed back as

```
return contractIDs
```

frontend/controller/actions/user.js
```diff
@@ -0,0 +1,102 @@
...
+  },
+
+  'gi.actions/user/logout': async function () {
+    await sbp('state/vuex/dispatch', 'logout')
```
taoeffect, line 106: Above this add:

```
// TODO: move the logout vuex action code into this function (see #804)
```

frontend/views/pages/BypassUI.vue
taoeffect: Sorry, I wasn't clear, I meant if `actionFn` doesn't exist then we should `return` here and not proceed to line 67.

taoeffect (changes requested):
See comments! ^_^

frontend/controller/actions/contract.js
sandrina-p: (Skip to edit part!)
Hum... still, this isn't working and I'm struggling to understand why :/
When we call `/signupAndLogin` on signup modal, the 2nd call `/login` inside it isn't resolved, preventing the form to finalize the submit (in this case, to close the modal).

Based on the logs, it seems we listen for `CONTRACT_IS_SYNCING` but it's never called during the login.

But the strange thing is that the login happens as you can see in this screenshot:
image

--- EDIT ---

I figured it out! (so happy!) The bug was at state action `syncContractWithServer`! The event `CONTRACT_IS_SYNCING` was only being emitted if the contract hash was different from the latest. We need to emit it always, even when the same. In this case, only "false" is enough. That way we can listen for contractSync even if the sync is the same.

test/cypress/support/commands.js
```diff
@@ -16,46 +16,98 @@ Cypress.Commands.add('getByDT', (element, otherSelector = '') => {
...
+  // cy.getByDT('app').then(([el]) => {
+  //   cy.get(el).should('have.attr', 'data-sync', '')
+  // })
+  // cy.visit(`/app/bypass-ui?action=${action}${query}`)
+
+  // Option 2 - hardcore wait. it works, but it isn't the ideal solution.
+  // cy.wait(400); // eslint-disable-line
+  // cy.visit(`/app/bypass-ui?action=${action}${query}`)
+
+  // Option 3 - Navigate to /bypass-ui using Vue instead of cy.visit!
```
sandrina-p: @taoeffect  and @pieer please read this!

#### A recap of this PR journey:

We have been trying to use `/bypass-ui` page to speed up our tests. The only issue is that cypress is so fast visiting `/bypass-ui` that sometimes information gets lost.

So we thought it was some problem with our code (contracts not fully synced)... but after all... it's not! Why do I say this? Because after cypress fails to load some data (ex: profile info) and I manually refresh it a second time, the information is there again! So this means, the contracts were indeed synced. It has to be something else... maybe the way how `forageStorage` (indexedDB) works?

#### A new approach:
I had an idea to overcome this page load issue on cypress: Use Vue to navigate to `/bypass-ui` instead of cypress visit command, to avoid a full page refresh (Read Option 3 explanation).

Now when we run cypress, a little button shows up on the corner, all the time. It's our "magic gate" to `/bypass-ui` page.

Please let me know if some further explanation about this is needed. I'll add it directly to this code.

image

test/cypress/integration/group-payments.spec.js
```diff
@@ -109,8 +109,6 @@ describe('Payments', () => {
   })
 
   it('user1 decides to switch income details to needing $100', () => {
-    cy.visit('/app/contributions')
-
```
sandrina-p: NOTE: There was no need for this. we are already at /contributions.

frontend/controller/actions/group.js
taoeffect: I can see this was resolved in a few other areas but the lines above were missed!

frontend/controller/actions/contract.js
taoeffect: Congrats on figuring it out! 🙌 

~~Since that was the issue, can you test again with `'gi.actions/contract/sync'` and see if that has the same effect as `'gi.actions/contract/syncAndWait'`?~~

Oops, just realizing that this change you made shouldn't have any effect on that, so probably not worth the trouble to do that.

test/cypress/support/commands.js
taoeffect: That's an interesting way of doing it. Avoiding the need for a `.visit` might also speed things up too, right? If it works, it works, I see no major issue of doing it this way at least.

frontend/main.js
```diff
@@ -122,6 +124,7 @@ async function startApp () {
     },
     computed: {
       showNav () {
+        console.log(':: groupsByName', this.$store.getters.groupsByName)
```
taoeffect: Should this be here? Either should be removed before being merged, or reasoning should be given for why this is being logged. Also, this is debug logging, so `console.debug` should be used

frontend/model/state.js
```diff
@@ -361,8 +362,10 @@ const actions = {
         // this must be called directly, instead of via enqueueHandleEvent
         await dispatch('handleEvent', GIMessage.deserialize(events[i]))
       }
-      sbp('okTurtles.events/emit', EVENTS.CONTRACT_IS_SYNCING, contractID, false)
+    } else {
+      console.log(`Contract ${contractID} was already synchronized`)
```
taoeffect: `console.debug`

frontend/views/pages/Home.vue
```diff
@@ -80,6 +80,9 @@ export default {
     SvgCreateGroup
   },
   mounted () {
+    console.log('::groupId:', this.$store.state.currentGroupId)
+    console.log('::groupsByName:', this.$store.getters.groupsByName)
```
taoeffect: Should be deleted unless needed. If needed, reasoning needed, and remember `console.debug`.

frontend/controller/actions/group.js
sandrina-p: ups, was the only one missing. changed

test/cypress/support/commands.js
sandrina-p: yes, it's slightly faster because we don't wait for the page to load all the stuff again.

frontend/main.js
sandrina-p: was only for debugging while working on this. removed

frontend/model/state.js
sandrina-p: If this is `debug`, should the other logs on this fn also be renamed to `debug`?

frontend/views/pages/Home.vue
sandrina-p: removed

frontend/controller/actions/user.js
taoeffect: Don't forget about this! ^_^

test/cypress/support/commands.js
```diff
@@ -165,11 +224,21 @@ Cypress.Commands.add('giAcceptGroupInvite', (invitationLink, {
...
+  actionBeforeLogout,
+  bypassUI
 }) => {
+  if (bypassUI && !isLoggedIn) {
+    // TODO - Create bypass action to join group. Until then
+    // do the other way around: signup before visiting the page.
+    // Then, the invitation is automatically accepted.
+    cy.giSignup(username, { bypassUI })
+    // use an hardcoded wait so indexedDB is synced before cy.visit()
+    cy.wait(400) // eslint-disable-line
```
taoeffect: I thought we no longer need hardcoded waits...?

frontend/controller/actions/user.js
sandrina-p: done!

test/cypress/support/commands.js
sandrina-p: We need it here because we do cy.visit() immediately after a bypass-ui action.

But nevermind, I removed this code to avoid confusion. We'll do the "join group" bypass-ui action in a proper way (without wait) on another MR. This was just a middle solution but it's not worth the confusion.

taoeffect (approved):
Stupendous, colossal, superb, spectacular job @sandrina-p! 😄 🏆

Issue #792, opened 2019-12-18
Finish Group Settings page


sandrina-p:
### Problem

The **Group Settings** page has an incomplete layout and it's not functional yet.

### Solution
- [x] 1. Update the page following the [design specs](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=123%3A1378)
- [x] 2. Group Profile (first card) - add logic
- [x] 3. Leave Group - add logic
- [ ] 4. ~Delete Group - add logic~ not needed for the prototype

sandrina-p:
I'll pick this issue and address points 1 and 2.

sandrina-p:
While developing #823, I realized something:

It doesn't make sense to show "Leave group" and "Delete Group" at the same time. We can only Leave the group if there are more members. And we can only delete the group if we are the only member.

@mmbotelho, can you review the UX of this part of the page?

mmbotelho:
@sandrina-p done!

https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=9720%3A0

Issue #802, opened 2020-01-20
Add loading state to form submission


sandrina-p:
### Problem

When submitting a form, there's no indication about what's happening (a loading state).

So far we never needed (or remembered them) because all the actions are done locally, so no network isn't required and everything is instantly done, but in the real world, these things will take time. (ex: login or group creation).

### Solution

Design (@mmbotelho) and implement the loading state where applicable.

taoeffect:
I don't think we need to make buttons themselves display a loading state beyond being disabled. The loading state can appear as a message and/or progress bar at the bottom or top of the window.

sandrina-p:
Okay, let's leave that for @mmbotelho. But just a note: using `disabled` (as HTML attribute) while the button is loading can be a bad practice because it hides the button from keyboard/screen readers, and that's confusing. In that scenario, a better approach is to use `aria-busy` or `aria-label="loading"` and then prevent the action with JS.

mmbotelho:
> I don't think we need to make buttons themselves display a loading state beyond being disabled. The loading state can appear as a message and/or progress bar at the bottom or top of the window.

@taoeffect the best practice in these situations is to always give feedback to users as close as possible to whatever element triggered it. For this reason, I do think showing a loader in the button itself is the best option and the one I would recommend.

[I made a very simple codepen with how I think the button should behave](https://codepen.io/mmbotelho/pen/KKpwQzP). Let me know what you think! 

@sandrina-p @pieer

sandrina-p:
@sandrina-p This UX is very much what I had in mind, I approve 👍

taoeffect:
@mmbotelho I like that, it looks great! 😄 👍

mmbotelho:
Ok, I'm removing myself from this issue, it's ready to be picked up by either @sandrina-p or @pieer for the implementation!

PR #817, opened 2020-01-31
Group Settings - implement "save changes".

sandrina-p:
Related to #792.

image

- Implement pixel perfect the "Group Settings" page, including the modals to leave/delete the group.
- Implement logic to update group settings. (profile, name, about and currency). Keep in mind that #658 wasn't done yet.
- `AvatarUpload` component created - it centralizes the layout and upload logic for both user and group profiles.
- Added necessary Cypress tests.

### scout improvements:
- "Invite links" card: the link "propose adding a new member" is now working correctly.
- Secondary/right sidebar is fixed when scrolling. (it wasn't for some reason!)
- `Banner` component - now it looks pretty when it has 1 or multiple lines. Before it was "broken" when it had just one line.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -13,6 +13,7 @@ import * as Errors from '../errors.js'
 import { merge, deepEqualJSONType } from '~/frontend/utils/giLodash.js'
 import { currentMonthTimestamp } from '~/frontend/utils/time.js'
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
+import currencies from '@view-utils/currencies.js'
 
 export const INVITE_INITIAL_CREATOR = 'INVITE_INITIAL_CREATOR'
 export const INVITE_STATUS = {
@@ -311,8 +312,11 @@ DefineContract({
       // OPTIMIZE: Make this custom validation function
       // reusable accross other future validators
       validate: objectMaybeOf({
+        groupName: x => typeof x === 'string',
+        groupPicture: x => typeof x === 'string',
+        sharedValues: x => typeof x === 'string',
         mincomeAmount: x => typeof x === 'number' && x > 0,
-        mincomeCurrency: x => typeof x === 'string'
+        mincomeCurrency: x => Object.keys(currencies).includes(x)
       }),
       process (state, { data }) {
         for (var key in data) {
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -13,7 +13,7 @@ import * as Errors from '../errors.js'
 import { merge, deepEqualJSONType } from '~/frontend/utils/giLodash.js'
 import { currentMonthTimestamp } from '~/frontend/utils/time.js'
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
-import currencies from '@view-utils/currencies.js'
+import currencies from '~/frontend/views/utils/currencies.js'
 
 export const INVITE_INITIAL_CREATOR = 'INVITE_INITIAL_CREATOR'
 export const INVITE_STATUS = {
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -13,7 +13,6 @@ import * as Errors from '../errors.js'
 import { merge, deepEqualJSONType } from '~/frontend/utils/giLodash.js'
 import { currentMonthTimestamp } from '~/frontend/utils/time.js'
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
-import currencies from '~/frontend/views/utils/currencies.js'
 
 export const INVITE_INITIAL_CREATOR = 'INVITE_INITIAL_CREATOR'
 export const INVITE_STATUS = {
@@ -316,7 +315,7 @@ DefineContract({
         groupPicture: x => typeof x === 'string',
         sharedValues: x => typeof x === 'string',
         mincomeAmount: x => typeof x === 'number' && x > 0,
-        mincomeCurrency: x => Object.keys(currencies).includes(x)
+        mincomeCurrency: x => typeof x === 'string'
       }),
       process (state, { data }) {
         for (var key in data) {
--- a/frontend/views/components/AvatarUpload.vue
+++ b/frontend/views/components/AvatarUpload.vue
@@ -4,7 +4,7 @@
     .c-avatar-wrapper
       label.c-avatar-label
         avatar.c-avatar-img(
-          :src='form.picture'
+          :src='avatar'
           ref='picture'
         )
         i18n.link.c-avatar-text Change avatar
@@ -29,9 +29,12 @@ import L from '@view-utils/translations.js'
 export default {
   name: 'AvatarUpload',
   props: {
-    variant: {
+    avatar: {
       type: String,
-      validator: (value) => ['user', 'group'].includes(value),
+      required: true
+    },
+    sbpParams: {
+      type: Object,
       required: true
     }
   },
@@ -40,34 +43,10 @@ export default {
     BannerScoped
   },
   data () {
-    const picture = {
-      user: this.$store.getters.ourUserIdentityContract.attributes.picture,
-      group: this.$store.getters.groupSettings.groupPicture
-    }[this.variant]
-
     return {
-      form: {
-        picture
-      },
       ephemeral: {
         isSubmitting: false
       }
-    }
-  },
-  computed: {
-    sbpParams () {
-      return {
-        user: {
-          selector: 'gi.contracts/identity/setAttributes/create',
-          contractID: this.$store.state.loggedIn.identityContractID,
-          key: 'picture'
-        },
-        group: {
-          selector: 'gi.contracts/group/updateSettings/create',
-          contractID: this.$store.state.currentGroupId,
-          key: 'groupPicture'
-        }
-      }[this.variant]
     }
   },
   methods: {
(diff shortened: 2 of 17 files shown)


frontend/views/pages/GroupSettings.vue
```diff
@@ -1,147 +1,199 @@
...
-          name='groupName'
-          :class='{ error: $v.form.groupName.$error }'
-          :value='groupSettings.groupName'
-          @input='update'
-          @keyup.enter='next'
+          :class='{error: $v.form.groupName.$error}'
+          v-model='form.groupName'
+          @input='debounceField("groupName")'
+          @blur='updateField("groupName")'
+          v-error:groupName=''
```
sandrina-p: cc @pieer and @taoeffect 

Note: These 5 lines are needed in every input that needs validation besides "required". So far we have it in >10 places around...

What do you think of creating a directive that handles it for us? What name do you suggest?
- `v-form:groupName=''`
- `v-model-debounce:groupName=''`
- `v-validate:groupName=''`
- _what else?_

If you agree I can create an issue for that.

sandrina-p, line 21: cc @pieer and @taoeffect 

Note: These 5 lines are needed in every input that needs validation besides "required". So far we have it in >10 places...

What do you think of creating a directive that handles it for us? What name do you suggest?
- `v-form:groupName=''`
- `v-model-debounce:groupName=''`
- `v-validate:groupName=''`
- _what else?_

If you agree I can create an issue for that.

taoeffect: I think, since it's so specific to what we're doing, a name like `v-gi-form:groupName=''` or something similar would make sense. Minor issue, but yeah, feel free to create an issue if you'd like

frontend/model/contracts/group.js
```diff
@@ -311,8 +312,11 @@ DefineContract({
       // OPTIMIZE: Make this custom validation function
       // reusable accross other future validators
       validate: objectMaybeOf({
+        groupName: x => typeof x === 'string',
+        groupPicture: x => typeof x === 'string',
+        sharedValues: x => typeof x === 'string',
         mincomeAmount: x => typeof x === 'number' && x > 0,
-        mincomeCurrency: x => typeof x === 'string'
+        mincomeCurrency: x => Object.keys(currencies).includes(x)
```
taoeffect: This is an `O(n)` operation. You can keep it `O(1)` by doing something along the lines of

```js
x in currencies
```

However, we probably want to keep it simply checking for `typeof 'string'` to be safe since we're not sure how client updates will affect things. Imagine some clients getting updated with a new currencies list, and some old clients still running an old version. If we go with the stricter approach above, the old clients wouldn't accept the new currencies from the new clients, and would initiate an autoban.

sandrina-p: Hum... I don't think that way of handling "versions" is well structure. The way I see it, our code should be retro compatible. So version 1 does:
```
mincomeCurrency: x => !!currencies[x]
```
and version 2 does:

```
mincomeCurrency: x => !!currencies[x] || !!oldCurrencies[x]
```

And so on until we start to deprecate version and "force" users to update their version.

But I'm not an expert on that, so I reverted the change and kept the type of validation only.

frontend/views/pages/GroupSettings.vue
sandrina-p: issue created #819

frontend/views/containers/group-settings/GroupLeaveModal.vue
```diff
@@ -79,82 +48,80 @@
...
       }
     }
   },
+  computed: {
+    ...mapGetters([
+      'groupSettings',
+      'ourUsername'
+    ]),
+    code () {
+      return `LEAVE ${this.groupSettings.groupName.toUpperCase()}`
```
taoeffect: Needs to use `L` function/string

frontend/views/containers/group-settings/GroupDeletionModal.vue
```diff
@@ -1,106 +1,101 @@
...
-          console.log(value)
-          return value === 'Delete The Dreamers'
-        }
-      }
+  computed: {
+    ...mapGetters([
+      'groupSettings'
+    ]),
+    code () {
+      return `DELETE ${this.groupSettings.groupName.toUpperCase()}`
```
taoeffect: `L`

frontend/views/components/AvatarUpload.vue
```diff
@@ -0,0 +1,165 @@
...
+  },
+  components: {
+    Avatar,
+    BannerScoped
+  },
+  data () {
+    const picture = {
+      user: this.$store.getters.ourUserIdentityContract.attributes.picture,
+      group: this.$store.getters.groupSettings.groupPicture
+    }[this.variant]
```
taoeffect: So this AvatarUpload is a neat way to DRY things, however, I see two issues with this specific design:

1. Using `data()` means these images will not be reactive and won't update for other clients when changed.
2. Tying the design of a generic component like this to Group Income's internals defeats the purpose of reusable components, and means that anytime we want to use this component for other types of image uploads, the component itself has to be modified. It should be designed in a way that is completely agnostic to internals, as if you were designing a pluggable library that can be used in any app

sandrina-p: 1. That's not true... It's being used the exact same logic as before, where we replace data value on `fileChange` using `setFromBlob`. I just doubled-check it and it works fine across clients.
2. I understand your concern and that was my first approach. The component accepting 2 props:
- `sbpParams` - Object `{ selector, contractID, key }`
- `avatar` - String - Vue "path" to the current picture.
But then I decided to centralize that info in one single place (computed property `sbpParams`). But anyway, I think that's indeed a better approach. I'll change it back.

mmbotelho (dismissed):
Perfect @sandrina! Approved!

frontend/views/containers/user-settings/UserProfile.vue
```diff
@@ -2,25 +2,10 @@
...
-            name='profilePicture'
-            accept='image/*'
-            @change='fileChange($event.target.files)'
-            placeholder='http://'
-            data-test='profilePicture'
-          )
-        // TODO #658
-        banner-scoped.c-pictureMsg(ref='pictureMsg' data-test='pictureMsg')
+    avatar-upload(
+      :avatar='$store.getters.ourUserIdentityContract.attributes.picture'
```
taoeffect: Instead of `$store.getters.ourUserIdentityContract.attributes.picture`, which might result in an exception because `.attributes` could return undefined, use `attributes.picture` (there's already a computed property in this Vue component that will return an empty object if `.attributes` is undefined)

frontend/views/pages/GroupSettings.vue
```diff
@@ -1,147 +1,209 @@
...
+      i18n.is-danger.is-outlined(
+        tag='button'
+        ref='delete'
+        @click='openProposal("GroupDeletionModal")'
+        data-test='deleteBtn'
+      ) Delete group
+
+    banner-simple(severity='info' v-else)
+      i18n(
+        :args='{ count: membersLeft, groupName: groupSettings.groupNname, ...LTags("b")}'
```
taoeffect: typo: `groupSettings.groupNname` => `groupSettings.groupName`

frontend/views/containers/access/LoginForm.vue
```diff
@@ -67,6 +70,8 @@ export default {
   },
   methods: {
     async login () {
+      if (this.ephemeral.isSubmitting) { return }
+      this.ephemeral.isSubmitting = true
```
taoeffect, line 74: Is it possible to disable the submit button using vuelidate? Something like this (copied from elsewhere in this PR):

```
:disabled='$v.form.$invalid'
```

?

taoeffect (changes requested):
Great job @sandrina-p! Only minor changes requested. After those are addressed, and if @pieer finds nothing, I think this is GTM.

frontend/views/containers/access/LoginForm.vue
sandrina-p: Hum, but we already do that. (and [I don't agree](https://axesslab.com/disabled-buttons-suck/) with it)

taoeffect: I mean disabling the button after it is pressed. I skimmed that link just now and it seems to be referring to buttons that are disabled initially, rather than buttons that are initially enabled, and disabled when clicked, to indicate that an action has taken place, and that the user needs to wait for it to complete before clicking it again.

The concern in that link is that clicking the button does nothing. And yet that is exactly what would happen in this scenario above, where the button is enabled, but clicking it multiple times does nothing if it takes a while to perform the submit action.

sandrina-p: > The concern in that link is that clicking the button does nothing. 

That's why we need visual loading states on this scenarios as per #802.

So, I'll change to `:disabled='$v.form.$invalid || ephemeral.isSubmitting'` does that sound ok to you?

sandrina-p: But still, if we press "Enter" instead of clicking the button, it will attempt to submit the form again, so we still need the line `if (this.ephemeral.isSubmitting) { return }`. In other words, doing what I suggested in the last comment is not useful.

sandrina-p: I would leave it as it is and wait for @mmbotelho  solution on #802.

taoeffect: OK, all of that sounds reasonable 👍

sandrina-p: Cool! Resolving this conversation, since there's nothing left to change.

pieer (approved):
All good to me!

mmbotelho (approved):

taoeffect (approved):
Great job @sandrina-p!

PR #818, opened 2020-02-01
Task / 561 Implement the "Group members" page #809

pieer:


mmbotelho (changes requested):
This is looking great! I noticed a couple of details:

- It's missing a "clear field" button (see images below);
- The number of results and query entered should be highlighted (see image below);
- When there are 0 results, can we change the copy to "Sorry, we couldn't find anyone called [xyz]."

Captura de ecrã 2020-02-04, às 17 33 50
Captura de ecrã 2020-02-04, às 17 34 33

Other than that, it's looking great! 😄

frontend/views/containers/dashboard/GroupMemberMenu.vue
```diff
@@ -0,0 +1,51 @@
...
+      )
+        i18n Remove member
+</template>
+
+<script>
+import { MenuParent, MenuTrigger, MenuContent, MenuItem } from '@components/menu/index.js'
+import { OPEN_MODAL } from '@utils/events.js'
+import sbp from '~/shared/sbp.js'
+export default {
+  name: 'GroupMembers',
```
sandrina-p: `GroupMembersMenu`

frontend/views/containers/dashboard/GroupMemberMenu.vue
```diff
@@ -0,0 +1,51 @@
...
+      this.$refs.modal.close()
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+.c-content {
+  width: 200px;
+  margin-left: -182px;
```
sandrina-p: use rem here on both 2 rules.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,227 @@
...
+  .c-container
+    .c-header
+      i18n.is-title-2.c-title(tag='h2') Group members
+
+    .card.c-card
+      form(
+        @submit.prevent='submit'
+      )
+        label.field
+          .input-combo
```
sandrina-p, line 13: An input should always have a label, even when invisible, for screen readers.
```
i18n.sr-only Search for a member
```

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,227 @@
...
+  i {
+    margin-right: .5rem;
+  }
+
+  @include tablet {
+    display: block;
+  }
+}
+
+.c-menu {
```
sandrina-p: Where's this class used? Seems lost.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,227 @@
...
+
+        i18n.c-member-count.has-text-1(
+          v-else
+          tag='div'
+          :args='{ groupMembersCount }'
+        ) {groupMembersCount} members
+
+      table.table.c-table(v-if='searchResult')
+        transition-group.c-group-list(name='slide-list' tag='tbody')
+          tr.c-group-member(
```
sandrina-p: This for me is a list of results, not a table. In terms on A11Y there are some issues here (ex: using `display: flex` in a table element removes its semantics (the same way we use `display: none` to hide something))

So, I would advise to just replace `table` for `ul li` and everything gets easier.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,227 @@
...
+}
+
+.c-menu {
+  @include tablet {
+    display: none;
+  }
+}
+
+.button + .button {
+  margin-left: .5rem;
```
sandrina-p: we have the class `.buttons` for this same purpose. In this case we just need to remove the `margin-top` that comes from it.

sandrina-p (commented):
Besides the table thing, I didn't notice any other major issue.

mmbotelho (changes requested):
Captura de ecrã 2020-02-05, às 13 07 17

I thought I had specified the interaction for the search input with the clear button but it wasn't on Figma, my mistake! I added it there, here's how it should look: 
Captura de ecrã 2020-02-05, às 13 09 58

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,255 @@
...
+    searchResult () {
+      return Object.keys(this.groupProfiles)
+        .map(username => {
+          const inList = (n) => n.toUpperCase().indexOf(this.form.search.toUpperCase()) > -1
+          const { name, displayName } = this.globalProfile(username)
+          if (!this.form.search || inList(name) || (displayName ? inList(displayName) : false)) {
+            return { username, name, displayName }
+          }
+        })
+        .filter(profile => profile, [])
```
sandrina-p: What's the 2nd parameter `[]` for?

sandrina-p (dismissed):
looks good, great work @pieer!

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,255 @@
...
+              aria-label='Clear search'
+              @click='form.search = null'
+            )
+              i.icon-times
+
+        i18n.c-member-count.has-text-1(
+          v-if='form.search && searchCount > 0'
+          tag='div'
+          :args='{ searchCount: `<strong>${searchCount}</strong>`, result: `<strong>${searchCount === 1 ? L("result") : L("results")}</strong>`, searchTerm: `<strong>${form.search}</strong>`}'
+          compile
```
taoeffect, line 35: If the HTML you're using doesn't contain any Vue-specific components (as is the case here and below on line 41), you don't need to add the `compile` to the attributes. It should render correctly and faster without it.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,255 @@
...
+        ) Showing {searchCount} {result} for "{searchTerm}"
```
taoeffect, line 36: You did a great job constructing this string a really readable and translator-friendly manner 👍

frontend/views/containers/dashboard/GroupMemberMenu.vue
```diff
@@ -0,0 +1,51 @@
...
+    i.icon-ellipsis-v
+
+  menu-content.c-actions-content
+    ul
+      menu-item(
+        tag='router-link' to='/chat'
+        item-id='message'
+        icon='comment'
+      )
+        i18n Send message
```
taoeffect, line 13: Strangely, clicking this does nothing?

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,255 @@
...
+import sbp from '~/shared/sbp.js'
+import { mapGetters } from 'vuex'
+import { OPEN_MODAL } from '@utils/events.js'
 import ModalBaseTemplate from '@components/modal/ModalBaseTemplate.vue'
+import UserImage from '@components/UserImage.vue'
+import GroupMemberMenu from '@containers/dashboard/GroupMemberMenu.vue'
 
 export default {
-  name: 'GroupMembersAllModal',
+  name: 'IncomeDetails',
```
taoeffect: The name of this component should be the same as the file

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,255 @@
...
+      'globalProfile',
+      'groupMembersCount'
+    ]),
+    searchResult () {
+      return Object.keys(this.groupProfiles)
+        .map(username => {
+          const inList = (n) => n.toUpperCase().indexOf(this.form.search.toUpperCase()) > -1
+          const { name, displayName } = this.globalProfile(username)
+          if (!this.form.search || inList(name) || (displayName ? inList(displayName) : false)) {
+            return { username, name, displayName }
```
taoeffect: Note that `name` is the same as `username`, so it can and probably should be ignored in favor of `username`.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
taoeffect: Yeah, I don't think the `[]` is needed there. It's also not super clear to me what this `.filter` is doing..? Maybe the condition should be something like `!!profile` or `profile !== undefined` to be a little clearer?

taoeffect (changes requested):
Nice job @pieer, left some comments. I wasn't able to finish my review because I couldn't figure out how to bring up the modal during manual testing.

Speaking of tests, this PR is missing them! Please add a test that brings up the modal, does a search, verifies the search results are correct, and then closes the modal.

sandrina-p:
Regarding the cypress tests, one suggestion for you @pieer:
Since we have bypass-ui for signup/login but we don't for `acceptInvitation`, I would do this flow:

- loop to signup 12 users (with bypassUI) and logout,
- user1 creates a group,
-  loop to login userX and accept Invitation.

or the most correct way: - create acceptInvitation bypass-ui action and do it directly without the need to signup/logout and login again.

frontend/views/containers/dashboard/GroupMemberMenu.vue
taoeffect: In `router.js` this is currently bound to the route `'/messages/:chatName'`

frontend/views/containers/dashboard/GroupMembersAllModal.vue
taoeffect: Yep, just tested this and it's not needed. I see you already removed the one on line 41, don't forget about this one too!

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -1,18 +1,254 @@
...
+  },
+  methods: {
+    openModal (name) {
+      sbp('okTurtles.events/emit', OPEN_MODAL, name)
+    },
+    closeModal () {
+      this.$refs.modal.close()
+    },
+    toChat () {
+      this.$router.push({ path: '/chat' })
```
taoeffect: Same issue [as above](https://github.com/okTurtles/group-income-simple/pull/818/files#r376741116), I don't think `/chat` is a valid route, but `/messages` seems to be? Dunno, that stuff is also not implemented, so it's up to you whether you want to change it or leave it as-is for now.

taoeffect (changes requested):
@pieer this is looking excellent, I just tested it and it works.

Minor changes requested, see comments. The other thing is, can you add a single test that does a quick search for a `displayName` and verifies the results list returns the correct number of rows?

mmbotelho (approved):
Approved! Great job @pieer 🎉

taoeffect (approved):
Excellent work @pieer!

PR #820, opened 2020-02-01
Task/672 uniformize avatar size

### LINKED ISSUE #672: Uniformize avatar sizes across app
### Problem
@sandrina-p noticed that we didn't have a size scale for avatars, so I decided to make one! 

Captura de ecrã 2019-09-19, às 18 05 40

https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1876%3A17778

### Solution

1. Make this avatar scale a part of our front-end style guide;
2. Make sure that all avatars implemented so far have one of these sizes

**When in doubt:**
- Check the Figma mockups for the page you have doubts - I have updated all pages with this new scale;
- If you do find an avatar in the mockups that doesn't respect this scale, let me know!

### LINKED ISSUE #800: Fix exception when using "Anyone" join link twice
### Problem

When you've already joined a group using the "Anyone" invite link, if you use it again to join the same group while you're logged in, there's an exception in the console:

Screen Shot 2020-01-14 at 12 37 57 PM

### Solution

Track down and fix the exception.

sandrina-p:
Closes #672 

Besides uniformizing the Avatar size, some minor tweaks were also done:
- Rename `UserProfile` to `AvatarUser`. That way it's visually next to `Avatar` component on the tree-view, which makes sense because they are related.
- Add default group avatar image when creating a group (on first step).
- Fixes #800
- Convert Cypress "bypass-ui" button to a Vue component.
   It prevents the button from rendering on the page mount (you can see the button for a split of a second and then disappear. that's because it isn't a Vue component.


mmbotelho (changes requested):
The only thing I noticed was that the avatar is missing from the left-hand sidebar:
Captura de ecrã 2020-02-02, às 17 32 46

mmbotelho (dismissed):
It's working now! Great job! 🎉

frontend/views/components/AvatarUser.vue
```diff
@@ -2,47 +2,59 @@
...
     pictureURL () {
-      return this.profile ? this.profile.picture : this.ephemeral.url
-    }
-  },
-  data () {
-    return {
-      ephemeral: {
-        url: null
-      }
+      return this.profilePicture || this.ephemeral.url
```
taoeffect: I'm seeing two issues with this change:

1. By renaming `profile` computed property to `profilePicture`, you're introducing a bug on line 37 which checks for `this.profile`
2. I do not understand why `profile` was modified in the first place, as it was doing a good job and in fewer lines...

taoeffect (changes requested):
Just seeing one issue, see comment.

frontend/views/components/AvatarUser.vue
sandrina-p: 1. Fixed
2. Yes, this isn't correct, but the old solution isn't fully correct either.  `globalProfile` isn't truly getting the global profile. It only searches for profiles in the current group. If we use this getter on a chat when talking with someone that's not part of our group, that globalProfile will return undefined. But that's a problem to fix later... I reverted to `globalProfile()` and later we can improve `globalProfile` to work properly.

taoeffect (approved):
Approved! Excellent work @sandrina-p! 👍

Issue #788, opened 2019-11-30
Standardized error messages in scoped banners


taoeffect:
### Problem

In #779 we began to tackle the problem of user-facing error messages more seriously, by adding generalized tools to our disposal (the addition of the new scoped banners).

However, the error messages that are being shown to users in that PR still can be improved. For reference, here's what they looked like:

```js
this.$refs.formMsg.danger(L('Failed to upload the group picture, please try again. {codeError}', { codeError: error.message }))
```

Or the more verbose 3-liner:

```js
this.$refs.formMsg.danger(L('Failed to upload the group picture, please try again. {codeError}', {
  codeError: error.message
}))
```

There are a couple problems here:

1. "Please try again" is not good advice to give to the user. Trying again almost always results in failure, with the same error being displayed. So we should be telling them something more helpful than "try again".
2. There is inconsistent repetition with the variable that's being used to display the more detailed error message, where the variable is somethings called `codeError` and sometimes named something else. Furthermore, `{ codeError: error.message }` is unnecessarily long.

### Solution

Replace all scoped banner error messages with error messages that follow the following conventions.

To address problem 1, error messages should tell the users to do something useful. Something useful would be to send us a bug report along with all the information we need to debug it (or to follow an existing, open issue).

We can accomplish this by telling users [to visit the debug page as suggested here](https://github.com/okTurtles/group-income-simple/pull/779#discussion_r351491389).

To address problem 2, we can use an `LError` function to create a standardized error object.

So this:

```js
this.$refs.formMsg.danger(L('Failed to upload the group picture, please try again. {codeError}', { codeError: error.message }))
```

Becomes this:

```js
this.$refs.formMsg.danger(L('Error uploading group picture: {error}. {troubleshootLink}.', LError(error)))
```

Where `LError` is a helper function that returns an object that looks something like this:

```js
function LError (error) {
  return {
    error: error.message,
    troubleshootLink: L('{link_}Click to troubleshoot{_link}.', {
      'link_': `<router-link class='link' :to="{ path: '/debug', query: { error: '${error.message}' } }">`,
      '_link': '</router-link>'
    })
  }
}
```

- Note that this depends on #595 being closed
- Also the Style Guidelines.md documentation needs to have its example updated
- The debug page can use the passed in `error` query parameter to generate a draft of a message that the user can submit to us

sandrina-p:
I agree with this approach. It's a mature solution based on all our previous discussions during #779

sandrina-p:
[...]

Issue #409, opened 2018-05-19
Improve Voting Rules UX


taoeffect:
### Problem

![screen shot 2018-05-19 at 6 04 31 pm](https://user-images.githubusercontent.com/138706/40273512-1e1fc118-5b8f-11e8-842e-11f2db33764d.jpg)

1. The circle/handle/nob is too small
2. The nob doesn't visually indicate it's selected when clicked / dragged
3. The nob is not adjustable via the keyboard tab-selection
4. **(BUG)** The settings allow 1% to be selected, potentially letting anyone change any setting. This is confusing and probably wrong in all cases.

### Solution

1. Make the circle/handle/nob bigger, along with the circle line width itself
2. The nob needs to "depress" itself (perhaps by making itself a darker shade) to indicate it's selected
3. When `<tab>` is pressed, the knob should be highlightable so that the visually impaired can use screen readers to select it and adjust it (via the up/down arrows)
4. We should probably not allow the percentage to go below a certain amount, or we should say + show in the UI that a minimum of 2 votes is required no matter the threshold amount

Please note, **(1), (2), and (3)** all require forking the widget, which is good, because we need to remove all dependencies (#372).

mmbotelho:
Hey! As mentioned in the issue #343 I've made a few changes to this component. I followed @taoeffect 's solution proposal and also added some ideas of my own:

![screen shot 2018-11-19 at 13 15 59](https://user-images.githubusercontent.com/22004566/48707842-0357ac80-ec01-11e8-8121-e7f185021912.png)

- The handler is now bigger, and added a "drag" icon to it;
- Added an actual number input to the center of the circle. The focus / active states are no longer a problem, and the user can input a fixed amount directly, if they prefer to do so;
- Made it start from the top center, like a clock.
- Added a reset link, that should only be visible after the user changed the value, in case they want to return to the recommended percentage.

I am not sure if this will take a lot of effort to implement, but I believe it will significantly help the usability. I would love to hear @sandrina-p and Simon's input!

dotmacro:
I really like the accessibility-friendly warning icon 👍 
(Likewise with the other mockups in #343)

I also really like the warning text. 
Because users will still be able to propose simple majority (and even minority) voting thresholds, I think we should keep the "at least x members need to agree" text. We could word it as "only x members need to agree", so it still reads as a warning.

> Added a reset link, that should only be visible after the user changed the value, in case they want to return to the recommended percentage.

[...]

PR #823, opened 2020-02-06
#560 remove member proposal

### LINKED ISSUE #560: Implement the "Remove a member" proposal
### Problem

At the moment, it's not possible to remove a member from a group.

### Solution

It should be possible to do so, from 2 different places in the Dashboard:
1. From the Dashboard's sidebar;
2. From the "All members page";

Group

Removing a member is never a direct action - the group always needs to vote on it.

[You can check out the full userflow for proposing to remove a member on Figma](https://www.figma.com/file/ZR0DRPhhUuHdJtf7PbGAM6tW/Group-Income?node-id=2371%3A0)

sandrina-p:
Closes #560 

image

These are the scenarios covered in this PR.

##### Group with 2 members - user1 and user2
- User1 and user2 are both online. user1 removes user2. user2 is redirected to Home.
- User2 is offline. user1 removes user2. user2 logs in and sees screen to create a new group.
- User2 rejoins the group. user1 and user2 see correctly the group members.
###### When user2 also has a 2nd group (groupB):
- User1 and user2 are both online. user1 removes user2. user2 is redirected to groupB dashboard.
- User2 is offline. user1 removes user2. user2 logs in and sees groupB dashboard.
- User2 rejoins the group. user1 and user2 see correctly the group members.

The same scenarios but in a group with 3 members (to trigger a proposal).

The scenarios that don't involve having all users online at the same time are covered by Cypress tests.


frontend/views/components/modal/Modal.vue
```diff
@@ -1,6 +1,6 @@
 <template lang='pug'>
   div
-    component(:is='content' ref='content')
+    component(:is='content' ref='content' v-bind='contentProps')
```
sandrina-p, line 3: @taoeffect and @pieer. Here (#560) I initially proposed to pass "specific data" to a component through URL query. But with that approach can limit us if we need to pass not primitive values in the future.
Then, I found another way (the one made here) to pass that data as props (which sounds better to me, because it's more "Vue'ish").

What do you think?

frontend/model/contracts/group.js
```diff
@@ -256,6 +256,30 @@ DefineContract({
...
+        }
+      },
+      async sideEffect (message) {
+        const rootState = sbp('state/vuex/state')
+        const data = message.data()
+        if (data.username === rootState.loggedIn.username) {
+          console.warn('TODO sideEffect - Remove all related contracts. How?')
+        } else {
+          console.warn('TODO sideEffect - Remove member from state', data.username)
+        }
```
sandrina-p: @taoeffect a little guidance from you about what's your vision for a possible `enqueueRemoveContractSync` would be helpful here :)


Another scenario I remember: John and Bob have 2 different groups in common. Bob is removed from one of the groups. In that case, we can't remove Bob's `identityContract` from John's state because he would still need it on the other group. What's the best approach to detect/avoid this conflict?

frontend/model/contracts/group.js
```diff
@@ -256,6 +256,33 @@ DefineContract({
         archiveProposal(state, data.proposalHash)
       }
     },
+    'gi.contracts/group/removeMember': {
```
sandrina-p, line 351: @taoeffect so, some updates about this:

1. I figured out how to remove contracts. After all, `removeContract` already existed! I think that's what we need right?
2. The member is removed as expected (I think!). It's only missing to redirect them to home or `/dashboard`. How can we do that with sbp?
3. Now, a very weird thing happens. If the removed member joins the group again, it doesn't work! Because when syncing the contract, one of the actions is to remove them... so it kinda reverts the whole thing. Any suggestions to overcome this?
4. The proposal layout is also done, it's only missing the action [VOTE_FOR] when the proposal gets accepted, but to do that we need to first address the previous points.

Another thing:
- When a member is removed, there's an error on the remaining members' console: `uncaught Vue error in render: TypeError: "_vm.member is undefined"` Couldn't find yet what's causing this, but working on it.

frontend/views/containers/proposals/RemoveMember.vue
```diff
@@ -1,64 +1,138 @@
...
+      'groupMembersCount'
+    ]),
     member () {
-      // TODO - connect to real store.
-      return {
-        id: 1000,
-        displayName: '[Attila the Hun]',
-        picture: '/assets/images/default-avatar.png'
-      }
+      return this.username ? this.globalProfile(this.username) : {}
```
taoeffect: One of the issues with JS is that it lacks typing. To get around this, we use FlowType, but that also is insufficient. So the other trick we do is we use very descriptive names for everything (variables, computed properties, functions, etc.).

This computed property needs a better name, because `member` is too vague. It doesn't tell us anything about what is being returned. Let's therefore rename this to `memberGlobalProfile`. That then makes it clear what is being returned (the profile for member's global identity contract, as opposed to their group profile).

If their group profile should be returned instead, then it would be renamed to `memberGroupProfile` (and the `groupProfile` function would be used instead).

frontend/model/contracts/group.js
taoeffect: ### Recap

To remove a member, we basically need to undo everything that happens when they join, i.e. everything in `'gi.contracts/group/inviteAccept'`.

`'gi.contracts/group/inviteAccept'` does two things (that I know you noticed, based on this PR):

1. In the `/process` function, it adds their group profile to `state.profiles` (and updates the invite, although the invite is irrelevant for our purposes here)
2. In the `sideEffect`, we call `state/enqueueContractSync` on their identity contract, so that we have access to their global profile, which includes things like their `displayName` and their profile picture URL

In `proposals.js`, we see a TODO comment:

```js
    [VOTE_FOR]: function (state, { proposalHash, passPayload }) {
      console.error('unimplemented!')
      // TODO: unsubscribe from their mailbox and identity contract
      //       call commit('removeContract'), etc.
    },
```

Note that this comment is a guide only, as it contains one mistake: we never subscribe to anyone's mailbox contract other than our own, so we don't need to unsubscribe from that.

We are subscribed to their identity contract, however, as well as our own identity contract. We are subscribed to theirs any time we interact with the user in any way, including outside of a group, and we are subscribed to our own identity contract always (unless we are logged out). So if a group kicks us out, we do not need to unsubscribe from our identity contract (or our mailbox contract). But if someone else is kicked out, we *might* need to unsubscribe from their identity contract — **IF we have no other connection to this person.**

Imagine if a friend of ours gets kicked out of a group, but we have an open DM chat with them. We need to be subscribed to their identity contract still, if the chat is outside of the group. Otherwise we won't be able to see their profile picture or their `displayName`. Basically, if we're interacting with someone at all in any way, we need to be subscribed to their identity contract.

So for now, as a temporary solution, let's *not* unsubscribe from their identity contract if they're kicked out of the group, and open a separate issue to deal with keeping track of when we should unsubscribe from a user's identity contract.

### Also, some points of note:

- Subscribing and unsubscribing from contracts is, as you've observed correctly in this PR, something that must be done in a `sideEffect`, because it is impacting state that is outside of this current contract. Only state related to this contract directly can be modified in the `/process` function.

### Answering your questions

> 1. I figured out how to remove contracts. After all, `removeContract` already existed! I think that's what we need right?

Yes, we can use that to stop listening for events on the group contract if we are the ones being removed.

> 2. The member is removed as expected (I think!). It's only missing to redirect them to home or `/dashboard`. How can we do that with sbp?

The dashboard is only something visible in the group. We can redirect them home, sure. That is something that can be done via `sbp` if it is an action that is made available through `sbp`. Since we are using Vue, we can tell the root vue router to navigate us to `/` in the `sideEffect`.

> 3. Now, a very weird thing happens. If the removed member joins the group again, it doesn't work! Because when syncing the contract, one of the actions is to remove them... so it kinda reverts the whole thing. Any suggestions to overcome this?

That's a funny bug. Some observations:

- It is OK if during the `'gi.contracts/group/proposalVote/process'` the user's profile is removed from `state.profiles`, because it should be added back again in a future message (once we get to `'gi.contracts/group/inviteAccept/process'`).
- So the problem is likely in the fact that the `sideEffect` is asking us to unsubscribe from the group that we're currently in the middle of syncing

The issue is, if we're offline when someone proposes to remove us from a group, and the proposal passes, then we do in fact want to unsubscribe ourselves from the group once `enqueueContractSync` gets called after we login to GI.

In other words, sometimes we want to ignore it, and sometimes we don't. We want to ignore it only when we are in the middle of a contract sync that is a direct result of joining the group. 

So this suggests a few things:

1. We need to look into the code that's in Join.vue. In that code, we need to set a global flag that says we're calling `enqueueContractSync` because we're just joining a group, and unset that flag after the sync is finished.
2. This updated code should be moved out of Join.vue and into `controller/actions/group.js`.
3. Then in the `sideEffect` function for `'gi.contracts/group/proposalVote'`, if we see that a vote has passed removing us from this same group, we need to NOT remove ourselves if that flag is currently set

For example, I had a similar problem to deal with when implementing `autoBanSenderOfMessage` in `state.js`, and I did this:

```js
      if (contractIsSyncing[contractID]) {
        console.warn(`skipping autoBanSenderOfMessage since we're syncing ${contractID}`)
        return // don't do anything, assume the existing users handled this event
      }
```

Hopefully your solution is a bit cleaner than that. :-)  Try to keep any global variables stored in the same file as the functions that are using them (if you decide to use global variables).

> 4. The proposal layout is also done, it's only missing the action [VOTE_FOR] when the proposal gets accepted, but to do that we need to first address the previous points.

Yes, also, per our convo on Slack, this contract action `'gi.contracts/group/removeMember'` should be removed, since we want this code tot be done only in the `[VOTE_FOR]` (so that it's only run as a result of a successful proposal). The sideEffect can be moved into the sideEffect for `'gi.contracts/group/proposalVote'`

frontend/views/components/modal/Modal.vue
taoeffect: That seems reasonable to me, although @pieer would know the most Vue-correct way of doing this

pieer: Sounds good to me.

frontend/model/contracts/group.js
sandrina-p: Great overview, it explained a lot and was enough to help me finish the member removal! Here's some notes about it:
- I used `sbp('okTurtles.data/set', 'JOINING_GROUP', true|false)` to fix the _funny_ bug when rejoining a group.
- I created #828 (and pointed it on the code) to later address the condition to remove an identity contract.
- I manually tested the following scenarios where user1 and user2 have a group in common:
  - -- when user2 has only 1 group:
  - ✅ User1 and user2 are both online. user1 removes user2. user2 is redirected to Home.
  - ✅ User2 is offline. user1 removes user2. user2 logs in and sees screen to create a new group.
  - -- when user2 also has a 2nd group:
  - ✅ User1 and user2 are both online. user1 removes user2. user2 is redirected to the dashboard of their 2nd group. 
  - ✅ User2 is offline. user1 removes user2. user2 logs in and sees the dashboard of its 2nd group. 
   - ✅ User2 rejoins the group. user1 and user2 see correctly the group dashboard. 
  - any other cases we need to cover? I'll write the cypress tests later for the scenarios where user2 is offline during the removal.

You said this:
 > contract action 'gi.contracts/group/removeMember' should be removed, since we want this code to be done only in the [VOTE_FOR] (so that it's only run as a result of a successful proposal). The sideEffect can be moved into the sideEffect for 'gi.contracts/group/proposalVote'

Okay, I can move it to there... but then, when there are only 2 members in group, how do we reach that code, taking in account this isn't a proposal?

frontend/views/containers/proposals/RemoveMember.vue
sandrina-p: I agree. `member` renamed to `memberGlobalProfile`

frontend/model/contracts/group.js
taoeffect: Excellent work on using `okTurtles.data` for this! 💯 

Those tests sounds great to me!

> Okay, I can move it to there... but then, when there are only 2 members in group, how do we reach that code, taking in account this isn't a proposal?

Right now we don't. All I am saying is that this code needs to be moved for now. We can create an issue to address this in greater detail later, when we add a permissions and authorizations system. I.e. it would make sense to have this code here *if* we had such a system, where someone, e.g. an `admin`, can unilaterally send this message without a vote, and then in this code we would have a check to see if the user who sent this message has the `admin` permission.

But let's leave that for an issue for later, and for now move this into the proposal.js section of `[VOTE_FOR]`

sandrina-p: So, @taoeffect what you are saying is that for now, we cannot remove a member, unless it's a proposal. Is that right?

taoeffect: Yes exactly, for now (while we haven't implemented permissions), removal should only be done through proposals. This is something we can rectify later, once permissions are implemented, for groups that want to be able to designate an "admin" with the authority to remove whoever. For now, while we don't have a permissions system, we're placing all users on equal footing, and so that involves voting.

sandrina-p: Okay, so groups with 2 members will never be able to exclude each other. The only way will be by removing themselves in group settings with "Leave Group". Roger, I'll let you know when the change is made.

frontend/model/contracts/group.js
```diff
@@ -256,6 +334,46 @@ DefineContract({
...
+        // But there we don't have access to state and meta.
+
+        // If we do a validation here and it fails,
+        // will this try to ban/propose to remove only "meta.username" <-- i guess!
+        // or everyone in the group who proccesses this? <-- I saw this happening O.o
+
+        // Well... meanwhile I added validateRemoveMember
+        // on the component side too, before group/removeMember.
+        validateRemoveMember(state, { data, meta })
+        Vue.delete(state.profiles, data.member)
```
sandrina-p: @taoeffect this was my major question in this whole thing related with validations.

As far as I understand, we should have validation on the Vue part (UI) and also here in case someone tries to send a message hardcoded.

But if `/validate` fails, the message won't be sent, correct? Or at least it shouldn't I guess?...

Well, my question here is: When/Where should the validation be done? Multiple places? If so, that's why I created `validateRemoveMember ` and call it on the multiple places before calling `group/removeMember`.

Not sure if it's the correct way though.

frontend/model/contracts/voting/proposals.js
```diff
@@ -122,10 +122,48 @@ const proposals = {
...
+      // Q: Calling this here, so in case it fails, doesn't propose to remove the member.
+      // And instead it catches (and shows) the error to the user.
+      validateRemoveMember(state, { data, meta: proposal.meta })
+
+      sbp('gi.contracts/group/removeMember/process', state, {
+        meta: proposal.meta,
+        data
+      })
+
+      removeMemberSideEffect(data)
```
sandrina-p: Read this code comment here.

frontend/views/containers/proposals/RemoveMember.vue
```diff
@@ -1,58 +1,155 @@
...
+      })
+    },
+    async submit (form) {
+      this.$refs.formMsg.clean()
+      const member = this.username
+      const memberID = this.groupProfiles[member].contractID
+
+      if (this.groupShouldPropose) {
+        // Look for existing proposal
+        for (const hash in this.currentGroupState.proposals) {
```
sandrina-p: This is similar to the validation we "have" (I commented it) on `contract/group.js` but specific for removeMember proposal. Should we do it there too?

frontend/model/contracts/group.js
```diff
@@ -194,15 +261,26 @@ DefineContract({
             prop.meta.username === meta.username &&
             deepEqualJSONType(prop.data.proposalData, data.proposalData)
           ) {
-            throw new TypeError(`proposal: is identical to proposal ${hash}`)
+            // BUG: cc @taoeffect
```
sandrina-p: @taoeffect, please read this code comment bellow 👇

frontend/model/contracts/group.js
```diff
@@ -63,6 +69,67 @@ function initGroupProfile (contractID: string) {
...
+    sbp('state/vuex/commit', 'setCurrentGroupId', groupIDToSwitch)
+    sbp('state/vuex/commit', 'removeContract', data.groupID)
+
+    sbp('state/router').push({ path: groupIDToSwitch ? '/dashboard' : '/' })
+    // TODO - #828 remove other group members contracts if applicable
+  } else {
+    // TODO - #828 remove the member contract if applicable.
+    // sbp('state/vuex/commit', 'removeContract', data.memberID)
+  }
+  // TODO - verify open proposals and see if they need some re-adjustments.
```
sandrina-p: @taoeffect, please read this code comment bellow 👇

sandrina-p (commented):
@taoeffect please take a look at the new code (and comments!). To make things easy you can centralize your overall feedback in one/two single threads because it's all very related.

frontend/model/contracts/group.js
```diff
@@ -194,15 +261,26 @@ DefineContract({
             prop.meta.username === meta.username &&
             deepEqualJSONType(prop.data.proposalData, data.proposalData)
           ) {
-            throw new TypeError(`proposal: is identical to proposal ${hash}`)
+            // BUG: cc @taoeffect
+            // If I try to create the same proposal twice (imagine: remove member)
+            // it will create a proposal to remove me.
+            // That doesn't make sense....
+            console.warn(`proposal: is identical to proposal ${hash}`)
+            // throw new TypeError(`proposal: is identical to proposal ${hash}`)
```
taoeffect: Why doesn't it make sense? Creating the same exact proposal twice should not be allowed for any reason that I can think of. If it is, then can be abused to spam the group with identical proposals.

So it's a hard ban for a misbehaving client. There should be added error-checking on the client side (in the Vue code), to make sure that an identical proposal isn't sent. Instead the user should be told to vote on the existing proposal.

frontend/model/contracts/group.js
```diff
@@ -194,15 +261,26 @@ DefineContract({
...
           data,
           meta,
-          votes: { [meta.username]: VOTE_FOR },
+          votes: {
+            [meta.username]: VOTE_FOR,
+            ...(data.proposalType === PROPOSAL_REMOVE_MEMBER ? {
+              // Member to be removed cannot vote against.
+              [data.proposalData.member]: VOTE_FOR
+            } : {})
+          },
```
taoeffect: A few issues here:

- Only you can vote for yourself, and you cannot vote for anyone else. In the UI the member would see themselves wanting to ban themselves, this would generate anger and confusion. They would also simply attempt to change their vote.
- Another issue is that let's say the threshold for passing a vote is 65%. Well that threshold would be immediately reached in a group of 3 people with a single person voting, even though two people are required to pass the vote according to the threshold.
  - Ex: the group has users `[A, B, C]`. `A` proposes to remove `C`. The threshold is 65%. That means we want `B` to vote `VOTE_FOR` in order to remove `C`. It wouldn't be OK if `A` could simply remove `C` by simply making the proposal, without waiting for `B` to vote. But if the code were implemented like the above, they would be able to do it.
- This also isn't the way or place to handle this (so the above should be reverted to what it was before)

What situation are we trying to avoid here? Well, that should be clearly described in a comment (in the code) right next to the fix.

The situation we're trying to avoid is where the threshold is too high to allow the member to be removed because they won't vote to remove themselves. This can happen for example if the group doesn't have enough members, or loses members and the threshold isn't adjusted to account for it, or simply if the threshold was set too high to begin with.

Note that this problem exists with both `RULE_THRESHOLD` and `RULE_DISAGREEMENT`:

- `RULE_THRESHOLD`: if the threshold and size of the group are such that everyone must agree to remove the member, then member being removed can simply not vote, or vote against the proposal. So the threshold must be artificially lowered to allow for that possibility.
- `RULE_DISAGREEMENT`: if a single user can block a proposal by disagreeing, then the member being kicked out can simply disagree. So for removing members, the minimum setting for the threshold has to be "at least 2 disagreements".

So what you want to do is to modify the file `rules.js`, and for each rule type (`RULE_THRESHOLD` and `RULE_DISAGREEMENT`), check if `proposalType === PROPOSAL_REMOVE_MEMBER`, and if so, **lower or increase** the threshold appropriately (depending on the rule).

- For `RULE_DISAGREEMENT`, you'll want something along the lines of `disagreementThreshold = Math.max(disagreementThreshold, 2)`
- For `RULE_THRESHOLD`, you'll want something along the lines of `threshold = Math.min(threshold, (population - 2) / population)` (double-check my math there)

frontend/model/contracts/group.js
taoeffect: Think about it this way:

- The UI must have validations to prevent users from using the UI from ever creating and sending bad messages (messages that will result in invalid state in the contract chain)
- The contract chain must have validations to ban users that attempt to create invalid state using modified client software

> But if `/validate` fails, the message won't be sent, correct? Or at least it shouldn't I guess?...

Well, `/validate` is called in two cases:

1. By our unmodified client software when creating a message to be sent out to everyone else ("to the contract chain")
2. Again by our unmodified clients upon receiving a message from anyone else

A modified client (since everything we're doing is open source), can remove validations and generate bad messages. But as long as everyone else is running the right software, such attacks will be detected and that user will be auto-banned.

See my other comment for a more detailed response, the one that mentions issue #836 and says that yes, validation should be done in `/validate`.

frontend/model/contracts/group.js
```diff
@@ -256,6 +334,46 @@ DefineContract({
...
+        // QUESTION: Shouldn't these validatons be on validate() method above?
+        // But there we don't have access to state and meta.
+
+        // If we do a validation here and it fails,
+        // will this try to ban/propose to remove only "meta.username" <-- i guess!
+        // or everyone in the group who proccesses this? <-- I saw this happening O.o
+
+        // Well... meanwhile I added validateRemoveMember
+        // on the component side too, before group/removeMember.
+        validateRemoveMember(state, { data, meta })
```
taoeffect: You're absolutely right, validations should be done in the `validate` function, not in the `process` function. This is so that developers get automatic validation upon creation of a message, and upon receiving it for processing, without having to write extra lines of code like the call to `validateRemoveMember` in `RemoveMember.vue`.

You are also right, that it's a problem currently that the `validate` function is only passed the `message.data` and is missing the contract state and the metadata. This is a bug that I need to fix.

So for now in this PR, can you leave the `validateRemoveMember` defined, but don't use it anywhere? I've created an issue and self-assigned: #836 — feel free to mention that issue in a comment here in the code.

frontend/model/contracts/group.js
```diff
@@ -63,6 +69,67 @@ function initGroupProfile (contractID: string) {
...
+  if (membersCount === 2) {
+    if (meta.username !== groupState.settings.groupCreator) {
+      console.error('validateRemoveMember - not the group creator')
+      throw new GIErrorUIRuntimeError(L('In a group of 2, only the group creator can remove the other member.'))
+    }
+  } else {
+    const { payload } = groupState.proposals[data.proposalHash] || {}
+    if (payload && payload.secret === data.proposalPayload) {
+      console.error('validateRemoveMember - invalid associated proposal')
+      throw new GIErrorUIRuntimeError(L('The proposal associated to remove {memberToRemove} is not valid.', { memberToRemove }))
```
taoeffect: Note that `GIErrorUIRuntimeError` is generally reserved for use outside of contract definitions, because those errors are specifically for errors that happen via Vue.js code that is triggered by a user action (e.g. clicking a button).

Validation errors should throw `TypeError`, which is a neutral error that can happen both on creation of a GIMessage (which can happen when a user clicks a button, but generally only during development because of some mistake), or during processing of messages (because of an attack), so all of the above should be `throw new TypeError`.

taoeffect (changes requested):
Partial review finished! Some feedback left!

frontend/model/contracts/group.js
```diff
@@ -256,6 +334,46 @@ DefineContract({
...
+      },
+      process (state, { data, meta }) {
+        const rootState = sbp('state/vuex/state')
+        if (data.member === rootState.loggedIn.username) {
+          // If this member is re-joining the group, ignore the rest
+          // so the member doesn't remove themself again.
+          if (sbp('okTurtles.data/get', 'JOINING_GROUP')) {
+            return false
+          }
+        }
```
taoeffect: Two things:

1. Per the comment above, this code should be moved into the sideEffect, not here. It's OK for the process to delete the profile, because it should be re-added once we join the group. (Of course, verify that's true..)
2. I think the outer `if` shouldn't be needed. It should be sufficient to check if `'JOINING_GROUP'` is set

frontend/model/contracts/group.js
sandrina-p: I agree with you, it shouldn't be allowed.
But as it is, if I try to create a proposal similar to another one, I get banned. The correct UX should be to show a UI error.

But, now with your new change #838, we can do that at both `validate()`and  `process()`. The diff between both is that `validate()` doesn't ban the user but `process()` does.

frontend/model/contracts/group.js
sandrina-p: 1. we still the verification 'JOINING_GROUP', because when re-joining the group, the member isn't re-added to the list.
2. will remove.

sandrina-p: EDIT:
2. It cannot be removed.  When rejoining a group that removed another member in the past (besides us) we want to perform the whole removal action. If we remove the outer if, the action gets "ignored".

frontend/model/state.js
```diff
@@ -140,6 +140,11 @@ const mutations = {
     sbp('okTurtles.events/emit', EVENTS.CONTRACTS_MODIFIED, state.contracts)
   },
   setContractHEAD (state, { contractID, HEAD }) {
+    const contract = state.contracts[contractID]
+    if (!contract) {
+      console.error(`This contract ${contractID} doesnt exist anymore. Probably you left the group just now.`)
+      return
```
sandrina-p: This console.error is always triggered when we are the members being removed. I didn't find a way to avoid it. Need your help @taoeffect

test/cypress/integration/group-payments.spec.js
```diff
@@ -220,7 +220,7 @@ describe('Payments', () => {
   })
 
   it('user2 pledges $100 and sees their contributions.', () => {
-    cy.giSwitchUser(`user2-${userId}`, { bypassUI: true })
+    cy.giSwitchUser(`user2-${userId}`)
```
sandrina-p: Note: By default giSwitchUser always uses `bypassUI`, so there's no need to pass the argument.

frontend/model/contracts/members/removeMemberSideEffect.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+import sbp from '~/shared/sbp.js'
```
sandrina-p: NOTE: @taoeffect 
I had to move this function out of `contracts/group.js`, otherwise, it would create a circular dependency when imported at `proposals.js` because that file also imports `contracts/group.js`.

Not sure if this is the best place... any suggestion?

frontend/model/contracts/members/removeMemberSideEffect.js
```diff
@@ -0,0 +1,31 @@
...
+    sbp('state/vuex/commit', 'removeContract', data.groupID)
+
+    sbp('state/router').push({ path: groupIDToSwitch ? '/dashboard' : '/' })
+    // TODO - #828 remove other group members contracts if applicable
+  } else {
+    // TODO - #828 remove the member contract if applicable.
+    // sbp('state/vuex/commit', 'removeContract', data.memberID)
+  }
+  // TODO - verify open proposals and see if they need some re-adjustments.
+  // ex 3 of 4 members voted -> one member left -> 3 of 3 members voted -> need to resync and mark it as passed!
```
sandrina-p: I think we could create an issue for this and address it in another PR. What do you think @taoeffect?

frontend/model/contracts/voting/proposals.js
```diff
@@ -122,10 +123,41 @@ const proposals = {
...
+        groupID,
+        proposalHash,
+        proposalPayload: passPayload
+      }
+      const message = { data, meta: proposal.meta }
+      sbp('gi.contracts/group/removeMember/process', message, state)
+
+      removeMemberSideEffect(data)
+      // // NOTE_1: I tried to call selector sideEffect manually but it didnt work
+      // // TypeError: selectors[selector] is not a function - sbp.js
```
sandrina-p: Please read this code comment.

frontend/model/contracts/voting/rules.js
```diff
@@ -2,6 +2,10 @@
 
 import { literalOf, unionOf } from '~/frontend/utils/flowTyper.js'
 
+// Error: Circular dependency:
+// import { PROPOSAL_REMOVE_MEMBER } from '@model/contracts/voting/proposals.js'
+const PROPOSAL_REMOVE_MEMBER = 'remove-member'
```
sandrina-p: TODO - Where should we create the file with all these `const`?

frontend/model/state.js
sandrina-p, line 146: This console.error is triggered when we are the member being removed (through a proposal). More details about how I ended up here at proposal.js L150
I didn't find a way to avoid it. Need your help @taoeffect

test/cypress/integration/group-payments.spec.js
```diff
@@ -251,7 +251,7 @@ describe('Payments', () => {
   })
 
   it('user3 pledges $100 and sees who they are pledging to - $50 to user1 (Greg)', () => {
-    cy.giSwitchUser(`user3-${userId}`, { bypassUI: true })
+    cy.giSwitchUser(`user3-${userId}`)
```
sandrina-p, line 254: Note: By default giSwitchUser always uses bypassUI, so there's no need to pass the argument.

frontend/model/contracts/members/removeMemberSideEffect.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+import sbp from '~/shared/sbp.js'
+
+export default async function removeMemberSideEffect (data) {
+  const rootState = sbp('state/vuex/state')
```
sandrina-p: NOTE: @taoeffect
I had to move this function out of contracts/group.js, otherwise, it would create a circular dependency when imported at proposals.js because that file also imports contracts/group.js.

Not sure if this is the best place... any suggestion?

test/cypress/integration/group-member-removal.spec.js
```diff
@@ -0,0 +1,188 @@
...
+      username: `userBot-${userId}`,
+      groupName: groupNameB,
+      bypassUI: true
+    })
+  })
+
+  it('user1 proposes to remove userBot', () => {
+    cy.giLogin(`user1-${userId}`, { bypassUI: true })
+
+    openRemoveMemberModal('userBot', 3)
```
sandrina-p, line 142: @taoeffect we are having random Cypress fails again where some data is not synced at time.
In [this run spec](https://dashboard.cypress.io/projects/q6whky/runs/299/specs) go see the video. User1 logins but doesn't see userBot on the group members' list, although userBot successfully joined the group on the previous test.

The `data-sync` we added to #app during the hackathon doesn't seem to be working correctly. Any suggestions to fix this?

frontend/controller/router.js
```diff
@@ -178,4 +179,8 @@ router.beforeEach((to, from, next) => {
   next()
 })
 
+sbp('sbp/selectors/register', {
+  'state/router': () => router
```
taoeffect: `state/` is the namespace for things in our `model/` folder (i.e. "state" and "model" mean basically the same thing).

So I would place this instead under `controller/`, e.g. `controller/router`.

frontend/controller/actions/group.js
```diff
@@ -119,5 +119,22 @@ export default sbp('sbp/selectors/register', {
...
+      await sbp('backend/publishLogEntry', message)
+    } catch (e) {
+      const mapMsgs = {
+        not_a_member: () => L('{memberToRemove} is not part of the group.', { memberToRemove }),
+        cannot_remove_yourself: () => L('You cannot remove yourself.'),
+        invalid_associated_proposal: () => L('The associated proposal is invalid.'),
+        default: () => L('Failed to remove member: {codeError}', { codeError: e.message })
+      }
+
+      throw new GIErrorUIRuntimeError((mapMsgs[e.message] || mapMsgs.default)())
```
taoeffect: I do not see any reason to do it this way, and I think we should avoid doing it this way few reasons:

1. This is a departure from how it's done everywhere else. We want consistency in our error handling (see #788)
2. This introduces an unnecessary and confusing intermediate step between the writing of the error message and its actual instantiation, forcing developers to write and think about code in two different places for error messages. There is no reason to do it like this. A simpler approach exists.

Instead, the error message should be written where it occurs, as we've been doing. If you'd like to add something in front of it, you can do so using `e.message` (as you do with the `default` here, and as is done elsewhere in the codebase).

frontend/model/contracts/group.js
```diff
@@ -65,6 +69,37 @@ function initGroupProfile (contractID: string) {
...
+  }
+}
+
+export function validateRemoveMember ({ data, meta }, { state }) {
+  const memberToRemove = data.member
+  const membersCount = Object.keys(state.profiles).length
+
+  if (!state.profiles[memberToRemove]) {
+    console.error(`removeMember - ${memberToRemove} isn't part of the group`)
+    throw new TypeError('not_a_member')
```
taoeffect: Note the duplication here. The text for this appears in the key for `not_a_member` and we don't need the `console.error`s here because we can have one instead (up above in the exception handler).

Instead, simply put the exact error message here inside the `TypeError` constructor, and remove each of the `console.error` logs in this function. Then, per DRY, have a single `console.error('removeMember', e)` at the top most exception handler. This will result in the stack trace being logged as well to the console (so that coupled with the unique error message, we'll know exactly where the error came from).

frontend/model/contracts/group.js
```diff
@@ -262,30 +297,45 @@ DefineContract({
...
-      }),
-      process ({ data, meta, hash }, { state }) {
-        // make sure this isn't a duplicate proposal
+      validate: (data, { state, meta }) => {
+        objectOf({
+          proposalType: proposalType,
+          proposalData: object, // data for Vue widgets
+          votingRule: ruleType,
+          expires_date_ms: number // calculate by grabbing proposal expiry from group properties and add to `meta.createdDate`
+        })
```
taoeffect: The way `objectOf` (and the other validation functions) work is they return a function, and that function is then called with `data`. So this is missing on line `306`:

```js
})(data)
```

frontend/model/contracts/group.js
```diff
@@ -65,6 +69,37 @@ function initGroupProfile (contractID: string) {
...
+
+  if (membersCount >= 3) {
+    const { payload } = state.proposals[data.proposalHash] || {}
+    if (!payload || payload.secret !== data.proposalPayload.secret) {
+      console.error('removeMember - invalid associated proposal', data.proposalHash)
+      throw new TypeError('invalid_associated_proposal')
+    }
+  }
+
+  return true
```
taoeffect: This does nothing so this line isn't needed

frontend/model/contracts/group.js
```diff
@@ -262,30 +297,45 @@ DefineContract({
...
         }
+      },
+      process ({ data, meta, hash }, { state }) {
         Vue.set(state.proposals, hash, {
           data,
           meta,
-          votes: { [meta.username]: VOTE_FOR },
+          votes: {
+            [meta.username]: VOTE_FOR
+          },
```
taoeffect: Personally I like 1-liner brevity but if you have strong preferences for this style I'm OK with it

frontend/model/contracts/group.js
```diff
@@ -339,6 +389,37 @@ DefineContract({
...
+          member: string,
+          memberID: string,
+          groupId: string,
+          // In case it happens in a big group (by proposal)
+          // we need to validate the associated proposal.
+          proposalHash: optional(string),
+          proposalPayload: optional(objectOf({
+            secret: string // NOTE: simulate the OP_KEY_* stuff for now
+          }))
+        })
```
taoeffect: `(data)`

frontend/model/contracts/group.js
```diff
@@ -339,6 +389,37 @@ DefineContract({
...
+        validateRemoveMember({ data, meta }, { state })
```
taoeffect: This function is now no longer being called from anywhere else so the code should be moved into here.

taoeffect (changes requested):
Great job on this PR! 💪

I wasn't able to finish the entire review, but to not keep you waiting I decided to leave this partial review. I'll have more feedback hopefully later tomorrow, but don't wait up for it if you feel like immediately addressing this partial review!

frontend/model/state.js
```diff
@@ -310,6 +315,12 @@ const getters = {
       })()
     }
   },
+  userDisplayName (state, getters) {
```
pieer, line 318: We have few userDisplayName in the app already. Should we clean this in the same time?

frontend/views/containers/dashboard/GroupMemberMenu.vue
```diff
@@ -1,10 +1,18 @@
 <template lang="pug">
-menu-parent
-  menu-trigger.is-icon-small
+menu-parent(data-test='menuActions')
+  menu-trigger.is-icon-small(aria-label='Show actions')
     i.icon-ellipsis-v
 
   menu-content.c-actions-content
-    ul
+    ul(v-if='username === ourUsername')
```
pieer, line 7: Should we have the v-if on the option instead?

pieer (dismissed):
Awesome work! Just left few minor comments

frontend/controller/router.js
sandrina-p: But this file is `controller/router.js`. I didn't understand where you want to place this. 🤔

I placed it here because it's the closest possible place to the `router` creation.

frontend/controller/actions/group.js
sandrina-p: I understand. reverted.

frontend/model/contracts/group.js
sandrina-p: Done. maybe `validate()` should fail if data is empty.

frontend/model/state.js
sandrina-p: Actually I did that and didn't found any place to replace it. Everywhere is already using `ourDisplayName`. Maybe I missed something you saw?

frontend/views/containers/dashboard/GroupMemberMenu.vue
sandrina-p: I prefered here to have only 1 v-if. If we added it to the option, we would do it 3 times (or more if we add more options in the future). I find this way cleaner. But if you have high preference, I can change it to individual ifs.

frontend/controller/router.js
taoeffect: Sorry, I wasn't referring to the file location, but to the SBP namespace. My suggestion here is to rename line 183 to `'controller/router'`. The file/folder path is also named this way because conceptually the router is part of the C in MVC. SBP namespaces are also based on conceptual hierarchies, so I think it would make sense to also have the SBP namespace here be `'controller/`'

frontend/model/contracts/group.js
taoeffect: Not sure what you mean? `objectOf` will fail if `data` is null

frontend/model/state.js
```diff
@@ -310,6 +315,12 @@ const getters = {
...
+    return (username) => {
+      const profile = getters.globalProfile(username) || {}
+      return profile.displayName || profile.name || username
```
taoeffect: `profile.name` is the same as `username` (and we should open an issue to rename it to `username` remove it entirely... ok done: #848), so the check for `profile.name` is not needed, and since we might remove that property later it should be removed.

frontend/model/contracts/group.js
```diff
@@ -262,30 +272,41 @@ DefineContract({
...
+          }
+
+          if (prop.meta.username === meta.username &&
             deepEqualJSONType(prop.data.proposalData, data.proposalData)
           ) {
-            throw new TypeError(`proposal: is identical to proposal ${hash}`)
+            throw new TypeError('This proposal is identical to another one.')
           }
         }
+      },
```
taoeffect: A few notes:

- Both here and in several other places in this file (do a search for `TypeError('`), I forgot to wrap the error messages in `L` functions. Please add them so that the errors get translated (here and elsewhere in the file)
- We want this function to be as generic as possible, meaning to keep it clean, ideally it does not check for a specific `proposalType`. I don't remember why I added the check for `prop.meta.username === meta.username`, but if we remove that then I think it's possible to remove the proposal-specific check on lines 290-294, and instead throw an exception for an "identical proposal" if `deepEqualJSONType(prop.data, data)` ... although double-check me on that (I saw that you added a `removeMemberSecret()` which might break that. I still need to review that part and see if there's a way to not have to use it)

frontend/model/contracts/group.js
taoeffect: > we still the verification 'JOINING_GROUP', because when re-joining the group, the member isn't re-added to the list.

I don't understand, what do you mean the member isn't re-added to the list? Why not? They should be added later on via the `group/inviteAccept` message.

> It cannot be removed. When rejoining a group that removed another member in the past (besides us) we want to perform the whole removal action. If we remove the outer if, the action gets "ignored".

Maybe you can explain this to me on a call, as I'm not understanding.

I know intuitively that it does not make sense to not run the process function and yet to run the sideEffect, which seems to be what you're suggesting.

As per above, the same reasoning should apply, that `group/inviteAccept` should re-add their profile to the group, and it's only the sideEffect that we should be concerned about.

On a separate note, I'm realizing that the way the `if` is written can be rewritten as a single `if` (e.g. `if (data.member === rootState.loggedIn.username && sbp('okTurtles.data/get', 'JOINING_GROUP')) { return }`), but this doesn't matter if the `if` should be removed entirely.

frontend/model/contracts/members/removeMemberSideEffect.js
```diff
@@ -0,0 +1,31 @@
...
+  const contracts = rootState.contracts || {}
+
+  if (data.member === rootState.loggedIn.username) {
+    // If this member is re-joining the group, ignore the rest
+    // so the member doesn't remove themself again.
+    if (sbp('okTurtles.data/get', 'JOINING_GROUP')) {
+      return false
```
taoeffect: Recommend just `return`, since the return value isn't used by anything, and we don't want to imply it is.

frontend/model/contracts/members/removeMemberSideEffect.js
taoeffect: Definitely not this PR 👍

frontend/model/contracts/group.js
```diff
@@ -339,6 +360,52 @@ DefineContract({
...
+        }
+        if (membersCount === 1) {
+          throw new TypeError('Cannot remove yourself.')
+        }
+        if (membersCount >= 3) {
+          const { payload } = state.proposals[data.proposalHash] || {}
+          if (!payload || payload.secret !== data.proposalPayload.secret) {
+            throw new TypeError('Invalid associated proposal.')
+          }
+        }
```
taoeffect, line 392: I do not get this invalidation function or why there's a secret.

I remember us reaching the conclusion on Figma that check for `group/removeMember` would be if the group creator was the one sending the message, and only in the case that voting isn't required (i.e. less than 3 members). That's not what this function is doing.

frontend/model/contracts/group.js
```diff
@@ -339,6 +360,52 @@ DefineContract({
...
+          // so the member doesn't remove themself again.
+          if (sbp('okTurtles.data/get', 'JOINING_GROUP')) {
+            return false
+          }
+        }
+
+        Vue.delete(state.profiles, data.member)
+      },
+      sideEffect ({ data }) {
+        removeMemberSideEffect(data)
```
taoeffect: 1. `sideEffect` must be marked `async`
2. `removeMemberSideEffect` must have an `await` before it (here and anywhere else it's called, e.g. proposals.js)

frontend/model/contracts/voting/proposals.js
taoeffect: yeah it's a bit confusing how it was defined in `Contract.js`, but if you have a look there you'll see it needs to be `'gi.contracts/group/removeMember/process/sideEffect'`. I'm considering rewriting as you've done here, as I think that might be more intuitive, feel free to open an issue for that if you'd like.

frontend/model/contracts/voting/proposals.js
```diff
@@ -122,10 +123,41 @@ const proposals = {
...
+
+      // NOTE_2: Everything went fine to all members expect the one being removed.
+      /*
+       There was an error:
+       - "processMutation: error TypeError TypeError: "state.contracts[contractID] is undefined"
+       - state.js:653 -> :172 at setContractHEAD()
+
+       I think this happens because the member received a new event where
+       the proposal was accepted. But the contract was already removed, so it breaks...
+       - A quick fix was to put a if() there but I feel like the issue is bigger than that.
```
taoeffect: I think that might in fact be all that's needed to fix this, but let me think about that a bit more... Can you create an issue and assign it to me to double-check this?

frontend/model/contracts/members/removeMemberSideEffect.js
taoeffect: I think having this inside of `group.js` and calling it via `await sbp(...` as you tried should be fine. You just were calling the wrong selector before (see comment below about that)

taoeffect (changes requested):
Draft 1 of my review. This was a really tough PR to review, some of the code is kinda hard to think about it. Great job @sandrina-p on this! Very impressive. 💪

mmbotelho (changes requested):
Captura de ecrã 2020-02-25, às 11 33 27
Captura de ecrã 2020-02-25, às 11 16 20
Captura de ecrã 2020-02-25, às 11 18 35

That's all I found so far!

frontend/model/state.js
sandrina-p: Checking `profile.name || username`, is needed in case `profile` returns an empty object (a.k.a. this user is not part of the group anymore).

frontend/model/contracts/group.js
sandrina-p: - done!
- `deepEqualJSONType` can catch proposals that are an exactly match, but I think we should be smarter than that. For example if I create one proposal to remove Bob and after if I create another one but with a different reason,  `deepEqualJSONType` wouldn't catch it. So that's why in the line 291-292 we check only for the `proposalData.member`. If it's a match, it's a duplicated, even if the reason for it is different. Does it make sense? Maybe we could compare the proposal data and exclude `reason` from it, so it's a generic "catch" for any proposal. What do you think?

frontend/model/contracts/group.js
sandrina-p: Yes, we agreed that on Figma, but we still need a validation somehow when the group has >= 3 members, right? So we can prevent someone from calling this method hardcoded and "bypassing the proposal".

I thought to do the same as we do with invite a member. We create the secret at the moment the proposal is approved and send it through `proposals.js`. Does it make sense?

frontend/model/contracts/voting/proposals.js
sandrina-p: Done! #849

frontend/model/state.js
taoeffect: I don't think that's correct. Even if they aren't part of this group anymore, you're passing in `username` into this function. So it's presumed you already have it, therefore you don't need to check `profile.name` which is identical to `username`

frontend/model/contracts/group.js
taoeffect: > Maybe we could compare the proposal data and exclude reason from it, so it's a generic "catch" for any proposal. What do you think?

@sandrina-p brilliant solution!

frontend/model/state.js
taoeffect: You opened an issue and assigned it to me, I think that's good for now. 👍

frontend/model/contracts/voting/rules.js
taoeffect: How about under `voting/constants.js` ?

test/cypress/integration/group-member-removal.spec.js
taoeffect: @sandrina-p can you find a way to get the browser console log output from that session? It might be something else..

frontend/model/contracts/group.js
```diff
@@ -65,6 +70,12 @@ function initGroupProfile (contractID: string) {
   }
 }
 
+export function removeMemberSecret () {
+  return {
+    secret: `${parseInt(Math.random() * 10000)}` // TODO: this
+  }
+}
+
```
taoeffect: 1. Let's keep group.js uncluttered. It should ideally be only what's inside `DefineContract`. We already have too much stuff defined outside of it.
2. This function is only called once, and it doesn't do anything significant. Therefore it shouldn't be defined as a function, and instead its call to `(Math.random() * 10000).toFixed(0)` inlined.

frontend/model/contracts/voting/proposals.js
```diff
@@ -122,10 +123,41 @@ const proposals = {
...
+        member,
+        memberId,
+        groupId,
+        proposalHash,
+        proposalPayload: passPayload
+      }
+      const message = { data, meta: proposal.meta }
+      sbp('gi.contracts/group/removeMember/process', message, state)
+
+      await removeMemberSideEffect(data)
```
taoeffect:
```js
await sbp('gi.contracts/group/removeMember/process/sideEffect', message)
```

frontend/model/contracts/voting/proposals.js
```diff
@@ -122,10 +123,41 @@ const proposals = {
...
+      // // NOTE_1: I tried to call selector sideEffect manually but it didnt work
+      // // TypeError: selectors[selector] is not a function - sbp.js
+      // sbp('gi.contracts/group/removeMember/sideEffect', {
+      //   // Q: Why sideEffect(message), message is not a simple object?
```
taoeffect: (This comment can be removed now I think; See my comment above)

frontend/views/containers/proposals/ProposalVoteOptions.vue
```diff
@@ -97,14 +100,18 @@ export default {
...
         const payload = {}
-        if (inviteMember && oneVoteToPass(proposalHash)) {
+
+        if (isOneVoteToPass && this.type === PROPOSAL_INVITE_MEMBER) {
           payload.passPayload = createInvite({
             invitee: this.proposal.data.proposalData.member,
             creator: this.proposal.meta.username
           })
+        } else if (isOneVoteToPass && this.type === PROPOSAL_REMOVE_MEMBER) {
+          payload.passPayload = removeMemberSecret()
```
taoeffect: See comment above about inlining this as:

```js
{
  secret: (Math.random() * 100000).toFixed(0)
}
```

frontend/model/contracts/group.js
```diff
@@ -262,30 +273,39 @@ DefineContract({
...
+          }
+
+          if (prop.data.proposalType === PROPOSAL_REMOVE_MEMBER &&
+            prop.data.proposalData.member === data.member
           ) {
-            throw new TypeError(`proposal: is identical to proposal ${hash}`)
+            throw new TypeError(L('There is an open proposal to remove them.'))
+          }
+
+          if (deepEqualJSONType(prop.data.proposalData, data.proposalData)) {
```
taoeffect: Just a reminder to remove the proposal reason from the comparison (you can use the new `_.omit` func in giLodash)

frontend/model/contracts/group.js
taoeffect: > We create the secret at the moment the proposal is approved and send it through proposals.js. Does it make sense?

OK, the issue is that this validation function is only ever called as a result of a `'gi.contracts/group/removeMember'` message being created and sent to the group only. This does not happen anywhere in this PR. When the proposal passes, we bypass this validation code and call `'gi.contracts/group/removeMember/process'` and `'gi.contracts/group/removeMember/process/sideEffect'` directly from within `proposals.js`, and that does not happen as a result of a `gi.contracts/group/removeMember` message, but as a result of a `'gi.contracts/group/proposalVote'` message.

So this validation code doesn't get run at all anywhere in this PR.

And maybe that's one issue to consider. In other words, whether `proposal.js` should be modified to call the validation code as well. Maybe that should be a separate issue, because `Contract.js` does not expose the validate function via SBP (there are no `group/invite/validate` or `group/removeMember/validate` selectors). EDIT: OTOH, maybe not because... see last paragraph below.

However, that isn't even the biggest issue I'm noticing. The biggest issue is that `removeMember` is not really designed to be called from a proposal. It is designed to be called by the admin to remove a member without a proposal. So with the way this is implemented, it would in fact throw an exception, because `payload` would be undefined.

So these lines 386-389 should be removed, and instead the check should be whether `getters.groupSettings.groupCreator === meta.username`. (EDIT: note that this validation would mean that actually we shouldn't call validate from proposals.js, since the value for `meta.username` wouldn't necessarily be the group creator.)

frontend/model/contracts/group.js
sandrina-p: Let me rephrase: Maybe validate() should fail if data is not passed at all.

frontend/model/state.js
sandrina-p: Oh I see, changed!

frontend/model/contracts/voting/proposals.js
sandrina-p: Oh, okay, it makes sense now, because `sideEffect` is part of process. I'm okay with keeping it as it is for now.

frontend/model/contracts/members/removeMemberSideEffect.js
sandrina-p: Created issue https://github.com/okTurtles/group-income-simple/issues/850

test/cypress/integration/group-member-removal.spec.js
sandrina-p: I'll research about it tomorrow!

frontend/controller/router.js
sandrina-p: Ah! 🤦‍♀ done!

frontend/model/contracts/group.js
sandrina-p: > It is designed to be called by the admin to remove a member without a proposal. 

Hum... no, it's designed to be called by user1 (creator) or user2 as far as the group has only 2 members. ATM user2 can remove user1 if they want to. (in a group with 2 members).

The same can happen in a group with >=3 members, but through a proposal. (anyone can propose to remove the group creator)

So that's why the validation is `membersCount >= 3` and checks for the proposal secret to make sure it was triggered by a valid proposal.

frontend/model/contracts/group.js
sandrina-p: Done!

frontend/model/contracts/group.js
taoeffect: > Hum... no, it's designed to be called by user1 (creator) or user2 as far as the group has only 2 members. ATM user2 can remove user1 if they want to. (in a group with 2 members).

Yes, so on Figma we agreed that we would treat the creator as the admin in this case.

This code needs to be added here.

The proposal secret should not be checked here, because of what I wrote above. The proposal secret is never validated with the way this PR is written. That if statement is not only doing nothing, it will cause the code to break should we ever implement a UI for the creator to remove members when the member count is less than 3 members.

sandrina-p: > We agreed that we would treat the creator as the admin in this case.

I changed the code to do that, but what's the problem of user2 being able to remove user1? The validation of a group having less than 3 members isn't enough?


~I also removed the proposal secret, since it's not needed after all.~

> The proposal secret should not be checked here, because of what I wrote above.

So where should it be done?
The way I see it, it works fine here because this verification is indeed executed when a proposal is approved. It's called at proposals.js L39 by `removeMember/process`

P.S. Make sure to check the latest if statements because this comment may be attached to stale code.

frontend/model/contracts/group.js
sandrina-p: > I don't understand, what do you mean the member isn't re-added to the list? Why not? They should be added later on via the group/inviteAccept message.

I'm sorry, we were talking about different things. I thought you suggested to remove the `if(data.member === rootState.loggedIn.username)` from `process` and `sideEffect`. But you meant to remove it only from `process`.

Removed.

test/cypress/support/output-logs.js
```diff
@@ -0,0 +1,50 @@
+// ***********
+// Output App logs in case a test fails!
+// Copied directly from: https://github.com/cypress-io/cypress/issues/3199#issuecomment-466593084
+// ***********
```
sandrina-p, line 4: @taoeffect This is the solution I found to make cypress output the app logs in case a test spec fails again.

I added a commit to force the fail. (and meanwhile reverted it).
Here's its Cypress run: https://dashboard.cypress.io/projects/q6whky/runs/318/specs (click on output icon next to the first spec failing.

I noticed sometimes it says `[object, object]`. I tried to use `JSON.stringify()` as an attempt to show the original value but it didn't work. Do you have any other idea?

frontend/model/contracts/group.js
```diff
@@ -262,30 +266,35 @@ DefineContract({
...
-        // make sure this isn't a duplicate proposal
+      validate: (data, { state, meta }) => {
+        objectOf({
+          proposalType: proposalType,
+          proposalData: object, // data for Vue widgets
+          votingRule: ruleType,
+          expires_date_ms: number // calculate by grabbing proposal expiry from group properties and add to `meta.createdDate`
+        })(data)
+
+        const dataToCompare = omit(data.proposalData, 'reason')
```
taoeffect, line 277: This `dataToCompare` is only used once, in the `if`, and it would look cooler I think if there were two `omit` calls in the `if`, but if you think that makes the line too long I'm ok with keeping it as a variable (my preference though is to inline them both in the `if` though). However, if you decide you like it as a variable more, then can you move this line right above the `if` so that it's right next to the other call to `omit`? I feel like that would also help with readability/clarity.

frontend/model/contracts/group.js
taoeffect: > I changed the code to do that, but what's the problem of user2 being able to remove user1? The validation of a group having less than 3 members isn't enough?

The issue is that it would be a poor user experience for users creating groups if they created a group and invited someone to join the group, and could be immediately removed by the person they invited. As unlikely as that would be to ever happen, we still shouldn't allow it.

> So where should it be done?

It shouldn't be done at all, and all the code related to the proposal secret should be removed. :)

> The way I see it, it works fine here because this verification is indeed executed when a proposal is approved. It's called at proposals.js L39 by removeMember/process

I just had a look at the code in `proposals.js` in this PR and it does not result in any calls to the validate function. The proposal secret isn't needed in this PR.

The check here should simply throw an exception in the else. Do not worry, it will not get triggered, because the proposal code doesn't result in a call to this validation stuff anyway.

Add a `// TODO: check whether meta.username has required admin permissions` comment in the else as well, and we can deal with that in the future

taoeffect (changes requested):
See comments! Almost ready to merge! 😄

test/cypress/support/output-logs.js
```diff
@@ -0,0 +1,50 @@
...
+         * The args parameter will be all of the values passed as arguments to
+         * the console method. ( If your not using es6 then you can use `arguments`)
+         * Example:
+         *       If your app uses does `console.log('argument1', 'arument2');`
+         *       Then args will be `[ 'argument1', 'arument2' ]`
+         */
+      // Save everything passed into a variable or any other solution
+      // you make to keep track of the logs
+      // Use JSON.stringify to avoid [object, object] in the output
+      logs += JSON.stringify(args.join(' ')) + '\n'
```
taoeffect: Hmm, can you please make sure to log the `consoleProperty` as a prefix (e.g. `[log]`, `[error]`, etc.)?

This way one can search to see what type of log is being outpout.

frontend/model/contracts/group.js
sandrina-p: It’s used once but executed N times based on how many proposal there are. So instead of creating a new object every loop, we create it once and reuse it :)

test/cypress/support/output-logs.js
taoeffect: Hmm, did you try not doing `args.join(' ')` before passing the args to stringify? e.g. just `JSON.stringify(args)`?

Or alternatively, something like `args.map(a => JSON.stringify(a)).join(', ')`

frontend/model/contracts/group.js
sandrina-p: > I just had a look at the code in proposals.js in this PR and it does not result in any calls to the validate function. The proposal secret isn't needed in this PR.

Well... that's not true: I just saw it by myself. On the `else` I added a warn:
```
console.warn('This secret for this proposal is:', payload.secret)
```

And after the proposal gets approved, it is logged into the console, 1x per member:

image

So... that's why I think the secret is needed.

taoeffect: Where did you add that console.warn? Can you push your changes? It doesn't make sense to me why the validate function would be called for a remove member proposal, since the validate function for `'gi.contracts/group/removeMember'` ~~is only ever called when a `'gi.contracts/group/removeMember'` action is sent directly, and that only happens when there's no proposal.~~

EDIT: I'm completely wrong. Lol. Anytime `/process` is called the validate function gets called.

taoeffect: *Edit to reflect how wrong I am. OK, now let me re-process this and re-think if it's still needed...

sandrina-p: right here:  (not pushed yet)
image

taoeffect: OK. Then here's what needs to be done: the condition should throw in two cases:

1. If there is a proposal, then check if there is a proposal secret and throw if they don't match. I think what you're currently doing here is throwing if there is no proposal too... but we need a slightly different error in that case.
2. If there is no corresponding proposal, then we need to throw a different error, which says that `'admin credentials needed and not implemented yet'`, or something to that effect

frontend/model/contracts/group.js
taoeffect: ah ok, great point!

frontend/model/contracts/group.js
sandrina-p: done, please see again

test/cypress/support/output-logs.js
sandrina-p: Actually it's simpler than that: Just passing `args` but using template literals \` \` worked perfectly :D.

test/cypress/support/output-logs.js
sandrina-p: Added!

taoeffect (dismissed):
@sandrina-p Excellent job!!!

frontend/model/contracts/group.js
taoeffect: Un-resolving this since we need the `RULE_THRESHOLD` modification too

frontend/views/containers/proposals/RemoveMember.vue
```diff
@@ -1,58 +1,129 @@
...
+      'groupShouldPropose',
+      'groupMembersCount',
+      'ourUsername',
+      'userDisplayName'
+    ]),
+    memberGlobalProfile () {
+      return this.globalProfile(this.username) || {}
+    },
+    rule () {
+      const { threshold } = this.groupSettings.proposals['remove-member'].ruleSettings.threshold
```
taoeffect: Instead of `'remove-member'`, we must use the constant `PROPOSAL_REMOVE_MEMBER`.

Also, can you please check here and elsewhere (below and it seems even in other files), what's gong on with the value of `.ruleSettings`?

The idea was originally to make it so that each proposal could have a different voting rule applied to it, and that this could be changed by groups.

Which rule is applied to which proposal is stored in `getters.groupSettings.proposals.[<RULE_TYPE>].rule`.

And therefore to get the threshold this line actually needs to be:

```js
const proposalRule = this.groupSettings.proposals.[PROPOSAL_REMOVE_MEMBER]
const { threshold } = proposalRule.ruleSettings[proposalRule.rule]
```

I've noticed that this is an error in may other files. In this PR, can you also do some boyscoutting by searching the project for `ruleSettings.threshold` ?

These errors also happens in:

- AddMembers.vue
- Mincome.vue

There is one more issue: note that depending on the `.rule` that's used, the way that `threshold` is used is different. For `RULE_THRESHOLD`, you multiply the threshold by the member count (as done below), but for `RULE_DISAGREEMENT`, the threshold itself is the number of members that must say "no" in order to prevent a proposal from passing. Since this will involve changing the line below, `ProposalTemplate.vue`, and the text in it, can you create an issue for this part of it?

taoeffect (changes requested):
Oops, missed: https://github.com/okTurtles/group-income-simple/pull/823/files#r385417097

frontend/views/containers/proposals/RemoveMember.vue
sandrina-p: > I've noticed that this is an error in may other files. In this PR, can you also do some boyscoutting by searching the project for ruleSettings.threshold ?

Done.

Regarding the `RULE_DISAGREEMENT` needed modifications. This also involves design work from @mmbotelho. So I'd rather wait for her part (#409) and address it there, instead of creating another "related-duplicated" issue. Is that okay?

mmbotelho (changes requested):
Copy should be "Are you sure you want to remove [member-name] from the group?"

Captura de ecrã 2020-02-28, às 15 06 02

frontend/views/containers/proposals/RemoveMember.vue
taoeffect: It's OK, I'll create the issue just to make sure we don't forget, since #409 doesn't mention this. Adding this to my TODO list. Thanks for pushing the recent fixes, about to approve!

taoeffect (approved):
Looking groovy! Only changes that seem to be needed are the copy that @mmbotelho mentioned. I'm giving my approval for the code!

frontend/views/containers/proposals/RemoveMember.vue
taoeffect: Please leave this thread unresolved so that it can be linked to from issue #853

mmbotelho (approved):

frontend/model/contracts/group.js
taoeffect: Unresolving this not because there are any issues, but because this thread is important for helping us remember a summary of all of the major concerns and issues that we discovered with respect to removing members.

So it's a useful reference to not lose.

PR #824, opened 2020-02-08
Update styles of checkbox and radio inputs

### LINKED ISSUE #794: Update radio buttons UI and create checkbox component
# Problem

I have updated the UI for the radio buttons and created a new UI for checkboxes that will be needed for the payments page. The change to the radio buttons was made so that the radios and checkboxes:
- Have more contrast;
- Have more distinct selected / not selected states;
- Work well with each other.

I have also added the focused / active states for both the checkbox and radio components to our Figma design system.

# Solution

- [ ] Update radio buttons UI;
- [ ] Implement new checkbox component;
- [ ] Make sure that the active / focused states for both components are according to design.

You can find the [relevant design system documentation here](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=48%3A4) (should scroll down a little, it's close to the end of the artboard). Let me know if you have any questions!

sandrina-p:
Closes #794 

@mmbotelho Based on Figma discussion 171, I removed the bold style. Now, I think it's missing some spacing between each input when used vertically. What do you think?

image

### DIFF
--- a/frontend/assets/style/components/_forms.scss
+++ b/frontend/assets/style/components/_forms.scss
@@ -291,6 +291,11 @@ p.error {
     transform: translateY(4px);
   }
 
+  .input:disabled + :last-child {
+    color: $text_1;
+    cursor: default;
+  }
+
   .is-column & { 
     display: block;
 
--- a/frontend/assets/style/components/_forms.scss
+++ b/frontend/assets/style/components/_forms.scss
@@ -296,6 +296,19 @@ p.error {
     cursor: default;
   }
 
+  .input:not(:disabled):hover + :last-child {
+    &::before {
+      border-color: $primary_0;
+    }
+  }
+
+  .input:not(:disabled):focus + :last-child {
+    &::before {
+      border-color: $primary_0;
+      box-shadow: 0 0 0 2px $primary_1;
+    }
+  }
+
   .is-column & {
     display: block;
 
@@ -348,14 +361,6 @@ p.error {
       transform: translate(-50%, -50%) scale(1);
     }
   }
-
-  .input:not(:disabled):hover + :last-child,
-  .input:not(:disabled):focus + :last-child {
-    &::before {
-      border-color: $primary_0;
-      box-shadow: 0 0 0 2px $primary_1;
-    }
-  }
 }
 
 .checkbox {
@@ -405,14 +410,6 @@ p.error {
 
     &::after {
       transform: translate(-50%, -50%) scale(1);
-    }
-  }
-
-  .input:not(:disabled):hover + :last-child,
-  .input:not(:disabled):focus + :last-child {
-    &::before {
-      border-color: $primary_0;
-      box-shadow: 0 0 0 2px $primary_1;
     }
   }
 }
--- a/frontend/assets/style/components/_forms.scss
+++ b/frontend/assets/style/components/_forms.scss
@@ -15,6 +15,7 @@ form {
 
 legend.legend {
   font-weight: 600;
+  margin-bottom: $spacer-sm;
 }
 
 .field {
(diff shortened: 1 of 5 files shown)


mmbotelho (dismissed):
@sandrina-p 
> Now, I think it's missing some spacing between each input when used vertically. What do you think?

At the moment, there is no scenario where these components are used like you have them on the DS. I don't think it's useful to set a margin at this stage since you would have to remove them every time the components are used on the app. Does that make sense?

Anyway, if you want a margin in case we need to use them vertically, 16px seems right to my eyes. 

Let me know what you think! I'm approving this anyways :D good job!

mmbotelho (approved):
Approved! Good job!

mmbotelho (changes requested):
Sorry to revoke my approval on this PR @sandrina-p but I just realized the hover style for checkboxes and radio buttons was not properly specified on Figma, and that the current implementation of it is not consistent with how other form components behave.

I added the hover state to Figma. You can see it [here](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=48%3A4) or on the screenshot below:

Captura de ecrã 2020-02-11, às 14 35 02

sandrina-p:
@mmbotelho, hover style added, please take a look!

frontend/assets/style/components/_forms.scss
```diff
@@ -300,29 +336,76 @@ p.error {
...
+      transition: box-shadow 150ms, background-color 150ms;
     }
 
     &::after {
       position: absolute;
       width: 6px;
       height: 6px;
-      background-color: $primary_0;
+      background-color: $background_0;
       top: 11px;
```
pieer, line 347: Do we keep pixels here?

frontend/assets/style/components/_forms.scss
```diff
@@ -300,29 +336,76 @@ p.error {
...
+      border: 1px solid $general_0;
+      transform: translateY(2px);
+      transition: box-shadow 150ms, background-color 150ms;
+    }
+
+    &::after {
+      content: "\f00c"; // icon-check
+      text-rendering: auto;
+      line-height: 1;
+      font-family: "Font Awesome 5 Free";
```
pieer, line 390: should we use icon-check instead?

pieer (approved):
Just few questions, but all good otherwise :)

frontend/assets/style/components/_forms.scss
sandrina-p: I didn't convert these to `rem` because all form stuff is in pixels. So maybe we could create a task to convert all form stuff to rem in one go. What do u think?

frontend/assets/style/components/_forms.scss
sandrina-p: What do you mean? We use it on line 387 `content: "\f00c"; // icon-check`.

Or do you mean to add the element in the DOM? `<i class="icon-check">` I think it'd be too much verbose because each time we need to use .checkbox, we would need to conditionally hide that element.

frontend/assets/style/components/_forms.scss
pieer: We could. I wonder if it will look good if replace 11px by 0,6875 however...

frontend/assets/style/components/_forms.scss
pieer: I was talking about the icon-check. But I think you're right. Maybe a mixin could be a good compromise, but no need to do that in the PR I guess

mmbotelho (approved):
Approved @sandrina-p ! Thank you!

frontend/assets/style/components/_forms.scss
sandrina-p: yah a mixin sounds better, AFAIK, we have 3-4 places in our codebase using the icons this way. I can address all those places in a different PR

PR #830, opened 2020-02-10
Fix title classes

pieer:
Remove unnecessary class when subtitle + is-title=[x]
+ fix bug on c-title

### DIFF
--- a/frontend/views/components/modal/ModalTemplate.vue
+++ b/frontend/views/components/modal/ModalTemplate.vue
@@ -14,7 +14,7 @@
       .c-modal-content(ref='card' v-if='modalIsActive')
         header.c-modal-header(
           :class='{ "has-subtitle": $scopedSlots.subtitle }'
-          v-if='$scopedSlots.title || $scopedSlots.is-subtitle'
+          v-if='$scopedSlots.title || $scopedSlots.issubtitle'
         )
           modal-close(@close='close' :back-on-mobile='backOnMobile')
           h2.is-subtitle(v-if='$scopedSlots.subtitle')
--- a/frontend/views/components/modal/ModalTemplate.vue
+++ b/frontend/views/components/modal/ModalTemplate.vue
@@ -14,7 +14,7 @@
       .c-modal-content(ref='card' v-if='modalIsActive')
         header.c-modal-header(
           :class='{ "has-subtitle": $scopedSlots.subtitle }'
-          v-if='$scopedSlots.title || $scopedSlots.issubtitle'
+          v-if='$scopedSlots.title || $scopedSlots.subtitle'
         )
           modal-close(@close='close' :back-on-mobile='backOnMobile')
           h2.is-subtitle(v-if='$scopedSlots.subtitle')
--- a/frontend/views/components/modal/ModalTemplate.vue
+++ b/frontend/views/components/modal/ModalTemplate.vue
@@ -17,7 +17,7 @@
           v-if='$scopedSlots.title || $scopedSlots.subTitle'
         )
           modal-close(@close='close' :back-on-mobile='backOnMobile')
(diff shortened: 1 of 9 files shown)


frontend/views/containers/contributions/GroupPledgesGraph.vue
```diff
@@ -4,7 +4,7 @@
     :slices='mainSlices'
     :inner-slices='innerSlices'
   )
-    i18n.is-title-1.c-title(tag='h3' :args='LTags("span")') {span_}Group{_span} goal
+    i18n.has-text-1.is-title-3.c-title(tag='h3' :args='LTags("span")') {span_}Group{_span} goal
```
sandrina-p: We don't need `.is-title-x` here, this doesn't need to look like a title, is just a "normal" test with `.has-text-1` color. 

image

pieer: Mmm we actually stipulate in the guide that we need to tell the class all the time isn'
t it?

sandrina-p: We need to add the class if the UI matches. In this case, the UI doesn't look like a title, so there's no need for it:
image

pieer: Confusing.. anyway, there is no class is-text-[x] for 16px or 1rem

pieer: A 16px h3 title with no class

frontend/views/components/modal/ModalTemplate.vue
```diff
@@ -14,10 +14,10 @@
       .c-modal-content(ref='card' v-if='modalIsActive')
         header.c-modal-header(
           :class='{ "has-subtitle": $scopedSlots.subtitle }'
-          v-if='$scopedSlots.title || $scopedSlots.subTitle'
+          v-if='$scopedSlots.title || $scopedSlots.is-subtitle'
```
sandrina-p: I think this was a wrong find-replace. it should be subtitle as in the line 20

frontend/views/containers/contributions/GroupPledgesGraph.vue
sandrina-p: Okay, so another approach, that makes more sense:
Leave this "Group Goal" as a normal text (without heading) as we usually do.
Add the title of this graphic/summary top of the component and hide it using `.sr-only`

```pug
i18n.sr-only(tag='h3') Group Pledging Status
```

That way we serve both worlds visually and semantically.

sandrina-p (changes requested):
There was a wrong replacement on the code with `subtitle`. check the comment I left.

frontend/views/components/modal/ModalTemplate.vue
pieer: Fixed. Thanks

frontend/views/components/modal/ModalTemplate.vue
```diff
@@ -14,10 +14,10 @@
       .c-modal-content(ref='card' v-if='modalIsActive')
         header.c-modal-header(
           :class='{ "has-subtitle": $scopedSlots.subtitle }'
-          v-if='$scopedSlots.title || $scopedSlots.subTitle'
+          v-if='$scopedSlots.title || $scopedSlots.issubtitle'
```
sandrina-p: @pieer this should be `$scopedSlots.subtitle`

sandrina-p (changes requested):
you fixed the typo with another typo 😆

frontend/views/components/modal/ModalTemplate.vue
pieer: Of course... I'm tired today...

sandrina-p (approved):
okay, it seems all fine now! I'm merging this since it's something small

PR #835, opened 2020-02-16
Create "group_join" cypress bypassUI action

### LINKED ISSUE #814: Cypress - bypass UI remaining actions.
### Problem

Now that #801 is done, more actions can be converted to an approach that bypasses the UI, so we can reduce tests time. A good candidate for bypass-ui is an action that requires written input from the user because typing stuff is time-consuming.

### Solution
The following actions may be good candidates:
- accept group invitation when not logged in (`cy.giAcceptGroupInvite()`)
- set profile display name (`cy.giSetDisplayName()`)

sandrina-p:
Closes #814 

Add option `bypassUI` to cypress command `cy.giAcceptGroupInvite()`.

This is a command used 18 times on our tests. `bypassUI` was enabled everywhere expect on `group-proposals.spec` (4 times) so we make sure that the "join group" flow works properly.

#### Test speed comparison before and now)
Group 96
- Before: 102s
- Now: 85s
- Diff: 16s faster (~4/5 of the time).

Note 1: These tests were run locally on my machine, the values may be different compared to your machine.

Note 2: Initially I thought about also add option `bypassUI` to `cy.setDisplayName()` but then I noticed it was only used 4 times and it's a quick flow (compared to login or join a group). So I think for now it's not worth the effort. But if you disagree, let me know and I can do it too.

## Update:
On our Cypress dashboard, compared to our last 2 builds, the current build is **~1min faster!** 🚀 

image

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -96,11 +96,14 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/join': async function ({ groupId, inviteSecret }) {
     try {
+      // post acceptance event to the group contract
       const acceptance = await sbp('gi.contracts/group/inviteAccept/create',
         { inviteSecret },
         groupId
       )
+      // let the group know we've accepted their invite
       await sbp('backend/publishLogEntry', acceptance)
+      // sync the group's contract state
       await sbp('state/enqueueContractSync', groupId)
       return groupId
     } catch (e) {
@@ -110,6 +113,7 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/joinAndSwitch': async function (joinParams) {
     const groupId = await sbp('gi.actions/group/join', joinParams)
+    // after joining, we can set the current group
     sbp('gi.actions/group/switch', groupId)
     return groupId
   },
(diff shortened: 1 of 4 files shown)


frontend/views/pages/BypassUI.vue
```diff
@@ -119,6 +119,14 @@ export default {
           finalize: () => {
             this.$router.push({ path: '/dashboard' }) // eslint-disable-line
           }
+        },
```
taoeffect, line 120: Boyscoutting: noticing that on line 115 a callback function is being passed in that is no longer being used. Callback should be removed.

frontend/controller/actions/group.js
```diff
@@ -94,6 +94,25 @@ export default sbp('sbp/selectors/register', {
...
     return groupID
   },
+  'gi.actions/group/join': async function ({ groupId, inviteSecret }) {
+    try {
+      const acceptance = await sbp('gi.contracts/group/inviteAccept/create',
+        { inviteSecret },
+        groupId
+      )
+      await sbp('backend/publishLogEntry', acceptance)
+      await sbp('state/enqueueContractSync', groupId)
```
taoeffect, line 107: Don't forget to copy over the helpful comments from Join.vue

taoeffect (changes requested):
Looking good! Just very minor issues noticed

sandrina-p:
@taoeffect, changes addressed!

taoeffect (approved):
Looking great @sandrina-p!

PR #838, opened 2020-02-17
pass in contract state + metadata to validation functions (Closes #836)

### LINKED ISSUE #836: Pass in contract state + full message into validate function for contract actions
### Problem

`Contract.js` is currently defined in a way where the `validate:` function for contract messages is only passed in `message.data`. This is a problem, because for example in the PR for removing members (#823), it's useful to validate a message using information from both the metadata as well as the contract state.

However, simply passing that information in is not enough, because we want our code to "look pretty" and work well with validation functions defined by `flowTyper.js`, and those are not equipped to handle passed in group state or `GIMessage`s.

### Solution

- In `Contract.js`, for all the `/process` definitions, pass in the contract state (`null` for the constructor) and `GIMessage` into the validate function.
- Make it work well with the `flowTyper.js` definitions

taoeffect:
This should be merged before #823 (because it helps with that).

You should now be able to write contract validation functions for contract actions that accept the contract state and metadata.

Example:

```js
      validate: (data, {state, meta}) => {
        // validate 'data'
        objectOf({
          member: string,
          memberID: string,
          groupID: string,
          // In case it happens in a big group (by proposal)
          // we need to validate the associated proposal.
          proposalHash: optional(string),
          proposalPayload: optional(objectOf({
            secret: string // NOTE: simulate the OP_KEY_* stuff for now
          }))
        })(data)
        // validate something else using 'state'
        if (state.member === data.member) {
          throw new TypeError(`blah blah blah ${state.member} isn't supposed to something!`)
        }
      },
```

Note also that from now on, only the following functions from flowTyper may be used directly with `validate` (without wrapping them in another function; copied from flowTyper.js):

```
//   - arrayOf
//   - mapOf
//   - object
//   - objectOf
//   - objectMaybeOf (this is a custom function that didn't exist in flowTyper)
```

### DIFF
--- a/frontend/model/contracts/Contract.js
+++ b/frontend/model/contracts/Contract.js
@@ -15,13 +15,13 @@ export function DefineContract (contract: Object) {
   sbp('sbp/selectors/register', {
     [`${contract.name}/create`]: function (data) {
       const metadata = meta.create()
-      contract.contract.validate(data)
       meta.validate(metadata)
+      contract.contract.validate(data, { state: null, meta: metadata })
       return GIMessage.create(null, null, undefined, `${contract.name}/process`, data, metadata)
     },
     [`${contract.name}/process`]: function (state, message) {
-      contract.contract.validate(message.data)
       meta.validate(message.meta)
+      contract.contract.validate(message.data, { state, meta: message.meta })
       contract.contract.process(state, message)
     }
   })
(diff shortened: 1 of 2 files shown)


sandrina-p (approved):
This will be helpful at #823!

pieer (approved):
All good :)

Issue #804, opened 2020-01-23
State structure improvements


sandrina-p:
### Problem

This project started with a simple file structure, right for its needs. 
- A folder `views/` with Vue components, each one on its own file.
- A folder `model/contracts` with contracts, each one on its own file.
- A single file `model/state.js` with all the Vuex state management inside.

Now that we are close to finishing all the core features for the prototype, `state.js` is getting too big. One of the struggles is readability and naming inconsistencies/conflicts/purpose as new getters appear. (ex: `ourIdentityContractId` vs `ourUserIdentityContract` or `groupProfile` vs `groupProfiles`)

### Solution

Discuss the next step to evolve this structure and how it co-exists with `sbp`.

Here are some references about this.
- [Vuex - modules](https://vuex.vuejs.org/guide/modules.html)
- [Article - Large-scale Vuex applications structures](https://medium.com/3yourmind/large-scale-vuex-application-structures-651e44863e2f)

So, what's your vision for the next step?

cc @taoeffect and @pieer

pieer:
I believe that:
- sbp event registration and handleEvent functions could be move into a different file. Maybe model/utility/eventHandler.js
- Move model/errors.js to model/utility/error.js
- Move model/database.js to model/utility/database.js
- Remove model/constants.js

- In this file model/state.js we should have only mutations, getters and actions left.

Then I guess it's a good idea to split the state in 3 files: settings / user / group

So my recommendation:
model/state/user.js
model/state/group.js
model/state/settings.js
model/state/colors.js
model/utility/error.js
model/utility/database.js
model/utility/eventHandler.js

taoeffect:
Some thoughts:

I like the introduction of `controller/actions` that #801 introduces, and think this can help move some of the Vuex actions out of state.js. For example, `syncContractWithServer`, `login`, and `logout` can be migrated to SBP controller actions. This is in line with the TODO that's been in the state.js source for quite some time:

```
// TODO: convert all these to SBP... and/or call dispatch through SBP only!
```

Doing this leaves only one Vuex action, `handleEvent`, and I think that that action and everything related to it can be moved into its own file under either `model/` or `controller/`, but probably `model/` makes more sense since it's integral to the building of the Vuex state in the app.

[...]

Issue #749, opened 2019-10-28
Create language-agnostic CLI and library for interacting with contract chains


taoeffect:
*(Continued from #430)*

### Problem

It can be difficult sometimes to debug or visualize what exactly is happening in a contract. It would be useful for us and for third-party developers to have a generalized CLI interface for parsing, viewing, and interacting with contracts, especially as we solidify our open protocol.

### Solution

Create an SBP-based command-line tool that lets you:

- Display the entire list of events in a contract given a contractID and either a remote server or a directory
- Display contract state
- Delete contracts and all messages
- Generating a contract bundle from a contract file (#1214)
- Uploading a contract bundle to a server via either the `/file` POST API or a more primitive content-addressing API

And possibly other features like adding messages to an existing contract.

EDIT: this is now the [`chel` command](https://github.com/okTurtles/chel/)

PR #842, opened 2020-02-20
Replaced 'state/groupContractSafeGetters' with new getter support in DefineContract!

taoeffect:
This PR partially addresses #749 and #804.

- Contracts can now have Vuex-compatible getters defined on them! They can only access state that's inside of a contract (no outside state like `state.loggedIn.username`. This means we can get rid of `'state/groupContractSafeGetters'` finally and re-use getter definitions across our contract chains and Vuex!
- `validate` functions now have access to contract state, getters, and message metadata!
- The `state` and `message` parameters to `[action]/process` have been swapped! If you call `process` directly (as for example, some code in `proposals.js` did), make sure to update it or things will break! This PR makes these changes for two `[VOTE_FOR]` calls to `process` functions.
- Note that the contract chain constructor, formally defined in the `contract` property of the object passed to `DefineContract`, has been moved into the `actions` property, and is defined by naming the action the same as the contract name.
- Contract `sideEffect` functions are now passed in more data, like the contract's state, just like the `process` and `validate` functions
- Now `validate`, `process`, and `sideEffect` contract functions all take their parameters in the same way, with the "main data thing" being passed in as the first parameter, and all "optional supporting things" being passed in an object as the second parameter

### DIFF
--- a/frontend/model/contracts/Contract.js
+++ b/frontend/model/contracts/Contract.js
@@ -18,11 +18,11 @@ export const ACTION_REGEX = /^(([\w.]+)\/([^/]+)\/(?:([^/]+)\/)?)process$/
 // TODO: define a flow type for contracts
 export function DefineContract (contract: Object) {
   const metadata = contract.metadata || { validate () {}, create: () => ({}) }
+  const getters = contract.getters
   sbp('sbp/selectors/register', {
-    // expose getters for Vuex integration
-    [`${contract.name}/getters`]: function () {
-      return contract.getters
-    }
+    // expose getters for Vuex integration and other conveniences
+    [`${contract.name}/getters`]: () => getters,
+    [`${contract.name}/state`]: contract.state
   })
   for (const action in contract.actions) {
     if (action.indexOf(contract.name) !== 0) {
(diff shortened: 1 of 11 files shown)


frontend/model/state.js
```diff
@@ -214,22 +185,33 @@ const mutations = {
...
+  // The getters in DefineContract are designed to be compatible with Vuex!
+  // When they're used in the context of DefineContract, their 'state' always refers
+  // to the state of the contract whose messages are being processed, regardless
+  // of what group we're in. That is why the definition of 'currentGroupState' in
+  // group.js simply returns the state.
+  //
+  // Since the getter functions are compatible between Vuex and our contract chain
+  // library, we can simply import them here, while excluding the getter for
+  // `currentGroupState`, and redefining it here based on the Vuex rootState.
+  ..._.omit(sbp('gi.contracts/group/getters'), ['currentGroupState']),
```
sandrina-p, line 211: What is this `.omit()` method? I searched it on MDN but couldn't find it...

frontend/model/contracts/Contract.js
```diff
@@ -6,54 +6,74 @@ import { GIMessage } from '~/shared/GIMessage.js'
...
-        contract.actions[action].validate(message.data, { state, meta: message.meta })
-        contract.actions[action].process(state, message)
+      [`${action}/process`]: function (message: Object, state: Object) {
+        const { meta, data, contractID } = message
+        // TODO: optimize so that you're creating a proxy object only when needed
+        const gProxy = gettersProxy(state, getters)
+        state = state || contract.state(contractID)
+        metadata.validate(meta, { state, ...gProxy, contractID })
+        contract.actions[action].validate(data, { state, ...gProxy, meta, contractID })
+        contract.actions[action].process(message, { state, ...gProxy })
```
sandrina-p, line 55: I think this approach where parameters are consistent across all fns. 👌

sandrina-p (approved):
I understand and liked this refactor! Just left a small question

frontend/model/state.js
sandrina-p: Oh nevermind, this is calling lodash! `..._.` was a really tricky "morse code" combination here 😅

pieer (approved):
Awesome work and great addition!

Issue #832, opened 2020-02-11
Implement new toggle component


mmbotelho:
### Problem

Some of our new pages are using a new component - a toggle!

### Solution

Implement this component according to the [specifications on Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=48%3A4). Please note that whatever label is used along with the toggle, should activate it as well (same as checkboxes and radio buttons). 

Let me know if you have any questions!

sandrina-p:
I'll pick it up during this week

Issue #763, opened 2019-11-05
Implement PayGroup page (V1)


taoeffect:
### Problem

The PayGroup page is unimplemented.

### Solution

Implement the first version of the PayGroup page. The second version is being discussed here: #757

Implement having multiple frozen payments distributions (see comment in Payments.vue).

Closing this issue implies closing #206 (cash-based payments)

-----

#### What doesn't work
- [ ] Figuring out what to do when a payment is marked as not received (no proper followup to this) — @mmbotelho @dotmacro please play with it and see if you can offer suggestions
- [ ] Proper display of currencies based on the group's currency setting
- [ ] Conversion of currencies or handling currency change proposals
- [ ] Payments list pagination
- [x] Display and record Late payments (started at #918)
- [x] Partial payments that are done in more than 2 parts. (e.g. $100 paid 25 + 40 + 35) 
- [ ] Support fetching old payments (+1 month ago)

#### TODO:
- [ ] Sorting and reverse sorting the payments list by clicking the tabs
- [ ] Implement and use the `prevMonthDueDate` computed property for late payments in `PaymentRowTodo.vue`
- [ ] Cancelling payments? Not sure how that flow is supposed to work, wasn't clear from Figma (cc @mmbotelho)
- [ ] Maybe make it so that payments can be marked as completed immediately when `paymentType === PAYMENT_TYPE_MANUAL` (see TODO in `/payment` in `group.js`), although this one isn't that important, it works well enough as it is now without it (two messages sent, one `/payment` followed by a `/paymentUpdate`)
- [ ] Making the payment date information use the time given by the server instead of the system clock (so no `new Date()` usage, but creating dates based on an API call, see #531 and the `/time` API defined on the backend)
- [ ] Implement anything I forgot to mention from the "what doesn't work" section above

#### Done #951
- [x] Making the `ESC` key clear the search field
- [x] Fixing a bug where if we switched from sending payments to receiving payments, we suddenly lose the "Sent" tab
- [x] Add tests to received payments and a scenario where a user changes from giving to receiving mincome

#### Done #917
- [x] Someone needs to go through and clean up all of the CSS — deduplicate the all of the copy/pasted junk in `PaymentRow*.vue` files and DRY it all so that it's all stored in one of the `.scss` files and not repeated
- [x] Create mixins and remove no longer used computed properties/methods from the `PaymentRow*.vue` files (and any other payment related files)
- [x] Delete `humanDate.js` and move the `humanDate` function into `time.js`
[...]

PR #847, opened 2020-02-23
#763 Implement pay group page frontend

pieer:


mmbotelho (changes requested):
Here's a list of details I found:

Captura de ecrã 2020-02-23, às 17 43 10
Captura de ecrã 2020-02-23, às 18 12 34

Everything else, UI-wise, is looking great!

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -243,9 +243,9 @@ export default {
 }
 
 .c-table {
-  width: 100%;
   table-layout: fixed;
-  margin: $size_2 0;
+  margin-top: $size_2;
+  margin-bottom: $size_2;
```
sandrina-p: spacings should use `$spacer` variable. `$size` is used for typography

frontend/views/pages/GroupSettings.vue
```diff
@@ -206,4 +207,18 @@ export default {
...
+.p-descritpion {
+  display: none;
+  margin-top: $spacer-xs;
+  padding-bottom: $spacer-md*3;
+
+  @include desktop {
+    display: block;
+  }
+
+  &.show-mobile {
```
sandrina-p: Is this class used?

frontend/views/pages/PayGroup.vue
```diff
@@ -3,14 +3,21 @@ page(
   pageTestName='payGroupPage'
   pageTestHeaderName='payGroupTitle'
 )
-  template(#title='') {{ L('Pay Group') }}
+  template(#title='') {{ L('Payment') }}
```
sandrina-p: Based on design it should be _Payments_

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+    tag='p'
+    data-test='openIncomeDetailsModal'
+    @click='openModal("IncomeDetails")'
+    :args='{ receive_or_send: needsIncome ? L("receiving") : L("sending"), r1: `<button class="link js-btnInvite">`, r2: "</button>"}'
+  ) You are currently {receive_or_send} mincome. You can change this at any time by updating your {r1}income details{r2}.
+
+  section.card
+    nav.tabs(
+      v-if='!needsIncome'
+      aria-label='navigation'
```
sandrina-p: a `nav` element is already announced as "navigation". the `aria-label` should be used to specify the navigation. I think `payments type` could be a nice description.

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+      @click='open = false'
+    )
+      a.tabs-link(
+        v-for='(link, index) in tabItems'
         :key='index'
+        :class='{ "tabs-link-active": activeTab === link.url}'
+        :data-test='`link-${link}`'
+        @click='activeTab = link.url'
```
sandrina-p: This clickable element behaves like a "tab" that controls some other part of the UI, without changing pages. For that reason, it should be a `button`, not an `a`. And as a button, it should have `aria-expanded="true|false"` depending if it's active or not.

CSS-Tricks published very recently a [Complete Guide to Links and Buttons](https://css-tricks.com/a-complete-guide-to-links-and-buttons/). Check it out any time you have doubts between one or another.

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+        span.tabs-notification(v-if='link.notification') {{ link.notification }}
+
+    form.c-search-form(
+      v-if='needsIncome || activeTab !== "todo"'
+      @submit.prevent='submit'
+    )
+      label.field
+        i18n.sr-only Search payments
+        .input-combo
+          .is-icon(:aria-label='L("Search")')
```
sandrina-p: We already have the description on line 58 "Search payments", I think there's no need to add it here too. It would be redundant.

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+          input.input(
+            type='text'
+            name='search'
+            data-test='search'
+            placeholder='Search payments...'
+            v-model='form.search'
+          )
+          button.is-icon-small(
+            v-if='form.search'
+            aria-label='Clear search'
```
sandrina-p: Missing L here

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+            @click='form.search = null'
+          )
+            i.icon-times
+
+    section.tab-section
```
sandrina-p: We are already inside a section (the .card), this can be just a normal div.

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+                    .button.is-icon-smaller.c-tip
+                      i.icon-info
+
+                  span.c-status Not received
+
+              td
+                .c-actions
+                  .c-actions-month(:class='!(index !== 0 && activeTab === "todo") ? "has-text-1" : "c-status-danger c-status"') {{ randomMonth() }}
+                  menu-parent
+                    menu-trigger.is-icon-small
```
sandrina-p: Missing aria-label at menu-trigger.

frontend/views/pages/PayGroup.vue
```diff
@@ -318,65 +515,196 @@ export default {
 <style lang="scss" scoped>
 @import "@assets/style/_variables.scss";
 
+.p-descritpion {
+  margin-top: -1.8rem;
```
sandrina-p: we should aim for values that match the DS unless extremely needed. I think here it should be -1.5rem.

frontend/views/pages/PayGroup.vue
```diff
@@ -318,65 +515,196 @@ export default {
...
 
+  @include desktop {
+    padding-top: 2.5rem;
+  }
+
   .c-title {
-    margin: $spacer;
+    margin: 1.5rem 0 0 0;
+    font-size: $size_4;
+    color: $text_1;
```
sandrina-p: why not using `.is-title-4` on the element?

frontend/views/pages/PayGroup.vue
```diff
@@ -3,14 +3,21 @@ page(
   pageTestName='payGroupPage'
```
sandrina-p: @pieer you are doing a great job here handling all the details and edge cages around payments, not an easy task!

Obviously, this file is getting ginormous and it's still missing some UI and connecting to the store. The bigger it gets, the harder it will be to modify/catch a bug in the future.

Because of that, I think we should extract some parts of the page into smaller components, maybe the sidebar, the search, and the footer, and focus only on the "tab/table view".

Another concern I have, but already discussed during the hackathon when implementing "Income Details" page: I think the amount of if statements and ternary around is cluttering the component markup way too much. I gave a suggestion on a segment of the component, let me know on that component what you think.

frontend/views/pages/PayGroup.vue
```diff
@@ -20,65 +27,205 @@ page(
...
+        table.table.table-in-card.c-payments
+          thead
+            tr(v-if='activeTab === "todo"')
+              i18n(tag='th') Send to
+              i18n.c-payments-amount(tag='th') Amount
+              i18n.c-payments-date(tag='th') Due in
+            tr(v-else)
+              i18n(tag='th') Sent to
+              i18n.c-payments-amount(tag='th') Amount
+              i18n.c-payments-date(tag='th') Date
```
sandrina-p: Heres a suggestion to get rid of this ternary that duplicates the markup twice:

```js
tr
  th {{ tableTitles.one }}
  th.c-payments-amount {{ tableTitles.two }}
  th.c-payments-date {{ tableTitles.three }}

// computed property
tableTitles () {
  return this.activeTab === "todo" ? {
    one: 'Send to', // with L() obviously.
    two: 'Amount',
    three: 'Due in'
  } : {
    one: 'Sent to'
    two: 'Amount'
    three: 'Date'
  }
}
```

Personally I like to keep the markup as clean as possible and deal with the logic (in this case text logic) on the JS side. What do you think?

frontend/views/pages/PayGroup.vue
pieer: I'm still not sure if we should add parameter to the url for the tabs. But ok, for the time being I'll change to buttons

frontend/views/pages/PayGroup.vue
pieer: I agree, that was only a first version

frontend/views/pages/PayGroup.vue
pieer: I personally don't think that "th {{ tableTitles.one }}" is much easier to read, but sure

frontend/views/pages/PayGroup.vue
sandrina-p: Awesome, was just to make sure we are on the same page 😊

frontend/views/pages/PayGroup.vue
sandrina-p: That can be a possibility too, but not sure if it's worth at the moment as an MVP. Keep them as buttons is for now simpler

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -0,0 +1,170 @@
...
+    ]),
+    currency () {
+      return currencies[this.groupSettings.mincomeCurrency].symbol
+    },
+    paymentSummary () {
+      const { sent, confirmed, amoutSent, amountTotal, paymentsTotal } = this.paymentStatus
+
+      return [
+        {
+          title: this.L('Payments completed'),
```
taoeffect: As a matter of style consistency and readability, could you replace instances of `this.L` with `L`?

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,291 @@
...
+        strong(v-else) {{payment.amountFormatted}}
+        .c-payments-amount-info(
+          v-if='index === 0 && (needsIncome || paymentsType === "received")'
+          :class='"c-status-warning"'
+        )
+          tooltip.c-name-tooltip(
+            v-if=''
+            direction='top'
+            :isTextCenter='true'
+            :text='payment.to + L(" marked this payment as not received.")'
```
taoeffect: User facing strings should be fully encapsulated sentences, since you never know where a language will place a person's name.

So this should be something like `L("{personName} marked this payment as not received.", { personName: payment.to })`

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,291 @@
...
+          )
+            .button.is-icon-smaller.c-tip
+              i.icon-info
+
+          span.c-status Not received
```
taoeffect: i18n

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,291 @@
...
+            v-else
+            tag='button'
+            type='button'
+            @click='reset'
+          ) Reset
+      td(v-if='paymentsType === "edit"')
+        label.field
+          .input-combo
+            input.input
+            .suffix $ USD
```
taoeffect: Use `currency.js` to get these string values, since that will help us localize things and adjust for whatever the user's native currency is

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -22,7 +22,7 @@ modal-base-template.has-background(ref='modal' :fullscreen='true')
             )
             button.is-icon-small(
               v-if='form.search'
-              aria-label='Clear search'
+              :aria-label='L("Clear search")'
```
taoeffect: Nice catch!

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,291 @@
...
+          strong.c-name {{payment.to}}
+        .c-user-month(:class='index === 0 ? "has-text-1" : "c-status-danger c-status"') {{ randomMonth() }}
+      td.c-payments-amount(v-if='paymentsType !== "edit"')
+        .c-payments-amount-info(v-if='index === 0 && (!needsIncome && paymentsType === "todo")')
+          i18n.c-payments-amount-text.has-text-1(
+            tag='span'
+            :args='{partial_amount: `<strong class="has-text-0">$20</strong>`, partial_total: payment.amountFormatted}'
+          ) {partial_amount} out of {partial_total}
+
+          span.c-status.c-status-primary Partial
```
taoeffect: i18n

taoeffect (commented):
Just leaving some quick i18n related feedback. Hope to have more in depth feedback tomorrow!

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -0,0 +1,170 @@
...
+          })
+        },
+        {
+          title: this.L('Amout sent'),
+          value: amoutSent,
+          max: amountTotal,
+          hasMarks: false,
+          label: this.L('{value} of {max}', {
+            value: `${this.currency}${amoutSent}`,
+            max: `${this.currency}${amountTotal}`
```
taoeffect: Whenever you display currency values, use the `displayWithCurrency` function of currencies.js. That function will place the currency symbol in the correct position (sometimes it comes before the amount, and sometimes it comes after).

frontend/views/containers/payments/PaymentsSearch.vue
```diff
@@ -0,0 +1,59 @@
...
+)
+  label.field
+    .input-combo
+      .is-icon(:aria-label='L("Search")')
+        i.icon-search
+      input.input(
+        type='text'
+        name='search'
+        data-test='search'
+        placeholder='Search payments...'
```
taoeffect: `L` function

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,232 @@
...
+            i18n.has-text-small.has-text-1(
+              tag='p'
+            ) Leave a message to the group members selected above.
+          .c-footer-action
+            input.switch#showComment(
+              type='checkbox'
+              name='displayComment'
+              @change='displayComment = !displayComment'
+            )
+            label.sr-only(for='displayComment') Toggle comment box
```
taoeffect: i18n?

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,232 @@
...
+    return {
+      displayComment: false,
+      form: {},
+      donePayment: false,
+      // Temp until modal params are merged
+      fakeStore: {
+        paymentsDistribution: [{
+          to: 'pierre',
+          amount: 910.99,
+          amountFormatted: '$910.99'
```
taoeffect: Although it's a fakeStore, we should still remind ourselves to use the `displayWithCurrency` function by using it here too

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,143 @@ page(
...
-      v-for='(item, index) in paymentSummary'
-      :key='index'
+    month-overview
+
+  i18n.p-descritpion.has-text-1(
+    tag='p'
+    data-test='openIncomeDetailsModal'
+    @click='openModal("IncomeDetails")'
+    :args='{ receive_or_send: `<strong>${needsIncome ? L("receiving") : L("sending")}</strong>`, r1: `<button class="link js-btnInvite">`, r2: "</button>"}'
+  ) You are currently {receive_or_send} mincome. You can change this at any time by updating your {r1}income details{r2}.
```
taoeffect: Hmm, this is a tricky sentence to localize. The issue here is that I don't think every language will be able to support this sort of translation style. In the first sentence, there are two separate translations goes on that are being combined together:

1. `You are currently {receive_or_send} mincome.`
2. The words `receiving` and `sending`, translated individually, and being placed into the first sentence.

The issue is that in some languages, it's possible that "receiving" might get translated one way by itself, but in the context of the full sentence, it might need to be translated another way.

So I recommend using a function to return one of two sentences for the first sentence:

1. `You are currently {b_}receiving{_b} mincome.`
2. `You are currently {b_}sending{_b} mincome.`

One way of doing this is to define a computed property that returns one of those sentences, and then insert it as a variable here.

taoeffect (changes requested):
OK, looks good. Just finished my first pass review. Please see these comments as well as the ones I left before (I forgot to set "request changes" on the previous review).

There is still some stuff to add that is based on @mmbotelho making designs. Specifically, we are waiting for the design for what the Payments page is supposed to look like before any income details have been added.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+          //-   :class='payment.paymentClass'
+          //- ) {{ payment.paymentStatusText }}
+
+        strong(v-else) {{currency(payment.amount)}}
+        .c-payments-amount-info(
+          v-if='index === 0 && (needsIncome || paymentsType === "received")'
+          :class='"c-status-warning"'
+        )
+          tooltip.c-name-tooltip(
+            v-if=''
```
taoeffect, line 57: What's going on with this `v-if`?

frontend/views/containers/payments/PaymentDetail.vue
```diff
@@ -0,0 +1,123 @@
...
+      i18n.has-text-1(tag='label') Mincome at the time
+      strong {{ currency(groupSettings.mincomeAmount) }}
+    li.c-payment-list-item.c-column
+      i18n.has-text-1(tag='label') Notes
+      p {{ fakeStore.note }}
+  form.buttons(novalidate @submit.prevent='submit')
+    i18n.button.is-danger.is-outlined.is-small(
+      tag='button'
+      @click='submit'
+    ) Cancel payment
```
taoeffect: There are two `@click` event handlers here, are they both necessary?

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
+        two: L('Amount'),
+        three: L('Due in')
+      } : {
+        one: L('Sent to'),
+        two: L('Amount'),
+        three: L('Date')
+      }
+    },
+    introTitle () {
+      let copy = this.needsIncome ? L('You are currently {strong_}receiving{_strong} mincome. ', this.LTags('strong')) : L('You are currently {strong_}sending{_strong} mincome. ', this.LTags('strong'))
```
taoeffect: Nicely done!

Only request here is if, for readability, can you import `LTags` at the top so that you can call it directly without needing to do `this.LTags`?

frontend/views/containers/payments/PaymentsList.vue
pieer: Logic to implement with real data if the payment is marked as not received. At the moment I just want to display it all the time to look at the design implementation

pieer: NB: all the conditions at the moment are temporary

frontend/assets/style/components/_forms.scss
```diff
@@ -413,3 +413,53 @@ p.error {
...
+    &:focus {
+      box-shadow: 0 0 0 2px $primary_1;
+    }
+  }
+}
+
+/* The toggle knob */
+.switch::after {
+  content: "";
+  border-radius: 50%;
```
sandrina-p: Great switch! Please, add it to the DS page.

Edit: I noticed it's missing the hover state.

frontend/assets/style/components/_tabs.scss
```diff
@@ -0,0 +1,52 @@
...
+    @include tablet {
+      padding: $spacer 1.5rem;
+    }
+
+    @include desktop {
+      padding: 1.5rem 2.5rem;
+    }
+
+    &:hover,
+    &-active {
```
sandrina-p: instead of `tab-active`, we could just us `.is-active` as we use in other places.

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
   pageTestName='payGroupPage'
   pageTestHeaderName='payGroupTitle'
 )
-  template(#title='') {{ L('Pay Group') }}
+  template(#title='') {{ L('Payments') }}
```
sandrina-p: This file should also be renamed to "Payments"

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
+  p.p-descritpion.has-text-1(
+    data-test='openIncomeDetailsModal'
+    @click='openModal("IncomeDetails")'
+    v-html='introTitle'
+  )
+
+  section.card(v-if='!(needsIncome && !hasPayments)')
+    nav.tabs(
+      v-if='!needsIncome'
+      aria-label='payments type'
```
sandrina-p: missing L()

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
-      :key='index'
+    month-overview
+
+  p.p-descritpion.has-text-1(
+    data-test='openIncomeDetailsModal'
+    @click='openModal("IncomeDetails")'
+    v-html='introTitle'
+  )
+
+  section.card(v-if='!(needsIncome && !hasPayments)')
```
sandrina-p: Took me a while to understand this if! 🤦‍♀ 
Suggestion: Remove the negation on the if `v-if='needsIncome && !hasPayments'` and switch the v-else content with this.

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
     )
-      h4.is-title-4 {{ item.title }}
-      progress-bar.c-progress(
-        :max='item.max'
-        :value='item.value'
-        :hasMarks='item.hasMarks'
+      button.is-unstyled.tabs-link(
+        v-for='(link, index) in tabItems'
+        :key='index'
+        :class='{ "tabs-link-active": activeTab === link.url}'
```
sandrina-p: I noticed that when I click a tab the content changes but the tab is not "active" visually.

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
+        :data-test='`link-${link}`'
```
sandrina-p: `link` is an object, this won't work on tests.

frontend/views/pages/PayGroup.vue
```diff
@@ -318,83 +230,84 @@ export default {
...
-    margin-right: $spacer;
-    margin-bottom: $spacer-sm;
+  @include desktop {
+    padding-top: 1.5rem;
   }
-}
 
-.c-info {
-  @include tablet {
+  > div {
```
sandrina-p: Coul you use a class instead of using a global selector?

frontend/views/pages/GroupSettings.vue
```diff
@@ -1,7 +1,8 @@
 <template lang='pug'>
 page.c-page
   template(#title='') {{ L('Group Settings') }}
-  template(#description='') {{ L('Changes to these settings will be visible to all group members') }}
+  template(#description='')
+    p.p-descritpion.has-text-small.has-text-1 {{ L('Changes to these settings will be visible to all group members') }}
```
sandrina-p: The text isn't small. It should be the default size.

frontend/views/pages/PayGroup.vue
```diff
@@ -318,83 +230,84 @@ export default {
...
+  margin-top: 1.5rem;
+}
+
+// Empty
 .c-container-empty {
   max-width: 25rem;
   margin: 0 auto;
-  padding: $spacer-xl $spacer $spacer;
   text-align: center;
+  padding-top: 2.5rem;
```
sandrina-p: Some places use `$spacer-[X]`, others don't. We should keep it consistent everywhere as much as possible... Or we use it, or we don't...

frontend/assets/style/components/_combobox.scss
```diff
@@ -0,0 +1,38 @@
+.combobox {
```
sandrina-p, line 1: Any reason to keep this out of `_forms.scss`? It's there where we keep all form input styles.

frontend/views/containers/payments/PaymentsSearch.vue
```diff
@@ -0,0 +1,59 @@
+<template lang='pug'>
+form.c-search-form(
+  @submit.prevent='submit'
+)
```
sandrina-p: This can totally be an agnostic `Search.vue` component to be reused at `Payments` and `GroupMembersAllModal`.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
+<template lang='pug'>
+table.table.table-in-card.c-payments(:class='{"c-payments-edit": paymentsType === "edit"}')
```
sandrina-p: same as `.is-active`, here can be just `.is-edit(ting)`

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+  thead
+    tr
+      th(v-if='paymentsType === "edit"')
```
sandrina-p, line 5: How can I see the `paymentsType` as "edit"? It only has "todo", "sent" and "received". 🤔 

Edit: Nevermind, found it! It's in case it's used at `RecordPayment.vue`. roger

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,234 @@
...
+        transition(name='slidedown')
+          label.field(v-if='displayComment')
+            i18n.sr-only.label Leave a message
+            textarea.c-comment(
+              rows='4'
+            )
+
+        .buttons
+          i18n.is-outlined(
+            tag='button'
```
sandrina-p, line 47: Inside a `<form>`, all buttons that are not a "submit" should explicitly have a `type="button"`, otherwise when you press Enter while focused in another button (example: Cancel or Add a note Switch), it will always submit the form instead. Another approach (simpler and better in my opinion) is to change form submit to just `@submit.prevent=''` (do nothing). That way the user has to explicitly click "Register payments".

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,234 @@
...
+            @click='closeModal'
+          ) Cancel
+
+          i18n.is-success(
+            tag='button'
+            @click='submit'
+            :disabled='$v.form.$invalid'
+            :args='{number: "0"}'
+          ) Register {number} payment
```
sandrina-p: if the number is 1 it should say "payment", otherwise it should say "payments". Weird language 🤔

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,234 @@
...
+
+  .c-payment-success(v-else)
+    svg-success
+    i18n.is-title-2.c-title(tag='h2') Your payments were recorded
+    banner-simple(severity='info')
+      i18n Please consider supporting the development of Group Income by sending a donation!
```
sandrina-p: `sending a donation!` is a link

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,234 @@
...
+
+export default {
+  name: 'RecordPayment',
+  mixins: [validationMixin],
+  data () {
+    return {
+      displayComment: false,
+      form: {},
+      donePayment: false,
+      // Temp until modal params are merged
```
sandrina-p: It was already merged! 😄

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+              i.icon-info
+
+          i18n.c-status Not received
+
+      td
+        .c-actions
+          .c-actions-month(:class='!(index !== 0 && paymentsType === "todo") ? "has-text-1" : "c-status-danger c-status"') {{ moment(payment.date).format('MMMM') }}
+          payments-list-menu(
+            v-if='paymentsType !== "edit"'
+            :paymentsType='paymentsType'
```
sandrina-p, line 73: When paymentsType is "received" the menu is empty. I think it should show "Payment details" and "I did not receive this"

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+  tbody
+    tr(
+      v-for='(payment, index) in payments'
+      :key='index'
+    )
+      td(v-if='paymentsType === "edit"')
+        input.input.c-checkbox(type='checkbox' v-model='payment.checked')
+      td
+        .c-user
+          avatar-user.c-avatar(:username='payment.to' size='xs')
```
sandrina-p: `:username='payment.to'` isn't working... it should be sent the username, not the display name.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+            tag='button'
+            type='button'
+            @click='reset'
+          ) Reset
+
+      td(v-if='paymentsType === "edit"')
+        label.field
+          .input-combo
+            input.input(:value='payment.amount')
+            .suffix {{symbolWithCode}}
```
sandrina-p: On mobile, it should display only the symbol.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+      th.c-payments-amount {{ titles.two }}
+      th.c-payments-date {{ titles.three }}
+
+  tbody
+    tr(
+      v-for='(payment, index) in payments'
+      :key='index'
+    )
+      td(v-if='paymentsType === "edit"')
+        input.input.c-checkbox(type='checkbox' v-model='payment.checked')
```
sandrina-p: You're missing our own checkbox based on DS.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+        background: var(--#{$name}_2);
+        color: var(--#{$name}_0);
+      }
+
+      .c-tip {
+        background: var(--#{$name}_0);
+      }
+    }
+  }
+}
```
sandrina-p: These would give a nice `.pill` CSS component, similar to `.button`

sandrina-p (changes requested):
Looking good! Please see the comments I left!

mmbotelho (changes requested):
Here's what I found:

- The name of the page, in the sidebar, is still **Pay Group**. It should be **Payments**;
- I noticed that I didn't specify the hover and active styles for the tabs. [I added the correct styles to Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=294%3A249) - the implemented version should look the same;

Captura de ecrã 2020-03-02, às 15 15 27

Captura de ecrã 2020-03-02, às 15 14 24

Captura de ecrã 2020-03-02, às 15 13 06

Captura de ecrã 2020-03-02, às 13 54 27

Captura de ecrã 2020-03-02, às 13 53 18

Captura de ecrã 2020-03-02, às 13 52 47

Captura de ecrã 2020-03-02, às 13 46 58

mmbotelho:
Forgot to mention that the implemented mobile versions are missing a cancel button. Here's the designed version:

Captura de ecrã 2020-03-02, às 15 30 20

mmbotelho:
As @sandrina-p mentioned today on a Figma comment, the button below should have a copy that is consistent with the page title. Can you please change the copy as well? 

Captura de ecrã 2020-03-02, às 15 54 56

frontend/assets/style/components/_combobox.scss
pieer: It's a selector, not a form element. Like tab is not in form either

frontend/assets/style/components/_forms.scss
sandrina-p: One thing: Add to the PR description or commit message that this closes #832

frontend/views/pages/PayGroup.vue
taoeffect: And the sidebar nav menu also be renamed :)

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -0,0 +1,307 @@
...
+<script>
+import { mapGetters } from 'vuex'
+import AvatarUser from '@components/AvatarUser.vue'
+import Tooltip from '@components/Tooltip.vue'
+import PaymentsListMenu from '@containers/payments/PaymentsListMenu.vue'
+import currencies from '@view-utils/currencies.js'
+import moment from 'moment'
+
+export default {
+  name: 'PaymentTable',
```
taoeffect: This and the filename must match.

frontend/views/containers/payments/PaymentsList.vue
taoeffect: Then there needs to be a `// TODO: ` comment explain exactly what needs to go here

frontend/views/containers/payments/PaymentsListMenu.vue
```diff
@@ -0,0 +1,100 @@
...
+        @click='cancelPayment(payment.to, payment.hash)'
+      )
+        i18n Cancel this payment
+
+      menu-item(
+        v-if='needsIncome'
+        tag='button'
+        item-id='message'
+        icon='times'
+        @click='cancelPayment(payment.to, payment.hash)'
```
taoeffect, line 40: This `payment` variable needs to be defined somewhere. Clicking this menu will result in an exception. Make sure that this PR doesn't generate any exceptions in the console by using the `fakeStore`.

If it's impossible to avoid generating exceptions in the JS methods, then comment out the offending JS code and write `// TODO: ` psuedo code about what you expect the code to do, so that you or others can then fill it out with the appropriate code.

frontend/views/pages/PayGroup.vue
```diff
@@ -3,157 +3,147 @@ page(
...
-    .c-summary-item(
-      v-for='(item, index) in paymentSummary'
-      :key='index'
+    month-overview
+
+  p.p-descritpion.has-text-1(
+    data-test='openIncomeDetailsModal'
+    @click='openModal("IncomeDetails")'
+    v-html='introTitle'
+  )
```
taoeffect: Per the updated design from @mmbotelho, there is no text displayed at all when the income details have not been entered yet. Instead we need the standard "Add your income details" card here that exists in other pages (e.g. the same one that's on the Contributions page when no income details are entered).

taoeffect (changes requested):
Great work @pieer! I think between myself, @mmbotelho, and @sandrina-p this PR now has enough feedback on it for you. :)

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,234 @@
...
+          i18n.is-outlined(
+            tag='button'
+            @click='closeModal'
+          ) Cancel
+
+          i18n.is-success(
+            tag='button'
+            @click='submit'
+            :disabled='$v.form.$invalid'
+            :args='{number: "0"}'
```
taoeffect: This should reflect the number of checked boxes. I think you have the logic for doing this in this PR

frontend/views/containers/payments/RecordPayment.vue
taoeffect: Yeah, this is another are where a computed property can be used to generate the string, e.g.: something like:

```js
return number === 1 ? L('Register 1 payment') : L('Register {number} payments', { number })
```

frontend/views/pages/PayGroup.vue
pieer: What do you suggest... $spacer * 2.5 instead of 2.5rem?

frontend/views/containers/payments/PaymentsList.vue
pieer: It sending the username I believe..

pieer: I just tested and it's working. You were probably looking at the fake data in the popup

frontend/views/pages/PayGroup.vue
pieer: I'm happy to update all $spacer-[X] when possible, but i find $spacer * 2.5 very ugly

mmbotelho (changes requested):
- The toggle component has the wrong hover styles.[ Please check this out, if you scroll down abit, you'll find the specifications for the toggle.](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=48%3A4). Also feel free to leave it as it is for now and we can fix it later on issue 823;

Captura de ecrã 2020-03-08, às 16 59 44
Captura de ecrã 2020-03-08, às 17 01 54
Captura de ecrã 2020-03-08, às 17 10 49

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -5,48 +5,33 @@ modal-base-template.has-background(ref='modal' :fullscreen='true')
...
-          :args='{ groupMembersCount }'
-          data-test='memberCount'
-        ) {groupMembersCount} members
+      i18n.c-member-count.has-text-1(
+        v-if='searchText && searchCount > 0'
+        tag='div'
+        :args='{ searchCount: `<strong>${searchCount}</strong>`, result: `<strong>${searchCount === 1 ? L("result") : L("results")}</strong>`, searchTerm: `<strong>${searchText}</strong>`}'
+        data-test='memberSearchCount'
+        compile
+      ) Showing {searchCount} {result} for "{searchTerm}"
```
taoeffect: This string has the same issue as the other one, where variable is being used to insert an individually translated word. In general, this should be avoided, because we cannot assume to know how other languages will pluralize works.

For example, in English, if the `searchCount === 1`, we say "1 result", but another language might have the condition backwards, where it says effectively "1 results", or some other weirdness. Similarly, to the translator, the variables are assumed to be values, not words, so they might have difficulty figuring out what to do with `{result}`, thinking it represents a number or some other value.

So in these situations it's again best to build full sentences:

- `Showing 1 result for "{searchTerm}"`
- `Showing {searchCount} results for "{searchTerm}"`

This gives the translators all the information they need to figure out what sort of sentence this is and how it should be translated natively.

So here I recommend not using `i18n`, and instead using the `L` function inside of a computed property, to return one of the two strings in the unordered list above ☝️

taoeffect: This string has the same issue as the other one, where variable is being used to insert an individually translated word. In general, this should be avoided, because we cannot assume to know how other languages will pluralize works.

For example, in English, if the `searchCount === 1`, we say "1 result", but another language might have the condition backwards, where it says effectively "1 results", or some other weirdness. Similarly, to the translator, the variables are assumed to be values, not words, so they might have difficulty figuring out what to do with `{result}`, thinking it represents a number or some other value.

In these situations it's again best to build full sentences:

- `Showing 1 result for "{searchTerm}"`
- `Showing {searchCount} results for "{searchTerm}"`

This gives the translators all the information they need to figure out what sort of sentence this is and how it should be translated natively. Notice we do not use `{searchCount}` for the first case.

So here I recommend not using `i18n`, and instead using the `L` function inside of a computed property, to return one of the two strings in the unordered list above ☝️ 

As a general rule, I would look at the use of `L()` functions inside of the `:args` of `i18n` tags as a sign that whole-sentence translation through computed properties should be used for this sentence instead.

taoeffect (changes requested):
In this quick review (which is still ongoing, not quite finished but I'll let you know as soon as it is), I left [an i18n-related comment](https://github.com/okTurtles/group-income-simple/pull/847/files#r389421863) above (that I posted before this review), and during testing also identified the following issues:

### console warning still appears

When entering payment details and visiting the Payments page, this still appears:

> Relative positioning of table rows and row groups is now supported. This site may need to be updated because it may depend on this feature having no effect.

### Search + clear input icons disappears on hover

When hovering the mouse over the search field, the search icon disappears:

Screen Shot 2020-03-08 at 6 26 26 PM

Becomes:

Screen Shot 2020-03-08 at 6 26 28 PM

Same happens for the "X" clear input icon.

### Other notes

Per your request @pieer, I'll continue looking through this (very large) PR and posting comments on anything else that I notice as I go through the files. I've looked through most of it it already, but there are still some files remaining. On tomorrow's call I'll let you know whether or not I managed to finish!

frontend/views/components/Search.vue
```diff
@@ -0,0 +1,82 @@
...
+        @click='$emit("input", "")'
+      )
+        i.icon-times
+</template>
+
+<script>
+import L from '@view-utils/translations.js'
+
+export default {
+  name: 'PaymentsSearch',
```
taoeffect: Must match filename

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,252 @@
...
+          :payments='paymentsDistribution'
+          paymentsType='edit'
+        )
+
+        .c-footer
+          .c-footer-info
+            i18n.has-text-bold(
+              tag='h4'
+              :args='{ optional: `<span class="has-text-small has-text-1 has-text-normal">(${L("optional")})</span>`}'
+            ) Add a note {optional}
```
taoeffect: This needs to be:

```
Add a note {span_}optional{_span}
```

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,252 @@
...
+          ) {{ registerPaymentCopy }}
+
+  .c-payment-success(v-else)
+    svg-success
+    i18n.is-title-2.c-title(tag='h2') Your payments were recorded
+    banner-simple(severity='info')
+      i18n(
+        compile
+        :args='{ r1: `<a class="link" href="https://donorbox.org/okturtles-donation" target="_blank">`, r2: "</a>"}'
+      ) Please consider supporting the development of Group Income by {r1}sending a donation{r2}!
```
taoeffect: A few things: here `compile` attribute isn't necessary because the HTML being used is not a Vue.js component but normal HTML.

We also don't want them to just consider donating, but to actually donate, so please update this sentence to read:

```
Please support the development of Group Income by {r1}sending a donation{r2}!
```

(I know the copy in the design is different, but let's just improve it now) cc @mmbotelho can you update the design docs as well?

frontend/views/containers/payments/RecordPayment.vue
pieer: How do you specify the class?

frontend/views/containers/payments/RecordPayment.vue
mmbotelho: Done @taoeffect @pieer !

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -0,0 +1,251 @@
...
+          :payments='paymentsDistribution'
+          paymentsType='edit'
+        )
+
+        .c-footer
+          .c-footer-info
+            i18n.has-text-bold(
+              tag='h4'
+              :args='{ span_: `<span class="has-text-small has-text-1 has-text-normal">`, _span: "</span>"}'
+            ) Add a note {span_}(optional){_span}
```
taoeffect: `optional` instead of `{optional}`

frontend/views/components/Search.vue
```diff
@@ -0,0 +1,82 @@
...
+        :value='value'
+        @input='$emit("input", $event.target.value)'
+      )
+
+      button.is-icon-small(
+        v-if='value !== ""'
+        :aria-label='L("Clear search")'
+        @click='$emit("input", "")'
+      )
+        i.icon-times
```
taoeffect, line 26: @pieer Everything is coming together! It's all looking really good!

Your recent changes fixed the search icon disappearing, but this X icon is still disappearing when the mouse is over the input field:

Screen Shot 2020-03-09 at 1 58 00 PM

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -5,48 +5,30 @@ modal-base-template.has-background(ref='modal' :fullscreen='true')
       i18n.is-title-2.c-title(tag='h2') Group members
 
     .card.c-card
-      form(
-        @submit.prevent='submit'
+      search(
+        :placeholder='L("Search payments...")'
+        :label='L("Search for a payment")'
+        v-model='searchText'
```
taoeffect, line 11: This text copy is for the payments search, not for the Group Members modal.

taoeffect (changes requested):
@pieer this PR is really coming together, I'm only finding very small issues! See the comments.

Also, I while testing the changes in this PR, I noticed an issue in the `GroupMembersAllModal.vue` that is unrelated to payments, and it's something you can either fix as part of boyscouting for this PR or leave it for a later issue (totally up to you!). And that's that in Group Members modal, clicking the "Remove Member" button on someone doesn't bring up the proposal modal, instead it displays this warning in the console: `Missing username to display RemoveMember modal`.

frontend/assets/style/components/_tabs.scss
sandrina-p: Could you tell me why you do you prefer to use `.tab-link-active` here instead of `.is-active` as we do in other places?

frontend/views/components/Search.vue
```diff
@@ -0,0 +1,83 @@
...
+
+    label: {
+      type: String,
+      required: true
+    }
+  },
+
+  watch: {
+    value () {
+      this.$emit('input', this.value)
```
sandrina-p, line 54: maybe `emit` should be called `change` to follow the native event name `@change`.

frontend/views/components/tabs/TabWrapper.vue
```diff
@@ -17,7 +17,7 @@
         a.tab-link.no-border(
           v-for='(links, index) in tabItem.links'
           :key='index'
-          :class='{ "tab-active": activeTab === links.index, "has-text-white": isDarkTheme}'
+          :class='{ "is-active": activeTab === links.index, "has-text-white": isDarkTheme}'
```
sandrina-p, line 20: Hum... here we use `is-active`, but at `tabs.scss` in only has `.tab-link-active`. Maybe `.tab-link-active` is stale?

frontend/views/components/Search.vue
sandrina-p: Oh, nevermind. I got it now. by using `input`, the `v-model` bind happens automatically!

frontend/assets/style/components/_pills.scss
```diff
@@ -1,9 +1,29 @@
...
+    font-weight: 100;
+  }
+
+  &.has-background-dark {
+    background: $text_0;
+    color: $general_2;
+  }
+
+  @each $name in $colors {
+    &--#{$name} {
```
sandrina-p: all the other components with modifiers, follow the naming convention `.is-#{$same}`, (ex: button.is-primary). I believe we should do the same here: `.pill.is-primary` instead of `.pill--primary` to keep consistency.

And there's no need to do `@extend .pill;` following that approach.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -2,7 +2,7 @@
 page(pageTestName='dashboard' pageTestHeaderName='groupName' v-if='groupSettings.groupName')
   template(#title='') {{ groupSettings.groupName }}
 
-  add-income-details-widget(v-if='!hasIncomeDetails')
+  add-income-details-widget(v-if='!hasIncomeDetails' :welcomeMessage='true')
```
sandrina-p, line 5: rename to `hasWelcomeMessage`, so it's more intuitive that's a Boolean, not a String.

frontend/views/pages/Home.vue
```diff
@@ -111,12 +111,12 @@ export default {
 }
 
 .logo {
-  width: 8rem;
-  margin: 2rem auto;
+  width: $spacer-xxl;
```
sandrina-p, line 114: `$spacer` should be used for spacing only: margin and padding.
I think using it to define an element size (width, height) is incorrect because it's not the same thing.

I'd leave `8rem`.

frontend/views/pages/Join.vue
```diff
@@ -177,8 +177,8 @@ export default {
 @import "@assets/style/_variables.scss";
 
 .c-logo {
-  min-width: 8rem;
-  width: 8rem;
+  min-width: $spacer-xxl;
+  width: $spacer-xxl;
```
sandrina-p, line 181: Same here as the previous comment

frontend/views/pages/Mailbox.vue
```diff
@@ -338,13 +338,13 @@ export default {
   display: flex;
   align-items: center;
   padding: 0 1rem;
```
sandrina-p, line 340: 0 $spacer-md;

sandrina-p (changes requested):
Some details around. Also, #832 is not linked to this PR yet.

frontend/assets/style/components/_pills.scss
pieer: I was sure you would say that.. I'm thinking to change the other way around not following the boostrap style, but BEM actually.. because I never been a big fan of having multiple class on top of each others where you depend on order of the css.

frontend/views/components/tabs/TabWrapper.vue
pieer: Same as next comment, is-active is problematic especially when you have the fake impression that that style is encapsulated but the actual child component wrapper would be affected by .is-active class... or here where it will decrease the readability of the code where everything is encapsulated, but now can't anymore. 
But ok, I'll change to is-active

frontend/views/pages/Home.vue
pieer: Next PR..

frontend/views/components/Search.vue
pieer: Correct, it was a test i forgot to replace by shorthand

frontend/assets/style/components/_pills.scss
sandrina-p: Hum... Depend on order of the css? Here we can do .pill.is-primary or .is-primary.pill, it would work.
Or are you talking about when writing  the css? So far we didn’t have any issue with CSS order, have we?

pieer: It's not because you didn't had issue that we should use a technique which is not bug proof.. has you probably know .pill.is-primary or .is-primary.pill is the same, that's not what i'm talking about.
If you have a generique class like .is-primary in your colors.css, you're dependent of where the tabs.css is located in main.css compare to the colors.css
I agree with you that this won't be the case here because we redefine .is-primary in the tab class.. But that's just assumption and therefore less effective that the BEM solution. How many time you had to redefine on the component level the priorities?
So now the question is: do we want to improve or stick to bad habit?

pieer: Another advantage of using BEM is separation of concern between markup and css. What do you reckon?

pieer: Anyway probably a different PR to do that

sandrina-p (dismissed):
One more page ready to rock! Well done @pieer 😎

frontend/assets/style/components/_pills.scss
sandrina-p: We always use `.is-primary` within the context of the component. We never use `.is-primary` has a standalone class, so we are never dependent on where the CSS is placed or what's its order. Which means, we are bug-proof as in BEM.

```scss
// .is-primary is always a modifier "suffixed" to its block.
// CSS will always output as excepted regardless of its order.

.button.is-primary { ... }

.tabs.is-primary { ... }

.pills.is-primary { ... }

// We never declare .is-primary alone
.is-primary { ... }
```

If there's something I'm still missing, we can continue this discussion in a new issue and refactor our approach in a new PR.

mmbotelho (dismissed):
Perfect @pieer! 🎉

taoeffect (changes requested):
### Error while searching payments

Screen Shot 2020-03-10 at 9 08 32 PM

I created a group and invited one user. User A (that created the group) is sending income.

I went to the SENT tab with User A and performed a search. A console error (shown above) appeared.

I did the same with User B, searching for payments, and a similar console error appeared.

Likewise the error appeared for the GroupMembersModal when doing a search, so I'm guessing it's originating in the Search component?

### Payments in TODO hidden while a search is performed

Screen Shot 2020-03-10 at 9 09 14 PM

While a search is being done in the SENT tab, it appears to be applied to the contents of both the TODO and RECEIVED tabs as well, when I think they should be totally unrelated searches?

taoeffect (approved):
Approved! Fantastic job @pieer!! Especially with the quick fixes! 🎉

frontend/assets/style/components/_forms.scss
```diff
@@ -219,7 +219,6 @@ textarea,
     align-items: center;
     justify-content: center;
     display: flex;
-    z-index: 1; // to be above input, always
```
sandrina-p, line 221: Why did you remove this z-index?

I also noticed when hovering the selectcombo the icon disappears:
image

Please verify

sandrina-p (approved):
Only noticed a small issue when hovering the .selectcombo Besides that all seems fine!

mmbotelho (approved):

frontend/assets/style/components/_forms.scss
pieer: I didn't remove the z-index, I **increase** the z-index. After I pull the last branch from master lots of element started to be hidden like this...

pieer: I didn't remove the z-index, I **increase** the z-index. After I pull the last branch from master lots of element started to be hidden like this...

PR #854, opened 2020-02-29
#802 Add loading state to form submission

### LINKED ISSUE #802: Add loading state to form submission

sandrina-p:
Closes #802

spin_final

### Loading state added to the following forms:
- Log in
- Sign up
- Join Group
- Group Creation
- Income Details (Save)
- Propose to Add Members
- Change Mincome (proposal or not)
- Group Settings
- User Settings
- Non-monetary Contributions (add, edit, remove)
- Design System

Did I miss any form? Please let me know!

###  🐌 How to simulate a slower internet:
1. Open the Devs Tools
2. Go to Network 
3. Change Throttling from "No Throttling" / "Online" to something else slower (ex: Slow 3G)

**Note:** The spin icon only appears after 500ms. This way the user won't even see the loading in case the load is fast enough.

image

### DIFF
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -79,15 +79,13 @@ References to help you:
 - [A11Y Style Guide Resources](https://a11y-style-guide.com/style-guide/section-resources.html)
 
 ### Submitting a Form
-When performing an action that requires data to be sent, the application UI should be updated to reflect this "submitting state". The most common scenario is to add a "spin" animation to the submit button, through `data-loading="true|false"`.
-
-Note: `data-loading` value should be "true" or "false", as a `String` instead of `Boolean`, so [Vue keeps it on the DOM even when "false"](https://github.com/vuejs/vue/issues/5860).
+When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. The button needs the CSS class `.is-loader` by default. When it's submitting it needs `data-loading="true"`. The button should also disabled to prevent duplicated submitions by click.
 
 ```pug
-  i18n(
+  i18n.is-loader(
     tag='button'
     :data-loading='ephemeral.isSubmitting'
-    :disabled='$v.form.$invalid || ephemeral.isSubmitting === "true"'
+    :disabled='$v.form.$invalid || ephemeral.isSubmitting'
   ) Login
 ```
 
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -78,8 +78,8 @@ References to help you:
 - [Accessibility Insights Extension](https://accessibilityinsights.io/)
 - [A11Y Style Guide Resources](https://a11y-style-guide.com/style-guide/section-resources.html)
 
-### Loading State
-When performing an action that requires to send data (ex: submit a form), the application UI is aupdated to reflect the "loading state". The most common case is to add a "spin" animation to the submit button through `data-loading`.
+### Submitting a Form
+When performing an action that requires data to be sent, the application UI should be updated to reflect this "submitting state". The most common scenario is to add a "spin" animation to the submit button, through `data-loading="true|false"`.
 
 Note: `data-loading` value should be "true" or "false", as a `String` instead of `Boolean`, so [Vue keeps it on the DOM even when "false"](https://github.com/vuejs/vue/issues/5860).
 
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -80,13 +80,15 @@ References to help you:
 
 ### Submitting a Form
 
-When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. All the logic needed to do it is under `ButtonSubmit` component.
+When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. Use `ButtonSubmit` component to do it:
 
 ```pug
   button-submit(
     @click='handleSubmit'
   ) Login
 ```
+
+All the needed logic to toggle the spinning information is handled by `ButtonSubmit` based on `@click` handler event. Read the source code for further explanation.
 
 
 ## Testing Style Guide
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -78,6 +78,20 @@ References to help you:
 - [Accessibility Insights Extension](https://accessibilityinsights.io/)
 - [A11Y Style Guide Resources](https://a11y-style-guide.com/style-guide/section-resources.html)
 
+### Loading State
+When performing an action that requires to send data (ex: submit a form), the application UI is aupdated to reflect the "loading state". The most common case is to add a "spin" animation to the submit button through `data-loading`.
+
+Note: `data-loading` value should be "true" or "false", as a `String` instead of `Boolean`, so [Vue keeps it on the DOM even when "false"](https://github.com/vuejs/vue/issues/5860).
+
+```pug
+  i18n(
+    tag='button'
+    :data-loading='ephemeral.isSubmitting'
+    :disabled='$v.form.$invalid || ephemeral.isSubmitting === "true"'
+  ) Login
+```
+
+
 ## Testing Style Guide
 
 We use [Mocha](https://mochajs.org/) for the unit tests and [Cypress](https://www.cypress.io/) for end-to-end (E2E) tests. **All new functionality must have corresponding tests!**
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -80,13 +80,11 @@ References to help you:
 
 ### Submitting a Form
 
-When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. The button needs the CSS class `.is-loader` by default. When it's submitting it needs `data-loading="true"`. The button should also disabled to prevent duplicated submitions by click.
+When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. All the logic needed to do it is under `ButtonSubmit` component.
 
 ```pug
-  i18n.is-loader(
-    tag='button'
-    :data-loading='ephemeral.isSubmitting'
-    :disabled='$v.form.$invalid || ephemeral.isSubmitting'
+  button-submit(
+    @click='handleSubmit'
   ) Login
 ```
 
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -79,6 +79,7 @@ References to help you:
 - [A11Y Style Guide Resources](https://a11y-style-guide.com/style-guide/section-resources.html)
 
 ### Submitting a Form
+
 When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. The button needs the CSS class `.is-loader` by default. When it's submitting it needs `data-loading="true"`. The button should also disabled to prevent duplicated submitions by click.
 
 ```pug
--- a/docs/Style-Guide.md
+++ b/docs/Style-Guide.md
@@ -80,7 +80,7 @@ References to help you:
 
 ### Submitting a Form
 
-When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. Use `ButtonSubmit` component to do it:
+When performing an async action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. Use `ButtonSubmit` component to do it:
 
 ```pug
   button-submit(
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -11,7 +11,7 @@
 
 <script>
 export default {
-  name: 'ButtonSubmit', // TODO rename file to ButtonSubmit
+  name: 'ButtonSubmit',
   props: {
     disabled: Boolean
   },
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -50,7 +50,7 @@ export default {
     }
   }),
   methods: {
-    async submit (e) {
+    async submit (event) {
       if (this.ephemeral.isSubmitting) {
         return
       }
@@ -60,7 +60,11 @@ export default {
       // this.$listeners can await for async handlers.
       // An advantage over using $emit().
       // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
-      await this.$listeners.click(e)
+      try {
+        await this.$listeners.click(event)
+      } catch (error) {
+        console.error('ButtonSubmit exception:', error)
+      }
 
       this.ephemeral.isSubmitting = false
     }
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -12,7 +12,7 @@
 
 <script>
 /*
-Use ButtonSubmit on buttons that will trigger an async action.
+Use ButtonSubmit on buttons that will trigger an **async** action.
 
 button-submit(
 @click='handleLogin'
@@ -62,9 +62,12 @@ export default {
       // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
       try {
         await this.$listeners.click(event)
+        console.log('passou aqui? 1')
       } catch (error) {
         console.error('ButtonSubmit exception:', error)
       }
+
+      console.log('passou aqui? 2')
 
       this.ephemeral.isSubmitting = false
     }
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -1,15 +1,44 @@
 <template lang='pug'>
   button.is-loader(
+    type='submit'
     v-bind='$attrs'
     v-on='bindListeners'
-    type='submit'
     :data-loading='ephemeral.isSubmitting'
     :disabled='disabled || ephemeral.isSubmitting'
   )
     slot
+    i18n.sr-only(v-if='ephemeral.isSubmitting') Loading
 </template>
 
 <script>
+/*
+Use ButtonSubmit on buttons that will trigger an async action.
+
+button-submit(
+@click='handleLogin'
+) Login
+
+This will display a "loading state" on the button while the login is happening.
+
+The button has type="submit" to catch any way of form submittion (ex: press Enter).
+
+That way we don't need to use $refs on the parent.
+
+// ❌ DO NOT do this, it's INCORRECT:
+form(@submit.prevent='$refs.btnSubmit.click')
+
+// ✅ DO this instead:
+form(@submit.prevent='')
+
+`@submit.prevent` prevents the original behavior of a form submit (page reload)
+but it will still look for a submit element and its event handler (@click).
+
+In this case ButtonSubmit is the submit element thanks to type="submit".
+So, when pressing Enter, buttonSubmit(@click) gets called directly too.
+
+More details about this approach:
+https://github.com/okTurtles/group-income-simple/pull/854/files#r388638068
+*/
 export default {
   name: 'ButtonSubmit',
   props: {
@@ -25,11 +54,11 @@ export default {
       if (this.ephemeral.isSubmitting) {
         return
       }
-
       this.ephemeral.isSubmitting = true
 
-      // this.$listeners can await for async events.
-      // An advantage over $emit()
+      // Call the original @click handler.
+      // this.$listeners can await for async handlers.
+      // An advantage over using $emit().
       // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
       await this.$listeners.click(e)
 
@@ -40,8 +69,8 @@ export default {
     bindListeners () {
       return {
         ...this.$listeners,
-        // bind passed @click to custom submit fn
-        // so it goes through isSubmitting "guard"
+        // overrides passed @click handler with a custom @click handler
+        // so it goes through "isSubmitting guard"
         click: this.submit
       }
     }
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -63,12 +63,9 @@ export default {
       // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
       try {
         await this.$listeners.click(event)
-        console.log('passou aqui? 1')
       } catch (error) {
         console.error('ButtonSubmit exception:', error)
       }
-
-      console.log('passou aqui? 2')
 
       this.ephemeral.isSubmitting = false
     },
--- a/frontend/views/components/ButtonSubmit.vue
+++ b/frontend/views/components/ButtonSubmit.vue
@@ -1,6 +1,7 @@
 <template lang='pug'>
   button.is-loader(
     type='submit'
+    ref='btn'
     v-bind='$attrs'
     v-on='bindListeners'
     :data-loading='ephemeral.isSubmitting'
@@ -70,6 +71,10 @@ export default {
       console.log('passou aqui? 2')
 
       this.ephemeral.isSubmitting = false
+    },
+    focus () {
+      // Used to focus the btn from the parent (ex:GroupCreationModal)
+      this.$refs.btn.focus()
     }
   },
   computed: {
--- a/frontend/views/components/group-creation-steps/GroupRules.vue
+++ b/frontend/views/components/group-creation-steps/GroupRules.vue
@@ -130,9 +130,6 @@ export default {
              this.group.memberApprovalThreshold >= SUPERMAJORITY &&
              this.group.memberRemovalThreshold >= SUPERMAJORITY
     }
-  },
-  mounted () {
-    this.$emit('focusref', 'finish')
   }
 }
 </script>
--- a/frontend/views/components/group-creation-steps/GroupRules.vue
+++ b/frontend/views/components/group-creation-steps/GroupRules.vue
@@ -130,6 +130,9 @@ export default {
              this.group.memberApprovalThreshold >= SUPERMAJORITY &&
              this.group.memberRemovalThreshold >= SUPERMAJORITY
     }
+  },
+  mounted () {
+    this.$emit('focusref', 'finish')
   }
 }
 </script>
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -26,10 +26,10 @@ form(
   banner-scoped(ref='formMsg' data-test='loginError')
 
   .buttons.is-centered
-    i18n(
+    i18n.is-loader(
       tag='button'
       :data-loading='ephemeral.isSubmitting'
-      :disabled='$v.form.$invalid || ephemeral.isSubmitting === "true"'
+      :disabled='$v.form.$invalid'
       data-test='loginSubmit'
       type='submit'
     ) Login
@@ -62,7 +62,7 @@ export default {
         password: null
       },
       ephemeral: {
-        isSubmitting: 'false'
+        isSubmitting: false
       }
     }
   },
@@ -71,8 +71,8 @@ export default {
   },
   methods: {
     async login () {
-      if (this.ephemeral.isSubmitting === 'true') { return }
-      this.ephemeral.isSubmitting = 'true'
+      if (this.ephemeral.isSubmitting) { return }
+      this.ephemeral.isSubmitting = true
       try {
         await sbp('gi.actions/user/login', {
           username: this.form.name,
@@ -83,7 +83,7 @@ export default {
         console.error('FormLogin.vue login() error:', e)
         this.$refs.formMsg.danger(e.message)
       }
-      this.ephemeral.isSubmitting = 'false'
+      this.ephemeral.isSubmitting = false
     },
     forgotPassword () {
       // TODO: implement forgot password
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -28,7 +28,8 @@ form(
   .buttons.is-centered
     i18n(
       tag='button'
-      :disabled='$v.form.$invalid'
+      :data-loading='ephemeral.isSubmitting'
+      :disabled='$v.form.$invalid || ephemeral.isSubmitting === "true"'
       data-test='loginSubmit'
       type='submit'
     ) Login
@@ -61,7 +62,7 @@ export default {
         password: null
       },
       ephemeral: {
-        isSubmitting: false
+        isSubmitting: 'false'
       }
     }
   },
@@ -70,8 +71,8 @@ export default {
   },
   methods: {
     async login () {
-      if (this.ephemeral.isSubmitting) { return }
-      this.ephemeral.isSubmitting = true
+      if (this.ephemeral.isSubmitting === 'true') { return }
+      this.ephemeral.isSubmitting = 'true'
       try {
         await sbp('gi.actions/user/login', {
           username: this.form.name,
@@ -82,7 +83,7 @@ export default {
         console.error('FormLogin.vue login() error:', e)
         this.$refs.formMsg.danger(e.message)
       }
-      this.ephemeral.isSubmitting = false
+      this.ephemeral.isSubmitting = 'false'
     },
     forgotPassword () {
       // TODO: implement forgot password
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -74,6 +74,8 @@ export default {
       }
 
       try {
+        this.$refs.formMsg.clean()
+
         await sbp('gi.actions/user/login', {
           username: this.form.name,
           password: this.form.password
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -69,7 +69,7 @@ export default {
   methods: {
     async login () {
       if (this.$v.form.$invalid) {
-        this.$refs.formMsg.danger('The fields are incompleted.')
+        this.$refs.formMsg.danger(L('The form is invalid.'))
         return
       }
 
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -4,7 +4,7 @@ form(
   ref='form'
   name='formData'
   data-test='login'
-  @submit.prevent='login'
+  @submit.prevent=''
 )
   label.field
     i18n.label Username
@@ -26,13 +26,11 @@ form(
   banner-scoped(ref='formMsg' data-test='loginError')
 
   .buttons.is-centered
-    i18n.is-loader(
-      tag='button'
-      :data-loading='ephemeral.isSubmitting'
+    button-submit(
+      @click='login'
       :disabled='$v.form.$invalid'
       data-test='loginSubmit'
-      type='submit'
-    ) Login
+    ) {{ L('Login') }}
 </template>
 
 <script>
@@ -42,6 +40,7 @@ import sbp from '~/shared/sbp.js'
 import { nonWhitespace } from '@views/utils/validators.js'
 import PasswordForm from '@containers/access/PasswordForm.vue'
 import BannerScoped from '@components/banners/BannerScoped.vue'
+import ButtonSubmit from '@components/Button.vue'
 import L from '@view-utils/translations.js'
 import validationsDebouncedMixins from '@view-utils/validationsDebouncedMixins.js'
 
@@ -53,16 +52,14 @@ export default {
   ],
   components: {
     PasswordForm,
-    BannerScoped
+    BannerScoped,
+    ButtonSubmit
   },
   data () {
     return {
       form: {
         name: null,
         password: null
-      },
-      ephemeral: {
-        isSubmitting: false
       }
     }
   },
@@ -71,8 +68,6 @@ export default {
   },
   methods: {
     async login () {
-      if (this.ephemeral.isSubmitting) { return }
-      this.ephemeral.isSubmitting = true
       try {
         await sbp('gi.actions/user/login', {
           username: this.form.name,
@@ -83,7 +78,6 @@ export default {
         console.error('FormLogin.vue login() error:', e)
         this.$refs.formMsg.danger(e.message)
       }
-      this.ephemeral.isSubmitting = false
     },
     forgotPassword () {
       // TODO: implement forgot password
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -40,7 +40,7 @@ import sbp from '~/shared/sbp.js'
 import { nonWhitespace } from '@views/utils/validators.js'
 import PasswordForm from '@containers/access/PasswordForm.vue'
 import BannerScoped from '@components/banners/BannerScoped.vue'
-import ButtonSubmit from '@components/Button.vue'
+import ButtonSubmit from '@components/ButtonSubmit.vue'
 import L from '@view-utils/translations.js'
 import validationsDebouncedMixins from '@view-utils/validationsDebouncedMixins.js'
 
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -68,6 +68,11 @@ export default {
   },
   methods: {
     async login () {
+      if (this.$v.form.$invalid) {
+        this.$refs.formMsg.danger('The fields are incompleted.')
+        return
+      }
+
       try {
         await sbp('gi.actions/user/login', {
           username: this.form.name,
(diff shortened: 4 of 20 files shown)


mmbotelho (dismissed):
Looking good @sandrina-p ! Thank you for leaving instructions on how to simulate a slow internet 😄

frontend/views/containers/access/SignupForm.vue
```diff
@@ -72,17 +72,17 @@ export default {
         email: null
       },
       ephemeral: {
-        isSubmitting: false
+        isSubmitting: 'false'
```
pieer: Why did you use a string and not a boolean?

pieer (dismissed):
Just left a question, otherwise all good for me! Great work and great addition!

frontend/views/containers/access/SignupForm.vue
taoeffect: Have a look at the Style-Guide.md update in this PR it seems to explain it?

pieer: Yes, but I'm not sure why we keep this in the dom. I mean I understand that we use [data-loading] has a selector, but that's what I question.

sandrina-p: We need to keep `data-loading="false"` in the DOM so the loading animation fades in/out properly. See CSS `[data-loading]` selector - it contains the styles needed

frontend/views/containers/access/LoginForm.vue
```diff
@@ -28,7 +28,8 @@ form(
   .buttons.is-centered
     i18n(
       tag='button'
-      :disabled='$v.form.$invalid'
+      :data-loading='ephemeral.isSubmitting'
+      :disabled='$v.form.$invalid || ephemeral.isSubmitting === "true"'
```
sandrina-p: @pieer and @taoeffect there's another way to keep `data-loading` and using Boolean. Transform the Bool into a String the moment it's added to the DOM. Everywhere else it can be a bool as usual.

```
:data-loading='(ephemeral.isSubmitting).toString()'
:disabled='$v.form.$invalid || ephemeral.isSubmitting'
```

What approach do you prefer?

frontend/views/containers/access/SignupForm.vue
pieer: Maybe use a class for the button instead of the [data-loading] selector?

sandrina-p: I tried with classes, but we would need two classes. One to say that is a "loader button" and another to apply the actual loading. 

```
button.button.loader(:class='{"is-loading": isSubmitting"}
```

What do you prefer? With `[data-loading]` at least is just one new addition instead of two.

pieer: How about .loader and use the attribute [data-loading]? The only thing to change is add class loader to the button that have a loader and the boolean attribute will trigger the animation?
Sorry I should have say what I was thinking from the beginning

sandrina-p: Oh yes, that's an option too. I'll do that then and let you know when it's ready

sandrina-p: Done! `.is-loader` + `data-loading="Bool"`

frontend/views/pages/Contributions.vue
```diff
@@ -241,17 +241,21 @@ export default {
...
         )
+        console.log('send msg!', groupProfileUpdate)
         await sbp('backend/publishLogEntry', groupProfileUpdate)
+        // TODO/BUG - Discover why "value" gets added to the list
+        // but Contribution input takes longer to unmount. (in Slow 3G).
       } catch (e) {
         console.error('handleNonMonetary', e)
         alert(e.message)
       }
+      done()
```
taoeffect: - Why was this added? Is it truly necessary? This is already an asynchronous method. Callbacks are usually not used with them.
- If for some reason a callback is really needed, then it's usually called with the error object when there's an error. However, I think this should be removed, and I'm not seeing where it's used

frontend/views/pages/Contributions.vue
```diff
@@ -241,17 +241,21 @@ export default {
...
       sbp('okTurtles.events/emit', OPEN_MODAL, modal)
     },
-    async handleNonMonetary (type, value) {
+    async handleNonMonetary (type, value, done) {
       try {
         const groupProfileUpdate = await sbp('gi.contracts/group/groupProfileUpdate/create',
           { [type]: value },
           this.$store.state.currentGroupId
         )
+        console.log('send msg!', groupProfileUpdate)
```
taoeffect: Seems like this should be removed?

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -123,7 +123,7 @@ export default {
           this.$refs.formMsg.danger(L('Failed to change mincome. {codeError}', { codeError: e.message }))
           this.ephemeral.currentStep = 0
         }
-
+        done()
```
taoeffect: Same concerns as with the other callback function in Contributions.vue (see comment below)

frontend/views/containers/proposals/AddMembers.vue
```diff
@@ -139,10 +139,10 @@ export default {
           break
         }
       }
-
       if (!hasFailed) {
         this.ephemeral.currentStep += 1 // Show Success step!
       }
+      done()
```
taoeffect: Same concerns as with the other callback function in Contributions.vue (see comment below)

docs/Style-Guide.md
```diff
@@ -78,6 +78,18 @@ References to help you:
 - [Accessibility Insights Extension](https://accessibilityinsights.io/)
 - [A11Y Style Guide Resources](https://a11y-style-guide.com/style-guide/section-resources.html)
 
+### Submitting a Form
```
taoeffect, line 81: Remember headers need an empty line after them.

frontend/views/containers/group-settings/GroupCreationModal.vue
```diff
@@ -105,6 +106,9 @@ export default {
         return
       }
 
+      if (this.ephemeral.isSubmitting) { return }
+      this.ephemeral.isSubmitting = true
```
taoeffect: So this is a lot of the same code repeated everywhere these buttons need to be used. It reminds me of the same issue we had with the scoped-banners, and how we created the scoped banner component that is self-removing and therefore doesn't need any new variables or logic added to the parent component.

If we had a custom button component, it could automatically check itself to see if it's in the submitting state, by accepting a submit handler and waiting for the promise to finish (`async` functions are syntactic sugar for wrapping the body in a promise). If it's clicked while the promise hasn't finished, it does nothing. This would remove the need for creating and checking this `ephemeral.isSubmitting` variable.

Do you think this is a component you could create? You could pass down all the other attributes that are passed to it to the `button` using [`v-bind="$attrs"`](https://vuejs.org/v2/api/#vm-attrs) and maybe [`v-on="$listeners"`](https://vuejs.org/v2/api/#vm-listeners) (although you'd have to be careful with that one, since we want to catch and wrap the `@click` submit handler ourselves for this component to do its self-contained magic).

taoeffect (changes requested):
I assume this PR is still being worked on (per our call), but I also noticed some early feedback I could leave so I went ahead to not keep you waiting.

frontend/views/containers/access/LoginForm.vue
taoeffect: Is this question still relevant given what you wrote above re

> Done! `.is-loader` + `data-loading="Bool"`

?

sandrina-p: No, it's not. The question was written before finding the new solution ;)

frontend/views/containers/group-settings/GroupCreationModal.vue
sandrina-p: > Do you think this is a component you could create? 

When I joined Group Income this was one of the very first things I pointed about buttons. Not being a component. GI is the only "modern framework based" project I ever work on where a button isn't a Vue component.

That way we could just pass `isLoading` prop and it would handle [data-loading] correctly everywhere. The only ugly part is we would need to use `L` instead of `i18n` when using the component

```pug
button(:isLoading='isSubmitting' @click={submit})
  | {{ L('Submit') }}
```


However, we'd still need to add the verification on the submit handler (in the parent)  because the form may be submitted through another way (example: press Enter button) 

So... Your suggestion would work 90% of the places, but not everywhere (ex: on Login we can use Enter to login).

frontend/views/containers/proposals/Mincome.vue
sandrina-p: `done()` is used to let the parent know when the submit is finished and set back `isSubmitting` to false. In this case (and the other 3 proposals), it reports to `ProposalTemplate` L178-180.

If we remove `done()` after the `submit` completes the button will keep showing the loading state. This isn't noticeable if the submit is successful because the modal closes, but if it fails, that's a problem. The user cannot try to submit again because the parent state wasn't updated. That's the purpose of `done()`

frontend/views/pages/Contributions.vue
sandrina-p: As explained in the other comment below, this is needed to let the child (`Contribution.vue`) know when to reset the `isSubmitting` back to false.

frontend/views/pages/Contributions.vue
```diff
@@ -241,17 +241,22 @@ export default {
...
+    async handleNonMonetary (type, value, done) {
+      console.log('handle new-value')
       try {
         const groupProfileUpdate = await sbp('gi.contracts/group/groupProfileUpdate/create',
           { [type]: value },
           this.$store.state.currentGroupId
         )
         await sbp('backend/publishLogEntry', groupProfileUpdate)
+        // TODO/BUG - Discover why "value" gets added to the list
+        // but Contribution input takes longer to unmount. (in Slow 3G).
```
sandrina-p: @taoeffect see the [video recordings I left on Slack](https://okturtles.slack.com/files/U9SGRFKT6/FUJS6KSE5/inspect_flow.mp4) (on a thread).

This is a Slow 3G flow of adding one contribution and adding a second contribution with logs strategically placed before and after the `this.$emit`, on `handleNonMonetary` and `beforeUpdate`.
Compare the logs with their placement in the code (I pushed the logs).

So, it seems the state is updated in the moment `events/emit GI Message` is processed. At this moment, the components receive the new list of contributions. But we still need to process `settings/save user-1`. Until then, the `Contribution.vue` will show that "buggy state" because `done()` wasn't called yet. `done()` will be called only after `settings/save` is completed.

As I initially suspected and told you during our video call on the weekend, there was something odd with the data flow, not related to Vue.

What if we saved the user settings before processing the GI Message? That way the moment the state updates, there are no more network requests to make and we don't risk showing incorrect states.

Note: This is only perceivable on 3G networks... if the effort to fix is high, it can be done later, I'd say.

frontend/views/containers/group-settings/GroupCreationModal.vue
taoeffect: > That way we could just pass `isLoading` prop 

My suggestion is to pass only the submit handler, e.g. `@click='submit'`. `isSubmitting` / `isLoading` doesn't need to be passed I don't think. Again the point is to avoid having to create these variables repeatedly in the parent components. The button will call `submit` and will change its state based on when the promise resolves or fails.

> However, we'd still need to add the verification on the submit handler (in the parent) because the form may be submitted through another way (example: press Enter button)

There's still gotta be a way to avoid introducing `isSubmitting`. For example, you could tag the button instance as a `ref` and then include a method on the button component that is called by the Enter handler.

e.g. something like:

```
@keyup.enter='$refs.submitButton.submit()'
```

sandrina-p: Hum... Using `$refs` could be an option. @pieer what do you think?

pieer: So pressing 'enter' would call the button action which handle the loading state and call the parent submit function... That works for me.

sandrina-p: Exactly. I'll work on it then! 👩‍💻

sandrina-p: So, @pieer and @taoeffect, I think I came up with a _neat solution_, but I only did it in 3 places. I need your approval before replacing it everywhere.

The places with this new approach are:
- LoginForm
- Change group mincome (not a proposal)
- Add a contribution.

There's a new component `ButtonSubmit`. This button should only be used, as the name says, to perform a submit action. We have tons of buttons around the app and most of them don't need a _Loading State_, so I think we can keep using them as usual (native HTML).

**How to use ButtonSubmit:**

```pug
form(@submit.prevent='')
  // ...
  button-submit(@click="login") {{ L('Login') }}
```

That's all! By using `form(@submit.prevent='')` + `button-submit`, all submit techniques will be handled, even if the user presses "Enter" instead of clicking the button. This is possible thanks to `type='submit'` inside `ButtonSubmit`.

**Quick question**: How do you prefer to pass the button text?

```pug
// Slot: - More flexible (ex: pass an icon)
button-submit(@click="login") {{ L('Login') }}

// Prop: - Cleaner. No need to use {{ }}
button-submit(@click="login" :text='L("Login")'
```

Now... Go read `button-submit` code. While I was exploring how to "guard" the @click event and how to approach asynchronous event handlers I came up with _one line solution_:

```js
await this.$listeners.click(e)
```

Not sure if it's the best, but it works! Meanwhile I [asked it on Stackoverflow](https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners)

What do you think?

taoeffect: > Quick question: How do you prefer to pass the button text?

Definitely the first style, though you can use `i18n` even there too (see examples below), but possibly even better would be to continue using `i18n`, just remember to add the `compile` attribute to get it to work.

So I would say either of these work:

```pug
i18n(
  tag='button-submit'
  compile 
  ...other-attrs...
) Login

// OR

button-submit(
  ...other-attrs...
)
  i18n Login
```

I'm still not seeing where you've used `$refs` though, and perhaps you could send another commit with this component being used for the lines referenced in the code in this file too?

I'll have a closer look at how you're handling `@click`, `$attrs` and `$listeners`, and definitely could use @pieer's help here too. Great work! 👍

sandrina-p: With this solution, there's no need to use `$refs`! 😄 (or did I miss something?)  I'll change the code to use i18n.

taoeffect: > With this solution, there's no need to use $refs! 😄 (or did I miss something?)

Good question, I've no idea. This is what I'm gonna try to figure out! :D

kamerat: > Hum... Using $refs could be an option. @pieer what do you think?

Just dropping by a comment:

Using a ref would be considered bad practice due to the fact that if the internal API of the button changes at some point, the function that you are trying to call could not exist, hence creating a bug. So IF you were to go for a ref solution, you really need to be aware of changes.

taoeffect: @BlitZz96 - thanks for dropping by! The internal API is controlled entirely by us, so that's not an issue.

Right now I'm trying to figure out whether `$listeners.click` makes sense or not because I think multiple `@click` handlers can be registered in the parent component. In which case I'm not sure that approach would work, and so $refs might be the correct way to go about it. We're already using the `$refs` approach in another component, and it works great.

taoeffect: OK, I think I get it, the `$listeners` refers only to the event handlers registered on this instance of the component.

However, @sandrina-p, the use-case for `$refs` was when `@keyup.enter` was registered on another part of the form. In that case you'd need a `$refs` reference (as shown in my earlier comment above), wouldn't you?

sandrina-p: No, we don't need `$refs`. When pressing Enter, `form(@submit.prevent='')` gets called. And when a form is submitted, it automatically triggers its submit element action. In this case, `button-submit` is the submit element thanks to `type="submit"`. So, `@click` gets called directly.

taoeffect: Ah! I see, that's interesting, thanks for the explanation! Worthy of a comment in the component as that's kinda surprising behavior / not super clear.

sandrina-p: Yes, I can add an extensive comment in ButtonSubmit explaining this. Just to clarify, this "press enter -> submit"  is not a "Vue" thing, this behavior would be the same everywhere, even with just plain javascript. It's how HTML forms work.

docs/Style-Guide.md
```diff
@@ -78,6 +78,17 @@ References to help you:
...
+
+When performing an action that requires data to be sent, the application UI should be updated to reflect this "loading state". The most common scenario is adding a "spin" animation to the submit button. All the logic needed to do it is under `ButtonSubmit` component.
+
+```pug
+  button-submit(
+    @click='handleSubmit'
+  ) Login
+```
+
```
taoeffect, line 90: Per the convo [here](https://github.com/okTurtles/group-income-simple/pull/854/files#r388620817), can you also include here an explanation about why `$refs` isn't needed (e.g. because of `form(@submit.prevent='')` behavior)?

> No, we don't need `$refs`. When pressing Enter, `form(@submit.prevent='')` gets called. And when a form is submitted, it automatically triggers its submit element action. In this case, `button-submit` is the submit element thanks to `type="submit"`. So, `@click` gets called directly.

frontend/views/components/ButtonSubmit.vue
```diff
@@ -0,0 +1,50 @@
...
+    async submit (e) {
+      if (this.ephemeral.isSubmitting) {
+        return
+      }
+
+      this.ephemeral.isSubmitting = true
+
+      // this.$listeners can await for async events.
+      // An advantage over $emit()
+      // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
```
taoeffect, line 63: This stackoverflow link isn't needed, since I think you figured out the perfect solution all on your own! 😄 

Well done thinking to use `this.$listeners` like this, very clean, and yes, means we don't need to use `$refs`. 👍 

Can you instead include a link to the section in our style guide that talks about how to use this component?

And also a link to [the discussion on GH](https://github.com/okTurtles/group-income-simple/pull/854/files#r388638068) which includes an interesting alternative directive-based solution from @BlitZz96?

frontend/views/containers/group-settings/GroupCreationModal.vue
kamerat: Would this solve your problem https://codepen.io/BlitZZ96/pen/RwPLEXN
This example uses a custom directive that acts like the default vue `@click`, but it updates the element that uses it's HTML to a loader while the promise is unresolved.

frontend/views/components/ButtonSubmit.vue
```diff
@@ -0,0 +1,50 @@
...
+      await this.$listeners.click(e)
+
+      this.ephemeral.isSubmitting = false
```
taoeffect: So here we need to be careful, because if an exception is thrown on `.click(e)`, then `isSubmitting` will not be reset back to `false`, so we need to wrap the call to `.click` in a try/catch.

```js
try {
  await this.$listeners.click(e)
} catch (err) {
  console.error('ButtonSubmit exception:', err)
}

this.ephemeral.isSubmitting = false
```

frontend/views/containers/group-settings/GroupCreationModal.vue
taoeffect: > This example uses a custom directive

@BlitZz96 Interesting solution! I think that approach might work too! I think @sandrina-p's approach also works well though, particularly since we want to make sure it works well with the `<enter>` key when a form is submitted.

By using `@click`, `$listeners`, and `form(@submit.prevent='')`, @sandrina-p has ensured that the button will behave correctly and in the same way regardless of whether the form was submitted by clicking on the button, or because the user pressed enter.

frontend/views/containers/proposals/Mincome.vue
taoeffect: We can remove `done()` now with `ButtonSubmit`?

frontend/views/containers/group-settings/GroupCreationModal.vue
taoeffect: @BlitZz96 OTOH, looking more closely at your solution, I think yours could also work with `form(@submit.prevent='')` and the `<enter>` key.

So I honestly don't know which approach is better. Both could work well. If there's no clear advantage to either one, I say we go with what @sandrina-p came up with, but if there's a reason to prefer the directive please let us know!

taoeffect: One thing about directives: they don't have the ability to access and update component state right? We need to make sure that this button does nothing when it's clicked while it's in the loading state. So if the directive isn't able to toggle component state, then the component-based approach is the only thing that works.

kamerat: @taoeffect 
Okey, then these simple changes would do the job
https://codepen.io/BlitZZ96/pen/WNvZmZW

If simplicity is what you are after.. This is definitly simpler than using `@click`, `$listen` and `form(@submit.prevent='')`

kamerat: > We need to make sure that this button does nothing when it's clicked while it's in the loading state

This could be achieved by doing `el.disabled = true;` before the callback and false in `.then()`. or something similar

Well, I want you guys to do what is simpler for you, but ofc, both solutions has its pros and cons. One pro with this one is that you do not need a "loading" state in the parent, because the promise does that for you. Other plus is that you do not need to add a loader everywhere. A con might be that It needs some polishing in order to work IDENTICAL to `@click` (with the event modifiers .prevent, .native, .stop etc...)


> they don't have the ability to access and update component state right?

They do. You can access this via `vnode.context` [ref](https://stackoverflow.com/a/41590520/6178832)

taoeffect: @BlitZz96 Think you might've missed my comment above about the state being used to prevent double submissions. In your example, pressing the enter key twice in rapid succession will result in the button permanently in a loading state, and `handleSubmit` getting called twice.

taoeffect: (Deleted my other comment, we posted at the same time.)

> This could be achieved by doing `el.disabled = true;` before the callback and false in `.then()`. or something similar

Ah, yes that looks like it could work!

> both solutions has its pros and cons.

Yeah, I like both solutions! Will be curious to hear what @sandrina-p thinks when she's back tomorrow.

If we go with @sandrina-p's approach, we should still bookmark yours, maybe via a link to these comments in the source. Thank you so much @BlitZz96 for sharing this! Always amazing to see so many different ways to solve a problem with Vue.js!

kamerat: No problem @taoeffect! This was probably more fun for me than you guys 😂 Thanks for the interesting problem.

taoeffect: @BlitZz96 I learned something new thanks to you! You are always welcome to drop by! ^_^

frontend/views/pages/Contributions.vue
taoeffect: @sandrina-p I don't think this related to `'gi.db/settings/save'` or `done()`.

`'gi.db/settings/save'` does not affect the UI at all. All it does is save the latest state to localStorage. It doesn't impact the UI in any way besides the loading of the state the next time the user logs in or refreshes the page.

The `done()` callback seems to have been removed in a recent commit (except in `AddMembers`, where I think it also should be removed), so that's no longer relevant here either.

AFAICT, the issue seems to be that in `Contribution.vue`, `contributionsList` (which is bound to `ourGroupProfile.nonMonetaryContributions`), is updated, and the input form is still shown.

So because `contributionsList` is updated, and the form is still shown, `vuelidate` seems to get triggered, and it evaluates the form, sees that the same input is being added to the list, and shows the error.

How to solve this? I'm not entirely sure. I can see at least two different approaches:

1. Delay vuelidate's evaluation by manually using `$touch()` and `$dirty()` at the appropriate times (see [Without v-model](https://vuelidate.js.org/#sub-without-v-model) section of vuelidate docs)
2. Immediately hide the form after submit is hit (as long as the form is valid), so that the vuelidate error isn't visible (this might be the simpler approach, so try it first)

frontend/views/containers/contributions/Contribution.vue
```diff
@@ -143,24 +150,34 @@ export default {
...
     },
-    handleSubmit () {
+    async handleSubmit () {
       if (this.$v.form.$invalid) {
         this.cancel()
       } else {
         if (this.isAdding) {
-          this.$emit('new-value', 'nonMonetaryAdd', this.form.contribution)
+          console.log('new-value called!', this.contributionsList)
+          await this.$listeners['new-value']('nonMonetaryAdd', this.form.contribution)
```
taoeffect: Think some of these changes can be reverted now to use the `$emit` as before, see [comment below](https://github.com/okTurtles/group-income-simple/pull/854/files#r388661801)

frontend/views/containers/group-settings/GroupCreationModal.vue
sandrina-p: Hi @BlitZz96  👋 , thanks for your help

> Using a ref would be considered bad practice due to the fact that if the internal API of the button changes at some point.

I used to think the same. But then I realized that using `props` suffers from the same problem. The moment you change/delete a prop, the component can break. At the moment we only have one component that is used through `$refs` and it's documented to prevent any internal API accidental change.  You can check `components/banners/BannerScoped.vue`.

Now, back to the loading state. Using a `directive` as you proposed is an interesting solution! However, I find the `directive` not as efficient as the `ButtonSubmit` component because:
- It contains additional logic to handle the form using `el.querySelector`. `ButtonSubmit` doesn't need it.
- We need to perform more DOM changes than just adding a spinner. One of them is to change `disabled`. However, when it's not loading (a.k.a `false`), it requires some conditional logic based on the context. For example, at `LoginForm` the `disabled` status also depends if the form is valid or not. This would be the API for each approach:

```pug
    button(
      v-save='{ handler: login, disabled: "$v.form.$invalid" }'
    ) Login

    button-submit(
      @click='login'
      :disabled='$v.form.$invalid'
    ) Login
```

Using a solution based on component + props is cleaner to use. I think it's also easier to write thanks to the declarative approach that Vue components offer us. The same isn't true when writing a `directive`.

TLDR: I prefer using the component `ButtonSubmit` over the directive `v-save` because it's more flexible and cleaner.

frontend/views/pages/Contributions.vue
sandrina-p: > 'gi.db/settings/save' does not affect the UI at all. 

I didn't say it changed the UI. I'm saying it causes the click handler to take longer to resolve, so `Contribution` can hide...

1. No, that's a workaround solution that doesn't solve the root of the problem: Why isn't the click handler resolved as soon as the state gets updated.
2. This is our current approach (in master), without using a loading state. This kind of approach has a name: ["Optimistic UI"](https://uxplanet.org/optimistic-1000-34d9eefe4c05).

But this still doesn't explain why we have a delay between when the state is updated and when the click handler resolves. If you see the Network, there's a request being made during that period. That request is what's preventing the handler to finish as soon as the state is updated. And that request doesn't seem necessary because the state was already updated. So why does this extra request exists and what it is?

frontend/views/containers/proposals/Mincome.vue
sandrina-p: It was removed yes

frontend/views/components/ButtonSubmit.vue
sandrina-p: Done, please check it out. I also added an example on our Design System page. :)

frontend/views/pages/Contributions.vue
sandrina-p: Here's a GIF showing the extra request that takes time and it's preventing the click handler to resolve immediately after the state gets updated.
flow_net2

frontend/views/containers/contributions/Contribution.vue
sandrina-p: Please see the latest changes and comments.

frontend/views/pages/Contributions.vue
taoeffect: Shower thought: realized there's a simple fix: clear the input field as soon as the return key is pressed (and validations pass).

sandrina-p: What if there's an error? Then the user would need to add the text again... Another downside is while it's submitting the user doesn't see the text what can be strange.

taoeffect: > What if there's an error

There is no error because it's cleared after the validations have passed. This seems like the right solution, unless you can think of an even better one. If you don't like it but can't think of a better one, feel free to open an issue.

sandrina-p: An error related to the server might happen... And this isn't the right solution because this "state / request order issue" might happen in other scenarios too. I'll create an issue about it and linking it to this discussion.

taoeffect: > An error related to the server might happen...

That's OK. That will still appear and be visible.

frontend/views/containers/contributions/Contribution.vue
taoeffect: Hmm, what are you referring to specifically? Are you saying this way is the way it needs to be done, perhaps because of the `await`? (Did you double-check that's still true?)

frontend/views/containers/proposals/AddMembers.vue
taoeffect: Sorry, I'm not following... why is this `done()` needed?

sandrina-p: it isn't anymore. Was the only one I forgot to remove.

frontend/views/pages/Contributions.vue
sandrina-p: If we clean the input, an inline error will still appear saying the field is required.

I think we should move on as it is. This issue is bigger than this Contribution scenario. I'm sure it happens in other places. We just didn't detect it yet.

Issue opened: https://github.com/okTurtles/group-income-simple/issues/858

frontend/views/containers/contributions/Contribution.vue
sandrina-p: Yes, this way is the way it needs to be done. So we wait for the message to be processed before the loading state gets changed to `false` again.

frontend/views/pages/Contributions.vue
sandrina-p: @taoeffect, as discussed in our weekly video call, I changed the code to clean (reset) the input when submitting it. It doesn't solve the issue, but now at least the user doesn't see any "wrong " state nor inline error. Still, I think we should understand what's the purpose of that 2º network requests and if we can avoid it...

frontend/views/containers/access/LoginForm.vue
```diff
@@ -70,8 +68,11 @@ export default {
   },
   methods: {
     async login () {
-      if (this.ephemeral.isSubmitting) { return }
-      this.ephemeral.isSubmitting = true
+      if (this.$v.form.$invalid) {
+        this.$refs.formMsg.danger('The fields are incompleted.')
```
taoeffect: `formMsg.clean()` needs to be called at the beginning of this function to clear this error right?

Also the `L` function needs to wrap this string.

frontend/views/containers/group-settings/GroupCreationModal.vue
```diff
@@ -100,7 +101,7 @@ export default {
     },
     async submit () {
       if (this.$v.form.$invalid) {
-        // TODO: more descriptive error message, highlight erroneous step
+        // TODO: more descriptive error message. Perhpahs highlight error step
```
taoeffect: Typo, `Perhpahs` => `Perhaps`

frontend/views/pages/DesignSystem.vue
```diff
@@ -1149,6 +1174,15 @@ export default {
...
       this.articles[last].link.classList.add('last')
+    },
+    async saveIn2500ms () {
+      const dumbPromiseThatTakes2500ms = () => new Promise((resolve, reject) => {
+        setTimeout(() => {
+          resolve()
+        }, 2500)
+      })
+
+      await dumbPromiseThatTakes2500ms()
```
taoeffect: Async/await is unnecessary here, can be written simply:

```js
saveIn2500ms () {
    setTimeout(resolve, 2500)
  })
}
```

frontend/views/containers/access/SignupForm.vue
```diff
@@ -62,27 +61,24 @@ export default {
...
-        isSubmitting: false
       }
     }
   },
   methods: {
     async signup () {
-      // Prevent autocomplete submission when empty field
-      if (!this.form.name || !this.form.email || !this.form.password || this.ephemeral.isSubmitting) {
+      if (this.$v.form.$invalid) {
+        this.$refs.formMsg.danger('The fields are incompleted.')
```
taoeffect: Same two concerns as above in LoginForm (see comment there)

frontend/views/containers/contributions/Contribution.vue
```diff
@@ -142,21 +143,24 @@ export default {
       }
       return this.isFilled ? this.handleSubmit() : this.handleDelete()
     },
-    handleDelete () {
-      this.$emit('new-value', 'nonMonetaryRemove', this.initialValue)
+    async handleDelete () {
+      await this.$listeners['new-value']('nonMonetaryRemove', this.initialValue)
```
taoeffect, line 147: One potential issue with this new way of calling the `new-value` handler is that it will cause an exception/error if the handler is not an asynchronous function.

So either you need to check for the existence of `this.$listeners['new-value'].then` before calling `await`, or you need to clearly name and document (with comments) this event as something that must be bound to an asynchronous function only.

taoeffect (changes requested):
See comments! Also, I wasn't able to test this because when I got to the final step of the "Create Group" flow I got the following error in Firefox Dev Edition:

```
uncaught Vue error in v-on handler: TypeError: "this.$refs[ref].focus is not a function"
    focusRef GroupCreationModal.vue:96
```

Screen Shot 2020-03-10 at 7 45 15 PM

I don't know how the tests managed to pass...

taoeffect:
Also, now that the other Leave Group PR has been merged, can you update the buttons for that one too? (And maybe as many others as you can find?)

frontend/views/pages/DesignSystem.vue
sandrina-p: I explicitly used async/await to be more similar to the real-world...

frontend/views/containers/contributions/Contribution.vue
sandrina-p: > it will cause an exception/error if the handler is not an asynchronous function

I just did the test by calling a normal function and everything worked correctly, even if it's a sync function.

Based on [this explanation](https://stackoverflow.com/a/55263084/4737729) there isn't any perceivable consequence.

Also, at `ButtonSumibt.vue` docs, this is the first sentence:

> Use ButtonSubmit on buttons that will trigger an **async** action.

frontend/views/components/group-creation-steps/GroupRules.vue
```diff
@@ -130,9 +130,6 @@ export default {
              this.group.memberApprovalThreshold >= SUPERMAJORITY &&
              this.group.memberRemovalThreshold >= SUPERMAJORITY
     }
-  },
-  mounted () {
-    this.$emit('focusref', 'finish')
```
sandrina-p: > Also, I wasn't able to test this because when I got to the final step of the "Create Group" flow I got the following error in Firefox Dev Edition:
>> uncaught Vue error in v-on handler: TypeError: "this.$refs[ref].focus is not a function"
    focusRef GroupCreationModal.vue:96

@taoeffect, I also see the error but it didn't prevent me from creating the group as normal.

These lines were the cause: it tried to focus the submit button once the final step is visible, but that's a bad practice because it's messing with the default keyboard navigation (here are other important interactive elements before the submit button).

A more correct solution would be to instead focus on the beginning of the step each time we move between steps.

Anyway, that's a problem for another PR. For now, I just removed the forced focus.

taoeffect: Hmm, that explanation does not make sense to me. This exception does not happen in `master`, and yet these lines are present in `master`.

The problem has to be something with this PR.

frontend/views/containers/contributions/Contribution.vue
taoeffect: > Also, at ButtonSumibt.vue docs, this is the first sentence:

OK, but that has nothing to do with `Contribution.vue` and this `new-value` event?

> I just did the test by calling a normal function and everything worked correctly, even if it's a sync function.

Hmm.. that's odd. OK, I will test it out myself later today and if it works as you say then we can ignore this.

frontend/views/pages/DesignSystem.vue
taoeffect: This is real world too, and way simpler. You're creating a promise in either case. Might as well not over-complicate things. Save lines please!

frontend/views/components/group-creation-steps/GroupRules.vue
sandrina-p: These lines are calling the ref "finish" at `GroupCreationModal.vue` that is the submit button. But now the button is a component `ButtonSubmit`, not an element `<button />`, so you cannot focus a component. But as I said, focusing the submit button here is a bad practice. So, I removed the code responsible for triggering the button focus.

frontend/views/containers/contributions/Contribution.vue
sandrina-p: `ButtonSubmit` calls the listener `@click`. That listener triggers `async handleDelete` which then awaits for the $listener `'new-value'`. That's the relationship between both.

frontend/views/pages/DesignSystem.vue
sandrina-p: okay, changed.

frontend/views/components/group-creation-steps/GroupRules.vue
taoeffect: > so you cannot focus a component

Ah I see. Being able to focus a component though seems like a potentially legitimate use-case, can you check to see if there's a way to make this work anyway (even if you feel that it shouldn't be focused in this case)? I.e. is there a way to add the `focus` method to the component so that an exception isn't thrown when it's used like this?

taoeffect: In other words, I don't think we should lose the ability to focus buttons just because we're using `ButtonSubmit.vue` in some places.

sandrina-p: Yes, ofc it's possible. Added it back and did the needed change to `ButtonSubmit.vue` :)

frontend/views/containers/access/LoginForm.vue
taoeffect: I think this was missed. Still needs an `L` function.

While we're updating this, can you please update the copy too to read: `L('The form is invalid.')`?

frontend/views/containers/access/SignupForm.vue
taoeffect: I think this was missed. Still needs an `L` function.

While we're updating this, can you please update the copy too to read: `L('The form is invalid.')`?

frontend/views/pages/DesignSystem.vue
```diff
@@ -1254,6 +1279,11 @@ export default {
         el.link.classList.remove('last')
       })
       this.articles[last].link.classList.add('last')
+    },
+    async saveIn2500ms () {
```
taoeffect: One last thing (might as well fix this since the other `L` -related string stuff needs to be fixed too): this is a minor detail, and I don't think it ultimately affects the behavior of the code, but when you are returning a promise like, and not using `await` inside the function, the `async` modifier should not be added to the function.

Having the `async` modifier there, in this case, will (I think), cause it to return a promise, that returns this promise, requiring two promise resolutions, which is just kinda unnecessary. So the `async` modifier isn't needed, since we're already returning a promise (and not calling `await` anywhere else).

frontend/views/components/ButtonSubmit.vue
```diff
@@ -0,0 +1,91 @@
...
+      }
+      this.ephemeral.isSubmitting = true
+
+      // Call the original @click handler.
+      // this.$listeners can await for async handlers.
+      // An advantage over using $emit().
+      // More at: https://stackoverflow.com/questions/60554270/vuejs-difference-between-emit-and-listeners
+      try {
+        await this.$listeners.click(event)
+        console.log('passou aqui? 1')
```
taoeffect: Dev logging needs to be removed

frontend/views/components/ButtonSubmit.vue
```diff
@@ -0,0 +1,91 @@
...
+      } catch (error) {
+        console.error('ButtonSubmit exception:', error)
+      }
+
+      console.log('passou aqui? 2')
```
taoeffect: Dev logging needs to be removed

frontend/views/containers/contributions/Contribution.vue
taoeffect: OK, I just verified that there's no issue with non-async handlers! 👍

taoeffect (changes requested):
Minor changes requested for some missed feedback from previous reviews + some minor new changes. After this I think it's GTM!

pieer (approved):
Look good to me!

taoeffect (approved):
Excellent!

PR #856, opened 2020-03-05
Implement Leave Group

### LINKED ISSUE #792: Finish Group Settings page

sandrina-p:
Closes #792 

- Implement "Leave Group" feature and add e2e tests.
- "Delete Group" card was removed since it won't be used for the prototype.

Note: You can only leave the group if the group has >1 member. (to avoid having groups without members)


frontend/model/contracts/group.js
```diff
@@ -69,6 +69,33 @@ function initGroupProfile (contractID: string) {
   }
 }
 
+function removeMemberSideEffect (username, groupId) {
```
taoeffect: OK, I'm noticing a few issues, but they're definitely fixable.

1. Having a large function at the top of this file results in clutter. Ideally these files only contain a single call to `DefineContract`, some imports at the top, and that's it.
2. I think it does make sense to separate out and define only the parameters that are needed for this function, as you've done here, however, we need to make sure that this code is called in a single consistent way. Right now it's being called in two completely different ways: (1) via the function, as in the sideEffect for `removeOurselves` and `removeMember`, and (2), via `await sbp('gi.contracts/group/removeMember/process/sideEffect', message)`, as in `proposals.js` via `proposalVote`. This inconsistency must be fixed (it's poor style and results in confusion/unclarity).

I propose the following solution to both problems:

1. Create a folder called `sideEffects` here: `frontend/controller/sideEffects`
2. Create this file: `frontend/controller/sideEffects/group.js` (make sure this file is imported early on so that the sbp select in (step 4) is  registered)
3. Just like you did for `frontend/controller/actions/*`, use `sbp` to register selectors
4. Define the following selector and map it to this function: `'gi.sideEffects/group/removeMember'`
5. Then call this side effect from all of the other places (there should now be 3 calls: in the side effect functions for `removeMember` and `removeOurselves`, and in `proposals.js`)

frontend/model/contracts/group.js
```diff
@@ -395,30 +422,18 @@ DefineContract({
...
+    },
+    'gi.contracts/group/removeOurselves': {
+      validate: objectOf({
+        groupId: string
+      }),
+      process ({ data, meta }, { state }) {
+        Vue.delete(state.profiles, meta.username)
+      },
+      async sideEffect ({ data, meta }) {
+        removeMemberSideEffect(meta.username, data.groupId)
```
taoeffect: If you don't have an `await` in front of the call it won't wait for it to finish, and things will mysteriously break. One way to remember this is that every `async` function must have at least one call to `await` inside of it. I think this used to be something the linter would catch, and am not sure why it doesn't anymore.

(Same on line 425 above)

Then, per the comment at the very top of this review, convert this to

```js
await sbp('gi.sideEffects/group/removeMember', meta.username, data.groupId)
```

Don't forget to do the same in `proposals.js`.

frontend/views/pages/GroupSettings.vue
```diff
@@ -60,7 +60,7 @@ page.c-page
 
   invitations-table
 
-  page-section(:title='L("Leave Group")')
+  page-section(:title='L("Leave Group")' v-if='groupMembersCount > 1')
```
taoeffect: You should still be able to leave the group if you're the only member in it, but it should probably delete the group in that case.

Suggestion: since delete group is unimplemented, instead throw up an `alert(L('Leaving the group when you're the only person in it will delete it, but deleting groups is unimplemented at that moment.')` or something to that effect?

frontend/views/containers/group-settings/GroupLeaveModal.vue
```diff
@@ -93,7 +99,22 @@ export default {
...
+      }
+
+      try {
+        const message = await sbp('gi.contracts/group/removeOurselves/create',
+          {
+            groupId: this.currentGroupId
+          },
+          this.currentGroupId
+        )
+        await sbp('backend/publishLogEntry', message)
```
taoeffect: All this type of stuff should no longer be done in Vue files. Instead it should be moved into `frontend/controller/actions/group.js`.

Feel free to boyscout other vue.js files and move any leftover `/create` messages into the actions folder too (there are a few I saw that are still in vue files).

taoeffect (changes requested):
Great work @sandrina-p! First review done!

frontend/model/contracts/group.js
sandrina-p: I thought `await` was needed only if the function being called is async. In this case, `gi.sideEffects/group/removeMember` is not an async function (or at least ATM is not doing any async work). That's why I didn't add the `await`.

Anyway. I added it now, but would be nice to understand why it's really needed in this case.

ESlint has a rule to catch when we use `await` inside a function that isn’t `async`. I think a linter rule to detect if we call an async function without using `await` doesn’t exist.

frontend/views/containers/group-settings/GroupLeaveModal.vue
sandrina-p: Changed to `actions`. Did the same to Save Group Settings.

frontend/model/contracts/group.js
taoeffect: > Anyway. I added it now, but would be nice to understand why it's really needed in this case.

You are right that it isn't needed - I either didn't notice or forgot that `'removeMember'` did nothing asynchronous.

Side effects generally tend to do something asynchronous, since they are modifying state outside of contract, however this apparently isn't always the case, as shown here.

Perhaps it's still good convention though to get used to marking them all as async and using them as such, since this is how side effects are treated in general. To see what I mean by that, have a look in `state.js` to see how these side effects get called. You'll see this:

```js
      if (sbp('sbp/selectors/fn', `${selector}/sideEffect`)) {
        await sbp(`${selector}/sideEffect`, mutation)
      }
```

In other words, all side effects are awaited on. So this is why they must be defined as asynchronous, because it's assumed they are, and they're invoked as such.

sandrina-p: Hm.. okay, it can be for convenience/consistency, although it isn't needed

mmbotelho (changes requested):
There seems to be a problem with the password input, on the "Leave group" modal. The eye button, to make the password visible, disappears when I hover it and makes it impossible to click on it.

sandrina-p:
@mmbotelho that was a global problem with password input and it's now fixed!

mmbotelho (approved):

taoeffect (approved):
Tested and works great! Excellent job @sandrina-p!!!! 🎉

Issue #861, opened 2020-03-12
Payments.vue: hovering the selectcombo the icon disappears


taoeffect:
### Problem

Copying from [here](https://github.com/okTurtles/group-income-simple/pull/847#discussion_r391081221):

> Why did you remove this z-index?
> 
> I also noticed when hovering the selectcombo the icon disappears:
![image](https://user-images.githubusercontent.com/14869087/76437372-626ea100-63b1-11ea-9177-9d1acd2f4cbb.png)

### Solution

Fix.

sandrina-p:
I can fix this at the same time I fix #833 and #831

taoeffect:
Closed in #862

Issue #833, opened 2020-02-13
Fix missing border on combo inputs


mmbotelho:
### Problem

All combination inputs (except the `.input-shifted`) have a missing gray border when they are focused, but not hovered:

<img width="1047" alt="Captura de ecrã 2020-02-13, às 12 31 29" src="https://user-images.githubusercontent.com/22004566/74430901-16861600-4e5e-11ea-9809-c7dd5c41b0f2.png">

### Solution

Keep the gray border when the inputs are focused but not hovered.

taoeffect:
Closed in #862

PR #862, opened 2020-03-13
Improve form fields UI - #831 #833 #861

### LINKED ISSUE #831: Fix hover style on the Selectbox component
### Problem

The current hover state of the `Selectbox` component is looking like this:
Captura de ecrã 2020-02-11, às 14 40 39

### Solution
Make the borders overlap / change color correctly.

sandrina-p:
Close #831, #833 and #861 

**Important note:** The way we were creating a Select Group, was not working correctly on iPhone. A `<label/>` can only have one `<input/>` element associated. If it has two (ex: Add Payment method), it's impossible to focus the second child input on a iPhone/iPad. I documented it on our DS page with a codepen demo to test in on a iPhone. [See link]( https://codepen.io/sandrina-p/pen/oNXyGPE)

### Scouting rule:
- Reviewed all form elements. There were some with incorrect error styling.
- Add to [DS page](http://localhost:8000/app/design-system) the form fields: `switch`, `combobox` and `search`.
- Inputs that need a decimal keyboard (ex: mincome) changed from `type="number"` to `inputmode="decimal"`. See [reference] and [available inputmodes](https://inputmodes.com/).
- Add  `@mixin icon` to use when a icon is used as pseudo element.
- Fix the `menu` example in the DS page.


frontend/views/components/Search.vue
```diff
@@ -4,11 +4,9 @@ form.c-search-form(
 )
   label.field
     .sr-only {{label}}
-
-    .input-combo
-      .is-icon(:aria-label='L("Search")')
+    .inputGroup.c-search
```
pieer: Why this one is camel case

pieer (dismissed):

frontend/views/components/Search.vue
sandrina-p: I was undecided too... Changed to `.inputgroup` as it's done with `.selectbox`.

frontend/assets/style/components/_combobox.scss
```diff
@@ -1,38 +1,26 @@
-.combobox {
+.selectsolo {
```
sandrina-p: Note: A combo field in the web usually means it is a select box where each child has multiple elements, (ex: a list of countries with flags) or it has some built it functionality (ex: a search within the select list).

This element is just a simple select box that can be used solo in the page (ex: nr of items per page). That's why I think `.selectsolo` is a better name than `.combobox`.

frontend/assets/style/components/_combobox.scss
```diff
@@ -1,38 +0,0 @@
-.combobox {
```
sandrina-p, line 1: Note: It wasn't deleted. It was renamed to `_selectsolo.scss`

frontend/assets/style/components/_selectsolo.scss
```diff
@@ -0,0 +1,26 @@
+.selectsolo {
```
sandrina-p, line 1: combobox usually refers to a custom select element with aditional behavior, just as a built-in search within the select options or with custom options (ex: a list of countries with flags).
This element is just a simple selectbox but it can be used anywhere as a solo element. That's why I believe `selectsolo` is a better name than `combobox`

mmbotelho (changes requested):
Here's what I found:

- The password input is not working as expected on the DS (it's working properly on the login though). It's not very important, just a heads up:
Captura de ecrã 2020-03-19, às 16 33 48

Captura de ecrã 2020-03-19, às 16 32 31

sandrina-p:
@mmbotelho,
The DS page is focused in showing the UI and sometimes the functionality is dumb or missing (ex: the input to remove the invitation does not work). But anyway, I made the password toggle work there, just for you :p
Regarding the `SelectGroup`, each scenario needs to manually define the select width for technical reasons. If you go to "the real places" where it's used, it's all fine. But, once again, I fixed it on DS page too and documented this constraint on the code too.

mmbotelho (approved):
Good job, approved!

pieer (approved):
Great work!

Issue #372, opened 2018-04-10
Get rid of dependencies + Adopt/fork rest


taoeffect:
### Problem

We have too many dependencies / third-party libraries. Dependencies are liabilities almost as much as they are assets.

Related issues:

- #361 - "You Might Not Need Lodash"
- #360 - Get rid of NPM
- #349 - Update dependencies
- #340 - Prevent package-lock.json from breaking when deps-of-deps update
- #21 - Bundle dependencies

### Solution

`for (each dependency) {`

- Either get rid of it by writing our own slimmed down implementation of it (e.g. #361)
- Or, adopt it by forking it to our Gitlab, trimming it down to remove all unnecessary functionality, and using our version of it (#360, #340, #21)

`}`

This includes Bulma.

In the end, the only third-party dependencies we should be left with are `devDependencies`, and even those might not be necessary. As much as possible should be removed or adopted, i.e. we don't need a minifier if our codebase is so small that it doesn't require minification.

In the end, we should get rid of `npm` and the `package-lock.json` file as well (since it's so problematic + not helpful). By having our own trimmed-down, audited forks of everything we don't need to worry about new issues arising from updates.

We then don't need to worry about what happens if "npm is down" or "npm is compromised", and we can rely on GPG + git signatures to verify the codebase and commits - Issue #153.

- *Related:* update contributing docs to note all of this and discourage new dependencies from being added.

EDIT: this issue can be closed if #586 and #1101 are closed.

CrazedCoding:
Running:

```
found 450 vulnerabilities (447 low, 3 high)
  run `npm audit fix` to fix them, or `npm audit` for details
phenomenon@phenomenon:~/group-income-simple-work$ npm audit
```

Yields:

![image](https://user-images.githubusercontent.com/18431780/103252832-7c7f6d00-494c-11eb-90c9-72f39d534f7f.png)
![image](https://user-images.githubusercontent.com/18431780/103252772-2e6a6980-494c-11eb-8a0d-532fcdccbca6.png)

Probably should prioritize dropping these packages soon if at all possible.

snowteamer:
When attempting to use our own GitHub fork of a dependency, Travis CI complains about not being able to access it, apparently because of some missing access rights:
![image](https://user-images.githubusercontent.com/64228468/130435133-b1896b3b-324d-49cd-b7d6-1a0178be4080.png)

taoeffect:
[...]

PR #866, opened 2020-03-18
#827 finishing touches to group members list

### LINKED ISSUE #827: Add finishing touches to GroupMembers.vue
### Problem

Screen Shot 2020-02-08 at 3 13 14 PM

The screenshot above shows what the group members list currently looks like with 13 members. Note that some things are missing from the Figma design:

- `displayName` isn't shown for users, only `username`
- The user that's you doesn't have `(you)` next to their name
- New users do not have the `new` tag next to their name (and it's also not clear to me whether they should be at the top or not, that's something @mmbotelho can comment on)

### Solution

Address each of these issues in turn, either as multiple PRs or one big PR.

sandrina-p:
Closes #827 

image

- Members show `displayName` if available.
- Show "(you)" when the member is us.
- Add pill "new" next to members that joined after me. The pill is visible for **1 week**.
- Include "pending members" that were approved to be invited with the pill "pending".
- Update Cypress tests to coverage
- Add "pending members" to modal "all members".
- Note: #869 is not done in this PR. It will be another one.

#### Scouting fixes:
- Improve button loading state when reduced motion is activated.
- Remove unused .pill variants and review its margins where used.


mmbotelho (dismissed):
Perfect @sandrina-p thank you!

frontend/model/state.js
```diff
@@ -334,7 +334,7 @@ const getters = {
     return username => {
       const groupProfile = getters.groupProfile(username)
       const identityState = groupProfile && state[groupProfile.contractID]
-      return identityState && identityState.attributes
+      return identityState ? identityState.attributes : {}
```
taoeffect: Why was this added?

frontend/controller/actions/group.js
```diff
@@ -98,7 +99,7 @@ export default sbp('sbp/selectors/register', {
     try {
       // post acceptance event to the group contract
       const acceptance = await sbp('gi.contracts/group/inviteAccept/create',
-        { inviteSecret },
+        { inviteSecret, joined_ms: Date.now() },
```
taoeffect: Remove (see other comments)

frontend/controller/actions/group.js
```diff
@@ -48,6 +48,7 @@ export default sbp('sbp/selectors/register', {
         invites: {
           [initialInvite.inviteSecret]: initialInvite
         },
+        joined_ms: Date.now(),
```
taoeffect: Remove (see other comments)

frontend/model/contracts/group.js
```diff
@@ -423,7 +437,8 @@ DefineContract({
     },
     'gi.contracts/group/inviteAccept': {
       validate: objectOf({
-        inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
+        inviteSecret: string, // NOTE: simulate the OP_KEY_* stuff for now,
+        joined_ms: number
```
taoeffect: Remove (see other comments)

frontend/model/contracts/group.js
```diff
@@ -199,7 +213,7 @@ DefineContract({
             groupCreator: meta.username
           },
           profiles: {
-            [meta.username]: initGroupProfile(meta.identityContractID)
+            [meta.username]: initGroupProfile(meta.identityContractID, data.joined_ms)
```
taoeffect: use `meta.createdDate`

frontend/model/contracts/group.js
```diff
@@ -436,7 +451,7 @@ DefineContract({
         if (Object.keys(invite.responses).length === invite.quantity) {
           invite.status = INVITE_STATUS.USED
         }
-        Vue.set(state.profiles, meta.username, initGroupProfile(meta.identityContractID))
+        Vue.set(state.profiles, meta.username, initGroupProfile(meta.identityContractID, data.joined_ms))
```
taoeffect: use `meta.createdDate`

frontend/views/components/Avatar.vue
```diff
@@ -1,11 +1,13 @@
 <template lang='pug'>
   img.c-avatar(
+    v-if='imageURL'
```
taoeffect, line 3: Why is needed? Seemed to be working fine before?

frontend/views/components/AvatarUser.vue
```diff
@@ -38,7 +38,7 @@ export default {
       console.debug(`Looking for ${this.username} profile picture`)
       const userContractId = await sbp('namespace/lookup', this.username)
       if (!userContractId) {
-        console.error(`AvatarUser: ${this.username} doesn't exist!`)
+        console.warn(`AvatarUser: ${this.username} doesn't exist!`)
```
taoeffect, line 41: Why the change?

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -54,45 +42,69 @@
...
-      'ourUsername'
+      'ourUsername',
+      'userDisplayName',
+      'currentGroupState',
+      'groupMembersPending'
     ]),
+    dateNow () {
+      console.log('calculate dateNow')
+      return Date.now()
+    },
```
taoeffect: no reason for creating `dateNow`, please remove

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -89,17 +96,31 @@ export default {
...
       'globalProfile',
-      'groupMembersCount'
+      'groupMembersCount',
+      'groupMembersPending',
+      'ourUsername'
     ]),
+    dateNow () {
+      console.log('calculate dateNow')
+      return Date.now()
+    },
```
taoeffect: no reason for creating `dateNow`, please remove

frontend/views/containers/dashboard/GroupMembersTooltipPending.vue
```diff
@@ -0,0 +1,34 @@
+<template lang="pug">
+  tooltip(direction='bottom-end')
+    span.button.is-icon-small(data-test='pendingTooltip')
+      i.icon-question-circle
+    template(slot='tooltip')
+      i18n(
+        tag='p'
+        :args='{ invitedBy: `${invitedBy} ${invitedBy === ourUsername ? `(${L("you")})` : ""}` }'
```
taoeffect: Simply:

```
:args='{ username }'
```

This sentence will not translate into all language properly otherwise.

Alternatively, you can construct two full sentences:

1. `This member did not use their invite link to join the group yet. This link should be given to them by {username}`
2. `This member did not use their invite link to join the group yet. This link should be given to them by {username} (you)`

By using computed properties and the `L` function

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -113,6 +134,15 @@ export default {
...
   },
   methods: {
+    isNewMember (username) {
+      if (username === this.ourUsername) {
+        return false
+      }
+      const weJoined = this.currentGroupState.profiles[this.ourUsername].joined_ms
+      const memberJoined = this.currentGroupState.profiles[username].joined_ms
+      const joinedAfterUs = weJoined < memberJoined
+      return joinedAfterUs && this.dateNow - memberJoined < 604800000 // joined less than 1w (168h) ago.
```
taoeffect: You can use `moment` on the ISOString

frontend/model/contracts/group.js
```diff
@@ -149,6 +150,18 @@ DefineContract({
     groupMembersCount (state, getters) {
       return getters.groupMembersByUsername.length
     },
+    groupMembersPending (state, getters) {
+      const invites = getters.currentGroupState.invites
+      const invitesValid = Object.keys(invites).filter(id => invites[id].status === INVITE_STATUS.VALID && invites[id].creator !== INVITE_INITIAL_CREATOR)
+      const pendingMembers = {}
+      for (const secretId of invitesValid) {
```
taoeffect: This is currently doing 3 loops:

1. `Object.keys(invites)`
2. `.filter(...)`
3. `for (const secretId of ...`

Replace 3 loops with a single loop:

```js
for (const inviteId in invites) {
  let invite = invites[inviteId]
  if (invite.status === INVITE_STATUS.VALID && invite.creator !== INVITE_INITIAL_CREATOR) {
    // stuff below
  }
}
```

frontend/model/contracts/group.js
```diff
@@ -149,6 +150,18 @@ DefineContract({
...
+        pendingMembers[invites[secretId].invitee] = {
+          isPending: true,
```
taoeffect: No reason to add `isPending: true`, remove. Replace with maybe the `inviteId`.

frontend/model/contracts/group.js
```diff
@@ -173,6 +186,7 @@ DefineContract({
     'gi.contracts/group': {
       validate: objectMaybeOf({
         invites: mapOf(string, inviteType),
+        joined_ms: number,
```
taoeffect: remove (rely on existing `meta.createdDate`) (avoid creating new variables if they are not needed)

taoeffect (changes requested):
First draft review done!

frontend/views/components/Avatar.vue
sandrina-p: note the `v-else`. an image must always have a `src`, otherwise it's invalid. so, when it doesn't have a src, it only renders a placeholder (a div)

frontend/views/components/AvatarUser.vue
sandrina-p: Because I think this is not an error... Nothing wrong happened with the app... It's just an edge case. This can happen if we try to look for a user that doesn't exist.
Example1: looking for a member name that was invited but it didn't join the group yet.) Example2: looking for a username that doesn't exist anymore (deleted the account)

frontend/views/containers/dashboard/GroupMembersTooltipPending.vue
sandrina-p: Using the L function? How? I came up with this...
```pug
      i18n(
        ref='msgYou'
        v-if='invitedBy === ourUsername'
        tag='p'
        :args='{ invitedBy }'
      ) This member did not use their invite link to join the group yet. This link should be given to them by {invitedBy} (you).
      i18n(
        v-else
        ref='msg'
        tag='p'
        :args='{ invitedBy }'
      ) This member did not use their invite link to join the group yet. This link should be given to them by {invitedBy}.
```

taoeffect: That uses way too many lines, so in this case a computed property is better. See for example `introTitle` in Payments.vue.

frontend/model/contracts/group.js
sandrina-p: I didn't remember we had `meta.createdDate`, will change!

sandrina-p: I didn't remember we had meta.createdDate, will change!

frontend/views/containers/dashboard/GroupMembersAllModal.vue
sandrina-p: I'm not familiar with `moment`. What method do you think it's the best for this case?

frontend/views/containers/dashboard/GroupMembersTooltipPending.vue
sandrina-p: That was what I thought. doing it

frontend/views/containers/dashboard/GroupMembersAllModal.vue
taoeffect: - https://momentjs.com/docs/#/manipulating/
- https://momentjs.com/docs/#/parsing/date/
- https://momentjs.com/docs/#/parsing/string/

You can `import moment from 'moment'` at the top (it seems to come from one of our dependencies... not sure how)

sandrina-p: Hum... it seems an elephant for what we need here. But found a way more or less intuitive:

```js
moment(memberJoined).isBefore(this.now, 'week')
```

this.now is a computed property returning `moment()`. It should be a computed property to avoid "calculating" now N times in a N loop. Specially on the list with all members (and search).

Btw, I looked into our source and `moment` is an indirect development dependency (it means we don't import it on package.json). The package importing it is Cypress!

So... if we really want to use `moment` it should be part of our dependencies. Just a reminder that this package is [65kb minified](https://bundlephobia.com/result?p=moment@2.24.0). I believe it can be useful to display a more "human time"  on proposals and chat... but it's way heavier than I thought.

taoeffect: I agree that `moment` is probably something we can live without. If you would like to switch back to using the microseconds approach feel free, just initialize the date object using the string (`new Date(timestamp).getTime()`) and then you can use it with `Date.now()`.

I still do not think a computed property is needed. The caching isn't needed in this case I do not think.

sandrina-p: I'm okay with leaving it with `moment`. For now, it's more than okay. I just wanted to show my concern...

> I still do not think a computed property is needed. The caching isn't needed in this case I do not think.

Well, it's the difference between calculating "Date.now()" n times vs 1 time. When n times is the number of members and every time we do a search. I think some cache doesn't hurt.

taoeffect: > Well, it's the difference between calculating "Date.now()" n times vs 1 time

That's not an issue since `Date.now()` doesn't take any significant resources.  Also, we want to make sure that it's updated eventually and isn't cached forever, which is what might happen with `this.now` since that doesn't depend on any Vue.js variables, and so it might return the same value forever.

taoeffect: If you can implement this without `moment` it would save us time in the future too, since we want to get removing unnecessary dependencies the closer we get to releasing this.

sandrina-p: > Also, we want to make sure that it's updated eventually and isn't cached forever

That's not how component's computed properties work. They are "cached" only while the component is mounted. Once it's destroyed (ex: the page changes or the modal is closed), it will re-calculate everything again.

But anyway... changed back to "old school `Date.now()`" and without using a computed property. We have other performance issues more important than this micro-optimization ;)

frontend/model/contracts/group.js
```diff
@@ -62,10 +62,11 @@ function initMonthlyPayments () {
   }
 }
 
-function initGroupProfile (contractID: string) {
+function initGroupProfile (contractID: string, joined: ?number) {
```
taoeffect: `joined: string`

frontend/model/contracts/group.js
```diff
@@ -62,10 +62,11 @@ function initMonthlyPayments () {
...
   return {
     contractID: contractID,
-    nonMonetaryContributions: []
+    nonMonetaryContributions: [],
+    joined_ms: joined
```
taoeffect: `joinedDate`

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -54,45 +42,64 @@
...
+    invite () {
+      this.$router.push({ path: '/invite' })
+    },
+    openModal (modal, queries) {
+      sbp('okTurtles.events/emit', OPEN_MODAL, modal, queries)
+    },
+    isNewMember (username) {
+      if (username === this.ourUsername) { return false }
+
+      const memberJoined = this.currentGroupState.profiles[username].joined_ms
```
taoeffect: This needs to be converted to milliseconds

```js
const joined_ms = new Date(this.currentGroupState.profiles[username].joinedDate).getTime()
```

taoeffect (changes requested):
quick feedback re date usage

frontend/views/containers/dashboard/GroupMembers.vue
sandrina-p: Changed where needed

package.json
```diff
@@ -101,6 +101,7 @@
     "buffer": "5.4.3",
     "form-data": "2.5.1",
     "localforage": "1.7.3",
+    "moment": "2.24.0",
```
taoeffect: `moment` is something we probably won't end up using since it's such a big dependency, so plz remove & reset `package-lock.json`

frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -89,17 +96,30 @@ export default {
...
-      'groupMembersCount'
+      'groupMembersCount',
+      'groupMembersPending',
+      'ourUsername'
     ]),
+    weJoinedMs () {
+      return new Date(this.currentGroupState.profiles[this.ourUsername].joinedDate).getTime()
+    },
+    allMembers () {
+      return Object.keys({ ...this.groupMembersPending, ...this.groupProfiles })
```
taoeffect, line 110: If you just want the keys of these two objects, then there's a faster way of doing this:

```js
return Object.keys(this.groupMembersPending).concat(Object.keys(this.groupProfiles))
```

package.json
sandrina-p: I added `moment` to our dependencies because it's something we are actually using at MonthOverview, PaymentDetails and PaymentsList components. And all direct dependencies must be linked on `package.json`

If we really want to stop using it (removing it), that's an issue for another PR, not this one. :)

frontend/views/containers/dashboard/GroupMembersAllModal.vue
sandrina-p: Hum... that's what I'm doing here. 😅 The difference to your suggestion is that I'm using destructuring (`...`) to create the array instead of using `Object.keys` twice and `.concat`.

And using destructuring is faster: https://jsperf.com/object-keys-concact-vs-destructuring/1

package.json
taoeffect: It was working fine before this PR and it will work fine after you remove it here... please remove. We want to get rid of as many dependencies as humanely possible.

taoeffect: That other issue exists in the form of #372. However I've created a new issue specifically for moment (#874). Please still remove it here tho.

frontend/views/containers/dashboard/GroupMembersAllModal.vue
taoeffect: > And using destructuring is faster: https://jsperf.com/object-keys-concact-vs-destructuring/1

Huh. That's surprising. OK we'll keep it your way 😄  👍

package.json
sandrina-p: > It was working fine before this PR and it will work fine after you remove it here

It works fine because we have Cypress. If tomorrow Cypress doesn't need `moment` anymore, GI installation will break. Btw, there is an existing JS rule on eslint and node community called "no-extraneous-dependencies". See [ref 1](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) and [ref 2](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-missing-import.md)

But anyway... removed.

frontend/model/contracts/group.js
```diff
@@ -423,7 +437,7 @@ DefineContract({
     },
     'gi.contracts/group/inviteAccept': {
       validate: objectOf({
-        inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
+        inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now,
```
taoeffect: random comma?

frontend/model/contracts/group.js
sandrina-p: done

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -13,34 +13,22 @@
...
-          i.icon-question-circle
-        template(slot='tooltip')
-          i18n(
-            tag='p'
-            :args='{ username }'
-          ) We are waiting for {username} to join the group by using their unique invite link.
-
-      group-member-menu(v-else :username='username')
+        | {{ userDisplayName(username) }}&nbsp;
+        i18n(v-if='username === ourUsername') (you)
```
taoeffect: To guard against weirdness in other languages (like the positioning of `(you)`), can use a method instead?

E.g.

```pug
.c-name.has-ellipsis(data-test='username') {{ localizedName(username) }}
```

Then:

```js
methods: {
  localizedName (username) {
    const name = this.userDisplayName(username)
    return username === this.ourUsername ? L('{name} (you)', { name }) : name
  }
}
```

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -54,45 +42,64 @@
...
     ]),
+    weJoinedMs () {
+      return new Date(this.currentGroupState.profiles[this.ourUsername].joinedDate).getTime()
+    },
     firstTenMembers () {
       const profiles = this.groupProfiles
-      const usernames = Object.keys(profiles).slice(0, 10)
-      return usernames.reduce((acc, username) => {
+      const sliceIndex = 10 - Math.min(10, Object.keys(this.groupMembersPending).length) // avoid slicing too many members.
+      const usernames = Object.keys(profiles).slice(0, sliceIndex)
```
taoeffect, line 73: Is this change necessary? AFAICT `.slice` will automatically bounds check for you, you don't need to do it yourself.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -29,7 +29,7 @@ table.table.table-in-card.c-payments(:class='{"is-editing": paymentsType === "ed
 
         // TODO: replace condition to indicate whether or not the payment date is < or > than the current date using payment.paymentStatusText
         i18n.c-user-month(
-          :class='index === 0 ? "has-text-1" : "pill is-danger is-small"'
+          :class='index === 0 ? "has-text-1" : "pill is-danger"'
```
taoeffect, line 32: Are the changes in this file (e.g. removal of `is-small`) because the previous version didn't follow the designs?

frontend/model/contracts/group.js
taoeffect: Remove random comma?

taoeffect (changes requested):
Tested, works great. 💪 

Only minor changes requested.

frontend/views/containers/dashboard/GroupMembers.vue
sandrina-p: No... Let's imagine we have an array `[1,2,3,4,5]` and we do `slice(0, -3)`, the output will be `[1, 2]`. IRL, this can happen if we have more than 10 members pending. Run the code on your computer and see it by yourself.

frontend/views/containers/payments/PaymentsList.vue
sandrina-p: `is-small` isn't needed. It doesn't exist a "small" variant of the `pill` component. They are all the same size.

frontend/views/containers/dashboard/GroupMembers.vue
taoeffect: Oh, I see what you mean. Oops, misread the code. Nice job! 👍

taoeffect (approved):
Great job @sandrina-p!

PR #871, opened 2020-03-21
set 'Poppins' as the font-tamily on button

### LINKED ISSUE #870: Restore correct font on buttons
### Problem

@mmbotelho observed that buttons should have "poppins" font but currently don't.

### Solution

Fix.

SebinSong:
closes #870


mmbotelho (approved):
Perfect, thank you @SebinSong !

PR #875, opened 2020-03-26
#848 Rename identity name to username

### LINKED ISSUE #848: Rename 'name' property in Identity contract to 'username' or delete it
### Problem

It's appeared a few times now in the code that the global profile's username is stored in the `name` property, and it's caused confusion in a few PRs where this was viewed as different from `username`. In fact the two are the same, and `username` should be preferred. It's not clear that we need the one in the global profile, and if we do it should be renamed to `username` to avoid confusion.

### Solution

One way of guaranteeing that they're the same is to simply remove the `name` property entirely, and rely on the namespace mapping (that is what in fact gets us access to the identity contract in the first place) for the username. This might be the best approach, as it means we don't need to write any code to keep the two in sync.

However, if for some reason there is a reason to keep the username in the identity contract as a property, then at the very least it should be renamed from `name` to `username` to make it clear that the two are the same.

sandrina-p:
Closes #848 

Summary of the change to `[identityContract].attributes`:

```diff
{
- name: (...),
+ username: (...),
  email: (...),
  picture: (...),
  mailbox: (...),
  displayName: (...)
}
```

### DIFF
--- a/frontend/controller/actions/user.js
+++ b/frontend/controller/actions/user.js
@@ -23,7 +23,7 @@ export default sbp('sbp/selectors/register', {
     const user = await sbp('gi.contracts/identity/create', {
       // authorizations: [Events.CanModifyAuths.dummyAuth()],
       attributes: {
-        name: username,
+        username,
         email: email,
         picture: `${window.location.origin}/assets/images/default-avatar.png`
       }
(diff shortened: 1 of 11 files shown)


taoeffect (approved):
LGTM!

PR #879, opened 2020-04-02
Rename actions/user to actions/identity

sandrina-p:
Related to #848 

> > Another thing I noticed is how we say "identity contract" and "user contract" and they mean the same thing. We actually have 2 files: contracts/identity.js(gi.contracts/identity/*) and actions/user.js (gi.actions/user/*)
> > Which one should we stick with? Perhaps identity because most of the code uses it.
> 
> Yeah agreed... feel free to do that in another PR


taoeffect (approved):
👍

PR #880, opened 2020-04-03
Remove moment usage from main codebase

### LINKED ISSUE #874: Remove moment usage everywhere
### Problem

`moment` dependency is used in a few places in the project, and while it's nice, it's also some ~65k+ (or more?). We want to remove it as well as a bunch of other dependencies (#372) to minimize attack surface and speed up the app.

### Solution

Remove it and replace with `Date`-based usage, e.g.:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/format

sandrina-p:
Closes #874 

Now dates are displayed based browser language (`navigator`) using [`Date .toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString).

A nice thing I like about this is that it handles date translations for us automatically! For example, in "en-US" it says "April 3" but in PT-PT it says "3 de Abril" (as expected).

### DIFF
--- a/frontend/views/containers/payments/PaymentDetail.vue
+++ b/frontend/views/containers/payments/PaymentDetail.vue
@@ -10,12 +10,10 @@ modal-template(ref='modal' v-if='payment')
   ul.c-payment-list
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Date & Time
-      // TODO remove new Date() when dealing with real data.
-      strong {{ humanDate(this.payment.date || new Date(), { year: 'numeric', month: 'long', day: 'numeric', hour: '2-digit', minute: '2-digit' }) }}
+      strong {{ humanDate(this.payment.date, { year: 'numeric', month: 'long', day: 'numeric', hour: '2-digit', minute: '2-digit' }) }}
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Relative to
-      // TODO remove new Date() when dealing with real data.
-      strong {{ humanDate(payment.relativeTo || new Date(), { month: 'long' }) }}
+      strong {{ humanDate(payment.relativeTo, { month: 'long' }) }}
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Mincome at the time
       strong {{ currency(groupSettings.mincomeAmount) }}
--- a/frontend/views/containers/payments/PaymentDetail.vue
+++ b/frontend/views/containers/payments/PaymentDetail.vue
@@ -10,10 +10,12 @@ modal-template(ref='modal' v-if='payment')
   ul.c-payment-list
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Date & Time
-      strong {{ moment(payment.date).format('hh:mm - MMMM DD, YYYY') }}
+      // TODO remove new Date() when dealing with real data.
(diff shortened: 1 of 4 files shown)


frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -157,6 +155,10 @@ export default {
       this.payments.map(payment => {
         payment.checked = this.tableChecked
       })
+    },
+    dueDate (datems) {
+      const date = datems || new Date() // remote new Date() when dealing with real data.
```
taoeffect: Can you enforce consistency here and everywhere else in the type usage?

I.e. either `humanDate` accepts milliseconds or a `Date` object, but not both.

So in this case, `new Date()` would be replaced with `Date.now()`, or even better, simply add a default value to the `humanDate` function (e.g. `humanDate(datems = Date.now()`, I think that might be possible, double-check me on that)

frontend/views/utils/humanDate.js
```diff
@@ -0,0 +1,5 @@
+export function humanDate (datems, opts = {}) {
+  const date = new Date(datems)
+  const locale = navigator.languages ? navigator.languages[0] : navigator.language
+  return date.toLocaleDateString(locale, opts)
```
taoeffect: We don't need to fetch the locale each time this function is called, that can be done once at the top of this file.

Then this function can be a one-liner:

```js
return new Date(datems).toLocaleDateString(locale, opts)
```

frontend/views/containers/payments/PaymentDetail.vue
```diff
@@ -10,10 +10,12 @@ modal-template(ref='modal' v-if='payment')
...
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Date & Time
-      strong {{ moment(payment.date).format('hh:mm - MMMM DD, YYYY') }}
+      // TODO remove new Date() when dealing with real data.
+      strong {{ humanDate(this.payment.date || new Date(), { year: 'numeric', month: 'long', day: 'numeric', hour: '2-digit', minute: '2-digit' }) }}
     li.c-payment-list-item
       i18n.has-text-1(tag='label') Relative to
-      strong {{ moment(payment.relativeTo).format('MMMM')}}
+      // TODO remove new Date() when dealing with real data.
+      strong {{ humanDate(payment.relativeTo || new Date(), { month: 'long' }) }}
```
taoeffect: See comment below re `Date.now()` usage

taoeffect (changes requested):
Great work! 👍 See comments!

frontend/views/containers/payments/PaymentsList.vue
sandrina-p: The default value is a clean solution. Changed, as well as the other comments!

taoeffect (approved):
Fantastic! 👍

Issue #263, opened 2017-06-01
Investigate Electrino and Weex and Flutter and NodeGUI etc as Electron alternatives for Desktop


taoeffect:
### Problem

["Electron is Flash for the desktop"](https://josephg.com/blog/electron-is-flash-for-the-desktop/), although see this [sort-of rebuttal](https://github.com/workflowproducts/postage/blob/master/Facts_About_Electron_Performance.md).

### Solution

167 kB app vs 115 MB by using OS's native browser.

https://github.com/pojala/electrino (note, see this issue on merging Electrino with MacGap: `https://github.com/pojala/electrino/issues/10`)

Downside is fewer desktop integration features, and the fact that it uses Microsoft Edge on windows... But that might not turn out to be a big deal?

Another possibility is [Weex](https://blog.learnstartup.net/p/native-mobile-apps-weex-vuejs-2-0/), which is like React Native, except for VueJS.

Still yet another possibility is [volt](https://volt.ws), which account to its site will be open sourced in 2021.

EDIT (2026-06): [Deno Desktop](https://docs.deno.com/runtime/desktop/) ! ([hn](https://news.ycombinator.com/item?id=48626137)). Problem: doesn't support mobile.

EDIT: see also: https://jlongster.com/secret-of-good-electron-apps

EDIT: also see [NodeGUI](https://hibbard.eu/node-gui/) -> https://github.com/nodegui/nodegui -> https://github.com/nodegui/vue-nodegui

EDIT: [Neutralino!](https://neutralino.js.org/)

EDIT: [Muon](https://github.com/ImVexed/muon)

EDIT: [Tauri](https://github.com/tauri-apps/tauri) !! => https://tauri.studio/en/ - See: [Tauri vs Electron](https://www.levminer.com/blog/tauri-vs-electron), [example app](https://github.com/cardo-podcast/cardo)

- Thread on native push notifications: https://github.com/tauri-apps/tauri/issues/11651

EDIT: [Capacitor](https://capacitorjs.com/) by Ionic (used to be [Cordova](https://capacitorjs.com/cordova) I think?)

EDIT: [Sciter](https://sciter.com/) ! => https://sciter.com/

EDIT: **[Socket SDK](https://sockets.sh/) !!** - looks really promising!! Desktop & mobile & P2P!

EDIT: [Flutter](https://medium.com/flutter-community/flutter-on-desktop-a-real-competitor-to-electron-4f049ea6b061)

[...]

PR #881, opened 2020-04-03
#788 Standardized scoped message

### LINKED ISSUE #788: Standardized error messages in scoped banners

sandrina-p:
Closes #788

demo

Based on #788, the original idea to normalize the error messages is to create a `L()` function with a link `<router-link>` pointing to the Application Logs page. That approach has one major problem: `L` does not support `compile` and even if it does, it can't be a `router-link`. It needs to be `sbp` to open `ApplicationLogs` modal correctly with the needed props.

Since our design is consistent and all error messages appear inside the banner scoped, I thought about using it as our "bridge" to the Application logs. I made a POC in two places: Change Mincome modal and User Settings form. **Please go give it a try on your computer**

This is the only change needed to be done everywhere to standardize the error messages:

```js
// Note as danger() accepts 2 arguments: A `L()` function with a normal string
// and as the 2nd parameter, we pass the `error`. 
this.$refs.formMsg.danger(L('Failed to change mincome.'), { error })
```

The tricky part (already done in this PR) was to update `BannerScoped` so it redirects correctly to the Application Logs page through `OPEN_MODAL` sbp call. There both custom props and url query are passed down to the modal component. This way the modal is open correctly at ApplicationLogs section with the error message.

What do you guys think? @taoeffect and @pieer

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -16,7 +16,7 @@ import { GIErrorUIRuntimeError } from '@model/errors.js'
 
 import imageUpload from '@utils/imageUpload.js'
 import { merge } from '@utils/giLodash.js'
-import L from '@view-utils/translations.js'
+import L, { LError } from '@view-utils/translations.js'
 
 export default sbp('sbp/selectors/register', {
   'gi.actions/group/create': async function ({
@@ -122,10 +122,12 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/updateSettings': async function (settings, groupId) {
     try {
+      this.dumbError.forced = 123 // DELETE THIS BEFORE MERGE
       const message = await sbp('gi.contracts/group/updateSettings/create', settings, groupId)
       await sbp('backend/publishLogEntry', message)
     } catch (e) {
-      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {codeError}', { codeError: e.message }))
+      console.error('gi.actions/group/updateSettings failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {reportError}', LError(e)))
     }
   },
   'gi.actions/group/removeMember': async function (params, groupID) {
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -122,7 +122,6 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/updateSettings': async function (settings, groupId) {
     try {
-      this.dumbError.forced = 123 // DELETE THIS BEFORE MERGE
       const message = await sbp('gi.contracts/group/updateSettings/create', settings, groupId)
       await sbp('backend/publishLogEntry', message)
     } catch (e) {
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -1,8 +1,6 @@
 import sbp from '~/shared/sbp.js'
 import { GIErrorUIRuntimeError } from '@model/errors.js'
-import L from '@view-utils/translations.js'
-
-// import { CONTRACT_IS_SYNCING } from '@utils/events.js'
+import L, { LError } from '@view-utils/translations.js'
 
 export default sbp('sbp/selectors/register', {
   'gi.actions/identity/create': async function ({
@@ -65,7 +63,7 @@ export default sbp('sbp/selectors/register', {
     } catch (e) {
       await sbp('gi.actions/identity/logout') // TODO: should this be here?
       console.error('gi.actions/identity/signup failed!', e)
-      throw new GIErrorUIRuntimeError(L('Failed to signup: {codeError}', { codeError: e.message }))
+      throw new GIErrorUIRuntimeError(L('Failed to signup: {reportError}', LError(e)))
     }
   },
   'gi.actions/identity/login': async function ({
@@ -92,7 +90,7 @@ export default sbp('sbp/selectors/register', {
       return userId
     } catch (e) {
       console.error('gi.actions/identity/login failed!', e)
-      throw new GIErrorUIRuntimeError(L('Failed to login: {codeError}', { codeError: e.message }))
+      throw new GIErrorUIRuntimeError(L('Failed to login: {reportError}', LError(e)))
     }
   },
   'gi.actions/identity/signupAndLogin': async function (signupParams) {
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -298,7 +298,7 @@ DefineContract({
           }
 
           if (deepEqualJSONType(omit(prop.data.proposalData, 'reason'), dataToCompare)) {
-            throw new TypeError(L('There is an identical open proposal'))
+            throw new Errors.GIErrorUIRuntimeError(L('There is an identical open proposal.'))
           }
         }
       },
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -298,7 +298,7 @@ DefineContract({
           }
 
           if (deepEqualJSONType(omit(prop.data.proposalData, 'reason'), dataToCompare)) {
-            throw new TypeError(L('There is an identical open proposal.'))
+            throw new TypeError(L('There is an identical open proposal'))
           }
         }
       },
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -298,7 +298,7 @@ DefineContract({
           }
 
           if (deepEqualJSONType(omit(prop.data.proposalData, 'reason'), dataToCompare)) {
-            throw new Errors.GIErrorUIRuntimeError(L('There is an identical open proposal.'))
+            throw new TypeError(L('There is an identical open proposal.'))
           }
         }
       },
--- a/frontend/views/components/banners/BannerScoped.vue
+++ b/frontend/views/components/banners/BannerScoped.vue
@@ -3,7 +3,10 @@
     .c-container(v-if='ephemeral.text')
       banner-simple(class='c-banner' :severity='ephemeral.severity')
         .c-inner
-          .c-inner-text(:data-test='dataTest' role='alert') {{ ephemeral.text }}
+          .c-inner-text(:data-test='dataTest' role='alert' @click='handleMsgClick')
+            | {{ ephemeral.text }}
+            |
+            span(v-if='ephemeral.error' v-html='linkToErrorReport')
           button.is-icon-small.c-button(
             type='button'
             :class='`is-${ephemeral.severity}`'
@@ -14,6 +17,9 @@
 </template>
 
 <script>
+import sbp from '~/shared/sbp.js'
+import { OPEN_MODAL } from '@utils/events.js'
+import L from '@view-utils/translations.js'
 import BannerSimple from '@components/banners/BannerSimple.vue'
 import TransitionExpand from '@components/TransitionExpand.vue'
 
@@ -32,17 +38,43 @@ export default {
   data: () => ({
     ephemeral: {
       text: null,
-      severity: null
+      severity: null,
+      error: null
     }
   }),
+  computed: {
+    linkToErrorReport () {
+      return L('Please check the {b1}error logs{b2} for more info.', {
+        'b1': '<button class="link js-btnError">',
+        'b2': '</button>'
+      })
+    }
+  },
   methods: {
+    handleMsgClick (e) {
+      if (this.ephemeral.error && e.target.classList.contains('js-btnError')) {
+        sbp('okTurtles.events/emit', OPEN_MODAL, 'UserSettingsModal',
+          { // custom props
+            tabProps: {
+              'application-logs': {
+                referalError: this.ephemeral.error
+              }
+            }
+          },
+          { // custom url query
+            section: 'application-logs'
+          }
+        )
+      }
+    },
     // To be used by parent. Example:
     // this.$refs.BannerScoped.success(L('Changes saved!'))
     clean () {
       this.updateBanner('', '')
     },
-    danger (text) {
+    danger (text, opts) {
       this.updateBanner(text, 'danger')
+      if (opts.error) { this.ephemeral.error = opts.error }
     },
     success (text) {
       this.updateBanner(text, 'success')
@@ -50,6 +82,7 @@ export default {
     updateBanner (text, severity) {
       this.ephemeral.text = text
       this.ephemeral.severity = severity
+      this.ephemeral.error = null
     }
   }
 }
--- a/frontend/views/components/banners/BannerScoped.vue
+++ b/frontend/views/components/banners/BannerScoped.vue
@@ -72,7 +72,7 @@ export default {
     clean () {
       this.updateBanner('', '')
     },
-    danger (text, opts) {
+    danger (text, opts = {}) {
       this.updateBanner(text, 'danger')
       if (opts.error) { this.ephemeral.error = opts.error }
     },
--- a/frontend/views/components/banners/BannerScoped.vue
+++ b/frontend/views/components/banners/BannerScoped.vue
@@ -3,7 +3,7 @@
     .c-container(v-if='ephemeral.text')
       banner-simple(class='c-banner' :severity='ephemeral.severity')
         .c-inner
-          .c-inner-text(:data-test='dataTest' role='alert' v-html="ephemeral.text")
+          .c-inner-text(:data-test='dataTest' role='alert' v-html='ephemeral.text')
           button.is-icon-small.c-button(
             type='button'
             :class='`is-${ephemeral.severity}`'
@@ -84,9 +84,7 @@ export default {
   }
 }
 
-$severities:
-  "success" $success_0 $success_1,
-  "danger" $danger_0 $danger_1;
+$severities: "success" $success_0 $success_1, "danger" $danger_0 $danger_1;
 
 .c-button {
   transition: box-shadow 150ms ease-in;
--- a/frontend/views/components/banners/BannerScoped.vue
+++ b/frontend/views/components/banners/BannerScoped.vue
@@ -3,10 +3,7 @@
     .c-container(v-if='ephemeral.text')
       banner-simple(class='c-banner' :severity='ephemeral.severity')
         .c-inner
-          .c-inner-text(:data-test='dataTest' role='alert' @click='handleMsgClick')
-            | {{ ephemeral.text }}
-            |
-            span(v-if='ephemeral.error' v-html='linkToErrorReport')
+          .c-inner-text(:data-test='dataTest' role='alert' v-html="ephemeral.text")
           button.is-icon-small.c-button(
             type='button'
             :class='`is-${ephemeral.severity}`'
@@ -17,9 +14,6 @@
 </template>
 
 <script>
-import sbp from '~/shared/sbp.js'
-import { OPEN_MODAL } from '@utils/events.js'
-import L from '@view-utils/translations.js'
 import BannerSimple from '@components/banners/BannerSimple.vue'
 import TransitionExpand from '@components/TransitionExpand.vue'
 
@@ -38,43 +32,17 @@ export default {
   data: () => ({
     ephemeral: {
       text: null,
-      severity: null,
-      error: null
+      severity: null
     }
   }),
-  computed: {
-    linkToErrorReport () {
-      return L('Please check the {b1}error logs{b2} for more info.', {
-        'b1': '<button class="link js-btnError">',
-        'b2': '</button>'
-      })
-    }
-  },
   methods: {
-    handleMsgClick (e) {
-      if (this.ephemeral.error && e.target.classList.contains('js-btnError')) {
-        sbp('okTurtles.events/emit', OPEN_MODAL, 'UserSettingsModal',
-          { // custom props
-            tabProps: {
-              'application-logs': {
-                referalError: this.ephemeral.error
-              }
-            }
-          },
-          { // custom url query
-            section: 'application-logs'
-          }
-        )
-      }
-    },
     // To be used by parent. Example:
     // this.$refs.BannerScoped.success(L('Changes saved!'))
     clean () {
       this.updateBanner('', '')
     },
-    danger (text, opts = {}) {
+    danger (text) {
       this.updateBanner(text, 'danger')
-      if (opts.error) { this.ephemeral.error = opts.error }
     },
     success (text) {
       this.updateBanner(text, 'success')
@@ -82,7 +50,6 @@ export default {
     updateBanner (text, severity) {
       this.ephemeral.text = text
       this.ephemeral.severity = severity
-      this.ephemeral.error = null
     }
   }
 }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -118,7 +118,7 @@ export default {
 
           this.ephemeral.currentStep += 1 // Show Success step
         } catch (e) {
-          this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
+          this.$refs.formMsg.danger(L('Failed to propose mincome change: {reportError}', LError(e)))
           this.ephemeral.currentStep = 0
         }
         return
@@ -134,7 +134,7 @@ export default {
         this.$refs.proposal.close()
       } catch (e) {
         console.error('Mincome.vue submit() error:', e)
-        this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
+        this.$refs.formMsg.danger(L('Failed to change mincome: {reportError}', LError(e)))
       }
     }
   }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -118,15 +118,14 @@ export default {
 
           this.ephemeral.currentStep += 1 // Show Success step
         } catch (e) {
-          console.error(`Failed to change mincome to ${mincomeAmount}`, e.message)
-          this.$refs.formMsg.danger(L('Failed to change mincome. {codeError}', { codeError: e.message }))
+          console.error('Mincome submit() error:', mincomeAmount, e.message)
+          this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
           this.ephemeral.currentStep = 0
         }
         return
       }
 
       try {
-        this.foo.dumbError = 'force-an-error' // DELETE THIS BEFORE MERGE
         const updatedSettings = await sbp(
           'gi.contracts/group/updateSettings/create',
           { mincomeAmount },
@@ -134,9 +133,9 @@ export default {
         )
         await sbp('backend/publishLogEntry', updatedSettings)
         this.$refs.proposal.close()
-      } catch (error) {
-        console.error('Mincome.vue submit() error:', error)
-        this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(error)))
+      } catch (e) {
+        console.error('Mincome.vue submit() error:', e)
+        this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
       }
     }
   }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -135,12 +135,8 @@ export default {
         await sbp('backend/publishLogEntry', updatedSettings)
         this.$refs.proposal.close()
       } catch (error) {
-        console.error('Mincome.vue submit() error:', error.message)
-        // @taoeffect, please try to make this work!
-        this.$refs.formMsg.danger(`${L('Failed to change mincome.')} ${LError(error)}`)
-
-        // The original implemented solution:
-        // this.$refs.formMsg.danger(L('Failed to change mincome.'), { error })
+        console.error('Mincome.vue submit() error:', error)
+        this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(error)))
       }
     }
   }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -30,7 +30,7 @@ import { validationMixin } from 'vuelidate'
 import { required } from 'vuelidate/lib/validators'
 import { mapGetters, mapState } from 'vuex'
 import { decimals } from '@view-utils/validators.js'
-import L from '@view-utils/translations.js'
+import L, { LError } from '@view-utils/translations.js'
 import ProposalTemplate from './ProposalTemplate.vue'
 import BannerScoped from '@components/banners/BannerScoped.vue'
 import { PROPOSAL_GROUP_SETTING_CHANGE } from '@model/contracts/voting/constants.js'
@@ -126,7 +126,7 @@ export default {
       }
 
       try {
-        this.foo.mincome = 'force-an-error' // DELETE THIS BEFORE MERGE
+        this.foo.dumbError = 'force-an-error' // DELETE THIS BEFORE MERGE
         const updatedSettings = await sbp(
           'gi.contracts/group/updateSettings/create',
           { mincomeAmount },
@@ -136,7 +136,11 @@ export default {
         this.$refs.proposal.close()
       } catch (error) {
         console.error('Mincome.vue submit() error:', error.message)
-        this.$refs.formMsg.danger(L('Failed to change mincome.'), { error })
+        // @taoeffect, please try to make this work!
+        this.$refs.formMsg.danger(`${L('Failed to change mincome.')} ${LError(error)}`)
+
+        // The original implemented solution:
+        // this.$refs.formMsg.danger(L('Failed to change mincome.'), { error })
       }
     }
   }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -118,8 +118,6 @@ export default {
 
           this.ephemeral.currentStep += 1 // Show Success step
         } catch (e) {
-          this.dumbError.dumbError = 'forcer-error' // DELETE BEFORE MERGE!!
-          console.error('Mincome submit() error:', mincomeAmount, e.message)
           this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
           this.ephemeral.currentStep = 0
         }
--- a/frontend/views/containers/proposals/Mincome.vue
+++ b/frontend/views/containers/proposals/Mincome.vue
@@ -126,6 +126,7 @@ export default {
       }
 
       try {
+        this.foo.mincome = 'force-an-error' // DELETE THIS BEFORE MERGE
         const updatedSettings = await sbp(
           'gi.contracts/group/updateSettings/create',
           { mincomeAmount },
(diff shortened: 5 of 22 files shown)


taoeffect:
> That approach has one major problem: `L` does not support `compile` and even if it does, it can't be a `router-link`. It needs to be `sbp` to open `ApplicationLogs` modal correctly with the needed props.

Hmm... OK, the idea of putting `compile` into `L` is indeed an issue, but something that I think can be solved potentially. I can suggest some steps later below. Per #788, everything would be done inside of `LError`, not `L` (so `compile` attribute wouldn't be needed).

I believe it can be a `router-link` as well. Have a look at the [#to](https://router.vuejs.org/api/#to) API for `router-link`. It supports passing both parameters to a component via `params`, or via `query` (preferred, I think, since it should work with Browser's back/forward ).

To get that working you simply need to add a router for it, see again the Vue Router documentation, for example [this](https://router.vuejs.org/guide/advanced/meta.html) or something else (there docs describe many options).

I see a couple issues with the approach in this PR:

1. It restricts our ability to create these links to the application logs to the BannerScoped component, and that limits our ability to link to it via other means. It would give us more flexibility if we could link to it anywhere.
2. It uses a hack (`.c-inner-text(:data-test='dataTest' role='alert' @click='handleMsgClick')`), not a real link, to navigate the user. This is annoying, because it means that these links will require special styling and will behave differently than regular links in the app.

Can you see if you can implement the `LError` function as described in #788 to see if it's possible?

It's been a while but I think you can extract this code into a function and use it in both locations:

```js
      var result = Vue.compile('<wrap>' + translation + '</wrap>')
      // console.log('TRANSLATED RENDERED TEXT:', context, result.render.toString())
      return result.render.call({
        _c: (tag, ...args) => {
          if (tag === 'wrap') {
            return h(context.props.tag, context.data, ...args)
          } else {
            return h(tag, ...args)
          }
        },
        _v: x => x
      })
```

Check to see what is returned by `result.render.call`. If it's simply a string consisting of regular HTML, then you're all good, very easy to integrate that directly into `LError`. If it's not a string, but some type of object, then see if there's a way to convert that object into HTML (there must be a way, since that is what Vue.js eventually does with everything returned from the `render` function).

sandrina-p:
@taoeffect, I tried what you suggested with `LError` but couldn't make it work. (see code.)

As far I understand `Vue.compile` can only be used inside render fn of a Vue component (as `i118n`)m but `LError()` is just a normal function. Could you please try it out on your computer and tell me in detail what code to use.

Once we make `LError` work correctly, everything else (router-link) is trivial.

Maybe @pieer knows how to do it...

--- EDIT ---

Assuming that it's impossible to make`LError()` returning a proper component, I think the best solution is to simply return the content as string and add `v-html` to the element that will render LError.

--- EDIT 2 ---

Actually using `v-html` wouldn't work neither because it [does not support Vue components inside](https://forum.vuejs.org/t/vue-components-elements-in-v-html/19000). That's why we use a custom link in the first place with `.js-btnError`.

> It uses a hack (.c-inner-text(:data-test='dataTest' role='alert' @click='handleMsgClick')), not a real link, to navigate the user. This is annoying, because it means that these links will require special styling and will behave differently than regular links in the app.

This is not a "real link" but neither it is any of other links that open a modal. However, all of them work properly by changing the routing through Modal.vue. Regarding styling, I didn't understand what you mean.

taoeffect (changes requested):
OK, I think I got it! It took me a while, I tried all kinds of tricks based on [this](https://css-tricks.com/creating-vue-js-component-instances-programmatically/) and other things, but realized that there was a fundamental problem: `LError` is defined outside of any sort of Vuejs context, and in order to render components via `$createElement` it needs that context. And since it doesn't have that context, it has no idea how to handle `<router-link>`.

So, the solution is to use regular HTML, which doesn't require compilation.

Using this method we are able to reproduce the behavior described in #788.

For file `Mincome.vue`, we have:

```js
this.$refs.formMsg.danger(L('Failed to change mincome: {error}. {troubleshootLink}.', LError(error)))
```

For file `translation.js`, we have:

```js
export function LError (error) {
  const translation = L('{link_}Click to troubleshoot.{_link}', {
    'link_': `<a class="link" href="/group-settings?error=${encodeURI(error.message)}">`,
    '_link': '</a>'
  })
  return {
    error: error.message,
    troubleshootLink: translation
  }
}
```

And for `BannerScoped.vue`, we have:

```pug
          .c-inner-text(:data-test='dataTest' role='alert' v-html='ephemeral.text')
```

I think that should work, and requires almost no changes to `translation.js` or `BannerScoped.vue`, in addition to working everywhere.

The only thing is that I think it will result in a "page refresh" instead of the normal routing behavior that VueRouter does, but I don't think that's a big deal? And if you don't like that, you might be able to fix it anyway by using tricks like `<a href="#" onclick=....` or `<a href="javascript:...`.

sandrina-p:
@taoeffect, that solution will cause a refresh, yes. I think thats a poor solution compared to our existing solutions. That’s why I used `js-btnError` to handle the link behavior correctly through sbp.
Tricks like `href='#'` are anti patterns. An href must always go somewhere by default, otherwise it should be a button. 

I’ll research a better solution.

taoeffect:
> Tricks like href='#' are anti patterns. An href must always go somewhere by default, otherwise it should be a button.

If you consider it an anti-pattern, have a look at using `href="javascript:"` in the link to run the router code. We need a simple and yet versatile solution to this.

sandrina-p:
@taoeffect, please check the code again. Based on the constraints and requirements, I believe the simplest solution is to create a raw anchor with the link to application logs page and `target="_blank"` to prevent the user from losing its context (previously caused by the page refresh). In summary:

```js
// translations.js
export function LError (error) {
  return {
    reportError: L('Please {a_}report to us{_a} the error.', {
      'a_': `<a class="link" target="_blank" href="/app/dashboard?modal=UserSettingsModal&section=application-logs&errorMsg=${error.message}">`,
      '_a': '</a>'
    })
  }
}
```

Then, wherever this error is used, it must be added to the DOM through `v-html`.

```js
// a sample error through BannerScoped. It has v-html by default.
this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(error)))
```

I added "forced errors" in 3 places: Change mincome, update profile settings and update group settings.

-----

**Observation 1:** We currently add a `console.error` at each `catch()` (on Vue components and at `controllers/` files) This cause us to end up with duplicated logs. Example: Update Group settings:

```js
// actions/group.js
    } catch (e) {
      console.error('gi.actions/group/updateSettings failed!', e)
      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {reportError}', LError(e)))
    }

// GroupSettings.vue
      } catch (e) {
        console.error('GroupSettings saveSettings failed!', e)
        this.$refs.formMsg.danger(e.message)
      }

```

Application logs:
```
2020-04-08T10:16:59.943Z [error] "gi.actions/group/updateSettings failed!" "TypeError: Cannot set property 'forced' of undefined\n"
2020-04-08T10:16:59.947Z [error] "GroupSettings saveSettings failed!" "GIErrorUIRuntimeError: Failed to update group settings. Please <a class=\"link\" target=\"_blank\" href=\"/app/dashboard?modal=UserSettingsModal&section=application-logs&errorMsg=Cannot set property 'forced' of undefined\">report to us</a> the error.\n 
```

Maybe we shouldn't add the `console.error` at `GroupSettings.vue`.

frontend/views/utils/translations.js
```diff
@@ -65,6 +65,15 @@ export default function L (
   return template(key, args)
 }
 
+export function LError (error) {
+  return {
+    reportError: L('Please {a_}report to us{_a} the error.', {
+      'a_': `<a class="link" target="_blank" href="/app/dashboard?modal=UserSettingsModal&section=application-logs&errorMsg=${error.message}">`,
```
taoeffect: 1. You must `encodeURI` the `error.message`, e.g. `encodeURI(error.message)`
2. Since we are not sure yet how this will behave when we create an app bundle out of it, please remove the `target="_blank"` part for now. You can add a comment and/or open an issue to add it back later once we are able to test it with Electron or whatever app-creator we use.

frontend/views/components/banners/BannerScoped.vue
```diff
@@ -3,7 +3,7 @@
     .c-container(v-if='ephemeral.text')
       banner-simple(class='c-banner' :severity='ephemeral.severity')
         .c-inner
-          .c-inner-text(:data-test='dataTest' role='alert') {{ ephemeral.text }}
+          .c-inner-text(:data-test='dataTest' role='alert' v-html="ephemeral.text")
```
taoeffect: This causes `grunt dev` to fail because of a linter error. Must single quotes

frontend/views/utils/translations.js
```diff
@@ -65,6 +65,15 @@ export default function L (
   return template(key, args)
 }
 
+export function LError (error) {
+  return {
+    reportError: L('Please {a_}report to us{_a} the error.', {
```
taoeffect: Please:

1. Add back `error` with `error.message`
2. Make this read `{a_}See error logs.{_a}`

frontend/views/containers/user-settings/UserProfile.vue
```diff
@@ -147,15 +147,16 @@ export default {
...
           this.$store.state.loggedIn.identityContractID
         )
         await sbp('backend/publishLogEntry', attributes)
         this.$refs.formMsg.success(L('Your changes were saved!'))
-      } catch (e) {
-        console.error('Failed to update profile', e)
-        this.$refs.formMsg.danger(L('Failed to update profile, please try again. {codeError}', { codeError: e.message }))
+      } catch (error) {
+        console.error('UserProfile.vue saveProfile() error:', error)
+        this.$refs.formMsg.danger(L('Failed to update profile. {reportError}', LError(error)))
```
taoeffect: Should read:

```js
this.$refs.formMsg.danger(L('Error updating profile: {error}. {reportError}', LError(error)))
```

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -126,16 +126,17 @@ export default {
...
           this.currentGroupId
         )
         await sbp('backend/publishLogEntry', updatedSettings)
         this.$refs.proposal.close()
-      } catch (e) {
-        console.error(`Failed to change mincome to ${mincomeAmount}`, e.message)
-        this.$refs.formMsg.danger(L('Failed to change mincome {codeError}', { codeError: e.message }))
+      } catch (error) {
+        console.error('Mincome.vue submit() error:', error)
+        this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(error)))
```
taoeffect: Should read:

```js
this.$refs.formMsg.danger(L('Error changing mincome: {error}. {reportError}', LError(error)))
```

frontend/controller/actions/group.js
```diff
@@ -122,10 +122,12 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/updateSettings': async function (settings, groupId) {
     try {
+      this.dumbError.forced = 123 // DELETE THIS BEFORE MERGE
       const message = await sbp('gi.contracts/group/updateSettings/create', settings, groupId)
       await sbp('backend/publishLogEntry', message)
     } catch (e) {
-      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {codeError}', { codeError: e.message }))
+      console.error('gi.actions/group/updateSettings failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {reportError}', LError(e)))
```
taoeffect, line 129: Should read:

```js
this.$refs.formMsg.danger(L('Error updating group settings: {error}. {reportError}', LError(e)))
```

Similar changes throughout rest of file.

frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -2,6 +2,8 @@
   .settings-container
     section.card
       .c-header
+        p.c-instructions(v-if='ephemeral.reportError')
+          | [Give some instructions to the user about this page. Ask @mmbotelho for help]
```
taoeffect: What you had before here was really good. Please add it back, or some version of it, showing the error message that was passed in via the query parameters.

E.g. `Search the logs below for "${errorMsg}" for details. Please send us the logs so that we can help you debug any problems.`

taoeffect (changes requested):
Seems to work! Some changes needed though to bring it in line with the requirements of #788, in order to make sure we present users with useful information so that they can potentially help themselves fix any problems.

frontend/controller/actions/group.js
sandrina-p: What if error, comes from {reportError}? That way we don’t need to repeat the 2 variables everywhere.

frontend/views/utils/translations.js
sandrina-p: 1. Will do
2. Will we create an app for the prototype? I thought this would be browser based. I can leave a comment to review it later but for now we need the target blank

frontend/controller/actions/group.js
taoeffect: @sandrina-p that's a good idea, ok, in that case this would read:

```js
this.$refs.formMsg.danger(L('Error updating group settings. {reportError}', LError(e)))
```

And then in `LError` you'd have:

```js
{errorMsg}. {a_}See error logs.{_a}
```

frontend/views/utils/translations.js
taoeffect: @sandrina-p OK if you really want to keep `target="_blank"` then please open an issue referencing this problem and issue #263.

frontend/views/containers/user-settings/AppLogs.vue
sandrina-p: Let’s wait for @mmbotelho feedback on this one :)

frontend/controller/actions/group.js
sandrina-p: Cool! Regarding the error itself. I find the wording ‘Error updating group settings’ kinda 🤖. Don’t you think ‘Failed to update group settings’ more human 👀?

taoeffect: sure 👍

frontend/views/containers/user-settings/AppLogs.vue
mmbotelho: Can you give me a bit more context on what this is? @sandrina-p

sandrina-p: Oh, I'm sorry @mmbotelho, this is in the Application page. 

As I said on the issue #788, I think it's missing some instructions there so the user know what to do. We thought something like:

Default instructions:
> If you are facing any problem with Group Income, download the logs and {a_}send to us{_a}, so that we can help you solve the problem.

Instructions if the user came from "report the error" link.
> You faced a recent error: "{errorMsg}". Please download the logs and {a_}send to us{_a}, so that we can help you solve the problem.

You can pull this PR and see it by yourself. Try to change the group mincome in a group with less than 3 members. It will appear a an error: _Failed to change the mincome. {errorMsg}. See error logs._

I think we had decided a few PR ago we wouldn't show the error message because it's not relevant to the average user. And "See error logs" doesn't explicit say that we want the user to report us the error. For that reason I think the following error msg is better: _Failed to change the mincome. Please, report the error._

frontend/views/utils/translations.js
sandrina-p: Issue created. https://github.com/okTurtles/group-income-simple/issues/885
I also added `LError()` everywhere it's needed.

frontend/controller/actions/identity.js
```diff
@@ -92,7 +90,7 @@ export default sbp('sbp/selectors/register', {
       return userId
     } catch (e) {
       console.error('gi.actions/identity/login failed!', e)
-      throw new GIErrorUIRuntimeError(L('Failed to login: {codeError}', { codeError: e.message }))
+      throw new GIErrorUIRuntimeError(L('Failed to login: {reportError}', LError(e)))
```
sandrina-p, line 93: @taoeffect here we have an interesting scenario: when this error happens, it doesn't make sense to ask the user to report us the error because it's a controlled / predictable / custom error.

image

So, my idea to overcome this problem is to throw a custom GIError instead of the default `TypeError`, so that `LError()` can detect it and not show the link. (See `translations.js`).

I remember a while ago you said we should not use `GIErrorUIRuntimeError` in this file but I don't remember why and now I even thing it's beneficial. Even if there's a problem, we can create another custom error. (a suggestion is welcome).

Anyway, with this solution, now the error is displayed correctly:

image

frontend/views/containers/proposals/Mincome.vue
sandrina-p: @taoeffect, here's a code review tip:

If you see that the same error is made multiple times, prefer to just leave one comment about it and remember the author to review it in the other places, too. Adding redundant comments only creates noise and might be contra-productive.

frontend/views/containers/user-settings/AppLogs.vue
mmbotelho: Ok that's clear now! Thank you for explaining. What about:

"Failed to change the mincome. You can report the error **here**."

As for the actual app log page, I think we need to add some instructions there, as you already have. Are you ok with not doing this on this PR though, and removing the message you added to the app logs page for now? We can create an issue for it, I can work on it and then pass it on to you. This is a bit of a chaotic way of working - I feel like the UX is being decided on code reviews and I don't think it's the most productive way of doing things. 

What do you think?

frontend/views/containers/proposals/Mincome.vue
taoeffect: OK, I will do my best to remember that for you. Different people seem to prefer (or need) different styles of code review.

frontend/views/containers/user-settings/AppLogs.vue
taoeffect: @mmbotelho I think we are in "crunch time" and that means a few things:

1. Yes, you are right, it is a chaotic way of doing things.
2. We should make minor changes like this anyway, but document them in issues to have them adjusted and fixed in future PRs.
3. The problem is that right now we need you @mmbotelho to focus on the UI design tasks you're focusing on, and *not* jump around to little things like this. When you get around to this (whenever that happens), then you can focus exclusively on this. But we need to move forward with this now and not be blocked on it. We want everyone to do a good job. The next few months are going to be a little bit more chaotic, but we will try to keep the chaos down to a minimum by keeping track of all such deviations via GH issues. This is why it's really important for @sandrina-p and me and all the other developers to open issues and assign them to you to review, if the changes are *both* important and minor.

I really like this text:

> You faced a recent error: "{errorMsg}". Please download the logs and {a_}send to us{_a}, so that we can help you solve the problem.

I would only change the first part to:

> Error: "{errorMsg}". Please download the logs and {a_}send to us{_a}, so that we can help you solve the problem.

@sandrina-p Please go ahead and implement this, but make sure to open an issue and assign to @mmbotelho to adjust UI for this as she sees fit.

mmbotelho: > The problem is that right now we need you @mmbotelho to focus on the UI design tasks you're focusing on, and not jump around to little things like this. When you get around to this (whenever that happens), then you can focus exclusively on this. But we need to move forward with this now and not be blocked on it. We want everyone to do a good job. The next few months are going to be a little bit more chaotic, but we will try to keep the chaos down to a minimum by keeping track of all such deviations via GH issues. This is why it's really important for @sandrina-p and me and all the other developers to open issues and assign them to you to review, if the changes are both important and minor.

Ok. Can we leave the adjustments to the App logs page out of it then? I think that's out of the scope of the original issue anyway

taoeffect: @mmbotelho 

> Ok. Can we leave the adjustments to the App logs page out of it then? I think that's out of the scope of the original issue anyway

Per our Slack chat, is it OK with you if we keep these changes in this PR and document them?

These changes are part of closing #788, as documented there:

> The debug page can use the passed in `error` query parameter to generate a draft of a message that the user can submit to us

We just forgot the design part of it, that's what was the mistake I think. Is it OK if in this case we move forward with it and fix it up in a future PR?

frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -55,6 +58,8 @@ export default {
       lastEntry = entry.prev
     } while (lastEntry)
     this.ephemeral.logs = logs.reverse() // chronological order (oldest to most recent)
+
+    this.ephemeral.reportError = this.$route.query.errorMsg
```
taoeffect: Since this variable is never re-assigned, I don't think it's needed? You should be able to access it via `this.$route.query.errorMsg` wherever it's needed right?

frontend/controller/actions/identity.js
taoeffect: For some reason your comment is appearing in `actions/identity.js`, so it doesn't make much sense.  I think you might have meant to write it under `contracts/group.js`?

> I remember a while ago you said we should not use GIErrorUIRuntimeError in this file

Not in `contracts/group.js`, but yes in `actions/identity.js`.

I don't see why the change is needed from TypeError in group.js

>  when this error happens, it doesn't make sense to ask the user to report us the error because it's a controlled / predictable / custom error.

Agreed with that, but again I'm confused because the lines about have nothing to do with changing the mincome but with logging in and signing up...?

frontend/model/contracts/group.js
```diff
@@ -298,7 +298,7 @@ DefineContract({
           }
 
           if (deepEqualJSONType(omit(prop.data.proposalData, 'reason'), dataToCompare)) {
-            throw new TypeError(L('There is an identical open proposal.'))
+            throw new Errors.GIErrorUIRuntimeError(L('There is an identical open proposal.'))
```
sandrina-p: @taoeffect here we have an interesting scenario: when this error happens, it doesn't make sense to ask the user to report us the error because it's a controlled / predictable / custom error.

image

So, my idea to overcome this problem is to throw a custom GIError instead of the default `TypeError`, so that `LError()` can detect it and not show the link. (See `translations.js`).

I remember a while ago you said we should not use `GIErrorUIRuntimeError` in this file but I don't remember why and now I even thing it's beneficial. Even if there's a problem, we can create another custom error. (a suggestion is welcome).

Anyway, with this solution, now the error is displayed correctly:

image

sandrina-p: @taoeffect now the comment is on the right place! Dunno what happened 😕

frontend/views/containers/user-settings/AppLogs.vue
mmbotelho: Sure! Can you please create an issue for it @sandrina-p ? Feel free to assign it to me

sandrina-p: So, let me recap this thread:
- The instructions I came up with will be implemented in this PR as asked by @taoeffect.
- I created an issue #890 about this change, so @mmbotelho can take her time to properly review these instructions and any copy related to the Applications pages or scoped error messages.
- We, as a team, will make an effort to avoid chaotic workflows in future tasks. When picking a new task/issue, let's make sure the UI/UX and technical requirements are clear as much as possible before moving to development phase. During the development, if something new needs to be added/changed, let's save it to another task to avoid chaotic decisions or bottlenecks in the process.

taoeffect: That sounds great to me @sandrina-p 😄  👍

frontend/controller/actions/group.js
```diff
@@ -122,10 +122,12 @@ export default sbp('sbp/selectors/register', {
   },
   'gi.actions/group/updateSettings': async function (settings, groupId) {
     try {
+      this.dumbError.forced = 123 // DELETE THIS BEFORE MERGE
       const message = await sbp('gi.contracts/group/updateSettings/create', settings, groupId)
       await sbp('backend/publishLogEntry', message)
     } catch (e) {
-      throw new GIErrorUIRuntimeError(L('Failed to update group settings. {codeError}', { codeError: e.message }))
+      console.error('gi.actions/group/updateSettings failed!', e)
```
taoeffect, line 128: Remember the `catch` in `GroupSettings.js` will also log this... so that's why we didn't include this `console.error` in the actions.

taoeffect: Ignore for now, no big deal, we can go back through this later.

frontend/model/contracts/group.js
taoeffect: It's really important for these to be TypeErrors in `contracts`... there's code in `state.js` that looks out for that. There's gotta be a better solution here. I don't see why changing this `GIErrorUIRuntimeError` matters, and it needs to be `TypeError` because of `state.js` stuff

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -118,8 +118,9 @@ export default {
 
           this.ephemeral.currentStep += 1 // Show Success step
         } catch (e) {
-          console.error(`Failed to change mincome to ${mincomeAmount}`, e.message)
-          this.$refs.formMsg.danger(L('Failed to change mincome. {codeError}', { codeError: e.message }))
+          this.dumbError.dumbError = 'forcer-error' // DELETE BEFORE MERGE!!
+          console.error('Mincome submit() error:', mincomeAmount, e.message)
+          this.$refs.formMsg.danger(L('Failed to change mincome. {reportError}', LError(e)))
```
taoeffect: Important rule I learned the hard way: to save Future Sandrina, make all user-facing errors unique.

So here it should be: `Failed to propose mincome change: {reportError}` (or, if you don't like the link, leave out `{reportError}` and replace it with `{errorMsg}` and set that to `e.message`

This way the error will be unique from the one below.

frontend/views/utils/translations.js
```diff
@@ -65,6 +66,20 @@ export default function L (
   return template(key, args)
 }
 
+export function LError (error) {
+  if (error instanceof GIErrorUIRuntimeError) {
+    // It's a controlled/predictable error, so we don't need the user to report it to us.
+    return { reportError: error.message }
+  }
+  return {
+    reportError: L('"{errorMsg}". {a_}Report the error{_a}.', {
```
taoeffect: If you want, I recommend additionally adding a key for `errorMsg` at the top level here too (along with the previous `troubleshootingLink`, so that `LError` usage can be made more flexible if you want it.

frontend/views/utils/translations.js
```diff
@@ -65,6 +66,20 @@ export default function L (
   return template(key, args)
 }
 
+export function LError (error) {
+  if (error instanceof GIErrorUIRuntimeError) {
+    // It's a controlled/predictable error, so we don't need the user to report it to us.
+    return { reportError: error.message }
+  }
```
taoeffect: Remove (see other comments for why)

taoeffect (changes requested):
See comments!

frontend/model/contracts/group.js
sandrina-p: It can't be `TypeError` so we can filter the type of error on `LError` as I explained on the previous comment. With it, we can choose to not show the "report" link within the error message. 

I'm sorry Greg, what's the problem with `TypeError`? What if we create a new type of error `GiUIErrorContract`, couldn't we modify state to adapt to it? I didn't find any other solution, good or bad.

frontend/views/utils/translations.js
sandrina-p: I didn't understand what you mean nor the flexibility advantage... 😕

taoeffect: So sometimes you want this exact string in `reportError` that you have here, but other times you might want to write a different string, e.g. have the positioning of `errorMsg` be different.

taoeffect: So sometimes you want this exact string in `reportError` that you have here, but other times you might want to write a different string, e.g. have the positioning of `errorMsg` be different.

frontend/model/contracts/group.js
taoeffect: > It can't be TypeError so we can filter the type of error on LError as I explained on the previous comment.

Needs to be a different way of doing it, can't do it this way, and `LError` must not check for error types.

> I'm sorry Greg, what's the problem with TypeError?

See `state.js` and search for `TypeError`

> What if we create a new type of error GiUIErrorContract, couldn't we modify state to adapt to it? 

Nope, doesn't make sense. Contract errors are not UI errors.

taoeffect: Feel free to check for the error type in this modal. The contract code uses very specific error types and changing those will break things.

sandrina-p: > Needs to be a different way of doing it, can't do it this way, and LError must not check for error types.

Any suggestion? I'm not seeing any way of detecting the type of message if we don't throw a custom error.

> Nope, doesn't make sense. Contract errors are not UI errors.

Hum... when we say `There is an identical open proposal.` it sounds very much as an UI error.

> Feel free to check for the error type in this modal

How do we check the error type if we can't change `TypeError`?

taoeffect: > Any suggestion? I'm not seeing any way of detecting the type of message if we don't throw a custom error

Yes, feel free to do these checks in the Modal code.

> How do we check the error type if we can't change TypeError?

You can check for TypeError in the Modal, can't you?

> Hum... when we say There is an identical open proposal. it sounds very much as an UI error.

It is. Doesn't mean we need to break the contract code or add unnecessary checks to LError.

Sorry for curt replies right now. I'm hoping you can figure this out while I try to figure out the payments/currency stuff!

sandrina-p: Yes, we can check the instance of Error on the modal, but for it to work we need to create a custom typeError, like `GIErrorContract`. If we just throw a normal `TypeError`, we can't detect the "custom error".

taoeffect: A `TypeError` is a custom error. You can check for it just like `state.js` checks for it

sandrina-p: Hum... I'll try it out @taoeffect and let you know when I have updates. Thanks and good luck on the payments stuff!

frontend/views/containers/proposals/Mincome.vue
sandrina-p: I didn't understand how this solves the current problem. And I never said I "don't like" the link. There are situations where the link is needed and others where it isn't. We can't simply exclude it from every cases.

frontend/views/utils/translations.js
sandrina-p: > but other times you might want to write a different string.

Do we? Where? Maybe I missed it. Anyway, I thought we agreed to do it as it is, to keep consistency and DRY. 

Quote from https://github.com/okTurtles/group-income-simple/pull/881#discussion_r405861003:
> What if error, comes from {reportError}? That way we don’t need to repeat the 2 variables everywhere.

frontend/model/contracts/group.js
sandrina-p: @taoeffect, both errors _"Cannot set property 'dumbError' of undefined"._ and _"There is an identical open proposal"_ return true when `instanceof TypeError`, so we cannot distinguish between both in the code, although they are different errors from a UX perspective.

So, since I couldn't find a solution that works in your POV and you don't have the time to properly try it in your computer, I suggest to leave it with the simple approach where we always show the link when there's an error and opened an issue (bug) assigned to you to fix it later. https://github.com/okTurtles/group-income-simple/issues/894

taoeffect: @sandrina-p brilliant resolution! 👍 🥇

frontend/views/containers/proposals/Mincome.vue
taoeffect: Oh, that's not exactly what I'm talking about here (only partly).

We can keep the link, that's fine, the main issue I'm concerned about here is that the two error messages are the same, so a user comes to us and tells us, "The app said 'Failed to change mincome.'", we won't be able to tell where the error occurred.

This is why you always make the wording of error messages unique if they're located in different locations. It's a "people hack" that's really helpful. :)

So can you rewrite this as `Failed to propose mincome change: {reportError}`?

frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -29,16 +30,18 @@ import { mapState, mapMutations, mapActions } from 'vuex'
...
-  components: {},
   data () {
     return {
       form: {
         filter: []
       },
       ephemeral: {
-        logs: []
+        logs: [],
+        reportError: null
```
taoeffect: This `reportError` variable isn't being used anymore, so it can be removed.

frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -82,6 +85,16 @@ export default {
...
+    },
+    instructions () {
+      const errorMsg = this.$route.query.errorMsg
+      const linkTag = {
+        'a_': '<a class="link" target="_blank" href="https://github.com/okTurtles/group-income-simple/issues">',
+        '_a': '</a>'
+      }
+      return errorMsg
+        ? L('You faced a recent error: "{errorMsg}". Please download the logs and {a_}send to us{_a}, so that we can help you solve the problem.', { errorMsg, ...linkTag })
+        : L('If you are facing any problem with Group Income, download the logs and {a_}send to us{_a}, so that we can help you solve the problem.', linkTag)
```
taoeffect: Minor wording change:

- `You faced a recent error: "{errorMsg}". Please download the logs and {a_}send to us{_a}, so that we can help you solve the problem.` => `Recent error: "{errorMsg}". Please download the logs and {a_}send them to us{_a} so we can help troubleshoot.`
- `If you are facing any problem with Group Income, download the logs and {a_}send to us{_a}, so that we can help you solve the problem.` => `If you encounter problems, please download the logs and {a_}send them to us{_a}.`

taoeffect (changes requested):
Great job! Just 3 minor change requests to the strings in this PR and it should be good to go IMO. 👍 

For convenience and to avoid losing track of them (because GitHub's UI is a bit cluttered), I link them here:

- https://github.com/okTurtles/group-income-simple/pull/881/files#r407103772
- https://github.com/okTurtles/group-income-simple/pull/881/files#r408386776
- https://github.com/okTurtles/group-income-simple/pull/881/files#r408390777

(Links don't work on mobile.)

frontend/views/containers/proposals/Mincome.vue
sandrina-p: Ah got it now, completely agree!

taoeffect (approved):
LGTM! 👍

Issue #778, opened 2019-11-15
Modals - Keyboard navigation is incorrect.


sandrina-p:
### Problem

I'll give the example of Login modal, but this happens to all the modals.

When we navigate in a modal with the keyboard, and we try to focus the last element of it (for example "Login" button, (after being enabled!), the focus position moves immediately to the close button. It should instead stay on the "Login" button.

### Solution
The way we trap focus needs to be done with a native 'keyup' eventListener to the document, not the modal itself (as it is nowadays). That way we can detect when the user starts to tab to outside of the modal and redirect them back to the first clickable element inside the modal.

PR #884, opened 2020-04-06
#798 Implement profile card

### LINKED ISSUE #798: Create a way to see other group members profile
### Problem

There should be a place where group members can see others:
- Payment information;
- Bio;
- Date they joined the group;
- [Action] Send message;
- [Action] Create proposal to remove member.

### Solution

Figure out how, where and when this information is displayed:
- Add a way for users to preview their profile. Probably somewhere in the User Settings;
- [Desktop] show this information on-hover;
- [Mobile] use a drawer component;

What I have in mind is very similar to the way Slack does it. 

Next steps:
- Create mockups and get feedback!

pieer:


frontend/assets/style/_transitions.scss
```diff
@@ -150,6 +150,18 @@
   }
 }
 
+@keyframes enterFromLeft {
+  from {
+    opacity: 0;
+    transform: translateX(-30px);
```
sandrina-p: `30px` -> `2rem`

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -43,6 +45,8 @@ import sbp from '~/shared/sbp.js'
 import Avatar from '@components/Avatar.vue'
 import AvatarUser from '@components/AvatarUser.vue'
 import GroupMembersMenu from '@containers/dashboard/GroupMembersMenu.vue'
+import Tooltip from '@components/Tooltip.vue'
```
sandrina-p: Tooltip is not being used in this file

frontend/views/components/UserName.vue
```diff
@@ -0,0 +1,49 @@
...
+    data-test='profileName'
+    v-if='displayName'
+  ) @{{ username }}
+</template>
+
+<script>
+import { mapGetters } from 'vuex'
+
+export default {
+  name: 'UserName',
```
sandrina-p, line 14: Do you plan to use this component anywhere else? As it is, it seems an extra file with low "value".

frontend/views/components/Tooltip.vue
```diff
@@ -1,85 +1,138 @@
...
+      if (this.manual && window.innerWidth < TABLET) {
+        y = windowHeight - this.tooltipHeight // Position at the bottom of the screen on mobile
+        x = -8 // Remove tooltip padding
+      } else {
+        y = scrollY + top + height / 2 - this.tooltipHeight / 2
+        if (y < 0) y = spacing
+        if (y + this.tooltipHeight > windowHeight) y = windowHeight - spacing - this.tooltipHeight
+
+        switch (this.direction) {
+          case 'right':
```
sandrina-p, line 107: Sweet! ✨

frontend/views/components/Tooltip.vue
```diff
@@ -90,15 +143,30 @@ export default {
       inserted (el, bindings, vnode) {
         document.body.appendChild(el)
 
-        const $this = vnode.context // Vue component instance
+        if (el.classList.contains('c-tooltip')) {
```
sandrina-p: In what scenarios would `c-tooltip` not exist? This code is executed when the tooltip is already `isActive`, so this condition would always be true

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+import { mapGetters } from 'vuex'
+
+export default {
+  name: 'ProfileCard',
+  props: {
+    username: String,
+    isSelf: {
+      type: Boolean,
+      default: false
+    },
```
sandrina-p: What if we calculate `isSelf` in this same component as a property, based on `username` prop?
This component will be widely used across the app and we would DRY a lot `username === ourUsername`.

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+  },
+  components: {
+    AvatarUser,
+    UserName,
+    Tooltip
+  },
+  data () {
+    return {
+      // Temp
+      fakeStore: {
```
sandrina-p: The `paymentMethods` is already fully implemented. You can access it at `this.groupProfiles[username].paymentMethods`.

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+
+      .buttons(v-if='!isSelf')
+        i18n.button.is-outlined.is-small(
+          tag='button'
+          @click='sendMessage'
+        ) Send message
+
+        i18n.button.is-outlined.is-small(
+          tag='button'
+          @click='removeMember'
```
sandrina-p: There's no need for `removeMember` method. It can be:
```
@click='openModal("RemoveMember", { username })'
```

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+      'groupMincomeFormatted',
+      'globalProfile',
+      'ourContributionSummary'
+    ]),
+    profile () {
+      return this.$store.getters.globalProfile(this.username)
+    },
+    isActiveGroupMember () {
+      return true
+    }
```
sandrina-p:
```
return !!this.groupProfiles[username]
```

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+      console.log('To do: implement send message')
+    },
+    removeMember () {
+      console.log('To do: implement remove member')
+    },
+    hasIncomeDetails () {
+      return !!this.ourGroupProfile.incomeDetailsType
+    },
+    receivingMonetary () {
+      return this.ourContributionSummary.receivingMonetary
```
sandrina-p: use `!!` to return a Bool as you did in `hasIncomeDetails`.

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
...
+
+<style lang="scss" scoped>
+@import "@assets/style/_variables.scss";
+
+.card {
+  padding: 1rem 1.5rem;
+  color: $text_1;
+  max-width: 100vw;
+  width: 24.3rem;
+  box-shadow: 0px 8px 20px rgba(54, 54, 54, 0.3);
```
sandrina-p: `0 0.5rem 1.25rem`

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,243 @@
+<template lang='pug'>
+tooltip(
+  :direction='direction'
+  :manual='true'
+  ref='tooltip'
+  :opacity='1'
+)
+  slot
```
sandrina-p, line 9: The designs show a light underline style when hovering the ProfileCard trigger (`slot`).

We have an accessibility problem here: We can't reach this component using the keyboard. I will ping @mmbotelho on Figma to double check the focus states and cc you.

On the technical side, the strategy to do an accessible "popover" menu is very similar to the one used for modals. On trigger click move the focus to the popover and then trap the keyboard navigation inside it. (See #778). I can help you if you need! Slack (browser version) do it perfectly. It's a great reference as A11Y best practices. ✨

sandrina-p (commented):
Yey @pieer this seems great!! Love the smoothness :D

My only concern is around accessibility. Besides that I didn't find any other major problem.

P.S. This PR seems to close #798. Don't forget to add it to the PR description so it gets closed automatically.

frontend/views/components/UserName.vue
pieer: Yes, everywhere we have the combination of name or name and displayName, which is in many places

mmbotelho (changes requested):
Here's what I found @pieer :

Captura de ecrã 2020-04-14, às 11 53 39

Captura de ecrã 2020-04-14, às 11 46 04

Captura de ecrã 2020-04-14, às 11 56 43
Captura de ecrã 2020-04-14, às 11 49 53

Suggestion:
- Can we make the animation similar to the one we have when the user settings (or any fullscreen modal) appears? 

Question:
- How can I test the versions of the card that don't have the payment methods there? Are you planning on doing that in this PR or in a future one? I'm ok with either, I just need to know if I should review them now or not 😄

mmbotelho (dismissed):
Approved @pieer ! Great job on this one!

frontend/views/containers/navigation/Profile.vue
```diff
@@ -2,18 +2,19 @@
...
-        :data-test='userDisplayName ? "profileDisplayName" : "profileName"'
-      ) {{ userDisplayName ? userDisplayName : ourUsername }}
+)
+  profile-card(:username='ourUsername' :isSelf='true' direction='top-left')
+    .c-avatar-user
+      avatar-user(:username='ourUsername' size='sm')
+      .c-user
+        strong(
+          :data-test='userDisplayName ? "profileDisplayName" : "profileName"'
+        ) {{ userDisplayName ? userDisplayName : ourUsername }}
```
taoeffect: (minor) can be written as:

```
{{ userDisplayName || ourUsername }}
```

frontend/views/containers/navigation/Profile.vue
```diff
@@ -2,18 +2,19 @@
...
 
-      span(
-        data-test='profileName'
-        v-if='userDisplayName'
-      ) @{{ ourUsername }}
+        span(
+          data-test='profileName'
```
sandrina-p, line 15: This `<span />` needs to be a `<button />`, so that someone with a keyboard can interact with it and open the profile card.

The same should happen on the members' list. A way to detect these issues is by using the keyboard only and try to open and close the profile card. Use `Tab` to go forward and `Shift + Tab` to go backward.

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -118,6 +125,10 @@ export default {
       // We access directly the modal here to avoid broacasting event to every possible modal
       this.$refs.modalTemplate.close()
     },
+    openModal (mode) {
+      sbp('okTurtles.events/emit', OPEN_MODAL, mode)
+      return false
```
sandrina-p: `return false` isn't needed, or is it?

frontend/views/components/modal/ModalBaseTemplate.vue
```diff
@@ -16,10 +16,11 @@
 
 <script>
 import modalMixins from './ModalMixins.js'
+import trapFocus from '@utils/trapFocus.js'
 
 export default {
   name: 'ModalBaseTemplate',
-  mixins: [modalMixins],
+  mixins: [modalMixins, trapFocus],
```
sandrina-p, line 23: Nice add! I completely forgot about this one.

frontend/views/components/UserName.vue
```diff
@@ -0,0 +1,49 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import "@assets/style/_variables.scss";
+
+.c-name {
+  display: flex;
+  flex-direction: column;
+  padding-left: 1rem;
+  font-size: 1rem;
```
sandrina-p: The default text size is `$size_4` (14px / 0.875rem), not 1rem.

frontend/views/components/Tooltip.vue
```diff
@@ -1,85 +1,150 @@
...
     hide () {
-      this.isActive = false
+      if (!this.manual) this.isActive = false
+    },
+    toggle () {
+      this.isActive = !this.isActive
+      if (this.$slots.tooltip) {
+        document[this.isActive ? 'addEventListener' : 'removeEventListener']('keydown', this.trapFunction)
+        // Wait for element to be present before auto focus
+        setTimeout(() => this.focusOnFirst(this.$slots.tooltip[0].elm))
```
sandrina-p: This works great on first try but if we close the profile card and then open it again it doesn't focus anymore. 🐛

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,241 @@
...
+
+  template(slot='tooltip')
+    .card.c-profile(v-if='profile')
```
sandrina-p: One major requirement to create an effective trapfocus system is to do it within a "dialog" element. As done in modal, we need to do it here too: `role="dialog"` [role dialog MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) and it needs an accessible label. Here it can be `aria-label="{username} profile"`

sandrina-p (changes requested):
Halfway there! Once we can open and close a profile card using the keyboard only, this is ready to go :D

frontend/views/components/Tooltip.vue
```diff
@@ -1,85 +1,150 @@
...
-    :class='isTextCenter ? "has-text-center" : ""'
+    :style='styles'
+    :class='{"has-text-center": isTextCenter, "is-active": isActive}'
     v-if='isActive || isVisible'
     v-append-to-body=''
   )
+    modal-close(
+      v-if='manual'
+      @close='toggle'
+    )
```
sandrina-p: This close button is part of the profile card, not the tooltip. Why did you choose to add it to the tooltip instead?

-- EDIT 1 --
It can't be outside the profile card, otherwise, we won't be able to close the tooltip using the keyboard. I suggest to add it to the profileCard as the last element

-- EDIT 2 --
I was about to ask to also move ".c-background" to ProfileCard but then we couldn't style it correctly :/ So, for now, let's keep it as it is under the `manual` prop. In the future, if needed we improve it.

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,241 @@
...
+
+      div(v-if='hasIncomeDetails')
+        ul.c-payment-list
+          li.c-payment-item(v-for='payment in fakeStore.paymentMethods')
+            span.c-payment-type.has-text-0.has-text-bold {{payment.name}}
+            span.has-text-1 {{payment.address}}
+
+        i18n.is-unstyled.link(
+          v-if='isSelf && receivingMonetary'
+          tag='button'
```
sandrina-p: A `<button>` does not need `.is-unstyled` class. Actually this breaks the :hover/:focus styling.

frontend/views/containers/navigation/Profile.vue
```diff
@@ -64,6 +67,13 @@ export default {
 .c-avatar-user {
   display: flex;
   align-items: center;
+
+  &:focus,
+  &:hover {
+    .c-user {
+      text-decoration: underline;
```
sandrina-p: The DS explicitly say the links should have `border-bottom` instead of `text-decoration`. The same should be made at `GroupMembers.vue`

frontend/views/components/Tooltip.vue
```diff
@@ -1,85 +1,150 @@
...
     isActive: false,
-    stylesPosition: null
+    styles: null
   }),
+  mounted () {
+    window.addEventListener('resize', this.adjustPosition)
+  },
+  beforeDestroy: function () {
+    window.removeEventListener('resize', this.adjustPosition)
+  },
```
sandrina-p: This event should be only added/removed when the tooltip opens/closes. The best place to do it is inside the directive `appendToBody`.

frontend/views/components/Tooltip.vue
sandrina-p: @pieer I found the issue and fixed it!

So, you were right: `this.$slots.tooltip[0].elm` is the problem. It's correct (the actual element) on the first try but after it unmounts, it's lost. As you said, I think it's because each time the element is appended to the body, it creates a new instance, but the `$slot` elm is still the original, so it's not exactly the same element. That's why it can't focus it.

If we do this logic (keydown eventlistener and to focus) directly inside the directive `appendToBody`, it works, because there we are dealing directly with the element, instead of a Vue $slot "shallow copy".

One extra note: When the element is created, we should not `focusOnFirst`. Instead we should focus the element itself as we do in the modal using `v-focus`. 

Anyway... all the changes needed to make this work 100% for keyboard users and screen readers are done in a single commit at [this branch (a copy of yours)](https://github.com/okTurtles/group-income-simple/commit/8e1a889e5a750b1d4e96c00272a8e8f52fd45f2b). Take special attention to the HTML attributes added such as `tabindex`, `role`, and `aria-label`. They are important.

If you need help to understand something in that code, I'm free to help you in a video call :)

sandrina-p (commented):
@pieer, I've found the bug cause and proposed a solution! See comments

frontend/utils/trapFocus.js
```diff
@@ -7,8 +7,12 @@ const trapFocus = {
     }
   },
   methods: {
-    focusOnFirst (el = this.$el) {
+    focusEl (el = this.$el) {
+      // prepare the element to be focused programmatically
       Array.from(el.querySelectorAll(this.focusableElements))[0].focus()
```
sandrina-p: This line can be removed. We do not need to focus the first focusable element, only the el itself.

frontend/views/components/Tooltip.vue
```diff
@@ -177,10 +161,24 @@ export default {
...
+          document.addEventListener('keydown', e => $this.trapFocus(e, el))
+          $this.lastFocus = document.activeElement
+          $this.focusEl(el)
+        }
       },
-      unbind (el) {
+      unbind (el, bindings, vnode) {
         if (el.parentNode) {
           el.parentNode.removeChild(el)
+          window.removeEventListener('resize', this.adjustPosition)
```
sandrina-p: The resize event should be removed always, even if there isn't a `parentNode`. Otherwise, we'll have a memory leak. And it should be `$this` instead of `this`.

frontend/views/components/ProfileCard.vue
taoeffect: Agreed makes more sense to calculate as a computed property in this component 👍

frontend/views/components/ProfileCard.vue
taoeffect: Yep, feel free to implement this part!

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,268 @@
...
+      'groupProfiles',
+      'groupMincomeFormatted',
+      'globalProfile',
+      'ourContributionSummary'
+    ]),
+    profile () {
+      return this.$store.getters.globalProfile(this.username)
+    },
+    isActiveGroupMember () {
+      return !!this.groupProfiles[this.username]
```
taoeffect: FYI in #900 (Payments) this is going to change, there will be a `.status` property on the groupProfile that you can check, e.g. `return this.groupProfiles[this.username].status === INVITE_STATUS.VALID`.

So either let's wait for that to be merged and update this, and/or add a `// TODO:` comment to that effect here to not forget to update this.

taoeffect (changes requested):
Great job @pieer! Looks almost perfect from my testing of it. In addition to what @sandrina-p noticed, I wrote some minor comments.

One issue I did notice had to do with the animation. During testing I noticed a weird "jump" behavior, which I recorded here:

out

sandrina-p (dismissed):
Yey! All good! The final touch with a fade-in was great!

mmbotelho (dismissed):
My last approval was stale, but I still approve this! 😄

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,254 @@
...
+        ) Edit bio
+
+      i18n.button.is-small.is-outlined.c-bio-button(
+        v-else-if='isSelf'
+        tag='button'
+        @click='openModal("UserSettingsModal")'
+        data-test='buttonEditBio'
+      ) Add a bio
+
+      div(v-if='hasIncomeDetails')
```
taoeffect: We should show a user's payment details regardless of whether or not we've filled in our own details.

taoeffect (changes requested):
Excellent work @pieer! You fixed the animation jitter I found before. This time I found two bugs and a UX issue:

### Bug 1: Showing incorrect payment details

Screen Shot 2020-05-05 at 12 50 25 PM

In this screenshot it shows "sam" as having bitcoin information, but that's wrong, because sam didn't enter their income details at all (didn't say they're pledging or receiving).

The bitcoin information displayed belongs to use "greg".

### Bug 2: Not showing payment details when they should be shown

Screen Shot 2020-05-05 at 12 50 44 PM

Screen Shot 2020-05-05 at 12 54 40 PM

As I played around with the app, user "joe" was shown as either not having any payment details (first screenshot above, which is what you see when looked at from "joe"'s perspective, who has not entered any payment details), or having the wrong payment details as shown in the 2nd screenshot above (when seen from user "greg"'s perspective).

The correct payment details of "Paypal: super@paypal.com" aren't shown.

### UX Issue: animation for user profile popup too slow

When bringing up the profile drawer on mobile, the profile takes too long to appear. Can you please reduce the animation time to make it appear nearly instantly? It should be about 100ms or so.

Am going to continue looking through the code and posting comments. Just getting this review to you as quickly as I can so that you can start working on it sooner than later. I'll let you know either here or on Slack (or both) when I'm done reviewing all the code, which should be in an hour or so.

EDIT: also, please add two tests to catch those bugs ☝️ so that they don't return in the future and so that we ensure that the profile cards will display the correct information.

taoeffect:
@pieer can you add tests to catch those two bugs also?

frontend/views/components/Tooltip.vue
```diff
@@ -91,19 +144,48 @@ export default {
...
+        }
       },
-      unbind (el) {
+      unbind (el, bindings, vnode) {
+        const $this = vnode.context
         if (el.parentNode) {
           el.parentNode.removeChild(el)
         }
+        if (bindings.value && bindings.value.manual) {
+          document.removeEventListener('keydown', e => $this.trapFocus(e, el))
```
taoeffect: My understanding is that `removeEventListener` must be called on the exact saved instance of a function.

Here it's being called on a new instance of a function, `e => $this.trapFocus(e, el)`, instead of a reference to an existing instance. This will result in a memory leak so it must be fixed.

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -8,6 +8,11 @@ modal-template(class='is-centered is-left-aligned' back-on-mobile=true ref='moda
     name='formData'
     @submit.prevent='changePassword'
   )
+    i18n.link(
+      tag='button'
+      data-test='passwordBtn'
+      @click.prevent='openModal("incomeDetails")'
+    ) Update Password
```
taoeffect: This clearly doesn't belong here. Not sure why this file was modified.

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -17,15 +17,17 @@
       :class='invitedBy && "is-pending"'
       :key='username'
     )
-      avatar(v-if='invitedBy' size='sm')
-      avatar-user(v-else :username='username' size='sm')
+      profile-card(:username='username' :isSelf='username === ourUsername')
```
taoeffect: I think @sandrina-p mentioned (and I agreed) that this `isSelf` property should be removed. Instead the profile card should have a computed property to do this.

taoeffect:
OK, finished code review as well, please see new comments.

Also, I noticed this wasn't implemented, yet @mmbotelho gave approval...

Screen Shot 2020-05-05 at 1 04 07 PM

mmbotelho (commented):
This is essentially the same thing @taoeffect pointed out, but I just want to add these screenshots with implemented and designed side by side,  so it's easier to understand:

Captura de ecrã 2020-05-06, às 11 41 26
Captura de ecrã 2020-05-06, às 11 40 18

sandrina-p (changes requested):
Can't wait to see this on master 🚀

sandrina-p (dismissed):

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -118,6 +120,9 @@ export default {
       // We access directly the modal here to avoid broacasting event to every possible modal
       this.$refs.modalTemplate.close()
     },
+    openModal (mode) {
+      sbp('okTurtles.events/emit', OPEN_MODAL, mode)
+    },
```
taoeffect: Unused code should be removed.

frontend/views/components/ProfileCard.vue
```diff
@@ -0,0 +1,252 @@
...
+        v-else-if='isSelf && hasIncomeDetails'
+        tag='button'
+        @click='openModal("UserSettingsModal")'
+        data-test='buttonEditBio'
+      ) Add a bio
+
+      div(v-if='hasIncomeDetails')
+        ul.c-payment-list
+          li.c-payment-item(v-for='(paymentMethod, name) in this.paymentMethods')
+            span.c-payment-type.has-text-0.has-text-bold {{name}}
```
taoeffect: `name` is the index, so this should be `paymentMethod.name`

taoeffect (changes requested):
Excellent work @pieer!

Just a few issues remaining:

- [This comment from the previous review about `removeEventListener`](https://github.com/okTurtles/group-income-simple/pull/884/files#r420399119) still needs to be addressed
- The speed of the mobile drawer appearing still needs to be made faster (I think 2x faster would keep the UI feeling snappy and not sluggish)
- There's a bug where a number is shown instead of the payment method:

Screen Shot 2020-05-08 at 2 09 29 PM

frontend/views/components/ProfileCard.vue
sandrina-p: Good catch. Probably @pieer didn't get the change to see this PR #912 that changed how the payments are stored, what's why this broke the page.

mmbotelho (approved):
Perfect @pieer !

taoeffect (approved):
Fantastic work @pieer, the new Profile card looks and works great! 🥇 💯

PR #888, opened 2020-04-11
#778 improve modal keyboard navigation

### LINKED ISSUE #778: Modals - Keyboard navigation is incorrect.

sandrina-p:
Closes #778 

Apr-11-2020 14-49-02

- Improved the keyboard navigation within a modal: It now works as expected: when reaching the last interactive element, the next time we press `Tab` it goes back to the first element and vice-versa.
- The "solid" buttons didn't change background color on `:focus`, Now they do, as in DS.
- Add a required prop to modal: `a11yTitle` so that SR (screen readers) can read correctly what's the modal about when they start to interact with it.

### DIFF
--- a/frontend/views/containers/access/LoginModal.vue
+++ b/frontend/views/containers/access/LoginModal.vue
@@ -1,5 +1,5 @@
 <template lang='pug'>
-  modal-template(class='has-background' ref='modal')
+  modal-template(class='has-background' ref='modal' :a11yTitle='L("Login")')
     template(slot='title')
       i18n Log in
 
--- a/frontend/views/containers/access/LoginModal.vue
+++ b/frontend/views/containers/access/LoginModal.vue
@@ -1,5 +1,5 @@
 <template lang='pug'>
-  modal-template(class='has-background' ref='modal' :a11yTitle='L("Login")')
+  modal-template(class='has-background' ref='modal' :a11yTitle='L("Log in")')
     template(slot='title')
       i18n Log in
 
@@ -9,7 +9,7 @@
       p
         i18n Not on Group Income yet?
         | &nbsp;
-        i18n.link(tag='button' data-test='goToSignup' @click='showSignupModal') Create an account
+        i18n.link(tag='button' @click='showSignupModal' data-test='goToSignup') Create an account
 </template>
 
 <script>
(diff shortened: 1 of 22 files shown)


pieer (dismissed):

dotmacro:
Log in modal works great! :-D 

Tab still misses the "Log in" _link_ in the footer of the sign up modal. (Should it also be a button, like "Create an account" in the footer of the Log In modal?)
Screen Shot 2020-04-13 at 7 25 49 PM

Scout rule: "Login" button on welcome screen and "Log in" modal should say "Log in" with a space, just like the modal title.

frontend/views/containers/access/LoginModal.vue
```diff
@@ -1,5 +1,5 @@
 <template lang='pug'>
-  modal-template(class='has-background' ref='modal')
+  modal-template(class='has-background' ref='modal' :a11yTitle='L("Login")')
```
taoeffect: Andrea noticed that we've been writing "Login" incorrectly, it should be "Log In"

taoeffect (changes requested):
Just one minor copy issue related to "Login" vs "Log In" (Andrea noticed we've been writing that incorrectly) + if you can see if you can fix the issue she noticed with consistent tab behavior between Log In and Sign Up modals.

sandrina-p:
@dotmacro the "Log in" link is fixed now and accessible by keyboard!

> Should it also be a button, like "Create an account" in the footer of the Log In modal?

Yes, it should be a button. Before it was a `<a>` tag without `href=''`. An `<a>` element is only keyboard accessible when it has a non-empty href attribute. `<a>` or `<a href="">` should not be used for links. In this case it must be a `<button>`. [webAIM reference](https://webaim.org/techniques/keyboard/)

taoeffect (approved):
👍

PR #892, opened 2020-04-11
Add a few translations

SGrondin:
Added a few French translations and ran v1.0.2, which prints the version number at the top, keeps file names sorted (to avoid useless git changes) and fixes duplicated file names.


taoeffect (approved):
Awesome 👍

PR #895, opened 2020-04-13
Close #685 Add the ability to have many subcontent modal in URL

### LINKED ISSUE #685: Cleanup and simplify Modal.vue implementation
### Problem

Modal.vue is really amazing, but is accumulating some cruft and it's possible to make a few simplifications. I noticed, for example:

1. There exist at least a couple methods that are used only once, these include at least: `initializeModals()` and `closeModal()`
2. At the moment the `data ()` method returns an object with 3 fields, and it's not clear whether `subcontent` should be an array. Additionally, it's not clear to me whether the implementation is correct, as `this.subcontent.splice(0, i)` seems to remove all previously opened modals from the array

### Solution

- For any methods that are called once, move their contents to the call site
- The URL currently only supports listing two modals in it, and it's unclear we ever have a use-case for there being more than 2 modals on top of one another, so `subcontent` probably shouldn't be an array, and if it does need to be an array then the implementation likely needs to be fixed, and the URL needs to be updateable to include all modals, in which case it's still not clear whether `subcontent` should be an array (as the array should be specified and available via the URL as a query parameter, e.g. `?modal=FirstModal+SecondModal+ThirdModal`)

pieer:


taoeffect:
@pieer just a heads up, this now has a (small) merge conflict

taoeffect:
@pieer It's a bit unclear to me what issue this PR is solving, I understand it's an improvement of some sort, but could you please open up a Problem/Solution issue for this? (Every PR needs one, and they're super helpful in making sure we're solving real problems.)

pieer:
@taoeffect it address part of the issue Cleanup and simplify Modal.vue implementation #685

frontend/views/components/modal/Modal.vue
```diff
@@ -70,19 +73,15 @@ export default {
...
       if (this.content) {
         this.$router.push({
           query: {
             ...this.$route.query,
-            modal: this.content,
-            subcontent: this.activeSubcontent()
+            ...{
+              modal: this.content,
+              subcontent: this.subcontent.length ? this.subcontent.join('+') : undefined
+            }
```
taoeffect: @sandrina-p removed the extra braces `{ }` around `modal` and `subcontent`, is there a reason they were added back?

taoeffect (changes requested):
Great job @pieer. This looks almost ready to merge, I just have two concerns:

1. See the comment about `...{` usage (@sandrina-p had removed it, if it needs to be added back please give a reason)
2. I've been going through the amazing Design System page you and @sandrina-p have been working on (it's really incredible!). I noticed that it has a demo for the modal. Can you make sure the existing demos work with this change, and add another demo that uses this new feature so that we can make sure it works now and in the future?

pieer:
The demo was already demonstrating the modal in the modal @taoeffect

taoeffect:
Yes I know, but it wasn’t demonstrating the feature added in this PR is what I meant.

frontend/views/components/modal/Modal.vue
```diff
@@ -39,7 +42,7 @@ export default {
       if (to.query.modal) {
         // We reset the modals with no animation for simplicity
         if (to.query.modal !== this.content) this.content = to.query.modal
-        const subcontent = to.query.subcontent
+        const subcontent = to.query.subcontent ? to.query.subcontent.split('+').pop() : []
```
taoeffect, line 45: Now that this PR supports `+`, does the modal still need `subcontent` variable, or can it do the same thing using just the `modal` variable? E.g. `?modal=Modal1+Modal2+Modal3` ?

pieer: We need subcontent because there is two kind of modals with different behavior according to the modal type

taoeffect: Ah, OK, makes sense 👍

taoeffect:
Pierre how does this work?

I tried visiting:

http://localhost:8000/app/design-system?modal=DSModalSimple&subcontent=DSModalNested+LoginModal

And that didn't bring up 3 modals, it brought it back down to 1.

Visiting the design page, I do see that it seems like it's possible to bring up 3 modals by clicking the buttons. It will load this link:

http://localhost:8000/app/design-system?modal=DSModalSimple&subcontent=DSModalNested%2BSignupModal

However, visiting that link (or refreshing the page) will bring it back down to one modal.

Furthermore, attempting to add a fourth modal manually via the URL also doesn't work when visiting the page:

http://localhost:8000/app/design-system?modal=DSModalSimple&subcontent=DSModalNested%2BSignupModal%2BLoginModal

pieer:
@taoeffect I'm not sure how to fix this, neither if it's a good idea to create modal history out of the blue. 
The purpose of having history in the navigation is to be able to go back on the previous modal. Going straight to some specific action like you did by hijacking the url is not the intended behavior.
I think it's good that we have the ability to have modal in modal in modal if we need, but It's safer to not assume too much on the initial loading as it can generate lot of bugs and create security issue.
When we load the modal in the modal, we keep the state of the previous modal.
What you're asking me is to recreate that state and hijack all the possible forms validation that the user should have gone through in the previous modal.

taoeffect (approved):
Works well enough!

PR #896, opened 2020-04-13
Implement sort of group members list

### LINKED ISSUE #869: Implement sort of group members list
### Problem

At the moment, the group members list, on the Dashboard sidebar as well as the "Group Members" page, are not sorted correctly.

### Solution
The sort should be, in both places:
1. Pending members (from A to Z);
2. New members (from A to Z);
3. All others (from A to Z).

sandrina-p:
Closes #869 

Screenshot 2020-04-13 at 15 00 43

#### Other improvements:
- Add Member Proposal: Fixed the "X" button to remove an invitee. It was out of place. 
- Search Members: Add missing _(you)_ when the member is ourselves.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -192,7 +192,7 @@ DefineContract({
           if (!userA.invitedBy && userB.invitedBy) { return 1 }
           // Then new members...
           if (userA.isNew && !userB.isNew) { return -1 }
-          if (!userA.isNew && userB.isNew) { return -1 }
+          if (!userA.isNew && userB.isNew) { return 1 }
           // and sort them all by A-Z
           if (nameA < nameB) { return -1 }
           if (nameA > nameB) { return 1 }
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -163,6 +163,42 @@ DefineContract({
       }
       return pendingMembers
     },
+    groupMembersSortedByTypeAndName (state, getters) {
+      const weJoinedMs = new Date(getters.currentGroupState.profiles[getters.ourUsername].joinedDate).getTime()
+      const isNewMember = (username) => {
+        if (username === getters.ourUsername) { return false }
+        const memberProfile = getters.currentGroupState.profiles[username]
+        if (!memberProfile) return false
+        const memberJoinedMs = new Date(memberProfile.joinedDate).getTime()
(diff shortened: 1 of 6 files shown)


frontend/model/contracts/group.js
```diff
@@ -163,6 +163,42 @@ DefineContract({
       }
       return pendingMembers
     },
+    groupMembersSortedByTypeAndName (state, getters) {
```
sandrina-p, line 166: NOTE: Advantages of creating this getter:
- It can be reused by across components (ex: `GroupMembers` and `GroupMembersAllModal`), making sure the displayed data is exactly the same in both views.
- The sort is calculated only once during the user visit, even if they visit the dashboard or members search multiple times during their session (major advantage compared to a component property that is recalculated in each component mount)

frontend/model/contracts/group.js
```diff
@@ -163,6 +163,42 @@ DefineContract({
       }
       return pendingMembers
     },
+    groupMembersSorted (state, getters) {
```
sandrina-p, line 166: Note: Advantages of creating this getter:

- It can be reused by across components (ex: `GroupMembers.vue` and `GroupMembersAllModal.vue`), making sure the displayed data is exactly the same in both views. DRY code and bugs avoided ✨
- The sort is calculated only once during the user visit, even if they visit the dashboard multiple times or do multiple members searches during their session (major advantage compared to a component property that is recalculated in each component mount)

mmbotelho (changes requested):
@sandrina-p I'm seeing this order:

Captura de ecrã 2020-04-14, às 11 29 29

The new ones are appearing at the bottom, as you can see.

sandrina-p:
Very strange @mmbotelho. I replicated your group and it works fine on my side. We'll need to schedule a video call to understand better what went wrong 🐛

image

mmbotelho:
Quick update: me and @sandrina-p had a quick call and we found that the culprit for the error is probably Firefox!

sandrina-p:
@mmbotelho I fixed it and tested locally on chrome, firefox and Safari. It seems all fine now! Please give it a try again

mmbotelho (approved):
It's working perfectly now @sandrina-p , good job!

pieer (approved):

taoeffect (approved):
Approved! Great job @sandrina-p!

PR #903, opened 2020-04-20
Close #898 Disable submit button on signup if the form is not valid

### LINKED ISSUE #898: "Create an Account" button should be disabled until form is complete
### Problem

[Figma designs](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=2737%3A154) show the "Log in" and "Create an Account" buttons as disabled when the form is empty. Currently, the "Create an Account" button is always enabled, even if the form is blank.

### Solution

Disable the "Create an Account" button until the form is complete, to match the designs, and to be consistent within the app.

pieer:
Disable submit button on signup if the form is not valid

### DIFF
--- a/frontend/views/containers/access/SignupForm.vue
+++ b/frontend/views/containers/access/SignupForm.vue
@@ -32,6 +32,7 @@ form(data-test='signup' @submit.prevent='')
     button-submit(
       @click='signup'
       data-test='signSubmit'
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Easy approve 👍

PR #905, opened 2020-04-23
Close #821: Make group description optional

### LINKED ISSUE #821: Make group description optional
### Problem

At the moment, the group description is a mandatory step. The problem with this is:
- It's not vital information;
- It can add friction when creating a group;
- Users might want to add a description in some other moment other than the group creation.

### Solution

- Make the group description step optional on the group creation flow;
- Add an empty state to the dashboard sidebar to prompt users to add a description, if they didn't on the group creation flow.

pieer:
Make group description optional


mmbotelho (changes requested):
Everything perfect except for the dashboard sidebar without description. It should look like this when there was no description added: 
Captura de ecrã 2020-04-23, às 16 13 03

When there is a description already, the styles are correct (the edit link). 👍

sandrina-p (approved):
Looks good!

frontend/views/components/group-creation-steps/GroupPurpose.vue
```diff
@@ -3,17 +3,20 @@
   i18n.is-title-4.steps-title(tag='h4') 2. Group Purpose
 
   .card
-    i18n.label(tag='label') How would you describe your group?
+    label.field
+      i18n.label(tag='label') How would you describe your group?
```
sandrina-p: Here it doesn't need the tag `label`. The parent is already a `<label>`.

sandrina-p (changes requested):
Ups, just saw this copy-paste mistake!

frontend/views/components/group-creation-steps/GroupPurpose.vue
```diff
@@ -3,17 +3,20 @@
...
+      textarea.textarea(
+        name='sharedValues'
+        ref='purpose'
+        :placeholder='L("Group Purpose")'
+        maxlength='500'
+        :class='{ error: $v.form.sharedValues.$error }'
+        :value='group.sharedValues'
+        @input='update'
+      )
+      i18n.helper This is optional.
```
sandrina-p, line 18: Should we add (or modify) a cypress test to create a group without description?
We already have a test to update the description.

pieer: A test to see if the group is created without a description? Sure, if you think that's usefull

sandrina-p: Yes, I believe it's a very important test, mainly because the contract was changed in this PR. (and creating a group is one of the main actions) If the change you made at `group.js` is reverted in the future, this e2e test will catch it. And also because our current "error message" in the group creation isn't very explicit.

pieer: Ok, will do a test to see if a group can be created without a description

mmbotelho (dismissed):
Perfect @pieer , thank you!

sandrina-p (dismissed):

taoeffect (changes requested):
Looks almost ready, only found one issue. When initially setting a description, you cannot clear it:

Screen Shot 2020-05-06 at 1 03 57 PM

But oddly enough it will still let you save the changes. That too is a bug. If the form is invalid it should not let me save the changes. So there are two bugs:

1. The field shouldn't be required anymore
2. If the field gives an error, the form shouldn't submit

And there's a third bug that I discovered a bug related to this: https://github.com/okTurtles/group-income-simple/issues/913

That third bug is on `master`, but since it's related to this you are welcome to fix it here too if you'd like, or save it for a future PR (up to you).

Finally, I don't think that the "Add +" button makes sense.

Screen Shot 2020-05-06 at 1 04 03 PM

It looks exactly the same as the "Add +" button for adding members, and I think that will confuse people. I preferred how it was before with the "Edit" link being where the group description is supposed to be. @mmbotelho are you OK with letting it be an "Edit" link?

mmbotelho:
@taoeffect it's supposed to have the same style as the add members, so the action is consistent. The "edit" link only makes sense if you already have a description and are editing it. I prefer leaving it as it is. Note that when there's a description, the edit link is displayed as before.

cc @pieer

mmbotelho (dismissed):

taoeffect:
> @taoeffect it's supposed to have the same style as the add members, so the action is consistent. The "edit" link only makes sense if you already have a description and are editing it. I prefer leaving it as it is. Note that when there's a description, the edit link is displayed as before.

I'm just voicing my objection. I find it very confusing to have those two buttons look the same to "Add" something. ("+ Add") is supposed to be to add a member, and yet in this case it means "add a description".

To me that is weird. One edits a description, not adds it. You can add something multiple times, as in add multiple units of something. But there is only one description to edit. I really think it should be restored to how it was.

Do you still disagree? Maybe we can have a vote on this if so?

sandrina-p:
> One edits a description, not adds it.

It adds a description because it does not exist yet. I'm okay with "+ Add", it makes sense to me.

taoeffect:
> It adds a description because it does not exist yet. I'm okay with "+ Add", it makes sense to me.

@dotmacro do you have a preference?

dotmacro:
I agree "+Add" is a little confusing. However, it's consistent with all similar instances of creating something that doesn't yet exist, e.g. adding a personal bio: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=9671%3A87438

taoeffect:
> I agree "+Add" is a little confusing. However, it's consistent with all similar instances of creating something that doesn't yet exist, e.g. adding a personal bio:

But there it's also clear as it says "Add a bio".

Here there are just two "Add +" buttons. The first one makes sense to me since it's next to "Members". But having the same exact "Add +" button just below the other one is very confusing to me.

@mmbotelho what do you think about changing it to a link that says "Add description"?

----

Unrelated to this, @pieer, I've restarted Travis several times now and it keeps failing, so there seems to be a real test that's failing.

pieer:
Weird the test are working locally..

mmbotelho:
@taoeffect There's not enough space to have "Add description" there. 

Maybe there's a misunderstanding here, but the only time the "Add" button is displayed is when there is no description. When there is a description, there's an edit link:
Captura de ecrã 2020-05-10, às 17 37 51


I think the best way to proceed here would be to continue as it is right now so @pieer can finally merge this and if you want you can open an issue to change it.

taoeffect:
> @taoeffect There's not enough space to have "Add description" there.

Not where the button is; I'm referring to where the "Edit" link was. There's plenty of room there.

> I think the best way to proceed here would be to continue as it is right now so @pieer can finally merge this and if you want you can open an issue to change it.

So, this is a trivial fix. It is less trivial to merge this knowing we're introducing a UI regression, open a new issue for it, describe the issue, open a new PR, test and merge the PR. We've been moving kind of fast lately, and this time we did not approve the design change before it was implemented.

This PR is ready to merge except for this one regression in the design. I would rather not knowingly introduce a new issue for us to deal with, so I propose two paths forward:

1. Restore it to how it was before, by removing the "Add +" button and having the "Edit" link as before, and then we can open a new issue to discuss alternatives that are not confusing.
2. Fix it now in this PR by changing the "Edit" link to "Add description" when there is no description.

Which would you like to do?

pieer:
How about we remove the "About the group" and only add a button Add group description?

sandrina-p:
> we did not approve the design change before it was implemented.

The design was given by @mmbotelho in February, everyone commented and nobody pointed out any issues.

> So, this is a trivial fix. 

Is it? If it was, we wouldn't be here discussing it for so long.

Here we are, losing time to figure out the position of a little button that is so `Priority:Low`, when we still have so many other more important things to decide. No wonder this prototype is taking forever to accomplish. I'll leave here a quote from PR #881 where we all clearly agreed with:

> We, as a team, will make an effort to avoid chaotic workflows in future tasks. When picking a new task/issue, let's make sure the UI/UX and technical requirements are clear as much as possible before moving to the development phase. **During the development, if something new needs to be added/changed, let's save it to another task to avoid chaotic decisions or bottlenecks in the process.**

mmbotelho:
I think there's a misunderstanding here @taoeffect 

> Restore it to how it was before, by removing the "Add +" button and having the "Edit" link as before, and then we can open a new issue to discuss alternatives that are not confusing.

There was no "before". Before there was no way for the description to be empty (because it was a mandatory field both in the group settings and group creation), so the only thing we had was an Edit (that is the same now). Because now it's possible to have no description, we added the "empty state" that has the Add button. 

I already voiced my opinion, and that is to keep the add button there. I think it's consistent with the rest of the app. Having 2 different patterns for the same action is weird and unnecessary and should be avoided.

If you want to change it to "Add group description" or even do as @pieer [suggested on his previous comment](https://github.com/okTurtles/group-income-simple/pull/905#issuecomment-626438723) however, feel free to, I'm ok with either solution.

taoeffect (approved):
Great job @pieer!

Issue #761, opened 2019-11-05
Create a recovery page


taoeffect:
### Problem

I have experienced very strange problems with the vuex state, possibly caused by accidentally clicking the wrong thing in the VueDevTools, where the vuex state is completely out of sync, and contract submodules are missing mutations. Messages like the following appear in the console:

```
[vuex] unknown mutation type: 21XWnNVY3JQGbGqNXyCwTXebDhKTRDYhmtN5Wh3mZeYsfbLChT/processMessage
```

### Solution

Part of the problem is that the error message above is a `console.error` message from vuex (when it should be an exception). I've already opened an issue in the Vuex repo for that. However, why that's happening to begin with is the real problem, and it's unclear that at this point catching the exception would fix anything, since the exception means that the vuex module hasn't been registered (even though it appears to be registered in the vuex dev tools!!).

For this particular error we can check for it by looking at the vuex internals themselves:

```js
Store.prototype.commit = function commit (_type, _payload, _options) {
    var this$1 = this;

  // check object-style commit
  var ref = unifyObjectStyle(_type, _payload, _options);
    var type = ref.type;
    var payload = ref.payload;
    var options = ref.options;

  var mutation = { type: type, payload: payload };
  var entry = this._mutations[type];
  if (!entry) {
    if (process.env.NODE_ENV !== 'production') {
      console.error(("[vuex] unknown mutation type: " + type));
    }
    return
  }
```

Here we can ourselves check make sure that `store._mutations[type]` exists, and if it doesn't, take steps to recover somehow (or take them to the recovery page).

The recovery page itself should have an option to unregister all of the vuex modules for every subscribed contract, and re-sync/re-register everything.

**Related issues**

- #432 - state machine

mmbotelho:
@taoeffect will this page be seen by users or is this just a tool for development purposes?

I'm wondering if this is the issue I should take on this week (the recovery page for when users encounter a critical error and need to recover from that), or if I should create a different one.

taoeffect:
@mmbotelho Yes users will see this page, they'll be automatically taken to it in certain situations.

mmbotelho:
@taoeffect I have a rough draft of what I believe needs to be on this page. Let me know what you think! https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=10216%3A0

I need some help with the copy there, I'm not sure what the right words are!

taoeffect:
[...]

PR #907, opened 2020-04-28
#761 Recovery page layout

sandrina-p:
Part of #761 

recovery-3

Implemented the layout with a dummy logic. The real logic will be done in a future PR by @taoeffect  

### Notes:
To see some specific scenario, the code needs to be modified.

- **App banner visibility:** Go to `frontend/main.js`, line 99. Change the key `isCorrupted` to true.

- **Troubleshooting initial status:** Go to `frontend/containers/user-settings/Troubleshooting.vue`. In line 80, change const `status` from `'ok'` to `'corrupted'`.
- **Recovery result:** Same file as before, line 124. Change the condition `false|true` to throw a dummy error and "break" the recovery process.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -129,7 +129,7 @@ async function startApp () {
       if (this.ephemeral.isCorrupted) {
         this.$refs.bannerGeneral.danger(
           this.L('Your app seems to be corrupted. Please {a_}re-sync your app data.{_a}', {
-            'a_': '<a class="link" href="/app/dashboard?modal=UserSettingsModal&section=troubleshooting">',
+            'a_': `<a class="link" href="${window.location.pathname}?modal=UserSettingsModal&section=troubleshooting">`,
             '_a': '</a>'
           }),
           'times-circle'
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -126,15 +126,15 @@ async function startApp () {
       // }, 2500)
 
       // #761 - Hardcoded banner
-      setTimeout(() => {
-        this.$refs.bannerGeneral.danger(
-          this.L('Your app seems to be corrupted and some functionality is unavailable. {a_}Click here to fix it.{_a}', {
-            'a_': '<a class="link" target="_blank" href="/app/dashboard?modal=UserSettingsModal&section=troubleshooting">',
-            '_a': '</a>'
-          }),
-          'times-circle'
-        )
-      }, 2500)
+      // setTimeout(() => {
+      //   this.$refs.bannerGeneral.danger(
+      //     this.L('Your app seems to be corrupted and some functionality is unavailable. {a_}Click here to fix it.{_a}', {
+      //       'a_': '<a class="link" target="_blank" href="/app/dashboard?modal=UserSettingsModal&section=troubleshooting">',
+      //       '_a': '</a>'
+      //     }),
+      //     'times-circle'
+      //   )
+      // }, 2500)
     },
     computed: {
       showNav () {
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -96,7 +96,7 @@ async function startApp () {
           syncs: [],
           // TODO/REVIEW page can load with already loggedin. -> this.$store.state.loggedIn ? 'yes' : 'no'
           finishedLogin: 'no',
-          isCorrupted: true // TODO #761
+          isCorrupted: false // TODO #761
         }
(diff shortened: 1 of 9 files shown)


mmbotelho (changes requested):
- Clicking on "re-sync your app data" should open on the same tab (current behavior is to open a new tab).

Other than that it's perfect!

sandrina-p:
> Clicking on "re-sync your app data" should open on the same tab (current behavior is to open a new tab).

@mmbotelho, that's a current technical limitation. :/ The same happens when we click "Report the error" in a scoped error message.
We did that because if we "open on the same tab" it will actually refresh the page and lose the current page context. For example, if it happens when trying to fill a formulary.

However, in this scenario, (re-sync data) it may be okay to lose the current page context because the App is already broken anyway, so it's not that important.

TLDR: Changed. It now opens on the same tab :)

mmbotelho (dismissed):
Approved, awesome work @sandrina-p !

frontend/main.js
```diff
@@ -95,7 +95,8 @@ async function startApp () {
         ephemeral: {
           syncs: [],
           // TODO/REVIEW page can load with already loggedin. -> this.$store.state.loggedIn ? 'yes' : 'no'
-          finishedLogin: 'no'
+          finishedLogin: 'no',
+          isCorrupted: true // TODO #761
```
taoeffect: Can't merge while this is set to true; although having it set to true has helped me find a bug! 😄

frontend/main.js
```diff
@@ -124,6 +125,16 @@ async function startApp () {
       // setTimeout(() => {
       //   this.$refs.bannerGeneral.show(this.L('Trying to reconnect...'), 'wifi')
       // }, 2500)
+
+      if (this.ephemeral.isCorrupted) {
+        this.$refs.bannerGeneral.danger(
+          this.L('Your app seems to be corrupted. Please {a_}re-sync your app data.{_a}', {
+            'a_': '<a class="link" href="/app/dashboard?modal=UserSettingsModal&section=troubleshooting">',
```
taoeffect: So... do we want to be able to access this page while we are not part of a group?

I think we might want to be able to do that... e.g. if the app is corrupted and therefore we can't access our group, it might make sense to be able to visit this as long as we're logged in.

That can be done either in this PR or in a future PR (in which case an issue is needed for that). Your choice!

taoeffect (changes requested):
Looks good! Almost ready to merge, just a few things:

- `isCorrupted` needs to be set back to false
- See comment about accessing recovery page while outside of a group... two ways to proceed there, up to you!
- When I clicked Re-Sync the progress bar looked weird/broken in Firefox Dev Edition:

Screen Shot 2020-05-06 at 4 37 14 PM

frontend/main.js
sandrina-p: Changed to false!

frontend/main.js
sandrina-p: Hum... nice question. I made it work on the last commit. Please take a look.

taoeffect: I think the file `router.js` needs to be updated... it just redirects back to welcome page to create a group. You can add a TODO for that or implement in this PR, your choice!

sandrina-p: Why do we need to update `router.js` exactly? I tested and it worked - it opens the recovery page.

Note that this is dumb: The URL is created on mount, so if we change pages, the URL is not updated (the pathname), and that's maybe why you are facing problems. This was just a dummy implementation of the banner link. IRL the banner would appear at the right moment, and there the URL would be correct.

taoeffect: > Why do we need to update router.js exactly? I tested and it worked - it opens the recovery page.

... weird. Doesn't work for me 🤷‍♂️ 

I copy/pasted the URL and manually loaded it. Was still redirected to the welcome page where it asks you to create a group. I'm guessing it's one of the guards that's causing it. Did you try in Firefox?

sandrina-p: When you open a page with no group and click the recovery link,
it should start with `/app/?modal=`.

When you open a tab with group, and click the recovery link,
it should start with `/app/dashboard/modal=`

taoeffect: Ah! You're right I must've been copying the wrong link! Will approve now!

taoeffect (approved):
Fantastic! 🎉

Issue #893, opened 2020-04-12
What happens when the disagreement number set by the group is larger than the group itself? [Discussion]


mmbotelho:
### Problem

As @taoeffect said on Slack:
> Eg: you set it (disagreement number) to 5, your group has shrunk to 4 members. Impossible to prevent proposals from passing. Everyone could say no and it will still pass.

This is a problem when:
- Groups increase in size;
- Groups shrink;
- The group creator sets a very high number on the group creation flow and the first proposal created in the group has a disagreement number higher than the group size;
- Someone creates a proposal to change the disagreement number to something higher than the group size.

### Solution

There are a number of possible solutions for this problem:
- Proposals fail if either the absolute threshold you set is lower than the group size OR 50% of the group has voted no;
- Prevent users from creating proposals when the disagreement number they set is higher than the total group size and prompting them to create a proposal to change voting system OR the disagreement number;
- Even if users set a disagreement number of 30, and their group has 20 people, we automatically adjust it to 19 (which is closer to their initial intent);
- Leave it as-is and let users assume responsibility for whatever number they set there;

These are the solutions I can think of at the top of my head, let me know if you can think of something else!

sandrina-p:
My vote goes to:
- Option 5: Change disagreement rule to be percentage based, avoiding this tricky scenario or any other possible bugs/misunderstandings/edge-cases.

taoeffect:
> revent users from creating proposals when the disagreement number they set is higher than the total group size and prompting them to create a proposal to change voting system OR the disagreement number;

This situation would result in a Catch-22, as the proposal won't be able to pass, but you need to make it to change the threshold.

> Even if users set a disagreement number of 30, and their group has 20 people, we automatically adjust it to 19 (which is closer to their initial intent);

The code is not designed in a way to automatically adjust the threshold like that. We would have to add special threshold-checking code anytime someone leaves, and that has "code smell".

> Leave it as-is and let users assume responsibility for whatever number they set there;

We can't leave users stuck in a hopeless situation, that's poor UX.

> Proposals fail if either the absolute threshold you set is lower than the group size OR 50% of the group has voted no;

This (slightly worded differently) is the solution I think makes the most amount of sense to me, and is based around [our first blog post](https://groupincome.org/2016/06/what-makes-a-good-voting-system/) on what makes a good voting system.

![](https://groupincome.org/wp-content/uploads/2016/06/voting-v6.2.png)

[...]

PR #909, opened 2020-04-30
#893 Rule Disagreement Layout

### LINKED ISSUE #897: Rename RULE_THRESHOLD to RULE_PERCENTAGE
### Problem

The word "threshold" is currently used in both `RULE_THRESHOLD` and `RULE_DISAGREEMENT` to mean to completely different things. This causes confusion because we also have a voting rule called `RULE_THRESHOLD`.

### Solution

Rename `RULE_THRESHOLD` to `RULE_PERCENTAGE` to get rid of the confusion.

### LINKED ISSUE #908: Verify color contrast ratio to comply Level AA
### Problem

The small grey text of light theme does not comply with [WCAG 2.0 Level AA contrast ratio](https://webaim.org/resources/contrastchecker/).

image

### Solution

- Improve the contrast ratio to comply with Level AA.
- Double-check all the other texts comply with Level AA.

sandrina-p:
Related to #893, closes #897, closes #908

image

Implemented all the layouts (3) related to Voting System, but with a dummy logic. The real logic will be done on a future PR. Here're some notes about each layout:

### 1. Group Creation: Voting Step
- Whatever option is chosen, it will not affect the app. It's just a dummy layout, It doesn't work.

### 2. Group Settings: Voting Card
- By default, the active voting system type is "Percentage based". If you want to see the other option, go to `containers/group-settings/GroupVotingSystem`, line 62. Change the `Array` order.
- Click the link to change the rule, to see the proposal modal.

### 3. Change Voting Rules/Setttings
- The title and footer copy are incorrect because it directly depends on the app logic. Will be done on the next PR.

The component `SliderContinuous` was added to the Design System page.


frontend/views/components/slider-circular/components/CircleSlider.vue
```diff
@@ -1,331 +0,0 @@
-<template lang='pug'>
```
taoeffect, line 1: Hmm, instead of deleting all of these potentially useful components, making it difficult to find them say 1 year from now, can you move all these deleted files into `historical/` in case we decide we'd like to bring them back? It will be much more time consuming to find them in the git history.

frontend/views/containers/group-settings/GroupCreationModal.vue
```diff
@@ -132,6 +132,11 @@ export default {
         groupName: '',
         groupPicture: '',
         sharedValues: null,
+
+        // randomize to reduce choice bias
+        ruleOrder: Math.round(Math.random()) === 1 ? ['threshold', 'disagreement'] : ['disagreement', 'threshold'],
```
taoeffect: Can you use the constants `RULE_DISAGREEMENT` and `RULE_THRESHOLD` here instead? (There are other places this needs to be done, but I won't point them out per your request ^_^)

Also, can you rename `RULE_THRESHOLD` to `RULE_PERCENTAGE` throughout the project so that we can also close #897?

frontend/views/containers/group-settings/GroupVotingSystem.vue
```diff
@@ -0,0 +1,133 @@
+<template lang='pug'>
+  ul.c-wrapper
+    li.cardBox(
+      v-for='option in votingSytemSorted'
+      :class='{ isActive: isVotingActive(option) }'
+      :aria-current='isVotingActive(option) ? "true" : "false"'
```
taoeffect: What's this `aria-current` thing and why is it `"true"` instead of `true`?

Can you add a comment explaining this above line 3? (Or a link to an explainer?)

frontend/views/containers/group-settings/GroupVotingSystem.vue
```diff
@@ -0,0 +1,133 @@
...
+    )
+      span.pill.is-primary.c-active(v-if='isVotingActive(option)') Active
```
taoeffect: i18n

frontend/views/containers/group-settings/GroupVotingSystem.vue
```diff
@@ -0,0 +1,133 @@
...
+
+      return option === 'disagreement' // TODO this
+    },
+    votingValue (option) {
+      const HTMLBold = (txt) => `<span class="has-text-bold">${txt}</span>`
+      const HTMLSmall = (txt) => ` <span class="has-text-1 has-text-small">(${txt})</span>`
+
+      if (option === 'threshold') {
+        const percent = '75%'
+        return `${HTMLBold(percent)} ${HTMLSmall(L('{count} out of {total} members', {
```
taoeffect: Here can you please use the standard way of constructing strings? The reason is the positioning can change depending on the language, and it's also important to have consistency in i18n/L usage.

So instead of using `HTMLBold`/`HTMLSmall`, use `{b_}` and `{_b}` as before. You don't need to use `LTags` for that.

e.g.

```js
L('{b_}{percent_value}{_b} {sm_}{count} out of {total} members{_sm}', {
  b_: '<span class="has-text-bold">',
  _b: '</span>',
  // etc.
})
```

taoeffect (changes requested):
Partial quick first draft review! Still more to come and need to test it locally, but getting this out to you now so that you have something to work with. Great job!

frontend/views/containers/proposals/ChangeVotingProposal.vue
```diff
@@ -0,0 +1,89 @@
...
+        currentStep: 0
+      }
+    }
+  },
+  created () {
+    this.form.value = this.type === 'threshold' ? 75 : 2 // TODO this.
+  },
+  computed: {
+    title () {
+      return L('Chage voting [xxxx]') // TODO this
```
pieer: *change

pieer (dismissed):
Great work @sandrina-p !

frontend/views/components/slider-circular/components/CircleSlider.vue
sandrina-p: If we decide to do a UI similar to that in the future, I hope they do it from scratch because the current ones are not very usable or accessible.

Why do you think it is time-consuming to find them on git history? We have a very clean git history. Everything is documented. Moving old files to a `historical/` folder goes against the whole concept of git and only creates noise in the repo. 

Maybe a middle step to avoid redundancy: Have a single file `historical.md` or a github issue "Code Archive" with permalinks pointing to the respective file. For example:

##### Stale components
- [CircleSlider](https://github.com/okTurtles/group-income-simple/blob/028c9c8ee6d7531f73f0b0fd5d395b9caa8971ce/frontend/views/components/slider-circular/components/CircleSlider.vue)

frontend/views/containers/group-settings/GroupVotingSystem.vue
sandrina-p: Here's [an article about aria-current](https://tink.uk/using-the-aria-current-attribute/). It's used to mark an element in a list as the "current/active" one. The same way we show it visually (with the ACTIVE pill), we should also say it programmatically.

Regarding `true/"true"` in this case does not make a difference.
Code simplified to `:aria-current='isVotingActive(option)'`.

mmbotelho (changes requested):
@sandrina-p here's what I found so far:

- The maximum number of "no" votes should start at 1 and not 0;

- The main content in the page moves when the warning message appears:
3zk38c

- The spacing inside the group creation cards is a bit off (group creation):

Captura de ecrã 2020-05-01, às 12 18 10

- The group creation header has the wrong styles in small screens:

Captura de ecrã 2020-05-01, às 12 22 46

- Group Settings, the invite links management section's table is a bit broken on smaller screens:

Captura de ecrã 2020-05-01, às 12 25 29

- Group Settings, smaller screens - the line breaks were designed a bit differently. There also seems to be more space than necessary at the bottom of the selectable cards:
Captura de ecrã 2020-05-01, às 12 25 54

- Group Settings, spacings:

Captura de ecrã 2020-05-01, às 12 48 21


Note: There are things here that I'm not 100% related to this PR, I'll leave it up to you if you fix them now or to open new issues. 

Thank you!

frontend/views/components/VotingSystemInput.vue
```diff
@@ -0,0 +1,95 @@
...
+      div(v-if='warnMajority') <!-- [1] -->
+        banner-simple.c-banner(severity='warning')
+          i18n The percentage value you are choosing is most likely too low for a decision that can have a potentially significant impact  on a person's life. Please consider using a
+          | &nbsp;
+          i18n(
+            tag='a'
+            class='link'
+            href='https://groupincome.org/2016/09/deprecating-mays-theorem/#when-majority-rule-can-harm'
+            target='_blank'
+          ) supermajority threshold.
```
taoeffect: Sentences must never be split like this, it makes translation impossible.

Please use `{a_} ... {_a}` instead

frontend/views/components/slider-circular/components/CircleSlider.vue
taoeffect: OK, that works, feel free to create that file! 👍

frontend/views/components/VotingSystemInput.vue
sandrina-p: True. The way it was before it made sense because it was 2 different sentences. Fixed.

sandrina-p:
@mmbotelho I've fixed all the UI issues you found, including the ones not related to this PR!

--- EDIT ---

Actually no, there's was one: "The main content in the page moves when the warning message appears:" that I couldn't replicate. Please confirm if it's still there after these new changes. If yes, I think we'll have to jump into a video call. Which browser are you using?

mmbotelho:
@sandrina-p FYI, I made the font bigger on the group creation! Here are the changes I made:

Captura de ecrã 2020-05-01, às 17 00 06

[All the mockups have been updated on Figma as well](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=11127%3A100535)

mmbotelho (changes requested):
Everything I requested is working great now. Thank you for fixing things that were not related to the PR as well! 

I did notice that the slider is behaving in an unpredictable way. Sometimes it's orange but does not display the warning, other times it's blue but the warning is visible. This happens at 60+% range.

Captura de ecrã 2020-05-02, às 17 31 38
Captura de ecrã 2020-05-02, às 17 29 09

As for the other weird "skipping" behavior, it's not happening anymore!

sandrina-p:
@mmbotelho nice catch, it's fixed now!
I also added a small animation on slider thumb `:hover`. Let me know if you like it, otherwise, I can remove it.

mmbotelho (dismissed):
It's working perfectly now, thank you @sandrina-p !

frontend/controller/actions/group.js
```diff
@@ -86,7 +86,7 @@ export default sbp('sbp/selectors/register', {
       return groupId
     } catch (e) {
       console.error('gi.actions/group/create failed!', e)
-      throw new GIErrorUIRuntimeError(L('Failed to create the group: {codeError}', { codeError: e.message }))
+      throw new GIErrorUIRuntimeError(L('Failed to create the group: {reportError}', LError(e)))
```
taoeffect, line 89: Nice catch 👍

frontend/views/components/VotingSystemInput.vue
```diff
@@ -0,0 +1,94 @@
...
+      :value='value'
+      @input='handleInput'
+    )
+    transition-expand
+      // inner el has paddings. A wrap is needed for a smooth transition.
+      div(v-if='warnMajority')
+        banner-simple.c-banner(severity='warning')
+          i18n(
+            :args='{ a_:`<a class="link" href="https://groupincome.org/2016/09/deprecating-mays-theorem/#when-majority-rule-can-harm" target="_blank">`, _a: "</a>" }'
+          ) The percentage value you are choosing is most likely too low for a decision that can have a potentially significant impact  on a person's life. Please consider using a {a_}supermajority threshold{_a}.
```
taoeffect: Two spaces after `impact` -> one space

frontend/views/components/VotingSystemInput.vue
```diff
@@ -0,0 +1,94 @@
...
+  data: () => ({
+    config: {
+      [RULE_PERCENTAGE]: {
+        slideLabel: L('What percentage of members need to agree?'),
+        slideMin: 0,
+        slideMax: 100,
+        slideUnit: '%'
+      },
+      [RULE_DISAGREEMENT]: {
+        slideLabel: L('Maximum number of “no” votes'),
```
taoeffect: Since it's potentially easy for a developer to accidentally change these angle-quotes from `“` to `"`, resulting in a broken translation, can you change them to regular quotes (`"`) here? (There's at least one other place in this PR where the same thing happens).

frontend/views/components/group-creation-steps/GroupRules.vue
```diff
@@ -96,13 +37,36 @@ export default {
...
+  data: () => ({
+    config: {
+      [RULE_PERCENTAGE]: {
+        optionLabel: L('Percentage based'),
+        optionHint: L('Define the percentage of members who will need to agree to a proposal.', LTags('b')),
+        slideDefault: 75 // TODO connect to store.
+      },
+      [RULE_DISAGREEMENT]: {
+        optionLabel: L('Disagreement number'),
+        optionHint: L('Define the number of people required to block a proposal.', LTags('b')),
```
taoeffect: Here and above `{b}` is not used, so shouldn't be included.

frontend/views/containers/group-settings/GroupVotingSystem.vue
```diff
@@ -0,0 +1,151 @@
...
+            tag='button'
+            @click='openVotingProposal(type)'
+          ) Propose change
+
+      banner-simple.c-banner(
+        severity='info'
+        v-if='isVotingRuleAdjusted(type)'
+      )
+        | * &nbsp;
+        i18n This value was automatically adjusted because your group is currently smaller than the disagreement number.
```
taoeffect: I don't know if other languages use a different symbol than `*` to indicate this. They might. They might also position it differently. So I think the `*` needs to be part of the `i18n`.

frontend/views/containers/group-settings/GroupVotingSystem.vue
```diff
@@ -0,0 +1,151 @@
...
+    votingSytemSorted () {
+      return [RULE_DISAGREEMENT, RULE_PERCENTAGE] // TODO sort order based on groupSettings.
+    }
+  },
+  methods: {
+    openVotingProposal (type) {
+      sbp('okTurtles.events/emit', OPEN_MODAL, 'ChangeVotingProposal', { type })
+    },
+    isVotingActive (option) {
+      return option === this.votingSytemSorted[0]
```
taoeffect, line 76: This should be based on the value of `store.getters.groupSettings.proposals[PROPOSAL_GROUP_SETTING_CHANGE].rule`.

Note: technically it can be based on the value of `store.getters.groupSettings.proposals[PROPOSAL_INVITE_MEMBER].rule` or `store.getters.groupSettings.proposals[PROPOSAL_REMOVE_MEMBER].rule` or any of the other proposal types, since we've decided that we are making it the same voting rule for all proposal types.

Since that's an easy change you can do it for this PR, or if you prefer you can add this as a TODO comment here.

taoeffect (changes requested):
Looks good! Only minor things noticed.

frontend/views/components/SliderContinuous.vue
```diff
@@ -0,0 +1,150 @@
...
+    /** Unique id to connect form and label */
+    uid: {
+      type: String,
+      required: true
+    },
+    label: String,
+    min: [String, Number],
+    max: [String, Number],
+    unit: String,
+    value: [String, Number]
```
taoeffect, line 38: Just curious, does it make sense for min/max/value to be either a String or a Numbers?

Math is performed on them directly below, so I think they should be numbers?

sandrina-p: Technically for HTML, every attribute is a String, not a Number.

We could pass as a number, but HTML would interpret it and return as a String. So this go-and-back will return a prop type error eventually. So we need to accept both types.

frontend/views/containers/group-settings/GroupVotingSystem.vue
sandrina-p: This PR does not contain any logic, even that simple thing. It will be done on a next PR :)

frontend/views/components/group-creation-steps/GroupRules.vue
sandrina-p: Removed

frontend/views/components/SliderContinuous.vue
taoeffect: @sandrina-p for Vue.js to pass numbers you do `:value='3'` (note the `:`). So can you change it to do it that way?

frontend/views/containers/group-settings/GroupVotingSystem.vue
taoeffect: OK but just to note forget can you add a TODO comment to this effect? You can copy/paste what I wrote above if you want.

frontend/views/components/SliderContinuous.vue
sandrina-p: Maybe I didn't explain myself well. You can pass as a Number, but HTML will retrieve a String. `@input(e => typeof e.target.value) // String`

So, that's why we need both. So that when the prop goes up to the parent and is updated back to the child, it can be a String (default) or a Number (if manipulated).

taoeffect: *sigh*... ok, sorry, didn't understand. OK thanks!

frontend/views/containers/group-settings/GroupVotingSystem.vue
sandrina-p: Sure... done

taoeffect (approved):
Approved! Fantastic work @sandrina-p, the new rules look great! 😄

PR #912, opened 2020-05-06
#910 Support multiple similar payment methods

### LINKED ISSUE #910: Cannot add more than one similar payment method.
### Problem

Go to "Income Details", add multiple "Other" payments and click save.
Screenshot 2020-05-02 at 11 35 35

Re-open Income Details, only the last "Other" payment was saved.

Screenshot 2020-05-02 at 11 36 19

### Solution

Allow similar payment methods to be stored.

Currently, the payments info are saved in an object where:

```js
{
   other: {
     value: 'mbway: 912345678'
  }
}
```

Because an object cannot have multiple similar keys (e.g. 3 "other"), only the last payment key will be stored.

The payments must be saved in an array:

```js
[
  {
     name: 'Other',
     value: 'mbway: 912345678'
  }
]
```

sandrina-p:
Fixes #910

Note: When PR #884 is merged (after or before this PR), the `ProfileCard.vue` needs to be updated to fetch correctly `ourGroupProfile.paymentMethods`.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -541,12 +541,12 @@ DefineContract({
           with: string
         }),
         nonMonetaryRemove: string,
-        paymentMethods: optional(arrayOf(
+        paymentMethods: arrayOf(
           objectOf({
             name: string,
             value: string
           })
-        ))
+        )
       }),
       process ({ data, meta }, { state }) {
         var groupProfile = state.profiles[meta.username]
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -3,7 +3,7 @@
 import sbp from '~/shared/sbp.js'
 import Vue from 'vue'
 import { DefineContract } from './Contract.js'
-import { mapOf, objectOf, objectMaybeOf, optional, string, number, object, unionOf } from '~/frontend/utils/flowTyper.js'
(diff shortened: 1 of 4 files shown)


frontend/model/contracts/group.js
```diff
@@ -536,17 +536,17 @@ DefineContract({
...
-          venmo: objectMaybeOf({ value: string }),
-          other: objectMaybeOf({ value: string })
-        }),
         nonMonetaryEdit: objectOf({
           replace: string,
           with: string
         }),
-        nonMonetaryRemove: string
+        nonMonetaryRemove: string,
+        paymentMethods: optional(arrayOf(
```
taoeffect: Since we're already inside of an `objectMaybeOf` the `optional` isn't needed here.

taoeffect (changes requested):
Excellent! Only one issue (very minor) noticed!

taoeffect (dismissed):
Approved! Excellent work! 😄

frontend/model/contracts/group.js
sandrina-p: Before it had `objectMaybeOf`. That's why I kept the `optional` thing. Removed it.

taoeffect (approved):
Approved again since my previous review was cleared and tests now pass!

Issue #878, opened 2020-04-01
UI doesn't support VOTE_INDIFFERENT


taoeffect:
### Problem

Currently you can only vote "yes" or "no" on a proposal, but cannot indicate that you don't care. A non-vote is not the same as a vote of not caring. Our `RULE_THRESHOLD` supports the math for handling `VOTE_INDIFFERENT`, but our UI does not.

### Solution

Implement in UI a way to indicate indifference about the outcome of a proposal.

mmbotelho:
@taoeffect isn't this the same as issue #596 ?

taoeffect:
@mmbotelho yep, you're right, it's effectively the same but focuses more on some details that #596 left out. In truth I totally forgot that was an existing issue. I'll close this but copy its text over.

Issue #877, opened 2020-04-01
Bug 2 in RULE_DISAGREEMENT


taoeffect:
### Problem

`RULE_THRESHOLD` has support for `VOTE_INDIFFERENT` votes, but `RULE_DISAGREEMENT` doesn't.

### Solution

Support `VOTE_INDIFFERENT` in `RULE_DISAGREEMENT`.

taoeffect:
Actually, I don't think it makes sense for `RULE_DISAGREEMENT` to support "don't care" votes, since it goes by absolute value rather than relative value. So this isn't a bug I don't think.

taoeffect:
@dotmacro brought up a good point:

> er, I think RULE_DISAGREEMENT should support indifferent votes. For example, (using the same "2 against" group as before) if 8/10 people vote "yes" asap and the other two don't care, then the proposal will sit unchanged until expirey waiting for those last two votes in case they're AGAINST. Alternatively, if they can vote INDIFERENT, then the proposal can pass asap. The point of INDIFFERENT is to get things passed/failed asap, non?

So, re-opening.

Issue #876, opened 2020-04-01
Bug 1 in RULE_DISAGREEMENT


taoeffect:
### Problem

```js
    if (totalAgainst >= threshold) {
      return VOTE_AGAINST
    }
    // consider proposal passed if more vote for it than against it and there aren't
    // enough votes left to tip the scales past the threshold
    return totalAgainst + absent < threshold && totalFor > totalAgainst ? VOTE_FOR : VOTE_UNDECIDED
```

In a group of 10 people, with a threshold of 2, if 8 people vote for a proposal, and 0 vote against it, then the proposal should pass upon expiry.

However, in the way it's currently written (shown above), it will remain `VOTE_UNDECIDED`, which equates to it not passing.

### Solution

Fix.

EDIT: but make sure there's at least one `VOTE_FOR` vote before passing. If there are 0 votes for or against, the result should be `VOTE_UNDECIDED`.

sandrina-p:
Fixed at #893

Issue #620, opened 2019-08-15
Implement invite expiry


taoeffect:
### Problem

The proposals PR (#611) creates a basic invite system, but doesn't fully polish it (#609), and one of the things that's missing is the `expires` field in `generateInvites`. Plus that function is just kinda hanging out by itself, and should probably be put under SBP.

### Solution

Fully implement invite expiry, including:

- Default expiry time
- ~~The ability for a group admin to change the default expiry time (and a proposal-based approach to change it without an admin)~~ EDIT: this is low priority and should be done later and in a separate issue.
- Displaying the fact that invites are expired both to existing users and to user who are trying to use that invite to join

@mmbotelho edit:
Remove any "pending" members from the members list (both on the dashboard sidebar and group members page) when the invite link expires.

mmbotelho:
@taoeffect I added one more task to this issue, hope that's ok!

taoeffect:
Yeah thanks for the heads up! 👍

dotmacro:
I acknowledge this hasn't been implemented yet.

Currently, links claim to expire in the past: 
![image](https://user-images.githubusercontent.com/13082264/144775483-3b2287dd-b58d-4b09-b6bc-e23d8ef57ad2.png)

Silver-IT:
Additionally, we need to use a different SVG from the one when it is invalid.

Silver-IT:
> The ability for a group admin to change the default expiry time (and a proposal-based approach to change it without an admin)
> Displaying the fact that invites are expired both to existing users

@leihla , Can you please take care of this design?

leihla:
@Silver-IT, yes I can take care of this, just assigned it do me and will assign back to you once is done

Silver-IT:
> The ability for a group admin to change the default expiry time (and a proposal-based approach to change it without an admin)

Just note that we are going to do this later when it is necessary.

taoeffect:
Yeah I've re-assigned to you @Silver-IT and edited it to cross it off the list, to note that it's a low-priority task right now:

> ~~The ability for a group admin to change the default expiry time (and a proposal-based approach to change it without an admin)~~ EDIT: this is low priority and should be done later and in a separate issue.

Silver-IT:
> Displaying the fact that invites are expired to existing users

@leihla , I don't think there is design for this. I guess there should be something here...
![image](https://user-images.githubusercontent.com/57976479/174201127-badd9563-7817-491a-9443-f79f8c947e60.png)

[...]

PR #916, opened 2020-05-08
#893 Voting System - Add "Disagreement" rule

### LINKED ISSUE #893: What happens when the disagreement number set by the group is larger than the group itself? [Discussion]

sandrina-p:
Closes #893

### What's done:
- A group can be created with a custom Voting System: _disagreement_ or _percentage_.
- The Voting System can be changed at Group Settings (directly or through a proposal)
- Cypress tests to make sure the Voting System is correct (in 6 different scenarios)
- Unit tests to the rule system. - Using Cypress to cover every single proposal scenario would increase the CI tests' time a lot. For that reason, I used unit tests instead for corner cases.

### How to test this:
There are a lot of possible scenarios based on the voting system config.

You can create multiple groups with different voting systems or create a single group and then change the voting system at the "Group Settings" page.
Here's a screenshot of how the Group Settings card looks like in 8 different scenarios:
**Note:** the last scenario is incorrect. it should say _adjusted to 4_, because the group has 5 members and the limit is `groupSize - 1`. The code was updated taking into account that.
ss

### Notes
- We have been naming this "Voting System", but the core code uses "Voting Rules". I changed all the Vue files names and props to follow the code "rule" naming.
- Some **layout decisions are still being discussed at Figma**. Once solved, I'll update this PR as well.


frontend/controller/actions/group.js
```diff
@@ -53,25 +61,29 @@ export default sbp('sbp/selectors/register', {
...
+            [PROPOSAL_INVITE_MEMBER]: merge(
+              merge({}, proposals[PROPOSAL_INVITE_MEMBER].defaults),
+              proposalSettings
             ),
-            [PROPOSAL_INVITE_MEMBER]: merge({},
-              proposals[PROPOSAL_INVITE_MEMBER].defaults,
-              { ruleSettings: { [RULE_PERCENTAGE]: { threshold: thresholdMemberApproval } } }
+            [PROPOSAL_REMOVE_MEMBER]: merge(
+              merge({}, proposals[PROPOSAL_REMOVE_MEMBER].defaults),
+              proposalSettings
```
sandrina-p, line 73: Note: `merge()` wasn't working as expected. It only accepts 2 parameters. So, we need to use `merge` twice: first from default settings to an empty object and then from (custom) settings to default settings.

frontend/model/contracts/voting/proposals.js
```diff
@@ -113,14 +117,10 @@ const proposals = {
...
-      ruleSettings: {
-        [RULE_PERCENTAGE]: { threshold: 0.8 },
-        // at least 2, since the member being removed shouldn't be able to
-        // block the proposal themselves
-        [RULE_DISAGREEMENT]: { threshold: 2 }
-      }
+      // REVIEW PR:
+      // - Before was 14 days instead of 7. Any major reason?
+      // - the default threshold was 2. Maybe we need some manual "adjustement" when running the rules.js?
+      ...proposalDefaults
```
sandrina-p: Clarify the code comment above ☝️

frontend/views/containers/group-settings/GroupRulesSettings.vue
```diff
@@ -0,0 +1,160 @@
...
+<script>
+import sbp from '~/shared/sbp.js'
+import { mapGetters } from 'vuex'
+import { RULE_PERCENTAGE, RULE_DISAGREEMENT } from '@model/contracts/voting/rules.js'
+import { OPEN_MODAL } from '@utils/events.js'
+import L from '@view-utils/translations.js'
+import BannerSimple from '@components/banners/BannerSimple.vue'
+
+export default {
+  name: 'GroupVotingSystem',
```
sandrina-p, line 45: Note: This file was not created, it was renamed: GroupVotingSystem -> GroupRulesSettings.

And added the needed logic to connect to the state. The markup is pretty much the same and the CSS wasn't modified at all.

frontend/model/contracts/voting/proposals.js
```diff
@@ -59,16 +59,20 @@ function voteAgainst (state, { meta, data, contractID }) {
   sbp('okTurtles.events/emit', PROPOSAL_RESULT, state, VOTE_AGAINST, data)
 }
 
+// NOTE: The code is ready to receive different proposals settings,
+// However, we decided to make all settings the same, to simplify the UI/UX proptotype.
+const proposalDefaults = {
+  rule: RULE_PERCENTAGE,
+  expires_ms: 7 * DAYS_MILLIS,
```
sandrina-p, line 66: I noticed that the proposals had different expire days.
- Add/Remove Member proposals expire in 14 days. 
- Group Setting and Proposal Setting expire in 7 days.

Any reason for that? Maybe we could also make it all the same to simplify.

cc @mmbotelho

sandrina-p: I noticed that the proposals have different expire days:
- Add/Remove Member proposals expire in 14 days. 
- Group Setting and Proposal Setting expire in 7 days.

Any reason for that? I suggest making them all the same, to simplify UI/UX as we did with the threshold.

cc @mmbotelho

mmbotelho: If we do, I would suggest keeping the upper limit of 14 days to give people more time!

sandrina-p: I'm okay with that.
Should we have a UI to change that setting too, or should we keep it "hidden" for now?

mmbotelho: I believe this is one of those things that we can set ourselves. We will find out if this timeframe works when the prototype is out and we can tweak it! I think adding a whole proposal to change this setting would be an overkill

mmbotelho: We probably need, however, to add a small text label indicating that a proposal is about to expire on the dashboard. Let me know what you think @sandrina-p @taoeffect and I can create an issue for it

sandrina-p: I like that idea, at least is informative. Later in the road, if requested, we can create the UI/logic to update that info.

taoeffect (commented):
This one is going to take me a few days to review, just a heads up

strings/french.strings
```diff
@@ -1101,6 +1106,9 @@
 /* MISSING TRANSLATION - views/pages/Home.vue */
 "Welcome to group income" = "Welcome to group income";
 
+/* MISSING TRANSLATION - views/containers/proposals/RuleAddMember.vue, views/containers/proposals/ChangeVotingRules.vue, views/containers/proposals/RuleRemoveMember.vue */
+"The reason why I' propositiong this change is..." = "The reason why I' propositiong this change is...";
```
taoeffect: Where are these strings coming from? I feel like they might be being re-introduced from an old commit. Make sure to merge `master`. If you merge `master` and they're still here, then run `./strings.exe frontend/` to regenerate them.

package-lock.json
```diff
@@ -11291,7 +11291,7 @@
             },
             "mkdirp": {
               "version": "0.5.1",
-              "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.1.tgz",
+              "resolved": false,
```
taoeffect: This file shouldn't be modified. Please make sure you're using the one from `master`.

frontend/views/utils/currencies.js
```diff
@@ -17,13 +17,15 @@ export function saferFloat (value: number): number {
 //       having to "recompile" a new version of the app.
 const currencies = {
   USD: {
+    code: 'USD',
```
taoeffect: The code is the key itself... so I don't think this should be added since it creates a redundancy that then needs to be kept in sync (and introduces bugs if it's out of sync). In other words, if you're already typing `.USD`, you don't need to type `.USD.code`, just write `'USD'`.

frontend/model/contracts/voting/rules.js
```diff
@@ -73,3 +73,10 @@ const rules = {
 export default rules
 
 export const ruleType = unionOf(...Object.keys(rules).map(k => literalOf(k)))
+
+export const thresholdAdjusted = (rule, threshold, groupSize) => {
+  if (rule === RULE_DISAGREEMENT) {
+    return { thresholdAdjusted: Math.min(threshold, groupSize) }
+  }
+  return {}
+}
```
taoeffect: Instead of creating this function, the adjustment should be done inside of the `RULE_DISAGREEMENT` function. All adjustments are done inside of there, not outside of it. For example, both `RULE_PERCENTAGE` and `RULE_DISAGREEMENT` make adjustments of the threshold for `PROPOSAL_REMOVE_MEMBER`.

frontend/controller/actions/group.js
```diff
@@ -44,6 +43,15 @@ export default sbp('sbp/selectors/register', {
 
     try {
       const initialInvite = createInvite({ quantity: 60, creator: INVITE_INITIAL_CREATOR })
+      const proposalSettings = {
+        rule: ruleName,
+        ruleSettings: {
+          [ruleName]: {
+            ...thresholdAdjusted(ruleName, +ruleThreshold, 1), // 1 = group initial size.
```
taoeffect: Threshold should not be adjusted here or with a `thresholdAdjusted` function, see comment below for details.

frontend/views/containers/group-settings/GroupRulesSettings.vue
```diff
@@ -0,0 +1,162 @@
...
+        const percent = this.votingRuleSettings.threshold * 100 + '%'
+
+        // TODO/BUG/REVIEW: in a 1 member group with 49%, it says: 0 out of 1 members. Is it rights?
+        return L('{b_}{percent}{_b} {sm_}({count} out of {total} members){_sm}', {
+          percent,
+          count: Math.round(this.groupMembersCount * this.votingRuleSettings.threshold),
+          total: this.groupMembersCount,
+          ...HTMLTags
+        })
+      } else {
```
taoeffect: `else if`, always check the rule type explicitly, don't assume there will only be two voting rule types since that's unlikely to be the case. The `else` case should show an error telling the user that they're probably using outdated software.

taoeffect (changes requested):
Great work @sandrina-p! Posting a quick first draft review (unfinished) with some quick feedback so that you have something to work with tomorrow while I continue to review this and #918.

frontend/model/contracts/group.js
```diff
@@ -299,6 +299,10 @@ DefineContract({
     groupShouldPropose (state, getters) {
       return getters.groupMembersCount >= 3
     },
+    groupVotingRule (state, getters) {
+      // It's okay to use "PROPOSAL_GENERIC" as an example because all the settings are the same.
+      return getters.groupSettings.proposals[PROPOSAL_GENERIC]
```
taoeffect: If memory serves, think this should be `.proposals[PROPOSAL_GENERIC].rule`?

frontend/model/contracts/voting/proposals.js
```diff
@@ -34,7 +34,7 @@ export const proposalSettingsType = objectOf({
   expires_ms: number,
   ruleSettings: objectOf({
     [RULE_PERCENTAGE]: objectOf({ threshold: number }),
-    [RULE_DISAGREEMENT]: objectOf({ threshold: number })
+    [RULE_DISAGREEMENT]: objectOf({ threshold: number, thresholdAdjusted: number })
```
taoeffect: ❌ `thresholdAdjusted`

mmbotelho (commented):
@sandrina-p I ran the tests you left in the PR description and found no problems at all with those specific flows. I found some other issues, however. I am not sure if they are related to this PR:

- Everyone voted on a proposal but it was not closed:
Captura de ecrã 2020-05-15, às 17 37 48


- Weird numbers showing up in the "New proposal" footer, as well as the "your proposal was submitted" view:
Captura de ecrã 2020-05-15, às 17 39 21

Captura de ecrã 2020-05-15, às 17 38 59

strings/french.strings
sandrina-p: What's exactly the command? That does not seem to work. I didn't find any documentation on our repo about it neither.

package-lock.json
sandrina-p: It was modified because my computer did generate a diff package-lock from yours. But yeah, reverted

frontend/views/utils/currencies.js
sandrina-p: So, I'd we solve the TODO at `GroupCreationModal.vue`?

```diff
- mincomeCurrency: 'USD' // TODO: grab this as a constant from currencies.js
+ mincomeCurrency: currencies.USD.code,
```

strings/french.strings
taoeffect: Documentation is here: https://github.com/okTurtles/strings

Download the binary here: https://github.com/okTurtles/strings/releases

frontend/model/contracts/voting/rules.js
sandrina-p: I see. Changing it

frontend/views/containers/group-settings/GroupRulesSettings.vue
sandrina-p: Gotcha. Updating code.

frontend/model/contracts/group.js
sandrina-p: We also need to access `ruleSettings`, besides `rule`. Maybe the getter could have a better name. Any suggestion?

sandrina-p:
@mmbotelho what proposal system do you have activated? It should work with the "percentage" system. If it's "disagreement", it will only work in the next PR, as I said in this PR description. Maybe I wasn't clear enough, I'm sorry about that. This PR only does the logic to choose a voting system and see it in the group settings.

mmbotelho (dismissed):
No worries @sandrina-p ! I just mentioned these things to be sure it was not part of this PR.

As I mentioned, all the flows you mentioned look great to me, I didn't find anything to fix there. Great work!

.Gruntfile.babel.js
```diff
@@ -112,7 +112,7 @@ module.exports = (grunt) => {
       //    - anything in /test folder, eg. integration tests
       //    - anything that ends with .test.js, eg. unit tests for sbp domains kept in the domain folder
       test: {
-        cmd: 'node node_modules/mocha/bin/mocha --require Gruntfile.js --exit -R spec --bail "{./{,!(node_modules|test)/**/}*.test.js,./test/*.js}"',
+        cmd: 'node node_modules/mocha/bin/mocha --require Gruntfile.js --watch frontend/model/contracts/voting/rules.test.js',
```
taoeffect: Not sure what's going on here but this needs to be reverted to how it was.

frontend/model/contracts/group.js
taoeffect: Since our code is designed to support individualized settings on each type of proposal, and since we might re-add UI support for that in the future, we need to make sure any code we write is forward-compatible with that.

So imagine we were to re-add support for individualized proposal settings. In such a world this getter wouldn't make any sense at all.

So the getter that needs to be defined is one that would make sense in both worlds (the current one, and the possible future one).

There are various things you can do, but here's one suggestion:

```js
proposalSettings (state, getters) {
  return (proposalType = PROPOSAL_GENERIC) => {
    return getters.groupSettings.proposals[proposalType]
  }
}
```

frontend/model/contracts/voting/rules.js
```diff
@@ -73,3 +83,39 @@ const rules = {
...
+  if (rule === RULE_DISAGREEMENT) {
+    // Limit number of maximum "no" votes to group size
+    return Math.min(groupSizeVoting - 1, threshold)
+  }
+
+  if (rule === RULE_PERCENTAGE) {
+    // Minimum threshold correspondent to 2 "yes" votes
+    const minThreshold = 2 / groupSizeVoting
+    return Math.max(minThreshold, threshold)
+  }
```
taoeffect: Always avoid `if`s if you can.

Here's another great place to use a dictionary/map/object instead:

```js
return {
  [RULE_DISAGREEMENT]: function (...) { .. }
  // ...
}[rule]()
```

This has several advantages:

1. It is more concise and easier to read
2. It is much faster
3. It will throw an exception automatically if the wrong rule type is passed in

frontend/model/contracts/voting/rules.js
```diff
@@ -15,14 +15,22 @@ export const RULE_MULTI_CHOICE = 'multi-choice'
...
 
+/* REVIVEW PR
+  the whole "state" being passed as argument to rules[rule]() is overwhelmed.
+  It would be simpler with just 2 simple args: "threshold" and "population".
+  Advantages:
+  - population: No need to do the logic to get it (and avoid import PROFILE_STATUS.ACTIVE from group.js).
+  - threshold: The selector to get the selector is huge.
+  - tests: Avoid the need to mock a complex state.
+  My suggestion: 1 parameter (object) with 4 explicit keys.
+  [RULE_PERCENTAGE]: function ({ votes, proposalType, population, threshold })
```
taoeffect, line 27: We keep it the way it is currently because you cannot know what you might need in the future for different rule types.

taoeffect (changes requested):
Partial review, unfinished. Getting it to you so that you can iterate while I continue to review.

.Gruntfile.babel.js
sandrina-p: ups, I staged it by accident. This is a modified command to run tests only at rule.js in watch mode.

frontend/model/contracts/group.js
sandrina-p: I know you like to predict the future when it comes to code, and although I do not agree, I can respect that.

So, to do it as you want, I have one question: At `GroupRuleSettings.js`, what `proposalType` do we pass to the getter?

frontend/model/contracts/voting/rules.js
sandrina-p: I understand that. When that happens in the future, why can't we just adapt the code to pass a new parameter?

At the moment, as I see it, we are passing down the whole jungle when we just need 4 bananas.

frontend/model/contracts/voting/rules.js
sandrina-p: Updated. Here and at other places with similar `if` usage.

frontend/model/contracts/voting/rules.js
sandrina-p: Besides that, when passing `state`, we need `PROFILE_STATUS` to get `population`. When importing it we create a `circular dependency`.
> Circular dependency: frontend/model/contracts/group.js -> frontend/model/contracts/voting/rules.js -> frontend/model/contracts/group.js

mmbotelho (changes requested):
Here's what I found @sandrina-p :

Captura de ecrã 2020-06-08, às 14 35 12

------

- There's a weird scenario where the group has 2 members, the disagreement number is 7, but the maximum # of "no" votes is ajusted to 2 (which is impossible). I wonder, should we not show ajustments at all when the group has less than 3 members? Showing that info is irrelevant because proposals don't need voting processes at that stage;

-------

Captura de ecrã 2020-06-08, às 14 42 22


-----

That weird bug where the slider does not have the correct color is happening again (see image below). Here's my flow to get here:

1. Created a group with percentage based voting system, set at 80%;
2. Changed the percentage to 30%;
3. Changed to a disagreement number voting system;
4. Clicked to change to a percentage based voting system again (that's when I saw this modal).

The app does not need to save that 30% value, it can just show the suggested 80% always. It's less confusing.

Captura de ecrã 2020-06-08, às 14 43 36

-----

I couldn't test the voting systems with a bigger group because the app crashed with this error:

Captura de ecrã 2020-06-08, às 15 29 18


I'll reset everything and test with a big group again!

mmbotelho:
Review part II - big groups:

I didn't find anything there! Seems to be working fine 😄

frontend/model/contracts/voting/rules.js
sandrina-p: Update: This took me a long time, but unit tests were failing exactly because of this circular dependency. (too complex to explain in a short comment)

Anyway, I added a const `PROFILE_STATUS` to this file as a short solution until you get back to this PR :)

sandrina-p:
@mmbotelho all fixed except 2 things:

- The only one not fixed was the "7 adjusted 2*" scenario. I'm okay with not showing that for small groups but would like to know @dotmacro opinion too.
- Modal threshold value: If it's the same system, it starts with the current value. (e.g. "30%"). If it's to change to a different system it starts with the default value (80%). That way is not confusing when updating the rule value.

dotmacro:
> The only one not fixed was the "7 adjusted 2*" scenario. I'm okay with not showing that for small groups but would like to know @dotmacro opinion too.

Yep, I think we can hide it for small groups because it's irrelevant (and potentially confusing) info.

sandrina-p:
> Yep, I think we can hide it for small groups because it's irrelevant (and potentially confusing) info.

Actually, I'm not sure after all if it's okay to hide/ignore that adjustment because then it shows values that are not the truth. For example _"40% (0 out of 1 member)"_.

We could create a "small group adjustment" to avoid these scenarios but still... I think it's better to show the adjustment right away with the banner message.

And, as Andrea suggested in Figma, this card could have a short explanation with a link to F.A.Q/blog post talking more about voting systems, so that the users understand this right away.

mmbotelho:
> Actually, I'm not sure after all if it's okay to hide/ignore that adjustment because then it shows values that are not the truth. For example "40% (0 out of 1 member)".

@sandrina-p why not just remove the "(0 out of 1 member)" part too? It doesn't make sense either.

We need to indicate somehow, on that page, that the voting system is not active yet and will only be once the group has 3 or more members. [I've created an issue for it so I can get to it eventually ](https://github.com/okTurtles/group-income-simple/issues/941).

So my suggestion would be:
- Remove the adjustment indicator (text label);
- Remove the "(x out of x members)" text label;
- Leave the rest for a different PR, sdo I have time to improve it.

Does that sound good?

sandrina-p:
@mmbotelho, it makes sense! So, these are the last changes:
- Group Settings - do not show any adjustment when the group is small.
- Proposals: Update the proposal descriptions. This [screenshot](https://user-images.githubusercontent.com/14869087/84495617-5ab9e380-aca3-11ea-8030-81e8b16ae999.png) shows 3 scenarios, however, there's one more (disagreement from X to Y) also implemented.

mmbotelho (changes requested):
There's something weird happening when I try to change from a percentage based voting system to a disagreement number:

Captura de ecrã 2020-06-12, às 16 55 50
Captura de ecrã 2020-06-12, às 16 56 44
Captura de ecrã 2020-06-12, às 16 57 19
Captura de ecrã 2020-06-12, às 16 57 37


This happened with a group that had 3 members and then was reduced to 2. I thought the problem might be related to that so I added one more member (to have 3 members again) and the problem persisted.

I also found a very minor spacing issue here:
Captura de ecrã 2020-06-12, às 16 48 02

mmbotelho:
Also, related to my previous comment - Shouldn't the error message be displayed BEFORE I move on to the proposal description? It's a bit weird to be notified of a form error 2 steps ahead.

Also, the error itself seems a bit odd, because I should be able to change from a percentage v.s. to a disagreement v.s. even if they result in the same "x out of x members",  right?

sandrina-p:
- Uh, nice catch @mmbotelho. We have a validation to prevent creating proposals with an identical threshold to the current one (i.e. it was 1 by default, even if "not active"). But here, it does not make sense because we are switching systems. So I disabled that validation for that case.
- Spacing bug fixed
- The validation by default happens always in the "submit" step (for all proposals). in this PR I added the validation to the first step too. Maybe we should create an issue to make the validation UX consistent across all proposals.

pieer (dismissed):

frontend/model/contracts/group.js
taoeffect: You don't need to pass in any proposalType, you can call it as an empty function. In the code I wrote above, the default, when no proposalType is passed in, will be `PROPOSAL_GENERIC`.

frontend/views/utils/currencies.js
taoeffect: Upon seeing this, I'm realizing it was probably a mistake to write that TODO. It just adds more work for us to have to maintain the same text in multiple places. Feel free to revert and remove the `// TODO` comment

frontend/model/contracts/group.js
sandrina-p: I suspected that, was just to double-check, thanks!

frontend/views/utils/currencies.js
sandrina-p: reverted

frontend/model/contracts/voting/rules.js
taoeffect: I gave this some thought, and think we can resolve it by doing the following:

1. Move the file `model/constants.js` into `model/contracts/contants.js`
2. Move the constants `INVITE_INITIAL_CREATOR`, `INVITE_STATUS`, and `PROFILE_STATUS` in `group.js` into `model/contracts/contants.js`
3. Then in `rules.js` and `group.js`, import those constants as needed.
4. Remove the comment above, since it's part of the design to pass `state` into the function. If you're worried about performance, don't be; it shouldn't have any effect on performance since JavaScript passes values by reference.

frontend/model/contracts/voting/rules.js
```diff
@@ -73,3 +89,49 @@ const rules = {
...
+export const getCountOutOfMembers = (groupSize, decimal) => {
+  const minGroupSize = 3 // when group can vote
+  return Math.ceil(Math.max(minGroupSize, groupSize) * decimal)
+}
+
+export const getPercentFromDecimal = (decimal) => {
+  // convert decimal to percentage avoiding weird decimals results.
+  // e.g. 0.58 -> 58 instead of 57.99999
+  // based on https://stackoverflow.com/a/11832950/4737729
+  return Math.round(decimal * 100 * 100 / 100)
```
taoeffect: This seems quite different from what's done in https://stackoverflow.com/a/11832950/4737729 ?

And `100 / 100` = `1`...

frontend/model/contracts/group.js
```diff
@@ -727,6 +734,20 @@ DefineContract({
         }
       }
     },
+    'gi.contracts/group/updateVotingRules': {
+      validate: objectOf({
+        ruleName: x => [RULE_PERCENTAGE, RULE_DISAGREEMENT].includes(x),
```
taoeffect, line 734: I don't think `objectOf` supports this type of syntax...? So this should be `string`.

frontend/model/contracts/group.js
```diff
@@ -727,6 +734,20 @@ DefineContract({
...
+        ruleThreshold: number
+      }),
+      process ({ data, meta }, { state }) {
+        // Update all types of proposal settings for simplicity
+        for (const ruleName in state.settings.proposals) {
+          Vue.set(state.settings.proposals[ruleName], 'rule', data.ruleName)
+          Vue.set(state.settings.proposals[ruleName].ruleSettings[data.ruleName], 'threshold', data.ruleThreshold)
```
taoeffect: Things should be named after what they do. If all this does is update the threshold for a rule, then the name should reflect that.

`'gi.contracts/group/updateVotingRules'` => `'gi.contracts/group/updateVotingRuleThreshold'`

frontend/model/contracts/voting/rules.js
sandrina-p: ah, the fn started as a different thing but then I simplified it and forgot to remove that redundancy. fixed.

frontend/model/contracts/group.js
sandrina-p: I believe it does. It's the same we did in the method `grouPprofileUpdate`.

frontend/model/contracts/group.js
sandrina-p: I only update the threshold because it's the only thing we need ATM. But it can be modified to support other things too. (e.g. `expires_ms`). I just didn't do it now because it's not needed.

frontend/model/contracts/group.js
taoeffect: That uses `objectMaybeOf`, which does support it

frontend/model/contracts/group.js
taoeffect: Remember, nothing here “can be modified”, that is why it’s a “contract”. These functions are set in stone after the prototype is released, so that they generate the same exact state forever.

sandrina-p: I understand that. Before the prototype is released, we will still touch this again to do #620, so I didn't think it was a big deal for now. But I'll improve it in this PR :)

frontend/model/contracts/voting/proposals.js
```diff
@@ -113,14 +117,7 @@ const proposals = {
...
     defaults: {
-      rule: RULE_PERCENTAGE,
-      expires_ms: 14 * DAYS_MILLIS,
-      ruleSettings: {
-        [RULE_PERCENTAGE]: { threshold: 0.8 },
-        // at least 2, since the member being removed shouldn't be able to
-        // block the proposal themselves
-        [RULE_DISAGREEMENT]: { threshold: 2 }
-      }
+      ...proposalDefaults
```
taoeffect: Hmm, this line is different from the other ones... maybe it was done this way to take into account `[RULE_DISAGREEMENT]: { threshold: 2 }` ?

That was never re-incorporated here though. The comment also needs to be preserved.

sandrina-p: It should be the same, yes.

I didn't re-incorporated it because there are other types of "adjustments" needed based on other conditions. So, all adjustments are made at `rules.js` to keep everything together.

frontend/controller/actions/group.js
```diff
@@ -53,25 +59,29 @@ export default sbp('sbp/selectors/register', {
           groupName: name,
           groupPicture: finalPicture,
           sharedValues,
-          mincomeAmount: +mincomeAmount, // ensure this is a number
+          mincomeAmount: +mincomeAmount,
```
taoeffect, line 60: Please preserve these types of comments. They might not mean much to you but they are super helpful for newbie volunteers.

frontend/model/contracts/group.js
taoeffect: So there are three options here:

1. Keep using `objectOf` and change `ruleName` verification to be `string` instead of a function
2. Keep using `objectOf` and add support for using custom functions
3. Switch it to `objectMaybeOf`; ~~**I think this is a bad choice**~~ Actually, it is the right choice! See section below. I will keep what I've written immediately below so that you can see how my thought-process evolved.

I think, for now, since we are renaming this to be a very precise and specific action called `updateVotingRuleThreshold`, that we should stick with `objectOf`, in which case you can do either (1) or (2) above.

The reason why I think `objectMaybeOf` (3) is a bad choice, is because this function **requires** both `ruleName` and `ruleThreshold` to work. So we cannot have options missing in the data sent.

If in the future this function is renamed to `updateVotingRuleSettings` (pre-prototype, or post-prototype it is simply added as a new action), then we can switch it to use `objectMaybeOf` and this custom validation function.

### Regarding `expires_ms`

Since another we are not updating the expiry settings of a voting rule in this function, ~~it is doubly-important to rename this to `updateVotingRuleThreshold`.~~ (Nope! see below)

### Oops! Almost ignore what I wrote above

Having looked deeper into this PR, I am now realizing that we have little choice about what to do here.

Since you are calling this action directly in `PROPOSAL_PROPOSAL_SETTING_CHANGE`, then you *must* support all of the different types of settings, including `expires_ms`, because there is only *one* kind of `PROPOSAL_PROPOSAL_SETTING_CHANGE` proposal, and it must be able to handle all of the proposal settings changes thrown at it. That includes changes to the expiry time.

What this means is:

1. Rename this to `'gi.contracts/group/updateAllVotingRules'`
2. Use (3) `objectMaybeOf` so that in the future this supports adjusting `expires_ms`
3. In the process function, only update settings that have been defined. E.g. check if `data.ruleName` is defined, and if it is, perform the change, otherwise skip the change. Same for `data.ruleThreshold` and `data.expires_ms`

frontend/model/contracts/voting/proposals.js
taoeffect: OK, I do see that `rules.js` protects against `{ threshold: 1 }`.

Can we have consistent styling though here, so that this is changed to `defaults: proposalDefaults,` just like it is everywhere else?

frontend/model/contracts/group.js
taoeffect: See my comment above for more details on the changes that must be made to this function.

frontend/model/contracts/voting/proposals.js
```diff
@@ -164,30 +154,23 @@ const proposals = {
...
-      rule: RULE_PERCENTAGE,
-      expires_ms: 7 * DAYS_MILLIS,
-      ruleSettings: {
-        [RULE_PERCENTAGE]: { threshold: 0.8 },
-        [RULE_DISAGREEMENT]: { threshold: 1 }
+    defaults: proposalDefaults,
+    [VOTE_FOR]: function (state, { meta, data, contractID }) {
+      const proposal = state.proposals[data.proposalHash]
+      const { ruleName, ruleThreshold } = proposal.data.proposalData
+      proposal.status = STATUS_PASSED // REVIEW/BUG This mutates the state and it's a Vue bad practice.
```
taoeffect: This is correct Vue.js / Vuex practice. Vuex state is allowed to be directly mutated here. All of this is done inside of the Vuex.commit function.

Anything asynchronous is done using `pushSideEffect`.

frontend/model/contracts/voting/proposals.js
```diff
@@ -164,30 +154,23 @@ const proposals = {
...
+      const message = {
+        meta,
+        data: { ruleName, ruleThreshold },
```
taoeffect: Very important to not destructure on line 161 and instead here do `data: proposal.data.proposalData`, since it's possible that `expires_ms` might (or might not) be included.

frontend/model/contracts/voting/proposals.js
```diff
@@ -164,30 +154,23 @@ const proposals = {
...
+        contractID
       }
-    },
-    [VOTE_FOR]: function (state, { meta, data }) {
-      throw new Error('unimplemented!')
+      sbp('gi.contracts/group/updateVotingRules/process', message, state)
```
taoeffect: Just a reminder to rename this to `'gi.contracts/group/updateAllVotingRules/process'` per the other large comment above.

frontend/model/contracts/voting/rules.js
```diff
@@ -15,14 +19,27 @@ export const RULE_MULTI_CHOICE = 'multi-choice'
...
+const getPopulation = (state) => Object.keys(state.profiles).filter(p => state.profiles[p].status === PROFILE_STATUS.ACTIVE).length
+
 const rules = {
   [RULE_PERCENTAGE]: function (state, proposalType, votes) {
     votes = Object.values(votes)
-    const population = Object.keys(state.profiles).length
+    const population = getPopulation(state)
     const defaultThreshold = state.settings.proposals[proposalType].ruleSettings[RULE_PERCENTAGE].threshold
-    const threshold = proposalType === PROPOSAL_REMOVE_MEMBER
-      ? Math.min(defaultThreshold, (population - 1) / population)
```
taoeffect, line 24: Can you keep the old formatting?

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -138,6 +136,45 @@ export default {
...
+      return getThresholdAdjusted(this.proposalSettings.rule, threshold, this.groupMembersCount)
+    },
+    footerVotingExplanation () {
+      return {
+        [RULE_DISAGREEMENT]: () => {
+          if (this.threshold === 1) {
+            return L('Your proposal will be accepted if {b_}no one{_b} disagrees.', LTags('b'))
+          }
+          // REVIEW - This copy is not very positive.
+          return L('Your proposal will be rejected if {b_}{n} members{_b} disagree.', { n: this.threshold, ...LTags('b') })
```
taoeffect: `Proposal will pass if fewer than {b_}{n} members{_b} disagree.`

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -138,6 +136,45 @@ export default {
...
     },
+    threshold () {
+      const threshold = this.proposalSettings.ruleSettings[this.proposalSettings.rule].threshold
+      return getThresholdAdjusted(this.proposalSettings.rule, threshold, this.groupMembersCount)
+    },
+    footerVotingExplanation () {
+      return {
+        [RULE_DISAGREEMENT]: () => {
+          if (this.threshold === 1) {
+            return L('Your proposal will be accepted if {b_}no one{_b} disagrees.', LTags('b'))
```
taoeffect: `Proposal will pass if {b_}no one{_b} disagrees.` (to match the copy below)

frontend/model/contracts/voting/rules.js
```diff
@@ -45,20 +59,22 @@ const rules = {
...
     const turnout = votes.length
     const absent = population - turnout
+
     console.debug(`votingRule ${RULE_DISAGREEMENT} for ${proposalType}:`, { totalFor, totalAgainst, threshold, turnout, population, absent })
     if (totalAgainst >= threshold) {
       return VOTE_AGAINST
     }
     // consider proposal passed if more vote for it than against it and there aren't
     // enough votes left to tip the scales past the threshold
-    return totalAgainst + absent < threshold && totalFor > totalAgainst ? VOTE_FOR : VOTE_UNDECIDED
```
taoeffect, line 61: Why was `&& totalFor > totalAgainst` removed?

I agree it's probably incorrect. Probably should check to make sure at least someone voted for it, so should be `&& totalFor > 0` is my thinking.

That's not quite enough to close these bugs but it's an improvement.

- https://github.com/okTurtles/group-income-simple/issues/876
- https://github.com/okTurtles/group-income-simple/issues/877

taoeffect (changes requested):
OK, I've done my best to review the critical bits (not everything!). Please see these comments as well as some of the unaddressed ones from before! I hope to be back in better shape soon!

frontend/views/containers/proposals/ProposalTemplate.vue
sandrina-p: I like it, it's shorter, but I think we should keep the pronoun "Your proposal...".

frontend/model/contracts/voting/rules.js
sandrina-p: I removed this while writing the tests. See the set `RULE_DISAGREEMENT - 4 - 5 members`, test `2VF vs 3VA returns for`. With the previous code, the proposal would return `VOTE_UNDECIDED `, but in reality, it should return `VOTE_FOR` because it's needed more `totalAgainst` than `totalFor` votes.

Why do we need to check if at least some voted for it?

Regarding #876, I didn't understand how to fix the bug. I understand the problem but I don't know what changes to make taking in account the expire date. We don't even have the code to automatically close expired proposals, yet, right? So, maybe that should be done in another PR.

Regarding #877, I prefer to leave it to a future PR. Probably at the same time as #878.

taoeffect: > Why do we need to check if at least some voted for it?

Well that's an interesting question... should a proposal pass if no one is in favor of it...? But I do see your point. Purely with the way the disagreement rule is described, it is implied that a proposal can pass with no one voting explicitly in favor of it.

> Regarding #876, I didn't understand how to fix the bug. I understand the problem but I don't know what changes to make taking in account the expire date. We don't even have the code to automatically close expired proposals, yet, right? So, maybe that should be done in another PR.

Agreed, let's leave for another PR.

> Regarding #877, I prefer to leave it to a future PR. Probably at the same time as #878.

👍

frontend/model/contracts/voting/rules.js
sandrina-p: Yes, keeping constants in its own file is much better. Changed

mmbotelho (dismissed):
Approved @sandrina-p ! Great job

I'll create an issue for the validation problem!

taoeffect (approved):
Approved! Let's move forward! 😄

PR #917, opened 2020-05-08
#763 Payments - dry payment row

sandrina-p:
Continue #763

### What's done:
- Solve 5 items from the list on #763:
> - [x] Someone needs to go through and clean up all of the CSS — deduplicate the all of the copy/pasted junk in `PaymentRow*.vue` files and DRY it all so that it's all stored in one of the `.scss` files and not repeated
> - [x] Create mixins and remove no longer used computed properties/methods from the `PaymentRow*.vue` files (and any other payment related files)
> - [x] Delete `humanDate.js` and move the `humanDate` function into `time.js`
> - [x] Delete outdated comments in the markup
> - [x] Cleanup and DRY the markup, CSS, and any JS, but especially the CSS
- Improve broken `Progress.vue` animation. Now it should work fine on chrome, firefox and safari.


frontend/utils/time.js
```diff
@@ -43,3 +43,11 @@ export function compareISOTimestamps (a: string, b: string): number {
   const B = new Date(b).getTime()
   return A > B ? 1 : (A < B ? -1 : 0)
 }
+
+export function humanDate (
+  datems = Date.now(),
+  opts = { month: 'short', day: 'numeric' }
+) {
+  const locale = navigator.languages ? navigator.languages[0] : navigator.language
```
taoeffect: Small: can you move this `const` outside of this function so that the logic is not performed on each invocation of the function?

frontend/views/containers/payments/PaymentRowRecord.vue
```diff
@@ -51,6 +43,9 @@ export default {
   },
   data () {
     return {
+      config: {
+        initialAmount: this.payment.amount
```
taoeffect, line 47: Nicely solved 👍

taoeffect (changes requested):
Fantastic work! Just one minor thing noticed. This is my "first draft review". I think that might be all that's necessary for this. Giving it to you right now so that you can quickfix while I test out the PR locally!

frontend/utils/time.js
sandrina-p: Done. But the locale logic still needs to be done inside the fn at least once, otherwise, the server will break (navigator does not exist there). Question. Will we support SSR on GI?

frontend/utils/time.js
```diff
@@ -44,10 +44,13 @@ export function compareISOTimestamps (a: string, b: string): number {
   return A > B ? 1 : (A < B ? -1 : 0)
 }
 
+// Do not initialize locale here so that it runs on the server (navigator is undefined there)
+let locale
```
taoeffect: I do not understand this comment. Nothing in this file runs on the server?

I think you can (and should) initialize it as a constant here.

sandrina-p: No, it will break. see the last Travis build. I'm rolling back.

taoeffect: Ah I see why travis is failing. It's failing because `navigator` isn't defined in mocha.

That's OK, so you can just write this as:

```js
// prevent mocha tests from breaking (navigator is undefined in them)
const locale = (navigator && navigator.languages ? navigator.languages[0] : navigator.language) || 'en-US'
```

sandrina-p: I'm ok-ish with that solution. Question: Will we support SSR on GI?

frontend/utils/time.js
```diff
@@ -43,3 +43,13 @@ export function compareISOTimestamps (a: string, b: string): number {
   const B = new Date(b).getTime()
   return A > B ? 1 : (A < B ? -1 : 0)
 }
+
+// TODO: Provide locale fallback in case navigator does not exist (e.g. server, Mocha, etc...)
+const locale = (navigator && navigator.languages ? navigator.languages[0] : navigator.language) || 'en-US'
```
taoeffect: Oops, it's still failing. It took me a bit but I finally figured out how to write this so that it doesn't fail:

```js
const locale = (typeof navigator === 'undefined' && 'en-US') || (navigator.languages ? navigator.languages[0] : navigator.language)
```
Verified locally that that works!

taoeffect (approved):
Real excellent work @sandrina-p! 👏

PR #918, opened 2020-05-11
#763 Payments - add logic to PaymentDetail, MonthOverview, RecordPayment and ContributionsWidget

sandrina-p:
Continue #763

image

image

image

image

### What's done:
- Solve 3 items from the list on #763:

> - [x] Correctly implement the `PaymentDetails.vue` file
> - [x] Correctly implement the `MonthOverview.vue` file
> - [x] Remove the `Math.random()` stuff in the `PaymentRow*.vue` files (mostly related to payment date stuff) and replace it with the actual data

- Improve `RecordPayment.vue`
  - Remove mutation to prop `paymentsList` (Vue anti-pattern). It was causing some weird bugs related to the payment amount and checkbox usage.
  - The moment the user modifies an amount, the checkbox is automatically checked.
  - Implement logic to leave a note (memo).
  - Add missing error handling. 
- Add missing real data to Payments footer. Ex: _"$20.51 in total, to 2 members"_ 
- Add missing real data to "Contributions Widget".
- Left comments with some bugs found / questions about Payment logic.


frontend/utils/time.js
```diff
@@ -44,12 +44,20 @@ export function compareISOTimestamps (a: string, b: string): number {
   return A > B ? 1 : (A < B ? -1 : 0)
 }
 
+export function lastDayOfMonth (date: Date): Strin {
```
taoeffect: If this returns a string the return value flow annotation should be `string`, but since it returns a `Date` the annotation should be that too here.

frontend/model/contracts/group.js
```diff
@@ -401,8 +401,7 @@ DefineContract({
         paymentHash: string,
         updatedProperties: objectMaybeOf({
           status: paymentStatusType,
-          details: object,
-          memo: string
```
taoeffect: This memo should be here (plz double-check before removing fields)

taoeffect:
Still doing my PR review, but a minor copy request for this one: can you please change line 83 in ContributionItem.vue from:

```js
          return L('{service} by {who}', html)
```

to:

```js
          return L('{service} from {who}', html)
```

Proper English says this should read "$270 from sue", not "$270 by sue":

Screen Shot 2020-05-13 at 1 27 28 PM

cc @mmbotelho please feel free to update the designs accordingly.

frontend/utils/time.js
```diff
@@ -44,12 +44,20 @@ export function compareISOTimestamps (a: string, b: string): number {
...
 
 export function humanDate (
-  datems = Date.now(),
+  datems,
   opts = { month: 'short', day: 'numeric' }
 ) {
+  if (!datems) {
+    console.error('humanDate:: 1st arg `datems` is required')
+    return ''
+  }
```
taoeffect, line 61: If it's required, it should throw `GIErrorUIRuntimeError`?

```js
if (!date) {
  throw new GIErrorUIRuntimeError('humanDate: missing first argument')
}
```

Also consider renaming from `datems` to `date` since it's usually a string that's passed in.

taoeffect (changes requested):
Partial unfinished review. Still more to be done. Just getting something to you!

mmbotelho:
> Still doing my PR review, but a minor copy request for this one: can you please change line 83 in ContributionItem.vue from:
> 
> ```js
>           return L('{service} by {who}', html)
> ```
> 
> to:
> 
> ```js
>           return L('{service} from {who}', html)
> ```
> 
> Proper English says this should read "$270 from sue", not "$270 by sue":
> 
> Screen Shot 2020-05-13 at 1 27 28 PM
> 
> cc @mmbotelho please feel free to update the designs accordingly.

@taoeffect I believe this was something we had a huge discussion about on the Bulgaria hackathon with @pieer, I can't remember exactly what it was but for some reason you mentioned (related to translations) we had to use "by" in all scenarios. Is this not a problem anymore?

taoeffect:
> but for some reason you mentioned (related to translations) we had to use "by" in all scenarios. Is this not a problem anymore?

I do have vague memories of some discussion... I have no concrete memory though of them, so can only say that from my understanding of English, it should be "from" in this case. I think it will work for services too, e.g. "Programming lessons from Sam", and multiple users, e.g. "$500 from sam and andrea".

frontend/model/contracts/group.js
sandrina-p: Why is it needed? I didn't see any place using it. The UI only needs `memo` at `group/payment`.

frontend/utils/time.js
sandrina-p: throwing an error would break the application because we don't have try/catch for each time we use `time.js`. I think that would be a very drastic outcome from a simple util fn. Logging the error and return empty string is better than crashing the app.

frontend/model/contracts/group.js
taoeffect: @sandrina-p Everything in this file involves thought that has been extended out like 5 years into the future. Just because something isn't being used right now doesn't mean it won't be used in the future. :P

Different payment systems support different levels of information, and if we want to be able to support all payment systems plus our own metadata and notes, we need this field. Please restore this much needed functionality 🙏

mmbotelho:
So should we replace every "by" with "from"? @taoeffect

frontend/model/contracts/group.js
sandrina-p: I did not know about that, it wasn't explicit anywhere. I'll add it back then.

sandrina-p:
> So should we replace every "by" with "from"?

I don't think so @mmbotelho, it was only asked to change line 83, and that's only the "RECEIVING" condition. If it's giving it stays the same.

pieer (dismissed):

sandrina-p:
Update: Added the logic to the dashboard "Contributions Widget" (Payments)

taoeffect:
Quick note @sandrina-p: while testing, I noticed that pressing "Enter" in the record payment input field closes the modal but doesn't record the payment. Can you update it so that it  records a payment?

taoeffect:
Also when viewing payment details there's an error in the console:

Screen Shot 2020-06-11 at 12 19 48 PM

sandrina-p:
@taoeffect, both issues fixed!

frontend/model/contracts/group.js
```diff
@@ -232,6 +232,158 @@ DefineContract({
...
+            ...p,
+            total: existingAmount,
+            partial: partialAmount > 0
+          })
+        }
+        return payments
+      })()
+
+      return { sent, todo, received, toBeReceived }
+    },
```
taoeffect: @sandrina-p Can you simplify and dry this entire function? It is huge. Right now it is hard for me to follow and read. It contains a billion loops. Can you please:

1. Reduce the number of loops that occur in this function? This means: using `for () { }` instead of higher-order-functions like `.map().find().filter()` <- each one of those function calls is a loop. You can replace them with a single loop instead, for example, look at how `paymentsTodo()` is implemented in `Payments.vue`
2. Call getters only once in the entire function and cache any results. For example, there are two calls to `getters.groupIncomeDistribution` in this function. There should be only one.

Using these types of techniques, see if you can reduce the size of this function by at least 50%

frontend/model/contracts/group.js
```diff
@@ -232,6 +232,158 @@ DefineContract({
...
+    ourPaymentsSummary (state, getters) {
+      const { todo, sent, toBeReceived, received } = getters.ourPayments
+      const isCompleted = (p) => p.data.status === PAYMENT_COMPLETED
+      const isPartialCount = (list) => list.filter(p => p.partial).length
+      const getUniqPayments = (users) => {
```
taoeffect: Maybe give this a more descriptive name, like `uniqLength`?

frontend/model/contracts/group.js
taoeffect: Ah, OK, sorry, yesterday I didn't see that you have moved the stuff from `Payments.vue` into here... ok, this is going to require seriously close review on my part... still, any simplifications and loop-removals that you see in the meantime please go ahead and do them!

frontend/utils/time.js
taoeffect: OK, fair enough 👍

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -34,48 +43,8 @@ export default {
...
-          {
-            name: 'Hugo Lil',
-            avatar: '/assets/images/default-avatar.png',
-            status: 'completed',
-            amount: 50
-          }
-        ]
-      }
-    }
-  },
```
taoeffect, line 75: 😍 🤩

frontend/model/contracts/group.js
```diff
@@ -227,6 +227,158 @@ DefineContract({
...
+        }
+        return {
+          paymentsDone: getUniqPayments(pByUser.sent) - pPartials,
+          hasPartials: pPartials > 0,
+          paymentsTotal: getUniqPayments([...pByUser.todo, ...pByUser.sent]),
+          amountDone: sentCompleted.reduce((total, p) => total + p.data.amount, 0),
+          amountTotal: todo.reduce((total, p) => total + p.amount, 0) + sent.reduce((total, p) => total + p.data.amount, 0)
+        }
+      }
+    },
```
taoeffect: @sandrina-p since both of these getters rely on data that does not come directly from the contract messages / current-contract-state (e.g. `currentMonthstamp()`), please move both getters into `state.js`

sandrina-p: Following that mindset, other getters should also be moved, such as `groupIncomeDistribution`, `groupIncomeAdjustedDistribution` and `thisMonthsPaymentInfo`. Should I move all of them too?

test/cypress/integration/group-contributions.spec.js
```diff
@@ -0,0 +1,533 @@
+const userId = Math.floor(Math.random() * 10000)
```
sandrina-p: NOTE: This file is not new, I just renamed `group-payments.spec.js` to `group-contributions.spec.js` and did not modify it at all.
Then, I created a new `group-payments.spec.js` to write the actual tests related to payments. (where we also override the time `Date` with `cy.clock`)

test/cypress/integration/group-paying.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      cy.log('Record one payment')
+      cy.get('button[type="submit"]').click()
+      cy.getByDT('successClose').click()
+      cy.getByDT('closeModal').should('not.exist')
+    })
+
+    // BUG - Making a late payment does not work as expected.
+    // Expected: Remove user3's late payment of $78.57.
+    // Reality: The late payment persists and the current month payment is reduced to partial ($100 out of $178.57)
+    // Once that is fixed, uncomment the assertions bellow:
```
sandrina-p, line 150: @taoeffect I started doing the tests for payments and it worked fine in normal payments.

But in late-payments, some parts are still broken: The list of payments is correct but the code to do the actual payment is broken. I think this should be left for another PR because this one is already large enough

frontend/model/contracts/group.js
taoeffect: @sandrina-p Yep! Excellent catch! And maybe you could add a comment about this above all of them (in `state.js`) explaining all of this (e.g. why they are there are not in group.js)

sandrina-p: @taoeffect could you explain again what's the problem with having `currentMonthstamp` at `group.js` and why is it okay to be at `state.js`? My memory is failing me here...

Also, I wonder if having `currentMonthstamp` in the getters isn't dangerous because it's not a "static value", but rather a dynamic value that can change over time without the user action. This may conflict with some contract message `process`... Imagine something related to the current month happens on 30 June and then another member logs one day after (1 July) to process the message from the previous month. A getter result won't be the same because of `currentMonthstamp`. I think there isn't any scenario that does this (yet), but it's something we need to take attention.

taoeffect: > Also, I wonder if having currentMonthstamp in the getters isn't dangerous because it's not a "static value", but rather a dynamic value that can change over time without the user action

It's dangerous precisely because of that, and that is why it must not be inside of the contract getters but only in the Vuex getters.

The getters in state.js are Vuex-only getters. These are getters that are not used to build up the contract state, which can only be built-up using the information that's coming from the messages being sent to the contract.

Since `currentMonthstamp` is, as you point out, a dynamic value that comes *not* directly from the data in contract messages, but from the end-user's computer, it can only be used in the Vuex-getters that are used to build up the ephemeral, non-contract state that is in the UI.

The *contract getters* can only access information from two places:

1. each `GIMessage` that comes in (e.g. that's passed to the `process` function)
2. the existing *contract state* (and **not** any other state, e.g. no Vuex-state is allowed to be used that exists outside of the contract state)

sandrina-p: gotcha, so we are on the same page. moved everything to `state.js` and created a new file with `groupIncomeDistribution` to be imported from both files (`state.js` and `group.js`)

frontend/model/contracts/group.js
```diff
@@ -99,44 +99,6 @@ function initFetchMonthlyPayments ({ meta, state, getters }) {
   return monthlyPayments
 }
 
-function groupIncomeDistribution ({ state, getters, monthstamp, adjusted }) {
```
sandrina-p, line 102: note: this function was moved to `utils/distribution/group-income-distribution.js`

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+                displayName: this.userDisplayName(payment.to),
+                amount: payment.amount, // TODO: include currency (what if it was changed?)
+                // partiaL: TODO add this info as it is in this.paymentsTodo
+                isLate: true,
+                date: dueIn
+              })
+            }
+          }
+        }
+        */
```
taoeffect: >       // QUESTION: Is this only from the previous month? What if there are payments from 2-3 months ago?

We only support late payments for the previous month. We ignore all other months because it would be too complicated to deal with. We could get away without supporting late payments at all, but we have support for them, just for one month.

> // This logic below is wrong (based on cypress tests). Commented for later analysis.

That logic needs to be uncommented. If you want to fix it in a future PR, you can comment out the tests for the late payment and update this code/fix the tests in another PR, but this code should be left uncommented.

test/cypress/integration/group-paying.spec.js
taoeffect: Well the commented out section above needs to be uncommented. If you want to comment out the late payment *tests* that's fine.

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+            continue
+          }
+
+          latePayments.push({
+            ...payment,
+            // partial: TODO add this info, but what's the cleanest way to do it?
+            date: dueIn
+          })
+        }
+      }
```
taoeffect, line 489: What's going on here? This code should probably be restored to how it was before..

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+    const allPayments = getters.currentGroupState.payments
+
+    const sent = (() => {
+      const payments = []
+      // QUESTION: Is this only from this month? What about the previous payments from months ago?
+      //           Same question for received.
+      for (const monthstamp of Object.keys(monthlyPayments).sort()) {
+        const { paymentsFrom } = monthlyPayments[monthstamp]
+
+        if (!paymentsFrom) { continue }
```
taoeffect: Am not a fan of this `if (!paymentsFrom) { continue }` style personally... prefer how it was before

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+        }
+      }
+      return payments
+    })()
+    const received = (() => {
+      const thisMonthPayments = monthlyPayments[cMonthstamp]
+      const paymentsFrom = thisMonthPayments && thisMonthPayments.paymentsFrom
+      const payments = []
+
+      if (!paymentsFrom) { return [] }
```
taoeffect: Same here. There would be a single `return` statement in this function if this was restored to how it was before (with the `if` wrapping the code below).

Same goes for lines 403 and 408 below. It would make it easier for me to read and follow the logic of the code, as well as compare it to the previous code to see if anything significant changed. Would appreciate if that style/logic can be revered.

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+
+        if (amount <= 0) { continue }
+
+        const partialAmount = existingAmount - amount
+        const existingPayment = {}
+        if (partialAmount > 0) {
+          // TODO/BUG this only work if the payment is done in 2 parts. if done in >=3 won't work.
+          const sentPartial = sent.find((s) => s.username === p.to && s.amount === partialAmount)
+          if (sentPartial) {
+            existingPayment.hash = sentPartial.hash // Q: Why do we need this hash?
```
taoeffect: All payment entries in `payments` have the hash. They need it because otherwise there's no way to know what payment this represents.

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+        const existPayment = unadjusted.find(({ to }) => to === p.to) || { amount: 0 }
+        const amount = +currency.displayWithoutCurrency(p.amount)
+        const existingAmount = +currency.displayWithoutCurrency(existPayment.amount)
+
+        if (amount <= 0) { continue }
+
+        const partialAmount = existingAmount - amount
+        const existingPayment = {}
+        if (partialAmount > 0) {
+          // TODO/BUG this only work if the payment is done in 2 parts. if done in >=3 won't work.
```
taoeffect: OK, can you fix this in a future PR? Maybe update the Payments issue to reflect this (e.g. add a checkbox)?

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+
+        payments.push({
+          ...existingPayment,
+          ...p,
+          total: existingAmount,
+          partial: partialAmount > 0,
+          dueIn
+        })
+      }
+      return payments
```
taoeffect, line 424: How are payments that were marked as not received handled? That code is missing from here. It should be added back, at least commented out with a TODO comment and a corresponding update to the Payments issue if there's an issue with that.

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+            }
+          }
+        }
+        */
+
+        // On the other hand this seems to work as expected. "partial" is the only thing missing
+        for (const payment of adjusted) {
+          if (payment.from !== ourUsername) {
+            continue
+          }
```
taoeffect: Not sure I understand what's going on here...

frontend/model/state.js
```diff
@@ -329,6 +335,235 @@ const getters = {
...
+        }
+        payments.push({
+          ...existingPayment,
+          ...p,
+          total: existingAmount,
+          partial: partialAmount > 0
+        })
+      }
+      return payments
```
taoeffect: Hmm. This part I still need to review since this is new. Don't quite fully understand yet. Maybe, could you add a comment above line 492 explaining why you found it necessary to implement this?

taoeffect (changes requested):
Some more review (still not fully reviewed)

frontend/model/contracts/group.js
```diff
@@ -264,7 +216,7 @@ DefineContract({
         const memberProfile = getters.currentGroupState.profiles[username]
         if (!memberProfile) return false
         const memberJoinedMs = new Date(memberProfile.joinedDate).getTime()
-        const joinedAfterUs = weJoinedMs < memberJoinedMs
+        const joinedAfterUs = weJoinedMs <= memberJoinedMs
```
taoeffect: Note: `groupMembersSorted` must also be moved to `state.js`. Notice it uses `Date.now()` below. That's a clear sign that dynamic code/data is being used — data not derived from the contract messages — so this entire function must be moved out of this contract getter.

test/cypress/integration/group-paying.spec.js
sandrina-p: It cannot be commented out because the code is still broken and it would fail the tests. That's for another PR (I'll update #763 to address that bug)

frontend/model/state.js
sandrina-p: How was it before? I'm not understanding what's wrong...

taoeffect: Look at the previous code that was deleted in Payments.vue.  e.g.

```js
    paymentsSent () {
      const monthlyPayments = this.groupMonthlyPayments
      const payments = []
      // TODO: support sort direction
      for (const monthstamp of Object.keys(monthlyPayments).sort()) {
        const { paymentsFrom } = monthlyPayments[monthstamp]
        if (paymentsFrom) {
          for (const toUser in paymentsFrom[this.ourUsername]) {
            for (const paymentHash of paymentsFrom[this.ourUsername][toUser]) {
              const { data, meta } = this.currentGroupState.payments[paymentHash]
              const payment = {
                hash: paymentHash,
                data,
                meta,
```

frontend/model/state.js
sandrina-p: Hum... can we leave as it is? This avoids having to many "nested conditions"

frontend/model/state.js
sandrina-p: But we don't use it everywhere in this scenario, that's what I meant...

taoeffect: Without passing in the hash it will be impossible to access information about the original payment.

frontend/model/state.js
sandrina-p: It's done at `Payments.vue` line ~210. I didn't do it here because at `MonthOverview` we need only the TODO payments without the not received.

frontend/model/state.js
sandrina-p: It's hard to explain TBH :/. My suggestion:
- pull the code and revert this change
- run the tests "group-paying.spec.js" - `npm run cy:open` and see how the late payments are displayed incorrectly in the last test.
- Put the code back as it was: Now the late payments are displayed correctly.

frontend/model/state.js
taoeffect: I would prefer it be restored to how it was before, because it represents a functional style that's easier to follow and modify. For example, the previous version of this code makes it easier to add an `else { .. }` condition in a natural fashion. Having a single `return` for the function also makes it easier to reason about how and when the code will exit from its task. That is why I wrote the code that way, so I'd really prefer it to be how it was written originally if it's not a big deal for you.

frontend/model/state.js
taoeffect: ah OK that makes sense

frontend/model/state.js
sandrina-p: We need to know what payments are expected to be received, to be displayed at `monthOverview` (the sidebar) on the view of a receiver. [see figma screen](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=9518%3A21)

frontend/model/contracts/group.js
sandrina-p: Moving it...

frontend/model/state.js
sandrina-p: I understand. reverted

taoeffect: 🙇

taoeffect:
While running the tests for `group-paying.spec.js` (directly, without running any other tests), and without making any changes to the code, ran into this error, which is weird because the tests failed locally but passed on Travis...

Screen Shot 2020-07-01 at 5 53 00 PM

This is referring to the "BypassUI" button in the lower right. It was expecting "Feb 10" but saw "Feb 9" for some reason.

sandrina-p:
That's very weird @taoeffect. Can you tweak the data at `group-paying.spec.js` (one of the first lines) and see how it behaves?

taoeffect:
@sandrina-p I changed this line:

```js
const timeStart = Date.UTC(2020, 1, 10)
```

To:

```js
const timeStart = Date.UTC(2020, 1, 20)
```

And got this corresponding/related error:

Screen Shot 2020-07-03 at 8 32 18 PM

Are you unable to reproduce locally? Maybe we have different versions of Cypress? Maybe I forgot to install some plugin?

sandrina-p:
@taoeffect that's very strange. Make sure you have cypress 3.5.0 installed. There's no need to install any additional plugins.

I created [this tiny repository](https://github.com/sandrina-p/cypress-issue-clock-date) with 2 cypress clock tests. Please clone it, run the tests, as explained in the `README.md` and let me know the results. If it keeps failing, I'll open an issue to cypress. What's your OS and browser? Maybe there's some additional extension or so that you have installed...

taoeffect:
Thanks @sandrina-p, I'll give it a try, replied on Slack!

taoeffect:
@sandrina-p So indeed, this line is what's causing the problem:

```js
const timeStart = Date.UTC(2020, 1, 10)
```

The tests pass if I change it to:

```js
const timeStart = Date.UTC(2020, 1, 11)
```

So I'm guessing this is a timezone difference thing? I.e. it must be the difference in timezone between where you are and where I am, so on my computer, because it expects this:

```js
    cy.get('[data-test-date]').should('have.attr', 'data-test-date', 'Feb 10')
```

That is Feb 10th in my timezone... but Feb 10th is the 11th I think in UTC? Or something. That's my guess.

So somehow we need to make sure that the local timezone doesn't interfere with UTC.

EDIT: Here's a proof. In node via terminal:

```
> new Date(Date.UTC(2020, 1, 10)).toLocaleDateString("en-US", { month: 'short', day: 'numeric' })
'Feb 9'
> new Date(Date.UTC(2020, 1, 10)).toLocaleDateString("en-US", { month: 'short', day: 'numeric', timeZone: 'UTC' })
'Feb 10'
```

So the solution is to update the Cypress test to something along the lines of:

```js
    const dateString = new Date(timeStart).toLocaleDateString("en-US", { month: 'short', day: 'numeric' })
    cy.get('[data-test-date]').should('have.attr', 'data-test-date', dateString)
```

Notice we omit the `timeZone` variable so that it defaults to using the computer's local timezone, which is what the browser will show.

frontend/model/state.js
```diff
@@ -329,6 +335,232 @@ const getters = {
...
+    const cMonthstamp = currentMonthstamp()
+    const pDate = dateFromMonthstamp(cMonthstamp)
+    const dueIn = lastDayOfMonth(pDate)
+    const monthlyPayments = getters.groupMonthlyPayments
+    const allPayments = getters.currentGroupState.payments
+
+    const sent = (() => {
+      const payments = []
+      // QUESTION: Is this only from this month? What about the previous payments from months ago?
+      //           Same question for received.
```
taoeffect: Think I answered this question but the answer was removed? Does this still need answering or can the comment be removed?

(Quick answer again: we only keep track of the current month, and the previous month's payments in memory. After this payments are removed from RAM and should be archived. That archiving part hasn't been implemented yet.)

frontend/model/state.js
taoeffect: It doesn't make much sense to modify the code, because you've already modified it, and so I don't know what modifications to make...

For example, whereas before a received payment contained this information:

```js
                const payment = {
                  hash: paymentHash,
                  data,
                  meta,
                  username: fromUser,
                  displayName: this.userDisplayName(fromUser),
                  date: meta.createdDate,
                  amount: data.amount // TODO: properly display and convert in the correct currency using data.currencyFromTo and data.exchangeRate
                }
                if (this.filterPayment(payment)) {
                  payments.push(payment)
                }
```

Now in this code you're pushing far less information into the payments array:

```js
                const { data, meta } = allPayments[paymentHash]
                payments.push({ hash: paymentHash, data, meta })
```

So I don't know how to modify the line above that reads:

```js
              latePayments.push({
                username: payment.to,
                displayName: this.userDisplayName(payment.to),
                amount: payment.amount, // TODO: include currency (what if it was changed?)
                // partiaL: TODO add this info as it is in this.paymentsTodo
                isLate: true,
                date: dueIn
              })
```

That is something you must do, and then uncomment that code, and comment out the code below.

frontend/model/state.js
sandrina-p: Yes, It can be removed. I added that task to to the payments issue "Support fetching old payments (+1 month ago)"

frontend/model/state.js
sandrina-p: Hum, [here's the change needed](https://user-images.githubusercontent.com/14869087/86582878-443d3a00-bf7a-11ea-9eb6-7cc20d3b828b.png). I've done it in the last commit and pushed it. Note that now the last test at `group-paying.spec` will fail. To make it work again is just to revert the changes as in the screenshot

taoeffect: But of course they will fail, they're completely different data? When uncommented `...payment` isn't pushed into `latePayments`. Are you making use of the data that's going in?

sandrina-p: I'm not understanding what you mean. Maybe we should do a video call about this, it would be easier to clarify these stuff and try to move forward this PR

taoeffect (approved):
Alright, I think we can merge this, even though I didn't get a chance to do a full / perfect review and might've missed something, since we need to make it possible for you to move on!

PR #921, opened 2020-05-16
#920 AppLogs - QuotaExceededError on LocalStorage

### LINKED ISSUE #920: BUG: QuotaExceededError for localStorage
### Problem

While testing out Group Income I've been running into `QuotaExceededError`s. These errors are being triggered because the `localStorage` is running out of room. When I check `localStorage.length`, I get a value that exceeds what it should be:

```
> localStorage.length
7072
> localStorage.getItem('giConsole/john/count')
"167"
> localStorage.getItem('giConsole/john/limit')
"5000"
> localStorage.getItem('giConsole/john/markers')
"[]"
> localStorage.getItem('giConsole/john/markerNth')
"1000"
```

### Solution

Call `clearLogs` when the user log in, and clear the entire logs for `giConsole`, not just the logs for the current user.

The problem in this case was that there were logs for other users stored as well, that for some reason weren't cleared properly (I guess because the "Logout" code was never called). So we want to clear the entire localStorage for all of `giConsole` upon login.

E.g. replace:

```js
if (key.indexOf(`giConsole/${username}`) >= 0)
```

With:

```js
if (key.indexOf('giConsole') >= 0)
```

And make sure `clearLogs` is called on login.

sandrina-p:
Fixes #920

### DIFF
--- a/frontend/model/captureLogs.js
+++ b/frontend/model/captureLogs.js
@@ -22,7 +22,6 @@ let entriesCount = null
 let appLogsFilter = []
 
 // LS = Local Storage
-// TODO: handle QuotaExceededError
 const giLSset = (key, value) => localStorage.setItem(`giConsole/${username}/${key}`, value)
 const giLSget = (key) => localStorage.getItem(`giConsole/${username}/${key}`)
 const giLSremove = (key) => localStorage.removeItem(`giConsole/${username}/${key}`)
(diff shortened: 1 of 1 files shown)


pieer (approved):

taoeffect (approved):
Great work!

Issue #637, opened 2019-08-27
Bring back low-contrast colors


taoeffect:
### Problem

Currently in Figma we have High Contrast A11y-compatible colors enabled as the default. For those who don't need high-contrast colors, these don't look as good as the colors we used to have.

### Solution

Make it so that by default Figma shows the regular, non-high contrast colors, and add a setting in the Appearance settings to enable high-contrast for all color themes (and create two versions for each color theme, a low-contrast and a high-contrast).

Alternatively, create only low-contrast colors, and see if each platform we support has a sufficiently good High Contrast mode that makes all colors on the screen high contrast.

I know for example both iOS and macOS have this feature, which may mean that @mmbotelho doesn't need to spend time creating two kinds of colors for each theme.

cc @dotmacro

taoeffect:
Note also: people who require High Contrast colors might already have this system-wide setting enabled, in which case our already-High-Contrast-colors will be superfluous even to these users, and they'll end up using the low-contrast colors too, because upping the contrast on high-contrast just looks ugly.

mmbotelho:
> Note also: people who require High Contrast colors might already have this system-wide setting enabled, in which case our already-High-Contrast-colors will be superfluous even to these users, and they'll end up using the low-contrast colors too because upping the contrast on high-contrast just looks ugly.

I assume you are referring to the iOs high contrast setting. This feature does not work automatically. It increases the contrast in Apple's native apps (like iMessage, Settings, Contacts, etc...) but does not have an effect on third-party apps. It also does not affect web pages. It is the responsibility of the developers and designers of third-party apps and websites to provide accessible web pages. @sandrina-p maybe your input would be valuable here as well.

> these don't look as good as the colors we used to have.

I am not sure if this is what you are suggesting (it's not very clear from reading the issue), but I don't think going back to the old colors is a good idea. I understand that the ones we have now are very dark, but the ones before were extremely light. Maybe finding a middle ground would be best?

FYI - here are the colors we had before and their respective contrast ratios:

<img width="361" alt="Captura de ecrã 2019-08-28, às 19 20 19" src="https://user-images.githubusercontent.com/22004566/63885945-f30f0200-c9c8-11e9-9158-1d60bab9e05d.png">
<img width="400" alt="Captura de ecrã 2019-08-28, às 19 22 17" src="https://user-images.githubusercontent.com/22004566/63886148-5d27a700-c9c9-11e9-8435-064e8ea691f9.png">
[...]

PR #922, opened 2020-05-18
Close #844: Implement updates to the User Settings page

### LINKED ISSUE #844: Implement updates to the User Settings page
### Problem

In order to address the issues #825 and #637 , I have made some changes to the **User Settings / Appearance** page. Here's how it looks now:

user-settings

### Solution

Here are the changes made:
- Updated the theme thumbnails;
- Added the option to reduce motion;
- Added the option to enable high-contrast colors, for users with low vision. _Note: ATM, it's not clear what these colors will look like, so for now, the toggle will not do anything._ 

[Here are the Figma mockups fo the page.](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=9550%3A0)

Here's how it looks now:
Captura de ecrã 2020-02-21, às 14 02 43

pieer:


frontend/assets/style/components/_forms.scss
```diff
@@ -418,22 +418,22 @@ p.error {
...
+      font-size: 0.625rem;
       position: absolute;
-      width: 6px;
-      height: 6px;
-      top: 9px;
-      left: 5px;
+      width: 0.375rem;
+      height: 0.375rem;
+      top: 0.5625rem;
+      left: 0.3125rem;
```
sandrina-p: In this component, let's keep the px. Using `rem` doesn't work well because things start being misaligned :/

frontend/views/containers/user-settings/UserSettingsModal.vue
```diff
@@ -57,7 +57,7 @@ export default {
 
 .settings-container {
   position: relative;
-  margin: 0 26px;
+  margin: 0 1.625rem;
```
sandrina-p: `1.5rem` (24px) should work fine too, to follow DS spacing guidelines.

mmbotelho (changes requested):
@pieer this is looking great! Looks good in smaller devices as well. The only thing I'm noticing is that the slider is behaving a bit weird. Here's what happened:

1. Started server, created a group;
2. User settings, played around with the slider;
3. Restarted the server (my attempt to go back to the default value 😂);
4. When I got to the appearance page again, the slider was on the smallest possible font;
5. Changed it to something bigger and then set it to small again;
6. The small is now way smaller then it was on step 4

I'm using chrome btw! I have 2 screenshots here that hopefully make it clearer to see - these images have different text sizes even though the setting is the same:
Captura de ecrã 2020-05-19, às 16 46 01
Captura de ecrã 2020-05-19, às 16 39 38

### Other notes
- I like your suggestion of adding a delay to apply the setting! Can we do that?
- Can we make the clickable area of the slider a bit bigger? The slider should not increase in size, but if we can increase the clickable area (vertically) that would be great!
- The default font size is not the corrrect one. There's only one smaller option than the default, and 2 bigger options:
Captura de ecrã 2020-05-19, às 16 58 06



### Style:

Captura de ecrã 2020-05-19, às 16 27 12

frontend/views/containers/navigation/Profile.vue
```diff
@@ -98,7 +98,7 @@ export default {
 
 button {
   .icon-cog {
-    font-size: 0.9rem;
+    font-size: $size_4;
```
sandrina-p, line 101: These tiny adjustments break the UI. You'll notice that now the icon cog is not centered and when animated, it doesn't stay in the center either.

frontend/views/containers/user-settings/Appearence.vue
```diff
@@ -1,45 +1,60 @@
...
-      label.checkbox
-        input.input(type='checkbox' :checked='$store.state.reducedMotion' @change='handleCheckbox')
-        i18n Reduced motion
-      i18n.help(tag='p') When enabled the amount of animations you see around are reduced.
-
-    section.card
-      h2.is-title-3.c-title Text settings
+      .c-subcontent
+        .c-text-content
+          i18n.c-smaller-title(tag='h3') Use high-contrast colors
```
sandrina-p: This part could be commented out until we do #637. What do you think?

frontend/views/containers/user-settings/FontSize.vue
```diff
@@ -39,26 +39,24 @@ export default {
...
       // Add data to the slider
       this.fontData.push(size)
-      size += 2.5
+      size += 2
 
       // Hise one label out of two
       this.fontRange.push({
-        label: 'Aa',
-        isHide: index % 2
+        label: 'Aa'
```
sandrina-p, line 57: It's odd that when clicking in a white space somewhere in the input range, that triggers the font-size change. IMO it should only change the font-size when we click exactly at "aA" or at the input range bar.

sandrina-p (changes requested):

frontend/assets/style/components/_forms.scss
pieer: Unfortunately if we leave some element in px, they will not scale with the rest

sandrina-p: I see. It's a tradeoff. But I guess that not scaling a radio button isn't a big deal.

mmbotelho (dismissed):
Perfect, thank you @pieer !

mmbotelho:
There's one small detail missing (not a blocking issue!). The default font-size should be highlighted in blue (even when the user selected something else):

Captura de ecrã 2020-05-21, às 18 38 05

frontend/views/containers/user-settings/UserSettingsModal.vue
```diff
@@ -1,43 +1,25 @@
...
-        tab-placeholder(name='Privacy & Safety')
-      tab-item
-        tab-placeholder(name='Notifications')
-      tab-item
-        appearence
-      tab-item
-        app-logs
-      tab-item
-        troubleshooting
+    tab-wrapper(
```
sandrina-p, line 4: While testing this, I noticed 2 things in this modal:

1. Card spacings are not correct. (mobile it's missing right margin and desktop the card is too large)
2. The animations work fine but I'm afraid that they might be too "expressive" because there are two elements doing the slide up/down in different timings. Maybe a simple fade-in would work better.

sandrina-p (commented):
Great work in simplifying User settings modal!

mmbotelho (dismissed):
Approved, thank you! 🎉

sandrina-p (changes requested):
Looks good!

sandrina-p (approved):

sdegutis (approved):
Looks good to me, great job.

PR #923, opened 2020-05-19
#538 better avatar placeholders

### LINKED ISSUE #538: Implement improved avatar placeholder image
### Problem

captura de ecra 2019-02-28 as 17 11 28

The current avatar placeholder does not match the UI of the app. Also, because we only have one placeholder, and if users don't upload their own avatars, the interface can become very dull in places where a lot of avatars are displayed together.

E.g.: Using a chat where all avatars look the same is actually more harmful than not having the avatars there at all.

### Solution

We can create a series of avatars that are attributed randomly to users, making it easier to distinguish users that haven't uploaded their own avatars yet. This is also a great opportunity to use more gender and race neutral images. 

If you want to read more about the fascinating world of avatar placeholders, I recommend [this article](https://uxdesign.cc/design-avatars-that-make-sense-and-be-more-inclusive-in-the-process-d4dd6a486ea6).

### LINKED ISSUE #891: Implement new avatar placeholders
## Problem

The problem stated in #538.

## Solution

There should be 3 different avatar placeholders:

1. **For members**;
2. **For groups**;
3. **For members that have been accepted (via proposal) into the group but have NOT used their invite link yet.**

------------

### 1. For members
We are using Blockies, an identicon generator. You'll need these 2 links:
- [Blockies GitHub repository](https://github.com/download13/blockies);
- [Codepen with the correct configuration](https://codepen.io/mmbotelho/pen/OJVeVxo).

You will notice, at the bottom of the JS file of the codepen linked above, there are 6 configurable options. These are:

1. `seed` - a string, that should be unique for each user. This string is what generates the image;
2. `bgcolor`;
3. `color`;
4. `spotcolor`;
5. `size` - set it to 6;
6. `scale` - used 12;

**Colors**
Each user should be assigned one of our colors (blue, orange, red or green). After that, the 3 color options should be:

1. `bgcolor` - variant 0 (darkest);
2. `color` - variant 1 (medium);
3. `spotcolor`- variant 2 (lightest);

Here are some examples of what they look like:
Captura de ecrã 2020-04-11, às 19 03 38

------

### 2. For groups
For groups, we use the groups initials. We have a minimum of 1 character and a maximum of 2. Here's what they look like:
Captura de ecrã 2020-04-11, às 19 05 04

------

### 3. Members that have been accepted (via proposal) into the group but have NOT used their invite link yet.

Same as for groups, but with a question mark. Here's an example:
Captura de ecrã 2020-04-11, às 19 07 06

------

### Other useful links:

- [Figma Design System with specifications](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1876%3A17778);
- [Example page with placeholder avatars](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=11556%3A97);

And that's it! I hope I didn't miss anything, but let me know if you have any questions or if you find other cases / issues with this approach!

sandrina-p:
Closes #538, closes #891.

image

- The user avatar placeholder it's generated during the signup. A hidden image is inserted in the 
`<form>` DOM (generated by `blockies`). You can see it by removing the inline `display:none;` ([see screenshot](https://user-images.githubusercontent.com/14869087/82350489-575c7080-99f3-11ea-8574-5e00be1e944b.png)). This image is then uploaded/saved when the user submits the form. (Same for joining a group or creating an account)
 
- When creating a group, its avatar placeholder is dynamically generated based on group name input value. [See demo](https://user-images.githubusercontent.com/14869087/82346145-0302c200-99ee-11ea-873f-41dd03368678.gif)

### Scouting Rule
- Fix a bug on group getter `groupMembersByUsername`. It was returning all members, including removed ones, creating bugs around the app. E.g.: A proposal's status "1 out of X members voted.". Now the getter includes only the active members.


frontend/views/components/AvatarGenerator.vue
```diff
@@ -0,0 +1,48 @@
+<template lang='pug'>
+  // Leave in the DOM for debug
```
pieer, line 2: Do you want to keep this now?

frontend/views/components/AvatarUpload.vue
```diff
@@ -105,7 +105,7 @@ export default {
     }
   }
 
-  &-img.is-xl {
+  &-img.is-xl.is-xl { // need .is-xl 2x for specificity
```
pieer, line 108: This is very confusing

pieer (dismissed):
Excellent work, just left optionnal comments

frontend/views/components/AvatarGenerator.vue
sandrina-p: At least during CR so @mmbotelho can see it too easily :)

frontend/views/components/AvatarUpload.vue
sandrina-p: Yes, indeed. without the 2nd class, these rules are not strong enough. Maybe you could take a look and suggest another way (besides `!important`)

mmbotelho (changes requested):
Here's what I found:
Captura de ecrã 2020-05-21, às 14 56 55
Captura de ecrã 2020-05-21, às 14 58 32
Captura de ecrã 2020-05-21, às 15 04 49

mmbotelho:
Nevermind the green avatar comment! Just saw one! 😄

sandrina-p:
@mmbotelho,
- The avatar bug already existed in master (when selecting an image). Now it should work with both methods (custom image and group initials) 
- Image resolution increased to 256px.
- The avatar color is random, you just had bad luck on your first try :D

mmbotelho (dismissed):
Approved! Thank you @sandrina-p !

frontend/views/components/AvatarUpload.vue
pieer: I never used that hack! But this is really !important ;)

pieer (dismissed):

sdegutis (dismissed):
Looks good to me.

frontend/controller/actions/group.js
```diff
@@ -27,13 +27,13 @@ export default sbp('sbp/selectors/register', {
...
   } = {}) {
-    let finalPicture = `${window.location.origin}/assets/images/default-group-avatar.png`
+    let finalPicture = `${window.location.origin}/assets/images/group-avatar-default.png`
 
     if (picture) {
       try {
         finalPicture = await imageUpload(picture)
       } catch (e) {
-        console.error('Failed to upload the group picture', e)
+        console.error('actions/group.js to upload the group picture', e)
```
taoeffect: `'actions/group.js to upload the group picture'` => `'actions/group.js: failed to upload the group picture'`

frontend/controller/actions/identity.js
```diff
@@ -17,13 +19,25 @@ export default sbp('sbp/selectors/register', {
...
+    let finalPicture = `${window.location.origin}/assets/images/user-avatar-default.png`
+
+    if (picture) {
+      try {
+        finalPicture = await imageUpload(picture)
+      } catch (e) {
+        console.error('actions/identity.js picture upload error:', e)
+        throw new GIErrorUIRuntimeError(L('Failed to upload the profile picture. {codeError}', { codeError: e.message }))
+      }
+    }
```
taoeffect, line 32: If they don't provide a picture, wouldn't this be a great place to generate it?

frontend/utils/image.js
```diff
@@ -1,7 +1,19 @@
...
+// Copied from https://stackoverflow.com/a/27980815/4737729
+export function b64toBlob (dataURI) {
+  var byteString = atob(dataURI.split(',')[1])
+  var ab = new ArrayBuffer(byteString.length)
+  var ia = new Uint8Array(ab)
+
+  for (var i = 0; i < byteString.length; i++) {
+    ia[i] = byteString.charCodeAt(i)
+  }
+  return new Blob([ab], { type: 'image/jpeg' })
```
taoeffect, line 15: Please copy this part of from the StackOverflow answer into a comment between lines 4 and 5:

```
// Accepts only JPEGs and assumes dataURI begins with: "data:image/jpeg;base64,"
```

It's not clear to me that it should work if, for example, a PNG is used. Even if it does work, the type listed here and the image type actually provided should match.

Also, since `b64toBlob` sounds like it will accept any Base64 text and convert it to a Blob, this function should be renamed since that is not what it does. This function takes a JPEG data URI and converts it to a blob, so a better name would be something like: `jpegDataURItoBlob`.

Alternatively, if you parse the `dataURI` for the image type and use it here, you can give the function a more generic name, like `imageDataURItoBlob`

frontend/views/components/group-creation-steps/GroupName.vue
```diff
@@ -51,11 +50,51 @@ export default {
...
+      this.$refs.pictureAvatar.setFromBlob(this.$assistant.ephemeral.groupPictureFile)
+    } else {
+      this.updatePictureCanvas(this.groupInitials)
+    }
+  },
+  beforeDestroy () {
+    if (!this.$assistant.ephemeral.groupPictureFile) {
+      const pictureBase64 = this.$refs.pictureCanvas.toDataURL('image/png')
+      this.$v.form.groupPicture.$touch()
+      this.$assistant.ephemeral.groupPictureFile = b64toBlob(pictureBase64)
```
taoeffect: Here is an example of where naming the function correctly would have caught the bug, since `pictureBase64` is a dataURI that uses `image/png` and `b64toBlob` only handles creating `Blob` objects that are of `{type: 'image/jpeg'}`.

Even if it mysteriously still works, it might not work in all browsers/environments, so we should make sure that the encoding type is the same as the data source type.

frontend/views/components/AvatarGenerator.vue
```diff
@@ -0,0 +1,48 @@
...
+  div(style='display: none;')
+    img(:src='ephemeral.avatarBase64')
```
taoeffect, line 4: It doesn't make sense to me why this code is inside of a Vue Component that does nothing to display anything. If the entire purpose of this component is to call `blockies.create`, then I think we only need a function. It could be called via `sbp` inside of SignupForm.

frontend/views/components/group-creation-steps/GroupName.vue
```diff
@@ -51,11 +50,51 @@ export default {
...
+      this.$assistant.ephemeral.groupPictureType = 'canvas'
+    }
+  },
+  computed: {
+    groupInitials () {
+      return this.group.groupName ? this.group.groupName.match(/\b(\w)/g).join('').slice(0, 2).toUpperCase() : ''
```
taoeffect, line 81: Nicely done 👍

frontend/controller/actions/identity.js
```diff
@@ -17,13 +19,25 @@ export default sbp('sbp/selectors/register', {
       console.warn(`deleting settings for pre-existing identity ${username}!`, oldSettings)
       await sbp('gi.db/settings/delete', username)
     }
+
+    let finalPicture = `${window.location.origin}/assets/images/user-avatar-default.png`
```
taoeffect, line 23: If they don't provide a picture, would this be a good place to generate it instead of in the SignupForm?

Then `user-avatar-default.png` wouldn't be needed ...

taoeffect (changes requested):
Review finished! Great job @sandrina-p! Minor things noticed, and a possible/potential bug.

frontend/utils/image.js
sandrina-p: Gotcha. Renamed to `imageDataURItoBlob` and added a TODO for later to support other formats :)

frontend/views/components/AvatarGenerator.vue
sandrina-p: Yes, it could be adapted to a normal function, but I found it very useful to be inside a Vue component to be more "vue-sh format". I also find it very handy to be the image in the DOM for debug (or even tests) purposes. I'll leave it as it is to be quicker, feel free to change it in a next MR

frontend/controller/actions/identity.js
sandrina-p: This is used in case, for some reason, `blockies` fail for some reason. That way we still have a very basic default avatar. This is called _graceful degradation_

sandrina-p:
@taoeffect changes made. This can be merged in any order.

taoeffect:
Merging in favor of newly opened issue #981 which can continue this PR!

Issue #925, opened 2020-05-20
Implement archiving of proposals


taoeffect:
### Problem

We don't have archiving of proposals implemented.

#### Old Solution

Copying from Slack:

> @Sandrina ok yeah, just looked at the dashboard. The designs don't specify how/when the proposals should be moved to the "All proposals" modal. I think there should be a button to clear/remove/archive the passed proposals. And in the code itself, it has to be implemented in a very specific way with respect to the data. E.g., when a proposal passes in the contract, it is immediately removed from the contract state, and a side effect adds it to the app's vuex state outside of the contract state to allow it to still be displayed in the dashboard. Then when the clear/archive button is clicked, it's removed from there too and placed only in IndexedDB (where it already exists, but you look it up by hash).

This issue is related to #924 — Design clear proposals button

The reason for immediately removing it from the contract state is also related to #813  — consensus, since it would break consensus for proposals to be removed from the contract state at different times by different users.

#### New Solution

See @leihla's comments here for how it should work:

- https://github.com/okTurtles/group-income/issues/924#issuecomment-1155064163
- https://github.com/okTurtles/group-income/issues/1233#issuecomment-1121157731

taoeffect:
See @leihla's comment here for new solution on how archiving of proposals should work: https://github.com/okTurtles/group-income/issues/924#issuecomment-1155064163

PR #929, opened 2020-05-22
#664 Invite links - Implement "Revoke link" action

### LINKED ISSUE #664: Implement invite management section (group settings page)

sandrina-p:
Closes #664.

screenshot

### Things done:
- The menu item "See original proposal" was removed. It will be added back at #925.
- The "time left" now shows the correct timing. Note: all invitations still have the same static expiration date. (4 December 2021)
- The revoke action works.
- Simplified the "Join Group" page. When visiting a revoked/expired/incorrect link, it shows a message saying the invite is invalid.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -50,7 +50,7 @@ export function createInvite (
     invitee,
     status: INVITE_STATUS.VALID,
     responses: {}, // { bob: true } list of usernames that accepted the invite.
-    expires: 1638588240000 // 04 december 2021.
+    expires: 1638588240000 // 04 december 2021. // TODO this
   }
 }
 
@@ -603,7 +603,6 @@ DefineContract({
         }
       },
       process ({ data, meta }, { state, getters }) {
-        console.debug('inviteRevoke:', data, state.invites)
         const invite = state.invites[data.inviteSecret]
         Vue.set(invite, 'status', INVITE_STATUS.REVOKED)
       }
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -50,7 +50,7 @@ export function createInvite (
     invitee,
     status: INVITE_STATUS.VALID,
     responses: {}, // { bob: true } list of usernames that accepted the invite.
-    expires: 123456789 // TODO: this
+    expires: 1638588240000 // 04 december 2021.
   }
 }
 
@@ -590,6 +590,22 @@ DefineContract({
           // new member has joined, so subscribe to their identity contract
           await sbp('state/enqueueContractSync', meta.identityContractID)
         }
+      }
+    },
+    'gi.contracts/group/inviteRevoke': {
+      validate: (data, { state, getters, meta }) => {
+        objectOf({
+          inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
+        })(data)
+
+        if (!state.invites[data.inviteSecret]) {
+          throw new TypeError(L('The link does not exist.'))
+        }
+      },
+      process ({ data, meta }, { state, getters }) {
+        console.debug('inviteRevoke:', data, state.invites)
+        const invite = state.invites[data.inviteSecret]
+        Vue.set(invite, 'status', INVITE_STATUS.REVOKED)
       }
     },
     'gi.contracts/group/updateSettings': {
--- a/frontend/views/containers/group-settings/InvitationsTable.vue
+++ b/frontend/views/containers/group-settings/InvitationsTable.vue
@@ -48,14 +48,23 @@ page-section.c-section(:title='L("Invite links")')
             i.icon-ellipsis-v
           menu-parent.c-webshare-fallback
             menu-trigger.is-icon-small.c-fallback-trigger(ref='webShareFallbackBtn')
-
             menu-content.c-dropdown-fallback
               ul
                 menu-item(
                   tag='button'
                   icon='link'
                 )
+                  // #780 TODO test/verify if webShare works in a real mobile device.
+                  //      And refactor this markup to only have one menu-parent
                   i18n Copy link
+                menu-item(
+                  v-if='item.status.isActive'
+                  tag='button'
+                  item-id='revoke'
+                  icon='times'
+                  @click='handleRevokeClick(item.inviteSecret)'
+                )
+                  i18n Revoke Link
         td.c-state
           span.c-state-description {{ item.description }}
           span.c-state-expire(
--- a/frontend/views/containers/group-settings/InvitationsTable.vue
+++ b/frontend/views/containers/group-settings/InvitationsTable.vue
@@ -24,7 +24,7 @@ page-section.c-section(:title='L("Invite links")')
         th.c-action(
           :aria-label='L("action")'
         )
-    tbody
+    transition-group(name='slidedown' tag='tbody')
       tr(
         v-for='(item, index) in invitesToShow'
         :key='item.inviteSecret'
@@ -128,7 +128,9 @@ export default {
         selectbox: {
           focused: false,
           selectedOption: 'Active'
-        }
+        },
+        // keep invite in "Active" list for a few seconds after being revoked
+        inviteRevokedNow: null
       }
     }
   },
@@ -149,7 +151,7 @@ export default {
         .filter(invite => invite.creator === INVITE_INITIAL_CREATOR || invite.creator === this.ourUsername)
         .map(this.mapInvite)
       const options = {
-        Active: () => invitesList.filter(invite => invite.status.isActive),
+        Active: () => invitesList.filter(invite => invite.status.isActive || (invite.status.isRevoked && invite.inviteSecret === this.ephemeral.inviteRevokedNow)),
         All: () => invitesList
       }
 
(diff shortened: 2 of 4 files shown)


pieer (dismissed):

mmbotelho (commented):
This seems to be working perfectly @sandrina-p ! Good job!

I have one suggestion to improve the UX of this action. When the user clicks "Revoke this link", can we show the revoked table row there for a few moments (maybe 1 or 2 seconds) and then fade it out? Just so they get clear feedback that the action was successful and that it was removed from the active links list accordingly?

Let me know if we can do this on this PR or if you want us to leave it for later, so I can approve it or wait for this change 😄

sandrina-p:
@mmbotelho nice idea, it's done! Now the revoked link is kept in the list for 2s until it slides up (smoother than fade out)

gif

mmbotelho (changes requested):
Perfect @sandrina-p ! Thank you for implementing that in this PR!

I found one other issue:
Captura de ecrã 2020-05-24, às 16 18 45

sandrina-p:
@mmbotelho, done!

mmbotelho (dismissed):
Approved, thank you!

frontend/model/contracts/group.js
```diff
@@ -50,7 +50,7 @@ export function createInvite (
     invitee,
     status: INVITE_STATUS.VALID,
     responses: {}, // { bob: true } list of usernames that accepted the invite.
-    expires: 123456789 // TODO: this
+    expires: 1638588240000 // 04 december 2021.
```
taoeffect: The `TODO: this` comment still needs to be there, because `TODO`s are used to create future issues.

frontend/model/contracts/group.js
```diff
@@ -592,6 +592,22 @@ DefineContract({
...
+        objectOf({
+          inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
+        })(data)
+
+        if (!state.invites[data.inviteSecret]) {
+          throw new TypeError(L('The link does not exist.'))
+        }
+      },
+      process ({ data, meta }, { state, getters }) {
+        console.debug('inviteRevoke:', data, state.invites)
```
taoeffect: Extraneous debug comments not needed here. All of the existing SBP logging will automatically log this information anyway.

frontend/views/pages/Join.vue
```diff
@@ -86,7 +75,7 @@ export default {
     pageStatus: {
       get () { return this.ephemeral.pageStatus },
       set (status) {
-        const posibleStatus = ['LOADING', 'WELCOME', 'SIGNING', 'LOGGING', 'EXPIRED', 'INVALID']
```
taoeffect, line 89: What happened to expired links?

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -145,42 +184,47 @@ export default {
...
+      remainder = timeLeft % MIL_DAY
       const hours = Math.floor(remainder / MIL_HR)
       remainder = remainder % MIL_HR
       const minutes = Math.ceil(remainder / MIL_MIN)
 
       if (days) return L('{days}d {hours}h {minutes}m left', { days, hours, minutes })
       if (hours) return L('{hours}h {minutes}m left', { hours, minutes })
       if (minutes) return L('{minutes}m left', { minutes })
-      return L('Expired')
+      return L('Not used')
```
taoeffect: `'Expired'` makes more sense to me and is consistent with the rest of the code.

taoeffect (changes requested):
Partial review!

frontend/views/pages/Join.vue
sdegutis, line 89: @sandrina-p It looks like this PR merges the concepts of expired/invalid invites, is that right?

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -124,10 +137,36 @@ export default {
...
+      'currentGroupId'
+    ]),
+    invitesToShow () {
+      const { invites } = this.currentGroupState
+
+      if (!invites) { return [] }
+
+      const invitesList = Object.values(invites)
+        .filter(invite => invite.creator === INVITE_INITIAL_CREATOR || invite.creator === this.ourUsername)
+        .map(this.mapInvite)
```
sdegutis, line 161: I think `this.mapInvite` needs to be bound to this with `.bind(this)` first, otherwise `this` in the method will not be bound to the right object.

sdegutis:
If you revoke an invite, and then a person uses it, it seems to work, by showing the "Welcome to {group}" screen. But then that person's links are all broken, and the console has tons of errors. And the person isn't really added to the group, according to what people see who are added in the group. (I'm not sure yet what happens in the database, still investigating that.)

frontend/model/contracts/group.js
```diff
@@ -592,6 +592,22 @@ DefineContract({
...
+        const invite = state.invites[data.inviteSecret]
+        Vue.set(invite, 'status', INVITE_STATUS.REVOKED)
```
sdegutis, line 607: We probably need to change `inviteGuard` to check if status is revoked.

frontend/views/containers/group-settings/InvitationsTable.vue
sdegutis: Then again it seems to work, so Vue probably binds `this` automatically in methods. In which case, disregard my last comment on this line :)

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -189,39 +233,41 @@ export default {
...
+        this.ephemeral.inviteRevokedNow = inviteSecret
+        const message = await sbp('gi.contracts/group/inviteRevoke/create', {
+          inviteSecret
+        }, this.currentGroupId)
+        await sbp('backend/publishLogEntry', message)
+        setTimeout(() => {
+          this.ephemeral.inviteRevokedNow = null
+        }, 2000)
+      } catch (e) {
+        this.ephemeral.inviteRevokedNow = null
```
sdegutis, line 267: Does this also hide the invite-link entry? If so, does it make sense to keep it in the list so the user can try again after the issue is resolved, without having to refresh (at which point it will be there again)?

sdegutis (changes requested):
Looks good, just have a few requests as marked above. Thanks again.

frontend/views/pages/Join.vue
sandrina-p: The layout is the same for expired an invalid links to it was a redundant status.

frontend/views/containers/group-settings/InvitationsTable.vue
sandrina-p: Changed
cc @mmbotelho

frontend/views/pages/Join.vue
sandrina-p: Exactly, it was asked by @mmbotelho. For further details you can reach her, I confess my memory is foggy 4 months later after I did this.

frontend/model/contracts/group.js
sandrina-p: Maybe yes. You can create an issue and tackle that after. Shouldn't be hard!

frontend/views/containers/group-settings/InvitationsTable.vue
sandrina-p: No, it does not because when failed, the invite won't turn `revoked`. So the `inviteRevokeNow` does not have any effect. 

Here's the condition again, the last comparison guarantees that.

```
Active: () => invitesList.filter(invite => invite.status.isActive || (invite.status.isRevoked && invite.inviteSecret === this.ephemeral.inviteRevokedNow)),
```

sandrina-p:
@taoeffect and @sdegutis changes made based on your reviews.

sdegutis:
Looks good to me, thanks @sandrina-p.

taoeffect (approved):
Approved! I didn't fully review this but I trust either @sdegutis found all the bugs there were to find or we shall discover them later! 😄

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -189,39 +233,41 @@ export default {
...
       }
 
-      return options[this.ephemeral.selectbox.selectedOption]()
+      try {
+        this.ephemeral.inviteRevokedNow = inviteSecret
+        const message = await sbp('gi.contracts/group/inviteRevoke/create', {
+          inviteSecret
+        }, this.currentGroupId)
+        await sbp('backend/publishLogEntry', message)
+        setTimeout(() => {
```
sdegutis, line 263: There's a very unlikely but real race condition here. If an exception is thrown, which sets `inviteRevokedNow` to null, and the user takes an action that sets it to non-null within these 2 seconds, then the timer created on this line is going to override the user's action by setting it to null. To solve it we'd just have to store this timer in a variable, and cancel it in the catch block.

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -189,39 +233,41 @@ export default {
...
-      const invitesList = Object.values(invites)
-        .filter(invite => invite.creator === INVITE_INITIAL_CREATOR || invite.creator === this.ourUsername)
-        .map(this.mapInvite)
-      const options = {
-        Active: () => invitesList.filter(invite => invite.status.isActive),
-        All: () => invitesList
+    },
+    async handleRevokeClick (inviteSecret) {
+      if (!confirm(L('Are you sure you want to revoke this link? This action cannot be undone.'))) {
+        return null
```
sdegutis, line 254: Is it necessary to return null here?

sdegutis (commented):
Just two more small things noticed in this PR. One is a potential race condition, and the other is a very minor concern about a possibly superfluous return value.

frontend/views/containers/group-settings/InvitationsTable.vue
taoeffect: The only thing that can throw an exception though is on lines 259 and 262, right? So if either of those lines threw an exception we wouldn't get to line 263, and if we did get to this line it implies no exception was thrown...

sdegutis: Oh good point. Oops.

Issue #663, opened 2019-09-12
Notification system [UX / UI Discussion]


mmbotelho:
#  Problem

At the moment, there is no notification system in the app. We need this in order to alert users about a variety of events, like a new proposal that needs to be voted on or a new member that has joined or left the group (among others).

# Solution

Create a notification system, that users can easily access by clicking on the notification bell 🔔 on the upper left side of the sidebar. 

### 1. What is a notification? 
A notification belongs in the notification bell if **all the following statements are true:**
- The notification can be linked to a place in the app where the user can take action or read more about it;
- Time is relevant to the information provided, either because the date when it happened is relevant OR because there's a time limit to act. _E.g.: when a new proposal is open, being notified in time is very important, so that the user can vote on it before it expires._
- There is an **Agent** (either a user, the group itself or the app), an **Event** and an **Action**. These can be implicit or explicit, depending on the context. _E.g.: "Margarida (agent) wants to add Sandrina to your group (Event). Vote now! (Action)"_

### 2. What states can a notification have?
A notification can have the following states:
- **New** - means that the user was first aware of them less than one hour ago. Note that this does not mean that the notification was created less than one hour ago. Being "aware" of the notification means that they clicked on the bell icon 🔔;
- **Unread** - means that the user has not clicked on the notification itself (not the notification bell) yet;
- **Read** - means that the user has clicked on the notification.

A notification can have the following state combinations:
- **Unread** and **New**;
- **Read** and **New**;
- **Unread**;
- **Read**.

Here's an example of how this looks on the interface:
<img width="1217" alt="Captura de ecrã 2019-09-12, às 21 05 46" src="https://user-images.githubusercontent.com/22004566/64813201-1f1dac00-d5a1-11e9-86b0-f8c7582bc415.png">

Here's an example of read and unread notifications (without the **new** and **older** separation):
<img width="1169" alt="Captura de ecrã 2019-09-12, às 21 08 58" src="https://user-images.githubusercontent.com/22004566/64813419-95baa980-d5a1-11e9-8af6-6711018946be.png">

### 3. When should the notification bell have a red badge?
Whenever there are **one or more than one** New AND Unread notifications, the bell icon will have a red badge with the number of notifications.

Clicking the 🔔 should remove the red badge immediately, as shown below. If there are more notifications than vertical space to read then, the user needs to scroll through all of them for the red badge to disappear.

[...]

PR #934, opened 2020-05-30
#663 notifications - layout only

sandrina-p:
Starting #663

83329066-1a0fa280-a27f-11ea-9882-056ca1c59245

### What's done:
- All the layout with dummy data. Focus only on its visual aspect. Don't expect any info to be correct (e.g. names, avatars, dates, etc...) That's for another PR.

### Notes:
- For now, the dummy data is based on the number of members in a group:
  - 1 member: No notifications.
  - 2 members: 1 notification.
  - +2 members: 7 notifications.
- @mmbotelho I left some questions in Figma about the layout. Once answered, I'll update this PR.
- @mmbotelho, if you use the keyboard to open/navigate through the notification card, eventually the close button will show up (similar to the profile card). But its placement is not the best one. What do you suggest?
- Ignore all the code related to logic. That's for another PR.
- I think we should do two video calls about this topic:
  - Review each notification message in detail (copy, type, icons, etc...) and analyze possible technical restrictions for the prototype. 
  - Discuss the notification system architecture (logic, backend, etc...)

### Bugs found (to be fixed in other PR):
- Open the notifications card/modal, click "Settings". The User Settings modal won't open at the "Notifications" section.


mmbotelho (changes requested):
Here's what I found @sandrina-p :

Captura de ecrã 2020-06-10, às 17 40 24
Captura de ecrã 2020-06-10, às 17 42 04
Captura de ecrã 2020-06-10, às 17 55 17
Captura de ecrã 2020-06-10, às 18 20 57

mmbotelho:
> @mmbotelho I left some questions in Figma about the layout. Once answered, I'll update this PR.

I looked at Figma and I believe I answered everything. Let me know if I missed something!

> @mmbotelho, if you use the keyboard to open/navigate through the notification card, eventually the close button will show up (similar to the profile card). But its placement is not the best one. What do you suggest?

I suggest leaving it as it is (as a fallback) but having the ESC key closing the popup as well.

> Review each notification message in detail (copy, type, icons, etc...) and analyze possible technical restrictions for the prototype.
I'm not sure if you saw this, but [I made a table with all the messages I could think of](https://github.com/okTurtles/group-income-simple/issues/663#issue-492988756). Let me know what's missing! We can also schedule a call, whatever works best for you!

sandrina-p:
@mmbotelho 

- Extra gray spaces fixed. This was a tricky one to understand. It's the scrollbar, visible by default (I forget that this is the default setting on windows). Now, all scrollbars should be hidden when not needed.
- Alignments fixed.
- Yes, I saw that list and I still think it's incomplete or it should have a new review. For example "Pending payment confirmations" does not make sense anymore. And maybe some of these notifications should be similar to the [emails list](https://github.com/okTurtles/group-income-simple/issues/843)... Anyway, that's not for this PR. This is just dummy layout for now. Later, when @taoeffect has more time, we can talk about this.

frontend/views/components/modal/Modal.vue
```diff
@@ -110,7 +110,8 @@ export default {
         this.content = null
         this.contentProps = {}
         // Refocus on button that open the modal
-        this.lastFocus.focus()
+        // TODO/BUG It might not exist... (e.g. profile card was closed). What's the best approach then?
```
pieer: The focus should be on the button that open the modal. In case of modal opening a modal, meaning that the last focus element is not present, the best approach I guess is to focus on the button that open the first modal. We can do that by recording the previous focused element if the last one doesn't exist. Or a simpler solution is to go back to the first focusable element in the page

frontend/views/containers/notifications/NotificationsBell.vue
```diff
@@ -0,0 +1,72 @@
+<template lang='pug'>
+div
+  router-link.button.is-icon-small.c-btn.hide-tablet(:to='{ query: { modal: "NotificationsModal" }}')
+    i.icon-bell(:class='notificationsCount ? "" : "is-active"')
```
pieer: i.icon-bell(:class='{"is-active": notificationsCount}')

frontend/views/containers/notifications/NotificationsCard.vue
```diff
@@ -0,0 +1,140 @@
+<template lang='pug'>
+tooltip(
```
pieer, line 2: 👍

frontend/views/containers/notifications/NotificationsModal.vue
```diff
@@ -0,0 +1,99 @@
...
+    ModalBaseTemplate,
+    NotificationsList
+  },
+  data () {
+    return {
+      searchText: ''
+    }
+  },
+  computed: {
+    ...mapGetters([
```
pieer: Empty?

pieer (changes requested):

mmbotelho:
> Yes, I saw that list and I still think it's incomplete or it should have a new review. For example "Pending payment confirmations" does not make sense anymore.

@sandrina-p you're right! My bad, I forgot a lot of things changed since I wrote that list! Definitely needs to be reviewed

sandrina-p:
Okay, @mmbotelho. Later in the road, maybe after #916 and #918 are merged, we can discuss this again. ATM we are already stretching our "multi-tasking work" with too many tasks at the same time.

frontend/views/containers/notifications/NotificationsModal.vue
sandrina-p: Forgotten 🙈 Removed!

frontend/views/components/modal/Modal.vue
sandrina-p: Here the trick is the profile|notifications cards. For example, clicking "Remove Member" in the profile card, it opens the modal and closes the card. This means that `this.lastFocus` does not exist anymore. When we close the modal, it will go back to the top of the page because of that. But it should go back to the respective user in the members' list. 
I didn't find any clean way to fix this... but I'll keep in the back of my head

mmbotelho:
I agree @sandrina-p !

mmbotelho (changes requested):
Only 2 things missing:

Captura de ecrã 2020-07-20, às 14 49 46

frontend/views/containers/notifications/NotificationsList.vue
```diff
@@ -0,0 +1,250 @@
...
+    flex-grow: 1;
+    height: 1.5rem;
+    border-radius: 0.7rem;
+    background-color: $general_2;
+  }
+}
+
+.c-title {
+  display: block;
+  padding: 0 0 0.5rem 0.5rem;
```
sdegutis: I think changing the left padding here to 0 is enough to fix the alignment issue.

sdegutis (changes requested):
Found the fix to the alignment issue above. Otherwise, looks good!

frontend/model/notifications/templates.js
```diff
@@ -0,0 +1,118 @@
...
+// - avatar, name, linkTo, read
+export const MEMBER_ADDED = ({ timestamp, data }) => {
+  const name = data.username // getDisplayName()
+  const avatar = ''
+  return {
+    avatar,
+    icon: 'user-plus',
+    level: 'info',
+    body: L('The group has a new member. Say hi to {name}!', {
+      name: `<strong>${name}</strong>`
```
taoeffect, line 20: Note: created an issue for this: #975

frontend/views/containers/notifications/NotificationsList.vue
sandrina-p: yes it did. updated

taoeffect:
Closing in favor of squashed version #995. Review continues there and the PR will be picked up by @F1LT3R.

PR #938, opened 2020-06-04
#913 - Prevent InvitationsTable to break in case there are no invites

### LINKED ISSUE #913: Bug: exception in group settings on log out
### Problem

Screen Shot 2020-05-06 at 1 23 45 PM

Discovered while testing #905. Here are the steps to reproduce on `master` as well:

1. Create a new user
2. Create a new group, make sure to to fill out something in the group description
3. Click "Edit" link in Dashboard to edit the group description, this will bring up the group settings
4. In the `textarea` for "About this group", delete the group description so that it's empty, hit tab to exit the textarea and so that the red error *This field is required* appears, **but do NOT click "Save changes"!**
5. Click the wheel icon in the lower-left to bring up the user settings
6. Click the "Log Out" link

You should see the error appear in the console.

It points to line 214 in InvitationsTable.vue:

```js
computed: {
    ...mapGetters(['currentGroupState', 'ourUsername']),
    ...mapState(['currentGroupId']),
    invitesToShow () {
      const { invites } = this.currentGroupState
      const invitesList = Object.values(invites) // <-- error here
```

### Solution

Fix!

sandrina-p:
Closes #913

The error happens because the state is cleared on logout before redirecting the user to the homepage. The fix for this particular error is to prevent the `InvitationsTable` from breaking in case there are no invitations.

### DIFF
--- a/frontend/views/containers/group-settings/InvitationsTable.vue
+++ b/frontend/views/containers/group-settings/InvitationsTable.vue
@@ -210,6 +210,9 @@ export default {
     ...mapState(['currentGroupId']),
     invitesToShow () {
       const { invites } = this.currentGroupState
+
+      if (!invites) { return [] }
+
       const invitesList = Object.values(invites)
         .filter(invite => invite.creator === INVITE_INITIAL_CREATOR || invite.creator === this.ourUsername)
         .map(this.mapInvite)


taoeffect (approved):
Great job! Easy fix!

PR #940, opened 2020-06-06
#914 fix bug pressing enter to login

### LINKED ISSUE #914: Bug: pressing enter in Log In modal doesn't log user in
### Problem

After creating a user, logging out, and then using the login modal to log back in, if you press "Enter" while the cursor is inside the password field, you'll get this:

Screen Shot 2020-05-06 at 1 01 44 PM

### Solution

Bring back previous behavior where it will log you in.

sandrina-p:
Fix #914

Pressing enter wasn't submitting the form (login) because the first button inside the form is the recover password. A button inside a form needs `type="button"` to prevent being interpreted as a submit button.


taoeffect (approved):
Great job @sandrina-p !

Issue #927, opened 2020-05-21
Update confetti colors


mmbotelho:
### Problem

The confetti (group creation and when joining a group) still have the old GI colors.

<img width="723" alt="Captura de ecrã 2020-05-21, às 16 07 25" src="https://user-images.githubusercontent.com/22004566/82574206-a97dcd00-9b7e-11ea-9345-0b274be50251.png">

### Solution

Update the color values, both for the light and dark themes!

SebinSong:
@mmbotelho Hi Margarida,
Do you happen to have particular colorsets to use for each theme in mind?
(colors you want to apply to the confetti particle under each theme, also how do you want the gi logo to look?)

mmbotelho:
Hello @SebinSong ! Sorry it wasn't clear, my bad. 

The idea is to replace all the old colors with the new ones. I believe the change on your side should only be to use the css variables instead of hardcoded values? But I'm not sure, this is just a suggestion.

<img width="252" alt="Captura de ecrã 2020-06-08, às 10 13 17" src="https://user-images.githubusercontent.com/22004566/84013573-03e59e80-a971-11ea-813f-a881860215aa.png">

So, here's what should happen:
- **Old blue** - replace with **--primary_0**;
- **Old orange** - replace with **--warning_0**;
- **Old red** - replace with **--danger_0**;
- **Old green** - replace with **--success_0**;

The logo remains the same! Hope that's clearer now 😄

SebinSong:
@mmbotelho No it wasn't unclear at all. I was just wondering if you have particular colors in mind as the color expert of the project. 
Thanks for the detailed info. Since the svg is part of the html there in animation, I think it's absolutely possible to replace the color values with the the css variables. 
Let me start working on it soooon.

c.c @taoeffect

pieer:
Implemented by @SebinSong in #942

PR #942, opened 2020-06-11
Update confetti colors

SebinSong:
closes issue #927 

replacing raw color values with css-variables so it displays the **correct colors** under each theme.


sandrina-p (approved):
well done!

pieer (approved):

PR #900, opened 2020-04-18
Task/763 payments implementation

taoeffect:
Closes almost, but not entirely: #763 #206

PR #945, opened 2020-06-13
fix: Exclude inactive members from members count

### LINKED ISSUE #944: Can't remove members when group size < 3
Note: I found this issue while reviewing PR #916 but it also happens on master.

### Problem

Here's what I did to find this bug:
1. Create group;
2. Add 2 more members to it (proposals are now enabled);
3. Remove 1 member (via proposal that was accepted with 1 yes vote);
4. Try to remove 1 more member, so my group can have only 1 member, created a proposal;
5. Impossible to close that proposal because the proposal creator can’t vote and the member being removed can’t vote either (and the group only has 2 members);

### Solution

This is what I expected to happen:

- Remove first member via proposal, because the group has 3 members;
- Remove the second member immediately, without proposal.

Let me know if you agree with my suggestion!

sandrina-p:
Closes #944.

The original issue is about "Can't remove members when group size less than 3". This bug only happens if we try to remove a member in a group of 2 members that already had 3 or more members in the past. We are prompt to create a proposal instead of immediate removal.

Actually, this "proposal issue" happens with any kind of change (e.g. try to change the mincome and it will prompt you to create a proposal instead of being an immediate change.)

The **cause** was quite simple. Since #900, removed members are kept on the list of members but with a status `"removed"`. We were counting all members regardless of their status. From now on, the counting excludes members that are not `"active"`.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -244,7 +244,7 @@ DefineContract({
       return monthstamp => groupIncomeDistribution({ state, getters, monthstamp, adjusted: true })
     },
     groupMembersByUsername (state, getters) {
-      return Object.keys(getters.currentGroupState.profiles || {})
+      return Object.keys(getters.groupProfiles)
     },
     groupMembersCount (state, getters) {
       return getters.groupMembersByUsername.length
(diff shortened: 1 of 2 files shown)


pieer (approved):

taoeffect (approved):
Excellent work @sandrina-p !

Issue #935, opened 2020-05-30
Using modals


sandrina-p:
### Links vs Buttons
When a user clicks a link, and they are moved to another page (or are moved to another place within the same page), and the URL is updated, it's advised to use a proper link `<a>` instead of a `<button>` because it's what makes more sense semantically speaking.

Besides that, when a `<button>` is used instead of a proper link `<a>`, the user cannot do a task in a new tab if they want to (e.g: create a new group, change income details, etc...).

So far, we have been opening modals using a `<button>` and `sbp`. The modal is opened and the URL is also updated manually using `$router`.

```js
button(@click='openModal("GroupCreationModal")') Add a group

// ...

openModal (name) {
  sbp('okTurtles.events/emit', OPEN_MODAL, name)
}
```

A **solution** is to use a proper link instead of a button when opening new modals. And we can use `router-link` without `sbp`.

```pug
router-link.link(:to='{ query: { modal: "GroupCreationModal" }}') Add a group
```

This approach is used by some other big websites: (e.g Twitter when creating a new tweet).

### Props vs Queries
Some modals need specific data to be displayed correctly. Currently, we pass that data through props. 

```js
openModal () {
  sbp('okTurtles.events/emit', OPEN_MODAL, 'RemoveMember', { username: 'rick' })
}
```

And the URL gets updated, but without the data passed.

```
http://localhost:8000/app/dashboard?modal=RemoveMember
```

This approach has a flaw: when we refresh/open this link, the modal does not load because the prop is missing. After all, what's the point of changing the URL if the content (modal) does not get loaded as expected?

There are **two solutions** here:

1. Use queries instead of props.

I checked all modals and there are only 2 modals that need parameters (e.g. remove member and payment detail). That can be passed through a query:

```
http://localhost:8000/app/dashboard?modal=removeMember&username=rick
http://localhost:8000/app/dashboard?modal=paymentDetail&id=12345
```

There are some websites that follow this approach (e.g Twitter and Coinbase)

All routing libraries out there that I know do not allow passing props from route to route. Data can only be passed through queries. I guess it's because any route can be opened standalone without a previous context (e.g. open a link directly, or refresh the page). So, their approach is by using URL queries.

2. Do not update the URL when opening a modal.

This solves all the issues with URLs (and it's okay to use a `<button>` in that situation). However, we lose the possibility of opening in a new tab / sharing / refreshing a page with a modal. (related to bug #472).

[...]

PR #946, opened 2020-06-15
#935 Using modals - queries instead of props

sandrina-p:
Relative to #935

### Usage
The way we open modals is the same. Pass the queries through the 2nd parameter as before.

```js
openModal (name, queries) {
  sbp('okTurtles.events/emit', OPEN_MODAL, 'RemoveMember', {username: 'bob' })
},
```

In the modal, get the query at `created()` method. Make sure to call `SET_MODAL_QUERIES` with the query passed. This will guarantee that the query `username` gets removed when the modal is closed, even when the modal is opened through a direct link or page refresh.

```js
  created () {
    const username = this.$route.query.username

    if (username) {
      this.username = username
      // Do not forget to do this in modals that receive queries
      sbp('okTurtles.events/emit', SET_MODAL_QUERIES, 'RemoveMember', { username })
    } else {
      console.warn('RemoveMember: Missing valid query "username".')
      sbp('okTurtles.events/emit', CLOSE_MODAL)
    }
  },
```

### Examples
Opening modals through queries instead of props allowed us to **fix some current bugs**. Here's a list with different scenarios and the excepted behavior:
-  [Remove member "bob" from the group](http://localhost:8000/app/dashboard?modal=RemoveMember&username=bob): Make sure bob is part of the group, otherwise the modal will not open.
    - Modal is opened with correct data even after refreshing the page.
    - Close modal. `username` query is removed.
- [User Settings - Notifications](http://localhost:8000/app/dashboard?modal=UserSettingsModal&section=notifications): Make sure you are logged in.
    - The settings modal is opened in the correct tab.
    - Close the modal. `section` query is removed.
    - Change to another tab (ex. App Logs) Close modal. `section` query is removed.
- [User Settings - App Logs with error msg](http://localhost:8000/app/dashboard?modal=UserSettingsModal&section=application-logs&errorMsg=Cannot%20set%20property%20%27x%27%20of%20undefined) - Make sure you are logged in.
    - The settings modal is opened in the correct tab and with an error log
    - Close modal. `section` and `errorMsg` queries are removed.
- Payment Detail: Make a payment and open the payment detail.
  - Modal is opened with the correct data, even after refreshing the page.
- [Example DS query](http://localhost:8000/app/design-system?modal=DSModalQuery&name=John&subcontent=SignupModal) - A nested modal (SignUp) opened after a modal with a query "john"
  - Both modals (Hello John and Signup) are opened in the correct order.
  - Closing the Signup modal removes the `subcontent` query but keeps the `name` query.
  - Closing the "Hello John" modal, removes `name` query.


pieer (dismissed):
Great work @sandrina-p 👌

sdegutis (dismissed):
Looks good and works as expected! Great job, thanks.

sandrina-p:
Update with master, @taoeffect  can this be the first one to be merged?

sdegutis (approved):
The latest changes look good to me! 👍

frontend/views/components/modal/Modal.vue
```diff
@@ -22,13 +24,13 @@ export default {
     sbp('okTurtles.events/on', OPEN_MODAL, this.openModal)
     sbp('okTurtles.events/on', CLOSE_MODAL, this.unloadModal)
     sbp('okTurtles.events/on', REPLACE_MODAL, this.replaceModal)
+    sbp('okTurtles.events/on', SET_MODAL_QUERIES, this.setModalQueries)
```
taoeffect, line 27: Any event listener that's enabled must have a corresponding call to disable it (see `beforeDestroy ()`). However, see also comment below as this event may not be needed at all. Ideally we do this without it.

frontend/views/components/modal/Modal.vue
```diff
@@ -124,6 +135,9 @@ export default {
       // At the moment you can only replace a modal if it's the main one by design
       // Use direct children instead of sbp to wait for animation out
       this.$refs['content'].$children[0].close()
+    },
+    setModalQueries (componentName, queries) {
+      this.queries[componentName] = queries
```
taoeffect, line 140: > In the modal, get the query at `created()` method. Make sure to call `SET_MODAL_QUERIES` with the query passed. This will guarantee that the query `username` gets removed when the modal is closed, even when the modal is opened through a direct link or page refresh.

It seems rather cumbersome to have to always remember to do this, and feels to me like a DRY violation, because the same redundant code must be copy/pasted and repeated in every modal. Ideally it should be possible to use the given information to automatically do this.

When a modal is loaded, whether via `sbp('okTurtles.events/emit', OPEN_MODAL, 'RemoveMember', {username: 'bob' })`, or via a link, our modal system should automatically handle setting this variable. Surely there is a way for our modal system to listen to a page load / route change and set this query variable automatically?

Can any one of [these router guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-resolve-guards) be used to achieve this?

- via the `watch` key on `'$route'` above on line 43
- [Per-Route Guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#per-route-guard) `beforeEnter`
- [In-Component Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards): `beforeRouteEnter`, `beforeRouteUpdate`
- [Global Before Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards): `beforeEach`

taoeffect (changes requested):
Nice job @sandrina-p, I like the idea but for some parts of the implementation I have some questions/concerns on, see comments.

taoeffect:
OK, per conversation on Slack I will merge this and open an issue based on the review above for this to be fixed in a followup.

frontend/views/components/modal/Modal.vue
sdegutis: At the very least, each modal will have to record which query params it "owns" and thus deletes when closed, right? I mean, if you have `?foo=1&bar=2` but this modal only needs `foo`, when closed it should have `?bar=2`.

While `["foo"]` as a list of query params to delete could be obtained from the OPEN_MODAL call, when opening it from a URL there seems to be no way to distinguish this information, unless the modal itself were to provide it.

I'm not yet sure how/where to structure this, but these are at least my initial thoughts @taoeffect.

taoeffect: @sdegutis Is it possible to structure the query/URL so that the parameters are associated with each modal in the URL itself?

sdegutis: One of the simplest solutions is to rename each key like `${comp-name}-${prop-name}` e.g.:

* `?modal=UserSettingsModal&UserSettingsModal-action=my-account`
* `?modal=PaymentDetail&PaymentDetail-id=abc123`

It's more verbose, but gives us basic association. This would allow us to figure out more safely which query params to remove when the component closes, assuming we always retain a global namespace of component names. It would also protect against property name collisions, e.g. if the page you're on and the modal it opens both have a `filter` property for some reason.

sdegutis: > > In the modal, get the query at `created()` method. Make sure to call `SET_MODAL_QUERIES` with the query passed. This will guarantee that the query `username` gets removed when the modal is closed, even when the modal is opened through a direct link or page refresh.
> 
> It seems rather cumbersome to have to always remember to do this, and feels to me like a DRY violation, because the same redundant code must be copy/pasted and repeated in every modal. Ideally it should be possible to use the given information to automatically do this.
> 
> When a modal is loaded, whether via `sbp('okTurtles.events/emit', OPEN_MODAL, 'RemoveMember', {username: 'bob' })`, or via a link, our modal system should automatically handle setting this variable. Surely there is a way for our modal system to listen to a page load / route change and set this query variable automatically?
> 
> Can any one of [these router guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-resolve-guards) be used to achieve this?
> 
> * via the `watch` key on `'$route'` above on line 43
> * [Per-Route Guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#per-route-guard) `beforeEnter`
> * [In-Component Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards): `beforeRouteEnter`, `beforeRouteUpdate`
> * [Global Before Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards): `beforeEach`

In order for the modal system to automatically "know" that certain query parameters belonged to the modal that's now exiting, some kind of association between modal component and its query parameters is needed:

1. When the modal component that sets queries mounts, and thus changes the routes, the modal system would need to know that this modal component is the one that set it, so that it can associate these queries with it; and when it unmounts, it needs to know which modal component is unmounting, so that it can remove them.

2. A similar solution is to put static data somewhere on the modal component itself, so that, when the modal component unmounts, it can look up which query fields to remove. This is basically what I had in mind in my first comment on this subthread.

Both of these solutions require that the router hooks have some way of identifying a specific *component* during route change events, which does not seem to be possible based on Vue Route docs. If I'm mistaken, then these two options are back on the table.

In the meantime, either the current solution, or the one I proposed yesterday in my previous comment, are all I can think of that would be practical. I agree with your concerns @taoeffect about this solution, but my other solution (renaming query params) is more invasive and therefore might have other ramifications on the rest of the system which, in my current lack of familiarity with the whole system, I may not yet be able to predict.

sdegutis: One other solution I thought of is to use `beforeRouteLeave` on each component to manually remove its own query params. This solution is less clean, more of a DRY violation, and easier to get wrong. But it could be improved with abstraction.

The first layer of abstraction could be to create a shared function and pass the parameters to that:

```js
beforeRouteLeave() {
  removeQueryParams('section', 'filter') // alternatively, SBP could be used
}
```

This is roughly the same complexity as the current solution, and might actually be cleaner both in interface and implementation.

This could further be abstracted by some kind of "inheritance" in order to make this "automatic":

```js
inheritFrom: 'ModalThing', // pseudo-code
queryParamsToRemove: ['section', 'filter']
```

This is less clean IMO and should be avoided, but putting it out here in case it might give anyone more ideas, or maybe my judgment about it being a poor solution is generally not agreed.

taoeffect: We really need a simple, elegant, automated way of doing this that makes it difficult for developers to "forget" to do it correctly. We have all of the information we need in the URL. Over on Slack you came up with an interesting method that could show promise, by using the hash: `/app/dashboard#UserSettingsModal?section=appearance`

For example, instead of a very ugly and verbose way of associating variables (e.g. `UserSettingsModal-action`), we could try something along these lines:

```
/app/dashboard?modal#Modal1?var1=val1&var2=val2_Modal2?var1=val1_Modal3
```

Or something to that effect. I am honestly not familiar with what is and isn't possible in these URL schemes, and I wonder if with enough creativity/experimentation we could figure out a solution that allows us to specify:

1. The modals and their order
2. The variables for each modal (in a non-ugly way)

And still have everything automatically cleared when the back/forward buttons are used, or when modals are closed.

If a solution is found, it shouldn't be too difficult to go through the entire codebase and update all modal usage to use this new method.

taoeffect: The only issue I see is how do you then handle using `router-link` in a clean way? I'm honestly not sure.

This was the example Sandrina gave:

```pug
router-link.link(:to='{ query: { modal: "GroupCreationModal" }}') Add a group
```

But I'm not sure what to do if you want this sort of push/pop behavior for adding modals with parameters.

sdegutis: Are the only variables we ever want to remove from a modal just "identifying" information?

In the PaymentDetails modal we only have `id`, in ChangeVotingRules we only have `rule`, in RemoveMember we only have `username`, in UserSettingsModal we only have `section`. So it seems to be the case.

If that's the case, we can skip query params altogether for this, and just use a dot-separated hashtag, where removing the hashtag is sufficient to remove the modal's "navigational" information, and the modal itself can parse it from the tag:

* `#PaymentDetails.123` - it can parse `{ id: 123 }`
* `#RemoveMember.fred` - it can parse `{ username: 'fred' }`
* etc

sdegutis: The only problems that still remain here are that:

* The variable of the hashtag can't have a dot in it, or we have to have a foolproof way of escaping it.
* Variables will have to be unnamed and strongly-ordered, but that's probably okay.
* There's still no way to remove actual query params that the modal may use, e.g. if the modal contains a filterable list, it may have added `?filter=foo`, which we can't automatically know to remove.

taoeffect: @sdegutis That looks much nicer. I feel like that is a worthwhile direction to explore. You can play with the syntax, and consider what it might look like if you need to pass in multiple parameters to a modal.

taoeffect: I don't know what you mean by removing (I'm forgetting atm the details of this, just remember what I didn't like 😆).

What about: `#PaymentDetails.id=123.id2=345#RemoveMember.username=fred` or something to that effect?

You'd have to figure out whether multiple `#` are OK after the first `#`, and if not, find a different separator, etc., and same goes for `.` and `=`, different characters can be used to do the separation.

There is still the remaining item of figuring out a push-based syntax for "click this link to add a new modal and a new corresponding set a params to the URL" to replace the previous pug stuff:

```pug
router-link.link(:to='{ query: { modal: "GroupCreationModal" }}') Add a group
```

taoeffect: @sdegutis Oh, and another crazy idea: you go a completely different way too: instead of trying to effectively re-create a URL syntax to encode an object-like structure, you could pass JSON directly in the URL as base64 that's encoded and decoded.

E.g. `/app/dashboard?modal=9823uadfjop8983ru9823rf98hf982h0390gh39r0ghehsdhfsdkjfidjdfghlksdjfghkasdhff982hf22983hfdsjflj` (with or without a hash).

Just another thought that popped into my head, to kinda illustrate how much freedom there is to explore design here.

sdegutis: By removing, I mean removing the modal's info from the URL when the modal closes. This is the reason @sandrina-p added the current solution, so that it can track which params to remove.

The router link will get even easier:

```js
router-link.link(:to='"#RemoveMember." + username')
```

It could use the object-based API instead but that's not much cleaner/shorter, it just replaces "#..." with "{ hash: ... }":

```js
router-link.link(:to='{ hash: "RemoveMember." + username' })
```

I think modals are inherently supposed to be small, short-lived, and single-purpose, so we can probably just not worry about ordering and naming of modal-specific variables, and live with the small but reasonable caveats listed above.

sdegutis: > Oh, and another crazy idea: base64 JSON directly in the URL

Crazy indeed. One of the best benefits of URLs is that they're human readable and thus self-describing. The Java communities tried many experiments in (ab)using URLs like this, and I think there's general consensus that we just shouldn't.

sdegutis: > to kinda illustrate how much freedom there is to explore design here.

Absolutely, thanks for putting it out there. I do think we're on a good path with the hashtag idea, but I've found that brainstorming like this has been very helpful in the past in coming up with out-of-the-box ideas that we can merge with other ideas to have ideas that are half-in and half-out of the box and work wonderfully.

PR #948, opened 2020-06-15
Improve the UI stack of each page in the app. Close #562

### LINKED ISSUE #562: Improve the UI stack of each page in the app
### Problem

From Scott Hurff's [How to fix a bad user interface](http://scotthurff.com/posts/why-your-user-interface-is-awkward-youre-ignoring-the-ui-stack):

> Every screen you interact with in a digital product has multiple personalities.
> Five, to be exact.
> Depending on the context, these personalities are revealed to your customer. In designer-speak, we call these states. And you should consider these states for every screen you make.
> That's because following the rules of the UI Stack and the five states helps you create a cohesive interface that’s forgiving, helpful, and human.

At the moment, pages in the Group Income app **do not have** all the states described in the UI stack:

ui-stack

### Solution

- Define how each of these states should look and feel in our app - create the templates.
- Take those templates and adapt it to each page, in case it's needed;

_Notes: Related to #496_

pieer:


sandrina-p:
Hey @pieer could you add some instructions to the PR, please?
For example, how can I simulate the loading state?

The way I see, when I visit the dashboard, it still fetches all the other pages (they are in the network tab).

And I have another question: How do we run the project in production mode?

mmbotelho (dismissed):
This is looking great @pieer, good job! My comments are not blocking issues, and for that reason, I'm approving the PR. 

The only change I would suggest is that the gradient is a bit harsh - can we make it more subtle? Maybe making the lighter color in the gradient a bit darker would help. 

But anyway, approved! Great work 🎉

pieer:
> For example, how can I simulate the loading state?

At the moment I don't know how to simulate loading on the local

> The way I see, when I visit the dashboard, it still fetches all the other pages (they are in the network tab).

This is correct, the webpack config is pre fetching them at the moment

> And I have another question: How do we run the project in production mode?

No idea

mmbotelho (dismissed):
Perfect @pieer, thank you!

mmbotelho:
@pieer I'm having some trouble viewing this pull request, I always get this:

Captura de ecrã 2020-08-12, às 12 46 19

mmbotelho:
Fixed, it was my (very noob) mistake! 😅

sdegutis:
You can simulate this locally by changing `frontend/controller/router.js` line 20:

```diff
-  const AsyncHandler = () => ({ component, loading, error })
+  const AsyncHandler = () => ({ component: new Promise(resolve => setTimeout(() => resolve(component), 1000)), loading, error })
```

sdegutis (dismissed):
Looks good to me!

frontend/utils/lazyLoadedView.js
```diff
@@ -1,16 +1,35 @@
 import Vue from 'vue'
-// NOTE: any modals opened with `OPEN_MODAL` go here
```
sdegutis, line 2: Just a small thought: Should this comment be left in? It still seems relevant, unless it's well enough implied by this point.

frontend/utils/lazyLoadedView.js
```diff
@@ -1,16 +1,35 @@
...
+  const AsyncHandler = () => ({ component, loading, error })
+
+  return () =>
+    Promise.resolve({
+      functional: true,
+      render (h, { data, children }) {
+        return h(AsyncHandler, data, children)
+      }
+    })
+}
```
taoeffect: DRY violation: the only difference between this and the other version seems to be `error = LoadingPage`. Please DRY this so that only a single definition of `lazyLoadView` is defined, exported, and imported.

taoeffect: In my local experiments, and from what I read of the Vue documentation, it is not necessary to have this function wrap the component inside of a functional component. So returning an immediately resolved promise isn't necessary, and the code below works just as well:

```js
const lazyLoadView = ({ component, loading = LoadingPage, error = LoadingPage }) => {
  return () => ({ component, loading, error })
}
```

However, additionally, neither the original version, nor the version above, showed me any loading screens/pages when I used Firefox's GPRS throttling mode in the Network tab:

Screen Shot 2020-09-11 at 6 33 19 PM

Instead what it did was it simply waited to load a component (without showing a loading screen) and then displayed the final component.

taoeffect (changes requested):
Great job @pieer! In my tests I wasn't able to reproduce the loading screen locally, see comments!

frontend/utils/lazyLoadedView.js
taoeffect: You could probably save a few characters and make the lines shorter by passing in the options as an options object, e.g.:

```js
const lazyLoadView = (component, { loading = LoadingPage, error = ErrorPage }) => {
  return () => ({ component, loading, error })
}

// ... and then ...

const GroupDashboard = lazyLoadView(import('@pages/GroupDashboard.vue'))

// or with options
const GroupDashboard = lazyLoadView(import('@pages/GroupDashboard.vue'), { loading: LoadingModal })
```

pieer: ... @taoeffect like we mention before, I don't think you can emulate network throttling when the file are already on your computer.
Concerning you refactoring, I have no idea what you're trying to do.
Maybe this vue-router page doc could help you to understand the bit code written here: https://github.com/vuejs/vue-router/pull/2140/files#diff-7d999265ce5b22152fdffee108ca6385

frontend/utils/lazyLoadedView.js
```diff
@@ -1,16 +1,32 @@
...
-
-Vue.component('GroupCreationModal', () => import('../views/containers/group-settings/GroupCreationModal.vue'))
-Vue.component('GroupJoinModal', () => import('../views/containers/group-settings/GroupJoinModal.vue'))
+import LoadingPage from '@views/containers/loading-error/LoadingPage.vue'
+import ErrorPage from '@views/containers/loading-error/ErrorPage.vue'
+import LoadingModal from '@views/containers/loading-error/LoadingModal.vue'
+import ErrorModal from '@views/containers/loading-error/ErrorModal.vue'
+import LoadingModalFullScreen from '@views/containers/loading-error/LoadingBaseModal.vue'
+
+const lazyLoadView = ({ component, loading = LoadingPage, error = ErrorPage }) => {
```
taoeffect, line 19: To help us remember where this code is from and why we're doing it this way, above line 8, can you please include a comment that says:

```js
/*
This method of loading components is documented here and is used to ensure compatibility
with lazy-loaded routes:
https://github.com/vuejs/vue-router/pull/2140/files#diff-7d999265ce5b22152fdffee108ca6385

WARNING: Components loaded with this strategy will **not** have access to in-component guards,
such as `beforeRouteEnter`, `beforeRouteUpdate`, and `beforeRouteLeave`. If you need to use 
these, you must either use route-level guards instead or lazy-load the component directly, without 
handling loading state.
*/
```

taoeffect (changes requested):
I successfully managed to get the loading component to display using the strategy @sdegutis mentioned:

```diff
-  const AsyncHandler = () => ({ component, loading, error })
+  const AsyncHandler = () => ({ component: new Promise(resolve => setTimeout(() => resolve(component), 1000)), loading, error })
```

However I noticed that for some reason on some pages the loading component is displayed multiple times:

multiple-loading-components

I also am very troubled by the fact that I truly test this PR without modifying the code. @pieer & @sdegutis can you guys try to figure out how to get it to work with some web browser's tools? E.g. I tried to use Firefox Developer Edition's network throttling tool and that simply didn't work. Is this a bug in Vue.js? What is the reason it's not working?

How can we know that this code works if in order to get it to display the loading stuff you have to modify the code?

sdegutis:
> I successfully managed to get the loading component to display using the strategy @sdegutis mentioned: [...] I also am very troubled by the fact that I [cannot] truly test this PR without modifying the code. [...] How can we know that this code works if in order to get it to display the loading stuff you have to modify the code?

All the code does is delay loading the component for a given duration. It can thus be proven logically that the code works without the code. I think it's practically equivalent to using devtools to delay loading a file, except that we have more control over this way and are less dependent on third party tools (that we can't currently get working anyway). This also allows testing it on more browsers than just Firefox, e.g. if Safari and Chrome don't (and they probably don't) have such devtool features.

taoeffect:
> All the code does is delay loading the component for a given duration. It can thus be proven logically that the code works without the code. I think it's practically equivalent to using devtools to delay loading a file, except that

Except it doesn't work the same way, so it's clearly not equivalent. Since it's not equivalent, nothing is proven. We therefore have a mystery on our hands, and reason to believe this code might not work in practice. In order to be certain that it actually works, we need to find out what's going wrong. At the very least, we should be able to demonstrate that it works over a slow connection (artificially shown or otherwise), without modifying the code.

sdegutis:
There's no mystery. If you understand that this line of code is just delaying the resolution of a promise after a `setTimeout`, and that the promise returns a component, then it's completely understandable, and it perfectly imitates a slow connection. I've used this technique in the past to test loading components the same way. It also is the only way that works for all browsers including mobile Safari.

taoeffect:
> There's no mystery.

Then explain why it doesn't work when Firefox delays loading the component?

sdegutis:
I just tested it in Firefox and Chrome with throttling and you're right, it does behave differently than my code, my mistake. I will look further into this.

sdegutis:
It seems to be a problem with how the lazy component is defined.

According to [these Vue docs](https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State), the `lazyLoadView` returns the right structure, eventually.

In fact, it can even be defined as:

```js
const lazyLoadView = ({ component, loading = LoadingPage, error = ErrorPage }) => {
  return () => ({ component, loading, error })
}
```

But testing this doesn't work.

However, I realized that this:

```js
Vue.component('LoginModal', lazyLoadView({ component: import('../views/containers/access/LoginModal.vue'), loading: LoadingModal, error: ErrorModal }))
```

Is effectively the same as this:

```js
Vue.component('LoginModal', () => ({ component: import('../views/containers/access/LoginModal.vue'), loading: LoadingModal, error: ErrorModal }))
```

And now, when I try this with throttling on in Chrome and Edge, I definitely see the loading component.

Which means we can just get rid of the `lazyLoadView` function altogether, especially since it wasn't working.

The one benefit of `lazyLoadView` left was that it had default values for the loading/error components. (And perhaps this is exactly why it was failing to work.) But personally I think it's better to be explicit about which components to use for these scenarios when defining a new component/route.

taoeffect:
Awesome @sdegutis, thanks for investigating! Did you check to see if that worked with the vue-router routes too? Because according [to the link](https://github.com/vuejs/vue-router/pull/2140/files#diff-7d999265ce5b22152fdffee108ca6385) Pierre shared above, it seems to be a different situation there.

frontend/utils/lazyLoadedView.js
pieer: @taoeffect ok, I will add the comment.
The multiple box is not a bug but the intentional design to express the different elements that are loading.
It's like that on pretty every loading element
image
image

pieer:
I'm not sure to understand all your supposition @sdegutis .
How this:
`  
const lazyLoadView = ({ component, loading = LoadingPage, error = ErrorPage }) => {
const AsyncHandler = () => ({ component, loading, error })
return () =>
Vue.component('GroupCreationModal', () => import('../views/containers/group-settings/GroupCreationModal.vue'))	    Promise.resolve({
Vue.component('GroupJoinModal', () => import('../views/containers/group-settings/GroupJoinModal.vue'))	      functional: true,
      render (h, { data, children }) {
        return h(AsyncHandler, data, children)
      }
    })`

is the same as this:
`
const lazyLoadView = ({ component, loading = LoadingPage, error = ErrorPage }) => {
  return () => ({ component, loading, error })
}
`

By the way this code is coming from posva (that I know and has been in charge of the vue router for few years). You can see the source here: https://github.com/vuejs/vue-router/pull/2140/files#

sdegutis:
Thanks @pieer and @taoeffect for clarifying the purpose of `lazyLoadView`. I wasn't aware that vue-router wouldn't work without it.

When using `lazyLoadView`, the component loads immediately, not lazily. This explains why we're not seeing the correct behavior with network throttling enabled.

After reading the docs you both linked to a little more carefully, I found the problem: The `lazyLoadView` function has to return the Promise, not a function that returns the Promise. Where the call to `lazyLoadView` is passed to `Vue.component`, we need to instead pass a function that returns the call to `lazyLoadView`.

Specifically:

```js
const lazyLoadView = ({ component, loading = LoadingPage, error = ErrorPage }) => {
  const AsyncHandler = () => ({ component, loading, error })

  return Promise.resolve({
    functional: true,
    render (h, { data, children }) {
      return h(AsyncHandler, data, children)
    }
  })
}

Vue.component('LoginModal', () => lazyLoadView({ component: import('../views/containers/access/LoginModal.vue'), loading: LoadingModal, error: ErrorModal }))
```

The reason it would always load immediately before this change is *probably* because the promise would start to load eagerly when invoked during the invocation of `lazyLoadView`, so that it was ready and resolved immediately by the time you clicked through to the given component.

Either way, this fixes it on my end reliably.

pieer:
Good catch @sdegutis ! Feel free to leave a comment on the vue router doc, it hasn't been merged yet

pieer:
It actually worked, but I changed it anyway for consistency

frontend/controller/router.js
```diff
@@ -87,8 +87,8 @@ const router = new Router({
     },
     {
       path: '/design-system',
-      component: DesignSystem,
-      name: DesignSystem.name,
+      component: () => DesignSystem,
+      name: 'DesignSystem',
```
sdegutis, line 91: Interesting idea. I'm on the fence personally in terms of whether to use `DesignSystem.name` or `'DesignSystem'`. The first one is safer against renames, and facilitates project-wide renames in the IDE. But the second one is less "magical", and allows you to move the arrow function back up into the `const`s above, which I think is technically more "correct". Just some thoughts, not an official request for changes. I think a case could be made for both sides on this.

sdegutis:
> It actually worked, but I changed it anyway for consistency

Oops, my mistake. Assigning a function like that `const foo = () => ...` will give it `.name === 'foo'` correctly. Dunno what I was thinking yesterday. Anyway good call changing it for consistency. 👍

taoeffect (changes requested):
Woohoo! Almost!

It looks like this version now works with the Firefox Dev Tools throttling! Great job @pieer and @sdegutis for figuring that out!

I did however run into a problem while testing. Here are the steps to reproduce:

1. Set the throttling to "Regular 2G" or slower
2. Sign up
3. Create a group

You'll see a weird behavior where the modal appears and then immediately disappears. You might have to try it a few times in a new container tab if it doesn't happen the first time. But then you should see something like this:

bug mov

taoeffect (commented):
Tested it locally and it seems to work with the network tool in FF! Great job @pieer!

I noticed a few things that I'm wondering if you could fix:

1. The IncomeDetails modal isn't set up to do the loading thing, is that intentional?
2. Can you undo the changes to the `package-lock.json` file?

taoeffect (approved):
Great work @pieer! 😄 💯

PR #950, opened 2020-06-29
#865 Improve mincome inputs ux

### LINKED ISSUE #865: Bug: Improve UX of mincome inputs
### Problem

Currently, the inputs to insert the mincome don't work correctly for everyone:

- Some inputs only accept numbers up to 2 decimals. But if the currency is BTC, 2 decimals aren't enough.
- In some countries (ex: Portugal) we use a comma instead of dot to write numbers. Currently, we don't support comma (it shows an inline error). On a PT mobile keyboard, it only shows the comma option, so it's impossible to write decimals there.

### Solution

Implement correctly the inputs to support all scenarios regardless of the currency or browser.
- It should limit the decimals based on currency (limit TBD, cc @taoeffect )
- It should accept both comma or dots. `10.5` and `10,5` are both valid mincomes.
- Bonus: Improve input usability with `inputmode='decimal'`. [See reference](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/)

sandrina-p:
Closes #865

image

- Support mincome with decimals based on group currency, using comma or dots

#### Places updated:
- Group Creations (mincome step) 
- Proposal to change mincome
- Update Income Details
- _somewhere missing? let me know!_

#### Scenarios covered:
- Decimals using comma or dots. E.g. `1,55` and `1.55` are both valid.
- Add decimals limit
  - EUR/USD: 2 (e.g. `100.50`)
  - BTC: 8 (e.g. `0.12345678`)
- 0 (zero) or negative mincome not allowed.

For future development reference:

- Use `saferFloat` when saving an amount based on an input value.
- Use `currencies[currency].validate()` to validate a currency amount value

```js
import currencies, { mincomePositive, saferFloat } from '@view-utils/currencies.js'

// ex: change the mincome...
const updatedSettings = await sbp('gi.contracts/group/updateSettings/create', {
  mincomeAmount: saferFloat(this.form.mincomeAmount) // e.g. "1,5" is converted to 1.5 
}, this.currentGroupId)

// ex: validating the mincome
validations: {
     mincomeAmount: {
        positive: mincomePositive,
        decimals: function (amount) {
          // e.g. USD - "1,55" is valid but "1.789" is not valid.
          return currencies["USD"].validate(amount)
        }
      },
    }
}
```

### DIFF
--- a/frontend/views/utils/currencies.js
+++ b/frontend/views/utils/currencies.js
@@ -7,8 +7,38 @@
 // this value, switch to a different currency base, e.g. from BTC to mBTC.
 export const DECIMALS_MAX = 8
 
-export function saferFloat (value: number): number {
-  return parseFloat(value.toFixed(DECIMALS_MAX))
+function commaToDots (value) {
+  // ex: "1,55" -> "1.55"
+  return typeof value === 'string' ? value.replace(/,/, '.') : value
+}
+
+function isNumeric (nr) {
+  return !isNaN(nr - parseFloat(nr))
+}
+
+function isInDecimalsLimit (nr, currency) {
+  const decimals = nr.toString().split('.')[1]
+  return !decimals || decimals.length <= currencies[currency].decimalsMax
+}
+
+function validateMincome (value, currency) {
+  const nr = commaToDots(value)
+  return isNumeric(nr) && isInDecimalsLimit(nr, currency)
+}
+
+function decimalsOrInt (num: number, currency: string): string {
+  const fixed = parseFloat(num.toFixed(currencies[currency].decimalsMax))
+  const integerPart = /^(.+)\.0+$/.exec(fixed)
+  return integerPart ? integerPart[1] : fixed
+}
+
+export function saferFloat (value): number {
+  // ex: "1,333333333333333333" -> 1.33333333
+  return parseFloat(parseFloat(commaToDots(value)).toFixed(DECIMALS_MAX))
+}
+
+export function mincomePositive (value): boolean {
+  return parseFloat(commaToDots(value)) > 0
 }
 
 // NOTE: if we needed for some reason, this could also be defined in
@@ -21,29 +51,25 @@ const currencies = {
     symbolWithCode: '$ USD',
     decimalsMax: 2,
     displayWithCurrency: n => '$' + decimalsOrInt(n, 'USD'),
-    displayWithoutCurrency: n => decimalsOrInt(n, 'USD')
+    displayWithoutCurrency: n => decimalsOrInt(n, 'USD'),
+    validate: n => validateMincome(n, 'USD')
   },
   EUR: {
     symbol: '€',
     symbolWithCode: '€ EUR',
     decimalsMax: 2,
     displayWithCurrency: n => '€' + decimalsOrInt(n, 'EUR'),
-    displayWithoutCurrency: n => decimalsOrInt(n, 'EUR')
-
+    displayWithoutCurrency: n => decimalsOrInt(n, 'EUR'),
+    validate: n => validateMincome(n, 'EUR')
   },
   BTC: {
     symbol: 'Ƀ',
     symbolWithCode: 'Ƀ BTC',
     decimalsMax: DECIMALS_MAX,
     displayWithCurrency: n => decimalsOrInt(n, 'BTC') + 'Ƀ',
-    displayWithoutCurrency: n => decimalsOrInt(n, 'BTC')
+    displayWithoutCurrency: n => decimalsOrInt(n, 'BTC'),
+    validate: n => validateMincome(n, 'BTC')
   }
-}
-
-function decimalsOrInt (num: number, currency: string): string {
-  const fixed = num.toFixed(currencies[currency].decimalsMax)
-  const integerPart = /^(.+)\.0+$/.exec(fixed)
-  return integerPart ? integerPart[1] : fixed
 }
 
 export default currencies
--- a/frontend/views/components/group-creation-steps/GroupMincome.vue
+++ b/frontend/views/components/group-creation-steps/GroupMincome.vue
@@ -3,9 +3,12 @@
   i18n.is-title-4.steps-title(tag='h4') 3. Minimum Income
 
   .card
-    fieldset
+    fieldset.field
       i18n.label What is the minimum each member should receive monthly?
-      .selectgroup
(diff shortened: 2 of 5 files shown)


dotmacro:
Issue:
The mincome step of group creation will not allow the user to proceed if the mincome includes a thousands separator and a decimal mark, but does not give an error or any info indicating why the user can't continue.

To duplicate, enter "1,234.56" or "1.234,56" in the mincome input field and see the "next ->" button is disabled.

Screen Shot 2020-06-29 at 9 25 16 PM
Screen Shot 2020-06-29 at 9 28 56 PM

Works fine with one separator 👍
Screen Shot 2020-06-29 at 9 25 34 PM

sandrina-p:
@dotmacro good catch. I only did the validation for decimal. Doing the validation for the "thousands separator" is trickier. What error copy should appear if this happens?

dotmacro:
I'm having difficulty thinking of error copy that clearly communicates that "the decimal mark is allowed but the thousands separator is not" for two main reasons. First, those terms aren't commonly used (though we could localize "decimal point" and "comma"). Second, I've never seen that behavior before because all the places I can recall (tax office, bank, PayPal, etc) either accept both or neither.

I think allowing a thousands separator is a much better solution, if possible, because errors telling me to write normal numbers in non-standard ways are especially frustrating to me as a user. 
 
Something like https://regex101.com/r/9iYQIp/1, without requiring the thousands separator, if it works with currencies.js.

If that doesn't fit in the scope of this PR, I can open a separate issue.

sandrina-p:
The trick here would be to combine this with the other errors because the error is shown on keyup instead of input blur... And showing an error before the user finishes typing is also annoying. For example, the number "1,888.50"

-`1` - ok
-`1,` - ok
-`1,8` - ok
-`1,888` - "max decimals are 2"
-`1,888.` - "max decimals are 2"
-`1,888.5` - "ok"
-`1,888.50` - "ok"

We could only show the error when the input is blurred.

sandrina-p:
@dotmacro I looked for other websites and couldn't find one that had the validations for both decimals and thousands separator. For example, Coinbase, only supports decimals: 
image

taoeffect:
Yes, do not use or add support for "1,000.50". The only time commas may be used is if they are required for representing the decimal. E.g. I think some countries use `,` instead of `.` for the decimal. Otherwise do not add support for the comma.

sandrina-p:
Yes, that was one of the main purposes for this PR as well, adding comma support only for decimals. Here in Portugal, it's common to use a comma instead of dots.

dotmacro:
> "max decimals are 2"  

If we're going [up to 8 decimal places](https://github.com/okTurtles/group-income-simple/issues/865#issuecomment-612086843), I don't understand this error.  

> We could only show the error when the input is blurred.  

This makes more sense, so users can enter the number either before or after switching the currency. 

> do not add support for the [thousands separator]  

Agreed that if no one else supports thousands separator in your experiences, no need to add support.  👍

> What error copy should appear

"Amount must be a number" works, but isn't the clearest error message when the entry is something the user would consider a number. An alternative is "Please only use a decimal mark", which can be localized to "...decimal point" and "...decimal comma", or even "Please enter a whole number".

sandrina-p:
> If we're going up to 8 decimal places, I don't understand this error.

We are only at BTC. The example I gave was USD/EUR, I forgot to clarify that. :)

Regarding the error, we can give an example in the error. "_The amount must be a number. (E.g. 1500.50)_". An example is worth a thousand words :p

dotmacro:
> > If we're going up to 8 decimal places, I don't understand this error.  

> We are only at BTC. The example I gave was USD/EUR, I forgot to clarify that. :)

Ok, then users will just need to change the currency before they can enter the number. That's a bit of friction for LTR, but not a big deal by itself.

> Regarding the error, we can give an example in the error. "_The amount must be a number. (E.g. 1500.50)_". An example is worth a thousand words :p

If we show this same error message for all currencies, it can lead users to believe we only support 2 decimal places when that's not always the case. Doubly so if we're validating the number input before users are ready to submit the form (e.g. they enter the number before switching EUR->BTC).

Note: Not showing _any_ relevant error message is an issue, but PR doesn't need to be blocked by perfecting the wording. We can figure that out elsewhere (e.g. new issue, Slack) / later if needed.

sandrina-p:
@dotmacro, I think this is fine as it is for the prototype. I only created the issue to support decimals because, without it, it would be impossible to add decimals on iOS mobile, since the keyboard adds a comma instead of a dot. [see screenshot](https://user-images.githubusercontent.com/14869087/86511565-61470100-bdf2-11ea-81de-7fe7009c6046.png)

dotmacro:
> I think this is fine as it is for the prototype  

I'm confused. Does this mean no error message explaining the error state is needed?

sandrina-p:
> I'm confused. Does this mean no error message explaining the error state is needed?

@dotmacro, no. I was referring to the thousand separator validation and the inline validation instead of blur. Regarding the error message, I've added where it was missing. Now it shows an explicit error message in group creation, proposal, and income details.

dotmacro:
Ah, it looks like those changes just hadn't been pushed yet when I checked. Sounds good!

pieer (approved):

sdegutis (approved):
Clean code, and works as described, great job. Looks good to me.

sandrina-p:
@taoeffect this has 2 approvals and all is green. Can it be merged right away @taoeffect?

frontend/views/utils/currencies.js
```diff
@@ -7,8 +7,38 @@
 // this value, switch to a different currency base, e.g. from BTC to mBTC.
 export const DECIMALS_MAX = 8
 
-export function saferFloat (value: number): number {
-  return parseFloat(value.toFixed(DECIMALS_MAX))
+function commaToDots (value) {
+  // ex: "1,55" -> "1.55"
+  return typeof value === 'string' ? value.replace(/,/, '.') : value
```
sdegutis, line 12: Just a quick question about this: is there ever a time this wouldn't be a string or a number? It's hard to trace all the code that calls it, but it always seems to be either a string or a number. Is it expected to ever be undefined or null?

sandrina-p: I want to believe it's always a string or a number :D. I don't remember being otherwise.

PR #951, opened 2020-07-09
#763 Payments - improve search, tabs and tests

sandrina-p:
Continue #763

- Making the ESC key clear the search field
- Fixing a bug where if we switched from sending payments to receiving payments, we suddenly lose the "Sent" tab
- Continue tests
- @mmbotelho can you review if the UI and data are correct when the user changes between receiving and giving mincome?

### Update 10 July

- Make payments pagination functional.

**Tip:** You can create N payments manually and see the pagination work... or you can create *dummy sent payments* by following these steps:
1. Create a group with members and **send at least one payment**.
2. Go to `views/pages/Payments.vue`. Inside the function `paymentsSent ()` (~line 260-275) there's a block of code commented. Uncomment it from `/*` to `*/`. You can change the variable `copies` to customize the number of dummy payments.
3. Save it and the page should refresh within a few seconds. _Voilà!_

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -347,6 +347,7 @@ const getters = {
     return monthstamp => groupIncomeDistribution({ state, getters, monthstamp, adjusted: true })
   },
   ourPayments (state, getters) {
+    // Payments relative to the current month only
     const currency = currencies[getters.groupSettings.mincomeCurrency]
     const ourUsername = getters.ourUsername
     const cMonthstamp = currentMonthstamp()
@@ -354,26 +355,23 @@ const getters = {
     const dueIn = lastDayOfMonth(pDate)
     const monthlyPayments = getters.groupMonthlyPayments
     const allPayments = getters.currentGroupState.payments
+    const thisMonthPayments = monthlyPayments[cMonthstamp]
+    const paymentsFrom = thisMonthPayments && thisMonthPayments.paymentsFrom
 
     const sent = (() => {
       const payments = []
-      for (const monthstamp of Object.keys(monthlyPayments).sort()) {
-        const { paymentsFrom } = monthlyPayments[monthstamp]
 
-        if (paymentsFrom) {
-          for (const toUser in paymentsFrom[getters.ourUsername]) {
-            for (const paymentHash of paymentsFrom[getters.ourUsername][toUser]) {
-              const { data, meta } = allPayments[paymentHash]
-              payments.push({ hash: paymentHash, data, meta })
-            }
+      if (paymentsFrom) {
+        for (const toUser in paymentsFrom[getters.ourUsername]) {
+          for (const paymentHash of paymentsFrom[getters.ourUsername][toUser]) {
+            const { data, meta } = allPayments[paymentHash]
+            payments.push({ hash: paymentHash, data, meta })
           }
         }
       }
       return payments
     })()
     const received = (() => {
-      const thisMonthPayments = monthlyPayments[cMonthstamp]
-      const paymentsFrom = thisMonthPayments && thisMonthPayments.paymentsFrom
       const payments = []
 
       if (paymentsFrom) {
--- a/frontend/views/containers/payments/PaymentsPagination.vue
+++ b/frontend/views/containers/payments/PaymentsPagination.vue
@@ -7,14 +7,14 @@
     label.selectsolo.c-select
       i18n.sr-only Show per page
       select.select(
-        ref='select'
+        :value='rowsPerPage'
         @change='updatePagination'
       )
         option(
-          v-for='numberPerPage in [10, 20, 30]'
-          :index='numberPerPage'
-          :value='numberPerPage'
-        ) {{ `${numberPerPage} ${L('results')}` }}
+          v-for='count in config.options'
+          :index='count'
+          :value='count'
+        ) {{ `${count} ${L('results')}` }}
     i18n.has-text-1(
       tag='span' aria-hidden='true'
     ) per page
@@ -23,17 +23,19 @@
     i18n.has-text-1(
       tag='p'
       data-test='paginationInfo'
-      :args='{ currentPage: `<span class="has-text-0">1 — 5</span>`, maxPages: `<span class="has-text-0">5</span>`}'
-    ) {currentPage} out of {maxPages}
+      :args='paginationInfo'
+    ) {currentPage} out of {count}
 
     .c-previous-next
-      button.c-control.is-unstyled(
+      button.is-icon-small.c-btn(
+        :disabled='page === 0'
         @click='previousPage'
         :aria-label='L("Previous page")'
       )
         i.icon-chevron-left
 
-      button.c-control.is-unstyled(
+      button.is-icon-small.c-btn(
+        :disabled='page + 1 >= maxPages'
         @click='nextPage'
         :aria-label='L("Next page")'
       )
@@ -43,15 +45,41 @@
 <script>
 export default {
   name: 'PaymentsPagination',
+  props: {
+    count: Number,
+    page: Number,
+    rowsPerPage: Number
+    // @changePage('next'|'prev')
+    // @changeRowsPerPage(String) // Page Number
+  },
+  data: () => ({
+    config: {
+      options: [10, 20, 30]
+    }
+  }),
+  computed: {
+    maxPages () {
+      return Math.ceil(this.count / this.rowsPerPage)
+    },
+    paginationInfo () {
+      const style = text => `<span class="has-text-0">${text}</span>`
+      const start = this.rowsPerPage * this.page
+      const currentRows = `${start + 1} - ${Math.min(start + this.rowsPerPage, this.count)}`
+      return {
+        currentPage: style(currentRows),
+        count: style(this.count)
+      }
+    }
+  },
   methods: {
-    updatePagination () {
-      console.log('Todo: implement update pagination settings')
+    updatePagination (e) {
+      this.$emit('changeRowsPerPage', +e.target.value)
     },
     previousPage () {
-      console.log('Todo: implement previous page functionality')
+      this.$emit('changePage', 'prev')
     },
     nextPage () {
-      console.log('Todo: implement next page functionality')
+      this.$emit('changePage', 'next')
     }
   }
 }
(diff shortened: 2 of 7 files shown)


mmbotelho (changes requested):
Here's what I found @sandrina-p !

Captura de ecrã 2020-07-09, às 21 04 36
Captura de ecrã 2020-07-09, às 21 05 13
Captura de ecrã 2020-07-10, às 12 39 10
Captura de ecrã 2020-07-10, às 12 40 13

sandrina-p:
@mmbotelho all fixed. In addition, I also made pagination work! (Read PR description for more details)

pieer (approved):

mmbotelho:
I'm getting this error when trying to record payments. Clicking the button did absolutely nothing visible, but when I closed the modal, the payments were recorded with wrong amounts:

Record payments modal:
Captura de ecrã 2020-08-12, às 15 07 55

"Sent" tab, after closing the modal:
Captura de ecrã 2020-08-12, às 15 10 33

sdegutis:
Hmm, I'm not able to reproduce that error.

frontend/views/pages/Payments.vue
```diff
@@ -141,22 +157,38 @@ export default {
...
         items.push({
           title: L('Received'),
-          url: 'PaymentRowReceived',
-          notification: this.paymentsReceived.length
+          url: 'PaymentRowReceived'
+        })
+      }
+
+      if (doesNeedIncomeAndDidSentBefore) {
+        items.push({
```
sdegutis, line 186: This is created twice, here and on line 169, which are mutually exclusive but are in the same scope. This should probably be cleaned up if possible. I think removing the two variables on lines 175 and 176 would help, by making it possible to combine the conditions above and below it, which would also make semantics more clear.

frontend/views/pages/Payments.vue
```diff
@@ -224,12 +256,22 @@ export default {
           date: meta.createdDate,
           monthstamp: ISOStringToMonthstamp(meta.createdDate),
           amount: data.amount // TODO: properly display and convert in the correct currency using data.currencyFromTo and data.exchangeRate,
-        }
+        })
+      }
 
-        if (this.filterPayment(p)) {
-          payments.push(p)
+      // DELETE THIS BEFORE MERGE!!!
```
sdegutis, line 262: It might be good to instead create a script that can be run locally to generate fake payments in the actual system. Then these lines could safely be deleted permanently, and there would be more flexibility in testing out this feature.

sdegutis (changes requested):
(Partial review.)

frontend/model/state.js
```diff
@@ -347,33 +347,31 @@ const getters = {
...
+    // Payments relative to the current month only
     const currency = currencies[getters.groupSettings.mincomeCurrency]
     const ourUsername = getters.ourUsername
     const cMonthstamp = currentMonthstamp()
     const pDate = dateFromMonthstamp(cMonthstamp)
     const dueIn = lastDayOfMonth(pDate)
     const monthlyPayments = getters.groupMonthlyPayments
     const allPayments = getters.currentGroupState.payments
+    const thisMonthPayments = monthlyPayments[cMonthstamp]
+    const paymentsFrom = thisMonthPayments && thisMonthPayments.paymentsFrom
```
sdegutis, line 359: This would be the perfect spot to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to get rid of the `thisMonthPayments` variable, except that it requires [a slightly newer Rollup version](https://github.com/rollup/rollup/issues/3469).

frontend/views/containers/payments/PaymentsPagination.vue
```diff
@@ -43,15 +45,41 @@
...
+    config: {
+      options: [10, 20, 30]
+    }
+  }),
+  computed: {
+    maxPages () {
+      return Math.ceil(this.count / this.rowsPerPage)
+    },
+    paginationInfo () {
+      const style = text => `<span class="has-text-0">${text}</span>`
```
sdegutis, line 65: I think it makes more sense to put this span into the actual view on line 27. Even though it will be repeated twice over there, I think the increased consistency makes up for that, since it will live with the rest of the view, and use the same pug syntax as the rest of the view. Plus, the names of these view variables will match their contents more accurately, since they'll both be numbers, not HTML strings, which could surprise us later.

frontend/views/containers/payments/PaymentsPagination.vue
```diff
@@ -43,15 +45,41 @@
...
+      const start = this.rowsPerPage * this.page
+      const currentRows = `${start + 1} - ${Math.min(start + this.rowsPerPage, this.count)}`
```
sdegutis, line 67: Similarly to the previous comment, this might be good to split up into two numbers that could be interpolated into the view independently. That way, the " - " in the view is put with the rest of the view, so the whole of this view can be read all at once in the same spot.

test/cypress/integration/group-paying.spec.js
```diff
@@ -42,47 +52,35 @@ describe('Group Payments', () => {
...
+  it('user1 sends $71.43 to user2 (total)', () => {
     cy.giSwitchUser(`user1-${userId}`, { bypassUI: true })
 
     cy.getByDT('paymentsLink').click()
     cy.get('[data-test-date]').should('have.attr', 'data-test-date', humanDateToday)
 
-    assertPaymentsTabs(['Todo2', 'Sent'])
+    assertNavTabs(['Todo2', 'Sent'])
+    assertMonthOverview([
+      'Payments sent0 out of 2',
```
sdegutis, line 81: Even though the HTML gets squished down into this text string, is it possible to separate these into arrays of strings to look for within the `li` in the assertion on line 38? Then it might be easier to read and understand as we add, update, satisfy, and fix tests.

frontend/views/pages/Payments.vue
```diff
@@ -279,18 +318,37 @@ export default {
...
       const { amount, username, displayName } = payment
       return query === '' || `${amount}${username.toUpperCase()}${displayName.toUpperCase()}`.indexOf(query.toUpperCase()) !== -1
     },
+    paginateList (list) {
+      const start = this.ephemeral.rowsPerPage * this.ephemeral.currentPage
+      return list.slice(start, start + this.ephemeral.rowsPerPage)
+    },
+    handleTabClick (url) {
+      this.ephemeral.activeTab = url
+      this.form.search = ''
```
sdegutis, line 338: Should changing tabs remove the search string? The user may want to keep their filter active while temporarily checking something out, intending to come back and see it as-is with the search string filtered.

frontend/views/pages/Payments.vue
```diff
@@ -111,14 +118,23 @@ export default {
   },
   data () {
     return {
+      form: {
+        search: ''
```
sdegutis, line 122: Just a small suggestion: The name "search" here suggested an action, and I think "searchText" was more clear about the intent.

frontend/views/pages/Payments.vue
```diff
@@ -42,44 +42,51 @@ page(
...
-        .c-container-empty(v-else)
+            payments-pagination(
+              v-else
+              :count='paymentsFiltered.length'
+              :rowsPerPage='ephemeral.rowsPerPage'
+              :page.sync='ephemeral.currentPage'
+              @changePage='handlePageChange'
+              @changeRowsPerPage='handleRowsPerPageChange'
+            )
+        .c-container-noresults(v-else-if='paymentsListData.length && !paymentsFiltered.length' data-test='noResults')
```
sdegutis, line 85: Clever! I've always checked the presence of the search string. Your solution feels cleaner.

sdegutis (changes requested):
This finishes the review from earlier. Overall great code.

frontend/views/pages/Payments.vue
sandrina-p: That approach usually works very well, it's something I have in another project and works great

frontend/model/state.js
sandrina-p: A few months ago I tried to setup optional chaining in this project but couldn't make it work. Some error related with rollup and grunt (as you described). If you find a way to do it, please @ me in the MR so that I can see how it was done!

frontend/views/pages/Payments.vue
sandrina-p: Thank u!

frontend/views/pages/Payments.vue
sandrina-p: That's a nice way of thinking. It makes sense to keep the search indeed

test/cypress/integration/group-paying.spec.js
sandrina-p: I remember trying something similar a few months ago but couldn't make it work (The same problem happens with Testing Library too). If you find a solution please @ me so that I can learn it too :D

frontend/views/containers/payments/PaymentsPagination.vue
sandrina-p: It makes sense, I agree

sandrina-p:
@sdegutis, thank you for your review, I answered all your comments and agreed!

Unfortunately, I won't have the time to fix it this weekend :/

@taoeffect, to avoid waiting any longer, do you think it would be possible to merge this MR as it is and then @sdegutis could work on top of that and fix the remaining things?

Thank you all 🙌

frontend/views/pages/Payments.vue
mmbotelho: Agreed!

mmbotelho (approved):
[I can't replicate the error either](https://github.com/okTurtles/group-income-simple/pull/951#issuecomment-672931362).

In that case, PR approved!

frontend/model/state.js
sdegutis: Ah good to know, thanks 👍

sdegutis:
@sandrina-p Good idea.

@taoeffect I feel comfortable tackling the remaining issues per her suggestion, just let me know.

taoeffect:
> @taoeffect I feel comfortable tackling the remaining issues per her suggestion, just let me know.

@sdegutis Sure, in that case can you create an issue out of the issues you found here and self-assign? Once you do that reply back / let me know and I'll go ahead and merge this.

sdegutis:
👍 Will do.

sdegutis:
Okay @taoeffect it's at #983.

taoeffect (approved):
Great job @sandrina-p!

With #983 opened in leu of pending changes, I'm approving this and have marked #983 high-priority and assigned to @sdegutis to pick up!

PR #953, opened 2020-07-11
Fix cypress crash, probably related to cy.clock

sandrina-p:
Most of the recent tests after #918 was merged have been failing:
image

After some research, I didn't find the exact cause. But I suspect the problem was using `cy.clock()` in every single test. So I moved it so that is used only in the needed test. That seemed to fix the problem. (At least, locally in +10 runs, none crashed, compared to master where ~1/4 of the runs crashed)

### DIFF
--- a/test/cypress/integration/group-paying.spec.js
+++ b/test/cypress/integration/group-paying.spec.js
@@ -2,7 +2,7 @@
 // Similar to time.js but without the import.
 export function humanDate (datems, opts = { month: 'short', day: 'numeric' }) {
   const locale = navigator.languages ? navigator.languages[0] : navigator.language
-  return new Date(datems).toLocaleDateString(locale, { ...opts, timeZone: 'UTC' })
+  return new Date(datems).toLocaleDateString(locale, opts)
 }
 
 const userId = Math.floor(Math.random() * 10000)
--- a/test/cypress/integration/group-paying.spec.js
+++ b/test/cypress/integration/group-paying.spec.js
@@ -1,7 +1,16 @@
+
+// Similar to time.js but without the import.
+export function humanDate (datems, opts = { month: 'short', day: 'numeric' }) {
+  const locale = navigator.languages ? navigator.languages[0] : navigator.language
+  return new Date(datems).toLocaleDateString(locale, { ...opts, timeZone: 'UTC' })
+}
+
 const userId = Math.floor(Math.random() * 10000)
 const groupName = 'Dreamers'
 const mincome = 1000
-const timeStart = Date.UTC(2020, 1, 10)
+const timeStart = Date.now()
+const timeOneMonth = 60000 * 60 * 24 * 31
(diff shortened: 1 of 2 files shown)


test/cypress/integration/group-paying.spec.js
```diff
@@ -1,7 +1,16 @@
+
+// Similar to time.js but without the import.
+export function humanDate (datems, opts = { month: 'short', day: 'numeric' }) {
+  const locale = navigator.languages ? navigator.languages[0] : navigator.language
+  return new Date(datems).toLocaleDateString(locale, { ...opts, timeZone: 'UTC' })
```
taoeffect: If you're doing this here in the tests, then the change that was done to `humanDate` in the other file in the other PR needs to be reverted.

taoeffect (changes requested):
Minor fix needed

test/cypress/integration/group-paying.spec.js
sandrina-p: What change? It needs to be as it is

taoeffect: This change: https://github.com/okTurtles/group-income-simple/pull/918/files#diff-eb95d40daae3bff68bee21866026301eR63

```js
  const timezone = window.cypress ? { timeZone: 'UTC' } : {} // ensure tests work consistently anywhere in the world.
  return new Date(datems).toLocaleDateString(locale, { ...opts, ...timezone })
```

Keep the `humanDate` as it was before, and in the tests, display it without `timeZone: 'UTC'`.

So above, remove `timeZone: 'UTC'`, and revert the prior change. This will result in the time being displayed in the browser being in local timezone to whatever machine is running the tests, and it will make sure the string that's being compared in the tests is also local timezone.

sandrina-p: The timezone will be the same in both scenarios because it’s run through cypress.

taoeffect: Yes, so there is no need for `humanDate` to perform this extra computation by checking for `window.cypress` and destructuring/restructuring an object.

sandrina-p: It has because humanTime at time.js can be run outside cypress. If we remove it, the test will fail in your timezone. (Not now, but in a few hours when it’s a different day between here and in pacific time)

taoeffect: I did this locally. Removing the `timeZone` parameter keeps both strings displaying in my local timezone. So there is no problem. You don't need to display the time in UTC, you just need to make sure that the timezones match. The problem before was that they weren't matching because you had hand-written `'Feb 10'`. Since it's no longer being hand-written, there should be no problem.

sandrina-p: ahh, I finally understood what you meant. Changed!

taoeffect (approved):
Great job!

Issue #429, opened 2018-06-19
Add flow-typed


taoeffect:
### Problem

Currently, not only does flow now type check third-party dependencies/modules, but we have to manually add and ignore them via `shared/declarations.js`.

It turns out there's a much nicer solution...

### Solution

https://github.com/flowtype/flow-typed/blob/master/README.md

snowteamer:
Unfortunately, flow-typed is missing a lot of libdefs for our current dependencies.

Out of the 83 dependencies found in our package.json, running `flow-typed update` shows that it only covers the following ones:
- chalk
- register
- node-fetch
- localforage
- mocha
- sinon

So for all the other deps it would only generate stubs. This would still be an improvement, but less useful than fully typed module definitions.

taoeffect:
Not needed (considering switching to TypeScript instead)

Issue #528, opened 2019-02-04
Clean up flow errors


taoeffect:
### Problem

A bunch of new code has introduced a bunch more flow errors thanks to missing type annotations (and less often, due to bugs in the code).

To keep the software bug free, it's important to keep `npm run flow` happy. Although it can be a pain to do so, we should at the very least periodically do so.

### Solution

Go through all of the `npm run flow` errors and fix them. For bonus points, remove the `.*/test` related ignores in `.flowconfig` and fix the resulting flow errors.

*This issue depends on #521 being merged first, so check if it's merged before taking this on!*

Related: #429

PR #959, opened 2020-08-19
#429 Add flow typed

snowteamer:
This installs flow-typed and configures it to install third-party libdefs in the new folder `./shared/libdefs` alongside `./shared/declarations.js` 
  
### Additional information
- The [flow-typed docs](https://github.com/flow-typed/flow-typed/wiki/FAQs#why-do-i-need-to-commit-the-libdefs-that-flow-typed-installs-for-my-project) recommend to commit the libdefs, but this results in a large number of files being checked in all at once. I wouldn't disagree to reset a commit to let a maintainer take care of commiting the libdefs.
Edit:
In fact I think I disagree with theses recommendations, and that committing the libdefs is a bit unnecessary.

- It will be sometimes necessary to run `npm run flow-typed:install` or `npm run flow-typed:update` after the usual `npm install` and `npm update` to keep the libdefs in sync.

- When creating other npm scripts to run flow-typed, it will be important to always pass `--libdefDir ./shared/libdefs` to enable flow-typed to find the libdefs. It is not yet possible to configure this option in the .flowconfig file.


sdegutis:
Thanks for the PR.

* You make a good point, it seems better not to commit the type files.
* I wonder if we could automate creating the files in an NPM hook after `install` and `update` run?
* Perhaps there could be a common shell script that the different NPM commands all invoke.

snowteamer:
@sdegutis Thank you for your advice. So I've followed it and added the libdefs path to the .gitignore file.

I had some trouble correctly uncommiting the files but now it should be fine!

> I wonder if we could automate creating the files in an NPM hook after install and update run?

That would certainly help keep the type files up to date!

sdegutis:
Looks good. The next step would be to remove the newly-ignored files from the repo with `git rm --cached shared/libdefs/*` and commit that.

They also need to be re-created on every NPM install/update command. There's a [`postinstall` hook](https://docs.npmjs.com/misc/scripts). I don't see any `postupdate` hook, but the `postinstall` hook may be called after `npm update` too.

snowteamer:
> Looks good. The next step would be to remove the newly-ignored files from the repo with git rm --cached shared/libdefs/* and commit that.

You are right, the files were not properly ignored before I did that.

snowteamer:
> They also need to be re-created on every NPM install/update command.

Indeed, and that is the reason why I've added the `flow-typed:install` and `flow-typed:update` npm scripts.
However, I also agree that including them in a postinstall script would be more convenient.

package.json
```diff
@@ -14,6 +14,8 @@
     "eslintfix": "eslint \"**/*.{js,vue}\" --fix",
     "i18n": "node scripts/i18n.js",
     "flow": "flow",
+    "flow-typed:install": "flow-typed install --ignoreDeps bundled peer --libdefDir ./shared/libdefs",
```
sdegutis: Looks good. Once this is extracted into a `postinstall` hook, I can test it and merge it.

I'm not too concerned about the repetition of arguments at this point, since there's only going to be three instances of it in this file (these two and the new one), and it's not likely to change very often.

But if you're feeling up to it, feel free to extract a bash command that runs `flow-typed ${cmd} --ignoreDeps bundled peer --libdefDir ./shared/libdefs` and can be invoked as `some-cmd install` in `flow-typed:install` and `postinstall`, and `some-cmd update` in `flow-typed:update`.

Or better yet, if we find that the `postinstall` hook is sufficient for most cases, and if `npm update` runs it, we can just leave the one instance there.

sdegutis (changes requested):

sdegutis (approved):
Looks good. Thanks.

package.json
snowteamer: That makes sense.
Besides, given that the project uses only exact semver versions for its dependencies, for me it looks like `npm update` is never used.
If that's correct, then the postinstall script will likely be enough.

taoeffect:
If it's OK with you guys, can we save merging this PR for last? Perhaps after the other PRs are reviewed, merged, and also after #528 (clean up flow errors) is closed?

I am torn between having superior flow support through flow-typed, and the massive amount of new dependencies this PR adds through dependencies-of-dependencies, and the conflict that brings with #372. I'd like some more time to think this over if that's OK.

sdegutis:
Good idea. (I was not aware of #372 as a specific requirement, good to know.)

snowteamer:
Waiting is fine for me too. To be honest, the vast majority of the typings offered by this PR are stubs, so they don't bring much real type safety.

Ideally we should create the typings ourselves (and submit them to the flow-typed repo), but this would likely require a non-trivial amount of work.

snowteamer:
Closing this for now as it seems to make Travis do extra work

Issue #683, opened 2019-09-25
Create eslint plugin for i18n tag usage


taoeffect:
### Problem

1. We don't have documentation for how to use i18n tag (open issue: #607)
2. Even if we did, people still might not follow it, and a lot of PR reviews lately have been correcting its usage

### Solution

If we had an eslint plugin for how to use the i18n tag, we could enforce correct i18n usage and save a lot of time on PR reviews, and also guarantee that i18n is used in a way the works for the parser.

Related:

- https://twitter.com/swyx/status/1289306584938508288

snowteamer:
I could be interested in tackling this, once thorough documentation is available regarding allowed and disallowed usages of the tag.

But it will likely have to be very detailed, almost like pseudo-code for me to grasp the requirements, since I'm not yet familiar with Vue nor Pug myself.

taoeffect:
@snowteamer Yes at the moment we don't have documentation on its usage, however it is possible to discern the use-cases from the code and here's why: the `strings` utility that @SGrondin developed has support for all of the usages that we have, and I think it will output an error if the use-case is incorrect.

So if you decide to take on this issue, then as you're developing the eslint plugin for `i18n` usage for Pug (and ideally `L` function usage as well for JS), then you should be able to find all of the instances that you've missed, because your plugin (I'm guessing) will output linter errors for the parts you haven't yet supported.

snowteamer:
So if I understand correctly, that workflow would cover the missed usages, but I hope it can also detect the wrong usages as well

taoeffect:
Let's say there are 6 usages for i18n. If you implement 1 and tell it to error on anything else, you'll discover the other 5.

taoeffect:
Closed in #683 by @snowteamer !

Issue #607, opened 2019-08-08
Add documentation on how to use L function and i18n tag in Pug


taoeffect:
### Problem

It's unclear what the rules are for using the `L` function in Pug.

### Solution

Add documentation showing how and when to use the `L` function, along with `v-html` [as shown here](https://github.com/okTurtles/group-income-simple/pull/606/commits/f846ae8dead1ce7f1083ed3f2ad278876364d123) for dealing with markup.

Basically, `i18n` should be used except when:

**UPDATE: Some of the stuff written below is outdated! See [this pull request for latest usage rules](https://github.com/okTurtles/group-income-simple/pull/727)**

- Passing in string literals as props to a component (e.g. `page-section(:title='L("foo")')`)
- String needs to include markup, like `<b>foo</b>`, in which case ~~use `v-html`~~ (see edit below)
- `template` is used, in which case you need to use it like this:

   ```pug
   template(#title='') {{ L('Group Settings') }}
   
   //- OR like this, but note this inserts an extra <span> tag
   template(#title='')
     i18n Group Settings
   ```

EDIT: this has been changed in PR #611, see [comment](https://github.com/okTurtles/group-income-simple/pull/606/commits/f846ae8dead1ce7f1083ed3f2ad278876364d123#r313655392), copied here:

> NOTE: I have changed this (since the above doesn't work very well in my experience, especially not with variables as well), in my PR #611 — now the way to this is simple:
> 
> ```pug
> i18n(tag='p' html='This means you will stop having access to the <b>group chat</b> (including direct messages to other group members) and <b>contributions</b>. Re-joining the group is possible, but requires other members to vote and reach an agreement.')
> ```
> 
> **EDIT**: the above is now invalid! There should be no more use of the `html=` attribute. **See [this PR for how to do it now](https://github.com/okTurtles/group-income-simple/pull/727)**

**EDIT:** see if using [literal HTML in Pug](https://pugjs.org/language/plain-text.html#literal-html) makes this any cleaner/easier.

PR #960, opened 2020-08-25
#683 Add pug-lint plugin to validate i18n tag usage

snowteamer:
Related to #607 

* The companion test file should describe what this plugin allows and disallows.
* To keep things simple, this plugin currently defines a single new rule `validateI18n`, which doesn't take any option.
* I was not sure where the new files should go, so for now I have put them under a new `~/pug-lint-rules` folder.
* The error messages do not report exact error locations, only the location of the relevant `i18n` or `:args` token. This is intentional, as even the best linters sometimes suffer from issues with location reporting.

### Boyscout rule
* Fixed the corresponding Grunt task `exec:puglint` to also work on older Windows shells.

### Ideas for future work
- Check usages of the `L()` function.


sdegutis (approved):

sdegutis (dismissed):
Looks good, thanks.

taoeffect (changes requested):
Fantastic @snowteamer!

Just two file/organization related changes requested:

- Could you please move `validate-i18n.test.js ` into the `test/` folder and make sure that `grunt test` runs the file
- Place `validate-i18n.js` into the `scripts/` folder and delete the no longer needed `pug-lint-rules` folder to keep the top-level folder from getting too crowded 😄

taoeffect (approved):
Fantastic, professional job @snowteamer! Very impressive! Approved!

Issue #961, opened 2020-08-26
Finish i18n translation/localization toolchain — fetch JSON and update UI


taoeffect:
### Problem

Our localization toolchain is almost finished, and the final missing piece is for the frontend to fetch the proper `.json` file from the server for the user's local language and update the UI.

This will require updates to both the backend and frontend code.

### Solution

**Backend**

Be able to serve the translated .json file that the client requests.

**Frontend**

For now as an alpha version we will do this incorrectly and inefficiently, and in a future update we will do it properly.

So a PR for this issue will simply fetch the json file by using either `navigator.language` or `navigator.languages[0]` or `navigator.languages[1]` (do thorough research on this and find out what the most correct choice is!). It will then save this translation into memory and do the mapping from `original english` => `new language` inside of `translations.js`.

A future issue will be opened to make this more efficient by versioning the translations file and caching it locally in IndexedDB, but for now this simple method should get us up and running.

mmbotelho:
Question - how will we decide which language to display to each user by default?

Note: There should be a setting somewhere where users can change the language.

taoeffect:
> Question - how will we decide which language to display to each user by default?

This is decided by the user's system, which affects the values of those variables I mentioned in the issue.

mmbotelho:
Can you give me a bit more detail on that @taoeffect ? Is it the system language, the browser language, the location language...?

taoeffect:
The system language, which might or might not be overridable by the browser, but ultimately we're going to go off of the values of those variables, however they're set.

snowteamer:
The value of `navigator.language` can be split into a country code and a culture code, but currently we don't have culture-specific strings file so I guess that only the language part will be used

taoeffect:
> The value of `navigator.language` can be split into a country code and a culture code, but currently we don't have culture-specific strings file so I guess that only the language part will be used

That's right, whatever translation is available is what will be used. If it turns out that culture-codes are necessary then we will implement it in a way that looks up first the specific translation, and then the more general translation. I think to keep things simple we'll start out by only using the first part (e.g. `en`, not `en-US`).

snowteamer:
Also, please note that currently, the files `french.strings` and `english.strings` have a few different keys.

This is because I thought it was a good idea to fix a few copy errors while translating, but now I realize that it's ineffective without also patching the corresponding strings in the .vue files before re-running the `strings` tool.

[...]

PR #963, opened 2020-09-02
#961 Implement i18n

snowteamer:
### Changes:

- A new selector `sbp('translations/init', language)` registered in views/utils/translations.js,
 used to initialize the in-memory translation table, fetching it if necessary;
- A new route `'translations/get/:language'`,
 used to fetch a translation table for a given language, if one is available.
- A new selector `sbp('backend/translations/get', language)` registered in controller/backend.js,
 used inside the `translations/get` route.

I also wished to add a `language` setting in the Vuex store, so that users could change and save their preferred language. But if the translation table is changed, the strings that are already rendered don't get reactively updated, since the `L()` function and the `i18n` component are not reactive.

Is there a way to make the components which use these functions to call them again to update their displayed text?
Meanwhile, since I could not make this setting reactive I did not include it in the store.

### Other ideas for future work:

- Add Cypress tests. I'm not familiar enough with Cypress yet so any help is welcome.
- Allow more strings to be displayed in the locale language by only rendering them once the relevant ressource has been fetched.
- Maybe improve the L() and/or `template()` functions to handle localization of numbers and dates that are sometimes embedded in the text.

### DIFF
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -168,9 +168,9 @@ route.GET('/file/{hash}', {}, function (request, h) {
   }
 })
 
-route.GET('/translations/get/{language}', {}, function (request, h) {
+route.GET('/translations/get/{language}', {}, async function (request, h) {
   try {
-    return sbp('backend/translations/get', request.params.language)
+    return await sbp('backend/translations/get', request.params.language)
   } catch (err) {
     logger(err)
     return Boom.notFound()
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -8,6 +8,7 @@ import { blake32Hash } from '~/shared/functions.js'
 import { SERVER_INSTANCE } from './instance-keys.js'
 import chalk from 'chalk'
 import './database.js'
+import './translations.js'
 
 const Boom = require('@hapi/boom')
 const Joi = require('@hapi/joi')
@@ -161,6 +162,15 @@ route.POST('/file', {
 route.GET('/file/{hash}', {}, function (request, h) {
   try {
     return sbp('backend/db/readFile', request.params.hash)
+  } catch (err) {
+    logger(err)
+    return Boom.notFound()
+  }
+})
+
+route.GET('/translations/get/{language}', {}, function (request, h) {
+  try {
+    return sbp('backend/translations/get', request.params.language)
   } catch (err) {
     logger(err)
     return Boom.notFound()
--- a/backend/translations.js
+++ b/backend/translations.js
@@ -29,7 +29,9 @@ export default sbp('sbp/selectors/register', {
     try {
       return await readFileAsync(filePath)
     } catch (error) {
-      throw new Error(`No translation file was found for ${language}.`)
+      throw new Error(
+        `Could not read translation file for language '${language}': ${error.message}`
+      )
     }
   }
 })
--- a/backend/translations.js
+++ b/backend/translations.js
@@ -29,7 +29,7 @@ export default sbp('sbp/selectors/register', {
     try {
       return await readFileAsync(filePath)
     } catch (error) {
-      throw new Error(`No translation file was found for language ${language}.`)
+      throw new Error(`No translation file was found for ${language}.`)
     }
   }
 })
--- a/backend/translations.js
+++ b/backend/translations.js
@@ -1,0 +1,35 @@
+'use strict'
+
+import fs from 'fs'
+import path from 'path'
+import util from 'util'
+
+import sbp from '~/shared/sbp.js'
+
+const readFileAsync = util.promisify(fs.readFile)
+
+const defaultLanguage = 'en-US'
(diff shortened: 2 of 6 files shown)


backend/translations.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default sbp('sbp/selectors/register', {
+  'backend/translations/get': async function (language: string): Promise<string> {
+    if (language === defaultLanguage) return ''
+    const [languageCode] = language.toLowerCase().split('-')
+    const fileName = languageFileMap.get(languageCode) || ''
+
+    // If we don't have a matching file yet,
+    // returns empty so that the UI stays in the default language.
+    if (!fileName || !fileName.endsWith(languageFileExtension)) return ''
```
taoeffect, line 27: 1. `languageFileExtension` is unnecessary because if the mapping exists it is guaranteed to end in `.json`.
2. If there is an error, then return an error, e.g. `Boom.notFound()`, not an empty string

backend/translations.js
```diff
@@ -0,0 +1,35 @@
...
+    const filePath = path.join(languageDir, fileName)
+    try {
+      return await readFileAsync(filePath)
+    } catch (error) {
+      throw new Error(`No translation file was found for language ${language}.`)
```
taoeffect: This assumes the file was not found. That's not necessarily the error. There could be a bunch of other possible errors happening here. So include `error.message` (the reason) in the error message.

backend/routes.js
```diff
@@ -166,3 +167,12 @@ route.GET('/file/{hash}', {}, function (request, h) {
     return Boom.notFound()
   }
 })
+
+route.GET('/translations/get/{language}', {}, function (request, h) {
+  try {
+    return sbp('backend/translations/get', request.params.language)
```
taoeffect: You defined `'backend/translations/get'` as `async` so *I think* it needs to be treated that way here too in order to catch any thrown exceptions.

frontend/views/utils/translations.js
```diff
@@ -1,11 +1,58 @@
...
+      translationTable = defaultTranslationTable
+      return
+    }
+    try {
+      currentTranslationTable = await sbp('backend/translations/get', language)
+
+      // Only set `currentLanguage` if there was no error fetching the ressource.
+      currentLanguage = language
+    } catch (error) {
+      console.log(error)
```
taoeffect: All errors should be logged as `console.error`.

taoeffect (changes requested):
Fantastic @snowteamer! I noticed only a few minor issues. Great work!

backend/translations.js
taoeffect: Another possibility is to avoid the try/catch here, avoid making this function async/await, and simply `return readFileAsync(filePath)` and let the try/catch of the caller handle any exceptions from the returned promise.

snowteamer: That's right, this message is confusing.

backend/routes.js
snowteamer: That was for consistency with `route.GET('/file/{hash}'` where the return statement does not use `await` either, although the returned value is a Promise.

taoeffect: Hmm. Fair point. Both should be consistent. I think both versions will work because I'm pretty sure Hapi handles returning an error when a promise is returned that is rejected. I just think it means that if there's an error in the promise, the error handler that is run is Hapi's and not ours, so in that case both this and the implementation of `route.GET('/file/{hash}'` might need updating, but it's no big deal. We can ignore this for now.

taoeffect: Ah, I see you updated it. That's great! Leave it as-is with the latest commit you made. Great job! This will remind us to deal with `/file/{hash}` too

taoeffect (approved):
Perfect! Simple solution for #961. 👍 

We can then later on open issues to perfect our language loading support with caching and whatnot. Excellent work @snowteamer, very impressive! 🏆

backend/translations.js
snowteamer: 1. I agree that it's not very useful, in fact I've only used that constant to avoid embedding a string in the function body. Similarly, I've also defined `defaultLanguage` for the same reason, while the default language will likely never change from `en-US`.

2. I supposed it was better to return empty because I didn't consider that case to be an error or an exception, but rather to be the default. Quite in the same way that accessing a value from an object or map doesn't throw but returns a default value like `undefined` if the key was not found.
On the contrary, if a file name was found in `languageFileMap` but no real file was found or no data could be read, then an error certainly occured.

taoeffect: Ah, I missed that this comment wasn't addressed in the followup commit. Feel free to tweek/optimize this in a future PR if you'd like!

Also, I'm about to merge updated strings after running `strings.mac`, see PR #964... although there are some issues I think @SGrondin needs to have a look at .... keep an eye on that PR

PR #964, opened 2020-09-03
ran strings.mac to sync .strings and .json

taoeffect:


taoeffect:
Screen Shot 2020-09-02 at 8 21 15 PM

@SGrondin I was about to simply merge this PR since no code was changed, but have a look at the above, it seems to wipe out existing translations...? It would be horrible if we lost work done by translators, maybe you can figure out what's going on here?

snowteamer:
The translations could likely still be retrieved and re-applied, so they would not be lost, but I agree this would require a little extra work.

SGrondin:
The bug was due to a missing semicolon in `french.strings` and unfortunately the strings tool would just silently stop reading past that point, which is why it looked as if we had lost translations. I've fixed the semicolon, updated the tool to report syntax errors, and added support for .js files (instead of just .vue).

Issue #972, opened 2020-09-07
Update French translation


snowteamer:
### Problem

Following #964, new strings were generated in `french.strings`, meaning they currently have no translation.

Moreover, since #961 we can now review translations in the context of the running application, which should certainly help improve a few of them.

### Solution

Update the translation accordingly.

PR #973, opened 2020-09-07
#972 Update french.strings

snowteamer:
Here are a few issues I'm aware of:

- I am not sure how to correctly translate the names of the voting systems, "Disagreement number" and "Percentage based" in the context of the UI.

  So for now I used the following:
  - "Disagreement number" -> "Sysème basé sur le nombre de votes négatifs"
  - "Percentage based" -> "Système basé sur le pourcentage des votes positifs"

- There are various case inconsistencies througout the strings, it would be nice to have a clear rule.


strings/french.strings
```diff
@@ -1270,7 +1270,7 @@
 "Read more" = "Lire la suite";
 
 /* views/containers/proposals/ProposalBox.vue */
-"{strong_}You{_strong} are proposing:" = "{strong_}Vous{_strong} proposez:";
+"{strong_}You{_strong} are proposing:" = "{strong_}Vous{_strong} proposez :";
```
taoeffect, line 1273: Is it normal in French to have a space before colon?

taoeffect: Oh OK, I see that apparently in fr-FR it is: https://github.com/okTurtles/group-income-simple/pull/958#discussion_r475359117

taoeffect:
> I am not sure how to correctly translate the names of the voting systems, "Disagreement number" and "Percentage based" in the context of the UI.

I think we will change these strings in the future to "Disagreement rule" and "Percentage rule". You are free to translate them as such in this PR if you'd like, or update them later.

strings/french.strings
```diff
@@ -58,64 +58,64 @@
...
-"Not part of the group." = "Not part of the group.";
+"Not part of the group." = "Ne fait pas partie du groupe.";
 
 /* MISSING TRANSLATION - model/contracts/group.js */
-"Only the group creator can remove members." = "Only the group creator can remove members.";
+"Only the group creator can remove members." = "Seul le fondateur du groupe peut exclure des membres.";
 
 /* MISSING TRANSLATION - model/contracts/group.js */
-"There is an identical open proposal." = "There is an identical open proposal.";
+"There is an identical open proposal." = "Une proposition ouverte identique existe déjà.";
```
pieer: Une proposition similaire a déjà été créée.

strings/french.strings
```diff
@@ -58,64 +58,64 @@
...
-"{username} profile" = "{username} profile";
+"{username} profile" = "Profil de {username}";
 
 /* MISSING TRANSLATION - views/components/ProfileCard.vue, views/pages/DesignSystem.vue */
-"Remove member" = "Remove member";
+"Remove member" = "Exclure un membre";
 
 /* MISSING TRANSLATION - views/components/VotingRulesInput.vue */
-"\"No\" votes required to block a proposal" = "\"No\" votes required to block a proposal";
+"\"No\" votes required to block a proposal" = "Nombre de votes négatifs requis pour bloquer une proposition";
```
pieer, line 115: Aucun vote négatif n'est requis pour bloquer une proposition

strings/french.strings
```diff
@@ -133,40 +133,40 @@
...
 /* MISSING TRANSLATION - views/containers/chatroom/MainHeader.vue */
 "{{ description }}" = "{{ description }}";
 
 /* MISSING TRANSLATION - views/containers/contributions/ContributionItem.vue */
-"{service} from {who}" = "{service} from {who}";
+"{service} from {who}" = "{service} de la part de {who}";
 
 /* MISSING TRANSLATION - views/containers/dashboard/GroupMembers.vue */
-"See all {groupMembersCount} members" = "See all {groupMembersCount} members";
+"See all {groupMembersCount} members" = "Afficher tous les {groupMembersCount} membres";
```
pieer: "Afficher les {groupMembersCount} membres"

strings/french.strings
```diff
@@ -181,85 +181,85 @@
...
-"percentage based" = "percentage based";
+"percentage based" = "basé sur le pourcentage des votes positifs";
 
 /* MISSING TRANSLATION - views/containers/proposals/ProposalItem.vue */
-"Change disagreement number from {X} to {N}." = "Change disagreement number from {X} to {N}.";
+"Change disagreement number from {X} to {N}." = "Faire passer le seuil de rejet de {X} à {N}.";
 
 /* MISSING TRANSLATION - views/containers/proposals/ProposalItem.vue */
-"Change from a disagreement based voting system to a percentage based one, with minimum agreement of {percent}." = "Change from a disagreement based voting system to a percentage based one, with minimum agreement of {percent}.";
+"Change from a disagreement based voting system to a percentage based one, with minimum agreement of {percent}." = "Passer d'un système basé sur le nombre de votes "non" à un système basé sur les pourcentages, avec un seuil minimum d'approbation de {percent}.";
```
pieer: "Passer d'un système basé sur le nombre de votes négative à un système basé sur les pourcentages, avec un seuil minimum d'approbation de {percent}."

strings/french.strings
```diff
@@ -181,85 +181,85 @@
...
 
 /* MISSING TRANSLATION - views/containers/proposals/ProposalItem.vue */
-"Change from a percentage based voting system to a disagreement based one, with a maximum of {N} “no” votes." = "Change from a percentage based voting system to a disagreement based one, with a maximum of {N} “no” votes.";
+"Change from a percentage based voting system to a disagreement based one, with a maximum of {N} “no” votes." = "Passer d'un système basé sur les pourcentages à un système basé sur le nombre de votes "non", avec un maximum de {N}.";
```
pieer: Passer d'un système basé sur les pourcentages à un système basé sur le nombre de votes négatifs, avec un maximum de {N}.

pieer (commented):
Awesome work. Few details maybe to improve and one mistake on 'no' translation

strings/french.strings
snowteamer: I'm a bit unsure about this one. The text is displayed for a slider that can only accept values between 1 and 60, so at least one negative vote is always required.

Besides, if your interpretation is correct; then the quotes around the `"No"` in the English version are very confusing, and should be removed I think.

strings/french.strings
snowteamer: Is it possible to keep "identical"? I don't understand why two similar but still not identical proposals could not be created.

taoeffect: I think we should keep the original meaning of "identical" unless there is a strong reason not to

pieer: I still prefer similaire in french to be honest, but I'm ok with this: Une proposition identique a déjà été créée.

strings/french.strings
pieer: You're right, I didn't see that one in its context.
I think that the english version is a bit awkward then. @taoeffect ?

taoeffect: I think the idea was, `1 "no" votes required to block a proposal`, `2 "no" votes required to block a proposal`, etc.

So if whatever you write makes the same sense in French it's fine. Yes I agree it's a bit awkward but it works.

pieer: All right then! We can close that one

taoeffect (approved):
Awesome! Thank you @snowteamer, great job! And thanks @pieer for the review!

PR #976, opened 2020-09-09
Restore flags to map Docker ports.

### LINKED ISSUE #971: Fix docker script
### Problem

I [broke](https://github.com/okTurtles/group-income-simple/commit/2c8ec32de76b88f65f78e7f0ce2a1aa2fde3b369#diff-285f5a875bc5897c56a0b51aea6307ffL12-L27) the networking in a change by removing these lines:

```sh
PORT_SHIFT=${PORT_SHIFT:-0}	
FRONTEND_PORT=$(( 8000 + $PORT_SHIFT ))	
BACKEND_PORT=$(( 3000 + $PORT_SHIFT ))	
REFRESH_PORT=$(( 35729 + $PORT_SHIFT ))

# .... [snip] ...

  -p 127.0.0.1:${FRONTEND_PORT}:${FRONTEND_PORT} \
  -p 127.0.0.1:${BACKEND_PORT}:${BACKEND_PORT} \
  -p 127.0.0.1:${REFRESH_PORT}:${REFRESH_PORT} \
```

### Solution

I think I removed them because I didn't want to have to make assumptions in the script about what the base port was (e.g. whether it was 8000 or 3000). However, removing these lines breaks the networking entirely, and I probably mistakenly thought `--ipc=host` would make it work, but I probably meant to write `--network=host`.

So either change it to `--network=host` (and verify that that fixes the problem), or add back those lines (and later one figure out how to guarantee they are synced between code and script).

F1LT3R:
- Fixes #971 
- All tests passing locally.

I restored the previous code for mapping Docker port to host in `bin/docker.sh`.

Note: The `--network=host \` approach did not work. It appears [this feature is unsupported](https://docs.docker.com/network/host/) on macOS and Windows.

### DIFF
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -13,14 +13,23 @@ if ! docker images -a | egrep -q "\b$PROJECT_NAME\b"; then
   docker build -t "$PROJECT_NAME" --build-arg TIMEZONE="$TZ" .
 fi
 
+# TODO: sync/configure the ports below if they're changed
+# https://github.com/okTurtles/group-income-simple/issues/71
+PORT_SHIFT=${PORT_SHIFT:-0}
+FRONTEND_PORT=$(( 8000 + $PORT_SHIFT ))
+BACKEND_PORT=$(( 3000 + $PORT_SHIFT ))
+REFRESH_PORT=$(( 35729 + $PORT_SHIFT ))
+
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Simple PR, great job!

PR #988, opened 2020-09-17
#987 fix payment cypress tests

### LINKED ISSUE #987: Cypress tests failing on master branch
### Problem

The recent modals PR (#946) caused a change in how the Payment object was accessed in the Payment Details modal. This has led to the payment missing a `monthstamp` property, causing a test to fail because it sees "Invalid Date" in the modal.

### Solution

Somehow make these properties available to this component.

sdegutis:
Resolves #987. Then @sandrina-p's other PR can be merged and my other PR can be finished and merged.

### DIFF
--- a/frontend/views/utils/currencies.js
+++ b/frontend/views/utils/currencies.js
@@ -26,7 +26,7 @@ function validateMincome (value, currency) {
   return isNumeric(nr) && isInDecimalsLimit(nr, currency)
 }
 
-function decimalsOrInt (num: number, currency: string): string {
+function decimalsOrInt (num: number, currency: string): string | number {
   const fixed = parseFloat(num.toFixed(currencies[currency].decimalsMax))
   const integerPart = /^(.+)\.0+$/.exec(fixed)
   return integerPart ? integerPart[1] : fixed
--- a/frontend/views/utils/currencies.js
+++ b/frontend/views/utils/currencies.js
@@ -26,30 +26,8 @@ function validateMincome (value, currency) {
   return isNumeric(nr) && isInDecimalsLimit(nr, currency)
 }
 
-const reasonablePrecisionBlacklist = ['BTC']
-
-function withReasonableDecimals (num: string, currency: string): string {
-  // Skip BTC
-  if (reasonablePrecisionBlacklist.includes(currency)) return num
-
-  // Leave '250' as '250'
-  if (!num.includes('.')) return num
-
-  // Expand 12.5 to 21.50
-  const places = currencies[currency].decimalsMax
-  const match = num.match((/\.(\d+)/))
-  if (match && match[1].length < places) {
-    const integerPart = match.index
-    const decimalPoint = 1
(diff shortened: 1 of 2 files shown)


sdegutis:
Note: this doesn't yet resolve the duplication issue, but it adds a note to resolve it. I can try to tackle that on Monday before this PR is merged, or we can merge it and try to resolve it afterward.

frontend/views/utils/currencies.js
```diff
@@ -26,10 +26,30 @@ function validateMincome (value, currency) {
...
+    num = num.padEnd(integerPart + decimalPoint + places, '0')
+  }
+  return num
+}
+
+function decimalsOrInt (num: number, currency: string): string | number {
   const fixed = parseFloat(num.toFixed(currencies[currency].decimalsMax))
   const integerPart = /^(.+)\.0+$/.exec(fixed)
-  return integerPart ? integerPart[1] : fixed
+  return withReasonableDecimals(integerPart ? integerPart[1] : fixed.toString(), currency)
```
taoeffect: This file is getting unwieldy and there were some things in #950 that did not get reviewed properly. Let's fix all of that in this PR.

The function `withReasonableDecimals` is doing way too much, and looks to be entirely unnecessary in the face of `decimalsOrInt`, which is supposed to do what `withReasonableDecimals` does. So let's fix `decimalsOrInt` instead, and clean up some of the stuff that was missed in #950.

1. There is a bug in `decimalsOrInt`, which is on line 51 here, and that is that `fixed` is a float, and yet it is being regex'd on in line 51. That is wrong. `parseFloat` shouldn't have been called in line 50.  The call to `.toFixed` should do nearly everything that `withReasonableDecimals` attempts to do, the only thing missing is to account for `.00`, and that is what line 51 and 52 should be doing.
2. `withReasonableDecimals` should be removed unless for the some reason the short & sweet `decimalsOrInt` cannot do what it does (but AFAICT, that is exactly what `decimalsOrInt` was written to do)
3. `decimalsOrInt` should return a string, not a number, and the type definitions should be updated accordingly, and all type issue should be fixed.
4. The review of #950 missed that `commaToDots` should take a string or a number and return a number. Everything got confused in that PR because strings and numbers were being mixed together and without any consistent type annotations, an no clear rules being followed. Actually running flowtype would have caught many of those errors. `saferFloat`'s type definition should be restored. It should take a number or string and return a number.
5. `isInDecimalsLimit` is missing type annotations, and so are other functions. Every single function in this file should be type annotated, both for the parameters and the return value, and all of the actual internal usage of those parameters and the actual return values must match the type annotations.

taoeffect (changes requested):
Review done!

frontend/views/utils/currencies.js
sdegutis: Now I understand this file a lot better, thanks.

taoeffect (approved):
Approved and merging while opening #989 for the remaining unaddressed review!

PR #992, opened 2020-09-21
#989 clean up currencies file

### LINKED ISSUE #989: Finish cleaning up currencies.js
### Problem

To speed things along we decided to merge #988 before finishing addressing all of the remaining cleanup/boy-scouting for `currencies.js`. Here I'm copying [the comment](https://github.com/okTurtles/group-income-simple/pull/988#discussion_r491088408) that got partially addressed:

> This file is getting unwieldy and there were some things in #950 that did not get reviewed properly. Let's fix all of that in this PR.
> 
> The function `withReasonableDecimals` is doing way too much, and looks to be entirely unnecessary in the face of `decimalsOrInt`, which is supposed to do what `withReasonableDecimals` does. So let's fix `decimalsOrInt` instead, and clean up some of the stuff that was missed in #950.
> 
> 1. (NOTE: this was done ✅)  There is a bug in `decimalsOrInt`, which is on line 51 here, and that is that `fixed` is a float, and yet it is being regex'd on in line 51. That is wrong. `parseFloat` shouldn't have been called in line 50.  The call to `.toFixed` should do nearly everything that `withReasonableDecimals` attempts to do, the only thing missing is to account for `.00`, and that is what line 51 and 52 should be doing.
> 2. (NOTE: this was done ✅) `withReasonableDecimals` should be removed unless for the some reason the short & sweet `decimalsOrInt` cannot do what it does (but AFAICT, that is exactly what `decimalsOrInt` was written to do)
> 3. (NOTE: this was done ✅) `decimalsOrInt` should return a string, not a number, and the type definitions should be updated accordingly, and all type issue should be fixed.
> 4. The review of #950 missed that `commaToDots` should take a string or a number and return a number. Everything got confused in that PR because strings and numbers were being mixed together and without any consistent type annotations, an no clear rules being followed. Actually running flowtype would have caught many of those errors. `saferFloat`'s type definition should be restored. It should take a number or string and return a number.
> 5. `isInDecimalsLimit` is missing type annotations, and so are other functions. Every single function in this file should be type annotated, both for the parameters and the return value, and all of the actual internal usage of those parameters and the actual return values must match the type annotations.

### Solution

Finish cleaning up and addressing all type usage, while making sure that no new bugs are introduced by checking all usage across all files.

Clean up and simplify anything else you notice.

sdegutis:
Resolves: #989

**Actions taken:**

* Added types for *almost* everything in currencies.js
* Simplified parameters to remove redundant information (`currencies` key)
* Removed duplication in `currencies` entry definitions
* Removed an unnecessary `.toString()` in one spot and added a necessary `.toString()` in another.

**Concerns:**

* `mincomePositive` and `validateMincome` are called from `validations` but I can't figure out what exact types it *always* gets. It seems to always be a string from my usage of the app, and since it's coming from the UI, I assume so, but I haven't fully traced down how/where these validators are called.
* Tracked down many calls to `displayWithCurrency` and `displayWithoutCurrency` which are always numbers, but haven't found them all yet and not sure if it's ever a string.

### DIFF
--- a/frontend/views/containers/contributions/IncomeDetails.vue
+++ b/frontend/views/containers/contributions/IncomeDetails.vue
@@ -64,7 +64,7 @@ modal-base-template(ref='modal' :fullscreen='true' :a11yTitle='L("Income Details
 
     group-pledges-graph.c-graph(
       :type='form.incomeDetailsType'
-      :amount='form.amount === "" ? undefined : saferFloat(form.amount)'
+      :amount='form.amount === "" ? undefined : normalizeCurrency(form.amount)'
     )
 </template>
 
@@ -72,7 +72,7 @@ modal-base-template(ref='modal' :fullscreen='true' :a11yTitle='L("Income Details
 import { mapGetters } from 'vuex'
 import { validationMixin } from 'vuelidate'
 import { required } from 'vuelidate/lib/validators'
-import currencies, { mincomePositive, saferFloat } from '@view-utils/currencies.js'
+import currencies, { mincomePositive, normalizeCurrency } from '@view-utils/currencies.js'
 import sbp from '~/shared/sbp.js'
 import PaymentMethods from './PaymentMethods.vue'
 import GroupPledgesGraph from './GroupPledgesGraph.vue'
@@ -151,7 +151,7 @@ export default {
     }
   },
   methods: {
-    saferFloat,
+    normalizeCurrency,
     resetAmount () {
       this.form.amount = this.form.incomeDetailsType === this.ourGroupProfile.incomeDetailsType ? this.ourGroupProfile[this.ourGroupProfile.incomeDetailsType] : ''
       this.$v.form.$reset()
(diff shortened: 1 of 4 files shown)


sdegutis:
~~This isn't fully ready to merge.~~ It was failing because `commaToDots()` sometimes received null from `saferFloat`. By taking this into account and adding a simple fix, it passes tests now. But it shouldn't take null, and this function shouldn't allow that. (Edit: Fixed.)

sdegutis:
When tracing usage of `saferFloat` to fix that bug, it seems that #950 introduced a second behavior of `saferFloat` that I understand better now after having discussed this file with @taoeffect.

In particular, beforehand it seems to exclusively have been used to [reduce precision in an existing number](https://github.com/okTurtles/group-income-simple/issues/813#issuecomment-593680834). But now it's also used to normalize user-input strings.

So I've split out that function into two functions: `saferFloat` retains the original precision-reduction behavior, and `normalizeCurrency` calls `saferFloat` after converting commas to dots. I've also changed the usages introduced in #950 to use `normalizeCurrency`, and ensured all tests pass.

taoeffect (approved):
@sdegutis I have to say, this is your best PR yet!! Fantastic work! Everything looks great to me and you made the right call with `saferFloat`! Excellent job researching that!

PR #994, opened 2020-09-22
Implement Notifications Trigger, Card and Modal - desktop

taoeffect:
This should be a squashed version of #934.

@F1LT3R — please double-check that is indeed the case. Then, if you're certain there are no differences between this and #934, you make pull it locally and continue working on top of it. Have a look through the differences, see if there are any improvements you can make to the code to simplify it, to address TODOs, etc.

Below is the collective commit message history:

```
NotificationsList - layout

Create fakeDB with templates

Add link "See All"

Continue Notifications layout

Continue Notifications

add badge on mobile

Implement timeSince() with tests

Hide scrollbars when non-existent.

Notification title "new" spacing tweak

Changes based on @mmbotelho review

Changes based on @mmbotelho review

Add missing guards in router.js (#982)

* [#987] Fixing invalid date error in Cypress payment test.

* [#987] Correcting type annotation on decimalsOrInt() helper function.

* [#987] Fixing payment total's decimal precision.

* [#987] Adding TODO note about future payment helper functions.

* [#987] Simpler implementation of stringifying currency.

Co-authored-by: Steven Degutis <sbdegutis@gmail.com>

* [#989] Add types to currencies.js

* [#989] Removing duplication and type-cycles in currencies.js file.

* [#989] Caught and removed unnecessary .toString()

* [#989] Adding assumed types to remaining currencies.js file types.

* [#989] Remove trailing commas to satisfy linter.

* [#989] Allow commaToDots() to receive null.

* [#989] Don't allow commaToDots() to receive null.

* [#989] Start IncomeDetails's form.amount to '' for consistency.

* [#989] Separate user-input-normalization from currency-precision-reduction.

Co-authored-by: Steven Degutis <sbdegutis@gmail.com>

* No longer clear search when changing Payment tabs

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486665087

* Renamed search -> searchText for semantic clarity

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486666965

* Moved inline HTML/text into Vue view.

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486655374
Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486656616

* Simplified logic that determines which Payment tabs to show.

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486545255

* Return to i18n using a full sentence including variables.

* [#983] Separate some scriptio continua in payments test suite.

[...]

PR #995, opened 2020-09-22
Task/663 notifications layout 3

taoeffect:
This should be a squashed version of #934, and a fixed version of the first attempt at squashing the commits (#994).

Whoever would like to pick this up is welcome to! Simply continue working in the branch. I recommend also double-checking that there are no differences between this and #934. See if there are any improvements you can make to the code to simplify it, to address TODOs, etc.

Below is the collective commit message history:

```
NotificationsList - layout

Create fakeDB with templates

Add link "See All"

Continue Notifications layout

Continue Notifications

add badge on mobile

Implement timeSince() with tests

Hide scrollbars when non-existent.

Notification title "new" spacing tweak

Changes based on @mmbotelho review

Changes based on @mmbotelho review

Add missing guards in router.js (#982)

* [#987] Fixing invalid date error in Cypress payment test.

* [#987] Correcting type annotation on decimalsOrInt() helper function.

* [#987] Fixing payment total's decimal precision.

* [#987] Adding TODO note about future payment helper functions.

* [#987] Simpler implementation of stringifying currency.

Co-authored-by: Steven Degutis <sbdegutis@gmail.com>

* [#989] Add types to currencies.js

* [#989] Removing duplication and type-cycles in currencies.js file.

* [#989] Caught and removed unnecessary .toString()

* [#989] Adding assumed types to remaining currencies.js file types.

* [#989] Remove trailing commas to satisfy linter.

* [#989] Allow commaToDots() to receive null.

* [#989] Don't allow commaToDots() to receive null.

* [#989] Start IncomeDetails's form.amount to '' for consistency.

* [#989] Separate user-input-normalization from currency-precision-reduction.

Co-authored-by: Steven Degutis <sbdegutis@gmail.com>

* No longer clear search when changing Payment tabs

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486665087

* Renamed search -> searchText for semantic clarity

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486666965

* Moved inline HTML/text into Vue view.

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486655374
Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486656616

* Simplified logic that determines which Payment tabs to show.

Resolves https://github.com/okTurtles/group-income-simple/pull/951#discussion_r486545255

* Return to i18n using a full sentence including variables.

* [#983] Separate some scriptio continua in payments test suite.

Co-authored-by: Steven Degutis <sbdegutis@gmail.com>
```


frontend/assets/style/_layout.scss
```diff
@@ -91,7 +91,7 @@ $bannerGeneralHeight: 2.3rem;
 .l-page {
   position: relative;
   grid-area: l-page;
-  overflow-y: scroll;
+  overflow-y: auto;
```
pieer: Could you explain why this has been changed? I'm not sure to understand atm

frontend/assets/style/_utilities.scss
```diff
@@ -62,24 +62,24 @@ $fills: (
 
 .hide-desktop {
   @include desktop {
-    display: none;
+    display: none !important;
```
pieer: This should not happen. Could you tell me why you did this? I would like to check that this is not an error from somewhere else

frontend/views/components/modal/Modal.vue
```diff
@@ -120,8 +120,9 @@ export default {
         this.subcontent.pop()
       } else {
         this.content = null
-        // Refocus on button that open the modal
-        this.lastFocus.focus()
+        // Refocus on button that open the modal if it exists.
```
pieer: It's either the element is on the page and we can focus again or it is not, then we don't need to focus back on this element

frontend/views/components/modal/ModalClose.vue
```diff
@@ -97,7 +97,6 @@ export default {
     outline: none;
   }
 
-  &:focus,
```
pieer: Why did we removed the focus indication?

frontend/views/containers/navigation/Navigation.vue
```diff
@@ -242,4 +245,11 @@ export default {
 .c-toggle {
   left: 100%;
 }
+
+// Need this nesting because of order of CSS
+.c-navigation .c-toggle .c-toggle-badge {
```
pieer: I don't see anything that prevent using only .c-toggle-badge selector only on this page.

frontend/assets/style/_utilities.scss
snowteamer: Maybe the original author did this for consistency with other `hide-` rules here, which all have  `!important`. It doesn't seem to break anything.

frontend/views/components/modal/ModalClose.vue
snowteamer: I don't know, this seems to make keyboard-based navigation more difficult since there is no longer any visual feedback on the close button when tabbing to or from it. So I'm adding this back for now.

frontend/assets/style/_layout.scss
snowteamer: I'm not sure either. This doesn't seem related to notification layout (unless I'm missing something), and I haven't noticed any visible effect for now. Do you think this change should better be reverted?

pieer (dismissed):
Great work @snowteamer

taoeffect (commented):
Really coming along quite well! I seems almost GTM, only some minor issues noticed in my tests on Firefox Dev Edition:

*Requesting changes on this as a Comment because Github won't let me request changes for a PR I opened myself*

## 1 - Weird highlight issues on hover on Desktop

- Some items sometimes will *de*-highlight when you hover over them
- Have a look at the video for more weird glitches possibly, and reference Figma design notes for proper behavior in case of doubt

notification-highlight-bug

## 2 - On mobile, spacing between notification content and timestamp doesn't exist

Screen Shot 2021-11-19 at 4 52 01 PM

Per Figma, it should look like this:

Screen Shot 2021-11-19 at 4 58 06 PM

## 3 - merge `master` into this PR to make testing easier

This I think I'll do myself and will push the changes. It helps with testing so that there's no need to re-run `npm install` to install outdated deps.

taoeffect (commented):
Approved! Congrats @sandrina-p and @snowteamer!

PR #997, opened 2020-09-23
#935 modal improvements (part 2)

### LINKED ISSUE #935: Using modals

sdegutis:
Tentatively resolves #935 as a follow-up to #946 (see update in #593.

This PR specifically takes the shortest path to resolving [this architectural question](https://github.com/okTurtles/group-income-simple/pull/946#discussion_r493737696), by removing an event listener added in #946, as well as another bug fix and some cleanup. It leaves open the option of:

1. Merging it as-is and going with this as the main solution, while leaving it open to further thought and discussion
2. Merging it and adding extra work in a follow-up PR
3. Waiting to merge it until this solution is discussed further

### DIFF
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -37,6 +37,7 @@ export default {
     sbp('okTurtles.events/off', OPEN_MODAL)
     sbp('okTurtles.events/off', CLOSE_MODAL)
     sbp('okTurtles.events/off', REPLACE_MODAL)
+    sbp('okTurtles.events/off', SET_MODAL_QUERIES)
     window.removeEventListener('keyup', this.handleKeyUp)
   },
   watch: {
--- a/frontend/views/containers/payments/PaymentDetail.vue
+++ b/frontend/views/containers/payments/PaymentDetail.vue
@@ -64,13 +64,9 @@ export default {
   computed: {
     ...mapGetters([
       'currentGroupState',
-      'ourGroupProfile',
       'ourUsername',
       'userDisplayName'
     ]),
-    needsIncome () {
-      return this.ourGroupProfile.incomeDetailsType === 'incomeAmount'
-    },
     withCurrency () {
       return currencies[this.payment.data.currencyFromTo[0]].displayWithCurrency
     },
--- a/historical/PaymentsListMenu.vue
+++ b/historical/PaymentsListMenu.vue
@@ -10,7 +10,7 @@ menu-parent
         tag='button'
         item-id='message'
         icon='info'
-        @click='openModal("PaymentDetail", { payment, needsIncome })'
+        @click='openModal("PaymentDetail", { id: payment.hash })'
       )
         i18n Payment details
 


taoeffect (approved):
Approved!

PR #1000, opened 2020-10-07
WIP: #763 - Fixes "Chunk A" group of bugs

sdegutis:


frontend/model/state.js
```diff
@@ -355,6 +355,9 @@ const getters = {
   groupIncomeDistribution (state, getters) {
     return groupIncomeDistribution({ state, getters, monthstamp: currentMonthstamp() })
   },
+  groupAdjustedIncomeDistribution (state, getters) {
+    return groupIncomeDistribution({ state, getters, monthstamp: currentMonthstamp(), adjusted: true })
+  },
```
taoeffect, line 360: Why is this being created instead of using the existing `groupIncomeAdjustedDistribution` which does the same thing?

sdegutis: You're probably right, it's probably the same thing. I looked for this but didn't find it at first. I'll take a closer look at this code again Monday.

But the bigger question is about "ourContributionSummary" which is described in [this comment](https://github.com/okTurtles/group-income-simple/issues/763#issuecomment-704541338). That's the real thing I need feedback here, and the main reason I opened this WIP PR.

taoeffect (commented):
This PR is currently [missing this part](https://github.com/okTurtles/group-income-simple/issues/763#issuecomment-658225618):

>  Now the only thing left is to not make it a partial payment:

The tests should also verify that too, i.e. in Scenario 1 make sure that it's not displayed as a partial payment when it shouldn't be, and fail if it is.

sdegutis:
> This PR is currently [missing this part](https://github.com/okTurtles/group-income-simple/issues/763#issuecomment-658225618):
> 
> > Now the only thing left is to not make it a partial payment:
> 
> The tests should also verify that too, i.e. in Scenario 1 make sure that it's not displayed as a partial payment when it shouldn't be, and fail if it is.

Will work on that next.

In the meantime, feel free to check the code in this PR that solves the 5 scenarios.

sdegutis:
Although my patch fixes those scenarios in *the unit tests*, the Scenario 3 *Cypress test* is failing, in a strange way:

| Recipient | Unit Test | Cypress Test |
|-|-|-|
| u2 | $5.769230769230769 | $15 |
| u3 | $69.23076923076924 | $60 |

(The other Scenarios are passing in the Cypress tests.)

It doesn't make sense why the same function is having two different results. It seems there are more calculations being done here. This is an example of where I would reach for a debugger to step through the code and find out what's happening.

Besides that, at least two other Cypress tests are now also failing:

* user1 sends $71.43 to user2 (total)
* user1 sends $100 to user3 (partial)
* (and probably more)

They have different amounts of payments with different amounts, and they show "-1 out of 2" for payments sent.

I'll translate them into the unit tests, to see if we get a similar difference as the table above.

sdegutis:
Based on my understanding of the proportional-split algorithm:

* Add up all the *need*, which in Scenario 3 is $325 ($25 for u2 + $300 for u3)
* Divide the excess (u1's $75) based on each person's relative need based on the absolute need
  * u2 needs $25 out of $325 total need, which is 0.076923076923077 of the need, of the $75 excess = $5.769230769230769
  * u3 needs $300 out of $325 total need, which is 0.923076923076923 of the need, of the $75 excess = $69.230769230769231

Based on this understanding, correct or not, the unit tests are correct, and the Cypress tests show the wrong result.

I'm not sure yet how the Cypress tests came up with this answer.

sdegutis:
The next thing in this PR that needs a set of eyes is [this comment](https://github.com/okTurtles/group-income-simple/issues/763#issuecomment-711084166).

frontend/utils/giLodash.js
```diff
@@ -246,3 +246,7 @@ export function deepEqualJSONType (a: any, b: any): boolean {
   }
   return true
 }
+
+export function remapObject (o, f) {
+  return Object.fromEntries(Object.entries(o).map(([k, v]) => [k, f(v)]))
+}
```
taoeffect: Nicely done.

sdegutis:
AFAIK there was a destructive version at the top of the file called mapValues which matches the name of it in lodash, but I needed an immutable one. But I could be wrong on both counts.

frontend/utils/giLodash.js
taoeffect: > AFAIK there was a destructive version at the top of the file called mapValues which matches the name of it in lodash, but I needed an immutable one. But I could be wrong on both counts.

Two things:

1. I didn't realize that, thank you for pointing it out. Having looked them both over, it seems that indeed `remapObject` looks to be a re-implementation of `mapValues`, and I'm guessing it's slower because of the multiple loops. `mapValues` doesn't modify the original object passed in (the third parameter is modified, but it's initialized to an empty object by default, and normally isn't used). You should be able to replace the `remapObject` usage with it without any trouble. So I think this implementation doesn't have a reason to exist, as elegant as it appears.
2. Just letting you know about a Github quirk I just discovered thanks to your reply: Github doesn't place your reply in the correct place if the comment is an inline-code-comment, instead the reply goes to the PR, so to observers it looks weird.

frontend/utils/distribution/group-income-distribution.js
```diff
@@ -1,40 +1,253 @@
...
+/*
+JS is giving us 20 * (10 / 18) * (18 / 50) = 3.9999999999999996 instead of 4
+*/
+
+function distibuteFromHavesToNeeds({ haves, needs }) {
+  const totalHave = haves.reduce((a, b) => a + b.have, 0)
+  const totalNeed = needs.reduce((a, b) => a + b.need, 0)
+  const totalPercent = Math.min(1, totalHave / totalNeed)
+
+  for (const have of haves) have.percent = have.have / totalHave
```
taoeffect, line 62: Does this need its own loop or can it be done inside of the other `for (const have of haves)` loop down below on line 66?

frontend/utils/distribution/group-income-distribution.js
```diff
@@ -1,40 +1,253 @@
...
+  let payments = distibuteFromHavesToNeeds({ haves, needs })
+
+  for (const event of events) {
+    if (event.type === 'payment') {
+      handlePayment({ payments, needs, payment: event })
+    }
+    else if (event.type === 'join') {
+      const newPayments = handleJoin({ user: event, needs })
+      payments = payments.concat(newPayments)
+    }
```
taoeffect: Minor: I have made it a habit to recognize if/else blocks like this, and write them in a way that is slightly different. Since if/else always creates a branch, and since branches slow things down (and can also make code slightly less readable in some situations), I use a map in situations like this:

```js
for (const event of events) {
  ({
    payment: event => handlePayment({ payments, needs, payment: event }),
    join: event => {
      const newPayments = handleJoin({ user: event, needs })
      payments = payments.concat(newPayments)
    }
  })[event.type](event)
}
```

It also has the added benefit of automatically handling unknown types by throwing an exception.

You don't need to adopt this method if you don't like it. I bring it up only as a neat trick I picked up from my C days of function tables for creating branchless code. It starts to pay off especially when there is a large number of events, and the body of code for each event is large.

sdegutis: Thanks for letting me know. If you mean it slows things down in the JavaScript VM, looped function creation and execution is probably much slower than if-branching, although performance doesn't really matter here.

frontend/utils/distribution/group-income-distribution.js
sdegutis: We could recalculate it every time, sure. But I did this partly to help with debug-logging so that I could see the variables in the console.

frontend/utils/giLodash.js
sdegutis: Good points, thanks.

frontend/utils/distribution/group-income-distribution.js
taoeffect: Oh yeah, I don't know if the JS compiler would actually do that, but if it did then yeah it would make sense to move the definition of that object outside of the loop.

taoeffect:
Closing in favor of #1009.

PR #1003, opened 2020-10-17
[#980] Still compile despite linter errors.

### LINKED ISSUE #980: Change linter to still compile
### Problem

When writing code, if there are stylistic errors, the linter will refuse to compile changes. This interrupts the cognitive flow of the developer, by changing focus away from the bigger-picture change to minor details about code style. This has the potential to slow down development significantly.

### Solution

The linter should emit warnings by default, but still compile. During a CI phase, a flag could be passed to change these to errors that fail the CI tests. This ensures that we can still see whether the linter is satisfied by the time code gets to PRs, but allows developers more freedom to code naturally and flowingly, and fix up stylistic differences at a more convenient time.

sdegutis:
Fixes #980. Works with both scenarios:

* Start project with a semicolon in something, see that it still compiles and runs
* Start project with no lint errors, add a lint error while it runs, see that it still recompiles and runs

Linter errors still show up in console, but this time without the unnecessary Grunt stack trace. So it's actually improved.

### DIFF
--- a/.Gruntfile.babel.js
+++ b/.Gruntfile.babel.js
@@ -116,7 +116,7 @@ module.exports = (grunt) => {
         options: { env: { LOAD_NO_FILE: 'true', ...process.env } }
       },
       // https://github.com/standard/standard/issues/750#issuecomment-379294276
-      eslint: 'node ./node_modules/eslint/bin/eslint.js "**/*.{js,vue}"',
+      eslint: 'node ./node_modules/eslint/bin/eslint.js --plugin only-warn "**/*.{js,vue}"',
       eslintgrunt: "./node_modules/.bin/eslint --ignore-pattern '!.*.js' .Gruntfile.babel.js Gruntfile.js",
       puglint: '"./node_modules/.bin/pug-lint-vue" frontend/views',
       stylelint: 'node ./node_modules/stylelint/bin/stylelint.js "frontend/assets/style/**/*.{css,scss,vue}"',
@@ -337,7 +337,7 @@ module.exports = (grunt) => {
           match: /\.scss$/,
           recurse: true
         }),
-        eslint({ throwOnError: true, throwOnWarning: true }),
+        eslint({ throwOnError: false, throwOnWarning: false }),
         svgLoader(),
         transformProxy({
           plugin: VuePlugin({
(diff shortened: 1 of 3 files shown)


taoeffect (commented):
I'm looking through the code and am not sure where the change is to ensure that Travis will fail when there are linter errors, did those make it in here or are they missing?

Also, we are still struggling to reduce the dependencies in this project (see #372), and as far as I know we almost never run `npm run eslint`, so is the `eslint-plugin-only-warn` truly needed, or is possible to avoid that addition?

sdegutis:
Good questions. This doesn't take care of Travis, I'll look into that now. I'll also check if we can inline the logic of that plugin.

sdegutis:
@taoeffect The plugin is used in the initial setup phase when `grunt dev` is run. It lints the whole project before anything else, before the watching-linter runs and lints, which is separate (and only needed trues changed to falses). So if you want to avoid plugins, we need to inline the whole logic of that plugin, which is a bit complex and might take a few hours.

taoeffect:
It sounds like the main issue you're running into is that during development sometimes the linter errors prevented your code changes from taking effect, and because of our setup, the messages got lost anyway too, is that right? If this plugin is only for the initial run of `grunt dev`, then would it be OK to allow `grunt dev` to display the errors/warnings on the initial run? It would save us another dependency to worry about.

sdegutis:
Good idea, will do.

sdegutis:
@taoeffect Okay fixed, see last commit's message for details. But I think it's ready to merge now.

taoeffect (approved):
I think, if memory serves, @sandrina-p also wanted this, and I shot down the idea before. Well, given the demand for it, it seems she was right all along. My apologies Sandrina, and thank you Steven for sending this in!

Approved! Will merge as soon as Travis shows green!

sandrina-p:
That's true @taoeffect. Another thing I wanted (and did only locally back at the time) was enabling auto-fix when possible.

Add "fix" or "autofix", not sure, can't remember exactly... That way I didn't need to fix the issues manually. It works very similar to [prettier](https://prettier.io/)

```
eslint({ throwOnError: false, throwOnWarning: false, fix: true }),
```

PR #1014, opened 2020-11-10
WIP: Continuation of PR 1000 - various payment bugs

taoeffect:
Squashed version of all changes and commits from #1000 for #763, please see that PR for details and original commentary. This PR is being picked up by @CrazedCoding 

```
[#763] Resolving Scenario 1.
[#763] Adding Cypress test for Scenario 2 which passes.
[#763] Adding Cypress test for Scenario 3 which passes.
[#763] Adding Cypress test for Scenario 4 which fails.
[#763] Splitting Scenario 4 into two tests to match GH issue.
[#763] Putting scenarios into test groups.
[#763] Commenting out code that currently has no effect on any tests.
[#763] Never mind. Something else must have been wrong.
[#763] Skip Scenario 4.1's test for now in Cypress...
[#763] Extract paymentTotalFromUserToUser so it can be used outside Vue.
[#763] Inlining unnecessary variables.
[#763] Extracting non-Vue version of groupIncomeDistribution().
[#763] Extracting object-remapping logic into helper function.
[#763] Wrote first test for group income distribution logic.
[#763] Wrote basic test for adjusted group income distribution logic.
[#763] Cleaning up whitespace.
[#763] Renaming test and adding new test for group income distribution.
[#763] Simplifying interface to paymentTotalFromUserToUser().
[#763] Rewriting some logic to use functional programming.
[#763] Adding comment to explain relative -> absolute conversion.
[#763] Splitting out group income distribution and adjustment logic.
[#763] Whitespace again.
[#763] Turned Scenarios 4 and 4.1 into unit test cases.
[#763] This whitespace is fine to get rid of, from/to becomes clearer.
[#763] Writing down notes that may help us solve failing test case.
[#763] Added unit test for Scenario 3.
[#763] Added unit test for Scenario 2.
[#763] Added unit test for Scenario 1.
[#763] Recombining group income distributions functions to pass test.
[#763] Fixing linter error that silently caused my code to not run.
[#763] Got Scenario 4.1 passing, but Scenario 3 fails now.
[#763] Added Scenario 3's test case to mincome-proportional tests.
[#763] Fixed Scenario 3's test case in mincome-proportional tests.
[#763] Deleting code that isn't used anymore.
[#763] Cleaning up tests a bit.
[#763] Changing Scenario 3's test to reflect what "should" happen.
[#763] Removing unused import.
[#763] Adding unit test to help explain/test distribution logic.
[#763] Removing unused import that's causing project to not compile.
[#763] Changing quote style that's preventing code from compiling.
[#763] Removing extra comma that's preventing code from compiling.
[#763] Changing the "Payments sent" numerator to only sent payments.
[#763] Turning first Cypress scenario into unit test.
[#763] Tentatively updating Cypress test to reflect new findings:
See https://github.com/okTurtles/group-income-simple/issues/763#issuecomment-712166451
[#763] Fixed bug in group income distribution logic.
[#763] Turned next failing Cypress test into a unit test.
[#763] Fixing more Cypress payment specs.
[#763] Turning current broken situation into a unit test.
[#763] Putting each expected test result array's items on one line.
[#763] Fixing a better-understood Scenario 3 by reverting code.
[#763] Fixing possibly flaky tests (only fails on precision):
Note, these might not be flaky long-term, they might only be different
now because of the new algorithm code in a recent commit. Since this
commit message will only ever show up in a squashed commit, I mean
that a commit or two ago, I went back to Greg's algorithm initially
posted in the github comments for this message, which I temporarily
took out and tried a different (incorrect) algorithm with.
[#763] Adding member join-date into group income distribution test cases.
[#763] Fixing and skipping second failing test.
[#763] Passing member joinedDate to function from UI.
[#763] Passing full creation timestamp in payments to algorithm.
[#763] Adding helper function to transform data into events.
[#763] Moving helper function above main functions.
[#763] Splitting haves/needs in helper function for convenience.
[#763] Adding another test case for helper function.
[#763] Making withDate argument explicit.
[#763] Fixing indentation.
[#763] Making sure helper function works without adjustWith.
[#763] Starting to rewrite tests using more semantic data.
[#763] Verifying data->event transformations for existing tests.
[#763] Creating sub-test-suite for tests using new helper function.
[#763] Doesn't seem necessary to give new function the mincome.
[#763] Exclude users who have no excess from algorithm.
[#763] Cleaning up test by removing useless variable.
[#763] Wrote new algorithm.
[#763] Converting old tests for new algorithm.
[#763] Current payments are deducted from todo payments.
[#763] Starting to distribute excess payments back to other payments.
[#763] A few other tests already pass.
[#763] A few more tests pass with altered order/precision.
[#763] Refactoring into multiple functions and trying a fix.
[#763] Using mapValues instead of remapObject
[#763] Changing Scenario 1 to represent added-income as "join" event.
```


taoeffect (commented):
Glad to see this progressing!

A couple comments based on a quick scan:

- Now is a good time to create a helper function to cleanup the test file so that it generates the data that's needed, and takes in the minimum amount of data, so that the tests and their data are easier to read. See how Steven did it.
- The parameter after `it` should be a very tiny short string, single sentence, no newlines. See how all the other existing tests are named. If you want to preserve the detailed info about the scenario, keep it in the tests as a code comment

CrazedCoding:
@taoeffect Points understood. I will update this stuff tomorrow.

frontend/utils/distribution/group-income-distribution.test.js
```diff
@@ -15,33 +55,22 @@ describe('Chunk 0: Adjustment Tests',
...
+              'incomeAmount': 10
             }
           },
-          monthstamp: '2020-11',
-          adjusted: true
-        })
+          'payments': [],
+          'mincome': 12,
+          'monthstamp': '2020-11',
+          'adjusted': true
```
taoeffect: I think this can be simplified to make it even more readable. Here are some suggestions:

- Use defaults so that you can leave out data. For example, if the `payments` is empty (as it is here), don't require it to be included. If the `monthstamp` is always the same for the purpose of the tests, fill it in with a default value so that it too doesn't need to be specified. If `adjusted: true` is the value for that most of the time, then have that be the default and only specify it when it changes. DRY!
- Anywhere you have things spread across multiple lines, where they could be specified on a single line within 70 columns width, make them a single line (e.g. `u2: { 'incomeDetailsType': 'incomeAmount', 'incomeAmount': 10 }`)
- Anywhere something can be rewritten in a simpler way, do it (e.g. `u2: { need: 2 }` instead of the previous version)

Then these tests will read like a children's book and will be both super easy to write and read and understand.

frontend/utils/distribution/group-income-distribution.test.js
```diff
@@ -0,0 +1,1852 @@
...
+            'u2': {
+              'incomeDetailsType': 'incomeAmount',
+              'incomeAmount': 10
+            }
+          },
+          'payments': [],
+          'mincome': 12,
+          'monthstamp': '2020-11',
+          'adjusted': true
+        }))
```
taoeffect: I think this can be simplified to make it even more readable. Here are some suggestions:

- Use defaults so that you can leave out data. For example, if the `payments` is empty (as it is here), don't require it to be included. If the `monthstamp` is always the same for the purpose of the tests, fill it in with a default value so that it too doesn't need to be specified. If `adjusted: true` is the value for that most of the time, then have that be the default and only specify it when it changes. DRY!
- Anywhere you have things spread across multiple lines, where they could be specified on a single line within 70 columns width, make them a single line (e.g. `u2: { 'incomeDetailsType': 'incomeAmount', 'incomeAmount': 10 }`)
- Anywhere something can be rewritten in a simpler way, do it (e.g. `u2: { need: 2 }` instead of the previous version)

Then these tests will read like a children's book and will be both super easy to write and read and understand.

taoeffect: Copy/pasting from our call:

Scenario 4 TODOs:

1. Split Scenario 4.0 into two scenarios:
  - New Scenario 4.0:
    ```js
    expect({
      mincome: $1000
      u1: have: $100
      u2: need: $50
      u3: need: $1000
      payments: [{ u1: { u3: 50 }}]
    }).toEqual({
      // TODO: fill this in with the TODOs that appear
      //       in the UI for all users. (e.g. who pays
      //       who how much.)
    })
    ```
  - New Scenario 4.1:
    ```js
    expect({
      mincome: $1000
      u1: have: $50
      u2: need: $50
      u3: need: $1000
      payments: [{ u1: { u3: 50 }}]
    }).toEqual({
      // TODO: fill this in with the TODOs that appear
      //       in the UI for all users. (e.g. who pays
      //       who how much.)
      //       
      //       Note, u1 is pleding $50, and there is an
      //       existing payment from u1 of $50, so the
      //       TODO for u1 shouldn't exist anymore.
    })
    ```
  - New Scenario 4.2:
  ```js
  expect({
    mincome: $1000
    u1: have: $50
    u2: need: $50
    u3: need: $1000
    u4: have: $150
    payments: [{ u1: { u3: 50 }}]
  }).toEqual({
    // NOTE: we are only showing what Sandrina wrote
    //       here as is expected for u4, and we are
    //       not including the values (that in the real
    //       tests should be here) for the rest of the users
    // ... [snip] ...
    u4: { u2: 50, u3: 50 } // NOTE: the output will be
                           //       look more complicated
                           //       i'm just simplifying
                           //       it's ok to include the
                           //       complicated expected
                           //       output, which will look
                           //       something like:
                           //       
                           //       { from: 'u4', to: 'u2',
                           //        amount: 50 }
    // ... [snip] ...
  })
  ```

taoeffect:
Closing in favor of #1056

PR #1015, opened 2020-11-11
Adjust thresholds for Percentage Rule

### LINKED ISSUE #1010: Adjust thresholds for Percentage Rule
### Problem

1. For some reason the threshold got switched from 75% to 80%, and I don't remember why. 75% intuitively makes more sense to me because a group of 4 people, with 3 voting yes and one voting no, seems like it should pass, and 3/4 = 75%
2. The low-threshold warning is currently displayed a bit too early. 66% should be fine too, and not need a warning.

### Solution

1. Unless there are objections, let's switch the default back to 75%
2. Let's make the low-threshold warning only appear at 59% and lower.

SebinSong:
closes #1010 

made changes to the number values per the issue description.


SebinSong:
will fix the test failure it there's any.

EDIT: All cypress test have passed on my local environment.
image

Let me add a random commit and see if the app passes this time..!

taoeffect (approved):
Great job @SebinSong! Another easy and simple fix! Fantastic! 😄

PR #1018, opened 2020-11-22
Finish "Avatar Placeholders" review

### LINKED ISSUE #981: Finish "Avatar Placeholders" review
### Problem

To speed things along we decided to merge #923 even though there was a pending request for changes.

Specifically two things were left unaddressed:

1. [Making `imageDataURItoBlob` support extraction of data type from the URI](https://github.com/okTurtles/group-income-simple/pull/923#discussion_r485243638)
2. [Replacing AvatarGenerator with an SBP function/selector](https://github.com/okTurtles/group-income-simple/pull/923#discussion_r487380628)
3. (Possibly) [generating the avatar inside of `'gi.actions/identity/create'`, and *IF* it fails for some reason, then using the default URL](https://github.com/okTurtles/group-income-simple/pull/923#discussion_r487380791)

### Solution

Submit a PR that addresses the above remaining issues.

SebinSong:
closes #981 

Things that have been worked on in this PR are,

1. improving `imageDataURItoBlob` so it extracts the image type from the Data URI.
2. deleting `AvatarGenerator.vue` and replacing it with a sbp selector/function.


frontend/views/utils/avatar.js
```diff
@@ -0,0 +1,26 @@
...
+    const avatarBase64 = blockies.create({
+      bgcolor: theme[`${palette}_0`], // darkest
+      color: theme[`${palette}_1`], // medium
+      spotcolor: theme[`${palette}_2`], // lightest
+      size: 6,
+      scale: 12
+    }).toDataURL('image/png')
+    const avatarBlob = imageDataURItoBlob(avatarBase64)
+
+    return imageUpload(avatarBlob)
```
taoeffect: Nice, this PR looks like it has simplified and cleaned up the code a good amount!

Question, this code (1) used to have a try/catch around it, and (2) is currently called in the following manner: `const picture = await sbp('gi.utils/avatar/create') || defaultAvatar` — so, are you ensuring that if  there is an error anywhere here things will be properly handled?

SebinSong: Thanks for catching it. I made a change where in case avatar creation goes wrong the avatar sbp function returns null. That way, the default avatar is used if an error occurs. ( `const picture = await sbp('gi.utils/avatar/create') || defaultAvatar` )

I confirmed it by manually throwing an error inside the `try {}` block. But pls let me know if you think there's a better way.

taoeffect: @SebinSong Here's the thing... look through the sequence of events and possible errors in the line you quoted: `const picture = await sbp('gi.utils/avatar/create') || defaultAvatar`

Ask yourself the following questions:

Can something else go wrong?

Yes, in fact it can: `await sbp('gi.utils/avatar/create')` can throw an exception, because the promise that `imageUpload` returns can be rejected.

Do we handle that?

No.

How should we handle that?

Well, think about this from the perspective of the user and go through the workflow step-by-step:

1. This code gets called if we're signing up a new user.
2. If an exception occurs on `imageUpload`, it could be due to a variety of situations. Enumerate them all:
  1. The user supplied their own picture and it was too large, so the server rejected it. (*Oops!* Looks like we already caught a bug! This PR seems to have removed the ability for a user to supply their own `picture`?  Line 11 in `frontend/controller/actions/identity.js`?)
     - How should this situation be handled? Is it currently handled? Should the signup fail with a user-facing message? Should we proceed with a generated avatar? What do you think would be the behavior that made the most sense to you if you were signing up and this happened? How should the app handle it? Does it currently do that? If not, can you make it behave the ideal way?
  2. `imageUpload` failed for some other reason. Perhaps the server is down? How are we handling this right now? If you decide to simply use `defaultAvatar`, how can you do that if the way the current code happens it won't ever reach the `|| defaultAvatar` because an exception was thrown because a promise was rejected?

Since `imageUpload` is outside of the `try/catch`, we encounter problems like this. Should it be outside of it? If so, then the caller should also have an appropriate `try/catch` to handle these scenarios. I encourage you to really think through as a user using the app. What do you think should happen in all of these situations? Then make sure the code does it.

So these are some of the issues that jumped out at me.

taoeffect (changes requested):
Nice job @SebinSong! The error handling is much improved, though I think I [spotted some other issues](https://github.com/okTurtles/group-income-simple/pull/1018/files#r531799806), but please double check me! I could be misunderstanding something.

frontend/views/components/banners/BannerSimple.vue
```diff
@@ -16,13 +16,13 @@ export default {
   props: {
     severity: {
       type: String,
-      default: 'info'
+      default: 'info',
+      validator: function (value) {
```
SebinSong, line 20: This change is not directly relevant to the issue but thought it was obvious this prop validation function was misplaced, so boyscouted it. But pls let me know if it needs to be put back.

frontend/views/utils/avatar.js
SebinSong: @taoeffect Thanks for the review.
Just put back the `picture` argument to the `gi.actions/identity/create` sbp function and make the random avatar generation happen in `gi.actions/identity/signup` as we discussed. This way, the both error cases you mentioned above are handled.

1. **in case random avatar generation goes wrong**: `gi.utils/avatar/create` returns `null` and the default avatar will be used.
2. **in case imageUpload fails**: `new GIErrorUIRuntimeError` will be thrown and the user will be presented with the error message saying what's gone wrong (via `banner-scoped.vue` component in the `SignupForm.vue`).

Pls let me know if there should be additional steps taken.

pieer (dismissed):

frontend/views/components/banners/BannerSimple.vue
taoeffect: Your change looks correct to me, nice catch! 👍

test/cypress/integration/group-paying.spec.js
```diff
@@ -118,7 +122,7 @@ describe('Group Payments', () => {
       cy.getByDT('amount').should('contain', '$71.43')
       cy.getByDT('subtitle').should('contain', `Sent to user2-${userId}`)
 
-      cy.getByDT('details').find('li:nth-child(2)').should('contain', humanDate(timeStart, { month: 'long', year: 'numeric' }))
+      cy.getByDT('details').find('li:nth-child(2)').should('contain', humanDate(dateToMonthstamp(new Date(timeStart)), { month: 'long', year: 'numeric' }))
```
SebinSong, line 125: Apparently the test failed because it didn't take into account the transformation happening in `L:55 of PaymentDetail.vue` [here](https://github.com/okTurtles/group-income-simple/blob/master/frontend/views/containers/payments/PaymentDetail.vue#L55), where the date becomes a monthstamp.
This change made all the tests pass in my local.

taoeffect (approved):
Approved! Excellent work @SebinSong! 🎉 💯 🥇

PR #1020, opened 2020-11-29
Serve website through Hapi + Update dependencies + Ngrok support

### LINKED ISSUE #751: Serve website using Hapi
### Problem

Currently we use `grunt dev` which runs primitive grunt middleware to serve the website.

EDIT: also, copying from #852:

> For example, in #780, @SebinSong was not able to sign up a user because the API is being served on "localhost:3000", which his client tries to connect to directly, instead of through the already existing HTTPS host connection on the ngrok domain. So placing the API on the same server will make it easier to proxy the connection using services like ngrok.

### Solution

Write/use Hapi to serve files. Look into [spazy](https://github.com/vokeio/spazy) and [inert](https://github.com/hapijs/inert) plugins.

Closing this issue will also probably involve moving off of Primus (#576) and switching to [hapi-plugin-websocket](https://github.com/rse/hapi-plugin-websocket).

It will also require replacing `grunt-contrib-connect` with our own homebrew solution. Have a look at some of these files:

- https://github.com/huanz/express-static-livereload/blob/master/index.js
- https://github.com/dai-shi/easy-livereload/blob/master/index.js
- https://github.com/intesso/connect-livereload/blob/master/index.js

### LINKED ISSUE #780: Unable to communicate with server(on trying to log in) using Ngrok
### Problem

Trying to test `grunt dev` server on mobile phone using [Ngrok](https://ngrok.com/) produces some issues(it appears that it fails to communicate with the server upon requests).

1) When trying to log in, with an account already created by other device(PC).

ngrok_unable_to_login

2) When trying to sign up on mobile device.

ngrok1

P.S.
@taoeffect I tried the other way you advised me first(connecting the phone with the same wifi as the laptop and then trying http://ipv4_address:8000 which worked this time), and also was presented with the same error messages.

### Solution

Close #852

### LINKED ISSUE #936: Grunt watch reload gets too slow over time
### Problem

When the server is started `grunt dev`, each time a file is changed the reload time takes ~8-10sec, which is _ok'ish_. This is already too slow compared to other tools, but the normal time for this project.

Over time, as we do more reloads, the reload time starts to increase as well, reaching 30-40s to reload the page.

This causes a High Usage of CPU and my PC starts to overheating. (up to 90ºC easily).

Besides that, it decreases a lot my productivity over time.  A task that could take 3h probably will take +4h because of this slow reloading. (and I already do an extra effort to avoid saving a file too soon to spare some extra reloads)

So, our only solution is to restart the server and recreate the previous state (Cypress helps here) to get back to normal reload times. However, it's a matter of time to start having slow reloads again.

### Solution

I tried to understand what's the cause but without success. :/

taoeffect:
+ Now using browsersync to do livereload (Might fix #936? Please let us know!)
+ Hapi serves SPA and static files (Closes #751)
+ Full Ngrok support (Closes #780), easy to access site over a proxy or a tunnel now, both web sockets and API and static files are served over a single port. Run `ngrok http 8000` after `grunt dev` finishes (this is a direct connection to the hapi server instead of the browsersync proxy, which is on port 3000)
+ You can do `grunt dev --tunnel` for built-in tunneling support through `*.loca.lt`! This is pretty much just like `ngrok` except it seems to be less finicky about data caps, but maybe more finicky in other ways. 
+ Fixed: SignupForm previously would fail to display validation error if a valid username was supplied, deleted, the tab key was pressed.
+ `grunt test:unit` now always skips the rollup build (since there's no reason to do it)

### Updated dependencies

+    "@hapi/hapi": "20.0.3",
+    "buffer": "6.0.3",
+    "form-data": "3.0.0",
+    "localforage": "1.9.0",
+    "primus": "8.0.0",
+    "systemjs": "6.8.0",
+    "tweetnacl": "1.0.3",
+    "tweetnacl-util": "0.15.1",
+    "vue": "2.6.12",
+    "vue-router": "3.4.9",
+    "vue-slider-component": "3.2.11",
+    "vuelidate": "0.7.6",
+    "vuex": "3.6.0",
+    "wicg-inert": "3.1.0",
+    "ws": "7.4.0"
+    "@babel/core": "7.12.9",
+    "@babel/plugin-proposal-class-properties": "7.12.1",
+    "@babel/plugin-proposal-object-rest-spread": "7.12.1",
+    "@babel/plugin-transform-runtime": "7.12.1",
+    "@babel/preset-env": "7.12.7",
+    "@babel/preset-flow": "7.12.1",
+    "@babel/register": "7.12.1",
+    "@babel/runtime": "7.12.5",
+    "@rollup/plugin-eslint": "8.0.0",
+    "node-fetch": "2.6.1",
+    "vue-template-compiler": "2.6.12",
+    "acorn": "8.0.4",
+    "babel-plugin-root-import": "6.6.0",
+    "chalk": "4.1.0",
+    "chokidar": "3.4.3",
+    "eslint": "7.14.0",
+    "eslint-config-standard": "16.0.2",
+    "eslint-plugin-flowtype": "5.2.0",
+    "eslint-plugin-import": "2.22.1",
+    "eslint-plugin-standard": "5.0.0",
+    "eslint-plugin-vue": "7.1.0",
+    "flow-bin": "0.138.0",
+    "flow-remove-types": "2.138.0",
+    "grunt": "1.3.0",
+    "mocha": "8.2.1",
+    "rollup": "2.33.3",
+    "rollup-plugin-vue": "5.1.9",
+    "sinon": "9.2.1",
+    "stylelint": "13.8.0",
+    "stylelint-config-standard": "20.0.0",
+    "vue-cli-plugin-pug": "2.0.0"


taoeffect:
Note: when testing ngrok, you might run into "error 429" or error 500 errors -- this happens because ngrok caps the number of requests you can do per minute, and after loading up the app you'll probably hit that cap. So you have to wait like 30 seconds before it allows you to do stuff again. Or you can pay them money to remove the cap. :P

frontend/views/containers/access/SignupForm.vue
```diff
@@ -91,29 +96,51 @@ export default {
...
-          return !await sbp('namespace/lookup', value)
+  // we use dynamic validation schema to support accessing this.usernameAsyncValidation
+  // https://vuelidate.js.org/#sub-dynamic-validation-schema
+  validations () {
+    return {
+      form: {
+        username: {
+          [L('A username is required.')]: required,
+          [L('A username cannot contain spaces.')]: nonWhitespace,
+          [L('This username is already being used.')]: async (value) => {
```
SebinSong, line 114: Love how this part has been debounced. 👍

pieer (approved):
Good job!  Nothing to add

Issue #852, opened 2020-02-27
User accounts and real login


taoeffect:
### Problem

In order to close #780 we need a real user accounts system.

Also, in order to make it easier for sysadmins to run the server, by exposing only a single port, it might be wise to place the API and the static files (see #751) under same port, with `/api/*` serving the API, and everything else being static files.

For example, in #780, @SebinSong was not able to sign up a user because the API is being served on "localhost:3000", which his client tries to connect to directly, instead of through the already existing HTTPS host connection on the ngrok domain. So placing the API on the same server will make it easier to proxy the connection using services like ngrok.

### Solution

- Implement/enable the ability to persist contract messages on the server
- Implement persisting the namespace (the mapping from usernames to identity contracts)
- Create a master device that controls the main account via a keypair, and encrypt it using the user's password. (Create an issue for handing account recovery if they lose the encrypted blob)
- Later on add the ability to designate secondary devices that can use the account

Close related issues:

- Create base contract protocol #603
- Key management #92
- Implement Keychain interface #112
- Unable to communicate with server(on trying to log in) using Ngrok #780
- Serve website using Hapi #751 (likely involves closing #576 as well)

PR #1028, opened 2020-12-03
Persistence implemented

taoeffect:
This PR partially implements #852, adding persistence of groups and accounts.

To test locally, run like so:

```bash
$ GI_PERSIST=true grunt dev
```

To clear out the "database", quit the server and run:

```bash
$ rm -rf data/*
```

Missing still is the ability to associate an account with a password. That's coming after the issues with Cypress in this are fixed.

### WIP Notes

Although the Cypress tests pass when testing with the in-memory storage, they fail when running using persistent storage. Travis doesn't catch this because it's not yet configured to test the persistent storage method. Am currently working to address this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
 dist: trusty
 compiler: clang # per: https://docs.travis-ci.com/user/languages/c
 env:
+  - NO_PERSIST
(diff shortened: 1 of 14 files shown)


pieer (dismissed):
Nice one!

Issue #1027, opened 2020-12-02
Add i18n linter rules for using vue variables in i18n tags


taoeffect:
### Problem

While reviewing #1004 I [noticed](https://github.com/okTurtles/group-income-simple/pull/1004/files#r534499657) that [our `i18n` linter](https://github.com/okTurtles/group-income-simple/blob/master/scripts/validate-i18n.js) that @snowteamer wonderfully implemented appears to be missing a rule to catch the use of double-braces (Vue.js variables):

```pug
i18n Replying to {{replyingTo}}
```

### Solution

Throw an error that says Vue.js variables aren't allowed, and to use `:args` parameter with i18n tag and single-braces in the string to pass in variables.

/cc @snowteamer

snowteamer:
So shall we just disallow any occurrence of `{{` in the `i18n` tag?

This would be much simpler to identify than checking if there is a valid JavaScript expression between matching double curly braces, and as far as I know double braces are unlikely to be used in an `i18n` tag outside Vue.js variables.

taoeffect:
> So shall we just disallow any occurrence of `{{` in the `i18n` tag?

Yep!

taoeffect:
Closed by @snowteamer in ##1031!

PR #1031, opened 2020-12-05
#1027 Disallow double curly braces in i18n strings

snowteamer:
### Changes

- Make the `scripts/validate-i18n` script check whether i18n strings contain any occurence of the `{{` pattern, which should be enough in our use case to safely detect incorrect usage of Vue.js variables
- Add a corresponding test in `test/validate-i18n.test.js`

When running this updated linting plugin against the codebase, two occurences of double curlies were detected, causing the build to fail:

lint-errors


taoeffect:
> When running this updated linting plugin against the codebase, two occurences of double curlies were detected, causing the build to fail:

Hah! It's already paying off! 😄 

Either I or someone else will find those two instances and fix them and then merge this PR, thank you so much @snowteamer, fantastic work!

taoeffect (approved):
Approved!

PR #1035, opened 2020-12-11
#1026 Use an empty string instead of null as the default mincome

### LINKED ISSUE #1026: Console error in GroupCreationModal during GroupMincome step
### Problem

It is unclear to me whether this is in `master` or whether this is in @pieer's #1004, because the error doesn't occur every time, only sometimes:

Screen Shot 2020-12-02 at 1 08 07 PM

For whatever reason, this validation in `GroupCreationModal.vue:170` is passed in a `null` value sometimes:

```js
        [L('Oops, you entered 0 or a negative number')]: mincomePositive,
```

This leads to an exception thrown by `commaToDots`, which doesn't expect a `null` parameter:

```js
function commaToDots (value: string | number): string {
  // ex: "1,55" -> "1.55"
  return typeof value === 'string' ? value.replace(/,/, '.') : value.toString()
}
```

### Solution

Find out why sometimes `vuelidate` is passing in a null value, and either modify the validation or the function `mincomePositive` to handle `null` values.

snowteamer:
Closes #1026

- Use an empty string instead of `null` as the default value of the mincome input field in the `GroupCreationModal` Vue component.
- Modify the custom `giCreateGroup()` Cypress command to include a regression check

### DIFF
--- a/frontend/views/containers/group-settings/GroupCreationModal.vue
+++ b/frontend/views/containers/group-settings/GroupCreationModal.vue
@@ -135,7 +135,7 @@ export default {
         sharedValues: '',
         // randomize to reduce choice bias
         ruleOrder: Math.round(Math.random()) === 1 ? [RULE_PERCENTAGE, RULE_DISAGREEMENT] : [RULE_DISAGREEMENT, RULE_PERCENTAGE],
-        mincomeAmount: null,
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Thank you @snowteamer!

PR #1036, opened 2020-12-12
Fix null reference error in unloadModal()

### LINKED ISSUE #1019: Bug: TypeError: can't access property "focus", this.lastFocus is null
### Problem

When loading [this URL to bring up the `InvitationLinkModal`](http://localhost:3000/app/dashboard?modal=InvitationLinkModal) (when already signed in and part of a group), and then clicking on the grey background to dismiss the modal, an error is output to the console:

```
uncaught Vue error in v-on handler: TypeError: can't access property "focus", this.lastFocus is null
    unloadModal Modal.vue:125
    emit events.js:25
    sbp sbp.js:25
    unload ModalMixins.js:44
    VueJS 28
    close ModalMixins.js:41
    VueJS 33
```

### Solution

Find and fix the problem!

Note: this issue using this link with port 3000, which as of this writing is not yet the way we access the prototype when developing (but soon will be!).

So if you're testing this out on `master` before #1020 is merged, use this link instead: http://localhost:8000/app/dashboard?modal=InvitationLinkModal

snowteamer:
Closes #1019 

Not sure if an additional Cypress tests would be useful for this one, since I think its the kind of error which would be better prevented by static analysis.

In such a test, we might have to load every modal in turn and test whether it actually disappears without any error when the user tries to dismiss them, either by clicking outside or using the keyboard.

### DIFF
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -121,8 +121,8 @@ export default {
         this.subcontent.pop()
       } else {
         this.content = null
-        // Refocus on button that open the modal
-        this.lastFocus.focus()
+        // Refocus on the button that opened this modal, if any.
+        if (this.lastFocus) this.lastFocus.focus()
       }
       if (this.replacement) {
         this.openModal(this.replacement)


taoeffect (approved):
Great work @snowteamer, nice one-line fix! 🥇

PR #1037, opened 2020-12-12
Updated Core Protocol in preparation of accounts support (Closes #603)

### LINKED ISSUE #603: Create base contract protocol
### Problem

The way our contract chain messages are designed right now needs to be modified because they do not handle protocol upgrades or key management (#92). Of these two issues, handling protocol upgrades is the most urgent.

An example of a protocol upgrade is a new voting rule type being added to the GroupContract that didn't exist before.

**EDIT:** Additionally, we currently have no way to distinguish between publicly-visible and privately-visible data for a contract. For example, it would be useful for someone who wants to join a group to see what the group's profile picture is (if they set its visibility public), or other information about the group. This should probably be split out into a separate issue.

### Solution

```
    //       define a base set of protocol messages that are publicly readable
    //       OP_CONTRACT - create a contract with a given name (publicly readable),
    //                     and authorized keys, but all other data is encrypted.
    //       OP_ACTION - an action is applied to the contract. its name and its
    //                   data is encrypted.
    //       OP_KEY_* - key related ops that determine who can write to this contract
    //       OP_PROTOCOL_UPGRADE - bump the protocol version, clients that are less
    //                             than this version cannot read or write
```

At runtime, clients in a group will support the lowest protocol version supported by any group member only. This will be determined by looking at the lowest protocol version of all member IdentityContracts, and using that as the supported protocol version for the GroupContract. As clients upgrade, new features can become available. This ensures that all members are able to process all messages. New members cannot join a group unless their client supports its protocol version (which can be updated by anyone in the group - but see important note in #602).

Useful links:

- [dynamic `import()`](https://rollupjs.org/guide/en/#importing)
- https://vuex.vuejs.org/api/#hotupdate
- https://vuex.vuejs.org/guide/modules.html#preserving-state

**EDIT:** regarding the publicly-visible data issue in the above edit, we could use `OP_PROP_SET` to selectively reveal data by setting public properties on the contract, and reading those.

**EDIT:** per [@sandrina-p's comment](https://github.com/okTurtles/group-income-simple/pull/611#discussion_r313870216), modify `GIMessage.create` to take an object as parameters instead of a long list of unnamed parameters.

taoeffect:
### Upgraded Core Protocol (Closes #603)

This PR includes a major change to the design of the core Group Income protocol in preparation for full accounts support and end-to-end encryption.

The core contract class `GIMessage` has been modified to add the concept of operations. The following operations are supported:

- `OP_CONTRACT` — Creates a new contract chain
- `OP_ACTION` — Performs an end-to-end encrypted action on a contract (e2e cryptography not yet implemented)
- `OP_PUB_ACTION` — Performs a publicly-readable action on a contract
- `OP_KEY_AUTH` — Authorizes a new key to be able to write to a contract
- `OP_KEY_DEAUTH` — Invalidates a key (not implemented)
- `OP_PROTOCOL_UPGRADE` — Advertises a new version of the protocol (not implemented)
- `OP_PROP_SET` — Sets a publicly readable key-value pair on the contract
- `OP_PROP_DEL` — Removes a publicly readable key-value pair from the contract

One of the significant changes that has resulted from this is that the very first message to a contract is no longer the contract's constructor action. Instead, the constructor action is posted second in the chain. The first message merely advertises the existence of a new contract via `OP_CONTRACT`, and the constructor comes next via `OP_ACTION` or `OP_PUB_ACTION`.

### Miscellaneous Changes

- We're now using native browser modules instead of SystemJS. `systemjs` dependency removed.
- Removed `babel-plugin-transform-inline-environment-variables` and replaced it with `@rollup/plugin-replace`
- Re-enabled the babel cache
- Build process now has two entry points and two outputs: one for the main bundle, and another for a service worker (which is currently unused in the app). As part of this, our build process now correctly supports multiple inputs/outputs via rollup.
- Ran `npm audit fix` to update some deep dependencies
- Removed `fromPairs` in `giLodash` (use [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)) and added `mapValues`

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -9,7 +9,7 @@ import { makeResponse } from '~/shared/functions.js'
 import { RESPONSE_TYPE } from '~/shared/constants.js'
 import { SERVER_RUNNING } from './events.js'
 import { SERVER_INSTANCE, PUBSUB_INSTANCE } from './instance-keys.js'
-import { bold } from 'chalk'
+import chalk from 'chalk'
 
 const Inert = require('@hapi/inert')
 
@@ -44,7 +44,7 @@ sbp('sbp/selectors/register', {
     const contractID = entry.isFirstMessage() ? entry.hash() : entry.message().contractID
     await sbp('gi.db/log/addEntry', entry)
     const response = makeResponse(RESPONSE_TYPE.ENTRY, entry.serialize())
-    console.log(bold.blue(`broadcasting to room ${contractID}:`), entry.hash(), entry.type())
+    console.log(chalk.blue.bold(`broadcasting to room ${contractID}:`), entry.hash(), entry.type())
     sbp('okTurtles.data/apply', PUBSUB_INSTANCE, p => {
       p.room(contractID).write(response)
     })
@@ -54,10 +54,11 @@ sbp('sbp/selectors/register', {
   }
 })
 
-// NOTE: uncomment this or enable debug logging support to show all requests
-// hapi.events.on('response', (request, event, tags) => {
-//   console.debug(chalk`{grey ${request.info.remoteAddress}: ${request.method.toUpperCase()} ${request.path} --> ${request.response.statusCode}}`)
-// })
+if (process.env.NODE_ENV === 'development' && !process.env.CI) {
+  hapi.events.on('response', (request, event, tags) => {
+    console.debug(chalk`{grey ${request.info.remoteAddress}: ${request.method.toUpperCase()} ${request.path} --> ${request.response.statusCode}}`)
+  })
+}
 
 ;(async function () {
   // https://hapi.dev/tutorials/plugins
--- a/backend/server.js
+++ b/backend/server.js
@@ -28,11 +28,13 @@ const hapi = new Hapi.Server({
   // See: https://github.com/hapijs/discuss/issues/262#issuecomment-204616831
   routes: {
     cors: {
-      origin: [
-        process.env.API_URL,
-        // improve support for browsersync proxy
-        process.env.NODE_ENV === 'development' && 'http://localhost:3000'
-      ]
+      // TODO: figure out if we can live with '*' or if we need to restrict it
+      origin: ['*']
+      // origin: [
+      //   process.env.API_URL,
+      //   // improve support for browsersync proxy
+      //   process.env.NODE_ENV === 'development' && 'http://localhost:3000'
+      // ]
     }
   }
 })
--- a/backend/server.js
+++ b/backend/server.js
@@ -33,7 +33,7 @@ const hapi = new Hapi.Server({
       // origin: [
       //   process.env.API_URL,
       //   // improve support for browsersync proxy
-      //   process.env.NODE_ENV === 'development' && 'http://localhost:3000'
+      //   ...(process.env.NODE_ENV === 'development' && ['http://localhost:3000'])
       // ]
     }
   }
--- a/backend/server.js
+++ b/backend/server.js
@@ -69,6 +69,6 @@ if (process.env.NODE_ENV === 'development' && !process.env.CI) {
   require('./routes.js')
   require('./pubsub.js')
   await hapi.start()
-  console.log('API server running at:', hapi.info.uri)
+  console.log('Backend server running at:', hapi.info.uri)
   sbp('okTurtles.events/emit', SERVER_RUNNING, hapi)
 })()
--- a/backend/server.js
+++ b/backend/server.js
@@ -43,10 +43,10 @@ sbp('okTurtles.data/set', SERVER_INSTANCE, hapi)
 
 sbp('sbp/selectors/register', {
   'backend/server/handleEntry': async function (entry: GIMessage) {
-    const contractID = entry.isFirstMessage() ? entry.hash() : entry.message().contractID
+    const contractID = entry.contractID()
     await sbp('gi.db/log/addEntry', entry)
     const response = makeResponse(RESPONSE_TYPE.ENTRY, entry.serialize())
-    console.log(chalk.blue.bold(`broadcasting to room ${contractID}:`), entry.hash(), entry.type())
+    console.log(chalk.blue.bold(`broadcasting ${entry.description()}`))
     sbp('okTurtles.data/apply', PUBSUB_INSTANCE, p => {
       p.room(contractID).write(response)
     })
--- a/frontend/controller/backend.js
+++ b/frontend/controller/backend.js
@@ -114,7 +114,7 @@ sbp('sbp/selectors/register', {
       }
       if (r.status === 409) {
         if (attempt + 1 > maxAttempts) {
-          console.error(`publishLogEntry: failed to publish ${entry.description()} after ${attempt} of ${maxAttempts} attempts`, entry)
+          console.error(`publishLogEntry: failed to publish ${entry.description()} after ${attempt} attempts`, entry)
           throw new Error(`publishLogEntry: ${r.status} - ${r.statusText}. attempt ${attempt}`)
         }
         // create new entry
--- a/frontend/controller/backend.js
+++ b/frontend/controller/backend.js
@@ -8,7 +8,6 @@ import { CONTRACTS_MODIFIED } from '~/frontend/utils/events.js'
 import { intersection, difference, delay, randomIntFromRange } from '~/frontend/utils/giLodash.js'
 import pubsub from './utils/pubsub.js'
 import { handleFetchResult } from './utils/misc.js'
-import { ACTION_REGEX } from '~/frontend/model/contracts/Contract.js'
 
 // temporary identity for signing
 // const nacl = require('tweetnacl')
@@ -97,7 +96,7 @@ sbp('okTurtles.events/on', CONTRACTS_MODIFIED, async (contracts) => {
 
 sbp('sbp/selectors/register', {
   'backend/publishLogEntry': async (entry: GIMessage, { maxAttempts = 2 } = {}) => {
-    const action = ACTION_REGEX.exec(entry.type())[1]
+    const contractID = entry.contractID()
     let attempt = 1
     // auto resend after short random delay
     // https://github.com/okTurtles/group-income-simple/issues/608
@@ -115,18 +114,21 @@ sbp('sbp/selectors/register', {
       }
       if (r.status === 409) {
         if (attempt + 1 > maxAttempts) {
-          console.error(`publishLogEntry: failed to publish ${action} after ${attempt} of ${maxAttempts} attempts`, entry)
+          console.error(`publishLogEntry: failed to publish ${entry.description()} after ${attempt} of ${maxAttempts} attempts`, entry)
           throw new Error(`publishLogEntry: ${r.status} - ${r.statusText}. attempt ${attempt}`)
         }
         // create new entry
         const randDelay = randomIntFromRange(0, 1500)
-        console.warn(`publishLogEntry: attempt ${attempt} of ${maxAttempts} failed. Waiting ${randDelay} msec before resending ${action}`)
+        console.warn(`publishLogEntry: attempt ${attempt} of ${maxAttempts} failed. Waiting ${randDelay} msec before resending ${entry.description()}`)
         attempt += 1
         await delay(randDelay) // wait half a second before sending it again
-        // re-create it to get correct latestHash
-        entry = await sbp(`${action}create`, entry.data(), entry.contractID())
+        // if this isn't OP_CONTRACT, get latestHash, recreate and resend message
+        if (!entry.isFirstMessage()) {
+          const previousHEAD = await sbp('backend/latestHash', contractID)
+          entry = GIMessage.create(contractID, previousHEAD, entry.op())
+        }
       } else {
-        console.error(`publishLogEntry: ${r.status} - ${r.statusText}, failed to publish ${action}:`, entry)
+        console.error(`publishLogEntry: ${r.status} - ${r.statusText}, failed to publish ${entry.description()}:`, entry)
         throw new Error(`publishLogEntry: ${r.status} - ${r.statusText}`)
       }
     }
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -32,10 +32,13 @@
       <modal class="l-modal"></modal>
       <cypress-bypass-ui v-if="isInCypress"></cypress-bypass-ui>
     </div>
-    <script src="/assets/js/s.min.js"></script>
+    <script type="text/javascript">
+      import('/assets/js/main.js')
+    </script>
+    <!-- <script src="/assets/js/s.min.js"></script>
     <script type="text/javascript">
       System.import('/assets/js/main.js')
-    </script>
+    </script> -->
     <!-- https://github.com/rollup/rollup-starter-code-splitting -->
     <!-- TODO:
     <script type="text/javascript">
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -32,9 +32,7 @@
       <modal class="l-modal"></modal>
       <cypress-bypass-ui v-if="isInCypress"></cypress-bypass-ui>
     </div>
-    <script type="text/javascript">
-      import('/assets/js/main.js')
-    </script>
+    <script type="module" src="/assets/js/main.js"></script>
     <!-- <script src="/assets/js/s.min.js"></script>
     <script type="text/javascript">
       System.import('/assets/js/main.js')
--- a/frontend/model/contracts/Contract.js
+++ b/frontend/model/contracts/Contract.js
@@ -69,7 +69,7 @@ export function DefineContract (contract: Object) {
         metadata.validate(meta, { state, ...gProxy, contractID })
         contract.actions[action].validate(data, { state, ...gProxy, meta, contractID })
         return GIMessage.createV1_0(contractID, previousHEAD, [
-          GIMessage.OP_ACTION,
+          GIMessage.OP_ENCRYPTED_ACTION,
           // TODO: encryption happens here
           JSON.stringify({ type: actionSelector, data, meta })
         ])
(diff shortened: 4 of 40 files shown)


backend/server.js
```diff
@@ -28,11 +28,13 @@ const hapi = new Hapi.Server({
   // See: https://github.com/hapijs/discuss/issues/262#issuecomment-204616831
   routes: {
     cors: {
-      origin: [
-        process.env.API_URL,
-        // improve support for browsersync proxy
-        process.env.NODE_ENV === 'development' && 'http://localhost:3000'
-      ]
+      // TODO: figure out if we can live with '*' or if we need to restrict it
+      origin: ['*']
```
snowteamer, line 32: I'd rather start with a minimal set of origins, then whitelist additional origins later as necessary. However that method would require us to keep in mind that not all origins are accepted.

backend/server.js
```diff
@@ -28,11 +28,13 @@ const hapi = new Hapi.Server({
...
+      // origin: [
+      //   process.env.API_URL,
+      //   // improve support for browsersync proxy
+      //   process.env.NODE_ENV === 'development' && 'http://localhost:3000'
```
snowteamer: This expression could evaluate to `false` if it was no longer commented out, likely causing an error.

This can be avoided by using a ternary expression like:
```js
origin: (
  process.env.NODE_ENV === 'development' ?
  [process.env.API_URL, 'http://localhost:3000'] :
  [process.env.API_URL]
)
```
or the drier, but less readable:
```js
origin: [
  process.env.API_URL,
  ...(process.env.NODE_ENV === 'development' ? ['http://localhost:3000'] : []),
]
```

frontend/controller/backend.js
```diff
@@ -115,18 +114,21 @@ sbp('sbp/selectors/register', {
       }
       if (r.status === 409) {
         if (attempt + 1 > maxAttempts) {
-          console.error(`publishLogEntry: failed to publish ${action} after ${attempt} of ${maxAttempts} attempts`, entry)
+          console.error(`publishLogEntry: failed to publish ${entry.description()} after ${attempt} of ${maxAttempts} attempts`, entry)
```
snowteamer: Isn't `attempt` always equal to `maxAttempts` here, and therefore redundant?

frontend/model/contracts/group.js
```diff
@@ -227,13 +227,13 @@ DefineContract({
...
     groupMonthlyPayments (state, getters) {
-      return getters.currentGroupState.paymentsByMonth
+      return getters.currentGroupState.paymentsByMonth || {}
     },
     thisMonthsPaymentInfo (state, getters) {
       return getters.groupMonthlyPayments[currentMonthstamp()] || {}
     },
     withGroupCurrency (state, getters) {
-      return currencies[getters.groupSettings.mincomeCurrency].displayWithCurrency
+      return getters.groupSettings.mincomeCurrency && currencies[getters.groupSettings.mincomeCurrency].displayWithCurrency
```
snowteamer, line 240: If this group has no defined mincome currency, not even a default one like the USD, then calling this function is probably an error which should be reported.

frontend/model/contracts/group.js
```diff
@@ -507,7 +507,7 @@ DefineContract({
           }
 
           const groupIdToSwitch = Object.keys(contracts)
-            .find(cID => contracts[cID].type === 'group' &&
+            .find(cID => contracts[cID].type === 'gi.contracts/group' &&
```
snowteamer, line 514: Nit: I would prefer the name `contractsByID` here, since it does not refer to an array but a plain object.

frontend/model/state.js
```diff
@@ -532,7 +522,9 @@ const getters = {
     return { sent, todo, late, received, toBeReceived }
   },
   ourPaymentsSummary (state, getters) {
-    const { todo, sent, toBeReceived, received } = getters.ourPayments
+    const ourPayments = getters.ourPayments
+    if (!ourPayments) return
+    const { todo, sent, toBeReceived, received } = ourPayments
     const isCompleted = (p) => p.data.status === PAYMENT_COMPLETED
     const isPartialCount = (list) => list.filter(p => p.partial).length
```
snowteamer: This function returns a number but its name looks very much like a boolean function name.

I would rather suggest `countPartial()`, or `countPartialPayments()`, or `getNumberOfPartialPayments()`, since it finds the number of partial payments in a list.

frontend/model/state.js
```diff
@@ -364,12 +353,13 @@ const getters = {
   },
   ourPayments (state, getters) {
     // Payments relative to the current month only
+    const monthlyPayments = getters.groupMonthlyPayments
+    if (!monthlyPayments || Object.keys(monthlyPayments).length === 0) return
```
snowteamer, line 357: I'm not sure about returning undefined here, it's probably unsafe. It also makes the function"s return type permanently more complex to handle an edge case which should be exceptional. So I would favor throwing instead.

frontend/model/state.js
```diff
@@ -532,7 +522,9 @@ const getters = {
     return { sent, todo, late, received, toBeReceived }
   },
   ourPaymentsSummary (state, getters) {
-    const { todo, sent, toBeReceived, received } = getters.ourPayments
+    const ourPayments = getters.ourPayments
+    if (!ourPayments) return
```
snowteamer, line 526: Same point here

frontend/model/state.js
```diff
@@ -579,14 +571,16 @@ const getters = {
...
     // due to the same flow issue as https://github.com/facebook/flow/issues/5838
     return Object.keys(contracts || {})
-      .filter(contractID => contracts[contractID].type === 'group' && state[contractID].settings)
+      .filter(contractID => contracts[contractID].type === 'gi.contracts/group' && state[contractID].settings)
       .map(contractID => ({ groupName: state[contractID].settings.groupName, contractID }))
   },
   groupMembersSorted (state, getters) {
-    const weJoinedMs = new Date(getters.currentGroupState.profiles[getters.ourUsername].joinedDate).getTime()
+    const profiles = getters.currentGroupState.profiles
+    if (!profiles) return
```
snowteamer, line 579: Another suspicious return statement. If there are no members in the group, it's ambiguous whether their profiles are sorted, so it would be safer to avoid the question altogether if possible.

shared/GIMessage.js
```diff
@@ -1,44 +1,65 @@
...
-  static create (
+  static OP_CONTRACT = 'c'
+  static OP_ACTION = 'a' // e2e-encrypted action
+  static OP_PUB_ACTION = 'pa' // publicly readable action
+  static OP_KEY_AUTH = 'ka' // add this key to the list of keys allowed to write to this contract
+  static OP_KEY_UPDATE = 'ku' // update a key's context
+  static OP_KEY_DEAUTH = 'kd' // remove this key from authorized keys
+  static OP_PROTOCOL_UPGRADE = 'pu'
+  static OP_PROP_SET = 'ps' // set a public key/value pair
+  static OP_PROP_DEL = 'pd' // delete a public key/value pair
```
snowteamer, line 40: - Abbreviations in in public APIs should generally be avoided.
  For example, the PUB part in `OP_PUB_ACTION` is very easy to read as PUBLISH instead of PUBLIC.
- When sticking to abbreviations, they should arguably be used consistently.
  In this case, PROTOCOL could be abbreviated as PROTO.
- Since `OP_ACTION` means an encrypted action according to the comments,
  Shouldn't it be written `OP_ENCRYPTED_ACTION` to be more explicit,
  or PRIVATE_ACTION in contrast with `OP_PUB_ACTION`?
- Some operation identifiers are nouns (e.g. action, contract) while other are verbs (prop_set). Some can also be read as both (auth, update).
  Would it be possible to use only names or only verbs to improve consistency?

shared/GIMessage.js
```diff
@@ -74,7 +114,78 @@ export class GIMessage {
...
+  'chelonia/message/process': function (message: GIMessage, state: Object) {
+    sanityCheck(message)
+    const [opT, opV] = message.op()
+    const hash = message.hash()
+    const contractID = message.contractID()
+    const config = sbp('okTurtles.data/get', 'CHELONIA_CONFIG')
+    if (!state._vm) state._vm = {}
+    const opFns = {
+      [GIMessage.OP_CONTRACT] (v) {
+        if (!state._vm.authorizedKeys) state._vm.authorizedKeys = []
```
pieer, line 130: Could initialize directly without push here

shared/GIMessage.js
```diff
@@ -74,7 +114,78 @@ export class GIMessage {
...
+          }
+          sbp(type, { data, meta, hash, contractID }, state)
+        }
+      },
+      [GIMessage.OP_PROP_SET] (v) {
+        if (!state._vm.props) state._vm.props = {}
+        state._vm.props[v.key] = v.value
+      },
+      [GIMessage.OP_KEY_AUTH] (v) {
+        if (!state._vm.authorizedKeys) state._vm.authorizedKeys = []
```
pieer, line 153: Same

shared/GIMessage.js
```diff
@@ -74,7 +114,78 @@ export class GIMessage {
...
+}
+
+function sanityCheck (msg: GIMessage) {
+  const [type] = msg.message().op
+  switch (type) {
+    case GIMessage.OP_CONTRACT:
+      if (!msg.isFirstMessage()) throw new Error('OP_CONTRACT: must be first message')
+      break
+    case GIMessage.OP_ACTION:
+      // nothing forn ow
```
pieer: forn ow?

pieer (dismissed):
Great work @taoeffect. I could only find very tiny detail to maybe change.
Approved!

frontend/controller/backend.js
snowteamer: Isn't `attempt` always equal to `maxAttempts` here? If so, no need to print it twice in the error message.

frontend/index.html
```diff
@@ -32,10 +32,13 @@ <h1>Outdated browser</h1>
       <modal class="l-modal"></modal>
       <cypress-bypass-ui v-if="isInCypress"></cypress-bypass-ui>
     </div>
-    <script src="/assets/js/s.min.js"></script>
     <script type="text/javascript">
-      System.import('/assets/js/main.js')
+      import('/assets/js/main.js')
```
snowteamer: Is using a dynamic import statement better than using `type=module` together with a `src` attribute?

frontend/model/contracts/Contract.js
```diff
@@ -6,6 +6,9 @@ import { GIMessage } from '~/shared/GIMessage.js'
 // this must not be exported, but instead accessed through 'actionWhitelisted'
 const whitelistedSelectors = {}
 const sideEffectStacks = {} // [contractID]: Array<*>
+const cheloniaCfg = sbp('okTurtles.data/get', 'CHELONIA_CONFIG')
+
+cheloniaCfg.whitelisted = (sel) => !!whitelistedSelectors[sel]
```
snowteamer, line 11: So "Chelonia" is the name of the GI protocol?

Well that's cute. I like it! 😄

frontend/model/contracts/group.js
snowteamer, line 240: This will return a falsy value if the group has no mincome currency.

And in that case, calling this function doesn't make sense anyway, therefore throwing an error seems appropriate.

frontend/model/contracts/group.js
snowteamer, line 514: I would find it easier to remember that `contracts` is not an array but actually a plain object whose keys are contract IDs, if it was named `contractsByID` instead.

frontend/model/state.js
snowteamer, line 357: I would favor either returning an empty object or throwing an error here.

frontend/model/state.js
snowteamer, line 526: Same point here also.

frontend/model/state.js
snowteamer, line 579: If there are currently no members in the group, it's ambiguous whether the member list is sorted.

So it could be safer to avoid the question altogether, by avoiding to call the function in that case in the first place, and by having it throw an exception if it does get called.

snowteamer (dismissed):
Congratulations for the hard work on this one!

I have still a few files to review, namely the tests. So I will likely add a few more comments later, but am already approving so as not to block the merge process.

shared/GIMessage.js
taoeffect: Do you mean something like this?

```js
if (!state._vm.authorizedKeys) {
  state._vm.authorizedKeys = [{ key: v.authkey, context: 'owner' }]
  return
}
```

That is a few extra lines and it would violate DRY since `{ key: v.authkey, context: 'owner' }` is repeated, so I chose to do it this way

frontend/model/state.js
taoeffect: So, these getters are attached to UI elements... and with the modifications I made in this PR they will throw an exception and that breaks the rendering of the Vue components that bind to the getters. This return statement prevents the exception from occurring and allows the UI to render.

Cleaner versions of this that accomplish the same thing are certainly welcome!

shared/GIMessage.js
taoeffect: Oops. Meant "nothing for now", nice catch, will be fixed in the next commit, thanks!

frontend/model/state.js
taoeffect: The UI was breaking because an exception was thrown, so I returned `undefined` here to fix it. The issue with returning an empty object is that it's not clear to me personally whether or not `{}` is a falsey value. Even if it is, that requires some deep JS knowledge to know. And other getters might rely on the return value of this to perform additional computations. This way they can also check if this getter returned a false value and return immediately without throwing an exception that breaks UI rendering.

frontend/model/contracts/group.js
taoeffect: Hmm, this would be a complex refactor because the project has so many instances of "contracts", so if you'd like this change to be made, can you open an issue for it?

frontend/model/contracts/group.js
taoeffect: Perhaps in this case that does make sense for this getter... I'll add a TODO here for now.

frontend/model/contracts/Contract.js
taoeffect: Glad you like it! It's a working name, and might change... will probably bring it up on Slack to hear people's opinions about this name and another I'm considering.

frontend/index.html
taoeffect: Interesting, I've no clue which is better, but your suggestion looks better so I'll use that in a followup commit!

backend/server.js
taoeffect: Well, eventually we want servers to talk to each other ... so... 🤷‍♂️

shared/GIMessage.js
taoeffect: Think this should be resolved now? I left the PROPs as the same because I couldn't think of anything better to replace them with.

CrazedCoding (approved):
It all looks well written, reasonably efficient, and logically air-tight, ALTHOUGH I think I need to run some tests: I am very interested in seeing how, if at all, updates to the core protocol are going to affect the mocha tests I am making a PR for.

backend/server.js
snowteamer: In that case, I don't know if it's even possible to not use `'*'`

frontend/model/state.js
snowteamer: Alright, let's leave it as is for now. There might be something to improve in the calling UI code and getters, but it can be fixed in a separate issue.

frontend/model/contracts/group.js
snowteamer: If you think it's a good idea, then I'd be glad to (when I have time)

frontend/model/state.js
snowteamer: Thank you for the clarifications!

shared/GIMessage.js
snowteamer: I would tend to misread PROP as property, or proposal, while it actually means a public key/value pair according to the inline comments.
So maybe `OP_KEYPAIR_SET` and `OP_KEYPAIR_DELETE`? Though, these names still leave the "public" part out - but iI might be descriptive enough if there are no operations involving private keypairs.

shared/GIMessage.js
```diff
@@ -59,13 +81,31 @@ export class GIMessage {
     return instance
   }
 
+  decrypted (fn: ?Function): any {
```
snowteamer, line 84: The name `decrypted` sounds like a boolean or boolean function to me, but it returns a decrypted op value (correct me if I am wrong); so what about `decryptedValue` ?

snowteamer (approved):

test/backend.test.js
```diff
@@ -90,23 +86,25 @@ describe('Full walkthrough', function () {
...
+  async function createIdentity (username, email) {
     // append random id to username to prevent conflict across runs
     // when GI_PERSIST environment variable is defined
     username = `${username}-${Math.floor(Math.random() * 1000)}`
-    return sbp('gi.contracts/identity/create', {
+    const msg = await sbp('gi.contracts/identity/create', {
       // authorizations: [Events.CanModifyAuths.dummyAuth(name)],
       attributes: { username, email }
     })
+    msg.decrypted(JSON.parse)
```
snowteamer, line 106: This statement is confusing, why is the return value not used? Is it just to check that the call doesn't throw?

shared/GIMessage.js
taoeffect: OK, can make that change in a followup PR so as to not force another round of review!

test/backend.test.js
taoeffect: In order to perform the decryption on the message and store the value in `this._decrypted` for followup invocations via `msg.decrypted()` (or in your case, `msg.decryptedValue()`).

PR #1040, opened 2020-12-21
#1017 Improve income details validation

### LINKED ISSUE #1017: Zero is not negative / People should be able to pledge $0
### Problem

According to our leading zoologist of thought (mathematicians), zero is not negative:

image

### Solution

fixit

snowteamer:
### Summary of changes

- Input validation now checks if the specified income or contribution is a number *before* checking if it's positive.
  This should improve error messages.

- Users can now specify an income amount of zero in their income details.

- Users are now allowed to pledge 0, i.e. to not pledge any monetary amount.
  They can still offer valuable non-monetary contributions of course.

Closes #1017


cypress.json
```diff
@@ -2,14 +2,14 @@
...
-  "defaultCommandTimeout": 5000, 
+  "defaultCommandTimeout": 15000, 
   "fixturesFolder": "test/cypress/fixtures",
   "integrationFolder": "test/cypress/integration",
   "pluginsFile": "test/cypress/plugins",
   "screenshotsFolder": "test/cypress/screenshots",
   "supportFile": "test/cypress/support",
   "videosFolder": "test/cypress/videos",
-  "video": true,
+  "video": false,
```
taoeffect: Can you revert these two changes? I see no reason for extending the timeout (unless you're aware of one?), and we definitely want the video output for the Cypress dashboard. It helps with debugging.

strings/french.json
```diff
@@ -255,7 +255,7 @@
   "Income Details": "Détails de Revenu",
   "No, I don't": "Non",
   "Oops, you entered a negative number": "Oups, vous avez entré un nombre négatif",
-  "The amount cannot have more than 2 decimals": "Le montant ne peut pas avoir plus de deux décimales",
+  "The amount must be a number (e.g. 100.75)": "Le montant doit être un nombre (par exemple 100,75)",
```
taoeffect: Leave this unmodified, the `.json` is created from the `.strings` file, which is where the translation should happen, and the `.strings` in turn are created from the UI (`i18n` tag) or JS (`L` function). Then run the `strings` command per the [`strings` readme](https://github.com/okTurtles/strings/blob/master/README.md#help-translate-a-language).

taoeffect (changes requested):
Haven't had a chance to read the issue and this PR, but here's a pre-review with some easy change requests! Thanks for sending this in!

cypress.json
snowteamer: Oops, I forgot to revert these custom changes before committing. The extended timeout was apparently necessary on my end to pass some tests, and while video is a powerful debugging tool, it takes a very long time for me so I only enable it as a last resort

strings/french.json
snowteamer: Ok. My intent was to emulate the effects of running the `strings` tool so as to avoid the need for someone else to run it, since I'm unfortunately still on an unsupported platform.

taoeffect: Oh, this was actually OK. My bad, I shouldn't have written that comment, because you're right, you can actually modify these files directly like this. The only reason why you might not want to do it manually is that there might be some format change or some other automated thing that the strings command will do differently. Anyway, you are free to un-revert this change, or we'll run the strings command ourselves later. Apologies, today has been a long day and my brain clearly wasn't thinking straight when it came to this.

taoeffect (approved):
Approved! Great job @snowteamer! Will merge later today, either as-is or with your revert (if you decided to send one in) regarding that `.json` strings file.

strings/french.json
snowteamer: No problem, I needed some `git restore` practice anyway 😄

PR #1041, opened 2020-12-22
Proper Center for Logo & Badges

ktamas77:
Markdown doesn't supports center tag, but simplified HTML does.
Converted all badges to HTML so they can be precisely aligned to the center among with the logo.


taoeffect (approved):
Nice! Thank you @ktamas77!

PR #1042, opened 2020-12-26
#826 Update Cypress

### LINKED ISSUE #826: Upgrade us to Cypress 4.0.1 or greater
### Problem

We currently cannot use Cypress with Firefox Developer Edition because we're running an outdated version of Cypress.

### Solution

Update our Cypress version and update our npm scripts and anything else (e.g. Gruntfile.babel.js, if needed) based the relevant instructions at:

- https://cypress.io/blog/2020/02/06/introducing-firefox-and-edge-support-in-cypress-4-0/
- https://docs.cypress.io/guides/guides/cross-browser-testing.html
- **Migration guide:** https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-4-0

snowteamer:
### Summary of changes

- Updated Cypress to 6.2.0
- Had to change one line in `group-contributions.spec.js` to use `not.exist` instead of `not.be.visible` (see [nonexistent element assertions](https://docs.cypress.io/guides/references/migration-guide.html#Non-existent-element-assertions) in the Cypress migration guide).
- Temporarily downgrade Node from v15 to v14 to avoid a dependency compatibility issue that was found by running the tests. Some of our dependencies do not support Node v15 yet.

### Additional info

The overall Cypress run for this build took 3min 40s, which is 23s better than the run from #1041, and 14s better than the run from #1040.

Therefore, the performance issues mentionned in https://github.com/okTurtles/group-income-simple/pull/829 seem to have been resolved, or to no longer be relevant.

As for the other issues, I'm not sure. There are a lot of open issues on the Cypress repo, but I didn't go as far as investigating enough of them to guess if the situation was better in version 3.5.

Maybe @sandrina-p could tell whether we can safely upgrade?

Closes #1042

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "15"
+  - "14"
 dist: trusty
 compiler: clang # per: https://docs.travis-ci.com/user/languages/c
 env:
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This semi-centralized "Simple Edition" of Group Income is our way to rapidly pro
 
 ## Getting Started
 
-1. Install [Node.js](https://nodejs.org) (version 12 or greater).
+1. Install [Node.js](https://nodejs.org) (version 14 LTS).
 2. Install [Grunt](https://github.com/gruntjs/grunt): `npm install -g grunt-cli`
 3. Clone this repo (or a fork of it if you plan on [contributing](#contributing)) and `cd` into it.
 4. Install dependencies: `npm install`
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "chalk": "4.1.0",
         "cheerio": "1.0.0-rc.3",
         "chokidar": "3.4.3",
-        "cypress": "3.5.0",
+        "cypress": "6.2.0",
         "cypress-file-upload": "3.5.0",
         "eslint": "7.14.0",
         "eslint-config-standard": "16.0.2",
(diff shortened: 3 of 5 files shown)


taoeffect (changes requested):
Awesome PR @snowteamer!

One request: could you please update your version of node and npm? The reason I ask is because using the latest version will produce a different `package-lock.json` file. We'd like to use the latest version of it so as to avoid conflicts in PRs going forward.

E.g., when I test locally, I get a large diff after running `npm install` that has a different `lockfileVersion`:

```diff
-> % git diff   

diff --git a/package-lock.json b/package-lock.json
index 2bdb51d7..48d74d2c 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,20071 @@
 {
   "name": "group-income-simple",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```

When I run `npm version`, here are the values I get:

```
-> % npm version
{
  'group-income-simple': '0.0.1',
  npm: '7.0.15',
  node: '15.4.0',

 [ ... snip irrelevant values ... ]
```

You can update node to 15 using whatever package manager you're using, and if it doesn't update npm to at least 7.0.15 after doing that, then you can do it yourself by running `npm i -g npm@latest`.

Finally, feel free to bump the minimum node version requirements in the `README.md` from `12` to `15`.

snowteamer:
That's indeed a good idea, now that our Travis config no longer mandates using Node v14.

However, since this might be useful for other contributors, I will add that `npm i -g npm@latest` will not upgrade npm from v6.x to v7.x, but only to the latest v6.x version.

So I had to run `npm i -g npm@7`, explicitly specifying the major version number.

snowteamer:
Hmm this still causes a merge conflict and a rather large diff - should I temporarily use the exact Node and NPM versions listed in your console output?

taoeffect:
@snowteamer Let's try to troubleshoot this on Slack

taoeffect (approved):
Excellent work @snowteamer! Thank you for sending this in! 😄

Issue #1043, opened 2021-01-15
Switch from Rollup to Snowpack or esbuild to speed up build & other features [$1k bounty]


taoeffect:
### Problem

Apparently the build is still gets progressively slower on some people's machines (@pieer was running into this issue, see #936), but more importantly, the build is still slow on first-run. There's a new builder written in Go that's [100x faster than rollup, called esbuild](https://esbuild.github.io/).

Esbuild in turn is used by [snowpack](https://www.snowpack.dev/), which actually can use other builders as plugins. Like Rollup, Snowpack has its [own plugins](https://www.snowpack.dev/plugins?q=), and you can [create your own](https://www.snowpack.dev/reference/plugins). (Note: esbuild also has [its own plugins](https://github.com/esbuild/community-plugins), though fewer than Snowpack.)

Another neat feature of Snowpack [mentioned in this long blogpost](https://www.snowpack.dev/posts/2021-01-13-snowpack-3-0) (among other features), is that it avoids the need for us to use NPM (#360). Instead of trusting NPM, we trust these people instead, but with the upside that they run `npm install` for us and therefore any scripts that run during that phase are run on their machines instead of ours. So this is a genuine security improvement, and a kind of stepping stone on the way to Deno (#586, #933).

### Solution

Switch our build system to Snowpack.

### Bounty

$1k bounty for a clean solution to this (paid in cryptocurrency).

taoeffect:
Adding a $1k bounty for closing this issue for non-contractors (meaning, if you don't have a development contract with okTurtles, and you'd like to close this issue, you can work on it, and if we accept your PR, you'll receive the bounty).

Anyone else can add to the bounty too if they'd like by following these simple steps:

1. Donate the amount you'd like to dedicate to this issue [to the okTurtles Foundation](https://okturtles.org/donate/).
2. Make a comment here (or privately to us via Gitter or Slack) the same day letting us know you'd like your donation to be used for a bounty to close an issue, and which issue that is.

PR #1044, opened 2021-01-17
#528 Clean up Flow errors 

### LINKED ISSUE #528: Clean up flow errors

snowteamer:
Closes #528

### Summary of changes

In the Flow declaration file:
- Declare the same dependencies as our package.json (although I had to add `chalk`).
- No longer declare dependencies and global variables which are only meant to be accessed from testing or build environments.
 
In the Flow config file:
- Ignore the Grunt file since it belongs to the build environment.
- Temporarily ignore the WIP service worker related stuff.
- Enable the 'emoji' option for nicer output in IDEs. 😄

- Edit many .js files to fill in missing type annotations, mainly on module exports. 

Other miscellanneous changes:
- Fix a bug in the `gi.contracts/group/proposal` SBP function where the `reason` field would not be considered when checking whether a given proposal isn't a duplicate.
- Fill in a missing definition for `INVITE_STATUS.REVOKED`.
- Refactor the `template()` function in `frontend/utils/stringTemplate`.

As the goal was to to add only a minimal amount of annotations to make Flow usable again in our testing process, I tried to only annotate only module exports where possible, and made ample use of `any` where basic types would not be enough.

As a result, the Flow code coverage is still very low, which is likely the reason why only a couple of bugs were discovered in the process.


frontend/views/utils/currencies.js
```diff
@@ -64,24 +64,45 @@ function makeCurrency (options) {
...
     formatCurrency: amount => amount + 'Ƀ'
-  })
+  }): {|
+  decimalsMax: number,
+  displayWithCurrency: (n: number) => string,
+  displayWithoutCurrency: (n: number) => string,
+  symbol: string,
+  symbolWithCode: string,
+  validate: (n: string) => boolean,
+|})
```
taoeffect: Hmm, this has been repeated 3 times. To DRY this, could you please convert it to a named type?

frontend/views/containers/user-settings/settings.js
```diff
@@ -52,7 +52,7 @@ export default {
   }, {
     links: [
       {
-        title: L('Log Out'),
+        title: (L('Log Out'): any),
```
taoeffect: These should be `string`s right?

frontend/utils/image.js
```diff
@@ -27,14 +27,17 @@ export const imageUpload = async (imageFile) => {
...
-      const hash = blake32Hash(new Uint8Array(reader.result))
-      console.debug('picture hash:', hash)
-      fd.append('hash', hash)
-      fd.append('data', file)
-      fetch(`${sbp('okTurtles.data/get', 'API_URL')}/file`, {
-        method: 'POST',
-        body: fd
-      }).then(handleFetchResult('text')).then(resolve).catch(reject)
+      const { result } = reader
+      if (result !== null) {
```
taoeffect: If the result is null, should we `reject` the parent promise?

frontend/model/state.js
```diff
@@ -865,7 +867,9 @@ const handleEvent = {
       const contractID = message.contractID()
       const hash = message.hash()
       if (message.isFirstMessage()) {
-        const { type } = message.opValue()
+        // Flow doesn't understand that a first message must be a contract,
+        // so we have to help it a bit in order to acces to 'type' property.
+        const { type } = ((message.opValue(): any): GIOpContract)
```
taoeffect, line 872: Excellent job adding this comment here, and great fix 👍

frontend/model/state.js
```diff
@@ -758,7 +760,7 @@ const actions = {
 
       if (attemptToReprocessMessageID === message.hash()) {
         console.warn(`Successfully re-processed ${attemptToReprocessMessageID}!`)
-        attemptToReprocessMessageID = null
+        attemptToReprocessMessageID = ''
```
taoeffect, line 763: Is this the best way to fix this..? Or could doing something like this on line 39 work?

```js
let attemptToReprocessMessageID: ?string = null
```

frontend/model/contracts/group.js
```diff
@@ -376,7 +384,7 @@ DefineContract({
           expires_date_ms: number // calculate by grabbing proposal expiry from group properties and add to `meta.createdDate`
         })(data)
 
-        const dataToCompare = omit(data.proposalData, 'reason')
+        const dataToCompare = omit(data.proposalData, ['reason'])
```
taoeffect, line 387: Wow, you fixed a bug, nice!

.flowconfig
```diff
@@ -6,26 +6,38 @@
...
+module.ignore_non_literal_requires=true
 
 # https://github.com/entwicklerstube/babel-plugin-root-import#dont-let-flow-be-confused
 module.name_mapper='^~/\(.*\)$' -> '<PROJECT_ROOT>/\1'
+module.name_mapper='^@components/\(.*\)$' -> '<PROJECT_ROOT>/frontend/views/components/\1'
+module.name_mapper='^@model/\(.*\)$' -> '<PROJECT_ROOT>/frontend/model/\1'
+module.name_mapper='^@pages/\(.*\)$' -> '<PROJECT_ROOT>/frontend/views/pages/\1'
+module.name_mapper='^@utils/\(.*\)$' -> '<PROJECT_ROOT>/frontend/utils/\1'
+module.name_mapper='^@views/\(.*\)$' -> '<PROJECT_ROOT>/frontend/views/\1'
+module.name_mapper='^@view-utils/\(.*\)$' -> '<PROJECT_ROOT>/frontend/views/utils/\1'
```
taoeffect, line 43: Excellent

frontend/controller/backend.js
```diff
@@ -125,7 +125,7 @@ sbp('sbp/selectors/register', {
         // if this isn't OP_CONTRACT, get latestHash, recreate and resend message
         if (!entry.isFirstMessage()) {
           const previousHEAD = await sbp('backend/latestHash', contractID)
-          entry = GIMessage.create(contractID, previousHEAD, entry.op())
+          entry = GIMessage.createV1_0(contractID, previousHEAD, entry.op())
```
taoeffect: Will double-check this in my PR that I'm working on... we can merge this even if it's wrong though since it's a rare situation and I plan to fix it shortly.

If I verify the fix before your changes to my review are pushed, I'll try to remember to comment here with the real fix if it diverges from yours. This comment is mostly a reminder to myself.

frontend/controller/router.js
```diff
@@ -170,12 +170,14 @@ const router = new Router({
...
-      meta: { title: L('Error Testing') }
-    }]),
+    ...(process.env.NODE_ENV === 'development'
+      ? [{
+          path: '/error-testing',
+          name: 'ErrorTesting',
+          component: () => import('../views/pages/ErrorTesting.vue'),
+          meta: { title: L('Error Testing') }
+        }]
+      : []),
```
taoeffect, line 180: This is actually what I had before, and changed it to the `&&`  method. Curious, did you change it back because flow preferred this way?

frontend/model/contracts/payments/index.js
```diff
@@ -7,8 +7,8 @@ export const PAYMENT_CANCELLED = 'cancelled'
 export const PAYMENT_ERROR = 'error'
 export const PAYMENT_NOT_RECEIVED = 'not-received'
 export const PAYMENT_COMPLETED = 'completed'
-export const paymentStatusType = unionOf(...[PAYMENT_PENDING, PAYMENT_CANCELLED, PAYMENT_ERROR, PAYMENT_NOT_RECEIVED, PAYMENT_COMPLETED].map(k => literalOf(k)))
+export const paymentStatusType: any = unionOf(...[PAYMENT_PENDING, PAYMENT_CANCELLED, PAYMENT_ERROR, PAYMENT_NOT_RECEIVED, PAYMENT_COMPLETED].map(k => literalOf(k)))
 export const PAYMENT_TYPE_MANUAL = 'manual'
 export const PAYMENT_TYPE_BITCOIN = 'bitcoin'
 export const PAYMENT_TYPE_PAYPAL = 'paypal'
-export const paymentType = unionOf(...[PAYMENT_TYPE_MANUAL, PAYMENT_TYPE_BITCOIN, PAYMENT_TYPE_PAYPAL].map(k => literalOf(k)))
+export const paymentType: any = unionOf(...[PAYMENT_TYPE_MANUAL, PAYMENT_TYPE_BITCOIN, PAYMENT_TYPE_PAYPAL].map(k => literalOf(k)))
```
taoeffect: These are `Function`s I think

.Gruntfile.babel.js
```diff
@@ -125,7 +125,7 @@ module.exports = (grunt) => {
       sbp('backend/pubsub/setup', require('http').createServer(), true)
     }
     if (!grunt.option('skipbuild')) {
-      grunt.task.run(['exec:eslint', 'exec:puglint', 'exec:stylelint', 'copy', rollup])
+      grunt.task.run(['exec:eslint', 'exec:flow', 'exec:puglint', 'exec:stylelint', 'copy', rollup])
```
taoeffect, line 128: The issue with adding it to the build step is that it will slow down development on first run. We can maybe add it here once we close #1043, but for now a better place might be as part of running `grunt test`.

Also, I'm not sure we're ready yet to have it as a requirement. One way to ignore any failures from the command is to add `|| true` after it (although flow might have a command line flag to not fail on error too, which would be better to use if such a flag exists).

.flowconfig
```diff
@@ -6,26 +6,38 @@
...
+.*/frontend/controller/service-worker.js
 .*/frontend/controller/utils/primus.js
+.*/frontend/utils/blockies.js
 .*/frontend/utils/crypto.js
-.*/frontend/utils/vuexQueue.js
 .*/frontend/utils/flowTyper.js
-# .*/test/backend.js
+.*/frontend/utils/vuexQueue.js
+.*/frontend/views/containers/chatroom/.*
+.*/historical/.*
```
taoeffect, line 18: Can you also add:

```
.*/ignored/.*
```

shared/GIMessage.js
```diff
@@ -25,27 +25,27 @@ sbp('okTurtles.data/set', 'CHELONIA_CONFIG', {
...
-  static OP_PROP_DEL = 'pd' // delete a public key/value pair
+  static OP_CONTRACT: 'c' = 'c'
+  static OP_ENCRYPTED_ACTION: 'ea' = 'ea' // e2e-encrypted action
+  static OP_PUBLIC_ACTION: 'pa' = 'pa' // publicly readable action
+  static OP_KEY_AUTH: 'ka' = 'ka' // add this key to the list of keys allowed to write to this contract
+  static OP_KEY_UPDATE: 'ku' = 'ku' // update a key's context
+  static OP_KEY_DEAUTH: 'kd' = 'kd' // remove this key from authorized keys
+  static OP_PROTOCOL_UPGRADE: 'pu' = 'pu'
+  static OP_PROP_SET: 'ps' = 'ps' // set a public key/value pair
+  static OP_PROP_DEL: 'pd' = 'pd' // delete a public key/value pair
```
taoeffect, line 41: For these, since they might be adjusted in a future PR of mine, can you set their type to `GIOpType` instead? That way fewer replacements will need to be made.

taoeffect (changes requested):
Amazing PR @snowteamer! 🤯

Fantastic work! I tested it out and all seems to be working well. Found a few minor issues and made comments with them. Really impressive work! 🥇

frontend/views/utils/currencies.js
snowteamer: Good catch, let's define a `Currency` type here.

frontend/views/containers/user-settings/settings.js
snowteamer: Most certainly. It looks like `flow codemod annotate-exports` (which I ran to automatically annotate some exports) was a bit conservative on this one.

It is also a bit confusing that Flow doesn't appear to pick up the return type information included where `L()` was defined and exported.

frontend/utils/image.js
snowteamer: Yes that would be safer. Should we pass `reader.error` to `reject()` while doing so ?

frontend/model/state.js
snowteamer: Thank you! Now that I think about it, perhaps I could also have added that this "technique" is called [typecasting through any](https://flow.org/en/docs/types/casting/#toc-type-casting-through-any) in Flow parlance, along with a relevant link. If that doesn't make the comment too long of course. 😄

frontend/model/state.js
snowteamer: This would mean that `undefined` is also a valid value, since so-called [Maybe types](https://flow.org/en/docs/types/maybe) in Flow include both `null` and `undefined`.

frontend/utils/image.js
taoeffect: I have no idea, but I see no harm in doing so if the documentation says that is a thing!

frontend/model/contracts/group.js
snowteamer: That's the whole point of using Flow :p

frontend/model/state.js
taoeffect: OK, this is fine

frontend/controller/router.js
snowteamer: Yes, Flow complained about it, if I remember correctly (not 100% sure though)

.Gruntfile.babel.js
snowteamer: Agreed, let's not do type-checking when no file has changed yet.
I had mainly included Flow here for consistency, because of the three linting tasks that were already listed.

.flowconfig
snowteamer: It was there initially, but I noticed that no top-level `ignored` folder existed, so I thought it would be better to remove it; but I can add it back of course.

shared/GIMessage.js
snowteamer: I'm going to try that. Let's hope Flow doesn't complain!

.flowconfig
taoeffect: Thanks! Yes it does exist on my machine, and anyone else's machine who wants such a folder. 😄

.Gruntfile.babel.js
snowteamer: Besides, I'm too in favor of a progressive integration rather than a radical change in the requirements.
But I'm hesitating regarding the best place to include `exec:flow` inside the `grunt test` and/or `grunt test:unit` task suites.
Should it run after or before the actual tests? I can only tell that if a failed Flow check gets ignored, then it would better run last, so that its console output has a chance to actually be read by the user.

shared/GIMessage.js
snowteamer: Hum... Flow is asking me to put the literal annotations back, because these static values get later used
as computed properties:
image

.Gruntfile.babel.js
taoeffect: Hmm, that's a great question, and I'm not sure as to the answer.

The problem with running it after the tests is that we want to make sure that it *isn't* run if any of the tests fail, so that the program exits with a non-zero exit code and tells Travis things are broken.

And on the other hand, in most traditional development flows, these kinds of checks would run first thing, before the program ran. So maybe where you've placed it here is perfect. We can remove it in the future if it starts causing problems. And since you've actually fixed all the flow errors, maybe this is good hygiene for us to avoid massive PRs in the future as the flow errors diverge from the codebase.

So I'm going to resolve this thread. Let's keep it here for now.

shared/GIMessage.js
taoeffect: Jeez... does it really need the type annotations then? Can you simply remove them? (Is it that dumb that it needs them here?)

snowteamer: A possible solution would be to use a language construct that avoids the need for computed property syntax, for example a switch statement (somewhat ugly), or building `opFns` using `Object.fromEntries()` (might need to be extracted so that it'd be only built once, rather than on every invocation).

frontend/controller/backend.js
taoeffect: This actually is the correct fix. Great job @snowteamer! I made the same change in #1047.

frontend/utils/image.js
snowteamer: I am no longer sure about rejecting the promise, we should make sure to not break the UI somewhere because of a regular file I/O error. This likely needs further investigation, so perhaps printing a warning to the console would be fine for now?

shared/GIMessage.js
snowteamer: @taoeffect If I remove these type annotations, it starts complaining about not being able to build a typed interface for that module :(

taoeffect: Yeah we can keep the literals if that's the best way forward

frontend/utils/image.js
taoeffect: Sure

taoeffect (approved):
Absolutely incredible work @snowteamer!

PR #1049, opened 2021-02-10
Minimal changes to get to our old 'adjust-first' algorithm

CrazedCoding:
So I was made aware that I wasn't following proper team procedures. Sorry about that, friends.

Here is the originally requested PR, with the originally described adjust-first distribution function in `frontend/utils/distribution/grop-income-distribution.js`

Please see the broken values in "Chunk C" of the `frontend/utils/distribution/grop-income-distribution.test.js` mocha tests.

I think this should be sufficient proof that a more computationally complex distribution function is required. I would love to know everyone's thoughts on this!


frontend/model/contracts/group.js
```diff
@@ -370,7 +370,14 @@ DefineContract({
             toUser.push(data.paymentHash)
           }
           paymentMonth.lastAdjustedDistribution = groupIncomeDistribution({
-            state, getters, updateMonthstamp, adjusted: true
+            getters: { // For some reason "{ getters: getters, ... }" results in undefined getters. Fixed below.
+              groupProfiles: getters.groupProfiles,
+              groupSettings: getters.groupSettings,
+              currentGroupState: getters.currentGroupState,
+              monthlyPayments: getters.groupMonthlyPayments
+            },
```
taoeffect: You should be able to pass in simply `getters` as before.

frontend/model/contracts/group.js
```diff
@@ -370,7 +370,14 @@ DefineContract({
...
+            monthstamp: updateMonthstamp,
+            adjusted: false
```
taoeffect: This is setting `lastAdjustedDistribution`, so why are you passing in `adjusted: false`?

package-lock.json
```diff
@@ -3280,7 +3280,6 @@
       "version": "7.2.5",
       "resolved": "https://registry.npmjs.org/@babel/polyfill/-/polyfill-7.2.5.tgz",
       "integrity": "sha512-8Y/t3MWThtMLYr0YNC/Q76tqN1w30+b0uQMeFUYauG2UGTR19zyUtFrAzT23zNtBxPp+LbE5E/nwV/q/r3y6ug==",
-      "deprecated": "🚨 This package has been deprecated in favor of separate inclusion of a polyfill and regenerator-runtime (when needed). See the @babel/polyfill docs (https://babeljs.io/docs/en/babel-polyfill) for more information.",
```
taoeffect: What version of NPM are you running? This file shouldn't be modified...

frontend/views/containers/contributions/GroupPledgesGraph.vue
```diff
@@ -86,40 +88,40 @@ export default {
...
-          .filter(i => i.to === this.ourUsername)
+        ourIncomeToReceive = groupIncomeDistribution({
+          getters: {
+            groupProfiles: this.groupProfiles,
+            groupSettings: this.groupSettings,
+            currentGroupState: this.currentGroupState,
+            monthlyPayments: this.groupMonthlyPayments
+          },
+          monthstamp: currentMonthstamp,
+          adjusted: true
```
taoeffect: Same concern here as above. Also, please check the intended behavior when passing in variables, and if you aren't sure, feel free to ask on Slack. Here, for example, `adjusted` should be `false`.

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -157,19 +166,26 @@ export default {
...
+      groupIncomeDistribution({
+        getters: {
+          groupProfiles: this.groupProfiles,
+          groupSettings: this.groupSettings,
+          currentGroupState: this.currentGroupState,
+          monthlyPayments: this.groupMonthlyPayments
+        },
+        monthstamp: currentMonthstamp,
+        adjusted: true
+      }).filter((i) => i.to === this.ourUsername)
```
taoeffect: OK, I have no idea why this was modified to be this... `this.ourPayments.todo` is supposed to be used here, and that in turn is supposed to call `groupIncomeDistribution` (down the line). So putting it here doesn't make sense.

taoeffect (changes requested):
Nice PR @CrazedCoding! We're making steady progress on this.

Although I didn't have time today to dive down into the details of the tests, I did do a quick scan of the code and wrote some comments about things that stood out to me that you can work on (along with the other stuff mentioned in DMs) while I continue to study the tests to give you feedback regarding those.

package-lock.json
CrazedCoding: Hmmm I didn't see this sneak past me. Thanks for pointing it out.

I am using npm version 7.5.3

frontend/views/containers/contributions/GroupPledgesGraph.vue
CrazedCoding: I can change adjusted to be false here.

I don't think I can change the `getters` to be more DRY, though, because `getters` is `undefined` here.

frontend/views/containers/payments/RecordPayment.vue
CrazedCoding: Thanks for explaining this section of code.

I believe I modified it while attempting to make the cypress test work.

I think it snuck past, like the `package-lock.json`. I will change it back in the next commit.

frontend/model/contracts/group.js
CrazedCoding: Good point; fixed in next commit.

PR #1050, opened 2021-02-24
Task/763 various payment bugs 6

CrazedCoding:
Here is the originally requested PR, with the originally described adjust-first distribution function in frontend/utils/distribution/grop-income-distribution.js WITH the frontend/utils/distribution/mincome-proportional.js core distribution logic.

Please see the broken values in "Chunk A: Scenario 3" of the frontend/utils/distribution/grop-income-distribution.test.js mocha tests:
```
const dist = groupIncomeDistribution(unfoldParameters(
  {
    mincome: 1000,
    profiles: {
      u1: { have: 100 },
      u2: { need: 50 },
      u3: { need: 300 }
    },
    adjustWith: [
      { u1: { u2: 25 } }
    ]
  }))
should(dist).eql([
  { amount: 69.23076923, from: 'u1', to: 'u3' }
])
```
The above amount should have a small amount sent from u1 to u2.

Also, the adjust-first approach breaks the four-way distribution test:

```
1) Chunk C: 4-way distribution tests
[SCENARIO 2]: stops asking user to pay someone they fully paid their share to:

AssertionError: expected Array [
  Object { amount: 111.11111111, from: 'u1', to: 'u3' },
  Object { amount: 27.77777778, from: 'u4', to: 'u2' },
  Object { amount: 138.88888889, from: 'u4', to: 'u3' }
] to equal Array [
  Object { amount: 125, from: 'u1', to: 'u3' },
  Object { amount: 50, from: 'u4', to: 'u2' },
  Object { amount: 125, from: 'u4', to: 'u3' }
] (at '0' -> amount, A has 111.11111111 and B has 125)
```

The problem is that the adjust-first approach changes the proportions of the core distribution logic before payments are calculated. I think this should be sufficient proof that a more computationally complex distribution function is required.

I think the best solution to all this is to make it so that the income distribution payments _**accumulate**_ over time, and not be done on a strict monthly basis, otherwise if people were being honest about their haves/needs then the havers would always need to pay for the previous month (late) after they had time to accumulate the money for the previous month.


frontend/utils/distribution/group-income-distribution.js
```diff
@@ -19,7 +40,7 @@ export default function groupIncomeDistribution ({ state, getters, monthstamp, a
       const amount = adjustment + profile[incomeDetailsType]
       currentIncomeDistribution.push({
         name: username,
-        amount: saferFloat(amount)
+        amount: saferFloat(amount) - (adjusted ? totalPaymentsToOrFrom(paymentsFrom, allPayments, username) : 0)
```
taoeffect: Noticing a couple issues:

1. Whenever a final or "final-ish" amount is being calculated, that needs to be wrapped in `saferFloat`. Here, `saferFloat` is being undermined because the calculation for the adjustment is subtracted from `amount` outside of `saferFloat`.
2. The adjust-first approach is called such because the adjustment is done beforehand. Below, the code for the adjustment remains in place when it should be moved/removed (that was the adjust-after approach).

test/cypress/integration/group-paying.spec.js
```diff
@@ -183,9 +183,9 @@ describe('Group Payments', () => {
     cy.getByDT('link-PaymentRowSent').click()
     cy.getByDT('payList').find('tbody').children().should('have.length', 2)
     cy.getByDT('payList').within(() => {
-      cy.getByDT('payRow').eq(1).find('td:nth-child(1)').should('contain', `user3-${userId}`)
-      cy.getByDT('payRow').eq(1).find('td:nth-child(2)').should('contain', '$100')
-      cy.getByDT('payRow').eq(1).find('td:nth-child(3)').should('contain', humanDateToday)
+      cy.getByDT('payRow').eq(0).find('td:nth-child(1)').should('contain', `user3-${userId}`)
+      cy.getByDT('payRow').eq(0).find('td:nth-child(2)').should('contain', '$100')
+      cy.getByDT('payRow').eq(0).find('td:nth-child(3)').should('contain', humanDateToday)
```
taoeffect, line 188: It's unclear to me why this was changed. Could you elaborate, was this intentional? If so, why?

taoeffect (changes requested):
Nice job @CrazedCoding! It's looking cleaner and cleaner. 👍 

Preliminary review done!

frontend/utils/distribution/group-income-distribution.test.js
```diff
@@ -0,0 +1,408 @@
...
+              u2: { need: 50 },
+              u3: { need: 300 }
+            },
+            adjustWith: [
+              { u1: { u2: 25 } }
+            ]
+          }))
+        should(dist).eql([
+          { amount: 69.23076923, from: 'u1', to: 'u3' }
+        ])
```
taoeffect: This scenario's expected value is incorrect. Note the comment says:

```
// Expected Result: most of the $75 would go to u3, and some of it would go to u2
```

So what should happen is that the adjustment should remove $25 from u1's pledge, leaving him with $75 to split between u2 and u3. Meaning:

1. The "should" portion of this currently has a single entry. It should have two.
2. The amounts that u1 should be asked to give to u2 and u3 should sum up to $75 (the remaining amount that he has).
3. The amount sent to u3 should be less than $25, because now the algorithm is taking into account the increased need of u3.

Specifically, based off of my tests, it should equal:

```js
[
  { amount: 5.76923077, from: 'u1', to: 'u2' },
  { amount: 69.23076923, from: 'u1', to: 'u3' }
]
```

Note: This is with using `saferFloat` inside of `mincome-proportional.js`, e.g. `amount: saferFloat(paymentAmount)`.

I did not check the other scenarios, but I would imagine they might also need adjustment.

taoeffect:
Recommended next steps:

1. Review the comments above and implement the changes.
2. Modify `mincome-proportional.js` to use `saferFloat` when setting the amoung, e.g.:

   ```js
          const payment = {
            amount: saferFloat(paymentAmount),
            from: fromName,
            to: toName
          }
          payments.push(payment)
   ```

3. Per the review comments, modify the `groupIncomeDistribution` to remove the old post-facto adjustment:

    ```js
    export default function groupIncomeDistribution ({ state, getters, monthstamp, adjusted }: Object): any {
      // the monthstamp will always be for the current month. the alternative
      // is to allow the re-generation of the distribution for previous months,
      // but that approach requires also storing the historical mincomeAmount
      // and historical groupProfiles. Since together these change across multiple
      // locations in the code, it involves less 'code smell' to do it this way.
      // see historical/group.js for the ugly way of doing it.
      const mincomeAmount = getters.groupMincomeAmount
      const groupProfiles = getters.groupProfiles
      const currentIncomeDistribution = []
      const allPayments = getters.currentGroupState.payments
      const thisMonthPayments = getters.monthlyPayments ? getters.monthlyPayments[monthstamp] : null
      const paymentsFrom = thisMonthPayments && thisMonthPayments.paymentsFrom
      for (const username in groupProfiles) {
        const profile = groupProfiles[username]
        const incomeDetailsType = profile && profile.incomeDetailsType
        if (incomeDetailsType) {
          const adjustment = incomeDetailsType === 'incomeAmount' ? 0 : mincomeAmount
          const amount = adjustment + profile[incomeDetailsType]
          currentIncomeDistribution.push({
            name: username,
            amount: saferFloat(amount - (adjusted ? totalPaymentsToOrFrom(paymentsFrom, allPayments, username) : 0))
          })
        }
      }
      return incomeDistribution(currentIncomeDistribution, mincomeAmount)
    }
    ```

4. Update the tests accordingly, per review, and also other tests that likely will need to be double-checked and updated for anything missed.
5. Merge `master` once #1051 is reviewed, approved and merged.

test/cypress/integration/group-paying.spec.js
CrazedCoding: Yes, it was intentional. The new distribution function appeared to give the results in a different order, breaking this cypress test in the process.

The solution was to correct the row-index which cypress checked for payments by this user.

IE: cypress had correct values, but was checking in the wrong row for them.

Maybe we should rewrite this so that the of the payments listed doesn't matter, or maybe sort the payment distributions by amount.

frontend/utils/distribution/group-income-distribution.test.js
CrazedCoding: Thanks for pointing this out. I probably should have changed the scenario to a `skip` scenario instead of changing it's values.

I did mention this scenario at the very beginning of this PR:

> Here is the originally requested PR, with the originally described adjust-first distribution function in frontend/utils/distribution/grop-income-distribution.js WITH the frontend/utils/distribution/mincome-proportional.js core distribution logic.
> 
> Please see the broken values in "Chunk A: Scenario 3" of the frontend/utils/distribution/grop-income-distribution.test.js mocha tests:
> 
> ```
> const dist = groupIncomeDistribution(unfoldParameters(
>   {
>     mincome: 1000,
>     profiles: {
>       u1: { have: 100 },
>       u2: { need: 50 },
>       u3: { need: 300 }
>     },
>     adjustWith: [
>       { u1: { u2: 25 } }
>     ]
>   }))
> should(dist).eql([
>   { amount: 69.23076923, from: 'u1', to: 'u3' }
> ])
> ```
> 
> The above amount should have a small amount sent from u1 to u2.

frontend/utils/distribution/group-income-distribution.js
CrazedCoding: Point #1 Response: I totally agree. I'll fix this error ASAP

Point #2 Response: I also agree. I'll fix this error ASAP.

CrazedCoding:
Made all requested fixes.

The 4-way distribution tests are still broken with the adjust-first approach:

```
1) Chunk C: 4-way distribution tests
     [SCENARIO 2]: stops asking user to pay someone they fully paid their share to:

    AssertionError: expected Array [
Object { amount: 22.22222222, from: 'u1', to: 'u2' },
Object { amount: 111.11111111, from: 'u1', to: 'u3' },
Object { amount: 27.77777778, from: 'u4', to: 'u2' },
Object { amount: 138.88888889, from: 'u4', to: 'u3' }
] to equal Array [
Object { amount: 125, from: 'u1', to: 'u3' },
Object { amount: 50, from: 'u4', to: 'u2' },
Object { amount: 125, from: 'u4', to: 'u3' }
] (at length, A has 4 and B has 3)
```

We should discuss this over a call, I think.

test/cypress/integration/group-paying.spec.js
taoeffect: How is this possible? These tests are passing on `master`...

CrazedCoding: Damn. Good. Point.

My initial thoughts are that perhaps the adjust-last algorithm somehow swapped the two payments value, but the name of the "to-user" of the payments are also swapped.

Definitely a puzzle.

Looking into.

taoeffect:
Closing in favor of #1056

Issue #576, opened 2019-06-25
Remove primus dependency [$3k bounty]


taoeffect:
### Problem

- Primus is a large dependency (and we don't need it)
- Primus constantly regenerates `frontend/controller/utils/primus.js`, which is kinda annoying
- `npm audit` gives two low severity vulnerabilities, one for `primus-rooms` and one for `primus-responder` because they use an old version of `debug`

### Solution

- Research issues #358 and #539
- Based on that decide whether to use websockets or HTTP2 (EDIT: based on preliminary readying it's looking like sticking with websockets for live updates is the way to go)
- Then remove primus and rewrite all the code using either raw `ws` calls or whatever the HTTP2 equivalent is

Research:

- Hapi-specific integration:
  - https://github.com/rse/hapi-plugin-websocket <- this seems the most promising at the moment
  - https://github.com/hapijs/nes
- https://github.com/websockets/ws
- https://www.npmjs.com/package/@small-tech/express-ws
- https://github.com/versatica/protoo
- https://github.com/theturtle32/WebSocket-Node
- https://blog.jayway.com/2015/04/13/600k-concurrent-websocket-connections-on-aws-using-node-js/
- https://patchbay.pub/ <- interesting

The solution should rely on as few moving pieces as possible so that it works well with our future switch to Deno (#586).

### Bounty

$3k bounty for a clean solution to this (paid in cryptocurrency).

taoeffect:
Adding a $3k bounty for closing this issue for non-contractors (meaning, if you don't have a development contract with okTurtles, and you'd like to close this issue, you can work on it, and if we accept your PR, you'll receive the bounty).

Anyone else can add to the bounty too if they'd like by following these simple steps:

1. Donate the amount you'd like to dedicate to this issue [to the okTurtles Foundation](https://okturtles.org/donate/).
2. Make a comment here (or privately to us via Gitter or Slack) the same day letting us know you'd like your donation to be used for a bounty to close an issue, and which issue that is.

PR #1051, opened 2021-02-24
quickfix for autoBanSenderOfMessage + commented gettersProxy

taoeffect:
This PR:

- adds a quickfix for a bug in `state.js` related to `autoBanSenderOfMessage` usage of `GIMessage`
- adds comments explaining what the purpose of `gettersProxy` is in `Contract.js`
- includes a modified version of `primus.js` based on their latest changes (getting rid of Primus is an open issue with a bounty: #576)

### DIFF
--- a/frontend/model/contracts/Contract.js
+++ b/frontend/model/contracts/Contract.js
@@ -100,9 +100,15 @@ export function DefineContract (contract: Object) {
   }
 }
 
+// the gettersProxy is what makes Vue-like getters possible. In other words,
+// we want to make sure that the getters functions that we defined in each
+// contract that we create get passed the 'state' when a getter is accessed.
+// The only way to pass in the state is by creating a Proxy object that does
+// that for us. This allows us to maintain compatibility with Vue.js and integrate
+// the contract getters into the Vue-facing getters.
 function gettersProxy (state: Object, getters: Object) {
   const proxyGetters = new Proxy({}, {
-    get: function (obj, prop) {
+    get: function (target, prop) {
       return getters[prop](state, proxyGetters)
     }
   })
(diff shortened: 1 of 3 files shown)


frontend/model/contracts/Contract.js
```diff
@@ -100,9 +100,15 @@ export function DefineContract (contract: Object) {
   }
 }
 
+// the gettersProxy is what makes Vue-like getters possible. In other words,
+// we want to make sure that the getters functions that we defined in each
+// contract that we create get passed the 'state' when a getter is accessed.
+// The only way to pass in the state is by creating a Proxy object that does
+// that for us. This allows us to maintain compatibility with Vue.js and integrate
+// the contract getters into the Vue-facing getters.
 function gettersProxy (state: Object, getters: Object) {
```
snowteamer, line 109: Nit: most style guides mandate the use of a leading capital letter in comments.

Also, if I understand the text correctly, then there are two small typos:
- `the getters functions` -> `the getter functions`
- `each contract that we create` -> `each contract that we created`

frontend/model/contracts/Contract.js
```diff
@@ -100,9 +100,15 @@ export function DefineContract (contract: Object) {
...
   const proxyGetters = new Proxy({}, {
-    get: function (obj, prop) {
+    get: function (target, prop) {
```
snowteamer: The renaming of `obj` is nice since it was an abbreviation.

Besides, shorthand method syntax could be used here to get rid of the `function` keyword:
```js
{
  get (target, prop) {
    ... 
```

snowteamer (dismissed):
That's nice to see the `proxyGetters` stuff explained.

Not much else to say, maybe a nit or two, but overall LGTM.

snowteamer (approved):

PR #1052, opened 2021-02-27
#576 Drop dependencies to Primus and Primus plugins

### LINKED ISSUE #576: Remove primus dependency [$3k bounty]

snowteamer:
Closes #576

### Summary of changes:
- Drops Primus.js and the `primus-rooms` and `primus-responder` plugins in favor of a pubsub implementation using only the `ws` socket library in Node and the WebSocket API in the browser.
- Increased the initial WebSocket connection timeout from 3000 to 5000 ms (Primus defaults to 10000 ms).
- Correction of a bug that often caused redundant sub/unsub requests.
- Less verbose backend console logging (from 19 k lines of build output to ~10 500 lines).
- New `'backend/server/broadcastEntry'` SBP selector.

### Additional information:
The core of this proposed pubsub framework consists of only two files:
- `~/backend/utils/pubsub.js`  which exports a `createServer()` function.
- `~/shared/pubsub.js` which exports a `createClient()` factory suitable to create either ordinary pubsub clients in the browser or server-side clients in Node.
  Such server-side clients can be useful e.g. in testing or to talk to other servers.

Currently, it comes with a set of so-called "default" event handlers, designed to handle only basic pubsub protocol stuff and `ENTRY` messages as required by the application.

However this protocol could be extended by providing appropriate "custom" message handlers to the factories.

### In progress:
Primus has automatic reconnection features which are not all implemented in this PR yet:

- [x] Reconnect upon initial connection timeout
- [x] Reconnect when the server restarts
- [x] Reconnect when the `online` event occurs, including in case the web frontend was started in offline mode. It seems to be working now, but again tests are still missing.
- [x] Client-side connection loss detection independently of the `offline/online` events using ping/pong messages (https://github.com/primus/primus#heartbeats-and-latency).

All these features are options so they can be disabled if we do not need them.

Again, testing these features might be tricky, so any help or advice will be much appreciated.


backend/server.js
```diff
@@ -45,10 +44,13 @@ sbp('sbp/selectors/register', {
   'backend/server/handleEntry': async function (entry: GIMessage) {
     const contractID = entry.contractID()
     await sbp('gi.db/log/addEntry', entry)
-    const response = makeResponse(RESPONSE_TYPE.ENTRY, entry.serialize())
     console.log(chalk.blue.bold(`broadcasting ${entry.description()}`))
-    sbp('okTurtles.data/apply', PUBSUB_INSTANCE, p => {
-      p.room(contractID).write(response)
+    sbp('okTurtles.data/apply', PUBSUB_INSTANCE, pubsub => {
+      if (contractID in pubsub.subscribersByContractID) {
```
joepie91: The `... in ...` operator has surprising behaviour when dealing with existent properties that contain an `undefined` value:

```js
> "foo" in {foo: undefined}
true
> "foo" in {}
false
```

Is that really the behaviour we want? Especially with shorthand code along the lines of...

```js
return {
    // ... some other properties ...
    someProperty: (someBooleanCondition)
        ? makeThing(id)
        : undefined
};
```

... it's realistically possible for existent `undefined` properties to exist, and it seems that in those cases we would want this check to fail, not pass.

Approaches that don't suffer from this problem are:

```js
// Check fails only if the property does not exist or contains `undefined`
if (pubsub.subscribersByContractID[contractID] === undefined)

// In addition to the above, the check *also* fails if the property contains `null` -- this one is more common
if (pubsub.subscribersByContractID[contractID] == null)
```

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+  const server = new WebSocket.Server({
+    ...defaultOptions,
+    ...options,
+    server: httpServer
+  })
+  server.customClientHandlers = { ...options.clientHandlers }
+  server.customServerHandlers = { ...options.serverHandlers }
+  server.messageHandlers = { ...defaultMessageHandlers, ...options.messageHandlers }
+  server.pingInterval = undefined
+  server.subscribersByContractID = Object.create(null)
```
joepie91: Isn't a `Map` more appropriate here?

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+ * @returns {Object}
+ */
+export function createServer (httpServer: Object, options?: Object = {}): Object {
+  const server = new WebSocket.Server({
+    ...defaultOptions,
+    ...options,
+    server: httpServer
+  })
+  server.customClientHandlers = { ...options.clientHandlers }
+  server.customServerHandlers = { ...options.serverHandlers }
```
joepie91: Why are we cloning the objects here?

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+      ? xForwardedFor.split(/\s*,\s*/)[0]
+      : request.socket.remoteAddress
+  )
+  console.log(bold(`[pubsub] Socket ${socket.id} connected from ${remoteAddress}`))
+
+  // Create and attach socket event listeners.
+  ;['close', 'error', 'message', 'ping', 'pong'].forEach((eventName) => {
+    socket.on(eventName, (...args) => {
+      console.log(`[pubsub] Event '${eventName}' on socket ${socket.id}`, ...args)
+      const customHandler = socket.server.customClientHandlers['on' + eventName]
```
joepie91: Is there any reason why we can't just remove the `on` prefix for handlers entirely, and key the object's entries by their `eventName` directly? Like `{ message: ..., ping: ... }`, etc.

With the current format it doesn't follow naming conventions (`onmessage` instead of `onMessage`) and it's not super readable, and I don't feel that it's worth it to add "clever" capitalization logic here, if we can also just omit the prefix entirely.

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+  socket.id = generateSocketID()
+  socket.isAlive = true
+  socket.server = this
+  socket.subscriptions = new Set()
+
+  const xForwardedFor = request.headers['x-forwarded-for']
+
+  const remoteAddress = (
+    typeof xForwardedFor === 'string' && xForwardedFor !== ''
+      ? xForwardedFor.split(/\s*,\s*/)[0]
```
joepie91: This is a security issue; with this, *any* malicious client could send a spoofed `X-Forwarded-For` header, and have its contents blindly accepted as truthful, thereby allowing them to obscure their own IP address.

This is the reason that off-the-shelf HTTP libraries usually require you to explicitly specify 'trusted proxies', ie. specific IP addresses (typically reverse proxies operated by the ops department) whose `X-Forwarded-For` headers are trusted; they are ignored for everyone else.

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+  // Create and attach socket event listeners.
+  ;['close', 'error', 'message', 'ping', 'pong'].forEach((eventName) => {
+    socket.on(eventName, (...args) => {
+      console.log(`[pubsub] Event '${eventName}' on socket ${socket.id}`, ...args)
+      const customHandler = socket.server.customClientHandlers['on' + eventName]
+      const defaultHandler = defaultClientHandlers['on' + eventName]
+
+      try {
+        if (defaultHandler) {
+          defaultHandler.call(socket, ...args)
```
joepie91: Not a fan of smuggling the socket in through the `this` context - this will not work when the event handlers are specified as arrow functions, for example (as is increasingly common nowadays). IMO the socket should be passed as an explicit argument to the handler instead.

backend/utils/pubsub.js
```diff
@@ -0,0 +1,261 @@
...
+    typeof xForwardedFor === 'string' && xForwardedFor !== ''
+      ? xForwardedFor.split(/\s*,\s*/)[0]
+      : request.socket.remoteAddress
+  )
+  console.log(bold(`[pubsub] Socket ${socket.id} connected from ${remoteAddress}`))
+
+  // Create and attach socket event listeners.
+  ;['close', 'error', 'message', 'ping', 'pong'].forEach((eventName) => {
+    socket.on(eventName, (...args) => {
+      console.log(`[pubsub] Event '${eventName}' on socket ${socket.id}`, ...args)
```
joepie91: Should this not be `console.debug` instead?

joepie91 (dismissed):
Initial review; it's end of day for me now, I will likely post further review comments later :)

backend/utils/pubsub.js
snowteamer: Yes, at least in theory, because `Map` objects currently do not appear to be used anywhere in the codebase, despite likely more than one appropriate use case. Perhaps because of their API which might feel heavier than object property access.
So I've used an object for consistency. It's fine to be used as a map since all keys are going to be strings and a null prototype object doen't initially have any property.

backend/utils/pubsub.js
snowteamer: Thanks for pointing this, we'll definetely have to consider such security issues.

Fow now this remote address value is only used in server console output for debugging during development and testing; it doesn't get stored or exposed.

backend/utils/pubsub.js
snowteamer: Yes, possibly. I don't precisely understand when to use which logging level, so I generally just use `console.log` everywhere.

.Gruntfile.babel.js
```diff
@@ -116,14 +115,7 @@ module.exports = (grunt) => {
...
   grunt.registerTask('build', function () {
     const rollup = this.flags.watch ? 'rollup:watch' : 'rollup'
-    if (this.flags.watch) {
-      // if we are being run with 'grunt dev', tell Primus to generate the file
-      // ./frontend/controller/utils/primus.js, otherwise, this is 'grunt test', and this
-      // block will be skipped so that pubsub.js can be required through ./backend/index.js
-      // TODO: get rid of this when we remove Primus (#576)
-      require('~/backend/pubsub.js') // hack to register 'backend/pubsub/setup' selector
-      sbp('backend/pubsub/setup', require('http').createServer(), true)
-    }
```
taoeffect, line 119: Doesn't it feel great to be able to delete this hacky crap? 😄

backend/utils/pubsub.js
snowteamer: We can also do as you suggest. I just thought it was nicer to pass options like `{ onerror: ..., onmessage: ... }` rather than  `{ error: ..., message: ... }`, because the common "on" prefix would act as a visual clue indicating that all these properties form a group of related event handlers.

As for the naming convention, it's still common to see such property names on objects implementing the `EventTarget` interface. For example `WebSocket` objects still have `onclose`, `onopen`, etc., although nowadays the preferred way to add listeners is via the `addEventListener()` method used with names without the "on" prefix.

.Gruntfile.babel.js
```diff
@@ -321,7 +310,7 @@ module.exports = (grunt) => {
             // NOTE: uncommenting this saves ~1 second off build process
             //       while making it a massive pain to deal with dependencies
             //       and causing "ReferenceError: require is not defined"
-            // include: /(node_modules\/(blakejs|multihashes|tweetnacl|localforage|@babel|vue.+).*|primus\.js$)/,
+            // include: /(node_modules\/(blakejs|multihashes|tweetnacl|localforage|@babel|vue.+).*)/,
```
taoeffect, line 306: Nice job updating this comment too 👍

backend/utils/pubsub.js
snowteamer: Well, its a trade-off between the `EventEmitter` and `EventTarget` interfaces.

When using an `EventEmitter`, the event target doesn't get passed through the `this` binding, making arrow functions a cleaner choice for event callbacks in pure Node code.

However when using an `EventTarget`, e.g. a WebSocket object in the browser, the socket does get passed through the `this` binding. Frontend developers are used to this behavior, so they will likely not use arrow functions as listeners there.

Since I had to implement both server-side and client-side components, I have chosen to favor the `EventTarget` interface to avoid using two different interfaces.
Other benefits may include a higher similarity between those components, and an hopefully an easier migration path to Deno.

Also, arrow functions are not needed to represent event handler maps when using shorthand method syntax.

Quick comparison:

1. Example using shorthand method syntax and socket smuggling:
```ts
const defaultClientHandlers = {
  onclose (code: string, reason: string) {
    ...
    this.subscriptions.clear()
  },
  onmessage (data: string) {
    ...
  }
}
```
2. Example using shorthand method syntax and explicit socket passing:
```ts
const defaultClientHandlers = {
  onclose (socket: Object, code: string, reason: string) {
    ...
    socket.subscriptions.clear()
  },
  onmessage (socket: Object, data: string) {
    ...
  }
}
```
3. Example using arrow functions. Not really nice here:
```ts
const defaultClientHandlers = {
  onclose: (socket: Object, code: string, reason: string) => {
    ...
    socket.subscriptions.clear()
  },
  onmessage: (socket: Object, data: string) => {
    ...
  }
}
```

backend/utils/pubsub.js
snowteamer: To remove any coupling between them and the passed in `options` object.

This way, subsequent changes to the `options` object cannot alter server behavior as a side effect.

Even in nobody ever mutates the `options` object after using it, decoupling could still help the engine and various tools.

backend/server.js
snowteamer: I agree that an accompanying type annotation could have made it clearer, but yes this is the intended behavior.

Every property value found on the `pubsub.subscribersByContractID` object must be a `Set` of contract IDs.

Otherwise, if unexpected properties or property values can get here, an error should be reported, preferably at build time.

However, the check is still needed because a matching property for the given contract ID might not have been defined here yet, for example if there are no subscriber yet because the contract just got created and the `GIMessage` being handled is the very first one of the contract.

backend/utils/pubsub.js
joepie91: Nevertheless it's currently being presented as *the* client's IP address. If someone were to change the code in the future, using this variable for something that *does* have security impact, it is very unlikely that they will re-evaluate whether the safe-usage assumptions that are made here still hold - and so they will likely introduce a security issue.

If we're going to keep using this 'naive' implementation, then we should at least tag the resulting variable as being unsafe, eg. calling it `unsafeRemoteAddress` instead of just `remoteAddress`, and leave a comment explaining the issue at the definition site. This provides an explicit cue to any future developer working with the code that they need to carefully evaluate the implications of using it.

This also applies more generally to cases of "it's fine that we're doing an unsafe thing because we're only using it safely" - those usage assumptions almost always get invalidated over time, and this is a frequent source of real-world security issues in software. Unsafe things should therefore always be tagged to communicate this.

backend/utils/pubsub.js
joepie91: > We can also do as you suggest. I just thought it was nicer to pass options like `{ onerror: ..., onmessage: ... }` rather than `{ error: ..., message: ... }`, because the common "on" prefix would act as a visual clue indicating that all these properties form a group of related event handlers.

I would normally agree, but in this case it seems that all the handlers are explicitly defined in a property that has 'handlers' in the name, so I think that that would be sufficient visual cue to communicate this. Unless there's something I missed in how this is being used? I'm still getting used to the codebase :)

> As for the naming convention, it's still common to see such property names on objects implementing the `EventTarget` interface. For example `WebSocket` objects still have `onclose`, `onopen`, etc., although nowadays the preferred way to add listeners is via the `addEventListener()` method used with names without the "on" prefix.

Those property naming conventions are fairly widely considered to be a legacy holdover that everyone would rather forget about though :) For browser API spec people it makes sense to continue following it for consistency, but I don't think that we need to follow any such convention, especially if we're specifying them as options rather than setting them as properties anyway.

backend/utils/pubsub.js
joepie91: > Well, its a trade-off between the `EventEmitter` and `EventTarget` interfaces.
> 
> When using an `EventEmitter`, the event target doesn't get passed through the `this` binding, making arrow functions a cleaner choice for event callbacks in pure Node code.
> 
> However when using an `EventTarget`, e.g. a WebSocket object in the browser, the socket does get passed through the `this` binding. Frontend developers are used to this behavior, so they will likely not use arrow functions as listeners there.
> 
> Since I had to implement both server-side and client-side components, I have chosen to favor the `EventTarget` interface to avoid using two different interfaces.
> Other benefits may include a higher similarity between those components, and an hopefully an easier migration path to Deno.

Honestly, I don't think that emulating the event interface of browser APIs - which is notoriously annoying to work with - is a good goal to strive for, here. The broader JS ecosystem has essentially standardized on the `EventEmitter` approach and variants thereof, precisely because it has less weird design caveats and is much closer to how the rest of JS works semantically.

Plus, if we're going to try and make it consistent with the `EventTarget` design, then the current implementation would still not meet that bar - it's typical for browser-style event handlers to be called with a singular `Event` object that contains all the event data, rather than multiple arguments of varying types, as we are doing here. This is another reason that IMO the `EventEmitter` design is more useful.

(Speaking of, why *aren't* we just using Node's core EventEmitter? It should bundle fine.)

> Also, arrow functions are not needed to represent event handler maps when using shorthand method syntax.

Correct, but this doesn't hold when you start adding levels of abstraction. For example, consider:

```js
function createThing(options) {
	// implementation using the `this`-based API design for event handlers
	// ...
}

function createCustomThing(handler) {
	return createThing({
		handlers: {
			onStuff: handler
		}
	});
}

createCustomThing((eventArg) => {
	doWhatever(eventArg, 2)
});
```

In the `createCustomThing` invocation, it's not possible to use shorthand method syntax, because we're not in an object context. The same would be true if we first assign the handler to a variable, or process it in any way that isn't a literal specification within an object. 

Edit: ... which also makes it painful to write event handlers that are meant to interact with some other `this` context surrounding it, having to revert to `self = this` type hacks.

backend/utils/pubsub.js
joepie91: Is there a concrete reason to believe that that will occur here, or is this intended as a general precaution? In the latter case, it doesn't seem like this is currently done anywhere else in the code, from a quick search - so perhaps this should be checked for other code as well, then. Otherwise you still can't assume that options passed to an arbitrary function are 'locked in' at call time.

That having been said, having this as a general practice is likely to turn into a rabbit hole, if for whatever reason we end up working with nested data structures in options at some point. Deep copies are expensive and difficult to express sensibly, especially when you're dealing with structures that contain more than just arrays and object literals.

Deep nested structures are frequently also those most subject to mutation (because of aforementioned deep-copy cost), so you'd end up with a situation where you can safely assume that inputs are not mutated afterwards... *except* for those structures which are actually *likely* to get mutated.

Because of this, I've stopped doing this in my own projects - it turned out to just not be worth it in practice for me. Instead, I just handle data immutably wherever possible (so inputs are super unlikely to ever mutate at all), and carefully reason through unavoidable mutation, which has more or less the same result.

backend/utils/pubsub.js
joepie91: Okay, that makes sense. Probably something to evaluate/fix as a separate point in the future, then.

backend/utils/pubsub.js
snowteamer: Good points. Therefore, it might be simpler to just stop trying to log the remote address here.

Anyway there is already logging code in `backend/server.js` that does it on every Hapi request. Does the Hapi `request.info.remoteAddress` suffer from the same issues?

backend/utils/pubsub.js
snowteamer: It is only intended as a general practice.

You have a point that it doesn't prevent nested options to stay coupled. That's one reason why this implementation has only flat options, whereas Primus accepts nested options such as `recovery` and `strategy`.

Thank you for sharing your experience, that's interesting. Plain assignment can certainly be used if you think copying does more harm than good.

frontend/main.js
```diff
@@ -57,13 +58,16 @@ async function startApp () {
...
-  })
-
+  sbp('okTurtles.data/set', PUBSUB_INSTANCE, createGIPubSubClient(
+    sbp('okTurtles.data/get', 'API_URL'), {
+      // TODO: verify these are good defaults
+      // Temporarily disable automatic reconnection until it's robust enough.
+      reconnectOnDisconnection: false,
+      reconnectOnOnline: false,
+      reconnectOnTimeout: false,
+      timeout: 3000
```
taoeffect: Are all of these options implemented, or are some of them not? For example, I couldn't find where `reconnectOnDisconnection` is used.

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
+'use strict'
+
+import type { JSONObject, JSONType } from '~/shared/types.js'
+
+// ====== Types ====== //
+
+export type Message = {
+  [key: string]: JSONType,
+  +type: string
```
taoeffect, line 28: What is this + prefix thing? I've never seen it before. Can you add a comment explaining?

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
...
+}
+
+export type PubSubClient = {
+  connectionTimeoutID: TimeoutID | void,
```
taoeffect, line 32: Where does this `TimeoutID` come from? Similarly, can a comment be added explaining?

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
...
+    listeners: Object.create(null),
+    messageHandlers: { ...defaultMessageHandlers, ...options.messageHandlers },
+    options: { ...defaultOptions, ...options },
+    pendingSubscriptionSet: new Set(),
+    pendingUnsubscriptionSet: new Set(),
+    reconnectionDelayID: undefined,
+    socket: null,
+    subscriptionSet: new Set(),
+    connectionTimeoutID: undefined,
+    url: url.replace(/^http/, 'ws'),
```
taoeffect, line 137: What about https/wss? Should that be handled here too? AFAIK that's how most servers will be setup.

taoeffect (commented):
@snowteamer this is a really fantastic PR you've sent in! Because of its size, it will take me a bit longer to review fully, however I've gone ahead and left some feedback for now so that you can work on it / respond to it while I continue to review.

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
...
+  }
+  client.connect()
+  return client
+}
+
+export function createMessage (type: string, data: JSONType): string {
+  return JSON.stringify({ type, data })
+}
+
+export function createRequest (type: $Values<typeof REQUEST_TYPE>, data: JSONObject): string {
```
taoeffect: I love this `$Values<typeof REQUEST_TYPE>` stuff. You are really a Flowtype master! 👍

shared/pubsub.js
snowteamer: It's just the Flow way of telling that a property should be considered read-only after initialization, so that Flow should report an error if it ever notices that some code is overwriting it.

shared/pubsub.js
snowteamer: It's also a Flow thing, more precisely, an opaque type declared in its core definition file:
```js
declare opaque type TimeoutID;
```
And there is a lot of them...

shared/pubsub.js
snowteamer: The second "s", if present, will not be affected by this prefix replacement:
```js
'https://example.com'.replace(/^http/, 'ws') // 'wss://example.com'
```

frontend/main.js
snowteamer: Only `reconnectOnTimeout` is implemented for now, but I'd like to implement the other ones too.

.Gruntfile.babel.js
snowteamer: It feels awesome haha

shared/pubsub.js
snowteamer: Thank you. I hope this is not too verbose!😄

backend/utils/pubsub.js
snowteamer: So I have followed your advice, however I also had to append `|| {}` instead of just using a simple property assignment.

Now we have `options.clientHandlers || {}` instead of `{ ...options.clientHandlers }`. Both have downsides IMHO.

shared/pubsub.js
taoeffect: Ah I see. Well, that's pretty advanced flow usage for most people I feel. Could you add a line comment after this, something to the effect of:

```js
  +type: string // the '+' tells flow that this property is read-only after initialization
```

This will help people who are unfamiliar with flow quickly grasp what's going on without having to dive through the depths of the Flowtype documentation.

shared/pubsub.js
taoeffect: On the first instance of `TimeoutID` usage in the code, could you similarly add a comment about this? It will help save future developers or archeologists a lot of time when browsing through the code. :)

E.g. something along these lines (if `connectionTimeoutID` is the first place where `TimeoutID` appears):

```js
  connectionTimeoutID: TimeoutID | void, // TimeoutID is an opaque type declared in Flow's core definition file
```

shared/pubsub.js
taoeffect: Ah, perfect. Marking this as resolved.

shared/pubsub.js
taoeffect: Well, it is quite verbose, but also cool that JavaScript with Flow can do this kind of stuff if it's necessary. As long as it's sparingly used and not all over the place as it happens with C++, it's fine. :)

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
...
+  return JSON.stringify(Object.assign({ type }, data))
```
taoeffect, line 183: Is there a reason that `createMessage` does `JSON.stringify({ type, data })` and `createRequest` does `JSON.stringify(Object.assign({ type }, data))`?

shared/pubsub.js
```diff
@@ -0,0 +1,386 @@
...
+}
+
+// ====== Event Handlers ====== //
+
+const defaultHandlers = {
```
taoeffect: Since there are so many different types of handlers in this PR, is it possible to give this a more descriptive name, e.g. something like, `defaultClientSocketEventHandlers`?

shared/pubsub.js
snowteamer: Yes, an annoying Flowtype error, apparently because `data` can possibly have a `type` property which would conflict with the passed in `type` argument.

If I remember correctly, I first tried to solve it using the `$Values<>` trick but it was not enough, then I reluctantly tried `Object.assign()` and Flow stopped complaining.

frontend/main.js
taoeffect: OK yeah, for full socket support it would be good to have the equivalent of the options we were using before working, e.g. `strategy: ['disconnect', 'online', 'timeout']`, which I think amounts to the ability for the client to reconnect automatically in the event the internet goes out, or the server goes down briefly, etc.

shared/pubsub.js
taoeffect: Wow. OK, another opportunity for maybe a line comment, e.g. something like:

```js
return JSON.stringify(Object.assign({ type }, data)) // Use Object.assign to make Flow happy
```

Basically anything unusual and not self-evident is worth commenting.

shared/pubsub.js
snowteamer: Sure! I was trying to limit the verbosity so as to not bother other developers too much, but personally I would not mind a full-length name here.

shared/pubsub.js
snowteamer: Since there is going to be several Flowtype usage related comments, what about grouping them near the top of the file (before the type definitions), like:
```js
/*
 * Flowtype usage notes:
 *
 * - The '+' prefix indicates properties that should not be re-assigned or
 *   deleted after their initialization.
 *
 * - 'TimeoutID' is an opaque type declared in Flow's core definition file,
 *   used as the return type of the core setTimeout() function.
 */
 ```
 Of course feel free to suggest improvements!

Comment (taoeffect) [reply shared/pubsub.js:32]: That sounds like a great idea, go for it! 👍

File: backend/utils/pubsub.js [hunk 17]
```diff
@@ -0,0 +1,254 @@
...
+const defaultOptions = {
+  clientTracking: true,
+  maxPayload: 6 * 1024 * 1024,
+  pingInterval: undefined
+}
+
+const generateSocketID = (() => {
+  let counter = 0
+
+  return () => counter++
```
taoeffect: Consider a long running program that has a few 100k users. This value can become very big, especially after months or years of operation. That's problematic for a few reasons:

1. It will look quite ugly in the console, and will result in variable length lines
2. In my tests, setting a variable to `Number.MAX_VALUE` causes that number to kind of break. You can no longer increment it or decrement it. Although it's unlikely the counter will ever reach this high, it's still a theoretical ugliness.

OTOH, something like `crypto.randomBytes(6).toString('hex')` is unlikely to ever create a conflict, will always be of the same length in output, and can be called indefinitely without worry. It also gives the impression of a unique ID instead of being a counter representing how many currently active socket connections exist (e.g. the statement, "Socket 54 connected" can lead some to think that there are 54 open socket connections).

shared/pubsub.js
```diff
@@ -0,0 +1,397 @@
...
+    console.debug(`[pubsub] Ignoring ${msg.type} message:`, msg.data)
+  },
+  [RESPONSE_TYPE.ERROR] ({ data: { type, contractID } }) {
+    console.log(`[pubsub] Received ERROR response for ${type} request to ${contractID}`)
+  },
+  [RESPONSE_TYPE.SUCCESS] ({ data: { type, contractID } }) {
+    const client = this
+    switch (type) {
+      case REQUEST_TYPE.SUB: {
+        console.log(`[pubsub] Subscribed to ${contractID}`)
```
taoeffect, line 382: 1. Do the sockets on this side (the frontend) share the same socket ID as the one on the backend? E.g. can you access it through `this` or `client`?
2. If so, can you log that here?
3. If not, can it be added? This will help with debugging and being able to match console logs between the frontend and backend when there are multiple clients/sockets open.

backend/utils/pubsub.js
taoeffect: Eh, it shouldn't be an issue, even with a DDoS attack, let's leave it as is.

shared/pubsub.js
snowteamer: 1. No, these sockets have currently no ID at all. The client doesn't know which ID the server has chosen, so it's currently not accessible.
2. This would be very useful in the backend console when running unit tests. Not so much in the browser console though, since there is only one socket per frontend at a time.
3. It can be added but there is a few different ways:
- By implementing an equivalent to the [primus.id(callback)](https://github.com/primus/primus/#primusidcallback) method, which fetches the relevant socket ID from the server. The downside being the cost of the additional request-response cycle. Besides, Primus ID's are `uuid-v4` ID's, not plain integers starting from zero.
- By letting the front-end generate an ID and send it to server, e.g. as query parameter in the WebSocket URL.

Another idea to ease debugging is to log the current username along or instead the socket ID, or "anonymous" if the connection is not authenticated. But I've noticed that currently, only opaque socket unique IDs or identity contract IDs are ever logged in the backend. If there any specific reason about this?

taoeffect: > Another idea to ease debugging is to log the current username along or instead the socket ID, or "anonymous" if the connection is not authenticated.

Something alone these lines sounds like the most reasonable way to go forward to me. Either logging the username or, more likely, the pubkey hash (which we don't have yet). In other words, something the backend already has to avoid the request/response cycle.

> But I've noticed that currently, only opaque socket unique IDs or identity contract IDs are ever logged in the backend. If there any specific reason about this?

If I understand your question (and I might be misunderstanding), the reason is that we don't have real authentication built-in yet.

shared/pubsub.js
```diff
@@ -0,0 +1,397 @@
...
+        this.getNextRandomReconnectionDelay()
+      )
+    }
+  },
+
+  // Do not manually close the socket here as it will be done automatically.
+  error (event: Event) {
+  },
+
+  message (event: MessageEvent) {
```
taoeffect, line 234: What is this `MessageEvent` type? I don't see where it's defined.

backend/utils/pubsub.js
```diff
@@ -0,0 +1,254 @@
...
+// ====== Server socket event handlers ====== //
+
+const defaultServerHandlers = {
+  /**
+   * Emitted when a connection handshake completes.
+   *
+   * @param socket - The client socket that connected.
+   * @param request - The http GET request sent by the socket.
+   */
+  connection (socket, request) {
```
taoeffect: Should these parameters have their type specified? (Either here in the arguments, or in the comment above?)

backend/utils/pubsub.js
```diff
@@ -0,0 +1,254 @@
...
+
+const generateSocketID = (() => {
+  let counter = 0
+
+  return () => counter++
+})()
+
+// ====== Server socket event handlers ====== //
+
+const defaultServerHandlers = {
```
taoeffect: I'm finding it a challenge to follow along with the code because `this` is bound in every one of the handlers, and depending on which handler I'm looking at, the type of `this` that I'm looking at changes.

Can you add a comment above each of the handler definitions what type `this` will refer to in the handlers that follow?

backend/utils/pubsub.js
snowteamer: This can be done to improve the Flowtype coverage, but is not required to pass the check according to the current .flowconfig.

taoeffect: Right, I'm less concerned about Flowtype coverage, and more interested in simply understanding what these parameters are as I don't have this Websocket API memorized. Maybe in the comment you could include a link to where one could read up about these parameters? Or just a name or something to make it easier to look up the reference to these?

shared/pubsub.js
snowteamer: It's a standard interface defined in the HTML spec: https://html.spec.whatwg.org/#the-messageevent-interface

backend/utils/pubsub.js
```diff
@@ -0,0 +1,254 @@
...
+
+  pong () {
+    this.isAlive = true
+  }
+}
+
+// ====== Server-side message handlers ====== //
+
+const defaultMessageHandlers = {
+  [PUB] (msg: Object) {
```
taoeffect: Should this be a `Message` type?

frontend/controller/backend.js
```diff
@@ -23,70 +24,34 @@ function signJSON (json, keypair) {
...
-              break
-            default:
-              console.error('SOCKET UNHANDLED EVENT!', msg) // TODO: this
-          }
-        }
+export function createGIPubSubClient (url: string, options: Object): Object {
+  return createClient(url, {
+    ...options,
+    messageHandlers: {
+      [RESPONSE_TYPE.ENTRY] (msg) {
```
taoeffect: There is already a `RESPONSE_TYPE` defined in `~/shared/pubsub.js`. We should not have two `RESPONSE_TYPE`'s, that is very confusing.

So, one of those `RESPONSE_TYPE`s should be deleted, and everything affected by that fixed, and then I think `NOTIFICATION_TYPE` should be used here..?

backend/utils/pubsub.js
snowteamer: However to keep it simple it's probably best to annotate them with just `Object` for now. Which is unfortunate for the type checking accuracy, but we don't have a Flowtype definition file for the `ws` library, and the exact type of the request argument is something like `http$IncomingMessage<net$Socket | tls$TLSSocket>`

test/backend.test.js
```diff
@@ -92,7 +94,22 @@ describe('Full walkthrough', function () {
...
+    return new Promise((resolve, reject) => {
+      createGIPubSubClient(process.env.API_URL, {
+        handlers: {
+          error (event) {
+            reject(event)
+          },
+          open (event) {
+            resolve(this)
+          }
+        },
```
taoeffect, line 106: The API is quite complex and confusing, so I'm not sure, but is this missing a `NOTIFICATION_TYPE.ENTRY` handler?

shared/pubsub.js
```diff
@@ -0,0 +1,397 @@
...
+    minReconnectionDelay * reconnectionDelayGrowFactor ** (this.failedAttempts + 1),
+    maxReconnectionDelay
+  )
+  return minDelay + Math.random() * (maxDelay - minDelay)
+}
+
+function pub (contractID: string, data: JSONType) {
+  const { socket } = this
+
+  if (socket && socket.readyState === WebSocket.OPEN) {
```
taoeffect: Is it possible for the `readyState` here to not be `.OPEN`, and if so, should an error handler be added here (and elsewhere)?

taoeffect (changes requested):
A few more comments (sorry it's taking me so long to go through this PR, but I guess that's to be expected for a PR of this size).

backend/utils/pubsub.js
taoeffect: Yeah, like I said, the flowtype annotation is less important to me than maybe a link to some place where more can be read up about these parameters so that people who are unfamiliar with the API have some place to go to if they're interested in knowing what these are.

frontend/controller/backend.js
snowteamer: Good catch!

shared/pubsub.js
snowteamer: I don't know, this is just a port of the unused `pub()` implementation which was there.

A comment still mentions that `PUB` can be used for ephemeral notifications outside any contract log, but It didn't appear to be used anywhere.

test/backend.test.js
snowteamer: I'm sorry if the API isn't simple enough. Is there something in particular that should better be removed?

Regarding these functions, they are `ws` socket event handlers; perhaps it was more obvious when using `onerror` and `onopen` as keys?

On the other hand, `NOTIFICATION_TYPE.ENTRY` denotes a pubsub message type.
Such a type constant can be used as a key in `messageHandlers` to define a processing routine ("message handler") that should be run whenever an `ENTRY` message is received.

It's still possible to not use the `handlers` options and just bind any event handler manually upon object instantiation.
Similarly, it's also possible to not use the `messageHandlers` option and just put any message processing logic in a rather large switch statement inside a `message` socket event handler.

taoeffect: Ah, my apologies, please ignore my comment. I understand now better how this works and why `ENTRY` doesn't need to be specified here. It's because it is already specified as part of the implementation of `createGIPubSubClient`. That was just me getting confused and not looking closely enough. Resolving this.

backend/utils/pubsub.js
snowteamer: I understand that a plein integer counter is not the most robust way to assign an ID to something. But it's very nice in development and testing. We can use real uuid's later in production.

taoeffect: Using a counter is fine. I don't think we'll need to use UUIDs. If we need to associate sockets between backend and frontend, that might be useful, and we can use the approach we discussed on Slack (that of creating a `socketDescription` function that logs some data that is shared between frontend and backend, e.g. a username or a key hash, and only if logging such things is enabled or we're in development mode).

backend/utils/pubsub.js
snowteamer: > (Speaking of, why aren't we just using Node's core EventEmitter? It should bundle fine.)

This indeed could make the API more convenient, but I would rather avoid adding any new dependency to the frontend, given that one of the main goals of this PR is to drop dependencies.

joepie91: > given that one of the main goals of this PR is to drop dependencies.

I don't agree that this is a useful goal, stated as such. From what I recall, the problem with Primus was specifically that it is unreliable / unpleasant to work with, and difficult to support outside of JS, and those are definitely valid reasons to remove or avoid a dependency. It's the same reason I usually recommend people to rip out Socket.io before doing anything else.

But avoiding dependencies for the sake of being able to say that you're not using dependencies, while then reinventing them yourself, is not a net positive thing - at that point you're just moving from well-tested complexity to untested homegrown complexity. The complexity is *still there*, but now you're no longer benefiting from the shared work and experience of the broader community.

Dependencies are not a bad thing at all, you only need to make sure that you don't add dependencies which have higher costs than benefits (which is almost always the case for monolithic/kitchensink dependencies in particular, and rarely for modular dependencies). But Node's core EventEmitter is IMO more than reliable enough to depend on it, and it has a well-defined scope.

taoeffect: On the topic of dependencies the goal of this PR: I agree with you @joepie91 that it doesn't make sense to re-invent the wheel for dependencies that work perfectly fine, which is why the issue title says to "adopt/fork rest". That makes it so that we can benefit from the collective labor without surprise updates that introduce vulnerabilities. With respect to Primus, the problems were:

- It required a very weird/ugly API to generate a client-side file that would randomly change periodically, making it difficult to be sure we weren't introducing vulnerabilities
- The size of this client-side file was monstrous, as it included dependencies for several other projects. By implementing our own solution we wouldn't have to duplicate existing dependencies, allowing us to keep the bundle size much smaller, speeding up loading speeds on slow connections
- Some of the Primus plugins were were using had outdated dependencies of their own with vulnerabilities that `npm` would complain about

And probably some others.

With respect to the EventEmitter API, I have no opinion, and will leave that up to your judgements.

backend/utils/pubsub.js
```diff
@@ -0,0 +1,253 @@
+/* globals logger */
```
taoeffect: One issue with file path: this is the only file under a `utils` folder, and it isn't a utility. Can you move the file under `backup/` with the rest of the core server files and delete the `utils` folder since there won't be anything else under it?

backend/utils/pubsub.js
joepie91: > which is why the issue title says to "adopt/fork rest". That makes it so that we can benefit from the collective labor without surprise updates that introduce vulnerabilities.

For what it's worth, forking isn't necessary here; a lockfile will already pin a dependency to a specific hash, and prevent automatic upgrades.

taoeffect: Yeah, hopefully it works as advertised.

snowteamer: By the way, if we wanted to keep Primus for its convenient API and other benefits, some of the associated problems could likely be mitigated by taking a few measures:
- The client-side Primus library file could be git-ignored so as not to disrupt git diffs.
- The Primus-related Grunt task could minimize that file to reduce the impact on the bundle size.
- It could also try to generate it only when it's outdated, not on every build.
- The `debug` dependency vulnerability could be solved by patching it ourselves or forcing Primus to use an up-to-date version.

It must be noted that all these measures would still require a non-trivial amount of work.

In my opinion, the best argument in favor of dropping Primus is the difficulty to make it work in other runtimes like Deno. Sure there is still a compatibility issue even when only using the `ws` library, but much smaller.

snowteamer: @joepie91
> In the createCustomThing invocation, it's not possible to use shorthand method syntax, because we're not in an object context.

I'm not sure to understand correctly, because shorthand method syntax seems to work just fine e.g. in `createGIPubSubClient()` implemented in `~/frontend/controller/backend.js`:
```js
export function createGIPubSubClient (url: string, options: Object): Object {
  return createClient(url, {
    ...options,
    messageHandlers: {
      [NOTIFICATION_TYPE.ENTRY] (msg) {
        // We MUST use 'state/enqueueHandleEvent' here to ensure handleEvent()
        // is called AFTER any currently-running calls to syncContractWithServer().
        // Calling via SBP also makes it simple to implement 'test/backend.js'
        sbp('state/enqueueHandleEvent', GIMessage.deserialize(msg.data))
      }
    }
  })
}
```
Granted, it was not even necessary to use the `this` binding in the single message handler defined here, but I think it would have nonetheless be correctly bound to the client instance if used, namely because of the explicit binding via `handler.call(client)`.

shared/pubsub.js
snowteamer: After investigation, I think passing the username to the server would not work as the same connection seems to be reused when a user logs out then logs in again as another user without reloading the page, which happens a lot in cypress tests.

That behavior should be fixed, but for now I guess it's ok to use a dummy devmode-only random socket ID with an explicit comment.

backend/server.js
```diff
@@ -45,10 +44,13 @@ sbp('sbp/selectors/register', {
...
+        pubsub.subscribersByContractID[contractID].forEach((client) => {
+          client.send(createMessage('entry', entry.serialize()))
+        })
+      }
```
taoeffect: I feel like in most of your code, we have taken several steps forward, without retrace. In this particular section however, it feels to me like a step backward. We went from a single elegant line "to this room, send this message", to complicated logic.

Can you look into reproducing the original API we had with the rooms? Or something similarly elegant so that we can keep this a flexible 1-liner?

frontend/controller/backend.js
```diff
@@ -23,70 +23,34 @@ function signJSON (json, keypair) {
...
-      const res = await serverSocket.unsub(contractID)
-      delete contractSubscriptions[contractID]
-      console.debug(`[Backend] unsubscribed ${contractID}`, res)
+      client.unsub(contractID)
     }
     for (const contractID of toSubscribe) {
-      const res = await serverSocket.sub(contractID)
-      contractSubscriptions[contractID] = true
-      console.debug(`[Backend] subscribed ${contractID}`, res)
+      client.sub(contractID)
```
taoeffect, line 53: I'm a bit surprised... are neither Flow nor Standard giving you an error here?

It is my understanding that ever function labeled `async` must have at least one corresponding `await` within it. And yet I see no `await` any longer inside of this function? So if `sub` and `unsub` do not return promises, and therefore no `await` needs to be called, then `async` should be removed from the function identifier on line 41.

shared/pubsub.js
```diff
@@ -198,7 +198,11 @@ const defaultSocketEventHandlers = {
       default: break
     }
     if (this.shouldReconnect && this.options.reconnectOnDisconnection) {
-      this.scheduleConnectionAttempt()
+      if (this.failedConnectionAttempts <= this.options.maxRetries) {
+        this.scheduleConnectionAttempt()
+      } else {
+        this.destroy()
+      }
```
taoeffect: So, does this mean that there's a situation where the user must refresh the page in order for the connection to re-establish itself?

What if, for example, their internet goes down, but it comes back up later. Does your code handle instantly re-connecting when the `online` event is fired?

backend/server.js
snowteamer: Agreed this code is a bit low-level, so I am planning to add something like a `broadcast()` method to implement this logic.

Though one single callsite was perhaps not enough to justify the addition of a new method with its own edge cases etc., I have since realized that it was necessary to hide `subscribersByContractID` which should not be accessed outside of the file where it is defined.

taoeffect: Sounds good 👍

frontend/controller/backend.js
snowteamer: No error for now, likely because our Flow and Standard setups are configured to be quite lenient.

While every function using `await` must be labeled `async`, the opposite is not strictly required, although it's arguably less confusing.

According to the [require-await](https://eslint.org/docs/rules/require-await) rule (which could be enabled to report such cases):
> Asynchronous functions that don't use await might not need to be asynchronous functions and could be the unintentional result of refactoring.

So it's probably best to avoid, however removing the `async` keyword will change the return type of the function.

shared/pubsub.js
snowteamer: > does this mean that there's a situation where the user must refresh the page in order for the connection to re-establish itself?

Yes that's the point. If we really do not want that, `maxRetries` can be set to `Infinity`, but I would be careful about it.
Similarly I guess the connection should be closed without autoreconnection in case of e.g. unrecoverable errors or malicious users being autobanned.

> Does your code handle instantly re-connecting when the online event is fired?

It is supposed to do so if possible, unless there is a good reason not to.

taoeffect: > It is supposed to do so if possible, unless there is a good reason not to.

Forgive me, this is a complicated PR and I haven't wrapped my mind over ever piece of it yet, so my question for you is: will it be able to still do that if you've called `this.destroy()` here?

snowteamer: Once the pubsub instance has been destroyed, it should not reconnect again.

However we can still listen to network events on the `window` object, then maybe use the `BannerGeneral` component to notify the user once they are back online and tell them to reload the page to resync the data.

This might sound inconvenient, but for now it is likely necessary anyway in order to get contract updates and pubsub messages that were missed while being offline.
Well, unless we make resynchronization work on the `online` event like it currently works on page reload, but this is likely out of the scope of this PR, as this issue likely already existed with Primus.

Also, this `destroy()` call will happen after `maxRetries` reconnection attempts, which already amounts to ~6 minutes with `maxRetries: 10`, and to ~15 minutes with `maxRetries: 20`.

This is a lot of time compared to other socket apps like Slack or Discord which completely hide their UI after roughly one minute of offline time, likely to protect their users from the stale data.

snowteamer: Also, this PR is not really supposed to add new functionality: all these reconnection capabilities were already available with Primus, albeit untested in the context of the app.

backend/server.js
```diff
@@ -41,17 +41,16 @@ const hapi = new Hapi.Server({
...
     await sbp('gi.db/log/addEntry', entry)
-    console.log(chalk.blue.bold(`broadcasting ${entry.description()}`))
-    sbp('okTurtles.data/apply', PUBSUB_INSTANCE, pubsub => {
-      if (contractID in pubsub.subscribersByContractID) {
-        pubsub.subscribersByContractID[contractID].forEach((client) => {
-          client.send(createMessage('entry', entry.serialize()))
-        })
-      }
-    })
+    await sbp('backend/server/broadcastEntry', entry)
```
taoeffect, line 53: Lovely choice of an SBP selector 👍

taoeffect: Oh wait nm this was removed later on with a function call. Oh well :)

backend/pubsub.js
```diff
@@ -1,127 +1,285 @@
...
+        client.send(message)
+      }
+    }
+  },
+  // Enumerates the subscribers of a given contract.
+  * enumerateSubscribers (contractID: string): Iterable<Object> {
+    if (contractID in this.subscribersByContractID) {
+      yield * this.subscribersByContractID[contractID]
+    }
+  }
```
taoeffect: Neat API 👍

backend/server.js
snowteamer: I've added it back! 😄 

However I'm not sure whether the `console.log()` call should take place on the first or the fourth line.

taoeffect: @snowteamer well, the only reason to have it is if you'd like to explicitly call `entry.description()` in the logging. The advantage of the SBP selector is that logging happens automatically, but it it won't result in a call to `entry.description()`. Is it possible for JS objects to specify how they are logged to the console? If so, then we could have it so that `this.description()` is called whenever they're logged

backend/server.js
```diff
@@ -42,14 +41,16 @@ const hapi = new Hapi.Server({
 sbp('okTurtles.data/set', SERVER_INSTANCE, hapi)
 
 sbp('sbp/selectors/register', {
+  'backend/server/broadcastEntry': async function (entry: GIMessage) {
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+    const pubsubMessage = createMessage('entry', entry.serialize())
```
taoeffect: Should this `'entry'` be a constant that's referenced from somewhere instead of a string literal?

snowteamer: The string literal was less code overall, but we can use an imported constant for consistency.

backend/server.js
snowteamer: > Is it possible for JS objects to specify how they are logged to the console?

Yes it is possible in the server console by defining a [`Symbol.for('nodejs.util.inspect.custom')`](https://nodejs.org/api/util.html#util_util_inspect_custom) method on said JS objects.
The colors and formatting can also be specified. We could possibly use it to improve logging for many objects.

taoeffect: Awesome, great find! Go for it! 😄

snowteamer: However, the nested SBP calls cause redundant logging. I've tried to mitigate this by adding the `'backend/server/broadcastEntry'` selector to the `dontLog` object.

taoeffect: Ah, yeah didn't realize that would be a problem, but it makes sense. Good call on adding it to the `dontLog` object. 👍

frontend/controller/backend.js
taoeffect: @snowteamer Feel free to both enable the `require-await` rule, and also remove the `async` keyword, because the return type here doesn't matter. This is just an event handler for an event. The return value is ignored (you can confirm that per `okTurtles.events` source).

shared/pubsub.js
taoeffect: OK, in that case can you add a comment here to make sure we don't forget to deal with this, since at some point in the future we probably will? I.e. ideally the app shouldn't require a refresh if there isn't a connection for 15 minutes.

So something like:

```
this.destroy() // TODO: handle app reconnecting when connection returns
```

snowteamer: If the app shouldn't require a refresh after only 15 minutes then we can increase that number to whatever appropriate value, even 1 hour for example. However after calling `destroy()` the web socket client won't check for connectivity anymore, which is why this should only be called after `maxRetries` attempts or when an irrecoverable error has occured, since it basically means to give up the reconnection process.

frontend/controller/backend.js
snowteamer: Enabling this rule globally results in four other similar errors being found in files that are unrelated to this PR, so maybe it would be better to just remove the `async` keyword here for now?

shared/pubsub.js
taoeffect: Ah, I see. OK that makes sense.

frontend/controller/backend.js
taoeffect: Can you check if fixing those other errors breaks anything? If not, feel free to remove `async` in those other locations too and enable the rule. It seems (to me) like a useful rule that used to be enabled, and maybe got disabled by a change in Standard.

snowteamer: Sure. In this PR or in a quick new one?

taoeffect: I guess we can do it in a separate PR. Just opened #1058 for it.

shared/pubsub.js
```diff
@@ -426,15 +468,30 @@ const publicMethods = {
...
+      throw new Error(`emit(): argument 'type' must not be '${type}'.`)
+    }
+    // This event object partially implements the `CustomEvent` interface.
+    const event = { type, detail }
+    const listener = this.listeners[type]
+    if (listener) {
+      listener(event)
+    }
+  },
+
```
taoeffect: I can't tell for sure, but this code seems like it only supports one listener per event?

Is there a specific reason you chose to re-implement this different type of event system, or do you think you can use the existing `okTurtes.events` domain instead (as is used throughout the app)? I feel like that would be cleaner and avoid re-inventing the wheel. `okTurtles.events` also handles multiple listeners per event.

snowteamer: Yes there is a single listener per event, but that listener can actually wrap several handler functions.

However, to keep the API simple for now it only allows one default "hard-coded" handler and one optional custom (e.g. "user-defined") handler though, whereas traditional API's allow to dynamically add and remove an unlimited number of custom handlers.

taoeffect: So is there an issue with using `okTurtles.events`?

backend/pubsub.js
```diff
@@ -1,127 +1,295 @@
...
-        })
-      })
-
-      spark.on('data', function (data) {
-        console.log(bold.red('[pubsub] received UNHANDLED DATA from client:', data))
-      })
     })
+  },
+  error (error: Error) {
+    console.log('[pubsub] Server event: error', error)
```
taoeffect: This should be `console.error`, with a clearer message, something like:

```
console.error('[pubsub] Server Error:', error)
```

backend/pubsub.js
```diff
@@ -1,127 +1,295 @@
...
+  }
+}
 
-    primus.on('disconnection', function (spark) {
-      // the spark that disconnected
-      console.log(bold.yellow(`[pubsub] ${spark.id} disconnection`))
-    })
-  })
+// Default handlers for server-side client socket events.
+// The `this` binding refers to the connected `ws` socket object.
```
taoeffect, line 180: Can you add a link to the corresponding MDN page that represents the web socket object here in the comments?

backend/pubsub.js
```diff
@@ -1,127 +1,295 @@
...
-    primus.on('joinroom', function (room, spark) {
-      console.log(bold.yellow(spark.id + ' joined ' + room))
-    })
+const defaultOptions = {
+  maxPayload: 6 * 1024 * 1024,
+  pingInterval: 30_000
+}
+
+// Default handlers for server events.
+// The `this` binding refers to the server object.
```
taoeffect, line 122: Can you add a link to the corresponding MDN page that represents the server object class here in the comments?

backend/pubsub.js
```diff
@@ -1,127 +1,295 @@
...
+    this.activeSinceLastPing = true
+    const handler = this.server.messageHandlers[msg.type]
+
+    if (handler) {
+      try {
+        handler.call(this, msg)
+      } catch (error) {
+        // Log the error message and stack trace but do not send it to the client.
+        logger(error)
+        // Should we call 'this.terminate()' instead?
```
taoeffect: What are the tradeoffs?

backend/server.js
```diff
@@ -42,14 +41,16 @@ const hapi = new Hapi.Server({
 sbp('okTurtles.data/set', SERVER_INSTANCE, hapi)
 
 sbp('sbp/selectors/register', {
+  'backend/server/broadcastEntry': async function (entry: GIMessage) {
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+    const pubsubMessage = createMessage(NOTIFICATION_TYPE.ENTRY, entry.serialize())
+    const subscribers = pubsub.enumerateSubscribers(entry.contractID())
+    console.log(chalk.blue.bold(`[pubsub] Broadcasting ${entry.description()}`))
```
taoeffect, line 48: Didn't you mention that there was a way for the `GIMessage` object to get logged with `entry.description()` automatically when being logged?

If so, wouldn't it be possible to remove this `console.log` line and then remove `'backend/server/broadcastEntry'`, resulting in the same number of log lines as before, but automated? Or was there a reason you did it this way instead?

frontend/main.js
```diff
@@ -57,12 +58,25 @@ async function startApp () {
...
+
+  if (process.env.NODE_ENV === 'development') {
+    // This is temporarily used in development mode to help the server improve
+    // its console output until we have a better solution. Do not use for auth.
+    const debugID = Array.from(
+      // $FlowIssue doesn't support crypto yet: https://github.com/facebook/flow/issues/5019
+      crypto.getRandomValues(new Uint8Array(3)),
+      (byte) => byte.toString(16).padStart(2, '0')
+    ).join('')
+    pubsubURL += `?debugID=${debugID}`
```
taoeffect: I think these 6 lines can be reduced to 1:

```js
    pubsubURL += `?debugID=${crypto.randomBytes(6).toString('hex')}`
```

taoeffect (changes requested):
@snowteamer OK this is my final review I think. Only minor things remaining, posted comments, plus a request to use `okTurtles.events` where possible unless there's a reason not to. You're already using it in some parts of the code, I'm not sure why we need to re-implement event emission, and if there isn't a strong reason to, let's not do that.

shared/pubsub.js
snowteamer: Say we want to use `okTurtles.events` to make the general banner be cleared and dissappear upon successful reconnection.

For now this can be done like this, from anywhere in the frontend code:

```js
sbp('okTurtles.events/on', 'reconnection-succeeded', (event) => {
  sbp('okTurtles.data/get', 'BANNER').clean()
})

sbp('okTurtles.data/apply', PUBSUB_INSTANCE, (client) => {
  client.customEventHandlers['reconnection-succeeded'] = (event) => {
    sbp('okTurtles.events/emit', 'reconnection-succeeded', event)
  }
})
```

The `reconnection-succeeded` event type name can also be made a constant for consistency.

backend/pubsub.js
snowteamer: It's the web socket object from the `ws` library, not the native WebSocket object which only exists in the frontend (and in Deno).
But yes a link to `ws` apidocs can be added somewhere.

backend/pubsub.js
snowteamer: Ok, in that case the `listening` handler should also be modified accordingly I guess.

backend/pubsub.js
snowteamer: Same as above.

backend/pubsub.js
snowteamer: Although we are also currently using many other dependencies without including links to their documentation, for example `chalk` in this module.

frontend/main.js
snowteamer: Awesome!

snowteamer: But I was using only three bytes, is it ok? That is still enough for 16 million IDs.

backend/server.js
snowteamer: Well I noticed than the default console output format was also being used in many places, likely providing valuable debug information in the build output, for example:
```js
 GIMessage {
    _decrypted: undefined,
    _mapping: {
      key: '21XWnNHXUB2pLXf3Y1wZ2aJny2sZGz4Hd4zrACUuA77tZqn9tC',
      value: '{"message":"{\\"version\\":1,\\"previousHEAD\\":\\"21XWnNUqk6LEdCAQ1jQTY5Tro1YPMx5DxAmXewwK295JT9qqo5\\",\\"contractID\\":\\"21XWnNUqk6LEdCAQ1jQTY5Tro1YPMx5DxAmXewwK295JT9qqo5\\",\\"op\\":[\\"ae\\",\\"{\\\\\\"type\\\\\\":\\\\\\"gi.contracts/identity/process\\\\\\",\\\\\\"data\\\\\\":{\\\\\\"attributes\\\\\\":{\\\\\\"username\\\\\\":\\\\\\"Bob-568\\\\\\",\\\\\\"email\\\\\\":\\\\\\"bob@okturtles.com\\\\\\"}},\\\\\\"meta\\\\\\":{}}\\"],\\"nonce\\":0.34964236633868184}","sig":{"type":"default","sig":"21XWnNN5w9SZ7kGvZBH3e2XwUSXXWEVGfkgbSSLWcZNAvdrj7i"}}'
    },
    _message: {
      version: 1,
      previousHEAD: '21XWnNUqk6LEdCAQ1jQTY5Tro1YPMx5DxAmXewwK295JT9qqo5',
      contractID: '21XWnNUqk6LEdCAQ1jQTY5Tro1YPMx5DxAmXewwK295JT9qqo5',
      op: [Array],
      nonce: 0.34964236633868184
    }
  }
  ```

Now if `entry.description()` is made the default logging format like I mentioned it was possible to do, then it will be automatically used everywhere by default, and that debug information will be lost.

Unless we implement and use another logging function for callsites where more debug information should be displayed.

File: backend/pubsub.js (hunk 37 above)
Comment (snowteamer) [reply backend/pubsub.js]: According to the author in https://github.com/websockets/ws/issues/1596:
> ws.terminate() immediately destroys the connection, ws.close() waits for close timer to expire before destroying the connection if the other peer is not responsive.

So in my understanding, `terminate()` call `close()` internally, but doesn't wait before removing event listeners and releasing memory. Some investigation is needed to see how this affects automatic reconnection of the client.

File: shared/pubsub.js (hunk 33 above)
Comment (snowteamer) [reply shared/pubsub.js]: After thinking a bit more about it, it looks like a tighter sbp events integration could also be achieved by deleting my `emit()` method and using `okTurtles.events/emit()` instead.

File: backend/server.js (hunk 38 above)
Comment (snowteamer) [reply backend/server.js:48]: It would be interesting to identify a clean minimal subset of debug info that should be included so that GIMessage console output can be improved.

File: frontend/main.js (hunk 39 above)
Comment (taoeffect) [reply frontend/main.js]: @snowteamer sure thing!

File: shared/pubsub.js (hunk 33 above)
Comment (taoeffect) [reply shared/pubsub.js]: > After thinking a bit more about it, it looks like a tighter sbp events integration could also be achieved by deleting my emit() method and using okTurtles.events/emit() instead.

Go for it 👍

File: backend/pubsub.js (hunk 35 above)
Comment (taoeffect) [reply backend/pubsub.js:180]: > But yes a link to ws apidocs can be added somewhere.

Yeah, see, you made it clear that I was confused what this was referring to, so I bet other developers might get confused too. A link would be super helpful then.

> Although we are also currently using many other dependencies without including links to their documentation, for example chalk in this module.

True... it's just that when `chalk` is used, there's a clear variable that is bound from an import, and you can trace it directly, whereas with these handlers it gets a bit more confusing because it's not clear where the `this` is bound just by reading the code. It's not easy to trace, so hence the documentation links on each one of them would be greatly helpful.

File: backend/pubsub.js (hunk 37 above)
Comment (taoeffect) [reply backend/pubsub.js]: > So in my understanding, terminate() call close()

Based on the text you quoted, wouldn't it be the other way around (`close()` calling `terminate()` internally)?

File: backend/server.js (hunk 38 above)
Comment (taoeffect) [reply backend/server.js:48]: Hmm. OK, you make a fair point. Resolving this, let's leave it as-is.

File: backend/pubsub.js (hunk 37 above)
Comment (snowteamer) [reply backend/pubsub.js]: `close()` does not need to remove all event listeners and clear references like terminate() does, so it should not call it I guess.

Comment (taoeffect) [reply backend/pubsub.js]: Ah, I see

File: shared/pubsub.js (hunk 33 above)
Comment (snowteamer) [reply shared/pubsub.js]: While it is easy to implement in the browser, it becomes more complicated in the backend unit tests, where there are commonly several pubsub client instances running at once and pubsub events are not global but are only scoped to their respective client instance.

Therefore the instance must be passed somehow when emitting pubsub events, so what would be the best way to do that in SBP?
- Passing the instance as an argument. This requires to define and remember which argument is used, e.g. it could the first positional argument, or its `target` property (assuming the argument is an object).
- Passing an instance identifier in the selector, e.g. `okTurtles.events/pubsub-5d498f/emit`

Comment (taoeffect) [reply shared/pubsub.js]: @snowteamer I'm not fully sure I understand the issue, maybe you can link me (either here or on Slack) to the part in the tests where this is happening?

FYI the implementation of `emit` allows for the passing of various variables to the listeners, so you could even pass in `this` and have the listener take in parameters. Copying the implementation of `emit` here for reference:

```js
  'okTurtles.events/emit': function (event: string, ...data: any) {
    for (const listener of sbp('okTurtles.data/get', listenKey(event)) || []) {
      listener(...data)
    }
  },
```

Not sure if that helps you or not.

shared/pubsub.js
```diff
@@ -0,0 +1,579 @@
...
+ * {boolean?} debug
+ * {object?} handlers - Custom handlers for WebSocket events.
+ * {boolean?} manual - Whether the factory should call 'connect()' automatically.
+ *   Also named 'autoConnect' or 'startClosed' in other libraries.
+ * {object?} messageHandlers - Custom handlers for different message types.
+ * {number?} pingTimeout - How long to wait for the server to send a ping, in milliseconds.
+ * {boolean?} reconnectOnDisconnection - Whether to reconnect after a server-side disconnection.
+ * {boolean?} reconnectOnOnline - Whether to reconnect after coming back online.
+ * {boolean?} reconnectOnTimeout - Whether to reconnect after a connection timeout.
+ * {number?} timeout - Connection timeout duration in milliseconds.
```
taoeffect, line 103: I don't see what the default `timeout` is, it would be very helpful if you could add that here to the comments (along with the defaults for the other values as you feel is relevant)

frontend/main.js
```diff
@@ -57,12 +58,25 @@ async function startApp () {
...
+  }
+  sbp('okTurtles.data/set', PUBSUB_INSTANCE, createGIPubSubClient(
+    pubsubURL, {
+      // This option can be enabled since we are not doing auth via web sockets.
+      reconnectOnTimeout: true
```
taoeffect: Here we lost the `timeout` value of 3000, could that be added back in please (especially since the default appears to be 5000)?

snowteamer: That was on purpose, 3000 is likely too short:
- We have seen that timing issues can cause the Travis build to fail.
- All web socket libraries that I have checked use a higher default timeout or do not seem to specify any at all:
  - [reconnecting-websocket](https://github.com/pladaria/reconnecting-websocket#default-values): 4000
  - [Primus](https://github.com/primus/primus/#connecting-from-the-browser): 10000
  - [socket.io](https://socket.io/docs/v4/client-initialization/#timeout): 20000

taoeffect: ok, it's still useful to specify it here though to remind us that's an option and what its current value is without diving into the implementation. To your first point, we can also extend it for travis:

```js
```

taoeffect: ok, it's still useful to specify it here though to remind us that's an option and what its current value is without diving into the implementation. To your first point, we can also extend it for travis, e.g. something like this:

```js
timeout: process.env.NODE_ENV === 'development' ? 10000 : 5000 // extend for Travis
```

snowteamer:
Unfortunately `crypto.randomBytes()` doesn't seem to exist in the browser, so I will likely have to revert to`crypto.getRandomValues()` :(

taoeffect:
> Unfortunately crypto.randomBytes() doesn't seem to exist in the browser, so I will likely have to revert tocrypto.getRandomValues() :(

Ok, I replied with a suggestion on Slack

shared/pubsub.js
```diff
@@ -0,0 +1,608 @@
...
+    client.pendingUnsubscriptionSet.add(contractID)
+  }
+}
+
+// Register custom SBP event listeners before the first connection.
+for (const name of Object.keys(defaultClientEventHandlers)) {
+  if (name === 'error' || !socketEventNames.includes(name)) {
+    sbp('okTurtles.events/on', `pubsub-${name}`, (target, detail?: Object) => {
+      target.listeners[name]({ type: name, target, detail })
+    })
```
taoeffect, line 598: This looks kinda odd to me, not sure what's going on here? This appears to be re-implementing an event system ... using an event system.

taoeffect: Copying my reply from Slack:

```
 FYI, I think it might update #dev, but it seems like sometimes it drops the messages. Anyway, this might be a better place to discuss it because it's faster and when there's a lot of back & forth Slack is better suited to it
5:45
I got a bit distracted by something but now am looking closer at it. I now see clearer what's going on... it's really kind of a conflict between two different approaches / coding styles... I guess what you have here is OK. It just looks weird because the okTurtles.events API is meant to be a more global events system, and here you have handlers for a specific instance of a pubsub client, and only a single handler per instance. A very different sort of approach
5:46
but it makes sense
5:47
In fact, we can keep this weirdness around to remind us to in the future maybe convert the API to a more SBP-style approach (as a possibility / experiment, not a necessity)
```

Feel free to continue the conversation there. I'll resolve this here in the meantime.

taoeffect (approved):
Approved! Excellent work @snowteamer! 😄 🥇 🎉

PR #1056, opened 2021-03-16
Task/763 payments algorithms

CrazedCoding:
I am preliminarily creating this PR so people can start commenting as changes are being made, in real-time.

Here is the start of the mocha tests that I am going to create for the event-based distribution approaches:

```
// Please note: "income"s are the user pledges if positive, OR the needed income amount ABOVE mincome if NEGATIVE.
const distributionEvents = []

// User1 creates a group at the beginning of the month of January, 2021, and then sets an income of 100 at 10% of the way throgh the monthly cycle.
distributionEvents.push(
  {
    "type": "incomeDeclaredEvent",
    "data": {
      "name": "u1",
      "income": 100,
      "cycle": 0.1
    }
  })

// User2 joins and sets income of -50 at 20% of the way through the monthly cycle.
distributionEvents.push(
  {
    "type": "incomeDeclaredEvent",
    "data": {
      "name": "u2",
      "income": -50,
      "cycle": 0.2
    }
  })

// User3 joins and sets income of -50 at 30% of the way through the monthly cycle.
distributionEvents.push(
  {
    "type": "incomeDeclaredEvent",
    "data": {
      "name": "u3",
      "income": -50,
      "cycle": 0.3
    }
  })

// User1 pays User2 10 (out of <???>) at 40% of the way through the monthly cycle.
distributionEvents.push(
  {
    "type": "paymentEvent",
    "data": {
      "from": "u1",
      "to": "u2",
      "amount": 10,
      "cycle": 0.4
    }
  })

// User4 joins and sets income of -100 at 50% of the way through the monthly cycle.
distributionEvents.push(
  {
    "type": "incomeDeclaredEvent",
    "data": {
      "name": "u4",
      "income": -100,
      "cycle": 0.5
    }
  })

// User3 removes themselves from the group at 60% of the way through the monthly cycle. TODO: This causes UI glitches/errors for other members...
distributionEvents.push(
  {
    "type": "leaveGroup",
    "data": {
      "name": "u3"
      "cycle": 0.6
    }
  })

// User1 pays User4 5 (out of <???>) at 70% of the way through the monthly cycle.
distributionEvents.push({
    "type": "paymentEvent",
    "data": {
      "from": "u1",
      "to": "u4",
      "amount": 5,
      "cycle": 0.7
    }
  })

// User3 RE-joins the group at 75% of the way through the monthly cycle, and sets their income to -100 at 80% of the way through the monthly cycle.
distributionEvents.push({
    "type": "incomeDeclaredEvent",
    "data": {
      "name": "u3",
      "income": -100,
      "cycle": 0.8
    }
  })
```

These scenarios are mathematically/algorithmically designed to throw a monkey-wrench into any so-called turing-complete-solution anyone comes up with. If these can't break your solution, nothing can, BUT:

PLEASE don't just answer these scenarios with: "SOME should go to u2, and MORE/LESS should go to u3"

Terms like SOME, MORE, and LESS do not help us approach an algorithmic/arithmatic/mathematical solution.

Until we all agree on the distribution outcome (for any point in time) of these scenarios for the group-consolidated-approach, I will continue to try to create a "base-line" pro-rating/time-weighted solution (so we at least have some sort of reference point).


frontend/utils/distribution/payments-minimizer.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+// TODO: this algorithm will be responsible for "balancing" payments
+// such that the least number of payments are made.
+function minimizeTotalPaymentsCount (distribution: Array<Object>): Array<any | {|from: string, to: string, amount: number|}> {
```
taoeffect, line 5: Quick comment: AFAIK the use of `any |` negates anything that follows it, rendering moot any kind of type checking. In other words, AFAIK, `any | ...` and `any` are the same thing.

frontend/model/contracts/group.js
```diff
@@ -672,6 +721,8 @@ DefineContract({
               Vue.set(groupProfile, key, value)
           }
         }
+        const income = data.incomeDetailsType === 'incomeAmount' ? data.incomeAmount - state.settings.mincomeAmount : data.pledgeAmount
```
taoeffect: Why are you subtracting the mincome from their income? Did you mean to add the mincome to the pledge?

CrazedCoding:
VERY good question!

I *kind of* (but not entirely) answer this at the very beginning of the
as-of-incomplete mocha tests:

```
// Please note: "income"s are the user pledges if positive, OR the needed
income amount ABOVE mincome if NEGATIVE.
```

(Negative values above a value are below).

From a CS/math perspective: this approach uses the 'sign-bit' of our
floating-point values to preserve the information about whether they are
pledging (positive) or needing (negative).

These positive/negative values are how above/below the mincome a
haver/needer is.

I should mention: this calculation is entirely reversible, and will NOT
affect the heuristics of our group-consolidated approach, however:

I assure you: this encoding is compatible with both the strictly-pro-rated
solution, as well as the group-consolidated version (because it is
reversible).

Also:instead of passing the minCome as a time-dependant-parameter to our
algorithm, it needs to be modeled as an event (including upon group
creation). This is because the minCome can change at any time during a
monthly cycle.

I should also mention: this above/below bit-encoding-of-the-sign-bit as an
indication of having/needing is a feature for making the pro-weighted
approach less computationally complex, and more immediately obvious as a
heuristic approach.

frontend/model/contracts/group.js
CrazedCoding: Apparently answering by email puts the github replies into the wrong discussion thread. As replied (in the wrong thread), here is my answer:

VERY good question!

I *kind of* (but not entirely) answer this at the very beginning of the
as-of-incomplete mocha tests:

```
// Please note: "income"s are the user pledges if positive, OR the needed
income amount ABOVE mincome if NEGATIVE.
```

(Negative values above a value are below).

From a CS/math perspective: this approach uses the 'sign-bit' of our
floating-point values to preserve the information about whether they are
pledging (positive) or needing (negative).

These positive/negative values are how above/below the mincome a
haver/needer is.

I should mention: this calculation is entirely reversible, and will NOT
affect the heuristics of our group-consolidated approach.

I assure you: this encoding is compatible with both the strictly-pro-rated
solution, as well as the group-consolidated version (because it is
reversible).

Also:instead of passing the minCome as a time-dependant-parameter to our
algorithm, it needs to be modeled as an event (including upon group
creation). This is because the minCome can change at any time during a
monthly cycle.

I should also mention: this above/below bit-encoding-of-the-sign-bit as an
indication of having/needing is a feature for making the pro-weighted
approach less computationally complex, and more immediately obvious as a
heuristic approach.

frontend/utils/distribution/payments-minimizer.js
CrazedCoding: Ok... you caught me...

I just copied another part of GI's code to fix the flow errors. I forgot where (unfortunately); it might have been Steven's old distribution code.

TBH: I am actually quite unfamiliar with this type of "strict-typing" in JS, which is why I adopted existing code. This fixed flow-errors before pushing the changes (you _completely_ busted me! [perhaps I was trying to publish too fast?]).

I would mad appreciate it if you can help find the correct syntax for removing the first 'any' key-word from this puzzle! I did *brefly* try to do this, but gave up quite quick. Right now, I am way more concerned with the core-distribution algorithm (as I am balancing many perspectives).

For now, I will add solving this syntax puzzle to my TODO list.

frontend/model/contracts/group.js
taoeffect: OK, then since this isn't actually their income, let's rename this parameter everywhere it appears. Both here, in the function name, and in the data that is set in the event.

This isn't an income, it's something else. It is a pledge or a need. Or a have/need. So maybe let's call it `haveOrNeed`.

CrazedCoding: Will change this in next push.

frontend/utils/distribution/group-income-distribution.js
```diff
@@ -1,29 +1,16 @@
...
-      const adjustment = incomeDetailsType === 'incomeAmount' ? 0 : mincomeAmount
-      const amount = adjustment + profile[incomeDetailsType]
-      currentIncomeDistribution.push({
-        name: username,
-        amount: saferFloat(amount)
-      })
-    }
-  }
-  let dist = incomeDistribution(currentIncomeDistribution, mincomeAmount)
+  const distributionEvents = getters.currentGroupState.distributionEvents
```
taoeffect: I don't think this is a big deal, but whenever you have direct access to `state`, just access it directly via that, instead of calling `getters.currentGroupState`.

taoeffect: Actually, ignore that comment. In fact, using `state` at all in this function will cause problems. So it should probably be removed from the parameters, and only accessed the way you are accessing it.

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -0,0 +1,226 @@
...
+        p.amount = Math.abs(p.amount)
+        return p
+      })
+
+      const latePayments = monthlyDistribution.filter((p) => {
+        return p.amount > 0
+      })
+
+      lastStartCycleEvent = event
+      lastStartCycleEvent.data.latePayments = redistributOverToLatePayments(overPayments, latePayments)
```
taoeffect: This causes a Vue error (the one we talked about on the call), just marking this for reference (since the data passed in is being modified...)

frontend/utils/distribution/group-income-distribution.test.js
```diff
@@ -0,0 +1,186 @@
...
+  it('Test empty distirbution event list for un-adjusted distribution.', function () {
+    should(groupIncomeDistribution(setup, { adjusted: false, minimizeTxns: false, mincomeAmount })).eql([])
+  })
+  it('EVENTS: [u1, u5, u2 and u3] join the group and set haveNeeds of [100, 100, -50, and -50], respectively. Test unadjusted.', function () {
+    setup = [setup,
+      { type: 'haveNeedEvent', data: { name: 'u1', haveNeed: 100 } },
+      { type: 'haveNeedEvent', data: { name: 'u5', haveNeed: 100 } },
+      { type: 'haveNeedEvent', data: { name: 'u2', haveNeed: -50 } },
+      { type: 'haveNeedEvent', data: { name: 'u3', haveNeed: -50 } }
+    ].flat()
```
taoeffect: Recommend adjustments / alternatives to doing it with `.flat()`

- Option 1: `setup = [...setup, { type: 'haveNeedEvent' .. }]`
- Option 2: `setup = setup.concat([{ type: 'haveNeedEvent' .. }, ..])`

frontend/utils/distribution/group-income-distribution.js
```diff
@@ -1,40 +1,21 @@
...
-    }
-  }
-  let dist = incomeDistribution(currentIncomeDistribution, mincomeAmount)
-  if (adjusted) {
-    // if this user has already made some payments to other users this
-    // month, we need to take that into account and adjust the distribution.
-    // this will be used by the Payments page to tell how much still
-    // needs to be paid (if it was a partial payment).
+export default function groupIncomeDistribution (distributionEvents: Array<Object>, { mincomeAmount, adjusted = false, minimizeTxns = false, monthlyRated = true }: Object): any {
+  let dist = parseMonthlyDistributionFromEvents(distributionEvents, mincomeAmount, monthlyRated)
```
taoeffect: So the lines 7 - 19 can be removed, and the `adjusted` parameter can be passed forward to the `parseMonthlyDistributionFromEvents` function. E.g. something along these lines:

```js
function distribution (events, options = { adjusted: false, minimized: false }) {
  dist = unadjustedDistributionFromEvents(events)
  // NOTE: ignore this `options.adjusted` stuff below, it was from an old chat we had. You are
  // already doing the adjusted version way better than here.
  if (options.adjusted) {
    dist = parseMonthlyDistributionFromEvents(/* ... */)
    /* old crap // payments = events.filter((evt) => evt.type === 'payment')
    for (const payment of payments) {
      for (const todo of dist) {
        if (todo.from === payment.from) {
          dist.amount = Math.max(0, dist.amount - payment.amount)
        }
      }
    }*/
  }
  if (options.minimized) {
    dist = minimize(dist)
  }
  return dist
}
unadjustedDistributionFromEvents = unadjustedDistributionFromEventsObjectVersion
function unadjustedDistributionFromEventsObjectVersion (events) {
  const users = {}
  for (const { data } of events.filter(evt => evt.type === 'haveNeed')) {
    users[data.who] = data.haveNeed
  }
  return mincomeProportional(users)
}

// Note: you can either pass in `mincome` as a parameter
// or, you can pass in `mincome` as an event in the `events` and extract it
function unadjustedDistributionFromEventsArrayVersion (events /* , mincome (?) */) {
  const users = []
  for (const { data } of events.filter(evt => evt.type === 'haveNeed')) {
    users.removeWhere(u => u.name === data.who)
    users.push({ name: data.who, haveNeed: data.haveNeed })
  }
  return mincomeProportional(users, mincome)
}
```

E.g. when it's unadjusted you just ignore the payment events entirely, and only pay attention to the most recent `haveNeed` events for each user, calculating what you need to pass in to the mincome-proportional algo.

taoeffect: As an alternative to passing in the `adjusted` to `parseMonthlyDistributionFromEvents`, you can do the check on that parameter here in this function. E.g. something like this:

```js
export default function groupIncomeDistribution (distributionEvents: Array<Object>, options: Object): any {
  // NOTE: what I'm calling the "adjustedDistribution" function below _is_ your current
  // parseMonthlyDistributionFromEvents function (renamed).
  const dist = (options.adjusted ? adjustedDistribution : unadjustedDistribution)(distributionEvents, options)
  return options.minimizeTxns ? minimizeTotalPaymentsCount(dist) : dist
}
```

Where `unadjustedDistribution` is effectively something akin to `unadjustedDistributionFromEventsArrayVersion` in the previous comment.

frontend/model/contracts/group.js
```diff
@@ -504,8 +552,7 @@ DefineContract({
         }
       },
       process ({ data, meta }, { state, getters }) {
-        state.profiles[data.member].status = PROFILE_STATUS.REMOVED
-        state.profiles[data.member].departedDate = meta.createdDate
+        memberLeaves(state, data.member, meta.createdDate)
```
CrazedCoding, line 568: I think maybe the following two lines got accidentally deleted when adding this line:
```
        state.profiles[data.member].status = PROFILE_STATUS.REMOVED
        state.profiles[data.member].departedDate = meta.createdDate
```
@taoeffect Can you confirm that these lines were not removed by mistake?

frontend/model/contracts/group.js
```diff
@@ -539,8 +586,7 @@ DefineContract({
         reason: string
       }),
       process ({ data, meta, contractID }, { state, getters }) {
-        state.profiles[meta.username].status = PROFILE_STATUS.REMOVED
-        state.profiles[meta.username].departedDate = meta.createdDate
+        memberLeaves(state, meta.username, meta.createdDate)
```
CrazedCoding, line 602: Again, I think maybe the following some lines got accidentally deleted when adding this line:
```
        state.profiles[meta.username].status = PROFILE_STATUS.REMOVED
        state.profiles[meta.username].departedDate = meta.createdDate
```
@taoeffect Can you confirm that these lines were not removed by mistake?

frontend/utils/distribution/group-income-distribution.js
CrazedCoding: @taoeffect I am going to push my fix for cypress first. This is next on the TODO list.

frontend/utils/distribution/group-income-distribution.test.js
CrazedCoding: @taoeffect Again, this is (after) the next item on the TODO list. Going to push cypress fixes, first.

frontend/utils/distribution/payments-minimizer.js
CrazedCoding: @taoeffect Update: this is ([after] after) the next item on the TODO list. Going to push cypress fixes, first, and fix a few of the other items, too.

CrazedCoding (commented):
Almost done ^^

frontend/model/contracts/group.js
taoeffect: @CrazedCoding haven't these two lines been moved into `memberLeaves`?

CrazedCoding: @taoeffect This is correct, actually. Still getting used to reading the git diff's ><

Thanks for confirming.

frontend/utils/distribution/group-income-distribution.js
```diff
@@ -1,40 +1,7 @@
...
-    for (const p of dist) {
-      const alreadyPaid = getters.paymentTotalFromUserToUser(p.from, p.to, monthstamp)
-      // if we "overpaid" because we sent late payments, remove us from consideration
-      p.amount = saferFloat(Math.max(0, p.amount - alreadyPaid))
-    }
-    dist = dist.filter(p => p.amount > 0)
-  }
-  return dist
+export default function groupIncomeDistribution (distributionEvents: Array<Object>, { mincomeAmount, adjusted = false, minimizeTxns = false, monthlyRated = true }: Object): any {
+  const dist = parseMonthlyDistributionFromEvents(distributionEvents, mincomeAmount, monthlyRated, adjusted)
```
taoeffect: See this discussion for how this function should look (it should call two different functions, based on the value of `options.adjusted`): https://github.com/okTurtles/group-income-simple/pull/1056#discussion_r603473612

taoeffect: After discussion with @CrazedCoding I think we possibly came to an agreement, that we're going to leave the handling of the `options.adjusted` parameter inside of the `parseMonthlyDistributionFromEvents` function, so that if we ever enable pro-rating, we can get an unadjustedDistribution that accounts for the pro-rating of the incomes.

frontend/controller/actions/group.js
```diff
@@ -185,5 +185,15 @@ export default (sbp('sbp/selectors/register', {
...
     }
+  },
+  'gi.actions/group/resetMonth': async function (groupId) {
+    try {
+      const message = await sbp('gi.contracts/group/resetMonth/create', '', groupId)
+      await sbp('backend/publishLogEntry', message)
+      return message
+    } catch (e) {
+      console.error('gi.actions/group/resetMonth failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to resetMonth. {reportError}', LError(e)))
```
taoeffect: `GIErrorUIRuntimeError`s are user-facing errors. Therefore the error message inside of these must always be carefully considered and ensured that it is user-friendly.

So let's give this a better error, something like `L('Failed to reset monthly stats: {reportError}', ', LError(e))` perhaps.

frontend/controller/monthlyCycleTimer.js
```diff
@@ -0,0 +1,25 @@
...
+let timerStartInterval
+
+export function startMonthlyCycleCheckInterval (store: Object) {
+  if (timerStartInterval) clearInterval(timerStartInterval)
+  timerStartInterval = setInterval(() => {
+    if (store.state.currentGroupId) {
+      const events = store.getters.currentGroupState.distributionEvents
+      const firstEvent = events[0]
+      const lastEvent = events[events.length - 1]
+      if (Math.floor(lastEvent.data.cycle) !== Math.floor(cycleAtDate(new Date(), firstEvent.data.when))) {
```
taoeffect: Can you add a comment above this line explaining in plain english what this check is doing? (I assume it's checking whether we're in a new month)?

Also, doesn't this need a check to see whether the month has already been reset (by us or someone else)?

frontend/controller/monthlyCycleTimer.js
```diff
@@ -0,0 +1,25 @@
...
+        console.table(lastEvent)
```
taoeffect: Had no idea this was a thing, cool!

frontend/utils/time.js
```diff
@@ -87,3 +91,16 @@ export function getTime (date: Date): string {
   const t = new Date(date)
   return `${t.getHours()}:${t.getMinutes()}`
 }
+
+// Returns the ratio of (the current day of the month) / (days in current month);
+// The current day / month are calculated with respect to the atDate parameter.
+export function cycleAtDate (atDate: string | Date, startDate: string | Date): any | number {
```
taoeffect: 1. Remove `any` cop-out return value :P
2. Recommendation: decomplicificate the types - I think it's safe to make the parameter types be `string`, and use `new Date().toISOString()` where you pass in `new Date()` elsewhere

Also, what is the point of `startDate`? The comment says this function does the following:

> Returns the ratio of (the current day of the month) / (days in current month)

If that were true, then there is no need for the second `startDate` parameter, because you can figure out how many days are in the current month at `atDate`. So yeah, it's not clear to me what this function is doing. Either the comment should be fixed or the parameter `startDate` removed and the function simplified.

frontend/model/contracts/group.js
```diff
@@ -697,7 +767,26 @@ DefineContract({
         // }
       }
     },
-
+    'gi.contracts/group/resetMonth': {
+      validate: optional(string),
+      process (_, { state, getters }) {
+        // Loop through missing monthly cycle events that happen before the 'event' parameter's cycle
+        let lastEvent = state.distributionEvents[state.distributionEvents.length - 1]
+        const currentCycle = cycleAtDate(new Date(), state.distributionEvents[0].data.when)
```
taoeffect: `new Date()` inside of contract is a big no-no

frontend/model/contracts/group.js
```diff
@@ -107,6 +107,50 @@ function initFetchMonthlyPayments ({ meta, state, getters }) {
...
+      cycle: cycleAtDate(createdDate, state.distributionEvents[0].data.when)
+    }
+  })
+}
+
+function memberLeaves (state, username, dateLeft) {
+  state.profiles[username].status = PROFILE_STATUS.REMOVED
+  state.profiles[username].departedDate = dateLeft
+  insertMonthlyCycleEvent(state, {
+    type: 'userExitsGroupEvent',
```
taoeffect, line 151: Do you need the `userExitsGroupEvent`? Or can you modify your algorithm to simplify it by adding a `haveNeedEvent` that sets the pledge to 0 for the user that left?

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -0,0 +1,199 @@
...
+      groupMembers = groupMembers.filter((v) => { return v.name !== event.data.name })
+    }
+  }
+
+  return lastStartCycleEvent.data.latePayments.filter((payment) => {
+    return payment.from !== payment.to // This happens when a haver switches to being a needer; remove neutral distribution payments.
+  }) // TODO: return late-payments as well.
+}
+
+export default parseMonthlyDistributionFromEvents
```
taoeffect: In this file, can you:

- Remove as many `any`s as you can?
- Simplify and dry the code as much as you can? (E.g. reduce the linecount)?

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -0,0 +1,199 @@
...
+  let groupMembers = []
+
+  // Convenience function for retreiving a user by name:
+  const getUser = function (userName) {
+    for (const member of groupMembers) {
+      if (member.name === userName) {
+        return member
+      }
+    }
+  }
```
taoeffect: This looks like it could be written in a single line with `.find()`?

E.g. someting like:

```js
const getUser = name => groupMembers.find( ... )
```

taoeffect (changes requested):
First quick-pass review.

frontend/controller/actions/group.js
CrazedCoding: Roger. Changed to this.

frontend/controller/monthlyCycleTimer.js
CrazedCoding: That is what these lines are actually for: these lines check to see if the last event of the `distributionEvents` has not already been reset.

This check also happens again in `model/contracts/group.js`.in the condition of the `while` loop of `resetMonth`.

I will add these comments.

frontend/controller/monthlyCycleTimer.js
CrazedCoding: Yepperoni ^^

Very useful.

frontend/utils/time.js
CrazedCoding: > Remove any cop-out return value :P

:P Fine. Lol.

> Recommendation: decomplicificate the types - I think it's safe to make the parameter types be string, and use new Date().toISOString() where you pass in new Date() elsewhere

I did not do this because it was DRYer to do this internally. Can you double verify this?

> Also, what is the point of startDate? The comment says this function does the following:

Good catch. I totally didn't update this comment/annotation. Fixing.

The old way of doing this function only returned the fraction of the way through the month. Now it adds 1 to the final cycle integer returned for each month that passes, and 12 for every year that passes. This is necessary for `monthlyCycleTimer.js` to track the month changes as well as update the distributionEvents accordingly.

frontend/model/contracts/group.js
CrazedCoding: Is it OK if I pass this in as a parameter to the process function?

Knowing current month is absolutely necessary.

frontend/model/contracts/group.js
CrazedCoding: I designed it this way because, eventually, `monthlyEventParser.js` needs to know who is present in order to handle mincome adjustments. Having a needer-pledger would cause this value of 0 to be adjusted with mincome changes.

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: Good observation. Fixed.

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: > Remove as many anys as you can?

Done.

> Simplify and dry the code as much as you can? (E.g. reduce the linecount)?

See response to next comment. It's a start.

frontend/model/contracts/group.js
taoeffect: It's already passed in to the first parameter. See the other actions for examples, where they use `meta.createdDate`

frontend/model/contracts/group.js
```diff
@@ -107,6 +107,50 @@ function initFetchMonthlyPayments ({ meta, state, getters }) {
...
+      type: 'startCycleEvent',
+      data: {
+        cycle: lastEvent.data.cycle + 1,
+        latePayments: [] // List to be populated later, by the events-parser
+      }
+    }
+    state.distributionEvents.push(monthlyCycleEvent)
+    lastEvent = monthlyCycleEvent
+  }
+
```
taoeffect, line 124: Convert to `if` statement per call.

frontend/model/contracts/group.js
```diff
@@ -697,7 +767,25 @@ DefineContract({
...
+          const monthlyCycleEvent = {
+            type: 'startCycleEvent',
+            data: {
+              cycle: Math.floor(lastEvent.data.cycle + 1),
+              latePayments: [] // List to be populated later, by the events-parser
+            }
+          }
+          state.distributionEvents.push(monthlyCycleEvent)
+          lastEvent = monthlyCycleEvent
+        }
```
taoeffect: Convert this whole thing to an `if` statement per call.

frontend/model/contracts/group.js
```diff
@@ -697,7 +767,25 @@ DefineContract({
...
+      validate: optional(string),
+      process ({ meta }, { state, getters }) {
+        let lastEvent = state.distributionEvents[state.distributionEvents.length - 1]
+        const currentCycle = cycleAtDate(meta.createdDate, state.distributionCycleStartDate)
+        // Add 'startCycleEvent' events for every month missed.
+        while (Math.floor(lastEvent.data.cycle) < Math.floor(currentCycle)) {
+          const monthlyCycleEvent = {
+            type: 'startCycleEvent',
+            data: {
+              cycle: Math.floor(lastEvent.data.cycle + 1),
```
taoeffect: This needs to be modified to `currentCycle` I think.

taoeffect (commented):
Comments per call.

frontend/model/contracts/group.js
```diff
@@ -107,6 +107,50 @@ function initFetchMonthlyPayments ({ meta, state, getters }) {
...
 
+function insertMonthlyCycleEvent (state, event) {
+  // Loop through missing monthly cycle events that happen before the 'event' parameter's cycle
+  let lastEvent = state.distributionEvents[state.distributionEvents.length - 1]
+  while (Math.floor(lastEvent.data.cycle) !== Math.floor(event.data.cycle)) {
+    // Add the missing monthly cycle event
+    const monthlyCycleEvent = {
+      type: 'startCycleEvent',
+      data: {
+        cycle: lastEvent.data.cycle + 1,
```
taoeffect: Think this also needs to be based on `meta.createdDate`

frontend/model/contracts/group.js
```diff
@@ -369,9 +428,18 @@ DefineContract({
...
-            state, getters, updateMonthstamp, adjusted: true
+          insertMonthlyCycleEvent(state, {
+            type: 'paymentEvent',
+            data: {
+              from: meta.username,
+              to: payment.data.toUser,
+              hash: data.paymentHash,
+              amount: payment.data.amount,
+              when: meta.createdDate,
+              cycle: cycleAtDate(meta.createdDate, state.distributionEvents[0].data.when)
```
taoeffect: Think `state.distributionEvents[0].data.when` is incorrect...

frontend/utils/time.js
```diff
@@ -87,3 +91,17 @@ export function getTime (date: Date): string {
   const t = new Date(date)
   return `${t.getHours()}:${t.getMinutes()}`
 }
+
+// Returns months passed + the ratio of (the current day of the month) / (days in current month);
+// The current day / month are calculated with respect to the atDate parameter. Months passed
+// is caclculated from the startDate parameter.
+export function cycleAtDate (atDate: string | Date, startDate: string | Date): number {
```
taoeffect: The second parameter `startDate` is causing so much confusion in the code, and bugs. Please remove it and to compare months, compare the `atDate` directly to another date using `compareMonthstamps`, or create a similar function to compare ISOStrings using months...

E.g. you can create a function that effectively does this:

```js
const month1 = dateToMonthstamp(isoString1)
const month2 = dateToMonthstamp(isoString2)
return compareMonthstamps(month1, month2)
```

Then use the `meta.createdDate` instead of the integer part of the `cycle` (and remove the integer part from this function).

You can name the function something like `compareDatesAsMonths`, which uses the above code inside of it.

frontend/controller/monthlyCycleTimer.js
```diff
@@ -0,0 +1,26 @@
...
+export function startMonthlyCycleCheckInterval (store: Object) {
+  if (timerStartInterval) clearInterval(timerStartInterval)
+  timerStartInterval = setInterval(() => {
+    if (store.state.currentGroupId) {
+      const events = store.getters.currentGroupState.distributionEvents
+      const lastEvent = events[events.length - 1]
+      const lastCycle = dateToMonthstamp(lastEvent.data.when)
+      const currentCycle = currentMonthstamp()
+      // Add 'startCycleEvent' events if a month has passed.
+      if (compareMonthstamps(currentCycle, lastCycle) > 0) {
```
taoeffect, line 21: I believe we need one additional check here, and that is to see whether or not we have already reset the month.

frontend/model/contracts/group.js
```diff
@@ -246,6 +293,12 @@ DefineContract({
       //       Just make sure the UI doesn't break if an exception is thrown, since this is
       //       bound to the UI in some location.
       return getters.groupSettings.mincomeCurrency && currencies[getters.groupSettings.mincomeCurrency].displayWithCurrency
+    },
+    groupCreationDate (state, getters) {
+      return getters.currentGroupState.distributionCycleStartDate
+    },
```
taoeffect: I believe `groupCreationDate` and `distributionCycleStartDate` and everything related to them can now be deleted since they're no longer used/needed.

frontend/model/contracts/group.js
```diff
@@ -697,7 +768,29 @@ DefineContract({
...
+        let lastEvent = state.distributionEvents[state.distributionEvents.length - 1]
+        // Fills in multiple missing months when `while` instead of `if`.
+        while (compareCycles(meta.createdDate, lastEvent.data.when) > 0) {
+          // Add the missing monthly cycle event
+          const monthlyCycleEvent = {
+            type: 'startCycleEvent',
+            data: {
+              latePayments: [], // List to be populated later, by the events-parser
+              when: dateToMonthstamp(addMonthsToDate(dateToMonthstamp(lastEvent.data.when), 1))
+            }
```
taoeffect: AFAICT, if on line 783 we use a `when` = `meta.createdDate` instead of using `addMonthsToDate`, then this `while` loop can be replaced with an `if`.

Is there any reason to not do that?

taoeffect (changes requested):
Some additional review!

frontend/utils/distribution/group-income-distribution.test.js
CrazedCoding: Went with Option 1. You'll see it in the next push.

frontend/model/contracts/group.js
CrazedCoding: Changed.

frontend/model/contracts/group.js
CrazedCoding: Changed.

frontend/model/contracts/group.js
CrazedCoding: Good observation. Fixed.

frontend/model/contracts/group.js
CrazedCoding: ```// Fills in multiple missing months when `while` instead of `if`.```

Think of the ``distributionEvents`` as a way of saving the previous two month's, in addition to the current month's information (such as ```latePayments```). If we don't have this information being propagated and preserved here, it will have to be inferred (or worse: entirely lost) later when the final distribution function is called.

I have changed many cases of using an ```while``` to being ```if```, for now... I 100% believe this results in the problem mentioned above, and edge-case problems mentioned on our calls (even though they have a rare probability of occurring)!

I think we might need to keep this comment open issue so we don't forget about it and we can discuss it further, while we try to create cypress/mocha tests that can actually prove these rare edge-cases exist.

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -20,7 +18,6 @@ function reduceDistribution (payments: Array<Object>): Array<Object> {
         // Add or subtract paymentB's amount to paymentA's amount, depending on the relative
         // direction of the two payments:
         paymentA.amount += (paymentA.from === paymentB.from ? 1 : -1) * paymentB.amount
-
         // Remove paymentB from payments, and decrement the inner sentinal loop variable:
         payments = payments.filter((_, paymentIndex) => { return paymentIndex !== j })
```
snowteamer: Unless the array contains duplicate references, then this could likely be simplified into:
```js
payments = payments.filter((payment) => payment !== paymentB)
 ```

File: frontend/utils/distribution/monthly-event-parser.js [hunk 28]
```diff
@@ -1,14 +1,12 @@
 'use strict'
 import incomeDistribution from '~/frontend/utils/distribution/mincome-proportional.js'
-import { cycleAtDate } from '~/frontend/utils/time.js'
+import { lastDayOfMonth, dateFromMonthstamp, dateToMonthstamp, prevMonthstamp } from '~/frontend/utils/time.js'
 
 // Flatten's out multiple payments between unique combinations of users
```
snowteamer: IIUC, the payment array is not a nested array and therefore "Flatten out" is confusing. I'd rather say something about the payments being merged together in some way by the function

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -50,66 +47,40 @@ function subtractDistributions (paymentsA: Array<Object>, paymentsB: Array<Objec
 }
 
 // This algorithm is responsible for calculating the monthly-rated distribution of
-// payments. monthlyRated = True if time-weighted monthly, false if purely pro-rated.
-function parseMonthlyDistributionFromEvents (distributionEvents: Array<Object>, minCome: number, monthlyRated: Boolean, adjusted: Boolean): Array<Object> {
+// payments.
+function parseMonthlyDistributionFromEvents (distributionEvents: Array<Object>, minCome: number, adjusted: Boolean): Array<Object> {
```
snowteamer: `mincome` doesn't need a capital `C`

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: It shouldn't have any duplicate references, as each payment object is parsed from a string:
`payments = JSON.parse(JSON.stringify(payments))`

Also: your suggestion reads nicer, and doesn't break any of the tests, so I'm using it; 'preciate it, @snowteamer

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: Good point. I changed the comment to:

`// Merges multiple payments between any combinations two of users:`

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: However minor, I actually totally agree with your (de)capitalization.

I can't remember why it is even capitalized, except it was a typo by a previous coder.

@taoeffect if you don't mind I'm fixing/closing this one before the next update.

taoeffect: @CrazedCoding don't mind, go for it 👍

frontend/utils/distribution/payments-minimizer.js
snowteamer: I would suggest to create a Payment type so that it can be used to make such function signatures shorter and possibly more useful too:
```js
type Payment = {|
  from: string;
  to: string;
  amount: number;
  total: number;
|}

function minimizeTotalPaymentsCount (distribution: Payment[]): Payment[] {
  ...
 ```
 Maybe a Distribution or MonthlyDistribution type could help even further, e.g.:
 ```js
 ...
 type Distribution = Array<Payment>;
 
 function minimizeTotalPaymentsCount (distribution: Distribution): Distribution {
  ...
 ```
 ```

File: frontend/utils/distribution/monthly-event-parser.js [hunk 30]
```diff
@@ -0,0 +1,241 @@
...
+  })
+
+  const lastWhen = distributionEvents[distributionEvents.length - 1].data.when
+
+  const artificialEnd = {
+    type: 'startCycleEvent',
+    data: {
+      when: lastWhen,
+      monthlyDistribution: [] // List to be populated later, by the events-parser
+    }
```
snowteamer: Flow complains that `completedPayments` property is missing in `data`

frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -0,0 +1,241 @@
...
+  startCycleEvent.data.monthlyDistribution.forEach((payment) => {
+    payment.amount = Math.min(payment.amount, payment.total)
+    payment.isLate = false
+    payment.dueOn = dateToMonthstamp(lastDayOfMonth(dateFromMonthstamp(dateToMonthstamp(new Date(lastWhen)))))
+  })
+
+  const previousTwoCycles = [
+    lastStartCycleEvent.data.monthlyDistribution,
+    startCycleEvent.data.monthlyDistribution
+  ].flat()
```
snowteamer: I would prefer using a pair of `...` operators here rather than `.flat()`, although that's no big deal so feel free to ignore this if you prefer `.flat()`:
```js
const previousTwoCycles = [
  ...lastStartCycleEvent.data.monthlyDistribution,
  ...startCycleEvent.data.monthlyDistribution
]
```

CrazedCoding: Thanks for the comment @snowteamer. I marked this as resolved because I have now switched to using your way, in most cases.

frontend/utils/distribution/payments-minimizer.js
CrazedCoding: Added to local changes, will update PR later.

Thanks for the suggestions. It reads much nicer.

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: `redistributOverToLatePayments ` has been removed in later updates, and there don't seem to be any Vue errors in the latest.

Marking resolved, for now.

frontend/utils/distribution/group-income-distribution.js
CrazedCoding: Ok.

frontend/utils/distribution/group-income-distribution.js
CrazedCoding: Correct me if I am wrong, but I think this is outdated because the `forgiveWithoutForget` replaces much of this code.

Marking resolved, for now.

frontend/controller/monthlyCycleTimer.js
CrazedCoding: I'm still confused by this suggestion.

There don't seem to be any other variables to check, except maybe a redundant check done inside `frontent/model/group.js`; the SBP call to `gi.actions/group/resetMonth` has this additional checks.

Let me know your thoughts, @taoeffect

taoeffect: Alright looking more closely, I trust that if a previous `startCycleEvent` was already pushed, then `currentCycle` and `lastCycle` should be equal the next time around, and so this check should prevent `resetMonth` spam from entering the log.

frontend/model/contracts/group.js
taoeffect: OK, trust you on this.

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: Looks like this section of code has been updated, @snowteamer .

Let me know if you spot anything like this in future.

frontend/model/contracts/group.js
```diff
@@ -368,9 +422,18 @@ sbp('chelonia/defineContract', {
             const toUser = vueFetchInitKV(fromUser, data.toUser, [])
             toUser.push(data.paymentHash)
           }
-          paymentMonth.lastAdjustedDistribution = groupIncomeDistribution({
-            state, getters, updateMonthstamp, adjusted: true
+          console.error(payment)
```
taoeffect: This shouldn't be here...

Screen Shot 2021-11-27 at 11 35 44 AM

taoeffect (changes requested):
Excellent work @CrazedCoding!

Think this is very close to ready to merge. The charts and graphs seem to not be broken, which is great, and during my testing I only found an extraneous `console.error` (see recent comment & screenshot - I recommend testing yourself in Firefox container tabs manually to catch these things).

Can you also merge `master` into your PR? It will just make it easier for others to test it without having to change the dependencies multiple times.

frontend/model/contracts/group.js
CrazedCoding: Nice catch.

Fixed.

taoeffect (approved):
Approved!! 🎉🎉🎉🎉

PR #1059, opened 2021-03-26
#1058 Enable require-await rule and fix lint errors

### LINKED ISSUE #1058: Enable `require-await` eslint rule
### Problem

Per https://github.com/okTurtles/group-income-simple/pull/1052#discussion_r601597165

There are apparently several places where `async` is used to describe a function with no corresponding `await`.

This can create confusion as to what the return value is supposed to be (e.g. a promise or a raw value) and could lead to bugs.

### Solution

Add `require-await` and fix all existing errors, making sure to double-check the return values are properly handled when `async` is removed.

snowteamer:
Closes #1058


taoeffect (approved):
Fantastic @snowteamer! Thanks for the quick turnaround on this one!

Issue #1060, opened 2021-04-08
Implement "Internet connection lost" & "Server down" banners (BannerGeneral) [$150 bounty]


taoeffect:
### Problem

At the moment we do not display the global `BannerGeneral` when an internet connection is lost, or the server is down.

Please see these links for full context:

- https://github.com/okTurtles/group-income-simple/issues/688#issuecomment-553473328
- https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=426%3A6

### Solution

Using the latest events implemented in @snowteamer's recent PR #1052, implement event handlers to display two different kinds of general banners (and hide them as appropriate):

1. When the Internet connection is down
2. When the Internet is up, but the server is down (e.g. via the ping/pong messages)

See `frontend/main.js` for examples of how to display the `BannerGeneral` component. E.g.:

```js
      sbp('okTurtles.data/set', 'BANNER', this.$refs.bannerGeneral)
      // call from anywhere in the app:
      // sbp('okTurtles.data/get', 'BANNER').show(L('Trying to reconnect...'), 'wifi')
      // sbp('okTurtles.data/get', 'BANNER').danger(L('message'), 'icon-type')
      // sbp('okTurtles.data/get', 'BANNER').clean()
```

### Bounty

$150 bounty for a clean solution to this (paid in cryptocurrency).

snowteamer:
We can also use native `online` and `offline` events.

There are several possibilities as to where to implement the relevant event handlers:
- In the `handlers` option when the `GiPubSubClient` instance is created, using plain functions. That's probably the solution requiring the fewest lines of code.
- In the same file but just after the `createGiPubSubClient()` call, using more SBP calls.
- In `main.js` where the banner usage comments are currently located.
- In yet another file, maybe something like `general-banner-events.js`, if we do not want to clutter `main.js` with a bunch of event handlers.

taoeffect:
@snowteamer if you'd like to go for this issue, then wherever you implement it, I recommend making sure all of the event handlers for these types of events (ping/pong stopped, online/offline) are grouped together in the same place. Fewer lines of code is better. So organized & simple is what we're aiming for. I agree that we shouldn't be cluttering up `main.js`. 👍

snowteamer:
Another point: once an agreement has been reached regarding the messages to display, their French translation strings should be added in `french.strings`.

snowteamer:
[...]

PR #1061, opened 2021-04-10
#1060 Implement some status messages using BannerGeneral

snowteamer:
Related to #1060

This PR is currently intended as a working draft, so as to have actual code to discuss and experiment with.

Important: design, code organization, and message contents still need to be discussed.

Corresponding French string translations are currently not included, waiting for banner message contents to be agreed upon.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -136,13 +136,40 @@ async function startApp () {
         this.ephemeral.finishedLogin = 'no'
         router.currentRoute.path !== '/' && router.push({ path: '/' }).catch(console.error)
       })
-
       sbp('okTurtles.data/set', 'BANNER', this.$refs.bannerGeneral)
       // call from anywhere in the app:
       // sbp('okTurtles.data/get', 'BANNER').show(L('Trying to reconnect...'), 'wifi')
       // sbp('okTurtles.data/get', 'BANNER').danger(L('message'), 'icon-type')
       // sbp('okTurtles.data/get', 'BANNER').clean()
+      sbp('okTurtles.data/apply', PUBSUB_INSTANCE, (instance) => {
+        const banner = this.$refs.bannerGeneral
+        // Allow to access `L` inside event handlers.
+        const L = this.L.bind(this)
 
+        Object.assign(instance.customEventHandlers, {
+          offline () {
+            banner.show(L('Your device appears to be offline.'), 'wifi')
+          },
+          online () {
+            banner.clean()
+          },
+          'reconnection-attempt' () {
+            banner.show(L('Trying to reconnect...'), 'wifi')
+          },
+          'reconnection-failed' () {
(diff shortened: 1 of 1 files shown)


frontend/main.js
```diff
@@ -136,13 +136,40 @@ async function startApp () {
...
+          },
+          'reconnection-failed' () {
+            banner.show(L('Cannot connect to the server.'), 'wifi')
+          },
+          'reconnection-succeeded' () {
+            banner.clean()
+          }
+        })
+      })
+      // Useful in case the app is started in offline mode.
```
taoeffect, line 167: So this is where I think we come to a situation that we didn't fully address in the previous PR. Correct me if I'm wrong, but AFAICT assigning to `customEventHandlers` like this will override any other handlers. And it's not even clear to me whether assigning to them after instance creation will work..

Can you modify the way that events work in the pubsub client so that this doesn't happen? One suggestion I have is to make full use of `okTurtles.events` for all of the event handling, since it doesn't have this problem. It would involve modifying the way events are handled in `shared/pubsub.js`.

taoeffect (commented):
Thanks @snowteamer for the quick PR! I tested it out and it works. However I have one question about the way the event listeners are implemented, see comment!

frontend/main.js
taoeffect: > And it's not even clear to me whether assigning to them after instance creation will work..

OK, I verified that it works, but I'm not sure why.

```js
  for (const name of Object.keys(defaultClientEventHandlers)) {
    client.listeners[name] = (event) => {
      const customHandler = client.customEventHandlers[name]
```

This code is from `createClient`, and I assume this is the only place where `customEventHandlers` is used. Initially it is `null`, so don't `const customHandler` get assigned to `null`? Why is it that assigning `customEventHandlers` still works after calling `createClient`? I must be missing something.

taoeffect:
> Corresponding French string translations are currently not included, waiting for banner message contents to be agreed upon.

The strings get a 👍 from me.

frontend/main.js
snowteamer: > Correct me if I'm wrong, but AFAICT assigning to customEventHandlers like this will override any other handlers

It's true that it will override a previously assigned *custom* handler if there is one.

However, it will never override any so-called *default* event handler since they are hardcoded in `shared/pubsub.js` and cannot be mutated from outside that file.

taoeffect: Alright, we can go with this. It works well enough.

I also deleted/retract my second question, as I see now why it works. (I must be having an off day).

taoeffect (approved):
Approved! Great job @snowteamer!

frontend/main.js
snowteamer: Well I have to admit that this code was not the easiest to follow. Dynamic event programming is a bit tricky! 😄

Issue #902, opened 2020-04-19
Limit length on all user-provided strings/input into contracts


taoeffect:
### Problem

Currently we do not place hard limit checks in the contracts on the length of strings, so this means things like usernames, profile summaries, bitcoin addresses, etc., are all technically allowed to be unlimited in size, presenting a DOS issue.

### Solution

Replace all `string` validators in the contracts with a `stringMax()` function that takes a byte length and throws if the string is greater than that many bytes.

If it throws, make sure to ban the user from the group. Also place a limit on the maximum size of any single GIMessage, and make sure the server enforces it. Server-side ban users that exceed this limit.

sandrina-p:
> If it throws, make sure to ban the user from the group.

Why banning? Can't we just reject it?

taoeffect:
> Why banning? Can't we just reject it?

Because in this case it would mean the frontend validators were purposefully bypassed, e.g. someone is hacking the app.

sandrina-p:
Fair enough

snowteamer:
Most of the front end validators don't seem to work at all on my end. The first thing I did upon launching the app for the mode was typing a very long username and password, and they got accepted by both the client and server.

Is this because the app was not running in production mode?

PR #1064, opened 2021-05-01

### LINKED ISSUE #975: Check the project for XSS [$250 bounty]
### Problem

```js
    }),
```

And later on we have:

```pug
```

### Solution

### Bounty

snowteamer:
Closes #975

### Summary of changes:
- Usage of `innerHTML` assignment in `SendArea.vue` has been replaced by `textContent` assignment.
- Usage of `rel="noopener noreferrer"` on outgoing hyperlinks can be enforced by creating them with the new `safeLinkTag()` function.


taoeffect (changes requested):
Very nice PR @snowteamer!

I tested it out to see if it would stop `xss"'>` from working in the contributions field or in the group name and it didn't. :-\

Screen Shot 2021-05-04 at 3 39 35 PM
Screen Shot 2021-05-04 at 3 39 13 PM
Screen Shot 2021-05-04 at 3 34 25 PM

snowteamer:
It looks like another kind of vulnerability is involved here, unrelated to `v-html` or `innerHTML` usage but rather to pug templates: 

For example, the issue with group name vulnerability seem to arise from `GroupWelcome.vue`:
https://github.com/okTurtles/group-income-simple/blob/7693b98575f8746fd26dd0e6fab9afc2ac63d899/frontend/views/components/GroupWelcome.vue#L9-L13

 where `groupName` gets included in the HTML despite coming from unsanitized user input.

I think a possible solution would be the sanitization of every textual user input. Patching places where  `v-html` or `innerHTML` are used is not enough to secure the application.

taoeffect:
@snowteamer I was a bit confused by the example you gave, but then realized what the confusion was. `{groupName}` is not Vue.js [template syntax](https://vuejs.org/v2/guide/syntax.html), which AFAIK doesn't have this problem, [but rather it is our custom `i18n` translation stuff](https://github.com/okTurtles/group-income-simple/blob/master/frontend/views/utils/translations.js#L137-L160).

That's what needs to be updated to address this. I see there is at least one assignment there to `innerHTML`, and there might be something else that needs to be modified.

Regarding the assignment to `innerHTML`, I think that needs to be kept as `innerHTML` so that rendering properly occurs when we pass in HTML ourselves, however `dompurify` can be used on any arguments to restrict the HTML tags to things like `<b>`, `<br>`, `<i>`, `<a>`, making sure to remove any potential XSS type `onclick` `onhover` etc stuff.

snowteamer:
Some of our i18n component instances contain an embedded anchor link, so we have to allow `<a>` tags as well as the `href` attribute in the `dompurify` config when sanititizing those strings, which might however be unsafe if user input is injected.

Besides, things such as group names or user names are plain text, and should likely not contain any character specific to JS or HTML such as `<>/\()"'&` (although `&` and `/` might be allowed if followed by a plain white space, to enable e.g. groupe names such as Turtles & Friends).

Therefore I think we need both
- whole i18n string sanitization (allowing only tags and attributes required for correct rendering),
- and thorough user input validation, both client-side and server-side. We probably don't want to store malicious input, even if we sanitize it every time before rendering. This could also solve #902.

@joepie91 I'm not really a security expert, so your expertise would be much appreciated here!

joepie91:
@snowteamer I won't be available going forward, but I'll try to answer your question as best as I can here:

> We probably don't want to store malicious input, even if we sanitize it every time before rendering. 

Storage-wise, there is no such thing as "malicious input", and so input should be stored as-is. Until the point of use, it's just a string of characters with no special meaning, and importantly, *which* characters have special meaning will change over time as you target different output forms.

The typical example of this is people who HTML-escape values before storing them; then later they decide to build a JSON API, and suddenly their JSON responses incorrectly contain HTML entities, because an assumption of "only HTML" as the output format was encoded into the stored data. Now HTML-escaping is reversible, so it's at least fixable, but many other 'input validation' approaches are lossy and essentially result in irrevocable data corruption.

Therefore, the stored data should always be the original data as-is, and it's the rendering logic that should process it in a manner appropriate to the rendering context. In most cases (eg. modern templaters, React, JSON encoders, etc.) that does not require any special actions, as you are *already* specifying the text separately from the markup/structure, and so the encoder will automatically escape as-needed.

However...

> Some of our i18n component instances contain an embedded anchor link, so we have to allow <a> tags as well as the href attribute in the dompurify config when sanititizing those strings, which might however be unsafe if user input is injected.

That's a trickier case. Essentially, your rendering target is now not "the page", but "the i18n implementation", and the actual rendering-to-page will have to be a direct raw HTML insert due to the i18n implementation outputting structurally significant HTML.

In practice, however, that just means that the "point of rendering" changes - now, instead of escaping content prior to injection into the page, you need to escape content *prior to injection into the i18n string* (which presumably also happens at rendering time). That way, the user-specified parts of the resulting string have been 'immunized' upfront and so any actual HTML output can be assumed to have come from the i18n templates.

> Besides, things such as group names or user names are plain text, and should likely not contain any character specific to JS or HTML such as <>/\()"'& (although & and / might be allowed if followed by a plain white space, to enable e.g. groupe names such as Turtles & Friends).

I would recommend against introducing character set restrictions for this reason. It's a much less reliable approach than just doing eg. HTML-escaping, while also inconveniencing the user, and it's risky from a future perspective - you might decide today to not allow 'dangerous characters', but a future developer won't be aware of the reasons for that decision, and in the future might go "hmm, strange, brackets are not allowed, dunno why, let's enable them because this user asked for it" and then you have a security issue.

As a general rule, assumptions are the most dangerous thing in building secure software; every assumption you make that isn't enforced in some way, is likely to bite you later when everybody has forgotten about the original assumption. Ideally your code should continue to function safely *in spite of* such assumptions.

> whole i18n string sanitization (allowing only tags and attributes required for correct rendering),

I do think this is useful as a defense-in-depth measure (being essentially zero-cost), so long as it's explicitly documented (eg. with a comment) that it's an *additional* safeguard, and should not be relied upon. In particular, things like `javascript:` URLs as the `href` are still risky, and this should mainly serve to eliminate malicious input that survived HTML-escaping and/or URL validation due to a hypothetical implementation bug.

snowteamer:
@joepie91 Though you might not read this until you are back, I wish to thank you for taking the time to write up this thorough and insightful answer.

Things are much clearer now thanks to your advice!

frontend/views/utils/safeLinkTag.js
```diff
@@ -0,0 +1,16 @@
...
+  ['FAQ_PAGE', 'https://groupincome.org/faq]']
+]))
+
+export default function safeLinkTag (key: string): string {
+  if (!Object.prototype.hasOwnProperty.call(urlsByKey, key)) {
+    throw new Error(`Unknown URL key: ${key}`)
+  }
+  // Make sure to include `noopener` and `noreferrer` in the `rel` attribute.
+  return `<a class="link" href="${urlsByKey[key]}" target="_blank" rel="noopener noreferrer">`
+}
```
taoeffect, line 10: Nice, looks bullet-proof to me. 👍

frontend/views/utils/vSafeHtml.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import dompurify from 'dompurify'
+import Vue from 'vue'
+
+// See https://github.com/cure53/DOMPurify#can-i-configure-dompurify
+export const defaultConfig = {
+  // Allow 'href' and 'target' attributes to avoid breaking our hyperlinks,
+  // but keep sanitizing their values.
+  ALLOWED_ATTR: ['class'],
```
taoeffect: I am a bit confused by the comment above here, as it refers to `'href'` and `'target`' attrs, but they don't appear in this list here.

OTOH they are included and merged in the file `translations.js`, so should this comment be moved there?

taoeffect (commented):
Fantastic work @snowteamer! Tested & verified! 😄 🎉 

Consider this comment an approval on the PR and the approach!

Only a minor issue I noticed with a placement of a comment. As you mentioned that you wanted first to OK to approve the approach before modifying other files using this approach, I'm leaving this as a comment instead of marking the PR approved. Let us know on Slack when it's ready for the final approval!

frontend/views/utils/vSafeHtml.js
snowteamer: Indeed, I overlooked this.

Issue #1002, opened 2020-10-14
Document the part of the app that can be viewed as "the framework"


taoeffect:
### Problem

There is a core part of the app that can be viewed as a framework that currently doesn't have documentation, this includes:

- GIMessage
- GIMessage + Vuex + SBP integration
- Contract.js `DefineContract` function
- Frontend <-> Backend communication, including the RESTful routes and websocket stuff

### Solution

Document how all of this works, and in the future perhaps, create an issue to split it out into an actual framework.

The closest thing we have to this currently is the **[Information Flow](https://github.com/okTurtles/group-income-simple/blob/master/docs/Information-Flow.md)** document.

Also, pick a name for the framework. Some [ideas are here](https://gitlab.okturtles.org/okturtles/group-income-simple/-/wikis/E2E-Protocol/Framework.md).

sdegutis:
> The closest thing we have to this currently is the **[Information Flow](https://github.com/okTurtles/group-income-simple/blob/master/docs/Information-Flow.md)** document.

Also tru of the [current SBP docs](https://github.com/okTurtles/group-income-simple/blob/master/docs/Style-Guide.md#sbp).

PR #1071, opened 2021-06-20
New Framework API + E2E Prep

taoeffect:
## Changes

- **NEW:** Chelonia API for defining contracts, creating them, and creating and sending actions, including stubs for managing keys. See `chelonia.js` for details
- **NEW:** SBP features:
  - Domains now have their own state that can be accessed via `this` within all selectors of that domain
  - Special `<domain>/_init` selector now allows "constructor-like" initialization of domains at the moment of first selector registration
- **IMPROVED:** Implementation of `_.debounce` significantly simplified based on underscore version
- **IMPROVED:** Updated `flow-bin` to `0.154.0` version, added type annotations and fixed some bugs it caught
- **IMPROVED:** Update `ws` to `7.5.2` (and closed #1067)
- **IMPROVED:** Thanks to new standardized Chelonia API for creating actions, all code under `frontend/controller/actions` has been greatly simplified with a new `encryptedAction` helper function for easily defining actions in a single line
- **IMPROVED:** Actions no longer require first creating a GIMessage and sending it. These two steps are done for you by Chelonia now.
- **MISC:** beginnings of e2e implementation under `gi.e2e` domain
- **MISC:** re-ran `./strings.mac frontend` to update error strings


backend/pubsub.js
```diff
@@ -115,7 +115,7 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
 
 const defaultOptions = {
   maxPayload: 6 * 1024 * 1024,
-  pingInterval: 30_000
+  pingInterval: 30000
```
CrazedCoding, line 118: Curiously, the former version was actually syntactically correct.

Big fan of "if it ain't broke, don't fix it" approach, but this does actually read nicer.

backend/routes.js
```diff
@@ -3,7 +3,7 @@
 'use strict'
 
 import sbp from '~/shared/sbp.js'
-import { GIMessage } from '~/shared/GIMessage.js'
+import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
```
CrazedCoding, line 6: As mentioned in DM: documentation on Chelonia is necessary.

frontend/controller/actions/group.js
```diff
@@ -10,23 +12,26 @@ import {
...
+    data: {
+      name,
+      picture,
+      sharedValues,
+      mincomeAmount,
+      mincomeCurrency,
+      ruleName,
+      ruleThreshold
+    },
+    options: { sync = true } = {},
```
CrazedCoding, line 32: Exotic syntax from my perspective. What does this part actually do?

options: { sync = true } = {}

CrazedCoding (dismissed):
Looks good, but I'll need to see if it is compatible with our distribution algorithm / PR after a merge into master.

Nothing immediately obvious is broken, but more Chelonia docs would be nice later down the line.

backend/pubsub.js
taoeffect: Well it was broke on my end, rollup would refuse to work unless I made this change. Not sure why

backend/routes.js
taoeffect: Open issue: #1002 :)

frontend/controller/actions/group.js
taoeffect: JS syntax sucks, I agree. It says: "`options` is a key for an optional object parameter that can be passed in, and by default if no object is passed in we set it to an object that looks like this: `{ sync: true }`"

frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -191,19 +191,18 @@ export default {
...
-          this.$store.state.currentGroupId
-        )
-        await sbp('backend/publishLogEntry', groupProfileUpdate)
+          }
+        })
         this.closeModal()
       } catch (e) {
         console.error('IncomeDetails submit() error:', e)
-        this.$refs.formMsg.danger(L('Failed to update income details. {reportError}', LError(e)))
+        this.$refs.formMsg.danger(e.message)
```
snowteamer, line 205: Why no longer translate the displayed error message?

package.json
```diff
@@ -142,14 +142,14 @@
...
     "eslint-plugin-import": "2.22.1",
     "eslint-plugin-node": "11.1.0",
     "eslint-plugin-promise": "4.2.1",
     "eslint-plugin-standard": "5.0.0",
     "eslint-plugin-vue": "7.1.0",
-    "flow-bin": "0.138.0",
-    "flow-remove-types": "2.138.0",
+    "flow-bin": "0.153.0",
+    "flow-remove-types": "2.153.0",
     "form-data": "3.0.0",
```
snowteamer, line 154: Current flow-bin version is already 0.154.0

shared/domains/chelonia/GIMessage.js
```diff
@@ -1,29 +1,29 @@
...
-export type GIKeyType = { type: string; key: string }
+import { blake32Hash } from '~/shared/functions.js'
+import type { JSONType, JSONObject } from '~/shared/types.js'
+
+export type GIKeyType = ''
+
+export type GIKey = {
+  type: GIKeyType;
+  data: Object; // based on GIKeyType this will change
+  meta: Object;
```
snowteamer, line 13: Why not `metadata`?

shared/domains/chelonia/GIMessage.js
```diff
@@ -1,29 +1,29 @@
...
+export type GIOpContract = { type: string; keyJSON: string, parentContract?: string }
+export type GIOpActionEnc = string // encrypted version of GIOpActionUnenc
+export type GIOpActionUnenc = { action: string; data: JSONType; meta: JSONObject }
+export type GIOpKeyAdd = { keyHash: string, keyJSON: ?string, context: string }
 export type GIOpPropSet = { key: string, value: JSONType }
 
-export type GIOpType = 'c' | 'ae' | 'au' | 'ka' | 'kd' | 'ku' | 'pu' | 'ps' | 'pd'
-export type GIOpValue = GIOpContract | GIOpAction | GIOpPubAction | GIOpKeyAuth | GIOpPropSet
+export type GIOpType = 'c' | 'ae' | 'au' | 'ka' | 'kd' | 'pu' | 'ps' | 'pd'
+export type GIOpValue = GIOpContract | GIOpActionEnc | GIOpActionUnenc | GIOpKeyAdd | GIOpPropSet
```
snowteamer: I would prefer unabbreviated action names even if they are longer, especially since `Enc` and `Unenc` are not so obvious.

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+
+// TODO: define ChelContractType for /defineContract
+
+export type ChelRegParams = {
+  contractName: string;
+  data: Object;
+  hooks?: {
+    prePublishContract?: (GIMessage) => void;
+    prePublish?: (GIMessage) => void;
+    postPublish?: (GIMessage) => void;
```
snowteamer: The camel-casing is not needed for pre- and postpublish.

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+  };
+  publishOptions?: { maxAttempts: number };
+}
+
+// eslint-disable-next-line no-useless-escape
+export const ACTION_REGEX: RegExp = /^((([\w.]+)\/([^\/]+))(?:\/(?:([^\/]+)\/)?)?)\w*/
```
snowteamer: Escape backslashes inside character classes (i.e. square brackets) can be removed here

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+// 2 => 'gi.contracts/group'
+// 3 => 'gi.contracts'
+// 4 => 'group'
+// 5 => 'payment'
+
+sbp('sbp/selectors/register', {
+  // https://www.wordnik.com/words/chelonia
+  // https://gitlab.okturtles.org/okturtles/group-income-simple/-/wikis/E2E-Protocol/Framework.md#alt-names
+  'chelonia/_init': function () {
+    this.cfg = {
```
snowteamer: I would find `config` easier to remember than `cfg`

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+      decryptFn: JSON.parse, // override!
+      encryptFn: JSON.stringify, // override!
+      whitelisted: (action: string): boolean => !!this.whitelistedActions[action],
```
snowteamer, line 49: I would name this `isWhitelistedAction()`, keeping the `whitelisted` identifier available to name e.g. the boolean result:
```js
let whitelisted = isWhitelistedAction(action)
...
if (whitelisted) {
  ...
}
```

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+      latestHashSelector: 'backend/latestHash',
+      publishSelector: 'backend/publishLogEntry',
+      skipActionProcessing: false
+    }
+    this.contracts = {}
+    this.whitelistedActions = {}
+    this.sideEffectStacks = {} // [contractID]: Array<*>
+    this.sideEffectStack = (contractID: string): Array<*> => {
+      let stack = this.sideEffectStacks[contractID]
```
snowteamer, line 58: Since Flowtype doesn't distinguish between strings, I would add a guard here to make sure the given string has the shape of a contract ID and is not e.g. an empty string.

frontend/views/containers/contributions/IncomeDetails.vue
taoeffect: This error message comes already translated within `e.message`, which comes from the lower-down `GIErrorUIRuntimeError` string, so there's no reason to build a super-long redundant sentence.

shared/domains/chelonia/chelonia.js
```diff
@@ -0,0 +1,263 @@
...
+      if (!stack) {
+        this.sideEffectStacks[contractID] = stack = []
+      }
+      return stack
+    }
+  },
```
snowteamer, line 64: I usually tend to think of configuration and state objects as plain data objects without methods. This makes working with them overall easier, and allow to serialize or clone them easily if need be.
In this regard, perhaps methods might better be defined on another object, e.g. `this.api` or `this.methods`?

frontend/controller/actions/types.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+// keep in sync with ChelActionParams
+export type GIActionParams = {
+  contractID?: string; // always set on contract actions, but not when creating a contract
+  data: Object;
+  options?: Object; // these are options for the action wrapper
+  hooks?: Object;
+  publishOptions?: Object
+}
```
snowteamer, line 10: Should not this type definition go in `shared/types.js`?

frontend/model/state.js
```diff
@@ -353,6 +353,7 @@ const getters = {
   groupIncomeAdjustedDistributionForMonth (state, getters) {
     return monthstamp => groupIncomeDistribution({ state, getters, monthstamp, adjusted: true })
   },
+  // TODO: this is insane, rewrite it and make it cleaner/better
```
snowteamer, line 356: Agreed the code below is a bit complex. 😄

frontend/model/state.js
```diff
@@ -984,26 +987,28 @@ const handleEvent = {
...
+            proposal = await sbp('gi.actions/group/proposal', {
+              contractID: groupID,
+              data: {
+                proposalType: PROPOSAL_REMOVE_MEMBER,
+                proposalData: {
+                  member: username,
+                  reason: L("Automated ban because they're sending malformed messages resulting in: {error}", { error: error.message })
+                },
+                votingRule: store.getters.groupSettings.proposals[PROPOSAL_REMOVE_MEMBER].rule,
+                expires_date_ms: Date.now() + store.getters.groupSettings.proposals[PROPOSAL_REMOVE_MEMBER].expires_ms
```
snowteamer, line 1008: Could we name this property e.g. `expirationTimestamp` to avoid the snake-casing?

frontend/utils/giLodash.js
```diff
@@ -19,6 +19,12 @@ export function pick (o: Object, props: string[]): {...} {
   return x
 }
 
+export function choose (a: Array<*>, indicies: Array<number>): Array<*> {
+  const x = []
+  for (const idx of indicies) { x.push(a[idx]) }
+  return x
+}
+
```
snowteamer, line 27: A few points here:
- The `a` name is too short; `array` is nicer.
- "indicies" is better spelled as "indices".
- Perhaps disallowing out-of-bound indices would be useful,
  as they could cause `undefined` values to creep in the returned array.
- How should negative indices be handled? Ignore them, throw an exception, or handle them like `.slice()` does?

frontend/views/containers/contributions/IncomeDetails.vue
snowteamer: I couldn't find any translation logic in `frontend/model/errors.js`, and the `.danger()` method doesn't seem to induce any translation either.

So I've tried registering an invalid payment while using the French locale, and the error message was indeed not translated:

Screenshot 2021-07-05 at 22-05-32 Payments

frontend/controller/e2e/keys.js
```diff
@@ -0,0 +1,62 @@
...
+
+function genSeed (): string {
+  return bytesToB64(nacl.randomBytes(nacl.box.secretKeyLength))
+}
+
+function genDeviceSecretFromSeed (
+  { seed, scope, devIdx }: {
+    seed: string,
+    scope: string,
+    devIdx: number
```
snowteamer: What is `devIdx` for? `deviceIndex`?

frontend/controller/actions/types.js
snowteamer: Also, I'm not sure about grouping the `data` and `options` together in the same parameter.
- Like their names suggest, `data` is generally required, while `options` is optional.
- In callsites where only `data` is needed, having to use an extra object nesting level to explicitly specify it (as a property of the input argument) makes the code more verbose than before.

snowteamer (dismissed):
Approved. It is always nice to see core improvements like this!

frontend/controller/e2e/keys.js
taoeffect: Right

frontend/views/containers/contributions/IncomeDetails.vue
taoeffect: Hmm. For this particular error it seems like it might be coming from vuelidate... probably worth opening a separate issue for this.

However for the other errors that might occur, such as anything related to sending the GIMessage itself, `e.message` will contain the translated string from `controller/actions/group.js`.

shared/domains/chelonia/GIMessage.js
taoeffect: because look how beautifully the characters line up together :D

(This type is still a work-in-progress also)

package.json
taoeffect: OK will bump

frontend/model/state.js
taoeffect: I've gone ahead and made some of the other name changes but this one I'm not sure about. I want to make sure the unit that is being used (milliseconds) appears in the name. Open to other suggestions though.

frontend/utils/giLodash.js
taoeffect: OK, changed `a` to `array`, and `indicies` to `indices`.

However, out-of-bounds, and negative indices I'm not going to handle because JS already handles those its own way. I don't want to add duplicate error-handling code because it's redundant and slows down performance.

frontend/controller/e2e/keys.js
taoeffect: Yes, sure, we can change it to that.

CrazedCoding (approved):
Not quite spaghetti code, yet.

Consider removing this PR and replacing with multiple smaller ones.

frontend/controller/e2e/keys.js
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable */
+
+'use strict'
+
+import sbp from '~/shared/sbp.js'
+import { blake32Hash, bytesToB64 } from '~/shared/functions.js'
+import nacl from 'tweetnacl'
+import scrypt from 'scrypt-async'
+
+import type { GIKey, GIKeyType } from '~/shared/domains/chelonia/GIMessage.js'
```
pieer, line 10: GIKey is not used I believe

frontend/model/state.js
```diff
@@ -3,14 +3,14 @@
...
-import type { GIOpContract } from '~/shared/GIMessage.js'
+import type { GIOpContract } from '~/shared/domains/chelonia/GIMessage.js'
 
 import sbp from '~/shared/sbp.js'
 import Vue from 'vue'
 import Vuex from 'vuex'
 import L from '~/frontend/views/utils/translations.js'
 // import incomeDistribution from '~/frontend/utils/distribution/mincome-proportional.js'
-import { GIMessage } from '~/shared/GIMessage.js'
+import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
```
pieer, line 13: Could add GIOpContract here

shared/domains/chelonia/GIMessage.js
snowteamer: Ok, but only because it is still a work in progress. 😄

frontend/views/utils/vSafeHtml.js
```diff
@@ -29,7 +29,7 @@ export const defaultConfig = {
 
 const transform = (el, binding) => {
   if (binding.oldValue !== binding.value) {
-    const config = { ...defaultConfig, ...(binding.arg ?? {}) }
+    const config = { ...defaultConfig, ...(binding.arg || {}) }
```
snowteamer, line 32: It looks like Rollup definetely doesn't like recent JS syntax

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -226,15 +226,20 @@ export default {
             paymentType: PAYMENT_TYPE_MANUAL,
             ...(memo ? { memo } : {}) // TODO/BUG with flowTyper validation. Empty string '' fails.
           }
-          const msg = await sbp('gi.actions/group/payment', paymentInfo, this.currentGroupId)
+          const msg = await sbp('gi.actions/group/payment', {
```
pieer, line 229: This variable is not used after

frontend/views/pages/Join.vue
```diff
@@ -145,8 +145,8 @@ export default {
       sbp('okTurtles.data/set', 'JOINING_GROUP', true)
       try {
         await sbp('gi.actions/group/joinAndSwitch', {
-          groupId,
-          inviteSecret: secret
+          contractID: groupId,
```
pieer, line 148: Are we going from groupId to contractID?

pieer (approved):

frontend/views/pages/Join.vue
taoeffect: The `groupId` and the `contractID` are the same thing, i.e. a `contractID` can be an ID for any kind of contract, and `groupId` is the `contractID` of a specific Group contract that we're joining

PR #1073, opened 2021-06-21

### LINKED ISSUE #975: Check the project for XSS [$250 bounty]

snowteamer:
Closes #975

### Summary of changes:

- Usage of innerHTML assignment in `SendArea.vue` has been replaced by textContent assignment.
- Usage of `rel="noopener noreferrer"` on outgoing hyperlinks can be enforced by creating them with the new safeLinkTag() function.


frontend/views/components/LinkToCopy.vue
```diff
@@ -10,7 +10,6 @@ component.c-wrapper(
     :value='link'
   )
   a.link.has-ellipsis.c-link(
-    :href='link'
```
taoeffect, line 13: Picking up from [this conversation](https://github.com/okTurtles/group-income-simple/pull/1074#discussion_r656689397):

> As far as I've tested, clicking the invite link still copies it, and deleting the href has only two visible effects:
> 
> - disables the hyperlink context menu in favor of the default menu;
> - and disables link URL display at the bottom of the viewport when the link is hovered.
> 
> Furthermore, I guess it doesn't really make sense for this to be a real hyperlink, since it isn't performing any navigation. It's more like a button that has to be stylized like a hyperlink.
For example, trying to open it in a new tab would display the dashboard without copying anything to the clipboard.
> 
> Therefore I have tried to refactor it as a `button.link.has-ellipsis.c-link`, but the ellipsis no longer worked, so I left the `a` tag for now.

OK, I see what you're saying, however, I really think it should be kept since it is designed to look like a link, not a button.

Also, what is the danger here in leaving it? As far as I can tell, this is not a user-generated link, but an app-generated link, and so it should be safe. Therefore it seems to me like what's going on here is dealing with linter errors, than dealing with a real problem. If so, then we should deal with the linter errors somehow without sacrificing useful and expected behavior like being able to right-click on a link and choosing "Copy Link Address".

Just brainstorming here: if the problem is the linter, is there some way we can flag this, perhaps via a directive or something, as being safe? E.g. create a directive like `linter-ignore-safe-link` or `linter-ignore-app-generated-link`?

taoeffect: Per further consideration and discussion with @snowteamer on this over on Slack, we've decided to go ahead with removing the `:href` here.

```
Visually, the invite link still looks like an ordinary link even without the href, thanks to the .link CSS class.Agreed this particular link is safe. But determining this fact requires manual verification, and it would be difficult to do it automatically since it's a generated link (the href is not a literal string but a JavaScript expression).So if we keep allowing such generated links, we'll have to manually review them. That's also a valid choice. After all, there are onle a few of them atm, so we could likely manage them just fine.
However it would go against one of the main goals of this issue, namely automation, since you wrote

>   ... We should not have to police our PRs for this, we should have an automated way of dealing with it.

8:48
Also FWIW, Discord invite links and GitHub repo clone links are not generated as real hyperlinks with href, but as plain text in a readonly text input.
So they have no underline and likely no href attribute.
However the whole URL gets selected on mousedown or touchdown, so that users can easily bring up the context menu if they'd rather copy via that menu.
Browsers recognize the text as a URL and display a suitable popup menu accordingly.
```

frontend/views/utils/translations.js
```diff
@@ -141,8 +159,12 @@ Vue.component('i18n', {
       console.warn('The following i18n text was not translated correctly:', text)
       return h(context.props.tag, context.data, text)
     }
+    // Prevent reverse tabnabbing by including `rel="noopener noreferrer"` when rendering as an outbound hyperlink.
+    if (context.props.tag === 'a' && context.data.attrs.target === '_blank') {
+      context.data.attrs.rel = 'noopener noreferrer'
+    }
```
taoeffect, line 165: Well done 👍

frontend/views/utils/vSafeHtml.js
```diff
@@ -0,0 +1,53 @@
...
+  RETURN_DOM_FRAGMENT: true
+}
+
+const transform = (el, binding) => {
+  if (binding.oldValue !== binding.value) {
+    const config = { ...defaultConfig, ...(binding.arg ?? {}) }
+
+    while (el.firstChild) {
+      el.firstChild.remove()
+    }
```
taoeffect, line 36: I noticed that this code [was removed](https://gitlab.com/gitlab-org/gitlab-ui/-/commit/3f026eb9c3916bd76e711a20021a9985652e98d0) in the original and replaced with `el.textContent = ''` for performance reasons. I also like it because it's a 1-liner too. ~~Can we include that change here too?~~

EDIT: let's do this in a followup PR. I'll open an issue for it. EDIT: here it is: #1075

taoeffect (approved):
Approved!

Excellent job @snowteamer! I noticed that we could incorporate [this change](https://gitlab.com/gitlab-org/gitlab-ui/-/commit/3f026eb9c3916bd76e711a20021a9985652e98d0) from upstream `v-safe-html`, but someone can do that in a followup PR. I'll open up [a minor issue for it](https://github.com/okTurtles/group-income-simple/issues/1075). Don't want to wait any longer to merge these changes.

PR #1074, opened 2021-06-22

### LINKED ISSUE #975: Check the project for XSS [$250 bounty]

snowteamer:
Closes #975

### Summary of changes:

- Usage of innerHTML assignment in `SendArea.vue` has been replaced by textContent assignment.
- Usage of `rel="noopener noreferrer"` on outgoing hyperlinks can be enforced by creating them with the new safeLinkTag() function.


frontend/views/components/LinkToCopy.vue
```diff
@@ -10,7 +10,6 @@ component.c-wrapper(
     :value='link'
   )
   a.link.has-ellipsis.c-link(
-    :href='link'
```
taoeffect, line 13: I'm not sure about this change... did you test it and verify that it works as it should everywhere it's used?

snowteamer: As far as I've tested, clicking the invite link still copies it, and deleting the href has only two visible effects:
- disables the hyperlink context menu in favor of the default menu;
- and disables link URL display at the bottom of the viewport when the link is hovered.

Furthermore, I guess it doesn't really make sense for this to be a real hyperlink, since it isn't performing any navigation. It's more like a button that has to be stylized like a hyperlink.
For example, trying to open it in a new tab would display the dashboard without copying anything to the clipboard.

Therefore I have tried to refactor it as a `button.link.has-ellipsis.c-link`, but the ellipsis no longer worked, so I left the `a` tag for now.

PR #1083, opened 2021-08-15
Switch from Rollup to esbuild

### LINKED ISSUE #1043: Switch from Rollup to Snowpack or esbuild to speed up build & other features [$1k bounty]

### LINKED ISSUE #1069: Figure out why Browsersync isn't refreshing page upon file change [$100 bounty]
### Problem

In Firefox Developer Edition (at least, probably other browsers too), browser sync no longer causes a page refresh when files change even though it appears to detect it in the console:

```
Running "exec:puglint" (exec) task

Running "keepalive" task
created dist/assets/js from main.js in 6.9s
[Browsersync] Reloading Browsers...
[Browsersync] Reloading Browsers... (buffered 134 events)
```

### Solution

Find and fix the problem!

### Bounty

$100 bounty for a clean solution to this (paid in cryptocurrency).

snowteamer:
Closes #1043
Closes #1069

## Summary of changes:

- Use esbuild instead of Rollup to bundle our frontend code.
- Fix progressive cluttering of the build output folder over time.
- Fix BrowserSync not reloading browsers on file change.
- Disable code splitting until related issues are resolved or a workaround is found.
- Drop a few Babel plugins.
- Drop .Gruntfile.babel.js.
- Drop the exec:eslintgrunt task.
- No longer use Babel to transpile our frontend code since esbuild can transpile many ESNext features.
- Refactor the Grunt file to gather config options at the top and to show tasks in lexical order.
- Performance improvements:
  - Build time reduced from 8 min 17s to 6 min 16s.
  - Time to create `dist/assets/js/main.js` reduced from 45.7s to 13.6s.


shared/sbp.js
```diff
@@ -8,6 +8,12 @@ const globalFilters: Array<TypeFilter> = []
 const domainFilters: {[string]: Array<TypeFilter>} = {}
 const selectorFilters: {[string]: Array<TypeFilter>} = {}
 
+// Makes the `Buffer` global available in the browser.
+if (typeof window === 'object') {
+  const { Buffer } = require('buffer')
+  window.Buffer = Buffer
+}
+
```
taoeffect: Hmm, this belongs somewhere else, as `sbp.js` should only be used for bare-minimum SBP setup.

Gruntfile.js
```diff
@@ -5,58 +5,400 @@
...
+
+const backendIndex = './backend/index.js'
+const distAssets = 'dist/assets'
+const distCSS = 'dist/assets/css'
+const distDir = 'dist'
+const distJS = 'dist/assets/js'
+// const serviceWorkerDir = 'frontend/controller/serviceworkers'
+const srcDir = 'frontend'
+
+const chompLeft = (s, what) => s.startsWith(what) ? s.slice(what.length) : s
```
taoeffect, line 67: Why is this being rewritten instead of imported as before?

scripts/esbuild-vue-plugin.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+let requireDepth = 0
+let usedFiles = new Set()
+
+editModule('fs', (fs) => {
+  fs.readFileSync = new Proxy(fs.readFileSync, {
+    apply (target, thisArg, args) {
+      if (usedFiles && requireDepth === 0) usedFiles.add(args[0])
+      return Reflect.apply(target, thisArg, args)
```
taoeffect: Interesting, I've never seen `Reflect.apply` before even though I've been coding JS for many years. Crazy!

I'm curious, what is the reason you chose to do this instead of `target.apply`?

Oh - this is probably not your code, right? I have no problem with this btw, just curious.

taoeffect (changes requested):
Fantastic PR @snowteamer! This must've been quite a lot of in-depth research on your part, really impressive!

I went ahead and did a preliminary review. I also tested it out locally and at least on first build it works great -- however, on I wasn't able to get it to make any subsequent builds. Saving files wouldn't get it to recompile. At least on my machine I saw no notification of changed files in the terminal, and the browser didn't refresh. Can you investigate that to see if you can get it working similarly to how it worked before?

shared/sbp.js
snowteamer: That's right, however I couldn't put it in `main.js` as no code is allowed before import statements, and I didn't really want to create a new file just for this.

Gruntfile.js
snowteamer: Importing wouldn't work here as babel is not loaded yet at this point.

scripts/esbuild-vue-plugin.js
snowteamer: Yea that code comes from the original plugin, namely from https://github.com/apeschar/esbuild-vue/blob/master/src/worker.js

Gruntfile.js
snowteamer: Anyway, `chompLeft()` is only used in the Sass plugin `importer` option, which has no effect for some reason (it looks like the Vue plugin takes precedence over the Sass one except for the `main.scss` file), so we might remove it later.

shared/sbp.js
taoeffect: OK, so, here's what I recommend:

1. Create a file just for this called buffer.js or something with the word "buffer" in it.
2. Check to see if Buffer already exists. If so, do not do anything. This is so that if we ever switch back to rollup or something else, and that something else already includes Buffer, we won't import it twice.
3. Import it above sbp.js in main.js

taoeffect: *Actually, why do you  have to import it before sbp.js?

Just import it afterward like all the other imports. Import it near the top. sbp.js should be the first import always unless there's something that won't work otherwise.

Gruntfile.js
taoeffect: ah I see. Can you add a comment to that effect above it?

shared/sbp.js
snowteamer: Would it be acceptable to put this snippet in `~/shared/functions.js`? This file defines a lot of buffer functions, and it could avoid the need of an extra file

taoeffect: Sure!

shared/functions.js
```diff
@@ -5,6 +5,12 @@ import multihash from 'multihashes'
 import nacl from 'tweetnacl'
 import blake from 'blakejs'
 
+// Makes the `Buffer` global available in the browser.
+if (typeof window === 'object') {
+  const { Buffer } = require('buffer')
```
taoeffect, line 11: Remember, per previous review, need to check to see if `Buffer` is already loaded.

snowteamer: I think the `require()` call already does that internally, and will not load a given module twice unless its entry in `require.cache` has been deleted.

taoeffect: @snowteamer I'm not so sure. The thing is `Buffer` can be defined by something other `require('buffer')`. For example, the way we previously included it is via `'rollup-plugin-node-globals'`, so in that case I don't think `require` would catch it. Think you need to check whether `Buffer` already exists, e.g. via `globals.Buffer` or `window.Buffer` or `typeof Buffer !== 'undefined'` (that last one might make the most sense)

snowteamer: Ok, it makes sense in that case

snowteamer: @taoeffect Is the following snippet suitable?
```js
// Makes the `Buffer` global available in the browser if needed.
if (typeof window === 'object' && !window.Buffer) {
  if (typeof Buffer === 'undefined')
    // Only import `Buffer` to hopefully help treeshaking.
    const { Buffer } = require('buffer')
    window.Buffer = Buffer
  } else {
    window.Buffer = Buffer
  }
}
```

taoeffect: @snowteamer Looks ok, can be shortened I think:

```js
// Makes the `Buffer` global available in the browser if needed.
if (typeof window === 'object' && !window.Buffer && typeof Buffer === 'undefined') {
  window.Buffer = require('buffer').Buffer
}
```

The shorter the better (usually) :)

snowteamer: Sure! :smile:
I was thinking about the case where the global `Buffer` has been defined elsewhere but not yet assigned to `window`; however in a browser context, that assignment would have already been performed by the browser itself I think.

So if you don't mind, I will just keep the `const Buffer = ...` declaration in case it helps treeshaking in the future (that would be nice because the `buffer` module is actually quite big, yielding more than 1k LOC in `dist/main.js`).

taoeffect: @snowteamer Yeah np 👍

frontend/assets/style/_icons.scss
```diff
@@ -81,8 +80,7 @@ $icons: (
     url("../fonts/Icons/fa-regular-400.eot?#iefix") format("embedded-opentype"),
     url("../fonts/Icons/fa-regular-400.woff2") format("woff2"),
     url("../fonts/Icons/fa-regular-400.woff") format("woff"),
-    url("../fonts/Icons/fa-regular-400.ttf") format("truetype"),
-    url("../fonts/Icons/fa-regular-400.svg#fontawesome") format("svg");
```
taoeffect, line 85: @snowteamer can you help me understand this change?
@pieer is this safe?

taoeffect (approved):
Approved! Excellent work @snowteamer! 😄 🎉 

I'm gonna hold off on merging tho just to double-check with your and/or @pieer that this change is safe: https://github.com/okTurtles/group-income-simple/pull/1083/files#r693405654

frontend/assets/style/_icons.scss
snowteamer: For some reason I couldn't bundle the app with SVG fonts URLs containing a fragment (the part after the `#` sign).
I was hesitant to remove them, but after some research it turned out that even old IE browsers should be able to handle at least the .eot font format, and that woff and woff2 are much more efficient anyway.

More precisely, according to https://www.caniuse.com/svg-fonts and https://www.caniuse.com/woff2, some versions of Safari and Safari & Chrome for iOS do still support SVG fonts; but since they also support the better WOFF formats, then providing SVG fonts does not seem necessary.

taoeffect: Fantastic reasoning. Thank you @snowteamer. I'm gonna go ahead and merge your PR now. @pieer if you disagree let us know and if necessary we can re-add in another PR.

pieer: Sounds good to me

PR #1086, opened 2021-08-25
Optimize rebuilds

### LINKED ISSUE #1007: Windows `grunt dev` pug-lint-vue Error
### Problem

I followed the "Getting Started" instructions on my Windows 10 machine, and ended up getting the following error when attempting to run `grunt dev`:

image

It seems to have something to do with the behavior of the `grunt-puglint` feature.

### Solution

First, do research into OS-dependent behavior of `pug-lint-vue`. Once the difference is understood, then suggest a more detailed solution.

### LINKED ISSUE #1084: Optimize rebuild times [$400 bounty]
### Problem

Although build times have been improved by switching from Rollup to Esbuild, application rebuild times after changing a file are not much better, or even slightly worse than before.

### Solution

Improve rebuild times to make them at least on par with the previous situation.
Better yet: optimize to make them take no longer than a few seconds.

Hints:
- Esbuild doesn't cache results from its plugin calls, so plugins should do it themselves.
- Esbuild is optimized for code transpilation and bundling rather than tranformation through plugins. Therefore, plugin usage should be reduced as much as possible.
- Linting and other checking operations on changed files could be sped up by using JavaScript APIs on the relevant files instead of spawning processes via `grunt exec`.

### Bounty

$400 for a clean solution (paid in cryptocurrency).

snowteamer:
Closes #1084
Closes #1007 (by integrating our `pug-lint-vue` fork)

### Summary of changes:
- Bump Node and NPM to latest v16 and v7 respectively.
- Use our `pug-lint-vue` GitHub fork instead of the NPM dependency.
- In watch mode, saving a file with lint errors no longer crashes the Grunt process or the Hapi server.
### Tasks:
- [x]  Lint new files and changed files using available linter API's instead of spawning new processes via `grunt:exec`.
- [x]  Improve caching.


taoeffect (changes requested):
PR looks really good @snowteamer 👍 

Something bugging me though is the massive change to `package-lock.json` even though on a single plugin changed.

So I created a PR just to prevent that file from changing (plus added some `npm audit fix` stuff): https://github.com/okTurtles/group-income-simple/pull/1092

That PR needs your review/approval. After it's merged, can you please merge `master` back into this PR, and have the `package-lock.json` from `master` override entirely the one here, so that there is no diff on that file?

Gruntfile.js
```diff
@@ -340,29 +368,69 @@ module.exports = (grunt) => {
...
+              vuePluginOptions.cache.clear()
+            }
+          }
           // Only rebuild the relevant entry point.
           if (filePath.startsWith(serviceWorkerDir)) {
             await buildServiceWorkers.run()
           } else {
             await buildMain.run()
           }
+          grunt.task.run(tasks.filter(task => typeof task === 'string'))
```
taoeffect, line 431: Hmm, previously this line would run before this `if` block. Should it be moved up?

snowteamer: I've moved this line after the `if` block so that tasks like `backend:relaunch` do not delay rebuilds and only run if the rebuild was successful

snowteamer: Although I might have made a mistake here. If a `./backend` file was changed, perhaps we want the backend server to reload before the browser refreshes

taoeffect (approved):
Fantastic work @snowteamer! Rebuild speeds have been significantly improved!! 🥇 😄 

Approved!

PR #1098, opened 2021-09-14
Serve translation files like other assets

### LINKED ISSUE #1030: Cache translations / localizations
### Problem

Currently, each time the page loads it must request the fairly large [local json](https://github.com/okTurtles/group-income-simple/blob/master/strings/english.json) (at the moment on the order of 38kb).

### Solution

Explore ways of *not* loading this file.

There are two main approaches:

- **Using caching headers, the traditional way.** Check to see if the Hapi server is using proper headers to prevent the client from downloading the json each time. It might be that browsers can automatically cache this stuff for us without having to write any code. It might even already do that. I just haven't checked.
- **Use ServiceWorker caching.** This is an advanced technique that manually stores the json locally using the [`Cache`](https://developer.mozilla.org/en-US/docs/Web/API/Cache) API of ServiceWorkers. This requires manually storing it in the cache, keeping track of new versions of the json, etc.

To summarize: this might not be a real issue. Someone needs to verify what the current behavior is with the Hapi.js server and see if the traditional approach suffices for our needs.

We want the app to work well offline, and for this we might need to use the ServiceWorkers API. See related issue: #590

snowteamer:
~~Clxxxx #1030~~ (EDIT by @taoeffect: I don't think this closes the issue because it doesn't modify any headers)

### Summary of changes:
- Translation-related JSON files (e.g. `french.json`) are now copied to `dist/assets/strings` at build time so that they can be served via the the `/assets` route like any other asset.
- In particular, these files now benefit from the same cache policy as other assets.
- The no longer used `/translations` route has been removed.
- The whole `backend/translations.js` file could also be deleted.


frontend/controller/backend.js
```diff
@@ -13,6 +15,11 @@ import { PUBSUB_INSTANCE } from './instance-keys.js'
 // const nacl = require('tweetnacl')
 import nacl from 'tweetnacl'
 
+// Used in the 'backend/translations/get()' SBP function.
```
taoeffect: `get()` shouldn't have a `()`, this should read simply:

```
// Used by 'backend/translations/get'
```

Gruntfile.js
```diff
@@ -228,6 +228,11 @@ module.exports = (grunt) => {
         src: ['**/*', '!style/**', '!svgs/**'],
         dest: distAssets,
         expand: true
+      },
+      strings: {
+        src: ['strings/*.json', '!strings/english.json'],
+        dest: distAssets,
+        expand: true
```
taoeffect: Why is `english.json` being excluded here? I see no reason not to copy it just in case.

frontend/controller/backend.js
```diff
@@ -111,8 +118,22 @@ sbp('sbp/selectors/register', {
...
+   *
+   * @param language - A BPC-47 language tag like the value
+   * of `navigator.language`.
+   *
+   * @see The 'translations/init' SBP selector in `~view-utils/translations.js`.
+   */
+  async 'backend/translations/get' (language: string): Promise<?JSONObject> {
+    // The language code is usually the first part of the language tag.
+    const [languageCode] = language.toLowerCase().split('-')
+    const languageFileName = languageFileMap.get(languageCode) || ''
```
taoeffect, line 133: What's the need for ` || ''` at the end here?

taoeffect (changes requested):
Nice optimization @snowteamer! Only noticed some minor things.

Are you sure though that this closes #1030? Some things to check:

- Did you test the caching policy locally and observe via the browser tools that the files (in this case, `french.json`) were no longer being requested after they were successfully loaded the first time?
- Did you verify that if the file `french.json` is updated with new translations, that it *is* loaded anew and that the new translated string appears on the page?

Gruntfile.js
snowteamer: Since `english.json` is the default file, it kinda means "no translation at all": all its values are identical to its keys. Therefore correct translation code should not need it.

frontend/controller/backend.js
snowteamer: It allows `languageFileName` to always be a string by disallowing the `undefined` value.
This can be removed, I just prefer to avoid mixing types when it's not necessary.

frontend/controller/backend.js
snowteamer: Ok :+1:

Gruntfile.js
taoeffect: It's true that at the moment `english.json` isn't being used. However, that doesn't necessarily mean that it won't ever be used at any point in the future. When we write code with exceptions like this, it creates traps / time-bombs for the future if the code changes at any point in time to need it. In this case, such an exception isn't needed. It doesn't save us anything, and instead creates cognitive overhead by creating an exceptional circumstance. The app won't be slower if we include this file, nor will it break anything, and it will result in less logic if you remove the exception, so I think it's a good idea to copy over that file and leave the logic simple.

frontend/controller/backend.js
taoeffect: I see. It's a quirk of JavaScript that `''` is falsey. In other languages the `if` would pass... so that's the only reason why I'd consider removing it. Not a strong reason, so can leave as is.

Gruntfile.js
snowteamer: The main intent was to avoid copying one more .json file on every build, but I agree it is no big deal

frontend/controller/backend.js
snowteamer: We can be even more explicit by writing `languageFileName !== ''` in the `if` clause rather than relying on implicit boolean conversion.

snowteamer:
@greg I did run some tests locally, but I still might have missed something.

Gruntfile.js
snowteamer: I also want to make sure that file won't be downloaded by the browser unless we actively use it

taoeffect: > I also want to make sure that file won't be downloaded by the browser unless we actively use it

OK, that part makes sense. It's still OK to copy the file though so that it's available if we decide we want it for some reason.

frontend/controller/backend.js
taoeffect: Totally up to you. I'm ok with it either way, just was curious about why you didn't rely on the `undefined`.

taoeffect (commented):
Did you look at the existing headers and conclude that no changes needed to be made?

taoeffect (approved):
Approved! Thanks @snowteamer, this is definitely a good PR to merge because it DRYs our code with respect to serving assets. This is really how it should have been implemented in the first place, but I don't think it closes the issue #1030, because no headers are modified. We need to research and determine the proper headers that should be served for these files.

Copying from Slack some notes:

- All assets served under `/file/{hash}` need headers that indicate the files don't change and can be strongly cached
- Assets served by `/assets/` can change, and in fact probably will change every update. The ideal caching method would cache them in between updates, and invalidate the cache when the files change.

Issue #1085, opened 2021-08-22
Make our build system agnostic to bundlers


taoeffect:
### Problem

PR #1083 is going to switch us to esbuild, and with this it will significantly improve the first-build time and Travis CI builds.

However, subsequent builds on file change are slower than rollup (our previous build system). Furthermore, it might be the case that our current bundler becomes no longer maintained, or a nicer build tool comes along that we want to switch to, or we decide we want to switch back to rollup for some reason. This will be difficult to do with the way #1083 is currently implemented.

### Solution

- Modify the build system so that it is trivial to build using a different bundler (provided the correct dependencies are setup in `package.json` - by default only 1 build system should be supported at a time).
- To prove that the design supports adding another build system by implementing our previous rollup setup using this new modular design (but in the PR, do not actually add back the rollup dependencies - just test it out locally to verify).

Example bundlers:

- [Rollup](https://rollupjs.org/guide/en/)
- [Parcel](https://github.com/okTurtles/group-income/issues/1107)
- [Vite](https://vitejs.dev/guide/why.html)
- [Esbuild](https://esbuild.github.io/)

### Bounty

~~$300 bounty for a clean solution to this (paid in cryptocurrency).~~

There was a $300 bounty on this, but it was created at a time before LLMs. Now if we're going to accept an LLM PR for this I would rather it be done either in-house by someone on our team, or if a volunteer is interested, that they please join [our Slack](https://join.slack.com/t/okturtles/shared_invite/zt-10jmpfgxj-tXQ1MKW7t8qqdyY6fB7uyQ) and let us know they're interested in working on this so we can have a human-to-human relationship as opposed to a human-to-claw relationship.

abdulazizali77:
@taoeffect Greg, am interested in taking this up/investigating this please assign me :)

taoeffect:
Here's another system to consider: https://vitejs.dev/guide/why.html

Updated the issue with examples of other bundling systems.

PR #1100, opened 2021-09-16
Fix new CSS assets not being used after a rebuild

### LINKED ISSUE #1099: New CSS assets are not used after a rebuild
### Problem

After changing a .scss file, new versions of `main.css` and `main.css.map` are created and and output to `main.js`'s directory.
However, they remain unused since our `index.html` tries to load them from `dist/assets/css`, where only outdated versions are available.

### Solution

Some ideas:
- Make esbuild redirect CSS asset output to the correct directory.
- Implement logic to move the relevant files after a rebuild occured.
- Change `index.html` to load both `main.css` and `main.js` from the same directory.

snowteamer:
Closes #1099

Fixes the relevant issue by copying new `main.css` and `main.css.map` over to the correct location.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -44,7 +44,7 @@ Object.assign(process.env, applyPortShift(process.env))
 const chalk = require('chalk')
 const crypto = require('crypto')
 const { exec, fork } = require('child_process')
-const fs = require('fs')
+const { copyFile, readFile } = require('fs/promises')
 const path = require('path')
 const { resolve } = path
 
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -427,6 +426,14 @@ module.exports = (grunt) => {
             await buildServiceWorkers.run()
           } else {
             await buildMain.run()
+            if (['.css', '.sass', '.scss'].includes(extension)) {
+              // Esbuild outputs `main.css` and `main.css.map` along `main.js`,
+              // but `index.html` wants to load them from `dist/assets/css`.
+              await fs.promises.copyFile(`${distJS}/main.css`, `${distCSS}/main.css`)
+              if (buildMain.options.sourcemap) {
+                await fs.promises.copyFile(`${distJS}/main.css.map`, `${distCSS}/main.css.map`)
+              }
```
taoeffect: This seems very granular of a way to do it, very specific to esbuild as well, which will make it more difficult to close #1085, I think. Have you considered doing it using the existing `grunt-contrib-copy` stuff that we have, or in a more generic way that is less specific to esbuild integration?

taoeffect (commented):
Good catch @snowteamer. I left a question.

Gruntfile.js
snowteamer: Indeed, and it also introduces a bit of code duplication since the same copying operation occurs on the first build.

So I've considered specifying this post-build operation as a custom option, then thought it was better to keep things simple for now; but I should be able to add it if you don't mind the extra option.

I it is also possible via `grunt copy`, but I think it would still be as much specific to esbuild as plain code, and likely a bit slower. I'm also slightly reluctant to increase our dependency to Grunt since we might want to switch to Snowpack later.

taoeffect: Would the post-build option would help with #1085 and DRY the code more?

If so, then I think it's worth doing.

snowteamer: I think so since this copying operation might not be needed with other build tools, or they might require doing something else.

taoeffect (changes requested):
Per the convo here: https://github.com/okTurtles/group-income-simple/pull/1100/files#r710204644

Implement the copying part using a post-build option to DRY things and help with #1085

scripts/esbuild-commands.js
```diff
@@ -3,32 +3,51 @@
...
+    async rerun ({ fileEventName, filePath }) {
+      const { state } = this
+
+      if (!state.result || !esbuildOptions.incremental) {
+        throw new Error('Cannot use rerun() on this task. Use run() instead.')
+      }
+      await state.result.rebuild()
+
+      if (otherOptions.postoperation) {
+        await otherOptions.postoperation({ fileEventName, filePath })
```
taoeffect: These two functions, `run` and `rerun`, are nearly identical except for the difference between:

```js
      state.result = await esbuild.build(esbuildOptions)
```

And:

```js
      await state.result.rebuild()
```

Couldn't therefore the logic be simplified so that there is only a single `run` function that keeps track of whether or not it has already been called before, and proceeds accordingly?

snowteamer: There was indeed a single `run()` method before, which used `state.result` to check whether it had already been called.

Since additional arguments must now be accepted when a rebuild occurs, I've been considering whether `run()` should be split in two distinct methods with different signatures, rather than a single catch-all method which accepts optional arguments.
Besides, esbuild itself also has both `build()` and `rebuild()` methods.

It's a tradeoff between two simpler functions and a more complex one.
So I've tried this approach somewhat as an experiment, but I don't mind also trying the monolithic method.

taoeffect: In your estimation, which approach would make more sense from the perspective of creating a build system that is agnostic to the underlying bunder? (i.e., issue #1085)?

snowteamer: Both should be roughly equivalent in that respect at first guess (I might be wrong though).
In any case, issue #1085 will likely involve several refactorings.

taoeffect (approved):
Approved! Great catch @snowteamer! (My bad for not catching it originally!)

PR #1103, opened 2021-10-27
Add Flow support to Vue files

### LINKED ISSUE #273: Add Flow support to .vue files
### Problem

Our static analyzer currently doesn't analyze the JS in .vue components, this can lead to bugs not being caught.

### Solution

[Writing Vue.js Components with Flow](https://alligator.io/vuejs/components-flow/)

snowteamer:
Closes #273

#### Pros:
- Flowtype syntax can now be used in Vue components without preventing the app to build.
- New `npm run flow:vue` command available to typecheck our .vue files.
#### Cons:
- New dependency [eslint-plugin-flowtype-errors](https://github.com/amilajack/eslint-plugin-flowtype-errors) installed.
- Travis build time slightly increased by ~2s because of the new syntax support, although no additional typechecking is done by default. Rebuilds might also be a bit slower, something like a tenth of a second per .vue file changed.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -225,7 +225,8 @@ module.exports = (grunt) => {
     // This map's keys will be relative Vue file paths without leading dot,
     // while its values will be corresponding compiled JS strings.
     cache: new Map(),
-    debug: false
+    debug: false,
+    flowtype: flowRemoveTypesPluginOptions
   }
 
   grunt.initConfig({
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,7 @@
         "eslint-config-standard": "16.0.2",
         "eslint-plugin-cypress": "2.11.3",
         "eslint-plugin-flowtype": "5.7.2",
+        "eslint-plugin-flowtype-errors": "4.4.0",
         "eslint-plugin-import": "2.22.1",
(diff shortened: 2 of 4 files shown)


taoeffect (approved):
This is some very clean code, great job @snowteamer!

I ran `npm run flow:vue` locally and it does seem to run flow on the vue files. There is one issue though, which is that while flow runs on the files, it basically prints out the same error a thousand times, things along these lines:

```
  73:28  error  Cannot resolve module `@svgs/create-group.svg`. [cannot-resolve-module]                                                                                                                flowtype-errors/show-errors
  74:26  error  Cannot resolve module `@svgs/join-group.svg`. [cannot-resolve-module]                                                                                                                  flowtype-errors/show-errors
  88:16  error  Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type annotation at array pattern: [signature-verification-failure]  flowtype-errors/show-errors
```

So it never really gets to actually analyzing the files for bugs, and worse still, it for some reason tries to lint files that it shouldn't even be looking into, e.g. `.svg` files. I feel like these should be added to `.flowconfig` so that they're ignored.

I'm going to approve the PR anyway though, because it's a big step forward. And then I'll create another issue to address the actual linting of the files with flow.

PR #1105, opened 2021-10-29
Cleanup trivial Flow errors in Vue components

### LINKED ISSUE #1104: Get Flow to generate meaningful linter results on Vue files [$50 bounty]
### Problem

Continuing from Issue #273 (and corresponding PR #1103) - thanks to @snowteamer we've got flow running on Vue files, however right now it doesn't actually lint them because before it gets to that part it isn't able to deal with the imports that we're using, so it generates thousands of errors like these:

```
  73:28  error  Cannot resolve module `@svgs/create-group.svg`. [cannot-resolve-module]                                                                                                                flowtype-errors/show-errors
  74:26  error  Cannot resolve module `@svgs/join-group.svg`. [cannot-resolve-module]                                                                                                                  flowtype-errors/show-errors
  88:16  error  Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type annotation at array pattern: [signature-verification-failure]  flowtype-errors/show-errors
```

### Solution

A proper solution to this doesn't need to actually fix all flow errors that the vue files might generate - but instead it needs to get to the point where it's actually showing them and not generating the sorts of errors that have to do with flow setup.

So:

- Ignore in `.flowconfig` all files that shouldn't be linted (e.g. `.svg`s like in the example above)
- Fix all remaining import-related errors, and any other errors that prevent the actual errors from showing

### Bounty

$50 bounty for a clean solution to this (paid in cryptocurrency).

snowteamer:
Closes #1104

#### Pros:
- The default export of every .vue file has been annotated with `: Object` so that Flow no longer complains about untyped module interfaces. A more specific type could be used in the future though.
- The Flow config file has been updated to use the `module.name_mapper.extension` option, which makes no longer necessary to update `shared/declarations.js` every time a .vue file is added, removed or renamed.
- The `shared/declarations.js` file has also been updated and greatly simplified thanks to the above option.
#### Cons:
- A small `vueComponentStub.js.flow` file has been added in the `@view-utils` directory to act as a stub for Vue component imports. The  `module.name_mapper.extension` option tells Flow's module resolution to use this stub when encountering a  `*.vue` or `*.svg` import statement.

### DIFF
--- a/frontend/views/containers/proposals/RemoveMember.vue
+++ b/frontend/views/containers/proposals/RemoveMember.vue
@@ -112,7 +112,7 @@ export default ({
         })
         this.$refs.proposal.close()
       } catch (e) {
-        console.error(`Failed to remove member ${member}.`, e)
+        console.error('Failed to remove member %s:', member, e.message)
         this.$refs.formMsg.danger(e.message)
       }
     }
--- a/frontend/views/containers/proposals/RemoveMember.vue
+++ b/frontend/views/containers/proposals/RemoveMember.vue
@@ -25,7 +25,7 @@ import { PROPOSAL_REMOVE_MEMBER } from '@model/contracts/voting/constants.js'
 import BannerScoped from '@components/banners/BannerScoped.vue'
 import ProposalTemplate from './ProposalTemplate.vue'
 
-export default {
+export default ({
   name: 'RemoveMember',
   components: {
     Avatar,
@@ -117,7 +117,7 @@ export default {
       }
     }
   }
-}
+}: Object)
 </script>
 <style lang="scss" scoped>
 @import "@assets/style/_variables.scss";
--- a/frontend/views/utils/vueComponentStub.js.flow
+++ b/frontend/views/utils/vueComponentStub.js.flow
@@ -1,0 +1,3 @@
+const x: Object = {}
+export default x
+
--- a/.flowconfig
+++ b/.flowconfig
@@ -10,7 +10,6 @@
 .*/dist/.*
 .*/frontend/assets/.*
 .*/frontend/controller/service-worker.js
-.*/frontend/controller/utils/primus.js
 .*/frontend/utils/blockies.js
 .*/frontend/utils/crypto.js
 .*/frontend/utils/flowTyper.js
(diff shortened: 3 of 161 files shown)


frontend/views/containers/proposals/RemoveMember.vue
```diff
@@ -112,12 +112,12 @@ export default {
         })
         this.$refs.proposal.close()
       } catch (e) {
-        console.error(`Failed to remove member ${member}.`, e)
+        console.error('Failed to remove member %s:', member, e.message)
```
taoeffect, line 115: @snowteamer Am curious why this change was made?

snowteamer: @greg Because of a CodeQL alert which was blocking the PR. It was marked as "critical" although the only potential risk seemed to be garbled console output.

frontend/views/utils/vueComponentStub.js.flow
```diff
@@ -0,0 +1,3 @@
+const x: Object = {}
+export default x
```
taoeffect, line 2: @snowteamer Is it possible to mess with either this file or with the `.flowconfig` to prevent having to add `: Object)` everywhere?

For example, did you try adding that here instead?

```js
export default (x: Object)
```

or

```js
export default ({}: Object)
```

Or something to that effect?

snowteamer: I don't think so, as this file only acts as a stub for import statements, not for exports. Upon encountering a `*.vue` import statement, Flow will use this stub instead of the Vue file (which might even not exist) and assume the resulting type must be `Object`.

Therefore the config file only allows to mess with imports, not exports.

Flow always requires explicit typing at so-called module boundaries (i.e. module exports).

Quoted from the [Flow blog](https://medium.com/flow-type/asking-for-required-annotations-64d4f9c1edf8):

> #### Why are annotations required?
>
> Flow has very few limitations on its type inference, but there is one big one: input positions reachable from exports must be annotated. An inferred type in such a position is insufficient, as it would not account for a value that might “flow in” from a dependent module. Indeed, these types are often never constrained by the defining module.

taoeffect (approved):
Alright, we'll go with this method - great job @snowteamer!

PR #1112, opened 2021-11-24
Correct index to select user inside the pay list

### LINKED ISSUE #1111: Find out why cypress test case breakes
### Problem

The cypress test case *`test/cypress/integration/group-paying.spec.js`* breaks. Seems like an incorrect index to select user.

### Solution

Find out the reason and fix so that it won't make error again!

Silver-IT:
Closes [#1111](https://github.com/okTurtles/group-income-simple/issues/1111)

Have found the incorrect index to select user and fixed it.

### DIFF
--- a/test/cypress/integration/group-paying.spec.js
+++ b/test/cypress/integration/group-paying.spec.js
@@ -183,9 +183,9 @@ describe('Group Payments', () => {
     cy.getByDT('link-PaymentRowSent').click()
     cy.getByDT('payList').find('tbody').children().should('have.length', 2)
     cy.getByDT('payList').within(() => {
-      cy.getByDT('payRow').eq(0).find('td:nth-child(1)').should('contain', `user3-${userId}`)
-      cy.getByDT('payRow').eq(0).find('td:nth-child(2)').should('contain', '$100')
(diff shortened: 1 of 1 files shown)


test/cypress/integration/group-paying.spec.js
```diff
@@ -183,9 +183,9 @@ describe('Group Payments', () => {
     cy.getByDT('link-PaymentRowSent').click()
     cy.getByDT('payList').find('tbody').children().should('have.length', 2)
     cy.getByDT('payList').within(() => {
-      cy.getByDT('payRow').eq(1).find('td:nth-child(1)').should('contain', `user3-${userId}`)
-      cy.getByDT('payRow').eq(1).find('td:nth-child(2)').should('contain', '$100')
-      cy.getByDT('payRow').eq(1).find('td:nth-child(3)').should('contain', humanDateToday)
```
taoeffect: [`.eq(1)` returns the second DOM element](https://docs.cypress.io/api/commands/eq) from the list of elements returned by the previous selector... I don't think changing that to `1` would select the second row. However, per [your screenshot here](https://user-images.githubusercontent.com/57976479/143189896-85efd93c-bf75-4a99-a2a7-0c3e29cbc921.png), it's the first row that you want selected.

taoeffect:
As #1111 was closed I'm going to close this.

PR #1113, opened 2021-11-29
Correct mistake of uploaded file path

Silver-IT:
Fixed
PS: it is not mobile platform issue. This happens when we use server hosted online (not local).
The reason of this issue is that there is a mistake to set uploaded file path.


taoeffect (approved):
Great work @Silver-IT!

PR #1106, opened 2021-11-02
Fix most NPM audit alerts

snowteamer:
### Summary of changes:
- Remove unused local grunt (we are only using a globally installed grunt).
- Bump browser-sync to 2.27.7.
- Uninstall unused cheerio and chokidar (however they are still there as sub-dependencies).
- Bump mocha to v8.4.0.
- Bump eslint to v7.32.0.
- Remove eslint-plugin-standard as suggested in the following notice:
> npm WARN deprecated eslint-plugin-standard@5.0.0: standard 16.0.0 and eslint-config-standard 16.0.0 no longer require the eslint-plugin-standard package. You can remove it from your dependencies with 'npm rm eslint-plugin-standard'. More info here: https://github.com/standard/standard/issues/1316
- Replace babel-eslint with @babel/eslint-parser as per notice:
> npm WARN deprecated babel-eslint@10.1.0: babel-eslint is now @babel/eslint-parser. This package will no longer receive updates.
- Bump eslint-plugin-vue to v7.20.0

- The number of audit alerts has been reduced from 37 (13 moderate, 19 high, 5 critical) to 12 (12 high).
This is still quite a lot, however they all seem to arise from only two packages, namely `set-value` and `union-value`. From the dependency graph displayed by `npm audit` it looks like fixing these remaining issues would involve either patching `check-dependencies' or forking `grunt-check-dependencies', in order to update their vulnerable `findup-sync' sub-dependency.

PR #1114, opened 2021-11-29
re-add local grunt

taoeffect:
Removed in #1106, didn't catch that this broke a fresh clone. /cc @snowteamer


snowteamer (approved):
LGTM

Good catch, I incorrectly thaught that a local Grunt package was no longer necessary once the Grunt CLI has been installed globally.

Issue #1117, opened 2021-12-03
Several UI issues especially in mobile view.


Silver-IT:
### Problem

* Group picture and description are overlapped
- [x] Fixed and the commit is inside the [PR 1121](https://github.com/okTurtles/group-income-simple/pull/1121)

![image](https://user-images.githubusercontent.com/57976479/144561853-2d4342cb-b36f-4ab2-87a1-5ea7b413050e.png)
* Drawer is overflowed and `Logout` button is hidden
![image](https://user-images.githubusercontent.com/57976479/144561873-6ae15bd2-2216-4ddb-ab58-24d5a5eee721.png)
* Can not open the drawer in mobile landscape view
![image](https://user-images.githubusercontent.com/57976479/144561966-4d7a63df-fe75-43b1-b462-939ad1e513a0.png)

### Solution

*I think we should remove fixed values such as `padding`, `margin`, `left`, `top`, ... and make it responsive using `display: flex`, `overflow: auto`, ...*

Silver-IT:
New issues found.

* Sections inside modal are overlapped in mobile view.
![image](https://user-images.githubusercontent.com/57976479/144680684-631c12fe-c709-4999-902c-411e4062b38a.png)
* Drawer is not closed when user clicks any item inside (**compose** and **inbox** in this example)
![image](https://user-images.githubusercontent.com/57976479/144683934-6d41f8d4-7b2a-4d40-a1c5-4f9af6eec265.png)

Silver-IT:
I can see that this issue is related to the #1024. This is not related to the `ngrok` or `tunnel` and it is only related to the Mobile Safari browser.

Silver-IT:
Also related to the #1023

Silver-IT:
Avatar is not displaying for the `PENDING` users.
![image](https://user-images.githubusercontent.com/57976479/145159135-eacb5c48-4b0c-4046-8364-f4d2e7789e59.png)

### Solution

The file name is `user-avatar-pending.png` not `avatar-pending.png`

- [x] Already Fixed on 9th of December

Silver-IT:
### Problem
User name/displayName overflow issue.
![image](https://user-images.githubusercontent.com/57976479/147209597-b99b65eb-00ec-4f75-a7b0-ef7e70edf81f.png)

### Solution
I think we need to truncate it.

**EDITED: DONE**

Silver-IT:
**Another Issue**

[...]

PR #1121, opened 2021-12-06
NavigationDuplicated error when we log out

### LINKED ISSUE #1079: Fix all console errors (e.g. Navigation Duplicated errors) [$125 bounty]
### Problem

In trying to upgrade Cypress, @snowteamer noticed in #1078, that NavigationDuplicated errors were preventing the tests from passing.

### Solution

Find and fix all of these errors. They're visible in the console during many of the tests.

See also previous related issue: #686

### Bounty

$125 bounty for a clean solution to this (paid in cryptocurrency).

Silver-IT:
Fixed #1079

### DIFF
--- a/frontend/utils/errorHandlers.js
+++ b/frontend/utils/errorHandlers.js
@@ -1,3 +1,0 @@
-export function logExceptNavigationDuplicated (err: Object) {
-  err.name !== 'NavigationDuplicated' && console.error(err)
-}
--- a/frontend/utils/errorHandlers.js
+++ b/frontend/utils/errorHandlers.js
@@ -1,0 +1,3 @@
+export function logExceptNavigationDuplicated (err: Object) {
+  err.name !== 'NavigationDuplicated' && console.error(err)
+}
--- a/frontend/views/components/GroupWelcome.vue
+++ b/frontend/views/components/GroupWelcome.vue
@@ -72,6 +72,7 @@ export default ({
   @include phone {
     justify-content: left;
     padding-left: 1rem;
+    padding-top: 8rem;
   }
 
   .c-title,
@@ -86,11 +87,6 @@ export default ({
 
   .c-description {
     margin: 0 0 0.5rem;
-
-    @include phone {
-      position: absolute;
-      top: 5.5rem;
-    }
   }
 
   @include phone {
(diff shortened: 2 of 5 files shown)


snowteamer:
While working on the notifications I've incidentally found another Vue navigation error.

It occured after I created an account, logged out without having a group yet, then logged in again:

image

Althought that is not strictly a `NavigationDuplicated` error, that is still an uncaught navigation error, so feel free to handle it in this PR if you would like to.

In case that may help, here is the patch I've used locally to fix the error:

image

frontend/utils/errorHandlers.js
```diff
@@ -0,0 +1,3 @@
+export function logExceptNavigationDuplicated (err: Object) {
+  err.name !== 'NavigationDuplicated' && console.error(err)
+}
```
taoeffect: This function [used to exist in `frontend/controller/utils/misc.js`](https://github.com/okTurtles/group-income-simple/pull/690/files#diff-b13300ef486717f682409d5ff465964df6f67442ecb77abc6a51da8e8256ecd3R10).

I would put it back there instead of creating a new file just for it.

taoeffect (changes requested):
Thanks @Silver-IT!

Two requests:

1. See comment about moving a function
2. Can you implement @snowteamer's fix too?

frontend/utils/errorHandlers.js
Silver-IT: Sure. I will

Silver-IT:
Regarding the @snowteamer's solution, I don't prefer to solve it in that way.
That's because this error is what the guard(`groupGuard` in this situation) makes, and if we handle all cases before to enter guard, there is no need to setup guard.

frontend/views/components/GroupWelcome.vue
```diff
@@ -86,11 +87,6 @@ export default ({
 
   .c-description {
     margin: 0 0 0.5rem;
-
-    @include phone {
-      position: absolute;
-      top: 5.5rem;
-    }
```
taoeffect, line 93: @Silver-IT Can you explain these CSS changes?

Silver-IT: This commit shouldn't be here in this PR because it is for the first problem of #1117.
`Group picture and description are overlapped`
Anyway, it fixes the overlapping issue in mobile view.

taoeffect (approved):
Approved! Great job @Silver-IT! 😄 🥇

PR #1133, opened 2021-12-09
Implement expiry time for the invite links

Silver-IT:
Need to add function to update expiry time in Group Settings page.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -22,7 +22,8 @@ import L from '~/frontend/views/utils/translations.js'
 import {
   INVITE_INITIAL_CREATOR,
   INVITE_STATUS,
-  PROFILE_STATUS
+  PROFILE_STATUS,
+  MAX_INVITE_AGE_IN_DAYS
 } from './constants.js'
 
 export const inviteType: any = objectOf({
@@ -50,7 +51,7 @@ export function createInvite (
   responses: {...},
   status: string,
 |} {
-  const expiresInDate = creator === INVITE_INITIAL_CREATOR ? 30 : 7
+  const expiresInDate = creator === INVITE_INITIAL_CREATOR ? MAX_INVITE_AGE_IN_DAYS.INITIAL : MAX_INVITE_AGE_IN_DAYS.PROPOSAL
   return {
     inviteSecret: `${parseInt(Math.random() * 10000)}`, // TODO: this
     quantity,
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -14,7 +14,10 @@ import {
 import { paymentStatusType, paymentType, PAYMENT_COMPLETED } from './payments/index.js'
 import * as Errors from '../errors.js'
 import { merge, deepEqualJSONType, omit } from '~/frontend/utils/giLodash.js'
-import { currentMonthstamp, prevMonthstamp, ISOStringToMonthstamp, compareMonthstamps, addMonthsToDate, dateToMonthstamp, compareCycles } from '~/frontend/utils/time.js'
+import {
+  currentMonthstamp, prevMonthstamp, ISOStringToMonthstamp, compareMonthstamps,
+  addMonthsToDate, dateToMonthstamp, compareCycles, DAYS_MILLIS
+} from '~/frontend/utils/time.js'
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
 import groupIncomeDistribution from '~/frontend/utils/distribution/group-income-distribution.js'
 import currencies, { saferFloat } from '~/frontend/views/utils/currencies.js'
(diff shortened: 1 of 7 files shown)


Silver-IT:
Closes #620

frontend/model/contracts/group.js
```diff
@@ -50,14 +50,15 @@ export function createInvite (
   responses: {...},
   status: string,
 |} {
+  const expiresInDate = creator === INVITE_INITIAL_CREATOR ? 30 : 7
```
snowteamer: Since a number of days is not actually a date, I would rather name this constant e.g. `maxAgeInDays`

Silver-IT: I agree. Where should I place those constants?

Silver-IT: Is there any recommended file to store those constants?
What about this file `frontend/model/contracts/constants.js`?

Silver-IT:
```
export const MAX_INVITE_AGE_IN_DAYS = {
  INITIAL: 30,
  PROPOSAL: 7
}
```

What about this in `frontend/model/contracts/constants.js`?

frontend/views/pages/Join.vue
```diff
@@ -75,7 +80,7 @@ export default ({
     pageStatus: {
       get () { return this.ephemeral.pageStatus },
       set (status) {
-        const posibleStatus = ['LOADING', 'WELCOME', 'SIGNING', 'LOGGING', 'INVALID']
+        const posibleStatus = ['LOADING', 'WELCOME', 'SIGNING', 'LOGGING', 'INVALID', 'EXPIRED']
         if (!posibleStatus.includes(status)) {
```
snowteamer: Here we have the opportunity to fix that "posibleStatus" -> "possibleStatus" typo.

Silver-IT: Sure. Will fix soon

frontend/model/contracts/group.js
snowteamer: In fact I was rather thinking about renaming the `expiresInDate` constant. Sorry I should have been more explicit about that.

Nonetheless, I think defining e.g. `MAX_INVITE_AGE_IN_DAYS `  in `constants.js` is a good idea; it improves consistency since that file already contains invite-related constants.

frontend/model/contracts/group.js
```diff
@@ -50,14 +54,15 @@ export function createInvite (
   responses: {...},
   status: string,
 |} {
+  const expiresInDate = creator === INVITE_INITIAL_CREATOR ? INVITE_EXPIRES_IN_DAYS.INITIAL : INVITE_EXPIRES_IN_DAYS.PROPOSAL
```
taoeffect: This PR is looking really good! One request that I have is to store the expiry time values in the group settings so that in the future if we need to change them (or allow the group to change them), we can.

Here's how I recommend doing that:

Modify the group constructor to include these values:

```js
        // TODO: checkpointing: https://github.com/okTurtles/group-income-simple/issues/354
        const initialState = merge({
          payments: {},
          paymentsByMonth: {},
          distributionEvents: [],
          invites: {},
          proposals: {}, // hashes => {} TODO: this, see related TODOs in GroupProposal
          settings: {
            groupCreator: meta.username,
            inviteExpiryInitial: INVITE_EXPIRES_IN_DAYS.INITIAL,
            inviteExpiryProposal: INVITE_EXPIRES_IN_DAYS.PROPOSAL
          },
```

Then, modify this function to take the expiry time as a variable:

```js
export function createInvite (
  {
    quantity = 1,
    creator,
    expires,
    invitee
  }: { quantity: number, creator: string, expires: number, invitee?: string }
): {|
```

And when it is called in the UI, pass in the expiry for the proposal based on the current value:

```js
          if (this.type === PROPOSAL_INVITE_MEMBER) {
            passPayload = createInvite({
              invitee: this.proposal.data.proposalData.member,
              creator: this.proposal.meta.username,
              expires: this.currentGroupState.settings.inviteExpiryProposal
            })
```

test/cypress/integration/group-proposals.spec.js
```diff
@@ -2,6 +2,10 @@ const userId = Math.floor(Math.random() * 10000)
 const groupName = 'Dreamers'
 const groupMincome = 250
 const groupNewMincome = 500
+const groupInviteLinkExpiry = {
+  anyone: 30,
+  proposal: 7
+}
```
taoeffect: So that they don't get out-of-sync, can you import these values from `frontend/model/contracts/constants.js`?

taoeffect (changes requested):
Looking really great @Silver-IT! Two changes requested, see comments.

test/cypress/integration/group-proposals.spec.js
Silver-IT: Sure. This needs to be done.

frontend/model/contracts/group.js
Silver-IT: Your recommendation is actually nice. But you didn't mention how to set pass expires when we create the initial invite.
We call this sbp function inside `frontned/controller/actions/group.js` when we create the group itself.
So you mean, we can only manage the expiresInDate for the proposal-based invitations.
In other words, we don't use the variable defined as `group.settings.inviteExpiryInitial`.

taoeffect (approved):
Excellent work @Silver-IT !

PR #1138, opened 2021-12-23
SBP Notification API

### LINKED ISSUE #1116: Create SBP API to post notifications [$200 bounty]
### Problem

As part of polish being added to the notifications PR (#995), we need a way for any part of the app to deliver a notification that appears in the notifications drop down.

### Solution

- Design an elegant SBP API for delivering notifications
- ~~Internally, have the notifications saved to a new/different localstorage instance, (EDIT: see note below)~~ and make sure it is created per-user (since each user will have their own notifications). Store up to 90 notifications locally, for up to 1 month, and clear out the old ones as new ones are posted - *unless* they are unread. In other words, unread notifications older than 1 month are allowed, but start removing the oldest unread ones to keep within the notification limit, and prioritize removing read notifications first. In other words, read notifications are removed first to stay within the notification limit. If that isn't enough, then next the oldest unread ones are removed.

After chatting with @snowteamer, it seems like we can save these in the state and have them saved automatically along with everything else.

### Bounty

Issue is open to contractors or non-contractors. Non-contractors can get paid via bounty. We're posting a bounty of $200 for a clean solution to this issue.

snowteamer:
Closes #1116

### Summary of changes
- New `gi.notifications` SBP domain, currently providing the following selectors:
  - 'gi.notifications/emit': create and emit a notification;
  - 'gi.notifications/markAsRead': mark a notification as read;
  - 'gi.notifications/markAllAsRead': mark all notifications from the current group as read, as well as notifications which are not tied to any group;

- New `notifications` Vuex module, located in the `@model/notifications` folder.
 Its available action and mutation identifiers are exported from `notifications/vuexModuleKeys.js`, while the implementation itself can be found in `./vuexModule.js`.
 - The `sbp` namespace is now available in Cypress tests and in the devtools console when in development mode.
 This can be used to emit artificial notifications from within the console. For example, to notify the user that `bob` has joined the current group:
  ```js
  const { currentGroupId } = sbp('state/vuex/state')
  sbp('gi.notifications/emit', 'MEMBER_ADDED', { groupID: currentGroupId, username: 'bob' })
  ```
-  Signatures for the available notification types can be found in `./templates.js`.
- Type definitions can be found in `./types.flow.js`.

### TODO
- Tests for multiple groups scenarios (partially done)
- Read/unread status synchronization between multiple tabs/devices (Can be implemented later)

Please feel free to discuss and suggest any improvement, better naming, or whatever makes sense.


frontend/model/notifications/vuexModuleKeys.js
```diff
@@ -0,0 +1,14 @@
...
+export const LOGOUT = 'logout'
+
+/* Mutation keys */
+
+export const ADD_NOTIFICATION = 'addNotification'
+export const MARK_ALL_NOTIFICATIONS_AS_READ = 'markAllNotificationsAsRead'
+export const MARK_NOTIFICATION_AS_READ = 'markNotificationAsRead'
+export const REMOVE_ALL_NOTIFICATIONS = 'removeAllNotifications'
+export const REMOVE_NOTIFICATION = 'removeNotification'
+export const SET_NOTIFICATIONS = 'setNotifications'
```
taoeffect: Here, I can't tell, is it mixing the existing actions login and logout with the notifications?

If so, then as `'login'` and `'logout'` are defined elsewhere, this is a DRY violation. To DRY it, we'd need to make sure that the action and mutation names are defined once only, and then used consistently throughout the app. So for this reason I suggest moving this file up a directory, and then modifying `state.js` accordingly.

Otherwise .. I couldn't find where these actions are interconnected with the app login and logout..

It seems to be mixing these with the main Vuex actions.. and not as a submodule.. unless I'm misunderstanding.

```js
sbp('state/vuex/commit', keys.ADD_NOTIFICATION, notification)
```

frontend/utils/time.js
```diff
@@ -107,15 +144,19 @@ export function timeSince (datems: number, dateNow: number = Date.now()): string
...
+    return L('1d')
   }
   if (interval >= HOURS_MILLIS) {
     return Math.floor(interval / HOURS_MILLIS) + 'h'
   }
-  return Math.max(1, Math.floor(interval / MINS_MILLIS)) + 'm'
+  if (interval >= MINS_MILLIS) {
+    return Math.max(1, Math.floor(interval / MINS_MILLIS)) + 'm'
+  }
+  return '<1m'
```
taoeffect: Hmm. These are user all also user-facing strings. So maybe we should add `L`'s here too?

Only thing to check though: I *think* there might currently be a bug where parameters have to have a space separating them from surrounding content. So please check if `L('{minutes}m', { minutes: ... })` works before implementing that. (And if it doesn't - bonus points if you can figure out the issue! But if you don't feel like it, then this part can be untranslated for now, with maybe a comment explaining why)

frontend/model/notifications/fakeDB.js
```diff
@@ -9,17 +9,17 @@ export default {
   },
   '1032': {
     timestamp: 1590822227327,
-    type: 'SEND_CONTRIBUTION',
+    type: 'CONTRIBUTION_REMINDER',
```
taoeffect: Should this entire file be removed now?

taoeffect (changes requested):
Fantastic work @snowteamer! Here's a quick preliminary review.

Can you give some examples on how to create these notifications from the console?

frontend/model/notifications/vuexModuleKeys.js
snowteamer: Yes, this module is not namespaced, so that Vuex will effectively mix its actions and mutations with main ones defined in the root module `state.js`.
So for example upon login, the `login` action handler from `notifications/vuexModule.js` will be automatically called alongside its eponym from `state.js`. This works because they have the same name.

frontend/utils/time.js
snowteamer: The minute has an international symbol `min` which wouldn't need to be translated if it was used. Otherwise I agree. Also thank you for the hint!

frontend/model/notifications/fakeDB.js
snowteamer: I don't know yet;  it might become useful as a basis for test data.

frontend/model/notifications/vuexModuleKeys.js
snowteamer: However exporting them might have been a mistake. The intent was to allow usage of LOGIN and LOGOUT constants in `vuexModule.js`, but whether to export them would rather be the responsability of `state.js`.

taoeffect: So my concern is the inconsistency and potential DRY issues.

DRY is designed to prevent multiple copies of the same code from appearing in the codebase, which over time leads to chaos, disorganisation, and an unmaintainable codebase. It's entropy that has to always be fought against.

So if in one part of the codebase we're using `'login'` as defined in `state.js`, and in another part of the codebase we're using `LOGIN`, that's the kind of thing to avoid.

I am definitely OK with using constants, but if you introduce them, then you have to modify all of the rest of the code to do the same thing.

snowteamer: Agreed, randomly using either `LOGIN` or `'login'` would be inconsistent. Therefore `LOGIN` will no longer be exported for now, leaving only `login`. We can still think later about exporting constants from `state.js` and refactor as necessary to use them.

frontend/model/notifications/templates.js
```diff
@@ -1,133 +1,99 @@
...
+  +level: NotificationLevel;
+  +linkTo: string;
+  +creator?: string;
+  +subtype?: string;
 }
 
-export const PROPOSAL_NEW = ({ timestamp, data }: TemplateArguments): TemplateDescription => {
-  const avatar = ''
-  const member = `<strong>${data.creator}</strong>`
+const strong = (text) => `<strong>${text}</strong>`
```
corrideat: This was already from before (this makes a more general implementation), but I wonder if it is a good idea to apply HTML like this. I'm unsure of what a good solution is when one wants markup inside a string that must be translated, but I'd start by escaping `&` with `&amp;`, `>` with `&gt;` and `<` with `&lt;`.

frontend/model/notifications/templates.js
```diff
@@ -1,133 +1,99 @@
...
+        month: strong(monthName(data.monthstamp))
+      }),
+      icon: 'coins',
+      level: 'info',
+      linkTo: '/payments'
+    }
+  },
+  INCOME_DETAILS_OLD (data: {}) {
+    return {
+      body: L("You haven't updated your income details in more than 6 months. Would you like to review them now?"),
```
corrideat: Maybe make the `6 months` an argument? It seems like the kind of thing better kept as a constant.

frontend/model/notifications/utils.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export function compareOnPriority (notificationA: Notification, notificationB: Notification): -1 | 0 | 1 {
+  const TWO_DAYS = 2 * ONE_DAY
+  const FIFTEEN_DAYS = 15 * ONE_DAY
+  const THREE_WEEKS = 21 * ONE_DAY
+  const preferA = -1
+  const preferB = 1
+
+  if (notificationA.read && !notificationB.read) {
```
corrideat: May it be useful to keep this DRY to avoid inconsistencies? E.g.,
```
if (notificationA.read !== notificationB.read) {
  const unread = notificationA.read ? notificationB : notificationA
  const read = notificationA.read ? notificationA : notificationB
  if (age(read) < ONE_DAY && age(unread) > FIFTEEN_DAYS) {
    return ...
   }
   ...
   }
     return age(notificationA) <= age(notificationB) ? preferA : preferB
}

File: frontend/model/notifications/vuexModule.js [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+    const notifications = await sbp('gi.db/notifications/load', user.username)
+
+    if (notifications && notifications.length) {
+      context.commit(keys.SET_NOTIFICATIONS, notifications)
+    }
+  },
+  async logout (context: VuexModuleContext) {
+    if (context.rootState.loggedIn) {
+      // Make sure to save the notifications *before* clearing them.
+      await sbp('gi.notifications/save')
```
corrideat: Isn't this missing the username?

frontend/model/notifications/vuexModule.js
```diff
@@ -0,0 +1,104 @@
...
+  },
+
+  unreadNotificationCount (state, getters) {
+    return getters.unreadNotifications.length
+  }
+}
+
+const mutations = {
+  [keys.ADD_NOTIFICATION] (state, notification: Notification) {
+    if (state.includes(notification)) {
```
corrideat, line 74: Unsure I follow the entire logic here, but if I might say, maybe these shouldn't be 'hard' errors, because they might for some reason come repeated or out of order?

frontend/model/notifications/vuexModule.js
```diff
@@ -0,0 +1,104 @@
...
+  [keys.MARK_ALL_NOTIFICATIONS_AS_READ] (state) {
+    state.forEach(item => {
+      item.read = true
+    })
+  },
+
+  [keys.REMOVE_NOTIFICATION] (state, notification: Notification) {
+    const index = state.indexOf(notification)
+
+    if (~index) {
```
corrideat: Nothing wrong here, but it seems a bit of an odd way (as in, I'm unaccustomed to see bitwise operations in JS) to check. I'd _subjectively_ say `index !== -1` or `index >= 0` are more idiomatic.

corrideat (commented):
Overall it seems good. I have a few general comments and some others that I think might be issues (the most salient being missing the username for storing the notifications).

frontend/model/notifications/vuexModule.js
snowteamer: These errors should normally not occur since `emit` always creates a new object, with a new timestamp.

Should such error nonetheless happen for some reason, then it's likely the symptom of a bug in the API or incorrect usage thereof somewhere. For example, manual notification programming using plain JS rather than the SBP API.

In such case, I think it's better to be notified of any issue immediately so as to be able to fix it. Otherwise, other developers might start relying on unsupported behavior, possibly leading to more complicated issues.

frontend/model/notifications/vuexModule.js
snowteamer: Since there is always a single logged-in username per session, I've made `sbp('gi.notifications/save')` just look it up by itself, so that there should be no need to specify it.

frontend/model/notifications/vuexModule.js
snowteamer: Yes, this idiom is less common than its counterparts.

Which I subjectively think is a bit unfortunate, since it is arguably less visual clutter, but also avoids having to consistently choose between` index !== -1`, `index >= 0`, but also `index > -1`, which is nice because none of them is clearly better than the other two.

So I've used this idiom here, waiting for feedback, but I will replace it if required.

frontend/model/notifications/templates.js
snowteamer: Yea that's a good idea. Every "reminder" notification type should likely be associated with some kind of configurable threshold constant. However, I'm not sure about the details yet.

frontend/model/notifications/templates.js
snowteamer: Good catch!

frontend/model/notifications/utils.js
snowteamer: Maybe. Thank you for your input, I'm going to spend a bit more time on this function.

frontend/model/notifications/templates.js
Silver-IT: In my experience, it is good to use this way since that number could be configurable later.
```
...
import { MAX_INCOME_PERSIST_IN_MONTHS } from '.../*.js'
...
INCOME_DETAILS_OLD (data: {}) {
    const maxIncomePersistInMonths = data.maxIncomePersistInMonths || MAX_INCOME_PERSIST_IN_MONTHS
    ...
}
```

frontend/model/notifications/utils.js
```diff
@@ -1,6 +1,6 @@
 import type { Notification } from './types.flow.js'
 
-import { notificationStorageLimits } from '@model/database.js'
+import { notificationStorageLimits } from '~/frontend/model/database.js'
```
Silver-IT: Is there any reason not to use the absolute path?
As far as I know, it's good to use absolute path using `@`, because it will work wherever the files are located and the project folder structure isn't changed very often.

frontend/model/database.js
```diff
@@ -66,3 +68,24 @@ sbp('sbp/selectors/register', {
...
+  'gi.db/notifications/save': function (username: string, items: Notification[]): Promise<void> {
+    return notificationStorage.setItem(username, items)
+  },
+  'gi.db/notifications/load': function (username: string): Promise<Notification[]> {
+    return notificationStorage.getItem(username)
+  },
+  'gi.db/notifications/delete': function (username: string): Promise<void> {
+    return notificationStorage.removeItem(username)
+  }
+})
```
taoeffect: Are these never used? If so, shouldn't they be deleted? Or is this PR unfinished?

frontend/model/notifications/fakeDB.js
taoeffect: Then those should be added in tests, and this file deleted.

taoeffect (changes requested):
Looking really good!

I added some additional comments. Also, can you create a new Cypress integration spec file for testing notifications? I recommend a single test of the following sort:

- Create a group and add two users to it
- Post one notification for each type of notification there is, and verify they appear, and verify the notification count is as it should be
- Click on one or more notifications, and verify the correct action is taken and the notification count goes down

frontend/model/notifications/vuexModule.js
```diff
@@ -0,0 +1,101 @@
...
+  [keys.SET_NOTIFICATIONS] (state, notifications: Notification[]) {
+    state.splice(0, state.length, ...notifications)
+  }
+}
+
+export default ({
+  state: () => [],
+  getters,
+  mutations
+}: Object)
```
taoeffect, line 118: Just wanted to say @snowteamer - great job splitting this up into its own vuex module, that's definitely the way to go 👍

frontend/model/database.js
snowteamer: Good catch. I'm deleting this for now.

frontend/views/containers/notifications/NotificationList.vue
```diff
@@ -1,136 +1,95 @@
...
+      return timeSince(item.timestamp)
     },
-    iconBg: (level) => {
+    iconBg (level: string): string {
       return {
         info: 'has-background-primary has-text-primary',
         success: 'has-background-success has-text-success',
         danger: 'has-background-danger has-text-danger'
       }[level]
+    },
```
Silver-IT, line 83: I think these(`title`, `ageTag`, and `iconBg`) shouldn't be methods. They should be `computed` values.

frontend/model/notifications/vuexModule.js
Silver-IT: Actually, I am not sure `state.includes(notification)` will return the expected value.
Since the state is Array of Object and notification is Object, I have some misunderstandings.
Even though the state includes a notification with the same keys and values, I think `state.includes(notification)` won't return true.

frontend/model/notifications/utils.js
```diff
@@ -66,36 +66,44 @@ export function applyStorageRules (notifications: Notification[]): Notification[
...
-    return preferB
-  }
-  if (!notificationA.read && notificationB.read) {
-    if (age(notificationA) > FIFTEEN_DAYS && age(notificationB) < ONE_DAY) {
-      return preferB
-    }
-    if (age(notificationA) > THREE_WEEKS && age(notificationB) < TWO_DAYS) {
-      return preferB
+  // If the given notifications are both read or both unread, then prefer the newer one.
+  if ((notificationA.read && notificationB.read) || (!notificationA.read && !notificationB.read)) {
```
Silver-IT: Is this same as the following?
`if (notificationA.read === notificationB.read)`

frontend/views/containers/navigation/GroupsList.vue
```diff
@@ -12,6 +12,7 @@ ul.c-group-list(v-if='groupsByName.length' data-test='groupsList')
     )
       button.c-group-picture.is-unstyled(@click='handleMenuSelect(group.contractID)')
         avatar.c-avatar(:src='$store.state[group.contractID].settings.groupPicture')
+      badge(v-if='unreadNotificationCountForGroup(group.contractID)') {{ unreadNotificationCountForGroup(group.contractID) }}
```
Silver-IT: Is it better to create a `computed` value instead of calling a function twice?
`unreadNotificationCountForGroup` is a function returned form the `getter`.

frontend/model/notifications/selectors.js
```diff
@@ -11,15 +11,20 @@ import templates from './templates.js'
 sbp('sbp/selectors/register', {
   // Creates and dispatches a new notification.
   'gi.notifications/emit' (type: string, data: NotificationData) {
-    const template = templates[type]
+    const template = templates[type](data)
```
Silver-IT, line 14: What about to rename `templates` to `templateFactory` or `templateGenerator`?
Actually, it's not a list of templates and it's a list of functions to generate template.

frontend/model/state.js
```diff
@@ -103,7 +103,8 @@ sbp('sbp/selectors/register', {
   },
   'state/vuex/state': () => store.state,
   'state/vuex/commit': (id, payload) => store.commit(id, payload),
-  'state/vuex/dispatch': (...args) => store.dispatch(...args)
+  'state/vuex/dispatch': (...args) => store.dispatch(...args),
+  'state/vuex/getters': () => store.getters
```
Silver-IT, line 108: It's really necessary to use `state/vuex/getters`. Thanks.

frontend/views/containers/navigation/GroupsList.vue
```diff
@@ -12,7 +12,7 @@ ul.c-group-list(v-if='groupsByName.length' data-test='groupsList')
     )
       button.c-group-picture.is-unstyled(@click='handleMenuSelect(group.contractID)')
         avatar.c-avatar(:src='$store.state[group.contractID].settings.groupPicture')
-      badge(v-if='unreadNotificationCountForGroup(group.contractID)') {{ unreadNotificationCountForGroup(group.contractID) }}
+      badge(v-if='unreadNotificationCountFor(group.contractID)') {{ unreadNotificationCountFor(group.contractID) }}
```
Silver-IT: We also should create a `computed` value and shouldn't call the function twice.

Silver-IT (commented):
Awesome. Great work, @snowteamer. 🥇

frontend/model/notifications/utils.js
snowteamer: Yea, if I remember correctly I had to change `@model/` to `~/frontend/` because the former caused an error.

frontend/views/containers/notifications/NotificationList.vue
snowteamer: That makes sense, however to be able to use computed values here we'd need something like a distinct Notification component, since these values differ per notification.

frontend/model/notifications/utils.js
snowteamer: This check is intended to prevent any duplicate references to the same notifification object to end up in the state Array.

Now you're right that two distinct notification objects bearing the same data would pass this check, but they would have been created through separate calls to `gi.notifications/emit`, so I think that's fine. Moreover this is quite unlikely to happen since a timestamp is included in the data, so both calls would have to occur in the same millisecond.

However, now that I'm thinking about it thanks to your input, maybe an extra check would be useful to make sure we never issue equivalent notifications twice during the same millisecond or short time interval.

frontend/views/containers/navigation/GroupsList.vue
snowteamer: I'm not sure how to replace the function by a computed value here, since the notification count can be different for each group.
Maybe with a getter like `group.unreadNotificationCount`, but while that syntax would be convenient, modularity would be decreased as `notification.js` and `group.js` would no longer be as much decoupled.

frontend/model/notifications/selectors.js
snowteamer: Yea you're right. I've been keeping this name because that list was exported from `templates.js`, but that file might as well be renamed too.

frontend/model/state.js
snowteamer: You're welcome! :smile:

frontend/views/containers/navigation/GroupsList.vue
snowteamer: Implementing a GroupBadge component should also allow for this, but I'm not sure it's worth adding a new file to the codebase. Do you have any suggestion?

Silver-IT: I missed that you were calling this function inside the loop. That should be defined in methods. Right.

frontend/views/containers/notifications/NotificationList.vue
Silver-IT: Yeap. They are calling inside the loop. You are right. 😉

frontend/model/notifications/utils.js
Silver-IT: But you are only comparing `boolean` typed `read` values here.

snowteamer: I was in fact replying to your previous comment above, regarding `state.includes(notification)`; for some reason no text field showed up there so I had to reply in the next available field.

As for the latter comment, after thinking about it I think that indeed `if (notificationA.read === notificationB.read)` turns out to be the same, but it wasn't obvious at all for me. So the longer version is likely easier to understand.

frontend/model/notifications/utils.js
Silver-IT: Now, I can understand what you mean. I have just tracked the Travis error which is very much similar to your case.
Travis fails because of the absolute path. I converted it into a related path and it seems working. Weird Travis...

frontend/model/notifications/utils.js
Silver-IT: @snowteamer. what I am concerned about is that `state.includes(notification)` will always be false unless we retrieve a notification from the state itself and just to check using the `includes` prototype function. Because they won't be compared by their values, they will be compared by the references.

frontend/model/notifications/vuexModule.js
taoeffect: Yes @snowteamer I think @Silver-IT is correct, `state.includes` will only compare by reference:

```js
> var a = {a: 4}
undefined
> b = []
[]
> b.push(a)
1
> b.includes({a: 4})
false
```

snowteamer: @greg That's intended. This check is meant to prevent duplicate references from ever being included in the array. However distinct objects with equal keys and values are allowed for now. Please see https://github.com/okTurtles/group-income/pull/1138#discussion_r804097925

frontend/model/notifications/utils.js
snowteamer: @Silver-IT Yea it is supposed to always be false, unless we later change the code in a dubious way. That's the idea. It also gives a clue that `state` is logically an ordered set of objects, although being implemented as an array. However I will remove it if it turns out to be too confusing.

frontend/model/database.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
-import sbp from '~/shared/sbp.js'
 import localforage from 'localforage'
+import sbp from '~/shared/sbp.js'
```
taoeffect, line 3: If there isn't a good reason to move this, can you restore it to where it was?

frontend/model/notifications/templates.js
```diff
@@ -1,133 +1,100 @@
...
+      level: 'info',
+      linkTo: '/contributions/TODO-LINK-MODAL',
+      scope: 'user'
+    }
+  },
+  MEMBER_ADDED (data: { groupID: string, username: string }) {
+    return {
+      body: L('The group has a new member. Say hi to {name}!', {
+        name: strong(data.username)
+      }),
```
taoeffect, line 46: Can you double-check that this is safe?

I know that the `i18n` component uses `dompurify` (see `views/utils/translations.js`), but I don't think the `L` function does. 

So if we're passing in user provided data like `data.username` here, is this a potential issue? Or was the username already sanitized? Or...?

taoeffect (changes requested):
Looking good @snowteamer! I think to get this merged we need remaining:

- Removal of `[WIP]` from title
- Merging `master`
- Passing tests
- Double-check the comments that I just added, especially the one about the potential XSS
- Resolve any outstanding comments

frontend/model/notifications/templates.js
snowteamer: I think `data.username` here gets sanitized later by using the `v-safe-html` directive:

https://github.com/okTurtles/group-income/blob/188a33d9a3b666881f9fe58c5e1029a70e6bd436/frontend/views/containers/notifications/NotificationList.vue#L24

frontend/model/database.js
snowteamer: I've moved this import for two reasons:
- in principe, third-party module imports should come before our own imports
- when it's easy to do, lexical ordering is usually nicer than no ordering at all

taoeffect: > in principe, third-party module imports should come before our own imports

Ah ok. Well, I've already made SBP into [a third-party module](https://www.npmjs.com/package/@sbp/sbp), it just hasn't been imported yet into this project. That's an open issue: #1153

So can we revert this change?

frontend/model/notifications/templates.js
taoeffect: OK good. @snowteamer, then can you add a comment here letting people know that they must include this in the page using `v-safe-html` to protect against XSS?

frontend/model/notifications/utils.js
taoeffect: @snowteamer Yeah I agree with @Silver-IT's observation, if this `if` could be simplified to `if (notificationA.read === notificationB.read)` that would be preferable. Save a few CPU cycles and some battery life :)

The comment (great job on that!) makes it clear what the intent is.

frontend/model/notifications/vuexModule.js
taoeffect: I've seen people do this but I agree with @corrideat personally... because it's not clear to a human what the value of `~index` is. By bounding it with `index > -1` or `index >= 0` it's clear what the intent behind the `if` is. So if it were me I'd change it, but @snowteamer if you feel strongly that you want to keep it we can.

taoeffect (changes requested):
@snowteamer ok I posted a few more comments.

I also found two issues:

### Issue 1 - some kind of exception going on

This only happened for the second user I added to a group (the first one to be invited).

I first created a group, and then posted some test notifications for users that didn't exist in the group:

```js
sbp('gi.notifications/emit', 'MEMBER_ADDED', { groupID: currentGroupId, username: 'joe' })
```

Then I invited `u2` to the group. When they got invited for some reason the navbar on the left didn't appear. I'm guessing an exception happened and it prevented it from appearing. I didn't think to open the console at that point, but later on after adding some more users to see if it would happen again (it didn't) I refreshed `u2`'s tab and saw an exception. This part is recorded in a video shared with you on Slack.

### Issue 2 - 3 `undefined` lookups.

I posted a total of 3 notifications - 2 for users that didn't exist, and one for one that does exist.

When I bring up the notifications window, there are 3 GETs to name lookups for AvatarUser with `undefined`... see screenshot below:

Screen Shot 2022-03-22 at 2 23 28 PM

frontend/model/notifications/templates.js
snowteamer: I have already included a similar comment above `strong()`"s definition. Maybe it was not the best place. Would a reference to that comment be fine, or do you think it's better to duplicate or move that comment to here?

taoeffect: Oh! I forgot about that. Sorry! Resolving.

frontend/model/database.js
snowteamer: Sure. This will still break lexical ordering though. Unless we insert a blank line after the `sbp` import:
```js

import localforage from 'localforage'

Comment (snowteamer) [reply frontend/model/database.js:3]: This would even make it slightly more obvious that we're going to use SBP in this file.

Comment (taoeffect) [reply frontend/model/database.js:3]: ok sure, it sounds like it's more important to you than it is to me, and it really isn't a big deal either way, so feel free to do it this way

File: frontend/model/notifications/vuexModule.js (hunk 9 above)
Comment (snowteamer) [reply frontend/model/notifications/vuexModule.js]: That's no big deal, we can change it. Which variant would you prefer?

Comment (snowteamer) [reply frontend/model/notifications/vuexModule.js]: Going with `index > -1` for now.

**Review by taoeffect — CHANGES_REQUESTED** [review]
(So that I don't forget, per Slack, needs a fix for repeated calls to saveSettings due to recent change)

File: frontend/model/state.js [hunk 21]
```diff
@@ -600,11 +596,17 @@ const actions = {
   },
   // persisting the state
   async saveSettings (
-    { state }: {state: Object}
+    { commit, state }: { state: Object}
```
taoeffect: Could we remove this `commit` from here just to be safe, so that we don't give the impression that it's safe to use in this function? (Maybe even add a comment to that effect...)

snowteamer: Sure. Good catch.

taoeffect (approved):
Approved!! 🎉🎉🎉🎉🎉🎉 😄

Issue #1107, opened 2021-11-08
Consider switching to Parcel 2? [$400 bounty]


taoeffect:
### Problem

Our current build setup lost code-splitting because esbuild doesn't support it (yet). This is an open issue: #1093

Recently [Parcel 2 was announced](https://parceljs.org/blog/v2/), and it looks pretty awesome, is supposedly a lot faster, now with a Rust-based "compiler"/bundler, and it supports code-splitting.

### Solution

This is something I was expecting to happen (and to continue to happen), and it's why we have an open issue #1085 for making our build system bundler-agnostic.

We want to be able to switch to a new bundler anytime it pulls ahead significantly of our existing setup (because the state of JavaScript bundlers remains somewhat lacking, and it impacts our productivity).

So closing this issue would involve switching us to Parcel 2, while doing it such a way that we can still switch back to esbuild (or anything else) should we want to (#1085).

**NOTE:**

- Before fully committing to this, we should evaluate the performance tradeoffs involved. If Parcel 2 significantly slows down our build times, then it might not be worth it to switch. Find out what the reality of this is.
- Make sure that all of the features we currently support are still supported with Parcel.
- **Double-check if we can use es6 modules instead. If we can - do that!**
  - https://medium.com/geekculture/using-native-unbundled-ecmascript-modules-with-a-simple-build-pipeline-351574a32815
  - https://github.com/guybedford/es-module-shims

### Bounty

$400 bounty for a clean solution to this (paid in cryptocurrency).

EDIT! It seems like we might be able to get this working with esbuild - see #1093. We're going to try that first, and if that works then we'll close this issue, especially since the next step from esbuild would be preferably Deno (#586).

taoeffect:
Closing in favor of #1196.

PR #1140, opened 2021-12-26
Issue #691 light client

### LINKED ISSUE #691: "Light client" for storage-challenged environments (i.e. browsers)
### Problem

Local storage on my browsers (especially mobile) will not let us store all of the log history.

### Solution

For non electron-type versions of the app, create a light client mode that discards the history. This may be as simple as not storing the history in the first place.

Edit: suggested solution should modify `shared/domains/gi/db.js` to do the following:

- Between [lines 67-68](https://github.com/okTurtles/group-income-simple/blob/master/shared/domains/gi/db.js#L67-L68), right before `return entry.hash()`, call a function that deletes the old entries from storage if we are a light client
- Implement the function that deletes the old entries
- Implement a `'gi.db/delete` selector that by default calls `'okTurtles.data/remove'`, and is [overwritten in `frontend/model/database.js`](https://github.com/okTurtles/group-income-simple/blob/master/frontend/model/database.js#L13-L17) to remove from localstorage

The reasoning for doing it this way instead of simply not storing it in storage in the first place is to preserve the useful logic checks that `'gi.db/log/addEntry'` performs.

corrideat:
Implement support for 'light' clients. This is a frontend: if using a 'light' client, the state is not stored in the local indexedDB store. Instead, the state is only kept in VueX and reprocessed as needed.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,7 @@ const { resolve } = path
 // We will rather load it later, and only if necessary.
 // require('@babel/register')
 
-const { NODE_ENV = 'development', VUEX_STRICT = 'true' } = process.env
+const { NODE_ENV = 'development', VUEX_STRICT = 'true', LIGHTWEIGHT_CLIENT = 'true' } = process.env
 
 const backendIndex = './backend/index.js'
 const distAssets = 'dist/assets'
@@ -119,7 +119,8 @@ module.exports = (grunt) => {
       define: {
         'process.env.BUILD': "'web'", // Required by Vuelidate.
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
-        'process.env.VUEX_STRICT': VUEX_STRICT
+        'process.env.VUEX_STRICT': VUEX_STRICT,
+        'process.env.LIGHTWEIGHT_CLIENT': LIGHTWEIGHT_CLIENT
       },
       external: ['crypto', '*.eot', '*.ttf', '*.woff', '*.woff2'],
       incremental: true,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,7 @@ const { resolve } = path
 // We will rather load it later, and only if necessary.
 // require('@babel/register')
 
-const { NODE_ENV = 'development', VUEX_STRICT = 'true', LIGHTWEIGHT_CLIENT = 'true' } = process.env
+const { LIGHTWEIGHT_CLIENT = 'true', NODE_ENV = 'development', VUEX_STRICT = 'true' } = process.env
 
 const backendIndex = './backend/index.js'
 const distAssets = 'dist/assets'
@@ -118,9 +118,9 @@ module.exports = (grunt) => {
       bundle: true,
       define: {
         'process.env.BUILD': "'web'", // Required by Vuelidate.
+        'process.env.LIGHTWEIGHT_CLIENT': LIGHTWEIGHT_CLIENT,
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
-        'process.env.VUEX_STRICT': VUEX_STRICT,
-        'process.env.LIGHTWEIGHT_CLIENT': LIGHTWEIGHT_CLIENT
+        'process.env.VUEX_STRICT': VUEX_STRICT
       },
       external: ['crypto', '*.eot', '*.ttf', '*.woff', '*.woff2'],
       incremental: true,
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -63,8 +63,6 @@ async function startApp () {
     })
   }
 
-  sbp('okTurtles.data/set', 'LIGHTWEIGHT_CLIENT', process.env.LIGHTWEIGHT_CLIENT ?? true)
-
   // this is to ensure compatibility between frontend and test/backend.test.js
   sbp('okTurtles.data/set', 'API_URL', window.location.origin)
 
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -63,6 +63,8 @@ async function startApp () {
     })
   }
 
+  sbp('okTurtles.data/set', 'LIGHTWEIGHT_CLIENT', process.env.LIGHTWEIGHT_CLIENT ?? true)
+
   // this is to ensure compatibility between frontend and test/backend.test.js
   sbp('okTurtles.data/set', 'API_URL', window.location.origin)
 
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -21,7 +21,7 @@ sbp('sbp/selectors/overwrite',
             return null
           }
           const state = sbp('state/vuex/state')
-          if (state?.contracts?.[contractId]?.HEAD) {
+          if (state.contracts[contractId]?.HEAD) {
             return state.contracts[contractId].HEAD
           }
           return null
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -35,14 +35,14 @@ sbp('sbp/selectors/overwrite',
       }
     : {
         'gi.db/get': key => {
-          return log.getItem(`${key}`)
+          return log.getItem(key)
         },
         // TODO: handle QuotaExceededError
         'gi.db/set': (key, value) => {
           return log.setItem(key, value)
         },
         'gi.db/delete': (key: string) => {
-          return log.removeItem(`${key}`)
+          return log.removeItem(key)
         }
       })
 
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -9,14 +9,14 @@ const log = localforage.createInstance({
   storeName: 'Contracts'
 })
 
-const isLightweightClient = process.env.LIGHTWEIGHT_CLIENT ?? true
+const isLightweightClient = process.env.LIGHTWEIGHT_CLIENT === 'true'
 
 // make gi.log use localforage for storage
 sbp('sbp/selectors/overwrite',
   isLightweightClient
     ? {
         'gi.db/get': key => {
-          const contractId = sbp('gi.db/log/isLogHEAD', key)
+          const contractId = sbp('gi.db/log/contractIDFromLogHEAD', key)
           if (!contractId) {
             return null
           }
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -16,7 +16,7 @@ sbp('sbp/selectors/overwrite',
   isLightweightClient
     ? {
         'gi.db/get': key => {
-          const contractId = sbp('gi.db/log/contractIDFromLogHEAD', key)
+          const contractId = sbp('gi.db/log/contractIdFromLogHEAD', key)
           if (!contractId) {
             return null
           }
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -9,13 +9,42 @@ const log = localforage.createInstance({
   storeName: 'Contracts'
 })
 
+const isLightweightClient = process.env.LIGHTWEIGHT_CLIENT ?? true
+
 // make gi.log use localforage for storage
-sbp('sbp/selectors/overwrite', {
-  'gi.db/get': key => log.getItem(key),
-  // TODO: handle QuotaExceededError
-  'gi.db/set': (key, value) => log.setItem(key, value),
-  'gi.db/delete': (key: string) => log.removeItem(key)
-})
+sbp('sbp/selectors/overwrite',
+  isLightweightClient
+    ? {
+        'gi.db/get': key => {
+          const contractId = sbp('gi.db/log/isLogHEAD', key)
+          if (!contractId) {
+            return null
+          }
+          const state = sbp('state/vuex/state')
+          if (state?.contracts?.[contractId]?.HEAD) {
+            return state.contracts[contractId].HEAD
+          }
+          return null
+        },
+        'gi.db/set': (key, value) => {
+          return Promise.resolve(value)
+        },
+        'gi.db/delete': () => {
+          return Promise.resolve()
+        }
+      }
+    : {
+        'gi.db/get': key => {
+          return log.getItem(`${key}`)
+        },
+        // TODO: handle QuotaExceededError
+        'gi.db/set': (key, value) => {
+          return log.setItem(key, value)
+        },
+        'gi.db/delete': (key: string) => {
+          return log.removeItem(`${key}`)
+        }
+      })
 
 // =======================
 // App settings to persist state across sessions
--- a/frontend/model/database.js
+++ b/frontend/model/database.js
@@ -13,7 +13,8 @@ const log = localforage.createInstance({
 sbp('sbp/selectors/overwrite', {
   'gi.db/get': key => log.getItem(key),
   // TODO: handle QuotaExceededError
-  'gi.db/set': (key, value) => log.setItem(key, value)
+  'gi.db/set': (key, value) => log.setItem(key, value),
+  'gi.db/delete': (key: string) => log.removeItem(key)
 })
 
 // =======================
--- a/shared/domains/gi/db.js
+++ b/shared/domains/gi/db.js
@@ -43,6 +43,9 @@ export default (sbp('sbp/selectors/register', {
   'gi.db/log/logHEAD': function (contractID: string): string {
     return `${contractID}-HEAD`
   },
+  'gi.db/log/isLogHEAD': function (key: string): string | boolean {
+    return key.slice(-5) === '-HEAD' ? key.slice(0, -5) : false
+  },
   'gi.db/log/getEntry': async function (hash: string): Promise<GIMessage> {
     try {
       const value: string = await sbp('gi.db/get', hash)
@@ -68,10 +71,6 @@ export default (sbp('sbp/selectors/register', {
       await sbp('gi.db/set', sbp('gi.db/log/logHEAD', contractID), entry.hash())
       console.debug(`[addLogEntry] HEAD for ${contractID} updated to:`, entry.hash())
       await sbp('gi.db/set', entry.hash(), entry.serialize())
-      const isLightweightClient = sbp('okTurtles.data/get', 'LIGHTWEIGHT_CLIENT')
-      if (isLightweightClient && previousHEAD) {
-        await sbp('gi.db/delete', previousHEAD)
-      }
       return entry.hash()
     } catch (e) {
       if (e.name.indexOf('ErrorDB') === 0) {
--- a/shared/domains/gi/db.js
+++ b/shared/domains/gi/db.js
@@ -45,8 +45,8 @@ export default (sbp('sbp/selectors/register', {
   'gi.db/log/logHEAD': function (contractID: string): string {
     return `${contractID}${headSuffix}`
   },
-  'gi.db/log/contractIdFromLogHEAD': function (key: string): string | boolean {
-    return key.endsWith(headSuffix) ? key.slice(0, -headSuffix.length) : false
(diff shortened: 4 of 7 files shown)


frontend/main.js
```diff
@@ -63,6 +63,8 @@ async function startApp () {
     })
   }
 
+  sbp('okTurtles.data/set', 'LIGHTWEIGHT_CLIENT', process.env.LIGHTWEIGHT_CLIENT ?? true)
```
snowteamer, line 66: No need to provide another fallback value for `process.env.LIGHTWEIGHT_CLIENT` here since one is already specified in the Grunt file.

shared/domains/gi/db.js
```diff
@@ -37,9 +37,15 @@ export default (sbp('sbp/selectors/register', {
...
   },
+  'gi.db/delete': function (key: string) {
+    return Promise.resolve(sbp('okTurtles.data/delete', key))
+  },
   'gi.db/log/logHEAD': function (contractID: string): string {
     return `${contractID}-HEAD`
   },
+  'gi.db/log/isLogHEAD': function (key: string): string | boolean {
+    return key.slice(-5) === '-HEAD' ? key.slice(0, -5) : false
+  },
```
taoeffect: Questions / concerns:

- This appears to be a helper function used only in the frontend. As such, to prevent excess logging in the console, and to give a slight speedup, I think it would make more sense to define it as a helper function and call it directly in `frontend/model/database.js`.
- Does it make sense to have this at all? What purpose is it serving, and should it be doing it at all? See related comment above about `'gi.db/get'` override.

test/cypress/support/index.js
```diff
@@ -16,6 +16,12 @@
 import './commands.js'
 import './output-logs.js'
 
+before(function () {
+  cy.clearCookies()
+  cy.clearLocalStorage()
+  indexedDB.deleteDatabase('Group Income')
+})
+
```
taoeffect, line 24: Did you double-check the Cypress docs to see how many of these are necessary? I feel like Cypress does some of this already between integration tests, doesn't it?

frontend/main.js
corrideat: @snowteamer Did you change your mind about this not being needed?

snowteamer: I still think the `?? true` part is not necessary assuming a default value of `'true'` is already given in the Grunt file.
But also, to me it looks a bit unclear here if the intent is to store a string or a boolean.

Gruntfile.js
```diff
@@ -52,7 +52,7 @@ const { resolve } = path
 // We will rather load it later, and only if necessary.
 // require('@babel/register')
 
-const { NODE_ENV = 'development', VUEX_STRICT = 'true' } = process.env
+const { NODE_ENV = 'development', VUEX_STRICT = 'true', LIGHTWEIGHT_CLIENT = 'true' } = process.env
```
snowteamer: To me it would look slightly better if the lexical ordering was kept by inserting the new name before the other ones.

Gruntfile.js
```diff
@@ -119,7 +119,8 @@ module.exports = (grunt) => {
       define: {
         'process.env.BUILD': "'web'", // Required by Vuelidate.
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
-        'process.env.VUEX_STRICT': VUEX_STRICT
+        'process.env.VUEX_STRICT': VUEX_STRICT,
+        'process.env.LIGHTWEIGHT_CLIENT': LIGHTWEIGHT_CLIENT
```
snowteamer: Same remark here, although that's no big deal

frontend/model/database.js
```diff
@@ -9,12 +9,42 @@ const log = localforage.createInstance({
   storeName: 'Contracts'
 })
 
+const isLightweightClient = process.env.LIGHTWEIGHT_CLIENT ?? true
```
snowteamer: Again this expression should no longer need a fallback value at this point.
Besides, to convert it to a boolean I would use
```js
process.env.LIGHTWEIGHT_CLIENT === 'true'
```
or maybe (more explicit but less concise)
```js
process.env.LIGHTWEIGHT_CLIENT === 'true' ? true : false
```
In any case, `Boolean(process.env.LIGHTWEIGHT_CLIENT)` would not work as `Boolean('false')` is `true`.

shared/domains/gi/db.js
snowteamer, line 50: I would suggest for this function to keep a single responsability and stay consistent with its name by only checking whether the passed in key is a log head. Extracting the corresponding contract ID is another task.

snowteamer (changes requested):
Looks good overall, with just a few points to address.

shared/domains/gi/db.js
corrideat: What do you say about changing the name of the selector to `getLogHEAD` or such? I had this same thought while writing it, but I think also that perhaps two selector calls are a bit excessive for the required use (as an aside, I'm not sure I like the current implementation that changes the key name for contract heads like, i.e., `gi.db/log/logHEAD`, although I don't have a proposal for that at the moment)

frontend/main.js
corrideat: The intent was to store a boolean, with the default being a light client. I realise now that the code as is has no clear opt out from the light client implementation.

Gruntfile.js
corrideat: Agreed

frontend/model/database.js
corrideat: This is a good point (I realised that the opting out from being a light client is less than ideal).

frontend/model/database.js
```diff
@@ -9,12 +9,42 @@ const log = localforage.createInstance({
...
+          const contractId = sbp('gi.db/log/isLogHEAD', key)
+          if (!contractId) {
+            return null
+          }
+          const state = sbp('state/vuex/state')
+          if (state?.contracts?.[contractId]?.HEAD) {
+            return state.contracts[contractId].HEAD
+          }
+          return null
+        },
```
taoeffect: So the point of `'gi.db/get'` is to return the value for a key. That doesn't seem to be what this is doing, and what it is doing is entirely unclear to me.

It seems to be some kind of weird version of `'gi.db/log/logHEAD'`...? Can you help me figure what you're  trying to do here and why?

taoeffect (changes requested):
Nice job @corrideat! Hope your holidays have been good. This is a quick preliminary review to move the ball forward.

shared/domains/gi/db.js
corrideat: I think it does make sense, because of separation of concerns (see also my reply to @snowteamer, as I'm not entirely happy with having this selector). If this were done on the frontend, then the frontend would have to have some logic about what a HEAD key looks like, which is determined by `'gi.db/log/logHEAD`. So, we would have logic for the same thing spread across different places.

test/cypress/support/index.js
corrideat: I checked and it seems 'necessary'. If this isn't here, then if a test fails (say, because the user state was corrupted, as it happened as I tried different approaches here), then the issue carries over. IMO, Cypress should start each test with a blank / fresh state, but it doesn't. By having this, then an error in one test doesn't affect the other tests, which I think is more sensible.

frontend/model/database.js
corrideat: The implementation is a bit different from what we discussed earlier. I figured that storing messages just for deleting them afterwards was a bit redundant and added unnecessary complexity. Hence, with this change light clients will not store messages at all, only the contract state. So, this is returning `null` for normal message keys (as there is nothing to return) and returning the current HEAD for HEAD keys. See also the `set` counterpart, which simply discards normal messages.

test/cypress/support/index.js
corrideat: The necessity of this is however something that can be handled / discussed in a separate issue, as it's not technically required for this particular PR

frontend/model/database.js
taoeffect: @corrideat OK, is there a reason to return the head instead of `null` as well?

And can you create a helper function instead of defining a selector? (Selectors are meant to be used as an API... and I'm not sure we should add `isLogHEAD` as an API...)

taoeffect:
@corrideat - per today's earlier call, can you check to see if it's possible to implement this PR in such a way that it doesn't break if we switch back and forth between light client and "full node"?

EDIT: and if it's not an easy thing to do, then let's save it for a future issue/PR.

frontend/model/database.js
corrideat: @taoeffect  Yes, there's a reason. The rest of the client logic expects the HEAD to have a valid value (note that with this implementation it's only stored in the VueX state).

I think that `gi.db/log/isLogHEAD` is as deserving of a selector as `gi.db/log/logHEAD` is. A helper function works, but I feel like it makes things a bit unbalanced, if you think of `gi.db/log/isLogHEAD` as the reverse operation of `gi.db/log/logHEAD`.

While answering this, I'm thinking of a cleaner implementation (but with more changes required), which is about how HEAD storage is handled. On the client side (talking about non-light clients or the current implementation without this PR), the HEADs are stored twice client-side. First, they are stored in the 'Contracts' store, and then they are stored in the VueX user store. This is some unnecessary duplication, and the client doesn't really need the HEADs in the 'Contracts' store (and having it is what actually caused issues with the first implementation that deleted messages).

The issue is that the server does need to store the HEAD, and that there is no server-side state the way that there is for a client/user. So, one way of making this a bit cleaner might be _not_ using `gi.db/get` and `gi.db/set` for HEADs, but using some specialised method (say we call them `gi.db/getHEAD` and `gi.db/setHEAD`). This way, we address several issues at once:

1. We no longer need `gi.db/log/logHEAD` nor `gi.db/log/isLogHEAD`
2. We can handle these operations differently in the client and in the server. The server could continue to append `-HEAD` to the keys used for storage. The client could for simply store the HEAD in VueX and (for non-light clients) store messages in the 'Contracts' store as it does now.
3. I believe it makes non-light more more robust (because the HEAD state is not unnecessarily duplicated)

corrideat:
> @corrideat - per today's earlier call, can you check to see if it's possible to implement this PR in such a way that it doesn't break if we switch back and forth between light client and "full node"?

It might already work (one potential issue is I guess the HEAD inconsistency between VueX and the contracts storage, and another one is if for some reason there are some messages missing), it's more like I didn't test for this scenario. In particular, it's the full->light->full switch that is more prone to cause issues, as switching from full to light shouldn't result in an invalid state, as the client-side contract state is plainly ignored.

Before I can test this (and thus determine if how much work it might be), how do you see the switch happening? By a user setting on the UI or something like that? Because I guess that the one change that we will need is that we can no longer rely on the `LIGHTWEIGHT_CLIENT` environment variable. Then, the other change we'll need is that I should write some automated tests for this case when one switches between modes.

Edit: A quick-and-dirty fix that addresses any potential issues but is suboptimal wrt. caching is emptying the 'Contracts' store either upon switching from light to full mode or vice versa. This becomes even more suboptimal if the switch meant to work at the user level (as opposed to at the application level), because then one user in light mode might clear the contract state for full mode users on the same client app.

taoeffect:
@corrideat Yeah I was thinking it could be a hypothetical user-facing toggle switch in the UI. But it sounds like a difficult thing to implement this right now. Let's avoid that for now then for this PR, and it can be separate issue that we tackle later.

frontend/model/database.js
taoeffect: > I think that `gi.db/log/isLogHEAD` is as deserving of a selector as `gi.db/log/logHEAD` is.

Alright, we'll keep it.

> The issue is that the server does need to store the HEAD, and that there is no server-side state the way that there is for a client/user. So, one way of making this a bit cleaner might be not using gi.db/get and gi.db/set for HEADs, but using some specialised method (say we call them gi.db/getHEAD and gi.db/setHEAD). This way, we address several issues at once:

OK, these are some fair observations. Code duplication of any kind does get on my nerves of course... however, if we want to dedup things, we might want to use localstorage over Vuex in this instance, instead of the other way around.

The reason being is that when I designed this entire system, I designed it with decentralized networks IPFS, Dat, et. al. in mind. I found the idea of being able to boot an entire app given a single hash (key) very appealing. The entire app could be represented and verified and loaded using this one hash with such an approach. These networks worked with key/value pairs, and if we designed the app to use key/value pairs, we could work with any such network. And so, the idea would be to do as much though a key/value API as possible. Moving things to specialized selectors like `getHEAD` and `setHEAD`, would potentially complicate any switch over to a decentralized backend. On the other hand, keeping everything built around a single key/value API would simplify any such migration.

That's the main reason. The other reason I'm hesitant to split things up into getHEAD/setHEAD (instead of just get/set), is that part of the reasoning here seems to be that we could make there be a bigger difference between the client and server. Well, that would increase the complexity of the system, because you'd have to think about the way the client and server behave differently, and we want to minimize the difference between them. The reason for this is, again, partly to support a fully decentralized "backend", but also, there was another potential vision, which was that anyone could trivially convert their client into a server if they wanted to. The idea was that at some point we'd release a desktop app version of GI, and if someone in the group decided that they want to be the server for the group, they could use the same client they were using, to run it as a server with a flip of a switch.

Let me know your thoughts given all of that.

corrideat: In this case, I think we can let `isLogHEAD` stay (I'll rename it to something more descriptive that explains why it returns a contract ID), address the issues with the env variable and call this issue closed. I have some ideas still for how to handle HEAD, but these are secondary to this issue and can be addressed separately.

frontend/model/database.js
```diff
@@ -9,12 +9,42 @@ const log = localforage.createInstance({
...
+sbp('sbp/selectors/overwrite',
+  isLightweightClient
+    ? {
+        'gi.db/get': key => {
+          const contractId = sbp('gi.db/log/contractIDFromLogHEAD', key)
+          if (!contractId) {
+            return null
+          }
+          const state = sbp('state/vuex/state')
+          if (state?.contracts?.[contractId]?.HEAD) {
```
snowteamer: Are the `?` checks on `state?` and `state.contracts?` necessary? I don't see how they could be null unless something went really wrong.

shared/domains/gi/db.js
snowteamer: It seems to be difficult to come up with a better name for this one. While `contractIdFromLogHEAD` is a bit long, it's easy to understand.
- Afaik, `get-` and `set-` prefixed names like `getLogHEAD` are not common in the codebase. I'm not saying these names are inherently bad, but maybe avoid them here for consistency.
- `key.endsWith('-HEAD') ?` could be used instead of `key.slice(-5) === '-HEAD' ?`. There should be no compatibility problem with older browsers thanks to build tools, therefore we can use newer JS functions.
- However, I think the name must be changed to `contractIdFromLogHEAD` if you wish to keep it valid camelCase. IIRC, in camel case acronyms may be written all-caps at the end of a name, but not in the middle.

shared/domains/gi/db.js
```diff
@@ -37,9 +37,15 @@ export default (sbp('sbp/selectors/register', {
   'gi.db/set': function (key: string, value: string): Promise<void> {
     return Promise.resolve(sbp('okTurtles.data/set', key, value))
   },
+  'gi.db/delete': function (key: string) {
```
snowteamer: Shouldn't there be a `Promise<void>` return type annotation here like for 'gi.db/set'?

snowteamer (commented):
Nice improvements; I have only a few questions left.

frontend/model/database.js
snowteamer: > they could use the same client they were using, to run it as a server with a flip of a switch.

That's a very interesting idea!

shared/domains/gi/db.js
corrideat: I fixed this. It'll be in the next commit

Gruntfile.js
corrideat: I've addressed this now.

frontend/model/database.js
corrideat: Perhaps not, although I'm not sure what the alternative would be to not doing the checks (a runtime exception?). I've replaced this to `if (state.contractsXXX[contractId].HEADXX) {` for testing, and, as expected, no compile time errors occur. I guess that I've grown used to being a bit defensive in JS when it comes to these things. I can remove the null / undefined checks if you think a hard fail is better.

corrideat: @snowteamer What are your thoughts on this? Should I just remove the `?`s?

shared/domains/gi/db.js
```diff
@@ -37,8 +39,14 @@ export default (sbp('sbp/selectors/register', {
...
   },
+  'gi.db/delete': function (key: string): Promise<void> {
+    return Promise.resolve(sbp('okTurtles.data/delete', key))
+  },
   'gi.db/log/logHEAD': function (contractID: string): string {
-    return `${contractID}-HEAD`
+    return `${contractID}${headSuffix}`
+  },
+  'gi.db/log/contractIdFromLogHEAD': function (key: string): string | boolean {
+    return key.endsWith(headSuffix) ? key.slice(0, -headSuffix.length) : false
```
taoeffect: Wouldn't it make more sense to have the return type be `?string` and return `null` / `undefined` instead of `false`?

frontend/model/database.js
taoeffect: I had no idea that we could do this now. Is this a feature of `esbuild`? Because last I remember this required a special babel plugin...

I'm personally OK with it (Edit: actually, not really, see below), but I would remove some of the `?`s that you can know for sure don't need to be there. For example, checking `state?` isn't necessary, that is guaranteed to be there by the time this function is called. Each `?` leads to a branch, and the more branches you add to the code, the more things slow down (it adds up eventually).

Similarly, I think it's safe to remove the `?` for `.contracts?`, because if `state` is defined, that should be defined too as it's initialized with it.

Finally... maybe a hard fail is better, since by adding these `?`s I count the following downsides:

- Slightly slower code
- Loss of awareness of bugs
- Potential breakage if we switch bundlers (e.g. to Parcel #1107)

And only the following upside:

- `?.` is fantastic in situations where you *know* that values can be non-existent, and want to avoid uglifying your code with massive if/else blocks

Since this situation doesn't fit that — we know that all of these values should exist — I vote for removing them to catch bugs.

taoeffect (changes requested):
I think this is the final review, I see no other issues.

taoeffect (dismissed):
Approved! Looks good to me.

frontend/model/database.js
snowteamer: Yea I agree with greg here

frontend/model/database.js
```diff
@@ -9,12 +9,42 @@ const log = localforage.createInstance({
...
+        'gi.db/set': (key, value) => {
+          return Promise.resolve(value)
+        },
+        'gi.db/delete': () => {
+          return Promise.resolve()
+        }
+      }
+    : {
+        'gi.db/get': key => {
+          return log.getItem(`${key}`)
```
snowteamer: Since `key` should already be a string, there is no need for a template literal here; a type annotation should be enough. The same goes for `'gi.db/delete'` below.

shared/domains/gi/db.js
snowteamer: Personally I would have used `key.slice(0, -'HEAD'.length)` without defining a `headSuffix` constant. But that's no big deal. Do you have any thought on this @greg ?

snowteamer (dismissed):
LGTM. Minor comments left, but nothing blocking

taoeffect (approved):
Approved!

PR #1160, opened 2022-02-19
Fix app name references

snowteamer:


test/cypress/support/output-logs.js
```diff
@@ -3,18 +3,23 @@
 // Copied directly from: https://github.com/cypress-io/cypress/issues/3199#issuecomment-466593084
 // ***********
 
+const APPLICATION_NAME = 'group-income'
```
taoeffect, line 6: Will merge but in the future we should get this using `JSON.parse(fs.readFileSync(pathToPkgJson)).name`

taoeffect (approved):

test/cypress/support/output-logs.js
snowteamer: That's a better way indeed. Or `require(pathToPkgJson).name` if we can

taoeffect: Oh yeah, I totally forgot that `require` might be able to do that

PR #1164, opened 2022-02-23
Updated event parser and mocha tests.

CrazedCoding:
This version passes all the old mocha tests, and will pass the new ones created from clamping haver payments.

### DIFF
--- a/frontend/utils/distribution/monthly-event-parser.js
+++ b/frontend/utils/distribution/monthly-event-parser.js
@@ -1,6 +1,6 @@
 'use strict'
 import incomeDistribution from '~/frontend/utils/distribution/mincome-proportional.js'
-import { lastDayOfMonth, dateFromMonthstamp, dateToMonthstamp, addMonthsToDate } from '~/frontend/utils/time.js'
+import { lastDayOfMonth, dateFromMonthstamp, dateToMonthstamp } from '~/frontend/utils/time.js'
 import { cloneDeep } from '~/frontend/utils/giLodash.js'
 import minimizeTotalPaymentsCount from '~/frontend/utils/distribution/payments-minimizer.js'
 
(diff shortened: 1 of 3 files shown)


frontend/utils/distribution/monthly-event-parser.js
```diff
@@ -54,11 +52,49 @@ function subtractDistributions (paymentsA: Distribution, paymentsB: Distribution
...
+    const totalDistFrom = distributionFrom.reduce((a, p) => a + p.amount, 0)
+    const scalar = (totalDistFrom) / (haver.haveNeed - totalPaidFrom)
+    ratio = Math.max(ratio, scalar)
+  }
+
+  distribution = distribution.map(p => {
+    p.amount /= ratio
+    p.total /= ratio
+    return p
+  })
```
taoeffect: Is this the clamping code? If so, then is it possible to create a single clamping function to pass the distribution to and have it return a clamped version? Or something to that effect? (Have a function with the word "clamped" / "clamp" in it will also make it clear what the code is doing).

taoeffect (changes requested):
Nice job @CrazedCoding!

Left some comments, and also, two things:

(1) This PR surprisingly still doesn't handle the 100v100 scenario in my manual tests. After the $50 payment is made and the fifth 100 needer joins, here's the TODO list for u1:

Screen Shot 2022-02-23 at 2 55 28 PM

For reference, what we're calling "100v100" is this setup:

```
    setup = [{ type: 'haveNeedEvent', data: { name: 'u1', haveNeed: 100 } },
      { type: 'haveNeedEvent', data: { name: 'u5', haveNeed: 100 } },
      { type: 'haveNeedEvent', data: { name: 'u2', haveNeed: -100 } },
      { type: 'haveNeedEvent', data: { name: 'u3', haveNeed: -100 } },
      { type: 'paymentEvent', data: { from: 'u1', to: 'u2', amount: 50 } },
      { type: 'haveNeedEvent', data: { name: 'u4', haveNeed: -100 } }
    ]
```

(2) There are some Flowtype errors that need to be fixed, possibly by better defining JSONType. @snowteamer if you have suggests, I know you're really good at that stuff, they are welcome!

frontend/utils/distribution/monthly-event-parser.js
CrazedCoding: Yes and yes. I'll get right on it.

Also: is it ok if we say "scaling"? Clamping A between B and C refers to following operation (in JS):
`Math.max(Math.min(A, C), B)`

Where as this is actually scalar multiplication.

taoeffect:
I went ahead and opened #1169 based on this PR, so closing this in favor of that one. @CrazedCoding it needs your review!

Issue #1126, opened 2021-12-07
Design and implement question prompt


taoeffect:
### Problem

We do not have a way of forcing the user to answer an important question.

Our modal component is dismissable, and what's needed is a centered rectangle that asks a question and puts up a grey overlay over the background so that it's unclickable and to focus the prompt question.

### Solution

Design this component. /cc @leihla 

Implement it such that it can be displayed immediately in the app using a single line, e.g.:

```js
sbp('gi.ui/prompt', L('my question'), { okButton: L('Yes'), cancelButton: L('No') })
```

EDIT: this has now been designed here: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=21598%3A194029

leihla:
Looking at this scenario, and reading #1125, I've designed the question prompted assuming that we want to force the user to re-enter their income details after a change to the mincome has happened. However the undismissable modal can be used for other scenarios.

The modal is similar to other modals, without the close button, and should not be dismissed when clicking outside the modal area.

Link for figma: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=21526%3A193805

<img width="1441" alt="Screen Shot 2022-01-10 at 15 59 05" src="https://user-images.githubusercontent.com/158178/148797168-22de3289-3694-41bc-8dff-329954b99c5c.png">

taoeffect:
@pieer As Leihla has designed this now: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=21598%3A194029

I'm assigning the issue to you to implement.

taoeffect:
Closed by #1168

PR #1168, opened 2022-03-07
Feature #1126: Design and implement question prompt 

pieer:


taoeffect (changes requested):
Per the followup call and issue #1126, this PR is missing an API for putting up and displaying a question to the user.

Issue #1126 gave an example of such an API, but it isn't a very good one. Here's a better one, can you please implement this?

Example:

```js
const boolYesOrNo = await sbp('gi.ui/prompt', {
  question: L('Would you like fresh tacos?'),
  heading: L('Update info'),
  yesButton: L('Yes'),
  noButton: L('No')
})

if (boolYesOrNo) {
  // user said yes
} else {
  // user said no
}
```

Some notes:

- If that parameter for `yesButton` isn't passed in, by default make it use `L('Yes')`
- If that parameter for `noButton` isn't passed in, by default make it use `L('No')`
- If that parameter for `heading` isn't passed in, by default make it blank / empty / not visible

frontend/views/components/modal/README.md
```diff
@@ -74,3 +74,28 @@ A full-width Modal with just `transition` as wrapper. Note that this Modal has "
...
+```js
+sbp('okTurtles.events/on', MODAL_RESPONSE, function (response) {
+  sbp('okTurtles.events/off', MODAL_RESPONSE)
+  if (response) {
+    console.log('Yes')
+  } else {
+    console.log('No')
+  }
+})
+```
```
taoeffect: @pieer The intention behind issue #1126 is to create our own, better version of `window.prompt()`. So, that's why the API [that was requested](https://github.com/okTurtles/group-income/pull/1168#pullrequestreview-902530557) involves a single SBP call instead of 3.

Do you think you can create that API?

Also, do you think you could add an example to the DesignSystem? E.g. a button, when clicked, that brings up the prompt?

taoeffect (changes requested):
Nice, getting there! Left one comment.

shared/domains/gi/ui.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+import sbp from '~/shared/sbp.js'
+import { OPEN_MODAL, MODAL_RESPONSE } from '@utils/events.js'
+
```
taoeffect: As this file is not shared with the backend, it should be placed somewhere appropriate under `frontend/views`, for example, under `frontend/views/utils/` might be a good place.

taoeffect: As this file is not shared with the backend, it should be placed somewhere appropriate under `frontend/views`, for example, under `frontend/views/utils/` might be a good place.

taoeffect (changes requested):
@pieer Tested and it works well! Great job!

To merge need to fix:

1. Flow errors. There are two of them in `shared/domains/gi/ui.js`. (Both of them relate to using `Object` annotation)
2. Moving `shared/domains/gi/ui.js` into `frontend/views/utils/`
3. Prompt UI, see below

### 3. Prompt UI

Here's what it looks like in this PR:

Screen Shot 2022-03-30 at 9 09 53 AM

Here's what it should look (same color buttons, spaced close together):

Screen Shot 2022-03-30 at 9 09 24 AM

frontend/utils/events.js
```diff
@@ -9,6 +9,7 @@ export const OPEN_MODAL = 'open-modal'
 export const CLOSE_MODAL = 'close-modal'
 export const REPLACE_MODAL = 'replace-modal'
 export const SET_MODAL_QUERIES = 'set-modal-queries'
+export const MODAL_RESPONSE = 'modal_response'
```
taoeffect: Should we maybe keep this consistent with the others and use a dash instead of an underscore?

frontend/views/utils/ui.js
```diff
@@ -0,0 +1,17 @@
...
+import sbp from '~/shared/sbp.js'
+import { OPEN_MODAL, MODAL_RESPONSE } from '@utils/events.js'
+
+export default (sbp('sbp/selectors/register', {
+  'gi.ui/prompt': function (params: Object): Promise<*> {
+    sbp('okTurtles.events/emit', OPEN_MODAL, 'Prompt', null, params)
+
+    return new Promise((resolve, reject) => {
+      sbp('okTurtles.events/on', MODAL_RESPONSE, function (response) {
+        sbp('okTurtles.events/off', MODAL_RESPONSE)
```
taoeffect: You can avoid having to do `/off` here by doing `/once` instead of `/on`

frontend/views/utils/ui.js
```diff
@@ -0,0 +1,17 @@
...
+        resolve(response)
+      })
+    })
+  }
+}))
```
taoeffect: Fix flow error here - Flow needs at the end here:

```js
}): Object)
```

taoeffect (changes requested):
Thanks @pieer for the fixes!

There's a merge conflict with `master` and I left another review.

Also, the two buttons are very close to each other compared to how Leihla designed it. This can lead to users accidentally tapping on the wrong one on mobile, could make them further apart?

PR: 

Screen Shot 2022-04-26 at 3 57 40 PM

Figma:

Screen Shot 2022-04-26 at 3 58 30 PM

taoeffect (approved):
Great work @pieer! Pixel-perfect!

PR #1169, opened 2022-03-10
Non-prorated distribution algorithm + Minimization implemented

### LINKED ISSUE #1167: Remove `gi.actions/group/resetMonth` if possible
### Problem

While testing the live site I noticed that messages would be repeatedly created for this, possibly preventing the site from loading due to too much traffic going on.

### Solution

Find out if this message is necessary.

- If not, remove it
- If it is, find out why it kept getting sent over and over and make sure that can't happen

taoeffect:
### Changes

- Cleaned up distribution algorithm code
- Added optimizations to distribution algorithm
- Implemented non-prorated algorithm
- Implemented greedy minimization algorithm
- Removed `state.distributionEvents` - generated on-the-fly now via `geters.distributionEventsForMonth`. This makes it possible to get rid of `'gi.actions/group/resetMonth'` and related code.
- Removed `'gi.actions/group/resetMonth'`, `startMonthlyCycleCheckInterval`, and `startCycleEvent` (Closes #1167)
- Payments older than 2 months are now removed from the Vuex state
- Rewrote `mincomeProportional` so that it's cleaner and based on haveNeeds instead of incomes


CrazedCoding:
All checks passed, including mocha.

I'm app

CrazedCoding (approved):
Read through all of it. Seems sensible, mocha test pass, but I doubt that all cases maintain `mincomeProportional.js`'s total froms/tos. I'll adapt this to the visualizer for further validation. Approving.

PR #1175, opened 2022-03-14
Fix #1109: Darkmode color fixes

### LINKED ISSUE #1109: Darkmode color fixes
### Problem

Some of the colors for Darkmode are a bit off.

#### Dashboard bar colors

Screen Shot 2021-11-19 at 5 13 47 PM

Screen Shot 2021-11-19 at 5 14 28 PM

Something about these just doesn't look right.

#### Tooltip background

Screen Shot 2021-11-19 at 4 44 31 PM

The text is barely readable, and I can't quickly find in Figma what this was supposed to be. Was it designed? If so what should it be? If not, can it please be designed?

### Solution

Address above with a new color scheme.

pieer:
Update color and label


snowteamer:
Thank you for fixing this @pieer!

I've found a few places where the colors still need to be updated:
- When hovering over the info icon in Income Details
- When hovering over the info icon in the Invite Links List section in Group Settings
- When clicking on the default invite link or its accompanying clipboard icon, eiter in the same section or in the `InvitationLinkModal` modal

frontend/views/components/graphs/Overview.vue
```diff
@@ -7,7 +7,7 @@ div
         :amount='withCurrency(totalCovered)'
         color='warning-solid'
         variant='inline'
-      ) {{ L('Total covered') }}
+      ) {{ L('Total needed') }}
```
taoeffect: Why was this change made? It seems wrong, resulting in two "Total needed":

Screen Shot 2022-03-15 at 5 54 27 PM

taoeffect (changes requested):
Looking good @pieer! Only one issue found.

@snowteamer, I checked what you listed:

> I've found a few places where the colors still need to be updated:
> - When hovering over the info icon in Income Details
> - When hovering over the info icon in the Invite Links List section in Group Settings
> - When clicking on the default invite link or its accompanying clipboard icon, eiter in the same section or in the `InvitationLinkModal` modal

And all of these look fine to me? So what are you referring to?

Screen Shot 2022-03-15 at 5 52 21 PM
Screen Shot 2022-03-15 at 5 52 33 PM
Screen Shot 2022-03-15 at 5 52 49 PM

snowteamer:
@pieer @taoeffect I've checked these places again today, and now they all look fine for me too. Not sure why. :laughing: 

The notification badges have a darker red tint than before though. I think I liked the older red better. Would it be possible to keep it, e.g. by assigning red badges the old hex color #FF938C rather than` danger_0`?

frontend/views/components/graphs/Overview.vue
pieer: Request from @leihla ? https://github.com/okTurtles/group-income/issues/1109#issuecomment-997913747

leihla:
I've added both versions side by side: on left old #FF938C and on the right the new danger_0.

I would keep everything consistent and use danger_0. If you guys prefer the pink version then we need to add a new color swatch for that. @taoeffect?

Screenshot 2022-03-22 at 10 58 33
Screenshot 2022-03-22 at 10 58 24

taoeffect:
@leihla My personal preference is the darker red for the light theme, and the lighter red for the dark theme, at least as far as these screenshots go.

However, the lighter red - **IF** that is the old red that the original issue was complaining about for the bar graph - might not go well with the other bar graph colors...

~~Also, did you see @pieer's question here regarding the label? https://github.com/okTurtles/group-income/pull/1175#discussion_r830779891~~ EDIT: ah sorry, - per the Monday call, we decided it needs to be reverted.

frontend/views/components/graphs/Overview.vue
taoeffect: @pieer I asked @leihla about this just now and she said that the conclusion on the Monday call was to revert it back to what it was, i.e. leave it as `Total covered`.

snowteamer:
Same opinion as taoeffect on this point. If keeping different badge colors according to whether the theme is dark or not is not feasible, then maybe the dark theme badge digits should be white rather than dark gray?

pieer:
What's the conclusion on the colors @leihla @taoeffect ?

leihla:
@pieer as per my conversation with @taoeffect and @snowteamer preference, the goal is to keep the old color for the dark-theme badge #FF938C and the new color for the dark-theme graphs. Is that clear?

taoeffect (changes requested):
Plz fix failing tests

taoeffect (changes requested):
@pieer Thanks for fixing the tests!

I just tested this in FF Dev Edition, and the "total needed" bar color is black... definitely not right:

Screen Shot 2022-04-08 at 11 23 27 AM

taoeffect (approved):
Nice. 👍

PR #1177, opened 2022-03-16
Fix pubsub breaking after sleep and other improvements

### LINKED ISSUE #1127: Pubsub breaks with computer sleep... [$150 bounty]
### Problem

Context: I had the deployed site open in a pinned tab in the background (in other words, the pinned tab was not selected, I was doing other browsing, I put the computer to sleep, left it there for a few minutes, *stuff happened in the group while I was away*, woke the computer, went back to browsing, and then clicked on the pinned tab to see if saw any of the changes - in this case someone had joined the group).

These repeated errors appeared in Firefox console:

sleep issue

It's not clear to me what the issue is because the code that's linked to is related to error logging... and that could mean an error happened in the pubsub code *and* in the logging code:

```js
function captureLogsStart(userLogged) {
  isCapturing = true;
  username = userLogged;
  clearLogs();
  verifyLogsConfig();
  appLogsFilter = sbp_default("state/vuex/state").appLogsFilter || [];
  sbp_default("okTurtles.events/on", SET_APP_LOGS_FILTER, (filter) => {
    appLogsFilter = filter;
  });
  if (!isConsoleOveridden) {
    isConsoleOveridden = true;
    window.console = new Proxy(window.console, {
      get: (obj, type) => (...args) => {
        if (appLogsFilter.includes(type)) {
          obj[type](...args);                      // <---- console says error in 'get' on this line
          if (isCapturing) {
            captureLog(type, ...args);
            verifyLogsSize();
          }
        }
      }
    });
```

### Solution

Find the problem in both logging and pubsub and fix.

### Bounty

Issue is open to contractors or non-contractors. Non-contractors can get paid via bounty. We're posting a bounty of $150 for a clean solution to this issue.

snowteamer:
Closes #1127

### Summary of changes:
- Fix data synchronization issue after waking up from slep mode or somehow getting disconnected by the server.
- Less console output related to pings and pongs unless explicitly set in the options.
- `process.env.CI` is now available in shared and frontend files. Useful e.g. to enable or disable logging according to the environment.
- Updated `ws` dependency to v8.5.0.
- Socket `error` events are now logged using `console.log()` rather than `console.error()` to avoid possible confusion with runtime errors.

### DIFF
--- a/backend/pubsub.js
+++ b/backend/pubsub.js
@@ -252,7 +252,7 @@ const defaultMessageHandlers = {
       subscribers.add(socket)
       // Broadcast a notification to every other open subscriber.
       const notification = createNotification(type, { contractID, socketID })
-      server.broadcast(notification, { to: subscribers })
+      server.broadcast(notification, { to: subscribers, except: socket })
     }
     socket.send(createResponse(SUCCESS, { type, contractID }))
   },
@@ -270,7 +270,7 @@ const defaultMessageHandlers = {
         subscribers.delete(socket)
         // Broadcast a notification to every remaining open subscriber.
         const notification = createNotification(type, { contractID, socketID })
-        server.broadcast(notification, { to: subscribers })
+        server.broadcast(notification, { to: subscribers, except: socket })
       }
     }
     socket.send(createResponse(SUCCESS, { type, contractID }))
--- a/backend/pubsub.js
+++ b/backend/pubsub.js
@@ -50,6 +50,8 @@ export function createResponse (type: ResponseTypeEnum, data: JSONType): string 
  *
  * @param {(http.Server|https.Server)} server - A Node.js HTTP/S server to attach to.
  * @param {Object?} options
+ * {boolean?} logPingRounds - Whether to log ping rounds.
+ * {boolean?} logPongMessages - Whether to log received pong messages.
  * {object?} messageHandlers - Custom handlers for different message types.
  * {object?} serverHandlers - Custom handlers for server events.
  * {object?} socketHandlers - Custom handlers for socket events.
@@ -68,8 +70,8 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
     ...{ clientTracking: true },
     server: httpServer
   })
-  server.customServerEventHandlers = options.serverHandlers || {}
-  server.customSocketEventHandlers = options.socketHandlers || {}
+  server.customServerEventHandlers = { ...options.serverHandlers }
+  server.customSocketEventHandlers = { ...options.socketHandlers }
   server.messageHandlers = { ...defaultMessageHandlers, ...options.messageHandlers }
   server.pingIntervalID = undefined
   server.subscribersByContractID = Object.create(null)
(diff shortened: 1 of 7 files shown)


shared/pubsub.js
```diff
@@ -224,7 +238,7 @@ const defaultClientEventHandlers = {
   error (event: Event) {
     const client = this
 
-    console.error('[pubsub] Event: error', event)
+    console.log('[pubsub] Event: error', event)
```
taoeffect, line 245: errors should be logged with `.error`

shared/pubsub.js
```diff
@@ -359,15 +370,15 @@ const defaultMessageHandlers = {
...
   },
 
   [NOTIFICATION_TYPE.SUB] (msg) {
-    console.debug(`[pubsub] Ignoring ${msg.type} message:`, msg.data)
+    console.warn(`[pubsub] Ignoring ${msg.type} message:`, msg.data)
   },
 
   [NOTIFICATION_TYPE.UNSUB] (msg) {
-    console.debug(`[pubsub] Ignoring ${msg.type} message:`, msg.data)
+    console.warn(`[pubsub] Ignoring ${msg.type} message:`, msg.data)
```
taoeffect, line 388: Do any of these warnings happen often in our current usage?

taoeffect (changes requested):
Minor comments related to logging

shared/pubsub.js
snowteamer: It's not a programming error, only a so called "error" event. It happens e.g. when connectivity is lost or during the reconnection process when it takes more than one attempt. There is usually nothing to fix in such cases.

Meanwhile, our past Slack conversations regarding the break-after-sleep issue made me realize that this event could easily look more important than it actually is, when logged as an error. So I've been thinking about reducing the log level of these events.

shared/pubsub.js
snowteamer: AFAIK, never, but it happened during development of the fix, because the `except` parameter was missing in the pubsub server `broadcast()` call. It was nor really harmfull but still suboptimal, so I thought `warn` would be an appropriate severity level.

shared/pubsub.js
taoeffect: @snowteamer ok, then that needs to be explained in a comment here

taoeffect (changes requested):
In my testing of commit ca1170c via the live website, it appears to not work:

a

Here's what I did:

1. I created a new group with user A on machine A
2. Invited and added a user B on machine B
3. On machine B I turned off the wifi and put it to sleep
4. On machine A I updated user A's profile info (their display name and profile description)
5. Waited ~5 min
6. Woke up machine B and turned on wifi, expecting to see the updated changes, but instead saw the error in the screenshot above

corrideat (dismissed):

backend/pubsub.js
```diff
@@ -25,11 +25,27 @@ import type { JSONType } from '~/shared/types.js'
...
+const generateSocketID = (() => {
+  let counter = 0
+
+  return (debugID) => String(counter++) + (debugID ? '-' + debugID : '')
+})()
+
+const log = console.log.bind(console, tag)
+log.bold = (...args) => console.log(bold(tag, ...args))
+log.debug = console.debug.bind(console, tag)
+log.error = (...args) => console.error(bold.red(tag, ...args))
```
taoeffect, line 45: Do these lines end up modifying `console.log` for the rest of the server?

If so, does it make more sense to create a special function just for this file?

snowteamer: > Do these lines end up modifying console.log for the rest of the server?
No they don't, calling `.bind()` creates a new function and does not mutate anything.

By the way I've been considering defining this logger in its own shared module, but it turns out that console styling works quite differently in the browser and in Node.

taoeffect (approved):
Fantastic work @snowteamer!

PR #1188, opened 2022-03-25
Quick fix for undefined `state.notifications` after reloading the app using a newer version

snowteamer:


taoeffect (approved):
Looks good!

PR #1191, opened 2022-03-26
Fix response headers related to caching when serving avatar files

### LINKED ISSUE #1186: Add caching for images/files
### Problem

When the page is refreshed, it seems any user profile avatars that are redownloaded from the server even though they don't change.

Looking at the response header I see:

```
cache-control: no-cache
```

### Solution

Files retrieved via `'/file/{hash}'` should always be cached since they never change. Add appropriate caching headers to Hapi server's response and verify via the browser's network monitor that they are indeed being cached and not re-fetched.

snowteamer:
Closes #1186


taoeffect (approved):
Fantastic!

Issue #1192, opened 2022-03-26
Blank white page when loading many events


taoeffect:
### Problem

When loading a very long contract (using eventsSince `/events/{contract}/{since}`), the page will remain completely blank, but activity is going on.

### Solution

Hopefully when snapshotting is implemented (#354) this won't be as big of an issue, but it's still incorrect behavior.

What should happen is that the site fully loads the UI elements as much as it can, but tells the user it is loading data (#1185).

This issue is related to #1185.

snowteamer:
As an improvement I can suggest an optimization of `'state/latestContractState'`.

Currently it clones the whole contract state again for every processed message, in order to not break on possible errors. On error it would skip the offending GIMessage and just output a warning:
```js
console.warn(`latestContractState: ignoring mutation ${message.description()} because of ${err.name}`)
```
However, I don't remember observing such a warning in practice yet, so those errors must be rare.
Therefore I guess it would make sense to not clone the state as long as no error occurs, even if that requires starting the message processing over from scratch using the old cloning behavior as fallback in case of error.

taoeffect:
Juts a note for reference, the optimization was merged, but this bug is entirely unrelated to `latestContractState`. It has to do with waiting on `syncContractWithServer` instead of proceeding with loading the UI to the best of its ability and displaying a loading message of some sort.

taoeffect:
Note: this might be fixed by #1224 - will need to test it somehow

PR #1193, opened 2022-03-26
Optimize 'state/latestContractState'

snowteamer:
Related to #1192


taoeffect (approved):
OK, sure, this is an interesting optimization, although I don't think it is related to #1192 since that is not related to latestContractState AFAICT, but to the regular `handleEvent` path.

PR #1194, opened 2022-03-26
#1183 Fix behavior on reconnection failure

### LINKED ISSUE #1183: Pubsub doesn't reconnect after server has been restarted [$50 bounty]
### Problem

If the server has been restarted (say due to an update), the client doesn't reconnect once it's back up.

### Solution

Track down the bug and fix it.

EDIT: based on a convo on Slack w/@snowteamer, it looks like this is an issue that can occur when the client gives up reconnecting.

There are two issues here:

1. We should keep trying to reconnect indefinitely.
2. In the case where the setting for pubsub tells it to give up after some period, it should update the banner to reflect that we've given up and the user should try refreshing the page.

Screen Shot 2022-03-22 at 6 06 07 PM

### Bounty

Issue is open to contractors or non-contractors. Non-contractors can get paid via bounty. We're posting a bounty of $50 for a clean solution to this issue.

snowteamer:
Closes #1183

### Summary of changes:
- The WebSocket client now keeps retrying to reconnect indefinetely.
- In case the retry behaviour is modified to become limited again (e.g. via the `maxRetry` option), the corresponding banner message should now be correct.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -167,7 +167,7 @@ async function startApp () {
             banner.show(L('Trying to reconnect...'), 'wifi')
           },
           'reconnection-failed' () {
-            banner.show(L('Cannot connect to the server.'), 'wifi')
+            banner.show(L('We could not connect to the server. Please refresh the page.'), 'wifi')
           },
           'reconnection-succeeded' () {
             banner.clean()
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
LGTM!

taoeffect:
Actually, while the code changes make sense to me, I should test this on the live server to be extra sure. If it works I'll go ahead and merge.

PR #1195, opened 2022-03-29
#1184 Enable auto page reloading on GI update available

### LINKED ISSUE #1184: Have pubsub system send GI version & have clients refresh page on new version [$100 bounty]
### Problem

If the server restarts because of a new version of GI, clients should fetch the new code from it but currently don't.

### Solution

- For clients who've just woken up - have the new version sent to them upon reconnection and reload the page
- For clients that are active and online during the server restart - have them also receive the new version and reload the page as soon as the server is back online

### Bounty

Issue is open to contractors or non-contractors. Non-contractors can get paid via bounty. We're posting a bounty of $100 for a clean solution to this issue.

snowteamer:
Closes #1184

### Summary of changes:

- A new `process.env.GI_VERSION` variable has been made available to both the backend and the frontend. It is assigned a new value on every fresh build, but not on partial rebuilds. 
- In production mode, the pubsub server sends the backend `GI_VERSION` to the pubsub client as soon as it reconnects. This allows the frontend to emit a `GI_UPDATE_AVAILABLE` event if the versions don't match, which for now results in an automatic page reload.
  Note that this behavior only applies in production.  In development, BrowserSync already reloads the page as necessary, so there would be a conflict.
 - Minification via ESBuild options has been enabled in production. 
 
 ### Additional info:
 - To test these changes locally:
  1. Set `NODE_ENV` to `production` before running `grunt`.
  2. Open the app in the browser using port 8000, not 3000, because we don't want BrowserSync to interfere.
  3. Shutdown the server and run `grunt` again to rebuild everything. This will make the client start trying to reconnect.
  4. Once the server is up again, the client should need less than one more minute to reconnect.
  5. Upon reconnection the browser should log "New GroupIncome version available" and automatically reload the page.


Gruntfile.js
```diff
@@ -299,7 +307,7 @@ module.exports = (grunt) => {
     const fork2 = function () {
       grunt.log.writeln('backend: forking...')
       child = fork(backendIndex, process.argv, {
-        env: process.env,
+        env: { ...process.env, GI_VERSION },
```
taoeffect: Why was this modified? This doesn't look correct, `GI_VERSION` should simply be part of `process.env`.

Gruntfile.js
```diff
@@ -362,7 +370,7 @@ module.exports = (grunt) => {
 
   grunt.registerTask('default', ['dev'])
   // TODO: add 'deploy' as per https://github.com/okTurtles/group-income/issues/10
-  grunt.registerTask('dev', ['checkDependencies', 'backend:relaunch', 'build:watch', 'keepalive'])
+  grunt.registerTask('dev', ['checkDependencies', 'build:watch', 'backend:relaunch', 'keepalive'])
```
taoeffect, line 371: Why were these re-arranged?

frontend/controller/backend.js
```diff
@@ -31,6 +31,9 @@ function signJSON (json, keypair) {
   }, json)
 }
 
+console.info('GroupIncome version:', process.env.GI_VERSION)
+sbp('okTurtles.data/set', 'GI_VERSION', process.env.GI_VERSION)
```
taoeffect: I do not see a reason to create this, we should be able to just use `process.env.GI_VERSION` right?

Gruntfile.js
```diff
@@ -289,6 +295,8 @@ module.exports = (grunt) => {
     grunt.log.writeln('backend: launching...')
     // Provides Babel support for the backend files.
     require('@babel/register')
+    // Make `GI_VERSION` available via `process.env` before loading our server.
+    process.env.GI_VERSION = GI_VERSION
```
taoeffect: This should probably be set up at the top of this file.

Gruntfile.js
```diff
@@ -55,6 +55,7 @@ const { resolve } = path
 const {
   CI = '',
   LIGHTWEIGHT_CLIENT = 'true',
+  GI_VERSION = `v0.0.1@${new Date().toISOString()}`,
```
taoeffect: Version should be fetched from `package.json`.

taoeffect (changes requested):
Nice PR @snowteamer! Cool that esbuild has built-in minification features too, thanks for adding those! Left some feedback.

Gruntfile.js
snowteamer: It's not a member of `process.env` yet at this point, unless either it could be imported from the environment or `backend:relaunch` was already run, which should not be relied upon in my opinion. 

I've left the possibility to import `GI_VERSION` from the environment just in case, but by default it is supposed to be assigned a dynamic value.

Then I had to find a way to pass it down to the child process, without mutating `process.env` if possible. Hence this code in line 310, although the same was not possible in `backend:launch` so I added `process.env.GI_VERSION = GI_VERSION` there.

Gruntfile.js
snowteamer: Previously, the backend would have finished launching quite long before the new frontend assets were ready, so the client would already reconnect but get the old assets when reloading the page.

frontend/controller/backend.js
snowteamer: That's right. I added this to allow access to the version from within the console by doing `sbp('okTurtles.data/get', 'GI_VERSION')`. I've used it a few times while debugging this PR, but it could also be removed.

Gruntfile.js
snowteamer: In that case it would currently not change between builds.

Gruntfile.js
snowteamer: That's what I did initially. Then I thought it would be cleaner to not mutate `process.env` if possible.

Now since it's still going to be mutated in at least one case, one could argue that it would be better to be clear about it upfront and just put that mutation at the top of the file.

However, this looked a bit ugly. And `backend:launch` is the only place where the environment mutation is necessary, because of a local optimization to avoid creating a child process when launching the backend for the first time.

So I ended up moving this line here, but I can undo that if you think it's important.

Gruntfile.js
snowteamer: But I agree that the semver part of the version should come from the` package.json` file. Only the build-specific suffix should be dynamically assigned, assuming we can't fetch it from somewhere for now.

frontend/controller/backend.js
taoeffect: Yeah it's OK to have stuff that isn't necessary while debugging, but before the code is merged such things should be removed to keep the codebase as clean as we can. It's a constant battle to keep the codebase clean.

shared/pubsub.js
```diff
@@ -74,6 +74,7 @@ export type UnsubMessage = {
 
 export const NOTIFICATION_TYPE = Object.freeze({
   ENTRY: 'entry',
+  GI_VERSION: 'gi_version',
```
taoeffect: Can we make this more generic? The pubsub system should ideally be usable from any app.

Gruntfile.js
```diff
@@ -55,6 +58,7 @@ const { resolve } = path
 const {
   CI = '',
   LIGHTWEIGHT_CLIENT = 'true',
+  GI_VERSION,
```
taoeffect, line 61: This variable is no longer needed. Can replace all instances with `process.env.GI_VERSION`

Gruntfile.js
```diff
@@ -41,12 +49,7 @@ const applyPortShift = (env) => {
 
 Object.assign(process.env, applyPortShift(process.env))
 
-const chalk = require('chalk')
-const crypto = require('crypto')
-const { exec, fork } = require('child_process')
-const { copyFile, readFile } = require('fs/promises')
-const path = require('path')
-const { resolve } = path
+process.env.GI_VERSION = `${version}@${new Date().toISOString()}`
```
taoeffect, line 52: What's the reasoning for having this date thing tacked on at the end?

taoeffect (changes requested):
Re-reviewed!

Gruntfile.js
snowteamer: Its purpose is to ensure every build gets a distinct version, so that we don't have to bump our `package.json` version everytime a change should be deployed for live testing. Maybe I was wrong and it's actually not an issue.

Sure it will no longer be useful once we fully commit to using semver, unless maybe we keep doing private deployments for testing.

Also another suffix format for this thing might be nicer, if we keep one. Maybe a Unix timestamp rounded to the nearest second - just a suggestion.

Gruntfile.js
snowteamer: That's true, but then the `define` option block will become a bit inconsistent:

```js
  define: {
    'process.env.BUILD': "'web'", // Required by Vuelidate.
    'process.env.CI': `'${CI}'`,
    'process.env.GI_VERSION': `'${process.env.GI_VERSION}'`, // <-- Only line accessing `process.env`!
    'process.env.LIGHTWEIGHT_CLIENT': LIGHTWEIGHT_CLIENT,
    'process.env.VUEX_STRICT': VUEX_STRICT
  }

File: shared/pubsub.js (hunk 6 above)
Comment (snowteamer) [reply shared/pubsub.js]: Defining this type as just a plain string would allow to remove this `GI_VERSION` reference, at the cost of a small loss of Flowtype coverage.

File: Gruntfile.js (hunk 8 above)
Comment (taoeffect) [reply Gruntfile.js:52]: Hmm, ok, yeah we can keep this. Resolving

File: Gruntfile.js (hunk 7 above)
Comment (taoeffect) [reply Gruntfile.js:61]: OK, we'll keep it, resolving.

File: Gruntfile.js (hunk 2 above)
Comment (taoeffect) [reply Gruntfile.js:371]: Hmm. OK, as long as it doesn't break anything. Changes like these make me nervous as I feel there was a rhyme & reason to the order...

File: shared/pubsub.js (hunk 6 above)
Comment (taoeffect) [reply shared/pubsub.js]: I mean, can't it be modified to just `VERSION` without issue?

File: Gruntfile.js (hunk 7 above)
Comment (snowteamer) [reply Gruntfile.js:61]: By the way I should likely fix the `LIGHTWEIGHT_CLIENT` expression missing quotes.

File: shared/pubsub.js (hunk 6 above)
Comment (snowteamer) [reply shared/pubsub.js]: It can, that's what it was initially. Then I remembered what you said about identifiers like `id`, i.e. that they were too generic and many things in the app can have one. I guess many things can have their own version too, e.g. the browser, or js libraries. So I went with GI_VERSION instead.

Comment (taoeffect) [reply shared/pubsub.js]: @snowteamer then can you modify it to `APP_VERSION`?

File: frontend/controller/backend.js [hunk 9]
```diff
@@ -68,6 +76,14 @@ sbp('okTurtles.events/on', CONTRACTS_MODIFIED, (contracts) => {
   }
 })
 
+sbp('okTurtles.events/on', GI_UPDATE_AVAILABLE, (version) => {
+  console.info('New GroupIncome version available:', version)
```
taoeffect: I think we've been writing it as "Group Income" when it appears in text

taoeffect (changes requested):
@snowteamer fantastic! Left two very minor requests:

1. Switching `GI_VERSION` to `APP_VERSION` in `shared/pubsub.js` to make the pubsub stuff more generic
2. minor request for a `console.info` modification

Let's also merge #1196 first (since that will result in a conflict with this one), and then merge `master` into here, and once the review is address this should be good to be merged too.

taoeffect (dismissed):
Approved! Will merge after #1196 is merged & `master` is merged back into this to fix conflicts.

Gruntfile.js
taoeffect: I knew my spidy-sense was tingling!

Screen Shot 2022-04-03 at 4 42 09 PM

taoeffect (changes requested):
EDIT: oops - ignore this - this is not related to your PR, but to the fact that during testing the backend wipes everything that happened when it's relaunched. So - should be good!

Re-arranging `'backend:relaunch', 'build:watch'` breaks app development when a backend file changes:

Please try the following:

1. run `grunt dev`
2. connect in a browser to the app
3. create a group
4. open a file that exists on the backend and save it, making sure to see that `grunt dev` relaunched the backend server in the console
5. now try setting your income details, you should see errors:

Screen Shot 2022-04-03 at 4 42 09 PM

Try reverting that change and repeating the steps. It should work now.

taoeffect (approved):
Approved!

PR #1196, opened 2022-04-01
#1093 Re-enable esbuild code splitting

### LINKED ISSUE #1093: Restore code splitting to build system [$50 bounty]
### Problem

Since switching from rollup to esbuild our build system is ~6x faster on first build, and almost ~3x faster on rebuilds.

This is awesome. However, esbuild doesn't support code splitting, so all our `lazyLoadView` pages aren't being code-split into separate files. It's all going into the `main.js` bundle. This will slow down initial page load.

### Solution

~~Either wait for esbuild to add code-splitting and make any needed changes, or implement it ourselves.~~

EDIT: apparently esbuild does have code-splitting according to @snowteamer. So I'm assigning this to him. He says that the only issue is that the old chunks aren't deleted. This can be fixed by deleting them ourselves by creating a list of them prior to the build running, and deleting them after it finishes.

snowteamer:
Closes #1093

### Summary of changes

- Re-enable esbuild's code splitting, which had been disabled due to various issues (e.g. https://github.com/evanw/esbuild/issues/399).
  Note that the number of output chunks and their minimal size cannot be controlled and might not always be optimal (https://github.com/evanw/esbuild/issues/1716).
- New CLI option `--no-chunks` to disable code splitting and revert to the previous bahavior.
- Updated `esbuild` to v0.14.29.
- The `dist/` folder should now be emptied whenever a fresh build is started, to prevent progressive cluttering.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -332,7 +332,7 @@ module.exports = (grunt) => {
     const esbuild = this.flags.watch ? 'esbuild:watch' : 'esbuild'
 
     if (!grunt.option('skipbuild')) {
-      grunt.task.run(['exec:eslint', 'exec:flow', 'exec:puglint', 'exec:stylelint', 'copy', esbuild])
+      grunt.task.run(['exec:eslint', 'exec:flow', 'exec:puglint', 'exec:stylelint', 'clean', 'copy', esbuild])
     }
   })
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -139,7 +139,8 @@ module.exports = (grunt) => {
       },
       outdir: distJS,
       sourcemap: development,
-      splitting: !grunt.option('no-chunks'), // Warning: split mode has still a few issues.
+      // Warning: split mode has still a few issues. See https://github.com/okTurtles/group-income/pull/1196
+      splitting: !grunt.option('no-chunks'),
       watch: false // Not using esbuild's own watch mode since it involves polling.
     },
     // Native options used when building the main entry point.
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -121,6 +121,7 @@ module.exports = (grunt) => {
     // Native options that are shared between our esbuild tasks.
     default: {
       bundle: true,
+      chunkNames: '[name]-[hash]-cached',
       define: {
         'process.env.BUILD': "'web'", // Required by Vuelidate.
         'process.env.CI': `'${CI}'`,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -129,6 +129,7 @@ module.exports = (grunt) => {
         'process.env.VUEX_STRICT': VUEX_STRICT
       },
       external: ['crypto', '*.eot', '*.ttf', '*.woff', '*.woff2'],
+      format: 'esm',
       incremental: true,
       loader: {
         '.eot': 'file',
@@ -136,22 +137,19 @@ module.exports = (grunt) => {
         '.woff': 'file',
         '.woff2': 'file'
       },
+      outdir: distJS,
       sourcemap: development,
-      splitting: false, // Split mode has still a few issues so don't enable it yet.
+      splitting: !grunt.option('no-chunks'), // Warning: split mode has still a few issues.
       watch: false // Not using esbuild's own watch mode since it involves polling.
     },
     // Native options used when building the main entry point.
     main: {
       assetNames: '../css/[name]',
-      entryPoints: [`${srcDir}/main.js`],
-      format: 'esm',
-      outfile: `${distJS}/main.js`
+      entryPoints: [`${srcDir}/main.js`]
     },
     // Native options used when building our service worker(s).
     serviceWorkers: {
-      entryPoints: ['./frontend/controller/serviceworkers/primary.js'],
-      format: 'iife',
-      outdir: distJS
+      entryPoints: ['./frontend/controller/serviceworkers/primary.js']
     }
   }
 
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -129,29 +129,27 @@ module.exports = (grunt) => {
...
       loader: {
         '.eot': 'file',
         '.ttf': 'file',
         '.woff': 'file',
         '.woff2': 'file'
       },
+      outdir: distJS,
       sourcemap: development,
-      splitting: false, // Split mode has still a few issues so don't enable it yet.
+      splitting: !grunt.option('no-chunks'), // Warning: split mode has still a few issues.
```
taoeffect: Great idea to add this warning! Think it would help also to link to exactly what those issues are, for example, could link to this PR.

taoeffect (changes requested):
@snowteamer Close to perfect!

Two issues (one minor, one medium):

1. Minor: see request for link in comment
2. Medium: could you leverage [`--chunk-names`](https://esbuild.github.io/api/#chunk-names) to add `[hash]` and then have the server enable caching on all chunks (that have this unique `[hash]` appended to them?

snowteamer:
@taoeffect The `[hash]` placeholder is included by default in generated chunks when not using `chunk-names`, although it is only eight characters long.

In my tests, shared output chunk names looked like this: `chunk-DEFJT7KY.js`. Other chunks had names such as `Payments-6SBPA7KC.js`.
However, neither `main.js` nor `main.css` included a hash in their names. Anyway if they did, then `index.html` would have to be dynamically updated every build in order to reference them correctly. Mayb

As per the `chunk-names` docs,
> Note that this option only controls the names for automatically-generated chunks of shared code. It does not control the names for output files related to entry points.

Other assets like fonts and translations strings are also currently not hash-suffixed. IIRC I couldn't make it work without issues when switching from Rollup to Esbuild, at least not without changing the directory structure. In fact we copy these assets manually from `frontend/assets` to `dist/assets` using `grunt exec:copy`. But it might be worth investigating again.

taoeffect:
@snowteamer replied on Slack!

backend/routes.js
```diff
@@ -179,12 +179,25 @@ route.GET('/assets/{path*}', {
...
-    path: path.resolve('./dist/assets'),
-    redirectToSlash: true
+}, function (request, h) {
+  const { subpath } = request.params
+  const basename = path.basename(subpath)
+  const extension = path.extname(subpath)
+  console.debug(`GET /assets/${subpath}`)
+  // In the build config we told our bundler to use the `-cached-[hash]` template to name immutable assets.
+  // This is useful because `dist/assets/` currently includes a few files without hash in their name.
+  if (basename.endsWith(`-cached${extension}`)) {
```
taoeffect: I think `.indexOf('-cached') !== -1` would be more appropriate here, because this doesn't handle files with multiple dots near the end, for example, `/assets/js/chunk-P6QBAJFH-cached.js.map` isn't matched here, and neither would any other file that uses multiple `.`'s near the end.

taoeffect (changes requested):
Looking real good @snowteamer!

Caught just one minor bug, after that's fixed, will merge!

backend/routes.js
snowteamer: Good catch! I did think about the possibility of multiple extensions, but somehow got confused in the execution.

backend/routes.js
```diff
@@ -179,12 +179,25 @@ route.GET('/assets/{path*}', {
...
+    return h.file(subpath, { etagMethod: false })
+      .etag(basename.slice(0, -extension.length))
```
taoeffect: I think it would simplify and slightly speed up things here to just pass in the `basename` directly for the `.etag` here

snowteamer: Yea that would be simpler, however this also results in more bytes to be sent. Since the full extension if difficult to extract (it would require a regex), I now think that stripping the `-chunk` part instead is a good compromise.

backend/routes.js
```diff
@@ -179,12 +179,25 @@ route.GET('/assets/{path*}', {
...
+}, function (request, h) {
+  const { subpath } = request.params
+  const basename = path.basename(subpath)
+  console.debug(`GET /assets/${subpath}`)
+  // In the build config we told our bundler to use the `[name]-[hash]-cached` template
+  // to name immutable assets. This is useful because `dist/assets/` currently includes
+  // a few files without hash in their name.
+  if (basename.includes('-cached')) {
+    return h.file(subpath, { etagMethod: false })
+      .etag(basename.replace('-cached', ''))
```
taoeffect, line 195: My OCD & perfectionism is being triggered by this... there's no reason (that I can see) to perform additional operations here on the `basename` vs passing it in unmodified. Will approve though as soon tests pass, and modify it myself in a different PR, unless there are objections

taoeffect: Ah, just saw your comment in the other thread (posted after I wrote that above ☝️ )

> Yea that would be simpler, however this also results in more bytes to be sent. Since the full extension if difficult to extract (it would require a regex), I now think that stripping the -chunk part instead is a good compromise.

I don't think sending ~7 bytes more is going to make a difference. To be fair, I also don't think removing the `.replace()` will make much of a difference either. But as they say good coders are lazy - they do the minimum to get the desired results to avoid unnecessary complexity - this is really hard to do in practice (as I just proved by spending ~1 year chasing a more complicated distribution algorithm that in the end we ended up throwing away anyway because in practice it performed worse than the simpler posthumous distribution algorithm).

snowteamer: In my other local branch related to this issue, every file in `dist/assets/js` is immutable so the `-cached` infix is no longer needed. But for now you can always remove it if you'd like

taoeffect (approved):
Approved! Small future note about `.replace()` usage here: https://github.com/okTurtles/group-income/pull/1196/files#r841288306

backend/routes.js
snowteamer: Yea that's definetely not easy, I can only agree!

PR #1199, opened 2022-04-04
Display a global banner message while syncing contracts

### LINKED ISSUE #1185: Display banner while contracts are syncing
### Problem

Currently there's no visual indicator that the app is loading anything when we're syncing contracts.

### Solution

While a specific loading design might need to be made for this at some point (cc @leihla), in the meantime we can get away with using the global BannerGeneral.vue to display a message to the users letting them know that we're loading data (see main.js):

```js
sbp('okTurtles.data/set', 'BANNER', this.$refs.bannerGeneral)
```

A message like, "Loading events from server..." should appear. Have a look in `state.js` for clues on how to close this issue.

snowteamer:
Closes #1185

### Summary of changes:
- A simple message is now displayed using BannerGeneral when contracts are syncing.
  Please note that this behavior only activates when contracts are actually syncing, not while the client just checks with the server whether sync is needed.

### Additional info:
- Special testing should be done to ensure the banner doesn't sometimes flash too quickly. If this happens then additional code will be necessary to throttle or debounce the banner visibility changes.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -138,13 +138,12 @@ async function startApp () {
         // Make it possible for Cypress to wait for contracts to finish syncing.
         if (isSyncing) {
           this.ephemeral.syncs.push(contractID)
-        } else {
-          this.ephemeral.syncs = this.ephemeral.syncs.filter(id => id !== contractID)
-        }
-        if (this.ephemeral.syncs.length > 0) {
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Great job @snowteamer! Approved.

PR #1201, opened 2022-04-08
Workaround for esbuild bug and fix for lazy route loading

snowteamer:


frontend/utils/time.js
```diff
@@ -28,8 +28,9 @@ export function ISOStringToMonthstamp (date: string): string {
 }
 
 export function dateFromMonthstamp (monthstamp: string): Date {
-  // this is a hack to prevent new Date('2020-01').getFullYear() => 2019
-  return new Date(`${monthstamp}-01T00:01`)
+  // Append a time zone indicator (`Z`) to avoid unintuitive behaviors in some
+  // edge cases like `new Date('2020-01').getFullYear() => 2019`.
+  return new Date(`${monthstamp}-01T00:01Z`)
```
taoeffect: note: it's not just the Z that matters here, it's also the `-01T00:01` that was there before that's important to prevent the 2020->2019 thing from happening. As I've also updated this section in my upcoming PR, would it be OK to replace this function definition with the following?

```js
export function dateFromMonthstamp (monthstamp: string): Date {
  // this is a hack to prevent new Date('2020-01').getFullYear() => 2019
  return new Date(`${monthstamp}-01T00:01:00.000Z`) // the Z is important
}
```

This also has the important feature of showing the full skeleton of what an ISOString is supposed to look like, and it will help avoid a merge conflict.

snowteamer: Yea no problem

taoeffect (changes requested):
This PR is excellent! Almost perfect! Everything seems to be working correctly with `grunt dev`, but with `grunt dev --no-chunks`, I get:

Screen Shot 2022-04-07 at 8 03 18 PM

taoeffect (approved):
Excellent! Really phenomenal work here - greatly improved syntax for lazily loading modals and pages - and I tested `--no-chunks` - that works too. A++ would merge again!

PR #1204, opened 2022-04-09
Posthumous distribution algorithm implemented

### LINKED ISSUE #1159: Create getter for correct total payments needed value
### Problem

I'm fairly sure that right now we're not correctly calculating the total number of payments needed value (the "3" in these screenshots):

Screen Shot 2022-02-16 at 10 47 41 AM
Screen Shot 2022-02-16 at 10 47 04 AM

I think currently we're using the unadjusted distribution to get this value, although I could be misremembering. If we are, then that's incorrect.

### Solution

A naive solution would be to use the length of the TODO list (the adjusted distribution).

However, as the TODO list gets shorter when payments are made, that's incorrect too.

So the actual value should be: `length(TODO list) + length(total payments sent this month)`

Create a getter that returns that value, and bind it to the UI in the appropriate place.

### LINKED ISSUE #1171: Enable payment minimization by default and update Cypress tests
### Problem

Currently we don't have payment minimization enabled.

### Solution

Payment minimization is implemented in #1169. After it's merged, we need to:

- Verify that the distribution algorithm still correctly works with partial payments and full payments with minimization enabled
- Enable payment minimization by default
- Create a non-user-accessible way to toggle the app between minimized and unminimized payments
- Create new Cypress tests for minimized payment scenarios

### LINKED ISSUE #1197: Replace BypassUI.vue & all related Cypress tests with SBP
### Problem

BypassUI.vue was a smart way to perform actions in Cypress while speeding up the tests invented by @sandrina-p.

However, it represents a very loose and still somewhat clunky way of manipulating the app. Improvements could be made.

### Solution

 @snowteamer recently had the brilliant idea to include SBP directly in the Cypress tests as part of his updates to the Notification PR. This genius move does several things:

1. It makes it possible to directly initiate any app action from within the Cypress tests.
2. It makes it possible to completely remove the BypassUI mechanism.
3. It should speed up all of the Cypress tests.
4. It makes it much simpler to wait on events to complete - without having to rely on watching DOM attribute changes, by simply directly listening to SBP events emitted by the app. For example:

```js
    cy.window().its('sbp').then(sbp => {
        sbp('gi.actions/group/forceDistributionDate', {
          hooks: {
            prepublish: (message) => {
              sbp('okTurtles.events/on', message.hash(), resolve)
            }
          }
        })
      })
    })
```

To close this issue, send in a PR that:

1. Replaces all usage of BypassUI.vue with this new method
2. Removes BypassUI from the app, you can put it in `historical/` if you want to keep it around

taoeffect:
### Changes

+ Implements the posthumous distribution algorithm with a default distribution date of 3 days after group creation. In development mode you can force the distribution date to the current time using `'gi.actions/group/forceDistributionDate'`. In tests the same can be done with `cy.giForceDistributionDateToNow()`.
+ BypassUI replaced with direct SBP calls in tests. (Closes #1197)
+ Payments minimized by default (Closes #1171)
+ Getter for correct total payments needed value (Closes #1159)
+ `.etag` uses full filename
+ All monthstamp related stuff should now be migrated to a "period stamp" as part of the way the new distribution algorithm works.
+ Some strings have been updated
+ Copy during group creation has been updated to make it clearer what the mincome is and what the group is trying to accomplish with it
+ A vue-related console error related to `Contributions.vue` has been fixed
+ A vue-related console error when visiting the Payments page before any income details have been set has been fixed
+ Added `pickWhere` to `giLodash.js`
+ Mocha when run from Grunt now properly ignores the folders it needs to ignore
+ Removed `getTime()` from `time.js` and replaced its usage with `humanDate(time, { hour: 'numeric', minute: 'numeric' })`
+ Greatly simplified `monthName()` implementation without need of our own translations.
+ The payments page has a new view that lets users know when the distribution period begins. There are no designs for this, so I just added some text. /cc @leihla 
+ Fixed a bug in `objectMaybeOf()` related to validating empty strings
+ Whether to minimize the distribution is now a group setting that can be changed

If you'd like to test the changes locally in a browser I recommend setting up the group, setting the income details on all users, and when ready, running this in the console:

```js
```

That will trigger the payment TODOs to appear.


frontend/views/containers/contributions/ContributionsOverviewWidget.vue
```diff
@@ -3,10 +3,10 @@ page-section(v-if='showHistory' :title='L("Support History")')
   i18n.support-history.button.is-outlined.is-small(
     tag='span'
     @click='toggleHistory'
-  ) This month's overview
+  ) This period's overview
   support-history
 
-page-section(v-else :title='L("This months overview")')
+page-section(v-else :title='L("This period overview")')
```
taoeffect, line 9: Hmm, maybe we should keep these as "This month's overview"? (If we revert the change, be sure to revert it elsewhere.

Only thing is - by leaving it as "period" instead of month, we leave open the possibility for shorter or longer periods in the future if people want that.

CrazedCoding (approved):
Very interesting distribution algorithm.

frontend/controller/router.js
```diff
@@ -91,12 +90,6 @@ const router: any = new Router({
       meta: { title: L('Design System') }
       // beforeEnter: createEnterGuards(designGuard)
     },
-    {
-      path: '/bypass-ui',
-      component: BypassUI,
-      name: BypassUI.name,
-      meta: { title: L('Cypress - BypassUI') }
-    },
```
Silver-IT, line 99: If we are going to use `sbp` in the cypress for the ByPass, we don't need `views/pages/BypassUI.vue`, do we?

frontend/model/captureLogs.js
```diff
@@ -172,11 +172,10 @@ export function captureLogsStart (userLogged: string) {
   console.log(isNewSession ? 'NEW_SESSION' : 'NEW_VISIT', 'Starting to capture logs of type:', appLogsFilter)
 }
 
-export function captureLogsPause ({ wipeOut }: { wipeOut: boolean }): void {
+export function captureLogsPause ({ wipeOut, username }: { wipeOut: boolean, username: ?string }): void {
```
Silver-IT, line 175: I can see `SET_APP_LOGS_FILTER` event is emitted every time when `form.filter` is changed in AppLogs.vue.
But I don't know what it is exactly. Can you please tell me what it is for?

frontend/utils/time.js
```diff
@@ -4,61 +4,105 @@ import L from '~/frontend/views/utils/translations.js'
...
 
 export function addMonthsToDate (date: string, months: number): Date {
   const now = new Date(date)
   return new Date(now.setMonth(now.getMonth() + months))
 }
 
+// It might be tempting to deal directly with Dates and ISOStrings, since that's basically
+// what a period stamp is at the moment, but keeping this abstraction allows us to change
+// our mind in the future simply by editing these two functions.
+// TODO: We may want to, for example, get the time from the server instead of relying on
```
Silver-IT, line 17: Correct. Nice catch.

Silver-IT (commented):
Great work, @taoeffect.

frontend/model/captureLogs.js
taoeffect: Search the app for `SET_APP_LOGS_FILTER`. It's used by `AppLogs.vue` to set what kinds of logging the app will enable / listen for / display in the logs view.

frontend/controller/router.js
taoeffect: You're right we don't need that file anymore, but I've kept it in the project for now in case we decide we need to restore it for whatever reason. Currently it's not being imported, so it doesn't appear in the generated bundle.

Silver-IT: Why don't we save this in `historical` folder instead of the current location?

taoeffect: @Silver-IT 👍 Yes, good idea, feel free to move it in your PR after merging `master` 🙂

Silver-IT: Got it. I will do.

PR #1205, opened 2022-04-09
Fix 'process is undefined' error breaking autorefresh on upgrade

### LINKED ISSUE #1184: Have pubsub system send GI version & have clients refresh page on new version [$100 bounty]

snowteamer:
Closes #1184

- Fix an error which was introduced in the `APP_VERSION` socket message handler when applying review feedback in PR #1195.


taoeffect (approved):
LGTM!

PR #1207, opened 2022-04-12
Fix linting styles in modified .vue files

### LINKED ISSUE #1203: Fix stylelint error when linting .vue files on file save
### Problem

When saving a `.vue` file after `grunt dev` has been run, an error appears:

```
<input css 1>
 9:26  ✖  Unknown word   CssSyntaxError
```

This happens regardless of which `.vue` file is saved:

Screen Shot 2022-04-08 at 1 02 59 PM

### Solution

This appears to be an issue with `stylelint` in `Gruntfile.js`. The following code is relevant:

```js
      [['frontend/views/**/*.vue'], [puglint, stylelint, eslint]]
```

and from `scripts/esbuild-plugins/utils.js`:

```js
exports.createStylelinter = (options = {}) => {
```

This error **does not** appear when we run stylelint in any of the following ways:

- `npm run stylelint`
- `grunt exec:stylelint`

My guess is that when run via the stylelint API, the `"stylelint"` config in `package.json` isn't getting used.

So one possible solution is to make sure it uses that config too. Other possibly useful links include:

- https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/node-api.md#config
- https://github.com/stylelint/stylelint/blob/main/docs/user-guide/configure.md#processors
- https://github.com/stylelint/stylelint/blob/main/docs/user-guide/configure.md#customsyntax

snowteamer:
Closes #1203

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -216,6 +216,7 @@ module.exports = (grunt) => {
   // https://github.com/stylelint/stylelint/blob/master/docs/user-guide/usage/node-api.md#options
   const stylelintOptions = {
     cache: true,
+    config: require('./package.json').stylelint,
     formatter: 'string',
     syntax: 'scss',
     throwOnError: false,
--- a/frontend/views/components/Page.vue
+++ b/frontend/views/components/Page.vue
@@ -206,5 +206,4 @@ $pagePaddingDesktop: 5.5rem;
     }
   }
 }
-
 </style>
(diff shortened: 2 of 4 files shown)


snowteamer:
Closes #1203

taoeffect (approved):
Tested and it seems to work well.

Great job also on supporting mulitple `<style>` blocks per `.vue` file!

PR #1212, opened 2022-04-18
Quickfix for building the app on Windows

snowteamer:


scripts/esbuild-plugins/utils.js
```diff
@@ -58,7 +58,7 @@ exports.createAliasReplacer = (aliases) => {
 
     return source.replace(re, (match, capture) => {
       const resolvedPathSegment = aliases[capture]
-      const replacement = join(relativeDirPath, resolvedPathSegment)
+      const replacement = join(relativeDirPath, resolvedPathSegment).replace(/\\/g, '/')
```
taoeffect: Woah! Isn't `join` supposed to be platform agnostic? (The entire point of having a `join` function?)

snowteamer: `join` accepts paths containing either `/` or `\\`, but returns Windows paths when on Windows, which we don't want here since that whould change the `replacement` value

taoeffect: Oh ok, @snowteamer, could you add a comment explaining this then?

snowteamer: @greg Done :)

taoeffect (approved):
Quickfix approved! Thanks @snowteamer for the Windows support fix! 😎

PR #1218, opened 2022-04-22
Make sure to clear the global banner on every successful connection

snowteamer:


shared/pubsub.js
```diff
@@ -308,10 +308,9 @@ const defaultClientEventHandlers = {
     const client = this
     const { options } = this
 
-    if (!client.isNew) {
-      sbp('okTurtles.events/emit', PUBSUB_RECONNECTION_SUCCEEDED, client)
-    }
     client.clearAllTimers()
+    sbp('okTurtles.events/emit', PUBSUB_RECONNECTION_SUCCEEDED, client)
```
taoeffect, line 312: What listens for this event? I can't find any corresponding `/on` listener.

snowteamer: It's dynamically assigned in `shared/pubsub.js` along a bunch of similar events, so grepping for `PUBSUB_RECONNECTION_SUCCEEDED` won't work:
https://github.com/okTurtles/group-income/blob/a62d7c99cb8da0a2eb81ec339284a802ddf1413a/shared/pubsub.js#L645

Then the corresponding `reconnection-succeeded` shorthand is used in `main.js`:
https://github.com/okTurtles/group-income/blob/a62d7c99cb8da0a2eb81ec339284a802ddf1413a/frontend/main.js#L177-L179

taoeffect (approved):
Approved! Thanks for the quickfix @snowteamer, and for the explanation!

PR #1222, opened 2022-04-23
replace sbp with the npm pkg version

### LINKED ISSUE #1153: Migrate to NPM SBP stuff
### Problem

SBP is now published as an [npm package](https://github.com/okTurtles/sbp-js), along with [all of the okTurtles SBP domains](https://github.com/okTurtles/sbp-js/blob/master/docs/libraries.md).

We currently have a copy of all of this in the `shared/` folder.

### Solution

We should delete the following files and re-import them as dependencies:

- `shared/sbp.js`
- `shared/okTurtles/*`

SebinSong:
closes #1153


shared/domains/gi/db.js
```diff
@@ -32,6 +32,13 @@ export class ErrorDBConnection extends Error {
   }
 }
 
+const overritableSelectors = [
+  'gi.db/set',
+  'gi.db/get',
+  'gi.db/delete'
+]
+sbp('sbp/selectors/unsafe', overritableSelectors) // make these selectors overritable before registering
+
```
taoeffect, line 41: Nice! Really impressed you saw to do this! 😄 👍

frontend/model/database.js
```diff
@@ -13,39 +13,38 @@ const log = localforage.createInstance({
...
+        return log.getItem(key)
+      },
+      // TODO: handle QuotaExceededError
+      'gi.db/set': (key, value) => {
+        return log.setItem(key, value)
+      },
+      'gi.db/delete': (key: string) => {
+        return log.removeItem(key)
+      }
+    })
```
taoeffect, line 47: Great job. We should also lock the selectors after overwriting them so that they cannot be overwritten again:

```js
sbp('sbp/selectors/lock', ['gi.db/get', 'gi.db/set', 'gi.db/delete'])
```

backend/database.js
```diff
@@ -1,12 +1,12 @@
...
 
-import sbp from '~/shared/sbp.js'
+import sbp from '@sbp/sbp'
 import { strToB64 } from '~/shared/functions.js'
 import { Readable } from 'stream'
 import fs from 'fs'
 import util from 'util'
 import path from 'path'
-import '~/shared/domains/okTurtles/data.js'
+import '@sbp/okturtles.data'
```
taoeffect, line 9: After line 125 in this file, we should also lock the selectors after overwriting them so that they cannot be overwritten again:

```js
sbp('sbp/selectors/lock', ['gi.db/get', 'gi.db/set', 'gi.db/delete'])
```

taoeffect (changes requested):
Thank you @SebinSong! Looks really good!

Other than two minor changes requested, I don't see any issues, really impressive!

snowteamer (commented):
Great refactoring job @SebinSong!

I just have a question regarding `shared/domains/okTurtles/data.test.js` and shared/domains/okTurtles/events.test.js` - shouldn't these files be removed from this repo?

taoeffect:
Oh yeah, those test files should be removed too, good catch @snowteamer 👍

frontend/model/state.js
```diff
@@ -67,6 +67,15 @@ sbp('okTurtles.events/on', EVENTS.CONTRACT_IS_SYNCING, (contractID, isSyncing) =
...
 
+const isInMochaTest = typeof globalThis.it === 'function'
+
+if (isInMochaTest) {
+  sbp('sbp/selectors/unsafe', [
+    'state/enqueueHandleEvent',
+    'state/vuex/state'
+  ])
+}
+
```
SebinSong, line 78: Thanks @taoeffect, @snowteamer 

I thought these lines might need your look too. 
Have added this change for L:63 in `~/test/backend.test.js` which is another place where `sbp/selectors/overwrite` is called.
(used `globalThis` instead of `global` of node so that this script doesn't fail in browser. )

what's your thoughts?

SebinSong:
Hi @taoeffect,

I just got back for your CRs :), 
and it seems like `master` branch is currently getting this runtime error. (it worked fine yesterday, fyi)

image

by a quick look, 
it seems like `sbp('backend/translations/get', language)` call in `translations.js` returns `undefined`.

Thought I would give you a heads up. thanks.

taoeffect:
@SebinSong 

>  it seems like sbp('backend/translations/get', language) call in translations.js returns undefined.

Woah, can you please open an issue for that? I haven't seen that before, so maybe it's only reproducible in some cases, you could really help with this if that's true!

SebinSong:
In favor of @taoeffect 's PR (https://github.com/okTurtles/group-income/pull/1224) that contains the same work,
I'm closing this one.

PR #1223, opened 2022-04-23
Fix missing `state.currentChatRoomIDs` field on older clients upon upgrade

snowteamer:


taoeffect (approved):
Nice, thanks @snowteamer!

Issue #1214, opened 2022-04-21
Figure out versioning rules for upgrading / updating contracts and document them


taoeffect:
### Problem

Before we deploy contracts live we must have versioning all figured out.

Contracts cannot simply be updated and deployed again. There are multiple problems with simply overwriting old code with new code:

1. If a contract action is modified it will result in different state among clients because old clients processed v1 of that action, and new clients process v2, this can lead to two different states being generated for that contract.
2. If contract actions are removed it will wreak havoc on old messages that rely on them.
3. If a contract action is added, and there exist clients who have not upgraded to the new code, they will not be able to comprehend what to do, and the correct action in that case is to first load the latest code, not to ignore the action and move on ahead.
4. If we ever need to interact with contracts across servers (and this will almost certainly be the case in the future), we must be able to safely load the proper versions of those contracts without creating conflicts with similarly (or identically) named contracts on our side.

All of this is a potential issue because we may *need* to be able to upgrade contracts in order to fix bugs that weren't caught.

### Solution

We must make sure that clients are always using the correct code to read and process contract events.

A possible way to address all of these concerns would be to store, reference, and load contracts by their hash, and include that hash with every single ChelMessage. This method may be preferable to creating an `OP_CONTRACT_UPGRADE` opcode because it would make it possible for old clients, that haven't yet upgraded, to still (somewhat) safely send messages using the old version of the contract. However, see **Risks & Considerations** section below.

This would make it possible to "live-upgrade" an existing deployed contract without worry of conflicting state. If every message references the code that is used to process it, we can load that code before processing the message. The contract can initially be deployed based on `hash1` version of the contract, and future messages can be appended to it based on `hash2` version of that same contract, supporting both old and new versions of the same contract actions.

Note that this issue depends on dynamic contract loading & sandboxing (#1206).

It may also be necessary to create a CLI tool that "packages" up contracts under their hash so that they can be deployed. There's an existing issue for creating a CLI tool (#749) and this feature would need to be added to it.

We'd need to do at least two things:

1. Modify ChelMessage so that it includes the hash of the deployed contract
2. Create a special file format for defining a contract-to-be-deployed that includes its source, author keys, and a signature

Either the "special file" itself, or the ChelMessage, should support including two versions of the contract: a self-contained version (so that the contract can be processed on its own without the app that uses it), and an app-specific version that doesn't include external dependencies whose behavior won't meaningfully change (so that apps that already have the external dependencies loaded don't have to download massive redundant bundles).

[...]

PR #1224, opened 2022-04-24
Improve the state.js contract error handling logic (Closes #1165)

### LINKED ISSUE #731: Instead of skipping autoban during contract sync, do it after sync is finished
### Problem

Currently `autoBanSenderOfMessage()` will skip banning users if it happens during a contract sync. This is incorrect behavior, since it should still vote in the autoban if the proposal is open.

### Solution

1. Enqueue the autoban so that it happens after the contract sync
2. Vote in an open corresponding proposal if one exists, otherwise, open a proposal only if no previous proposal for this particular violation has been opened before, and the violation happened on a message within the past remove-member-proposal-period days

### LINKED ISSUE #1153: Migrate to NPM SBP stuff

### LINKED ISSUE #1165: Improve the state.js contract error handling logic
### Problem

At the moment, the way state.js works for contract errors (with the automated banning system) has a few issues:

- It's kinda whacky
- It's extraordinarly contract-specific: it assumes we're dealing with identities (the identity contract) whom we can ban from a group (the group contract)
- It kinda breaks sometimes

### Solution

- We must separate the contract error handling into a framework-level and an application-level
- We need to handle contract errors in a more generic way while still allowing the app itself to handle them in a more specific way (which includes the possibility of automated bans)
- We need to make sure that messages that lead to contract errors are not discarded and re-processed, but ignored
- We need to still allow application-specific error handling for errors
- We need to allow developers to create and invent their own error classes for exceptions and not rely on any sort of specific error classes - leaving that to the application-specific logic

Note: instead of complicated auto-banning logic, we could simply ignore those messages, and post a notification to the UI stating which user's client sent a bad message, and what they were trying to do, and leave it up to the group to deal with banning them.

### Additional considerations and examples

- Users might have situations where they modify a contract to remove a no longer needed action. For example, we did this with `'gi.contracts/group/resetMonth'`. This resulted in the entire contract being unusable because:

    ```
    processMutation: error Error Error: chelonia: action not whitelisted: 'gi.contracts/group/resetMonth'
    [ERROR] exception GIErrorDropAndReprocess in handleEvent while processing <op_ae|21XWnNM8SQzDsW6gMZLwhVfd8i2SXRc6rXHAzmdhSMXjDGrfj1 of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1>! Error during processMutation: chelonia: action not whitelisted: 'gi.contracts/group/resetMonth' GIErrorDropAndReprocess: Error during processMutation: chelonia: action not whitelisted: 'gi.contracts/group/resetMonth'
    reverting to previous state! 
    [ERROR] exception GIErrorDropAndReprocess in handleEvent while processing <op_ae|21XWnNW2vZzVq2ekurqehMKNdcVMQ5bEnu9841ocAjzHJeqoH5 of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1>! ignoring message until page refresh: <op_ae|21XWnNW2vZzVq2ekurqehMKNdcVMQ5bEnu9841ocAjzHJeqoH5 of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1> GIErrorDropAndReprocess: ignoring message until page refresh: <op_ae|21XWnNW2vZzVq2ekurqehMKNdcVMQ5bEnu9841ocAjzHJeqoH5 of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1>
    reverting to previous state! 
    [ERROR] exception GIErrorDropAndReprocess in handleEvent while processing <op_ae|21XWnNM6Df5tWeBJvNB9emnqMd5jeuyiXuFk5BuX4M7QHUCWDa of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1>! ignoring message until page refresh: <op_ae|21XWnNM6Df5tWeBJvNB9emnqMd5jeuyiXuFk5BuX4M7QHUCWDa of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1> GIErrorDropAndReprocess: ignoring message until page refresh: <op_ae|21XWnNM6Df5tWeBJvNB9emnqMd5jeuyiXuFk5BuX4M7QHUCWDa of 21XWnNP6RkUF7sXAUZ5pZE9PgN7RnXdwo5M1yyjMKXrmdRovW1>
    ...etc...
    ```

- See also issue #1176 for a different error
- Sometimes we get a `[CRITICAL ERROR] NOT EXPECTING EVENT!` because things happen too fast - removing a contract while still having not processed some messages for it, as for example [here](https://github.com/okTurtles/group-income/pull/1143/files#diff-0e4600578c772d3d71d59082adec0332af88c00c5dacdb5d9ed97e5cdf40fc28R88) and [here](https://okturtles.slack.com/archives/C0EH7P20Y/p1649298837056349).

This issue also depends on #1214 being closed:

> [Slack] Greg: the issue is that right now the data and computation are separated, which means for example, if the rule is that if exceptions cause events to be ignored and skipped over (which likely seems to be the case for #1165), then once the issue in the contract has been fixed to prevent the exception, anyone who joins afterward and resyncs the old event (that existing clients skipped), they won’t encounter the exception, they will process the message, and therefore will have a different state from the other clients. This is a huge no-no, so I need to close #1214 along with #1165, so that each event specifies the exact code it uses to process itself

### LINKED ISSUE #1176: Contract sync error is there in group contract
### Problem

Let's think about the following workflow

1. user1 creates a group
2. user2 joins the group and then user3 joins too with the same invitation link
3. user2 proposes to remove user3 and user1 approves. IMPORTANT thing is that user3 needs to be offline at this moment meaning that user3 doesn't sync the contract at this moment.
4. user2 leaves the group by himself
5. user3 tries to login and join with the invitation link but get message of `Oh no! This invite is not valid` and fails to join group.

image

### Solution

I guess there is a mistake in syncing group contract inside `gi.actions/group/join` action.
Need to track the contract actions backwards one by one and find the reason out.

### LINKED ISSUE #1217: Remove the deprecated Mailbox/Inbox
### Problem

This page and its navbar item should be removed:

Screen Shot 2022-04-21 at 10 58 36 PM

### Solution

Remove from the project.

### LINKED ISSUE #1220: debounce the banner on contract sync
### Problem

Sometimes the page kinda "jumps" because the contract sync is happening but it happens too fast.

### Solution

The banner should appear if the contract sync is taking more than 2 seconds to finish, and should remain open for at least 2 seconds if it appears.

Use the `debounce` function from `giLodash.js` to accomplish this.

taoeffect:
### Changes

+ Moved all Chelonia-related Vuex stuff into Chelonia SBP framework (Closes #1165)
+ Autoban after contract sync (Closes #731)
+ Removed Mailbox.vue (Closes #1217)
+ May have fixed blank white page on large amount of events (#1192)
+ Added debounce to contract sync banner (Closes #1220)
+ Migrate to NPM SBP stuff (Closes #1153 - my mistake for not assigning to myself when @SebinSong did the same thing!)
+ Added a-link support to `v-safe-html` via `v-safe-html:a`
+ Moved `logExceptNavigationDuplicated` into `@view-utils` since it's a vue-related utility
+ Removed TimeTravel.vue
+ Added some more hacks for esbuild failing to load chunks for chatroom.js contract
+ Changed pubsub logging from `.log` to `.debug` and `.warn` and `.error` as necessary
+ I don't know if this PR closes #1176 or not, but the error is gone. However, u3 is now able to rejoin the group after they've been kicked using the original invite link. Should open a new issue for that
+ Display user notifications on contract errors
+ Updated `esbuild` to `0.14.38` and `esbuild-sass-plugin` to `2.2.6` 

The new contract error handling rules are:

- Exceptions thrown during mutation processing (e.g. either by `validate` or by `process` functions) will result in that message being ignored and any side-effects skipped. Unless `ChelErrorUnrecoverable` is thrown, in which case that contract comes to a permanent halt (usually not something you want).
  - Until #1214 is closed, this rule can lead to divergent state among clients. Example: a process function throws an exception and is ignored. Later that contract is updated to not throw the exception. Clients who join later will process the mutation, resulting in a different state from older clients.
- Exceptions thrown during side-effect processing (e.g. either by a pushed side effect or in the `sideEffect` function or in the `postHandleEvent` hook or in an event listener) will result in that message getting dropped and the application coming to a halt until the bug in the side effect is fixed, after which a page refresh will resume normal functioning of the app.

Additionally, applications can modify the behavior of exceptions of process functions by implementing the `processError` hook. For example, if `group.js` throws `GIErrorIgnoreAndBan`, it will autoban that user. Chelonia can be configured with various hooks and event listeners for further customization, see changes in `main.js` for more examples.


backend/database.js
```diff
@@ -1,13 +1,13 @@
...
+import sbp from '@sbp/sbp'
 import { strToB64 } from '~/shared/functions.js'
 import { Readable } from 'stream'
 import fs from 'fs'
 import util from 'util'
 import path from 'path'
-import '~/shared/domains/okTurtles/data.js'
-import '~/shared/domains/gi/db.js'
+import '@sbp/okturtles.data'
+import '~/shared/domains/chelonia/db.js'
```
snowteamer, line 10: So the DB stuff is now a part of Chelonia?

backend/routes.js
```diff
@@ -92,7 +92,7 @@ route.GET('/name/{name}', {}, async function (request, h) {
 
 route.GET('/latestHash/{contractID}', {}, async function (request, h) {
   try {
-    const entry = await sbp('gi.db/log/lastEntry', request.params.contractID)
+    const entry = await sbp('chelonia/db/lastEntry', request.params.contractID)
```
snowteamer: Shouldn't there be a `'chelonia/db/latestHash'` selector? This way a route handler wouldn't have to know it needs to call `.hash()` on an entry to get its hash.

frontend/controller/actions/group.js
```diff
@@ -345,6 +348,80 @@ export default (sbp('sbp/selectors/register', {
...
+          const prop = contractState.proposals[hash]
+          if (prop.status === STATUS_OPEN &&
+            prop.data.proposalType === PROPOSAL_REMOVE_MEMBER &&
+            prop.data.proposalData.member === username
+          ) {
+            proposal = prop
+            proposalHash = hash
+            break
+          }
+        }
```
snowteamer: This could be made more concise, for example:
```suggestion
        // Find existing proposal if it exists.
        let [ proposalHash, proposal ] = Object.entries(contractState.proposals)
          .find(([hash, prop]) => (
            prop.status === STATUS_OPEN &&
            prop.data.proposalType === PROPOSAL_REMOVE_MEMBER &&
            prop.data.proposalData.member === username
          )) ?? []
```

frontend/controller/backend.js
```diff
@@ -2,81 +2,18 @@
...
-      client.sub(contractID, dontBroadcast)
-    }
-  } catch (e) {
-    // TODO: handle any exceptions!
-    console.error('CONTRACTS_MODIFIED: error in pubsub!', e, { toUnsubscribe, toSubscribe })
-  }
-})
-
-sbp('okTurtles.events/on', GI_UPDATE_AVAILABLE, (version) => {
+sbp('okTurtles.events/on', NOTIFICATION_TYPE.APP_VERSION, (version) => {
```
snowteamer, line 16: Strictly speaking, `APP_VERSION` is a message type not an event type, but I guess that's an acceptable shortcut.

frontend/views/components/banners/BannerScoped.vue
```diff
@@ -27,6 +28,10 @@ export default ({
     dataTest: {
       type: String,
       default: 'feedbackMsg'
+    },
+    allowA: {
```
snowteamer, line 32: What does `allowA` mean?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -83,7 +83,7 @@ import { MESSAGE_TYPES, MESSAGE_ACTION_TYPES, MESSAGE_VARIANTS } from '@model/co
 import { createMessage, getLatestMessages } from '@model/contracts/chatroom.js'
 import { proximityDate, MINS_MILLIS } from '@utils/time.js'
 import { CHATROOM_MESSAGE_ACTION, CHATROOM_STATE_LOADED } from '~/frontend/utils/events.js'
-import { CONTRACT_IS_SYNCING } from '@utils/events.js'
+import { CONTRACT_IS_SYNCING } from '~/shared/domains/chelonia/chelonia.js'
```
snowteamer: No `'~/shared/domains/chelonia/events.js'` file from where to import event names?

frontend/views/utils/vSafeHtml.js
```diff
@@ -29,8 +30,12 @@ export const defaultConfig = {
...
 const transform = (el, binding) => {
   if (binding.oldValue !== binding.value) {
-    const config = { ...defaultConfig, ...(binding.arg || {}) }
-
+    let config = defaultConfig
+    if (binding.arg === 'a') {
+      config = cloneDeep(config)
+      config.ALLOWED_ATTR.splice(config.ALLOWED_ATTR.length, 0, 'href', 'target')
+      config.ALLOWED_TAGS.push('a')
+    }
```
snowteamer, line 38: Hmm now I understand that `allowA` was meant for allowing anchor tags in `v-safe-html `directives.
Now that I remember, there is a configurable safe URL whitelist in `utils/allowedUrls.js`, couldn't it have been used?

shared/domains/chelonia/GIMessage.js
```diff
@@ -46,8 +46,7 @@ export class GIMessage {
     op: GIOp,
     signatureFn?: Function = defaultSignatureFn
   ): this {
-    const instance = new this()
-    instance._message = {
+    const message = {
       version: 1.00,
```
snowteamer: Shouldn't `version` be a SemVer string instead?

backend/database.js
taoeffect: Yep. Some selectors are designed to be overwritten by the app

frontend/views/components/banners/BannerScoped.vue
taoeffect: It means allow `a` tags to be rendered

frontend/views/utils/vSafeHtml.js
snowteamer: There is a configurable whitelist in `utils/allowedUrls.js`, couldn't have been used?

shared/domains/chelonia/chelonia.js
```diff
@@ -62,18 +91,60 @@ sbp('sbp/selectors/register', {
...
+  'chelonia/configure': function (config: Object) {
+    merge(this.config, config)
+    // merge will strip the hooks off of config.hooks when merging from the root of the object
+    // because they are functions and cloneDeep doesn't clone functions
+    merge(this.config.hooks, config.hooks || {})
+  },
+  'chelonia/connect': function (): Object {
+    if (!this.config.connectionURL) throw new Error('config.connectionURL missing')
+    if (!this.config.connectionOptions) throw new Error('config.connectionOptions missing')
+    if (this.socket) {
```
snowteamer: Since `this.socket` here is actually not a WebSocket object, I think something like `this.client` or `this.pubsub` would be more correct

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: Yeah that would be nicer, good suggestion

frontend/controller/actions/group.js
taoeffect: Yeah that looks like a good improvement, but is it me or would this code throw some kind of an error?

```
> d = { a: 4 }
{ a: 4 }
> Object.entries(d)
[ [ 'a', 4 ] ]
> let [i, j] = Object.entries(d).find(() => false)
Uncaught:
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

Or am I doing something wrong?

Silver-IT: > let [i, j] = Object.entries(d).find(() => false)

Two variables `i` and `i` are set from the result of `find` which returns `undefined` always.
So it's just like `let [i, j] = undefined;`

If you try with the following code, it won't make any errors if object `d` is not empty.
`let [i, j] = Object.entries(d).find(() => true)`

taoeffect: Ah, but that's the thing, `proposals` might not contain the proposal we're looking for, so in that case this would error.

frontend/controller/backend.js
taoeffect: Good point. I updated this since we're listening for a framework-provided (not GI-provided) event

shared/pubsub.js
```diff
@@ -243,7 +244,7 @@ const defaultClientEventHandlers = {
   error (event: Event) {
     const client = this
 
-    console.log('[pubsub] Event: error', event)
+    console.error('[pubsub] Event: error', event)
```
snowteamer: Not all `error` events should be logged with `console.error`, for example every failed connection attempt generates one such event.

docs/Information-Flow.md
```diff
@@ -70,19 +70,19 @@ In our example above, the name of the *contract action* generated by `sbp('gi.ac
...
 
 ```js
-sbp('gi.actions/contract/sync', contractId)
+sbp('chelonia/contract/sync', contractID)
 
-// OR
+// OR wait until it finishes syncing:
 
-await sbp('gi.actions/contract/syncAndWait', contractId)
+await sbp('chelonia/contract/sync', contractID)
```
snowteamer, line 82: Nice API :)

frontend/main.js
```diff
@@ -127,21 +146,41 @@ async function startApp () {
...
+          } else {
+            // after first 2 seconds, check every half a second for sync to finish
+            setTimeout(clearBanner, 500)
+          }
+        }
+        if (!clearBannerTimer && this.ephemeral.syncs.length) {
+          bannerGeneral.show(L('Loading events from server...'), 'wifi')
+          clearBannerTimer = setTimeout(clearBanner, 2000)
+        }
+      }, 2000)
```
snowteamer: I was working on that issue too. Checking whether the banner has already a higher severity status is a great idea!

For my part, my solution filters out contract IDs from `state.pending`, to avoid triggering such banner activity when signing up or creating a new group.

frontend/main.js
```diff
@@ -194,6 +232,29 @@ async function startApp () {
...
+          //         lookup failed! TypeError: "state[state.currentGroupId] is undefined"
+          //         memberUsernames state.js:231
+          //
+          //       Which doesn't mean that the lookup actually failed!
+          await sbp('gi.db/settings/delete', username)
+        }
+      }
+      // useful to delay the initialization of various pages like Join.vue until
+      // we have finished setting up and logging in
+      sbp('okTurtles.events/emit', VUE_LOADED)
```
snowteamer: Awesome idea too!

frontend/model/notifications/templates.js
```diff
@@ -12,6 +12,33 @@ const escapeForHtml = (text) => text.replace(/[<>&]/g, '\\$&')
...
+  },
+  ERROR (data: { contractID: string, message: string }) {
+    return {
+      body: data.message,
+      icon: 'exclamation-triangle',
+      level: 'danger',
+      linkTo: `/app/dashboard?modal=UserSettingsModal&section=application-logs&errorMsg=${encodeURI(data.message)}`,
+      scope: 'app'
+    }
+  },
```
snowteamer, line 41: At first guess I would say that errors and warnings would rather belong in the global banner or something similar. But maybe it's already used for too many things.

frontend/views/utils/vSafeHtml.js
```diff
@@ -29,8 +30,12 @@ export const defaultConfig = {
 
 const transform = (el, binding) => {
   if (binding.oldValue !== binding.value) {
-    const config = { ...defaultConfig, ...(binding.arg || {}) }
-
+    let config = defaultConfig
+    if (binding.arg === 'a') {
+      config = cloneDeep(config)
+      config.ALLOWED_ATTR.splice(config.ALLOWED_ATTR.length, 0, 'href', 'target')
```
snowteamer: This `splice` call doesn't seem to remove any element, only to add `href` and `target`. So did you mean the following?
```suggestion
      config.ALLOWED_ATTR.push('href', 'target')
```

frontend/main.js
snowteamer: In the related issue however, the delays were different, like "show the banner after 1s then don't hide it again before another 2s"

frontend/controller/actions/group.js
snowteamer: Sorry for leaving this case out.

Update: appended `?? []` to the suggested code to make it return and assign `undefined` values in case no proposal is found. But I find it a bit less clean now.

frontend/main.js
snowteamer: Also I feel like more of this banner-related code should be moved to `BannerGeneral.vue` rather than stay in `main.js`.

docs/Information-Flow.md
taoeffect: Thanks :)

frontend/main.js
taoeffect: > In the related issue however, the delays were different, like "show the banner after 1s then don't hide it again before another 2s"

Yeah, I edited the issue because I decided they should be slightly tweaked. 1 second to read a message seemed a bit short to me.

> Also I feel like more of this banner-related code should be moved to BannerGeneral.vue rather than stay in main.js.

Good point, I'll add a new `debouncedShow` method to BannerGeneral.vue and will push

frontend/views/utils/vSafeHtml.js
taoeffect: Lol yes, for some reason I thought `.push` only accepted one argument, learn something new every day. Thanks! :)

shared/pubsub.js
taoeffect: noted! will be changed to `console.warn`

test/cypress/support/commands.js
```diff
@@ -167,9 +167,6 @@ Cypress.Commands.add('giCreateGroup', (name, {
     cy.getByDT('app').then(([el]) => {
       cy.get(el).should('have.attr', 'data-sync', '')
     })
-
-    checkIfJoinedGeneralChannel(name)
-
```
Silver-IT, line 172: Actually, I am not sure if this removal breaks cypress test...

taoeffect: They passed though didn't they

Silver-IT (commented):
Actually, I can not say I have understood all the code lines but I agree it's a really impressive and you did a great work on this. Awesome~~~
Just one thing to ask regarding `checkIfJoinedGeneralChannel` function.

Silver-IT (approved):
Approved. ~~~

test/cypress/support/commands.js
Silver-IT: Yeap. Nice catch. It should not be here. It's for ByPassUI.

Issue #1208, opened 2022-04-12
Implement payment distribution date dropdown


taoeffect:
### Problem

The new posthumous distribution needs a distribution date picker!

### Solution

Implement this design: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=22579%3A203497

Do not set a default date. Make sure that the user cannot move on until they select a date themselves.

taoeffect:
I think this has been closed in #1229

PR #1229, opened 2022-04-25
Feature #1208: Implement payment distribution date dropdown

pieer:


frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,21 +38,59 @@
...
+          required=''
+          :value='group.distributionDate.toString()'
+          @change='update'
+        )
+          i18n(tag='option' disabled value='') Choose day
+          option(
+            v-for='(item, index) in ephemeral.distributionDayRange'
+            :key='index'
+            :value='item'
+          ) {{ new Date(item).getDate() }}
```
taoeffect: This should probably use `humanDate()` for a more friendly display of the date.

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,21 +38,59 @@
...
+            i.icon-question-circle
+          template(slot='tooltip')
+            i18n(tag='p') Select which day of the month should the distribution be calculated. Every group member will need to update their mincome details before this date to ensure that the algorythm can fairly distribute available funds between group members.
+
+      .selectbox
+        select.select(
+          :aria-label='L("Choose day")'
+          name='distributionDate'
+          required=''
+          :value='group.distributionDate.toString()'
```
taoeffect: I don't think this should have `.toString()` as the value, as then it's not bound to the variable

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,21 +38,59 @@
...
       i18n.helper This value can be adjusted in the future.
 
+    fieldset.field
+      .label.c-label-tooltip
+        i18n In what day should the first payment distribution be calculated?
+        tooltip(direction='bottom-end')
+          span.button.is-icon-small
+            i.icon-question-circle
+          template(slot='tooltip')
+            i18n(tag='p') Select which day of the month should the distribution be calculated. Every group member will need to update their mincome details before this date to ensure that the algorythm can fairly distribute available funds between group members.
```
taoeffect: `algorythm` => `algorithm`

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,21 +38,59 @@
...
+
+      .selectbox
+        select.select(
+          :aria-label='L("Choose day")'
```
taoeffect:
```suggestion
          :aria-label='L("Choose your group's distribution date")'
```

taoeffect (changes requested):
Nice job @pieer!

Left a few review comments! I recommend merging `master` into this PR too.

Also, an issue that seems to show up in Dark Mode:

Screen Shot 2022-04-26 at 4 16 04 PM

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,27 +38,66 @@
...
       i18n.helper This value can be adjusted in the future.
 
+    fieldset.field
+      .label.c-label-tooltip
+        i18n In what day should the first payment distribution be calculated?
+        tooltip(direction='bottom-end')
+          span.button.is-icon-small
+            i.icon-question-circle
+          template(slot='tooltip')
+            i18n(tag='p') Select which day of the month should the distribution be calculated. Every group member will need to update their mincome details before this date to ensure that the algorithm can fairly distribute available funds between group members.
```
taoeffect: > Select which day of the month should the distribution be calculated.

Should be:

> Select which day of the month the distribution should be calculated.

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,27 +38,66 @@
 
       i18n.helper This value can be adjusted in the future.
 
+    fieldset.field
+      .label.c-label-tooltip
+        i18n In what day should the first payment distribution be calculated?
```
taoeffect: > In what day should the first payment distribution be calculated?

To:

> On what day should the first payment distribution be calculated?

frontend/views/components/group-creation-steps/GroupMincome.vue
taoeffect: Don't forget this plz

frontend/views/components/group-creation-steps/GroupMincome.vue
```diff
@@ -38,27 +38,66 @@
...
     return {
-      currencies
+      currencies,
+      ephemeral: {
+        distributionDayRange: []
+      }
+    }
+  },
+  beforeMount () {
+    for (let index = 0; index < 30; index++) {
```
taoeffect: I noticed an issue in my testing, which is that with this method (at least in my timezone), the previous day will show up in the dropdown menu.

So if today is May 3rd, then May 2nd will show up.

To fix, could you update this to the following?

```js
for (let index = 1; index <= 30; index++) {
```

taoeffect (changes requested):
Looking really good, I think this is the last set of issues that I was able to find. There's also one comment from the previous review that didn't get addressed (for `aria-label`).

taoeffect (approved):
Approved! Excellent work @pieer. 👍

PR #1238, opened 2022-04-27
Closes #1236 + updates vue-router

### LINKED ISSUE #1236: "Error: Redirected when going from "/?modal=SignupModal" to "/dashboard" via a navigation guard."
### Problem

My PR #1224 both fixed an introduced an issue related to logging in.

Now after signing up there's an error in the console:

```
Error: Redirected when going from "/?modal=SignupModal" to "/dashboard" via a navigation guard.
```

This is the offending code:

```js
      sbp('okTurtles.events/on', LOGIN, () => {
        if (this.$store.getters.ourUsername) {
          router.currentRoute.path === '/' && router.push({ path: '/dashboard' }).catch(console.error)
        }
      })
```

If that's simply removed, however, and we're logged in and in a group, and load the `/app` page, then we aren't redirected to `/dashboard`

### Solution

Find and fix this error, undoing what was done in main.js #1224 without re-breaking the `/app` page behavior.

taoeffect:
Turns out closing #1192 was *a lot* trickier than expected.

+ Fixes `Error: Redirected when going from "/?modal=SignupModal" to "/dashboard" via a navigation guard.` (Closes #1236)
+ Fixes a few vue console errors
+ Updates vue-router to 3.5.3

### DIFF
--- a/frontend/controller/router.js
+++ b/frontend/controller/router.js
@@ -8,9 +8,7 @@ import store from '@model/state.js'
 import Home from '@pages/Home.vue'
 import Join from '@pages/Join.vue'
 import L from '@view-utils/translations.js'
-import { logExceptNavigationDuplicated } from '@view-utils/misc.js'
 import { lazyPage } from '@utils/lazyLoadedView.js'
-import { VUE_LOADED } from '@utils/events.js'
 
 /*
  * Lazy load all the pages that are not necessary at initial loading of the app.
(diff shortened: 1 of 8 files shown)


frontend/views/containers/access/SignupModal.vue
```diff
@@ -28,10 +28,7 @@ export default ({
   methods: {
     submit () {
       if (this.$route.query.next) {
-        // TODO: get rid of this timeout and fix/update tests accordingly
-        setTimeout(() => {
-          this.$router.push({ path: this.$route.query.next })
-        }, 1000)
+        this.$router.push({ path: this.$route.query.next }).catch(logExceptNavigationDuplicated)
```
snowteamer, line 31: I'm not a fan of `logExceptNavigationDuplicated` becoming commonplace. Although it's arguably convenient, I'd rather have an if-statement upfront to prevent the duplicate navigation.

frontend/controller/router.js
```diff
@@ -70,113 +68,133 @@ function createEnterGuards (...guards) {
...
+    : []),
+  {
+    path: '*',
+    redirect: '/'
+  }
+]
 const router: any = new Router({
   mode: 'history',
   base: '/app',
+  routes,
```
snowteamer: Why was `routes` extracted?

frontend/controller/router.js
```diff
@@ -70,113 +68,133 @@ function createEnterGuards (...guards) {
...
+sbp('okTurtles.events/once', VUE_LOADED, function () {
+  const route = routes.find(r => r.path === startRoute.path)
+  // $FlowFixMe
+  const next = route?.beforeEnter?.(startRoute, null, (next) => next)
+  // console.warn(`started at ${startRoute.path} and got next of ${JSON.stringify(next)}`)
+  if (next) {
+    router.push(next).catch(logExceptNavigationDuplicated)
+  }
+})
+
```
snowteamer: This seems a bit hacky and I have some trouble picturing the control flow here.

snowteamer: In fact I just don't understand why all this was necessary. :sweat_smile:

frontend/controller/router.js
taoeffect: This should be answered a few lines down bellow...

frontend/controller/router.js
taoeffect: It is a bit hacky.

It's necessary because of what's mentioned in the comment. In some cases, some pages will not redirect properly because `store.state` isn't initialized but the guards get triggered early on anyway

frontend/views/containers/access/SignupModal.vue
taoeffect: An `if` statement would turn what should be a 1-liner into 3 lines :-\

frontend/controller/router.js
taoeffect: To elaborate a bit more, `store.state.currentGroupId` will be `null` because the contracts haven't yet finished being loaded by `login`. We delay `login` because it performs a contract sync which would result in a blank white page if that took too long.

taoeffect: @snowteamer Actually, I've been thinking about this, and I think there is a cleaner way to implement it, so thanks for pointing this out.

I will work on separating out the contract sync part of the login from the rest of the login action, and that should fix it.

snowteamer: @greg Thank you for the additional details.
> I will work on separating out the contract sync part of the login from the rest of the login action, and that should fix it.

Sounds promising, I hope this works :+1:

snowteamer (dismissed):
Approved

frontend/controller/router.js
taoeffect: Seems to work. 🙂

PR #1256, opened 2022-05-01
Use experimental Fetch API instead of node-fetch and form-data

snowteamer:


taoeffect (approved):
Nice, great PR @snowteamer, thank you for helping us reduce our dependencies!

Issue #1227, opened 2022-04-24
"Duplicate keys detected:" in ChatMain.vue


taoeffect:
### Problem

I'm not sure if this happens in `master`, but it definitely happens in many of the Cypress tests in my PR (#1224).

An error will flash in the console during the tests. It won't prevent the tests from passing, but it clearly indicates something is wrong:

<img width="971" alt="Screen Shot 2022-04-24 at 3 53 09 PM" src="https://user-images.githubusercontent.com/138706/165000158-12d0a30b-03fa-4571-b24f-b6312a7c64bb.png">

### Solution

Based on the error message, something is affected by `setMessageEventListener`, possibly a `v-for` loop. I am guessing it might be `:key='message.id'`.

Silver-IT:
I can see that it happens only with options `{ BypassUI: true }`.

PR #1258, opened 2022-05-02
Fix  #1219: Emoticon not showing up after being sent

### LINKED ISSUE #1219: Emoticon not showing up after being sent
### Problem
In the file frontend/views/containers/chatroom/MessageReactions.vue, we can find the codes to render emoticons from the emoticonsList.

This emoticonsList is a prop which is sent from the frontend/views/containers/chatroom/ChatMain.vue thru the MessageBase.vue component.

The problem is that emoticonsList is updated every time user send/remove emoticons but UI doesn't keep updated.
Emoticon Updates

They are all MESSAGE_ACTION_TYPES.EDIT_MESSAGE event. And it handles well. But UI doesn't update...
image

### Solution

Find why the UI is not being updated and force to refresh

pieer:


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -28,7 +28,7 @@
 
         component(
           :is='messageType(message)'
-          :key='message.id'
+          :key='message.id + Math.random()'
```
taoeffect, line 31: This change is related to the "duplicate keys" bug of #1227, but a more correct fix should be used. /cc @Silver-IT

Silver-IT: Make sense.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -275,7 +275,7 @@ export default ({
         if (alreadyAdded >= 0) {
           emoticons[emoticon].splice(alreadyAdded, 1)
           if (emoticons[emoticon].length === 0) {
-            delete this.messages[emoticon]
+            delete this.messages[index].emoticons[emoticon]
```
Silver-IT, line 278: @pieer, I think you need to make a new branch from `chat-contract-messaging` branch and make pull request to `chat-contract-messaging` branch. Actually, making reactions is not implemented in the `main` branch.
I think this is my mistake that I didn't tell you before. You couldn't know.

Silver-IT (commented):
Nice work, @pieer. I think you worked in the wrong place.
You need to create a new branch from `chat-contract-messaging` brach and also make a PR to `chat-contract-messaging` branch. Making reactions are not implemented in `main` branch and it is implemented in `chat-contract-messaging` branch and the issue was related to the `chat-contract-messaging` branch. I was offline when you start working on this. I should tell you before. That's my mistake. Sorry for that.

pieer:
Better fix implemented chat-contract-messaging branch

PR #1259, opened 2022-05-02
Avoid linebreaks before punctuations in i18n text + quickfix for missing controller/backend.js import 

### LINKED ISSUE #1246: Make page refresh on new version more reliable
### Problem

It's very odd, the page refresh was working reliably but now it's not...

Screen Shot 2022-04-28 at 2 17 50 PM

What's weird is that I saw the lost server connection banner disappear. Yet the page didn't refresh, I'm guessing because for whatever reason the client didn't get the new version.

### Solution

Figure out what's going on?

Maybe upon reconnection (since at least that part is working) the client can request the version be re-sent?

### LINKED ISSUE #1253: French: add non-breaking space (&nbsp;) to copy with colons
### Problem

Some of the strings in the project have colons in them, and in French apparently you put a space before the colon, so that instead of `"Amount: {foo}"` it's `"Amount : {foo}"`.

This however could cause the colon to appear in some places on a new line when it shouldn't.

### Solution

Update the `french.strings` file to use `&nbsp;` before the colon. This should work for places that use `i18n` to render the string, since that uses `v-html`. However, in places where the `L` string is used instead of `i18n`, it might render the `&nbsp;` if `v-html` isn't used. In those places you'll need to either manually use `v-safe-html` or switch to using `i18n`.

snowteamer:
Closes #1246
Closes #1253

### Summary of changes:
- No more inopportune linebreaks before punctuations in languages that use blank spaces before certain punctuations (e.g. French in `fr-FR` and `fr-BE` locales).  This was done by replacing those blank spaces with `&nbsp;` at runtime.
- A missing `./controller/backend.js` import has been restored, resulting in the following quickfixes:
  - Fix broken translation table loading when the browser is not using the default English locale.
  - Fix broken page reloading when a new version is deployed.
- A bunch of French strings have been updated.


frontend/views/utils/translations.js
```diff
@@ -126,6 +126,8 @@ export default function L (
   args: Array<*> | Object | void
 ): string {
   return template(currentTranslationTable[key] || key, args)
+    // Avoid inopportune linebreaks before certain punctuations.
+    .replace(/\s(?=[;:?!])/g, '&nbsp;')
```
taoeffect, line 130: This is rather brilliant 👍

frontend/main.js
```diff
@@ -9,6 +9,7 @@ import { GIMessage } from '~/shared/domains/chelonia/chelonia.js'
 import { CONTRACT_IS_SYNCING } from '~/shared/domains/chelonia/events.js'
 import './controller/namespace.js'
 import './controller/actions/index.js'
+import './controller/backend.js'
```
taoeffect, line 12: Kinda shocked this got removed somehow and nobody noticed until now

taoeffect (approved):
Brilliant PR @snowteamer, thank you! Approved! 🥇

frontend/views/utils/translations.js
snowteamer: Thank you :slightly_smiling_face:

frontend/main.js
snowteamer: Yea, me too

PR #1262, opened 2022-05-03
Make sure the pubsub client tries to reconnect on any error

### LINKED ISSUE #1246: Make page refresh on new version more reliable

snowteamer:
Closes #1246

### DIFF
--- a/shared/pubsub.js
+++ b/shared/pubsub.js
@@ -203,6 +203,19 @@ const defaultClientEventHandlers = {
     client.socket = null
     client.clearAllTimers()
 
+    // This has been commented out to make the client always try to reconnect.
+    // See https://github.com/okTurtles/group-income/issues/1246
+    /*
+    // See "Status Codes" https://tools.ietf.org/html/rfc6455#section-7.4
+    switch (event.code) {
+      // TODO: verify that this list of codes is correct.
+      case 1000: case 1002: case 1003: case 1007: case 1008: {
+        client.shouldReconnect = false
+        break
+      }
+      default: break
+    }
+    */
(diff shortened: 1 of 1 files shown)


shared/pubsub.js
```diff
@@ -203,15 +203,6 @@ const defaultClientEventHandlers = {
...
 
-    // See "Status Codes" https://tools.ietf.org/html/rfc6455#section-7.4
-    switch (event.code) {
-      // TODO: verify that this list of codes is correct.
-      case 1000: case 1002: case 1003: case 1007: case 1008: {
-        client.shouldReconnect = false
-        break
-      }
-      default: break
-    }
```
taoeffect: Instead of outright deleting this code, as the RFC comment is useful, could you just comment it out and add another comment with a link to https://github.com/okTurtles/group-income/issues/1246 ?

taoeffect (changes requested):
Thanks @snowteamer! Let's hope this works! 🤞 

Very minor change requested.

shared/pubsub.js
snowteamer: Fair point. I'm going to think a bit about a good comment.

taoeffect (approved):
Am back from a vacation - sorry for taking a while to approve this! Thanks @snowteamer!

PR #1264, opened 2022-05-04
Messaging functions in group chat page

### LINKED ISSUE #1227: "Duplicate keys detected:" in ChatMain.vue

Silver-IT:
### Summary of changes

1. Edit / Delete messages
2. Make reactions ( Add / Remove emoticons )
3. Some UI updates
4. Closes #1227 

### TODO
1. `(edited)` label for the edited text ( in the next Pull Request #1274 )


Silver-IT (commented):
TODO: fix error regarding emoticons ( Not well reacted between UI and variable in ChatMain.vue )

frontend/model/contracts/chatroom.js
```diff
@@ -124,17 +124,24 @@ function createNotificationData (
...
-    if (state.messages[i].id === hash) {
-      sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, {
-        type,
-        data: { message: state.messages[i] }
-      })
-      break
+  let data = {}
+  if (type === MESSAGE_ACTION_TYPES.ADD_MESSAGE || type === MESSAGE_ACTION_TYPES.EDIT_MESSAGE) {
+    const messagesCount = state?.messages.length || 0
+    for (let i = messagesCount - 1; i >= 0; i--) {
```
snowteamer, line 149: What is the reason to use a reverse loop rather than a regular loop of `.find()` call?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -255,6 +258,39 @@ export default ({
       this.ephemeral.replyingMessage = message.text
       this.ephemeral.replyingTo = this.who(message)
     },
+    editMessage (message, newMessage) {
+      console.log('updating...')
```
snowteamer: Are these log statements still necessary?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -255,6 +258,39 @@ export default ({
...
+        }
+      })
+    },
+    sendMessage (index) {
+      this.ephemeral.replyingMessage = null
+      this.ephemeral.replyingTo = null
+      setTimeout(() => {
+        console.log('TODO $store send message')
+        // this.$set(this.ephemeral.pendingMessages[index], 'hasFailed', true)
+      }, 2000)
```
snowteamer: This doesn't seem to do anything

test/cypress/integration/group-chat-message.spec.js
```diff
@@ -0,0 +1,210 @@
...
+      cy.get('.c-message').eq(0).find('.c-text').should('contain', `Joined ${CHATROOM_GENERAL_NAME}`)
+
+      cy.get('.c-message').eq(1).find('.c-who > span:first-child').should('contain', user2)
+      cy.get('.c-message').eq(1).find('.c-text').should('contain', `Joined ${CHATROOM_GENERAL_NAME}`)
+
+      cy.get('.c-message').eq(2).find('.c-who > span:first-child').should('contain', user2)
+      cy.get('.c-message').eq(2).find('.c-text').should('contain', 'Can we have a meeting this evening?')
+
+      cy.get('.c-message').eq(3).find('.c-who > span:first-child').should('contain', user1)
+      cy.get('.c-message').eq(3).find('.c-text').should('contain', `Hi ${user2}. I am fine thanks.`)
```
snowteamer: Couldn't we reduce code duplication and line lengths by defining some other helper functions, for example `checkSender()`, `checkMessage()`, or something similar?

snowteamer (dismissed):
Great work once again!

I don't have much feedback to give, although I would appreciate:
- some loop optimizations (if possible and not too difficult, otherwise nevermind)
- a few more helper functions in Cypress tests, to make them nicer

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -104,8 +107,17 @@ export default ({
...
     isCurrentUser: Boolean
   },
+  computed: {
+    isEditable (): Boolean {
+      if (this.type === MESSAGE_TYPES.TEXT) {
+        return true
+      }
+      return false
+    }
+  },
```
pieer, line 117: Maybe we could simplify with online
return this.type === MESSAGE_TYPES.TEXT

frontend/model/contracts/chatroom.js
Silver-IT: As you know, the purpose is to find a `message` having same hash as it's `id`.
I think if we do reverse loop, we can save calculation and time to get the right message.
That's because in most cases, we will update message which we sent just before.
We don't update message which I sent long time ago.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: No. I will remove it. Nice catch.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: This `sendMessage` function is not necessary at all. I will remove it.

test/cypress/integration/group-chat-message.spec.js
Silver-IT: Correct. I will update.

test/cypress/integration/group-chat-message.spec.js
```diff
@@ -164,19 +171,12 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
...
-      cy.get('.c-message').eq(2).find('.c-text').should('contain', 'Can we have a meeting this evening?')
-
-      cy.get('.c-message').eq(3).find('.c-who > span:first-child').should('contain', user1)
-      cy.get('.c-message').eq(3).find('.c-text').should('contain', `Hi ${user2}. I am fine thanks.`)
     })
+
+    checkMessageBySender(0, user1, `Joined ${CHATROOM_GENERAL_NAME}`)
+    checkMessageBySender(1, user2, `Joined ${CHATROOM_GENERAL_NAME}`)
+    checkMessageBySender(2, user2, 'Can we have a meeting this evening?')
+    checkMessageBySender(3, user1, `Hi ${user2}. I am fine thanks.`)
```
snowteamer, line 179: Now this looks great!

snowteamer (dismissed):

frontend/model/contracts/chatroom.js
snowteamer: Fair point. I've also been thinking that keeping track of the `id` <-> `message` mappings could sometimes allow to do plain lookups such as `messagesById[hash]` instead of loops, but the mappings object would have to be updated every time a message is added or removed, so I'm not sure.

Silver-IT: In `chat-contract-optimize` branch, I am going to remove `state.messages` to save in the vuex store.
So we can save disk and performance too.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -150,8 +153,20 @@ export default ({
...
+      return objElement.id === 'mobileSendButton'
+    },
+    textAreaBlur (event) {
+      if (this.ephemeral.isPhone) {
+        if (this.checkIfMobileSendButton(event.relatedTarget)) {
+          this.sendMessage()
+        } else {
+          this.ephemeral.showButtons = true
+        }
+      }
```
pieer, line 165: This 10 lines could be simplified in 2 lines maybe:
if (!this.ephemeral.isPhone) return;
event?.relatedTarget?.id === 'mobileSendButton' ? this.sendMessage() : this.ephemeral.showButtons = true

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -199,7 +214,7 @@ export default ({
       this.$emit('stop-replying')
     },
     sendMessage () {
-      console.log('send')
+      console.log('send', this.$refs.textarea.value)
```
pieer: Remove console log now?

Silver-IT: Yeap.

frontend/views/containers/chatroom/MessageActions.vue
Silver-IT: Nice catch. Thanks

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: Yeap. That's right.

snowteamer (dismissed):

frontend/model/contracts/chatroom.js
```diff
@@ -129,17 +129,24 @@ function createNotificationData (
...
+      if (state?.messages[i].id === hash) {
+        data = { message: { ...state?.messages[i] } }
+        break
+      }
     }
+  } else if (type === MESSAGE_ACTION_TYPES.DELETE_MESSAGE) {
+    data = { id: hash }
```
taoeffect, line 156: This is confusing me, it suggests that there are two different shapes to the `data` object.

Sometimes `data` contains the `id` directly inside of it, as above: `{ id: hash }`

Other times the `id` is stored inside another object, as on line 142: `{ message: { id, ... } }`

Won't think cause confusion and potential bugs in the rest of the code? Should perhaps this `data` be made more consistent, so that `id` is in the same place in both cases?

frontend/model/contracts/chatroom.js
```diff
@@ -129,17 +129,24 @@ function createNotificationData (
...
 }): void {
-  for (let i = state.messages.length - 1; i >= 0; i--) {
-    if (state.messages[i].id === hash) {
-      sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, {
-        type,
-        data: { message: state.messages[i] }
-      })
-      break
+  let data = {}
+  if (type === MESSAGE_ACTION_TYPES.ADD_MESSAGE || type === MESSAGE_ACTION_TYPES.EDIT_MESSAGE) {
```
taoeffect, line 147: If `state` is `null` or `undefined`, then this section of the `if` statement won't do anything.

So instead of having constant checks for `state?.` below, couldn't you add `state` to the condition here? E.g. something like:

```js
  if (state && (type === MESSAGE_ACTION_TYPES.ADD_MESSAGE || type === MESSAGE_ACTION_TYPES.EDIT_MESSAGE)) {
```

BUT - an even better question - why are you making `state` optional here? Isn't it passed in everywhere this function is called? If so, then you can update the function parameters to indicate it mandatory, and remove the `state?` stuff below and just make it `state.messages`

frontend/model/contracts/chatroom.js
```diff
@@ -327,36 +334,94 @@ sbp('chelonia/defineContract', {
...
+        objectOf({
+          id: string,
+          text: string
+        })(data)
+        // TODO: need to check if the meta.username === message.from
+      },
       process ({ data, meta }, { state }) {
-
+        for (let i = state.messages.length - 1; i >= 0; i--) {
+          if (state.messages[i].id === data.id) {
```
taoeffect: I agree with @snowteamer  - this pattern is showing up in your code in several places, so let's DRY this.

I also agree with you @Silver-IT, that it's better to start the search from the end instead of the beginning, and so we should implement our own version of `.find`, e.g. something like:

```js
function findMessageIdx(id, messages) {
  for (let i = messages.length - 1; i >= 0; i--) {
    if (messages[i].id === id) {
      return i
    }
  }
  return -1
}
```

Then you can remove the loops you have everywhere else.

frontend/model/contracts/chatroom.js
Silver-IT: Got it. I just wanted to send only necessary data especially only `id`, however I needed to keep consistency of the data type. I agree with it.

But can we skip this in this PR? Because we are not sending those data anymore.
You can see what I am saying if you review PR #1274. 

> sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, { hash })

I send only `hash` and no other data in the next  PR.
But anyway, I understand your point and will keep it in my mind.

taoeffect: > But can we skip this in this PR? Because we are not sending those data anymore.

OK, sure 👍

frontend/model/contracts/chatroom.js
Silver-IT: Correct. Just make `state` parameter mandatory and remove `?`. Nice catch. Thanks

frontend/model/contracts/chatroom.js
Silver-IT: Nice catch. 10Q

frontend/model/contracts/chatroom.js
```diff
@@ -327,36 +343,88 @@ sbp('chelonia/defineContract', {
...
+              emoticons[emoticon].push(meta.username)
+            }
+          } else {
+            emoticons[emoticon] = [meta.username]
+          }
+          if (emoticons) {
+            Vue.set(state.messages[msgIndex], 'emoticons', emoticons)
+          } else {
+            Vue.delete(state.messages[msgIndex], 'emoticons')
+          }
```
taoeffect: This code could be simplified & shortened by leaving an empty list when there are no more emoticons left. I don't think it needs the code that completely deletes the entire emoticons list.

strings/french.json
```diff
@@ -420,8 +423,8 @@
   "*This value was automatically adjusted because your group is too small for the disagreement number.": "*Cette valeur a été ajustée automatiquement car la taille de votre groupe est trop petite pour le seuil de désaccord.",
   "Active": "Actif",
   "Change to this system": "Choisir ce système",
-  "Maximum number of \"no\" votes:": "Nombre maximal de votes négatifs :",
-  "Percentage of members that need to agree:": "Pourcentage de votes positifs requis :",
+  "Maximum number of \"no\" votes:": "Nombre maximal de votes négatifs:",
+  "Percentage of members that need to agree:": "Pourcentage de votes positifs requis:",
```
taoeffect: I think we decided that the spaces should exist for French, right @pieer @snowteamer ?

frontend/model/contracts/chatroom.js
Silver-IT: Of course, if there are `emoticons` empty list, there won't be any thing wrong.
All things will work well and nothing breaks. But no need to save redundant thing, don't we?
I think the problem is to shorten the code lines instead of leaving empty array.

strings/french.json
Silver-IT: I forgot to run `strings.linux`. After I run it and commit again, they are all solved.
Nice catch, @taoeffect.

frontend/model/contracts/chatroom.js
taoeffect: Well, shorter code is easier to read, understand, and has fewer bugs...

What is the harm with leaving an empty array? These messages will get cleared out of memory eventually anyway.

taoeffect (changes requested):
Looking really good @Silver-IT! I think this is almost ready to merge.

Something that's missing is the little `(edited)` grey text that should appear on messages that have been edited:

Screen Shot 2022-05-15 at 6 19 32 PM

Also, I do still think [this emoticon code](https://github.com/okTurtles/group-income/pull/1264#pullrequestreview-973312189) could be improved by leaving the empty list instead of deleting it. It would improve readability and reduce the likelihood of bugs, with little to no downside.

frontend/model/contracts/chatroom.js
Silver-IT: > These messages will get cleared out of memory eventually anyway.

This makes sense. I will remove the redundant codes then.

taoeffect (approved):
Great work @Silver-IT! Approved!! 😄 🎉

Issue #1255, opened 2022-04-30
Click link to copy doesn't work in Safari on Desktop


taoeffect:
### Problem

A recent PR messed up the behavior of Safari on Desktop when clicking to copy a link.

It should should display a tooltip saying the link was copied but instead a share menu appears:

<img width="683" alt="Screen Shot 2022-04-30 at 1 23 18 PM" src="https://user-images.githubusercontent.com/138706/166121464-c6b6d27d-6636-4a28-93e5-a90064e1f25d.png">

### Solution

Fix, while not breaking Safari on Mobile, which should bring up the share menu.

/cc @abdulazizali77

abdulazizali77:
@taoeffect this probably means that Safari has the Web Share API defined https://caniuse.com/?search=web%20share%20api
what should be investigated is if other browsers have 'Copy to clipboard' when the share menu comes up, i assume in terms of behaviour 'Share this link' should be inclusive of all methods including 'Copy to clipboard' and mail etc

taoeffect:
Yeah that's a fair point. The share menu on iPhone does have a "copy" button available, but it's not available on most desktop browsers afaik. Should come up with some kind of consistent rule for this.

abdulazizali77:
what does 'More...' in Safari show?

abdulazizali77:
apparently seems like an issue on Safari, though i am unsure what the version is
https://discussions.apple.com/thread/252923714

abdulazizali77:
![image](https://user-images.githubusercontent.com/23547158/166127764-c4cce26f-2044-40f8-95f0-a9ec9c477c7d.png)
Chrome

![image](https://user-images.githubusercontent.com/23547158/166127792-4f0dd420-dea4-4915-ab65-098d3eca48ea.png)
Edge

Consistent with the caniuse table, Firefox 80.0.1 doesnt have `navigator.share` hence copies to clipboard 

I guess one kludge would be to check if its Safari desktop and explicitly copy to clipboard. doesnt seem to be any immediately elegant/extensible/modular way to do that that doesnt involve a good degree of architecture.

taoeffect:
Oddly enough it brings up the System Preferences, which actually *does* have a "Copy link" thing in it, that I had to move up to the top:

<img width="780" alt="Screen Shot 2022-04-30 at 6 12 02 PM" src="https://user-images.githubusercontent.com/138706/166127952-f6a34309-6dd3-4701-bddd-d0fceb163345.png">

[...]

PR #1270, opened 2022-05-09
Fix clipboard copy Mac Safari corner case (#1255)

abdulazizali77:
Defined function just for the corner case `isMacSafari` in `LinkToCopy.vue` itself using `navigator.userAgent`. 
In terms of UX might be a bit weird that the copy notification shows at the same time that the share context menu pops up

- Tested OSX Catalina 10.15.6
(`navigator.share` defined) Safari/605.1.15 (14.0); Edge/101.0.1210.39
(`navigator.share` undefined) Chrome/90.0.4430; Firefox/99.01;

- Tested Windows 10
(`navigator.share` defined) Chrome/101.0.4951.54; Edge/101.0.1210.39
(`navigator.share` undefined) Firefox/80.0;

_Issues_
- Defined function specifically in `LinkToCopy.vue`. Probably best to have a central useragent util somewhere and relocate that there
- No actual Mac Safari version was tested and will return true for ALL versions. We could actually test for version ranges with `/Macintosh.*Safari\/(.*)/` 
- Probably the regex for machine and application should be separate 
- Firefox doesnt seem to tack on `Safari` on to the useragent hence was not tested

### DIFF
--- a/frontend/views/components/LinkToCopy.vue
+++ b/frontend/views/components/LinkToCopy.vue
@@ -47,13 +47,24 @@ export default ({
     }
   },
   methods: {
+    // FIXME: probably needs to be defined centrally somewhere else
+    isMacSafari () {
+      // NB: Chrome, Edge, Brave on Macintosh tacks on the 'Safari/537.36' string in its userAgent
+      // so we have to ensure that were not mistaking those browsers instead when matching for Safari on Macintosh
+      const userAgent = navigator.userAgent
+      const safari = /Macintosh.*Safari/
+      const chrome = /Macintosh.*Chrome/
+      const edge = /Macintosh.*Edg/
+      return userAgent.match(safari) &&
+              !(userAgent.match(chrome) || userAgent.match(edge))
+    },
     copyToClipboard () {
       // if the user is using the device that supports web share API, use it and then skip the other logics below.
-      if (navigator.share) {
+      if (navigator.share && !this.isMacSafari()) {
         navigator.share({
           title: this.L('Your invite'),
           url: this.link
-        })
+        }).catch((error) => console.error('navigator.share failed with:', error))
         return
       }
 


frontend/views/components/LinkToCopy.vue
```diff
@@ -47,14 +47,25 @@ export default ({
     }
   },
   methods: {
+    // FIXME: probably needs to be defined centrally somewhere else
+    isMacSafari (userAgent) {
+      const safari = /Macintosh.*Safari/
+      const chrome = /Macintosh.*Chrome/
+      const edge = /Macintosh.*Edg/
+      return userAgent.match(safari) &&
+              !(userAgent.match(chrome) || userAgent.match(edge))
```
taoeffect, line 59: A few questions here:

I am confused by this condition... why is it written this way instead of simplified to a single regex match?

Is it possible that `/Macintosh.*Safari/` and `/Macintosh.*Chrome/` can match at the same time? What are the actual user agent strings we're dealing with?

frontend/views/components/LinkToCopy.vue
```diff
@@ -47,14 +47,25 @@ export default ({
     }
   },
   methods: {
+    // FIXME: probably needs to be defined centrally somewhere else
+    isMacSafari (userAgent) {
```
taoeffect: Does it make sense to pass in the user agent as a parameter? Shouldn't it just directly check against `navigator.userAgent` in here?

frontend/views/components/LinkToCopy.vue
```diff
@@ -47,14 +47,25 @@ export default ({
...
+    },
     copyToClipboard () {
       // if the user is using the device that supports web share API, use it and then skip the other logics below.
       if (navigator.share) {
         navigator.share({
           title: this.L('Your invite'),
           url: this.link
-        })
-        return
+        }).catch((error) => console.log(error))
```
taoeffect: Errors should be logged with `console.error` and should always indicate which part of the code caused them, so something like this would be better:

```js
        }).catch((error) => console.error('navigator.share failed with:', error))
```

frontend/views/components/LinkToCopy.vue
```diff
@@ -47,14 +47,25 @@ export default ({
...
+        // we return immediately for any case other than mac safari
+        if (!this.isMacSafari(navigator.userAgent)) {
+          return
+        }
```
taoeffect: Wouldn't it make more sense to put this into the condition above, like so?

```js
if (navigator.share && !this.isMacSafari()) {
```

taoeffect (changes requested):
Thanks so much @abdulazizali77! Sorry for taking a while to review this, you caught me on my vacation but I'm back now! Review posted. 🙂

frontend/views/components/LinkToCopy.vue
abdulazizali77: I tested these browsers on my mac

Chrome
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36"

Edge
'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.4951.54 Safari/537.36 Edg/101.0.1210.39

Brave
'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.4951.54 Safari/537.36'

Firefox (no issue)
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:99.0) Gecko/20100101 Firefox/99.0"
​
Safari
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0 Safari/605.1.15"

Interesting article related to it 
https://blogs.windows.com/msedgedev/2015/06/17/building-a-more-interoperable-web-with-microsoft-edge/
Apparently using detection with userAgent as you are aware is rife with problems not the least due to spoofing

frontend/views/components/LinkToCopy.vue
abdulazizali77: mainly due to the compound conditions above, having a function at the very least allows us to relocate it later, i definitely dont think embedding it in the vuejs is the best in terms of usability at least

taoeffect: Sorry, I'm not following, what compound conditions above?

frontend/views/components/LinkToCopy.vue
taoeffect: Ah, I see. OK, yeah that's kinda crazy. Nice job 👍

taoeffect: Can you write a comment with a few sentences explaining that so that it doesn't get forgotten?

frontend/views/components/LinkToCopy.vue
abdulazizali77: i was under the impression that for Mac Safari (which we assume would have `navigator.share` defined), we would still want to display the share context menu AND copy the link to the clipboard (only because the copy to clipboard is missing in the share context menu)
which would sort of equal to other browsers ie Edge and Chrome (Windows) which have `navigator.share` defined AND has the option to copy the link to the clipboard, amirite? :)

but admittedly copying the url to clipboard might not be the best UX experience ("were doing something that the user didnt choose to do")

taoeffect: So, the issue was to disable the share the share context menu on macOS Safari entirely, right? Since it doesn't really work how we want it to.

We want it to behave exactly like Firefox on macOS behaves - simply displaying the "copied to clipboard" tooltip

frontend/views/components/LinkToCopy.vue
abdulazizali77: good point, changed

frontend/views/components/LinkToCopy.vue
abdulazizali77: safari == true &&  (edge || chrome) == false

frontend/views/components/LinkToCopy.vue
abdulazizali77: ok

frontend/views/components/LinkToCopy.vue
abdulazizali77: ok

frontend/views/components/LinkToCopy.vue
taoeffect: Right, but that doesn't mean we need to take a `userAgent` parameter here right? We can directly access `navigator.userAgent` in this function instead, for example, simply writing:

```js
const userAgent = navigator.userAgent
```

abdulazizali77: youre right, i think initially i wanted to have a more general two parameter function not an `isSpecificA` `isSpecificB` which is why i left the `userAgent` param there

abdulazizali77: changed

frontend/views/components/LinkToCopy.vue
abdulazizali77: changed

taoeffect (approved):
Great job @abdulazizali77, thank you!

PR #1274, opened 2022-05-15
Optimize memory usage for chatroom contract

Silver-IT:
### Summary of Changes

- `eventsBefore` backend API endpoint that respond `howMany` number of events before a specific event whose hash is `before`
- Removed messages from vuex storage for chatroom contract
- Set 40 as the default number of events to display in a page
- Implement `Infinite Scroll` and retrieve events from the backend and display messages when user scrolls up and tries to see more
- `(edited)` label for the edited text


taoeffect:
Note: this PR is pending review & merge of #1264.

taoeffect:
I'm guessing my change requests for removing the `emoticons` deletion code in the previous PR result in these `+` emoticons appearing when you add a reaction to a message and remove it. Probably the UI needs to hide them if `.length === 0`

Screen Shot 2022-05-15 at 8 42 39 PM

Silver-IT:
> I'm guessing my change requests for removing the `emoticons` deletion code in the previous PR result in these `+` emoticons appearing when you add a reaction to a message and remove it. Probably the UI needs to hide them if `.length === 0`
> 
> Screen Shot 2022-05-15 at 8 42 39 PM

Interesting. If we need to check all the emoticons if it has no users, then we should remove emoticons without any users?

> emoticons = { emoticon1: [...users], emoticon2: [...users] }

I think I need to revert that change to remove empty emoticons.

taoeffect:
> I think I need to revert that change to remove empty emoticons.

OK, maybe it does make more sense to do it that way then 👍

backend/database.js
```diff
@@ -51,6 +51,52 @@ export default (sbp('sbp/selectors/register', {
...
+    if (!currentHEAD) {
+      throw Boom.notFound(`contractID ${contractID} doesn't exist!`)
+    }
+
+    hash = hash || currentHEAD
+    let prefix = '['
+    if (hash === currentHEAD) {
+      const entry = await sbp('chelonia/db/getEntry', currentHEAD)
+      prefix = `["${strToB64(entry.serialize())}",`
+      numberOfActions--
```
snowteamer: What if `numberOfActions` hasn't been specified? Is it required whereas `hash` is not (as per the ` hash = hash || currentHEAD` line)?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -93,6 +112,7 @@ export default ({
     Avatar,
     Emoticons,
     Loading,
+    InfiniteLoading,
```
snowteamer, line 115: Why not add it before `Loading` so as to preserve the natural ordering?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -118,8 +138,11 @@ export default ({
       config: {
         isPhone: null
       },
+      latestEvents: [],
       messages: [],
       ephemeral: {
+        refreshMessages: true,
+        infiniteLoading: null,
```
snowteamer, line 145: Same point here about ordering. Agreed sometimes it's not practical to maintain, but in this case (and in many similar ones) I don't see why not keep it.

backend/database.js
```diff
@@ -51,6 +51,52 @@ export default (sbp('sbp/selectors/register', {
...
+          } else if (currentHEAD === hash) {
+            metBefore = true
+          }
+
+          if (!numberOfActions || currentHEAD === contractID) {
+            this.push(']')
+            this.push(null)
+          } else {
+            currentHEAD = entry.message().previousHEAD
+            this.push('')
```
snowteamer: Did you mean `this.push(prefix)` (or `this.push(',')`)?

backend/routes.js
```diff
@@ -66,6 +66,18 @@ route.GET('/events/{contractID}/{since}', {}, async function (request, h) {
   }
 })
 
+route.GET('/eventsBefore/{contractID}', {}, async function (request, h) {
+  try {
+    const { contractID } = request.params
+    const { before, howMany } = request.query
```
snowteamer: Shouldn't we check whether these query parameters are actually present and valid, and issue a `BadRequest` error response otherwise?
I'd also consider renaming `numberOfActions` in `backend/db/streamEntriesBefore'` to `howMany` for consistency, if they mean the same thing. Same point for `hash` and `before`.

frontend/model/contracts/chatroom.js
```diff
@@ -40,10 +40,8 @@ export const messageType: any = objectMaybeOf({
     type: unionOf(...Object.values(MESSAGE_NOTIFICATIONS).map(v => literalOf(v))),
     params: mapOf(string, string) // { username }
   }),
-  replyingMessage: objectOf({
+  replying: objectOf({
```
snowteamer: Although this looks nicer, I still prefer `replyingMessage` as it's more obvious that it's a message.

frontend/model/contracts/chatroom.js
```diff
@@ -352,38 +353,37 @@ sbp('chelonia/defineContract', {
...
+          return
+        }
         const msgIndex = findMessageIdx(data.id, state.messages)
         if (msgIndex >= 0) {
           state.messages[msgIndex].text = data.text
           state.messages[msgIndex].updatedDate = meta.createdDate
+          if (state.simulation && state.messages[msgIndex].pending) {
+            delete state.messages[msgIndex].pending
+          }
         }
```
snowteamer, line 368: It seems that only the found message itself is used in this block, not its index. So would it better to just search for the message?

frontend/model/contracts/chatroom.js
```diff
@@ -392,14 +392,23 @@ sbp('chelonia/defineContract', {
         emoticon: string
       }),
       process ({ data, meta, contractID }, { state }) {
+        if (!state.simulation) {
+          return
+        }
         const { id, emoticon } = data
         const msgIndex = findMessageIdx(id, state.messages)
```
snowteamer, line 401: Again only the found message appears to be necesary in the code below

pieer (dismissed):
Excellent work!
No changes from what I can see

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,43 @@ export default ({
...
+    async getLatestEvents (refresh = false) {
+      const eventsPerPage = this.chatRoomSettings?.actionsPerPage || CHATROOM_ACTIONS_PER_PAGE
+      const beforeActionHash = refresh || !this.latestEvents.length
+        ? ''
+        : GIMessage.deserialize(this.latestEvents[0]).hash()
+
+      const newEvents = await sbp('chelonia/contractEventsBefore', this.currentChatRoomId, beforeActionHash, eventsPerPage)
+
+      if (refresh) {
+        this.latestEvents = cloneDeep(newEvents)
```
snowteamer: Why is deep cloning necessary? If it wasn't then shallow copying would be more efficient

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,43 @@ export default ({
...
       } else {
-        this.messages = []
-        sbp('okTurtles.events/once', `${CHATROOM_STATE_LOADED}-${this.currentChatRoomId}`, (state) => {
-          this.messages = getLatestMessages({
-            count: this.chatRoomSettings.messagesPerPage, // TODO: this.chatRoomSettings could be {}
-            messages: state.messages
-          })
-        })
+        this.latestEvents = newEvents.concat(this.latestEvents)
```
snowteamer: As a suggestion, `this.latestEvents.splice(0, newEvents.length, ...newEvents)` might avoid old `this.latestEvents` needing garbage-collection, without losing reactivity.

backend/database.js
Silver-IT: `numberofActions` is **required** argument.
`hash` could be `undefined` as we set `currentHEAD` as default if it is `falsy` value.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Sorry, I don't understand the `natural ordering`. Do you mean `alphabetic` order?
If it is, I have just updated.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Also reordered in alphabetic.

backend/database.js
Silver-IT: I used this to wait without adding any text. Please let me know if there is better solution.

backend/routes.js
Silver-IT: Nice catch. Thanks.

frontend/model/contracts/chatroom.js
Silver-IT: I just wanted to distinguish between `replyingMessage` in ChatMain.vue.
`replyingMessage` in ChatMain.vue is just message text whereas here it is an object of `text` and `id`.

Silver-IT: Also in `mesage` object, we have `id` and `text` fields so `replyingMessage` makes sense here too.
@taoeffect, what do you think?

frontend/model/contracts/chatroom.js
Silver-IT: `findMessageIdx` function should return an index as we can use the message in the index and also we can use index itself when we remove message just like this `state.messages.splice(msgIndex, 1)`

frontend/model/contracts/chatroom.js
Silver-IT: Or do you mean we should make a variable `message` and use it below?
`const message = state.messages[msgIndex]`

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Is it `this.latestEvents.splice(0, 0, ...newEvents)`?

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Removed `cloneDeep`. It was not necessary. Thanks

frontend/views/containers/chatroom/ChatMain.vue
snowteamer: Yea, that's was I meant. Thank you.

backend/database.js
snowteamer: I was just worried that there could be missing separating commas in the output. If I was wrong then it's fine.

frontend/model/contracts/chatroom.js
snowteamer: I didn't catch the distinction. So maybe replyingMessage in ChatMain.vue should be renamed to replyingMessageText since it is plain text. Although I'm not sure because that name sounds a bit like it was about the text of the reply rather than the text of the original message. Good names are sometimes hard to find.

frontend/views/containers/chatroom/ChatMain.vue
snowteamer: Hmm yes, sorry ! 😅

snowteamer (dismissed):

frontend/model/contracts/chatroom.js
snowteamer: Why not. But it's not necessary if you don't want an extra line.

frontend/model/contracts/chatroom.js
taoeffect: @Silver-IT is this what we're replying to? If so what about `replyingTo` or `replyTo`?

taoeffect: `replyingMessage` also seemed fine to me. Or the more descriptive but longer, `replyingToMessage`.

backend/database.js
Silver-IT: Thanks for your deep observation.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Thanks. Done.

frontend/model/contracts/chatroom.js
Silver-IT: Okay. `replyingMessage`!

Silver-IT: Oh sorry. I didn't implement `message replying` function in this PR. It's in the next PR. #1279

backend/database.js
```diff
@@ -51,6 +51,52 @@ export default (sbp('sbp/selectors/register', {
...
+      throw Boom.notFound(`contractID ${contractID} doesn't exist!`)
+    }
+
+    const hash = before || currentHEAD
+    let prefix = '['
+    if (hash === currentHEAD) {
+      const entry = await sbp('chelonia/db/getEntry', currentHEAD)
+      prefix = `["${strToB64(entry.serialize())}",`
+      limit--
+    }
```
taoeffect: This section doesn't make much sense to me, and looks like a DRY violation.

`before` is not marked as being an optional parameter, and it shouldn't be an optional parameter, so this `if` statement and `const hash = before || currentHEAD` should be removed.

backend/database.js
```diff
@@ -51,6 +51,52 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
+  'backend/db/streamEntriesBefore': async function (contractID: string, before: string, limit: number): Promise<*> {
+    let currentHEAD = await sbp('chelonia/db/latestHash', contractID)
+    if (!currentHEAD) {
+      throw Boom.notFound(`contractID ${contractID} doesn't exist!`)
+    }
```
taoeffect: Why is this retrieving the latestHash for the contract? It's unclear to me why that's necessary. You just need to start at `before` and go backwards, right?

backend/database.js
```diff
@@ -51,6 +51,52 @@ export default (sbp('sbp/selectors/register', {
...
+          } else if (currentHEAD === hash) {
+            metBefore = true
+          }
+
+          if (!limit || currentHEAD === contractID) {
+            this.push(']')
+            this.push(null)
+          } else {
+            currentHEAD = entry.message().previousHEAD
+            this.push('')
```
taoeffect: I think this code could be greatly simplified. I do not see a need for the `metBefore` variable.

Something like:

```js
  'backend/db/streamEntriesBefore': async function (before: string, limit: number): Promise<*> {
    let prefix = '['
    let currentHEAD = before
    let entry = await sbp('chelonia/db/getEntry', currentHEAD)
    if (!entry) {
      throw Boom.notFound(`entry ${currentHEAD} doesn't exist!`)
    }
    currentHEAD = entry.message().previousHEAD
    return new Readable({
      async read (): any {
        try {
          if (!currentHEAD || limit === 0) {
            this.push(']')
            this.push(null)
          } else {
            entry = await sbp('chelonia/db/getEntry', currentHEAD)
            const json = `"${strToB64(entry.serialize())}"`
            currentHEAD = entry.message().previousHEAD
            limit -= 1
            this.push(prefix + json)
            prefix = ','
          }
        } catch (e) {
          // TODO: properly return an error - I don't think this lets the caller know there was an error
          console.error(`read(): ${e.message}:`, e)
          this.push(']')
          this.push(null)
        }
      }
    })
  },
```

(Warning: I didn't test or check the above code for bugs, so please double-check me!)

Notice:

- We don't need the `contractID`, so it's removed from the parameters
- We don't need `metBefore`
- Both `'backend/db/streamEntriesBefore'` and `'backend/db/streamEntriesSince'` fail to properly inform the caller of an error during the string. That's why I added the `TODO:` there. Maybe either you or @snowteamer know how to handle that properly? I'd check the Node.js documentation for this (and maybe the Hapi.js documentation)

backend/routes.js
```diff
@@ -66,6 +66,21 @@ route.GET('/events/{contractID}/{since}', {}, async function (request, h) {
   }
 })
 
+route.GET('/eventsBefore/{contractID}', {}, async function (request, h) {
```
taoeffect: - This doesn't need `contractID` in the parameters.
- We must create a uniform and consistent interface for our backend API. If you think switching to query parameters is better (and I'm inclined to agree), then you need to argue why it's better, and also update the other APIs to be consistent. For example, you could say that query parameters are more natural for this use-case and are better supported across different libraries and frameworks. I would believe that, as it's true (@snowteamer do you agree with that reasoning? I know you're working on Deno, so your input would be valuable here).
- And so, if we switch to query parameters, we also need to change the `'/events/{contractID}/{since}'` to be consistent. So for example, we'd change that API to be `/eventsSince?contractID={cID}&since={since}` and update all of the code accordingly.

frontend/model/contracts/chatroom.js
```diff
@@ -137,25 +129,11 @@ function createNotificationData (
...
-      if (state?.messages[i].id === hash) {
-        data = { message: { ...state?.messages[i] } }
-        break
-      }
-    }
-  } else if (type === MESSAGE_ACTION_TYPES.DELETE_MESSAGE) {
-    data = { id: hash }
-  }
-  sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, { type, data })
+  sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, { hash })
```
taoeffect, line 136: Nice simplification 👍

backend/routes.js
snowteamer: I'm not sure which is better between query parameters and route parameters. Query parameters are likely easier to implement in vanilla, but even a simple router will usually support route parameters too (like the router from the Pogo Deno framework which I'm experimenting with).

[This post](https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api) might give some insights.

backend/database.js
snowteamer: According to the [docs](https://nodejs.org/api/stream.html#errors-while-reading), if I understand correctly no error must be manually thrown here in `read()`, but [destroy(error)](https://nodejs.org/api/stream.html#readabledestroyerror) should be used instead. This will result in an `error` event to be emitted, with the given error argument as payload.

Also note that `destroy()` calls `_destroy()` internally, which can be [overriden](https://nodejs.org/api/stream.html#readable_destroyerr-callback) in case custom behavior is needed.

shared/domains/chelonia/chelonia.js
```diff
@@ -260,6 +260,13 @@ sbp('sbp/selectors/register', {
     // calling this will make pubsub unsubscribe for events on `contractID`
     sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, state.contracts)
   },
+  'chelonia/contractEventsBefore': async function (contractID: string, before: string, limit: number) {
+    if (limit <= 0) {
+      console.error('[chelonia] invalid params error: "limit" needs to be positive integer')
+      return
+    }
+    return await sbp('chelonia/private/out/eventsBefore', contractID, before || '', limit)
+  },
```
taoeffect: I'd make a few suggestions here:

1. Rename to `'chelonia/out/eventsBefore'`
2. Remove `contractID` from the parameters
3. Delete the `'chelonia/private/out/eventsBefore'` selector and move its implementation into here, since this is essentially just a wrapper around that

taoeffect: *edited above message to suggest rename to `'chelonia/out/eventsBefore'`.

Also, for consistency then, might be worth also moving `'chelonia/private/out/eventsSince'` into this file and renaming it to `'chelonia/out/eventsSince'`

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -29,6 +29,7 @@
         )
 
         p.c-text(v-else-if='text') {{ text }}
+          span.c-edited(v-if='edited') (edited)
```
taoeffect: This should use `i18n`

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,43 @@ export default ({
...
-      if (this.isJoinedChatRoom(this.currentChatRoomId)) {
-        this.messages = cloneDeep(this.chatRoomLatestMessages)
+    getSimulatedState (initialize = true) {
+      return {
+        settings: cloneDeep(this.chatRoomSettings),
+        attributes: cloneDeep(this.chatRoomAttributes),
+        users: cloneDeep(this.chatRoomUsers),
+        messages: initialize ? [] : this.messages,
+        simulation: true
+      }
```
taoeffect: This seems a bit repetitive, wouldn't it be easier to first clone `currentChatRoomState` and then update the individual properties that differ?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,43 @@ export default ({
...
+        this.latestEvents = newEvents
       } else {
-        this.messages = []
-        sbp('okTurtles.events/once', `${CHATROOM_STATE_LOADED}-${this.currentChatRoomId}`, (state) => {
-          this.messages = getLatestMessages({
-            count: this.chatRoomSettings.messagesPerPage, // TODO: this.chatRoomSettings could be {}
-            messages: state.messages
-          })
-        })
+        this.latestEvents.splice(0, 0, ...newEvents)
```
taoeffect: A bit cleaner I think: `this.latestEvents.unshift(...newEvents)`

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,43 @@ export default ({
...
+      }
+
+      const state = this.getSimulatedState(true)
+      for (const event of this.latestEvents) {
+        await sbp('chelonia/private/in/processMessage', GIMessage.deserialize(event), state)
+      }
+      this.messages = cloneDeep(state.messages)
```
taoeffect: Did you double-check and verify that this `cloneDeep` here is necessary? What happens if you remove it?

taoeffect (changes requested):
Review finished! Great job @Silver-IT! Very interesting PR!

After requested changes are implemented I'll go ahead and test it out locally and will give it a final review. Let me know if you have any questions.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch. I don't think it is necessary here. As a matter of fact, the emoticon issue was because of this `cloneDeep`. It made me to use `cloneDeep` but I should think twice before to use it. I will take care of it from the next time. :grin:

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Perfect. That's correct.

frontend/model/contracts/chatroom.js
Silver-IT: Yeap.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch. Need to be fixed.

frontend/views/containers/chatroom/MessageBase.vue
Silver-IT: Correct. Done.

backend/database.js
Silver-IT: If there is no `before` in API queries, `before` will be **undefined** and also `before` could be an empty string.

backend/database.js
Silver-IT: Correct. I also think it is weird. I need to check it again. I am not sure why I did it this way. :grin:

backend/database.js
Silver-IT: Improved and optimized this `streamEntriesBefore` sbp function
- removed `metBefore` variable indeed
- I didn't remove the `contractID` because there could be the cases when we are going to get 40 latest events. In this case, we need to know whether the event is the latest one or not. Also in this case, there is one thing different which is to set the default `currentHEAD`.

> Both 'backend/db/streamEntriesBefore' and 'backend/db/streamEntriesSince' fail to properly inform the caller of an error during the string. 

I have no idea right now when the errors occurs and how we should handle. I need to study this more.

shared/domains/chelonia/chelonia.js
Silver-IT: Got it. I think `contractID` should be there. I made coment for this above.

Silver-IT: Got it.

backend/database.js
taoeffect: > - I didn't remove the `contractID` because there could be the cases when we are going to get 40 latest events. In this case, we need to know whether the event is the latest one or not. Also in this case, there is one thing different which is to set the default `currentHEAD`.

@Silver-IT But the problem with this approach is that it unnecessarily complicates the API and creates a somewhat unclean interface.

For example, with the way the code is currently written, it's possible to call the API with a `before` value from `contractA`, and a `contractID` value that corresponds to `contractB`. It will still work, but no error will be thrown, and this sort of API is therefore confusing. I don't think the answer is to throw an error in this instance - that will only complicate the code unnecessarily even more.

We want the Chelonia API to be a **simple** as possible - with no unnecessary complexities. From our CONTRIBUTING guidelines:

> A PR SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.

By "minimal answer", it means if something can be simplified, it should.

In Chelonia, events are a linked list pointing backwards in time:

```
[ ContractA ] <- [ Event1 ] <- [ Event2 ]  <- [ Event3 ]
```

The way this API is written right now asks for 3 inputs:

```
[ ContractA ] <- [ Event1 ] <- [ Event2 ]  <- [ Event3 ]
     ^                                        ^
     |                                        |
Input1                                        Input2

40 <--- Input3
```

But here "Input1" (the `contractID`) is unnecessary. A simpler, more minimal API exists that only asks for 2 inputs:

```
[ ContractA ] <- [ Event1 ] <- [ Event2 ]  <- [ Event3 ]
                                              ^
                                              |
                                            Input1

40 <--- Input2
```

If you want the 40 latest events, then before calling this API, you can call `'chelonia/private/out/latestHash'`.

In fact, we should move that to be a public API (move it from `internals.js` to `chelonia.js`), and rename it to `'chelonia/out/latestHash'`. And then call `'chelonia/out/eventsBefore'`.

Silver-IT: > For example, with the way the code is currently written, it's possible to call the API with a before value from contractA, and a contractID value that corresponds to contractB. It will still work, but no error will be thrown, and this sort of API is therefore confusing. I don't think the answer is to throw an error in this instance - that will only complicate the code unnecessarily even more.

Make sense.

One more thing is that the reason why we need to know is that
If the `before` is the latest one, we need to get 40 events from `before` including **itself**.
If the `before` is not the latest one, we need to get 40 events before `before` hash.

taoeffect: Just making a note here to record that based on Slack conversation we figured out the solution to this would be to simply always include `before`

Silver-IT: Perfect.
So the return value should contain events like the following.
up to `limit` number of events before `before` + `before`

backend/database.js
```diff
@@ -51,6 +51,39 @@ export default (sbp('sbp/selectors/register', {
...
+            entry = await sbp('chelonia/db/getEntry', currentHEAD)
+            const json = `"${strToB64(entry.serialize())}"`
+            this.push(prefix + json)
+            prefix = ','
+
+            limit--
+            currentHEAD = entry.message().previousHEAD
+          }
+        } catch (e) {
+          console.error(`read(): ${e.message}:`, e)
```
taoeffect, line 80: Could you add above line 80 the following comment?

```js
// TODO: properly return an error to caller, see https://nodejs.org/api/stream.html#errors-while-reading
```

backend/database.js
```diff
@@ -51,6 +51,39 @@ export default (sbp('sbp/selectors/register', {
...
+        try {
+          if (!currentHEAD || !limit) {
+            this.push(']')
+            this.push(null)
+          } else {
+            entry = await sbp('chelonia/db/getEntry', currentHEAD)
+            const json = `"${strToB64(entry.serialize())}"`
+            this.push(prefix + json)
+            prefix = ','
+
```
taoeffect: Could you delete this empty line?

backend/database.js
Silver-IT: Sure. Why not.

backend/database.js
Silver-IT: Okay.

taoeffect (approved):
Approved! Great job @Silver-IT! Really impressive work!

I had some weird errors that I couldn't reproduce related to duplicate keys in ChatMain after joining a channel with several users that had over 120 events in it, but I couldn't reproduce it.

Screen Shot 2022-05-24 at 6 19 19 PM

Screen Shot 2022-05-24 at 6 19 26 PM

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -303,17 +326,45 @@ export default ({
...
-      if (this.isJoinedChatRoom(this.currentChatRoomId)) {
-        this.messages = cloneDeep(this.chatRoomLatestMessages)
+    getSimulatedState (initialize = true) {
+      return {
+        settings: cloneDeep(this.chatRoomSettings),
+        attributes: cloneDeep(this.chatRoomAttributes),
+        users: cloneDeep(this.chatRoomUsers),
+        messages: initialize ? [] : this.messages,
+        saveMessage: true
+      }
```
taoeffect, line 336: Oh, it looks like my comment here about using `cloneDeep` on the entire state instead of individually didn't get incorporated?

taoeffect: Nevermind - it turns out that it must be done this way because in a chatroom with many messages, when we previous the chatroom, the chatroom constructor event hasn't been processed, and so it hasn't set the `initialState` to fill in the settings/attributes/users etc. The getters on the other hand will set those to empty objects/arrays.

PR #1278, opened 2022-05-16
Create "Support History" widget for ContributionsOverviewWidget

pieer:


taoeffect (changes requested):
Pierre, this PR looks really good!

The only concern that I have is about the ordering of the months.

On Figma in Desktop mode, the months are ordered from left-to-right in ascending mode, whereas here they're ordered from left-to-right in descending mode (going backwards in time, with the oldest month on the right).

I think I prefer the way it's done on Figma, and the reason is, anytime a time-series graph is drawn with an X and a Y axis, time is almost always depicted as increasing to the right instead of to the left. As a simple example, look at charts that depict the price of a stock or crypto, time always increases to the right.

However, for mobile - I think the way it's currently implemented here is correct, with the most recent month shown at the top. The reason being - anytime on a social network or a news website or in a notifications list where time is displayed from top-to-bottom, newer events are always shown at the top, as it's done in this PR.

So my one request would be to just reverse the display of time in desktop mode only, but keep it the way it is right now for mobile.

taoeffect (approved):
Fantastic! Approved!

PR #1279, opened 2022-05-18
Reply message & Scolling inside the Group Chat

### LINKED ISSUE #1281: Show scrollbar in chat
### Problem

The chat currently doesn't show the scrollbar.

### Solution

See https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=21773%3A201263 for details on the scrollbar & implement.

Silver-IT:
### Summary of changes

- Reply Message
- Save scroll position for each channel
- `Jump to latest message` button
- Closes #1281


pieer (dismissed):
Excellent work!

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -259,18 +296,58 @@ export default ({
...
-      this.updateScroll()
     },
-    updateScroll () {
+    async scrollToMessage (messageId, focus = true) {
+      if (!messageId) {
+        return
+      }
+
+      const scrollAndHighlight = (index) => {
+        const eleMessage = document.querySelectorAll('.c-body-conversation > .c-message')[index]
```
pieer, line 308: Could you use ref here to access the Dom here?

pieer (commented):
Awesome work!
I was working on this yesterday... I'm glad i didn't spent too much time on it though.. but I came with similar approach.. and you did really good implementation :)

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Sorry @pieer, I have tried already.
I couldn't get the expected value of `allElements[i].offsetParent.offsetTop`.

```
const allElements = document.querySelectorAll('.c-body-conversation > .c-message')
for (let i = allElements.length - 1; i >= 0; i--) {
  if (allElements[i].offsetTop - allElements[i].offsetParent.offsetTop < curScrollTop) {
    sbp('state/vuex/commit', 'setChatRoomScrollPosition', {
      chatRoomId: this.currentChatRoomId,
      messageId: this.messages[i].id
    })
    break
  }
}
```

Can you please give me some idea about it?

taoeffect (changes requested):
Nice job @Silver-IT!

I only had a chance to test the scrollbar right now, and think I found two issues:

1. When there is no need to scroll (because there aren't enough messages to display a scrollbar), a dark bar still appears on the righthand side.
2. Saving the chatroom scroll position doesn't seem to work? See GIF:

scrolling bug mov

Silver-IT:
> 1. When there is no need to scroll (because there aren't enough messages to display a scrollbar), a dark bar still appears on the righthand side.

Okay, I will remove the scroll bar on the righthand side when there is no need to scrol...
Having checked, but it doesn't display the scroll bar when there is nothing to scroll.

> 2. Saving the chatroom scroll position doesn't seem to work? See GIF:

That's because we don't remove the scroll position when the difference between scrollTop and scrollTopMax is less than 500px.
When it is more than 500px, you can see the `Jump to the latest message` button.

taoeffect (changes requested):
Two bugs found:

### bug 1 - missing | in front of quote

There is a [bug in the Figma design](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=14384%3A166184). There is a [second design on Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=14387%3A166327) that shows how quotes are supposed to have a "|" in front of them, like this:

> Test quote

And actually, the second Figma design is also slightly wrong, as the "|" is slightly too high and should be lower (I opened up a [comment/request](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=14387:166327#203953698) on Figma to fix this). But it's more accurate.

Screen Shot 2022-05-31 at 5 37 00 PM

### bug 2 - The text "To latest message"

It should read "Jump to latest message" instead of "To latest message"

taoeffect (commented):
Two issues still seem to remain:

1. The scroll position still doesn't seem to be saved when I switch between channels, even with lots of messages. It always jumps back down to the latest message when I switch back.
2. When there's no reason to have a scrollbar, there's a rectangle that's still visible on the right (see screenshot below). Is it possible to hide it?

Screen Shot 2022-06-01 at 6 54 16 PM

Silver-IT:
> 1. The scroll position still doesn't seem to be saved when I switch between channels, even with lots of messages. It always jumps back down to the latest message when I switch back.

Yeap. It doesn't work in `Google Chrome`. Noticed that there isn't `scrollTopMax` attribute in the element inside Google Chrome browser. Need another solution. `scrollHeight` could be helpful.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTopMax

> 2. When there's no reason to have a scrollbar, there's a rectangle that's still visible on the right (see screenshot below). Is it possible to hide it?

Should use `overflow-y: auto` instead of `overflow-y: scroll`.
In Firefox browser, `overflow-y: scroll` also works, but not in Google Chrome browser.
So used `overflow-y: auto`. Done.

frontend/model/state.js
```diff
@@ -32,6 +32,7 @@ if (typeof (window) !== 'undefined' && window.matchMedia && window.matchMedia('(
 const initialState = {
   currentGroupId: null,
   currentChatRoomIDs: {}, // { [groupId]: currentChatRoomId }
+  chatRoomScrollPosition: {}, // [chatRoomId]: messageId
```
taoeffect, line 35: As you're introducing a new key, remember that old clients don't have this saved in their settings.

So you need to use `'state/vuex/postUpgradeVerification'` to set it.

taoeffect (changes requested):
Nice! The scrollbar is no longer visible when it shouldn't be. Great job @Silver-IT!

Posted a comment about one issue related to `'state/vuex/postUpgradeVerification'`, and also the scrolling seems to be sortof remembered, but it's really clumsy and not accurate. It doesn't behave like Slack for example. Maybe it's an issue with the vue-infinite-scroll plugin?

Here's what it looks like, it doesn't remember exactly where it should go:

scrolling mov

Silver-IT:
> Posted a comment about one issue related to 'state/vuex/postUpgradeVerification', and also the scrolling seems to be sortof remembered, but it's really clumsy and not accurate. It doesn't behave like Slack for example. Maybe it's an issue with the vue-infinite-scroll plugin?

Correct. I know that issue too. We scroll up the page to the saved position when user opens a chatroom.
Saved position means saved message. We can not save the scroll position, because there could be many more messages could be added.
So we retrieve all the events since the saved messageId and render and scroll up to it. So it works just like targetMessageId.
Because targetMessageId is the first message among the loaded messages the scroll position is almost 0 and infinite scroll works again. So it will load one more page of events and render.
At that point, that weird thing happens. The `infinite-vue-loading` automatically scrolls the page to the top when it completes the inifinite loading and no more data is there to be rendered,
That's why it looks something weird. I know that but I don't think that is the big problem.
As long as we use the channel, we don't normally scroll up to the top of the message by scrolling page by page.

frontend/model/state.js
Silver-IT: Nice catch...

Silver-IT:
I have found an issue regarding the scrolling to the saved position. As we are using messageId as saved position, we should think about the case when the message is removed.
....

frontend/views/containers/chatroom/ChatMain.vue
pieer: The ref should be on the component
component(is='messageType' ref='message.id' offset=''...)
Then loop on messages to get already calculted offset for quick comparison between curScrollTop and this.refs[message.id].offset

taoeffect:
> I have found an issue regarding the scrolling to the saved position. As we are using messageId as saved position, we should think about the case when the message is removed.

This is why I think it's better to think in terms of Y-axis offset instead of scrolling to a specific message... If that's possible. If the message is deleted, it won't matter, since it would be scrolling to an Y-axis offset value instead of to a message. It would look more like how Slack does it.

Silver-IT:
> This is why I think it's better to think in terms of Y-axis offset instead of scrolling to a specific message... If that's possible. If the message is deleted, it won't matter, since it would be scrolling to an Y-axis offset value instead of to a message. It would look more like how Slack does it.

How could it be? Y-axis offset is not absolute stuff, is it?
It depends on how many messages are displayed.
image
Neither offset from top nor offset from bottom are constant. I am not sure that can solve the problem...

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT:
```
component(
          :is='messageType(message)'
          :ref='message.id'
          ...
)
```
After define `ref` like the above, I can access `offsetTop` of each element in the following way.

`this.$refs[m.id][0].$el.offsetTop`

Is there any shortcut to the offsetTop? `this.$refs[m.id][0].$el.offsetTop` seems too complicated.

taoeffect (changes requested):
Nice job @Silver-IT!

The changes seem to be improving things a bit. I will need to do another review later to fully review the changes you've made to `streamEntriesSince`, but for now here's some quick feedback.

**Scroll jumping issue**

It is still "jumping" to the scroll position instead of instantly appearing there, but it seems to be faster. It would be nice if you could have it instantly appear there instead of doing the jump... but maybe this is a problem with `vue-infinite-loading`? (Have you tried seeing if maybe there's a better infinite scroll library that doesn't have this issue?)

**Errors on backend**

Are these supposed to be there?

```
[sbp] backend/db/streamEntriesSince [
  '21XWnNKGkzTVz7243Q7YZqeBBQX32wCkzk39Lk9hFaSdYVYMSz',
  '21XWnNL6y5K6f5dY4kANdrcBqL1iB9EJjNdsrjcKNeaY4d8NN4',
  20
]
read(): Error during getEntry: no entry for null!: ChelErrorDBConnection: Error during getEntry: no entry for null!
    at Object.chelonia/db/getEntry (group-income/shared/domains/chelonia/db.js:53:13)
```

**No longer scrolling to latest message**

I think maybe while trying to fix another issue, a bug was introduced. Now, when we're scrolled all the way down to the bottom of the chat, and someone sends a message, it no longer automatically scrolls down to the new message:

scrolling mov

If we're already scrolled to the bottom, it should automatically scroll down. If we aren't at the bottom, then it shouldn't scroll.

backend/database.js
```diff
@@ -35,12 +36,19 @@ export default (sbp('sbp/selectors/register', {
...
+            isMet = true
+          } else if (isMet) {
+            offset--
+          }
+
+          this.push(prefix + json)
+          currentHEAD = entry.message().previousHEAD
+          prefix = ','
+
+          if (isMet && offset <= 0) {
```
taoeffect: This is wrong... and this is why I saw these errors:

```
read(): Error during getEntry: no entry for null!: ChelErrorDBConnection: Error during getEntry: no entry for null!
    at Object.chelonia/db/getEntry (group-income/shared/domains/chelonia/db.js:53:13)
```

But not only that, this function shouldn't have been modified. Doesn't it make more sense to use `'backend/db/streamEntriesBefore'` instead of loading all of the events that we already have (+ 20) each time the user scrolls?

Silver-IT: I should remove this offset parameter here in this `streamEntriesSince` API and should make a new API `streamEntriesBetween`.

Silver-IT:
> It is still "jumping" to the scroll position instead of instantly appearing there, but it seems to be faster. It would be nice if you could have it instantly appear there instead of doing the jump... but maybe this is a problem with vue-infinite-loading? (Have you tried seeing if maybe there's a better infinite scroll library that doesn't have this issue?)

Yeap. It works that way as you guess. Of course, I am using the `vue-infinite-loading` library.
At the time of switching channel, it needs to display the saved position(saved message).
To display `Saved Message` requires to load events from the backend and render.
As soon as we load and render events and display messages in the UI, it will automatically jumps to the specific message.
I just removed the `behavior: smooth`. So it just jumps very fast now. So far, I haven't found any alternative solution for this.

_The other solution could be display `skeleton` until we jumps to the saved message after switching channel._

backend/database.js
```diff
@@ -92,6 +84,47 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
+  'backend/db/streamEntriesBetween': async function (startHash: string, endHash: string, offset: number): Promise<*> {
+    offset = parseInt(offset) === isNaN ? 0 : Math.max(0, parseInt(offset))
```
taoeffect: This comparison is incorrect, isNaN is a function, and typeof would make more sense here

Silver-IT: Correct. Poor concentration now.

Silver-IT:
> I think maybe while trying to fix another issue, a bug was introduced. Now, when we're scrolled all the way down to the bottom of the chat, and someone sends a message, it no longer automatically scrolls down to the new message:
> If we're already scrolled to the bottom, it should automatically scroll down. If we aren't at the bottom, then it shouldn't scroll.

Having tested on my side, I noticed that's because the tab is inactive inside the browser.
If you open two browsers and test it, you can see it works.

Example_2 (Not Working)
Example_1 (Working)

But one thing I have found too.
We need to scroll to the bottom only when new message is ADDED by ANOTHER.
Why ANOTHER? It scrolls down automatically as soon as send message even though it is successfully sent or not. (Just like how Slack works.)

Silver-IT:
> I think maybe while trying to fix another issue, a bug was introduced. Now, when we're scrolled all the way down to the bottom of the chat, and someone sends a message, it no longer automatically scrolls down to the new message:
> If we're already scrolled to the bottom, it should automatically scroll down. If we aren't at the bottom, then it shouldn't scroll.



I think this wo

taoeffect (changes requested):
The same problem exists:

scrolling mov

The logic is: if you are scrolled down to the bottom already, and anyone sends a chat (you or anyone else), it should scroll down to the bottom again automatically.

If you've scrolled upward and are not at the bottom, and someone enters a chat message, it shouldn't scroll down, but it should display (if it isn't already displayed) "Jump to latest message".

backend/database.js
```diff
@@ -84,6 +84,47 @@ export default (sbp('sbp/selectors/register', {
...
+    let entry = await sbp('chelonia/db/getEntry', currentHEAD)
+    if (!entry) {
+      throw Boom.notFound(`entry ${currentHEAD} doesn't exist!`)
+    }
+    // NOTE: if this ever stops working you can also try Readable.from():
+    // https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options
+    return new Readable({
+      async read (): any {
+        try {
+          entry = await sbp('chelonia/db/getEntry', currentHEAD)
```
taoeffect, line 100: The way this function is written, it will still throw an exception because `currentHEAD` can be `null` at this point. And indeed that's what I'm still getting in the logs:

```
read(): Error during getEntry: no entry for null!: ChelErrorDBConnection: Error during getEntry: no entry for null!
```

backend/routes.js
```diff
@@ -82,6 +82,22 @@ route.GET('/eventsBefore/{before}/{limit}', {}, async function (request, h) {
   }
 })
 
+route.GET('/eventsBetween/{startHash}/{endHash}', {}, async function (request, h) {
+  try {
+    const { startHash, endHash } = request.params
+    const { offset } = request.query
```
taoeffect: Here is where it should convert `offset` to a `number` if it's defined, or `0` if it isn't, so that `streamEntriesBetween` can expect a type of `number` for `offset`

backend/database.js
```diff
@@ -84,6 +84,47 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
+  'backend/db/streamEntriesBetween': async function (startHash: string, endHash: string, offset: number): Promise<*> {
+    offset = isNaN(parseInt(offset)) ? 0 : Math.max(0, parseInt(offset))
```
taoeffect: `offset` is defined as type `number`, so there is no reason to use `parseInt` here. See other comment in `routes.js`.

taoeffect (changes requested):
OK, I'm ignoring the [previous review issue](https://github.com/okTurtles/group-income/pull/1279#pullrequestreview-1002087993) on the basis of @Silver-IT's rationale that the problem only happens when the tab is inactive, and if their tab is inactive it's actually OK if it doesn't scroll down, so that they start off where they were before and can catch up on unread messages.

However, I found other issues in the code, related tot `streamEntriesBetween` implementation.

backend/database.js
Silver-IT:
```
const previousHEAD = entry.message().previousHEAD
if (!currentHEAD || !previousHEAD || (isMet && !offset)) {
  this.push(']')
  this.push(null)
} else {
  currentHEAD = previousHEAD
}
```
Nice catch. I think this will fix it.

backend/routes.js
Silver-IT: Okay.

backend/database.js
Silver-IT: Okay.

backend/routes.js
```diff
@@ -74,7 +74,24 @@ route.GET('/eventsBefore/{before}/{limit}', {}, async function (request, h) {
...
+})
+
+route.GET('/eventsBetween/{startHash}/{endHash}', {}, async function (request, h) {
+  try {
+    const { startHash, endHash } = request.params
+    const { offset } = request.query
+
+    if (!startHash) return Boom.badRequest('missing startHash')
+    if (!endHash) return Boom.badRequest('missing endHash')
+    if (isNaN(parseInt(offset)) || parseInt(offset) < 0) return Boom.badRequest('invalid offset')
```
taoeffect: It should be fine to not specify the `offset` as it's an optional parameter.

Silver-IT: Nice catch.

backend/routes.js
```diff
@@ -74,7 +74,24 @@ route.GET('/eventsBefore/{before}/{limit}', {}, async function (request, h) {
...
+route.GET('/eventsBetween/{startHash}/{endHash}', {}, async function (request, h) {
+  try {
+    const { startHash, endHash } = request.params
+    const { offset } = request.query
+
+    if (!startHash) return Boom.badRequest('missing startHash')
+    if (!endHash) return Boom.badRequest('missing endHash')
+    if (offset && (isNaN(parseInt(offset)) || parseInt(offset) < 0)) return Boom.badRequest('invalid offset')
+
+    const stream = await sbp('backend/db/streamEntriesBetween', startHash, endHash, parseInt(offset || 0))
```
taoeffect: `parseInt(offset)` is called 3 times here. DRY says it should only be called once.

So, something like this:

```js
const offset = parseInt(request.query.offset || 0)

if (!startHash) return Boom.badRequest('missing startHash')
if (!endHash) return Boom.badRequest('missing endHash')
if (isNaN(offset) || offset < 0) return Boom.badRequest('invalid offset')

const stream = await sbp('backend/db/streamEntriesBetween', startHash, endHash, offset)
```

backend/database.js
```diff
@@ -84,6 +84,46 @@ export default (sbp('sbp/selectors/register', {
...
+          prefix = ','
+
+          if (currentHEAD === startHash) {
+            isMet = true
+          } else if (isMet) {
+            offset--
+          }
+
+          const previousHEAD = entry.message().previousHEAD
+          if (!currentHEAD || !previousHEAD || (isMet && !offset)) {
```
taoeffect: I think this condition should be:

```js
          if (!previousHEAD || (isMet && offset <= 0)) {
```

Because:

1. How can `currentHEAD` ever be `null` here?
2. If an `offset` of `0` was passed in, then your last condition will fail, because `offset` will be `-1`, and `!-1` is `false`

taoeffect (changes requested):
Additional issues with `streamEntriesBetween` implementation.

backend/routes.js
Silver-IT: Nice catch. Resolved.
Since the expected type of parseInt function is `string`.
I would like to use this.

```
const offset = parseInt(request.query.offset || '0');
```

backend/database.js
Silver-IT: 1. I think that's right. currentHEAD couldn't be null here.
2. Even though offset of 0 is passed in, it couldn't be -1 here because of the following condition.
```
if (currentHEAD === startHash) {
  isMet = true
} else if (isMet) {
  offset--
}
```

frontend/model/contracts/chatroom.js
```diff
@@ -116,6 +117,23 @@ export function findMessageIdx (id: string, messages: Array<Object>): number {
...
+    action.endsWith('rename') ||
+    action.endsWith('changeDescription') ||
+    action.endsWith('leave')) {
+    // we add new pending message in 'handleSendMessage' function in ChatMain.vue
+    return me !== meta.username
+  }
+
+  return false
+}
+
```
taoeffect: Since this function isn't being used in this file, could you please move it into `ChatMain.vue` where it is used?

Also, this condition could be shortened and simplified, to something like:

```js
...
const me = rootState.loggedIn.username

return /.*(addMessage|join|rename|changeDescription|leave)$/.test(action) && me !== meta.username
```

(But verify that I got that regex correct...)

frontend/model/contracts/chatroom.js
```diff
@@ -22,6 +22,7 @@ import {
 } from './constants.js'
 import { CHATROOM_MESSAGE_ACTION } from '~/frontend/utils/events.js'
 import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
+import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
```
taoeffect: Similarly, this isn't needed here anymore (see comment below)

taoeffect (changes requested):
Very nice! Just verifying changes to chatroom contract now. Noticed a few minor issues.

backend/database.js
taoeffect: Nice catch 👍

frontend/model/contracts/chatroom.js
Silver-IT: Nice idea to use regular expression. And also agree to relocate the function.

frontend/model/contracts/chatroom.js
Silver-IT: Okay. Removed...

taoeffect (approved):
Approved! Nice job @Silver-IT!

PR #1282, opened 2022-05-21
Enable info-level logging by default

### LINKED ISSUE #1268: captureLogs.js should capture console.info too and enable by default
### Problem

1. `console.info` isn't captured by the `appLogsFilter` & `AppLogs.vue` (there is no checkbox for it)
2. `console.info` and `console.log` aren't logged by default and they should be

### Solution

Fix.

Also, consider converting `appLogsFilter` from an array to an object inside of `captureLogsStart` here (while keeping it an array in the settings):

```js
  appLogsFilter = sbp('state/vuex/state').appLogsFilter || []
  sbp('okTurtles.events/on', SET_APP_LOGS_FILTER, filter => { appLogsFilter = filter })
```

So that instead of `O(n)`:

```js
        if (appLogsFilter.includes(type)) {
```

we can do the much faster `O(1)`:

```js
        if (appLogsFilter[type]) {
```

snowteamer:
Closes #1268 

### Summary of changes:
- An `info` checkbox has been added in the Application Logs page. It is checked by default.

### Additional info:
- Note that `info` won't be retroactively added to `state.appLogsFilter` for older clients, as I didn't know how to determine whether it was missing or removed by the user via the Application Logs page.

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -40,8 +40,8 @@ const initialState = {
   increasedContrast: false,
   fontSize: 16,
   appLogsFilter: process.env.NODE_ENV === 'development'
-    ? ['error', 'warn', 'debug', 'log']
-    : ['error', 'warn']
+    ? ['error', 'warn', 'info', 'debug', 'log']
+    : ['error', 'warn', 'info']
 }
 
 sbp('sbp/selectors/register', {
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
LGTM! Great job @snowteamer!

PR #1287, opened 2022-05-30
Add system colors theme

pieer:


frontend/views/containers/user-settings/ThemeSvg.vue
```diff
@@ -0,0 +1,37 @@
...
+  circle(cx='12.5' cy='32.5' r='4.5' :fill='color.primary_0')
+  circle(cx='12.5' cy='52.5' r='4.5' :fill='color.primary_0')
+  circle(cx='12.5' cy='72.5' r='4.5' :fill='color.primary_0')
+  rect(x='8' y='8.00002' width='64' height='9' rx='4.5' :fill='color.general_0')
+  circle(cx='12.5' cy='92.5' r='4.5' :fill='color.primary_0')
+</template>
+
+<script>
+export default ({
+  name: 'SelectorThemeSVG',
```
taoeffect: Shouldn't this name match the name of the file?

frontend/views/containers/user-settings/Theme.vue
```diff
@@ -1,28 +1,29 @@
...
-  fieldset.theme(v-for='(color, label) in themes' :key='label' @click='setTheme(color.name)')
+  fieldset.theme(@click='setTheme("system")')
+    label(for='system')
+      .c-combined-color
+        ThemeSvg(:color='themes.light')
+        ThemeSvg(:color='themes.dark')
+
+      .radio
+        input.input(type='radio' name='theme' value='system' id='system' v-model='theme')
+        span Use system settings
```
taoeffect: Missing `i18n`

taoeffect, line 11: Missing `i18n`

taoeffect (changes requested):
Review done! Nice job @pieer! 👍 

Found some minor issues, but also, for some reason in Firefox the theme isn't getting updated when I switch the system theme:

Screen Shot 2022-05-31 at 6 06 00 PM

I even refreshed the page and it still isn't switching.

taoeffect (changes requested):
@pieer Same issue.

Refreshing page after changing os theme doesn't update.

Ideally it will also update without a refresh if set to system theme and system theme changes.

taoeffect (commented):
I know you said you're still working on this, but just an update, I tried out the latest changes you pushed and indeed, the problem still exists. I didn't notice a difference, refreshing the page still doesn't fix it.

pieer:
> I didn't notice a difference, refreshing the page still doesn't fix it.

Weird, it's working well on my mac on firefox when I change the system settings appearance...
Could you help me to find what's the difference?
I sent you a video on Slack

frontend/model/state.js
```diff
@@ -24,9 +24,22 @@ import notificationModule from '~/frontend/model/notifications/vuexModule.js'
...
+  }
+  return theme
+}
+const defaultTheme = checkSystemColor()
+
+if (window.matchMedia) {
+  window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
+    if (store.state.theme === 'system') {
+      store.commit('setTheme', THEME_DARK)
+      store.commit('setTheme', 'system')
```
taoeffect: Why is there `store.commit('setTheme', THEME_DARK)` here?

pieer: I found the solution I think in the new PR: update the state of the colors instead of the state of the theme

frontend/model/state.js
```diff
@@ -33,6 +45,7 @@ const initialState = {
   pending: [], // contractIDs we've just published but haven't received back yet
   loggedIn: false, // false | { username: string, identityContractID: string }
   theme: defaultTheme,
+  themeColor:
```
taoeffect: Bug here causing travis to fail

frontend/model/state.js
```diff
@@ -101,6 +116,10 @@ const mutations = {
   },
   setTheme (state, color) {
     state.theme = color
+    store.commit('setColors', checkSystemColor())
```
taoeffect: For consistency and easy-of-readability sake, and since `setColors` doesn't actually take a parameter, and so that we’re not storing the entire theme data in local storage, I recommend the following changes:

- Rename the parameter in this function to `theme`
- Remove `setColors` mutation, and set the colors directly inside this function:

```js
  setTheme (state, theme) {
    state.theme = theme
    state.themeColor = theme === 'system' ? checkSystemColor() : theme
  },
```

And then modify the lines up above to:

```js
const defaultTheme = 'system'
const defaultColor = checkSystemColor()

if (window.matchMedia) {
  window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
    if (store.state.theme === 'system') {
      store.commit('setTheme', 'system')
    }
  })
}
```

And the `colors` getter below to:

```js
  colors (state) {
    return Colors[state.themeColor]
  },
```

frontend/model/state.js
```diff
@@ -19,9 +19,23 @@ import notificationModule from '~/frontend/model/notifications/vuexModule.js'
...
-let defaultTheme = THEME_LIGHT
-if (typeof (window) !== 'undefined' && window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
-  defaultTheme = THEME_DARK
+const checkSystemColor = () => {
+  let theme = THEME_LIGHT
+  if (typeof (window) !== 'undefined' && window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+    theme = THEME_DARK
+  }
+  return theme
+}
```
taoeffect: Since `window` is guaranteed to be defined, we don't need to check if it's defined and this function could be simplified to:

```js
const checkSystemColor = () => {
  return window.matchMedia?.('(prefers-color-scheme: dark)').matches
    ? THEME_DARK
    : THEME_LIGHT
}
```

frontend/views/containers/user-settings/Theme.vue
```diff
@@ -44,7 +49,8 @@ export default ({
 
   computed: {
     ...mapGetters([
-      'colors'
+      'colors',
```
taoeffect: AFAICT, `colors` is no longer used in this file so doesn't need to be here.

frontend/model/state.js
```diff
@@ -19,9 +19,23 @@ import notificationModule from '~/frontend/model/notifications/vuexModule.js'
...
+const defaultTheme = checkSystemColor()
+const defaultColor = Colors[defaultTheme]
+
+if (window.matchMedia) {
+  window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
+    console.log(store.state.theme)
+    if (store.state.theme === 'system') {
+      store.commit('setColors', checkSystemColor())
+    }
+  })
```
taoeffect: - If debug logging is still needed, please change to `console.debug`, otherwise please remove logging
- It's possible to simplify the code by calling `setTheme` here instead of `setColors`, see comment below about `setTheme`

taoeffect (changes requested):
@pieer - Great work, and sorry again that I didn't realize that Firefox Developer Edition has a unique setting that prevented me from properly testing this PR before.

This time I've done a thorough test and also a thorough review of the code, and made some suggestions for how to simplify some of the code.

Also, I'm encountering a weird issue when changing the system theme in the System Preferences, and I verified this on both Firefox and Safari.

When I set it to "System" theme and use System Preferences to switch back and forth between the system theme while in the Appearance preferences, it doesn't switch the colors for everything and I'm not sure why.

Notice here when I switched to "Light Theme" in System Preferences the menu items didn't change their color:

Screen Shot 2022-07-17 at 6 15 07 PM

And similar issue when switching back to the dark theme (notice the background shadow is white), although this only happens sometimes (I can't always reproduce it):

Screen Shot 2022-07-17 at 7 03 53 PM

If you can't figure this out it's OK, not a big deal since refreshing the page fixes the problem! I'm mainly concerned with some code simplification (see comments).

frontend/model/state.js
pieer: Thanks for the feedbacks. I fix all the issues and simplify the code according to your help

frontend/model/state.js
pieer: Fixed

frontend/model/state.js
pieer: All done

taoeffect (approved):
Excellent!

PR #1291, opened 2022-05-31
hopefully fixes bad previousHEAD error (closes #1290)

### LINKED ISSUE #1290: Fix "bad previousHEAD" error + improve handling if it happens again
### Problem

1. Live server is throwing `bad previousHEAD` errors because somehow an identity contract forked with multiple `setLoginState` actions
2. Not entirely sure if this is a problem, but double-check that the exception path when two of the same `bad previousHEAD` errors are thrown (e.g. in `addMessageToDB` when `eventsToReinjest.includes(hash)` evaluates to `true`) operates currently. Currently the exception thrown at that point prevents the extra `reprocessDebounced` from resulting in an additional contract sync, because it causes the entire eventQueue in `'chelonia/contract/sync'` to throw and stop.

### Solution

1. Find out why the server allowed the fork to occur and fix the bug.
2. Evaluate whether the current behavior is correct or whether it should allow the `reprocessDebounced` contract sync to proceed ahead (although that could possibly lead to an infinite loop?)

taoeffect:


Silver-IT (approved):
I have found nothing wrong.

PR #1292, opened 2022-05-31
Unread Messages in Group Chat

### LINKED ISSUE #1216: Chat jumps from top to bottom
### Problem

When the chat is selected, instead of immediately starting at the bottom, it "jumps" to the bottom:

chat-issue mov

### Solution

Ideally, the behavior is similar to Slack, each chat channel remembers its scroll position and goes to it, unless there's a new message, according to the [figma designs](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=14747%3A184), it should try to place the oldest newest message as close to the top of the chat view as possible, using a `----------- new ------------` separator (see Figma).

Silver-IT:
### Summary of Changes

- Implemented unread messages
- Improve styles
- Made it available to add `reactions` while editing message
- Closes #1216 
- Did https://github.com/okTurtles/group-income/pull/1299#discussion_r903161733


package-lock.json
```diff
@@ -14360,6 +14337,20 @@
         "node": ">=8"
       }
     },
+    "node_modules/readable-stream": {
```
taoeffect: I don't think this file should have been modified since no new dependency was added or removed? Do you think you can revert these changes?

Silver-IT: Yeap. Sure.

taoeffect (approved):
Tested & verified. Excellent work @Silver-IT!

Issue #531, opened 2019-02-06
Make sure backend clock is synced with clients


taoeffect:
### Problem

For handling things like onboarding periods, message send/receive timestamps, etc, we need to make sure everyone's clocks are synchronized to the correct time.

### Solution

Create a server API to get the server's time. Make sure it's the same as the client's. Do not let the group to be created if the client's clock is out of sync with the server's time. The server's time is canonical time that all clients must adhere to. Make sure to account for different timezones.

Silver-IT:
We also should consider that the client's datetime could be changed in the middle of using the platform.
He can set the correct time to create group and login but he also can update his datetime while using it.

PR #1299, opened 2022-06-09
Fix issues regarding Group Chat and Group Invitation

### LINKED ISSUE #620: Implement invite expiry

### LINKED ISSUE #1215: Remove checkIfJoinedGeneralChannel from frequently used support commands in Cypress
### Problem

The Cypress tests are now a bit slower because of how much we're testing.

The Chat PR (#1143) added `checkIfJoinedGeneralChannel` to several frequently used support commands, including `giAcceptGroupInvite` and `giCreateGroup`.

### Solution

`checkIfJoinedGeneralChannel` should be removed from `giAcceptGroupInvite` and `giCreateGroup`, and only done a couple of times in `group-chat.spec.js`

### LINKED ISSUE #1237: Remove cy.wait() in Cypress chat tests
### Problem

In #1143 we left a few `cy.wait()`s because the error handling wasn't yet fully working.

It should be working now.

### Solution

@Silver-IT Please go ahead and try to remove all the remaining `cy.wait()`s that you had in the chat tests after merging `master`.

### LINKED ISSUE #1254: Do not allow duplicate chatroom names
### Problem

Currently duplicate channel names are allowed.

### Solution

Prevent channels with a duplicate name from being created. Do a check in two places:

1. A check in the chat creation modal using vuelidate if possible (so that the field is immediately marked red)
2. A check in either the validation or the process function the chatroom creation message

This is similar to doing logic checks on "both the frontend and the backend", except for us the "backend" is the contracts themselves.

### LINKED ISSUE #1269: You can add members to a channel even if you're not a part of it
### Problem

The UI lets you add members to a channel you haven't even joined yourself:

Screen Shot 2022-05-04 at 7 58 26 PM

### Solution

Probably should do two things at least:

1. Hide the "add members" button in the UI if you're not part of the channel
2. In the contracts, don't allow someone who isn't part of the channel themselves to add a member to it

### LINKED ISSUE #1296: People joining the chat channel multiple times
### Problem

On the live server, I see multiple join entries for both Leihla and Ricardo:

Screen Shot 2022-06-05 at 12 27 50 PM

### Solution

Find out why this is happening and fix.

Silver-IT:
### Summary of Changes

- Closes #1269 
- Closes #1254 
- Closes #1215 
- Closes #1296 
- Closes #1237 
- Closes #620


frontend/model/contracts/group.js
```diff
@@ -302,7 +297,11 @@ sbp('chelonia/defineContract', {
       const pendingMembers = {}
       for (const inviteId in invites) {
         const invite = invites[inviteId]
-        if (invite.status === INVITE_STATUS.VALID && invite.creator !== INVITE_INITIAL_CREATOR) {
+        if (
+          invite.status === INVITE_STATUS.VALID &&
+          invite.creator !== INVITE_INITIAL_CREATOR &&
+          invite.expires >= Date.now()
```
taoeffect: Nowhere inside the contract code — except for sideEffects — can you call `Date.now()`.

Getters and process functions must always behave the same no matter where they are run. `Date.now()` is unique to each individual client that calls it. Must be removed here.

Silver-IT: Understand. Nice catch...

Silver-IT: But if we think that way, what about the `createdDate` of metadata.

```
createdDate: new Date().toISOString()
```

This depends on the creator's local date and if it is not correct, the wrong date will be added to contract.
How can we handle this?
This also happends inside the `createInvite` function too.

Silver-IT: I think this is important problem... We should use the one datetime of backend, shouldn't we?

taoeffect: > This depends on the creator's local date and if it is not correct, the wrong date will be added to contract.
How can we handle this?

Issue #531 is actually about this. A `/time` route exists in the backend. The only issue is that we currently don't use it. We would have to change the contract `create()` function for metadata to be asynchronous for this to work...

(Although it should be noted that we cannot guarantee that a hacked client won't still send the wrong date.)

However, all of this is besides the point here - we cannot use `Date.now()` here regardless.

Inside of the contract getters and contract process functions you can only use the `data` and `meta`data.

Silver-IT: @taoeffect, I can't find the better place and solution to check expiry.
Sending all pending users (no matter their invites are expired) to the frontend and checking their expiry in frontend components, doesn't seems to be a good practice.
I can also find the `isNewMember` function in `groupMembersSorted` getter which using Date.now().
Line 436 in frontend/model/state.js

The best solution is to create a sbp function to get the Date.now() on backend and use it. Of course, I am not sure this could be also acceptable because we are not using `sbp` functions inside process and getters.

If it is the solution of the #531, and also if we should do that way, I think I can skip it in this PR and resolve it creating a new PR from the #531.

taoeffect: > The best solution is to create a sbp function to get the Date.now() on backend and use it.

... this really isn't necessary.

If all you want is to filter the pending members, you can modify this getter to include the expiry, and create a new getter in state.js that filters the results of this getter. The getters in state.js are allowed to use `Date.now()` and anything else.

Silver-IT: Okay. Understood why the solution should be like it.

frontend/controller/actions/group.js
```diff
@@ -185,26 +190,38 @@ export default (sbp('sbp/selectors/register', {
...
+        const rootState = sbp('state/vuex/state')
+        const chatRoomIds = Object.keys(rootState[params.contractID].chatRooms)
+
+        for (const cId of chatRoomIds) {
+          await sbp('chelonia/contract/sync', cId)
+        }
+        sbp('state/vuex/commit', 'setCurrentChatRoomId', {
+          groupId: params.contractID,
+          chatRoomId: rootState[params.contractID].generalChatRoomId
+        })
```
taoeffect: - Why is this code being added to the `else`?
- And why is it subscribing to all the chatrooms?

frontend/controller/actions/group.js
```diff
@@ -250,7 +274,14 @@ export default (sbp('sbp/selectors/register', {
...
     try {
       const rootState = sbp('state/vuex/state')
-      const username = params.data.username || rootState.loggedIn.username
+      const rootGetters = sbp('state/vuex/getters')
+      const me = rootState.loggedIn.username
+      const username = params.data.username || me
+
+      if (!rootGetters.isJoinedChatRoom(params.data.chatRoomID) && username !== me) {
+        throw new GIErrorUIRuntimeError(L('Failed to join chat channel.'))
+      }
```
taoeffect: What is this condition doing? It doesn't make sense to me.

taoeffect (commented):
Posted to questions and also I'm wondering if you can add a 1 test to test the invite link expiry using `cy.clock()`?

frontend/controller/actions/group.js
Silver-IT: > Why is this code being added to the else?

`else` part is for logging in in different device. User doesn't need to make `join` and `joinChatRoom` group actions, but he should sync the chatroom contracts which he already joined.

> And why is it subscribing to all the chatrooms?

Nice catch. User should sync the chatroom contracts he already joined. Not all contracts...

frontend/controller/actions/group.js
Silver-IT: This is for blocking `not-chatroom-members` to add another to the chatroom.

frontend/controller/actions/group.js
taoeffect: OK, and could you add a comment explaining all of that? To make it clear what scenario this section is for?

Silver-IT: Sure.

frontend/controller/actions/group.js
taoeffect: Then the error message should be a little clearer, maybe something like this:

```js
throw new GIErrorUIRuntimeError(L('Only channel members can invite others to join.'))
```

Silver-IT: Okay.

frontend/controller/actions/group.js
```diff
@@ -224,6 +241,13 @@ export default (sbp('sbp/selectors/register', {
     sbp('state/vuex/commit', 'setCurrentGroupId', groupId)
   },
   'gi.actions/group/addChatRoom': async function (params: GIActionParams) {
+    const contractState = sbp('state/vuex/state')[params.contractID]
+    for (const contractId in contractState.chatRooms) {
+      if (params.data.attributes.name.toUpperCase() === contractState.chatRooms[contractId].name.toUpperCase()) {
+        throw new GIErrorUIRuntimeError(L('Duplicated channel name'))
```
taoeffect: `'Duplicated channel name'` => `'Duplicate chatroom name'`

Silver-IT: What about in the UI? In the dialog to create new channel?

taoeffect: *Yes, sorry, meant `channel`

test/cypress/integration/group-proposals.spec.js
```diff
@@ -286,7 +277,7 @@ describe('Proposals - Add members', () => {
     })
   })
 
-  it(`proposal-based invitation link has ${groupInviteLinkExpiry.proposal} days of expiry`, () => {
+  it(`invitee can not join the group with the expired invitation link`, () => {
```
taoeffect: This may give a linter error as backticks should only be used when there's string interpolation. (e.g. ` needs to be changed to ')

Silver-IT: Done.

frontend/controller/actions/group.js
```diff
@@ -185,26 +190,44 @@ export default (sbp('sbp/selectors/register', {
...
+         * but he should sync all the contracts he was syncing in the previous device
+         */
+        const rootState = sbp('state/vuex/state')
+        const me = rootState.loggedIn.username
+        const chatRoomIds = Object.keys(rootState[params.contractID].chatRooms)
+          .filter(cId => rootState[params.contractID].chatRooms[cId].users.includes(me))
+
+        for (const cId of chatRoomIds) {
+          await sbp('chelonia/contract/sync', cId)
+        }
```
taoeffect, line 226: Two minor change requests here, that you can make in your other PR after this is merged:

1. Since `const rootState = ...` is used in both if/else clauses, DRY it and move it to the top of the function so that it's written only once.
2. Chelonia is able to take an array of contractIDs, so it would be better to remove the `for` loop and simply pass in `chatRoomIds`

taoeffect (approved):
Awesome job @Silver-IT! 🎉

PR #1301, opened 2022-06-17
bug fix: translation fails when locale is not en-US

SebinSong:
i18n.vue component fails to translate a string when the localization settings on the browser is different than en-US.
when there is no translation-table fetched from the back-end, it needs to resolve to the default table.

image


taoeffect (approved):
Approved! Nice job @SebinSong, thanks for the quickfix!

Issue #1298, opened 2022-06-07
Find out why settings are being saved twice


taoeffect:
### Problem

I've noticed that now for some reason the settings are being saved twice whenever contract changes are detected:

```
[sbp] state/vuex/save Array []
[sbp] 'gi.db/settings/save' u1
[sbp] state/vuex/save Array []
[sbp] 'gi.db/settings/save' u1
```

### Solution

Find out the problem and fix it so that they're only saved once.

The first place to look will be in `state.js`, see `debouncedSave`.

taoeffect:
Closed by @SebinSong in #1307

PR #1307, opened 2022-06-19
PR for issue #1294 and #1298

### LINKED ISSUE #1294: Set default threshold of percentage rule to 67%
### Problem

75% is a very high threshold to reach in groups whose users aren't very active.

See also related issue: #1179

### Solution

Set the default threshold to 67%.

SebinSong:
closes #1294 and #1298

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -515,7 +515,10 @@ const store: any = new Vuex.Store({
 
 // save the state each time it's modified, but debounce it to avoid saving too frequently
 const debouncedSave = _.debounce(() => sbp('state/vuex/save'), 500)
-store.subscribe(debouncedSave) // for e.g saving notifications that are markedAsRead
+store.subscribe((commit) => {
+  if (commit.type !== 'noop')
+    debouncedSave()
+}) // for e.g saving notifications that are markedAsRead
 // since Chelonia updates do not pass through calls to 'commit', also save upon EVENT_HANDLED
 sbp('okTurtles.events/on', EVENT_HANDLED, debouncedSave)
 // logout will call 'state/vuex/save', so we clear any debounced calls to it before it gets run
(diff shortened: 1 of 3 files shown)


frontend/model/state.js
```diff
@@ -515,7 +515,11 @@ const store: any = new Vuex.Store({
 
 // save the state each time it's modified, but debounce it to avoid saving too frequently
 const debouncedSave = _.debounce(() => sbp('state/vuex/save'), 500)
-store.subscribe(debouncedSave) // for e.g saving notifications that are markedAsRead
+store.subscribe((commit) => {
+  if (commit.type !== 'noop') {
```
SebinSong, line 519: (Pls correct me if I'm wrong)
['noop'](https://github.com/okTurtles/group-income/blob/master/frontend/model/state.js#L132) mutation triggered [here](https://github.com/okTurtles/group-income/blob/master/frontend/model/state.js#L529) in 'EVENT_HANDLED' handler apparently is some sort of a dummy(?) action created to help with the possible discrepancy in Vuex devtool state.

as this mutation doesn't actually perform any state change(it's an empty function), the Vuex state doesn't need to be cached in the local-storage in response to it.

frontend/utils/lazyLoadedView.js
```diff
@@ -8,7 +8,7 @@ import LoadingPage from '@views/containers/loading-error/LoadingPage.vue'
 
 type LazyImport = () => Promise<Object>;
 
-// See https://v2.vuejs.org/v2/guide/components-dynamic-lazy.html#Async-Components
+// See https://v2.vuejs.org/v2/guide/components-dynamic-async.html#Async-Components
```
taoeffect, line 11: Thank you for this link fix!

taoeffect (approved):
Perfect PR @SebinSong! First commit! 🥇 

Approved!

PR #1308, opened 2022-06-20
Contract versioning

### LINKED ISSUE #291: Create set-in-stone.js test files
### Problem

We can't have different versions of the software building up a different state from the log.

### Solution

Every release we must "set in stone" the latest behavior of all of the events/actions/contracts/states. Create numbered test files that build up a state using all of the available events, and verify the correct expected state is reached. Hash these test files and make sure they never change again. Create a new test file for each release that adds any new actions/contracts. This also means that the mutations must never change.

### LINKED ISSUE #475: Make contracts specify how state is built
### Problem

Currently we build up the state with code that's outside of the contract (based on data that's in the chain), this can impact portability.

### Solution

I'm keeping my research notes here: https://gitlab.okturtles.org/okturtles/group-income-simple/issues/41

### LINKED ISSUE #1214: Figure out versioning rules for upgrading / updating contracts and document them

taoeffect:
Group Income now keeps track of different versions of the contracts so that each GIMessage is processed using the version of the contract that generated it. This ensures consistent state among clients even as contracts are modified and upgraded.

### Changes

- Contracts are now loaded dynamically at runtime, and each `GIMessage` will specify the version of the contracts that were used to create it. GI will automatically load and use the correct version of the contract to process the message.
- Group Income is now very much versioned, meaning, any changes to contracts must result in a version bump using the `grunt pin` command. For example, if you've submitted a PR that modifies contracts and the current version is `0.0.1` as specified in `package.json`, then you need to run `grunt pin:0.0.2` (or whatever the new version number is) before submitting your PR. See section below on `grunt pin` for more details.
- `package.json` now has a new field called `"contractsVersion"` to keep track of the contracts version. This is separate from the `"version"` field because it's possible to update the frontend UI code and bump `"version"` without bumping the contracts version. This avoids generating unnecessary updates to the contracts.
- The experience of `grunt dev` has been preserved but its underlying behavior has been significantly modified. Now whenever a contract is modified while `grunt dev` is running, it will save two versions of the contract to `dist/contracts/` (the full bundle and the slim version), "deploy" the new contract to the `data/` folder, modify the `frontend/main.js` file with the new manifest hashes for the new version, and refresh the browser. To the developer, modifications to contract behavior should still become immediately active upon page refresh (which should happen automatically thanks to `browser-sync`).
- Closes #1214
- Closes #475
- Closes #291
- Updated Dependencies: `grunt`, `browser-sync`, `esbuild`, `flow-remove-types`
- Added Dependencies: [`@chelonia/cli`](https://github.com/okTurtles/chel), `babel-plugin-module-resolver` (so that `mocha` tests can handle files containing `@common` imports, see new `scripts/mocha-helpers.js` file)
- New Directories: `contracts/`, `frontend/common/`, and `frontend/model/contracts/shared/` (see section below for details)
- Our esbuild `alias-plugin.js` file can now handle aliases that are also marked as `external` (e.g. they don't get resolved). This is useful when dealing with things like the `@common` alias, which in some builds will be marked external (e.g. the slim contract builds), and in others will be resolved.

### New Grunt command: `grunt pin`

If you work on the contracts themselves (i.e. you modify their behavior), you must become familiar with the new `grunt pin` command, because using it will be necessary for Travis tests to pass.

`grunt pin` will take contracts that were saved to `dist/contracts/` (this is where contracts are saved to while `grunt dev` is running) and copy them into `contracts/<version>/`. It will also update `frontend/main.js` and `test/backend.test.js` to configure Chelonia to use that version of the contract when it sends new messages, and it will update the project's `"contractsVersion"` field in `package.json`.

`grunt pin` will ensure that your changes to the contracts will not get lost when a fresh clone of the project is made. This is why it's necessary to use it in order for Travis tests to pass, as otherwise Travis will not know about the changes you've made to the contracts.

All versions of the contracts are now stored under a new `contracts/` folder in the root of the project. The contracts that are stored here will get "deployed" by freshly cloned projects. Travis tests will pass because of this. `grunt pin` saves the temporary contracts that were stored in `dist/contracts/` during development to this `contracts/` folder.

Each version of the contracts is stored in a subdirectory of that folder that represents its version. For example, version `0.0.1` is stored in `contracts/0.0.1`.

To indicate a new release of GI, you use the `grunt pin` command. When you are certain that your changes are finished, the tests are passing, and you're ready to submit a new PR, increase the version number by running `grunt pin:<version>`, for example: `grunt pin:0.0.2`.

NOTE: **once a PR is merged with that version, the files in that folder can never be updated ever again.**

For this reason, you cannot run the command `grunt pin` twice on the same version. If you need to do so before your PR is merged, you can, by simply deleting the version folder. For example, say you made one last change after running `grunt pin:0.0.2`, then you can run `rm -rf contracts/0.0.2` and `grunt pin:0.0.2` again.

Again: once your PR is merged, no one can ever pin that version again.

### New Directories: `contracts/`, `frontend/common/`, and `frontend/model/contracts/shared/`

As mentioned already, `contracts/` will contain all of the versions of the contracts that GI will ever have. This ensures that they get "deployed" to new servers automatically.

Since contracts are loaded at runtime, and potentially multiple versions of the same contract could be loaded, the new contract versioning system supports loading two types of contracts: full bundles and "slim" versions of contracts that import shared code from an external module.

We are using these slim versions of contracts with a new `frontend/common/` folder that contains a file called `common.js` that is used to load code shared by the frontend UI code, and the contracts. The major dependencies that are shared are `vue` and `dompurify` (via our translations code). The code that is imported by this file must never change in its behavior in a way that affects the state of contracts because it does not get included in the slim versions of the contracts.

Please see the comment at the top of the file `fronend/common/common.js` for more detail.

Similarly, there is now also a new `fronend/model/contracts/shared/` folder that represents code that is shared by contracts and frontend code. Unlike the code in `frontend/common/` this code *does* get included in the slim versions of the contracts, but it still should remain relatively unchanged, preferring additions instead of modifications to functionality.

See the `README.md` file in the `fronend/model/contracts/shared/` for more details.

### Two versions of each contract: slim & bundle

Contracts are now saved in two different versions, linked by a "manifest" file (generated using the `chel manifest` command).

The slim version of the contract exists thanks to the shared `@common` code described above, and it is what gets loaded in Group Income.

The full bundle version of the contract contains all of the code needed to parse events independently of Group Income. In the future you'll be able to run the contracts using the `chel` command from the terminal. This allows for many possibilities, for example, creating other applications that use the Group Income state.


Gruntfile.js
```diff
@@ -47,9 +50,60 @@ const applyPortShift = (env) => {
   return { ...env, API_PORT, API_URL }
 }
 
+function pick (o, props) {
+  const x = {}
```
corrideat: `Object.create(null)` is better because it avoids prototype pollution.

Gruntfile.js
```diff
@@ -47,9 +50,60 @@ const applyPortShift = (env) => {
...
+  const { stdout, stderr } = await execP(cmd)
+  if (stderr) {
+    console.error(chalk`{red ${errMsg}:}`, stderr)
+    throw new Error(errMsg)
+  }
+  return { stdout }
+}
+
+async function generateManifests (contractsDir, version) {
+  const { stdout } = await execWithErrMsg(`ls ${contractsDir}/*-slim.js | sed -En 's/.*\\/(.*)-slim.js/\\1/p' | xargs -I {} node_modules/.bin/chel manifest -v ${version} -s ${contractsDir}/{}-slim.js key.json ${contractsDir}/{}.js`, 'error generating manifests')
```
corrideat: Is it possible that this could be done with JavaScript? Not only is it potentially unsafe (command injection) but it also reduces the portability of the code (it requires a POSIX shell and POSIX commands, meaning it won't work, e.g., on Windows)

Gruntfile.js
```diff
@@ -178,7 +255,17 @@ module.exports = (grunt) => {
       }
     }
   }
-
+  const esbuildOtherContractOptions = {
+    // since two builds get triggered each time any contract file is modified
+    // and either build might get triggered first, we run the postoperation on
+    // each build, with triggerEveryN it gets called every other time, ensuring
+    // it's called once per build
+    postoperation: triggerEveryN(2, async ({ fileEventName, filePath }) => {
```
corrideat: This is an unsafe call to an `async` function, because if it causes an exception it will result in node terminating, and, because it's asynchronous, it might happen in an unpredictable order. This function (from the comment) also seems to rely a bit too much on the specifics of the build process (what if later there's more / fewer operations)? Perhaps something using `Promise.all` is a better approach?

contracts/0.0.1/chatroom-slim.js
```diff
@@ -0,0 +1,666 @@
+"use strict";
```
corrideat, line 1: What's up with this file? It it autogenerated (looks like it)? If so, does it have to be part of the repository (as opposed to being generated during build)? And if it's manually generated, can it use more modern syntax and avoid the redefinitions of `Object.*`?

contracts/0.0.1/chatroom.js
```diff
@@ -0,0 +1,9768 @@
+"use strict";
```
corrideat, line 1: Same as my other comment to `chatroom-slim.js`

contracts/0.0.1/group.js
```diff
@@ -0,0 +1,10681 @@
+"use strict";
```
corrideat: Same comment, I guess for all files in `contracts/*`

frontend/common/common.js
```diff
@@ -0,0 +1,29 @@
...
+// expects the behavior to be.
+
+// EVERYTHING EXPORTED BY THIS FILE MUST ALWAYS AND ONLY BE IMPORTED
+// VIA "/assets/js/common.js"!
+// DO NOT CHANGE THE BEHAVIOR OF ANYTHING IMPORTED BY THIS FILE THAT AFFECTS THE
+// BEHAVIOR OF CONTRACTS IN ANY MEANINGFUL WAY!
+// Doing otherwise defeats the purpose of this file and could lead to bugs and conflicts!
+// You may *add* behavior, but never modify or remove it.
+
+export { default as Vue } from 'vue'
```
corrideat: Does `vue` really need to be part of the contracts? (I can understand it's easier for now, but I'm making this comment in light of the comment above not to remove functionality in the future. Also, since Vue is likely to be updated in ways that might not be API-compatible, having Vue seems to defeat the purpose of this file in the long run)

frontend/main.js
```diff
@@ -132,10 +148,13 @@ async function startApp () {
   // NOTE: setting 'EXPOSE_SBP' in production will make it easier for users to generate contract
   //       actions that they shouldn't be generating, which can lead to bugs or trigger the automated
   //       ban system. Only enable it if you know what you're doing and don't mind the risk.
+  // IMMPORTANT: setting 'window.sbp' must come *after* 'chelonia/configure' so that the Cypress
```
corrideat: There' a typo in `IMPORTANT`

scripts/esbuild-plugins/utils.js
```diff
@@ -79,9 +79,10 @@ exports.createAliasReplacer = (aliases) => {
  * @returns {RegExp}
  */
 exports.createFilterRegExpFromAliases = (aliases) => {
-  if (aliases.some(alias => alias.includes('/'))) {
-    throw new Error('Path aliases may not include slash characters.')
-  }
+  // TODO: why was this added?
```
corrideat: I'm guessing probably to avoid path traversal attacks? Although in that case a whitelist of characters might be better. There's `\\` (double backslash) which is a standard path separator, and single backslash will work on Windows as well.

shared/domains/chelonia/chelonia.js
```diff
@@ -93,12 +108,23 @@ sbp('sbp/selectors/register', {
...
-  'chelonia/configure': function (config: Object) {
+  'chelonia/config': function () {
+    return cloneDeep(this.config)
+  },
+  'chelonia/configure': async function (config: Object) {
     merge(this.config, config)
     // merge will strip the hooks off of config.hooks when merging from the root of the object
     // because they are functions and cloneDeep doesn't clone functions
-    merge(this.config.hooks, config.hooks || {})
+    Object.assign(this.config.hooks, config.hooks || {})
```
corrideat: Like said in the other comment, it's preferable to use `Object.create(null)` for objects that may have user-supplied keys.

shared/domains/chelonia/chelonia.js
```diff
@@ -93,12 +108,23 @@ sbp('sbp/selectors/register', {
...
+    // using Object.assign here instead of merge to avoid stripping away imported modules
+    Object.assign(this.config.contracts.defaults, config.contracts.defaults || {})
```
corrideat: Ditto about `{}`

shared/domains/chelonia/internals.js
```diff
@@ -1,19 +1,86 @@
...
+    const source = await fetch(`${this.config.connectionURL}/file/${contractInfo.hash}`)
+      .then(handleFetchResult('text'))
+    const sourceHash = blake32Hash(source)
+    if (sourceHash !== contractInfo.hash) {
+      throw new Error(`bad hash ${sourceHash} for contract '${contractInfo.file}'! Should be: ${contractInfo.hash}`)
+    }
+    function reduceAllow (acc, v) { acc[v] = true; return acc }
+    const allowedSels = ['okTurtles.events/on', 'chelonia/defineContract']
+      .concat(this.config.contracts.defaults.allowedSelectors)
+      .reduce(reduceAllow, {})
```
corrideat: Ditto (`{}`)

shared/domains/chelonia/internals.js
```diff
@@ -1,19 +1,86 @@
...
+    const allowedDoms = this.config.contracts.defaults.allowedDomains
+      .reduce(reduceAllow, {})
```
corrideat: Ditto (`{}`)

shared/domains/chelonia/internals.js
```diff
@@ -1,19 +1,86 @@
...
+    `)()
+    this.defContractSBP = contractSBP
+    this.defContractManifest = manifestHash
+    saferEval((dep) => {
+      return dep === '@sbp/sbp'
+        ? contractSBP
+        : this.config.contracts.defaults.modules[dep]
+    }, {
+      fetch: null,
+      XMLHttpRequest: null,
```
corrideat: Why are we using `XMLHttpRequest`? It seems quite redundant when we're already using `fetch` and all modern browsers support it. I doubt the code otherwise works in browsers that _don't_ support `fetch`, so this seems like a good opportunity to remove it, maybe?

shared/domains/chelonia/internals.js
```diff
@@ -46,7 +113,7 @@ sbp('sbp/selectors/register', {
         // if this isn't OP_CONTRACT, get latestHash, recreate and resend message
         if (!entry.isFirstMessage()) {
           const previousHEAD = await sbp('chelonia/out/latestHash', contractID)
-          entry = GIMessage.createV1_0(contractID, previousHEAD, entry.op())
+          entry = GIMessage.createV1_0(contractID, previousHEAD, entry.op(), entry.message().manifest)
```
corrideat: Nothing to change / improve here, but this will conflict with some of my changes.

corrideat (commented):
Great PR @greg! I think this is a great improvement over what we have now. It was also quite a lot to review, which I think it all looks mostly good. The main comments are (1) that IMO the contracts code (which seems generated) should not be in source control unless there's a compelling argument for it (answering to myself, I guess the compelling argument could be that it's the easiest way to ensure the code doesn't change) and (2) it'd be preferable to have Vue out of the shared contract code; I can live with it being there for now (which is an improvement from the current situation anyhow) but I believe it should be seen as technical debt in that case.

corrideat:
> Great PR @greg! I think this is a great improvement over what we have now. It was also quite a lot to review, which I think it all looks mostly good. The main comments are (1) that IMO the contracts code (which seems generated) should not be in source control unless there's a compelling argument for it (answering to myself, I guess the compelling argument could be that it's the easiest way to ensure the code doesn't change) and (2) it'd be preferable to have Vue out of the shared contract code; I can live with it being there for now (which is an improvement from the current situation anyhow) but I believe it should be seen as technical debt in that case.

This goes to say, these two I mentioned are the main changes I'd say can be considered, but none of them are blocking if there's good reason for keeping things as they are. Then, there are some issues I mentioned about `{}` vs `Object.create(null)`, which is something that probably _should_ be done where possible, as it protects against prototype pollution.

shared/domains/chelonia/internals.js
taoeffect: We aren't using it, we are making sure the contract cannot use it. This is a lightweight sandbox that undefines global variables we don't want the contract to have access to.

Soon I will push a commit that changes this section to look more like this:

```js
    saferEval((dep) => {
      return dep === '@sbp/sbp'
        ? contractSBP
        : this.config.contracts.defaults.modules[dep]
    }, {
      // not the greatest of sandboxes but something is better than nothing
      window: undefined,
      document: undefined,
      globalThis: undefined,
      fetch: undefined,
      XMLHttpRequest: undefined,
      // preserve these however, they're useful
      setTimeout,
      clearTimeout,
      setInterval,
      clearInterval,
      // TODO: add other possible global overrides here.
      //       Although again, the contracts are signed so not that big of a deal
      ...this.config.contracts.defaults.exposedGlobals,
      sbp: contractSBP
    })
```

scripts/esbuild-plugins/utils.js
snowteamer: I didn't allow slashes in path aliases because otherwise it would be more difficult to implement correctly, in particular regarding ordering of the given alias map. I wanted to make sure the given ordering would not be significant.

Suppose we have the following alias entries:
```js
{
  '@views': './frontend/views',
  '@views/utils': './frontend/views/utils'
}
```
The current code would build an incorrect regex in this case, one that would first match `@views` rather than `@views/utils` upon encoutering a path like `@views/utils/misc.js`.

Fixing that kind of edge cases would require either enforcing correct key ordering in the alias map, or more complex code.

Gruntfile.js
taoeffect: I hope it's OK based to leave this here based on our convo on Slack. If you want me to add `Object.freeze(Object) && Object.freeze(Object.prototype)` etc lemme know

Gruntfile.js
taoeffect: Yes, although I don't buy the potentially unsafe argument, I do think it'd be useful to add better Windows support.

However, Windows devs should still be able to use this via WSL, right?

contracts/0.0.1/chatroom-slim.js
taoeffect: It's autogenerated by `esbuild`. Don't know about how to get it to remove those redefinitions.

Yes it must be part of the project, as otherwise new installations won't be able to process old messages (see PR description for more info)

shared/domains/chelonia/chelonia.js
taoeffect: Resolving, plz reply here if followup needed: https://github.com/okTurtles/group-income/pull/1308#discussion_r906884592

frontend/common/common.js
taoeffect: Yes, we need it for `Vue.set` in the contracts, and we also need it to *not* be included in the slim bundle, so it goes here.

We should be fine though with regards to future Vue.js updates. Vue2 (which is what we're using) has reached EOL, so there won't be anymore updates to that. If we switch to Vue3, and that doesn't use `Vue.set`, then we'll add an additional import here for that and export it as `Vue3`.

frontend/main.js
```diff
@@ -101,11 +101,30 @@ async function startApp () {
...
+          'gi.actions/chatroom/leave', 'gi.notifications/emit'
+        ],
+        allowedDomains: ['okTurtles.data', 'okTurtles.events', 'okTurtles.eventQueue'],
+        preferSlim: true
+      },
+      manifests: {
+        'gi.contracts/group': '21XWnNSHSo5Dov3RwKFwSJzMNpt3KmpJNaVN6qsacysUnmxgq9',
+        'gi.contracts/identity': '21XWnNVUKkh1dWq1RoAAJbkPnGXxqFUpjXrzb9yo7NeGEEToin',
+        'gi.contracts/mailbox': '21XWnNKoSm7e2vnGJvoFSRB7heFTeg24FwDCkb6JAazkzhpNdY',
+        'gi.contracts/chatroom': '21XWnNXJkm2bkc8Wu7TC38zB845tR28qRGCs2HXCeMEJgvFozi'
```
Silver-IT: In my opinion, `grunt pin` should generate a manifest JSON file in a specific format and main.js should read it.
Instead of updating main.js.

Silver-IT (changes requested):
Very much impressive PR. It's a big evolution in Chelonia.

frontend/main.js
taoeffect: That's an excellent suggestion @Silver-IT! Not sure why I didn't think to do that!

shared/domains/chelonia/internals.js
```diff
@@ -66,10 +67,21 @@ export default (sbp('sbp/selectors/register', {
         ? contractSBP
         : this.config.contracts.defaults.modules[dep]
     }, {
-      fetch: null,
-      XMLHttpRequest: null,
-      // TODO: add other possible global overrides here.
-      //       Although again, the contracts are signed so not that big of a deal
+      // TODO: not a real sandbox, use webworkers, SES (see bottom of file), or something else
```
corrideat: As discussed in Slack, this seems like a pointless exercise as it's possible to get `window` one way or the other.

Gruntfile.js
corrideat: The unsafe part comes from `contractsDir`, etc. The build system mostly should be trusted, but there's still the possibility of intentional or unintentional shell injection.

And yeah, one could use WSL, but at that point it's no longer Windows; perhaps Mingw32 would be an option though. Also, I think it'd be messy to have a POSIX version, a Windows version, a Plan 9 version, etc., hence why I think doing this with JavaScript (or eventually Python, Perl, etc.) might be better.

taoeffect: > The unsafe part comes from `contractsDir`, etc. The build system mostly should be trusted, but there's still the possibility of intentional or unintentional shell injection.

I would not consider that a possibility personally, as it's a constant string, not a user-supplied string, and yes the build system is trusted. That would be like saying our own code is an exploit.

> And yeah, one could use WSL, but at that point it's no longer Windows

As long as people can develop on Windows I think it's fine.

> Also, I think it'd be messy to have a POSIX version, a Windows version, a Plan 9 version, etc., hence why I think doing this with JavaScript (or eventually Python, Perl, etc.) might be better.

Alright, we'll have to agree to disagree here. I can see how implementing it in JS might improve cross-compatibility, but using Python or Perl definitely would be unnecessary complexity. I'm going to leave this as-is because as-is we support all 3 major operating systems. If someone ever tries to build GI on a different OS, we can revisit this.

corrideat: > As long as people can develop on Windows I think it's fine.

Arguably it's not native development; I see it more as a workaround. But it works. It might not work if one is using Visual Studio or something like that.

> but using Python or Perl definitely would be unnecessary complexity

Of course, hence my suggestion to use JS (since one already needs Node.js to build the project). However, Python or Perl are more portable than Bash when it comes to Windows.

Like I said in my general review comment, this is merely a suggestion and not a blocking issue by any means.

That said, if going the POSIX route, I suggest using `find` instead of `ls`. Something like:

`find ${contractsDir} -type f -name '*-slim.js' | sed -En 's/.*\\/(.*)-slim.js/\\1/p' | xargs -I {} node_modules/.bin/chel manifest -v ${version} -s ${contractsDir}/{}-slim.js key.json ${contractsDir}/{}.js`

corrideat: > I would not consider that a possibility personally, as it's a constant string, not a user-supplied string, and yes the build system is trusted. That would be like saying our own code is an exploit.

I agree in this case. The reason I made the comment (similarly to the `Object.create(null)` one) is because this is being called as a different function. While it is fine _in this case_, someone calling this function in the future might call in a way that's unsafe. I think we agree that since is the build system, the risk is lower here than in other parts.

Silver-IT (dismissed):
Awesome. I completely approve this PR.
Thanks @taoeffect for your great work.

Silver-IT (approved):
Triple awesome. It's a big evolution.

PR #1311, opened 2022-06-21
Group chat mentions

### LINKED ISSUE #1128: Enhance gi.notifications domain to make notifications more visible [$80 bounty]
### Problem

The fact that there's an open proposal isn't advertised very well, and this can lead to stagnant proposals and a worse user experience.

### Solution

Enhance the notifications API (see the `gi.notifications` domain) so that it:

- Add an unread indicator to the browser tab. Research the best way(s) to do this. Feel free to post ideas below.
- Post an OS notification via the browser (https://developer.mozilla.org/en-US/docs/Web/API/notification ?)
- Add a red dot to the group icon in the groups sidebar when you're in multiple groups to show that a group has an active proposal pending. This involves closing  #1129

### Bounty

### LINKED ISSUE #1129: Implement notifications red dot on group image in group sidebar
### Problem

When a user is in multiple groups it's important to get their attention for certain events like a new proposal (#1128).

### Solution

Implement the red dot [as designed on Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=4666%3A36263)

Screen Shot 2021-12-07 at 5 02 37 PM

### LINKED ISSUE #1231: Group Chat link's red dot never disappears [$20 bounty]
### Problem

In the navigation bar on the left side of the app, the Group Chat link has a small dot indicator that currently never disappears, regardless of whether I've got unread chat messages or not:

image

After looking a bit in `Navigation.vue` I've found there was a hard-coded `:badgeCount` Vue binding which could be related to the issue:
https://github.com/okTurtles/group-income/blob/9936357727ea5ea76d5290bf55cddad6672d3ce2/frontend/views/containers/navigation/Navigation.vue#L31-L32

### Solution

Find a way to reactively show/hide the dot according to unread chat messages. 

### Bounty

$20 bounty for a clean solution to this (paid in cryptocurrency).

### LINKED ISSUE #1317: Non-functional buttons in chat
### Problem

For certain events in the chat, like when someone joins, when you however it shows you can add a reaction or reply:

Screen Shot 2022-06-26 at 7 37 11 PM

However, clicking them does nothing.

### Solution

Either remove them for events that cannot be interacted with, or implement them.

Silver-IT:
### Summary of Changes

- Implement @ mentions in the group chat
- Make notifications depends on the mentions
- Fix joining chatroom error in a very special case
- Closes #1231
- Truncate the channel name when it is too long
- Closes #1317 
- Closes #1129 
- Closes #1128


taoeffect (changes requested):
Very nice progress @Silver-IT!

I'm going to test the behavior first, and then review the code. So some issues with the behavior:

**Darkmode issues**

List isn't readable:

Screen Shot 2022-06-26 at 3 52 49 PM

Should have specific darkmode styling that looks nice.

**[Tab] key selects "Create poll"**

When the pop appears and you press "tab", it should fully complete the current selection, add a space after the username, and close the popup. Instead, right now it selects "Create poll" and leaves the popup:

Screen Shot 2022-06-26 at 3 53 42 PM

**Missing blue pill surrounding @mention**

Both within the edit field and within the chat log it should look something like this:

Screen Shot 2022-06-26 at 4 01 25 PM

**Notification sounds + browser notification + notification dot + notification count**

When you're @mentioned, several things should happen with regards to notifications:

1. In the left sidebar, the red dot should appear on "Chat" for unread mentions:

   Screen Shot 2022-06-26 at 4 00 53 PM

2. A counter of unread mentions should appear in the chat channel name on the right (don't forget to test dark mode!):

   Screen Shot 2022-06-26 at 4 03 08 PM

3. The browser should play an @mention sound (you can search for "royalty free sound effects" to find an appropriate one. E.g. I searched and found a [web page](https://www.epidemicsound.com/sound-effects/) with some. You can also maybe use one of the ones from [Adium sound effects](https://www.adiumxtras.com/index.php?a=search&cat_id=3&sort=ranking&user_id=0&s=&start=300) - but check if there's any license /cc @dotmacro).

4. The browser should display a system notification that says, `#<channel name>: @<user> mentioned you.`

taoeffect (changes requested):
1-group works really well! Am really impressed, great job @Silver-IT!

I found two issues:

1. The Cypress tests sometimes break, [see this one in particular](https://dashboard.cypress.io/projects/q6whky/runs/1180/test-results/e3b82a4f-7cca-4799-8448-1a2d11163511)
2. Multi-group scenario (described below)

### Multi-group scenario

Here's what I did to produce this error:

Screen Shot 2022-06-30 at 6 22 08 PM

1. I first created u1 account, and had them create a group
2. I created u2 in a separate container tab and joined them to the group
3. I tested that @mentioning works
4. As u1, I created a second group and switched to it
5. As u2, back in the original group, I did *not* join the other group, but instead @mentioned `@u1` in the chat

That is when the error in the screenshot appeared.

frontend/main.js
```diff
@@ -266,8 +271,28 @@ async function startApp () {
           'times-circle'
         )
       }
+
+      this.setBadgeOnTab()
+
+      this.$el.addEventListener('mousemove', this.requestPermissions, false)
```
taoeffect: We should request notification permissions not when they load the app, but when the user has entered a group (e.g. when the Group Dashboard becomes visible). This way they can get some context for why the permissions are necessary, and they will be already somewhat invested in receiving them - that way they are less likely to reflexively say "No" when they don't mean to.

Silver-IT: Make sense.

backend/database.js
```diff
@@ -108,12 +108,10 @@ export default (sbp('sbp/selectors/register', {
             offset--
           }
 
-          const previousHEAD = entry.message().previousHEAD
-          if (!previousHEAD || (isMet && !offset)) {
+          currentHEAD = entry.message().previousHEAD
+          if (!currentHEAD || (isMet && !offset)) {
```
taoeffect, line 112: Nice improvement, good catch 👍

frontend/controller/actions/group.js
```diff
@@ -220,18 +219,23 @@ export default (sbp('sbp/selectors/register', {
...
         sbp('state/vuex/commit', 'setCurrentChatRoomId', {
           groupId: params.contractID,
           chatRoomId: rootState[params.contractID].generalChatRoomId
         })
       }
       sbp('okTurtles.data/set', 'JOINING_GROUP', false)
-      sbp('okTurtles.data/set', 'READY_TO_JOIN_CHATROOM', false)
     } catch (e) {
       sbp('okTurtles.data/set', 'JOINING_GROUP', false)
-      sbp('okTurtles.data/set', 'READY_TO_JOIN_CHATROOM', false)
```
taoeffect, line 234: This should be added back, in case there was an exception on line 228.

Silver-IT: I don't understand this. Can you please explain more in detail?

taoeffect: Yeah, in the new file, on line 228 you have:

```js
        await sbp('chelonia/contract/sync', chatRoomIds)
```

If there's an exception there, then `sbp('okTurtles.data/set', 'READY_TO_JOIN_CHATROOM', true)` will not get reset because this line was removed.

Silver-IT: Got it. Thanks

frontend/model/contracts/chatroom.js
```diff
@@ -136,6 +141,80 @@ function emitMessageEvent ({ contractID, hash }: {
   sbp('okTurtles.events/emit', `${CHATROOM_MESSAGE_ACTION}-${contractID}`, { hash })
 }
 
+export function makeMentionFromUsername (username: string): {
+  me: string, all: string
+} {
+  return {
+    me: `@${username}`,
+    all: '@here'
```
taoeffect: `@here` is different from `@all`/`@group`.

The way Slack uses `@here` is to specifically mention everyone that is *online* at this specific moment.

Screen Shot 2022-07-07 at 3 18 51 PM

So there are two issues here:

1. `all` is being assigned `@here` - it should be consistent, since `all` != `here`, they mean two different things
2. `@here` should be renamed to `@channel` or `@all` if you are using it to mention everyone in the channel (regardless of whether or not they're online)

frontend/utils/notification.js
```diff
@@ -0,0 +1,26 @@
...
+
+export function makeNotification ({ title, body, icon }: {
+  title: string, body: string, icon: string
+}): void {
+  if (!checkNotification() || Notification.permission !== 'granted') {
+    return
+  }
+  const notification = new Notification(title, { body, icon })
+  // HACK: need to use variable (ERROR: assigned and never used)
+  console.log('Unread message', notification)
```
taoeffect: If you aren't using this variable, then don't create it. Just do `new Notification(title, { body, icon })` by itself.

Also, `console.debug` should be used instead of `console.log` for debugging. But in this case the line logging should just be removed.

frontend/utils/notification.js
```diff
@@ -0,0 +1,26 @@
+export function checkNotification (): void {
+  return !!window.Notification
+}
```
taoeffect: This function doesn't seem to be used anywhere else, so it shouldn't be exported.

frontend/utils/notification.js
```diff
@@ -0,0 +1,26 @@
...
+
+export async function requestNotificationPermission (): Promise<null | string> {
+  if (!checkNotification() || Notification.permission !== 'default') {
+    return null
+  }
```
taoeffect: The behavior we want is:

1. If the user blocked notifications on the dashboard, then don't ask again
2. If they blocked notifications on the dashboard, and explicitly request notifications again via the preferences, then ask again

frontend/utils/lazyLoadedView.js
```diff
@@ -91,6 +91,7 @@ lazyComponent('PaymentDetail', () => import('../views/containers/payments/Paymen
 lazyComponent('RecordPayment', () => import('../views/containers/payments/RecordPayment.vue'))
 
 lazyComponent('Appearence', () => import('../views/containers/user-settings/Appearence.vue'))
+lazyComponent('NotificationSettings', () => import('../views/containers/user-settings/Notification.vue'))
```
taoeffect: The component names and file names should always match.

frontend/views/containers/user-settings/Notification.vue
```diff
@@ -0,0 +1,30 @@
...
+      i18n.link(
+        tag='button'
+        @click.prevent='requestPermission'
+      ) Request Permission
+</template>
+
+<script>
+import { requestNotificationPermission } from '~/frontend/utils/notification.js'
+export default ({
+  name: 'NotificationSettings',
```
taoeffect: This file should be renamed to `NotificationSettings.vue`

frontend/model/contracts/chatroom.js
```diff
@@ -136,6 +141,80 @@ function emitMessageEvent ({ contractID, hash }: {
...
+  // }
+
+  const groupID = rootGetters.groupIdFromChatRoomId(contractID)
+  makeNotification({
+    title: `# ${chatRoomName}`,
+    body: text,
+    icon: rootGetters.globalProfile2(groupID, username).picture
+  })
+
+  sbp('okTurtles.events/emit', MESSAGE_RECEIVE, {})
```
taoeffect: It is not necessary to pass in `{}` if that parameter isn't used. Can be removed.

frontend/model/contracts/chatroom.js
Silver-IT: Nice Catch.

frontend/utils/notification.js
Silver-IT: Yeah. I tried to remove `console` and also `new Notification(title, { body, icon })`.
But there was an warning regarding `new Notification(title, { body, icon })` too.
I don't remember the name but similar to the error `Created and never used`.
I will try to find a solution...

frontend/views/containers/user-settings/Notification.vue
Silver-IT: Okay.

frontend/model/state.js
```diff
@@ -140,11 +144,45 @@ const mutations = {
...
   },
+  addChatRoomUnreadMentioning (state, { chatRoomId, messageId, createdDate }) {
+    const prevUnread = state.chatRoomUnread[chatRoomId]
+    if (!prevUnread) {
+      return
+    }
+    Vue.set(state.chatRoomUnread, chatRoomId, {
+      ...prevUnread,
+      mentionings: [...prevUnread.mentionings, { messageId, createdDate }]
+    })
```
taoeffect: Does this really need to be written this way?

Won't work just the same if you do it like this?

```js
prevUnread.mentionings.push({ messageId, createdDate })
```

frontend/model/state.js
```diff
@@ -140,11 +144,45 @@ const mutations = {
...
+  },
+  deleteChatRoomUnreadMentioning (state, { chatRoomId, messageId }) {
+    const prevUnread = state.chatRoomUnread[chatRoomId]
+    if (!prevUnread) {
+      return
+    }
+    Vue.set(state.chatRoomUnread, chatRoomId, {
+      ...prevUnread,
+      mentionings: prevUnread.mentionings.filter(m => m.messageId !== messageId)
+    })
```
taoeffect: Same question here...

```js
prevUnread.mentionings = prevUnread.mentionings.filter(m => m.messageId !== messageId)
```

frontend/views/components/sounds/Background.vue
```diff
@@ -0,0 +1,29 @@
+<template lang='pug'>
+.c-bgsound-wrapper
+  audio(ref='msgReceive' src='/assets/audio/msg-received.mp3' type='audio/mpeg')
+  audio(ref='msgSend' src='/assets/audio/msg-sent.mp3' type='audio/mpeg')
```
taoeffect, line 4: Nice job converting them to .mp3's 👍

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -100,6 +109,58 @@ export default ({
...
+                }
+              }).flat()
+            }
+          }
+        }).flat()
+
+        return splitByMentionings(objTexts, mentionings)
+      }
+
+      return splitByMentionings([{ type: textObjectType.Text, text: this.text }], ableMentionings)
```
taoeffect: This code seems really convoluted and complicated. Is there are simpler solution?

One possibly simpler method would be to use a regex to split the text:

```
> text = '@bob hello @bob, is your email bob@example.com? my friend @chris wants to chat with you @bob.'
'@bob hello @bob, is your email bob@example.com? my friend @chris wants to chat with you @bob.'
> text.split(/((?:^|\s)@[^\s]+\b)/)
[
  '',
  '@bob',
  ' hello',
  ' @bob',
  ', is your email bob@example.com? my friend',
  ' @chris',
  ' wants to chat with you',
  ' @bob',
  '.'
]
```

Then you can check if the array elements are proper @mentions.

taoeffect: @Silver-IT Even better Regex example:

```
> allMentions = ['@bob', '@chris', '@all']
[ '@bob', '@chris', '@all' ]
> text = '@bob hello @bob, is your email bob@example.com? my friend @chris wants to chat with you @bob. @bob? are you there? what do you think of @bob@foo.com?'
'@bob hello @bob, is your email bob@example.com? my friend @chris wants to chat with you @bob. @bob? are you there? what do you think of @bob@foo.com?'
> text.split(new RegExp(`(${allMentions.join('|')})`))
[
  '',
  '@bob',
  ' hello ',
  '@bob',
  ', is your email bob@example.com? my friend ',
  '@chris',
  ' wants to chat with you ',
  '@bob',
  '. ',
  '@bob',
  '? are you there? what do you think of ',
  '@bob',
  '@foo.com?'
]
```

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -244,6 +365,21 @@ export default ({
       this.$refs.textarea.value = this.$refs.textarea.value + emoticon.native
       this.closeEmoticon()
       this.updateTextWithLines()
+    },
+    startMentioning (keyword, position) {
+      this.ephemeral.mentioning.options = this.users.concat([{
+        // TODO: use group picture here or broadcast icon
+        username: 'here', displayName: 'here', picture: ''
```
taoeffect: Don't forget this also needs to be updated

taoeffect: And maybe it would be better if the possibilities were all retrieved from the same place to DRY things, so that you don't have to update multiple sections of code if things are changed

frontend/views/containers/user-settings/Appearence.vue
```diff
@@ -33,6 +33,20 @@
...
             @change='handleReducedMotion'
           )
+
+      i18n.is-title-2.c-title(tag='h2') Mute chat
+      .c-subcontent
+        .c-text-content
+          i18n.c-smaller-title(tag='h3') Blablablabla
+          i18n.help(tag='p') Blablabla blablabla blablabla blablabla blablabla
+        label
+          i18n.sr-only Blablablabla
```
taoeffect: What's going on here...?

taoeffect (changes requested):
Great job @Silver-IT! See comments.

frontend/utils/notification.js
taoeffect: Where was the warning? If it was from eslint, you can use `// eslint-disable-line` or `// eslint-disable-next-line`

If it was from Flow you can use `// $FlowFixMe`.

Just search the code for the above examples if unsure.

frontend/utils/notification.js
Silver-IT: Nice catch.

frontend/utils/notification.js
Silver-IT:
```
export async function requestNotificationPermission (force = false): Promise<null | string> {
  if (!checkNotification() || (!force && Notification.permission !== 'default')) {
    return null
  }

  try {
    return await Notification.requestPermission()
  } catch (e) {
    return null
  }
}
```

This could be the solution. Right?

frontend/views/containers/user-settings/Appearence.vue
Silver-IT: It's just saving the info to mute chat or not. Just save it in state but not implemented function yet.

taoeffect: OK, we cannot have "blahblahblah" going into `master`, that is not professional :P

Silver-IT: I will remove this. I think that's true also. That's because it's not the one from the design.

frontend/main.js
```diff
@@ -269,8 +273,26 @@ async function startApp () {
...
+        return Object.keys(this.ourUnreadMessages)
+          .map(cId => this.ourUnreadMessages[cId].mentionings.length)
+          .reduce((a, b) => a + b, 0)
+      },
+      ourNotificationsCount () {
+        return Object.keys(this.contracts)
+          .filter(cId => this.contracts[cId].type === 'gi.contracts/group')
+          .map(gId => this.unreadGroupNotificationCountFor(gId))
+          .reduce((a, b) => a + b, 0)
+      },
```
snowteamer, line 286: This count doesn't seem to include private notifications. In case they must be counted too, then the `totalUnreadNotificationCount()` getter could be used.

frontend/main.js
```diff
@@ -269,8 +273,26 @@ async function startApp () {
           'times-circle'
         )
       }
+
+      this.setBadgeOnTab()
     },
     computed: {
+      ...mapGetters(['ourUnreadMessages', 'unreadGroupNotificationCountFor']),
+      ...mapState(['contracts']),
+      ourUnreadMessagesCount () {
```
snowteamer, line 282: I'm not sure about the plural `s` in this name and similar ones. I could be wrong, but I've been thinking that names ending with e.g. `-Count` or `-List` don't need it.

frontend/model/notifications/templates.js
```diff
@@ -99,6 +99,16 @@ export default ({
       scope: 'group'
     }
   },
+  MENTION_ADDED (data: { username: string, groupID: string, chatRoomId: string }) {
+    return {
+      avatarUsername: data.username,
+      body: L('You have unread mentionings.'),
```
snowteamer: `mentioning` sounds a bit unusual to me. I would have written `mentions` instead.
EDIT: I noticed other occurences of `mentioning` in the new code, namely in function names. I would favor simplifying them by writing `mention`.

frontend/model/notifications/templates.js
```diff
@@ -99,6 +99,16 @@ export default ({
...
+      icon: 'comment',
+      level: 'danger',
```
snowteamer: Isn't `'danger'` level too much alarming? On the other hand, maybe `'info'` wouldn't be alarming enough. Just asking.

frontend/model/state.js
```diff
@@ -473,6 +502,13 @@ const getters = {
       return identityState && identityState.attributes
     }
   },
+  globalProfile2 (state, getters) {
```
snowteamer, line 506: What's the difference between `globalProfile2()` and `globalProfile()` ? The code looks quite similar in both functions. Would it be possible to use better names and/or add provide some documentation in comments?

frontend/model/state.js
```diff
@@ -508,10 +560,11 @@ const getters = {
       const chatRoom = state[contractID]
       if (chatRoom && chatRoom.attributes &&
         chatRoom.users[state.loggedIn.username]) {
+        const unreadMentioningsCount = getters.chatRoomUnreadMentionings(contractID).length
```
snowteamer: I'd suggest `unreadMentionCount`. It's also shorter.

frontend/utils/notification.js
```diff
@@ -0,0 +1,27 @@
...
+
+export function makeNotification ({ title, body, icon }: {
+  title: string, body: string, icon: string
+}): void {
+  if (!checkNotification() || Notification.permission !== 'granted') {
+    return
+  }
+  // eslint-disable-next-line
+  new Notification(title, { body, icon })
+}
```
snowteamer: When coming across an import statement referencing this file earlier in the review, I thought it was related to usual notifications. But it appears to be about native notifications, so I think naming this file `nativeNotifications.js` would be a good idea.

frontend/utils/notification.js
```diff
@@ -0,0 +1,27 @@
+function checkNotification (): void {
+  return !!window.Notification
```
snowteamer: This function's return type appears to be `boolean` rather than `void`.

frontend/views/containers/chatroom/ConversationsList.vue
```diff
@@ -76,6 +80,14 @@ export default ({
     },
     openModal (modal, queries) {
       sbp('okTurtles.events/emit', OPEN_MODAL, modal, queries)
+    },
+    limitedUnreadCount (n) {
+      const nLimit = 99
+      if (n > nLimit) {
+        return `${nLimit}+`
```
snowteamer, line 87: Appending a `+` sign to the unread counter is a nice idea!

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -190,9 +243,76 @@ export default ({
         return this.createNewLine()
       }
     },
+    updateMentioningKeyword () {
+      let value = this.$refs.textarea.value.slice(0, this.$refs.textarea.selectionStart)
+      const lastIndex = value.lastIndexOf('@')
+      const regExWordStart = /(\s|\n)/g // Space of Enter
```
snowteamer: `\s` does not only match the space character, but also `\n` and other ASCII whitespace. So the `|\n` part is not necessary.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -126,6 +130,11 @@ export default ({
     },
     updateBannerVisibility () {
       this.ephemeral.hideBanner = localStorage.getItem(this.bannerStorageKey) === 'true'
+    },
+    async requestPermissions () {
+      await requestNotificationPermission()
```
snowteamer: If this call ever throwed for some reason (for example in development when modifying the associated code) the `mousemove` listener would not get removed, likely causing a loop.
This can be avoided with a try/finally block.

frontend/main.js
Silver-IT: Interesting. I think we should use plural `s` in this kind of cases.
What do you think, @taoeffect?

frontend/model/notifications/templates.js
Silver-IT: Yeah. Make sense.

frontend/model/notifications/templates.js
Silver-IT: I used `danger` for the red dot.
But as we are not adding notifications for the mentions, I will remove this template and also other parts to add and remove notification.

frontend/model/state.js
Silver-IT: `globalProfile` gets profile from username who is part of current group.
`globalProfile2` gets profile from username who is part of the group identified by it's group ID.

frontend/utils/notification.js
Silver-IT: Yeah. Correct. Nice catch.

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: Correct. https://www.w3schools.com/jsref/jsref_regexp_whitespace.asp

frontend/views/pages/GroupDashboard.vue
Silver-IT: Nice catch. Thanks. @snowteamer

Silver-IT: What do you think if we do `removeEventListener` in `beforeDestroy` lifecycle?

Silver-IT: But `requestNotificationPermission` function can not throw any errors.
I think try/catch/finally isn't needed here.

frontend/utils/notification.js
Silver-IT: Yeah. Just to remove confusion. That's good idea.

frontend/main.js
Silver-IT: Yeah. `totalUnreadNotificationCount` is the right solution.

frontend/main.js
taoeffect: I think it's true that you don't need the `s`, it would work without it just fine, but I also think it doesn't matter, that either way is fine.

frontend/model/contracts/chatroom.js
```diff
@@ -136,6 +141,66 @@ function emitMessageEvent ({ contractID, hash }: {
...
+  // if (!rootState.chatRoomUnread[contractID] ||
+  //   !rootState.chatRoomUnread[contractID].mentions.length) {
+  //   const rootGetters = sbp('state/vuex/getters')
+  //   const notification = rootGetters.currentGroupNotifications.find(n =>
+  //     n.type === 'MENTION_ADDED' && n.linkTo === `/group-chat/${contractID}`
+  //   )
+  //   if (notification) {
+  //     sbp('gi.notifications/remove', notification)
+  //   }
+  // }
```
snowteamer: Indeed notifications usually do not need to be manually removed. The so-called `applyStorageRules()` function ensures they just do not get persisted to local storage anymore when they are both read and too old or when there are too many of them.

frontend/model/notifications/selectors.js
```diff
@@ -38,5 +38,9 @@ sbp('sbp/selectors/register', {
 
   'gi.notifications/markAllAsRead' (groupID: string) {
     sbp('state/vuex/commit', keys.MARK_ALL_NOTIFICATIONS_AS_READ, groupID)
+  },
+
+  'gi.notifications/remove' (notification: Notification) {
+    sbp('state/vuex/commit', keys.REMOVE_NOTIFICATION, notification)
   }
```
snowteamer: This selector and its mutation no longer seem to be used anywhere.

frontend/model/state.js
snowteamer: Thanks for adding comments.

frontend/model/contracts/chatroom.js
```diff
@@ -136,6 +141,66 @@ function emitMessageEvent ({ contractID, hash }: {
...
+
+  const rootGetters = sbp('state/vuex/getters')
+  const groupID = rootGetters.groupIdFromChatRoomId(contractID)
+  makeNotification({
+    title: `# ${chatRoomName}`,
+    body: text,
+    icon: rootGetters.globalProfile2(groupID, username).picture
+  })
+
+  sbp('okTurtles.events/emit', MESSAGE_RECEIVE)
```
taoeffect, line 182: At the moment this is used by only one listener that doesn't care what the message is (it just plays a sound), and so it doesn't need any additional parameters. However, it's possible some other listener might be interested in knowing specifically which message was received. In that case it might be useful to pass in the `messageId` along with the `chatRoomId`.

frontend/model/contracts/chatroom.js
```diff
@@ -136,6 +141,66 @@ function emitMessageEvent ({ contractID, hash }: {
...
+  //   const rootGetters = sbp('state/vuex/getters')
+  //   const notification = rootGetters.currentGroupNotifications.find(n =>
+  //     n.type === 'MENTION_ADDED' && n.linkTo === `/group-chat/${contractID}`
+  //   )
+  //   if (notification) {
+  //     sbp('gi.notifications/remove', notification)
+  //   }
+  // }
+}
+
```
taoeffect: Since I don't think we'll ever have chatroom mentions appear under the bell notifications menu, it's probably safe to delete this.

frontend/model/contracts/chatroom.js
```diff
@@ -223,8 +288,17 @@ sbp('chelonia/defineContract', {
...
-      sideEffect ({ contractID, hash }) {
+      sideEffect ({ contractID, hash, meta }) {
         emitMessageEvent({ contractID, hash })
+
+        if (sbp('okTurtles.data/get', 'READY_TO_JOIN_CHATROOM') || // Join by himself or Login in another device
+          sbp('okTurtles.data/get', 'JOINING_CHATROOM_ID') === contractID) { // Be added by another
+          sbp('state/vuex/commit', 'setChatRoomUnreadSince', {
+            chatRoomId: contractID,
+            messageId: hash,
+            createdDate: new Date(meta.createdDate).getTime()
```
taoeffect: I am seeing a lot of repetition of this code: `new Date(meta.createdDate).getTime()`

That seems like a DRY violation. You could avoid that by passing in `meta.createdDate` directly, and then inside of `setChatRoomUnreadSince` doing the conversion.

However, why are you converting it to UNIX time anyway instead of leaving it as an ISOString?

frontend/model/contracts/chatroom.js
```diff
@@ -318,8 +415,8 @@ sbp('chelonia/defineContract', {
         }
       },
       sideEffect ({ meta, contractID }, { state }) {
-        if (!state.saveMessage && state.attributes.creator === meta.username) { // Not sure this condition is necessary
-          if (sbp('okTurtles.data/get', 'JOINING_CHATROOM')) {
+        if (state.attributes.creator === meta.username) { // Not sure this condition is necessary
```
taoeffect: Is this `// Not sure this condition is necessary` still true?

frontend/model/contracts/chatroom.js
```diff
@@ -242,8 +316,16 @@ sbp('chelonia/defineContract', {
...
+      sideEffect ({ contractID, hash, meta }) {
         emitMessageEvent({ contractID, hash })
+
+        if (sbp('okTurtles.data/get', 'READY_TO_JOIN_CHATROOM')) {
+          sbp('state/vuex/commit', 'setChatRoomUnreadSince', {
+            chatRoomId: contractID,
+            messageId: hash,
+            createdDate: new Date(meta.createdDate).getTime()
+          })
+        }
```
taoeffect: I have seen these 7 lines repeated verbatim in several places now. That's a DRY violation => convert to function and minimize repetition as much as possible.

frontend/utils/nativeNotification.js
```diff
@@ -0,0 +1,27 @@
+function checkNotification (): boolean {
+  return !!window.Notification
+}
+
+export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
+  if (!checkNotification() ||
+    (!force && Notification.permission !== 'default') ||
+    (force && Notification.permission !== 'granted')) {
```
taoeffect: There seem to be two problems here:

1. It's unclear to me what `force` means here (it should be added in a comment what it does), but I assume it's intended to be used to re-request permissions even if they were previously not granted. If that's the case, this logic is wrong, since if `force = true` and `Notification.permission !== 'granted'` we want to proceed.
2. I see nowhere else in the code where `force = true`. It seems this variable isn't used. And yet it should be used in the settings.

frontend/views/pages/GroupDashboard.vue
taoeffect: @Silver-IT But @snowteamer raises a good point - at some point in the future someone else could rewrite your code for `requestNotificationPermission` so that it does throw an exception. In that case this part of the app will break in a serious way.

taoeffect (changes requested):
Great job @Silver-IT! This PR is shaping up to be really solid. 💪 

Another review, I think we are nearing the finish line on this one.

frontend/model/contracts/chatroom.js
Silver-IT: Is that over-engineering now?  If it is needed, we can add them.
But we don't need it at this time, do we?

taoeffect: Eh, yeah we don't need it now. It's OK to leave it as-is if you'd like.

frontend/model/contracts/chatroom.js
Silver-IT: It's not necessary. We should remove that condition.
All the members should leave chatroom if the chatroom is deleted.

frontend/views/pages/GroupDashboard.vue
Silver-IT: Since the current `requestNotificationPermission` function can not throw any errors, should we leave this without any changes? I think I should add comment here but to be honest, I don't think I should use try/finally at the moment.

taoeffect: It is common/encouraged for all `await` statements to be wrapped in a `try`/`catch` somewhere above where they're called, so that uncaught exceptions do not occur. Again, if this is not wrapped in such a handler, then we open ourselves to forgetting that this exists at all, and a future change to `requestNotificationPermission` can potentially break the app in a serious way.

We should catch all exceptions where they might occur, even if right now you're confident that no exceptions shouldn't happen. It's part of our policy to catch all possible errors and handle them appropriately.

taoeffect: And in this case, if there's an exception, I believe @snowteamer's suggestion is to remove the event listener so that it doesn't get stuck in a loop. That sounds like a reasonable/wise course of action to me.

Silver-IT: Understand. Thanks, @taoeffect , @snowteamer.

frontend/utils/nativeNotification.js
Silver-IT: 1. It is necessary when user wants to request permission if they have previously denied. You are right. Nice catch.
2. We don't call this function with `force` parameter right now. But if we have design to request permission, we will use this parameter.

frontend/model/contracts/chatroom.js
Silver-IT: Nice catch. That's correct. Not sure why I didn't think this way.

frontend/utils/nativeNotification.js
taoeffect: > We don't call this function with force parameter right now. But if we have design to request permission, we will use this parameter.

Well we do have a design: that's how the current behavior of the Notifications settings should work. It should pass in `force: true` so that if the user rejected notifications initially, they can re-request them

Silver-IT: Yeah. Just checked. Design is ready and UI implemented is missing.

Silver-IT: I think we should use `force: true` parameter only in the `NotificationSettings` page.
Right now, we request permission when users do mousemove in dashboard.
What do you think, @taoeffect?

Silver-IT: So we should implement NotificationSettings page first, and call this function passing `force: true` parameter.

taoeffect: > I think we should use force: true parameter only in the NotificationSettings page.

Right, that's what I was saying, that the NotificationSettings should use this, to allow users to re-request permissions if they initially said no.

Silver-IT: Should I implement NotificationSettings page and call this function passing `force: true` in this PR?
Or just call function with `force: true` parameter from the current `NotificationSettings` page?

taoeffect: Yes, that's what I've been saying, lol. You already have the link in NotificationSettings that says `Request Permission`. You just need to pass in `true` for `force` there, and fix the logic error here

frontend/model/contracts/chatroom.js
Silver-IT: > However, why are you converting it to UNIX time anyway instead of leaving it as an ISOString?

That's why I wanted to use it to compare with operator `>` or `<`.

taoeffect: Ah, understood. Yeah that makes sense. I'm curious, where are you doing that comparison? If it's only in one place then you could do the conversion there and leave it as an ISOString elsewhere...

test/cypress/integration/notifications.js
```diff
@@ -113,7 +113,7 @@ const cyCheckDashboardsBadge = (expectedCount) => {
 
 const cyCheckDreamersBadge = (expectedCount) => {
   if (expectedCount) {
-    cy.getByDT(`groupBadge-${dreamersGroupName}`).should('have.text', expectedCount)
+    cy.getByDT(`groupBadge-${dreamersGroupName}`).should('exist')
```
snowteamer, line 116: Why is it no longer feasible to check the badge text, to see whether it matches `expectedCount`?

frontend/model/contracts/chatroom.js
Silver-IT: That makes sense.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -94,13 +109,34 @@
...
+
+const caretKeyCodes = {
+  ArrowLeft: 37,
+  ArrowUp: 38,
+  ArrowRight: 39,
+  ArrowDown: 40,
+  Esc: 27,
+  End: 35,
+  Home: 36
+}
```
snowteamer, line 126: Should this also include `PageUp` and `PageDown`?

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -190,9 +243,76 @@ export default ({
...
+        this.addSelectedMention(this.ephemeral.mention.index)
+        e.preventDefault()
+      }
+    },
     handleKeyup (e) {
       if (e.keyCode === 13) e.preventDefault()
       else this.updateTextArea()
+
+      if (!Object.values(caretKeyCodes).includes(e.keyCode) &&
+        !Object.values(functionalKeyCodes).includes(e.keyCode)) {
```
snowteamer: Same remark here as in `handleKeydown()`.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -190,9 +243,76 @@ export default ({
...
+        return this.endMention()
+      }
+      value = value.slice(lastIndex + 1)
+      if (regExWordStart.test(value)) {
+        return this.endMention()
+      }
+      this.startMention(value, lastIndex)
+    },
+    handleKeydown (e) {
+      if (Object.values(caretKeyCodes).includes(e.keyCode)) {
```
snowteamer: This code is going to be run many times when the user is typing. So I think it would be nice to have a check without intermediary array creation here, to avoid garbage-collection of hundreds of small arrays when typing long messages.

Well maybe the engine is smart enough to not re-compute `Object.values(caretKeyCodes)` every time, but I'm not sure.

So in case it doesn't, I would try a switch statement or something like `if (caretKeyCodeValues[e.keyCode]))`, where `const caretKeyCodeValues = Object.values(caretKeyCodes)` would be defined next to `caretKeyCodes`.

frontend/model/notifications/selectors.js
Silver-IT: Yeah. I will remove it.

test/cypress/integration/notifications.js
Silver-IT: We don't display count there. We display only red dot.

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: No. Slack works that way. When we type `@` in the textarea and press `PageUp` or `PageDown`, the popup doesn't disappear.

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: Same as above.

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: > Well maybe the engine is smart enough to not re-compute Object.values(caretKeyCodes) every time, but I'm not sure.

But I am sure. :grin: . Nice catch, @snowteamer

Silver-IT:
```
const caretKeyCodes = {
  ArrowLeft: 37,
  ArrowUp: 38,
  ArrowRight: 39,
  ArrowDown: 40,
  Esc: 27,
  End: 35,
  Home: 36
}
const caretKeyCodeValues = Object.fromEntries(Object.values(caretKeyCodes).map(v => [v, true]))

if (caretKeyCodeValues[e.keyCode]) {
  ...
}
```

snowteamer: Ok! 😄 My example code was wrong, sorry

Silver-IT: Always appreciated.

frontend/model/contracts/chatroom.js
```diff
@@ -546,7 +521,7 @@ sbp('chelonia/defineContract', {
         if (rootState.chatRoomUnread[contractID].since.messageId === data.id) {
           sbp('state/vuex/commit', 'deleteChatRoomUnreadSince', {
             chatRoomId: contractID,
-            deletedDate: new Date(meta.createdDate).getTime()
+            deletedDate: meta.createdDate
           })
         }
```
snowteamer, line 527: Nice DRYing up throughout this file thanks to `updateUnreadPosition()`

frontend/utils/nativeNotification.js
```diff
@@ -5,7 +5,7 @@ function checkNotification (): boolean {
 export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
   if (!checkNotification() ||
     (!force && Notification.permission !== 'default') ||
-    (force && Notification.permission !== 'granted')) {
+    (force && Notification.permission === 'granted')) {
```
snowteamer, line 8: I was unsure about this too, but assumed I likely just didn't understand correctly

taoeffect (approved):
Excellent work @Silver-IT! Approved!

PR #1313, opened 2022-06-24
Use in-memory array for application logs

### LINKED ISSUE #1273: DOMException: Failed to execute `setItem` on `Storage` [$50 bounty]
### Problem

I am not sure whenever I've seen this error before.
But now, I have this error when I scroll up the messages in `Group Chat` page and try to log many events inside localStorage in a short period of time.
The expression `in a short period of time` is just my guess because we don't get this error when I scroll up slowly and give app the enough time to log and do something I don't know.
However, we actually get this error when I scroll up very quickly and tries to load several messages very soon.
image
Also when I refresh the page, I don't get this error too. So I guess this is not the problem sth like `Out of memory` or `Limit on memory usage`. And I used the expression `in a short period of time`.

### Solution

First things first, need to debug in `captureLogs.js`.
Not sure but need to log events synchronously.

EDIT by @taoeffect: I recommend [@snowteamer's solution](https://github.com/okTurtles/group-income/issues/1273#issuecomment-1133240119): storing the events in an in-memory array, and periodically serializing that array as the value of a single localstorage key.

### Bounty

$50 bounty for a clean solution to this (paid in cryptocurrency).

snowteamer:
Closes #1273

### Summary of changes
- `debug`-level logging is now enabled by default.
- An internal fixed-length circular array is used to store logs in memory
- The logs are flushed to local storage when the `onbeforeunload` event fires, i.e. upon page refresh (assuming the event does trigger), or when the new `logging/flush` selector is called.
- Other strategies should be easy to implement if requested.
- The per-user config object for log storage has been simplified to only include `maxEntries`. Other fields might be added if other log flushing strategies turn out to be useful.


frontend/main.js
```diff
@@ -91,6 +91,7 @@ async function startApp () {
           errMsg: e.message || '?'
         })
       })
+      sbp('logging/flush')
```
taoeffect: This is somewhat of a surprising addition to a function that is focused on displaying errors. To someone unfamiliar with the codebase, they will wonder why it is here. Could you add a comment explaining here why this is also done here?

frontend/model/captureLogs.js
```diff
@@ -220,20 +164,41 @@ export function downloadLogs (elLink: Object): void {
...
-      localStorage.removeItem(key)
+export function getLogger (): Object {
+  if (!logger) {
+    logger = createLogger(config)
+    const previousEntries = JSON.parse(getItem('entries'))
+
+    // If `maxEntries` is changed in a release, this will discard oldest logs as necessary.
+    if (config.maxEntries < previousEntries?.length) {
+      previousEntries.splice(0, previousEntries.length - config.maxEntries)
+    }
```
taoeffect: Nice thinking, and very elegant design for the circular list. I almost thought this was a bug because it doesn't consider the offset, but then I realized it doesn't need to thanks to how you implemented `.toArray()`.

frontend/model/captureLogs.js
```diff
@@ -220,20 +164,41 @@ export function downloadLogs (elLink: Object): void {
...
+  appLogsFilter = filter
+  // NOTE: Find a way to capture logs without messing up with log file location.
+  // console.log() doesnt include stack trace, so when logged, we can't access
+  // where the log came from (file name), which} difficults debugging if needed.
+  for (const methodName of appLogsFilter) {
+    console[methodName] = logger[methodName]
+    if (console[methodName] !== logger[methodName]) {
+      throw new Error(`could not overwrite console method ${methodName}`)
+    }
+  }
```
taoeffect: If a checkbox for a log level is checked in the UI, that logging should appear in both the console and in the UI.

And if a checkbox for a log level is unchecked, then it shouldn't appear in either.

I think there's a bug here because while it enables logging, I don't think it disables logging.

frontend/model/captureLogs.js
```diff
@@ -220,20 +164,41 @@ export function downloadLogs (elLink: Object): void {
...
 }
+
+window.addEventListener('beforeunload', event => {
+  setItem('entries', getLogger().entries.toArray())
+})
+
+sbp('sbp/selectors/register', {
+  'logging/getLogger' () { return getLogger() },
```
taoeffect: Is this used anywhere? If not, can safely remove.

taoeffect (changes requested):
@snowteamer - Fantastic PR!

Really beautiful implementation for the logging system. Love your decision to use a circular list and the clean and simple implementation of `createCircularList`.

Think I identified one minor bug. Otherwise really solid PR.

frontend/main.js
snowteamer: I included this SBP call here in an attempt to follow the suggestion you gave me on Slack, in response to one of my questions:
https://okturtles.slack.com/archives/C0EH7P20Y/p1656092332479119

frontend/model/captureLogs.js
snowteamer: I'm glad you like it! 🙂

frontend/model/captureLogs.js
snowteamer: That's true. Unchecked log levels aren't elements of `appLogsFilter` and therefore the corresponding console methods currently don't get overwritten.

This can be fixed if we only consider `debug`, `error`, `info`, `log` and `warn`, but please note that the console object has yet other methods which still log something if called. Although we currently don't use any of them I think.

frontend/model/captureLogs.js
snowteamer: This could be used in `AppLogs.vue` in place of `getLogger()`. I have to admit that's an inconsistency that should be fixed.

frontend/model/captureLogs.js
taoeffect: > but please note that the console object has yet other methods which still log something if called. Although we currently don't use any of them I think.

Totally OK. We'll just have to avoid using those other ones.

frontend/main.js
taoeffect: @snowteamer yeah, I'm not saying that you shouldn't have. I'm saying that an explanatory comment should go along with it on the line above to help others understand why it's here

frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -42,7 +42,7 @@ import sbp from '@sbp/sbp'
 import { mapState, mapMutations } from 'vuex'
 import safeLinkTag from '@views/utils/safeLinkTag.js'
 import { CAPTURED_LOGS, SET_APP_LOGS_FILTER } from '@utils/events.js'
-import { downloadLogs, getLog } from '@model/captureLogs.js'
+import '@model/captureLogs.js'
```
taoeffect: @snowteamer think this should go into `main.js` since `AppLogs.vue` is loaded lazyily... Even if it works right now, it might stop working if esbuild changes its behavior. Having it in `main.js` should ensure main.js's usage of it won't break.

frontend/controller/actions/identity.js
```diff
@@ -5,7 +5,7 @@ import { GIErrorUIRuntimeError } from '@model/errors.js'
 import L, { LError } from '@view-utils/translations.js'
 import { imageUpload } from '@utils/image.js'
 import { pickWhere, difference } from '@utils/giLodash.js'
-import { captureLogsStart, captureLogsPause } from '~/frontend/model/captureLogs.js'
+import '~/frontend/model/captureLogs.js'
```
taoeffect: And probably don't need this here if we're using SBP & main.js has it

frontend/model/captureLogs.js
```diff
@@ -220,20 +170,41 @@ export function downloadLogs (elLink: Object): void {
...
+  // NOTE: Find a way to capture logs without messing up with log file location.
+  // console.log() doesnt include stack trace, so when logged, we can't access
+  // where the log came from (file name), which} difficults debugging if needed.
+  for (const level of loggingLevels) {
+    consoleCopy[level] = appLogsFilter.includes(level) ? logger[level] : noop
+  }
 }
+
+window.addEventListener('beforeunload', event => {
+  setItem('entries', getLogger().entries.toArray())
```
taoeffect: Should this call `'appLogs/flush'` to DRY things?

taoeffect (commented):
Left 3 comments

taoeffect (changes requested):
Oh, there are several flow errors when I ran `grunt dev`, plz have a look at those

frontend/model/captureLogs.js
snowteamer: Good point

taoeffect (approved):
Fantastic PR @snowteamer. Really great job on this one. 👍 💯

PR #1322, opened 2022-06-30
Bug fixes for #1318 and #1320

### LINKED ISSUE #1318: Distribution date banner should re-appear as month nears end
### Problem

The banner in `GroupDashboard.vue` currently stays hidden if the user dismisses it:

Screen Shot 2022-06-26 at 7 28 59 PM

However, even if their income details are set, it should re-appear 7 days before the end of the current distribution period.

### Solution

Update the logic to make sure it re-appears without breaking the ability to dismiss it once seen (or once income details are added).

### LINKED ISSUE #1320: Fix "copy" icon for proposal invite link
### Problem

See design related issue #1314:

Screen Shot 2022-06-24 at 11 49 06 AM

### Solution

Implement the [correct copy icon design](https://github.com/okTurtles/group-income/issues/1314#issuecomment-1167351019) for dark mode.

SebinSong:
closes #1318
closes #1320

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -17,6 +17,7 @@ import { addTimeToDate, dateToPeriodStamp, dateFromPeriodStamp, isPeriodStamp, c
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
 import { unadjustedDistribution, adjustedDistribution } from './distribution/distribution.js'
 import currencies, { saferFloat } from '~/frontend/views/utils/currencies.js'
+import { GROUP_PROFILE_UPDATE } from '~/frontend/utils/events.js'
 import L from '~/frontend/views/utils/translations.js'
 import { chatRoomAttributesType } from './chatroom.js'
 import { INVITE_INITIAL_CREATOR, INVITE_STATUS, PROFILE_STATUS, INVITE_EXPIRES_IN_DAYS } from './constants.js'
@@ -866,6 +867,8 @@ sbp('chelonia/defineContract', {
           // someone updated their income details, create a snapshot of the haveNeeds
           updateCurrentDistribution({ meta, state, getters })
         }
+
+        sbp('okTurtles.events/emit', GROUP_PROFILE_UPDATE)
       }
     },
     'gi.contracts/group/updateAllVotingRules': {
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -17,7 +17,6 @@ import { addTimeToDate, dateToPeriodStamp, dateFromPeriodStamp, isPeriodStamp, c
 import { vueFetchInitKV } from '~/frontend/views/utils/misc.js'
 import { unadjustedDistribution, adjustedDistribution } from './distribution/distribution.js'
 import currencies, { saferFloat } from '~/frontend/views/utils/currencies.js'
-import { GROUP_PROFILE_UPDATE } from '~/frontend/utils/events.js'
 import L from '~/frontend/views/utils/translations.js'
 import { chatRoomAttributesType } from './chatroom.js'
 import { INVITE_INITIAL_CREATOR, INVITE_STATUS, PROFILE_STATUS, INVITE_EXPIRES_IN_DAYS } from './constants.js'
@@ -867,8 +866,6 @@ sbp('chelonia/defineContract', {
           // someone updated their income details, create a snapshot of the haveNeeds
           updateCurrentDistribution({ meta, state, getters })
         }
-
-        sbp('okTurtles.events/emit', GROUP_PROFILE_UPDATE)
       }
     },
     'gi.contracts/group/updateAllVotingRules': {
--- a/frontend/views/components/LinkToCopy.vue
+++ b/frontend/views/components/LinkToCopy.vue
@@ -68,12 +68,24 @@ export default ({
         return
       }
 
+      const displayTooltip = () => {
+        this.ephemeral.isTooltipActive = true
+
+        setTimeout(() => {
+          this.ephemeral.isTooltipActive = false
+        }, 1500)
+      }
+
       this.$refs.input.select()
-      document.execCommand('copy')
-      this.ephemeral.isTooltipActive = true
-      setTimeout(() => {
-        this.ephemeral.isTooltipActive = false
-      }, 1500)
+      if (navigator.clipboard) {
+        navigator.clipboard.writeText(this.link).then(displayTooltip)
+      } else {
+        // document.execCommand is deprecated, so only use it as a fallback of Clipboard API.
+        // reference: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
+        document.execCommand('copy')
+
+        displayTooltip()
+      }
     }
   }
 }: Object)
--- a/frontend/views/pages/GroupDashboard.vue
+++ b/frontend/views/pages/GroupDashboard.vue
@@ -40,7 +40,7 @@ page(pageTestName='dashboard' pageTestHeaderName='groupName' v-if='groupSettings
 <script>
 import { mapGetters, mapState } from 'vuex'
 import sbp from '@sbp/sbp'
-import { OPEN_MODAL, GROUP_PROFILE_UPDATE } from '@utils/events.js'
+import { OPEN_MODAL, INCOME_DETAILS_UPDATE } from '@utils/events.js'
 import Page from '@components/Page.vue'
 import AddIncomeDetailsWidget from '@containers/contributions/AddIncomeDetailsWidget.vue'
 import StartInvitingWidget from '@containers/dashboard/StartInvitingWidget.vue'
@@ -58,11 +58,16 @@ import { addTimeToDate, DAYS_MILLIS, humanDate } from '~/frontend/utils/time.js'
 export default ({
   name: 'GroupDashboard',
   beforeMount () {
+    sbp('okTurtles.events/on', INCOME_DETAILS_UPDATE, this.closeBanner)
+
     if (!this.isCloseToDistributionTime) {
       localStorage.setItem(this.bannerStorageKey, false)
     }
 
-    this.checkIfBannerBeenDismissed()
+    this.updateBannerVisibility()
+  },
+  beforeDestroy () {
+    sbp('okTurtles.events/off', INCOME_DETAILS_UPDATE, this.closeBanner)
   },
   data () {
     return {
@@ -114,17 +119,12 @@ export default ({
     humanDate,
     handleIncomeClick (e) {
       sbp('okTurtles.events/emit', OPEN_MODAL, 'IncomeDetails')
-
-      // firstly, make sure the listener is attached to 'GROUP_PROFILE_UPDATE' more than twice.
-      // e.g. user dismiss the modal without actually 'updating the income-details' and then click on the banner again.
-      sbp('okTurtles.events/off', GROUP_PROFILE_UPDATE, this.closeBanner)
-      sbp('okTurtles.events/once', GROUP_PROFILE_UPDATE, this.closeBanner)
     },
     closeBanner () {
       localStorage.setItem(this.bannerStorageKey, true)
       this.ephemeral.hideBanner = true
     },
-    checkIfBannerBeenDismissed () {
+    updateBannerVisibility () {
       this.ephemeral.hideBanner = localStorage.getItem(this.bannerStorageKey) === 'true'
     }
   },
@@ -132,7 +132,7 @@ export default ({
     bannerStorageKey () {
       // if the user performs switching to another group while he/she is still in this page[GroupDashboard.vue],
       // check if the distribution-warning banner has been dismissed for the switched group again.
-      this.checkIfBannerBeenDismissed()
+      this.updateBannerVisibility()
     }
   },
   components: {
--- a/frontend/views/pages/GroupDashboard.vue
+++ b/frontend/views/pages/GroupDashboard.vue
@@ -59,9 +59,10 @@ export default ({
   name: 'GroupDashboard',
   beforeMount () {
     if (!this.isCloseToDistributionTime) {
-      localStorage.setItem('giHideDistributionWarning', false)
+      localStorage.setItem(this.bannerStorageKey, false)
     }
-    this.ephemeral.hideBanner = localStorage.getItem('giHideDistributionWarning')
+
+    this.checkIfBannerDismissed()
   },
   data () {
     return {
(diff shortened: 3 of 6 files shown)


frontend/views/components/LinkToCopy.vue
```diff
@@ -68,12 +68,24 @@ export default ({
...
-      this.ephemeral.isTooltipActive = true
-      setTimeout(() => {
-        this.ephemeral.isTooltipActive = false
-      }, 1500)
+      if (navigator.clipboard) {
+        navigator.clipboard.writeText(this.link).then(displayTooltip)
+      } else {
+        // document.execCommand is deprecated, so only use it as a fallback of Clipboard API.
+        // reference: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
+        document.execCommand('copy')
```
SebinSong, line 85: I was reading through this LinkCopy.vue component which I was the original author of, while fixing hte issue #1320 and then noticed that document.execCommand() method being used here is deprecated according to [the MDN doc](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand).
So thought I would go ahead and do a bit of boyscouting here that replaces it with Clipboard API.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -93,8 +94,14 @@ export default ({
...
-      const warningDate = addTimeToDate(new Date(this.groupSettings.distributionDate), -7 * DAYS_MILLIS)
-      return Date.now() >= new Date(warningDate).getTime()
+      const dDay = new Date(this.groupSettings.distributionDate)
+      const warningDate = addTimeToDate(dDay, -7 * DAYS_MILLIS)
+
+      // when (D-day - 7d) <= today < D-day
+      return Date.now() >= new Date(warningDate).getTime() && Date.now() < dDay.getTime()
+    },
+    bannerStorageKey () {
+      return `giHideDistributionWarning.${this.currentGroupId}`
```
SebinSong, line 109: the reason for this change:
**'giHideDistributionWarning'** key to be used in local-storage here needs to be specific to a group. 
For example, a single user can create multiple groups that have different distribution dates, and we don't want those different groups to share the same flag.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -109,8 +116,18 @@ export default ({
       sbp('okTurtles.events/emit', OPEN_MODAL, 'IncomeDetails')
     },
     closeBanner () {
-      localStorage.setItem('giHideDistributionWarning', true)
+      localStorage.setItem(this.bannerStorageKey, true)
       this.ephemeral.hideBanner = true
+    },
+    checkIfBannerDismissed () {
+      this.ephemeral.hideBanner = localStorage.getItem(this.bannerStorageKey) === 'true'
```
SebinSong, line 128: a boolean value is converted to a string type when stored in local-storage.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -93,8 +94,14 @@ export default ({
       return Object.keys(this.currentGroupState.proposals).length > 0
     },
     isCloseToDistributionTime () {
-      const warningDate = addTimeToDate(new Date(this.groupSettings.distributionDate), -7 * DAYS_MILLIS)
-      return Date.now() >= new Date(warningDate).getTime()
+      const dDay = new Date(this.groupSettings.distributionDate)
+      const warningDate = addTimeToDate(dDay, -7 * DAYS_MILLIS)
+
+      // when (D-day - 7d) <= today < D-day
+      return Date.now() >= new Date(warningDate).getTime() && Date.now() < dDay.getTime()
```
SebinSong, line 106: This (making this computed-prop be false when **today > previous distribution date**)
 is actual fix for the issue #1318

frontend/views/pages/GroupDashboard.vue
taoeffect: @SebinSong Excellent catch 👍

frontend/views/pages/GroupDashboard.vue
```diff
@@ -93,8 +94,14 @@ export default ({
...
+    },
+    bannerStorageKey () {
+      return `giHideDistributionWarning.${this.currentGroupId}`
     },
     showBanner () {
       return !this.hasIncomeDetails && this.isCloseToDistributionTime && !this.ephemeral.hideBanner
```
taoeffect: This condition hasn't changed, and I think it's still broken, even with the new fix & improvement to `isCloseToDistributionTime`.

The issue being - the second the user adds their income details, the banner will never re-appear, because this is a chain of `&&` conditions, and `!this.hasIncomeDetails` will return `false`.

What we want is - even if their income details are set, we want the banner to still appear 7 days before the next distribution date. Changing the income details *within that 7 day period* (while the banner is visible) should dismiss the banner. So you need to be able to detect the change to the income details to get everything to work properly.

How do you do this?

Well, there are multiple ways this can be done:

- Inside of the modal for setting the income details, when "OK" is pressed (or whatever the confirmation button is), the modal can emit an event using `okTurtles.events/emit` that we can listen for inside of here, and then dismiss the banner using `closeBanner()`. This is probably the easiest way to do it as it doesn't involve touching any of the contract related code.
- We could also listen for messages to be processed and check if the message that was processed was an update to the income details data. This method is much more complicated and I don't recommend it because of that, but it would involve listening for the `EVENT_HANDLED` event and then checking the action and data of the `GIMessage` passed in to see if it was a message to update the income details for our user. Again, I don't recommend this method, it's too complicated, but I wanted to mention it just so that you're aware that such a thing is possible.

taoeffect (changes requested):
Fantastic PR @SebinSong, and great job on the boyscouting to improve the code! 😄 

Found one issue - I think some old code still needs updating. Left a comment.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -114,20 +114,25 @@ export default ({
     humanDate,
     handleIncomeClick (e) {
       sbp('okTurtles.events/emit', OPEN_MODAL, 'IncomeDetails')
+
+      // firstly, make sure the listener is attached to 'GROUP_PROFILE_UPDATE' more than twice.
```
SebinSong: a typo in this comment 'is attached to' -> is '**not**' attached to

SebinSong: @taoeffect 
Btw, I've got a small proposal here.

while working on updating this part, I felt it would've been great if there was a method in `@okTurtles/events` library that checks if a handler has already been attached to a particular event. something that looks like this.
```
if (!sbp('okTurtles.events/hasHandler', GROUP_PROFILE_UPDATE, this.closeBanner))
  sbp('okTurtles.events/once', GROUP_PROFILE_UPDATE, this.closeBanner)

```

I went ahead and tried to write those code myself that maybe can be added to [okTurtles.events/index.js ](https://github.com/okTurtles/okTurtles.events/blob/master/index.js)

```
'okTurtles.events/hasHandler': function (event, handler) {
    const handlerList = sbp('okTurtles.data/get', listenKey(event))

    return handlerList ? handlerList.includes(handler) : false
  }

```

Do you think it can be a useful addition to the library?

frontend/views/pages/GroupDashboard.vue
SebinSong: @taoeffect thanks for clarification on what is the needed change.
rather than emitting an event for 'submit button being pressed on the income-detail modal', I've created an event named GROUP_PROFILE_UPDATE (and one-time listener to it) which is emitted once the income-detail update gone successful.
Do you think it's a better approach?

taoeffect: @SebinSong 

> Do you think it's a better approach?

It's actually not a bad idea. You're thinking in the right direction, there are some issues with how you implemented that I'll comment in-line about elsewhere, but the general idea is sound.

The thing is, clicking "OK" in the Income Details modal isn't near the code that actually makes the update, and for that reason, emitting a notification there is a bit sketchy since the actual update to the income details might happen a few milliseconds later. EDIT: however, it would be simpler to do this in this instance! So it still might be the better way to go.

However, if such a notification is made in the contract, it needs to be done in a sideEffect. I'll explain elsewhere in more detail why this is.

frontend/model/contracts/group.js
```diff
@@ -866,6 +867,8 @@ sbp('chelonia/defineContract', {
           // someone updated their income details, create a snapshot of the haveNeeds
           updateCurrentDistribution({ meta, state, getters })
         }
+
+        sbp('okTurtles.events/emit', GROUP_PROFILE_UPDATE)
```
taoeffect: (We will explain all of the following in our documentation for the framework once it's released.)

### First issue - use `sideEffect`

Contracts take events (GIMessages) and build up a state out of them. To make sure they create the same state and behave consistently across devices we separate application code and contract code.

Anytime a contract interacts with application code - it must do so only in the `sideEffect` function. There are two ways to do this:

1. Add the sideEffect function for this action (search this file for `sideEffect` for examples).
2. Use `'gi.contracts/group/pushSideEffect'` inside of the process function.

It is usually simpler to implement the `sideEffect` function.

Inside of that function you can do whatever you want. You can access application data, call application functions, and emit events that might be listened to by application code (like you're doing there).

### Second issue - be specific

Right now it's simply emitting an event saying the group profile has been updated.

If we dismiss the banner based on this, then the banner will get dismissed anytime any change to the group profile is made, even if it isn't a change to the income details.

What we want is to dismiss the banner if the income details are modified, and there are 4 group settings that would indicate the income details were modified:

1. `incomeDetailsType`
2. `incomeAmount`
3. `pledgeAmount`
4. `paymentMethods`

If any of these keys exist in `data`, it means the income details were modified.

So we should include `data` along with the event.

```js
sbp('okTurtles.events/emit', GROUP_PROFILE_UPDATE, data)
```

And check for those keys in our listener.

However, there's another problem: if we decide that this is the way we want to listen for updates to the income details, then we need to make sure we aren't listening to old events.

What do I mean?

Say you log in on a new device to your account and it is near the end of the month.

GI will start syncing all the group events, and it will dismiss the banner if we code it this way for two reasons:

1. Our event listener doesn't currently check for who updated their income details, and this sideEffect will get called each time anyone updates group profile.
2. Even if we check to make sure the message came from us and not someone else, we could still have a bug where it loads a message from us updating our group profile / income details a few weeks ago, outside of the 7-day window at the end of distribution period. We are only interested in updates *from us* that happen *within the 7-day window*.

So we would have to also check in our listener those two things, and that means we need to include that information in the data that's emitted by the event.

Who created the event is stored in `meta.username`, and when the event was created is stored in `meta.createdDate`.

So our code now looks like this:

```js
sbp('okTurtles.events/emit', GROUP_PROFILE_UPDATE, { meta, data })
```

And our event listener will become more complicated, because we need to check if `meta.username === this.ourUsername` (where we got `ourUsername` via `mapGetters`), and we need to check if `meta.createdDate` is after `warningDate` and before `dDay`.

This might be more complicated than it's worth. It might be simpler to simply emit an event from the IncomeDetails modal when `OK` is clicked, as that would work without the event listener having to check either of these things.

### Summary

If you take this approach:

1. Move emitting the event into a `sideEffect`
2. Modify it to emit `data` and `meta`
3. Check if the updated keys are one of the 4 keys in our listener, whether it's our user who is making the update, and whether the update occurred within the 7-day window.

However, I think it would be simpler to emit an event when the user clicks the submit button in the IncomeDetails modal, as that would greatly simplify the event listener.

frontend/views/pages/GroupDashboard.vue
taoeffect: So far I haven't seen a valid use-case for such an API but if I do then yeah we can add it.

I will add another comment below about how you're adding the event listener below, because I'm pretty sure it needs to be done differently and not here.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -107,10 +114,25 @@ export default ({
...
+      // e.g. user dismiss the modal without actually 'updating the income-details' and then click on the banner again.
+      sbp('okTurtles.events/off', GROUP_PROFILE_UPDATE, this.closeBanner)
+      sbp('okTurtles.events/once', GROUP_PROFILE_UPDATE, this.closeBanner)
```
taoeffect: Instead of this, it makes more sense for there to be a single `/on` inside of either `beforeMount` or `mounted`, and a single corresponding `/off` in `beforeDestroy`.

frontend/views/pages/GroupDashboard.vue
```diff
@@ -107,10 +114,25 @@ export default ({
...
     },
     closeBanner () {
-      localStorage.setItem('giHideDistributionWarning', true)
+      localStorage.setItem(this.bannerStorageKey, true)
       this.ephemeral.hideBanner = true
+    },
+    checkIfBannerBeenDismissed () {
```
taoeffect: `checkIfBannerBeenDismissed` isn't quite correct English, and also it is doing more than simply checking, it is also updating the value - potentially dismissing the banner.

Another possible name that might be a bit clearer would be: `updateBannerVisibility`

taoeffect (changes requested):
Review finished!

frontend/model/contracts/group.js
taoeffect: I just thought of another reason why it may more more sense to do it in the IncomeDetails modal when "OK" is clicked. Say you opened your income details and decided that in fact everything is correct as it looks, and so you click "OK" to confirm them. Then you're still "updating your income details" in the sense that you're confirming they're still correct, and so the reminder should disappear. We can't do that with this method, but doing in the IncomeDetails modal allows for it and is also simpler.

SebinSong: @taoeffect Yeah, reading through these comments, It sounds like creating an event for a modal button click is a cleaner and sufficient enough approach. I will make a new change that way. 
Btw, This long detailed documentation on how a GI contract works in the app is very helpful in understanding which means what in the contracts. will take some to play around with them too. thanks.

frontend/views/pages/GroupDashboard.vue
SebinSong: @taoeffect yeah that sounds better. coming up with the best name is the hardest part in writing a JS code btw lol.

SebinSong:
Additional updates which are,
- renaming `checkIfBannerBeenDismissed` to `updateBannerVisibility`
- move the event(`INCOME_DETAILS_UPDATE`) emission from inside the group contract to `IncomeDetails.vue` modal.
- attaching/detaching the listener once in `beforeMount` and 'beforeDestroy` hooks

have been made.

taoeffect (approved):
Excellent work @SebinSong! Approved! 🥇 🏆 😄

PR #1326, opened 2022-07-04
Fix chat not working on mobile

pieer:


taoeffect (changes requested):
Thanks @pieer! Tested!

This is how it should look like on mobile and tablet:

Screen Shot 2022-07-07 at 6 07 13 PM

There is a blue "Send" button in the input field. See Figma for details.

In this PR this is how it looks:

Screen Shot 2022-07-07 at 6 07 07 PM

pieer:
> Thanks @pieer! Tested!
> 
> This is how it should look like on mobile and tablet:
> 
> Screen Shot 2022-07-07 at 6 07 13 PM
> 
> There is a blue "Send" button in the input field. See Figma for details.
> 
> In this PR this is how it looks:
> 
> Screen Shot 2022-07-07 at 6 07 07 PM

Did you test on your laptop? Because the button is supposed to appear only if it's a touch device.
NB: atm when I try to test on the emulator, I have an issue that say: "Unhandled Promise Rejection: ReferenceError: Can't find variable: Notification"

So I can't test until we fix this issue

taoeffect:
> Did you test on your laptop? Because the button is supposed to appear only if it's a touch device.

Yes, I tested in Firefox by using its tool to make the screen the same size as on mobile.

Is there a way to make it possible to test this on laptops?

> NB: atm when I try to test on the emulator, I have an issue that say: "Unhandled Promise Rejection: ReferenceError: Can't find variable: Notification"
> So I can't test until we fix this issue

Can you merge `master`? That should fix the issue, I think. If it doesn't, please post a screenshot to Slack of the full console error log with some context and let me know how to reproduce it so that I can debug.

pieer:
So the test is on touch screen, which make sense since we need a button only when we don't have keyboard.

Yes the bug appear with the last version of master

taoeffect (changes requested):
IMG_6510

I tried this on mobile and it similarly doesn't work there either.

But I guess per the call we're waiting on @leihla to make the chat UI a bit more consistent in it's behavior? So I guess this PR is on hold for a bit.

taoeffect (changes requested):
@pieer Excellent work getting this going even with broken designs!! 👍 

I noticed three issues based on this screenshot in my testing, will comment in detail further below:

Screen Shot 2022-08-26 at 12 34 28 PM

1. At least on mobile, there's a gap that shouldn't exist at the bottom of the chat between the bottom of the chat and the bottom of the page.
2. The send button is always grey, it should turn blue when some text is typed so that the user knows they can click it
3. You can see in @leihla's [notes on Figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=5159%3A0) that the text box is supposed to grow vertically when the user either presses shift+enter or adds enough text for the box to grow vertically. However in this PR the text box doesn't grow.
- Note: I think the behavior here is that: the text box should grow vertically up to a max-size that is calculated based on the current window's viewport size (which could change I guess if the user resizes the window), and after that it should stop growing and show a scrollbar instead.

taoeffect (approved):
Great work @pieer! Approved! 😄

PR #1331, opened 2022-07-07
Attempt to fix errors related to timezone differences on proposal date display

### LINKED ISSUE #1293: Display proposal dates in the local timezone of the user
### Problem

Currently, the date a contract is created (and also likely the date it expires, when we implement showing that), displays the date in UTC timezone (I think), instead of the user's local timezone.

For example, here it shows "June 1, 2022" instead of "May 31, 2022":

Screen Shot 2022-05-31 at 5 26 30 PM

### Solution

Display the date in the local timezone of the client.

snowteamer:
Closes #1293

### DIFF
--- a/frontend/utils/time.js
+++ b/frontend/utils/time.js
@@ -113,14 +113,20 @@ export function firstDayOfMonth (date: Date): Date {
   return new Date(date.getFullYear(), date.getMonth(), 1)
 }
 
-// TODO: Provide locale fallback in case navigator does not exist (e.g. server, Mocha, etc...)
-const locale = (typeof navigator === 'undefined' && 'en-US') || (navigator.languages ? navigator.languages[0] : navigator.language)
-
 export function humanDate (
   date: number | Date | string,
-  opts?: Intl$DateTimeFormatOptions = { month: 'short', day: 'numeric' }
+  options?: Intl$DateTimeFormatOptions = { month: 'short', day: 'numeric' }
 ): string {
-  return new Date(date).toLocaleDateString(locale, opts)
+  const locale = typeof navigator === 'undefined'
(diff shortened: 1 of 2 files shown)


frontend/utils/time.js
```diff
@@ -113,14 +113,20 @@ export function firstDayOfMonth (date: Date): Date {
...
-  return new Date(date).toLocaleDateString(locale, opts)
+  const locale = typeof navigator === 'undefined'
+    // Fallback for Mocha tests.
+    ? 'en-US'
+    // Flow considers `navigator.languages` to be of type `$ReadOnlyArray<string>`,
+    // which is not compatible with the `string[]` expected by `.toLocaleDateString()`.
+    // Casting to `string[]` through `any` as a workaround.
+    : ((navigator.languages: any): string[]) ?? navigator.language
+  // NOTE: `.toLocaleDateString()` automatically takes local timezone differences into account.
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
```
taoeffect, line 128: OK, essentially the same thing except we're allowing the full array to be passed in. That seems fine. However, why did you move this into this function instead of leaving it outside of it? Now this code will have to be run each time `humanDate` is called, and that seems unnecessary.

taoeffect (commented):
Great PR @snowteamer, I think this should fix it. One doubt though, see comments.

frontend/utils/time.js
snowteamer: Mainly because other functions in this file don't need access to that variable.
I agree recomputing it is unnecessay most of the time, unless the user change their browser locale.

taoeffect (approved):
Approved!

PR #1336, opened 2022-07-08
#1312 implement 'mark all as read' buttons for notifications

### LINKED ISSUE #1312: Implement new designs for "Mark all as read" for Notifications panel
### Problem

See problem description in #1289.

### Solution

Implement ["Option 5"](https://github.com/okTurtles/group-income/issues/1289#issuecomment-1155295920) ([figma link](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23055%3A391024))

Leihla also suggests we should also separate the red counter from from the number of read messages, and have it be the count of notifications seen - and have the background of the message correspond to whether the message is actually "read".

In other words, when "Mark all as read" is clicked - it does two things:

1. It clears the counter entirely (all notifications are "seen")
2. It also changes the background of each notification by marking it as read.

And there are 2 ways to mark a notification as read:

1. Clicking "Mark all as read"
2. Clicking on the individual notification

SebinSong:
closes #1312


taoeffect (approved):
@SebinSong - this looks like a perfect PR, I couldn't find any issues.

Fantastic work! 🥇

PR #1337, opened 2022-07-08
Make notification text easier to highlight

### LINKED ISSUE #1315: Some issues with notifications display
### Problem

Two problems:

1. Notifications sometimes look weird when their contents are too big
2. You cannot highlight/select the text of a notification. This can be a problem if someone wants to copy/paste an error they've seen.

Screen Shot 2022-06-24 at 12 26 41 PM

### Solution

1. Fix UI issues
2. If it is possible to preserve the ability to click on a notification and also highlight text within it, make it so

snowteamer:
Closes #1315

### Additional info
- Hitting the back button after following a notification link no longer automatically re-opens the notifications, which might be inconvenient when manually checking many notifications in order.
Has someone any idea on how to fix that?

### DIFF
--- a/frontend/model/notifications/templates.js
+++ b/frontend/model/notifications/templates.js
@@ -58,7 +58,7 @@ export default ({
       }),
       icon: 'coins',
       level: 'info',
-      linkTo: '/contributions/TODO-LINK-MODAL',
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Seems to work really well, nice one @snowteamer !!

PR #1338, opened 2022-07-09
work on issue #1189 - don't allow uppercase in the username

### LINKED ISSUE #1189: Force usernames to be lowercase [$40 bounty]
### Problem

Currently usernames allow lowercase and uppercase letters.

This can potentially lead to phishing, impersonation, and other problems.

### Solution

- Prevent users from using uppercase letters in their username in `SignupForm.vue`
- Use a validation in `identity.js` for `'gi.contracts/identity'` action to verify there are no uppercase letters in the username. See #1335 for details on how to write that validation, as it's a related issue.

### Bounty

SebinSong:
closes #1189

### DIFF
--- a/frontend/model/contracts/identity.js
+++ b/frontend/model/contracts/identity.js
@@ -6,6 +6,7 @@ import Vue from 'vue'
 import '~/shared/domains/chelonia/chelonia.js'
 import { objectOf, objectMaybeOf, arrayOf, string, object } from '~/frontend/utils/flowTyper.js'
 import { merge } from '~/frontend/utils/giLodash.js'
+import { noUppercase } from '~/frontend/views/utils/validators.js'
 import L from '~/frontend/views/utils/translations.js'
 
 sbp('chelonia/defineContract', {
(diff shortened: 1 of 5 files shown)


frontend/views/utils/validators.js
```diff
@@ -1,3 +1,5 @@
 export const nonWhitespace = (value: string): boolean => /^\S+$/.test(value)
 
 export const decimals = (digits: number): ((value: number) => boolean) => (value: number) => Number.isInteger(value * Math.pow(10, digits))
+
+export const noUppercase = (value: string): boolean => /[^A-Z]+/i.test(value)
```
taoeffect: This would work if the only characters were English characters, but we need to also test other languages.

I *think* (please double-check me), that something like `value.toLowerCase() === value` might work better. Please check me though on that.

frontend/model/contracts/identity.js
```diff
@@ -20,13 +21,19 @@ sbp('chelonia/defineContract', {
...
+        objectMaybeOf({
+          attributes: objectMaybeOf({
+            username: string,
+            email: string,
+            picture: string
+          })
+        })(data)
+
+        if (!noUppercase(data.attributes.username)) {
+          throw new TypeError(L('The username cannot contain uppercase letters.'))
```
taoeffect: Oh, this actually doesn't need to be wrapped in `L`, since most likely this won't be a user-facing string. If you do want to wrap it in `L` though, I think it makes sense to keep it the same sentence as the other one you have `SignupForm`, so that translators only need to translate one version.

taoeffect (changes requested):
Excellent work @SebinSong! See comments. :)

frontend/model/contracts/identity.js
SebinSong: @taoeffect Got U. Have removed the L function from it.

frontend/views/utils/validators.js
SebinSong: @taoeffect `value.toLowerCase() === value` does work with a string that contains non-english characters. update the validator with it.

taoeffect (approved):
Well done @SebinSong. 👏

PR #1340, opened 2022-07-11
No punctuation in usernames

### LINKED ISSUE #1335: Do not allow punctuation in usernames [$40 bounty]
### Problem

When mentioning users in chat, it would be weird if someone were named `@bob?` instead of `@bob` Because then you'd have mentions that looked like this:

```
@bob? are you there?
```

Instead of:

```
@bob are you there?
```

Also, punctuation could introduce bugs in how the @mentions are parsed.

### Solution

Create a validation function that returns `true` or `false` and is used to validate the username in both `SignupForm.vue` as part of the vuelidate validations, and in `contracts/identity.js` as part of the validation for `'gi.contracts/identity'`.

I recommend using regex's `\b` to check for this: https://www.regular-expressions.info/wordboundaries.html

And for the contract validation you'll need to create a function. Have a look at how this validation in `group.js` is written:

```js
    'gi.contracts/group/inviteRevoke': {
      validate: (data, { state, meta }) => {
        objectOf({
          inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
        })(data)

        if (!state.invites[data.inviteSecret]) {
          throw new TypeError(L('The link does not exist.'))
        }
      },
```

You'll need to create a similar type of validation.

See also related issue: #1189 - Force usernames to be lowercase

### Bounty

snowteamer:
Closes #1335

### Summary of changes
- Usernames can no longer include non-alphanumeric characters other than hyphens and underscores.
- Usernames can no longer exceed 80 characters.
- Minimum password length and maximum username length are now configurable constants in `@contracts/constants.js`.
- A few new validators in `utils/validators.js`. Some of them are currently unused but could be useful to enable accented Latin-1 letters (e.g. `àéëó` etc) in usernames if required.


frontend/model/contracts/identity.js
```diff
@@ -2,13 +2,16 @@
 
 import sbp from '@sbp/sbp'
 import Vue from 'vue'
+import { maxLength } from 'vuelidate/lib/validators'
```
taoeffect: With the changes coming in #1308, importing a third-party dependency inside of the contracts is no longer non-trivial, as it requires careful thought and inclusion via `common.js`, see #1308 and the comments in that file for details.

To save ourselves some headaches, let's avoid doing that if we can. `maxLength` really isn't needed here anyway, as that's a simple `.length <= ...` operation.

frontend/views/containers/access/SignupForm.vue
```diff
@@ -102,8 +106,12 @@ export default ({
       form: {
         username: {
           [L('A username is required.')]: required,
-          [L('A username cannot contain spaces.')]: nonWhitespace,
+          [L('A username cannot contain white space.')]: nonWhitespace,
```
taoeffect: I am pretty sure that `white space` should be spelled `whitespace` as it's a single thing, and "white space" makes it sound like the space is the color white or something. Even though it might trigger some spellchecks, it seems like they're just not updated: https://en.wikipedia.org/wiki/Whitespace_character

frontend/views/utils/validators.js
```diff
@@ -1,3 +1,19 @@
+// Only allows non-empty strings of plain ASCII letters and digits.
+export const alphanumeric = (value: string): boolean => /^[A-Za-z\d]+$/.test(value)
+
+// Same as `alphanumeric`, but also allows accented letters from the first 256 Unicode code points.
+export const alphanumericLatin = (value: string): boolean => /^[A-Za-zÀ-ÖØ-Ýà-öø-ÿ\d]+$/.test(value)
+
+// Same as `alphanumeric`, but also allows hyphens.
+export const alphanumericOrHyphens = (value: string): boolean => /^[A-Za-z\d-]+$/.test(value)
```
taoeffect: What about underscores? Can we have those? (I personally see no issue with them)

taoeffect (changes requested):
Great work @snowteamer! Reviewed!

frontend/views/utils/validators.js
snowteamer: Sure, but the name of this function is going to become a bit long. Any idea to avoid `alphanumericOrHyphensOrUnderscores` ?

taoeffect: To simplify things, maybe remove `alphanumericLatin` and then rename `alphanumericOrHyphens` to something more generic and simple like `isValidUsername`

snowteamer: There are more conditions for a string to actually be a valid username, so that name wouldn't really be warranted. Maybe `allowedUsernameCharacters`?

taoeffect: Yeah that sounds good 👍

snowteamer: Also, should we apply the same rules for underscores as for hyphens, i.e. no leading, trailing, or consecutive underscores?

taoeffect: Sure. And we can relax or modify these rules in the future if necessary.

snowteamer: Yeah. It's likely better to start strict and relax as necessary than the opposite, as otherwise previously valid usernames could become suddenly invalid.

taoeffect (changes requested):
@snowteamer Something I missed from the previous PR on forcing lowercase usernames: if a user clicks on the username field in the SignUp form, and then clicks away from it (or tabs out of it), then this exception is thrown:

Screen Shot 2022-07-15 at 12 52 31 PM

frontend/model/contracts/identity.js
```diff
@@ -1,10 +1,19 @@
...
 import { merge } from './shared/giLodash.js'
 import { objectOf, objectMaybeOf, arrayOf, string, object } from '~/frontend/model/contracts/misc/flowTyper.js'
-import { noUppercase } from '~/frontend/views/utils/validators.js'
+import {
+  allowedUsernameCharacters,
+  noConsecutiveHyphensOrUnderscores,
+  noLeadingOrTrailingHyphen,
+  noLeadingOrTrailingUnderscore,
+  noUppercase
+} from '~/frontend/views/utils/validators.js'
```
taoeffect: If code is (1) included in the slim version of contracts (and this code should be), and (2) is shared with the frontend UI code, then it belongs somewhere inside the `frontend/model/contracts/shared/` folder.

This helps us remember to think carefully when making modifications to the behavior of such code, so that if the code changes, we can make sure that the usage in the UI (usually but not always the latest version of such code) doesn't conflict with whatever version of the contract that's being used.

See `frontend/model/contracts/shared/README.md` for details.

frontend/model/contracts/manifests.json
```diff
@@ -1,8 +1,3 @@
 {
-  "manifests": {
-    "gi.contracts/chatroom": "21XWnNKjPmeTVSNfQZ2DWD7KY4Be55R737KXSu4DDZmdhTwUrZ",
-    "gi.contracts/group": "21XWnNXj2ryGCoQEWy2KrDXBkyThZdGSWtZLsETg78h6rrpmjt",
-    "gi.contracts/identity": "21XWnNXrKKtTmrMJZZbjUQUocAwtETsgE8Ri6kY3K9Thy5bvMx",
-    "gi.contracts/mailbox": "21XWnNKoSm7e2vnGJvoFSRB7heFTeg24FwDCkb6JAazkzhpNdY"
-  }
+  "manifests": {}
```
taoeffect: This file shouldn't be deleted.

And this reminds me: since you are making modifications to the ~~chatroom~~ identity contract, and since we haven't yet released version `0.0.1` of the contracts, you can safely delete `contracts/0.0.1` and regenerate them using `grunt pin:0.0.1`.

Just make sure to run `grunt dev` at least once immediately before running `grunt pin`, see #1352 (just created) for details.

taoeffect (changes requested):
Great work @snowteamer! Some comments related to the new way we're doing contracts.

frontend/model/contracts/manifests.json
snowteamer: I didn't touch or delete this file manually. Also not sure in which ways the chatroom contract is modified. I would have thought only the identity contract would be involved, or maybe also the group contracts because of some transitivity since groups are made of users.
Going to follow your instructions anyway, but I wish there would be some automated way to detect and perform the necessary steps.

snowteamer: Got this error after deleting `contracts/0.0.1` and running `grunt dev`:
image

taoeffect: Oops, I meant the identity contract, sorry. Let’s try to troubleshoot this over Slack?

frontend/model/contracts/identity.js
taoeffect: @snowteamer This was marked resolved but doesn't seem to be?

frontend/model/contracts/manifests.json
taoeffect: @snowteamer this was also marked resolved but doesn't seem to be?

taoeffect (changes requested):
See comments

frontend/model/contracts/identity.js
snowteamer: @taoeffect Not sure what needed to be done here. Did you mean the `validarors.js` file should be moved to `contracts/shared/`?

frontend/model/contracts/manifests.json
snowteamer: I think I did delete the `contracts/0.0.1` folder, then ran `grunt dev --no-chunks` followed by `grunt pin:0.0.1` before the latest commit

frontend/model/contracts/identity.js
taoeffect: @snowteamer yep!

frontend/model/contracts/manifests.json
taoeffect: @snowteamer this file isn't being generated, or it is but without the hashes. Could you please investigate?

taoeffect (approved):
Nice work @snowteamer !

PR #1342, opened 2022-07-12
General banner SBP API and minor refactorings in main.js

snowteamer:
### Summary of changes:
- New SBP selectors in `utils/ui.js` related to the "general banner" API:
  - 'gi.ui/showBanner': wraps `banner.show()`
  - 'gi.ui/dangerBanner': wraps `banner.danger()`
  - 'gi.ui/clearBanner': wraps `banner.clean()`.
  - 'gi.ui/seriousErrorBanner': replaces the `displaySeriousErrorBanner()` function in `main.js`.
- New `CHELONIA_ERROR` notification template in `notifications/templates.js`,  used in `main.js` to greatly simplify the `notificationError()` function.
- ~~Minor simplification of the global SBP filter setup code in `main.js`, mainly to avoid `reduce()` calls.~~


frontend/main.js
```diff
@@ -50,59 +51,40 @@ async function startApp () {
   // force debug output even in production
   const debugParam = new URLSearchParams(window.location.search).get('debug')
   if (process.env.NODE_ENV !== 'production' || debugParam === 'true') {
-    const reducer = (o, v) => { o[v] = true; return o }
```
taoeffect, line 55: This should remain unchanged. The problem is that Set is `O(n)` while maps are `O(1)`, and we need these to be as fast as possible since they will get triggered on just about every SBP invocation.

frontend/views/utils/ui.js
```diff
@@ -2,15 +2,35 @@
 
 import sbp from '@sbp/sbp'
 import { OPEN_MODAL, MODAL_RESPONSE } from '@utils/events.js'
+import L, { LError } from '@view-utils/translations.js'
+
+// Call from anywhere in the app:
```
taoeffect: The comment should mention that to use this `BANNER` needs to be set, and maybe give an example

taoeffect (changes requested):
Nice, `main.js` is cleaned up a bit, and it was getting a bit out of hand tbh. Some minor changes requested.

frontend/main.js
snowteamer: Most sources I could find say that Set `has()` is `O(1)` in time. I've used sets here instead of arrays precisely to avoid the `O(n)` cost of Array `includes()`.

frontend/views/utils/ui.js
snowteamer: A few usage exemples can be found just below. Do you mean there should be an example of explicitly setting BANNER before use?
By the way, the related selectors can still be used somewhat safely even if `BANNER` is not set yet, thanks to the `?.` operator.

frontend/main.js
taoeffect: ~~AFAICT, [Set.has is `O(n)`](https://stackoverflow.com/a/31092145/1781435).~~

>  to avoid the O(n) cost of Array includes().

... We aren't using includes()...?

frontend/views/utils/ui.js
taoeffect: >  Do you mean there should be an example of explicitly setting BANNER before use?

Yeah, I think that'd be useful to mention

> By the way, the related selectors can still be used somewhat safely even if BANNER is not set yet, thanks to the ?. operator.

Yeah I saw, nice one! 👍

frontend/main.js
taoeffect: > ~~AFAICT, [Set.has is `O(n)`](https://stackoverflow.com/a/31092145/1781435).~~

Oops! I must be tired. I misread that. Here's another interesting comment that includes benchmarks: https://stackoverflow.com/a/69125167/1781435

I got slightly different results for Firefox:

Screen Shot 2022-07-15 at 8 42 26 PM

It shows that `Object#hasOwnProperty` is faster than `Set.has`, sometimes by an order of magnitude. So we should stick with that. It's not surprising that should be so, because a Set is considered to be an array with checks, whereas an object/map is considered to be a hashtable with O(1) lookup.

snowteamer: If I understand correctly (which is a bit tricky here), according to this SO answer the spec mandates Set `has()` to be sublinear, i.e. `O(n)` _at worst_. So implementors are free to make it better, which should be feasible since Set implementations likely use a hashtable internally.

However, now that I think about it... remembering that JS sets are _ordered_ sets and not just plain sets, makes me have doubts about the efficiency of `has()`. Perhaps this constraint prevents it to be as fast as object property access (I would be kinda disappointed though).

So in doubt I will agree to revert this change.

>... We aren't using includes()...? 

True. I would have had to use it after removing the reducers, if I had let the blacklists be plain arrays. So I opted for sets.

snowteamer: Just read your latest comment. So yeah let's stick with plain objects.

taoeffect (changes requested):
This PR looks really good @snowteamer!

Think it's ready to merge except for the issue with `manifests.json` as we discussed on Slack.

corrideat (dismissed):

taoeffect (approved):
Thanks @snowteamer! In the future though please open an issue before submitting a PR so that we can first agree on the problem being fixed!

PR #1347, opened 2022-07-13
Move app settings into their own Vuex module

### LINKED ISSUE #1276: Debug logging should show up if the user enables debug logging
### Problem

Currently the SBP debug logging check in `main.js` looks like this:

```js
```

It doesn't enable debug logging of SBP calls if the user enabled debug logging in their user settings.

### Solution

This is tricky as there are a few considerations.

- One might be tempted to somehow integrate with `sbp('state/vuex/state').appLogsFilter`, but that isn't available at that point (only upon login)
- It's tempting to simply remove the `if (process.env.NODE_ENV !== 'production' || debugParam === 'true') {` check, as that would then display debug logging upon login if the `appLogsFilter` has it enabled. However, users might accidentally leave this on and forget to turn if off, so perhaps the `debugParam` is the way to go (and we should entirely remove the checkbox for enabling debug logging from the UI). Think about this.

### LINKED ISSUE #1344: App settings should use default values when not logged in
### Problem

Modified user settings don't seem to revert to their default values upon logging out.
So when another user comes in and wants to sign up, they will have to either refresh the page or use the Sign Up form with preferences from the previous user applied.

### Solution

Make sure to revert interface settings to their default values whenever a logout happens. This must not prevent correctly saving user preferences to offline storage though.

snowteamer:
Closes #1344
Closes #1276

### Summary of changes:
- Settings like `appLogsFilter`, `fontSize`, `theme` etc. have been moved to `@model/settings/vuexModule.js`.
- They should now correctly revert to their default values upon successfully logging out.
- If a `?debug` query parameter is present and truthy, then debug- and log-level applications logs will be enabled by default. Users can still disable them later via the Application Logs modal though.


frontend/views/containers/user-settings/AppLogs.vue
```diff
@@ -16,13 +16,13 @@
           .c-filters-inner
             legend.c-filters-legend Optional logs:
             label.checkbox
-              input.input(type='checkbox' name='filter' v-model='form.filter' value='debug')
+              input.input(type='checkbox' name='filter' v-model='form.filter' value='debug' checked)
```
taoeffect: Why is the checked attribute for Debug and Log being modified?

frontend/model/settings/vuexModule.js
```diff
@@ -0,0 +1,73 @@
+'use strict'
```
taoeffect, line 1: A folder is usually needed when there are more than one related files. But this folder contains only a single file... so why not just have it be `settings.js`?

taoeffect (commented):
Not a review but some questions for now.

frontend/views/containers/user-settings/AppLogs.vue
snowteamer: That's a mistake of mine - I've read that these attributes are ignored when v-model is provided, so I first removed them. Then I thought I'd better do that later, so I added them back, albeit with this mistake.

frontend/model/settings/vuexModule.js
snowteamer: For consistency with the notification module, which has the same folder structure, although it contains more files.

taoeffect: ok, make sense 👍

frontend/model/contracts/manifests.json
```diff
@@ -1,8 +1,3 @@
 {
-  "manifests": {
-    "gi.contracts/chatroom": "21XWnNKjPmeTVSNfQZ2DWD7KY4Be55R737KXSu4DDZmdhTwUrZ",
-    "gi.contracts/group": "21XWnNXj2ryGCoQEWy2KrDXBkyThZdGSWtZLsETg78h6rrpmjt",
-    "gi.contracts/identity": "21XWnNXrKKtTmrMJZZbjUQUocAwtETsgE8Ri6kY3K9Thy5bvMx",
-    "gi.contracts/mailbox": "21XWnNKoSm7e2vnGJvoFSRB7heFTeg24FwDCkb6JAazkzhpNdY"
-  }
+  "manifests": {}
```
taoeffect: (Somehow my comment here got lost.)

This needs to not be empty. After merging `master` (which may fix this problem all on its own), be sure to run `grunt dev` at least once. The hashes should match and this file should be without modifications since it doesn't contracts.

taoeffect (changes requested):
Nice work @snowteamer - just a heads up - recent changes to `master` added a new setting (`notificationEnabled`) that needs to be merged with this PR.

PR #1348, opened 2022-07-13
Group Chat Improvements

### LINKED ISSUE #1328: Scroll to the message which is already deleted
### Problem

If the target message is deleted, the reply message will lost the target to scroll when user clicks it.

### Solution

Do something in **sideEffect** function of **model/contracts/chatroom.js**, as soon as the message is deleted.
Since we don't save messages in store, I think message.createdDate could be used.

### LINKED ISSUE #1345: Pick a different user generated event to trigger Notification permissions request
### Problem

In the recent group mentions PR that was merged (#1311), the `mousemove` method of requesting notification permissions doesn't seem to work in Firefox on our production server:

Screen Shot 2022-07-13 at 8 36 03 PM

Somehow we missed this, perhaps due to different behavior between production and development, or different browsers.

### Solution

Try a different user-generated event. @snowteamer suggested a mouse click, so we can try that for when the user clicks on "Chat" in the nav.

### LINKED ISSUE #1346: Mentions popup should only show usernames
### Problem

Screen Shot 2022-07-13 at 8 52 57 PM

When pressing `[tab]` on a username shown in the @mentions popup, it doesn't complete them as expected.

In this case, pressing tab with `@Snowt` completes it as `@snow`, which is very surprising and confusing to users.

### Solution

Make the popup only show usernames (don't show display names).

Silver-IT:
### Summary of Changes
- Handler when the target message is removed, after someone replies message from it
- Closes #1345 
- Closes #1346 
- Added image to `@all` in mentions popup
- Closes #1328 
- Close mention popup when click outside
- Focus textarea after click a mention from the popup


frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -14,7 +14,7 @@
         @click='addSelectedMention(index)'
       )
         avatar(:src='user.picture' size='xs')
-        .c-username {{user.displayName}}
+        .c-username {{user.username}}
```
taoeffect, line 17: If they have a display name (and only if a display name is set), can you add it in a grey font in parenthesis after the username?

E.g. so that the mentions looks something like this:

```
[IMG] greg
[IMG] snow (Snowteamer)
[IMG] alex (Alex Jin)
```

EDIT: just make sure that the lighter colored name looks good in both regular mode and dark mode.

taoeffect (changes requested):
Looking really good @Silver-IT! Minor request for improvement.

taoeffect (approved):
Great work @Silver-IT! Approved! 🎉

Issue #1303, opened 2022-06-18
Implement modal for Generic Proposal


taoeffect:
### Problem

The implementation for the new generic proposal design (#867) doesn't exist.

### Solution

Once #1300 is closed, implement generic proposals in the contracts and the modal for it.

SebinSong:
- "create proposals" button design figma link:
https://github.com/okTurtles/group-income/issues/1288#issuecomment-1155061508
- "modal" design figma link:
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23051%3A213049

SebinSong:
@taoeffect 
Things that I've completed working on as part of this task atm are,

1. building the Generic Proposal Modal(`GenericProposal.vue`). 
![image](https://user-images.githubusercontent.com/17641213/179202815-7f698a22-17a5-45a9-940c-96a24b482483.png)

2. make sure the 'open' generic proposal item shows up in the proposal widget in the Dashboard.
![image](https://user-images.githubusercontent.com/17641213/179204266-2117fba6-f546-4497-84bb-464591eaed2d.png)

Do you think these two cover the scope of this task item and deserve a PR to master? (I didn't touch any contract file, as no idea what to do with them yet..!)

SebinSong:
@taoeffect 
I just sent a [PR here](https://github.com/okTurtles/group-income/pull/1353) for my work for this item.
I noticed there's no call-to-action element implemented yet in the app that triggers a generic proposal modal, so
created this **temporary** 'Create general proposal' button in `ProposalWidget.vue` in the dashboard.

![image](https://user-images.githubusercontent.com/17641213/179385631-0730eebc-bb86-49aa-bc4a-3cb2928a3b30.png)

clicking on it will pop out the modal that I've implemented as screenshot below.

![image](https://user-images.githubusercontent.com/17641213/179385689-26ffb13d-31df-4b7f-b218-7077ae8f0739.png)

Looking forwards to your review and change requests for further improvements in the work.

SebinSong:
@taoeffect I believe this is good to be closed along with  #1362

PR #1353, opened 2022-07-17
#1303 implment generic proposal modal

SebinSong:
work for #1303


frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -148,7 +148,7 @@ export default ({
             })
           }[ruleName]()
         },
-        [PROPOSAL_GENERIC]: () => L('TODO: Change [generic] from [current] to [new-value]', {})
+        [PROPOSAL_GENERIC]: () => `${this.proposal.data.proposalData.name}`
```
taoeffect: Since you're not returning a modified version of the string, there's no need to wrap this in backticks. Just return the value.

taoeffect (changes requested):
Excellent @SebinSong! Nicely done!

A few minor change requests:

### Icon

Here the icon looks like a bar graph:

Screen Shot 2022-07-17 at 5 13 56 PM
 
And to be honest, that seems like a decent choice for it. However, the designs have this icon (and corresponding dark version):

Screen Shot 2022-07-17 at 5 14 38 PM

How difficult would it be to switch to that?

### Regenerate `contracts/0.0.1`

Since this PR modifies the contracts code we need to freeze this version of the contracts in time.

We do this once for each version release of the contracts. Right now we are on version 0.0.1 of the contracts, which are saved here: `contracts/0.0.1`, but since we haven't released them we are overwriting that version each time the contracts code is modified.

So to do this with your PR (since it modifies the `group.js` contract by modifying `shared/voting/proposals.js`), do the following:

1. Delete the folder `contracts/0.0.1`
2. Run `grunt pin:0.0.1`

And this needs to be done each time you make modifications to the contracts part of the code. Unless we've begun working on a new release (in which case you'd do `grunt pin:0.0.2`).

### See comment

One minor improvement in ProposalItem.vue, see comments.

SebinSong:
@taoeffect worked on all your change requests in [this commit](https://github.com/okTurtles/group-income/pull/1353/commits/fafd0b09c94ee378e7a436f0c578944f5f743f9b). pls let me know if there's additional thing to be done.

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -18,14 +18,17 @@
...
-      label.field(v-if='isReasonStep' key='reason')
-        i18n.label Why are you proposing this change?
-        textarea.textarea(
-          ref='reason'
-          maxlength='500'
-          data-test='reason'
-        )
-        i18n.helper This is optional.
+      template(v-if='isReasonStep')
+        slot(name='reason')
```
taoeffect: This looks like a significant change, did you double-check everything related to this to make sure it still works?

Also, I don't know if this bug is related to this, or to a different recent PR having to do with the `Enter` key being pressed that I didn't catch in my review (#1360), but when I created a group with 1 user, and went to create a proposal to change the mincome, and pressed Enter after entering the new mincome, this error came up:

Screen Shot 2022-07-23 at 5 07 19 PM

Do you think you could track it down?

taoeffect (commented):
Nice work @SebinSong, have a question and found a weird bug... see comment

SebinSong:
@taoeffect
A fix for this issue has been made in [this commit](https://github.com/okTurtles/group-income/pull/1353/commits/d32b5975892c3dd31e8dbdda1abb6922a2109644)

Firstly, this issue isn't related to `slot(name='reason')` I created but while looking into the issue, I noticed this slot is not being used anywhere, so have deleted & reverted it back to what it was in `ProposalTemplate.vue`

Secondly, the cause of the issue was `Mincome.vue` trying to reference the the error-banner after it being destroyed. but that's only the case when `this.groupShouldPropose` is `false`, which matches your scenario where the group member is only one.

Pls let me know if you spot any other issue in this PR.

taoeffect (approved):
Approved! Great work @SebinSong! 😄

PR #1355, opened 2022-07-19
Basic notifications improvements

### LINKED ISSUE #1349: Basic notifications improvements
### Problem

Our notifications setup could still use some improvements.

- Our notifications settings don't allow you to toggle notifications on and off
- User feedback for notifications isn't the best
- When a user clicks on a notification for an @mention, it doesn't take them to the message

### Solution

h/t @snowteamer for the following suggestions:

> - Use a toggle switch like we already do e.g. for the reduced motion setting. This would also allow revoking a previously granted permission, which currently doesn't look easy to do.

Greg adds: since we don't have designs for this, a simple checkbox will do.

> - Emit a single notification when the permission is granted, with a nice feedback message. This would immediately reassure users that native notifications are now effectively enabled and working. A corresponding Vue notification is not necessary.

And from myself:

> clicking on the browser notifications => taking you to the corresponding message in chatroom or DM

### LINKED ISSUE #1350: Exception thrown in chat sideEffect
### Problem

When @dotmacro used the app, she encountered this error:

image

However, the error went away after she refreshed the page. Still we should track it down and fix it.

### Solution

Find the issue and fix.

Silver-IT:
### Summary of Changes

- Closes #1349 
- Tried to resolve #1350 ( But did not get this error. )
- Fix error to display unread position
  1. Regarding `createdDate`
  2. When switch channel very fast
- Display `Original message was removed.` in reply message
- Fix error in delete chatroom
- Improve to display the mentions count when user edit message


frontend/model/contracts/chatroom.js
```diff
@@ -393,6 +399,15 @@ sbp('chelonia/defineContract', {
         if (msgIndex >= 0) {
           state.messages.splice(msgIndex, 1)
         }
+        // filter replied messages and check if the current message is original
+        for (const message of state.messages) {
+          if (message.replyingMessage) {
+            if (message.replyingMessage.id === data.id) {
```
snowteamer: No big deal but these two conditions could be merged into one with `&&`

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -9,20 +10,39 @@ export async function requestNotificationPermission (force: boolean = false): Pr
...
-
   try {
-    return await Notification.requestPermission()
+    const permission = await Notification.requestPermission()
+    if (permission === 'granted') {
+      sbp('state/vuex/commit', 'setNotificationGranted', true)
+    } else {
+      sbp('state/vuex/commit', 'setNotificationGranted', false)
+    }
+    return permission
```
taoeffect: I do not think these changes are an improvement. They seem to me to be very unnecessary and only add the unnecessary complexity of managing more state for no reason.

It is perfectly fine to call `makeNotification` without checking whether this variable is set, because that function already checks if permissions are granted.

Please remove `setNotificationGranted`, `state.notificationGranted`, and everything associated with it.

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -9,20 +10,39 @@ export async function requestNotificationPermission (force: boolean = false): Pr
...
+    }
+  }
+}
+
+export function getNotificationPermission (): string | null {
+  if (typeof Notification === 'undefined') {
+    return null
+  }
+
+  return Notification.permission
```
taoeffect: This function does basically nothing. It should be removed.

Wherever you need to check for these permissions, you can simply do:

```js
Notification?.permission === 'granted'
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -449,7 +454,9 @@ export default ({
       } else {
         events = await sbp('chelonia/out/eventsBefore', before, limit)
       }
-
+      if (curChatRoomId !== this.currentChatRoomId) {
+        return
+      }
```
taoeffect: This needs a comment above it explaining why this is here

frontend/model/contracts/chatroom.js
```diff
@@ -59,13 +59,19 @@ function addMention ({ contractID, messageId, datetime, text, username, chatRoom
...
 
-  const rootGetters = sbp('state/vuex/getters')
-  const groupID = rootGetters.groupIdFromChatRoomId(contractID)
-  makeNotification({
-    title: `# ${chatRoomName}`,
-    body: text,
-    icon: rootGetters.globalProfile2(groupID, username).picture
-  })
+  const rootState = sbp('state/vuex/state')
+  if (rootState.notificationGranted) {
```
taoeffect: Please revert this change (see other comment below)

taoeffect (changes requested):
Great work @Silver-IT! Preliminary review finished!

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -9,20 +10,39 @@ export async function requestNotificationPermission (force: boolean = false): Pr
...
     return null
   }
-
   try {
-    return await Notification.requestPermission()
+    const permission = await Notification.requestPermission()
+    if (permission === 'granted') {
+      sbp('state/vuex/commit', 'setNotificationGranted', true)
+    } else {
+      sbp('state/vuex/commit', 'setNotificationGranted', false)
```
snowteamer: Could also slightly simplify here, with `sbp('state/vuex/commit', 'setNotificationGranted', permission === 'granted')`

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: The reason why I used this state.notificationGranted is because users can not switch Notification permission using the Switch button.
For example, if a user allows permission and switch on, the notifications should be working.
And once he switched off, the notifications should not be working.
But a simple switching off, this can not deny browser notification permission.

That's why it is necessary.

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: Yeah. Correct. Nice catch.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Got it.

frontend/model/contracts/shared/nativeNotification.js
taoeffect: @Silver-IT There has to be a better way to implement this.

For example, in `mounted` you could check to see what the value of it is, and then set an ephemeral boolean variable in the NotificationSettings.vue component based on it.

Then it will be sync'd from that point on.

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -9,20 +10,39 @@ export async function requestNotificationPermission (force: boolean = false): Pr
...
     return
   }
-  // eslint-disable-next-line
-  new Notification(title, { body, icon })
+
+  const notification = new Notification(title, { body, icon })
+  if (url) {
+    notification.onclick = function (event) {
+      event.preventDefault()
+      open(url, '_blank')
```
snowteamer: I don't feel really comfortable with the `open()` function. Wouldn't it be better to somehow push the URL onto the Vue router?

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: > For example, in mounted you could check to see what the value of it is, and then set an ephemeral boolean variable in the NotificationSettings.vue component based on it.

```
  mounted () {
    const permission = Notification?.permission
    if (permission === null) {
      this.disabled = true
    } else {
      this.granted = permission === 'granted' && this.notificationGranted
    }
  },
```

We use `granted` as a init value of the Switch component. Do you mean this?
Actually this value `this.granted` is the final boolean flag represents the if notification will be displayed or not.

frontend/model/contracts/chatroom.js
Silver-IT: if (message.replyingMessage?.id === data.id)
And this is better solution, isn't it?

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: Yeah. You are the KING of Code Optimization. 😁

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: Yeah. I am not sure why I didn't think that way. Really nice catch it is.

frontend/model/contracts/shared/nativeNotification.js
taoeffect: > We use granted as a init value of the Switch component. Do you mean this?

Yeah that looks good, something like that! 👍

frontend/model/contracts/shared/nativeNotification.js
taoeffect: @snowteamer See other discussion about this being removed completely: https://github.com/okTurtles/group-income/pull/1355#discussion_r926296100

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: `this.notificationGranted` is the Vuex state. So do you agree to use vuex state for this purpose?

taoeffect: Oops! Sorry I didn't see that part. No I'm still very much against using vuex for this :P

Why is that part even there? Couldn't it be just:

```js
  mounted () {
    const permission = Notification?.permission
    if (permission === null) {
      this.disabled = true
    } else {
      this.granted = permission === 'granted'
    }
  },
```

frontend/model/contracts/chatroom.js
snowteamer: Indeed (provided `data.id` cannot be undefined)

frontend/model/contracts/shared/nativeNotification.js
snowteamer: @greg Yea this makes my suggestion redundant

frontend/model/contracts/chatroom.js
Silver-IT: Right. Only `message.replyingMessage` could be undefined.
```
if (message.replyingMessage?.id === data.id)
```

frontend/model/contracts/chatroom.js
```diff
@@ -367,11 +371,10 @@ sbp('chelonia/defineContract', {
         const mentions = makeMentionFromUsername(me)
         const isIncludeMention = data.text.includes(mentions.me) || data.text.includes(mentions.all)
         if (!isAlreadyAdded && isIncludeMention) {
-          // TODO: Not sure createdDate should be this way
           addMention({
             contractID,
             messageId: data.id,
-            datetime: meta.createdDate,
+            datetime: data.createdDate, // TODO: Not sure createdDate should be this way
```
taoeffect: Can you explain why you are using `data.createdDate` here instead of `meta.createdDate`? What does this represent?

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -1,28 +1,33 @@
...
+import sbp from '@sbp/sbp'
 
 // NOTE: since these functions don't modify contract state, it should
 //       be safe to modify them without worrying about version conflicts.
 
 export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
   if (typeof Notification === 'undefined' ||
-    (!force && Notification.permission !== 'default') ||
-    (force && Notification.permission === 'granted')) {
+    (!force && Notification.permission !== 'default')) {
```
taoeffect: Why are we returning `null` in this case instead of just calling `await Notification.requestPermission()` below?

What if `Notification.permission` is === `'granted'` here? It would return `null` but we want this to return `'granted'` in that case...

frontend/model/contracts/chatroom.js
Silver-IT: That's because this `datetime` is used to decide if the this mention should be removed or not.
Let's say the last message was a mention.
If user reads the last message, the mention is removed from the list of mentions.
And also the `UnreadSince.messageId` is the ID of the last message which made mention.
Here, if user edits the message(the last one) for example, user can update mention from @all to @alex or just update original message content, the mention is created again.
If the new mention's datetime is the editing time, it is later than the UnreadSince.createdDate.
So we can not delete this mention unless user creates a new message and read it.
That's why I used the `data.createdDate` instead of `meta.createdDate`.

taoeffect: OK, I'm having difficulty following, and I don't think you answered what `data.createdDate` represents. What time does it represent? What time is stored in it?

Silver-IT: `meta.createdDate` : The date when edit this message
`data.createdDate` : The date when the message was created

taoeffect: >  data.createdDate : The date when the message was created

OK, I see.

Normally that information is stored on the message object itself when it was created via `'gi.contracts/chatroom/addMessage'`.

Is there a reason you can't do that here and have to instead pass it in as part of the message edit?

frontend/model/contracts/shared/nativeNotification.js
Silver-IT: > Why are we returning null in this case instead of just calling await Notification.requestPermission() below?

We don't need to request permission because Notification is undefined or user already reacted.
Since the `force` is false, we don't need to request if he is already replied.

frontend/model/contracts/chatroom.js
Silver-IT: Correct, we can not do this in `process` or `sideEffect` or addMessage.
This happens only when the message is edited.

frontend/model/contracts/shared/nativeNotification.js
taoeffect: @Silver-IT The problem though is that it's not specified what this function is supposed to return, and it should be specified.

In some places in the code you check the return value of this function and expect it to be a string, and check to see if it's `'granted'`. In other places you don't check the return the value - but if you're checking for `'granted'` even in a single place - then it's implied that this function returns `'granted'` when permissions are granted.

Yet here the function will return `null` in some cases when permissions are `'granted'`. That is very confusing and I think will lead to bugs.

Silver-IT: > What if Notification.permission is === 'granted' here? It would return null but we want this to return 'granted' in that case...

The main purpose of this condition is to skip when Notification is `undefined` or user has already replied.
Also it's good idea to return permission if Notification is not `undefined`.

frontend/model/contracts/chatroom.js
taoeffect: Right, but I asked if there was a reason that it has to be passed in here like this, instead of being stored when `addMessage` is called?

Can you add in a comment and explain the reason?

frontend/model/contracts/shared/nativeNotification.js
taoeffect: OK, then the behavior should be changed to return `null` when `Notification` is `undefined`, and the permission otherwise.

*Edit: And also re-request permissions if `force` is `true` + return the permission

frontend/model/contracts/chatroom.js
Silver-IT: Okay.

frontend/model/contracts/shared/nativeNotification.js
Silver-IT:
```
export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
  if (typeof Notification === 'undefined') {
    return null
  } else if (!force && Notification.permission !== 'default') {
    return Notification.permission
  }
  try {
    return await Notification.requestPermission()
  } catch (e) {
    return null
  }
}
```

So, is this fine with you?

taoeffect: How about?

```js
export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
  if (typeof Notification === 'undefined') {
    return null
  }
  if (force) {
    try {
      return await Notification.requestPermission()
    } catch (e) {
      console.error('requestNotificationPermission:', e.message)
      return null
    }
  }
  return Notification.permission
}
```

Silver-IT: Much better.

Silver-IT: But your code missed one situation. When force is false and the permission is `default`.

Silver-IT:
```
export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
  if (typeof Notification === 'undefined') {
    return null
  }
  if (force || Notification.permission === 'default') {
    try {
      return await Notification.requestPermission()
    } catch (e) {
      console.error('requestNotificationPermission:', e.message)
      return null
    }
  }
  return Notification.permission
}
```

This could be perfect.

taoeffect: > But your code missed one situation. When force is false and the permission is default.

Good catch!!

Is this right?

```js
export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
  if (typeof Notification === 'undefined') {
    return null
  }
  if (force || Notification.permission === 'default') {
    try {
      return await Notification.requestPermission()
    } catch (e) {
      console.error('requestNotificationPermission:', e.message)
      return null
    }
  }
  return Notification.permission
}
```

taoeffect: > This could be perfect.

We both wrote the same thing 😄

Silver-IT: Hahaha. Sure. Not only right. It's perfect.

contracts/0.0.1/group-slim.js
```diff
@@ -1464,7 +1464,7 @@ ${this.getErrorInfo()}`;
           }
         },
         process({ data, meta }, { state }) {
-          import_common2.Vue.delete(state.chatRooms, data.chatRoomID);
+          import_common2.Vue.delete(state.chatRooms[data.chatRoomID]);
```
taoeffect: Looks like this got reverted in the recent commit. Did you undo the `grunt pin:0.0.1` that you did earlier? Remember you need to delete the old `contracts/0.0.1` folder and run that command each time you update this PR.

Silver-IT: I removed `0.0.1` folder and run `grunt pin:0.0.1`. Also I tried it again just before but it works that way.
Weird...

Silver-IT: However, It's a really nice catch...

Silver-IT: Should I run `grunt dev` and then `grunt pin:0.0.1`?

Silver-IT: I did this way and it is resolved...

frontend/views/containers/user-settings/NotificationSettings.vue
```diff
@@ -1,20 +1,52 @@
...
+      if (permission !== null) {
+        permission = await requestNotificationPermission(true)
+      }
+      if (!this.granted && (permission === 'granted')) {
+        makeNotification({
+          title: L('Congratulations'),
+          body: L('You have granted browser notification!')
+        })
+      }
+      this.$refs.permission.checked = this.granted = permission === 'granted'
```
taoeffect: OK, this is my fault @Silver-IT - I gave you a somewhat problematic review earlier — my apologies!! 🙇‍♂️

You were right - we do need a setting in the user settings for whether or not notifications are enabled, because there's no way to tell Firefox to disable the notifications once the permissions have been granted.

Thankfully, I don't think `chatroom.js` needs to be modified.

The only modifications that are needed are in this file (to check and toggle the user setting), `state.js` (to add the setting), and two functions in `nativeNotifications.js`.

Something like this for `requestNotificationPermission`:

```js
export async function requestNotificationPermission (force: boolean = false): Promise<null | string> {
  if (typeof Notification === 'undefined') {
    return null
  }
  if (force || Notification.permission === 'default') {
    try {
      const permission = await Notification.requestPermission()
      sbp('state/vuex/commit', 'setNotificationsEnabled', permission === 'granted')
    } catch (e) {
      console.error('requestNotificationPermission:', e.message)
      return null
    }
  }
  return Notification.permission
}
```

And something like this for `makeNotification`:

```js
export function makeNotification ({ title, body, icon, path }: {
  title: string, body: string, icon?: string, path?: string
}): void {
  const notificationsEnabled = sbp('state/vuex/state').notificationsEnabled
  if (typeof Notification === 'undefined' || Notification.permission !== 'granted' || !notificationsEnabled) {
    return
  }

  const notification = new Notification(title, { body, icon })
  if (path) {
    notification.onclick = function (event) {
      event.preventDefault()
      sbp('controller/router').push({ path }).catch(console.warn)
    }
  }
}
```

taoeffect (changes requested):
Great work @Silver-IT - something was messed up by my earlier review - sorry! See comment:

frontend/views/containers/user-settings/NotificationSettings.vue
Silver-IT: > There's no way to tell Firefox to disable the notifications once the permissions have been granted.

Yeah. It's same in the Google Chrome. That's why I made this.
```
this.$refs.permission.checked = this.granted = permission === 'granted'
```
So switch will not be changed **unless user reset the notification permission and set the permission again**.

taoeffect: Right, but doing it this way means that there's no way to uncheck the button, and notifications are always enabled, even if the user wants to disable them.

Silver-IT: It's same as if user wants to enable after he denied the permission.
User needs to reset permission anyway.

taoeffect: @Silver-IT If we are offering the user a checkbox, then the user should be able to easily disable the notifications using this checkbox. Right now it doesn't work - clicking on the checkbox does nothing after notifications permissions have been granted.

We need this to behave in the way users expect: for notifications to stop when they uncheck the checkbox. And since there is no API to disable notifications permissions, we do have to add an extra variable to the state.

Silver-IT: Yeah. This is what I thought.
But users will see that switch component doesn't work unless they reset notification permission.
Could it be helpful for the users to know why it doesn't work and make him to reset browser permission?

taoeffect: Yep! Per the discussion on Slack, in such instances we can let them know via an `alert`

taoeffect (approved):
Great work @Silver-IT! Approved! 😄

PR #1360, opened 2022-07-21
make sure hitting 'enter' key work as expected in various places in the app

### LINKED ISSUE #1155: `Enter` keypress event makes error in some input fields
### Problem

`Enter` keypress event works wrong for some input fields.
For example, when we press `Enter` in the `Display Name` input field of `Account Profile` page, it opens `Change Password` modal which is not expected.
I am not sure this is the only issue in `Display Name` input field and I think we need to check the input fields of all the forms.

Screenshot 1 [ Before press `Enter` key ]
image
Screenshot 2 [ After press `Enter` key ]
image

### Solution

I guess the reason is related to the `form.submit` or attributes of buttons in the form.

### LINKED ISSUE #1356: Pressing Enter in proposals modal doesn't work
### Problem

Whether it's the modal for "Add new members", or the modal for "Change minimum income", pressing `<Enter>` doesn't result in the "Next ->" button being pressed.

### Solution

Find and fix.

SebinSong:
closes #1356
closes #1155 (possibly)

### DIFF
--- a/frontend/views/containers/proposals/ProposalTemplate.vue
+++ b/frontend/views/containers/proposals/ProposalTemplate.vue
@@ -12,7 +12,10 @@
       i18n(key='title1' v-if='isConfirmation') Your proposal was created
       span(v-else) {{ title }}
 
-    form.c-form(@submit.prevent='')
+    form.c-form(
+      @submit.prevent=''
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Seems to work! Great job @SebinSong!

PR #1363, opened 2022-07-24
add --overwrite option to 'grunt pin'

### LINKED ISSUE #1361: Add `--overwrite` option to `grunt pin`
### Problem

Each time we want to make an update to a contract version that we haven't released we need to run `rm -rf contracts/<version>` and `grunt pin:<version>`. It would be nice if we could run just one command instead.

### Solution

Implement `grunt pin:<version> --overwrite` to remove `contracts/<version>` before pinning the latest changes.

SebinSong:
closes #1361


taoeffect (approved):
Excellent!

Issue #1284, opened 2022-05-23
Design proposal for automated bans


taoeffect:
### Problem

Automated bans are system generated proposals to remove a member because their client might be broken / hacked. However, currently visually they look too much like a member consciously chose to ban someone (see #1242).

### Solution

Create a more visually distinct design for them so that users don't think they personally to ban someone, but that it's part of an automated system.

leihla:
Update automated ban to new layout. 

### Old layout:
<img width="772" alt="Screenshot 2022-06-14 at 12 46 05" src="https://user-images.githubusercontent.com/158178/173569964-6f02f2d8-71cb-4f86-94ca-ba1148481454.png">

---

### New Layout:

<img width="785" alt="Screenshot 2022-06-14 at 12 47 57" src="https://user-images.githubusercontent.com/158178/173570279-8567940f-0138-4bf6-9c58-f685819b4086.png">

**Changes**:
- Show group avatar next to proposal icon
- Add [Automated] before proposal title
- Keep description as is

---
@taoeffect can you assign this issue to the correct person?

taoeffect:
Closing in favor of #1304 and #1300.

SebinSong:
@taoeffect 
I would love to pick up this issue if no one else is already looking into it. :)
where in the [Group Income 2.0 figma ](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0)does this new design live in?

Also, could you help with describing the steps to take in the app to generate that 'automated ban proposal' UI?

cheers,

PR #1365, opened 2022-07-25
Fix #1304:  Implement new design for automated ban proposals.

### LINKED ISSUE #1304: Implement new design for automated ban proposals.
### Problem

See the problem description in the related design issue #1284.

### Solution

Implement the new design #1284.

This issue depends on #1300 being closed.

pieer:


frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -119,6 +119,11 @@ export default ({
       const username = this.proposal.meta.username
       const isOwnProposal = username === this.ourUsername
 
+      if (this.proposal.data.proposalData.automated) {
+        return isOwnProposal
+          ? L('Automated ban sent by your client')
+          : L('Automated ban sent by {username}\'s client', { username })
```
taoeffect: To avoid the backslash to escape the `'`, can you instead use double-quotes for this string?

frontend/controller/actions/group.js
```diff
@@ -436,7 +436,7 @@ export default (sbp('sbp/selectors/register', {
           if (!proposal.votes[getters.ourUsername]) {
             await sbp('gi.actions/group/proposalVote', {
               contractID: groupID,
-              data: { proposalHash, vote: VOTE_FOR, passPayload: { secret: '' } },
+              data: { proposalHash, vote: VOTE_FOR, passPayload: { secret: '', automated: true } },
```
taoeffect: `automated` doesn't need to be here, you have it correct below. Please revert the change above.

taoeffect (changes requested):
Thanks @pieer! Left a few comments.

Also, this doesn't seem to make the requested design changes. This PR looks like this:
Screen Shot 2022-08-10 at 1 47 26 PM

And as Leihla [said](https://github.com/okTurtles/group-income/issues/1284#issuecomment-1155080107), it needs to look like this:

> Screenshot 2022-06-14 at 12 47 57
> 
> **Changes**:
> - Show group avatar next to proposal icon
> - Add [Automated] before proposal title
> - Keep description as is

taoeffect (changes requested):
@pieer Great work! I've tested this and found two issues that need to be fixed for the automated ban system to actually work.

FYI, after applying the change requests below, you can verify yourself that the automated ban system works by creating a group with 3 users in it (I recommend using Firefox Dev Edition with Session Tabs) and then with one of the users sending a malformed mutation of type `GIErrorIgnoreAndBan` from this URL: http://localhost:3000/app/error-testing

1. Edit the `'gi.contracts/group/removeMember'` action in `frontend/model/contracts/group.js` to add `automated: optional(boolean)` to the validation, so that it looks like this:

    ```js
    'gi.contracts/group/removeMember': {
      validate: (data, { state, getters, meta }) => {
        objectOf({
          member: string, // username to remove
          reason: optional(string),
          automated: optional(boolean),
          // In case it happens in a big group (by proposal)
    ```

2. Replace the contents of `flowTyper.js` file with: https://gist.githubusercontent.com/taoeffect/d030684e64c0b0db0d223cb09f7f64f6/raw/fee39e18708622158d0060c3d18bcd831c73f4bc/flowTyper.js
3. Run `grunt dev` at least once.
4. Run `grunt pin:0.0.4`

Then commit & push the changes.

taoeffect (approved):
Great work @pieer!

Issue #1247, opened 2022-04-28
Emit notifications for all of the various important situations [$40 bounty]


taoeffect:
### Problem

Currently we do not emit notifications when important events happen in the group (e.g. members join/leave, proposals created, passed, rejected, etc.)

### Solution

For all of the features we currently support that notifications can be emitted for, use the table in #663 as guidance and emit them.

See also related sub-issue #1124. Closing this will close that issue too.

### Bounty

$40 bounty for a clean solution to this (paid in cryptocurrency).

SebinSong:
@taoeffect I'm currently working on this issue and have worked on **'Member added'**, **'Member removed'** and **'Member left'** from this table below([original source](https://github.com/okTurtles/group-income/issues/663#issue-492988756)).

| Event | Copy |Links to |
| ------------- | ------------- | ------------- |
| Member added | _Your group has a new member! Say hi to [new_member_name]!_  | General group chat channel | 
| Members added<sup>[1]</sup> | _Your group has [#] new members! Say hi!_  | General group chat channel | 
| Member removed (via proposal)| _[member_name] was kicked out of the group. Contributions were updated accordingly_  | Contributions page | 
| Members removed (via proposal)<sup>[1]</sup> | _[member_name] and [#] others were kicked out of the group. Contributions were updated accordingly_  | Contributions page | 
| Member left (by their initiative)| _[member_name] has left your group. Contributions were updated accordingly_  | Contributions page | 
| Members left (by their initiative)<sup>[1]</sup> | _[member_name] and [#] others left your group. Contributions were updated accordingly_  | Contributions page | 
| Income details are 6 months old or more | _You haven't updated your income details in [amout_of_months]. Would you like to review them now?_ | Income details modal | 
| 1 week to send monthly contributions | _Don't forget to send your [month_name] contributions!_ | Contributions page | 
| Pending payment confirmations<sup>[1]</sup> | _You have [#] pending payment confirmations._ | Pay group page | 
| New proposal open | _[member_name] wants to [proposal_type_description]<sup>[2]</sup>. Vote now!_ | Dashboard proposals (anchor link) | 
| Proposal closed verdict | _[member_name]'s proposal to [proposal_type_description] was approved / not approved by your group._ | Archived proposals | 


But as the works rather deeply touch the group contract, I thought it needs your look&opinion regarding how it needs to be implemented.
[...]

PR #1369, opened 2022-07-31
#1247 - emit various notifications

SebinSong:
Implement emitting notifications in the app for when

1. A member has been added to a group
2. A member has been kicked out of a group
3. A member has left a group
4. New proposal open
5. Proposal closed verdict


frontend/model/contracts/group.js
```diff
@@ -703,13 +721,28 @@ sbp('chelonia/defineContract', {
...
           }]
         )
+      },
+      sideEffect ({ meta, contractID }, { state }) {
+        const { loggedIn } = sbp('state/vuex/state')
+        const { profiles = {} } = state
+        const myProfile = profiles[loggedIn.username] || null
+
+        if (loggedIn.username !== meta.username &&
+          isContractYoungerThanUser(meta, myProfile)) {
```
SebinSong, line 740: A user who has joined a group later than other existing members doesn't need to get notified of the existing members. For example, a `user3` who has just joined the group doesn't need to get a notification for `user2` who joined the group (let's say)2 days ago.

contracts/0.0.1/group.js
```diff
@@ -10438,11 +10453,23 @@ ${this.getErrorInfo()}`;
         }),
         process({ data, meta, contractID }, { state, getters }) {
           memberLeaves({ username: meta.username, dateLeft: meta.createdDate }, { meta, state, getters });
+          meta.isRemoveOurselves = true;
```
SebinSong, line 10456: Pls let me know if this(mutating the structure of the meta object for sending down an information to `sideEffect()` of `gi/contracts/group/removeMember`) is an issue. Or if you have an advice regarding this.

frontend/model/contracts/group.js
```diff
@@ -703,13 +721,28 @@ sbp('chelonia/defineContract', {
           { username: meta.username, dateLeft: meta.createdDate },
           { meta, state, getters }
         )
+
+        meta.isRemoveOurselves = true
```
SebinSong: Pls let me know if this(mutating the structure of the meta object for sending down an information to `sideEffect()` of `gi/contracts/group/removeMember`) is an issue. Or if you have an advice regarding this.

frontend/model/contracts/group.js
```diff
@@ -741,28 +774,41 @@ sbp('chelonia/defineContract', {
...
           }
         } else {
+          const myProfile = profiles[loggedIn.username] || null
           // we're an existing member of the group getting notified that a
           // new member has joined, so subscribe to their identity contract
           await sbp('chelonia/contract/sync', meta.identityContractID)
+
+          if (!myProfile) { return }
+
+          if (isContractYoungerThanUser(meta, myProfile)) {
```
SebinSong: A user doesn't need to get notifications for the existing members that were previously emitted. For example, a `user3` who has just joined the group doesn't need to get a notification for `user2` who joined the group earlier than him/her, let's say, 2 days ago.
`isContractYoungerThanUser()` util function here is used across multiple places for a similar purpose.

frontend/model/contracts/group.js
taoeffect: Modifying the input data to these functions is definitely not correct. :)

In most situations it won't work anyway. The only reason why it works here is because below we immediately call the `removeMember/sideEffect`.

There is a far better solution, I'll comment elsewhere.

frontend/model/contracts/group.js
```diff
@@ -703,13 +721,28 @@ sbp('chelonia/defineContract', {
...
+        const { profiles = {} } = state
+        const myProfile = profiles[loggedIn.username] || null
+
+        if (loggedIn.username !== meta.username &&
+          isContractYoungerThanUser(meta, myProfile)) {
+          sbp('gi.notifications/emit', 'MEMBER_LEFT', { // emit a notification for a member addition.
+            groupID: contractID,
+            username: meta.username
+          })
+        }
```
taoeffect: You should be able to handle everything in the `removeMember` sideEffect.

So this function doesn't need to be created.

To check for the case where the member is removing themselves from the group, check for the condition where `data.member === meta.username && data.member === loggedIn.username`. Those 3 will equal each other in such instances.

frontend/model/contracts/group.js
```diff
@@ -120,6 +120,15 @@ function memberLeaves ({ username, dateLeft }, { meta, state, getters }) {
...
 
+function isContractYoungerThanUser (contractMetaData, userProfile) {
+  if (!contractMetaData || !userProfile) { return }
+
+  const contractCreated = new Date(contractMetaData.createdDate)
+  const userJoined = new Date(userProfile.joinedDate)
+
+  return contractCreated > userJoined
+}
+
```
taoeffect: I don't understand the purpose of this function. It seems unnecessary.

taoeffect (changes requested):
Great work @SebinSong! I think this is your first time really interacting with the contracts, huh? Pretty cool! 😄

frontend/model/contracts/group.js
taoeffect: Oh, I missed your comment here: https://github.com/okTurtles/group-income/pull/1369#discussion_r933960575

FYI, these sorts of comments would be much better placed inline in the code than on Github. I will re-evaluate this function in light of that comment, but if you could also add that comment in the codebase it would be appreciated!

frontend/model/contracts/group.js
```diff
@@ -686,6 +695,15 @@ sbp('chelonia/defineContract', {
             })
           // TODO - #828 remove other group members contracts if applicable
         } else {
+          const myProfile = state.profiles[username] || null
```
taoeffect: Instead of getting the profile like this, use `getters.groupProfile(username)`.

frontend/model/contracts/group.js
```diff
@@ -686,6 +695,15 @@ sbp('chelonia/defineContract', {
...
+
+          if (!meta.isRemoveOurselves &&
+            isContractYoungerThanUser(meta, myProfile)) {
```
taoeffect: Since, per the other comment I posted, you can do all the checks here in this function, you no longer need the `isContractYoungerThanUser` I don't think, and you should use `compareISOTimestamps` from `shared/time.js` instead.

SebinSong:
@taoeffect update according to your CR has been done.

SebinSong:
@taoeffect I have just updated([commit](https://github.com/okTurtles/group-income/pull/1369/commits/e25fa4c684b178f588c07d8b43fa7ba3b17940af)) this PR to include 

**4. New proposal open
5. Proposal closed verdict**


There is couple of things to note regarding this change.

**1)** A notification message for a generic proposal

<image width='355' src='https://user-images.githubusercontent.com/17641213/183281009-2544e77d-3c67-4b6b-880e-574be823fb37.png'>

A notification body message for a generic proposal is `{username} created a proposal. Vote now!` for now. But let me know if this needs to be rewritten to a better one.

**2)** Notification messages for closed proposals

<image width='355' src='https://user-images.githubusercontent.com/17641213/183281214-a85b59b6-804e-454d-be33-d97e7f770b31.png'>

A notification message for a closed proposal is for now either `{member}'s proposal has passed.` or `{member}'s proposal has failed.` with a `cog` icon, as I couldn't find message templates for this notification for each proposal type(looks like they haven't been discussed & determined yet). But pls let me know if the figma(or descriptions) for this does exist somewhere.

cheers,

frontend/model/contracts/group.js
```diff
@@ -547,6 +558,27 @@ sbp('chelonia/defineContract', {
...
+          [PROPOSAL_REMOVE_MEMBER]: 'REMOVE_MEMBER',
+          [PROPOSAL_GROUP_SETTING_CHANGE]: 'CHANGE_MINCOME',
+          [PROPOSAL_PROPOSAL_SETTING_CHANGE]: 'CHANGE_VOTING_RULE',
+          [PROPOSAL_GENERIC]: 'GENERIC'
+        }
+
+        const myProfile = getters.groupProfile(loggedIn.username)
+
+        if (meta.username !== loggedIn.username &&
+          isActionYoungerThanUser(meta, myProfile)) {
```
taoeffect: I think in this case we don't need to check `meta.username !== loggedIn.username`. We should display notifications for proposals we posted ourselves.

frontend/model/contracts/group.js
```diff
@@ -120,6 +120,17 @@ function memberLeaves ({ username, dateLeft }, { meta, state, getters }) {
...
+  // A util function that checks if an action (or event) in a group occurred after a particular user joined a group.
+  // This is used mostly for checking if a notification should be sent for that user or not.
+  // e.g.) user-2 who joined a group later than user-1 (who is the creator of the group) doesn't need to receive
+  // 'MEMBER_ADDED' notification for user-1.
+
+  if (!actionMeta || !userProfile) { return }
+
+  return compareISOTimestamps(actionMeta.createdDate, userProfile.joinedDate) > 0
+}
+
```
taoeffect: Can we change this function in the following way?

1. Since `actionMeta` should always be defined, can we remove the check for `!actionMeta` and also add a Flowtype annotation of `: Object` for it?
2. If `userProfile` an be undefined sometimes, then can you add an annotation of `: ?Object` for it?
3. And finally, can you make this a true predicate function by always returning a `boolean` (and annotating the function as such with `: boolean`) by doing: `return userProfile && compareISOTimestamps(...)` ?

frontend/model/contracts/group.js
```diff
@@ -722,7 +781,7 @@ sbp('chelonia/defineContract', {
       validate: objectOf({
         inviteSecret: string // NOTE: simulate the OP_KEY_* stuff for now
       }),
-      process ({ data, meta }, { state }) {
+      process ({ data, meta, contractID }, { state }) {
```
taoeffect: Why was this added back in if it's never used?

frontend/model/contracts/group.js
```diff
@@ -747,22 +806,34 @@ sbp('chelonia/defineContract', {
...
-          for (const name in state.profiles) {
-            if (name !== rootState.loggedIn.username) {
-              await sbp('chelonia/contract/sync', state.profiles[name].contractID)
+          for (const name in profiles) {
+            if (name !== loggedIn.username) {
+              await sbp('chelonia/contract/sync', profiles[name].contractID)
             }
           }
         } else {
+          const myProfile = profiles[loggedIn.username] || null
```
taoeffect: The `|| null` is unnecessary here

frontend/model/contracts/group.js
```diff
@@ -747,22 +806,34 @@ sbp('chelonia/defineContract', {
...
           // we're an existing member of the group getting notified that a
           // new member has joined, so subscribe to their identity contract
           await sbp('chelonia/contract/sync', meta.identityContractID)
+
+          if (!myProfile) { return }
```
taoeffect: So, some consistency here is needed. Either `isActionYoungerThanUser` doesn't also do this check, in which case we should move `myProfile &&` into the `if` below, or it does do the check, in which case this line should be deleted as redundant.

taoeffect (changes requested):
Excellent work @SebinSong! Reviewed!

Is this still a WIP? If not please update the title!

EDIT: also, because the contracts have been updated, please pin a new version, e.g. `grunt pin:0.0.2`. And then if more change requests are required, continue updating the pin before pushing with `grunt pin:0.0.2 --overwrite`

frontend/model/contracts/group.js
SebinSong: Got it. made that change.

frontend/model/contracts/group.js
SebinSong: Have updated the function with the requirements.

frontend/model/contracts/group.js
SebinSong: This check has been merged with the `if` below.

SebinSong:
@taoeffect Worked on all your change requests.

Also, Regarding my work for `5. Proposal closed verdict`

<image width='350' src='https://user-images.githubusercontent.com/17641213/184049714-12c6d56a-760c-4fda-b525-f9de68d4b1b4.png'>

I believe everyone would think & agree this message should be more specific. for example, rather than just saying "u2's proposal", it needs to give more information like, "u2's proposal **'to update mincome'** has failed" or "u2's proposal **'to remove u3 from the group'** has failed"

Once what message template is used for which proposal type has been decided and an issue ticket is created for it, I can come back to work on implementing it too.

frontend/model/contracts/group.js
```diff
@@ -120,6 +120,15 @@ function memberLeaves ({ username, dateLeft }, { meta, state, getters }) {
   updateCurrentDistribution({ meta, state, getters })
 }
 
+function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boolean {
+  // A util function that checks if an action (or event) in a group occurred after a particular user joined a group.
+  // This is used mostly for checking if a notification should be sent for that user or not.
+  // e.g.) user-2 who joined a group later than user-1 (who is the creator of the group) doesn't need to receive
+  // 'MEMBER_ADDED' notification for user-1.
+  return Boolean(userProfile) &&
```
taoeffect: I think `Boolean(userProfile) &&` can be removed and the type annotation above modified to `Object` instead of `?Object`.

Anywhere in the calling function where myProfile/userProfile is null, that's where the check can be added if necessary. And as I'm looking through the code, I don't see where that is even necessary.

frontend/model/contracts/group.js
```diff
@@ -747,22 +803,32 @@ sbp('chelonia/defineContract', {
...
+              await sbp('chelonia/contract/sync', profiles[name].contractID)
             }
           }
         } else {
+          const myProfile = profiles[loggedIn.username]
           // we're an existing member of the group getting notified that a
           // new member has joined, so subscribe to their identity contract
           await sbp('chelonia/contract/sync', meta.identityContractID)
+
+          if (myProfile && isActionYoungerThanUser(meta, myProfile)) {
```
taoeffect: Here we don't need `myProfile && `. `myProfile` was set above in the `process` function, it's guaranteed to not be undefined.

taoeffect (changes requested):

taoeffect (approved):
Excellent work @SebinSong!

I'm excited to see both these changes merged and also the first real test of the contract upgrade system. 😄

PR #1370, opened 2022-07-31
Fix #1240: Distribution date info on payments page

### LINKED ISSUE #1240: Design for: distribution date info on payments page
### Problem

As we're going with rolling 30-day periods, they can land on any day of the month. @leihla, you've already done a fantastic job of designing the warning on the dashboard for when the next distribution date is. Something that I think is missing though is that same info somewhere on the payments page, and maybe shown all of the time in small text somewhere, so that people know when they need to finish making their payments by.

Related Figma comment: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=7813:0#186592040

### Solution

Come up with a great design for addressing this problem and then open an issue to implement.

pieer:


frontend/views/pages/Payments.vue
```diff
@@ -301,6 +314,12 @@ export default ({
         amount: this.withGroupCurrency(amount),
         count: this.paymentsTodo.length
       }
+    },
+    isCloseToDistributionTime () {
+      const dDay = new Date(this.groupSettings.distributionDate)
```
taoeffect: `this.groupSettings.distributionDate` is the date for the current period.

The value we want here is for the next period. See how `nextDistribution()` works to get that date.

taoeffect (changes requested):
Nice work @pieer! Some date related comments.

frontend/views/pages/Payments.vue
```diff
@@ -63,6 +63,20 @@ page(
       )
 
       .tab-section
+        banner-simple(severity='info')(
```
taoeffect: During build time in `grunt dev`, this `)(` results in a warning:

```
frontend/views/pages/Payments.vue, line 66:
You should not have pug tags with multiple attributes.
```

frontend/views/pages/Payments.vue
```diff
@@ -301,6 +314,12 @@ export default ({
         amount: this.withGroupCurrency(amount),
         count: this.paymentsTodo.length
       }
+    },
+    isCloseToDistributionTime () {
+      const dDay = this.periodAfterPeriod(this.groupSettings.distributionDate)
+      const warningDate = addTimeToDate(dDay, -7 * DAYS_MILLIS)
+
+      return Date.now() >= new Date(warningDate).getTime() && Date.now() < dDay.getTime()
```
taoeffect: `dDay.getTime()` will throw an exception here because `periodAfterPeriod` returns a string, not a date, and needs to be converted using the `dateFromPeriodStamp` function from `time.js`.

So to fix line 319 should look like this:

```js
const dDay = dateFromPeriodStamp(this.periodAfterPeriod(this.groupSettings.distributionDate))
```

Sorry for not catching this in my previous review. You can use the Date & Time System Preferences to test this stuff locally.

frontend/views/pages/Payments.vue
```diff
@@ -63,6 +63,20 @@ page(
       )
 
       .tab-section
+        banner-simple(severity='info')(
+          v-if='!isCloseToDistributionTime'
```
taoeffect: This condition should be inverted:

`!isCloseToDistributionTime` => `isCloseToDistributionTime`

taoeffect (changes requested):
Nice work @pieer! I had a chance to test this locally now.

I noticed a few issues:

### Banner not shown before distribution period

It shows my old code instead:

Screen Shot 2022-08-03 at 6 57 05 PM

Can this be addressed without duplicating the banner code?

### Banner not shown after payments are completed during last 7 days of period

After completing a payment with a local test user, and then changing my computer's clock to be within 7 days before the end of the next period, the banner still didn't show. I found the issue and commented (`!isCloseToDistributionTime` => `isCloseToDistributionTime`).

pieer:
@leihla Here is the banner updated according to greg comment
image

What do you think we should do for that case?


> Nice work @pieer! I had a chance to test this locally now.
> 
> I noticed a few issues:
> 
> ### Banner not shown before distribution period
> It shows my old code instead:
> 
> Screen Shot 2022-08-03 at 6 57 05 PM
> 
> Can this be addressed without duplicating the banner code?
> 
> ### Banner not shown after payments are completed during last 7 days of period
> After completing a payment with a local test user, and then changing my computer's clock to be within 7 days before the end of the next period, the banner still didn't show. I found the issue and commented (`!isCloseToDistributionTime` => `isCloseToDistributionTime`).

pieer:
Also I'm not sure exactly when we should display this banner. Your indication were clear but tricky to implement and @taoeffect seems to have different point of view of when to display.
CC: "Banner not shown after payments are completed during last 7 days of period"

Final though, should we make this consistent somehow with the pills new design?

taoeffect:
Per the discussion on Slack and the calls, this PR was superseded by recently merged PR #1390.

PR #1373, opened 2022-08-03
Move app settings into their own Vuex module

### LINKED ISSUE #1276: Debug logging should show up if the user enables debug logging

### LINKED ISSUE #1344: App settings should use default values when not logged in

snowteamer:
Closes https://github.com/okTurtles/group-income/issues/1344
Closes https://github.com/okTurtles/group-income/issues/1276

### Summary of changes:

- Settings like appLogsFilter, fontSize, theme etc. have been moved to `@model/settings/vuexModule.js`.
- They should now correctly revert to their default values upon successfully logging out.
    If a `?debug` query parameter is present and truthy, then debug- and log-level applications logs will be enabled by default.
    Users can still disable them later via the Application Logs modal though.


taoeffect (changes requested):
Great work @snowteamer!

I tested adjusting some settings, and while most seem to work, changing the system theme from Dark to Light and back while the settings are set to "System" doesn't work. I tried in both Firefox Developer Edition (with preferences set to "System") and also with Safari.

Gruntfile.js
```diff
@@ -113,12 +113,12 @@ module.exports = (grunt) => {
...
   }
 
-  async function deployAndUpdateMainSrc (manifestDir) {
+  async function deployAndGenManifestsJSON (manifestDir) {
     grunt.log.writeln(chalk.underline("Running 'chel deploy'"))
     const { stdout } = await execWithErrMsg(`./node_modules/.bin/chel deploy ./data ${manifestDir}/*.manifest.json`, 'error deploying contracts')
     console.log(stdout)
     const r = /contracts\/([^.]+)\.(?:x|[\d.]+)\.manifest.*data\/(.*)/g
-    const manifests = Object.fromEntries(Array.from(stdout.matchAll(r), x => [`gi.contracts/${x[1]}`, x[2]]))
+    const manifests = Object.fromEntries(Array.from(stdout.replace(/\\/g, '/').matchAll(r), x => [`gi.contracts/${x[1]}`, x[2]]))
```
corrideat: Is this some Windows-related change? Either `/` or `\\` should be valid FS path separators

corrideat (dismissed):

taoeffect:
It's surprising that this is failing repeatedly, even after `master` is merged.

@snowteamer what's your thinking about what might be the problem? Maybe it's a real bug in the PR?

PR #1378, opened 2022-08-08
Fix the issue #1148 - Renaming group during creation results in wrong default group icon

### LINKED ISSUE #1148: Renaming group during creation results in wrong default group icon [$20 bounty]
### Problem

- Open a modal to create a group. The first step is to write the group name.
- Type a group name. For example "Real Madrid". You can see "RM" as the group picture. And click the `Next` button
- The next step is to add a description. Here, click the `Back` button.
- Rewrite the group name again. For example: "FC Barcelona". You can see "FB" as the group picture.
- Do all the steps to create the group and expect the group picture would be "FB". But it is "RM" 

### Solution

Check `groupPictureFile` ephemeral variable in `frontend/views/components/group-creation-steps/GroupName.vue` and track it.

### Bounty

$20 bounty for a clean solution to this (paid in cryptocurrency).

SebinSong:
closes #1148

### DIFF
--- a/frontend/views/components/group-creation-steps/GroupName.vue
+++ b/frontend/views/components/group-creation-steps/GroupName.vue
@@ -69,7 +69,7 @@ export default ({
     }
   },
   beforeDestroy () {
-    if (!this.$assistant.ephemeral.groupPictureFile) {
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
A++ would merge again!

PR #1388, opened 2022-08-19
Archive proposals implemented

### LINKED ISSUE #925: Implement archiving of proposals

taoeffect:
- Closes #925 - implements archiving of proposals
- Bumps `"contractsVersion"` to `0.0.3`
- A new dependency, `cypress-pipe` has been added to help with a specific test scenario where you want a jQuery query to be run multiple times until it produces the result you want. See `.pipe()` related commands in tests.
- Chelonia contracts can now have version-tracked SBP "methods" specified under the `methods:` key. See how this PR uses that feature in `group.js` and `proposals.js`
- Chelonia sandbox has been greatly improved/strengthened. Almost perfect now.

Once proposals are closed, they're moved to a section below open proposals and remain there until page refresh or 1 day has passed, whichever happens first.

Archived proposals can now be seen in a modal by clicking the "Archived Proposals" button. Only archived proposals appear there, open proposals remain on the dashboard only.

Many tests have been rewritten.


frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -150,6 +126,15 @@ export default ({
...
   },
   methods: {
+    onProposalArchived (hashWithProp) {
+      this.ephemeral.archivedProposals.unshift(hashWithProp)
+      // after a day, remove it from the list
+      setTimeout(() => {
+        this.ephemeral.archivedProposals = this.ephemeral.archivedProposals.filter(x => {
+          return x[0] !== hashWithProp[0]
+        })
+      }, 1000 * 60 * 60 * 24) // 1 day
```
snowteamer: Shouldn't these timeouts be cleared before destroying or unmounting the component? I'm afraid too many timers will remain active for example after logging out and in again several times

test/cypress/integration/group-member-removal.spec.js
```diff
@@ -180,9 +180,11 @@ describe('Group - Removing a member', () => {
     getProposalItems().eq(0).within(() => {
       cy.getByDT('typeDescription').should('contain', `Remove userbot-${userId} from the group.`)
       cy.getByDT('voteFor').click()
-      cy.getByDT('statusDescription')
-        .should('contain', 'Proposal accepted')
     })
+    // see explanatory comment in group-proposals.spec.js for this .pipe() thing
+    cy.get('body')
+      .pipe($el => $el.find('[data-test="proposalsWidget"]').children().eq(0).find('[data-test="statusDescription"]'))
```
snowteamer, line 186: Why do we have to start with `cy.get('body').pipe($body => ...)` instead of `cy.getByDT('[data-test="proposalsWidget"]').pipe($widget => ...)`?

frontend/views/containers/proposals/PropositionsAllModal.vue
```diff
@@ -46,20 +49,30 @@ export default ({
...
+      },
+      proposals: []
     }
   }),
+  async mounted () {
+    // const openProposals = Object.entries(this.currentGroupState.proposals)
+    //   .sort((a, b) => {
+    //     return compareISOTimestamps(a[1].meta.createdDate, b[1].meta.createdDate)
+    //   })
+    const key = `proposals/${this.ourUsername}/${this.currentGroupId}`
```
snowteamer, line 61: Currently `/` characters are still allowed in usernames I think. This might precludes interpolation in SBP selectors like here.

snowteamer (commented):
Really nice work!
I just have a thing or two about timers and username interpolation

frontend/views/containers/proposals/ProposalsWidget.vue
taoeffect: Good catch!

test/cypress/integration/group-member-removal.spec.js
taoeffect: I assume you meant `cy.getByDT('proposalsWidget').pipe(...)`, but yeah, I think it might be possible to start with that. The main thing is that `.children()` be inside of the `pipe`, since that's what can change.

frontend/views/containers/proposals/PropositionsAllModal.vue
taoeffect: This isn't an SBP selector, it's just a key into a key-value database, so it's OK.

test/cypress/integration/group-member-removal.spec.js
taoeffect: So it turns out this is necessary for some of the integration tests. Upon trying with `cy.getByDT('proposalsWidget')` the `group-chat.spec.js` test began to fail repeatedly.

I think the issue is that sometimes `proposalsWidget` doesn't exist in the DOM (when there are no proposals). So we need that part to be inside of the `.pipe()`.

test/cypress/integration/group-proposals.spec.js
```diff
@@ -151,6 +151,9 @@ describe('Proposals - Add members', () => {
     // This .pipe() function (from 'cypress-pipe' dependency)
     // let's use apply a should() over a jQuery query repeatedly
     // until it's true.
+    // The reason we start with 'body' is because in some tests
+    // (including in other integration tests), proposalsWidget isn't
+    // attached to the DOM because of the v-if (when there are no proposals)
```
snowteamer, line 156: Good comment :)

snowteamer (approved):
LGTM

PR #1392, opened 2022-08-23
Add 'X-Frame-Options: deny' header to server responses

### LINKED ISSUE #1383: Add 'X-Frame-Options: Deny' header to server
### Problem

We don't want the app to be embedded in frames on other websites as that could lead to click-jacking attacks and other problems.

Ideally, we also don't want the app to be used in in-app webviews - and it's possible [with enough pressure phone manufacturers will begin to start respecting the `X-Frame-Options header`](https://www.holovaty.com/writing/framebust-native-apps/).

### Solution

Add header `X-Frame-Options: Deny` to the backend.

snowteamer:
Closes #1383

### Summary of changes:
- New `X-Frame-Options: deny` header add to every server response for increased security.
 Currently this also includes static asset responses and 404's, but that could be changed later.
 
 ### Additional information:
- This change has been implemented using an [`onPreResponse()`](https://hapi.dev/api/#request-lifecycle) server hook. This way there was no need to alter individual route handlers.

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -52,7 +52,7 @@ hapi.ext({
         request.response.output.headers['X-Frame-Options'] = 'deny'
       }
     } catch (err) {
-      console.warn('[backend] Could not set X-Frame-Options header:', err.message)
+      console.warn(chalk.yellow('[backend] Could not set X-Frame-Options header:', err.message))
     }
     return h.continue
   }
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Nice work @snowteamer! Approved!

Issue #586, opened 2019-06-29
Switch off of nodejs to deno


taoeffect:
### Problem

Node.js is the past. [Deno](https://github.com/denoland/deno) is the future. See also #360 - 'get rid of NPM'.

### Solution

1. Switch to https://github.com/denoland/deno
2. Follow this article (or something more up-to-date) for using NPM packages: https://reactgo.com/deno-npm-modules/
3. If you switch us to TypeScript because of Deno's support, [avoid using these features](https://www.executeprogram.com/blog/typescript-features-to-avoid), and verify that our [`strings` translation tool](https://github.com/okTurtles/strings/blob/master/README.md#help-translate-a-language) still works, as it's designed to work with Flowtype (which is similar but not the same as TypeScript). If you choose TypeScript, and it breaks the tool, update the tool to work with both Flow and TypeScript. Most likely this is too difficult and we'll just stick with Flow - so see if you can get Deno working with Flow instesad. EDIT: If you decide to go with TypeScript, contact @SGrondin for help with this, since he wrote the `strings` tool. (Join our Slack!)
4. Make sure all of the existing features and functionality of the project work the same. Make sure the existing dev-flow works the same. Make sure code-splitting still works for `lazyLoadedView.js` (see [deno with esbuild](https://esbuild.github.io/getting-started/#deno), the note in #1107 about `es-module-shims`, and #1093).
5. Try to make sure the solution will work with sandboxing dependencies too, see issue #1101, which mentions one of 3 possible sandboxing solutions (lavamoat, confine, and endo)

Moar:

- https://dev.to/bogdanned/on-deno-and-the-future-of-node-1l0p
- https://www.stackbuilders.com/news/welcome-deno-does-this-mean-goodbye-to-node-js

Somewhat related to this issue is sandboxing: #319.

taoeffect:
Adding a $5k bounty for closing this issue for non-contractors (meaning, if you don't have a development contract with okTurtles, and you'd like to close this issue, you can work on it, and if we accept your PR, you'll receive the bounty).

Anyone else can add to the bounty too if they'd like by following these simple steps:

1. Donate the amount you'd like to dedicate to this issue [to the okTurtles Foundation](https://okturtles.org/donate/).
2. Make a comment here (or privately to us via Gitter or Slack) the same day letting us know you'd like your donation to be used for a bounty to close an issue, and which issue that is.

taoeffect:
[...]

PR #1395, opened 2022-08-28
WIP - Migrate backend to Deno

snowteamer:
Related to #586

### Summary of changes:
- New Deno backend, using [my fork](https://github.com/snowteamer/pogo) of the Pogo server framework.
  An okTurtles-based fork could be used as well.
- All files in `backend/` and `shared/` have been converted to TypeScript.
- Some Flowtype definitions in `shared/` were still necessary and have been moved to new `types.flow.js` files in their respective folders.
- Grunt tasks `backend:launch` and `backend:relaunch` have been replaced with `deno:start`.
- New `deno:stop` and `flow:stop` tasks, used to avoid dangling process issues.
- New `exec:ts` task for typechecking backend and shared files.

#### Known issues and limitation: 
- Everything but the backend is still based on Nodejs. I think migrating one thing at a time is likely easier.

### Remaining tasks:
- [x] Use Deno's `--watch` flag to implement `npm backend` script equivalent functionality without using Nodemon
- [x] Annotate backend files so they pass typechecking with `deno check`
- [x] Fix Flow errors in shared files
- [x] Replace Mocha by `deno test` for backend unit tests


backend/database.ts
```diff
@@ -0,0 +1,215 @@
...
+        const entry = await sbp('chelonia/db/getEntry', currentHEAD)
+        if (!entry) {
+          console.error(`read(): entry ${currentHEAD} no longer exists.`)
+          chunks[chunks.length] = ']'
+          break
+        }
+        if (chunks.length > 1) chunks[chunks.length] = ','
+        chunks[chunks.length] = `"${strToB64(entry.serialize())}"`
+        if (currentHEAD === hash) {
+          chunks[chunks.length] = ']'
```
taoeffect: Instead of this `chunks[chunks.length]` stuff, could you please preserve the previous behavior as much as possible? E.g. use `chunks.push` and keep the code identical to the previous code, except where changes are required?

taoeffect: The reason I'm making this request is because every single deviation from the original source has to be re-reviewed. The fewer such deviations there are, the better. The only exception to this is boy-scouting - fixing up other people's code / making it cleaner and simpler. I don't think this change counts because it's not making an improvement, and using `.length` to add elements isn't ... what's the word (idiomatic?) for adding elements to an array. In most languages such a thing would throw an exception.

snowteamer: Agreed with using `chunks.push()` - however I had to research and use a new stream API as the current one didn't work in Deno.

taoeffect (changes requested):
Really impressive work @snowteamer! This involved a deep understanding of the differences between Node and Deno.

This is just a preliminary review, nothing serious, two change requests:

- ~~Use `.push` instead of `.length` for streaming APIs, and in general get them to match the original code as much as possible~~ EDIT: oops, seems like you already did this!
- Go through the logging stuff and try to match the original logging as closely as possible

Gruntfile.js
```diff
@@ -628,27 +643,30 @@ module.exports = (grunt) => {
...
       grunt.log.writeln('Quitting dangling child!')
-      child.send({ shutdown: 2 })
+      child.kill()
     }
-    // Stops the Flowtype server.
-    exec('./node_modules/.bin/flow stop')
+    // Make sure to stop the Flowtype server in case `flow:stop` wasn't called.
+    exec('./node_modules/.bin/flow stop', () => {
+      grunt.log.writeln('Flowtype server stopped in process exit handler')
+    })
```
taoeffect, line 693: Just a note: Sebin recently updated the Gruntfile to make sure that `grunt pin` runs esbuild first. Check https://github.com/okTurtles/group-income/pull/1402/files for what to bring back here

.travis.yml
```diff
@@ -9,6 +9,14 @@ env:
 branches:
   only:
     - master # https://github.com/okTurtles/group-income/issues/58
+before_install:
+  # https://blog.travis-ci.com/2021-02-01-rundeno
+  - pwd
+  - curl -fsSL https://deno.land/x/install/install.sh | sh
```
corrideat: Granted this is only for Travis, but I'm not a big fan of `curl | sh`.

backend/server.ts
```diff
@@ -0,0 +1,118 @@
...
+  createNotification,
+  createServer,
+  isUpgradeableRequest,
+  NOTIFICATION_TYPE
+} from '~/backend/pubsub.ts'
+import { router } from './routes.ts'
+
+import { GIMessage } from '../shared/domains/chelonia/GIMessage.ts'
+
+const { default: { version } } = await import('~/package.json', {
```
corrideat: Does this really need to be an async import?

frontend/controller/namespace.js
```diff
@@ -1,11 +1,12 @@
+// @no-flow
 'use strict'
 
 import sbp from '@sbp/sbp'
 import { handleFetchResult } from './utils/misc.js'
 
 // NOTE: prefix groups with `group/` and users with `user/` ?
 sbp('sbp/selectors/register', {
-  'namespace/register': (name: string, value: string) => {
+  'namespace/register': (name, value) => {
```
corrideat: Is there a good reason to remove the type annotations here? Granted, the types _are_ misleading because sbp may be called with the wrong types, so a type assertion might be preferrable (or adding type checking to sbp somehow; I never got that deep into TypeScript, but Redux do something like that). If the reason is that this is a JS file, TypeScript supports JSDoc, see <https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html>

frontend/controller/namespace.js
```diff
@@ -14,9 +15,9 @@ sbp('sbp/selectors/register', {
       }
     }).then(handleFetchResult('json'))
   },
-  'namespace/lookup': (name: string) => {
+  'namespace/lookup': (name) => {
```
corrideat: ditto

frontend/controller/utils/misc.js
```diff
@@ -1,7 +1,8 @@
+// @noflow
 'use strict'
 
-export function handleFetchResult (type: string): ((r: any) => any) {
-  return function (r: Object) {
+export function handleFetchResult (type) {
```
corrideat: Ditto about removing type annotations

frontend/model/contracts/shared/giLodash.js
```diff
@@ -2,36 +2,36 @@
 // https://github.com/lodash/babel-plugin-lodash
 // additional tiny versions of lodash functions are available in VueScript2
 
-export function mapValues (obj: Object, fn: Function, o: Object = {}): any {
+export function mapValues (obj, fn, o = {}) {
```
corrideat: Ditto about types. Granted this is a tricky case because it relies on JS internals in a way that is difficult to express with TS. I'd perhaps use generics here with `Record`.

frontend/model/contracts/shared/giLodash.js
```diff
@@ -2,36 +2,36 @@
...
   for (const key in obj) { o[key] = fn(obj[key]) }
   return o
 }
 
-export function mapObject (obj: Object, fn: Function): {[any]: any} {
+export function mapObject (obj, fn) {
```
corrideat: Ditto about types. I won't comment on the rest of this file or other type removals, but the comment applies in general.

scripts/process-shim.ts
```diff
@@ -0,0 +1,13 @@
+const process = {
```
corrideat: What's the purpose of this? I understand what it does, but I don't see why it's needed. Is it to keep the app running on both Node.js and Deno? Or because not having it would be a significant rewrite? Or something else? I don't like the idea of polluting the global namespace.

shared/domains/chelonia/internals.ts
```diff
@@ -158,14 +164,14 @@ export default (sbp('sbp/selectors/register', {
       }
     }
   },
-  'chelonia/private/in/processMessage': async function (message: GIMessage, state: Object) {
+  'chelonia/private/in/processMessage': async function (this: CheloniaInstance, message: GIMessage, state: any) {
```
corrideat: This is an optional comment (as I understand migration can be done gradually), but I'd advise against using `any`. In most cases `unknown` is preferrable.

shared/domains/chelonia/internals.ts
```diff
@@ -208,24 +214,26 @@ export default (sbp('sbp/selectors/register', {
     if (config.preOp) {
       processOp = config.preOp(message, state) !== false && processOp
     }
-    if (config[`preOp_${opT}`]) {
-      processOp = config[`preOp_${opT}`](message, state) !== false && processOp
+    if (`preOp_${opT}` in config) {
```
corrideat, line 256: This should be done as a type assertion, same as the logic below.

shared/functions.ts
```diff
@@ -4,14 +4,17 @@ import multihash from 'multihashes'
 import nacl from 'tweetnacl'
 import blake from 'blakejs'
 
-// Makes the `Buffer` global available in the browser if needed.
-if (typeof window === 'object' && typeof Buffer === 'undefined') {
-  // Only import `Buffer` to hopefully help treeshaking.
-  const { Buffer } = require('buffer')
+import { Buffer } from 'buffer'
+
+if (typeof window === 'object') {
```
corrideat: Probably this whole block can be simplified as `self.Buffer = Buffer`.

frontend/controller/namespace.js
snowteamer: I had to remove Flow annotations here and also in giLodash because Deno choked on them. Converting these files to .ts was another option, but then the frontend would be in a mixed state where most files are .js while a few are .ts.

I can add JSDoc comments for now, but I think we'll still want to re-add "real" annotations later.

scripts/process-shim.ts
snowteamer: It helps to reduce the amount of changes, given many files are using `process.env` in conjunction with esbuild's `--define` option. When replacing it with `Deno.env.get()`, things become a bit tricky as `--define` doesn't work with function calls AFAIK (https://esbuild.github.io/api/#define)

shared/domains/chelonia/internals.ts
snowteamer: Yea I'm working on removing instances of `any` and use the `@typescript-eslint/no-explicit-any` rule to make sure it is generally no longer used.

shared/functions.ts
snowteamer: Do you mean just using Deno's `Buffer` instead of the third-party `buffer` module? That would be nice, but the frontend is also using this file, and there is no browser `Buffer` API yet. :/

backend/server.ts
snowteamer: Likely not. Good catch.

.travis.yml
snowteamer: Neither am I tbh. I think I should at least replace that `x` with a fixed version number. Do you know a better approach?

taoeffect: @snowteamer using a package manager? Sometimes those have better security than `curl | sh`

snowteamer: According to Deno's official installation [guide](https://deno.land/manual@v1.26.1/getting_started/installation), alternatives include using `nix-shell` or manually downloading the Deno executable from GitHub. I think I'd favor the latter one.

taoeffect: Yeah latter sounds better to me too 👍

corrideat: Usually, I'd see what the script does and then try to replicate it as a last resort if a package manager isn't an option. I'd also suggest checking the hash / signature of the downloaded file. But again, since this is for Travis perhaps it doesn't need to be especially secure as I don't see much damage happening besides being billed extra hours or maliciously breaking / fixing builds. The reason for a better approach would be to have reference build instructions.

scripts/process-shim.ts
corrideat: esbuild shims can probably address the lack of support for `Deno.env.get`, but yeah, I kind of get the idea of needing this

shared/functions.ts
corrideat: I assumed (but didn't test) that `globalThis` and `self` are the same (and `widnow` and `self` are the same), and hence the `if` is not needed by using `self`. Re. browser `Buffer` support, it's unlikely to ever happen because it's non-standard. The standard way is `Uint8Array` (`Buffer` is this plus some helpers for encoding and decoding). It might be worth it to move away from `Buffer` later but probably not relevant to this PR.

snowteamer: You were right, thank you!

taoeffect:
Wonder if these updates to Deno are relevant? https://deno.com/blog/v1.35

taoeffect:
This PR is an admirable attempt on @snowteamer's part to get us there. There are too much merge conflicts at this point though, and too many changes that need to be done for us to be able to merge this. I think it should also be done as part of splitting out Chelonia and the TypeScript conversion, which is something @corrideat plans to do, so closing. Thanks for your attempt at this @snowteamer!

PR #1402, opened 2022-09-07
Issue #1352 - Make `grunt pin` run esbuild first

### LINKED ISSUE #1352: Make `grunt pin` run esbuild first [$40 bounty]
### Problem

Currently you must make sure that `grunt dev` has been run before `grunt pin`, because `grunt pin` takes the contracts that are built and saved to `dist/contracts/` and moves them into a version-stamped folder in `contracts/`.

However, if a change has been made to the code while `grunt dev` isn't running, then `grunt pin:<version>` will miss that change and the old version that was put in `dist/contracts/` will be placed into `contracts/` - this is surprising behavior.

### Solution

Have `grunt pin:<version>` run the esbuild task first, but make sure that the esbuild task doesn't stay running in the background watching for changes.

### Bounty

SebinSong:
closes #1352 

@taoeffect 
I chose to run `build` because all the linters still need to be executed to clear possible syntax errors in the changed files but still wasn't 100% sure it is enough as the preparation of `pin` task.
please let me know of your opinion.

cheers,

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -455,8 +455,8 @@ module.exports = (grunt) => {
     cypress[command](options).then(r => done(r.totalFailed === 0)).catch(done)
   })
 
-  grunt.registerTask('pin', async function (version) {
-    if (typeof version !== 'string') throw new Error('usage: grunt pin:<version>')
+  grunt.registerTask('_pin', async function (version) {
+    // a task internally executed in 'pin' task below
     const done = this.async()
     const dirPath = `contracts/${version}`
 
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -481,6 +481,13 @@ module.exports = (grunt) => {
     done()
   })
 
+  grunt.registerTask('pin', function (version) {
+    if (typeof version !== 'string') throw new Error('usage: grunt pin:<version>')
+
+    grunt.task.run('build')
+    grunt.task.run(`_pin:${version}`)
```
SebinSong, line 488: created an internal task `_pin` to make sure `build` runs first before executing `pin`(now `_pin`).
(adding `grunt.task.run('build')` inside the original `pin` custom task actually queues this(`build`) task at the end of the current task which is not what we want. but pls let me know if there is a better way)

taoeffect (approved):
Seems to work! Great job @SebinSong!

Issue #1332, opened 2022-07-07
Design for displaying inactive users


taoeffect:
### Problem

It's a problem for the health of a group if pledgers are not making their pledges, and if monetary receivers are not logging in.

### Solution

Perhaps somewhere near the support history of pledges completed (#1286) design a list of inactive that users, something that contains the following information:

```
- Mark (monetary receiver, last login: <date>)
- Joe (pledger, last pledge: <date>, last login: <date>)
- ...
```

leihla:
This has been designed an implemented on the Dashboard widgets.

Issue #1286, opened 2022-05-26
Need a design to show percentage of TODOs that were actually completed in Support History


taoeffect:
### Problem

The Support History widget currently shows how close the group is to reaching its mincome goal each month. It doesn't show how well the group did in actually completing the payments they were instructed to make.

### Solution

Design something in the Support History widget to show for each month the % of the total amount that should have been sent.

Could simply be a second row of bars.

leihla:
This has been designed already https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=23342%3A251404&t=OYjm7ACensmK2Bix-1

PR #1404, opened 2022-09-10
Dashboard tab Skeleton

pieer:
Implement design solution found in "Design for displaying inactive users #1332" and "Need a design to show percentage of TODOs that were actually completed in Support History #1286"


frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -0,0 +1,45 @@
+<template lang='pug'>
+div(:class='isReady ? "" : "c-ready"')
+  i18n.is-title-3(
+    tag='h2'
+  ) Support history
+
+  i18n(tag='p') Percentage of the goal reached by the group.
```
taoeffect: Let's change this to:

> Percentage of the group income goal reached by the group.

I'll make a similar comment on Figma.

frontend/views/containers/contributions/TodoHistory.vue
```diff
@@ -0,0 +1,45 @@
+<template lang='pug'>
+div(:class='isReady ? "" : "c-ready"')
+  i18n.is-title-3(
+    tag='h2'
+  ) TODO history
+
+  i18n(tag='p') Percentage of the payments TODO reached by the group.
```
taoeffect: How about the following?

> Percentage of payments completed by those pledging.

I'll make a similar comment on Figma.

taoeffect (changes requested):
Excellent work @pieer! 💪 

Per the weekly call, the "Activity" Tab contains several rows that do not belong there because they are older designs that were rejected.

Check Figma for the latest and correct designs, as @leihla should have removed all the ones we're not using. If you have any doubts as to what should go there, ask her or myself.

Also, there's an error when loading the activity tab:

Screen Shot 2022-09-13 at 2 40 09 PM

Silver-IT (dismissed):
I can't find any issues here. Awesome~~~

frontend/views/components/graphs/BarGraph.vue
```diff
@@ -0,0 +1,185 @@
...
+      if (percentage < 0.6) return 'danger'
+      if (percentage < 1) return 'warning'
+      return 'success'
+    },
+    isLow (percentage) {
+      return percentage <= 0.2
+    }
+  },
+  filters: {
+    toPercent
```
SebinSong, line 50: great to know we can use vue filter this way.

SebinSong:
Looks good for me. approving it.

Silver-IT (approved):
New updates are also approved. Great work, @pieer

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -0,0 +1,160 @@
...
+                  div Ken M
+                  div Ines de Castro
+
+              //- Todo: discuss if tooltip better than toggle
+              //- i18n.link(
+              //-   tag='button'
+              //-   :aria-label='L("See members")'
+              //-   :args='{  members: 3 }'
+              //-   @click='openModal("todoSeeMembers")'
+              //- ) {members} members
```
taoeffect, line 36: So on Figma these are displayed as a toggle, and in this PR they are implemented as a tooltip:

Screen Shot 2022-09-27 at 5 01 18 PM

@pieer and @leihla - can you guys come to an agreement about what this should be?

Also:

- please consider what might look better when the list is 25+ members long
- note that on Figma the designs include more information, such as how long the streak is. Leihla, do you feel this extra info should be there? If so, do any of the other fields also need extra info? Because currently all of these are implemented simply as a list of names in a tooltip.

taoeffect (commented):
Looking really good @pieer!

I left a question for you and @leihla to come to a decision over. After that's figured out I think this is good to merge. Maybe Leihla may also want to try it out on her end too.

frontend/views/containers/dashboard/GroupMembersActivity.vue
leihla: @taoeffect @pieer I don't mind that it shows in a tooltip instead of the expandable, I was following the same pattern we have on the contributions page. 

Regarding the information of duration of the streak, I do think is important to show, but that can be shown on the tooltip as well. 

@pieer any reason you prefer the tooltip?

pieer: Few reasons:
- common pattern on the web to have more information on link with tooltip.
- quicker to get the info than with click
- less moving element

I can had the streak, yes

I don't think that the tooltip would look great with 25 members though

leihla: @pieer I think we can keep the tooltip for now. If we have that big of a list we'll need to change this pattern.

taoeffect (approved):
Excellent work @pieer! Approved!

Issue #1393, opened 2022-08-24
Redesign Voting System Step during Group Setup to Remove Disagreement Rule


taoeffect:
### Problem

There are a few issues with asking users to choose between the percentage rule and the disagreement rule during setup:

1. Most users won't understand the difference, and will find it confusing. This is unnecessary friction during setup. Anyone who really wants the disagreement rule can choose it later in the group settings if necessary.
2. Our implementation of the disagreement rule isn't complete anyway, there are a couple of bugs related to it: #1295 #877

### Solution

All we need is for users to choose the threshold for the percentage rule. So create a new design that's simpler, and just for the percentage rule for that step.

dotmacro:
To add (in agreement): 
Most commonly, I've seen the disagreement rule implemented as "consensus" (i.e. 100% threshold) or "consensus minus 1". Because groups will be small, any users who really want the disagreement rule could also choose a % threshold that approximates the same result.

leihla:
Update Group Creation flow to simplify voting system:

- Remove disagreement number option 
- Update copy 
- Remove disagreement number option from group settings


![5 1  Group Creation](https://user-images.githubusercontent.com/158178/189693930-fed3f617-befe-49e4-8885-17b09fb0e482.png)
![5 2  Group Creation](https://user-images.githubusercontent.com/158178/189693932-4fae4e13-b61a-471f-a506-d0e47cadcbdc.png)

PR #1407, opened 2022-09-14
Remove disagreement rule from group creation step

### LINKED ISSUE #1405: Remove Disagreement Rule from Group Creation setup
### Problem

See #1393 for details on both problem and solution.

Summary: there are problems with having the disagreement rule in the group setup, and there are currently also problems with the disagreement rule implementation.

### Solution

Implement the new designs for #1393.

SebinSong:
closes #1405


test/cypress/integration/group-settings.spec.js
```diff
@@ -119,15 +119,6 @@ describe('Group Voting Rules', () => {
...
 
-  it('user1 creates a group with rule "disagreement" of "2"', () => {
-    cy.giCreateGroup('groupDis_2', { ruleName: 'disagreement', ruleThreshold: 2 })
-
-    verifyRuleSelected('disagreement', {
-      status: '2',
-      ruleAdjusted: false
-    })
-  })
-
```
taoeffect: Could you either comment this out or add `it.skip` to the test?

We don't want to remove support for it completely. Just for now and just in the group creation step. We might allow users to switch to it manually themselves.

historical/test/commands-cyBypassUI.js
```diff
@@ -205,8 +205,7 @@ Cypress.Commands.add('giCreateGroup', (name, {
     cy.getByDT('nextBtn').click()
 
     cy.getByDT('rulesStep').within(() => {
-      const threshold = ruleName === 'percentage' ? ruleThreshold * 100 : ruleThreshold
-      cy.getByDT(ruleName, 'label').click()
```
taoeffect: Please allow the historical files to preserve history :P
(They should not be modified)

taoeffect (changes requested):
Excellent work @SebinSong!

I left a few comments, and also while testing this locally I realized that we need some design changes too, so I left a few comments on Figma for @leihla:

- https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=23903:270549#266397148
- https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=23903:270549#266397218

So before we merge this, let's also wait for the design updates to be made.

historical/test/commands-cyBypassUI.js
SebinSong: done it in this [commit](https://github.com/okTurtles/group-income/pull/1407/commits/1478f02b565169320ea860dc549665f64aff2401)

test/cypress/integration/group-settings.spec.js
SebinSong: done it in this [commit](https://github.com/okTurtles/group-income/pull/1407/commits/1478f02b565169320ea860dc549665f64aff2401)

leihla:
@SebinSong just reviewed this PR and it is indeed implemented as figma was, but as per Greg's comments, figma was updated to remove the icon and outside box, you can see it here: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=24083%3A269896

Screenshot 2022-09-26 at 16 22 38
Screenshot 2022-09-26 at 16 22 44


Also, not sure if is part of this PR, but we need to remove that option from the group settings as well, as you can see here: 
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=12717%3A2816
Screenshot 2022-09-26 at 16 23 59

SebinSong:
@leihla Thanks, will make the change per review.


@taoeffect 
regarding leihla's below comment,

> Also, not sure if is part of this PR, but we need to remove that option from the group settings as well, as you can see here:

are we completely removing the disagreement-rule implementation from the app here or is it that we are only hiding it from the group creation step but still let users be able to switch to this rule setup when they want?

am asking for your confirmation because you mentioned below i[n this ticket](https://github.com/okTurtles/group-income/issues/1393#issue-1350140343)

> Most users won't understand the difference, and will find it confusing. This is unnecessary friction during setup. **Anyone who really wants the disagreement rule can choose it later in the group settings** if necessary.

thanks,

taoeffect:
@SebinSong 

> are we completely removing the disagreement-rule implementation from the app here or is it that we are only hiding it from the group creation step but still let users be able to switch to this rule setup when they want?

To confirm, what I wrote in the issue is correct - we don't want to remove it completely, just hide it for now.

SebinSong:
@taoeffect Thanks, I won't make any change to the group-settings page then.

SebinSong:
Update for removing disagreement-rule from group-settings page has been made too, according to @taoeffect 's Slack comment below.

> is the disagreement rule still visible in the settings? If so can you just hide that section for now, or better yet, disable it and make it greyed out? It’s not fully functional at the moment

frontend/views/containers/group-settings/GroupRulesSettings.vue
```diff
@@ -1,42 +1,35 @@
...
+      type='button'
+      data-test='changeRule'
+      @click='openVotingProposal'
+    ) {{ groupShouldPropose ? L('Propose change') : L('Change') }}
+
+  banner-simple.c-banner(
+    serverity='info'
+    data-test='ruleAdjusted'
+    v-if='groupShouldPropose && votingRuleAdjusted()'
+  ) {{ votingRuleAdjusted() }}
```
taoeffect: @SebinSong I'm sorry I guess I wasn't clear enough. We don't want to *rewrite* the code so that this isn't here anymore, because we might want to add this feature back in the future. We just want to comment out the old code that shows that section, or disable it with a `v-if='false'`.

That way, whenever we do fix the issues with the disagreement rule, it is simple to re-enable it.

taoeffect (changes requested):

frontend/views/containers/group-settings/GroupRulesSettings.vue
```diff
@@ -58,6 +71,12 @@ export default ({
       // because the group disappears. it's not a big deal though
       return this.groupProposalSettings() || {}
     },
+    votingRulesSorted () {
+      // NOTE: temporarily hiding DISAGREEMENT_RULE from the settings.
```
SebinSong, line 75: @taoeffect just reverted the code back to what they were and hided the 'disagreement-rule' by making this change here with the comment explaining the context. Thought this is the way that requires the least amount of code change, hence making things easier when we want to re-enable it. but pls let me know if you have other thoughts.
cheers,

frontend/views/containers/group-settings/GroupRulesSettings.vue
```diff
@@ -58,6 +71,12 @@ export default ({
...
+      // TODO: once disagreement-rule implementation is ready, put this back to
+      //       return this.proposalSettings.rule === RULE_DISAGREEMENT ? [RULE_DISAGREEMENT, RULE_PERCENTAGE] : [RULE_PERCENTAGE, RULE_DISAGREEMENT]
+      return [RULE_PERCENTAGE]
+    },
```
SebinSong, line 80: @taoeffect just reverted the code back to what they were and hided the 'disagreement-rule' by making this change here with the comment explaining the context. Thought this is the way that requires the least amount of code change, hence making things easier when we want to re-enable it in the future. but pls let me know if you have other thoughts.
cheers,

taoeffect: @SebinSong Thanks - however this is insufficient to hide the setting from the user.

Now, users will just be confused as to what's going on. There's a setting that's broken completely, they can't even switch to it. The settings need to look like the Figma design (only show the one voting system we have).

When I said "hide" and "comment out" - I meant it. :P

Please keep this old code around, but the user must not see this alternative voting system until we get it working.

taoeffect (changes requested):
Review complete.

frontend/views/containers/group-settings/GroupRulesSettings.vue
taoeffect: Oops, so ... again it seems like the first time I reviewed this I didn't get the latest updates some how. Now it is working as expected.

Something odd going on with my `git pr-id` script. Ignore what I wrote above.

frontend/views/components/VotingRulesInput.vue
```diff
@@ -46,7 +46,7 @@ export default ({
   data: () => ({
     config: {
       [RULE_PERCENTAGE]: {
-        sliderLabel: L('What percentage of members need to agree?'),
+        sliderLabel: L('What percentage of members need to agree to a proposal?'),
```
taoeffect: One minor change request here (I made the same change request on Figma just now for @leihla as well): can we change this to:

> What percentage of members need to agree to pass a proposal?

SebinSong:
the PR has been updated with,
- hiding 'active' tag.
- copy update to 'What percentage of members need to agree to pass a proposal?'

taoeffect (approved):
Approved! Excellent work @SebinSong!! 🎉

PR #1408, opened 2022-09-14
Cache namespace lookups in Vuex

### LINKED ISSUE #1141: Cache 'namespace/lookup' results
### Problem

In some areas of the app, for example AvatarUser and elsewhere, we lookup the username => identity contract mapping, and this lookup could be done multiple times for the same user.

This is a problem for two reasons:

1. It kinda breaks the offline-first approach we want to take with the app
2. That lookup should never change, so it's safe to speed things up by caching that value, not just for performance reasons but for security reasons too (preventing the server from changing it on us)

### Solution

Create a cache in the local state for this and transparently use it everywhere it's needed in a simple fashion. Keep things DRY.

snowteamer:
Close #1141

### DIFF
--- a/frontend/controller/namespace.js
+++ b/frontend/controller/namespace.js
@@ -12,10 +12,17 @@ sbp('sbp/selectors/register', {
       headers: {
         'Content-Type': 'application/json'
       }
-    }).then(handleFetchResult('json'))
+    }).then(handleFetchResult('json')).then(result => {
+      sbp('state/vuex/state').namespaceLookups[name] = value
+      return result
+    })
   },
   'namespace/lookup': (name: string) => {
     // TODO: should `name` be encodeURI'd?
+    const cache = sbp('state/vuex/state').namespaceLookups
+    if (name in cache) {
+      return cache[name]
+    }
     return fetch(`${sbp('okTurtles.data/get', 'API_URL')}/name/${name}`).then((r: Object) => {
       if (!r.ok) {
         console.warn(`namespace/lookup: ${r.status} for ${name}`)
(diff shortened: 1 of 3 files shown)


frontend/controller/namespace.js
```diff
@@ -12,10 +12,17 @@ sbp('sbp/selectors/register', {
       headers: {
         'Content-Type': 'application/json'
       }
-    }).then(handleFetchResult('json'))
+    }).then(handleFetchResult('json')).then(result => {
+      sbp('state/vuex/state').namespaceLookups[name] = value
```
taoeffect, line 16: Remember to always use `Vue.set` when setting with Vue state on objects.

frontend/controller/namespace.js
```diff
@@ -25,6 +32,11 @@ sbp('sbp/selectors/register', {
         return null
       }
       return r['text']()
+    }).then(value => {
+      if (value === null) {
+        return null
+      }
+      return (cache[name] = value)
```
taoeffect, line 39: How about:

```js
if (value) {
  Vue.set(cache, name, value)
}
return value
```

A bit cleaner IMO

frontend/model/state.js
```diff
@@ -76,6 +77,9 @@ sbp('sbp/selectors/register', {
     if (!state.chatRoomUnread) {
       state.chatRoomUnread = {}
     }
+    if (!state.namespaceLookups) {
+      state.namespaceLookups = Object.create(null)
```
taoeffect, line 81: Nice job remembering 👍

taoeffect (changes requested):
Nice work @snowteamer!

Left a few comments. For making the changes, could you please do them in a branch of this repo, and close this PR and re-open from that branch so that if another test failure occurs, we can see why in Cypress dashboard?

PR #1409, opened 2022-09-15
Cache namespace lookups in Vuex state

### LINKED ISSUE #1141: Cache 'namespace/lookup' results

snowteamer:
Closes #1141

### DIFF
--- a/frontend/controller/namespace.js
+++ b/frontend/controller/namespace.js
@@ -1,6 +1,7 @@
 'use strict'
 
 import sbp from '@sbp/sbp'
+import Vue from 'vue'
 import { handleFetchResult } from './utils/misc.js'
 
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Excellent work @snowteamer!

Issue #1413, opened 2022-09-17
Multiple UI bugs in dashboard


taoeffect:
#### Bug 1: Safari can't scroll to the bottom of the dashboard

![a1](https://user-images.githubusercontent.com/138706/190874027-652e82df-837f-43f1-b5f0-2332c3ed8839.jpeg)

You must test this on Safari (and preferably on an iPhone) to see if it's fixed, because it doesn't show up in Responsive Mode on Firefox.

#### Bug 2: Right Sidebar doesn't display when scrolled to bottom of dashboard

EDIT: @SebinSong fixed Bug 2 in #1414. But Bug 1 still needs a fix!

![b1](https://user-images.githubusercontent.com/138706/190874031-3ef9a652-3df9-4b74-88a2-1d3f5dd8b27d.jpeg)

This actually is testable in Firefox responsive mode. You must first scroll down to the bottom and then make the right sidebar appear.

### Solution

Fix both!

EDIT: @SebinSong fixed Bug 2 in #1414. But Bug 1 still needs a fix!

PR #1414, opened 2022-09-18
fix the bug where side-bar is cropped at the bottom

SebinSong:
fix below issue in mentioned in #1413 

**Bug 2: Right Sidebar doesn't display when scrolled to bottom of dashboard**


taoeffect (changes requested):
I tested this by creating a group, adding income details, and in FF dev edition, scrolling to the bottom while in responsive mode.

Didn't seem to fix the issue- in fact the sidebar seemed to not appear at all:

Screen Shot 2022-09-24 at 2 34 07 PM

There was also using `Element.setCapture() is deprecated` in the console (see screenshot).

It was only after I scrolled up did I realize anything had happened at all:

Screen Shot 2022-09-24 at 2 34 24 PM

SebinSong:
Hi @taoeffect, 
Thanks for review. 
I did test my work again this is what it looks like for me while testing.



Did you try them in the correct feature branch which is `sebin/task/sidebar-cropped-issue`? 

I did see the same warning displayed in devtool console too (`Element.setCapture() is deprecated`)but don't see the method being used anywhere in the app code-base. :)

Thanks,

taoeffect (approved):
It works now! Not sure what happened! Thanks @SebinSong - great work!

Issue #1339, opened 2022-07-09
Updated Payments - Multiple Payment Methods


leihla:
## Overview

**Problem**
The current implementation only accounts for users to record manual payments that are performed out-of-platform.

 **Solution**
Add Lightning Network payments and update payments flow.

**What changed:**

- Distribution date (we now have a date where distributions are calculated and payments are expected to be done)
- Multiple payment methods (Manual + Lightning Network)

---

**Updated Flows:**

## 1. Updated - Payments with Distribution date + multiple payment methods

- [ ] Add distribution banner / chip
- [ ] Add Payment Method filters
- [ ] Add checkboxes for each payment row
- [ ] Add Accepted Methods to the "To Do" Payments Table
<img width="1440" alt="Screenshot 2022-07-09 at 11 23 28" src="https://user-images.githubusercontent.com/158178/178101703-214840fe-f663-4aa1-8233-ad88ac5ef0cf.png">

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23249%3A231771

---

## 2. New - Send Payments with Lightning Network

- [ ] Select Lighning Network Filter and check users 
- [ ] New LN Payment Modal - includes QR code for scanning payment
- [ ] New Donation fee - it's added to the transaction

<img width="1446" alt="Screenshot 2022-07-09 at 11 27 52" src="https://user-images.githubusercontent.com/158178/178101847-1d7aedaa-56be-4697-90bb-a4575f021d55.png">

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23123%3A224576

---

## 3. Updated - Record Manual Payments

- [ ] Update TODO table

<img width="1433" alt="Screenshot 2022-07-09 at 11 30 03" src="https://user-images.githubusercontent.com/158178/178104305-278224ae-74a7-4101-82c3-4cd209c6350e.png">

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23189%3A249357

---
## 4. New - Payment details Modal for Lightning

- [ ] Create Payment Details modal for LN Payments with Transaction ID

<img width="717" alt="Screenshot 2022-07-09 at 14 15 44" src="https://user-images.githubusercontent.com/158178/178107494-017c0920-9523-41ce-8f75-a249beea0e99.png">

[...]

PR #1421, opened 2022-09-22
#1339 Update payments page - subtask 5.

### LINKED ISSUE #1428: Payments - Design prompt for users to send thanks for receiving pledges
### Overview

**Problem**
Currently there's no automatic prompt for thanking pledgers who have sent their payments.

**Solution**
Add a simple way for a receiver of a contribution to send a thank you note to the pledger, and for the pledger to view it. 

**What changed:**

For receivers:
- Payments - Received tab show "Send thank you note" option on the more actions button
- Notifications - Add notification for payments received 
- Create thank you modal 

For pledgers:
- Notifications - Add a notification for thank you notes received 

----

**Updated Flows:**

## 1. Payments - Received tab (receivers)

- [ ] Add "Send thank you note" option on the more actions button
- [ ] Once clicked, user should be redirected to 4.Payment (receiving) - Thank you Modal 
Screenshot 2022-10-03 at 15 32 33
Screenshot 2022-10-03 at 15 32 43
Screenshot 2022-10-03 at 15 32 48

---
## 2. Notifications - Add notification for payments received (receivers)

- [ ] When a payment is received, the user should receive a notification
- [ ] Clicking the notification should show the Payment details modal
- [ ] Clicking "Send thanks" on the modal should open the "Thank you Modal"

Screenshot 2022-10-03 at 15 35 00
Screenshot 2022-10-03 at 15 35 11
Screenshot 2022-10-03 at 15 32 43
Screenshot 2022-10-03 at 15 32 48

Figma flow: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23482%3A253597

---
## 3. Notifications - Add a notification for thank you notes received (pledgers)

- [ ] When a thank you note is received, the user should receive a notification
- [ ] Clicking the notification

Screenshot 2022-10-03 at 15 39 24
Screenshot 2022-10-03 at 15 39 32
 should show the Thank you modal

Figma: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23482%3A253599

Related to issue #1333

SebinSong:
working on subtask 5. from #1339 
closes #1428


frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -0,0 +1,121 @@
...
+  methods: {
+    close () {
+      this.$refs.modal.close(0)
+    },
+    submit () {
+      try {
+        sbp('gi.actions/group/sendPaymentThankYou', {
+          contractID: this.$store.state.currentGroupId,
+          data: {
+            fromUser: this.ourUsername,
```
taoeffect: This should use `ourUserDisplayName` instead, which will show the display name (if set) or the username (if not set).

frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -0,0 +1,121 @@
...
+            toUser: this.$route.query.to,
+            memo: this.form.memo
+          }
+        })
+      } catch (err) {
+        console.error('send thank you note error: ', err)
```
taoeffect: Errors must always be displayed to users instead of silently ignored. Think about this from the user's perspective: they click a button and nothing happens. It's very frustrating. You can use either `v-error` to do that, or `alert`.

frontend/model/contracts/group.js
```diff
@@ -521,6 +525,27 @@ sbp('chelonia/defineContract', {
...
       }
     },
+    'gi.contracts/group/sendPaymentThankYou': {
+      validate: objectOf({
+        fromUser: string,
+        toUser: string,
+        memo: string
+      }),
+      process ({ data, meta, hash }, { state }) {
+        Vue.set(state.paymentThankYouNotes, hash, data)
```
taoeffect: By creating a `paymentThankYouNotes` this way - will save and store all payment thank yous forever. That seems excessive to me.

Instead of that, how about we just save the most recent thank-you for each user?

You can do this by creating a 2-dimentional lookup object, similar to how `paymentsFrom` works. You could call it, `thankYousFrom`, and access it like this:

```js
state.thankYousFrom[fromUser]?.[toUser]
```

I recommend also using `vueFetchInitKV` when implementing this. Have a look for example at how the process function for `'gi.contracts/group/payment'` uses it.

Then, for the `linkTo` in the notification, pass in the `fromUser` and `toUser` parameters to the modal.

Also, remember - if you are modifying contracts you need to pin a new version. For now it's OK that you haven't, because @Silver-IT is also modifying them in his PR, and his PR might be merged before yours. So his might be `0.0.5`, and yours might be `0.0.6`.

taoeffect (changes requested):
Nice work @SebinSong!

I haven't tested it yet but I added some comments based on the code.

Also see my question on Slack about `hello.svg`. Unclear to me why that's being updated here since there doesn't seem to be any change? I could be wrong though.

SebinSong:
@taoeffect Worked on your change requests in [this commit](https://github.com/okTurtles/group-income/pull/1421/commits/8207940bf3a73504871e8dac3a95a42ad30d2a13).
Pls let me know if you find any issue there. cheers,

leihla:
@SebinSong I reviewed the pull request and from what was on 1339, it seems that you added everything. As we changed the tabs for consistency, for the user who's receiving mincome, they should also have the "Received" tab and distribution pill and removed the "Received Payments". 

Question, should this account as well for the send thank you note changes or will that be on another PR?

SebinSong:
@leihla yes, 'Send thank you note' feature is part of this PR too.

SebinSong:
@taoeffect 
made a change to the payment table UI for a user who is 'receiving' mincome. (according to @leihla 's change request)

taoeffect (changes requested):
Nice work @SebinSong!

I found three issues in my tests:

### Issue 1 - red late payment label for non-late payments

Screen Shot 2022-09-30 at 2 18 19 PM

These should only appear in red for late payments.

### Issue 2 - Disappearing received tab

1. Create 2 users, one pledging (u1) and one receiving (u2)
2. u1 pays u2 everything for the month.
3. Next u2 switches to pledging.
4. Run `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` again to creating a new distribution date.

Suddenly there only 2 tabs for u2 - TODO and COMPLETED. The RECEIVED tab is missing.

### Issue 3 - Missing reminder notification to send "thank you"

Each time a user receives a payment from another user, they should receive a notification reminding them to send a thank you to that user. And when they click this notification, it should bring up the modal to send the thank you to that user.

Screen Shot 2022-09-30 at 2 24 52 PM

See the designs here for details: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=24085%3A275535

SebinSong:
@taoeffect Thanks for review and I have updated the PR with the work for issue 1. & 3.

I investigated the issue 2. and it doesn't seem be a bug brought in by this PR but rather an expected behaviour.

(Please correct me if I'm wrong about anything in below)
As you might be well aware too, our current group state records payment history in `paymentsByPeriod` object by setting the key to a date ISO string of the corresponding distribution period and the time stamp is currently as detailed as the milli-sec level.

"as detailed as the milli-sec level" here seems to cause the problem.
as soon as `gi.actions/group/forceDistributionDate` selector is processed, the distribution date of the group is reset to 'now'([reference](https://github.com/okTurtles/group-income/blob/17f6dd4adbc02d44f0c106d246bf30251889729f/frontend/model/contracts/group.js#L1026)) but that actually ends up creating two separate periods with the same 'date'(as you can see in the screenshots below), because certain period of time(can be as short as <= 1 min if going through the steps you described quickly) has passed since we first executed `gi.actions/group/forceDistributionDate` previously and the new ISO string takes that time into account.



But if it's that we only use `gi.actions/group/forceDistributionDate` for bebugging & testing purpose during development, (I think) the users will never encounter this situation, so we don't need to worry about this dual periods with the same date 
 issue.

But pls let me know if you think otherwise and further change is required.
Thanks,

taoeffect:
@SebinSong 

> But if it's that we only use gi.actions/group/forceDistributionDate for bebugging & testing purpose during development, (I think) the users will never encounter this situation, so we don't need to worry about this dual periods with the same date

Sorry, I'm not following what you're saying here.

Rather - I do understand what you're saying, but I don't understand why you're bringing it up in this situation.

Issue 2 is not about any issue with periods being on the same day, it's about users losing information about the payments they've received after switching to becoming a pledger. That doesn't have anything to do with the tabs issue...

> I investigated the issue 2. and it doesn't seem be a bug brought in by this PR but rather an expected behaviour.

I think we should show all 3 tabs for users who switched from being a receiver to being a pledger, so that users don't forget the payments they once received in case they want to look them up for their accounting purposes. Can you show all 3 tabs in this case? I do realize it's not explicitly mentioned in #1339, but it's a payments related issue, and we should take this opportunity to fix this issue if it's an easy fix. Is it not an easy fix? If so I can open a separate issue.

Feel free to reply here or on Slack.

SebinSong:
Hi @taoeffect 
Thanks for review and sorry I think my previous comment lacks some detailed explanation on why the periods on the same day has to do with this issue :) Let me try my best to describe what happens step by step below..!

**1.** 
Currently the `Payments.vue` page component determines to display `RECEIVED` and `COMPLETED` tabs based on whether the corresponding field (`sent` / `received` / `todo`) in `ourPayments` Vuex getter has >= 0 of the payment data (reference: [`tabItems` computed prop](https://github.com/okTurtles/group-income/blob/b03050c6cc6d419d4f8d23d3e8b0f8664b73fbef/frontend/views/pages/Payments.vue#L237)).

**2.**
The payment data of each field (`sent` / `received` / `todo`) in `ourPayments` are determined by taking two of the most recent distribution periods, and (**2-1.**)this is done by using `ourPaymentsReceivedInPeriod` and `ourPaymentsSentInPeriod` getters in `group.js` contract.



**3.**

**2-1.** works just fine in most cases but once the group distribution date is updated by `gi.actions/group/forceDistributionDate` to a different timestamp on the same day, these two getters don't fetch the data that we expect to see and here is why.

it's not that the previous payment history disappears from the vuex store. below is a screenshot of `paymentsByPeriod` vuex getter after `gi.actions/group/forceDistributionDate` is executed, and as you can see they are still there.



but current period stamp has now been updated to something other than "2022-10-05T05:20:16.127Z" (let's say, "2022-10-05T05:**21:00.000**Z") and likewise, the previous period stamp will now be calculated to "2022-09-08T05:**21:00.000**Z" in wherever it is used.
This results in the two getters(`ourPaymentsReceivedInPeriod` and `ourPaymentsSentInPeriod`) not being able to get the payment history we are targeting from `paymentsByPeriod`.

so the `received`, `sent` fields in `ourPayments` are now an empty array and the two corresponding tabs (`COMPLETED`, `RECEIVED`) are not shown in `Payments.vue` to reflect this.

Hope this makes things clearer for you and let me know of your opinion on what we can(or should) do about this. 
Thanks,

taoeffect:
Ah! Thank you @SebinSong! I see, yes, the issue is `pPeriod` gets set to a period that doesn't exist. Got it. Thanks for the explanation, glad to hear Issue 2 is not a problem. :)

frontend/model/contracts/group.js
```diff
@@ -519,6 +523,46 @@ sbp('chelonia/defineContract', {
...
+    'gi.contracts/group/sendPaymentThankYou': {
+      validate: objectOf({
+        fromUser: string,
+        toUser: string,
+        memo: string
+      }),
+      process ({ data }, { getters }) {
+        const thankYousFrom = getters.groupThankYousFrom
+
+        const fromUser = vueFetchInitKV(thankYousFrom, data.fromUser, {})
```
Silver-IT: I don't think we should make variable `thankYousFrom` as it is used only once and also not complex to make it.

frontend/model/contracts/group.js
```diff
@@ -519,6 +523,46 @@ sbp('chelonia/defineContract', {
...
+        Vue.set(fromUser, data.toUser, data.memo)
```
Silver-IT, line 551: I can see that only one **thank you** could be saved which sent from `fromUser` to `toUser`. Is that by the requirement?

frontend/views/components/modal/Modal.vue
```diff
@@ -128,16 +129,21 @@ export default ({
         if (this.lastFocus) this.lastFocus.focus()
       }
       if (this.replacement) {
-        this.openModal(this.replacement)
+        this.openModal(this.replacement, this.replacementQueries[this.replacement])
+
         this.replacement = null
+        delete this.replacementQueries[this.replacement]
```
Silver-IT: Seems weird here. We should delete `this.replacement` first and then set it `null`, shouldn't we?

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -35,6 +40,12 @@
           @click='markNotReceived'
         )
           i18n I did not receive this
+
+        menu-item(
+          tag='button'
+          icon='comment'
+          @click='openModal("SendThankYouModal", { to: payment.username })'
+        ) Send thank you
```
Silver-IT: This should be translated using `i18n`, shouldn't it?

frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -0,0 +1,126 @@
...
+          data: {
+            fromUser: this.ourUserDisplayName,
+            toUser: this.$route.query.to,
+            memo: this.form.memo
+          }
+        })
+      } catch (err) {
+        alert("something went wrong! couldn't process your request to send a thank you note.")
+
+        return
```
Silver-IT: I think this is unnecessary lines. One is `Line break` and the other is `return`.

frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -0,0 +1,126 @@
...
+    submit () {
+      try {
+        sbp('gi.actions/group/sendPaymentThankYou', {
+          contractID: this.$store.state.currentGroupId,
+          data: {
+            fromUser: this.ourUserDisplayName,
+            toUser: this.$route.query.to,
+            memo: this.form.memo
+          }
+        })
```
Silver-IT, line 98:
```
this.isConfirmation = true
```

We can add this here.

frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -0,0 +1,126 @@
...
+      } catch (err) {
+        alert("something went wrong! couldn't process your request to send a thank you note.")
+
+        return
+      }
+
+      this.isConfirmation = true
```
Silver-IT: And we can remove this.

frontend/views/containers/payments/ThankYouNoteModal.vue
```diff
@@ -0,0 +1,73 @@
...
+  computed: {
+    ...mapGetters([
+      'groupThankYousFrom'
+    ]),
+    from () {
+      return this.$route.query.from || ''
+    },
+    thankYouNote () {
+      const to = this.$route.query.to || ''
+      return (this.groupThankYousFrom[this.from] || {})[to]
```
Silver-IT: `this.groupThankYousFrom` always return object. So I think the following is enough.


```
return this.groupThankYousFrom[this.from][to]
```

frontend/views/containers/payments/ThankYouNoteModal.vue
```diff
@@ -0,0 +1,73 @@
...
+  },
+  computed: {
+    ...mapGetters([
+      'groupThankYousFrom'
+    ]),
+    from () {
+      return this.$route.query.from || ''
+    },
+    thankYouNote () {
+      const to = this.$route.query.to || ''
```
Silver-IT: Instead of setting empty string to `to`, I think we should return empty or exceptional value.

Silver-IT (changes requested):
Nice work, @SebinSong

frontend/model/contracts/group.js
taoeffect: Yep! From my previous review - storing all of the thank yous ever made is too much data in memory and not necessary.

frontend/views/components/modal/Modal.vue
SebinSong: @Silver-IT yeah right. good catch :)

SebinSong:
@Silver-IT thanks for review. updates been made accordingly in [this commit](https://github.com/okTurtles/group-income/pull/1421/commits/c79ebc1f2f2670a8d4770a0e6887ac698b5011d4)

Silver-IT (dismissed):
Thanks, @SebinSong . Nice work.

taoeffect (changes requested):
Almost! I tested this in FF Dev Edition using the mobile responsive mode, and found this bug:

bug mov

Not sure if it's related to #1431 or is a completely different issue.

frontend/views/components/modal/Modal.vue
```diff
@@ -43,9 +45,14 @@ export default ({
...
   watch: {
     '$route' (to, from) {
-      if (to.query.modal) {
+      const toModal = to.query.modal
+
+      if (toModal) {
         // We reset the modals with no animation for simplicity
-        if (to.query.modal !== this.content) this.content = to.query.modal
+        if (toModal !== this.content) {
+          if (this.content) this.replaceModal(toModal, omit(to.query, 'modal')) // if another modal is already open, replace it.
```
SebinSong, line 53: @taoeffect This is the fix for the bug you've spotted.

SebinSong:
@taoeffect 
fixed the bug in the following commits.

the bug is not something brought in by this PR but the `$route` watcher in `Modal.vue` needed some enhanced logic to cover the case where another modal is queried when one is already open (in our case, `PaymentDetails` modal is requested when `NotificationModal` is open)

to be more detailed, 
the cause of the issue is that the value of `this.content` in `Modal.vue` is updated to a new modal too early, because `CLOSE_MODAL` event is emitted only after the 'leave' animation of the currently open modal(`NotificationModal`) ends. the `Modal.vue` needs to sorts of enqueue the newly requested modal(`PaymentDetails`) and waits until the animation finishes before performing modal replacement .
Explicitly calling `replaceModal()` method in this situation elegantly does this job.

hope it sounds all good. :) thanks,

taoeffect (approved):
Excellent work @SebinSong! The modal bug seems to be fixed! 😄 

A weird thing happened while I was testing the thank-you workflow but I couldn't reproduce it after I tried to reproduce it twice, so I'm document this here in case it shows up again, and approving the PR. Great work!

Screen Shot 2022-10-07 at 3 46 16 PM

PR #1422, opened 2022-09-29
#1415 - Add 'confirm password' field to the sign-up form

### LINKED ISSUE #1415: Add "Confirm Password" field to Sign Up form
**Problem**
Since we don't have account recovery implemented, it's really important that users don't mistype their password when signing up, but we only have a single password field in the signup form instead of two.

**Solution**
Add "Confirm Password" field that the user must fill to create an account. Make sure that both passwords match.

1  Sign up
2  Sign up

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=2737%3A154

SebinSong:
closes #1415


taoeffect (changes requested):
It seems to work great - and great job adding that additional test to Cypress!

Only one request - the password field seems to have plenty of room to show the •••• characters, and they seem to be cut off a bit prematurely:

Screen Shot 2022-09-30 at 1 47 35 PM

Do you think you could make it show that they're cut off much closer to the 👁‍ icon, so that there's 0 padding? Not so that they're touching, but my understanding is that there's already about a ~10px box border around that icon.

SebinSong:
@taoeffect 
The PR has been updated with your requested change.

taoeffect (changes requested):
Looks like I caught one final bug. 😁

The dark-mode background for the second field is off when you click the eye icon to show the password (text should be white):

Screen Shot 2022-10-04 at 4 09 48 PM

SebinSong:
The fix has been made!

Silver-IT (approved):
Nice work, @SebinSong. I approve this pull request.

taoeffect (approved):
Excellent @SebinSong!

PR #1424, opened 2022-09-30
Add LRU cache to the backend

### LINKED ISSUE #1198: Add an LRU cache to the backend to speed things up a bit
### Problem

Reading from the filesystem is much slower than from memory.

### Solution

When in `NODE_ENV=production` mode or `GI_PERSIST=true`, we can add [node-lru-cache](https://github.com/isaacs/node-lru-cache) to speed up lookups, and enable a server-configurable value for the size of the LRU cache in MB that it reads from a config file (config can be just a `.json` file for now, but in the future it would make sense to switch to something like an [INI](https://github.com/npm/ini) file).

A good default for the LRU cache would be say, 20MB of memory.

snowteamer:
Closes #1198

### Summary of changes:
- Server-side in-memory caching has been enabled in 'chelonia/db/get' and 'chelonia/db/set'  when `NODE_ENV` is `'production'` or `GI_PERSIST` is `'true'`.

### Additional info:
- Default cache size is 1000
- Test using `export GI_PERSIST=true && grunt test` or `export GI_PERSIST=true && grunt dev`


backend/lru-cache.js
```diff
@@ -0,0 +1,24 @@
...
+      return map.has(key)
+    },
+    set (key: any, value: any): void {
+      map.set(key, value)
+      // Remove the oldest cache entry if necessary.
+      if (map.size > MAX_CACHE_ENTRIES) {
+        // Modern .keys() implementations return an iterator which respects key insertion order.
+        // This means the first iterated key will be the oldest one to have been added.
+        map.delete(map.keys().next().value)
+      }
```
taoeffect: This is not an LRU cache. Please use `node-lru-cache` instead of this file

taoeffect (changes requested):

backend/lru-cache.js
```diff
@@ -0,0 +1,24 @@
+// How many entries to keep at most.
+const MAX_CACHE_ENTRIES = 1000
```
taoeffect: Note: if we end up going with this (I need to run some benchmarks to figure this out), we'll want to make this a customizable option that you can pass in to `createCache`.

Also this file (like all js files) should have `'use strict'` at the top of the file.

taoeffect (changes requested):
In my benchmarks the implementation of LRU in this PR is 44x slower than node-lru-cache:

```
$ node homespun/index.mjs
Took: 63997ms
$ node node-lru/index.mjs
Took: 1449ms
```

[benchmark-lru.zip](https://github.com/okTurtles/group-income/files/9693940/benchmark-lru.zip)

While I commend your creativity, I'd ask to please use `node-lru-cache` (aka `lru-cache`).

backend/database.js
```diff
@@ -184,15 +185,31 @@ function throwIfFileOutsideDataDir (filename: string): string {
 }
 
 if (production || process.env.GI_PERSIST) {
+  // https://github.com/isaacs/node-lru-cache#usage
+  const cache = new LRU({
+    max: 1000
```
taoeffect: Can you make this 10k and add a `// TODO:` comment to read this value from a file / envvar?

In fact, you can also implement the envar now, e.g. with `process.env.GI_LRU_NUM_ITEMS`. Just remember envars need to be converted from strings to numbers first.

backend/database.js
```diff
@@ -184,15 +185,31 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
   sbp('sbp/selectors/overwrite', {
     // we cannot simply map this to readFile, because 'chelonia/db/getEntry'
     // calls this and expects a string, not a Buffer
     // 'chelonia/db/get': sbp('sbp/selectors/fn', 'backend/db/readFile'),
     'chelonia/db/get': async function (filename: string) {
-      const value = await sbp('backend/db/readFile', filename)
-      return Boom.isBoom(value) ? null : value.toString('utf8')
+      if (cache.has(filename)) {
+        return cache.get(filename)
+      }
```
taoeffect: One can and should avoid a double-lookup by just calling `.get` and checking for `undefined` or `null` (or whatever is considered "non-existent").

backend/database.js
```diff
@@ -184,15 +185,31 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
+      if (Boom.isBoom(bufferOrError)) {
+        return null
+      }
+      const value = bufferOrError.toString('utf8')
+      cache.set(filename, value)
+      return value
     },
-    'chelonia/db/set': sbp('sbp/selectors/fn', 'backend/db/writeFile')
+    'chelonia/db/set': async function (filename: string, data: any): Promise<*> {
+      cache.set(filename, data)
```
taoeffect: Since the `await` below can fail, we want to set the cache only if it succeeds.

taoeffect (changes requested):
Nice work @snowteamer, few remaining issues.

taoeffect (approved):
Approved! 😄👍

PR #1434, opened 2022-10-10
Support History Widget Implemented

### LINKED ISSUE #815: Create "Support History" widget for ContributionsOverviewWidget
### Problem

[figma link](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=1712%3A413)

Support history

This still needs to be implemented.

### Solution

Different parts assigned to different people:

1. Assigned to @Silver-IT: finish implementing the logic for storing the support history data
2. Assigned to @pieer: implement the design as a vue component
3. Assigned to @Silver-IT: connect the component to the state (after step 1 above is done)

### LINKED ISSUE #1433: Implement archiving payment history
### Problem

In `group.js` contract, `clearOldPayments` keeps only two 30 day periods worth of payments in the state, along with a corresponding `TODO` to archive payments:

```js
function clearOldPayments ({ state, getters }) {
  const sortedPeriodKeys = Object.keys(state.paymentsByPeriod).sort()
  // save two periods worth of payments, max
  while (sortedPeriodKeys.length > 2) {
    const period = sortedPeriodKeys.shift()
    for (const paymentHash of getters.paymentHashesForPeriod(period)) {
      Vue.delete(state.payments, paymentHash)
      // TODO: archive the old payments in a sideEffect, not here
    }
    Vue.delete(state.paymentsByPeriod, period)
  }
}
```

We currently don't have archiving payments implemented, but we need this so that people can keep track of their entire history.

### Solution

Payments should be archived to IndexedDB via the `'gi.db/archive/*` selectors in `frontend/model/database.js`.

Those selectors were created for implementing archiving of proposals, and researching how that PR was done is likely useful to anyone closing this issue. That PR can be found here: https://github.com/okTurtles/group-income/pull/1388

To close this issue:

1. Implement archiving of old payments (older than 2 months worth)
2. Make sure that pagination works and displays all payments in the completed payments tab (and the received payments tab).

Silver-IT:
### Summary of Changes
- Closes #815 
- Closes #1433 

- [ ] To display delayed payments is not implemented and also not tested.


frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -15,31 +15,65 @@ div(:class='isReady ? "" : "c-ready"')
...
+      return this.groupSettings.mincomeAmount
+    },
+    periods () {
+      const periods = []
+      let lastPeriod = this.currentPaymentPeriod
+      for (let i = 0; i < 5; i++) {
+        periods.unshift(lastPeriod)
+        lastPeriod = this.periodBeforePeriod(lastPeriod)
+      }
+      periods.unshift(lastPeriod)
```
taoeffect: I think this loop can be rewritten in a way so that the `.unshift` happens inside of it only instead of again outside of it... For example:

```js
for (let i = 0; i < MAX_SAVED_PERIODS - 1; i++) {
  periods.unshift(this.periodBeforePeriod(periods[0]))
}
```

And make sure to define `MAX_SAVED_PERIODS` under `shared/constants.js` and use it everywhere as needed. Hard coding constants is a DRY violation and will lead to trouble if they change.

frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -15,31 +15,65 @@ div(:class='isReady ? "" : "c-ready"')
...
+      return periods
+    },
+    history () {
+      return this.periods.map((period, i) => {
+        const payments = this.paymentsForPeriod(period)
+        const { totalDistributionAmount, numberOfNeedyPeople } = this.parsePayments(payments)
+        return {
+          total: numberOfNeedyPeople === 0 ? 1 : totalDistributionAmount * this.mincome / numberOfNeedyPeople,
```
taoeffect: In English, it sounds rather politically incorrect (and somewhat funny) to name this variable `numberOfNeedyPeople`. It is also a bit against the ethos of the project's core values - which is to recognize that everyone in the group has something of value to offer, even if it isn't money.

So can we rename this variable to `numReceivers` instead?

Same change requested below inside of `parsePayments`.

taoeffect (changes requested):
Preliminary review. Nice work @Silver-IT!

frontend/views/containers/contributions/SupportHistory.vue
Silver-IT: Sure. I agree with you. It's really better for me too.

frontend/views/containers/contributions/SupportHistory.vue
Silver-IT: Nice catch. That should be.

frontend/model/contracts/group.js
```diff
@@ -1121,6 +1128,28 @@ sbp('chelonia/defineContract', {
...
+      for (const period of Object.keys(paymentsByPeriod).sort()) {
+        const periodKey = `paymentPeriods/${username}/${contractID}`
+        const periods = await sbp('gi.db/archive/load', periodKey) || []
+        const paymentsKey = `paymentsByPeriod/${username}/${contractID}/${period}`
+        const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
+
+        periods.unshift(period)
+        merge(payments, paymentsByPeriod[period])
+
+        if (periods.length > MAX_ARCHIVED_PAYMENTS) {
```
taoeffect: `MAX_ARCHIVED_PAYMENTS` has payments in the name, not periods. So this variable - since it's being used to compare against periods, needs to be renamed, since it's referring to periods and not payments.

Something like `MAX_ARCHIVED_PERIODS`.

Also, since this value can change, this should be a `while` instead of an `if`.

frontend/model/contracts/shared/functions.js
```diff
@@ -22,6 +23,62 @@ import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
...
+  return periods
+}
+
+export async function getHistoricalPayments (numPeriods: number, skip?: number) {
+  const periods = getHistoricalPeriods(numPeriods, skip)
+
+  const payments = await Promise.all(periods.map(period => getPaymentsByPeriod(period)))
+
+  return { periods, payments }
+}
```
taoeffect: Where is  this function used?

frontend/model/contracts/shared/functions.js
```diff
@@ -22,6 +23,62 @@ import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
 
 // group.js related
 
+export async function getPaymentsByPeriod (period: string) {
```
taoeffect: Perhaps I'm mistaken, but this function seems to only be used in the frontend UI, and not in the contracts too. `shared/functions.js` is only for code that is used in both places.

frontend/model/contracts/shared/functions.js
```diff
@@ -22,6 +23,62 @@ import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
...
+      const payment = paymentsByHash[hash]
+      if (payment.data.status === PAYMENT_COMPLETED) {
+        payments.push({
+          from: payment.meta.username,
+          to: payment.data.toUser,
+          hash,
+          amount: payment.data.amount,
+          isLate: !!payment.data.isLate,
+          when: payment.data.completedDate
+        })
```
taoeffect: This code appears twice now, once here and once in the group contract getter `paymentsForPeriod`. We need to DRY this by creating a shared function that takes as input the payment and the payment hash, and call that function in both places. This will help prevent a divergence in the type of object returned.

taoeffect (changes requested):
Excellent work @Silver-IT!

I've left a few comments in this preliminary review.

Also - I don't think I saw pagination implemented for the payments screen? Is that already implemented or does it still need to be? Pagination is one of the requirements for closing #1433 (see the second bullet point).

frontend/model/contracts/group.js
Silver-IT: > Something like MAX_ARCHIVED_PERIODS.

Correct. That's more precise meaning.

> Also, since this value can change, this should be a while instead of an if.

I understand. Could be changed later like 20 to 10.

frontend/model/contracts/shared/functions.js
Silver-IT: I am going to use this in the `Payments` tab but never used it yet.

frontend/model/contracts/shared/functions.js
Silver-IT: Make sense. Will find a best place to save them.

frontend/model/contracts/shared/functions.js
Silver-IT: Nice catch.

frontend/model/contracts/shared/functions.js
```diff
@@ -22,6 +22,33 @@ import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
 
 // group.js related
 
+export function getPaymentHashes (periodPayments: Object): string[] {
```
taoeffect: Name suggestion: `paymentHashesFromPaymentPeriod`

frontend/model/contracts/shared/functions.js
```diff
@@ -22,6 +22,33 @@ import { logExceptNavigationDuplicated } from '~/frontend/views/utils/misc.js'
...
+      for (const toUser in paymentsFrom[fromUser]) {
+        hashes = hashes.concat(paymentsFrom[fromUser][toUser])
+      }
+    }
+  }
+
+  return hashes
+}
+
+export function simplifyPayment (paymentHash: string, payment: Object): {
```
taoeffect: Name suggestion: `createPaymentInfo`

frontend/model/contracts/group.js
```diff
@@ -1124,6 +1118,34 @@ sbp('chelonia/defineContract', {
...
+          for (const hash of paymentHashes) {
+            delete archPayments[hash]
+          }
+          delete archPaymentsByPeriod[shouldBeDeletedPeriod]
+        }
+
+        await sbp('gi.db/archive/save', archPaymentsByPeriodKey, archPaymentsByPeriod)
+        await sbp('gi.db/archive/save', archPaymentsKey, archPayments)
+      }
+      sbp('okTurtles.events/emit', PAYMENTS_ARCHIVED, { paymentsByPeriod, payments })
```
taoeffect: Well done! 👏

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -0,0 +1,97 @@
...
+      'currentPaymentPeriod',
+      'paymentHashesForPeriod',
+      'groupIncomeAdjustedDistribution'
+    ])
+  },
+  methods: {
+    async getHistoricalPaymentsInTypes () {
+      const paymentsInTypes = {
+        sent: cloneDeep(this.ourPayments?.sent),
+        received: cloneDeep(this.ourPayments?.received),
```
taoeffect: The way these are written suggests that `this.ourPayments` could be `null` or `undefined`. This presents a problem, because if that's true, then `cloneDeep` will return `null`, and an exception will be thrown then these are used later on.

However, looking at the implementation of `ourPayments`, it doesn't seem like this value can be null/undefined, so the `?` should be removed here.

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -0,0 +1,97 @@
...
+              for (const rHash of paymentsFrom[fromUser][toUser]) {
+                const { data, meta } = payments[rHash]
+                paymentsInTypes.received.push({ hash: rHash, data, meta, amount: data.amount, username: data.toUser })
+              }
+            } else if (fromUser === this.ourUsername) {
+              for (const sHash of paymentsFrom[fromUser][toUser]) {
+                const { data, meta } = payments[sHash]
+                paymentsInTypes.sent.push({ hash: sHash, data, meta, amount: data.amount, username: data.toUser })
+              }
+            }
```
taoeffect: This seems like it has some code duplication. We can DRY this more by doing something like this:

```js
if (toUser === this.ourUsername || fromUser === this.ourUsername) {
  const receivedOrSent = toUser === this.ourUsername ? 'received' : 'sent'
  for (const hash of paymentsFrom[fromUser][toUser]) {
      const { data, meta } = payments[hash]
      paymentsInTypes[receivedOrSent].push({ hash, data, meta, amount: data.amount, username: toUser })
    }
}
```

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -0,0 +1,97 @@
+import sbp from '@sbp/sbp'
+import { mapState, mapGetters } from 'vuex'
+import { PAYMENT_COMPLETED } from '@model/contracts/shared/payments/index.js'
+import { simplifyPayment, getPaymentHashes } from '@model/contracts/shared/functions.js'
+import { cloneDeep } from '@model/contracts/shared/giLodash.js'
+
+const PaymentsMixin: Object = {
```
taoeffect, line 9: I would add a comment above this explaining that this file combines payment information from both the current in-memory state and the archived payments on disk.

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    async getHistoricalPaymentByHash (hash: string) {
+      const paymentsKey = `payments/${this.ourUsername}/${this.currentGroupId}`
+      const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
+      return payments[hash]
+    }
+  }
+}
+
+export default PaymentsMixin
```
taoeffect, line 91: Nicely done btw, this file is a great addition!

frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -26,19 +29,58 @@ export default ({
...
+        list[from] = (list[from] || 0) + amount
+        list[to] = (list[to] || 0) - amount
       })
+      return {
+        numReceivers: Object.values(list).filter(amount => amount < 0).length,
+        totalDistributionAmount: Object.values(list).filter(amount => amount > 0).reduce((total, curValue) => total + curValue, 0)
+      }
+    },
+    async updateHistory () {
+      this.history = await Promise.all(this.periods.map(async (period, i) => {
```
taoeffect, line 70: OK, either in this function, or in `this.periods`, we need to show 0% for months that don't have payments for them (or don't even exist).

Remember to test locally before pushing, since a quick test on my side shows this when first starting up a group with 3 users and no payments:

Screen Shot 2022-10-26 at 4 49 37 PM

taoeffect (changes requested):
Excellent work @Silver-IT! Getting very close to merging. Found a few remaining issues, see comments!

frontend/views/containers/payments/PaymentsMixin.js
Silver-IT: I described the implementation of `ourPayments` getter below. It could be return `undefined on the third line.
```
ourPayments (state, getters) {
    const periodPayments = getters.groupPeriodPayments
    if (Object.keys(periodPayments).length === 0) return
    const ourUsername = getters.ourUsername
    const cPeriod = getters.currentPaymentPeriod
    const pPeriod = getters.periodBeforePeriod(cPeriod)
    const currentSent = getters.ourPaymentsSentInPeriod(cPeriod)
    const previousSent = getters.ourPaymentsSentInPeriod(pPeriod)
    const currentReceived = getters.ourPaymentsReceivedInPeriod(cPeriod)
    const previousReceived = getters.ourPaymentsReceivedInPeriod(pPeriod)

    // TODO: take into account pending payments that have been sent but not yet completed
    const todo = () => {
      return getters.groupIncomeAdjustedDistribution.filter(p => p.from === ourUsername)
    }

    return {
      sent: [...currentSent, ...previousSent],
      received: [...currentReceived, ...previousReceived],
      todo: todo()
    }
  }
```
So in my opinion, the `sent` and `received` should be defined this way.
```
sent: cloneDeep(this.ourPayments?.sent || []),
received: cloneDeep(this.ourPayments?.received || []),
```

taoeffect: Oh! Good point!! 👍

frontend/views/containers/payments/PaymentsMixin.js
Silver-IT: Great solution.

frontend/views/containers/contributions/SupportHistory.vue
Silver-IT: The following screenshot shows the example of support history graph for the last six months when I add payments only 2 months.
image

dotmacro: Seeing multiple months of red "0%" will be discouraging for new groups. I think there should be some other placeholder for non-existent months, so I've made a comment on Figma: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=1712:413#294246554

(Leaving this comment here because it's relevant, not because it should be blocking the PR)

taoeffect (changes requested):
Excellent work @Silver-IT! It seems to be working really well now.

While testing this and changing the date on my computer, I discovered a bug that resulted in a mismatch between the next distribution date as shown in the blue pill and in the "Due On" field:

Screen Shot 2022-11-30 at 3 28 10 PM

The "Due On" field is correct, but the blue pill is wrong.

And I found one more bug: the dates did not update when I updated the time. I had to refresh the page for it to happen. So I fixed both bugs.

Here is the diff from my patch for two files: `frontend/model/state.js` and `frontend/views/containers/payments/PaymentNextDistributionPill.vue`, please apply it to this PR and we can get this merged. 😄 

```diff
diff --git a/frontend/model/state.js b/frontend/model/state.js
index e20f4951..5bf2013f 100644
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -56,6 +56,11 @@ const initialState = {
     : ['error', 'warn', 'info']
 }
 
+const reactiveDate = Vue.observable({ date: new Date() })
+// every 5 minutes update the reactive date to force recomputation of getters using this
+setInterval(function () { reactiveDate.date = new Date() }, 5 * 60 * 1000)
+
 sbp('sbp/selectors/register', {
   // 'state' is the Vuex state object, and it can only store JSON-like data
   'state/vuex/state': () => store.state,
@@ -334,8 +339,9 @@ const getters = {
       return profile.displayName || username
     }
   },
+  // this getter gets recomputed automatically according to the setInterval on reactiveDate
   currentPaymentPeriod (state, getters) {
-    return getters.periodStampGivenDate(new Date())
+    return getters.periodStampGivenDate(reactiveDate.date)
   },
   thisPeriodPaymentInfo (state, getters) {
     return getters.groupPeriodPayments[getters.currentPaymentPeriod]
@@ -617,6 +623,15 @@ const store: any = new Vuex.Store({
   strict: false // we're intentionally modifying state outside of commits
 })
 
+// Somewhat of a hack, I'm not sure why this is necessary now, but ever since changing how
+// getters.currentPaymentPeriod works, this is necessary to force the UI to update immediately
+// after a payment is made.
+store.watch(function (state, getters) {
+  return getters.groupSettings.distributionDate
+}, function () {
+  reactiveDate.date = new Date()
+})
+
 // save the state each time it's modified, but debounce it to avoid saving too frequently
 const debouncedSave = debounce(() => sbp('state/vuex/save'), 500)
 store.subscribe((commit) => {
diff --git a/frontend/views/containers/payments/PaymentNextDistributionPill.vue b/frontend/views/containers/payments/PaymentNextDistributionPill.vue
index aef11b6f..de520164 100644
--- a/frontend/views/containers/payments/PaymentNextDistributionPill.vue
+++ b/frontend/views/containers/payments/PaymentNextDistributionPill.vue
@@ -13,10 +13,11 @@ export default ({
   computed: {
     ...mapGetters([
       'groupSettings',
-      'periodAfterPeriod'
+      'currentPaymentPeriod',
+      'dueDateForPeriod'
     ]),
     nextDistributionDateShort () {
-      const nextDisDate = this.periodAfterPeriod(this.groupSettings.distributionDate)
+      const nextDisDate = this.dueDateForPeriod(this.currentPaymentPeriod)
 
       return humanDate(nextDisDate, { month: 'short', day: 'numeric' })
     }
```

frontend/views/pages/Payments.vue
```diff
@@ -199,15 +201,24 @@ export default ({
...
   created () {
     this.setInitialActiveTab()
+    this.updatePayments()
   },
   watch: {
     needsIncome () {
       this.setInitialActiveTab()
+    },
+    ourPayments () {
+      this.updatePayments()
```
taoeffect: Note: because `ourPayments` gets recomputed each time `reactiveDate.date` gets updated, this causes loading all of the payments history at that interval frequency, which in the previous diff patch I posted was once per second. That is too frequent to load all of the payments from disk!

So we should change this. Instead we should construct a custom watch statement that looks at the payments for the current period using: `getters.paymentHashesForPeriod(getters.periodStampGivenDate(getters.groupSettings.distributionDate))?.length` - to avoid depending on `getters.currentPaymentPeriod` from state.js.

And we should add a comment explaining all of this so that we don't forget why we did it that way.

Silver-IT: Nice catch. Thanks, @taoeffect

Silver-IT: @taoeffect, your solution breaks the cypress error.
Even though we only watch the payments count, `currentPaymentPeriod` getter is used to make `todo` payments, so I don't think this could be the solution.
Please check my solution to get the historical payments only when `ourPayments` get updated.
I used `deepEqualJSONType` function from `giLodash`.

frontend/model/contracts/group.js
```diff
@@ -1124,6 +1118,34 @@ sbp('chelonia/defineContract', {
...
```
taoeffect, line 1148: @Silver-IT So this PR seems to be well written and ready to merge, however I have one remaining doubt about this approach that I mentioned on Slack (feel free to reply there or here), and that is:

- On slower and older devices, it might still be quite a lot to save and load all of the payments payments in one go.
- The whole point of archiving the payments in the first place is to not have to load all of them into memory. Yet with this PR we will still end up doing that.

Although storing the array of archived payments using pagination<sup>1</sup>  is more complicated...

So, what do you think, we could merge this now, and then work on archiving using pages in a future PR?


<sup>1</sup> e.g. by storing the length of the array and paginating the keys themselves, for example, by adding the periodstamp to the key name, or by adding numbers to the keyname.

Silver-IT: Yeah, it's really good point. I am 100% sure that we should archive payments using pagination.
I will create an issue for this and wanna implement in the future PR.

Silver-IT: Created [this issue](https://github.com/okTurtles/group-income/issues/1448).

frontend/model/state.js
```diff
@@ -56,6 +56,11 @@ const initialState = {
     : ['error', 'warn', 'info']
 }
 
+const reactiveDate = Vue.observable({ date: new Date() })
+// every 5 minutes update the reactive date to force recomputation of getters using this
+// TODO: can make this every minute after payments pagination no longer loads all of the payments
+setInterval(function () { reactiveDate.date = new Date() }, 5 * 60 * 1000)
```
taoeffect: I just realized there's a far better way of writing this, could you please update this to this:

```js
setInterval(function () {
  const date = new Date()
    reactiveDate.date = date
  }
}, 60 * 1000)
```

Note: you'll need to import `MINS_MILLIS` from `time.js`

frontend/model/state.js
```diff
@@ -477,8 +483,8 @@ const getters = {
   },
   groupMembersSorted (state, getters) {
     const profiles = getters.currentGroupState.profiles
-    if (!profiles) return []
-    const weJoinedMs = new Date(profiles[getters.ourUsername].joinedDate).getTime()
+    if (!profiles || !profiles[getters.ourUsername]) return []
+    const weJoinedMs = new Date(profiles[getters.ourUsername]?.joinedDate).getTime()
```
taoeffect: Since line 486 verifies that `profiles[getters.ourUsername]` exists, I don't think we need the `?` after `profiles[getters.ourUsername]` on line 487, right?

taoeffect (changes requested):
One final review - realized that my code could be improved a bit to be more responsive to date changes and more efficient at the same time.

frontend/model/state.js
Silver-IT: Correct. Nice catch.

frontend/model/state.js
Silver-IT: Got it.

taoeffect (approved):
Approved!! Fantastic work @Silver-IT! 💪 👍

PR #1437, opened 2022-10-13
remove disagreeing number option from the proposal dropdown

### LINKED ISSUE #1436: Forgot to hide disagreement rule in voting dropdown
### Problem

The disagreement rule still appears in the UI in one place we forgot:

Screen Shot 2022-10-12 at 2 05 45 PM

### Solution

Don't delete the code, but instead comment it out or something so that the remaining traces of the disagreement rule are visually no long present.

SebinSong:
closes #1436


taoeffect (approved):
Thanks @SebinSong!

Issue #1435, opened 2022-10-10
Various HTML issues


corrideat:
### Problem

1. The app sets `lang="eng"`, which is an invalid language.
2. IE conditional comments are used. Conditional comments have been removed since IE 10, released in 2012 (that's over 10 years ago). The last version of IE (IE 11) was released in 2013, and Microsoft have announced its end of life. It's unlikely that someone will be browsing the app using such an outdated browser.
3. Related to IE: meta x-ua-compatible is set.

### Solution

1. Set `lang="en"` instead
2. Remove conditional comments. If showing a warning for unsupported / outdated browsers is desired, this should be some additional JavaScript that does it. The reasoning is that the (1) app doesn't work at all in IE 10 or IE 11 (with no warning shown), (2) it's far more likely that someone will use IE 11 than IE <10, and it seems odd to show the warning only to these people using incredibly outdated systems (which probably won't even be able to connect to the app in the first place, as the latest OSes to support these don't support modern TLS) and (3) it'd be nice to show such a warning to users of other outdated browsers (e.g., old FF versions)
3. Remove x-ua-compatible.

taoeffect:
Closed!

PR #1439, opened 2022-10-17
Fix HTML issues discussed in #1435

snowteamer:


frontend/index.html
```diff
@@ -1,23 +1,14 @@
...
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="x-ua-compatible" content="ie=edge">
     <title>Group Income</title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
     <link rel="stylesheet" href="/assets/css/main.css">
     <link rel="icon" href="/assets/images/group-income-icon-transparent.png" sizes="32x32">
   </head>
   <body lang="eng">
```
corrideat: Remove `lang` from `body`, as it's an invalid value (and redundant)

snowteamer: Good catch - sorry for overlooking it

corrideat:
As per the conversation in Slack, the application should include a fallback message that is shown when browsers are unsupported (e.g., old browsers that don't understand modern JS) or when scripts are disabled. This message must then be removed with JavaScript.

In simple terms, this is removing the 'conditional' part of the conditional comment and making the markup part of the page for all browsers. Then, old browsers or browsers without JS will show this fallback message.

Note that `<noscript>` _can_ be used to provide additional context, but that it _should not_ wrap the entire block, as this will result in a blank page shown to old browsers or to browsers that selectively run some scripts.

Example:

```html
<body>
    <div class="modal is-active" id="fallback">
        <div class="modal-content">
            <h1>Outdated browser</h1>
            <p>You are using an <strong>outdated</strong> or unsupported browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
          <noscript><p>This application requires JavaScript to be enabled to correctly function.</p></noscript>
        </div>
    </div>
</body>
```

Then somewhere in the script:

```js
// do this once the DOM is available
document.getElementById('fallback')?.remove()
```

taoeffect (changes requested):
I think @corrideat left some good suggestions, no?

frontend/index.html
```diff
@@ -1,23 +1,20 @@
...
-    <div class="modal is-active">
-        <div class="modal-content">
-            <h1>Outdated browser</h1>
-            <p>You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
-        </div>
+  <body>
+    <div class="modal is-active" id="fallback" style="visibility: hidden;">
+      <div class="modal-content">
+        <h1>Outdated browser</h1>
+        <p>You are using an <strong>outdated</strong> or unsupported browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
```
taoeffect: Nice update! I have one request: at first I thought that we should change this to be an https link instead of http, but on second thought I think it would be better to simply remove the link. We don't know how long this website will stay up for, we don't know whether or not they'll get compromised and start serving malware, and we don't know if they'll start recommending bad browsers, so IMO it would be better to avoid all these possibilities and leave it at, "please upgrade your browser".

taoeffect (changes requested):
Minor change request - great work @snowteamer!

frontend/index.html
snowteamer: Agreed :+1:

taoeffect (approved):
Approved! Thanks @snowteamer & @corrideat!

PR #1441, opened 2022-10-19
Resolve frontend issues

### LINKED ISSUE #1137: TypeValidator objectMaybeOf makes error with boolean type
### Problem

`boolean` TypeValidator which is defined in `frontend/utils/flowTyper.js` returns `TypeError` when the data has boolean-typed false value. That's because it throws TypeError when `!validations[key](data[key])` is falsy value, and `boolean` TypeValidator returns `false` as boolean which is also falsy value.

The followings are the definition of the `boolean` and `objectMaybeOf` TypeValidators.
```
export const boolean = (
  function boolean (value, _scope = '') {
    if (isEmpty(value)) return false
    if (isBoolean(value)) return value
    throw validatorError(boolean, value, _scope)
  }
  : TypeValidator<boolean>
)
```

```
export function objectMaybeOf (validations: Object): Object {
  return function (data: any) {
    object(data)
    for (const key in data) {
      if (validations[key]) {
        if (!validations[key](data[key])) {
          throw new TypeError(`${key} has bad value: ${data[key]}`)
        }
      }
    }
    return data
  }
}
```

### Solution

I recommend to take the `objectOf` as a reference, it works fine with `boolean` TypeValidator.

### LINKED ISSUE #1142: Unexpected result when adding emoji
### Problem

* Dialog gets hidden, when we click the tabs at the top
image
* Several emojis are added once we try to add one after switching tabs several times
image

### Solution

* I have no idea at this moment, we should debug and find it out.

### LINKED ISSUE #1144: Validation error message is not displayed for the textarea
### Problem

- On the chat room page, there are dialogs for renaming and changing its description. If the validation is failed whatever `required` or `maxLength`, there should be an error message under the element. `Input` element is used for the renaming and the `Textarea` element is used for changing description. I can see the error message under the `Input` but can not see it for the `Textarea`.
image
image

### Solution

- I can not guess the reason at this moment. Need to debug and find the reason and fix it.

### LINKED ISSUE #1180: Spinner doesn't mean join in Group Join page
### Problem

If a new user tries to join group with the invitation link, he needs to create an account first things first.
So the invitation link leads him to a `Signup` page. See attached image.
image
Here, we have a submit button whose name is `Create an account`.
This submit button is inside the `Signup` form and it's submit actions is `Signup`.
So the spinner starts rolling as soon as clicks the button and stops rolling as soon as `Signup` is finished.
This spinner needs to be rolling until the whole joining process is finished since this page is `Join` page.

### Solution

There could be several solutions for this
But I think we could update `Signup` form just adding new attributes.
For example, we can add `postSubmit` function attribute so that it can be added inside submit action(which is `signup` function inside SignupForm.vue).
This solution is very similar to the way Chelonia works with hooks. (`prepublish` and `postpublish`)

### LINKED ISSUE #1425: Switching group doesn't work in Group Settings page
### Problem

Switching group works fine when we normally do.
But if we switch group just after update the group picture in Group Settings page, it doesn't work.
image

### Solution

Debug and find the reason and solve it.

Silver-IT:
### Summary of Changes

- Closes #1425 
- Closes #1144 
- Closes #1137 
- Closes #1180
- Closes #1142

### DIFF
--- a/frontend/views/components/AvatarUpload.vue
+++ b/frontend/views/components/AvatarUpload.vue
@@ -57,6 +57,10 @@ export default ({
       try {
         const { selector, contractID, key } = this.sbpParams
         await sbp(selector, { contractID, data: { [key]: picture } })
+        // calling `setFromBlob` seems unnecessary here since the bound :avatar
+        // parameter should get updated after the upload completes.
+        // also, calling it here prevents the avatar from switching in Group Settings if we
+        // just uploaded a new image and then switch groups (see #1425)
         // this.$refs.picture.setFromBlob(fileReceived)
         this.$refs.formMsg.success(L('Avatar updated!'))
       } catch (e) {
--- a/frontend/views/components/AvatarUpload.vue
+++ b/frontend/views/components/AvatarUpload.vue
@@ -57,7 +57,7 @@ export default ({
       try {
         const { selector, contractID, key } = this.sbpParams
         await sbp(selector, { contractID, data: { [key]: picture } })
-        this.$refs.picture.setFromBlob(fileReceived)
+        // this.$refs.picture.setFromBlob(fileReceived)
         this.$refs.formMsg.success(L('Avatar updated!'))
       } catch (e) {
         console.error('AvatarUpload fileChange() error:', e)
--- a/frontend/views/pages/GroupSettings.vue
+++ b/frontend/views/pages/GroupSettings.vue
@@ -200,6 +200,7 @@ export default ({
   },
   watch: {
     groupSettings () {
+      // re-fetch the latest correct values whenever the user switches groups
       this.refreshForm()
     }
   }
--- a/frontend/views/pages/GroupSettings.vue
+++ b/frontend/views/pages/GroupSettings.vue
@@ -181,6 +181,14 @@ export default ({
       } else {
         this.openProposal('GroupLeaveModal')
       }
+    },
+    refreshForm () {
+      const { groupName, sharedValues, mincomeCurrency } = this.groupSettings
+      this.form = {
+        groupName,
+        sharedValues,
+        mincomeCurrency
+      }
     }
   },
   validations: {
@@ -188,6 +196,11 @@ export default ({
       groupName: {
         [L('This field is required')]: required
       }
+    }
+  },
+  watch: {
+    groupSettings () {
+      this.refreshForm()
     }
   }
 }: Object)
--- a/frontend/views/pages/Join.vue
+++ b/frontend/views/pages/Join.vue
@@ -19,8 +19,8 @@ div
         h1.is-title-1.c-title(data-test='groupName') {{ ephemeral.invitation.groupName }}
         p.has-text-1(data-test='invitationMessage') {{ ephemeral.invitation.message }}
       .card
-        signup-form(v-if='isStatus("SIGNING")' :postEffect='accept')
-        login-form(v-else :postEffect='accept')
+        signup-form(v-if='isStatus("SIGNING")' :postSubmit='accept')
+        login-form(v-else :postSubmit='accept')
 
       p.c-switchEnter(v-if='isStatus("SIGNING")')
         i18n Already have an account?
--- a/frontend/views/pages/Join.vue
+++ b/frontend/views/pages/Join.vue
@@ -19,8 +19,8 @@ div
         h1.is-title-1.c-title(data-test='groupName') {{ ephemeral.invitation.groupName }}
         p.has-text-1(data-test='invitationMessage') {{ ephemeral.invitation.message }}
       .card
-        signup-form(v-if='isStatus("SIGNING")' @submit-succeeded='accept')
-        login-form(v-else @submit-succeeded='accept')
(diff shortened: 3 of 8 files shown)


frontend/views/components/AvatarUpload.vue
```diff
@@ -57,7 +57,7 @@ export default ({
       try {
         const { selector, contractID, key } = this.sbpParams
         await sbp(selector, { contractID, data: { [key]: picture } })
-        this.$refs.picture.setFromBlob(fileReceived)
+        // this.$refs.picture.setFromBlob(fileReceived)
```
taoeffect, line 64: Why is this being commented out?

I think that could cause problems... and it seems unrelated to the issues in the PR description.

frontend/views/pages/Join.vue
```diff
@@ -19,8 +19,8 @@ div
         h1.is-title-1.c-title(data-test='groupName') {{ ephemeral.invitation.groupName }}
         p.has-text-1(data-test='invitationMessage') {{ ephemeral.invitation.message }}
       .card
-        signup-form(v-if='isStatus("SIGNING")' @submit-succeeded='accept')
-        login-form(v-else @submit-succeeded='accept')
+        signup-form(v-if='isStatus("SIGNING")' :postSubmit='accept')
+        login-form(v-else :postSubmit='accept')
```
taoeffect, line 23: I don't understand how this change solves the problem.

The two lines `await this.postSubmit()` and `this.$emit('submit-succeeded')` are right next to each other, so what is the difference?

frontend/views/pages/GroupSettings.vue
```diff
@@ -181,6 +181,14 @@ export default ({
...
         this.openProposal('GroupLeaveModal')
       }
+    },
+    refreshForm () {
+      const { groupName, sharedValues, mincomeCurrency } = this.groupSettings
+      this.form = {
+        groupName,
+        sharedValues,
+        mincomeCurrency
+      }
```
taoeffect, line 191: I also don't understand which issue this is fixing and what it's doing.

Could you add a comment explaining in detail which issue this is fixing and why it works?

taoeffect (changes requested):
Nice work @Silver-IT! I have some questions, see review!

frontend/views/pages/Join.vue
Silver-IT: Nice question. What I did here is to make the `Submit` button loading until the user finish to join the group not only SignUp or SignIn. Actually, the new solution is related to the way of `ButtonSubmit` component works.
The loading spinner will be cycling until the `click` listener of that `ButtonSubmit` component is finished.
If the SignUp form emit an event and `Join` stuff is processed on `Join` page by listening this event, the `ButtonSubmit` spinner will be cycling only until that event is emitted, not until the `Join` process is finished.

frontend/views/pages/GroupSettings.vue
Silver-IT: In the previous way, the `form` variable is initiated once only when the component is created.
So when user switch group, the form is not updated to the details of new group so it displays the wrong data.
What I did here is to update this form every time the `groupSettings` is changed so that it can display the current group details.

frontend/views/components/AvatarUpload.vue
Silver-IT: It's one of the solution of the issue #1425.
As I described in the issue, the problem is that the group avatar is not updated when user switch group after upload picture once.
If user upload picture, the `Avatar` component loses the connect to the group profile picture because we set it from the Blob not from the group picture link.
Here, if user switch group, the group picture is not updated because it's already lose the connection

frontend/views/pages/Join.vue
taoeffect: Ah, I see, so it's because it waits on the async function `accept` before proceeding. Could you then add a comment here linking to that issue and explaining that we use `postSubmit` instead of `@submit-succeeded` because it supports waiting on async functions?

Silver-IT: Okay. Will do. I think I should make comments to the props of `SignUpForm` and `LoginForm` as well.

frontend/views/pages/GroupSettings.vue
taoeffect: Got it! Could you add a comment to that effect then?

Something along the lines of:

```js
// re-fetch the latest correct values whenever the user switches groups
```

frontend/views/components/AvatarUpload.vue
taoeffect: Ah, got it! Can you please add the following comment above this line then:

```js
// calling `setFromBlob` seems unnecessary here since the bound :avatar
// parameter should get updated after the upload completes.
// also, calling it here prevents the avatar from switching in Group Settings if we
// just uploaded a new image and then switch groups (see #1425)
```

frontend/views/pages/GroupSettings.vue
Silver-IT: I think I should add this comment in `watch` and did that way.

taoeffect (approved):
Approved! Great work @Silver-IT!

PR #1442, opened 2022-10-20
#658 Add a profile picture editor

### LINKED ISSUE #658: Add a profile picture editor and use it upon profile picture upload

SebinSong:
closes #658


taoeffect (commented):
Excellent work @SebinSong! This is looking really great! 😄 

Some quick feedback on things I noticed:

- User can drag the image anywhere - it should be bounded by the image itself, right?
- When testing mobile in FFDevEdition and updating my profile picture (in responsive mobile mode), I wasn't able to drag the zoom sliders for some reason.

SebinSong:
@taoeffect 
Updated the PR with the fix for those issues.

> When testing mobile in FFDevEdition and updating my profile picture (in responsive mobile mode), I wasn't able to drag the zoom sliders for some reason.

I did see this happening in Firefox devtool too but when I tried the same in chrome devtool everything works fine as you can see from below video.

https://user-images.githubusercontent.com/17641213/197938691-0b2d186e-757f-47a0-bd42-b02e5ec2eec0.mp4

More importantly, I tried it on my android phone (on chrome browser via the browsersync external link- http://172.30.1.56:3000), It did work fine there too. It seems like a bug on firefox dev edition itself. :)

Thanks,

frontend/views/components/avatar-editor/AvatarEditorModal.vue
```diff
@@ -0,0 +1,165 @@
...
+      button.is-outlined(
+        type='button'
+        @click='close'
+      ) Cancel
+
+      button-submit(
+        key='save'
+        data-test='saveBtn'
+        @click='submit'
+      ) Save
```
taoeffect: Missing i18n, try with `tag='button'` and `tag='button-submit'`, or include a child `i18n` element

taoeffect (changes requested):
Very clean and nicely done @SebinSong!

Minor comments-  also, it still allows me to drag the images beyond their borders / outside of their own bounding box.

Screen Shot 2022-10-26 at 2 13 24 PM

SebinSong:
@taoeffect  thanks for review. 

Thought I would need a help with clarifying what the below requirement means.
> it still allows me to drag the images beyond their borders / outside of their own bounding box.

Does this mean that the the clipping circle should never leave the boundary of the image? For example,

 

The red boundary is the bounding-box of the image on the canvas, whose size can change as the user zooms in & out of the image. and when the user moves the image around, Does the movement(or translation) of the image have to be limited to the points where the clipping circle never goes beyond the four edges of the red-boundary here?
To be detailed,


 

This is not ok because the image has been moved out of the clipping circle. and **the correct behaviour** would be,

 

The moving of image stops at this point, where the edges of the images are flush with the edges of the clipping circle.

Is this correct?
Hope my drawings here are being helpful here. :)

SebinSong:
@taoeffect 
here is the video demonstration of what I commented above.


https://user-images.githubusercontent.com/17641213/198231509-85525a99-cf6a-4e29-b92b-4ae0b43d2b1f.mp4


Is this the behaviour we want? cheers,

taoeffect:
@SebinSong Exactly right! 👍

SebinSong:
@taoeffect updated the PR with the work for the change request.

taoeffect (approved):
Brilliant! 🥇 😄 🎉

PR #1443, opened 2022-10-20
Direct Message Implementation

### LINKED ISSUE #1277: Implement DMs using chatroom.js contract
### Problem

DMs currently don't work.

### Solution

Implement DMs using the chatroom.js contract.

Replace the Mailbox contract with the Chatroom.js contract. In other words, the identity contract, instead of having a mailbox contract associated with it, should have a chatroom contract associated with it.

Silver-IT:
### Direct Message Implementation
Direct message doesn't depend on the group. Every user could have his own direct message users just like how it works in Discord. We use chatroom contract as direct message contract and every user's mailbox contract for saving his direct message contracts and their states.

#### GUI Guide
- In the right side of group chat page, there is a section of direct messages which displays the list of direct messages you already joined.
- At the top of this section, there is a `+ New` button. Clicking this button, you can see the direct messages which you haven't joined yet and you can join them by clicking.

#### How it works
Let's say `UserA` tries to do direct message with `UserB`.

- `UserA` tries to find direct message contract from his(`UserA's`) mailbox contract.
  - If exists, open the direct chatroom page with the chatroom contract he found.
- If `UserA` doesn't have direct message contract between `UserB` in his mailbox contract, he will create a new chatroom contract and join it.
- And then `UserA` saves this direct message contract in his mailbox contract and starts to synchronize this direct message contract.
- `UserA` makes `UserB` joins the direct message contract and also makes an event to save this direct message contract in `UserB`'s mailbox contract.
- Depending the `autoRequestAllownace` in `UserB`'s mailbox configuration, `UserB` could start synchronize this direct message contract or not.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "group-income",
   "version": "0.0.1",
-  "contractsVersion": "0.0.7",
+  "contractsVersion": "0.0.10",
```
taoeffect: We jumped from `.7` to `.10`?

I think this should be `.9`, right? (now that `.8` has been merged)

Silver-IT: [This PR](https://github.com/okTurtles/group-income/pull/1434) will make the contract version of 0.0.9. :grin:

frontend/model/contracts/mailbox.js
```diff
@@ -1,52 +1,162 @@
...
     validate: objectOf({
-      createdDate: string
+      createdDate: string, // action created date
+      username: optional(string), // action creator
+      identityContractID: optional(string) // action creator identityContractID
     }),
     create () {
+      if (!sbp('state/vuex/state').loggedIn) {
+        return { createdDate: new Date().toISOString() }
+      }
```
taoeffect, line 23: Could you add a comment explaining the situation when they might create a message for the mailbox but not be logged in?

frontend/controller/router.js
```diff
@@ -177,7 +177,9 @@ router.beforeEach((to, from, next) => {
     return next(from)
   }
 
-  document.title = to.meta.title
+  if (to.meta.title) {
+    document.title = to.meta.title
+  }
```
taoeffect: Ideally, we'd have the title correspond to whatever it is that we're doing.

So which view in the UI is this `if` statement here for? Wouldn't it be better to give it a title instead?

frontend/model/state.js
```diff
@@ -502,6 +489,41 @@ const getters = {
       return identityState && identityState.attributes
     }
   },
+  groupMembers (state, getters) {
```
taoeffect: How about: `globalProfilesForGroup` ?

frontend/model/state.js
```diff
@@ -502,6 +489,41 @@ const getters = {
...
+    }
+  },
+  ourContactProfiles (state, getters) {
+    const profiles = {}
+    const allProfiles = getters.groupsByName
+      .map(({ groupName, contractID }) => getters.groupMembers(contractID))
+      .flat()
+    const profilesSet = allProfiles
+      .filter((profile, pos) => profile && profile.username !== getters.ourUsername &&
+        allProfiles.findIndex(p => p.username === profile.username) === pos)
```
taoeffect: What is this doing and why? I am especially confused by `allProfiles.findIndex(p => p.username === profile.username) === pos`

Also, what is this getter supposed to be returning?

taoeffect (commented):
Some preliminary comments - still not finished going through, but getting these to you so that you don't wait until I finish reviewing everything

frontend/controller/actions/identity.js
```diff
@@ -195,6 +195,13 @@ export default (sbp('sbp/selectors/register', {
         // we must re-sync our identity contract again to ensure we don't rejoin a group we
         // were just kicked out of
         await sbp('chelonia/contract/sync', identityContractID)
+        const rootGetters = sbp('state/vuex/getters')
+        const { mailbox } = rootGetters.currentIdentityState.attributes
+        if (mailbox && !contractIDs.includes(mailbox)) {
+          sbp('okTurtles.data/set', 'SYNCING_MAILBOX', true)
+          await sbp('chelonia/contract/sync', mailbox)
+          sbp('okTurtles.data/set', 'SYNCING_MAILBOX', false)
```
taoeffect: Would be cool if we could DRY this somehow with Chelonia - since this seems to be a necessary trick that's used in multiple places.

For example, if you could check using Chelonia whether a specific `contractID` is currently being sync'd, would that allow us to remove these?

frontend/model/contracts/mailbox.js
```diff
@@ -1,52 +1,162 @@
...
-      validate: objectOf({
-        messageType: mailType,
-        from: string,
-        subject: optional(string),
-        message: optional(string),
-        headers: optional(object)
-      }),
-      process (message, { state }) {
-        state.messages.push(message)
+    'gi.contracts/mailbox/setAutoJoinAllowance': {
```
taoeffect: What is this?

Also - it is typically better / simpler to have a single action for setting all attributes, instead of having an action for each attribute. For example, see `'gi.contracts/identity/setAttributes'`

frontend/model/state.js
```diff
@@ -522,6 +544,18 @@ const getters = {
...
+    return (username: string) => getters.mailboxContract.users[username]?.contractID
+  },
+  usernameFromDirectMessageID (state, getters) {
+    return (chatRoomId: string) => {
+      if (!getters.isDirectMessage(chatRoomId)) {
+        return
+      }
+      return Object.keys(getters.mailboxContract.users)
+        .find(username => getters.directMessageIDFromUsername(username) === chatRoomId)
+    }
```
taoeffect, line 565: Couldn't there be multiple users in a DM?

frontend/controller/router.js
Silver-IT: > So which view in the UI is this if statement here for?

I noticed that the title is changed to `undefined` even for a few when we switching channel by channel.

> Wouldn't it be better to give it a title instead?

So you mean I should find where we don't set `meta.title` and set it. Right?

frontend/model/state.js
Silver-IT: That's possible. Will rename.

frontend/controller/router.js
taoeffect: > So you mean I should find where we don't set meta.title and set it. Right?

Yep!

frontend/model/state.js
Silver-IT: > What is this doing and why? I am especially confused by `allProfiles.findIndex(p => p.username === profile.username) === pos`

That's for removing duplications. User1 could be part of Group1 and also Group2. So Group1 and Group2 returns User1 simultaneously.

> Also, what is this getter supposed to be returning?

It supposed to return my contact profiles. This means the profiles of users who are any of my groups. In other words, the users I can create DM to.

Silver-IT: `allProfiles.findIndex(p => p.username === profile.username) === pos`

If there is better solution for removing duplications, I will gladly accept it. For example, using Set...

taoeffect: @Silver-IT You can loop through `allProfiles` in a single loop and set `profiles[profile.username] = profile` if `profile.username !== getters.ourUsername`.

In other words, you can modify the `for` loop on line 508 below to do that, and then delete the 3 lines above.

frontend/model/contracts/mailbox.js
Silver-IT: > What is this?

It's not in the requirement and there is no design for this yet. But I thought it could be better. No more syncing chatroom contracts if the user don't want new DM more.

> Also - it is typically better / simpler to have a single action for setting all attributes, instead of having an action for each attribute. For example, see 'gi.contracts/identity/setAttributes'

Yeah. That's great. Just like what we did in identity contract.

frontend/controller/router.js
Silver-IT: > Yep!

I will try to solve this that way.

frontend/model/contracts/mailbox.js
Silver-IT: We create mailbox contract before to create identity contract. That's when the user is not logged in and also registered either. I will add comments too.

frontend/model/state.js
Silver-IT: Adding guests is not implemented yet. And direct message is a channel between only two users.
If we invite guests, we should create a new chatroom (not DM) whose `type` is `CHATROOM_TYPES.INDIVIDUAL` and `privacyLevel` is `CHATROOM_PRIVACY_LEVEL.GROUP` in mailbox contract.

For a DM, there is only one partner.

shared/domains/chelonia/chelonia.js
```diff
@@ -270,9 +270,10 @@ export default (sbp('sbp/selectors/register', {
   },
   // 'chelonia/contract' - selectors related to injecting remote data and monitoring contracts
   // TODO: add an optional parameter to "retain" the contract (see #828)
-  'chelonia/contract/sync': function (contractIDs: string | string[]): Promise<*> {
+  'chelonia/contract/sync': async function (contractIDs: string | string[], syncDetector?: string): Promise<*> {
     const listOfIds = typeof contractIDs === 'string' ? [contractIDs] : contractIDs
-    return Promise.all(listOfIds.map(contractID => {
+    syncDetector && sbp('okTurtles.data/set', syncDetector, true)
+    await Promise.all(listOfIds.map(contractID => {
```
taoeffect: This is one way to do it, but I think there's another, simpler way, that doesn't require passing in `syncDetector` at all.

After line 274 we have `listOfIds` available, which tell us which contracts are about to be sync'd, and inside of the `group.js` and `chatroom.js` contracts, we are passed the `contractID` (or other contractIDs via other variables, like `data.chatRoomID`).

We can rewrite the contracts to directly check if their contractID is being sync'd, and then, I think, you will no longer need to create a string to pass in to `syncDetector`.

So, inside of `'chelonia/_init'`, you can add:

```js
this.currentSyncs = {}
```

Then, instead of modifying `'chelonia/contract/sync'`, inside of `'chelonia/private/in/syncContract'`, you can add next to `sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, true)`:

```js
this.currentSyncs[contractID] = true
```

And at the end (next to the other `CONTRACT_IS_SYNCING`), add:

```js
this.currentSyncs[contractID] = false
```

And finally, create a new selector in this file (`chelonia.js`): `'chelonia/contract/isSyncing'`:

```js
'chelonia/contract/isSyncing': function (contractID: string): ?boolean {
  return this.currentSyncs[contractID]
}
```

And then everywhere in the contracts replace the checks for strings like `'READY_TO_JOIN_CHATROOM'` with calls to `'chelonia/contract/isSyncing'` and pass in the `contractID`. `contractID` is passed in as a parameter to all `process` functions and all `sideEffect` functions.

taoeffect (changes requested):
Preliminary review

shared/domains/chelonia/chelonia.js
Silver-IT: @taoeffect, I tried this and finally found that this could not be solution.
Because of that we use that `okTurtles.data/set` for checking if the user is joining group or chatroom contract.
Actually, **joining** is quite different from **syncing**. We can sync the contracts when we login in another device.
So it breaks many test cases. I will add this `chelonia/contract/isSyncing` but don't think it's good fit.

Silver-IT: Using `chelonia/contract/isSyncing`, I have removed many `okTurtles.data/set` and `okTurtles.data/get` calls.
It is really awesome...

taoeffect: Awesome! 😄 🎉

shared/domains/chelonia/chelonia.js
```diff
@@ -270,9 +271,9 @@ export default (sbp('sbp/selectors/register', {
   },
   // 'chelonia/contract' - selectors related to injecting remote data and monitoring contracts
   // TODO: add an optional parameter to "retain" the contract (see #828)
-  'chelonia/contract/sync': function (contractIDs: string | string[]): Promise<*> {
+  'chelonia/contract/sync': async function (contractIDs: string | string[]): Promise<*> {
     const listOfIds = typeof contractIDs === 'string' ? [contractIDs] : contractIDs
-    return Promise.all(listOfIds.map(contractID => {
+    await Promise.all(listOfIds.map(contractID => {
```
taoeffect: Could you revert these changes in case having the return value ends up useful?

frontend/views/pages/Messages.vue
```diff
@@ -25,14 +25,14 @@ import Page from '@components/Page.vue'
 import { CHATROOM_TYPES } from '@model/contracts/shared/constants.js'
 import ChatNav from '@containers/chatroom/ChatNav.vue'
 import ChatMain from '@containers/chatroom/ChatMain.vue'
-import chatroom from '@containers/chatroom/chatroom.js'
+import ChatroomMixin from '@containers/chatroom/ChatroomMixin.js'
```
taoeffect, line 28: Nice improvement to the clarity of the codebase 👍

frontend/views/containers/chatroom/ChatroomMixin.js
```diff
@@ -61,30 +61,36 @@ const chatroom: Object = {
...
         description,
-        routerBack: type === CHATROOM_TYPES.INDIVIDUAL ? '/messages' : '/group-chat',
-        private: this.currentChatRoomState.attributes.privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE,
+        private: privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE,
         privacyLevel,
         general: this.generalChatRoomId === this.currentChatRoomId,
         joined: true,
-        creator,
-        picture
+        picture: type === CHATROOM_TYPES.INDIVIDUAL ? partner?.picture : undefined,
```
taoeffect, line 92: So... for this PR, based on the conversation we had on Slack about merging GMs & DMs, will you be updating this PR to do that? Or will you be leaving that for the other PR?

Because everything with `CHATROOM_TYPES.INDIVIDUAL` would need to be changed, and I want to make sure we don't forget about that

taoeffect (changes requested):
Partial review

shared/domains/chelonia/chelonia.js
Silver-IT: Okay.

frontend/views/containers/chatroom/ChatroomMixin.js
Silver-IT: Yeah. I am going to implement that in the #1463. Removing `dms` and `gms` in the mailbox contract.

By the way, even though we merge `dms` and `gms`, why do you think `CHATROOM_TYPES.INDIVIDUAL` need to be changed? Also we don't check `type` this way in the #1463, but I think there could be comparison between `type` and `CHATROOM_TYPES.INDIVIDUAL` anyway.

taoeffect: > By the way, even though we merge dms and gms, why do you think CHATROOM_TYPES.INDIVIDUAL need to be changed? 

Because if we are merging `dms` and `gms` - all chatroom types are of type `GROUP`. So to simplify you can also remove `CHATROOM_TYPES` entirely.

frontend/model/contracts/chatroom.js
```diff
@@ -246,7 +245,7 @@ sbp('chelonia/defineContract', {
         }
         Vue.delete(state.users, member)
 
-        if (!state.saveMessage) {
+        if (!state.saveMessage || state.attributes.type === CHATROOM_TYPES.INDIVIDUAL) {
```
taoeffect, line 248: So yeah, in this PR there's a bunch of references to `CHATROOM_TYPES.INDIVIDUAL`, but per the other comment, it seems like `CHATROOM_TYPES.INDIVIDUAL` and `CHATROOM_TYPES.GROUP` will be made the same, thus removing the need for this. Perhaps this could be done in this PR, instead of doing it in another PR?

frontend/views/containers/chatroom/ChatroomMixin.js
Silver-IT: To be honest, I don't think so.
There are two chatroom types; `CHATROOM_TYPES.INDIVIDUAL` and `CHATROOM_TYPES.GROUP`.
Chatroom of type `GROUP` is accessible only when they joined the specific group.
But chatroom of type `INDIVIDUAL` is accessible no matter they joined that group or not, just like DMs and GMs in Discord.

We can filter those chatrooms from group contract and mailbox contract.
group.chatRooms is listing the chatrooms of type `GROUP`, while chatrooms from mailbox are those of type `INDIVIDUAL`.
But I think we should set this difference as an attribute inside the chatroom contract.

Silver-IT: And also there is one more problem in merging `dms` and `gms` in mailbox contract.
Keys in `dms` are the usernames, while the keys in `gms` are contractID.
```
{
  dms: {
    alexjin: {},
    greg: {},
    sebin: {},
    ...
  },
  gms: {
    21XWnNHxRmhAxBPyr9aPB4VX1THB6d2S8qT1oGLREHNM4mNsuw: {},
    21XWnNVWxapZ9bGxcxqdoqbnSPHdEUccKucVqC5DaKeqKuYXDD: {},
    ...
  }
}
```
User opens DMs by clicking the username, not contractID.
But for the GMs, we can there is no other identifier except contractID.

Do you have any recommended solution for this?
Actually, after trying to merge these two, I have found this issue and also started to think that we should keep it and no merge `dms` and `gms`.

frontend/model/contracts/chatroom.js
Silver-IT: I think you misunderstood the usage and meaning of `CHATROOM_TYPES.INDIVIDUAL` and `CHATROOM_TYPES.GROUP`. It's not something to decide if the chatroom could let 2 or more people join to.
https://github.com/okTurtles/group-income/pull/1443#discussion_r1033123867

taoeffect: Ah, you're right, I did misunderstand. "Individual" means "direct message" and "group" is referring to the chatrooms in the group. 👍 

IMO, it would be clearer to rename `CHATROOM_TYPES.INDIVIDUAL` to `CHATROOM_TYPES.DIRECT_MESSAGE`, but if you strongly prefer `CHATROOM_TYPES.INDIVIDUAL` we can leave it at that

frontend/views/containers/chatroom/ChatroomMixin.js
Silver-IT: As @taoeffect suggetsed, the key could be like this `usernames.sort().join('|')` which is really awesome.
This will be implemented in the PR #1463 so I will resolve this conversation.

frontend/controller/actions/mailbox.js
```diff
@@ -24,5 +27,69 @@ export default (sbp('sbp/selectors/register', {
...
+          postpublish: params.hooks?.postpublish
+        }
+      })
+    } catch (e) {
+      console.error('gi.actions/mailbox/createDirectMessage failed!', e)
+      throw new GIErrorUIRuntimeError(L('Failed to create a new direct message channel.'))
+    }
+  },
+  ...encryptedAction('gi.actions/mailbox/setAutoJoinAllowance',
+    L('Failed to set Mailbox attributes. {attribute}', { attribute: 'autoJoinAllownace' })),
```
taoeffect: I think you forgot to delete this

taoeffect (changes requested):
Excellent work @Silver-IT!

It's looking pretty solid in my tests. :)

I left one comment, and also, two additional issues I'll note here:

1. Should be able to send a DM by clicking "Send Message" from the user profile card in the Dashboard:
  
   Screen Shot 2022-11-28 at 8 50 20 PM

2. When a DM is received by some user, they should see the red dot on the Chat bubble in the notification sidebar, but currently it doesn't show it:

   Screen Shot 2022-11-28 at 8 52 06 PM

The red dot should disappear when there are no unread messages from either chatrooms or DMs.

frontend/controller/actions/mailbox.js
Silver-IT: Nice catch. Thanks

taoeffect (approved):
Great work @Silver-IT!

PR #1444, opened 2022-10-20
No longer ignore @model/contracts/ in .flowconfig

### LINKED ISSUE #1440: Files in @model/contracts are ignored by Flow
### Problem

Files in `@model/contracts/` e.g. `~/frontend/model/contracts/group.js` have Flowtype annotations, but nonetheless appear to be skipped during typechecking with Flow.

### Solution

Use `<PROJECT_ROOT>/contracts/.*` rather than `.*/contracts/.*` in the `[ignore]` section of our `.flowconfig` file (see https://flow.org/en/docs/config/ignore/).
This way only the root `~/contracts/` folder will be skipped, rather than any subfolder named "contracts".

Then fix the newfound type errors if any!

snowteamer:
Closes #1440

### Summary of changes:
- Fixed two new Flow errors
- Bumped contracts version to `0.0.8` because of the Flow error fixed in `contracts/group.js`


taoeffect (approved):
Nice catch @snowteamer, and job well done! 👍

PR #1445, opened 2022-10-22
#1426 GroupCreationModal reachable even when no group is created

### LINKED ISSUE #1426: GroupCreationModal reachable even when no group is created
### Problem

The proper guards for joining a group are missing and little validation is done. Visiting <http://localhost:3000/app/?modal=GroupCreationModal&step=GroupWelcome> results in a success screen with empty group information (name), even when this screen was not reached as part of a group joining process.

### Solution

The state should be validated properly to ensure that this screen is reachable only upon joining a group successfully as part of the group creation flow.

SebinSong:
closes #1426

### DIFF
--- a/frontend/controller/router.js
+++ b/frontend/controller/router.js
@@ -168,6 +168,15 @@ const router: any = new Router({
 })
 
 router.beforeEach((to, from, next) => {
+  if (to.query.modal &&
+    !['SignupModal', 'LoginModal'].includes(to.query.modal) &&
+    !store.state.loggedIn) {
+    // if modal is queried and,
+    // the requested modal is only meant to be used post authentication and,
+    // the user is not logged in now, discard the navigation.
+    return next(from)
+  }
+
   document.title = to.meta.title
   next()
 })
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -4,7 +4,6 @@
     component(:is='subcontent[subcontent.length-1]')
 </template>
 <script>
-import { mapState } from 'vuex'
 import sbp from '@sbp/sbp'
 import { OPEN_MODAL, REPLACE_MODAL, CLOSE_MODAL, SET_MODAL_QUERIES } from '@utils/events.js'
 import { omit } from '@model/contracts/shared/giLodash.js'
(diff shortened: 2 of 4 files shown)


frontend/views/components/modal/Modal.vue
```diff
@@ -154,6 +158,11 @@ export default ({
     },
     setModalQueries (componentName, queries) {
       this.queries[componentName] = queries
+    },
+    isPreauthGuarded (toModal) {
```
SebinSong: While investigating this issue, noticed that we are able to open `GroupCreationModal.vue` even when the user is not logged in yet, when this modal is only meant to be used for authenticated users. So decided to go ahead and work on adding a guard logic in `$route` watcher of `Modal.vue` here. (Pls let me know if any of you think there is a better way of achieving this..!)

frontend/views/containers/access/SignupForm.vue
```diff
@@ -71,7 +70,6 @@ export default ({
     validationsDebouncedMixins
   ],
   components: {
-    ModalTemplate,
```
SebinSong, line 74: spotted an unused local component here, so removed it even though it's not related to this bug.

frontend/views/components/modal/Modal.vue
snowteamer: Can't this be achieved by adding a guard in `~/frontend/controller/router.js`?

SebinSong: @snowteamer agreed the `beforeEach()` guard in `router.js` is more suitable place to handle this logic so moved it there in the following commit.

frontend/controller/router.js
```diff
@@ -168,6 +168,15 @@ const router: any = new Router({
 })
 
 router.beforeEach((to, from, next) => {
+  if (to.query.modal &&
+    !['SignupModal', 'LoginModal'].includes(to.query.modal) &&
+    !store.state.loggedIn) {
+    // if modal is queried and,
+    // the requested modal is only meant to be used post authentication and,
+    // the user is not logged in now, discard the navigation.
```
taoeffect, line 176: Excellent comment 👍

taoeffect (approved):
LGTM!

PR #1452, opened 2022-11-03
#1447- prevent the payment method from being highlighted by double-clicking

### LINKED ISSUE #1447: Double-clicking on payment info shouldn't highlight payment method too
### Problem

When you double-click on the payment info to select it, it highlights both the payment info and the payment method:

double-click bug

### Solution

Make it so that only the payment info gets highlighted on double-click. Also test long-pressing on mobile (by using `grunt dev --tunnel` to access the site with your mobile phone).

SebinSong:
closes #1447

### DIFF
--- a/frontend/views/components/ProfileCard.vue
+++ b/frontend/views/components/ProfileCard.vue
@@ -220,6 +220,7 @@ export default ({
 
 .c-payment-type {
   padding-right: 0.5rem;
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Fancy CSS trick! Nice work @SebinSong! Love these 1-liners!

Issue #1449, opened 2022-11-01
Implement Activity Dashboard Tab


taoeffect:
### Problem

The [Activity Tab](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23766%3A357406) isn't implemented and needs to be.

### Solution

Go through an implement the streaks and inactivity fields (*except for the non-monetary contributions streak!* — that [is still being discussed](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=23766:357406#297122208) as we have no way to measure it at the moment).

Take care to:

- Use the group contract state to store information about streaks
- Store this information in an efficient way in the state (think to yourself: "how much information will this be to store in-memory for a group that's been running for 2 years?"). If the information is too big, then think about using `'gi.db/archive/*'` selectors instead of the in-memory state.
- Test thoroughly locally, and then bump the contracts versions before submitting a PR

You can split this issue up into multiple PRs, implementing different activity / inactivity labels.

SebinSong:
Will split the work into multiple PRs like below

<img src='https://user-images.githubusercontent.com/17641213/199736930-c9abd160-cf1b-46f7-aeec-da9922285d6a.png' width='560' >

SebinSong:
Hi @taoeffect 

Got questions regarding 'Streaks' part of this task.



Q1. **'Group streak of 100% support for N months'**
Does the streak have to be n 'most recent' histories? for examples,

Apr - 70%
**May - 100%
Jun - 100%
Jul - 100%**
Aug - 80%
Sep - 70%
**Oct - 100%
Nov - 100%**

if there is a support history like above, even though there was 3 streaks of 100% support in the past(May, Jun, Jul), do we still honour the 2 of the most recent streaks(Oct, Nov)? and if there is no >=2 number of latest 100% support records, Do we not show this sentence?

Q2. **'N members have on-time payment streaks'**
Likewise, does the streak we count here have to be the most recent ones? also, the count of streaks each member in N members has isn't necessarily the same, right?

Thanks,

taoeffect:
@SebinSong 

> Q1

That number counts the number of recent months (not including the current month) that had 100%. The second we have a month that doesn't have 100%, it resets back to 0. Does that make sense?

And keep in mind the Vuex state only stores 6 months worth of data for those numbers, so you will have to remember the current streak and increment by 1 each time we get a new 100%.

> Q2

[...]

Issue #1451, opened 2022-11-01
reactiveDate update code is wrong


taoeffect:
### Problem

This code is supposed to update `reactiveDate.date` only once we switch over to a new period, but that's not what it does:

```js
const reactiveDate = Vue.observable({ date: new Date() })
setInterval(function () {
  const date = new Date()
  // payments recalculation happen within a minute of day switchover
  if (Math.abs(reactiveDate.date.getTime() - date.getTime()) >= MINS_MILLIS) {
    reactiveDate.date = date
  }
}, 60 * 1000)
```

Instead, this code will cause `reactiveDate.date` to be updated every minute. That's not what we want. We want the getters to recalculate all of the payments within 1 minute of us entering a new period.

### Solution

The problem is, in this part of the code, we do not have access to any of the getters for calculating the current period stamp.. yet. So we need to carefully check for everything, and do something like this instead inside of `setInterval` (double-check this code):

```js
// We want the getters to recalculate all of the payments within 1 minute of us entering a new period.
const rememberedPeriodStamp = store.getters.periodStampGivenDate?.(reactiveDate.date)
const currentPeriodStamp = store.getters.periodStampGivenDate?.(new Date())
if (rememberedPeriodStamp !== currentPeriodStamp) {
  reactiveDate.date = new Date()
}
```

PR #1453, opened 2022-11-06
#1449 implement activity dashboard tab

SebinSong:
work on chunk 1 of #1449


frontend/model/contracts/group.js
```diff
@@ -141,6 +142,46 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
   return compareISOTimestamps(actionMeta.createdDate, userProfile.joinedDate) > 0
 }
 
+function updateGroupStreaks ({ state, getters }) {
+  const {
+    groupIncomeAdjustedDistribution,
+    thisPeriodPaymentInfo
+  } = sbp('state/vuex/getters')
```
taoeffect: Important to understand about contracts: they cannot access data outside of the contracts.

That is one immutable rule. Notice we're passing in `getters` via the function's parameters. Those are the only getters that are allowed.

Accessing `'state/vuex/getters'` is not allowed. If we had a way to detect data-access outside of the contracts, we would automate it as a hard error during linting. But we don't, so I'm acting as the linter. 😄 

The reason is because contracts must build up the same state consistently across clients. If we access data that is outside of the contracts, the state that will end up in the contracts will be inconsistent across different clients, and everything will explode.

As an example — `thisPeriodPaymentInfo` depends on `currentPaymentPeriod` Vuex getter, which in turn depends on `reactiveDate.date` — information that is different on each client. Therefore it cannot be used.

So - you can only access the getters that are defined on the contract - not the Vuex getters. The getters that are allowed are defined in this file and are accessible via the `getters` parameter that's being passed around.

taoeffect (changes requested):
Preliminary review.

frontend/model/contracts/group.js
taoeffect: > Accessing `'state/vuex/getters'` is not allowed.

*To be clear: the only place where accessing `'state/vuex/getters'` is allowed in contracts, is in sideEffects, because those do not affect the state of contracts themselves.

SebinSong: Thanks @taoeffect for a detailed explanation about this. I will keep the below in mind when changing any stuff related to the contract, moving forwards. :)

> Important to understand about contracts: they cannot access data outside of the contracts.

And the `updateGroupStreaks()` function has been updated so that it doesn't use vuex state for the logic.

frontend/model/contracts/group.js
```diff
@@ -141,6 +142,47 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
...
+  const cPeriod = getters.groupSettings.distributionDate
+  const thisPeriodPayments = getters.paymentsForPeriod(cPeriod)
+  const thisPeriodDistribution = adjustedDistribution({
+    distribution: unadjustedDistribution({
+      haveNeeds: getters.haveNeedsForThisPeriod(cPeriod),
+      minimize: getters.groupSettings.minimizeDistribution
+    }) || [],
+    payments: getters.paymentsForPeriod(cPeriod),
+    dueOn: getters.dueDateForPeriod(cPeriod)
+  })
```
taoeffect: Instead of re-calculating this, I *think* you should be able to get the value from the `lastAdjustedDistribution`, which is calculated and updated for you, see `updateAdjustedDistribution`.

The only question I have about that is what happens at the very beginning of the period, before any payments have been made. At that point, `lastAdjustedDistribution` hasn't been saved yet. Please consider this use-case and what the UI will show.

You might want to use the same trick I used in the `groupIncomeAdjustedDistribution` Vuex getter. It would also mean that you might need to create getters in `state.js`.

(Have you tested this functionality in the beginning of a 2nd payment period by setting up a group and adjusting your computer's clock?)

frontend/model/contracts/group.js
```diff
@@ -94,6 +94,7 @@ function updateCurrentDistribution ({ contractID, meta, state, getters }) {
   const noPayments = Object.keys(curPeriodPayments.paymentsFrom).length === 0
   // update distributionDate if we've passed into the next period
   if (comparePeriodStamps(period, getters.groupSettings.distributionDate) > 0) {
+    updateGroupStreaks({ state, getters })
```
taoeffect, line 108: You probably want to call `updateGroupStreaks` after `updateAdjustedDistribution` is called below, since that will save and update `lastAdjustedDistribution`

frontend/model/contracts/group.js
```diff
@@ -141,6 +142,47 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
...
+  })
+
+  // --- update 'fullMonthlyPledgesCount' streak ---
+  // if the group has made 100% pledges in this period, +1 the streak value.
+  // or else, reset the value to '0'
+  Vue.set(
+    streaks,
+    'fullMonthlyPledges',
+    thisPeriodDistribution.length === 0 ? streaks.fullMonthlyPledges + 1 : 0
+  )
```
taoeffect: This looks good 👍

frontend/model/contracts/group.js
```diff
@@ -141,6 +142,47 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
...
+
+  // --- update 'onTimePayments' streaks for 'pledging' members of the group ---
+  const filterMyItems = (array, username) => array.filter(item => item.from === username)
+  for (const username in getters.groupProfiles) {
+    if (getters.groupProfiles[username].incomeDetailsType !== 'pledgeAmount') continue
```
taoeffect: Because users can change this value during a payment period after it has already been locked, we shouldn't rely on / use this value.

Instead, use the `haveNeedsSnapshot` for the current payment period to figure out who was marked as being a pledger.

taoeffect (changes requested):
Preliminary review!

frontend/model/contracts/group.js
SebinSong: @taoeffect does sound good. 👍 

> Have you tested this functionality in the beginning of a 2nd payment period by setting up a group and adjusting your computer's clock?

No I actually repeatedly called `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` in dev console for testing this functionality, as I added this function call in there for dev purpose.

frontend/model/contracts/group.js
SebinSong: @taoeffect the reason why it is placed here is, I thought it would make sense to check & update the streaks status re payments right before the group's distribution period is updated to the next one, which is right below this line.
But like you mentioned in the above review comment, I can add logics where it uses the same trick in `groupIncomeAdjustedDistribution ` for a case when `lastAdjustedDistribution` has no value. :)

SebinSong: @taoeffect question here, I was actually looking for a place where the group's distribution period is set to the next one and all the state related to the payments are reset accordingly. does it seems like here(inside `updateCurrentDistribution()` and within `comparePeriodStamps(period, getters.groupSettings.distributionDate) > 0` block) is a correct place in that regard?

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -34,7 +33,7 @@
               //-   :args='{  members: 3 }'
               //-   @click='openModal("todoSeeMembers")'
               //- ) {members} members
-              i18n(:args='{ ...LTags("strong"), streaks: "4 weeks" }')  have {strong_} streaks of more than {streaks}{_strong}
+              i18n(:args='{ ...LTags("strong") }') have {strong_} on-time payment streaks{_strong}
```
Silver-IT, line 30: I can see the `i18n` sentence is updated, but couldn't find it from `strings` folder.

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -6,26 +6,25 @@
...
-        li.c-item-wrapper
+        li.c-item-wrapper(v-if='groupStreaks.fullMonthlyPledges > 1')
           .c-item
             .icon-star.icon-round.has-background-success.has-text-success
             .c-item-copy
               i18n(
-                :args='{ ...LTags("strong"), timeframe: "2 weeks", pourcent: 100 }'
-              ) Group has a streak of {strong_} {pourcent}% Support of {timeframe}{_strong}
+                :args='{ ...LTags("strong"), streak: groupStreaks.fullMonthlyPledges }'
+              ) Group has a streak of {strong_} 100% Support of {streak} months{_strong}
```
Silver-IT, line 15: Did you add this new string in `strings` folder?

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -88,18 +87,43 @@
...
+          membersOnStreak.push(getDisplayName(username))
+        }
+      }
+
+      const len = membersOnStreak.length
+      return len === 0
+        ? null
+        : {
+            members: membersOnStreak,
+            tooltipText: len === 1 ? this.L(`${len} member`) : this.L(`${len} members`)
```
Silver-IT: Does this `this.L()` works fine? Seems weird since we didn't define `L` in this component and also didn't import `L` from `@common/common`.

frontend/model/contracts/group.js
taoeffect: @SebinSong There are two places you might consider: one is in the Vuex getters (see `groupIncomeAdjustedDistribution` and how it relies on both `thisPeriodPaymentInfo` and `lastAdjustedDistribution`), and the other is either in `updateCurrentDistribution` or `updateAdjustedDistribution`.

I would look at all these places carefully to see how they're called and make a decision.

Silver-IT (changes requested):
Reviewed mostly focused on vue file because it would take some time to check the group contract monitoring how it works on my local. And that could break my concentration on my plan today so... Sorry. But I think @taoeffect made a great comments there.

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: @Silver-IT In case you didn't realise it, [here](https://github.com/okTurtles/group-income/blob/master/frontend/common/translations.js#L10) is the reference.

Silver-IT: Yeah, that makes sense.

frontend/model/state.js
```diff
@@ -44,6 +44,8 @@ setInterval(function () {
   // payments recalculation happen within a minute of day switchover
   if (Math.abs(reactiveDate.date.getTime() - date.getTime()) >= MINS_MILLIS) {
     reactiveDate.date = date
+
+    sbp('gi.actions/group/checkAndUpdateDistributionDate', { contractID: store.state.currentGroupId })
```
SebinSong: Please correct me if I'm wrong about anything in below explanations about this change I made.

so, we expect the app to update `distributionDate` of the group **as soon as the current period ends** but apparently that's **not** the case currently in the app, and this is a fix for that.

The only place where `distributionDate` is checked and updated currently is, inside `updateCurrentDistribution()` in group.js ([reference](https://github.com/okTurtles/group-income/blob/master/frontend/model/contracts/group.js#L96)). 
This means that until this function is executed by any further user action that runs it, the distribution period in the app will still remain the old one, even when the current period actually has ended. (I encountered this bug while changing around the clock of my machine locally but this issue holds true in real world scenario too I believe.)

What this new sbp selector (`gi.actions/group/checkAndUpdateDistributionDate`) does here is,
every time `reactiveDate.date` gets updated, we check whether the current distribution period has ended or not and if it has, we immediately update `distributionDate` accordingly. this way, we can minimise the discrepancy between the current period value and the actual time to **< 1 min**.

Hope this sounds as a good improvement and please let me know if there is any additional issue this might bring in. (I haven't encountered any yet.)

Thanks,

Thanks,

test/cypress/integration/group-chat-message.spec.js
```diff
@@ -214,7 +214,7 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
       cy.get('.c-message:nth-child(5) .c-emoticons-list>.c-emoticon-wrapper').should('have.length', 2)
     })
 
-    sendEmoticon(5, '+1', 1)
+    sendEmoticon(5, '+1', 2)
```
Silver-IT: Why this should be 2? I can not find any changes in test scenario.
Then, why is this changed?

frontend/model/contracts/group.js
taoeffect: @SebinSong OK, well, I would try testing by adjusting the computer clock, because a lot of code uses `periodBeforePeriod` and `periodAfterPeriod` getters - and those won't work with `forceDistributionDate`, because they work by adding 30 days to the period, so they would skip over the periods that were created because of `forceDistributionDate`

SebinSong: @taoeffect Thanks for providing more context, and yeah I did test it via adjusting the clock of my computer and it did seem to work well for me. please go ahead with reviewing that way. hope it works for you too.

test/cypress/integration/group-chat-message.spec.js
SebinSong: As we briefly had a chat about in the slack channel, This is a fix for below test failure. let me keep it for now until things get clarified and hope it sounds good.

frontend/controller/actions/identity.js
```diff
@@ -155,6 +155,8 @@ export default (sbp('sbp/selectors/register', {
             router.push({ path: '/dashboard' }).catch(console.warn)
           }
         }
+      } else {
+        sbp('gi.actions/group/updateLastLoggedIn', { contractID: state.currentGroupId })
```
taoeffect: If we put this here, then we'll miss the case where we switched to another group (line 150 above).

So I would suggest instead creating a second `if` statement, e.g. check for `state.currentGroupId` again:

```js
if (state.currentGroupId) {
  sbp('gi.actions/group/updateLastLoggedIn', { contractID: state.currentGroupId })
}
```

frontend/model/contracts/group.js
taoeffect: >  thought it would make sense to check & update the streaks status re payments right before the group's distribution period is updated to the next one, which is right below this line.

OK, I don't understand the reasoning for this though.

Doing it this way results in a DRY violation because you end up running the `adjustedDistribution` algorithm twice. So could we instead move this code below so that we avoid the DRY violation? Or is there some downside I'm not seeing?

taoeffect (changes requested):
Two comments/questions!

frontend/controller/actions/identity.js
SebinSong: Hi @taoeffect, the PR already does have this sbp call in `gi.actions/group/switch` too. :)
Pls check the line **248** below in this file.
The reason why I added it there too is, users might switch between multiple groups that they are part of after logging in and the `lastLoggedIn` field is for tracking user's activity for a 'group'.

SebinSong:
@taoeffect 

> Doing it this way results in a DRY violation because you end up running the adjustedDistribution algorithm twice. So could we instead move this code below so that we avoid the DRY violation? Or is there some downside I'm not seeing?

Yeah it would be redundant if the two times of calculation are for the same period. but they actually are not. `updateGroupStreaks()` has to be called **only** when the user just has passed into the next period, which is when `comparePeriodStamps(period, getters.groupSettings.distributionDate) > 0`. and as you can see from inside `updateGroupStreak` the distribution calculation is done for `groupSettings.distributionDate`, which is the **previous** period **at that point**. NOT for the **current** period (the new period the user just has passed into)

that calculated distribution data for the previous period is used for determining various things needed for counting streaks, such as, if there is any  unpaid pledge items, and if so who is the user that didn't complete the payments and so on..

But, like you advised in one of your previous comments, if `payments.lastAdjustedDistribution` field exists, that will be used instead so we don't unnecessarily run the distribution algorithm, like below.

`const thisPeriodDistribution = thisPeriodPayments.lastAdjustedDistribution || adjustedDistribution({ ... })`

Hope it helps clarifying everything.

Thanks,

frontend/controller/actions/identity.js
taoeffect: @SebinSong Oh! That's a great point! Since that isn't obvious, can you add a comment above line 159 explaining that?

SebinSong: @taoeffect done it!

frontend/model/contracts/group.js
```diff
@@ -141,6 +143,63 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
...
+  if (!thisPeriodPayments) return
+
+  const thisPeriodDistribution = thisPeriodPayments.lastAdjustedDistribution || adjustedDistribution({
+    distribution: unadjustedDistribution({
+      haveNeeds: getters.haveNeedsForThisPeriod(cPeriod),
+      minimize: getters.groupSettings.minimizeDistribution
+    }) || [],
+    payments: getters.paymentsForPeriod(cPeriod),
+    dueOn: getters.dueDateForPeriod(cPeriod)
+  })
```
taoeffect, line 181: One of the issues with DRY violations is, this code must be kept in sync with the other copy of it in `updateAdjustedDistribution`. For example, here we're missing a `.filter` to filter for active members only.

I'm still thinking on how to fix the DRY violation, but in case this helps you figure it out before me, I'm leaving this comment here now.

taoeffect: OK, yeah, I can't figure out a way to cleanly DRY this. However, that `.filter` should be copied over here so that it matches the other code.

Also, could you please a comment? something along the lines of:

```js
// IMPORTANT! This code must be kept in sync with updateAdjustedDistribution!
// TODO: see if it's possible to DRY this with the code inside of updateAdjustedDistribution
```

frontend/model/contracts/group.js
```diff
@@ -141,6 +143,63 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
   return compareISOTimestamps(actionMeta.createdDate, userProfile.joinedDate) > 0
 }
 
+function updateGroupStreaks ({ state, getters }) {
+  const streaks = state.streaks
```
taoeffect: If I'm not mistaken, to support the existing group we're using, you'll need to set `streaks` if it's not defined. E.g.

```js
Vue.set(state, 'streaks', initStreaks())
```

And use `initStreaks()` in the group constructor as well to DRY it.

frontend/model/contracts/group.js
```diff
@@ -647,6 +715,15 @@ sbp('chelonia/defineContract', {
           // handles proposal pass or fail, will update proposal.status accordingly
           proposals[proposal.data.proposalType][result](state, message)
           Vue.set(proposal, 'dateClosed', meta.createdDate)
+
+          // TODO: update 'streaks.noVotes'
```
taoeffect: Is this different from `groupStreaks.noVotes` below?

frontend/model/state.js
taoeffect: Hey @SebinSong - the issue is we don't want multiple clients to send this message all at once.

Also, this code I wrote for updating `reactiveDate` is wrong, see #1451.

But, even if #1451 is closed, we still would prefer to do things differently than have all the clients send `checkAndUpdateDistributionDate`, which is what this code currently does.

There are two ways to improve this:

1. Either come up with a scheme where only the first client to correctly notice the new period we're in sends the message.
2. Treat this more like the way we calculate the TODO list for new users when the new period rolls around.

I am in favor of (2) if it is possible (and I'm not sure if it is) because that doesn't require sending any updates to the contract at all. However, figuring out how to do (2) is more difficult. I will think on this in the meantime, but I encourage you to also consider both options. See how we handle the TODO list automatically updating without sending any messages (it's in `groupIncomeAdjustedDistribution` in `state.js`).

frontend/model/contracts/group.js
SebinSong: @taoeffect No the below logic is what this comment was meant for. forgot to delete it after updating it.

frontend/model/state.js
SebinSong: @taoeffect I used your reference to  `groupIncomeAdjustedDistribution` in `state.js` in the above comment as an inspiration and updated the logic to use `store.watch()` (pls refer to the change in `state.js`) instead of calling the contract action in the setInterval callback. I thought it would make more sense as a watcher. 
But I still haven't figured out how to resolve 
"1. Either come up with a scheme where only the first client to correctly notice the new period we're in sends the message." in here.

let me take more time trying to come up with a better way before/while you review my updated work, but pls let me know if yourself get to find a good solution for it. I will make changes accordingly.

frontend/model/contracts/group.js
SebinSong: done it!

frontend/model/contracts/group.js
SebinSong: done it!

frontend/model/state.js
SebinSong: Seems like I resolved it by [this commit](https://github.com/okTurtles/group-income/pull/1453/commits/7a79335a32d447ede770eed8b6bf36d42caa6276). it no longer sends multiple messages for `gi.actions/group/updateDistributionDate` when only one is enough. (but like I said, pls let me know if there is a room for improvement re this.)

frontend/model/state.js
```diff
@@ -623,6 +623,17 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
...
       getters: omit(contract.getters, omitGetters[contract.name] || [])
     })
+
+    if (contract.name === 'gi.contracts/group') {
+      store.watch(
+        (state, getters) => getters.currentPaymentPeriod,
+        (newPeriod, oldPeriod) => {
+          if (oldPeriod && newPeriod && (oldPeriod !== newPeriod)) {
+            sbp('gi.actions/group/updateDistributionDate', { contractID: store.state.currentGroupId })
+          }
```
taoeffect: Excellent improvement! This is a brilliant idea @SebinSong, a much better place to put this.

There is, I think, one remaining issue: this watcher is still likely to trigger one `updateDistributionDate` message per currently-online client, because they will all experience this transition, and therefore all will send `updateDistributionDate`.

So an addition `if` check is necessary: `getters.groupSettings.distributionDate !== newPeriod`

It is still possible that multiple `updateDistributionDate` messages will get sent, but with this check the likelihood of that should be significantly reduced.

Of course, you will need to obtain the getters in here somehow, and I recommend using `sbp('state/vuex/getters')` to do so.

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -34,72 +27,97 @@
...
+    ]),
+    onTimePayments () {
+      return Object.entries(this.groupStreaks.onTimePayments)
+        .filter(([username, streak]) => streak >= STREAK_ON_TIME_PAYMENTS)
+        .map(([username]) => this.userDisplayName(username))
+    },
+    missedPayments () {
+      return Object.entries(this.groupStreaks.missedPayments)
+        .filter(([username, streak]) => streak >= STREAK_MISSED_PAYMENTS)
+        .map(([username, streak]) => `${this.userDisplayName(username)} missed ${streak} payment${streak >= 2 ? 's' : ''}`)
```
taoeffect: \*cough\* i18n/L \*cough\*

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -34,72 +27,97 @@
...
+    },
+    noIncomeDetails () { // group members that haven't entered their income details yet
+      return Object.entries(this.groupProfiles)
+        .filter(([username, profile]) => !profile.incomeDetailsType)
+        .map(([username]) => this.userDisplayName(username))
+    },
+    noVotes () {
+      return Object.entries(this.groupStreaks.noVotes)
+        .filter(([username, streak]) => streak >= STREAK_MISSED_PROPSAL_VOTE)
+        .map(([username, streak]) => `${this.userDisplayName(username)} missed ${streak} vote${streak >= 2 ? 's' : ''}`)
```
taoeffect: here too :)

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -34,72 +27,97 @@
...
-                  div Rosalía
-                  div Ken M
-                  div Ines de Castro
-                  div Attila the Hun
-                  div Istralianda
-              i18n(:args='LTags("strong")')  haven´t {strong_} logged in past week {_strong}
-
-        li.c-item-wrapper
+              member-count-tooltip.c-member-count(:members='haventLoggedIn')
+              i18n(:args='{ ...LTags("strong"), daysCount: config.notLoggedInDays }') haven´t {strong_} logged in past {daysCount} days or more {_strong}
```
taoeffect: Two concerns here:

1. In some languages, the placement of "5 member" might be different, and by breaking up the translation, you may be generating a confusing / wrong translation. Sentences should always be translated as complete sentences, and never broken up for this reason. Sometimes it is technically difficult to do this, as might be the case here, but it's not unprecedented. For example, look at how `Payments.vue` handles opening the income modal when "income details" is clicked (lines 20-26)
2. Please use ' instead of ´ for the apostrophe (same issue below on line 63)

Similar concerns for the lines below.

frontend/model/contracts/group.js
```diff
@@ -647,6 +743,14 @@ sbp('chelonia/defineContract', {
           // handles proposal pass or fail, will update proposal.status accordingly
           proposals[proposal.data.proposalType][result](state, message)
           Vue.set(proposal, 'dateClosed', meta.createdDate)
+
+          const votedMembers = Object.keys(proposal.votes)
```
taoeffect, line 751: To make the code a little more readable, can you add a comment above this line to explain what the following lines are for?

taoeffect (changes requested):
Nice progress @SebinSong! A few more comments!

frontend/model/contracts/group.js
SebinSong: done it.

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: done it!

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: done it!

frontend/model/state.js
SebinSong: done it!

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: 2. has been resolved in the follow-up commit.

regarding 1., thanks for providing the reference to `Payments.vue`. it's actually more complicated case here than how things is handled in `payments.vue`, because the side effect we want to achieve is toggling the tooltip, and the way `Tooltip.vue` is built is it has to wrap around the whole element that triggers it. so I tried to figure out a way to inject `<tooltip></tooltip>` into the `i18n` component but didn't really work.
So I believe it requires building a new type of tooltip component that allows us to set a specific part of its content as the `trigger` element. And is it alright for me to work on this as a standalone issue (or a follow-up work of this issue) after this PR has been merged? would be easier to work with if it's handled as a separate concern. :)

hope it sounds good. Thanks,

taoeffect: @SebinSong Yeah, if you'd like to work on it in a followup please open up an issue and self-assign and reference this PR in it :)

frontend/controller/actions/identity.js
```diff
@@ -155,6 +155,9 @@ export default (sbp('sbp/selectors/register', {
             router.push({ path: '/dashboard' }).catch(console.warn)
           }
         }
+      } else {
+        // when the user has already switched into the group, record the log-in time for the group.
```
taoeffect: This doesn't quite say what you explained in your [previous comment](https://github.com/okTurtles/group-income/pull/1453#discussion_r1027051614).

If I were reading this comment for the first time, I would still be completely confused.

Something like this would be clearer:

```js
// We call updateLastLoggedIn in this else clause, instead of outside of it because
// in the `if` above, updateLastLoggedIn will get called by 'gi.actions/group/switch'
```

frontend/model/contracts/group.js
```diff
@@ -141,6 +153,83 @@ function isActionYoungerThanUser (actionMeta: Object, userProfile: ?Object): boo
...
+  const thisPeriodPayments = getters.groupPeriodPayments[cPeriod]
+  const noPaymentsAtAll = !thisPeriodPayments
+
+  if (streaks.lastStreakPeriod === cPeriod) return
+  else {
+    Vue.set(streaks, 'lastStreakPeriod', cPeriod)
+  }
+
+  // IMPORTANT! This code must be kept in sync with updateAdjustedDistribution!
+  // TODO: see if it's possible to DRY this with the code inside of updateAdjustedDistribution
```
taoeffect, line 171: Could you copy this same comment inside of `updateAdjustedDistribution` (except replace `updateAdjustedDistribution` with `updateGroupStreaks`)

frontend/model/contracts/group.js
```diff
@@ -208,6 +297,9 @@ sbp('chelonia/defineContract', {
           recentDate = recentDate.toISOString()
         }
         const { distributionDate, distributionPeriodLength } = getters.groupSettings
+
+        if (!distributionDate) return null
```
taoeffect, line 304: Why was this added?

As far as I know, this should never happen since the group contract is called with this initiated...

frontend/model/contracts/group.js
```diff
@@ -647,6 +743,15 @@ sbp('chelonia/defineContract', {
           // handles proposal pass or fail, will update proposal.status accordingly
           proposals[proposal.data.proposalType][result](state, message)
           Vue.set(proposal, 'dateClosed', meta.createdDate)
+
+          // update 'streaks.noVotes' which records the number of proposals that each member did NOT vote for
+          const votedMembers = Object.keys(proposal.votes)
+          for (const member of getters.groupMembersByUsername) {
+            const memberCurrentStreak = vueFetchInitKV(getters.groupStreaks.noVotes, member, 0)
+            const memberHasVoted = member === meta.username || votedMembers.includes(member)
```
taoeffect: Think this line could be simplified to simply `votedMembers.includes(member)`

frontend/model/state.js
```diff
@@ -623,6 +623,20 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
     store.registerModule(contract.name, {
       getters: omit(contract.getters, omitGetters[contract.name] || [])
     })
+
+    if (contract.name === 'gi.contracts/group') {
+      store.watch(
+        (state, getters) => getters.currentPaymentPeriod,
+        (newPeriod, oldPeriod) => {
+          if (!oldPeriod || !newPeriod) return
```
taoeffect, line 636: Because this code feels a bit out-of-place and random, could you add a comment above this line explaining why this function/code is here, including the specific scenario it's addressing?

taoeffect (changes requested):
Excellent work @SebinSong!

Some final cleanup / polishing steps before merging:

1. Please don't forget to open the issue [discussed here](https://github.com/okTurtles/group-income/pull/1453#discussion_r1032733561) and self-assign so that we don't forget about it.
2. Remember to re-run the `strings` command to generate the strings
3. After addressing the remaining comments remember to re-run `grunt pin:0.0.10 --overwrite`

frontend/controller/actions/identity.js
SebinSong: it sounds better. yeah.

frontend/model/contracts/group.js
SebinSong: @taoeffect This is actually a fix for below issue.



you can encounter this by removing this line and running `grunt dev`. pls let me know tho if there is better fix for this.

frontend/model/contracts/group.js
SebinSong: agree

frontend/model/contracts/group.js
taoeffect: Oh, ok, thanks, that looks like an OK fix then 👍

taoeffect (approved):
Approved! Excellent work @SebinSong!

Silver-IT (approved):
Awesome. Thanks for your work, @SebinSong

PR #1459, opened 2022-11-10
Improve archiving historical payments

### LINKED ISSUE #1448: Archive payments in indexed storage using pagination
### Problem

We have `gi.contracts/group/archivePayments` method in group contract which archives historical payments in Indexed Storage of the browser.
What this function does is to clone the state of historical `payments` and `paymentsByPeriod` to the indexed storage.
We need the payments in pagination, so we should find another way to save historical `payments` and `paymentsByPeriod` so that it could be easy to get them for a specific page and also in types(`Received` and `Sent`) easily.

Discussed here: https://github.com/okTurtles/group-income/pull/1434#discussion_r1008971602

### Solution

Need to divide `payments` and `paymentsByPeriod` into several parts using different keys so that it's easy to pick payments easily if necessary. Should TAKE CARE of the payments types of `Received` and `Sent`.
I think it's good to make keys for each of `Received` and `Sent`.

Silver-IT:
### Summary of Changes
- Closes #1448


Silver-IT:
@taoeffect, I made a comment [here](https://github.com/okTurtles/group-income/issues/1448#issuecomment-1313176378).
It describes what I did in this PR for the issue #1448.

taoeffect (changes requested):
@Silver-IT It's not clear to me what exactly this PR is doing, but from what I can tell it's not doing what it should.

The task is to make it so that historical payments (whether they are received or sent), are stored in batches, so that when we look through the UI via the pages and load a new page, if we've reached the end of the current batch, it loads the next batch.

This PR doesn't do that. It doesn't change how payments are stored, it just adds a new `archSentOrReceivedPayments` object that is saved. That's not what needs to happen.

Instead, we need to split up the payments into chunks, say 100 payments per chunk. Each of these chunks then needs to have a unique key.

Here's an example scenario:

1. User1 is looking at the payments they've sent, so they click on "Sent"
2. Chunk1 gets loaded with 100 of the most recent sent payments, creating 10 pages to go through.
3. User clicks page 2, it doesn't load anything from IndexedDB because we have 8 other pages worth of payments in memory already.
4. When they get to page 10, it sees that there are more archived payments in IndexedDB and loads those into memory too, creating say, 5 more pages (in other words, if there are 10 payments per page, then we loaded the last 50 payments).
5. The pagination gets updated. Now, instead of page 10 being the last page, page 15 is the last page.

The same happens when a User2 goes to have a look at the payments they've received, except they're showing just the payments they've received.

Silver-IT:
What I did so far is that we just filtered only `Sent` and `Received` payments and stored them individually.
That's because in `Payments` page, we don't display the payments unless they are `Sent` or `Received`.
We display only the payments of three types; `Sent`, `Received` and `Todo`.

Yes, you are right. I didn't implement **pagination** which is mentioned in the #1448.
That's because I thought there won't be too many `Sent` and `Received` payments in historical payments which are stored in Indexed Storage. And also I thought that doesn't affect the performance issue.

Another thing is that it could be messy.
Because every time the `currentPaymentPeriod` is changed, every time we archive payments, we should update all the payments chunks in Indexed Storage because payments inside chunks should be sorted by the `createdDate` and we should get the latest payments first when we access the historical payments.
**So every chunks needs to be updated if we archive the payments for the past period.**

@taoeffect, do you think we should really implement pagination in archiving payments?

In my opinion, just filtering `Sent` and `Received` payments and dividing them is quite enough.
And furthermore, we could slice the historical payments for the current page of type (`Sent`, `Received`, `Todo`) so that we can take only 10 payments in memory and display them.
To be honest, this is my best solution.

taoeffect:
@Silver-IT Replied on Slack!

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -80,8 +55,8 @@ const PaymentsMixin: Object = {
       }
       return payments
     },
-    async getHistoricalPaymentByHash (hash: string) {
-      const paymentsKey = `payments/${this.ourUsername}/${this.currentGroupId}`
+    async getHistoricalPaymentByHash (period: string, hash: string) {
```
SebinSong: @Silver-IT Could we maybe also update the name of this method to something that tells us passing a payment period is now required to use it? (like `getHistoricalPaymentByHashAndPeriod`)
(There is a large number of payment-related Vuex getters, group contract getters and utility methods across the front-end code-base with _more or less_ similar names, so the more detailed and explicit the names are, the more helpful it is, when trying to determine which one to use to solve a particular problem. :) )

SebinSong:
@Silver-IT Apart from one suggestion I commented, everything looks good. 
(Looking forwards to using the improved payment-related utilities in `paymentMixins.js` particularly)

frontend/views/containers/payments/PaymentsMixin.js
Silver-IT: That's reasonable. Will do that.

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -19,55 +19,30 @@ const PaymentsMixin: Object = {
...
-        const paymentsByPeriodKey = `paymentsByPeriod/${this.ourUsername}/${this.currentGroupId}`
-        const paymentsByPeriod = await sbp('gi.db/archive/load', paymentsByPeriodKey) || {}
-        const paymentHashes = paymentHashesFromPaymentPeriod(paymentsByPeriod[period])
-
-        const paymentsKey = `payments/${this.ourUsername}/${this.currentGroupId}`
-        const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
-        for (const hash of paymentHashes) {
-          detailedPayments[hash] = payments[hash]
-        }
+        return paymentHashes.map(hash => this.currentGroupState.payments[hash])
```
SebinSong: @Silver-IT There is two return statements in this function but apparently their types are different. it returns an array here whereas the one in below L:45 returns an object. Did you probably want to transform the array here to an object equivalent like `Object.fromEntries(paymentHashes.map(hash => [hash, this.currentGroupState.payments[hash]]))`

Silver-IT: It's really good catch. Thanks @SebinSong.

tq
```diff
@@ -0,0 +1,95 @@
+[1mdiff --git a/frontend/model/contracts/manifests.json b/frontend/model/contracts/manifests.json[m
```
taoeffect: What's this?

Silver-IT: It seems like a git something. I should remove it.

test/cypress/support/commands.js
```diff
@@ -90,6 +91,10 @@ Cypress.Commands.add('giLogin', (username, {
       }
       await sbp('gi.actions/identity/login', { username, password })
       await sbp('controller/router').push({ path: '/' }).catch(e => {})
+
+      return new Promise((resolve) => {
+        sbp('okTurtles.events/on', LOGIN, resolve)
+      })
```
taoeffect: I think this can be removed. And if for some reason it was needed, the promise should be created before `'gi.actions/identity/login'` is called, and it should be created with `/once` instead of `/on`. But I think this can be safely removed. The latest changes in `master` should have fixed recent heisenbugs

Silver-IT: Yeah. Removed. Nothing failed anyway.

frontend/model/contracts/group.js
```diff
@@ -1300,28 +1300,54 @@ sbp('chelonia/defineContract', {
...
-        const archPaymentsByPeriodKey = `paymentsByPeriod/${username}/${contractID}`
-        const archPaymentsByPeriod = await sbp('gi.db/archive/load', archPaymentsByPeriodKey) || {}
-        const archPaymentsKey = `payments/${username}/${contractID}`
-        let archPayments = await sbp('gi.db/archive/load', archPaymentsKey) || {}
+      const archPaymentsByPeriodKey = `paymentsByPeriod/${username}/${contractID}`
+      const archPaymentsByPeriod = await sbp('gi.db/archive/load', archPaymentsByPeriodKey) || {}
+      const archSentOrReceivedPaymentsKey = `sentOrReceivedPayments/${username}/${contractID}`
+      const archSentOrReceivedPayments = await sbp('gi.db/archive/load', archSentOrReceivedPaymentsKey) || { sent: [], received: [] }
+
+      const sortPayments = payments => payments.sort((f, l) => f.meta.createdDate < l.meta.createdDate ? 1 : -1)
```
taoeffect: Please use `compareISOTimestamps` here instead (but verify they're still sorted in the right order)

frontend/model/contracts/group.js
```diff
@@ -1300,28 +1300,54 @@ sbp('chelonia/defineContract', {
...
 
+      for (const period of Object.keys(paymentsByPeriod).sort()) {
```
taoeffect, line 1314: This function is very large and complicated and difficult to follow. It is made all the more difficult to follow by the lack of comments.

Could you improve its readability by using single-line comments above important sections and blocks explaining what they're about to do?

For example, a comment above this `for` loop could say something like this:

```js
// add the new archived payment data to the collection of all archived send and received payments
```

Good comments explain conceptually what's going on and what is the goal that we're trying to achieve.

taoeffect (changes requested):
Partial review

frontend/views/pages/Payments.vue
```diff
@@ -319,43 +318,21 @@ export default ({
...
-        })
-      }
-      // TODO: implement better / more correct sorting
-      return payments.sort((a, b) => a.date < b.date)
+      return this.historicalPayments.received.map(payment => ({
+        ...payment,
+        username: payment.meta.username, // fromUser
+        displayName: this.userDisplayName(payment.meta.username),
+        date: payment.meta.createdDate
+      }))
```
taoeffect, line 335: Nice simplification of this whole section! 👍

frontend/model/state.js
```diff
@@ -345,11 +345,11 @@ const getters = {
         for (const toUser in paymentsFrom[ourUsername]) {
           for (const paymentHash of paymentsFrom[ourUsername][toUser]) {
             const { data, meta } = allPayments[paymentHash]
-            payments.push({ hash: paymentHash, data, meta, amount: data.amount, username: toUser })
+            payments.push({ hash: paymentHash, data, meta, amount: data.amount })
           }
         }
       }
-      return payments
+      return payments.sort((paymentA, paymentB) => paymentA.meta.createdDate < paymentB.meta.createdDate ? 1 : -1)
```
taoeffect: Anywhere we're comparing ISOTimestamps, we should use `compareISOTimestamps`

Silver-IT: Okay.

frontend/model/state.js
```diff
@@ -367,13 +367,13 @@ const getters = {
...
                 const { data, meta } = allPayments[paymentHash]
-                payments.push({ hash: paymentHash, data, meta, amount: data.amount, username: toUser })
+                payments.push({ hash: paymentHash, data, meta, amount: data.amount })
               }
             }
           }
         }
       }
-      return payments
+      return payments.sort((paymentA, paymentB) => paymentA.meta.createdDate < paymentB.meta.createdDate ? 1 : -1)
```
taoeffect: Same here

taoeffect (approved):
Approved! Excellent work @Silver-IT!

PR #1462, opened 2022-11-17
Use nomodule script as fallback in main.html

### LINKED ISSUE #1460: Revert some changes to index.html's handling of unsupported browsers
### Problem

In attempting to fix issue #1435 (index.html issues) - PR #1439 displays the following screen to users on a supported browser who are on slow connections:

Screen Shot 2022-11-12 at 1 53 36 PM

Not catching this earlier was my fault btw!

### Solution

Revert to previous behavior. Users should only see that screen if they're actually on an unsupported browser, not because they're on a slow connection.

snowteamer:
Closes #1460

### Summary of changes:
- Added an inline script tag with the `nomodule` attribute in `main.html`: 
  -  Modern browsers with JS enabled should just run our main bundle and ignore the nomodule script.
  -  Browsers without ESM support will ignore the main bundle as it is a `type=module` script, and only run the nomodule script which will display an unsupported browser warning to the user.
  -  Browsers with JS disabled or without JS should just display the `noscript` contents.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -9,7 +9,7 @@
     <link rel="icon" href="/assets/images/group-income-icon-transparent.png" sizes="32x32">
   </head>
   <body>
-    <div class="modal is-active" id="fallback" style="visibility: hidden;">
+    <div class="modal is-active" id="fallback" style="display: none;">
       <div class="modal-content">
         <h1>Outdated browser</h1>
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
LGTM! Thanks for the PR @snowteamer - approved!

Issue #182, opened 2017-02-13
Implement SignUp.vue


taoeffect:
### Problem

It's not implemented.

### Solution

Implement it using new `TrustedNamespace` + `IdentityContract` APIs introduced in #180.

vijayee:
Fixed in #186

PR #1463, opened 2022-11-18
Adding guests to DMs

### LINKED ISSUE #1451: reactiveDate update code is wrong

### LINKED ISSUE #1461: Adding guests to direct message
### Problem

Just like in Slack, we should have a feature to add guests to the direct message channel.
In Group Chat, we already have this feature. Its name is `Invite member`.

What is the difference between `Invite members` to the group chat channel and `Add guests` to the direct message channel?
That's what it belongs to. Chatroom with guests belongs to the mailbox contract, meaning that users can access the chatroom no matter he left a group. But group chat channel is not accessible unless they are currently in the group.

### Solution

In the mailbox, create a method to add guests. What it does is to create a new chatroom whose `type` is `CHATROOM_TYPES.INDIVIDUAL` and `privacyLevel` is `CHATROOM_PRIVACY_LEVEL.GROUP` in mailbox contract.
Reference Link: https://github.com/okTurtles/group-income/pull/1443#discussion_r1021009639

### LINKED ISSUE #1488: Fix all usage of prepublish
### Problem

As @snowteamer discovered in #1487, relying on the message hash in the `prepublish` (or `prepublishContract`) hooks is not safe, because messages can be resent with different hashes during publishing.

Hence, code like this needs to be scrutinized for all usages of `message.hash()` and probably replaced with something else (if the hash is important):

```js
        hooks: {
          prepublish: (message) => {
            this.messages.push({
              pending: true
            })
            this.stopReplying()
            this.updateScroll()
          }
        }
      })
```

### Solution

Search the codebase for all usage of `prepublish` and `prepublishContract` that uses the messages `.hash()` and either rewrite it so that it doesn't rely on that hash, or uses a `postpublish` hook instead (where the hash is safe to rely on).

### LINKED ISSUE #1489: Implement notifications and sounds for the messages
### Problem
Currently, notifications and sounds are only working for the direct messages and mentions.
According to the updated design, it must be able to be configured in the notifications settings.
UI is already implemented in the PR #1463, so the functionality needs to be done.

### Solution
1. Save the configuration for notifications and sounds in the storage.
2. Make notifications and play sounds according to the configuration inside the function `messageReceivePostEffect` from `model/contracts/chatroom.js`.

### LINKED ISSUE #1492: Fix undefined exception in GroupMembers
### Problem

On our test server this code is throwing from `GroupMembers.vue`:

```js
    directMessages() {
      return this.ourContacts.filter(e=>Object.keys(this.mailboxContract.users).includes(e) && this.mailboxContract.users[e].joinedDate).map(e=>this.ourContactProfiles[e])
    }
```

Screen Shot 2023-02-11 at 3 36 45 PM

### Solution

Find and fix the bug. Based on the error message it might be that `this.mailboxContract.users` is `undefined`

### LINKED ISSUE #1494: Several weird issues in joining group and group chat
### Problem

1. Wrong/undefined `chatroomId` for each group as soon as a new group is created
  Ways to reproduce
  a) register a new user and create a new group
  b) go to the group chat page and open General chatroom
  c) create a new group
  d) go to group chat page
2. Possibility to leave channel when he is the creator
  It's possible but the function should be to `Leave and Archive`.
3. Wrong greeting as soon as a new group is created
  Ways to reproduce:
  a) register a new user and create a new group
  b) go to the group chat page and open General chatroom
  c) create a new group
4. It's broken when the incorrect chatRoomId is attached inside the URL

### Solution

1. Needs to debug and fix them.
2. Add these test cases

### LINKED ISSUE #1495: Chatroom contract uses message "pending" attribute incorrectly
### Problem

I'm not 100% sure, but it seems like the `chatroom.js` contract is handling state incorrectly.

Notice here a search for `msg.pending`:

```js
    'gi.contracts/chatroom/addMessage': {
      validate: messageType,
      process ({ data, meta, hash }, { state }) {
        if (!state.saveMessage) {
          return
        }
        const pendingMsg = state.messages.find(msg => msg.id === hash && msg.pending)
        if (pendingMsg) {
          delete pendingMsg.pending
        } else {
          state.messages.push(createMessage({ meta, data, hash, state }))
        }
      },
```

I could not find where `.pending` is set as part of the data sent to contracts. Instead, I found this in `ChatMain.vue`:

```js
    editMessage (message, newMessage) {
      sbp('gi.actions/chatroom/editMessage', {
        data: {
          id: message.id,
          createdDate: message.datetime,
          text: newMessage
        },
        hooks: {
          prepublish: (msg) => {
            message.text = newMessage
            message.pending = true
          }
        }
      })
    },
```

Here it appears like `.text` and `.pending` attributes are set on a `GIMessage` before it is published. This is incorrect - GIMessages should never be modified like this. And furthermore, as far as I can tell these modifications should do nothing, since the GIMessage should ignore these values (as they are unused during serialization and sending of the GIMessage).

So it's unclear to me where even the `.pending` is set in a way that it appears on messages in `state.messages`.

But the point is - the `state` (and therefore, `state.messages`) needs to be consistent across clients. And for this reason it can only be modified by the contracts, not outside of them. If the contract state becomes inconsistent between different clients, this will lead to serious bugs.

### Solution

Help me understand what's going on, via Slack or a video call, and come up with a solution to this, and make sure that `GIMessages` aren't modified in `prepublish`.

Silver-IT:
### Summary of Changes

- Closes #1461
- Closes #1451 
- Closes #1489 
- Closes #1492 
- Closes #1494 
- Closes #1495 
- Closes #1488 
- Implement test scenario for Adding guests to DM
- Updated Notifications section in User Settings page (Only UI Implementation)

### More things to be done regarding adding guests
- [x] #1493


frontend/controller/actions/group.js
```diff
@@ -306,10 +297,10 @@ export default (sbp('sbp/selectors/register', {
         // But when user joins the group, he needs to ignore all the actions about chatroom
         // Because the user is joining group, not joining chatroom
         // and he is going to make a new action to join 'General' chatroom AGAIN
-        // While joining group, we don't set this flag because Joining chatroom actions are all OLD ones, which needs to be ignored
-        // Joining 'General' chatroom is one of the step to join group
+        // While joining group, we don't set this flag because Joining chatroom actions are all OLD ones, which need to be ignored
+        // Joining 'General' chatroom is one of the steps to join group
```
taoeffect: This comment references `READY_TO_JOIN_CHATROOM` and `JOINING_GROUP`, neither of which are being used in the code here anymore, so seems like the comment needs to be updated to explain what's going on now with respect to `JOINING_GROUP_CHAT` instead?

frontend/model/contracts/group.js
```diff
@@ -1044,13 +1044,13 @@ sbp('chelonia/defineContract', {
...
         if (username === rootState.loggedIn.username) {
-          if (!sbp('okTurtles.data/get', 'JOINING_GROUP') || sbp('okTurtles.data/get', 'READY_TO_JOIN_CHATROOM')) {
+          if (!sbp('okTurtles.data/get', 'JOINING_GROUP') || sbp('okTurtles.data/get', 'JOINING_GROUP_CHAT')) {
             // while users are joining chatroom, they don't need to leave chatrooms
             // this is similar to setting 'JOINING_GROUP' before joining group
             sbp('okTurtles.data/set', 'JOINING_CHATROOM_ID', data.chatRoomID)
             await sbp('chelonia/contract/sync', data.chatRoomID)
             sbp('okTurtles.data/set', 'JOINING_CHATROOM_ID', undefined)
-            sbp('okTurtles.data/set', 'READY_TO_JOIN_CHATROOM', false)
+            sbp('okTurtles.data/set', 'JOINING_GROUP_CHAT', false)
```
taoeffect: I think `JOINING_CHATROOM_ID` can be removed too?

The only place it is checked is in the sideEffect for `'gi.contracts/chatroom/delete'`, and there, instead of checking for `JOINING_CHATROOM_ID`, it could use `'chelonia/contract/isSyncing'` right?

frontend/model/contracts/mailbox.js
```diff
@@ -1,52 +1,234 @@
...
+        username: string
+      }),
       process ({ data }, { state }) {
+        const initialState = merge({
+          attributes: {
+            creator: data.username,
+            autoJoinAllowance: true // this attribute could be used to block him to be joined direct/group messages automatically by another
+          },
+          dms: {}, // direct message
+          gms: {} // group message
```
taoeffect: Shouldn't it be possible to design this so that `dms` and `gms` are the same thing?

taoeffect (changes requested):
Partial review, leaving a few comments to get them to you quickly. Will continue reviewing!

Also - noticed that there are two versions of the contracts in this PR: `0.0.10` and `0.0.13` - should only be 1

frontend/model/contracts/mailbox.js
Silver-IT: Sorry, @taoeffect. I don't understand what you mean?
Can you please tell me more in detail? Or taking an example?

frontend/controller/actions/group.js
Silver-IT: Nice catch.

frontend/model/contracts/group.js
Silver-IT: Nice catch too.

frontend/model/contracts/mailbox.js
taoeffect: @Silver-IT let's move this to Slack, I'll reply there

Silver-IT:
For merging `dms` and `gms`, we discussed to use the way to name the key of each chatroom contract in mailbox contract.

The following was how the mailbox contract designed previously.
```
mailbox: {
  dms: {
    [username]: Object
  },
  gms: {
    [contractID]: Object
  }
}
```
And the following is how we are going to design mailbox contract.
```
mailbox: {
  chatRooms: {
    [sortedUsernames]: Object // e.g. 'alexjin|greg|sebin'
  }
}
```
I have found the lack of this format. If we set the key by the usernames, we should make an action to everyone's mailbox contract for every joining. So I think we should use `contractID` as a key.

frontend/controller/actions/identity.js
```diff
@@ -146,7 +146,22 @@ export default (sbp('sbp/selectors/register', {
...
         }
       }
-      // note: leaving groups will happen when we sync the removeOurselves message
+      // NOTE: should sync all the identity contracts which are not part of same group
+      // these users could be the ones from 12n direct messages
+      const chatRoomUsers = [...new Set(
+        Object.keys(
+          pickWhere(state.contracts, ({ type }) => type === 'gi.contracts/chatroom')
+        ).map(cID => Object.keys(state[cID].users)).flat())
+      ]
```
SebinSong: @Silver-IT (Assuming it's getting rid of any possible duplicated username from an array here) Seems like this can be DRYed by using `uniq` function [here ](https://github.com/okTurtles/group-income/blob/0439e89600e5af9eae07f7a2190144cdfa1d0264/frontend/model/contracts/shared/giLodash.js#L132)in `giLodash.js`

frontend/views/containers/chatroom/ChatNotificationSettingsModal.vue
```diff
@@ -0,0 +1,129 @@
...
+      fieldset.is-column
+        legend.legend Send notifications for:
+        label.radio
+          input.input(
+            type='radio'
+            name='messageNotification'
+            :value='options.ALL_MESSAGES'
+            v-model='$v.form.messageNotification.$model'
+          )
+          span All new messages
```
SebinSong: @Silver-IT is it alright not to translate sentence here (also other sentences below) using `i18n`?

frontend/views/containers/chatroom/ChatMembersAllModal.vue
```diff
@@ -212,18 +233,44 @@ export default ({
...
+              ? {
+                  displayName: this.globalProfile(username).displayName,
+                  username,
+                  departedDate: null
+                }
+              : {
+                  displayName: this.ourContactProfiles[username].displayName,
+                  username,
+                  departedDate: null
+                }
```
SebinSong: @Silver-IT Not a big deal but looks like the return block here can be simplified as
`return { username, departedDate: null, displayName: condition ? value1 : value2 }`

SebinSong:
Great work @Silver-IT. according to your request for a review, Left a few comments here.

frontend/views/containers/chatroom/ChatNotificationSettingsModal.vue
Silver-IT: Nice catch. I think I missed that part. We should use `i18n`.

frontend/views/containers/chatroom/ChatMembersAllModal.vue
Silver-IT: Correct, we can simplify. For better understanding, I am doing this way.
```
const profile = username === this.ourUsername ? this.globalProfile(username) : this.ourContactProfiles[username]
return {
  displayName: profile.displayName,
  username,
  departedDate: null
}
```

frontend/controller/actions/identity.js
Silver-IT: Yeah. We can do this way.
```
const chatRoomUsers = uniq(Object.keys(
  pickWhere(state.contracts, ({ type }) => type === 'gi.contracts/chatroom')
).map(cID => Object.keys(state[cID].users)).flat())
```

SebinSong (dismissed):

frontend/controller/actions/identity.js
```diff
@@ -146,7 +146,20 @@ export default (sbp('sbp/selectors/register', {
...
       }
-      // note: leaving groups will happen when we sync the removeOurselves message
+      // NOTE: should sync all the identity contracts which are not part of same group
+      // but from the direct messages invited by another
+      const chatRoomUsers = uniq(Object.keys(
+        pickWhere(state.contracts, ({ type }) => type === 'gi.contracts/chatroom')
+      ).map(cID => Object.keys(state[cID].users)).flat())
+      const additionalIdentityContractIDs = await Promise.all(chatRoomUsers.filter(username => {
+        return getters.ourUsername !== username && !getters.ourContacts.includes(username)
+      }).map(username => sbp('namespace/lookup', username)))
```
taoeffect, line 156: `sbp('namespace/lookup', username)` returns a Promise, so the resulting array will be full of promises, not strings.

*EDIT: it could also return strings (if the value was previously cached). but it should be treated as promises

taoeffect: One way to address this would be to, below, instead of:

```js
await sbp('chelonia/contract/sync', identityContractID)
```

Do:

```js
// 'identityContractID' could be a promise
await sbp('chelonia/contract/sync', await identityContractID)
```

frontend/controller/actions/mailbox.js
```diff
@@ -58,26 +58,28 @@ export default (sbp('sbp/selectors/register', {
...
       })
 
+      await sbp('chelonia/out/actionEncrypted', {
+        ...omit(params, ['options', 'data', 'hook']),
+        data: {
+          username: params.data.username,
+          contractID: message.contractID()
+        },
+        action: 'gi.contracts/mailbox/createDirectMessage'
+      })
```
taoeffect: What `contractID` is this message being sent to? It doesn't seem to be specified, unless it's part of `params`?

BTW, I don't see `'gi.actions/mailbox/createDirectMessage'` called anywhere in the code! Why is that?

frontend/controller/actions/identity.js
Silver-IT: @taoeffect, I think `additionalIdentityContractIDs` is the array of strings because `await` is used before `Promise.all()`.

taoeffect: Not sure how I missed that - you're right!

frontend/controller/actions/mailbox.js
Silver-IT: > BTW, I don't see 'gi.actions/mailbox/createDirectMessage' called anywhere in the code! Why is that?

That's because the action name is set dynamically according to the existence of direct message.
`Line #165 in ProfileCard.vue`
`Line #182 in NewDirectMessageModal.vue`

> What contractID is this message being sent to? It doesn't seem to be specified, unless it's part of params?

If you check those two lines above, you can see the contractID there.

frontend/model/contracts/mailbox.js
```diff
@@ -98,63 +127,108 @@ sbp('chelonia/defineContract', {
...
+        // TODO: need to validate if meta.username is part of the chatroom of contractID
+      },
+      process ({ meta, data }, { state }) {
+        if (state.chatRooms[data.contractID]) {
+          throw new TypeError(L('Already existing group message channel.'))
+        }
+
+        const joinedDate = state.attributes.autoJoinAllowance ? meta.createdDate : null
+        Vue.set(state.chatRooms, data.contractID, {
+          hidden: false, // TODO: this hidden attribute should be there in state.js as global
```
taoeffect: What does this comment mean?

Remember - global state in state.js cannot be used here, since inside of contracts the only state that can be used is the state that's passed in via the parameters to `process` (and constants that are defined in the contract file itself).

Silver-IT: Yeah. Correct, we can not use the global state inside the contract state because it could be different from user to user.
What I meant is that we should define `hidden` value in the global state not in the contract state.
I will re-write the comment or remove the `hidden` attribute then. Because `hidden` attribute is not using at the moment.

Silver-IT: I thought wrong. This shouldn't be there in global state. I will re-write the comment.

taoeffect: If you mean that the default is defined in state.js, and then it is passed in via the parameters to `process` (because it's included as part of the `data` in the message, then that makes sense 👍

frontend/model/contracts/mailbox.js
```diff
@@ -98,63 +127,108 @@ sbp('chelonia/defineContract', {
...
       }),
-      process ({ meta, data }, { state }) {
-        if (state.attributes.creator === data.username) {
-          if (state.users[meta.username]) {
-            throw new TypeError(L('Already existing direct message channel.'))
-          }
-        } else if (state.attributes.creator === meta.username) {
-          if (!state.users[data.username] || state.users[data.username].joinedDate) {
+      process ({ meta, data }, { state, getters }) {
+        const me = state.attributes.creator
```
taoeffect: It's confusing to call this `me` - as it's not clear who "me" is referring to.

I recommend, to avoid confusion, writing this as something like:

```js
const { creator } = state.attributes
```

frontend/model/contracts/mailbox.js
Silver-IT: Let's think about the workflow once more.

If I hide the direct message by clicking 'X' button or any other, that direct message will be invisible in the current device.
Then, I login in another device and check the direct message visibility.
What should it be? `Visible` or `Invisible`. I think it should be `Invisible`. So the `hidden` attribute should be there inside the `mailbox` contract I think.

That's why I think I should re-write the comment.

```
hidden: false, // NOTE: this attr is used to hide/show direct messages
```

frontend/model/contracts/mailbox.js
Silver-IT: Yeah, it's confusing actually. I think I should use `creator` instead and also add comment.

frontend/model/contracts/mailbox.js
taoeffect: Yes, that sounds good 👍

frontend/model/contracts/mailbox.js
```diff
@@ -98,63 +127,108 @@ sbp('chelonia/defineContract', {
...
+        if (me !== meta.username) {
+          Vue.set(state.chatRooms, data.contractID, {
+            partner: data.username,
             hidden: false,
             joinedDate
           })
         } else {
-          Vue.set(state.users[data.username], 'joinedDate', joinedDate)
+          const contractID = getters.ourOneToOneDirectMessages[data.username].contractID
+          Vue.set(state.chatRooms[contractID], 'joinedDate', joinedDate)
```
taoeffect: This isn't making much sense to me... I thought we agreed that DMs and GMs would be merged into the same thing?

Why then do we have a special getter for `ourOneToOneDirectMessages` and a separate action `'gi.actions/mailbox/createGroupMessage'`?

Notice in Slack 1-on-1 DMs can be turned into multi-user GMs:

Screen Shot 2023-02-14 at 5 11 02 PM

So we can simplify our code and treat all DMs as GMs, right?

frontend/model/contracts/mailbox.js
Silver-IT: I think I should add comment here. Actually, that is referring to `me` that's because `state.attributes.creator` is the creator who creates this mailbox contract when registers the group income account.

taoeffect: I would rename the variable to `creator` - keeping it as `creator`. "me", in English, is too confusing. These actions could potentially be called by anyone, in any context and any situation. And so for thinking clearly about this, `creator` makes more sense, because it's specific to what it's referring to. It doesn't even need a comment. But `me` doesn't make sense

Silver-IT: Okay, then. I will rename it to `creator` and add comment.

frontend/model/contracts/mailbox.js
Silver-IT: @taoeffect, as you described in Slack, we could add a list of users in the dms in mailbox contract state.
```
{ dms: { [chatroomID1]: { hidden: false, users: [ user1contractID, user2contractID, user3contractID ] } } }
```
While I am trying to do this, I found an issue. Let's imagine the following situation.
If someone create a direct message and add 4 people (so 5 people totally), 5 people would sync the chatroom contract together and 5 mailbox contracts should be up-to-date to the chatroom contract states (at least users inside).
At this moment, if anyone is added to this direct message (so totally 6 people), a new people would sync the chatroom contract and create a new `dms` field in his mailbox contract and the previous 5 people should update their mailbox contracts because there is a `users` field inside. I think this is a not good in terms of efficiency.

I wanted to design the mailbox contract that wouldn't be reactive to any of the chatroom contract state. So the problem is that I want to remove `users` field in the `dms` of mailbox contract. But I will try to find a solution to merge `createDirectMessage` and `createGroupMessage`.

frontend/controller/actions/mailbox.js
```diff
@@ -31,18 +31,26 @@ export default (sbp('sbp/selectors/register', {
...
-      if (!partnerProfile) {
+      const hasEmptyProfile = partnerProfiles.some(profile => !profile)
+      if (hasEmptyProfile) {
         throw new GIErrorUIRuntimeError(L('Incorrect username to create direct message.'))
+      } else if (!partnerProfiles.length) {
+        throw new GIErrorUIRuntimeError(L('Not enough usernames provided.'))
+      } else if (isPrivacyLevelPrivate && partnerProfiles.length !== 1) {
+        throw new GIErrorUIRuntimeError(L('Incorrect number of usernames provided.')) // TODO: need to add translation
+      } else if (isPrivacyLevelPrivate && rootGetters.ourPrivateDirectMessages[params.data.usernames[0]]) {
+        throw new GIErrorUIRuntimeError(L('Already existing direct message channel.'))
```
taoeffect: Is this an error that can be triggered by the user? If so, why? Couldn't we just immediately take them to that chatroom instead of giving an error? And if it's meant to be an error that only the programmer sees, then it doesn't need to be translated.

frontend/model/state.js
```diff
@@ -24,13 +24,15 @@ const initialState = {
...
+  notificationSettings: {}, // { messageNotification: MESSAGE_NOTIFY_SETTINGS, messageSound: MESSAGE_NOTIFY_SETTINGS }
   contracts: {}, // contractIDs => { type:string, HEAD:string } (for contracts we've successfully subscribed to)
   pending: [], // contractIDs we've just published but haven't received back yet
   loggedIn: false, // false | { username: string, identityContractID: string }
   namespaceLookups: Object.create(null) // { [username]: sbp('namespace/lookup') }
 }
 
-if (window.matchMedia) {
+if (window
+  .matchMedia) {
```
taoeffect: This looks like an accidental change

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -381,12 +381,7 @@ export default ({
     deleteMessage (message) {
       sbp('gi.actions/chatroom/deleteMessage', {
         contractID: this.currentChatRoomId,
-        data: { id: message.id },
-        hooks: {
-          prepublish: (msg) => {
-            // need to do something
-          }
-        }
+        data: { id: message.id }
```
taoeffect, line 385: Above this line you forgot to remove the `prepublish` and move the code for `'gi.actions/chatroom/editMessage'`

frontend/views/containers/chatroom/ChatMembersAllModal.vue
```diff
@@ -270,6 +312,51 @@ export default ({
...
+            hooks: {
+              postpublish: (message) => {
+                sbp('gi.actions/mailbox/joinDirectMessage', {
+                  contractID: profile.mailbox,
+                  data: {
+                    privacyLevel: CHATROOM_PRIVACY_LEVEL.GROUP,
+                    contractID: this.currentChatRoomId
+                  }
+                })
+              }
```
taoeffect: This seems like it could be simply placed after the first call to `await sbp('gi.actions/chatroom/join'` instead of being inside of `postpublish`

frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -42,21 +42,30 @@ export default ({
...
+        },
+        default: {
+          [MESSAGE_NOTIFICATIONS.ADD_MEMBER]: L('Added a member to {title}: {displayName}', { displayName, title: channelName }),
+          [MESSAGE_NOTIFICATIONS.JOIN_MEMBER]: L('Joined {title}', { title: channelName }),
+          [MESSAGE_NOTIFICATIONS.LEAVE_MEMBER]: L('Left {title}', { title: channelName }),
+          [MESSAGE_NOTIFICATIONS.KICK_MEMBER]: L('Kicked a member from {title}: {displayName}', { displayName, title: channelName }),
+          [MESSAGE_NOTIFICATIONS.UPDATE_NAME]: L('Updated the channel name to: {title}', { title: channelName }),
+          [MESSAGE_NOTIFICATIONS.UPDATE_DESCRIPTION]: L('Updated the channel description to: {description}', { description: channelDescription }),
+          [MESSAGE_NOTIFICATIONS.DELETE_CHANNEL]: L('Deleted the channel: {title}', { title: channelName }),
+          [MESSAGE_NOTIFICATIONS.VOTE]: L('Voted on “{}”')
```
taoeffect: This is missing a variable

frontend/model/contracts/mailbox.js
```diff
@@ -41,9 +50,9 @@ sbp('chelonia/defineContract', {
         const initialState = merge({
           attributes: {
             creator: data.username,
-            autoJoinAllowance: true
+            autoJoinAllowance: true // this attribute could be used to block him to be joined direct/group messages automatically by another
```
taoeffect: Could you please clarify what this attribute is for? The comment still isn't very clear to me

frontend/model/state.js
Silver-IT: Yeah. Correct. Will fix it.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: That is done in the PR #1496, because it's for the the Issue #1494.

frontend/views/containers/chatroom/ChatMembersAllModal.vue
Silver-IT: Okay.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -285,6 +285,9 @@ export default ({
             const { meta, data } = msgValue
             this.messages.push({
               ...createMessage({ meta, data, hash: message.hash() }),
+              // TODO: message.hash could be different from the message on event listener
+              // https://github.com/okTurtles/group-income/issues/1487
+              // so that the message.hash() can't be used as the identifier of this message
```
taoeffect: The issue mentioned in this comment is closed... so should a new issue be created if this isn't going to be fixed in this PR?

frontend/views/containers/chatroom/ChatMembers.vue
```diff
@@ -12,7 +12,7 @@
 
   ul.c-group-list
     list-item(
-      v-for='{username, displayName, picture } in directMessages'
+      v-for='{username, displayName, picture} in priateDirectMessages'
```
taoeffect: `priateDirectMessages` is misspelled, should be `privateDirectMessages` I think?

test/cypress/support/commands.js
```diff
@@ -90,6 +91,10 @@ Cypress.Commands.add('giLogin', (username, {
       }
       await sbp('gi.actions/identity/login', { username, password })
       await sbp('controller/router').push({ path: '/' }).catch(e => {})
+
+      return new Promise((resolve) => {
+        sbp('okTurtles.events/on', LOGIN, resolve)
+      })
```
taoeffect: I don't think this should be here. And if it were needed for some reason, the promise should be setup before `'gi.actions/identity/login'` is called, not after, as that might miss the event being fired.

But I think this just shouldn't be here.

taoeffect (changes requested):
Nice work @Silver-IT!

I tried testing this and found a few bugs:

I created a group with 3 users in it, as per the video below. The first time I tried this, the 2nd tab said "Loading" in the tab title. I couldn't reproduce this though when I made the recording.

Other bugs found:

- Multi-user DMs does not result in sound in other tab when a DM is sent. Sounds should happen on DMs just like when @mentions are done
- The message counter doesn't increment in the UI
- The red dot icon doesn't appear in the tab, nor the navigation, etc. when multi-user DMs are received

Video: https://okturtles.slack.com/archives/C0EH7P20Y/p1677642785935519

test/cypress/support/commands.js
Silver-IT: Make sense. I thought I removed this one. Nice catch, @taoeffect

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: After we do this issue(#1503), I think we should update this. I will update comment.

Silver-IT:
> The first time I tried this, the 2nd tab said "Loading" in the tab title.

I will try to find this issue and fix it out.

> Multi-user DMs does not result in sound in other tab when a DM is sent. Sounds should happen on DMs just like when @mentions are done

Even in Multi-user DMs? Actually in a DM between two people, all the messages make sounds by default. But I think in multi-user DM, we should handle this just like a group channel?

> The message counter doesn't increment in the UI

Yeah, that's a nice catch. It's a really bug. I will fix it out.

> The red dot icon doesn't appear in the tab, nor the navigation, etc. when multi-user DMs are received

Only mentions are counted in multi-user DM by default. Is this wrong?

taoeffect:
> Even in Multi-user DMs? Actually in a DM between two people, all the messages make sounds by default. But I think in multi-user DM, we should handle this just like a group channel?

Hmm. I think Slack handles multi-user DMs as regular DMs... Yeah, let's do it that way, and then we can update it later if necessary. For example, if it becomes a problem, we can allow users to adjust notifications for group DMs on an individual basis to only make sounds when an @mention occurs.

> Only mentions are counted in multi-user DM by default. Is this wrong?

Again - I think multi-user DMs should be handled exactly the same as 1-on-1 DMs

test/cypress/support/commands.js
taoeffect: Yeah it might have been from the merge of the 2 PRs

taoeffect (approved):
Excellent work @Silver-IT - this is a massive PR and a lot of issues fixed!

I don't think I've reviewed 100% of it, but I've reviewed and tested enough. Now let's make sure everyone tests it on our production server and see if we can find any additional bugs.

PR #1468, opened 2022-12-01
fix the broken translation in GroupMembersActivity.vue

### LINKED ISSUE #1449: Implement Activity Dashboard Tab

### LINKED ISSUE #1467: Sentences get broken into two pieces in translation in GroupMembersActivity.vue
### Problem
This is a follow up work of PR #1453. (the context [here](https://github.com/okTurtles/group-income/pull/1453#discussion_r1032733561))
Some sentences in the 'GroupMembersActivity.vue' components are broken into and two pieces of elements, each of which gets translated separately when they actually compose one sentence together.
e.g) 5 members haven't logged in past 2 weeks => "5 members" + "haven't logged in past 2 weeks" (2 pieces)

This can lead to an wrong translation in some languages. For example,
The position of "5 members" can be somewhere else that is not at the beginning of the sentence, as it is in English.
sentences in the app should always be translated as complete sentences.

### Solution
Figure out how to wrap the "N members" in the sentence with a tooltip component, while not breaking the translation.

SebinSong:
closes #1467 
closes #1449


SebinSong:
@taoeffect The solution made in this PR works good but if we think this can overcomplicate the `Tooltip.vue` component as a result, I think this UI (which I just found in the figma, which seemingly was suggested as one of the possible options there) can be a good option to avoid both 1) over-complicating `Tooltip.vue` 2) broken translation issues.



just letting you know, but if the existing tooltip is just good as is, don't worry about this comment. :)

cheers,

frontend/views/components/Tooltip.vue
```diff
@@ -62,17 +57,37 @@ export default ({
...
       default: false
+    },
+    triggerElementSelector: {
+      // Instead of taking the entire 'default-slot' as the trigger element(which is the default behaviour of this component),
+      // specifying this prop will bind the tooltip to 'a particular element within the default-slot' content.
+      // The value must be a valid css-selector string, which will be used in searching via HTMLElement.querySelector()
+      // (Refer to GroupMembersActivity.vue for the use case.)
+      type: String,
+      required: false,
+      default: ''
```
Silver-IT, line 68: Nice idea using `triggerElementSelector`.

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -92,12 +85,12 @@ export default ({
       'groupProfiles'
     ]),
     onTimePayments () {
-      return Object.entries(this.groupStreaks.onTimePayments)
+      return Object.entries(this.groupStreaks.onTimePayments || {})
```
Silver-IT, line 89: To keep consistency, I think we should return empty Object in getters.
Just like how we did for `groupStreaks`, `groupThankYousFrom`, `groupPeriodPayments`, ...

Silver-IT (changes requested):
Nice work, @SebinSong. Left only one feedback.

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: @Silver-IT `onTimePayments` here is one field that belongs to `groupStreaks` getter and I decided not to expose every streak field as `group.js` contract getters as it felt like an overkill when group streak values are currently only used in one place in the app. hope it makes sense.

Silver-IT: Yeah. It makes sense. As it's not a getter, it doesn't break the consistency.

Silver-IT (dismissed):
Approving this PR. Great work.

frontend/views/pages/DesignSystem.vue
```diff
@@ -613,29 +613,56 @@ page(
...
+        tr
+          td
+            pre
+              | tooltip(triggerElementCss='.trigger-target')
+              |   i.icon-* Basic
+          td
+            tooltip(text='A simple text inside' triggerElementSelector='.trigger-target' direction='bottom-end')
+              i18n(
+                :args='{ ...LTags("strong"), membercount: `<span class="link trigger-target">${5 + " members"}</span>` }'
+              ) {membercount} &nbsp;have {strong_} on-time payment streaks{_strong}
```
taoeffect: Nothing in the DesignSystem should contain translations or use of i18n/LTags because it's not user-facing. (Only user-facing strings use these)

frontend/views/pages/DesignSystem.vue
```diff
@@ -1364,6 +1391,7 @@ import SvgProposal from '@svgs/proposal.svg'
 import SvgVote from '@svgs/vote.svg'
 import { mapGetters, mapMutations } from 'vuex'
 import { THEME_LIGHT, THEME_DARK } from '~/frontend/model/settings/themes.js'
+import { LTags } from '@common/common.js'
```
taoeffect: So this can be removed (and same below)

frontend/views/containers/dashboard/SentenceWithMemberTooltip.vue
```diff
@@ -0,0 +1,31 @@
+<template lang="pug">
+component(
+  is='Tooltip'
+  :key='members.length'
+  triggerElementSelector='.t-trigger'
+  direction='bottom-end'
+)
```
taoeffect: Is there a reason you are using `<component>` here instead of `<tooltip>` directly? If so, could you add that explanation in a comment?

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -129,6 +122,38 @@ export default ({
...
+          : this.L('{span_}{membercount} members{_span} haven\'t {strong_} logged in past {days} days or more {_strong}', argsMap['haventLoggedIn']),
+        'noIncomeDetails': this.noIncomeDetails.length === 1
+          ? this.L('{span_}{membercount} member{_span} haven\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails'])
+          : this.L('{span_}{membercount} members{_span} haven\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails']),
+        'missedPayments': this.missedPayments.length === 1
+          ? this.L('{span_}{membercount} member{_span} have {strong_} missed payments {_strong}', argsMap['missedPayments'])
+          : this.L('{span_}{membercount} members{_span} have {strong_} missed payments {_strong}', argsMap['missedPayments']),
+        'noVotes': this.noVotes.length === 1
+          ? this.L('{span_}{membercount} member{_span} haven\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
+          : this.L('{span_}{membercount} members{_span} haven\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
```
taoeffect: Can you re-run the latest version of the `strings` utility on the `frontend/` folder to regenerate all of these strings?

taoeffect (changes requested):
Great work @SebinSong! Review ready for you. 😄 

One question: do you think you can update this PR to show all of activity labels, including all streaks and inactivity, even if the values are 0? This way users will know what the possibilities are.

If it is too difficult to do that in this PR, let me know, and we can create a separate issue for it.

frontend/views/containers/dashboard/SentenceWithMemberTooltip.vue
SebinSong: @taoeffect I did leave the explanation comment few lines below here but rephrased it in [this commit](https://github.com/okTurtles/group-income/pull/1468/commits/f1bcc70b31706c703d474a8348ed2d72ac6c820d) to make it more clear.

SebinSong:
@taoeffect updated this PR with the work on all your change requests.
Also regarding below,

> One question: do you think you can update this PR to show all of activity labels, including all streaks and inactivity, even if the values are 0?

Yes I can. please let me know either here or in the original issue page if the copies for the cases when the strikes are 0 are confirmed.

taoeffect:
> Yes I can. please let me know either here or in the original issue page if the copies for the cases when the strikes are 0 are confirmed.

Sorry, I'm not sure I understand what you mean by "copies for the cases when the strikes are 0 are confirmed"?

I'm requesting that things like `v-if='missedPayments.length > 0'` be removed, so that the user becomes immediately aware of what strikes and streaks are possible in the app.

frontend/views/containers/dashboard/SentenceWithMemberTooltip.vue
taoeffect: Interesting comment, thank you for adding that! Seems like an advanced usage of Vue!

frontend/views/containers/dashboard/GroupMembersActivity.vue
taoeffect: Interesting! I was wondering why these strings weren't appearing in the generated strings files.

Apparently using `this.L` will prevent the strings from getting extracted. But `L` by itself works.

So can you do `import { L } from '@common/common.js'` instead?

Also, I noticed that this isn't the only file where this is an issue. Can you do a project-wide search for all instances of `this.L()` usage and replace with direct `L()` function usage?

taoeffect (changes requested):
Very nice @SebinSong! I did some tests and found that `strings` wasn't extracting `this.L` (needs to be `L` by itself). Posted some comments.

frontend/views/containers/dashboard/GroupMembersActivity.vue
SebinSong: @taoeffect yes sure. will do that in the following commit.

taoeffect (approved):
Excellent.

🥇

Issue #952, opened 2020-07-10
What happens to income details when the group mincome changes?


mmbotelho:
### Problem

When the group's mincome changes, the information entered by members on their income details modal can become outdated. 

### Solution

Mincome changes, the user was **receiving**. We should automatically calculate:
- If they now need more money, and update the amount they need;
- If they now need less money, and update the amount they need;
- If they no longer need to receive a mincome (e.g.: the mincome is now lower than their stated income), we should change their income details to say "Do you make at least x a month? Yes I do" and leave the pledge at $0.

Mincome changes, user was **pledging**:

- Their pledge remains the same unless they specifically change it.

*EDIT by @taoeffect:* We automatically/programmatically adjust the user's need as above, and where it makes sense, we'll additionally ask the user if they make the new mincome $X amount, and then proceed through the normal "Income Details" flow to adjust them based on the user's answer. See for details on this: #1125

dotmacro:
Do mincome changes take place instantly, or for the next month? If it's instantly, then for "user was receiving", also add:  
- If they now need no money _for the current month_ because they have already received enough to put them at/over the new mincome threshold, update the amount they need _for the current month_ to be $0.

taoeffect:
Replying to @dotmacro:

> Do mincome changes take place instantly, or for the next month?

I think it should be instant if at all possible. This might require some changes to how the distribution algorithm works.. /cc @CrazedCoding 

>  If it's instantly, then for "user was receiving", also add:
> 
> - If they now need no money for the current month because they have already received enough to put them at/over the new mincome threshold, update the amount they need for the current month to be $0.

Unless I'm misunderstanding, I think the distribution algorithm should do this automatically...

EDIT: oh, your comment was opened a while ago, and yeah since then I think this has been addressed by @CrazedCoding 🙂

Replying to @mmbotelho:

> If they no longer need to receive a mincome (e.g.: the mincome is now lower than their stated income), we should change their income details to say "Do you make at least x a month? Yes I do" and leave the pledge at $0.

And:

>  Their pledge remains the same unless they specifically change it.

I think these are both instances that could use with prompting the user whether or not they want to edit their income details.

For example, in that last case, where they were pledging, I think it really does make sense to confirm with the pledger whether they still want to pledge because the very basis of their original pledge was based on a question we asked them: whether or not they made more than the minimum income. Since the mincome just increased, their answer to that question might've changed, so we should ask them again.

Does that seem right to you as well?

[...]

PR #1470, opened 2022-12-03
#1125 Ask users for income details update if mincome has been increased.

### LINKED ISSUE #1125: Ask users whether they want to change their income details when the mincome changes
### Problem

Related to #952 - this issue addresses the problem of what needs to be done when the mincome changes and the user is pledging.

At the moment we don't prompt the user to do anything, and this really isn't the right answer because the user set their mincome in the first place based on a question, which is whether or not they made less than the mincome (whatever it was originally).

So when a proposal passes to increase the mincome, their answer to that question might change too.

### Solution

Prompt the user a Yes/No question when a mincome proposal passes and ask them:

> **The mincome has changed to $X.
> 
> **Do you make at least $X per month?**
>
> Buttons: No / Yes

If they click "No", open the modal for the income details.

Two issues need to be closed to do this:

- #1124 - Emitting a notification in the `sideEffect` of `proposalVote` action
- #1126  The design and implementation of a non-dismissable prompt to ask the user a question

SebinSong:
closes #1125


frontend/controller/actions/group.js
```diff
@@ -508,6 +509,19 @@ export default (sbp('sbp/selectors/register', {
...
+    const isNoSelected = await sbp('gi.ui/prompt', {
+      question: L('Do you make at least {amount} per month?', { amount: params.amount }),
+      heading: L('Mincome increased'), // TODO: discuss if this heading is fine. currently just a placeholder.
+      yesButton: L('No'),
+      noButton: L('Yes')
+    })
+
+    if (isNoSelected) {
+      // if a user doesn't make more than the changed mincome, let them decide whether they want to update their income details.
+      sbp('okTurtles.events/emit', REPLACE_MODAL, 'IncomeDetails')
```
taoeffect, line 531: This should be `REPLACE_MODAL` and not `OPEN_MODAL` (you're sure?)

frontend/model/contracts/group.js
```diff
@@ -1012,14 +1012,34 @@ sbp('chelonia/defineContract', {
         groupName: x => typeof x === 'string',
         groupPicture: x => typeof x === 'string',
         sharedValues: x => typeof x === 'string',
-        mincomeAmount: x => typeof x === 'number' && x > 0,
+        mincomeAmount: objectOf({
+          from: x => typeof x === 'number' && x > 0,
+          to: x => typeof x === 'number' && x > 0
+        }),
```
taoeffect: A few concerns here:

1. This is less elegant than the validation we had before, and there's a way to do this without having to modify the validation
2. By allowing `from` to be specified, we allow nonsensical actions to be created with invalid `from` values, and we also force the calling code to remember what the previous value was when creating this action

Instead, you can do most of what you need to do inside of the `process` by:

1. Caching the old value in the `process` function
2. Creating a new method under `methods:` key (see bottom of file)
3. Calling that method using `'gi.contracts/group/pushSideEffect'` and passing in the cached old mincome value

frontend/model/contracts/shared/voting/proposals.js
```diff
@@ -154,10 +154,16 @@ const proposals: Object = {
       //       this code will need to be updated
       const message = {
         meta,
-        data: { [setting]: proposedValue },
+        data: {
+          [setting]: setting === 'mincomeAmount'
+            ? { from: sbp('state/vuex/getters').groupMincomeAmount, to: proposedValue }
+            : proposedValue
+        },
```
taoeffect: This can then be reverted

frontend/views/containers/notifications/NotificationList.vue
```diff
@@ -82,7 +82,12 @@ export default ({
     handleItemClick (item) {
       if (!this.ephemeral.isSelectingText) {
         this.markAsRead(item)
-        this.$router.push(item.linkTo).catch(console.warn)
+
+        if (item.sbpInvocation) {
+          sbp(...item.sbpInvocation)
```
taoeffect, line 87: This is a cool idea 👍

taoeffect (changes requested):
Excellent work @SebinSong 💪 

Review ready!

frontend/controller/actions/group.js
SebinSong: @taoeffect Yes, at the point in the flow, the `Prompt.vue` is active by the `sbp('gi.ui/prompt')` call few lines above.

taoeffect: OK, possibly worth adding a comment to explain that, as to me it's not intuitive why that is, and the comment will prevent someone from accidentally modifying this to `OPEN_MODAL`

frontend/model/contracts/group.js
SebinSong: great to know `methods` can be useful for similar cases (when certain data needs to be passed from `process` to `sideEffect`) in the future. 👍

SebinSong:
@taoeffect I got one question here.
I just found this sentence from the issue description.
> Related to https://github.com/okTurtles/group-income/issues/952 - this issue addresses the problem of what needs to be done when the mincome changes and **the user is pledging.**

The way I've implemented it now is the notification and the prompt modal is sent & displayed to any group members who has entered their income details. should we do these only for pledging members? Thanks,

taoeffect:
@SebinSong 

> The way I've implemented it now is the notification and the prompt modal is sent & displayed to any group members who has entered their income details. should we do these only for pledging members? Thanks,

Let's consider a scenario where the mincome was $1000/month, and a user is a receiver, with an income of $500/month:

- If the mincome increases to $1400/month, no action is needed for this user. The algorithm correctly recalculates everything
- If the mincome decreases to $600/month, no action is needed.
- If the mincome decreases to $400/month, we should automatically switch them to being a pledger with a $0 pledge, and we should prompt them to update their income details per this PR

Issue #952 is about this automatic sort of adjustment, that we currently don't do.

SebinSong:
@taoeffect Thanks for clarification.
Could you confirm if my summary below is correct again? :)

[ Pledging members ]
- When mincome has increased: prompt them to update their income details
- When mincome has decreased: No action needed

[ Receiving members ]
- When mincome has increased: No action needed
- When mincome has decreased: If the decreased value is still higher than their monthly income, no action is needed. But if the updated mincome is lower than their monthly income, automatically switch them to a pledging member and prompt them to update their income details.

Btw,
> if the updated mincome is lower than their monthly income, automatically switch them to a pledging member

is apparently not implemented yet either. I can work on it as part of this PR if there isn't any reason that it has to be dealt with separately.

Thanks,

taoeffect:
Yep! That all sounds good 👍 

> automatically switch them to a pledging member

Yes, and specifically a pledging member with a 0 pledge.

frontend/model/contracts/group.js
```diff
@@ -1275,6 +1307,24 @@ sbp('chelonia/defineContract', {
         await sbp('gi.db/archive/save', archPaymentsKey, archPayments)
       }
       sbp('okTurtles.events/emit', PAYMENTS_ARCHIVED, { paymentsByPeriod, payments })
+    },
+    'gi.contracts/group/sendMincomeChangedNotification': async function (contractID, data) {
+      if (data.memberType === 'receiving') {
+        await sbp('gi.actions/group/groupProfileUpdate', {
```
SebinSong: @taoeffect 
This part of the code is the source of below issue which I had trouble finding the cause for.



Steps to reproduce / encounter this issue is:
1) Sign up `user-1` and then create a `group-1`(whose mincome is set to, let's say, 150$). then set up his income details as a pledger.
2) Sign up `user-2` and then join  him `group-2`. then set up his income details as a receiver with the income as let's say, 120$.
3) switch to `user-1` and reset the mincome to an amount lower than `user-2`'s income (e.g. 110$)
4) In `user-2`'s screen, he is expected to have received a notification saying the mincome changed, but instead he gets this error.

frontend/model/contracts/group.js
```diff
@@ -1015,10 +1015,42 @@ sbp('chelonia/defineContract', {
...
+        const mincomeCache = 'mincomeAmount' in data ? state.settings.mincomeAmount : null
+
         for (const key in data) {
           Vue.set(state.settings, key, data[key])
         }
+
+        if (mincomeCache !== null) {
+          // if mincome has changed and the user has set up their income details,
+          // emit 'MINCOME_CHANGED' notification, which prompts the user to choose whether they want to update income details or not.
+          const { loggedIn } = sbp('state/vuex/state')
```
taoeffect: Can't access state outside of contract state inside of a `process` function. `sbp('state/vuex/state')` can only be called inside of a sideEffect.

taoeffect:
One request from @dotmacro (that I think is a good one): [in the scenarios you described here](https://github.com/okTurtles/group-income/pull/1470#issuecomment-1372831162), where it says "no action needed", can you still prompt the **receiving** members (with the original prompt) to verify their income hasn't changed?

It's good to periodically prompt them tbh.

SebinSong:
@taoeffect @dotmacro Updated the PR with the work for the latest minor requirement.

frontend/controller/actions/group.js
```diff
@@ -508,6 +509,28 @@ export default (sbp('sbp/selectors/register', {
...
+    const { withGroupCurrency } = sbp('state/vuex/getters')
+    const promptOptions = params.increased
+      ? {
+          heading: L('Mincome changed'),
+          question: L('Do you make at least {amount} per month?', { amount: withGroupCurrency(params.amount) }),
+          yesButton: params.memberType === 'pledging' ? L('No') : L('Yes'),
+          noButton: params.memberType === 'pledging' ? L('Yes') : L('No')
+        }
+      : {
+          heading: L('Automatically switched to pledging {zero}', { zero: withGroupCurrency(0) }),
```
taoeffect, line 522: Having tested this out, I think from a UX perspective, it might make more sense to automatically show this prompt the user, instead of waiting for them to click on the notification, which they might miss, and then be confused about what's happening and why they're suddenly no longer a receiver.

What do you think?

taoeffect (changes requested):
Excellent work @SebinSong! All of the code seems solid now.

I have now gone into the testing phase of the review, and have one concern / possible behavior change request. Have a look:

frontend/controller/actions/group.js
SebinSong: @taoeffect I agree it makes more sense that way. Do we still keep the notification along with this change?

taoeffect: @SebinSong sure, let's keep the notification

SebinSong:
@taoeffect updated this PR with the change for your request below

>  it might make more sense to automatically show this prompt the user, instead of waiting for them to click on the notification

taoeffect (changes requested):
Forgot to regenerate the saved contracts using `grunt pin 0.0.12 --overwrite`

taoeffect (approved):
Great work @SebinSong! Approved!!

PR #1472, opened 2022-12-09
#1458 Fix wrong notification ages and broken links

### LINKED ISSUE #1458: Time / Revew & Thank-You-Note in Notification UI Wrong / Broken
### Problem

1) The notifications don't show the amount of time they have existed, but rather: the amount of time since they have been received.

2) The "Review and send a thank you note" button in the notifications doesn't work.

Both of these issues can be seen in the following screenshot.

Screenshot 2022-11-09 at 9 53 08 PM

The time on the notifications should be many weeks by this point, not just the 2 minutes since opening the page.

Additionally, as mentioned, the "Review and send a thank you note" button in the notifications doesn't work.

### Solution

A) Show the time since the event occurred on earth in the notifications, not just the time since it was received by your browser.

B) (1) Find out why the button isn't responding, (2) fix accordingly.

### LINKED ISSUE #1487: Listening for message hash in prepublish hooks isn't safe!
### Problem

As shown in this code for `cy.giForceDistributionDateToNow`:

```js
Cypress.Commands.add('giForceDistributionDateToNow', () => {
  cy.window().its('sbp').then(sbp => {
      sbp('gi.actions/group/forceDistributionDate', {
        hooks: {
          prepublish: (message) => {
            sbp('okTurtles.events/on', message.hash(), resolve)
          }
        }
      })
    })
  })
})
```

We are listening for `message.hash()` event - BUT, as @snowteamer very astutely observed, this isn't safe! Messages can be re-created by `'chelonia/private/out/publishEvent'`, causing the hash of the successful message to be different, resulting in a hanging promise.

### Solution

Find all such uses of `prepublish` or `prepublishContract` in the code, and replace them with listeners for `EVENT_HANDLED` (as emitted by `internals.js`), and in the listener compare on the ~~`message.message().previousHEAD` instead of the message's hash.~~ Add a comment explaining all of this so that people who read the code aren't confused.

EDIT: NO! The previousHEAD will be different too (because a message got squeezed in between)! So, perhaps the best way to deal with this is to compare on the `JSON.stringify(message.opValue())`, as that should be the same

snowteamer:
Closes #1458
Closes #1487

### Summary of changes:
- The notification card popup should now correctly display when the corresponding event happened
- Clicking `PAYMENT_RECEIVED` or `PAYMENT_THANKYOU_SENT` notifications should now correctly redirect to the relevant part of the app.
- Fixes heisenbug in tests


frontend/views/containers/notifications/NotificationList.vue
```diff
@@ -76,7 +76,7 @@ export default ({
     }
   },
   methods: {
-    ageTag (item: Object): number {
+    ageTag (item: Object): string {
```
taoeffect, line 79: Nice catch

frontend/model/notifications/vuexModule.js
```diff
@@ -76,11 +76,9 @@ const mutations = {
...
       throw new Error('This notification is already in the store.')
     }
-    if (state[0] && age(notification) > age(state[0])) {
-      throw new Error('This notification is older than the latest one in the store.')
-    }
-    // Using `unshift()` here to insert the new item before older ones.
-    state.unshift(notification)
+    state.push(notification)
+    // Sort items in chronological order, newest items first.
+    state.sort(compareOnAge)
```
taoeffect: I'm looking at the implementation of `compareOnAge`, and its usage of `age` doesn't make much sense to me:

```js
// Compares notifications, so that newer ones come first.
export function compareOnAge (notificationA: Notification, notificationB: Notification): -1 | 1 {
  return age(notificationA) <= age(notificationB) ? -1 : 1
}
```

This really should be implemented like so:

```js
export function compareOnAge (notificationA: Notification, notificationB: Notification): number {
  return notificationA.timestamp - notificationB.timestamp

  // OR:
  // return age(notificationA) - age(notificationB)
}
```

Or:

And similarly, `age` doesn't make sense. Instead of:

```js
export function age (notification: Notification): number {
  return Date.now() - notification.timestamp
}
```

It should simply return `notification.timestamp`.

taoeffect (changes requested):
Excellent work @snowteamer! One minor issue found!

frontend/model/notifications/vuexModule.js
snowteamer: @taoeffect Ok, may I rename `compareOnAge` to `compareOnTimestamp` then?

taoeffect: Yep!

snowteamer: @taoeffect However removing the age function is going to make the `isExpired()` function and the conditions in `compareOnPriority()` a bit less readable

frontend/model/notifications/utils.js
```diff
@@ -96,8 +92,8 @@ export function compareOnPriority (notificationA: Notification, notificationB: N
     // However, under certain conditions we might want to keep the already read one instead.
     // Note: use a condition array so that it is easy to add/remove conditions later.
     const conditions = [
-      (read, unread) => age(read) < ONE_DAY && age(unread) > FIFTEEN_DAYS,
-      (read, unread) => age(read) < TWO_DAYS && age(unread) > THREE_WEEKS
+      (read, unread) => Date.now() - read.timestamp < ONE_DAY && Date.now() - unread.timestamp > FIFTEEN_DAYS,
+      (read, unread) => Date.now() - read.timestamp < TWO_DAYS && Date.now() - unread.timestamp > THREE_WEEKS
```
taoeffect: > @taoeffect However removing the age function is going to make the isExpired() function and the conditions in compareOnPriority() a bit less readable

Hmm, I agree. If you'd like to re-add it in order to use it here and below (in `isExpired`), feel free.

taoeffect (changes requested):
Nice work @snowteamer!

The 3rd test in `group-paying.spec.js` failed locally for me 3 times in a row, and as it's a test related to notifications, it suggests either a bug in this PR or a test that needs to be updated:

Screen Shot 2023-01-22 at 5 11 31 PM

Also replied regarding `age()`

frontend/model/notifications/vuexModule.js
```diff
@@ -76,11 +76,9 @@ const mutations = {
     if (state.includes(notification)) {
       throw new Error('This notification is already in the store.')
     }
-    if (state[0] && age(notification) > age(state[0])) {
-      throw new Error('This notification is older than the latest one in the store.')
-    }
-    // Using `unshift()` here to insert the new item before older ones.
-    state.unshift(notification)
+    state.push(notification)
```
taoeffect, line 80: One request - during my testing of a different PR where I changed the computer's clock, I got the `throw new Error('This notification is older than the latest one in the store.')` thrown, and this broke the app.

It turns out that we cannot throw errors here because these are being thrown inside of sideEffects, and if an error is thrown inside of a sideEffect it will stop the app completely from moving forward and doing anything. So errors thrown in sideEffects must be because something super bad happened.

So it's good that this PR removes that error, but there's still one on line 77. Can you change that to `console.error` + `return` instead?

snowteamer: Ok, will do

taoeffect (approved):
Excellent work @snowteamer !!!

Issue #400, opened 2018-05-13
Polls


taoeffect:
### Problem

It can often be useful to interact with your group members by asking them to vote in a poll, similar to how Twitter handles polls. We do have a proposals system where votes can be cast, but these have real consequences. It would be nice to have a simple polls feature in the chat with no consequences (simply to gage sentiment on whatever question members have).

### Solution

Implement this feature.

dotmacro:
[Figma link for polls design & flow](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=14088%3A154296)

SebinSong:
Hi @dotmacro , 
I was wondering if this task is ready to be implemented in the front-end and if so, could you please share the link to the design of this again please? 
the one you commented above seems to have been revoked. Thanks.

dotmacro:
The polls design and flow is here: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23766%3A273992&t=kIFu9duj35DntNRc-3

It looks ready to me! There is one question about checkbox & text positioning, which you may be able to figure out based on the [design notes](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23766%3A265613&t=kIFu9duj35DntNRc-3).

SebinSong:
# Dev TODO List

## Part 1. Build required UI components


 - [x] **Task 1-1. 'Create new polls' popover component**

<img src='https://user-images.githubusercontent.com/17641213/208851018-b119026a-cf18-4b2f-8c4a-ba91f76df1c5.png' width='340'>

- [x]  **Task 1-2. Message component with type='poll'** _- when the user hasn't voted yet_

<img src='https://user-images.githubusercontent.com/17641213/208851617-8c010879-9eab-496f-ad64-c2545f6b0f79.png' width='410'>

- [x]  **Task 1-3. Message component with type='poll'** _- after the user has voted (message variation, notification for vote)_

<img src='https://user-images.githubusercontent.com/17641213/208852174-85c85a2b-9c64-4d96-89bb-c32107807d6e.png' width='410'>

<img src='https://github.com/okTurtles/group-income/assets/17641213/c3d06918-08fc-4ccb-8c7f-137dfd5e4177' width='410'>

- [x]  **Task 1-4. Implement 'change vote' functionality**

[...]

PR #1474, opened 2022-12-18
Some code improvements in the Chatroom area

SebinSong:
- Remove the unused vue-components in `ChatNav.vue`.
- Fix an issue in `ChatMain.vue` where a `change` event listener for `window.matchMedia()` in the `created()` hook there is not destroyed event after the user navigates to another page. The number of this un-destroyed event listeners increases as the user repeats getting in & out of the `/group-chat` page, leading to a memory leak problem.


SebinSong:
Hi @taoeffect @Silver-IT ,
I've been actually going through the chatroom area in the code-base in the past few days to familiarise myself with it in preparation of my next task (#400)
and found some legacy codes in there that are no longer in use and also a bug related to an un-destroyed event listener in `ChatMain.vue`. Thought I would make this simple & small PR to resolve them.
Please let me know if you spot any issue in here.

Thanks,

frontend/views/containers/chatroom/ChatNav.vue
```diff
@@ -1,58 +1,18 @@
...
-          menu-item(tag='button' itemid='hash-1' icon='heart')
-            i18n Settings
-          menu-item(tag='button' itemid='hash-2' icon='heart' hasdivider='')
-            i18n Mute chat
-          menu-item(tag='button' itemid='hash-3' icon='heart')
-            i18n Report a problem
-
-  .c-chatnav-body
-    slot
-
```
taoeffect, line 19: @Silver-IT is Sebin right that this is no longer needed?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -169,10 +170,12 @@ export default ({
...
-    // TODO #492 create a global Vue Responsive just for media queries.
-    const mediaIsPhone = window.matchMedia('screen and (max-width: 639px)')
-    this.config.isPhone = mediaIsPhone.matches
-    mediaIsPhone.onchange = (e) => { this.config.isPhone = e.matches }
+    // TODO: #492 create a global Vue Responsive just for media queries.
+    this.config.mediaIsPhone = window.matchMedia('screen and (max-width: 639px)')
+    this.config.mediaIsPhone.onchange = (e) => {
+      this.config.isPhone = e.matches
+    }
+    this.config.isPhone = this.config.mediaIsPhone.matches
```
taoeffect: Why is `mediaIsPhone` being added to the config? Now there's two config variables to keep track of when we just want to know if we're a mobile device, right? So wouldn't it be better to just have one variable?

taoeffect (commented):
Added comments with questions for you and @Silver-IT !

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @taoeffect The purpose of this change is to destroy the `onchange` matchMedia handler along with destruction of `ChatMain.vue` (L188 below). So I wanted to expose this matchMedia instance as a component level variable instead of a local variable in `created()` hook. But now that you've mentioned that, it looks like it's not necessary to define it in `config` object. I could just attach it as `this.matchMediaPhone`. Does it sound good?

taoeffect: Ah I see, yes that sounds good. Nice catch! And I like `matchMediaPhone` as a variable name better too 👍

frontend/views/containers/chatroom/ChatNav.vue
SebinSong: @Silver-IT @taoeffect 
(As you might just know by looking at the code)`template(slot='actions')` here is silently ignored and removed from the DOM at runtime (Apparently there used to be a child component in here to which this `slot` was passed). so this code practically becomes,
`.c-chatnav`
`  .c-chatnav-body`
`     slot`
here. Then I decided to just turn it into
`.c-chatnav`
`  slot`
As it doesn't need to wrapped twice. Thanks.

Silver-IT: Currently I'm not sure. I will review it when I get back.

Silver-IT:
>

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Should we declare the `matchMediaPhone` variable inside the `data`?

frontend/views/containers/chatroom/ChatNav.vue
Silver-IT: @SebinSong, I think that's not needed. Maybe that part was copied from another component when it was created.
Actually, I think there could be some more unused lines of codes in chat part such as Messages.vue page and it's route. I could guess the previous developer was going to use Messages.vue for the direct messages.

Silver-IT (approved):
Nice catch, @SebinSong. Please check my comment above even though it's a tiny request.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @Silver-IT I think it's not necessary in this case because, (unlike `this.config.isPhone`) it's not the kind of state that is linked to the DOM update, let alone the fact that it's only used in two life cycle hooks in the component. Thanks.

taoeffect (approved):
Approved!

PR #1479, opened 2023-01-11
Use isMobile() instead of isMacSafari() in LinkToCopy.vue

### LINKED ISSUE #1478: Bug: LinkToCopy.vue needs to check for mobile, not isMacSafari
### Problem

On a call with @corrideat we noticed that we couldn't copy links that used `LinkToCopy.vue` component (e.g. invite links), because Microsoft Edge blocked some type of popup that was triggered by the usage of `navigator.share`.

`navigator.share` shouldn't be used on any desktop browser anyway (I think).

The problematic code is the way `copyToClipboard` function in `LinkToCopy.vue` checks whether it should use `navigator.share` or copy to clipboard (and display a tooltip that the link was copied).

### Solution

Instead of `isMacSafari`, we need a `isMobile` function that checks if we're on mobile.  Find the best solution for this (e.g. search Stackoverflow).

Then if we're on mobile, [and `navigator.share` is defined](https://github.com/okTurtles/group-income/issues/1021#issuecomment-1106801450), then use `navigator.share`. Otherwise use the other existing code path that's there to copy the link.

Desktop browsers should always use the copy-to-clipboard code path.

snowteamer:
Closes #1478

### DIFF
--- a/frontend/views/components/LinkToCopy.vue
+++ b/frontend/views/components/LinkToCopy.vue
@@ -47,20 +47,13 @@ export default ({
     }
   },
   methods: {
-    // FIXME: probably needs to be defined centrally somewhere else
-    isMacSafari () {
-      // NB: Chrome, Edge, Brave on Macintosh tacks on the 'Safari/537.36' string in its userAgent
-      // so we have to ensure that were not mistaking those browsers instead when matching for Safari on Macintosh
-      const userAgent = navigator.userAgent
-      const safari = /Macintosh.*Safari/
-      const chrome = /Macintosh.*Chrome/
-      const edge = /Macintosh.*Edg/
-      return userAgent.match(safari) &&
-              !(userAgent.match(chrome) || userAgent.match(edge))
+    isMobile () {
+      // https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser#comment95674193_51774045
+      return !window.matchMedia('(any-pointer:fine)').matches
     },
     copyToClipboard () {
-      // if the user is using the device that supports web share API, use it and then skip the other logics below.
-      if (navigator.share && !this.isMacSafari()) {
+      // if the device supports the Web Share API, use it and then skip other logic below.
+      if (navigator.share && this.isMobile()) {
         navigator.share({
           title: this.L('Your invite'),
           url: this.link
--- a/frontend/views/components/LinkToCopy.vue
+++ b/frontend/views/components/LinkToCopy.vue
@@ -26,6 +26,9 @@ component.c-wrapper(
 <script>
 import Tooltip from '@components/Tooltip.vue'
 
+// https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser#comment95674193_51774045
+const isMobile = !window.matchMedia('(any-pointer:fine)').matches
+
 export default ({
   name: 'LinkToCopy',
   components: {
@@ -47,13 +50,9 @@ export default ({
     }
   },
   methods: {
-    isMobile () {
-      // https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser#comment95674193_51774045
-      return !window.matchMedia('(any-pointer:fine)').matches
-    },
     copyToClipboard () {
       // if the device supports the Web Share API, use it and then skip other logic below.
-      if (navigator.share && this.isMobile()) {
+      if (navigator.share && isMobile) {
         navigator.share({
           title: this.L('Your invite'),
           url: this.link
--- a/test/cypress/integration/group-paying.spec.js
+++ b/test/cypress/integration/group-paying.spec.js
@@ -97,9 +97,15 @@ describe('Group Payments', () => {
   it('user1 sends $250 to user3 (total)', () => {
     cy.giSwitchUser(`user1-${userId}`, { bypassUI: true })
 
-    cy.giForceDistributionDateToNow()
+    // NOTE: TWO HEISENBUGS ARE IN THIS TEST! PLEASE LEAVE THESE COMMENTS FOR FUTURE
+    //       REFERENCE IN CASE WE RUN INTO MORE!
+    // NOTE: when we had this command here, Cypress would sometimes fail because cy.window() would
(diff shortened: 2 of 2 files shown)


frontend/views/components/LinkToCopy.vue
```diff
@@ -47,20 +47,13 @@ export default ({
...
-      // so we have to ensure that were not mistaking those browsers instead when matching for Safari on Macintosh
-      const userAgent = navigator.userAgent
-      const safari = /Macintosh.*Safari/
-      const chrome = /Macintosh.*Chrome/
-      const edge = /Macintosh.*Edg/
-      return userAgent.match(safari) &&
-              !(userAgent.match(chrome) || userAgent.match(edge))
+    isMobile () {
+      // https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser#comment95674193_51774045
+      return !window.matchMedia('(any-pointer:fine)').matches
```
taoeffect: Since Travis is failing (not because of this PR, but because of a heisenbug), I'm thinking maybe we should make a change here anyway, a small optimization (and plus trigger a re-run of the Travis tests): this value doesn't change, right? It should stay the same for the entire session?

If so, then we can define `isMobile` as a `const` variable at the top of the file instead of defining it as a function. This way we don't have to run the test each time `copyToClipboard` is called.

taoeffect (dismissed):
Looks great! Needs a small change to re-run the tests (Travis removed the ability for me to trigger the tests to re-run by clicking a button, so another commit is needed).

I've also added a small suggestion for what the change can be.

frontend/views/components/LinkToCopy.vue
snowteamer: Good point!

taoeffect (changes requested):
So - there are two **bizarre** heisenbugs in the `group-paying.spec.js` tests, which are currently failing in this PR.

I was able to reproduce what might be the same failure locally, and after much trial & error, was able to change the tests in odd ways to get them to start passing consistently locally.

Please apply the changes below to the `group-paying.spec.js` file, specifically to the `'user1 sends $250 to user3 (total)'` test, and hopefully we'll get the tests on Travis to finally pass:

```js
  it('user1 sends $250 to user3 (total)', () => {
    cy.giSwitchUser(`user1-${userId}`, { bypassUI: true })

    // NOTE: TWO HEISENBUGS ARE IN THIS TEST! PLEASE LEAVE THESE COMMENTS FOR FUTURE
    //       REFERENCE IN CASE WE RUN INTO MORE!
    // NOTE: when we had this command here, Cypress would sometimes fail because cy.window() would
    //       for some reason return a promise that wouldn't resolve.
    // cy.giForceDistributionDateToNow()

    cy.getByDT('paymentsLink').click()
    // Moving it here seems to have fixed things.
    cy.giForceDistributionDateToNow()
    cy.get('[data-test-date]').should('have.attr', 'data-test-date', humanDateToday)

    assertNavTabs(['Todo1', 'Completed'])
    assertMonthOverview([
      ['Payments sent', '0 out of 1'],
      ['Amount sent', '$0 out of $250']
    ])

    cy.getByDT('recordPayment').should('be.disabled')
    // NOTE: keep this comment around just to show the lengths we have to go to
    //       to get Cypress tests working consistently sometimes, before figuring
    //       out a cleaner workaround
    // cy.getByDT('payList').within(() => {
    //   cy.getByDT('todoCheck').find('[type="checkbox"]').check({ force: true })
    // })
    // NOTE: this is what I finally settled on to get around a timeout error
    //       related to a nonsensical "detached element" error in Cypress
    cy.getByDT('todoCheck').click({ force: true })
```


And for reference, here is the output of `git diff`:

```diff
diff --git a/test/cypress/integration/group-paying.spec.js b/test/cypress/integration/group-paying.spec.js
index 6d0c57a9..4dcd5318 100644
--- a/test/cypress/integration/group-paying.spec.js
+++ b/test/cypress/integration/group-paying.spec.js
@@ -97,9 +97,15 @@ describe('Group Payments', () => {
   it('user1 sends $250 to user3 (total)', () => {
     cy.giSwitchUser(`user1-${userId}`, { bypassUI: true })
 
-    cy.giForceDistributionDateToNow()
+    // NOTE: TWO HEISENBUGS ARE IN THIS TEST! PLEASE LEAVE THESE COMMENTS FOR FUTURE
+    //       REFERENCE IN CASE WE RUN INTO MORE!
+    // NOTE: when we had this command here, Cypress would sometimes fail because cy.window() would
+    //       for some reason return a promise that wouldn't resolve.
+    // cy.giForceDistributionDateToNow()
 
     cy.getByDT('paymentsLink').click()
+    // Moving it here seems to have fixed things.
+    cy.giForceDistributionDateToNow()
     cy.get('[data-test-date]').should('have.attr', 'data-test-date', humanDateToday)
 
     assertNavTabs(['Todo1', 'Completed'])
@@ -109,7 +115,15 @@ describe('Group Payments', () => {
     ])
 
     cy.getByDT('recordPayment').should('be.disabled')
-    cy.getByDT('todoCheck').click()
+    // NOTE: keep this comment around just to show the lengths we have to go to
+    //       to get Cypress tests working consistently sometimes, before figuring
+    //       out a cleaner workaround
+    // cy.getByDT('payList').within(() => {
+    //   cy.getByDT('todoCheck').find('[type="checkbox"]').check({ force: true })
+    // })
+    // NOTE: this is what I finally settled on to get around a timeout error
+    //       related to a nonsensical "detached element" error in Cypress
+    cy.getByDT('todoCheck').click({ force: true })
     cy.getByDT('recordPayment').should('not.be.disabled').click()
     cy.getByDT('modal').within(() => {
       cy.getByDT('payRecord').find('tbody').children().should('have.length', 1)
```

taoeffect (approved):
F$(* it. Overriding Travis failure and merging anyway.

At least `group-paying.spec.js` is passing on Travis. This time `group-proposals.spec.js` is failing.

Maybe they've fixed these bugs in the latest version of Cypress (and hopefully the latest version of Cypress isn't any slower than 7.7.0).

PR #1481, opened 2023-01-17
Closes #531 - message createdDate sync'd with backend server time

### LINKED ISSUE #531: Make sure backend clock is synced with clients

taoeffect:
This PR ensures that regardless of the user's local time settings, all messages will have correct metadata for `createdDate` that is based on the server's time.

Introduces a new global function for contracts to use for this purpose: `fetchServerTime`

Example:

```js
    async create () {
      const { username, identityContractID } = sbp('state/vuex/state').loggedIn
      return {
        createdDate: await fetchServerTime(),
        username,
        identityContractID
      }
    }
```


snowteamer (commented):
LGTM, but what's the reason to declare `fetchServerTime()` as a gobal rather as e.g. an SBP selector?

taoeffect:
@snowteamer:

> LGTM, but what's the reason to declare fetchSeverTime() as a gobal rather as e.g. an SBP selector?

That's a good question. Deciding what should be a selector and what should be a function is more of an art than a science. My thinking was along these lines:

- Having it be a selector would clog the logs unnecessarily
- Selectors tend to be reserved for things that can be considered actions / activities, and usually they consistent of more complicated logic than what's essentially a 1-liner.
- Functions are better reserved for things that can be considered "primitives". For example, you wouldn't make basic arithmetic operations a selector, or even algorithms like `qsort`.
- Having it be a selector would require also adding it to the whitelist of approved selectors.

PR #1496, opened 2023-02-15
Several Group Chat Issues

### LINKED ISSUE #1488: Fix all usage of prepublish

### LINKED ISSUE #1494: Several weird issues in joining group and group chat

### LINKED ISSUE #1495: Chatroom contract uses message "pending" attribute incorrectly

Silver-IT:
### Summary of Changes

- Closes #1494 
- Closes #1495 
- Closes #1488 
- Updated Notifications section in User Settings page


frontend/views/containers/user-settings/NotificationSettings.vue
```diff
@@ -1,19 +1,62 @@
...
+      //- .c-subcontent
+      //-   .c-text-content
+      //-     i18n.c-smaller-title(tag='h3') Chat
+      //-     i18n.c-description(tag='p') Get notified when other members sent you messages. You can also change frequency of these emails.
+      //-   label
+      //-     input.switch(
+      //-       type='checkbox'
+      //-       name='switch'
+      //-       :checked='chatGranted'
+      //-     )
```
SebinSong, line 59: If this block of commented lines are temporarily hiding this feature for a reason, it would be great if the context behind it is stated together here as well. :)

SebinSong:
@Silver-IT Looks good to me, Alex.

SebinSong (approved):

PR #1497, opened 2023-02-15
Emit notifications for various events

### LINKED ISSUE #1247: Emit notifications for all of the various important situations [$40 bounty]

SebinSong:
Closes #1247


frontend/model/state.js
```diff
@@ -674,15 +674,27 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
           // 'currentPaymentPeriod': gets auto-updated(t1) in response to the change of 'reactiveDate.date' when it passes into the new period.
           // 'groupSettings.distributionDate': gets updated manually by calling 'updateCurrentDistribution' function(t2) in group.js
           // This logic removes the inconsistency that exists between these two from the point of time t1 till t2.
-          if (!oldPeriod || !newPeriod) return
```
taoeffect, line 750: Why was this removed?

frontend/model/state.js
```diff
@@ -674,15 +674,27 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
...
 
           // Note: if this code gets called when we're in the period before the 1st distribution
           //       period, then the distributionDate will get updated to the previous distribution date
           //       (incorrectly). That in turn will cause the Payments page to update and display TODOs
           //       before it should.
           const distributionDateInSettings = store.getters.groupSettings.distributionDate
-          if (newPeriod !== oldPeriod && (newPeriod !== distributionDateInSettings)) {
+          if (oldPeriod && newPeriod && (newPeriod !== distributionDateInSettings)) {
```
taoeffect, line 757: Why was this modified?

frontend/model/state.js
```diff
@@ -674,15 +674,27 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
...
+            const userIncomeDetailsType = store.getters.ourGroupProfile?.incomeDetailsType
+
+            if (userIncomeDetailsType) {
+              sbp('gi.notifications/emit', 'NEW_DISTRIBUTION_PERIOD', {
+                creator: store.getters.ourUsername,
+                createdDate: new Date().toISOString(),
+                groupID: store.state.currentGroupId,
+                memberType: userIncomeDetailsType === 'pledgeAmount' ? 'pledger' : 'receiver'
+              })
+            }
```
taoeffect: A few issues here:

1. `NEW_DISTRIBUTION_PERIOD` is emitted in two different places in this PR
2.  Why is this logic part of a separate logical branch instead of being inside of the original `if`?

frontend/model/contracts/group.js
```diff
@@ -1242,13 +1291,32 @@ sbp('chelonia/defineContract', {
...
+        if (isActionYoungerThanUser(meta, myProfile) &&
+          myProfile.incomeDetailsType // if the user has entered their income details
+        ) {
+          sbp('gi.notifications/emit', 'NEW_DISTRIBUTION_PERIOD', {
+            creator: meta.username,
+            createdDate: meta.createdDate,
+            groupID: contractID,
+            memberType: myProfile.incomeDetailsType === 'pledgeAmount' ? 'pledger' : 'receiver'
+          })
+        }
```
taoeffect: The fact that we're in a new distribution period can be noticed and emitted locally without touching the contracts code

frontend/model/contracts/group.js
```diff
@@ -1242,13 +1291,32 @@ sbp('chelonia/defineContract', {
...
     ...((process.env.NODE_ENV === 'development' || process.env.CI) && {
       'gi.contracts/group/forceDistributionDate': {
         validate: optional,
-        process ({ meta }, { state, getters }) {
+        process ({ meta, contractID }, { state, getters }) {
           getters.groupSettings.distributionDate = dateToPeriodStamp(meta.createdDate)
+
+          sbp('gi.contracts/group/pushSideEffect', contractID,
+            ['gi.contracts/group/updateDistributionDate/sideEffect', { meta, contractID }]
+          )
```
taoeffect: I don't think these changes are needed here

frontend/model/contracts/group.js
```diff
@@ -1229,6 +1236,48 @@ sbp('chelonia/defineContract', {
...
+          (ourPayments && ourPayments.todo?.length > 0) &&
+          // Also, check if the notification has already been sent for the target period. - prevents sending the same notification multiple times.
+          !myNotificationHas(item => item.type === 'NEAR_DISTRIBUTION_END' && item.data.period === currentPeriod)
+        ) {
+          sbp('gi.notifications/emit', 'NEAR_DISTRIBUTION_END', {
+            createdDate: now,
+            groupID: contractID,
+            period: currentPeriod
+          })
+        }
```
taoeffect: This is a lot of logic to add to the contracts. Ideally - the contract code should be as simple as possible to avoid bugs and to keep the contract file small, as well as limiting the number of messages we send to the contracts (to not unnecessarily fill up servers with junk).

What this PR is trying to do is 2 things, I think:

1. Emit a notification when we've entered a new distribution period. This doesn't need to involve any contract code.
2. Emit a notification when it's 1 week before the end a distribution period.

AFAICT, neither of these things need to be done in the contract code. The logic for it can be done entirely locally.

And something that I've noticed is that this type of code currently exists in different places in the project. There is frequently a need to notify the user about something, and the checks for that are all over the place. This is a problem - because periodic checks to notify users when certain events are happening or about to happen should all be organized neatly in a single place in the app. There should be a single go-to file for adding such checks, and the logic for adding these checks should look and behave the same way.

For example, there is a function in `main.js` called `periodicStateChecks` that also has this problem. It's currently used to notify users about expiring proposals, and it also has the problem that it sends messages to the contracts (to mark globally sync'd state to set a boolean that says the notification was done, to prevent multiple notifications).

That check, and these checks, would be better done in a single file, and locally, without touching the contracts. And we should take this opportunity to do that correctly now, to prevent this problem from exploding into more spaghetti code down the line.

Here's what I propose:

Create a file that contains an array of objects that have this format:

```js
const every1MinQueue = [
  {
    notificationStateKey: '', // string or function that returns a string
                              // used to store a boolean in the Vuex
                              // rootState.periodicNotifications
    notificationCondition (): boolean {
      // returns a boolean that determins whether or not to
      // emit a notification
      return goodTimeToEmtiNotification() && !rootState.periodicNotifications[this.notificationStateKey]
    },
    notificationEmit () {
      // logic to emit a notifications
    },
    shouldClearNotificationStateKey (): boolean {
      // returns true to delete the
      // rootState.periodicNotifications[this.notificationStateKey]
      return enteredANewPeriod()
    }
  },
  // ... more periodic notification objects
]
const every5MinQueue = [
  // ... same format as above
]
const every30minQueue = [
  // ... same format as above
]

// to avoid using a lot of CPU, use the appropriate queue for tasks
// we don't want to perform intensive operations every minute unless
// absolutely necessary

setInterval(runCronQueue(every1MinQueue), 1 * MINS_MILLIS)
setInterval(runCronQueue(every5MinQueue), 5 * MINS_MILLIS)
setInterval(runCronQueue(every30MinQueue), 30 * MINS_MILLIS)
```

Then move all of the logic from this PR, and the logic from main's `periodicStateChecks` into that file, and delete the old way of doing things.

taoeffect (changes requested):
Nice work @SebinSong. After thinking about it for a while, I think the way we're solving this problem is going about it the wrong way, including some of the old code that I wrote. For periodic state checks that don't need to touch the contracts, we can do it all locally, and in a more organized way. Please see comments.

frontend/model/notifications/templates.js
```diff
@@ -219,5 +220,30 @@ export default ({
         }
       }]
     }
+  },
+  NEW_DISTRIBUTION_PERIOD (data: { creator: string, memberType: string }) {
+    const bodyTemplate = {
+      'pledger': L('A new distribution period has started. Please check Payment TODOs.'),
+      'receiver': L('A new distribution period has started. Please update your income details if they have changed.')
```
Silver-IT, line 227: I think @SebinSong would made the french translations, but I am not sure why they in English in french.json and french.string files.

Silver-IT (commented):
Nice work, @SebinSong. Since the @taoeffect made a detailed review, I couldn't find more things.

frontend/model/state.js
SebinSong: @taoeffect Logics here are simplified into `if (oldPeriod && newPeriod)` below.

frontend/model/state.js
SebinSong: @taoeffect I realised `newPeriod !== oldPeriod` here is not necessary as watcher being triggered already implies this.

frontend/model/contracts/group.js
SebinSong: Hi @taoeffect ,
Got a couple of questions regarding the above.

**Q1.** If I understood it right(please correct me if not), we would want to keep the `notificationStateKey` here in the root vuex state as boolean like below,

```js
{
  ...
  periodicNotifications: {
    key1: true,
    key2: false,
    ...
  }
}
```
What does each key being either `true` or `false` mean? are we keeping them to be able to toggle those notifications on/off when we want to in the app?

**Q2.** It's about how to compose the logic of `runCronQueue` function here. I have roughly written it down below based on what I understood from the comment. 

```js
function runCronQueue (queue) {

  for (const entry of queue) {
    if (rootState.periodicNotifications[entry.notificationStateKey] && entry.notificationCondition()) {
      entry.notificationEmit()

      if (entry.shouldClearNotificationStateKey()) {
        Vue.delete(rootState.periodicNotifications, entry.notificationKey)
      }
    }
  }
}

```

So on each interval, we go through the queue and take below steps for each entry.
**1)** check if the `notificationStateKey` is `true` in the rootState and also if `notificationCondition()` is truthy.
**2)** If conditions in the step **1)** are met, execute `notificationEmit()` logic
**3)** As the post action of **2)**, check if `shouldClearNotificationStateKey()` returns `true`, we clear the `notificationStateKey` from the root state so that the current entry is not executed from the next interval.

 is the above correct?
 
 Thanks,

taoeffect: > What does each key being either true or false mean? are we keeping them to be able to toggle those notifications on/off when we want to in the app?

They indicate whether or not the notification has already been fired. You can make this clearer by naming the rootState key something like `periodicNotificationsAlreadyFiredMap` instead of `periodicNotifications`. Using long descriptive variable names is a perfectly good stylistic way of coding that helps make the code more readable and avoids having to write a bunch of comments.

> Q2. It's about how to compose the logic of runCronQueue function here. I have roughly written it down below based on what I understood from the comment.

The first condition of the `if` that checks `notificationStateKey` needs to be negated, since we don't want to fire the same notification multiple times.

The other nested `if` condition should be outside of the first `if`, since if you keep it inside of the `if`, it won't get fired (because we go into the `if` only when a notification *hasn't* been fired - and the second condition — for clearing notifications — only happen after they've been fired)

SebinSong: @taoeffect 
> because we do into the if only when a notification hasn't been fired

Sorry, Could you help with what this mean?

taoeffect: *oops, I edited the comment to fix a typo. I meant "go" instead of "do"

SebinSong: Thanks for clarification @taoeffect ,
Just to make sure my understanding is correct before moving on, Does below looks good?

```js
function runCronQueue (queue) {
  const firedMap = sbp('state/vuex/state').periodicNotificationsAlreadyFiredMap

  for (const entry of queue) {
    if (!(entry.notificationStateKey in firedMap)) { continue } // if the key has already been cleared. ignore this entry.

    if (firedMap[entry.notificationStateKey] && entry.shouldClearNotificationStateKey()) {
      // if the notification has already been fired, check & clear the key
      Vue.delete(rootState.periodicNotifications, entry.notificationKey)
    } else if (entry.notificationCondition()) {
      entry.notificationEmit()
      Vue.set(firedMap, entry.notificationStateKey, true)
    }
  }
}
```

taoeffect: @SebinSong let's take this to Slack :)

frontend/model/notifications/templates.js
SebinSong: @Silver-IT I think that's what my last execution of `strings.mac` output and it looks like any missing translations falls back to their english scripts in the json files.

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -0,0 +1,156 @@
...
+           for defining more notifications
+
+  There is two types of notifications that can be defined here.
+  1) one-time notification: A type of notification that is checked & emitted once when user has signed in on their group.
+                            (Executed in response to 'SWITCH_GROUP' event)
+  2) periodic notification: A type of notication that is checked periodically and emitted when the conditions are met.
+
+  *-*-*-* notification entry structure *-*-*-*
+
+  1. oneTimeNotification: { name?: string, emitCondition: function, emit: function }
```
SebinSong, line 25: @taoeffect 
_'one-time'_ notification here was not part of what you requested in the previous review, but while working on 'INCOME_DETAILS_OLD' notification, which is to check if the user's income details are >= 6 months old, I realised this one only needs to be checked 'once' when user has logs in on their group, not needing to be implemented as a periodic notification. So I worked on adding this too here and hope it sounds good.

SebinSong:
@taoeffect I worked on your requests of adding 'periodic notifications' to the project and also re-implementing 'NEAR_PERIOD_END' and 'NEW_DISTRIBUTION_PERIOD' notifications using it. Tried my best to guess and write things according to your guidance up here and our discussions in Slack and hope it looks good.

frontend/model/notifications/periodicNotifications/index.js
```diff
@@ -0,0 +1,77 @@
...
+  emitCondition: isFunction,
+  emit: isFunction,
+  shouldClearStateKey: isFunction
+})
+
+async function runQueueRecursive (data) {
+  const rootState = sbp('state/vuex/state')
+  const rootGetters = sbp('state/vuex/getters')
+  const firedMap = rootState.periodicNotificationAlreadyFiredMap
+  const callWithStates = func => func.call(data.state, { rootState, rootGetters })
```
SebinSong: Calling `emitCondition()`, `emit()`, and `shouldClearStateKey()` this way was kind of inspired from the way we reference the contract states & getters inside the `process()`, `sideEffect()` functions. I wanted to be able to reference the vuex state and getters inside them like an example below

```js
emit ({ rootState, rootGetters }) {
  const { currentGroupId } = rootState
  const { ourUsername, ourPayments } = rootGetters
  ...
}
```

frontend/model/notifications/periodicNotifications/index.js
```diff
@@ -0,0 +1,77 @@
...
+// $FlowFixMe
+import { objectOf, string, isFunction } from '@model/contracts/misc/flowTyper.js'
+import { MINS_MILLIS } from '@model/contracts/shared/time.js'
+import { PERIODIC_NOTIFICATION_TYPE } from './constants.js'
+
+const { MIN1, MIN5, MIN30 } = PERIODIC_NOTIFICATION_TYPE
+
+const every1MinTimeout = { queue: [], state: {}, delay: 0.2 * MINS_MILLIS }
+const every5MinTimeout = { queue: [], state: {}, delay: 5 * MINS_MILLIS }
+const every30MinTimeout = { queue: [], state: {}, delay: 30 * MINS_MILLIS }
```
SebinSong: `state: {}` object here is for the case where we want to keep some data between different consecutive timeout calls. For example, it is used for implementing `NEW_DISTRIBUTION_PERIOD` notification, where we need to store and read the `distributionDate` value from the previous time call and compare it to the value in the current call. (`this.distPeriodRecords` there). hope it sounds good.

SebinSong: Oops, `0.2 * MINS_MILLIS` here is supposed to be `MILS_MILLIS` here. forgot to revert it back after testing. I will make the fix along with your upcomming change requests.

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -0,0 +1,156 @@
...
+        if (entry.emitCondition(arg)) {
+          await entry.emit(arg)
+        }
+      }
+    }
+  },
+  mounted () {
+    for (const entry of periodicNotificationEntries) {
+      sbp('gi.periodicNotifications/createEntry', entry)
+    }
```
taoeffect: While there is nothing particularly wrong about this, it will create a lot of log noise (the more notifications, the more noise).

To avoid the noise, I recommend moving the loop into the selector, and renaming the selector to `'gi.periodicNotifications/importNotifications'`

frontend/model/notifications/periodicNotifications/constants.js
```diff
@@ -0,0 +1,5 @@
+export const PERIODIC_NOTIFICATION_TYPE = {
+  MIN1: '1MIN',
+  MIN5: '5MIN',
+  MIN30: '30MIN'
+}
```
taoeffect: This is just a personal preference, but I don't see a reason to create a separate file, and therefore an `index.js` file and a folder. I think these could be included as part of `periodicNotifications.js` without problems. And then there will be a single `constants.js` file in the project that is used for more globally relevant constants.

frontend/model/notifications/periodicNotifications/index.js
```diff
@@ -0,0 +1,77 @@
...
+import sbp from '@sbp/sbp'
+import { Vue } from '@common/common.js'
+// $FlowFixMe
+import { objectOf, string, isFunction } from '@model/contracts/misc/flowTyper.js'
+import { MINS_MILLIS } from '@model/contracts/shared/time.js'
+import { PERIODIC_NOTIFICATION_TYPE } from './constants.js'
+
+const { MIN1, MIN5, MIN30 } = PERIODIC_NOTIFICATION_TYPE
+
+const every1MinTimeout = { queue: [], state: {}, delay: 0.2 * MINS_MILLIS }
```
taoeffect: A `queue` is the name given to a FIFO array (first-in-first-out). However, AFAICT, this array is not being used as a queue. The notifications are added to it, but aren't removed once they're processed.

So perhaps a better name is called for, maybe `notifications`?

frontend/main.js
```diff
@@ -244,7 +247,13 @@ async function startApp () {
       sbp('okTurtles.events/on', LOGOUT, () => {
         this.ephemeral.finishedLogin = 'no'
         router.currentRoute.path !== '/' && router.push({ path: '/' }).catch(console.error)
+        sbp('gi.periodicNotifications/destroyAll')
       })
+      sbp('okTurtles.events/on', SWITCH_GROUP, () => {
+        this.initOrResetPeriodicNotifications()
```
taoeffect: As far as I can tell, `destroyAll` is somewhat misleadingly named... there doesn't seem to be a way to clear the notifications once they're imported, as `destroyAll` only stops the timers from firing.

- Should the API be modified to add the ability to clear the imported notifications? (And if so, then the code needs to be updated to also re-import them)
- Or should `destroyAll` be renamed to something like `stopTimers`?

frontend/main.js
```diff
@@ -244,7 +247,13 @@ async function startApp () {
...
+        this.checkAndEmitOneTimeNotifications()
+      })
+
```
taoeffect, line 258: Remember that below here we still have the `periodicStateChecks` function, which now is no longer necessary, and its contents can be moved into the `periodicNotificationEntries` array. Do you think you can move that logic for this PR?

taoeffect (changes requested):
This PR creates a very elegant, centralized periodic notifications system for the project. Fantastic work, @SebinSong! 😄 

My comments are below.

frontend/main.js
SebinSong: @taoeffect I agree the name can be misleading. For the usage of periodic notifications in this PR, there is no need to clear the imported notifications but the timers & relevant states do need to be cleared and initialised once either the user switches to another group or logs out. These are currently taken care of in `/destroyAll` selector, so let me rename it to `stopTimers`(or `clearStatesAndStopTimers`) as you suggested.

frontend/main.js
```diff
@@ -242,14 +240,19 @@ async function startApp () {
...
+        }
       })
       sbp('okTurtles.events/on', LOGOUT, () => {
         this.ephemeral.finishedLogin = 'no'
         router.currentRoute.path !== '/' && router.push({ path: '/' }).catch(console.error)
-        sbp('gi.periodicNotifications/destroyAll')
+        sbp('gi.periodicNotifications/clearStatesAndStopTimers')
       })
       sbp('okTurtles.events/on', SWITCH_GROUP, () => {
+        console.log('@@@ hello')
```
taoeffect: Don't forget to remove debug logging before pushing :)

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -47,6 +48,9 @@ const oneTimeNotificationEntries = [
       const { incomeDetailsLastUpdatedDate } = (rootGetters.ourGroupProfile || {})
       const now = new Date().toISOString()
 
+      console.log('@@@ incomeDetailsLastUpdatedDate: ', incomeDetailsLastUpdatedDate)
+      console.log('@@@ now: ', now)
+      console.log('@@@ compareISOTimestamps(now, incomeDetailsLastUpdatedDate) > 6 * MONTHS_MILLIS: ', compareISOTimestamps(now, incomeDetailsLastUpdatedDate) > 6 * MONTHS_MILLIS)
```
taoeffect: Also here

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -127,13 +131,60 @@ const periodicNotificationEntries = [
...
+        }).filter(entry => entry.proposals.length)
+
+        return this.expiringProposalsByGroup.length
+      },
+      async emit () {
+        for (const { contractID, proposals } of this.expiringProposalsByGroup) {
+          await sbp('gi.actions/group/notifyExpiringProposals', {
+            contractID,
+            data: { proposals }
+          })
```
taoeffect, line 158: So the idea is ultimately to notify the user of something without touching the contract code.... and I'm trying to figure out what this code does and where the notification is ultimately sent. And I'm coming to the conclusion that the notification itself was never implemented!

Because this action sends a message to the contracts that ultimately just sets a `notifiedBeforeExpire` variable on a particular proposal to indicate that we've notified the user about this about-to-expire proposal - but I can't find where that notification is actually sent. I don't think it's sent?

So - we do need a way to keep track of whether or not we've notified about a particular proposal. So for that we can use the `state` that you've created, and use the `proposalId` to key an object on the state as to whether or not we've already emitted a notification for that proposal, e.g. `this.proposalNotificationSent[proposalId]`, and use that instead of `notifiedBeforeExpire`

taoeffect: One more thought - regarding the design of using `this.proposalNotificationSent[proposalId]` to keep track of notified expiring proposals - as far as I can tell, that state will not be cleared until the browser page is refreshed.

And in this case `shouldClearStateKey` isn't being used for this purpose, because this notification really has within it a whole bunch of other notifications (one for each expiring proposal).

Just something to be aware of, maybe deserving of a comment too, because it's different from how the other notifications behave.

SebinSong: @taoeffect 
> but I can't find where that notification is actually sent. I don't think it's sent?

Yes, it looks like the `gi.contracts/group/notifyExpiringProposals` only toggles `notifiedBeforeExpire` prop in the proposal object but doesn't actually send the notification as the action side-effect. So it seems like a new notification type (e.g. PROPOSAL_EXPIRING) has to be added and the logic for sending it needs to be implemented too.

`this.proposalNotificationSent[proposalId]` sounds like a good idea.
Btw, there is another way of checking if a particular notification has already been sent before that I've created while working on this PR, which is `myNotificationHas` method (L9 in `mainNotificationsMixin.js`). we can use this method too.

Lastly, this is a topic that I have never looked at so far, so even though this is fairly relevant to the original issue(which is implementing various types of notifications), but quite overwhelming. So is it alright for this additional work to be handled as a separate issue in the future either by me or someone else?

Thanks,

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -0,0 +1,204 @@
...
+
+        this.distPeriodRecords = {
+          prevPeriod: currentPeriod,
+          prevNow: now
+        }
+
+        return check
+      },
+      emit ({ rootState, rootGetters }) {
+        sbp('gi.notifications/emit', 'NEW_DISTRIBUTION_PERIOD', {
```
SebinSong, line 111: @taoeffect I have a question regarding this **'NEW_DISTRIBUTION_PERIOD'** notification here.
What the logic here does is, every 1 minute it checks if the distribution period has been updated and send notification if it has. But it's currently only targeting a very specific case where the distribution period changes _**while the user is online**_. (which happens quite frequently while we're testing/developing but I guess would rarely happens in real usage. mostly, the distribution date will be updated while users are not using the app and then they will notice it once they log in again)

So do you think we have to add this same entry to the _one-time_ notifications list so that once the user logs in to the group, the app checks if the user has been notified of the currently-active distribution period and if not, send the notification? :) (Or should we just move this entry to the one-time notifications?)

taoeffect: I don't know, I'm thinking maybe there's a way to improve the emitCondition logic that hasn't been explored?

For example, what about simply checking for `comparePeriodStamps(dateToPeriodStamp(now), currentPeriod) > 0` ?

Btw - remember to convert `now` to a period stamp using `dateToPeriodStamp` before using it inside of `comparePeriodStamps`.

frontend/model/notifications/mainNotificationsMixin.js
taoeffect: > Lastly, this is a topic that I have never looked at so far, so even though this is fairly relevant to the original issue(which is implementing various types of notifications), but quite overwhelming. So is it alright for this additional work to be handled as a separate issue in the future either by me or someone else?

Yep! 👍

frontend/model/notifications/mainNotificationsMixin.js
SebinSong: @taoeffect Updated the logic in `emitCondition()` to simplify it.

frontend/model/notifications/mainNotificationsMixin.js
SebinSong: Once, the issue is created, please assign myself to it.

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -0,0 +1,193 @@
...
+  {
+    type: PERIODIC_NOTIFICATION_TYPE.MIN1,
+    notificationData: {
+      stateKey: 'nearDistributionEnd',
+      emitCondition ({ rootGetters }) {
+        const currentPeriod = rootGetters.groupSettings?.distributionDate
+        if (!currentPeriod) { return false }
+
+        const nextPeriod = rootGetters.periodAfterPeriod(currentPeriod)
+        const now = new Date().toISOString()
```
taoeffect, line 76:
```js
const now = dateToPeriodStamp(new Date())
```

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -0,0 +1,193 @@
...
+      emit ({ rootState, rootGetters }) {
+        sbp('gi.notifications/emit', 'NEW_DISTRIBUTION_PERIOD', {
+          createdDate: new Date().toISOString(),
+          groupID: rootState.currentGroupId,
+          creator: rootGetters.ourUsername,
+          memberType: rootGetters.ourGroupProfile.incomeDetailsType === 'pledgeAmount' ? 'pledger' : 'receiver'
+        })
+      },
+      shouldClearStateKey ({ rootGetters }) {
+        return comparePeriodStamps(new Date().toISOString(), rootGetters.groupSettings.distributionDate) > 0
```
taoeffect, line 119: `dateToPeriodStamp(new Date())` here too (and anywhere else I might've missed, please double-check, this is to prevent possibility of future bugs)

taoeffect (approved):
Approved! Great work @SebinSong!

I'll create a followup issue for this PR as well to address:

- [Handling expiring proposals notification](https://github.com/okTurtles/group-income/pull/1497#discussion_r1121173815)
- Using `dateToPeriodStamp` where appropriate

Issue #1503, opened 2023-03-01
A new way needed to identify GIMessages


Silver-IT:
### Problem

For the discussion on the following link, we pointed that there should be a way to identify GIMessages which is not possible at the moment.
https://github.com/okTurtles/group-income/issues/1487#issuecomment-1449429880

EDIT by @taoeffect: see especially this code that needs to be fixed in ChatMain.vue:

```js
      sbp('gi.actions/chatroom/addMessage', {
        contractID: this.currentChatRoomId,
        data: !replyingMessage ? data : { ...data, replyingMessage },
        hooks: {
          prepublish: (message) => {
            const msgValue = JSON.parse(message.opValue())
            const { meta, data } = msgValue
            this.messageState.contract.messages.push({
              // TODO: message.hash() could be different from this message on event listener
              // https://github.com/okTurtles/group-income/issues/1487
              // so that the message.hash() can't be used as the identifier of this message
              // https://github.com/okTurtles/group-income/issues/1503
              // After this issue is fixed, we can fix this TODO and identity messages using message.id() instead of hash()
```

### Solution

Make a static GIMessage function to make a GIMessage instance with a provided `nonce`. The function works almost similar to `createV1_0` but the only different would be using `nonce`.

Issue #1371, opened 2022-08-02
Fix login after failure to sync when username exists


taoeffect:
### Problem

This problem occurs in the following scenario:

1. You've logged in on Browser A to your account and everything is working fine.
2. The server has lost its data or deleted your group's contracts
3. You've re-registered your account on Browser B
4. Now Browser A will show the following error when it refreshes:

![Group Dashboard copy](https://user-images.githubusercontent.com/138706/182317194-577360f5-c8b8-4fde-b332-9c78f94accaf.jpeg)

### Solution

When the server resets and you’ve recreated your username - but none of the contracts exist (including a different identity contract for yourself!) - there needs to be a way to delete your local contracts and reset everything. And in the future there needs to be a way to migrate your data to another server if you'd like, but for now it's more important that your browser doesn't get "stuck" in an unrecoverable way.

Maybe use `gi.ui/prompt` to ask if they’d like to reset everything, and if they say yes, delete the contracts that cannot be found and reset your local state afresh.

Silver-IT:
> When the server resets and you’ve recreated your username - but none of the contracts exist (including a different identity contract for yourself!) - there needs to be a way to delete your local contracts and reset everything.

@taoeffect, you mentioned `none of the contracts exist`. But what can we do if only a few contracts don't exist?
For example, UserA joined two groups; GroupA and GroupB. But accidently, only GroupB contract (or HEAD file) is deleted.
What can we do in this situation? Should we allow him to login or reset everything?

**In my opinion, if any of them doesn't exist, we should reset.**

taoeffect:
There is a problem with resetting everything, and that is we have to treat identity contracts differently from group contracts. It is a much bigger deal to reset an identity contract than it is to reset a group. In fact, at the moment we cannot reset the lookup from username to identity contracts at all because they are cached, and it might stay that way for security reasons.

This issue specifically focuses on recovering from a situation where the identity contract has been wiped, and then the username is re-registered (creating a new identity contract). Again - this is a serious problem because any users from before who have interacted with this user will no longer be able to interact with them, because of the way each user locally caches the username->identityContract lookup.

You're are right that we should also handle the case where group contracts become unavailable suddenly. And that should be handled in a totally different way I think. The UI should inform the user that their group is now unusable as it appears to no longer exist on the server, but it should give them the option to recreate the group (perhaps on a different server) with the same state.

[...]

PR #1508, opened 2023-03-02
Improve chat service & fix some more issues

### LINKED ISSUE #1374: Chat: do not refetch messages if not needed
### Problem

Currently, switching back and forth between the dashboard and the chat will cause the chat messages to be refetched from the server each time.

### Solution

Don't refetch messages if it's not needed.

### LINKED ISSUE #1507: Some chat issues in production server
### Problem

1. It makes error when try to create direct message.
2. In `General` chatroom, we should see the `Joined General` message (which is created automatically when a new member is joined the group). But we can not see them.

@taoeffect, described issues in detail here in Slack.
https://okturtles.slack.com/archives/C0EH7P20Y/p1677716969380129

### Solution

I am not sure what the reason is at this moment. Need to debug.

### LINKED ISSUE #1511: Incorrect title when route is changed in the channel which is not joined
### Problem

If the route is changed in the channel which is not part of, the title is set `'Loading'`.
Steps to Reproduce:
1. UserA creates a channel ChannelA
2. UserB views the ChannelA. Not yet joined and just view.
3. Click `+ New` button to create a new channel or click `Settings` button, so that a new modal would be opened and the route would be changed.
4. Check the title. It will display `'Loading'` which is wrong.

### Solution

Seems like the `refreshTitle` function in `ChatroomMixin.js` didn't catch all the cases.

Silver-IT:
### Summary of Changes

1. Closes #1374 
2. Closes #1507 
3. Closes #1511


Silver-IT (commented):
All seems to be ready for your review, @taoeffect

taoeffect (changes requested):
@Silver-IT could you remove your changes related to #1371 from this PR? Please see this comment for more details: https://github.com/okTurtles/group-income/issues/1371#issuecomment-1458993957

Also, to make it easier for us to coordinate around PR reviews in the future, please try submit separate PRs for separate issues, especially if the issues are very different. This is part of the [contributing guidelines](https://github.com/okTurtles/group-income/blob/master/CONTRIBUTING.md#pull-request-requirements-exclamation):

> 1. A PR SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.

It helps to speed up PR reviewing and merging for both contributors and maintainers. :)

Silver-IT:
> Also, to make it easier for us to coordinate around PR reviews in the future, please try submit separate PRs for separate issues, especially if the issues are very different.

Got it. Will do from the next time.

frontend/model/contracts/mailbox.js
```diff
@@ -63,7 +63,7 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/mailbox/setAttributes': {
       validate: (data, { state, meta }) => {
-        if (state.attributes.creator !== meta.username) {
+        if (state.attributes && state.attributes.creator !== meta.username) {
```
taoeffect, line 66: I am realizing that if `state.attributes` is not defined, we are calling `'gi.actions/mailbox/setAttributes'`, but that assumes `state.attributes` is defined.

So something like this needs to be added to the `process`:

```js
      process ({ meta, data }, { state }) {
        // TODO: remove this hack before releasing version 1
        if (!state.attributes) {
          Vue.set(state, 'attributes', {})
        }
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -402,13 +411,15 @@ export default ({
...
+      // NOTE: this state is rendered using the chatroom contract functions
+      // so should be CAREFUL of updating the fields
+      this.messageState.contract = {
         settings: cloneDeep(this.chatRoomSettings),
         attributes: cloneDeep(this.chatRoomAttributes),
         users: cloneDeep(this.chatRoomUsers),
-        messages: initialize ? [] : this.messages,
-        onlyRenderMessage: true
+        messages: [],
+        onlyRenderMessage: true // NOTE: DO NOT CHANGE THIS, OR CHATROOM COULD BREAK
```
taoeffect: Recommend `CHANGE` => `RENAME` to be more specific

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -642,13 +680,41 @@ export default ({
...
+        // this.currentChatRoomId could be wrong when the channels are switched very fast
+        // so it's good to initiate using input parameter chatRoomId
+        sbp('gi.db/archive/save', this.getArchiveKeyFromChatRoomId(chatRoomId), JSON.stringify(this.latestEvents))
+      }
+    },
+    getArchiveKeyFromChatRoomId (chatRoomId) {
+      const curChatRoomId = chatRoomId || this.currentChatRoomId
+      return `messages/${curChatRoomId}`
+    },
+    refreshContent: debounce(function (from, to) {
```
taoeffect, line 718: Below this I see a call to `setMessageEventListener`, which seems to set a lot of event listeners. But I don't see a corresponding `'okTurtles.events/off'` call for all of these:

```js
      if (force) {
        sbp('okTurtles.events/on', `${CHATROOM_MESSAGE_ACTION}-${to || this.currentChatRoomId}`, this.listenChatRoomActions)
      } else if (this.isJoinedChatRoom(to)) {
        sbp('okTurtles.events/on', `${CHATROOM_MESSAGE_ACTION}-${to}`, this.listenChatRoomActions)
      }
```

This seems like something I didn't catch in a previous PR review.

frontend/model/contracts/mailbox.js
Silver-IT: That's a nice catch. Correct.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Got it.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: I think we remove the event listeners in these two places by calling `sbp('okTurtles.events/off')`.
1. In `beforeDestroy` life cycle
2. in `setMessageEventListener` function

taoeffect: Hmm, you're right. I didn't understand that `setMessageEventListener` properly handled this

frontend/controller/actions/identity.js
```diff
@@ -228,7 +228,6 @@ export default (sbp('sbp/selectors/register', {
     } catch (e) {
       console.error('gi.actions/identity/login failed!', e)
       const humanErr = L('Failed to login: {reportError}', LError(e))
-      alert(humanErr)
```
taoeffect: I think this should be added back. Seems related to changes to #1371

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -283,7 +292,7 @@ export default ({
           prepublish: (message) => {
             const msgValue = JSON.parse(message.opValue())
             const { meta, data } = msgValue
-            this.messages.push({
+            this.messageState.contract.messages.push({
```
taoeffect, line 294: Regarding issue #1503 and the comment below, you could instead rename this handler from `prepublish` to `postpublish`. Then it would be safe to use `message.hash()` safely as the identifier.

Silver-IT: Yeah. It's a nice solution in terms of comparing GIMessages.
On the other hand, the pending could work a bit weird, if we add the pending message in `postpublish`.
Because the message won't be added immediately after users press `Enter`.

What do you recommend?
1. Keep it in `prepublish` and update later
2. Move it to `postpublish` and update later

frontend/controller/actions/identity.js
Silver-IT: Yeah, I know. But in my experience, this `alert` isn't needed. Because we log that `humanErr` in `main.js` exception handler.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: I think we should for now put it in `postpublish`, because at least will avoid weird bugs that would be caused by the `messageId` being wrong

Silver-IT: Sometimes the `sbp('okTurtles.events/once', hash, () => {...})` is called before the `postpublish`. So one event can add two messages if we add pending message in `postpublish`.
I think we should keep it here in `prepublish` and fix the Issue #1503 and then update again.

taoeffect: Oh, weird. OK 👍

Silver-IT: Thanks

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -531,9 +564,9 @@ export default ({
       }
 
       sbp('okTurtles.events/once', hash, async (contractID, message) => {
-        if (contractID === this.currentChatRoomId) {
-          const state = this.getSimulatedState(false)
-          await sbp('chelonia/private/in/processMessage', message, state)
+        // NOTE: while syncing the chatroom contract, we should ignore all the events
```
taoeffect, line 578: I don't quite understand this `sbp('okTurtles.events/once', hash` event listener.

This function, `listenChatRoomActions`, will be called by `emitMessageEvent` in the `sideEffect`. `emitMessageEvent` could easily pass in `contractID` along with `hash`. I do not see a reason for immediately setting up an event listener here on `hash`, because it will simply be immediately called after the side effect in `handleEvent`: `sbp('okTurtles.events/emit', hash, contractID, message)`

What I'm saying is, I'm pretty sure you can remove `sbp('okTurtles.events/once', hash, async (contractID, message) => {`, and place the code inside of it into this function directly, right? (as long as you modify `emitMessageEvent` to pass along the `contractID` too)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -642,13 +680,41 @@ export default ({
...
+      }
+      const unit = this.chatRoomSettings?.actionsPerPage || CHATROOM_ACTIONS_PER_PAGE
+      const from = this.latestEvents.length ? GIMessage.deserialize(this.latestEvents[0]).hash() : null
+      const to = this.latestEvents.length
+        ? GIMessage.deserialize(this.latestEvents[this.latestEvents.length - 1]).hash()
+        : null
+
+      // NOTE: save messages in the browser storage, but not more than 3 pages of events
+      if (this.latestEvents.length >= 3 * unit) {
+        sbp('gi.db/archive/delete', this.getArchiveKeyFromChatRoomId(chatRoomId))
```
taoeffect, line 707: The way these messages are deleted doesn't make much sense to me.

Instead of keeping the number of stored events limited to `3 * unit`, this just deletes all of these events. But when then will it save the events again?

 This just doesn't seem right to me.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: That's reasonable. The reasons why I made this way are because
1. I wanted to use `message` object in the function
2. As long as `listenChatRoomActions` is sync, there won't be any issues because `okTurtles.events/emit` is defined just like the following.
```
'okTurtles.events/emit': function (event: string, ...data: any) {
  for (const listener of sbp('okTurtles.data/get', listenKey(event)) || []) {
    listener(...data)
  }
}
```

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: 1. I need to define the constant value for the number 3. (Not sure 3 is the best choice. It could be 5 or any other)
2. Do you remember `MAX_ARCHIVED_PERIODS` which we use in group contract? We archive payments but no more than `MAX_ARCHIVED_PERIODS` periods of payments. Here, it works the same too. We are not going to save more than 3 pages of events for each channel. The events are always saved to the IndexedDB unless the loaded events are for more than 3 pages or user didn't load any events(old or new) after he opens the page.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: If I remove that event handler `sbp('okTurtles.events/once', hash)`, I am not sure how I can get the `message` object in `listenChatRoomActions` function.

taoeffect: Oh! Good point! Can you add this comment above line 569?

```js
// we use this listener here so that we can get access to the GIMessage
// it will be called immediately after this function is called
```

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: OK, I cannot follow everything you're doing here. However, there is an easy way to make sure everything works correctly: make it easy to detect bugs.

So, please for now set `CHATROOM_MAX_ARCHIVE_ACTION_PAGES` to `2`. This way, if there's a bug in the archiving code, we will catch it earlier. Then later we can set it back to `3` after we're sure that everything is working OK.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Sure. Can I make this comment a bit more informative like the following?

```
// we use this listener here so that we can get access to the GIMessage
// it will be called immediately after this function is called as long as listenChatRoomActions function is sync
```

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Okay. I will set `CHATROOM_MAX_ARCHIVE_ACTION_PAGES` to `2` then.

taoeffect (changes requested):
With even 1 user, when I switch between the Chat to the Dashboard and back, there are calls to `'chelonia/out/eventsBetween'` and `'chelonia/out/eventsBefore'`:


Screen Shot 2023-03-09 at 3 34 20 PM

Screen Shot 2023-03-09 at 3 34 30 PM

If this PR closed the issue, these calls shouldn't happen each time I switch between the Chat <-> Dashboard.

Silver-IT:
Yeah. Nice catch. I will explain the reason why those two `sbp` calls are made.
**eventsBetween**: Every time we enter the chatroom, we should check if the locally stored messages are up-to-date. So we call `eventsBetween` between the `latestHash` and the `lastHash from the store`. But you pointed the important thing. Since they are same, it shouldn't be called in this situation. I think I handled this but there might be a mistake.
**eventsBefore**: We need to display at least one page of messages. I guess there were a few messages when you tested. So it tried to get more messages to fill the page (at least one page).

Silver-IT (commented):
@taoeffect, It's ready for you to re-review.

Silver-IT (commented):
It seems ready for the final review.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: Yep! Of course, more informative is always better. :)

Although, that last part might have a typo:

>  as long as listenChatRoomActions function is sync

Did you mean "is syncing"? or "is synced"? Or what?

Silver-IT: `as long as listenChatRoomActions is sync function`. I will update the comment.

frontend/controller/actions/identity.js
taoeffect: @Silver-IT but I think we want the user to see the error in this case, because it's a serious problem that would need manual intervention to solve, and this way if anyone runs into it they'll be able to take a screenshot of the error and we'll know. Most normal users don't check the console.

Silver-IT: Yeah, that's reasonable. Then, let's add it again.

frontend/model/contracts/chatroom.js
```diff
@@ -224,10 +225,7 @@ sbp('chelonia/defineContract', {
       },
       sideEffect ({ contractID, hash, meta }) {
         emitMessageEvent({ contractID, hash })
-
-        if (sbp('chelonia/contract/isSyncing', contractID)) {
-          updateUnreadPosition({ contractID, hash, createdDate: meta.createdDate })
-        }
+        updateUnreadPosition({ contractID, hash, createdDate: meta.createdDate })
```
taoeffect, line 228: So the logic here completely reversed. Before, if it was called from a contract sync, we would update the unread position. Now, it no longer does.

Is this change intentional?

(Same question for the line above on line 208)

Silver-IT: Nice catch. Actually, I didn't notice that it reverses the previous workflow.
But the problem is that I can't understand the previous workflow to set `UnreadPosition` while the chatroom contract is syncing.
It makes sense to set `UnreadPosition` after joining the chatroom. So no need to set `UnreadPosition` while syncing the contract and the current solution is correct.

Silver-IT (commented):
Resolved your feedback.

taoeffect (approved):
Nice improvements @Silver-IT, well done!

PR #1512, opened 2023-03-13
Archive Chatroom

### LINKED ISSUE #1500: Archive Chatroom
### Problem

The Figma design says that the chatroom would be archived when the creator leaves it.
To archive chatroom means that the members are able to access and check the previous messages but no longer able to send any more messages. I don't think the Figma design is fully provided for the details.
1. The confirm button title when the creator tries to leave channel. `Leave and Archive` or `Leave and Archive Channel` or ...
2. How does it look like when a user is on the archived chatroom
3. How to unarchive the archived chatroom

### Solution

1. Confirm the workflow and make a design
2. Clone the Figma design in frontend
3. Implement workflow

Silver-IT:
### Summary of Changes

- Closes Issue #1500 
- Added `box-shadow` to Card when it's in dark theme


taoeffect:
https://github.com/okTurtles/group-income/issues/1500#issuecomment-1513885219

PR #1513, opened 2023-03-14
New way to identify GIMessage using id() function

### LINKED ISSUE #1503: A new way needed to identify GIMessages

### LINKED ISSUE #1507: Some chat issues in production server

Silver-IT:
### Summary of Changes

- Closes #1503 
- Removed hack for older versions regarding Create/Join DMs
- Closes #1507 
- Fix the Travis heisenbug regarding `updateLastLoggedIn` function call
- Fix the Travis heisenbug regarding `inifinite loading spinner` in group chat
- Refactoring test scenario for `GroupChat`


Silver-IT (commented):
@taoeffect, this PR seems ready for your review.

shared/domains/chelonia/GIMessage.js
```diff
@@ -143,6 +164,17 @@ export class GIMessage {
...
+  id (): string {
+    const data: string = JSON.stringify({
+      contractID: this.contractID(),
+      op: this.op(),
+      manifest: this.manifest(),
+      nonce: this._message.nonce
+    })
+
+    return blake32Hash(data)
+  }
```
taoeffect: So I get that the idea here is to include everything except for the `previousHEAD`, and  yeah that makes sense, however there is an important JavaScript quirk to be aware of that makes it so that we cannot do it this way. There is a comment on line 60 that discusses this a bit, although it doesn't explain why:

```js
    // NOTE: the JSON strings generated here must be preserved forever.
    //       do not ever regenerate this message using the contructor.
    //       instead store it using serialize() and restore it using
    //       deserialize().
```

The issue is that different implementations of JavaScript engines might generate different strings when serializing JS objects using `JSON.stringify`, and that would lead to different hashes resulting from `blake32Hash`. So to get around this we save the serialized string upon creation and keep a copy of it (instead of regenerating it as needed).

So one way to solve this would be to save this hash upon message creation and include it as part of the data that is signed. However, there is also an easier solution, and that is to just create a unique ID. There is a name for this: [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier):

> While the probability that a UUID will be duplicated is not zero, it is generally considered close enough to zero to be negligible

A UUID is 128-bits, that's twice the size of the value of `Math.random()`. You can add the dependency `uuid` to the project to do this, it is considered a high-quality dependency: https://github.com/uuidjs/uuid

Then you can simply replace the `nonce` with a `uuid`, and add an accessor function to access it. This would be the simplest and easiest approach with minimal changes, and you can do it in a way that is DRY. You would still need a a function to copy the GIMessage though.

shared/domains/chelonia/GIMessage.js
```diff
@@ -72,6 +72,27 @@ export class GIMessage {
     })
   }
 
+  // GIMessage.assign is necessary when make an GIMessage object having the same identity id()
+  // https://github.com/okTurtles/group-income/issues/1503
+  static assign (
```
taoeffect: I recommend renaming this to `clone` or `cloneWith`

taoeffect (changes requested):
Reviewed! Nice work @Silver-IT!

shared/domains/chelonia/GIMessage.js
Silver-IT: I used the name `assign` from the `Object.assign`. We can use `cloneWith` either.

shared/domains/chelonia/GIMessage.js
Silver-IT: Very interesting to hear that `JSON.stringify()` could make different strings according to the JavaScript version.
So we can make `id()` function which returns the `this._message.nonce` as we use `uuid` for `nonce`. This is absolutely a simple and easy approach.

> You would still need a a function to copy the GIMessage though.

I tried to do this but as I noted in the comment (Line 83), I don't think we should make a new static function for this.
The reason why the new function (which would be used to fix DRY) should be `static` function is because it couldn't return `this` type as long as it's not class function. Logically, type `this` is different from type `GIMessage`. So I used twice to create GIMessage in both functions; `assign` and `createV1_0`.

Silver-IT: I think we can make a function which returns only `{ mapping, message }`, not `this` typed object.

dotmacro: > There is a comment on line 60 that discusses this a bit, although it doesn't explain why:

> The issue is that different implementations of JavaScript engines might generate different strings when serializing JS objects using `JSON.stringify`, and that would lead to different hashes resulting from `blake32Hash`. So to get around this we save the serialized string upon creation and keep a copy of it (instead of regenerating it as needed).

@taoeffect should this "why" be added to the comment on line 60 in some form?

Silver-IT (commented):
@taoeffect, seems ready for your review again.

shared/domains/chelonia/GIMessage.js
taoeffect: Yeah probably a good idea, thanks @dotmacro

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -300,7 +300,8 @@ export default ({
               ...createMessage({ meta, data, hash: message.hash() }),
               // TODO: pending is useful to turn the message gray meaning failed (just like Slack)
               // when we don't get event after a certain period
-              pending: true
+              pending: true, // NOTE: do not RENAME this as it's being used inside the contract
```
taoeffect: Above this line we are using `createMessage` and are still passing in the `hash`. But is that right? Should be instead passing in the `id`? Or in addition to the `hash` also passing in the `id` to `createMessage`?

Also, there is the now outdated comment on lines 295 to 299. It should be rewritten to explain why `message.id()` is being used, instead of `.hash()` and the `TODO:` removed.

shared/domains/chelonia/GIMessage.js
```diff
@@ -55,21 +58,20 @@ export class GIMessage {
...
+
+  // GIMessage.cloneWith could be used when make a GIMessage object having the same id()
+  // https://github.com/okTurtles/group-income/issues/1503
+  static cloneWith (
+    target: GIMessage,
+    sources: Object, // TODO: type check is needed
+    signatureFn?: Function = defaultSignatureFn
+  ): this {
+    const message = Object.assign({}, target.message(), sources)
+    return new this(messageToParams(message, signatureFn))
```
taoeffect, line 74: Nice job DRYing this using the `messageToParams` function 👍

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch. Passing `hash` to the `createMessage` function is correct since it is used to create message.
But the problem is that in the chatroom contract, we should check if the hash is same.

Will remove TODO and update comment.

Silver-IT (commented):
@taoeffect, it's ready for your final review.

frontend/model/contracts/chatroom.js
```diff
@@ -310,13 +310,17 @@ sbp('chelonia/defineContract', {
...
           return
         }
-        const pendingMsg = state.messages.find(msg => msg.id === hash && msg.pending)
+        // NOTE: id(GIMessage.id()) should be used as identifier for GIMessages, but not hash(GIMessage.hash())
+        //       https://github.com/okTurtles/group-income/issues/1503
+        const pendingMsg = state.messages.find(msg => msg.giMsgID === id && msg.pending)
         if (pendingMsg) {
           delete pendingMsg.pending
+          delete pendingMsg.giMsgID
+          pendingMsg.id = hash // NOTE: hash could be different from pendingMsg.id
```
taoeffect: This isn't making sense. Why are you setting the `.id` to `hash`? The `id` should always be the same.

Silver-IT: As we discussed in [Slack](https://okturtles.slack.com/archives/C0EH7P20Y/p1680045474046089), we should have both `id` and `hash` in message object.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -292,15 +292,13 @@ export default ({
...
-              // After this issue is fixed, we can fix this TODO and identity messages using message.id() instead of hash()
               ...createMessage({ meta, data, hash: message.hash() }),
-              // TODO: pending is useful to turn the message gray meaning failed (just like Slack)
+              // NOTE: pending is useful to turn the message gray meaning failed (just like Slack)
               // when we don't get event after a certain period
-              pending: true
+              pending: true, // NOTE: do not RENAME this as it's being used inside the contract
+              // message.id() should be used as an identifier of GIMessage
+              // https://github.com/okTurtles/group-income/issues/1503
+              giMsgID: message.id() // NOTE: do not RENAME this as it's being used inside the contract
```
taoeffect: I think this can be deleted once `id` is passed in to `createMessage` :)

Silver-IT: Correct.

taoeffect (approved):
Approved! Excellent work @Silver-IT!

PR #1516, opened 2023-03-17
WIP - Re-add SQLite support

### LINKED ISSUE #1484: Return of Sqlite
### Problem

The way we're storing data right now - using the filesystem - is cool. But it might also be [35% slower than using sqlite](https://www.sqlite.org/fasterthanfs.html), and it might also use more disk space.

We will likely need to add sqlite back anyway for the backend dashboard stuff.

### Solution

Add back Sqlite. See some of the [old code](https://github.com/okTurtles/group-income/blob/master/historical/backend/objection.js) that we had. But this time don't simply copy that code - instead create an SBP API that is capable of choosing either an SQLite database or the filesystem for doing key/value lookups.

Of course, we will need a specific and more detailed API for backend dashboad stats, but that can come later as a separate issue.

Edit: note that using sqlite can also help with certain custom server setups that, for example, use networked versions of sqlite such as [sqld](https://github.com/libsql/sqld) or [rqlite](https://github.com/rqlite/rqlite), and [mvsqlite](https://github.com/losfair/mvsqlite)

snowteamer:
Closes #1484

### Summary of changes
- When set, `GI_PERSIST` can now be either `fs` or `sqlite`.
- In production,  persistence is always enabled and defaults to `fs`  if `GI_PERSIST` is not set or empty.
- Storage unit tests have been added in `test/database.test.js`. These tests also apply to the `fs` backend.
- The `backend/database.js` module must now be initialized before use.

### Additional info
- Blob storage is not affected by these changes, i.e. file uploads are still plain files in the `./data` folder, no matter the persistence mode.


backend/database.js
```diff
@@ -184,34 +230,52 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
+          reject(err)
+        } else {
+          resolve()
+        }
+      })
+    })
+  })()
+}
+
 if (production || process.env.GI_PERSIST) {
```
Silver-IT: Instead of `process.env.GI_PERSIST`, we can use `persistence`.

backend/database.js
```diff
@@ -172,6 +173,51 @@ export default (sbp('sbp/selectors/register', {
...
+            reject(err)
+          } else {
+            resolve(row?.value)
+          }
+        })
+      })
+    },
+    'backend/db/writeString': function (key, value) {
+      return new Promise((resolve, reject) => {
+        db.run('REPLACE INTO Strings(key, value) VALUES(?, ?)', [key, value], (err) => {
```
Silver-IT: Why do you think `REPLACE` should be used here instead of `INSERT`?
Do you think it would be necessary to update string?
Couldn't be there any security problem?

Silver-IT (changes requested):
Nice work, @snowteamer. Nice PR actually.
Additionally, I think there should be `.env.example` file since we use many environment variables.
We can create an issue for this and do in another PR.

backend/database.js
snowteamer: Wouldn't `INSERT` fail if an entry with the given key already exists?

Silver-IT: When would it be happening to add entry with the already existing key? I don't think there will be.

snowteamer: For example, when updating the current HEAD for a given contract ID key. I think this can happen many times for any contract.

backend/database.js
```diff
@@ -172,6 +173,51 @@ export default (sbp('sbp/selectors/register', {
...
+} else if (persistence === 'sqlite') {
+  sbp('sbp/selectors/register', {
+    'backend/db/readString': function (key: string): Promise<string> {
+      return new Promise((resolve, reject) => {
+        db.get('SELECT value FROM Strings WHERE key = ?', [key], (err, row) => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve(row?.value)
+          }
```
taoeffect: What happens if the key is not found? Is an error thrown or does it resolve as `null`?

I ask because the previous behavior needs to be preserved, as selectors like `'backend/db/lookupName'` depend on it.

taoeffect: OK, answered my own question, it returns `undefined`. This will be important later on, see my other comment.

backend/database.js
Silver-IT: I am not sure why I missed this. Hahaha. Totally agreed.

backend/database.js
```diff
@@ -184,34 +230,52 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
         return lookupValue
       }
-      const bufferOrError = await sbp('backend/db/readFile', filename)
-      if (Boom.isBoom(bufferOrError)) {
-        return null
-      }
-      const value = bufferOrError.toString('utf8')
-      cache.set(filename, value)
+      const value = await sbp('backend/db/readString', key)
+      cache.set(key, value)
```
taoeffect: This API must be designed to behave in the *exact same way* regardless of which backend database type is selected (including in-memory database).

Right now the behavior is inconsistent between `fs` and `sqlite`, and possibly in-memory.

If a value is not found, `fs` will return `null`, whereas `sqlite` will return `undefined`, and I think in-memory will also return `undefined`.

We need to make sure that they all return the same thing, so that the calling APIs, if their behavior depends on the return value in any way, remains consistent when the backend database is switched. I recommend having `fs` also return `undefined`.

Another thing to note, is that previously, the cache would not be set when the fs backend failed to find the file. Now it is set. This is OK, but just want to make sure that that is a conscious decision, because it does represent a minor change in behavior (that I think is fine).

taoeffect (changes requested):
Excellent work @snowteamer! One very minor (but important) technical detail (actually two, I second @Silver-IT's suggestion to replace `process.env.GI_PERSIST` with `persistence` in the `if` clause.

Silver-IT (dismissed):
Great work, @snowteamer.

shared/domains/chelonia/db.js
```diff
@@ -15,21 +15,21 @@ sbp('sbp/selectors/unsafe', ['chelonia/db/get', 'chelonia/db/set', 'chelonia/db/
...
   ? {
-      'chelonia/db/get': function (key): Promise<*> {
+      'chelonia/db/get': function (key: string): Promise<string | void> {
         const id = sbp('chelonia/db/contractIdFromLogHEAD', key)
-        return Promise.resolve(id ? sbp(this.config.stateSelector).contracts[id]?.HEAD : null)
+        return Promise.resolve(id ? sbp(this.config.stateSelector).contracts[id]?.HEAD : undefined)
       },
-      'chelonia/db/set': function (key, value): Promise<void> { return Promise.resolve(value) },
-      'chelonia/db/delete': function (): Promise<void> { return Promise.resolve() }
+      'chelonia/db/set': function (key: string, value: string): Promise<string> { return Promise.resolve(value) },
```
taoeffect: Why does this return signature return a `Promise<string>`, while the overwritten `'chelonia/db/set'` return `Promise<void>`? They should be consistent...

(Same comment for other lines, e.g. on line 29 below)

backend/database.js
```diff
@@ -223,53 +205,42 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
-          resolve()
-        }
-      })
+export default async () => {
+  // If persistence must be enabled:
+  // - load and initialize the selected storage backend
+  // - register `readString` and `writeString` selectors
+  // - overwrite 'chelonia/db/get' and '-set' to use an LRU cache
+  if (persistence) {
+    const { initStorage, readString, writeString } = await import(`./${persistence}-backend.js`)
```
taoeffect, line 214: Beautiful

test/backend.test.js
```diff
@@ -236,7 +239,9 @@ describe('Full walkthrough', function () {
     it('Should create mailboxes for Alice and Bob and subscribe', async function () {
       // Object.values(users).forEach(async user => await createMailboxFor(user))
       await createMailboxFor(users.alice)
+      console.log('alice ok')
       await createMailboxFor(users.bob)
+      console.log('bob ok')
```
taoeffect, line 244: I don't think these console.logs should be here?

taoeffect (changes requested):
Really beautiful reorganization of how the database loads its backend.

I did notice some minor issues related to fs module:

- The old code for the fs stuff is still under `// Filesystem API` comment in main file. It doesn't seem used anymore and should be removed.
- The new code for the fs stuff should be using the `throwIfFileOutsideDataDir` function just as the old code did (that function should be moved into `fs-backend.js`

test/backend.test.js
snowteamer: Good catch! :sweat_smile:

backend/database.js
snowteamer: Thank you! :slightly_smiling_face:

PR #1517, opened 2023-03-20
Set default focus and submit actions in modals

### LINKED ISSUE #1515: Set default focus and default submit actions
### Problem

1. Some modals don't set the focus by default.
E.g. `Create a channel` modal has the `Name` input field and it should be focused by default.
2. When we press `Enter` in some input fields, it makes error.
E.g. `Search` input field in `New Direct Message` modal.

### Solution

1. Check all the modals and their default focus.
2. Set the default focus. The reference modal is `Create a new group` modal.
3. And check the submit action which would be running when we press `Enter` button.

Silver-IT:
### Summary of Changes

- Closes #1515 
- Fix errors handling `undefined` displayName
- Fix DRY issue to filter users
- Skip unnecessary call to change profile attributes


frontend/utils/filters.js
```diff
@@ -0,0 +1,34 @@
+/**
```
taoeffect: There is already a filters file in `frontend/views/utils/filters.js`, please move this there

taoeffect (changes requested):
This is a great idea for a PR @Silver-IT! Nice work also on DRYing the filter!

I did a quick run through the UI to see how many modals I could find that needed focus, and I found a few, plus also I think that in some circumstances the input field (SendArea) for the Chat should be focused too when we first enter into a new chatroom or into a DM with someone. So I took screenshots of all of these scenarios and include them here for reference:

Screen Shot 2023-03-30 at 2 26 14 PM
Screen Shot 2023-03-30 at 2 24 47 PM
Screen Shot 2023-03-30 at 2 24 24 PM
Screen Shot 2023-03-30 at 2 23 29 PM
Screen Shot 2023-03-30 at 2 23 14 PM
Screen Shot 2023-03-30 at 2 23 07 PM
Screen Shot 2023-03-30 at 2 22 36 PM

Also added a comment about the filters file already existing in a different part of the project.

frontend/utils/filters.js
Silver-IT: Okay.

taoeffect (approved):
Nicely done @Silver-IT! Approved!

PR #1521, opened 2023-03-29
E2E protocol

### LINKED ISSUE #92: Key management
EDIT: we're not longer using Mileposts. So ignore most of what's here. But this issue is still useful as "the tracking issue" key management stuff.

The primary branch for this [milepost](https://github.com/taoeffect/mileposts) is located at: **[taoeffect/group-income-simple{identity}](https://github.com/taoeffect/group-income-simple/tree/identity).**

#### Description

This milepost concerns itself with design of Group Income's authentication/login mechanism.

- Instead of using cookies/sessions we're doing an identity-contract system.
- We need to manage keys securely to do this.

#### Files

Potentially relevant files/directories include:

- `frontend/simple/js/backend/*`
- `test/*`
- `shared/*`
- `backend/auth.js` (and related)

#### Issues

See the issues associated with the milestone.

EDIT: random links follow:

- https://thehftguy.com/2020/02/18/jwt-is-awesome-heres-why/

### LINKED ISSUE #112: Implement Keychain interface
_Part of #92 MP._
## Problem

There are a bunch of keys that need to be managed. 
## Solution

Figure out and fill out a generic API for it. This needs to be done in a sufficiently generic fashion to support transition from GIS to GI.

EDIT: quoting comment on #92:

> So for this we've decided that the beginning stuff (the planning on what exactly to do and how to design this thing), will be a joint task between myself and @vijayee (e.g. come to consensus on plan forward), because it's so important to get this right from the start.

Research:

- https://gun.eco/docs/SEA

Libraries:

- [nacl-blob](https://github.com/bcomnes/nacl-blob)
- [nacl-stream](https://github.com/dchest/nacl-stream-js)

### LINKED ISSUE #839: Make group activity history private
## Problem

When visiting the `/join` page to join a group through an invitation link, the group activity history is public. In this example, we can see the group already invited 2 members and proposed to remove one of them.
 
image

### Solution

Keep the group activity and any other sensitive information private from people outside of the group.

### LINKED ISSUE #852: User accounts and real login

### LINKED ISSUE #1765: Clean up and improve code related to calling 'gi.actions/identity/login'
### Problem

In a recent commits 0a1cd00f3d47b82650ce7ecf50af2c7dda12ccb6 and 8e1701e4698bf88531cc83836b751f58e3cbda4b to `e2e-protocol`, @corrideat had to modify and complicate the way that `LoginForm.vue`'s `login` method worked in order to make sure that it waited for `'gi.actions/identity/login'` to finish loading all of the contracts.

There are 4 files in the reasons behind this:

1. `frontend/main.js` — this file calls `'gi.actions/identity/login'`, but in order to prevent a blank white screen from appearing while contracts are loading, `'gi.actions/identity/login'` was modified so that it wouldn't await on a promise that waits for all contracts to finish loading (see comment above call to `sbp('chelonia/contract/sync', contractIDs)` in `actions/identity.js`)
2. `frontend/controller/actions/identity.js` — this is where `'gi.actions/identity/login'` is defined.
3. `Join.vue` — this file is what loads `LoginForm.vue`, and it needs the login to finish loading all of the contracts so that it can get the correct and complete `userState` before calling `'gi.actions/group/joinAndSwitch'`
4. `LoginForm.vue` — This is the file that @corrideat had to complicate in order to re-introduce waiting-for-contracts-to-sync behavior to `'gi.actions/identity/login'`. We essentially want to undo those chains and have it call `await sbp('gi.actions/identity/login', ..)` as it did before without all the complicated event-listener stuff.

### Solution

There are two possible solutions. Both of them will involve restoring `LoginForm.vue` to the way that it was, and making sure to `await` on the call to `'gi.actions/identity/login'`.

#### Solution 1 (Hard)

There might be some way to modify `'gi.actions/identity/login'` so that it awaits on the call to `sbp('chelonia/contract/sync', contractIDs)`, and then inside of `main.js`, some way to call `'gi.actions/identity/login'` within the `mounted` method (but not awaiting on it), so that the UI does not freeze.

However, this might be difficult to achieve without breaking tests, as the comment in `main.js` suggests:

```js
    // NOTE: important to do this before setting up Vue.js because a lot of that relies
    //       on the router stuff which has guards that expect the contracts to be loaded
    const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
    try {
      if (identityContractID) {
        sbp('okTurtles.events/on', CONTRACT_IS_SYNCING, initialSyncFn)
        // TODO: 'gi.actions/identity/login' will return before the contract
        // state is synced
        // This is to continue with the following block that is setting up
        // Vue.js, but as a result the login events listened for below may
        // have already fired. Instead, this block should be moved below the
        // Vue.js set up, and the router be updated to handle not having a login
        // state initially.
        await sbp('gi.actions/identity/login', { identityContractID })
      }
```

#### Solution 2 (Easy)

The second solution is simply to pass in a boolean option to `'gi.actions/identity/login'` called `dontWaitForSync`.

The call to `sbp('chelonia/contract/sync', contractIDs)` is modified like so:

```js
const p = sbp('chelonia/contract/sync', contractIDs)...
```

And then after the catch block, just before returning from the function, we await on `p` unless `dontWaitForSync` is `true`.

Then `main.js` is modified to pass in `dontWaitForSync: true`, and the changes to `LoginForm.vue` can be reverted.

taoeffect:
- Closes #92 
- Closes #112 
- Closes #1765
- Closes #839
- Closes #852


shared/domains/chelonia/internals.js
```diff
@@ -217,22 +472,66 @@ export default (sbp('sbp/selectors/register', {
...
+
+      verifySignature(signingKey.data, signedPayload, signature.data)
+    }
+
     if (config[`preOp_${opT}`]) {
       processOp = config[`preOp_${opT}`](message, state) !== false && processOp
     }
-    if (processOp && !config.skipProcessing) {
-      opFns[opT](opV)
+    opFns[opT](opV)
```
taoeffect: @corrideat Why was this moved out of the `if`?

shared/domains/chelonia/internals.js
```diff
@@ -283,7 +714,7 @@ export default (sbp('sbp/selectors/register', {
     try {
       preHandleEvent && await preHandleEvent(message)
       // verify we're expecting to hear from this contract
-      if (!state.pending.includes(contractID) && !state.contracts[contractID]) {
+      if (isNaN(1) && !state.pending.includes(contractID) && !state.contracts[contractID]) {
```
taoeffect: Need to not forget about this

shared/domains/chelonia/internals.js
```diff
@@ -447,11 +890,11 @@ function loadScript (file, source, hash) {
     // script.type = 'application/javascript'
     script.type = 'module'
     // problem with this is that scripts will step on each other's feet
-    script.innerHTML = source
+    script.text = source
     // NOTE: this will work if the file route adds .header('Content-Type', 'application/javascript')
     // script.src = `${this.config.connectionURL}/file/${hash}`
     // this results in: "SyntaxError: import declarations may only appear at top level of a module"
-    // script.innerHTML = `(function () {
+    // script.text = `(function () {
```
taoeffect, line 1770: Undo?

frontend/controller/actions/group.js
```diff
@@ -56,6 +48,33 @@ async function saveLoginState (action: string, contractID: string) {
...
+              chatRoomID: generalChatRoomId
+            }
+          })
+        })
+      }
+    }
+  }
+})
+*/
+
```
taoeffect: Can be removed?

shared/domains/chelonia/utils.js
```diff
@@ -0,0 +1,151 @@
...
+        this.setPostSyncOp(contractID, 'pending-keys-for-' + keyRequestContractID, ['okTurtles.events/emit', CONTRACT_IS_PENDING_KEY_REQUESTS, { contractID: keyRequestContractID }])
+      }
+    }
+
+    // Is this a foreign key in this contract? If so, flag it in the
+    // other contract to mirror operations
+    if (key.foreignKey) {
+      const fkUrl = new URL(key.foreignKey)
+      const foreignContract = fkUrl.pathname
+      const foreignKeyName = fkUrl.searchParams.get('keyName')
```
taoeffect, line 389: As a user-provided parameter, can this be abused in any way?

corrideat: I'm not sure that this is 'user-provided' any more than the rest of the parameters, but I guess it would result in perhaps syncing the wrong key at worst.

shared/domains/chelonia/GIMessage.js
```diff
@@ -102,24 +159,33 @@ export class GIMessage {
...
     }
     return this._decrypted
   }
 
+  head (): Object { return this._head }
+
   message (): Object { return this._message }
 
-  op (): GIOp { return this.message().op }
+  op (): GIOp { return this._decrypted ? [this.head().op, this.message(), this.decryptedValue()] : [this.head().op, this.message()] }
```
taoeffect: Why is this `this.decryptedValue()` being included in the op?

Gruntfile.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
 
+// TODO: REMOVEME. This prevents tests from running
+if (process.env['CI']) process.exit(1)
+
```
taoeffect: This can now be removed

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -160,6 +227,8 @@ export default ({
   },
   data () {
     return {
+      GIMessage,
+      JSON,
```
taoeffect, line 231: Why are these here?

EDIT: will be cleared out in a separate PR

corrideat (approved):

PR #1527, opened 2023-04-05
Re-add SQLite support

### LINKED ISSUE #1484: Return of Sqlite

snowteamer:
Closes #1484

### Summary of changes
- Updated Node version to v18 LTS (was necessary to get `FormData` and `Blob` support in backend tests without pulling in new dependencies).
- User file uploads are now stored as blobs in Chelonia DB using `blob:${hash}` keys, whereas previously they were always stored in the `./data` folder. This means that file uploads will no longer be persisted unless persistence is actually enabled.
- When set, `GI_PERSIST` can now be either `fs` or `sqlite`.
- In production,  persistence is always enabled and defaults to `fs`  if `GI_PERSIST` is not set or empty.
- A `name:` prefix is now used in namespace keys instead of `name=`.
- A `head:` prefix is now used in contract log head keys instead of a `-HEAD` suffix.
- The `'backend/db/readFile'`, `'-writeFile'`, and `'-writeFileOnce'` selectors have been removed.
- The `backend/database.js` module must now be initialized before use.
- Storage unit tests have been added in `test/database.test.js`. These tests apply to both the `fs` and `sqlite` backends in a single run, so that you do not have to do multiple test runs with different `GI_PERSIST` values.


backend/database.js
```diff
@@ -148,34 +162,41 @@ export default (sbp('sbp/selectors/register', {
...
+    return await readFile(filepath)
   },
-  'backend/db/writeFile': async function (filename: string, data: any): Promise<*> {
+  'backend/db/writeFile': async function (filename: string, data: any): Promise<void> {
     // TODO: check for how much space we have, and have a server setting
     //       that determines how much of the disk space we're allowed to
     //       use. If the size of the file would cause us to exceed this
     //       amount, throw an exception
-    return await writeFileAsync(throwIfFileOutsideDataDir(filename), data)
+    return await writeFile(throwIfFileOutsideDataDir(filename), data)
```
taoeffect: Can you get rid of these?

backend/database.js
```diff
@@ -148,34 +162,41 @@ export default (sbp('sbp/selectors/register', {
...
   },
-  'backend/db/writeFileOnce': async function (filename: string, data: any): Promise<*> {
+  'backend/db/writeFileOnce': async function (filename: string, data: any): Promise<void> {
```
taoeffect: Could you get rid of this and consolidate the `/file` API in `routes.js` to use `'chelonia/db/get` and `'chelonia/db/set'` instead?

backend/database.js
```diff
@@ -184,39 +205,42 @@ function throwIfFileOutsideDataDir (filename: string): string {
...
-        return result
-      } catch (err) {
-        throw err
+export default async () => {
+  // If persistence must be enabled:
+  // - load and initialize the selected storage backend
+  // - register `readString` and `writeString` selectors
+  // - overwrite 'chelonia/db/get' and '-set' to use an LRU cache
+  if (persistence) {
+    const { initStorage, readString, writeString } = await import(`./${persistence}-backend.js`)
```
taoeffect: One request: could you please rename the files to have this format? `database-{persistence}.js`?

This will keep them sorted next to each other nicely in file viewers and code editors.

backend/fs-backend.js
```diff
@@ -0,0 +1,27 @@
...
+
+export async function initStorage (options: Object = {}): Promise<void> {
+  dataFolder = path.resolve(options.dirname)
+  await mkdir(dataFolder, { mode: 0o750, recursive: true })
+}
+
+// eslint-disable-next-line require-await
+export async function readString (key: string): Promise<string | void> {
+  checkKey(key)
+  return readFile(`${dataFolder}/${key}`)
```
taoeffect: Please use the system-independent `path.join`.

I'm also not 100% sure this is safe, even with `checkKey`, because I'm not sure if specially encoded entities could encode `../` and somehow escape. Might be a question for @corrideat? This is why I used `path.resolve` in `throwIfFileOutsideDataDir`, as with `path.resolve` you can be 100% sure it's safe.

taoeffect (changes requested):
Great work @snowteamer! Some more minor issues and after these are addressed I think it should be good to go.

backend/fs-backend.js
corrideat: 1. `throwIfFileOutsideDataDir` seems like it would work. 
2. Re. system-independent `path.join`: `/` or `\\` should work as path separators in any system
3. specially encoded entities could encode `../`: I didn't test, but I'm guessing not, as Node.js should decode the URI. In any case, one could check the `key` doesn't contain any `%` (actually, it'd be better to have a whitelist of characters, like `/^[a-zA-Z0-9=+_-]$/`).

taoeffect: I think we do want to allow support for foreign characters though, so I'm not sure about the whitelist. Sounds like the above code should work then...

backend/database.js
snowteamer: `'chelonia/db/set'` currently only support strings. Should it be extended for blob support?
Or should we define `'chelonia/db/getBlob'` and `'-setBlob'` similarly to `'chelonia/db/getEntry'` and `'-addEntry'`?

backend/fs-backend.js
snowteamer: Node `fs` methods don't seem to interpret strings as URIs:
image

image

taoeffect: OK, it's probably fine. If you're sure Windows is happy with `/`... we can stick with this

backend/database.js
taoeffect: Let's figure this out on Slack, I'll reply there

Gruntfile.js
```diff
@@ -365,7 +365,7 @@ module.exports = (grunt) => {
       // - anything that ends with `.test.js`, e.g. unit tests for SBP domains kept in the domain folder.
       // The `--require` flag ensures custom Babel support in our test files.
       test: {
-        cmd: 'node --experimental-fetch node_modules/mocha/bin/mocha --require ./scripts/mocha-helper.js --exit -R spec --bail "./{test/,!(node_modules|ignored|dist|historical|test)/**/}*.test.js"',
```
snowteamer, line 368: No longer required as of Node 18

backend/routes.js
```diff
@@ -162,30 +162,50 @@ route.POST('/file', {
       console.error('failAction error:', err)
       return err
     },
-    maxBytes: 6 * MEGABTYE, // TODO: make this a configurable setting
+    maxBytes: 6 * MEGABYTE, // TODO: make this a configurable setting
```
taoeffect, line 165: Hah! Nice typo catch!

shared/domains/chelonia/db.js
```diff
@@ -15,34 +27,43 @@ sbp('sbp/selectors/unsafe', ['chelonia/db/get', 'chelonia/db/set', 'chelonia/db/
...
 
 export default (sbp('sbp/selectors/register', {
   ...dbPrimitiveSelectors,
   'chelonia/db/logHEAD': function (contractID: string): string {
-    return `${contractID}${headSuffix}`
+    return `${headPrefix}${contractID}`
   },
   'chelonia/db/contractIdFromLogHEAD': function (key: string): ?string {
-    return key.endsWith(headSuffix) ? key.slice(0, -headSuffix.length) : null
+    return key.startsWith(headPrefix) ? key.slice(headPrefix.length) : null
```
taoeffect, line 64: @corrideat remember you were wondering why I had these as functions? 😄

backend/routes.js
```diff
@@ -162,30 +162,50 @@ route.POST('/file', {
...
+    return Boom.notFound()
+  }
+  return h.response(blob).etag(hash)
+})
+
+// Used to serve contract source files and contract manifests.
+// These files are stored in the file system rather than the Chelonia db,
+// therefore `GET /file` could not be used since a different handler was needed.
+// This route is very similar to `GET /assets`, but for now a separate route must be used.
+route.GET('/contract/{hash}', {
```
taoeffect: I don't understand what's going on here and the comment isn't enough to help me understand.

> // These files are stored in the file system rather than the Chelonia db,

This doesn't seem to be true?  `POST /file` uses:

```js
await sbp('chelonia/db/set', `blob:${hash}`, data)
```

That is the Chelonia DB, which can be either the filesystem, memory, or sqlite.

The source files should be stored wherever `POST /file` stores them, and that should be equivalent to the above 3 possible backends. Am I missing something?

snowteamer: These files are only stored at build time using `chel deploy` or `deployAll`, and these commands will put them in the file system. Yea `POST /file` uses Chelonia DB, `GET /file `does too, that's why a new route endpoint was needed, one which uses Hapi's `h.file()` instead.

taoeffect: OOHHHHHH!!!

OK, that's an issue... ideally `chel deploy`/`chel deployAll` would deploy it to the database, whatever that happens to be. In this case we're mixing and matching databases (filesystem and sqlite), and that's the problem. `chel deploy` should be instructed which one to use, then this wouldn't be a problem.

We could use `/contract` temporarily, but only as a hack until we fix `chel deploy` to take an argument to accept which type of database we're deploying to

backend/database-sqlite.js
```diff
@@ -0,0 +1,65 @@
...
+  await new Promise((resolve, reject) => {
+    if (db) {
+      reject(new Error(`The ${filename} SQLite database is already open.`))
+    }
+    db = new sqlite3.Database(join(dataFolder, filename), (err) => {
+      if (err) {
+        reject(err)
+      }
+    })
+    db.run('CREATE TABLE IF NOT EXISTS Data(key TEXT UNIQUE NOT NULL, value TEXT NOT NULL)', [], (err) => {
```
taoeffect: What was the verdict on adding a third column for metadata for dealing with `lastModified` header for fs?

taoeffect: Also, `key` should be marked as a `PRIMARY KEY` instead of `UNIQUE`. This might also speed up performance as primary keys should be automatically indexed in sqlite. At least according to GPT4 (and my memory).

snowteamer: That turned out to be a bit complex because `lastModified` has to be stored somewhere and kept up-to-date for all three storage backends, and also because we want a consistent API regardless the data type but only want that field for files.

backend/database-sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Useful in test hooks.
+export function clear (): Promise<void> {
+  return run('DELETE FROM Data')
+}
+
+export function exportToJSON (): Promise<Object> {
+  return new Promise((resolve, reject) => {
+    db.all('SELECT key, value FROM Data', [], (err, rows) => {
```
taoeffect, line 56: This would read the entire database contents into memory, right? That seems like an extreme amount of data to read into memory (potentially hundreds of gigabytes worth). What is this function for?

taoeffect: Copying a note from @snowteamer from Slack about what this is:

> These are database migration helpers, which I was planning to maybe use either in the preloading code or  in chel migrate.
 I've left them here so that they can be unit-tested in database.test.js and reviewed, but yeah actually they are not used (at least yet).

backend/database-sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function readData (key: string): Promise<Buffer | string | void> {
+  return new Promise((resolve, reject) => {
+    checkKey(key)
+    readStatement.get([key], (err, row) => {
+      if (err) {
+        reject(err)
+      } else {
+        resolve(row?.value)
```
taoeffect, line 91: Why is there no corresponding `key.startsWith('blob=')` check here?

backend/database.js
```diff
@@ -139,84 +153,84 @@ export default (sbp('sbp/selectors/register', {
...
-      return
-    }
-    return await writeFileAsync(filepath, data)
   }
-}): any)
+})
+
+// Used to thwart path traversal attacks.
+export function checkKey (key: string): void {
+  if (/[/\\]/.test(key)) {
```
taoeffect: One of the genuinely good ideas from that [windows filesystem link](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) you mentioned is:

> Characters whose integer representations are in the range from 1 through 31,

See the unprintable ASCII characters here: https://www.rapidtables.com/code/text/ascii-table.html

I asked the bots for implementations and here are some:

```js
function hasUnprintableChars(string) {
    for (let i = 0; i < string.length; i++) {
        const charCode = string.charCodeAt(i);
        if (charCode < 32 || charCode === 127) {
            return false;
        }
    }
    return true;
}
```

```js
function containsUnprintableChars(str) {
  // Regular expression that matches any characters other than printable ASCII characters
  const unprintableCharsRegExp = /[^\x20-\x7E]/;

  return !unprintableCharsRegExp.test(str);
}
```

Just some ideas. The hard part is picking one that's fast and also accurate.

backend/database-sqlite.js
snowteamer: Not all keys have to start with `=blob`. Besides, I think these checks are only needed when writing to the DB (e.g. in `'chelonia/db/set'`, not when reading from it since the database is assumed to cnotain valid key-value pairs.

taoeffect (changes requested):
Nice work @snowteamer! I left a few comments.

Also, I tried testing this with `GI_PERSIST=sqlite grunt dev`, and ran into two problems:

1. I couldn't load the site, neither on port `:3000` nor `:8000`
2. I didn't see the message `[chelonia.db] Preloaded ${newKeys.length} new entries` in the console either

This is (some of) the output I got:

```
Running "backend:relaunch" task
backend: forking...

Running "keepalive" task
[Browsersync] Proxying: http://127.0.0.1:8000
[Browsersync] Access URLs:
 ----------------------------------------
       Local: http://localhost:3000
    External: http://192.168.100.201:3000
 ----------------------------------------
          UI: http://localhost:3002
 UI External: http://localhost:3002
 ----------------------------------------
[Browsersync] Watching files...
[SBP WARN]: registering unsafe selector: 'chelonia/db/get' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia/db/set' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia/db/delete' (remember to lock after overwriting)
Connected to the groupincome.db SQLite database.
[SBP WARN]: registering unsafe selector: 'chelonia/db/get' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia/db/set' (remember to lock after overwriting)
```

backend/database.js
snowteamer: Initially `checkKey` was only there to prevent path attacks. But if keys have to be valid Windows file names, then besides valid characters we'll also have to check for max length, reserved names, no space at the end, etc. If necessary I already have such a function lying around

backend/database-sqlite.js
taoeffect: >  Besides, I think these checks are only needed when writing to the DB (e.g. in 'chelonia/db/set', not when reading from it since the database is assumed to cnotain valid key-value pairs

But it is done in `readData` in `database-fs.js`?

backend/database.js
taoeffect: @snowteamer I think we can have a rule that is similar to but not quite the same as the Windows filesystem. I see no reason why we should ban keys with `:` in them, for example. We don't care about the reserved names like `COM1` etc. But some things really don't belong in key names, and unprintable characters are one of them.

backend/database-sqlite.js
snowteamer: Do you mean this line?
```js
    .then(buffer => key.startsWith('blob=') ? buffer : buffer.toString('utf8'))
 ```
 It's not a data validation check, just a test to know whether the value must be decoded to UTF-8 before returning

taoeffect: Yeah, so I'm wondering why this `resolve(row?.value)` doesn't do a similar decoding

snowteamer: That's because sqlite is able to remember data types, in order to return correctly deserialized values when queried. That's actually quite convenient: if a Node buffer was stored, `row.value` will be a Node buffer, but if a string was passed in, then `row.value` will already be a string. 
The same cannot be said for `fs` functions, Node's `fs.readFile()` won't know by itself whether the data should be decoded, the caller has to know it and specify the right encoding (or none if it wants a buffer)

taoeffect (approved):
Approved! Great work @snowteamer!!

PR #1528, opened 2023-04-05
Stop Cypress upon first failed spec

### LINKED ISSUE #1520: Cypress tests on Travis should stop immediately upon first failure
### Problem

Travis uses up our credits, and if we are running the entire test suite even though there was a failure, that's a lot of credits we're wasting. We could be saving a lot of Travis credits if we had the Cypress tests stop after the first failure.

### Solution

In the Travis environment only, have Cypress stop immediately after the first failure.

snowteamer:
Closes #1520

### DIFF
--- a/test/cypress/plugins/index.js
+++ b/test/cypress/plugins/index.js
@@ -22,8 +22,24 @@
 //   on('file:preprocessor', browserify(options))
 // }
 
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Fantastic @snowteamer !

PR #1529, opened 2023-04-06
Added comments + Minor refactoring

taoeffect:
Also:

- Renamed `update` function to `updateContractSalt`
- Renamed some ZKPP related API calls in `routes.js`

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -1166,12 +1166,19 @@ sbp('chelonia/defineContract', {
         if (rootState[data.chatRoomID]?._volatile) {
           return
         }
-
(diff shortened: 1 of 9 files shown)


frontend/model/contracts/group.js
```diff
@@ -1166,12 +1166,19 @@ sbp('chelonia/defineContract', {
         if (rootState[data.chatRoomID]?._volatile) {
           return
         }
-
+        // the specific scenario here is joining chatrooms (should be the general channel in this
```
corrideat, line 1169: The comments here are fine. It'll cause a merge conflict with my changes as this is removed, but I'm not sure that's avoidable.

corrideat (approved):

PR #1533, opened 2023-04-11
Wrong due date shown in UI

### LINKED ISSUE #1450: Wrong due date shown in UI
### Problem

The screenshot below illustrates the problem. It shows how the UI will tell the user that the due-date is "October 31st", but when November 1st arrives, the new TODOs are not visible in the Payments screen, and it also explains why.

Note: in the screenshot below, `g` refers to the global Vuex getters, and `date` refers to the current date when I took the screenshot (in this case, `"2022-11-01T22:11:40.457Z"`):

Screen Shot 2022-11-01 at 3 22 22 PM

The problem is that the due date is saved as `"2022-11-01T00:00:00.000Z"` (which signifies the very end of October 31st / the very beginning of November 1st), while the period-stamp is saved as `"2022-11-02T00:00:00.000Z"`, a full day later.

The getter `currentPaymentPeriod` is what determines what TODO list is displayed, and it, in turn, depends on `periodStampGivenDate`.

As such, there's a 24-hour period after the due-date shown to the user during which `periodStampGivenDate` will return the period stamp for the previous period. In other words, we're plugging in `date` to `periodStampGivenDate`, and it's returning the period stamp for the previous period:

```js
getters.periodStampGivenDate("2022-11-01T22:11:40.457Z")

// => returns "2022-10-03T00:00:00.000Z"
```

### Solution

Here are the various dates we're dealing with in this example:

- Current date: `"2022-11-01T22:11:40.457Z"`
- Previous period stamp: `"2022-10-03T00:00:00.000Z"`
- Next period stamp: `"2022-11-02T00:00:00.000Z"`.
- Date returned by `dueDateForPeriod`: `"2022-11-01T00:00:00.000Z"`
- Due date shown to user (in Pacific Time Zone): `October 31st`

Now, technically, all of the values returned by the getters seem correct - but the date shown to the user is clearly wrong.

In this example, the switchover with a fresh set of TODOs for pledgers happens once the current date is `>= "2022-11-02T00:00:00.000Z"`.

I think fixing this should be as simple as making `dueDateForPeriod` return the `periodAfterPeriod` for that period stamp.

FWIW, in that case, `humanDate(g.periodAfterPeriod(g.periodStampGivenDate(date)), { month: 'short', day: 'numeric' })` returns `"Nov 1"` as the due date, which would work. However, if we make this the due date shown to the user, there is some hour of November 1st (today) at which payments for the current period will no longer be accepted, and we currently do not show the hour of the due date to the user (cc @leihla).

Silver-IT:
### Summary of Changes

- Closes #1450


frontend/model/contracts/shared/time.js
```diff
@@ -128,7 +129,10 @@ export function humanDate (
     : ((navigator.languages: any): string[]) ?? navigator.language
   // NOTE: `.toLocaleDateString()` automatically takes local timezone differences into account.
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
-  return new Date(date).toLocaleDateString(locale, options)
+  if (dateOrTime) {
+    return new Date(date).toLocaleDateString(locale, options)
+  }
+  return new Date(date).toLocaleTimeString(locale, options)
```
taoeffect: I'm pretty sure these changes to the `humanDate` function can (& should) be undone because they aren't necessary.

The `toLocaleDateString` supports displaying the current time, you simply need to pass in the options telling it to, e.g. here's from me experimenting in the terminal:

```js
> new Date().toLocaleDateString('en-US', { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' })
'Apr 12, 4:46 PM'
```

frontend/views/containers/payments/PaymentRowTodo.vue
```diff
@@ -86,7 +86,15 @@ export default ({
...
   methods: {
-    humanDate,
+    humanDateTime (dateTime) {
+      return humanDate(dateTime, {
+        month: 'short',
+        day: 'numeric'
+        // hour: '2-digit',
+        // minute: '2-digit',
+        // hour12: false
+      })
```
taoeffect: Could you remove the comment if it's no longer needed and place the options on 1 line?

frontend/views/containers/payments/PaymentNextDistributionPill.vue
```diff
@@ -18,8 +18,13 @@ export default ({
...
       const nextDisDate = this.dueDateForPeriod(this.currentPaymentPeriod)
-
-      return humanDate(nextDisDate, { month: 'short', day: 'numeric' })
+      const localeTimeString = humanDate(nextDisDate, {
+        hour: '2-digit',
+        minute: '2-digit',
+        hour12: false
+      }, false)
+      const localeDateString = humanDate(nextDisDate, { month: 'short', day: 'numeric' })
+      return `${localeTimeString} - ${localeDateString}`
```
taoeffect: It is much simpler to just pass these options without modifying how `humanDate` is implemented:

```js
return humanDate(nextDisDate, { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' })
```

You also do not need to use `hour12`, because that should be handled correctly by `locale` value inside of `humanDate`.

For example:

```
> new Date().toLocaleDateString('en-US', { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' })
'Apr 12, 4:50 PM'
> new Date().toLocaleDateString('ja', { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' })
'4月12日 16:50'
```

taoeffect (changes requested):
Nice work @Silver-IT!

frontend/views/containers/payments/PaymentRowTodo.vue
Silver-IT: Okay.

frontend/model/contracts/shared/time.js
Silver-IT: Okay. Sure.

frontend/views/containers/payments/PaymentNextDistributionPill.vue
Silver-IT: Yeah. I want to do this way. But the problem is that I don't think we can make the same format using `toLocaleDateString`.
That's why I updated the `humanDate` function and make it possibly return only time.
image

taoeffect: Either way, in that case you still do not need to modify `humanDate`, you can just call it twice with different `options`. But in that case, still `hour12` should not be passed, it should always be decided by the locale.

However, I think the way it formats `{ month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' }` right now is perfectly acceptable, and by calling it once with those options, we are ensured that it will always do it in a locale-appropriate way, whereas creating our own string use ` - ` might not be appropriate for some locales

Silver-IT: Make sense. I should make that dateTime string. And split and merge again to get the right format.
Not sure why I didn't think this way. Hahaha

frontend/views/containers/payments/PaymentNextDistributionPill.vue
```diff
@@ -18,8 +18,13 @@ export default ({
...
       const nextDisDate = this.dueDateForPeriod(this.currentPaymentPeriod)
-
-      return humanDate(nextDisDate, { month: 'short', day: 'numeric' })
+      const localeDateTimeStrings = humanDate(nextDisDate, {
+        month: 'short',
+        day: 'numeric',
+        hour: '2-digit',
+        minute: '2-digit'
+      }).split(', ')
+      return `${localeDateTimeStrings[1]} - ${localeDateTimeStrings[0]}`
```
taoeffect: We cannot assume that the string will have a `,` that can be split like this. Please return the `humanDate` directly, without this ` - `

Silver-IT (commented):
Seems ready for the final review.

taoeffect (dismissed):
Looks good to me! Only thing is Travis has failed...

taoeffect (approved):
Great work @Silver-IT!

Issue #1506, opened 2023-03-02
Create staging server


taoeffect:
### Problem

Multiple times now we've had the issue of everything appearing to work during testing and Travis, etc., but once merged into production everything breaks.

This is caused because of the way the contracts need to be upgraded with missing data sometimes (and failed to be upgraded). Or other reasons.

### Solution

Create a staging server URL and update that before merging changes to production.

PR #1534, opened 2023-04-12
New design to create DM and optimize ChatMixin

### LINKED ISSUE #1493: Design needed for adding guests to DM (Direct Message) and GM (Group Message)
### Problem

1. We can only add **ONE** person to DM or GM at a time at the moment. But we should be able to add several people at a time.
2. The workflow is not 100% clear to add people to DM or GM.

    For example, let's see the following situation.
    I (Alex) have a `DM with Greg`, and another `DM with Sebin`. Now, on a `DM with Greg`, I am trying to add Sebin
    **First problem**: There could be two choices; to clone all the actions from the `DM with Greg` to a new chatroom, and the other is to create a new chatroom with no messages. We don't have any design for this part.
    **Second problem**: In the above situation, to create a new chatroom by adding Sebin to DM with Greg is the same to create a new chatroom by adding Greg to DM with Sebin? In other words, is it possible to exist **several** GMs for the same users?

* The following image describes the types of chatrooms visually.
image

### Solution

Part 1. Clarify the workflow
Part 2. Make design
Part 3: Create a new issue to implement this workflow and design

### LINKED ISSUE #1532: Optimize ChatMixin
### Problem

We have `ChatMxin` which is created earlier and is only being used in `GroupChat.vue`.
From time to time, I think it is not well-designed.

For example, `ChatMixin` watches the `currentChatRoomId` and update details and summary according to its change.
But `ChatMain.vue` gets `summary` and `details` props from `ChatMixin` and also watches the `currentChatRoomId` and do something according to its change.

### Solution

Optimize `ChatMixin`.

### LINKED ISSUE #1549: Clear vuex state on logout
### Problem

As @Silver-IT observed, `'gi.actions/identity/logout'` doesn't completely clear the vuex state on logout, and this could be causing issues in some Cypress tests, and just in general is bad hygiene.

### Solution

Delete the `logout` Vuex mutation, and create `'state/vuex/reset'` instead, and call `store.replaceState(initialState)` in there. Then, in `'gi.actions/identity/logout'`, instead of `sbp('state/vuex/commit', 'logout')` call `sbp('state/vuex/reset')`

### LINKED ISSUE #1553: ChatRoom UnreadPosition should not be updated while logging in
### Problem

If users join the chatroom, we don't care in the previous messages if they've read or not and we check the only new messages to decide if they are read or unread.
Meaning that while joining the chatroom, we should update unread message position to the latest message.
But the problem is that `sbp('chelonia/contract/isSyncing', contractID)` is used in `setReadUntilWhileJoining` function to check if it's joining in chatroom contract.
But we should take care that we also sync contracts when we log in. So after log in, the unread position is always set to the latest message. (This is bug)
Additionally, I think we should set the unread message position to the latest when user log in in a new device.

### Solution

1. `Joining` chatroom comes from the sideEffect functions of the group and mailbox contract.
2. ~~Login in a new device means `sbp('gi.db/settings/load', username)` returns `undefined`. We check this in `gi.actions/identity/login`~~.

We can use **vuex** or **okTurtles.data** to save if it's sign-in in a new device or join.

### LINKED ISSUE #1554: Initialize value when start editing non-monetary pledge
### Problem

Steps to reproduce
1. Add 3 non-monetary pledges in `Contributions` page. Let's say those names are `1`, `2`, `3`.
2. Remove `2` by clicking `Edit` button on the second pledge and also `Remove` button.
3. Click the `Edit` button for the pledge `3`, and you can see the input field is filled by `2` instead of `3`.
4. Remove `3` and type `333` and click `Cancel` button.
5. Click `Edit` button again for the pledge `3` and you can see `333` instead of `3`.

### Solution

Try to debug and set the correct value when user click `Edit` button.

Silver-IT:
### Summary of Changes

- Closes #1493 
- Closes #1532 
- Closes #1549 
- Closes #1553 
- Closes #1554


Silver-IT:
@taoeffect, I don't think there is solution to fix the error which you (and also we) can reproduce when you are using another device.

1. It happens because `rootState.chatRoomUnread[contractID]` is not initialized. I updated the contract to initialize it when the sideEffect of `gi.contracts/chatroom` constructor is being called. And also added `Optional Chaining`(?.) in the `gi.contracts/chatroom/editMessage`, `gi.contracts/chatroom/deleteMessage` sideEffects. 😔

2. But the problem is that all of these **can't** be the solution to fix the error. Because these updates can be used only for new events, not for the previous events which are already created with the previous version of chatroom contract.
And also `postUpgradeVerification` **can't** be a solution either. Because this error happens only in the new device and `postUpgradeVerification` won't be called while the first login in new device.
And the error happens when the chatroom contract is being synced in the sideEffect of group contract which would be also created in the previous version. 😔

3. The another solution is to initialize `chatRoomUnread` in the sideEffect of `join` and `addMessage` chatroom contract events by modifying the global state mutation functions, without modifying the contract code. But unfortunately, in the previous version, they are all blocked by `isSyncing` sbp function, and no longer executed while syncing. 😔

Considering all these, I am very, very sorry to tell you that the only solution is to reset the database again.

And I think we should have this kind of test environment in the local. We should take more care working in the contracts, and also even though we can't catch these errors the Cypress should catch in the local test.
Is this something possible? Then, should we create an issue for this?

taoeffect:
> the only solution is to reset the database again.

OK, that actually is OK this time because I messed up on choosing distribution date anyway.

>  Is this something possible? Then, should we create an issue for this?

Currently the issue that we have open related to this is creating a staging server and testing there: https://github.com/okTurtles/group-income/issues/1506

I should also think a bit more about maybe how to deal with this better in the error handling in `internals.js`. I've created an issue for that too and self-assigned: https://github.com/okTurtles/group-income/issues/1544

frontend/utils/filters.js
```diff
@@ -0,0 +1,34 @@
+/**
```
taoeffect: What's this file doing here? Remember from another PR it was already moved to `frontend/views/utils/filters.js`, what happened?

Silver-IT: Should remove it. Correct.

taoeffect: Maybe `master` needs to be merged?

Silver-IT: I was working after merging the previous branch. I manually removed this file because merging master branch won't delete this file, I think.

test/cypress/support/commands.js
```diff
@@ -261,9 +262,6 @@ Cypress.Commands.add('giAcceptGroupInvite', (invitationLink, {
 }) => {
   if (bypassUI) {
     if (!isLoggedIn) {
-      cy.visit('/')
-      cy.url().should('eq', 'http://localhost:8000/app/')
-      cy.getByDT('welcomeHome').should('contain', 'Welcome to Group Income')
```
taoeffect: @Silver-IT why was this removed?

frontend/views/components/UsersSelector.vue
```diff
@@ -0,0 +1,189 @@
...
+
+      if (keyCode === 13 || keyCode === 39) { // Enter
+        this.$emit('submit')
+        return
+      } else if (keyCode === 8 && !this.value && this.usernames.length) { // Backspace
+        this.remove(this.usernames[this.usernames.length - 1])
+      }
+
+      // NOTE: this.$refs.input.innerHTML could contains HTML entities
+      this.value = decode(this.$refs.input.innerHTML)
```
taoeffect: What is going on here? This is very strange...

- why aren't you using `v-model`?
- if you must do it this way for some reason, it should be clearly commented and explained why `v-model` isn't used
- why add a new dependency instead of [using `textContent`](https://medium.com/@stevriss22/differences-between-innertext-textcontent-innerhtml-23268b2ac7b2)?

Silver-IT: > why aren't you using v-model?

We can use `v-model` for only `input`, `select`, `textarea` elements. Reference in [here](https://vuejs.org/api/built-in-directives.html#v-model)
> if you must do it this way for some reason, it should be clearly commented and explained why v-model isn't used

Okay
> why add a new dependency instead of using textContent?

Correct. `textContent` works fine. I used `html-entities` library to handle HTML entities as it's not an input field. To be honest, never knew `textContent` before. Thanks for letting me know.

frontend/views/components/UsersSelector.vue
```diff
@@ -0,0 +1,189 @@
...
+    autofocus: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data () {
+    return {
+      // NOTE: v-model can't be used here since it's only for limited elements; input, select, textarea
+      //       https://vuejs.org/api/built-in-directives.html#v-model
+      value: ''
```
taoeffect, line 66: @Silver-IT why isn't this using an `input` element?

Silver-IT: Having looked the Slack, I can see it is automatically resizing and repositioning.

I don't think we can implement this using `input` or `textarea` elements.

frontend/model/state.js
```diff
@@ -126,51 +125,44 @@ const mutations = {
...
+  setNotificationSettings (state, { chatRoomId, settings }) {
+    if (chatRoomId) {
+      if (!state.notificationSettings[chatRoomId]) {
+        Vue.set(state.notificationSettings, chatRoomId, {})
+      }
+      for (const key in settings) {
+        Vue.set(state.notificationSettings[chatRoomId], key, settings[key])
+      }
+    }
+  },
```
taoeffect: We already have notification settings. They're in the `notificationModule` that's imported in this file, in a separate Vuex module.

If these are different than those, then this has to be renamed so that it doesn't conflict and cause confusion.

Perhaps renaming this to `setChatroomNotificationSettings` would be a good idea

frontend/model/state.js
```diff
@@ -126,51 +125,44 @@ const mutations = {
...
   },
   addChatRoomUnreadMention (state, { chatRoomId, messageHash, createdDate }) {
     const prevUnread = state.chatRoomUnread[chatRoomId]
-    if (!prevUnread) {
+    if (!prevUnread.readUntil) {
       Vue.set(state.chatRoomUnread, chatRoomId, {
-        since: { messageHash, createdDate, deletedDate: null, fromBeginning: true },
-        mentions: [{ messageHash, createdDate }]
+        readUntil: { messageHash, createdDate, deletedDate: null },
+        mentions: [...prevUnread.mentions, { messageHash, createdDate }]
```
taoeffect: Just a note: changes like this cannot be done without corresponding code in `postUpgradeVerification` once we release

frontend/model/contracts/group.js
```diff
@@ -1215,10 +1216,16 @@ sbp('chelonia/defineContract', {
         name: string
       }),
       process ({ data, meta }, { state, getters }) {
-        Vue.set(state.chatRooms, data.chatRoomID, {
-          ...getters.getGroupChatRooms[data.chatRoomID],
-          name: data.name
-        })
+        Vue.set(state.chatRooms[data.chatRoomID], 'name', data.name)
```
taoeffect, line 1219: This looks like a nice simplification 👍

frontend/controller/actions/group.js
```diff
@@ -373,6 +373,33 @@ export default (sbp('sbp/selectors/register', {
...
+      })
+
+      return await sbp('chelonia/out/actionEncrypted', {
+        ...omit(params, ['options', 'action', 'hooks']),
+        action: 'gi.contracts/group/changeChatRoomDescription',
+        hooks: {
+          prepublish: null,
+          postpublish: params.hooks?.postpublish
+        }
+      })
```
taoeffect: Why are there two messages being sent?

Can you refactor this so that only `'gi.actions/chatroom/changeDescription` is called?

frontend/model/contracts/shared/functions.js
```diff
@@ -156,3 +156,30 @@ export function makeMentionFromUsername (username: string): {
...
+
+export async function syncContract (contractID: string, logJoining: boolean = false) {
+  if (logJoining) {
+    setContractJoining(contractID, false)
+  }
+  await sbp('chelonia/contract/sync', contractID)
+  if (logJoining) {
+    setContractJoining(contractID, true)
+  }
+}
```
taoeffect: What are these functions?

Why did you create them?

Can you use `'chelonia/contract/isSyncing'` instead?

taoeffect (changes requested):
Nice work @Silver-IT, left some questions for you

frontend/controller/actions/group.js
Silver-IT: The reason why I created this `gi.actions/group/changeChatRoomDescription` function is to call `gi.contracts/group/changeChatRoomDescription` function so that the group could save the attributes of its chatrooms.
The reason why the group should save the attributes of its chatrooms is that the group members should see the group chatrooms and their details even though they are not part of. It's not good to sync those chatrooms every time they are trying to see the details.

frontend/model/contracts/shared/functions.js
Silver-IT: These functions are needed for #1553.
When a user tries to log in, he tries to sync all the contracts no matter it's the first login in a new device or the contracts were already fully/half synced. We need to separate these two and `'chelonia/contract/isSyncing'` is not enough for this since those two are all syncing.
And also users sync the chatroom contract when they join it and we need to differentiate this too.

I differentiate those by calling the `syncContract` function with different value of parameter `logJoining` and check them by calling `checkContractJoining` function in `model/contracts/chatroom.js`.

I will add comments to these functions.

Silver-IT: By the way, should these functions be here in `'model/contracts/shared/functions.js'`?
Or should we update `'chelonia/contract/sync'` function?

Silver-IT: I have just got an idea to create `'chelonia/contract/isNewSyncing'` instead of these functions.

frontend/model/state.js
Silver-IT: Why do you think so?
`state.chatRoomUnread[contractID]` is initiated in the sideEffect of `'gi.contracts/chatroom'` constructor. And it must be before to call any other vuex mutations for this chatroom unread state.

taoeffect: Because, before the data stored in Vuex was using other names for keys. And now you're changing those keys. So when clients upgrade to the new code, their existing data is stored on the old keys, not the new keys. You need to properly migrate everything if you change how and where things are stored.

frontend/controller/actions/group.js
taoeffect: OK, makes sense.

taoeffect: OK, makes sense. Could you add a comment here explaining that then?

Silver-IT: Okay. Sure.

frontend/model/state.js
Silver-IT: I can understand what you mean, and I don't think it would break because of this change. And if the `readUntil` is not set, users will automatically be redirected to the latest message and then `readUntil` will be set to the latest.


Also, we agreed to reset the database [here](https://github.com/okTurtles/group-income/pull/1534#issuecomment-1513972639) after we merge this PR.

taoeffect (changes requested):
Nice work @Silver-IT! It's great to see so many simplifications and improvements to the chat stuff! 🙌

I'd like to request the following modifications to `chelonia.js` and `internals.js` to replace `isNewSyncing` with a different version of `isSyncing`:

```js
  'chelonia/contract/isSyncing': function (contractID: string, { firstSync = false }): boolean {
    const isSyncing = !!this.currentSyncs[contractID]
    return firstSync
      ? isSyncing && this.currentSyncs[contractID].firstSync
      : isSyncing
  },
```

And then in `internals.js`:

```js
  'chelonia/private/in/syncContract': async function (contractID: string) {
    const latest = await sbp('chelonia/out/latestHash', contractID)
    console.debug(`[chelonia] syncContract: ${contractID} latestHash is: ${latest}`)
    // there is a chance two users are logged in to the same machine and must check their contracts before syncing
    let recent
    if (state.contracts[contractID]) {
      recent = state.contracts[contractID].HEAD
    } else {
      // we're syncing a contract for the first time, make sure to add to pending
      // so that handleEvents knows to expect events from this contract
      if (!state.contracts[contractID] && !state.pending.includes(contractID)) {
        state.pending.push(contractID)
      }
    }
    sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, true)
    this.currentSyncs[contractID] = { firstSync: !state.contracts[contractID] }
    try {
      if (latest !== recent) {
        console.debug(`[chelonia] Synchronizing Contract ${contractID}: our recent was ${recent || 'undefined'} but the latest is ${latest}`)
        // TODO: fetch events from localStorage instead of server if we have them
        const events = await sbp('chelonia/out/eventsSince', contractID, recent || contractID)
        // remove the first element in cases where we are not getting the contract for the first time
        state.contracts[contractID] && events.shift()
        for (let i = 0; i < events.length; i++) {
          await sbp('chelonia/private/in/handleEvent', GIMessage.deserialize(events[i]))
        }
      } else {
        console.debug(`[chelonia] contract ${contractID} was already synchronized`)
      }
    } catch (e) {
      console.error(`[chelonia] syncContract error: ${e.message}`, e)
      this.config.hooks.syncContractError?.(e, contractID)
      throw e
    } finally {
      delete this.currentSyncs[contractID]
      sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, false)
    }
  },
```

Then instead of `sbp('chelonia/contract/isNewSyncing', contractID)`, you'd do: `sbp('chelonia/contract/isSyncing', contractID, { firstSync: true })`

frontend/model/state.js
taoeffect: > Also, we agreed to reset the database here after we merge this PR.

Yes 👍

Silver-IT:
By the way, should we use `delete this.currentSyncs[contractID]` instead of `this.currentSyncs[contractID] = false`?
Since `this.currentSyncs[contractID]` could be two values (`{ firstSync: !state.contracts[contractID] }`, `false`) of two types, one is `Object` and the other is `Boolean`.

taoeffect:
@Silver-IT yes, I updated the code. Also, I just realized that it should happen before `sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, false)`, so I updated that code too.

Silver-IT:
Great. Additionally, we should update the parameter declaration of isSyncing function.
```
'chelonia/contract/isSyncing': function (contractID: string, { firstSync = false } = {}): boolean {
...
}
```

frontend/assets/style/components/_cards.scss
```diff
@@ -19,7 +19,7 @@
 }
 
 .is-dark-theme .card {
-  box-shadow: none;
+  box-shadow: 0 0 20px rgba(0, 0, 0, 0.25);
```
taoeffect: Why did you make this change?

As far as I can tell this is not approved by the designs.

How this looks:

Screen Shot 2023-04-28 at 3 31 38 PM

How it should look:

Screen Shot 2023-04-28 at 3 31 44 PM

taoeffect (changes requested):
Great work @Silver-IT. Only thing left before I can merge is checking the chat-related tests for instances where Cypress isn't waiting properly for the app to "come to a completely stop", e.g. like the one we discussed on Slack:

```js
Cypress.Commands.add('giWaitUntilMessagesLoaded', (isJoined = true) => {
  cy.getByDT('conversationWrapper').within(() => {
    // NOTE: '.infinite-status-prompt:first-child' is the spinner css selector
    cy.get('.infinite-status-prompt:first-child')
      .invoke('attr', 'style')
      .should('not.include', 'display: none')
    cy.get('.infinite-status-prompt:first-child')
      .invoke('attr', 'style')
      .should('include', 'display: none')
  })
```

Then once I verify that everything passes both locally on my machine and on Travis I can merge!

taoeffect (approved):
Congrats @Silver-IT! All tests passing locally and on Travis!

Excellent work, this was a tough one!

PR #1541, opened 2023-04-16
Create public chatroom option on chat creation flow

### LINKED ISSUE #1380: Create Public Chatroom option on chat creation flow
### Problem

- The current UI accounts for group channels (visible to all members of a group) and private channels (visible to selected members of the group)
- We want to allow for the creation of a third privacy level: Public Channel (people from outside the group can see the channel’s content)

### Solution

1. Group Settings Page - Create a way for group admin's to choose if the group may have public channels or not 
2. Public Chatroom option is inactive by default
3. If Public Chatroom option is active - update the channel creation flow show the 3 privacy level options
4. If the channel is public, clearly show that in the chat area of that channel, so users don't share sensitive information
5. Update channel list icons

---
### 1. Group Settings Page - Create a way for group admin's to choose if the group may have public channels or not 

- [x] On the Group Settings Page, add a new panel for Public Chatrooms that allows group admin's to choose if the group members will be able to create public chatrooms or not
- [x] By default, Public Chatroom option is inactive, which means that the chat creation flow remains the same as we have today. [See figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=15649%3A89)

Screenshot 2022-08-10 at 10 13 00

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23589%3A258298

EDITED by @Silver-IT: I want to rename `Chatroom` to `Channel` as we use `Chatroom` only in the code and we use `Channel` in the design. Users are more familiar with `Channel`. cc: @taoeffect, @leihla

---
 
### 2. Public Chatroom option is inactive by default
- [x] When Public Chatroom option is inactive, the create new channel flow remains unfollowed and the user can toggle between group channel or private channel

Screenshot 2022-08-10 at 10 21 30

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=15649%3A89

---

### 3. If Public Chatroom option is active 
- [x] When Public Chatroom option is active, update the create new channel flow and instead of the toggle, replace with a dropdown with 3 options: Group Channel, Private Channel and Public Channe

Screenshot 2022-08-10 at 10 31 51

Screenshot 2022-08-10 at 10 23 46

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23551%3A253664

---

### 4. If the channel is public, clearly show that in the chat area of that channel, so users don't share sensitive information

- [x] Add warning background and notice to users on public chat rooms

Screenshot 2022-08-10 at 10 27 46

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23551%3A253706

---

### 5. Update channel list icons

- [x] On channel list, add icon for public chatroom

Screenshot 2022-08-10 at 10 33 38

**Figma**: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23551%3A253706

---

### 6. Cypress Test Scenario

- [x] Add test case for creating public channel and check the styles

---

**Related design issue:** https://github.com/okTurtles/group-income/issues/1364

Silver-IT:
### Summary of Changes

- Closes #1380 
- Updated color palette for dark theme


test/cypress/support/commands.js
```diff
@@ -444,8 +443,11 @@ Cypress.Commands.add('giRedirectToGroupChat', () => {
 })
 
 Cypress.Commands.add('giWaitUntilMessagesLoaded', () => {
+  cy.get('.c-initializing').should('not.exist')
   cy.getByDT('conversationWrapper').within(() => {
-    cy.get('.c-initializing').should('not.exist')
+    cy.get('.infinite-status-prompt:first-child')
+      .invoke('attr', 'style')
+      .should('include', 'display: none')
```
taoeffect, line 450: Did you mean to introduce this change?

Silver-IT: `.c-initializing` is always existing while loading the first page of the channel. It works just like a skeleton and it could be replaced with the awesome skeleton component later.
So when user opens or switches channels, it is always there in the chat so we should wait until it would be disappeared.
While fixing the Cypress error in this PR, I noticed that `.c-initializing` doesn't exist **inside** the `conversationWrapper` element. They are siblings each other, so I take it out.

And `.infinite-status-prompt:first-child` is the infinite loading spinner. In many cases of our test scenario, we don't make many messages which can't fill even one page. So infinite loading tries to load more and also we should wait until it is fully done.

Silver-IT: I tried 5 times in row after update this code, Cypress didn't break.

frontend/views/pages/GroupSettings.vue
```diff
@@ -146,8 +159,17 @@ export default ({
...
       }
+    },
+    isGroupAdmin () {
+      // TODO: group admin doesn't mean group creator
+      // https://github.com/okTurtles/group-income/issues/202
+      return this.groupSettings.groupCreator === this.ourUsername
     }
   },
+  mounted () {
+    // TODO: need to remove double exclamation mark after release version 1
```
taoeffect: This comment is unclear. What needs to be check?

frontend/views/pages/GroupSettings.vue
```diff
@@ -146,8 +159,17 @@ export default ({
         contractID: this.$store.state.currentGroupId,
         key: 'groupPicture'
       }
+    },
+    isGroupAdmin () {
+      // TODO: group admin doesn't mean group creator
+      // https://github.com/okTurtles/group-income/issues/202
+      return this.groupSettings.groupCreator === this.ourUsername
```
taoeffect: For now please comment this out and `return false` always.

The reason is we need to also check whether the Chelonia server admin has allowed us to create public chatrooms.

frontend/model/contracts/group.js
```diff
@@ -548,7 +548,8 @@ sbp('chelonia/defineContract', {
             groupCreator: meta.username,
             distributionPeriodLength: 30 * DAYS_MILLIS,
             inviteExpiryOnboarding: INVITE_EXPIRES_IN_DAYS.ON_BOARDING,
-            inviteExpiryProposal: INVITE_EXPIRES_IN_DAYS.PROPOSAL
+            inviteExpiryProposal: INVITE_EXPIRES_IN_DAYS.PROPOSAL,
+            publicChannelCreateAllowance: false
```
taoeffect: `publicChannelCreateAllowance` is bad English. Could you please rename this everywhere to something like, `allowPublicChannels` ?

taoeffect (changes requested):
Nice work Alex! Left a review.

frontend/views/pages/GroupSettings.vue
Silver-IT: Wanted to say that _Double exclamation mark should be removed when we release this version and no more previous versions are using. That's because groupSettings.allowPublicChannels could be undefined._

frontend/views/pages/GroupSettings.vue
Silver-IT:
```
isGroupAdmin () {
  // TODO: https://github.com/okTurtles/group-income/issues/202
  return false
}
```
Is this fine?

frontend/views/pages/GroupSettings.vue
Silver-IT: Can I remove this comment and also two exclamation mark too?

taoeffect: Sure

taoeffect (approved):
Excellent work @Silver-IT! Well done! Tested & approved!

PR #1550, opened 2023-04-21
#1545 Expired proposals are not archived

### LINKED ISSUE #1545: [Bug] Proposals aren't expiring, should become archived upon expiring
### Problem

A proposal should archive after it expires, but for some reason it's not, and also the text should say that it `Expired on [date]` but it doesn't:

Screen Shot 2023-04-19 at 2 46 06 PM

### Solution

Find and fix the bug!

SebinSong:
closes #1545


frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -53,6 +53,10 @@ export default ({
     ProposalItem,
     ButtonDropdownMenu
   },
+  created () {
+    // check if there are expired proposals to archive
+    sbp('gi.actions/group/checkAndArchiveExpiredProposals', { contractID: this.$store.state.currentGroupId })
+  },
```
taoeffect: This will result in a GIMessage being created each time this widget is loaded. That would flood the contract chain.

Please only send any messages if and when they are necessary.

taoeffect (changes requested):
Reviewed!

SebinSong:
@taoeffect 
updated the PR so that it checks for the expired proposals first and then makes the sbp call if there is any.

frontend/model/contracts/group.js
```diff
@@ -819,6 +819,21 @@ sbp('chelonia/defineContract', {
...
+        if (data.proposalIds.length) {
+          for (const proposalId of data.proposalIds) {
+            const proposal = state.proposals[proposalId]
+
+            Vue.set(proposal, 'status', STATUS_EXPIRED)
+            archiveProposal({ state, proposalHash: proposalId, proposal, contractID })
+          }
+        }
+      }
+    },
```
taoeffect: Let's imagine 2 users were online at the same time and their clients both sent this message at the same time. In that case we'd call `archiveProposal` twice on the same proposal, even though it was already archived. That could lead to problems.

So inside here we also need to have a check to make sure that `proposal` actually exists on `state.proposals[proposalId]` (because it wouldn't exist if 2 messages were sent - the 2nd message would result in an error upon processing)

frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -53,6 +53,10 @@ export default ({
     ProposalItem,
     ButtonDropdownMenu
   },
+  created () {
+    // check if there are expired proposals to archive
+    this.checkExpiredProposals()
```
taoeffect: This does a single check once when then ProposalsWidget is loaded. But what if we are already loaded and the proposal expires while we are looking at the page?

We want to make sure:

1. The user cannot vote on a proposal that has actually expired. So when they vote the check should happen there.
2. That the check is performed periodically, say every 3 minutes or so. This might be a useful candidate to combine with your periodic notifications code perhaps

taoeffect (changes requested):
Excellent improvement @SebinSong! Some more final issues spotted!

frontend/views/containers/proposals/ProposalsWidget.vue
SebinSong: @taoeffect Agree with below

> This might be a useful candidate to combine with your periodic notifications code perhaps

As you might remember, we currently have a periodic notification logic that checks for expiring proposals that recurs every 15mins. Since we are dealing with a very similar topic(`expiring` and `expired` proposals) here, I decided to make additions to that existing logic that it checks both 'expired' and 'expiring' proposals in it and then reduced the recurrence period down to `5min` from the previous `15min`. Hope it sounds good.

taoeffect: @SebinSong That sounds like a great idea!  👍

frontend/model/notifications/periodicNotifications.js
```diff
@@ -14,7 +14,7 @@ export const PERIODIC_NOTIFICATION_TYPE = {
 }
 
 const every1MinTimeout = { notifications: [], state: {}, delay: MINS_MILLIS }
-const every5MinTimeout = { notifications: [], state: {}, delay: 5 * MINS_MILLIS }
+const every5MinTimeout = { notifications: [], state: {}, delay: 0.15 * MINS_MILLIS }
```
taoeffect: What happened here? This seems like a mistake maybe?

SebinSong: yes, I think I forgot to change it back after testing the periodic notification. will fix it.

taoeffect (approved):
Excellent work @SebinSong! 😄 🎉

PR #1552, opened 2023-04-22
Use key prefixes only in 'chel/db/get', not in actual DB keys

### LINKED ISSUE #1548: Changes to how keys are stored in db and queried
### Problem

With the way #1527 added prefix support to keys (e.g. `blob=${hash}`), we now have a slight confusion / ugliness where the same key could theoretically be stored twice (e.g. `blob=hash1`, `=hash1`).

It makes sense to store the prefix when it is a literal part of the key (as with `name=` and `head=`), but otherwise it doesn't make sense.

### Solution

Add the ability to query the value via a specific encoding a key prefix in `'chelonia/db/get'`, so that you can for example do:

```js
sbp('chelonia/db/get', `blob:${hash}`)
```

But don't store that prefix as the key. In this case it would return a blob. Assume utf8 string by default. Note that sqlite3 will actually return a buffer by default if the key was stored using `/file`, so if`blob:` isn't present on the query, it would need to convert the buffer to `utf8`.

snowteamer:
Closes #1548

#### Changes in `'chelonia/db/get'`:
-    If no prefix is given, only blobs are decoded to strings before returning. All other values are returned as-is, whether they are strings or not.
-    If `blob:` is passed in but the DB backend returns a string, an error is thrown. The alternative would be to encode the string to a buffer.
-    The `any:` prefix is used in `GET file/` to pass data from the DB backend to `h.response()` without transforming the data ourselves. I didn't want to decode buffers here since Hapi or the OS would likely encode them back anyway when sending, and `blob:` would throw with strings.

#### Other changes:
- The internal `readData() `and `writeData() `don't take any encoding or key prefix in their arguments. They're just the lowest-level DB access layer we have, and may behave differently for different backends (unlike the 'chelonia/db/' functions, which _must_ be consistent). An alternative would be to allow an `encoding` parameter which would be used when the raw data is a buffer, but ignored otherwise.
- Key validation with `checkKey()` has been moved from the internal backend functions into `'chelonia/db/set'`, to make sure the same constraints are applied whatever the DB backend.
 Note that `'chelonia/db/get'` and `'-delete'` don't use prefixable keys, so they don't need this check. Although in the `fs` backend an additonnal check is done in `readData()` against path traversal attacks. Other backends do not need it.

### DIFF
--- a/backend/database-fs.js
+++ b/backend/database-fs.js
@@ -2,8 +2,7 @@
 
 import { mkdir, readdir, readFile, unlink, writeFile } from 'node:fs/promises'
 import { join, resolve } from 'node:path'
-
-import { checkKey } from './database.js'
+import { checkKey } from '~/shared/domains/chelonia/db.js'
 
 // Initialized in `initStorage()`.
 let dataFolder = ''
@@ -19,28 +18,15 @@ export function clear (): Promise<void> {
     .then(keys => Promise.all(keys.map(key => unlink(join(dataFolder, key)))))
 }
 
-export function exportToJSON (): Promise<Object> {
-  return readdir(dataFolder)
-    .then(keys => Promise.all(keys.map(key => readData(key).then(value => ([key, value])))))
-    .then(Object.fromEntries)
-}
-
-export function importFromJSON (json: Object): Promise<void> {
-  // TODO: rollback the operation if an error occurs?
-  return Promise.all(Object.keys(json).map(key => writeData(key, json[key])))
-    .then(() => undefined)
-}
-
 // eslint-disable-next-line require-await
-export async function readData (key: string): Promise<Buffer | void> {
+export async function readData (key: string): Promise<Buffer | string | void> {
+  // Necessary here to thwart path traversal attacks.
   checkKey(key)
   return readFile(join(dataFolder, key))
-    .then(buffer => key.startsWith('blob=') ? buffer : buffer.toString('utf8'))
     .catch(err => undefined) // eslint-disable-line node/handle-callback-err
 }
 
 // eslint-disable-next-line require-await
 export async function writeData (key: string, value: Buffer | string): Promise<void> {
-  checkKey(key)
   return writeFile(join(dataFolder, key), value)
 }
--- a/backend/database-sqlite.js
+++ b/backend/database-sqlite.js
@@ -4,8 +4,6 @@ import { mkdir } from 'node:fs/promises'
 import { join, resolve } from 'node:path'
 import sqlite3 from 'sqlite3'
 
-import { checkKey } from './database.js'
-
 let db: any = null
 let readStatement: any = null
 let writeStatement: any = null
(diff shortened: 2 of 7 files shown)


backend/database-fs.js
```diff
@@ -19,28 +18,15 @@ export function clear (): Promise<void> {
...
+  // Necessary here to thwart path traversal attacks.
   checkKey(key)
   return readFile(join(dataFolder, key))
-    .then(buffer => key.startsWith('blob=') ? buffer : buffer.toString('utf8'))
     .catch(err => undefined) // eslint-disable-line node/handle-callback-err
 }
 
 // eslint-disable-next-line require-await
 export async function writeData (key: string, value: Buffer | string): Promise<void> {
-  checkKey(key)
```
taoeffect, line 44: We do also want to make sure no path traversal attacks happen here (files outside of the data directory must not be written to)

backend/database-sqlite.js
```diff
@@ -96,7 +65,6 @@ export function readData (key: string): Promise<Buffer | string | void> {
 
 export function writeData (key: string, value: Buffer | string): Promise<void> {
   return new Promise((resolve, reject) => {
-    checkKey(key)
```
taoeffect, line 99: Similar comment here. While we are still supporting `fs`, and even after, we do want to make sure improperly named keys are not being passed in.

backend/database-sqlite.js
```diff
@@ -51,37 +49,8 @@ export function clear (): Promise<void> {
...
-          resolve()
-        }
-      })
-    })
-  })
-}
-
 export function readData (key: string): Promise<Buffer | string | void> {
   return new Promise((resolve, reject) => {
-    checkKey(key)
```
taoeffect, line 84: And here

backend/database.js
```diff
@@ -186,18 +178,22 @@ export default async () => {
...
+        let value = await readData(key)
+        if (value === undefined) {
+          return
         }
+        value = prefixHandlers[prefix](value)
+        cache.set(prefixableKey, value)
         return value
       },
       'chelonia/db/set': async function (key: string, value: Buffer | string): Promise<void> {
+        checkKey(key)
```
taoeffect, line 196: I'm not sure about putting this here. For example, for the `readData` in `fs` the `checkKey` is still in `readData`, so that's an inconsistency, but more to the point, by keeping it "closer to the action" we can be more sure that it was called before the actual setting happens.

taoeffect, line 196: So, we need to decide whether to have this here or have this somewhere else. At the moment, `readData` in `fs` the `checkKey` is still in `readData`, so that's an inconsistency. Should we have this here, or "closer to the action"?

taoeffect, line 196: So, we need to decide whether to have this here or have this somewhere else. At the moment, `readData` in `fs` the `checkKey` is still in `readData`, so that's an inconsistency. Should we have this here, or "closer to the action"?

Whatever is decided, it needs to be done consistently everywhere. For example, right now in the in-memory `'chelonia/db/get'` and `set`, it's not called. But it should be if we decided that it's better to have it here than in `readData` / `writeData`.

taoeffect (changes requested):
@snowteamer excellent work!

Left a few comments, primarily we need to figure out how to make `checkKey` usage consistent everywhere.

backend/database-fs.js
snowteamer: Sure, that's done in 'chelonia/db/set', This writeFile() function is not meant to be called directly (except in tests)

backend/database.js
snowteamer: Yea I was trying to improve consistency and performance, and thought maybe it'd be better to have one key check in `'chelonia/db/set'` to be sure no bad key can enter the DB, no matter what every individual backend's `writeData()` would do. Then the key check was no longer needed when getting or deleting entries, since the DB could now be assumed free from any invalid key... Except that in `fs` we also have to prevent path attacks, so I added a key check in `fs`'s `readData()` (actually I could have used a simpler check than checkKey here) . If there was a `deleteData()` function it would need the same key check.

snowteamer: If you'd rather have all the checks in the storage backends and none in chelonia/db then I will undo this change, except that I think I won't put a key check back in sqlite's `readData()` since it's not vulnerable to path attacks.

taoeffect: @snowteamer let me think on this. It might be fine.

taoeffect (approved):
Approved! Nicely done @snowteamer 👍

PR #1565, opened 2023-04-29
Fix issue #1564

### LINKED ISSUE #1564: `grunt dev` sometimes reloads the app while the backend is down after a `shared` file change
#### Steps to reproduce:
1. Run `export GI_PERSIST= && grunt` or `export GI_PERSIST=sqlite && grunt dev`, with enough files in `./data` to make a server restart slow enough, i.e. at least slower than a frontend rebuild. The issue won't show up otherwise.
2. Visit `http://localhost:3000`
3. Make a trivial change in a shared file, e.g. remove a character in a comment in `./shared/functions.js`

#### Expected behavior:
The app should display a "Trying to reconnect..." banner for a while, then reload normally without that banner nor errors in the console.

#### Actual behavior:
The app quickly reloads, but starts displaying a "Trying to reconnect..." banner for a while, then the banner disappears but the page is unable to finish loading (the page load indicator stays cross-shaped). There are connection errors and reconnection attempt messages in the console:

image

### Solution

Make sure to only reload the browser when the backend is fully ready to accept connections and serve assets.

snowteamer:
Closes #1564

### Additional changes:
- No longer copy all our assets again to `./dist/assets` when a single one has changed. Only copy the changed one.


Gruntfile.js
```diff
@@ -50,6 +50,24 @@ const applyPortShift = (env) => {
...
 }
 
+const untilServerIsReady = () => {
+  console.log('waiting for the server to be ready')
+  // Wait for the server to be ready.
+  const t0 = Date.now()
+  const timeout = 30000
+  return new Promise((resolve, reject) => {
+    (function ping () {
+      fetch(process.env.API_URL).then(resolve).catch(() => {
```
taoeffect: I think you can [directly cause `fetch` to timeout](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#aborting_a_fetch):

Screen Shot 2023-04-29 at 3 26 17 PM

Gruntfile.js
```diff
@@ -50,6 +50,24 @@ const applyPortShift = (env) => {
   return { ...env, API_PORT, API_URL }
 }
 
+const untilServerIsReady = () => {
```
taoeffect: Maybe this should be moved below under the comment that says `// Task functions`?

taoeffect (changes requested):
Excellent work @snowteamer! Really interesting changes. I left 2 comments + 1 on Slack showing that there's an issue with how long the browsersync plugin takes to load.

Gruntfile.js
snowteamer: This technique could be useful in the client, for example to implement a button allowing users to stop fetching a large ressource when it takes too long.

In this function, the specified timeout is not meant for individual fetch requests, but is the total time we're willing to spend polling the server before giving up. That's why it's quite generous (30s).
Aborting individual requests would defeat their purpose, which is to test whether the server is ready. The ressource is not used.
If the server is not up, then any fetch quickly fails by itself. According to my tests it won't stay pending waiting to be aborted. So In that case the `catch()` clause runs and schedules another fetch request, unless the timeout has been exceeded.

Gruntfile.js
snowteamer: It's true that it's used in some task e.g. `backend:relaunch`, but it's not meant to be directly wrapped into a grunt task (there is no `grunt untilServerIsReady` task).

taoeffect: But there is no `grunt copyTo` and `grunt fork2` tasks either, so I'm not sure I understand your reply...

Gruntfile.js
taoeffect: Ah, I have a better understanding of how this function works now, thanks, resolving this

Gruntfile.js
snowteamer: Hmm let's move it then. You're right about `copyTo` and `fork2`. I thought they belonged here, but in fact that's not so much consistent.

taoeffect:
Closing (stale)

PR #1585, opened 2023-05-09
Chat Improvements

### LINKED ISSUE #1582: Can not create channel
### Problem

Steps to reproduce
1. Focus on the channel which is not part of.
2. Try to create a new channel.
3. We would get the following error message and can not create channel.
image

### Solution

Debug and fix it.

### LINKED ISSUE #1583: Archived data is shared to everyone
### Problem

Since `gi.db/archive/save` saves data to the IndexedStorage, we should point the whose data is it.

### Solution

Try to find all the codes using `gi.db/archive/save` function and add `ourUsername` in the key.

Silver-IT:
### Summary of Changes

- Closes #1582 
- Closes #1583 
- Fix cypress bug in join channel
- Fix error in watching summary


Silver-IT (commented):
Ready for your review, @taoeffect

frontend/model/contracts/group.js
```diff
@@ -1359,7 +1359,7 @@ sbp('chelonia/defineContract', {
         archSentOrReceivedPayments.sent = [...sortPayments(newSentOrReceivedPayments.sent), ...archSentOrReceivedPayments.sent]
         archSentOrReceivedPayments.received = [...sortPayments(newSentOrReceivedPayments.received), ...archSentOrReceivedPayments.received]
 
-        const archPaymentsKey = `payments/${period}/${username}/${contractID}`
+        const archPaymentsKey = `payments/${username}/${period}/${contractID}`
```
taoeffect, line 1362: So the username was being saved to the key...?

How does this materially change anything?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -701,40 +702,59 @@ export default ({
...
+        if (sbp('chelonia/contract/isSyncing', toChatRoomId)) {
+          this.archiveMessageState(fromChatRoomId)
+          this.setMessageEventListener({ to: toChatRoomId, from: fromChatRoomId })
+          if (toIsJoined && sbp('chelonia/contract/isSyncing', toChatRoomId)) {
+            initAfterSynced(toChatRoomId)
+          }
+        } else {
+          this.refreshContent(toChatRoomId, fromChatRoomId)
+        }
+      } else if (toIsJoined && toIsJoined !== fromIsJoined) {
```
taoeffect, line 751: I don't understand this `isJoined` thing.

In `ChatMixins.js` it is set to `true`, so this condition doesn't make sense.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -701,40 +702,59 @@ export default ({
...
           }
         })
       }
+
+      if (toChatRoomId !== fromChatRoomId) {
+        this.ephemeral.messagesInitiated = false
+        if (sbp('chelonia/contract/isSyncing', toChatRoomId)) {
+          this.archiveMessageState(fromChatRoomId)
+          this.setMessageEventListener({ to: toChatRoomId, from: fromChatRoomId })
+          if (toIsJoined && sbp('chelonia/contract/isSyncing', toChatRoomId)) {
```
taoeffect: You already performed `sbp('chelonia/contract/isSyncing', toChatRoomId)` check up above

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -701,40 +702,59 @@ export default ({
...
+      const toIsJoined = to.isJoined
+      const fromIsJoined = from.isJoined
+
+      const initAfterSynced = (chatRoomId) => {
+        // TODO: `okTurtles.events/once` shouldn't be used here
+        //       since another contract could be started or finished being synced in advance
         sbp('okTurtles.events/once', CONTRACT_IS_SYNCING, (contractID, isSyncing) => {
-          if (contractID === this.currentChatRoomId && isSyncing === false) {
-            this.setMessageEventListener({ from: contractID, to: contractID })
+          if (contractID === chatRoomId && isSyncing === false) {
```
taoeffect: Let's address this TODO comment in this PR.

By saving the event handler to a variable, you can use `okTurtles.events/on` and `okTurtles.events/off`

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -701,40 +702,59 @@ export default ({
...
         // so it's good to initiate using input parameter chatRoomId
-        sbp('gi.db/archive/save', this.getArchiveKeyFromChatRoomId(chatRoomId), JSON.stringify(this.latestEvents))
+        sbp('gi.db/archive/save', this.archiveKeyFromChatRoomId(chatRoomId), JSON.stringify(this.latestEvents))
       }
     },
-    getArchiveKeyFromChatRoomId (chatRoomId) {
+    archiveKeyFromChatRoomId (chatRoomId) {
       const curChatRoomId = chatRoomId || this.currentChatRoomId
-      return `messages/${curChatRoomId}`
+      return `messages/${this.ourUsername}/${curChatRoomId}`
```
taoeffect, line 714: Just to be clear: in the future, once we release, changes like this will require a migration step when the app is starting up. Otherwise without it you'll be polluting their local IndexedDB with data that is then "stuck" there

taoeffect (changes requested):
Reviewed! Nice work @Silver-IT 👍 

Left a few comments

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Correct. We should take care of the `bad-formatted` data in the storage either.

I think we can redefine the three functions (`'gi.db/archive/delete'`, `'gi.db/archive/save'`, `'gi.db/archive/load'`) after the user logs in so that we don't need to take care of username in making keys.
```
sbp('sbp/selectors/overwrite', {
  'gi.db/archive/delete': function (key: string): Promise<Object> {
    const usernamePrefix = sbp('state/vuex/state').loggedIn.username
    return archive.removeItem(`${usernamePrefix}/${key}`)
  },
  'gi.db/archive/save': () => { ... },
  'gi.db/archive/load': () => { ... }
}
```

Silver-IT: I mean we can redefine those three functions after the user logs in so that we don't need to take care of username in making keys.

taoeffect: That's an interesting idea. If we only ever use those functions when logged in, then yes we could do that. I'm not sure if we do.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -701,39 +702,58 @@ export default ({
...
+          toIsJoined && initAfterSynced(toChatRoomId)
+        } else {
+          this.refreshContent(toChatRoomId, fromChatRoomId)
+        }
+      } else if (toIsJoined && toIsJoined !== fromIsJoined) {
+        if (sbp('chelonia/contract/isSyncing', toChatRoomId)) {
+          // NOTE: toChatRoomId equals to fromChatRoomId here
+          this.setMessageEventListener({ to: toChatRoomId, from: fromChatRoomId })
+          initAfterSynced(toChatRoomId)
+        }
```
taoeffect: And what if `isSyncing` returns `false` here?

Silver-IT: This is when user joins the channel by himself clicking the `Join Channel` button.
So the `join` action would be the last action and it couldn't return `false` here.

Silver-IT: This is when the user joins the channel by himself clicking the `Join channel` button, so the `join` action would be the last action of the chatroom contract.

I tested to see the `isSyncing` value, and it returned `true` at this time. I guess that's because chatroom contract is state is changed before the contract is finished to be synced. However, we should take care of the `false` value since the `join` action is the last action and the syncing would be immediately finished.

taoeffect (approved):
Excellent work @Silver-IT! Approved!

PR #1593, opened 2023-05-12
E2e protocol greg

taoeffect:


frontend/controller/actions/group.js
```diff
@@ -407,6 +407,8 @@ export default (sbp('sbp/selectors/register', {
             chatRoomId: rootState[params.contractID].generalChatRoomId
           })
         }
+
+        sbp('okTurtles.events/emit', PENDING_TO_WELCOME)
```
corrideat, line 467: I think this needs to be done at two different locations instead of here: at the end of the `if (sendKeyRequest) {}` block and at the next else (with `console.log('@@@@@@@@ AT join[invalid state] for ' + params.contractID)`)

frontend/views/pages/PendingApproval.vue
```diff
@@ -12,18 +13,37 @@ div
...
+      }
+    }
   },
   computed: {
     ...mapGetters([
       'groupSettings'
     ])
+  },
+  mounted () {
+    sbp('okTurtles.events/on', PENDING_TO_WELCOME, () => {
```
taoeffect, line 41: Ricardo says I could watch for `!state.profiles?.[username]` where `state` is the group state

taoeffect: we can create a computed property that looks at the specific group state by saving the currentGroupId when mounted

frontend/controller/actions/group.js
taoeffect: >  at the end of the `if (sendKeyRequest) {}` block

I don't think that's right... This needs to happen after we've received `OP_KEY_SHARE` and have successfully sync'd the group. At the end of `if (sendKeyRequest) {}` block we have not received `OP_KEY_SHARE`

corrideat: I think you're right

corrideat (approved):

PR #1597, opened 2023-05-16
Add validation in Login Form

### LINKED ISSUE #1525: Validations on SignUp form aren't triggered when LastPass fills in fields
### Problem

When a user is using LastPass to fill in their information on the Sign Up form, the validations aren't triggered when they should be.

This can lead to confusion, where their username is automatically filled in with a capital letter, but they cannot sign up because the sign up button is greyed out.

It appears that the sign up button is disabled correctly by the validators, but the red error message telling the user that they cannot use capital letters isn't shown for some reason.

### Solution

Find out why the red error message isn't shown when the username field is autofilled by LastPass with a capital letter and fix it.

Silver-IT:
### Summary of Changes

- Closes #1525

### DIFF
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -82,6 +82,10 @@ export default ({
   },
   methods: {
     async login () {
+      if (typeof document.body.dispatchEvent === 'function') {
+        this.$refs.username.dispatchEvent(new Event('input'))
+        // TODO: password input too
+      }
       if (this.$v.form.$invalid) {
         this.$refs.formMsg.danger(L('The form is invalid.'))
         return
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -30,14 +30,22 @@ form(data-test='login' @submit.prevent='')
 <script>
 import sbp from '@sbp/sbp'
 import { validationMixin } from 'vuelidate'
-import { required } from 'vuelidate/lib/validators'
+import { required, maxLength } from 'vuelidate/lib/validators'
 import { L } from '@common/common.js'
 import BannerScoped from '@components/banners/BannerScoped.vue'
 import ButtonSubmit from '@components/ButtonSubmit.vue'
 import PasswordForm from '@containers/access/PasswordForm.vue'
 import { requestNotificationPermission } from '@model/contracts/shared/nativeNotification.js'
 import validationsDebouncedMixins from '@view-utils/validationsDebouncedMixins.js'
-import { noWhitespace } from '@model/contracts/shared/validators.js'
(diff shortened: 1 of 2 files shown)


Silver-IT (commented):
@taoeffect, as we discussed [here](https://github.com/okTurtles/group-income/issues/1525#issuecomment-1547015334), I just added validation rules so that it could be checked when user click `Login` button.

It's ready for your review.

taoeffect (changes requested):
Per our discussion on Slack: https://okturtles.slack.com/archives/C0EH7P20Y/p1684454697388439

frontend/views/containers/access/LoginForm.vue
```diff
@@ -104,7 +112,13 @@ export default ({
     form: {
       username: {
         [L('A username is required.')]: required,
-        [L('A username cannot contain whitespace.')]: noWhitespace
+        [L('A username cannot contain whitespace.')]: noWhitespace,
```
taoeffect: For all of the validations that this has in common with `SignupForm.vue`, could you store the on an object in `SignupForm.vue` that you import here?

e.g. in `SignupForm.vue`:

```js
export const signupValidations = {
  [L('A username is required.')]: required,
  // ...
}
```

Then here:

```js
import { signupValidations } from ...
// ...
username: {
  ...signupValidations,
  // ...
}
```

taoeffect (changes requested):
Nicely done @Silver-IT. A DRY request

frontend/views/containers/access/LoginForm.vue
```diff
@@ -74,6 +74,10 @@ export default ({
   },
   methods: {
     async login () {
+      if (typeof document.body.dispatchEvent === 'function') {
+        this.$refs.username.dispatchEvent(new Event('input'))
+        // TODO: password input too
+      }
```
taoeffect: Hmm, what is the purpose of this code?

This looks like it either needs a comment explaining what it's doing, or needs to be removed. I'm noticing a weird console error now whenever I bring up the LoginModal and dismiss it while the username field is empty:

https://github.com/okTurtles/group-income/assets/138706/e090684f-fafa-44b6-8280-371e281f87dd

taoeffect (changes requested):
Nicely done! Almost ready to merge I think. Found a new error:

frontend/views/containers/access/LoginForm.vue
Silver-IT: > Hmm, what is the purpose of this code?

It's for making `Input` event manually when user click `Login` button so that the `this.form.username` represents the current value of input element even though the injected script updated its value. If all the chrome extension emit this `Input` event after they update the value of input element, this isn't needed.
I'll remove this because I think `LastPass` extension emits the `Input` event and let us have the correct username.

taoeffect (approved):
Nicely done @Silver-IT!

PR #1601, opened 2023-05-18
Resolved DRY in chatroomUnreadMessages

### LINKED ISSUE #1590: DRY chatRoomUnread
### Problem

In PR #1574 I noticed a lot of DRY violations related to `chatRoomUnread`:

```js
Vue.set(state.chatRoomUnread[chatRoomId], 'mentions', prevMentions.filter(m => m.messageHash !== messageHash))
Vue.set(state.chatRoomUnread[chatRoomId], 'others', prevOthers.filter(m => m.messageHash !== messageHash))

```js
mentions: prevMentions.filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime()),
others: prevOthers.filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
```

```js
    if (isDMOrMention) {
      state.chatRoomUnread[chatRoomId].mentions.push(newItem)
    } else {
      if (!('others' in state.chatRoomUnread[chatRoomId])) {
        Vue.set(state.chatRoomUnread[chatRoomId], 'others', [newItem])
      } else {
        state.chatRoomUnread[chatRoomId].others.push(newItem)
      }
    }
```

### Solution

After discussing with @SebinSong & @Silver-IT [here](https://github.com/okTurtles/group-income/pull/1584#pullrequestreview-1421549371) and on Slack, we decided that `mentions` and `others` could be combined by adding a `type` attribute that says the type (e.g. `type: "mention"`, `type: "other"` or whatever)

From: https://okturtles.slack.com/archives/C0EH7P20Y/p1683772080647849

```
export type GOOD_NAME = {
  messageHash: string;
  createdDate: string;
  type: string;
  deletedDate?: string;
}
```

```
readUntil: GOOD_NAME,
messages: GOOD_NAME[] // [ mentions + others ]
```

Don't forget to convert the old `state.chatRoomUnread` format in `postUpgradeVerification`

### LINKED ISSUE #1592: Chat: "Add description" not linked
### Problem

If a user is not the chatroom creator, we still show an "Add description" link, but it is not clickable:

Screen Shot 2023-05-12 at 10 08 19 AM

This is confusing.

### Solution

In this situation, we should do one of two things to address this. Either:

1. Remove the link/text completely
2. Allow it to be clicked but display an error saying "Only the chatroom creator is allowed to set the description"

Silver-IT:
### Summary of Changes

- Closes #1590 
- Closes #1592


frontend/model/contracts/chatroom.js
```diff
@@ -75,13 +76,18 @@ function messageReceivePostEffect ({ contractID, messageHash, datetime, text, is
   const rootGetters = sbp('state/vuex/getters')
   const isDirectMessage = rootGetters.isDirectMessage(contractID)
   const isDMOrMention = isMentionedMe || isDirectMessage
+  const type = messageType === MESSAGE_TYPES.INTERACTIVE
+    ? UNREAD_MESSAGE_TYPE.INTERACTIVE
+    : messageType === MESSAGE_TYPES.TEXT && isDMOrMention
+      ? UNREAD_MESSAGE_TYPE.MENTION
+      : null
```
taoeffect: What about `MESSAGE_TYPES.POLL` that was recently merged? Should that be considered here?

frontend/model/state.js
```diff
@@ -64,6 +64,24 @@ sbp('sbp/selectors/register', {
     // if (!state.notifications) {
     //   state.notifications = []
     // }
+
+    for (const chatRoomId in state.chatRoomUnread) {
+      if (state.chatRoomUnread[chatRoomId].messages) {
```
taoeffect: Is this missing a `!` ?

```
if (!state.chatRoomUnread[chatRoomId].messages)
```

taoeffect (changes requested):
Excellent work @Silver-IT!

Left 2 comments

frontend/model/state.js
Silver-IT: Oh, nice catch.

frontend/model/contracts/chatroom.js
Silver-IT: Yeah. We could consider `MESSAGE_TYPES.POLL` here too.

frontend/model/contracts/chatroom.js
```diff
@@ -75,13 +76,20 @@ function messageReceivePostEffect ({ contractID, messageHash, datetime, text, is
...
   const isDMOrMention = isMentionedMe || isDirectMessage
-
-  if (!isAlreadyAdded && (isDMOrMention || messageType === MESSAGE_TYPES.INTERACTIVE)) {
+  const type = messageType === MESSAGE_TYPES.INTERACTIVE
+    ? UNREAD_MESSAGE_TYPE.INTERACTIVE
+    : messageType === MESSAGE_TYPES.POLL
+      ? UNREAD_MESSAGE_TYPE.POLL
+      : messageType === MESSAGE_TYPES.TEXT && isDMOrMention
+        ? UNREAD_MESSAGE_TYPE.MENTION
+        : null
```
taoeffect: Is there a way to simplify this so that it's not as ugly?

For example, can you change the code so that this works?

```js
const type = messageType
```

taoeffect (changes requested):
Left a comment + tested with the existing database. If this is deployed it will break the website with the following error:

Screen Shot 2023-05-24 at 10 26 35 AM

I recommend, before working on any PRs that involve changes to the state, to first create a local group using `GI_PERSIST=sqlite grunt dev`, so that you have a local `data/groupincome.db` that uses the old version of the site. Then you can test your changes for upgrading the previous version and catch bugs like this.

frontend/model/contracts/chatroom.js
Silver-IT: Does it look ugly to you? Hahaha. I will simplify this surely.

Silver-IT:
```
const type = messageType
    ? {
        [MESSAGE_TYPES.TEXT]: isDMOrMention ? MESSAGE_TYPES.TEXT : undefined,
        [MESSAGE_TYPES.INTERACTIVE]: MESSAGE_TYPES.INTERACTIVE,
        [MESSAGE_TYPES.POLL]: MESSAGE_TYPES.POLL
      }[messageType]
    : undefined
```

Is this much better?

taoeffect: That code will throw an exception when `NOTIFICATION: 'notification'` is passed in.

Is it possible to use the message type directly instead of creating another (very similar) type?

Silver-IT: > That code will throw an exception when NOTIFICATION: 'notification' is passed in.

The type will be `undefined` when `messageType` is `MESSAGE_TYPES.NOTIFICATION`.

> Is it possible to use the message type directly instead of creating another (very similar) type?

Yes we can. But please see the next line of `if` statement which decides if we should call `addChatRoomUnreadMessage`.

taoeffect: > The type will be `undefined` when `messageType` is `MESSAGE_TYPES.NOTIFICATION`.

Oops, you're right!

> Yes we can. But please see the next line of `if` statement which decides if we should call `addChatRoomUnreadMessage`.

Ah, I see. Ok yeah this code should be good. I think.

Also, I don't think you need the `?:` to check `messageType`. You can just directly index the object. `{ a: 5 }[undefined]` still returns `undefined`

Silver-IT: > The type will be undefined when messageType is MESSAGE_TYPES.NOTIFICATION.

Oops, I have  found an interesting issue.
I have tested the following three in the browser console and got an error for the first one unexpectedly.
```
{ name: 'Alex', age: 31 }['name'] // Uncaught SyntaxError
({ name: 'Alex', age: 31 })['name'] // 'Alex'
({ name: 'Alex', age: 31 })['height'] // undefined
```

taoeffect: Hmm, ok I guess make sure to wrap it in parenthesis 🤔

frontend/model/state.js
```diff
@@ -108,37 +127,59 @@ const mutations = {
...
+      }
+      if (state.chatRoomUnread[chatRoomId].mentions) {
+        Vue.set(state.chatRoomUnread, chatRoomId, {
+          readUntil: { messageHash, createdDate, deletedDate: null },
+          mentions: [],
+          messages: state.chatRoomUnread[chatRoomId].messages
+            .filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
+        })
+        return
+      }
```
taoeffect, line 151: @Silver-IT For users already in the group who've been mentioned, won't this block of code prevent the new code from kicking in because they have `state.chatRoomUnread[chatRoomId].mentions` defined and therefore this block will `return`?

frontend/model/state.js
```diff
@@ -108,37 +127,59 @@ const mutations = {
...
+      if (!state.chatRoomUnread[chatRoomId].messages) {
+        Vue.set(
+          state.chatRoomUnread[chatRoomId],
+          'messages',
+          [].concat(
+            (state.chatRoomUnread[chatRoomId].mentions || []).map(m => Object.assign({ type: MESSAGE_TYPES.TEXT }, m)),
+            (state.chatRoomUnread[chatRoomId].others || []).map(o => Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
+          )
+        )
+      }
```
taoeffect, line 142: I don't understand the purpose of this block... shouldn't this have already been taken care of by `'state/vuex/postUpgradeVerification'`?

frontend/model/state.js
```diff
@@ -108,37 +127,59 @@ const mutations = {
...
+    {
+      // TODO: need to remove this block after we release 0.2.*
+      if (state.chatRoomUnread[chatRoomId].mentions) {
+        Vue.set(
+          state.chatRoomUnread[chatRoomId],
+          'mentions',
+          state.chatRoomUnread[chatRoomId].mentions.filter(m => m.messageHash !== messageHash)
+        )
+      }
+    }
```
taoeffect, line 176: Again, this also doesn't make sense to me. Why is this here? `'state/vuex/postUpgradeVerification'` should have taken care of this and migrated everything to `messages`

frontend/model/state.js
```diff
@@ -64,6 +64,25 @@ sbp('sbp/selectors/register', {
...
+        state.chatRoomUnread[chatRoomId].mentions.forEach(m => {
+          state.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.TEXT }, m))
+        })
+        // Vue.delete(state.chatRoomUnread[chatRoomId], 'mentions')
+      }
+      if (state.chatRoomUnread[chatRoomId].others) {
+        state.chatRoomUnread[chatRoomId].others.forEach(o => {
+          state.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
+        })
+        // Vue.delete(state.chatRoomUnread[chatRoomId], 'others')
```
taoeffect, line 83: Why are these `Vue.delete` calls commented out?

frontend/model/state.js
```diff
@@ -108,37 +127,59 @@ const mutations = {
...
+        })
+        return
+      }
+    }
     Vue.set(state.chatRoomUnread, chatRoomId, {
       readUntil: { messageHash, createdDate, deletedDate: null },
-      mentions: prevMentions.filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime()),
-      others: prevOthers.filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
+      messages: state.chatRoomUnread[chatRoomId].messages
+        .filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
```
taoeffect, line 156: These lines are the problem, `state.chatRoomUnread[chatRoomId].messages` is `undefined` when this is called.

So maybe this line needs some `?.` and `|| []` type stuff?

frontend/model/state.js
taoeffect: These two `// Vue.delete(state.chatRoomUnread[chatRoomId], 'others')` lines can be uncommented.

frontend/model/state.js
```diff
@@ -108,37 +127,59 @@ const mutations = {
...
+      if (state.chatRoomUnread[chatRoomId].mentions) {
+        Vue.set(state.chatRoomUnread, chatRoomId, {
+          readUntil: { messageHash, createdDate, deletedDate: null },
+          mentions: [],
+          messages: state.chatRoomUnread[chatRoomId].messages
+            .filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
+        })
+        return
+      }
+    }
```
taoeffect: The above block will `return` if `state.chatRoomUnread[chatRoomId].mentions` is defined, which we do not want, because we want the new behavior.

Also, the above block can be deleted completely by adding something else after line 157. So below, after `Vue.set(state.chatRoomUnread, chatRoomId, {`, and after line 157, add this:

```js
    // eslint-disable-next-line no-lone-blocks
    {
      // hack: delete me after upgrade to 0.2.x!
      Vue.set(state.chatRoomUnread[chatRoomId], 'mentions', [])
      Vue.set(state.chatRoomUnread[chatRoomId], 'others', [])
    }
```

frontend/model/state.js
taoeffect: This can then be deleted

taoeffect (changes requested):
Nice job @Silver-IT! There are some issues with how `state.js` is implemented though. There is a way to simplify this greatly, while also fixing some bugs with how logic is handled in `setChatRoomReadUntil`.

I've added comments inline, and have also included my full set of changes here as a diff. Please apply and test to verify that this works also for you:

```diff
diff --git a/frontend/model/state.js b/frontend/model/state.js
index c233b89e..bfa46d67 100644
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -74,13 +74,13 @@ sbp('sbp/selectors/register', {
         state.chatRoomUnread[chatRoomId].mentions.forEach(m => {
           state.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.TEXT }, m))
         })
-        // Vue.delete(state.chatRoomUnread[chatRoomId], 'mentions')
+        Vue.delete(state.chatRoomUnread[chatRoomId], 'mentions')
       }
       if (state.chatRoomUnread[chatRoomId].others) {
         state.chatRoomUnread[chatRoomId].others.forEach(o => {
           state.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
         })
-        // Vue.delete(state.chatRoomUnread[chatRoomId], 'others')
+        Vue.delete(state.chatRoomUnread[chatRoomId], 'others')
       }
     }
   },
@@ -127,34 +127,17 @@ const mutations = {
     Vue.delete(state.chatRoomScrollPosition, chatRoomId)
   },
   setChatRoomReadUntil (state, { chatRoomId, messageHash, createdDate }) {
-    // eslint-disable-next-line no-lone-blocks
-    {
-      // TODO: need to remove this block after we release 0.2.*
-      if (!state.chatRoomUnread[chatRoomId].messages) {
-        Vue.set(
-          state.chatRoomUnread[chatRoomId],
-          'messages',
-          [].concat(
-            (state.chatRoomUnread[chatRoomId].mentions || []).map(m => Object.assign({ type: MESSAGE_TYPES.TEXT }, m)),
-            (state.chatRoomUnread[chatRoomId].others || []).map(o => Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
-          )
-        )
-      }
-      if (state.chatRoomUnread[chatRoomId].mentions) {
-        Vue.set(state.chatRoomUnread, chatRoomId, {
-          readUntil: { messageHash, createdDate, deletedDate: null },
-          mentions: [],
-          messages: state.chatRoomUnread[chatRoomId].messages
-            .filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
-        })
-        return
-      }
-    }
     Vue.set(state.chatRoomUnread, chatRoomId, {
       readUntil: { messageHash, createdDate, deletedDate: null },
       messages: state.chatRoomUnread[chatRoomId].messages
-        .filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime())
+        ?.filter(m => new Date(m.createdDate).getTime() > new Date(createdDate).getTime()) || []
     })
+    // eslint-disable-next-line no-lone-blocks
+    {
+      // hack: delete me after upgrade to 0.2.x!
+      Vue.set(state.chatRoomUnread[chatRoomId], 'mentions', [])
+      Vue.set(state.chatRoomUnread[chatRoomId], 'others', [])
+    }
   },
   deleteChatRoomReadUntil (state, { chatRoomId, deletedDate }) {
     Vue.set(state.chatRoomUnread[chatRoomId].readUntil, 'deletedDate', deletedDate)
@@ -163,18 +146,6 @@ const mutations = {
     state.chatRoomUnread[chatRoomId].messages.push({ messageHash, createdDate, type })
   },
   deleteChatRoomUnreadMessage (state, { chatRoomId, messageHash }) {
-    // eslint-disable-next-line no-lone-blocks
-    {
-      // TODO: need to remove this block after we release 0.2.*
-      if (state.chatRoomUnread[chatRoomId].mentions) {
-        Vue.set(
-          state.chatRoomUnread[chatRoomId],
-          'mentions',
-          state.chatRoomUnread[chatRoomId].mentions.filter(m => m.messageHash !== messageHash)
-        )
-      }
-    }
-
     Vue.set(
       state.chatRoomUnread[chatRoomId],
       'messages',
```

Silver-IT:
@taoeffect, I totally agree with your suggestion, which is quite clean and simplified. There could be a small difference between ours, but I am pretty sure we can ignore that change.
Thanks for your review and nice solution.

taoeffect (approved):
LGTM! Great job @Silver-IT! 🎉

Issue #1334, opened 2022-07-07
Bug in login code related to refresh during contract sync


taoeffect:
### Problem

In my testing of the live server login, I encountered a situation where login was taking too long. The white `Loading events from server...` banner appeared for a long time, so I refreshed the page.

Upon refresh, the group was no longer being joined, and the account I was logged in as was group-less.

When I re-added the account using the expired "anyone can join" invite link (which seemed to work because the group considered the account present), the account no longer had its income details specified. They also appeared in the "Members" list as a "NEW" member.

### Solution

This is really two separate issues, but I wanted to document both so that I wouldn't forget.

To recap, the two problems described:

1. Account leaving the group upon page refresh (this is the problem to fix for this issue)
2. Account losing income details (as a separate issue, figuring out why this happened, if this is a bug, and fixing it if so)

To fix the leaving group upon page refresh during login bug, I think requires modification to this code:

```js
  'gi.actions/identity/updateLoginStateUponLogin': async function () {
    const getters = sbp('state/vuex/getters')
    const state = sbp('state/vuex/state')
    const ourLoginState = generatedLoginState()
    const contractLoginState = getters.loginState
    try {
      if (!contractLoginState) {
        console.info('no login state detected in identity contract, will set it')
        await sbp('gi.actions/identity/saveOurLoginState')
      } else if (diffLoginStates(ourLoginState, contractLoginState)) {
        // we need to update ourselves to it
        // mainly we are only interested if the contractLoginState contains groupIds
        // that we don't have, and if so, join those groups
        const groupsJoined = difference(contractLoginState.groupIds, ourLoginState.groupIds)
        console.info('synchronizing login state:', { groupsJoined })
        for (const contractID of groupsJoined) {
          try {
            await sbp('gi.actions/group/join', { contractID, options: { skipInviteAccept: true } })
          } catch (e) {
            console.error(`updateLoginStateUponLogin: ${e.name} attempting to join group ${contractID}`, e)
            if (state.contracts[contractID] || state[contractID]) {
              console.warn(`updateLoginStateUponLogin: removing ${contractID} b/c of failed join`)
              try {
                await sbp('chelonia/contract/remove', contractID)
              } catch (e2) {
                console.error(`failed to remove ${contractID} too!`, e2)
              }
            }
          }
        }
      }
```

Specifically, there is clearly a call to `await sbp('chelonia/contract/remove', contractID)` there if an exception occurs.

[...]

PR #1603, opened 2023-05-22
Fix error in Joining group

### LINKED ISSUE #1575: Duplicate notifications when someone joins on both desktop and mobile
### Problem

When someone joins first on desktop, and then logs in on mobile, I got this duplicate notification.

Screen Shot 2023-05-04 at 9 19 18 PM

I also got the duplicate notification on my mobile device (where I was also logged in).

### Solution

Find and fix the problem.

### LINKED ISSUE #1579: Member loses their income details after logging in on mobile??
### Problem

A member logged into a group on desktop, entered their income details successfully and saved them.

They then logged in on the mobile device, and suddenly their income details were gone!

The messages this user sent to the database after updating their income details the first time included `'gi.contracts/group/inviteAccept'` (presumably from the mobile device), `'gi.contracts/group/joinChatRoom'` and `'gi.contracts/group/updateLastLoggedIn'`.

### Solution

Figure out why their income details got wiped. Keep in mind that during this time other users might have logged in as well and performed similar actions.

Silver-IT:
### Summary of Changes

- Closes #1575 
- Closes #1579

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -995,14 +995,10 @@ sbp('chelonia/defineContract', {
         if (Object.keys(invite.responses).length === invite.quantity) {
           invite.status = INVITE_STATUS.USED
         }
-        // NOTE: ensure `meta.username` is unique for the lifetime of the username
-        //       since we are making it possible for the same username to leave and rejoin the group.
-        //       All of their past posts will be re-associated with them upon re-joining.
-        const profile = state.profiles[meta.username]
-        if (profile && profile.status !== PROFILE_STATUS.REMOVED) {
-          console.error(`inviteAccept: ${meta.username} is currently active member in the group`)
-          throw new Errors.GIErrorIgnoreAndBan(`${meta.username} is already joined and active in the group`)
-        }
+        // TODO: ensure `meta.username` is unique for the lifetime of the username
+        //       since we are making it possible for the same username to leave and
+        //       rejoin the group. All of their past posts will be re-associated with
+        //       them upon re-joining.
         Vue.set(state.profiles, meta.username, initGroupProfile(meta.identityContractID, meta.createdDate))
         // If we're triggered by handleEvent in state.js (and not latestContractState)
         // then the asynchronous sideEffect function will get called next
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -995,10 +995,14 @@ sbp('chelonia/defineContract', {
         if (Object.keys(invite.responses).length === invite.quantity) {
           invite.status = INVITE_STATUS.USED
         }
-        // TODO: ensure `meta.username` is unique for the lifetime of the username
-        //       since we are making it possible for the same username to leave and
-        //       rejoin the group. All of their past posts will be re-associated with
-        //       them upon re-joining.
+        // NOTE: ensure `meta.username` is unique for the lifetime of the username
+        //       since we are making it possible for the same username to leave and rejoin the group.
+        //       All of their past posts will be re-associated with them upon re-joining.
+        const existingProfile = state.profiles[meta.username]
+        if (existingProfile && existingProfile.status !== PROFILE_STATUS.REMOVED) {
+          console.error(`inviteAccept: ${meta.username} is currently active member in the group`)
+          throw new Errors.GIErrorIgnoreAndBan(`${meta.username} is already joined and active in the group`)
+        }
         Vue.set(state.profiles, meta.username, initGroupProfile(meta.identityContractID, meta.createdDate))
         // If we're triggered by handleEvent in state.js (and not latestContractState)
         // then the asynchronous sideEffect function will get called next
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -998,8 +998,8 @@ sbp('chelonia/defineContract', {
         // NOTE: ensure `meta.username` is unique for the lifetime of the username
         //       since we are making it possible for the same username to leave and rejoin the group.
         //       All of their past posts will be re-associated with them upon re-joining.
-        const existingProfile = state.profiles[meta.username]
-        if (existingProfile && existingProfile.status !== PROFILE_STATUS.REMOVED) {
+        const profile = state.profiles[meta.username]
+        if (profile && profile.status !== PROFILE_STATUS.REMOVED) {
           console.error(`inviteAccept: ${meta.username} is currently active member in the group`)
           throw new Errors.GIErrorIgnoreAndBan(`${meta.username} is already joined and active in the group`)
         }
--- a/frontend/views/pages/Join.vue
+++ b/frontend/views/pages/Join.vue
@@ -19,8 +19,8 @@ div
         h1.is-title-1.c-title(data-test='groupName') {{ ephemeral.invitation.groupName }}
         p.has-text-1(data-test='invitationMessage') {{ ephemeral.invitation.message }}
       .card
-        signup-form(v-if='isStatus("SIGNING")' :postSubmit='accept')
-        login-form(v-else :postSubmit='accept')
+        signup-form(v-if='isStatus("SIGNING")' :postSubmit='acceptPostEffect')
+        login-form(v-else :postSubmit='acceptPostEffect')
 
       p.c-switchEnter(v-if='isStatus("SIGNING")')
         i18n Already have an account?
@@ -49,7 +49,7 @@ div
 import sbp from '@sbp/sbp'
 import { mapGetters, mapState } from 'vuex'
 import { INVITE_INITIAL_CREATOR, INVITE_STATUS } from '@model/contracts/shared/constants.js'
-import { LOGIN } from '@utils/events.js'
+import { LOGIN, JOIN_GROUP } from '@utils/events.js'
 import SignupForm from '@containers/access/SignupForm.vue'
 import LoginForm from '@containers/access/LoginForm.vue'
 import Loading from '@components/Loading.vue'
@@ -97,23 +97,11 @@ export default ({
   mounted () {
     // For some reason in some Cypress tests it loses the route query when initialized is called
     this.ephemeral.query = this.$route.query
-    if (syncFinished || !this.ourUsername) {
-      this.initialize()
-    } else {
-      sbp('okTurtles.events/once', LOGIN, () => this.initialize())
-    }
+    this.initialize()
   },
   methods: {
     async initialize () {
       try {
-        if (this.ourUsername) {
-          if (this.currentGroupId && this.$store.state.contracts[this.ephemeral.query.groupId]) {
-            this.$router.push({ path: '/dashboard' })
-          } else {
-            await this.accept()
-          }
-          return
-        }
         const state = await sbp('chelonia/latestContractState', this.ephemeral.query.groupId)
         const invite = state.invites[this.ephemeral.query.secret]
         if (!invite || invite.status !== INVITE_STATUS.VALID) {
@@ -150,6 +138,13 @@ export default ({
           message
         }
         this.pageStatus = 'SIGNING'
+
+        if (!syncFinished) {
+          sbp('okTurtles.events/once', LOGIN, () => this.accept())
+        } else {
+          this.pageStatus = 'LOGGING'
+          this.accept()
+        }
       } catch (e) {
         console.error(e)
         this.ephemeral.errorMsg = `${L('Something went wrong. Please, try again.')} ${e.message}`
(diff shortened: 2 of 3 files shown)


frontend/model/contracts/group.js
```diff
@@ -995,10 +995,14 @@ sbp('chelonia/defineContract', {
...
-        //       rejoin the group. All of their past posts will be re-associated with
-        //       them upon re-joining.
+        // NOTE: ensure `meta.username` is unique for the lifetime of the username
+        //       since we are making it possible for the same username to leave and rejoin the group.
+        //       All of their past posts will be re-associated with them upon re-joining.
+        const profile = state.profiles[meta.username]
+        if (profile && profile.status !== PROFILE_STATUS.REMOVED) {
+          console.error(`inviteAccept: ${meta.username} is currently active member in the group`)
+          throw new Errors.GIErrorIgnoreAndBan(`${meta.username} is already joined and active in the group`)
+        }
```
taoeffect: Nice job @Silver-IT considering this. However, I'm worried that this might get some members banned unintentionally if we make a mistake in the code...

For example, we could add additional statuses and forget to update other code, or they could be in `PENDING` status... or something else.

The thing is, once the `e2e-protocol` is implemented properly, that should take care of ensuring that all messages are sent by who they claim to be sent by. So do you think you can remove these updates?

taoeffect: OK, I see that this change is related to #1579. And indeed you are correct that calling `inviteAccept` twice is what resulted in the profile being reset (because the invite link was used twice).

I was able to track this down by using @snowteamer's `chel eventsSince` PR: https://github.com/okTurtles/chel/pull/9

For reference (for myself and others), this is the command on how to show all of the actions from the user `andrea` in the group:

```
chel eventsSince --limit 1000 groupincome.db 21XWnNWZzYYPfezAq9HN85b4by9TWt3o5GPUVzw9baSVjeQ49w 21XWnNWZzYYPfezAq9HN85b4by9TWt3o5GPUVzw9baSVjeQ49w | jq '.[] | select(.op[0] == "ae") | .op[1] | fromjson | select(.meta.username == "andrea")'
```

So, we still need to change this code because we don't want to ban Andrea for clicking the invite link twice. :P

Instead, if we are an existing user in the group, can we simply skip running the next line?

Silver-IT: @taoeffect, I have just removed that change in `group.js` contract file considering your opinion and one more reason that I think this can't be reached since it's prehandled in `Join.vue` page.

But to be honest, I am not 100% sure `it can't be reached` **at the moment**.
The meaning of **at the moment** is that I can say 100% only after we update the workflows of `Login`, `Join Group`, ... and resolve Issue #1334.

taoeffect (changes requested):
@Silver-IT The way `Join.vue` is written, especially how `mounted` calls `initialize`, is very important, and changing it can cause Cypress tests to break or introduce heisenbugs.

So I investigated this more deeply and found out that fixing these issues can be done with a 1-line change.

Change:

```pug
        login-form(v-else :postSubmit='accept')
```

To:

```pug
        login-form(v-else :postSubmit='goHome')
```

Alternatively, you can create a `goDashboard` function to directly go to the `/dashboard` (this would be better IMO).

Silver-IT:
@taoeffect, actually we can not set `postSubmit` of `Login` form to `goHome` as constant. That's because users should join the group after he logs in when he hasn't joined **this group**(the group that the invite link indicates to) yet. So after login,
- We should let users join the group if they have **not** already joined **this group**.
- We should let users go `Dashboard` page if they have already joined **this group**.

And `accept` action is not being called inside the `Login` or `SignUp` form. It's called by these codes.
```
if (!syncFinished) {
  this.pageStatus = 'SIGNING'
  sbp('okTurtles.events/once', LOGIN, () => this.accept())
} else {
  this.accept()
}
```
So, actually we don't need to take care of calling `accept` function in `postSubmit` of two forms.

The main reason I added `postSubmit` to the `Login` and `SignUp` forms is that it was weird when the spinner didn't spin even though `Submit` action was in progress. We used `ButtonSubmit` component in both forms and the spinner of `ButtonSubmit` component should spin and not disappear until the `Submit` action would be finished.
In `Login` and `SignUp` form inside `Join` page, the `Submit` action is not only `Login` or `SignUp` but also `Joining group`. So the spinner should spin until the user finishes joining the group.

taoeffect (changes requested):
Ah, ok you're right that we should join the group after login if the user has not already joined the group.

So please send a minimal PR that does that. The only file that needs to be modified is `Join.vue`, and the only function that needs to be modified is `accept` - add a check to see if `groupId` already exists: `this.$store.state.contracts[groupId]`, and if so proceed to the dashboard instead of calling `'gi.actions/group/joinAndSwitch'`.

Silver-IT (commented):
@taoeffect, I have reviewed this PR and no improvements needed I think.

frontend/views/pages/Join.vue
```diff
@@ -97,23 +97,11 @@ export default ({
...
-      this.initialize()
-    } else {
-      sbp('okTurtles.events/once', LOGIN, () => this.initialize())
-    }
+    this.initialize()
   },
   methods: {
     async initialize () {
       try {
-        if (this.ourUsername) {
```
Silver-IT, line 109: This doesn't mean `Login` is finished. We can not say he is part of the group (or not) before he finishes to login and completes to sync the contract.

frontend/views/pages/Join.vue
```diff
@@ -97,23 +97,11 @@ export default ({
   mounted () {
     // For some reason in some Cypress tests it loses the route query when initialized is called
     this.ephemeral.query = this.$route.query
-    if (syncFinished || !this.ourUsername) {
```
Silver-IT: Why should we take care of `this.ourUsername`? To say if he is logged in?
Even though `this.ourUsername` has its value and isn't **undefined**, the user could be in progress to log in and to sync contracts.

frontend/views/pages/Join.vue
```diff
@@ -97,23 +97,11 @@ export default ({
...
-      this.initialize()
-    } else {
-      sbp('okTurtles.events/once', LOGIN, () => this.initialize())
-    }
```
Silver-IT: We don't need to wait validating the invitation link until the user finishes login, since we always do validation no matter he has logged in or not. We can immediately validate if he opens this link.

frontend/views/pages/Join.vue
```diff
@@ -97,23 +97,11 @@ export default ({
...
+    this.initialize()
   },
   methods: {
     async initialize () {
       try {
-        if (this.ourUsername) {
-          if (this.currentGroupId && this.$store.state.contracts[this.ephemeral.query.groupId]) {
-            this.$router.push({ path: '/dashboard' })
-          } else {
-            await this.accept()
```
Silver-IT, line 113: We shouldn't call this `accept` function **here** for the following two reasons.

- The link is not yet validated
- At this time, we can not say the user is not part of this group. Because he could be in progress to log in and to sync contracts.

frontend/views/pages/Join.vue
```diff
@@ -149,7 +137,13 @@ export default ({
...
           message
         }
-        this.pageStatus = 'SIGNING'
+
+        if (!syncFinished) {
+          this.pageStatus = 'SIGNING'
+          sbp('okTurtles.events/once', LOGIN, () => this.accept())
+        } else {
+          this.accept()
+        }
```
Silver-IT: `accept` function should be called once only when the validation is passed and also the user finishes logging in.

frontend/views/pages/Join.vue
```diff
@@ -149,7 +137,13 @@ export default ({
           creatorPicture,
           message
         }
-        this.pageStatus = 'SIGNING'
```
Silver-IT: No need to set `this.pageStatus = 'SIGNING'`. We should only set this when the user is not logged in.
So replace this to the next line below inside the `if` statement.

frontend/views/pages/Join.vue
```diff
@@ -163,10 +157,13 @@ export default ({
       this.$router.push({ path: '/' })
     },
     async accept () {
+      this.pageStatus = 'LOADING'
```
Silver-IT: This line should be removed. I will remove this soon.

frontend/views/pages/Join.vue
```diff
@@ -163,10 +157,13 @@ export default ({
...
       this.ephemeral.errorMsg = null
+      const { groupId, secret } = this.ephemeral.query
+      if (this.currentGroupId && this.$store.state.contracts[groupId]) {
+        return this.$router.push({ path: '/dashboard' })
+      }
```
Silver-IT: This is for checking if the user is joined the group or not.

frontend/views/pages/Join.vue
```diff
@@ -176,7 +173,14 @@ export default ({
...
         this.ephemeral.errorMsg = e.message
         this.pageStatus = 'INVALID'
+      } finally {
+        sbp('okTurtles.events/emit', JOIN_GROUP)
       }
+    },
+    acceptPostEffect () {
+      return new Promise(resolve => {
+        sbp('okTurtles.events/once', JOIN_GROUP, resolve)
+      })
```
Silver-IT: `accpetPostEffect` is for waiting until the user finishes joining group.
If we use `accept` function in the `postSubmit`, the `accept` function would be called twice.

As we can see in `initialize` function, `accept` function is called after the user finishes logging in.
Users could log in by clicking `Sign Up` or `Login` button. So if we call `accept` function in `postSubmit` of those two forms, it will be called twice; once inside `initialize`, once in `postSubmit`.

Silver-IT (commented):
@taoeffect, I have described what they do in detail. Need your replies. Thanks

frontend/views/pages/Join.vue
taoeffect: If `this.ourUsername` is defined it does mean login has finished.

That is what this code in `mounted` does:

```js
    if (syncFinished || !this.ourUsername) {
      this.initialize()
    } else {
      sbp('okTurtles.events/once', LOGIN, () => this.initialize())
    }
```

frontend/views/pages/Join.vue
taoeffect: This code from `main.js` is run before `mounted` gets called:

```js
await sbp('gi.actions/identity/login', { username })
```

So if `this.ourUsername` is not defined, it means the user is not logged in, and won't be logged in when `mounted` is called.

frontend/views/pages/Join.vue
Silver-IT: Yes. It's correct for the previous `mounted` code, but not for the new `mounted` code.

frontend/views/pages/Join.vue
Silver-IT: That's correct. However, the whole `if-else` state isn't needed I think.

frontend/views/pages/Join.vue
taoeffect: Then you can re-arrange the code so that it looks like this:

```js
    async initialize () {
      try {
        const state = await sbp('chelonia/latestContractState', this.ephemeral.query.groupId)
        const invite = state.invites[this.ephemeral.query.secret]
        if (!invite || invite.status !== INVITE_STATUS.VALID) {
          console.error('Join.vue error: Link is not valid.')
          this.ephemeral.errorMsg = L('You should ask for a new one. Sorry about that!')
          this.pageStatus = 'INVALID'
          return
        } else if (invite.expires < Date.now()) {
          console.log('Join.vue error: Link is already expired.')
          this.ephemeral.errorMsg = L('You should ask for a new one. Sorry about that!')
          this.pageStatus = 'EXPIRED'
          return
        }
        if (this.ourUsername) {
          if (this.currentGroupId && this.$store.state.contracts[this.ephemeral.query.groupId]) {
            this.$router.push({ path: '/dashboard' })
          } else {
            await this.accept()
          }
          return
        }
```

frontend/views/pages/Join.vue
taoeffect: See my comment above: https://github.com/okTurtles/group-income/pull/1603#discussion_r1206179653

frontend/views/pages/Join.vue
taoeffect: See my comment above: https://github.com/okTurtles/group-income/pull/1603#discussion_r1206179653

frontend/views/pages/Join.vue
taoeffect: With the code like this https://github.com/okTurtles/group-income/pull/1603#discussion_r1206179653

Then it makes sense to have this.

Silver-IT (commented):
@taoeffect, ready for your FINAL review. :)

taoeffect (approved):
Approved! Nicely done @Silver-IT 👍

PR #1625, opened 2023-06-01
#1623 - Allow chat width to expand on desktop

### LINKED ISSUE #1623: Allow chat width to expand on desktop
### Problem

We should be able to take advantage of big monitors to display more of the chat, but at the moment the width of the chat view is restricted to a maximum with on Desktop:

Screen Shot 2023-05-31 at 2 20 06 PM

### Solution

On desktop-only, allow the chat view to expand. Keep about the same amount margin to the right as there is to the left nav bar. Whatever looks good.

SebinSong:
closes #1623

### DIFF
--- a/frontend/views/pages/GroupChat.vue
+++ b/frontend/views/pages/GroupChat.vue
@@ -222,6 +222,8 @@ export default ({
 
   .p-main {
     height: auto !important;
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Excellent!

PR #1628, opened 2023-06-02
2nd chunk of #400 - Poll

SebinSong:
This chunk of work contains implementation of below items

- Voting on the poll
- Changing previous votes on the poll
- Expiration of the poll
- Building all UI components that are related to above.

https://github.com/okTurtles/group-income/assets/17641213/a1ceadc1-f561-48c9-ac05-37a6f8e51b1b


frontend/model/contracts/shared/constants.js
```diff
@@ -51,6 +51,7 @@ export const CHATROOM_NAME_LIMITS_IN_CHARS = 50
 export const CHATROOM_DESCRIPTION_LIMITS_IN_CHARS = 280
 export const CHATROOM_ACTIONS_PER_PAGE = 40
 export const CHATROOM_MAX_ARCHIVE_ACTION_PAGES = 2 // 2 pages of actions
+export const CHATROOM_POLL_DURATION_DAYS = 7 // TODO: discuss with team and confirm this value
```
SebinSong: For now, a poll expires 7 days after its creation. '7' here is just a random number I chose for development and let me know what this should be. will tweak this value.

frontend/model/contracts/shared/constants.js
```diff
@@ -118,7 +120,6 @@ export const POLL_TYPES = {
 }
 
 export const POLL_STATUS = {
-  ACTIVE: 'active', // the poll is open
-  CLOSED: 'closed', // everyone has voted on the poll
-  EXPIRED: 'expired' // not everyone has voted on the poll but it has expired
```
SebinSong, line 124: While developing, I realised it doesn't make sense to close a poll as soon as everyone has voted because voters should be able to change their votes for some time until it expires. so the status values are either `ACTIVE` & `CLOSED` here now.

frontend/views/containers/chatroom/MessagePoll.vue
```diff
@@ -82,6 +99,58 @@ export default ({
...
+        } else {
+          this.expirationTimeoutId = setTimeout(checkAndSetTimer, MINS_MILLIS)
+        }
+      }
+      const timeDiff = Date.now() - this.pollData.expires_date_ms
+
+      if (timeDiff >= 0) {
+        // if the poll has been expired, mark it 'closed' immediately.
+        markPollClosed()
+      } else if (Math.abs(timeDiff) < DAYS_MILLIS) {
```
SebinSong, line 139: I defined **'poll expiring soon'** in below comment as **'within 24hrs'** and This is just an arbitrary period that I set for development. pls let me know if this should be other value. will make a tweak.
As I mentioned in below comment too, periodic check & expiring something is a good pattern to be exported into `mainNotificationMixin.js` entry, but currently there is a challenge of accessing a particular message outside `ChatMain.vue`. so implementing this logic here.

frontend/views/containers/chatroom/poll-message-content/PollToVote.vue
```diff
@@ -0,0 +1,131 @@
...
+  },
+  methods: {
+    submitVotes () {
+      sbp('gi.actions/chatroom/voteOnPoll', {
+        contractID: this.currentChatRoomId,
+        data: {
+          hash: this.messageHash,
+          votes: this.allowMultipleChoices ? this.form.selectedOptions : [this.form.selectedOptions]
+        }
+      })
```
taoeffect: Could you please make sure to catch all errors (here and below on line 75, and anywhere else) so that user know if there was an issue voting on the poll / performing some action?

frontend/views/containers/chatroom/MessagePoll.vue
```diff
@@ -82,6 +99,58 @@ export default ({
...
+    },
+    switchOffChangeMode () {
+      this.ephemeral.isChangeMode = false
+    },
+    checkAndSetupPollExpirationTimer () {
+      const markPollClosed = () => {
+        sbp('gi.actions/chatroom/closePoll', {
+          contractID: this.currentChatRoomId,
+          data: { hash: this.messageHash }
+        })
```
taoeffect: Please catch any errors and report to user

Silver-IT:
Great work, @SebinSong. Just a partial review done.
1. The empty space should be removed.
image
2. Should be single quotation mark, instead of double quotation mark.
image
3. Display name should be used instead of username, if it exists.
image
4. Emoticon should be selected if the current user reacted.
image
5. Messages should be same for all the users.
I can see that messages for `Poll` aren't rendered as soon as login, not only signup.
image

Silver-IT:
@SebinSong, another partial review done.
1. Should be `Reply` button to the polls.
image
2. Should deal with long messages, even truncate.
image
3. Can not find `Cancel` button when click `Change vote` button. [EDITED]
image

SebinSong:
@Silver-IT 
> 3. Can not view votes once click Change vote button and make no changes.

I don't think this is absolutely necessary. There is no need for a user to see other people's vote while changing their votes. Normally, we make up our mind first before going ahead and clicking `change votes`, right? (Plus, there is no design requirement particularly mentioning this in Figma)

This is my only objection. will make fixes for others later.

Silver-IT:
> I don't think this is absolutely necessary. There is no need for a user to see other people's vote while changing their votes. Normally, we make up our mind first before going ahead and clicking change votes, right?

I wanted to change my vote and click `Change vote` button. But suddenly changed my mind and decided to keep the current vote. But I could not find `Cancel` button.

SebinSong:
@Silver-IT Then the comment should've mentioned `There is a 'cancel' button missing` instead. agree that cancel btn is needed.

Silver-IT:
> @Silver-IT Then the comment should've mentioned `There is a 'cancel' button missing` instead. agree that cancel btn is needed.

Sorry for making you confused. :). Updated the comment.

taoeffect (changes requested):
I had started my review but then had to go - I'm leaving this partial review as it's better than nothing!

SebinSong:
@Silver-IT Worked on feedbacks.

> Should be Reply button to the polls.

The design doesn't have this requirement. once they decide to have this feature in, will work on adding it then. (Pls make sure a request to add a feature is based on design requirements or discussion, as opposed to one's preference.)

> Messages should be same for all the users.
I can see that messages for Poll aren't rendered as soon as login, not only signup.

I was not able to reproduce this issue you pointed out. every types of messages (`MESSAGE_TYPES.POLL` , `MESSAGE_TYPES.NOTIFICATION` informing member voting) were consistently rendered to all members, even for a member that joined after a poll has been created/active.

Worked on the rest of the feedbacks. great job spotting them.

SebinSong:
@taoeffect Updated the PR with the work for your feedbacks.

Silver-IT:
@SebinSong, great work in resolving feedback.

> The empty space should be removed.

Fixed

> Should be single quotation mark, instead of double quotation mark.

Fixed

> Display name should be used instead of username, if it exists.

Fixed

> Emoticon should be selected if the current user reacted.

Fixed

> Messages should be same for all the users.
I can see that messages for Poll aren't rendered as soon as login, not only signup.

Not yet fixed. Please try to make many messages so that there are more than 2 pages of messages in total.

> Should be Reply button to the polls.

Okay. Since there is no design for this, we can leave it for now.

> Should deal with long messages, even truncate.

Fixed

> Can not find Cancel button when click Change vote button. [EDITED]

Fixed

Silver-IT:
@SebinSong, another partial review done.
1. The `Title` input element in `New poll` popup isn't cleared after create a poll.
2. The `New poll` popup isn't closed when click the sidebar.
3. Display name should be truncated too.
image

Totally, it's awesome.

SebinSong:
@Silver-IT 
> Messages should be same for all the users. I can see that messages for Poll aren't rendered as soon as login, not only signup.

can you post a video demonstrating the issue? it's actually not clear what you are describing by this sentence.

>  Okay. Since there is no design for this, we can leave it for now.

Yes, I don't think it's something you are entitled to decide.. when Greg decides we want that feature and need design for it, I will work on it.

> The New poll popup isn't closed when click the sidebar.

Again, this sentence is not clear. when 'Create Poll' pop-up is open and I click outside the pop-up. it's closed.


And I actually would love to make this clear here. when you review and request me to add something, please make sure it's based on 1) design requirements or 2) existing discussions involving Greg or Andrea. (e.g. their comments saying we need those features).
I have noticed a few times from your review that I get requested to add something but the reason is neither 1) nor 2), Which means they are based on 'your own' preference. I don't think it's right for me to develop something just because another developer says so. It always has to be either the product owner's decision or the group decision involving their opinion.

SebinSong:
- The Title input element in New poll popup isn't cleared after create a poll.
- Display name should be truncated too.

have been fixed. 
@taoeffect @Silver-IT

Silver-IT:
@SebinSong, please have a look at this video for two issues.

https://github.com/okTurtles/group-income/assets/57976479/4187b010-81e2-4b1b-bdbb-764db2a4709d

Silver-IT:
@SebinSong, I think some of my reviews could be out of requirement and your expectations. I will try to restrict my review to the requirement.

But we have our user experience. We can guess what should be there, where should it be, what should be more.
For example, even though the Issue describes the UI error in mobile, we should take care of tablet and desktop while we are working on that issue. Of course, we could not think about it but I think we should agree to we should take care of it.
Just like the animation you added in Chelonia Dashboard. If we think it's better and it should be, we should share our opinions. We are not the owner of the project, but we are all contributors.

For the `Reply` button, it is not described in the Figma design, so it's true that we should not make it before we make our decisions to make it.

SebinSong:
@Silver-IT  Thanks for uploading the video.

Now I can see what you meant by below.

> The New poll popup isn't closed when click the sidebar.

And I also think this isn't absolutely necessary either. there is nothing blocking user to take any wanted action on the sidebar while the poll pop-up is open. so objecting it too.

> Messages should be same for all the users.
I can see that messages for Poll aren't rendered as soon as login, not only signup.

This is something I still haven't been able to encounter. Judging by the video, for me it looks more like an issue with the connection than an issue with implementation of poll in the code-base. If I can see it happen and can spot a cause in the code will make a fix.

SebinSong:
@Silver-IT 
For your second comment, I think you got my point wrong.

> Of course, we could not think about it.

This is absolutely not what I meant.
We all can have our opinion on what's better and how things can be improved. But, when you say, "I think this is right. so I tell you to work on it", it doesn't involve confirmation & agreement of Greg is what I meant to say. It's good that you care about user experience as I always do too, but I think you should bring it up to a discussion first that involves Greg instead of telling me you should work on it straight away. Working on something just because another coworker says 'this is right' is something I naver have done in my career. I always listen to my manager. at least this is a work ethic I believe is right to do..
(EDITTED: unless it's something that's so obvious that we all just know anyone will agree, just like adding the `Cancel` button we discussed before)

Animations on Chelonia dashboard is I something I put extra effort on but I literally put it to review of Greg first by sending a PR. If he had found anything inappropreate about it, then they could've been taken down of course, no doubt.

Appreciate your review anyways

SebinSong:
> Messages should be same for all the users.
I can see that messages for Poll aren't rendered as soon as login, not only signup.

@Silver-IT 
Been trying to reproduce this issue for about an hour and haven't been able to encounter this issue even after manually setting up a slow internet connection.. So dismissing this request too for now. Once the change goes live and the issue is spotted again with clearer root cause, we can come back to this discussion again.

https://github.com/okTurtles/group-income/assets/17641213/d4a702e9-9c00-412c-b7d5-9f584b579351

frontend/model/contracts/shared/constants.js
taoeffect: Most apps allow the user to specify the duration. If this wasn't in the designs... well, maybe add a dropdown where the user can select how many days the post expires in, from 1 day to 30 days. Can be done as part of a separate PR too, based on what you think looks good (with the possibility of having to redesign it if Leihla prefers something else), or waiting until after @leihla designs it. I'll leave the decision up to you @SebinSong.

taoeffect: Added a comment for @leihla on Figma regarding this: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN?node-id=23766:265445#468454194

SebinSong: @taoeffect Sure, I would love to work after @leihla 's design update on this.

Silver-IT:
@SebinSong 

**Steps to reproduce**
- user1, user2 are part of groupA
- user1 creates a poll and votes on it
- user1 creates 50 dummy messages (e.g. "1" to "50")
- user1 changes his vote
- user2 logs in using another device
- user2 goes to the chat page and sees the differene

Additionally, I got this weird duplication while I was trying to follow the above steps.
image

Silver-IT:
@SebinSong, for the `"Jump to poll"`, you can use the function `scrollToMessage` which you can find in `ChatMain.vue`.

SebinSong:
@Silver-IT 

> for the "Jump to poll", you can use the function scrollToMessage which you can find in ChatMain.vue.

Thanks for letting know, I can work on making this  work in the following PR (which will likely be the final PR for the Poll functionality)

SebinSong:
@Silver-IT 
I just followed through steps you wrote up above. but couldn't run into the same issue... (I opened a browse on my phone as a 'different device' in there)
Btw, can you explain the reason why you think that issue is particularly brought in by this specific PR..? Don't you think it's an issue more related to the performance of chatroom when loading large amount of messages rather than it's the poll message causing the problem? 
cuz I think this problem(loss of some messages when loading large amount) can/should be handled separately from this PR and it's more than welcome if you want to look into it, as it looks like you have worked on those part in `GroupChat.vue` before.

Silver-IT:
@SebinSong, I don't think it's related to the performance issue. I think it's related to the rendering first page. Many messages is what I meant for the 2 or 3 pages of messages so that the `Load More` could work.

I think we can think about this more until @taoeffect finishes the review on this PR.

SebinSong:
@Silver-IT Btw, I'm sorry I came off a bit emotional while trying to make my point during our discussion on this PR. I would love to say, most of all I thank your eyes & expertise for this PR.

Silver-IT:
Thanks, @SebinSong. I think I did too. Actually, I totally agree with your work ethic, and that's the right thing.
Also please excuse me for my unintended behavior that made you that emotional. :)

SebinSong:
@Silver-IT 
>  I don't think it's related to the performance issue. I think it's related to the rendering first page. Many messages is what I meant for the 2 or 3 pages of messages so that the Load More could work.

Yes, as there is no clue yet that this issue is particularly related to `MESSAGE_TYPE.POLL`, I think this should be handled as a separate issue. we can talk more about it after Greg's review.

Silver-IT:
@SebinSong

> Yes, as there is no clue yet that this issue is particularly related to MESSAGE_TYPE.POLL, I think this should be handled as a separate issue. we can talk more about it after Greg's review.

Okay. However, I hope to have time to have a deep look before @taoeffect finishes review.

SebinSong:
Yes sure, @Silver-IT Thanks for that. Unfortunately I'm not very familiar with those area yet so, please let me know if you spot the root cause while digging there yourself. will work on fixing that.

frontend/model/contracts/shared/constants.js
dotmacro: > For now, a poll expires 7 days after its creation.

The current design on Figma specifies that polls close after 2 months by default, so this should be `60` instead of `7`. (The note is on a later frame, here: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=23766%3A266542&t=B3OUjrU5YYYuhjZT-1)

SebinSong: updated the PR [here](https://github.com/okTurtles/group-income/pull/1628/commits/284c83fd6d9da5e88c27fef708f599c26e7431df) per @dotmacro 's comment

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,105 @@ sbp('chelonia/defineContract', {
...
+    },
+    'gi.contracts/chatroom/voteOnPoll': {
+      validate: objectOf({
+        hash: string,
+        votes: arrayOf(string)
+      }),
+      process ({ data, meta, hash, id }, { state }) {
+        const msgIndex = findMessageIdx(data.hash, state.messages)
+
+        if (msgIndex >= 0) {
```
Silver-IT, line 538: @SebinSong, this line is the reason of the error I reported.
`state.messages` could not have the message with that `hash`. Because `state.messages` doesn't save all the messages created from the beginning. I remember you couldn't see any messages in vuex contract state.
Chatroom contract creates messages only when `onlyRenderMessages` attribute is `true`, only inside Chat page.

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,105 @@ sbp('chelonia/defineContract', {
       sideEffect ({ contractID, hash }) {
         emitMessageEvent({ contractID, hash })
       }
+    },
+    'gi.contracts/chatroom/voteOnPoll': {
+      validate: objectOf({
+        hash: string,
+        votes: arrayOf(string)
+      }),
+      process ({ data, meta, hash, id }, { state }) {
```
Silver-IT, line 531: Need to create messages only when `onlyRenderMessages` is `true`. Please add the follow lines just below the `process` function just like `addMessage/process` function.
```
if (!state.onlyRenderMessage) {
  return
}
```

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,105 @@ sbp('chelonia/defineContract', {
...
+      sideEffect ({ contractID, hash }) {
+        emitMessageEvent({ contractID, hash })
+      }
+    },
+    'gi.contracts/chatroom/changeVoteOnPoll': {
+      validate: objectOf({
+        hash: string,
+        votes: arrayOf(string)
+      }),
+      process ({ data, meta, hash, id }, { state }) {
```
Silver-IT, line 575: Need to create messages only when `onlyRenderMessages` is `true`. Please add the follow lines just below the `process` function just like `addMessage/process` function.
```
if (!state.onlyRenderMessage) {
  return
}
```

Silver-IT (changes requested):
@SebinSong, just did a very partial review, for the error I reported.
I found the reason and added the comments. Will review more later this week.

SebinSong:
@Silver-IT Updated the PR with the fix related to `onlyRenderMessages`.

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,117 @@ sbp('chelonia/defineContract', {
...
+            MESSAGE_NOTIFICATIONS.VOTE_ON_POLL,
+            { votedOptions: votedOptNames.join(', ') }
+          )
+          const newMessage = createMessage({ meta, hash, id, data: notificationData, state })
+          state.messages.push(newMessage)
+        }
+      },
+      sideEffect ({ contractID, hash }) {
+        emitMessageEvent({ contractID, hash })
+      }
```
Silver-IT: Since it function creates new message, should call the function bellow after `emitMessageEvent` function.
You can check `addMessage` function as reference.
```
setReadUntilWhileJoining({ contractID, hash, createdDate: meta.createdDate })
```

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,117 @@ sbp('chelonia/defineContract', {
...
+            MESSAGE_NOTIFICATIONS.CHANGE_VOTE_ON_POLL,
+            { votedOptions: votedOptNames.join(', ') }
+          )
+          const newMessage = createMessage({ meta, hash, id, data: notificationData, state })
+          state.messages.push(newMessage)
+        }
+      },
+      sideEffect ({ contractID, hash }) {
+        emitMessageEvent({ contractID, hash })
+      }
```
Silver-IT: This too as above.
```
setReadUntilWhileJoining({ contractID, hash, createdDate: meta.createdDate })
```

frontend/model/contracts/chatroom.js
Silver-IT: Even though `msgIndex === -1`, `NOTIFICATION` message should be added.

```
if (msgIndex >= 0) {
  ...
  Vue.set(state.messages[msgIndex], 'pollData', { ...pollData, options: optsCopy })
}
...
state.messages.push(newMessage)
```

frontend/model/contracts/chatroom.js
```diff
@@ -520,6 +521,117 @@ sbp('chelonia/defineContract', {
...
+        votes: arrayOf(string)
+      }),
+      process ({ data, meta, hash, id }, { state }) {
+        if (!state.onlyRenderMessage) {
+          return
+        }
+
+        const msgIndex = findMessageIdx(data.hash, state.messages)
+
+        if (msgIndex >= 0) {
```
Silver-IT, line 582: Even though `msgIndex === -1`, `NOTIFICATION` message should be added.

```
if (msgIndex >= 0) {
  ...
  Vue.set(state.messages[msgIndex], 'pollData', { ...pollData, options: optsCopy })
}
...
state.messages.push(newMessage)
```

frontend/views/containers/chatroom/MessagePoll.vue
```diff
@@ -9,36 +9,41 @@ message-base(
...
+    MessageBase,
+    PollToVote,
+    PollVoteResult,
+    BannerScoped
   },
   props: {
     pollData: Object, // { creator, status, question, options ... }
     messageId: String,
+    messageHash: String,
+    currentUsername: String,
```
Silver-IT, line 46: Is this `currentUsername` used in this component?

frontend/views/containers/chatroom/MessagePoll.vue
Silver-IT: Yeah. The solution using `mainNotificationMixin.js` would be nice. But the problem is that we don't store any messages (including polls data) in contract state and we don't have any `getters` either.

So I think it's also fine to close the poll here in this part.

frontend/views/containers/chatroom/poll-message-content/VoterAvatars.vue
```diff
@@ -0,0 +1,251 @@
...
+    showTooltip () {
+      if (!this.ephemeral.isTooltipActive) {
+        this.ephemeral.isTooltipActive = true
+      }
+    },
+    closeTooltip () {
+      if (this.ephemeral.isTooltipActive) {
+        this.ephemeral.isTooltipActive = false
+      }
+    },
```
Silver-IT: If I were you, I will remove the `if` statements and reduce the LOC, and create only one function `setToolTip` instead of two functions.

This is fine too. Anyway, it's up to you.

Silver-IT (changes requested):
@SebinSong, Review done. Great work!!!

frontend/views/containers/chatroom/MessagePoll.vue
SebinSong: @Silver-IT Yes, this has to be explicitly specified here so that it's part of `$props` attribute used in its template. This was a fix for the issue where the current user's emoticon reaction is not highlighted.

frontend/views/containers/chatroom/poll-message-content/VoterAvatars.vue
SebinSong: @Silver-IT That's a simpler way but we are using `v-on-clickaway` 3rd-party directive here and this requires a single component method to be passed as the directive argument. Also, specifying an inline action / arrow function like `setTooltipActive(false)`, `() => setTooltipActive(false)` leads to a console error too. will remove the `if` conditionals btw.

frontend/model/contracts/chatroom.js
SebinSong: @Silver-IT 
updated per your comment. Btw,

> Chatroom contract creates messages only when onlyRenderMessages attribute is true, only inside Chat page.

Where exactly is this happening in the `GroupChat.vue`? I've been searching the source-code trying to locate the part where this takes place but couldn't easily spot it.

SebinSong:
@Silver-IT 
Updated the PR for your feedbacks again.

frontend/model/contracts/chatroom.js
Silver-IT: Since there could be really many messages in the chatroom contract, we couldn't save those messages inside contract. So we creates messages only when `onlyRenderMessages` attribute is set as true.

Silver-IT (changes requested):
Nice fix, @SebinSong!
Additionally I don't think you ran `grunt pin:0.1.5` with the latest contract source code. Please check it and make a commit if it's right.

SebinSong:
@taoeffect Updated the PR with the work of adding a select menu for the poll duration.

frontend/model/contracts/chatroom.js
taoeffect: @SebinSong 
> Where exactly is this happening in the `GroupChat.vue`?

`onlyRenderMessage` is set in `ChatMain.vue`

frontend/views/containers/chatroom/CreatePoll.vue
```diff
@@ -139,19 +161,29 @@ export default {
...
-      if (!element) {
+      if (!element && !blurredByDurationSelect) {
         this.close()
       }
     },
     optionPlaceholder (index) {
       return `${L('Option')} ${index}`
     },
-    addOption () {
+    addOption (e) {
```
taoeffect, line 173: Above this on line 171 it says:

```js
return `${L('Option')} ${index}`
```

My guess is this is saying something like "Option 1". This needs to be rewritten to include the number in the translated string, because in some languages it might be appropriate to write it like so: "1 Option".

So, something like:

```js
return L('Option {index}', { index })
```

frontend/views/containers/chatroom/MessagePoll.vue
```diff
@@ -82,6 +104,64 @@ export default ({
...
+    switchOffChangeMode () {
+      this.ephemeral.isChangeMode = false
+    },
+    checkAndSetupPollExpirationTimer () {
+      const markPollClosed = () => {
+        try {
+          sbp('gi.actions/chatroom/closePoll', {
+            contractID: this.currentChatRoomId,
+            data: { hash: this.messageHash }
+          })
```
taoeffect, line 120: This `try` does not catch any errors because there is no `await`

frontend/views/containers/chatroom/poll-message-content/PollToVote.vue
```diff
@@ -0,0 +1,173 @@
...
+    submitVotes () {
+      try {
+        sbp('gi.actions/chatroom/voteOnPoll', {
+          contractID: this.currentChatRoomId,
+          data: {
+            hash: this.messageHash,
+            votes: this.allowMultipleChoices ? this.form.selectedOptions : [this.form.selectedOptions],
+            votesAsString: `"${this.selectedVotesAsString}"`
+          }
+        })
```
taoeffect, line 88: This `try` does not catch any errors because there is no `await`

frontend/views/containers/chatroom/poll-message-content/PollToVote.vue
```diff
@@ -0,0 +1,173 @@
...
+      } catch (e) {
+        console.error('"voteOnPoll" action failed: ', e)
+        this.$refs.errBanner.danger(e.message)
+      }
+    },
+    async changeVotes () {
+      try {
+        await sbp('gi.actions/chatroom/changeVoteOnPoll', {
```
taoeffect, line 96: This `try` works though 🙂 👍

frontend/views/components/modal/ModalClose.vue
```diff
@@ -1,7 +1,7 @@
 <template lang='pug'>
   button.c-modal-close(
     :class='{ backOnMobile, fullscreen }'
-    @click.self='$emit("close")'
+    @click.self.stop='$emit("close")'
```
taoeffect, line 4: @SebinSong just curious if you remember what this change was for?

frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -61,18 +63,28 @@ export default ({
...
+        MESSAGE_NOTIFICATIONS.CHANGE_VOTE_ON_POLL
+      ].includes(this.notification.type)
     }
   },
   methods: {
     addEmoticon (emoticon) {
       this.$emit('add-emoticon', emoticon)
+    },
+    jumpToPoll () {
+      alert('TODO: implement "jumping to a particular message."')
```
taoeffect, line 87: ~~Could you open an issue for this?~~

frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -61,18 +63,28 @@ export default ({
           [MESSAGE_NOTIFICATIONS.UPDATE_DESCRIPTION]:
             L('Updated the channel description to: {description}', { description: channelDescription }),
           [MESSAGE_NOTIFICATIONS.DELETE_CHANNEL]: L('Deleted the channel: {title}', { title: channelName }),
-          [MESSAGE_NOTIFICATIONS.VOTE]: L('Voted on “{title}”', { title: '' }) // TODO: polls are not implemented yet
+          [MESSAGE_NOTIFICATIONS.VOTE_ON_POLL]: L('Voted on {options}', { options: votedOptions }),
+          [MESSAGE_NOTIFICATIONS.CHANGE_VOTE_ON_POLL]: L('Changed votes to {options}', { options: votedOptions })
```
taoeffect, line 67: Screen Shot 2023-07-25 at 11 14 17 AM

Probably should read "Changed vote" (singular)

taoeffect: Context:

Screen Shot 2023-07-25 at 11 15 00 AM

frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -2,7 +2,9 @@
 message-base(v-bind='$props' @add-emoticon='addEmoticon($event)')
   template(#body='')
     .c-notification
-      p.c-text(v-if='message') {{message.text}}
+      p.c-text(v-if='message')
+        | {{message.text}}
+        i18n.c-link(v-if='isPollNotification' @click='jumpToPoll') Jump to poll
```
taoeffect, line 7: This should probably be commented out if it's not functional.

frontend/views/containers/chatroom/MessageNotification.vue
taoeffect: I opened an issue: https://github.com/okTurtles/group-income/issues/1674

Can you add a comment here linking to that issue?

taoeffect (changes requested):
Nice work @SebinSong! Tested & left a few comments!

Also wondering, why is it that here sometimes the progress bars are a different color and sometimes they aren't?

Screen Shot 2023-07-25 at 11 22 44 AM

Screen Shot 2023-07-25 at 11 22 47 AM

EDIT: copying Sebin's answer from Slack:

> The brighter color in the progress bar indicates that the active user has voted on it and the dimmer color means the opposite.

frontend/views/containers/chatroom/poll-message-content/PollVoteResult.vue
```diff
@@ -0,0 +1,209 @@
...
+    i18n.c-poll-label(tag='label') poll
+    h3.is-title-4 {{ pollData.question }}
+
+    i18n.pill.is-neutral.c-poll-expired-badge(v-if='isPollExpired') Poll closed
+    menu-parent.c-poll-menu(v-else)
+      menu-trigger.is-icon.c-poll-menu-trigger
+        i.icon-ellipsis-v
+      menu-content.c-poll-menu-content
+        ul
+          menu-item(tag='button' icon='edit' @click='() => pollUtils.switchOnChangeMode()')
```
taoeffect, line 13: Did you test to see if you can write this `@click` handler without the function wrapper?

```pug
menu-item(tag='button' icon='edit' @click='pollUtils.switchOnChangeMode')
```

frontend/views/containers/chatroom/CreatePoll.vue
SebinSong: makes sense.

frontend/views/containers/chatroom/MessagePoll.vue
SebinSong: @taoeffect From my knowledge, it's not necessary for `try ... catch` to have `await` inside it, to catch any error. (right?) if there is any error or exception thrown during the `sbp` operation that leads to the failure of closing the poll, it will let the user know by showing the banner.

taoeffect: @SebinSong It is necessary:

```
$ node
> (async () => { try { (async () => { throw 'asdf'; })() } catch (e) { console.log('caught:', e) } })()
Promise {
  undefined,
  [Symbol(async_id_symbol)]: 946,
  [Symbol(trigger_async_id_symbol)]: 5
}
> Uncaught 'asdf'
```

The try/catch is just a rewritten syntax sugar over `new Promise(() => { .. }).catch(..)`, and if you don't put the `await` you don't properly chain the `.catch()`

frontend/views/components/modal/ModalClose.vue
SebinSong: @taoeffect I can't exactly remember but I think I was fixing one of the bugs Alex raised during his review. it only prevents the `click` event triggered by the `X` button in modal from bubbling up the DOM tree, so I don't think the change will break something anywhere else in the app but if an issue is encountered regarding this later on, I will remove it.

frontend/views/containers/chatroom/MessageNotification.vue
SebinSong: @taoeffect thnx

frontend/views/containers/chatroom/MessagePoll.vue
SebinSong: @taoeffect 
```
const throwError = () => {
  throw new Error('random error!')
}

const anotherFunction = () => {
  try {
    throwError()
  } catch (error) {
    console.error('error thrown: ', error);
  }
}

anotherFunction()
```

Thought it still contradicted my experience of using it, so I just wrote/run this test script and it does seem to catch exceptions in synchronous operation too.

frontend/views/containers/chatroom/poll-message-content/PollVoteResult.vue
SebinSong: @taoeffect good catch here, it does work without additional function wrapper.

frontend/views/containers/chatroom/MessagePoll.vue
taoeffect: @SebinSong Your example does not use a Promise. The reason we call `await` here is because a promise is involved.

SebinSong: @taoeffect got it. will attach the `await` keyword then.

taoeffect:
Closing in favor of #1675 due to Github breaking down and not reflecting latest changes.

PR #1643, opened 2023-06-05
Proposal to update distribution date

### LINKED ISSUE #1543: Create a proposal for changing the distribution date
### Problem

If the wrong distribution date is accidentally set when the group is first created, there is currently no way except by recreating the group that this can be fixed.

### Solution

Create a way to fix this by passing a proposal. Allow the change to be applied immediately without voting if the group is currently in the first distribution period and the proposal is being created by the group creator.

### LINKED ISSUE #1710: `Jump to latest message` button should be invisible
### Problem

Steps to reproduce
- Create two channels; **c1** and **c2**
- Create 30 messages in **c1** and scroll to the top
- The `Jump to latest message` button will be shown
- Switch to the **c2**
- `Jump to latest message` should be invisible, but it is there in **c2**

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1543 
- Closes #1710


Silver-IT (commented):
@taoeffect, seems ready for your review.

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -117,16 +121,24 @@ export default ({
...
     ...mapGetters([
+      'groupSettings',
+      'ourUsername',
       'groupMembersCount',
       'groupShouldPropose',
       'groupProposalSettings'
     ]),
+    canChangeDistributionDateImmediately () {
+      return this.variant === 'changeDistributionDate' && this.ourUsername === this.groupSettings.groupCreator
+    },
```
SebinSong: @Silver-IT I can't help but feel that this part is somehow violating a good practice.
`ProposalTemplate.vue`, which is a template component shared by multiple existing proposal modals, now has a computed prop that is specific to one proposal modal (`DistributionDate.vue`). Plus, `DistributionDate.vue` has the exact same computed prop in there, which makes it feel the related logic here should be moved to `DistributionDate.vue`
Also, because of this `canChangeDistributionDateImmediately `, this component ended up having two computed props with very similar names (`shouldPropose`, `groupShouldPropose`) which can lead to confusion in future maintenance.

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -82,11 +82,15 @@
...
+        span(v-if='shouldPropose' v-safe-html='footerVotingExplanation')
         i18n(
-          v-else
+          v-else-if='!groupShouldPropose'
           :args='LTags("strong")'
         ) Your group has less than 3 members, so {strong_}this change will be immediate{_strong} (no voting required).
+        i18n(
+          v-else
+          :args='LTags("strong")'
+        ) The first distribution period is not started yet, so {strong_}this change will be immediate{_strong} (no voting required).
```
SebinSong: @Silver-IT it looks like this last `i18n` block here is only used by `DistributionDate.vue`. Then I think instead of specifying it here (which is a template component shared by multiple other proposal modals), it should be passed as `template(#footer='')` slot from `DistributionDate.vue`.

frontend/views/containers/proposals/ProposalTemplate.vue
Silver-IT: I got your point. Actually, this is what I felt while I was working in this part, but I didn't find the better solution.
I will try to improve this part again.

frontend/views/containers/proposals/ProposalTemplate.vue
Silver-IT: Good idea.

frontend/views/containers/proposals/ChangeVotingRules.vue
```diff
@@ -74,6 +74,11 @@ export default ({
       sbp('okTurtles.events/emit', CLOSE_MODAL)
     }
   },
+  mounted () {
+    if (!this.groupShouldPropose && this.ourUsername !== this.groupSettings.groupCreator) {
+      this.$refs.proposal.close()
+    }
```
SebinSong: https://github.com/okTurtles/group-income/assets/17641213/06d6183c-517f-4497-96d8-69cd5db3290e

@Silver-IT Apparently, changes here causes this unwanted behavior of the modal opens and then closes immediately. I think hiding the `Change` button is better solution instead of this.
(If this is happening somewhere else as well, the same solution should be applied there too)

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -37,10 +37,38 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
...
   }
 
+  const settingChangeMessages = (options) => ({
+    [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to change the groups {setting}.', options),
+    [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to change the {setting} is expiring.', options),
+    [PROPOSAL_VARIANTS.ACCEPTED]: L('The groups {setting} changed.'),
+    [PROPOSAL_VARIANTS.REJECTED]: L('The group {setting} hasn\'t changed.'),
+    [PROPOSAL_VARIANTS.EXPIRED]: L('The group {setting} hasn\'t changed.')
+  })
+
```
SebinSong, line 71: Nice work DRYing repetitive logics here.

frontend/views/containers/proposals/DistributionDate.vue
```diff
@@ -0,0 +1,187 @@
...
+        )
+          i18n(tag='option' disabled value='') Choose your group's distribution date
+          option(
+            v-for='(item, index) in ephemeral.distributionDayRange'
+            :key='index'
+            :value='item'
+          ) {{ humanDate(item, { month: 'long', year: 'numeric', day: 'numeric' }) }}
+      i18n.helper(:args='{currentDistributionDate}') Current distribution date is on {currentDistributionDate}.
+
+    template(#skipToProposeFooter='' v-if='skipToProposeDistributionDate')
```
SebinSong: It seems you haven't noticed that `ProposalTemplate.vue` already has a slot for footer in there. So you can use `template(#footer='')` instead here. (and `skipToProposeFooter` here and there should be removed)

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -82,11 +82,12 @@
...
       .c-footer
         i.icon-vote-yea
-        span(v-if='groupShouldPropose' v-safe-html='footerVotingExplanation')
+        span(v-if='shouldPropose' v-safe-html='footerVotingExplanation')
         i18n(
-          v-else
+          v-else-if='!groupShouldPropose'
           :args='LTags("strong")'
         ) Your group has less than 3 members, so {strong_}this change will be immediate{_strong} (no voting required).
+        slot(v-else name='skipToProposeFooter')
```
SebinSong: Like I mentioned in `DistributionDate.vue`, slot should be removed and then these block can be reverted back to `v-if` and `v-else` pair.

SebinSong:
@Silver-IT 
There is one thing we need to note regarding this task.
Currently in `state.js`, we have `periodBeforePeriod` getter that has below logic.

```
periodBeforePeriod (state, getters) {
  return (periodStamp: string) => {
    const len = getters.groupSettings.distributionPeriodLength
    return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), -len))
  }
}
```
This logic is based on the assumption that distribution dates are all apart from each other by the same length of period(which is `distributionPeriodLength`), but as soon as a new distribution date is set, this assumption will become wrong and the multiple places in the app where this getter is used (e.g. `SupportHistory.vue`, `TodoHistory.vue` etc..) will start to break.
So I think this problem will have to be taken care of either as part of this PR or as a follow-up task of this PR.

frontend/views/containers/proposals/DistributionDate.vue
Silver-IT: I saw the `template(#footer='')`. But I didn't want to repeat the whole content of that footer. I just wanted to add additional footer only when it's needed.

frontend/views/containers/proposals/ChangeVotingRules.vue
Silver-IT: Yeah. That's reasonable. Nice catch.

frontend/views/containers/proposals/DistributionDate.vue
Silver-IT: Can you please share the example of those codes here?

SebinSong: @Silver-IT 
What I meant was,

in `DistributionDate.vue`

```
template(#footer='' v-if='skipToProposeDistributionDate')
  .c-footer
    i.icon-vote-yea
    i18n(
      :args='LTags("strong")'
    ) The first distribution period is not started yet, so {strong_}this change will be immediate{_strong} (no voting required).
```

also in the relevant part of `ProposalTemplate.vue`
```
template(slot='footer' v-if='!isConfirmation')
  .c-footer
    i.icon-vote-yea
    span(v-if='shouldPropose' v-safe-html='footerVotingExplanation')
    i18n(
      v-else
      :args='LTags("strong")'
    ) Your group has less than 3 members, so {strong_}this change will be immediate{_strong} (no voting required).
```

This way, we can remove the `slot` in another `slot` pattern in `ProposalTemplate.vue`. I think it's cleaner this way although it involves duplications of a couple of lines of code(`.c-footer` and `i.icon-vote-yea`). (The decision is up to you  though)

Silver-IT: I don't think we can use `footer` slot in `DistributionDate.vue` because it is from the `ModalTemplate`, not `ProposalTemplate.vue`.

Silver-IT:
@SebinSong, nice catch regarding the `periodBeforePeriod`.
@taoeffect and I have already discussed this too and agreed to leave it at the moment since it's only possible to update the distribution date before the first distribution period is started.

frontend/views/containers/proposals/DistributionDate.vue
Silver-IT: @SebinSong, in my honest point of view, it's better to leave it same and to make no other changes more.
@taoeffect, what do you think?

SebinSong: @Silver-IT I tried this change and it did work but if you don't agree, that's ok. up to your and Greg's opinion

SebinSong:
My apologies. I accidentally deleted the source branch of this PR, thinking it's an old working branch of mine while I was organising local branches in my laptop..! just restored it.

SebinSong (dismissed):
Looks good to me @Silver-IT, Great work!

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -37,10 +37,38 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
   const options = Object.assign({}, baseOptions)
   const { proposalType, proposalData, variant } = proposal
 
+  const groupSettingType = proposalData.setting
+  let proposalSettingType
+  if (!!proposalData.ruleName && proposalData.ruleName !== proposalData.current.ruleName) {
+    proposalSettingType = 'votingSystem'
+  } else if (!!proposalData.ruleThreshold && proposalData.ruleThreshold !== proposalData.current.ruleThreshold) {
```
snowteamer: `!!proposalData.ruleName` and `!!proposalData.ruleThreshold` are going to always be true. This won't check whether these properties are defined

frontend/model/contracts/group.js
```diff
@@ -413,6 +413,9 @@ sbp('chelonia/defineContract', {
     groupShouldPropose (state, getters) {
       return getters.groupMembersCount >= 3
     },
+    groupDistributionStarted (state, getters) {
+      return Object.keys(getters.groupPeriodPayments).length >= 2
```
snowteamer: I think `new Date().toISOString() >= getters.groupSettings?.distributionDate` might be more reliable in some edge cases where we've just left the waiting period but the change hasn't been registered yet

frontend/model/state.js
```diff
@@ -747,12 +747,9 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
           // 'groupSettings.distributionDate': gets updated manually by calling 'updateCurrentDistribution' function(t2) in group.js
           // This logic removes the inconsistency that exists between these two from the point of time t1 till t2.
 
-          // Note: if this code gets called when we're in the period before the 1st distribution
-          //       period, then the distributionDate will get updated to the previous distribution date
-          //       (incorrectly). That in turn will cause the Payments page to update and display TODOs
-          //       before it should.
           const distributionDateInSettings = store.getters.groupSettings.distributionDate
-          if (oldPeriod && newPeriod && (newPeriod !== distributionDateInSettings)) {
+          const isCypressOrDistributionStarted = process.env.CI || store.getters.groupDistributionStarted
```
snowteamer: `process.env.CI` means we're in Travis, not necessarily in Cypress. Therefore it won't be set if we're running Cypress tests outside Travis. If that's intended you might want to use another name for this constant

frontend/model/contracts/group.js
taoeffect: @snowteamer `new Date()` is never allowed inside of contract getters. They must be deterministic, same for `process` functions.

Silver-IT: We can make it this way, I think.
```
groupDistributionStarted (state, getters) {
  return (cDateInISO: string) => cDateInISO >= getters.groupSettings?.distributionDate
}
```

snowteamer: @taoeffect You are right. Then maybe we can use Silver-IT's idea, or find a way so that his previous suggestion (`Object.keys(getters.groupPeriodPayments).length >= 2`) will always be correct

frontend/model/state.js
```diff
@@ -747,10 +747,6 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
           // 'groupSettings.distributionDate': gets updated manually by calling 'updateCurrentDistribution' function(t2) in group.js
           // This logic removes the inconsistency that exists between these two from the point of time t1 till t2.
 
-          // Note: if this code gets called when we're in the period before the 1st distribution
-          //       period, then the distributionDate will get updated to the previous distribution date
-          //       (incorrectly). That in turn will cause the Payments page to update and display TODOs
-          //       before it should.
```
taoeffect: Why was this comment removed?

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -57,29 +85,12 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
...
-        [PROPOSAL_VARIANTS.REJECTED]: L('The groups voting rules hasn\'t changed.'),
-        [PROPOSAL_VARIANTS.EXPIRED]: L('The groups voting rules hasn\'t changed.')
-      },
-      votingSystem: {
-        [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to change the groups voting system.', options),
-        [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to change the voting system is expiring.', options),
-        [PROPOSAL_VARIANTS.ACCEPTED]: L('The groups voting system changed.'),
-        [PROPOSAL_VARIANTS.REJECTED]: L('The groups voting system hasn\'t changed.'),
-        [PROPOSAL_VARIANTS.EXPIRED]: L('The groups voting system hasn\'t changed.')
-      }
```
taoeffect, line 82: Good job DRYing this 👍

frontend/model/contracts/group.js
```diff
@@ -1057,14 +1063,26 @@ sbp('chelonia/defineContract', {
...
+          allowPublicChannels: x => typeof x === 'boolean'
+        })(data)
+
+        const isGroupCreator = meta.username === getters.groupSettings.groupCreator
+        if ('allowPublicChannels' in data && !isGroupCreator) {
+          throw new TypeError(L('Only group creator can allow public channels.'))
+        } else if ('distributionDate' in data && !isGroupCreator) {
+          throw new TypeError(L('Only group creator can update distribution date.'))
+        } else if ('distributionDate' in data && getters.groupDistributionStarted(meta.createdDate)) {
+          throw new TypeError(L('Distribution is already started.'))
```
taoeffect: `'Distribution is already started.'` => `"Can't change distribution date because distribution period has already started."`

frontend/model/contracts/group.js
```diff
@@ -1073,6 +1091,10 @@ sbp('chelonia/defineContract', {
           Vue.set(state.settings, key, data[key])
         }
 
+        if ('distributionDate' in data) {
+          const period = dateToPeriodStamp(addTimeToDate(data.distributionDate, -getters.groupSettings.distributionPeriodLength))
+          Vue.set(state, 'paymentsByPeriod', { [period]: Object.values(getters.groupPeriodPayments)[0] })
+        }
```
taoeffect: I'm not sure about this code. Why is this here and what is it doing?

frontend/views/containers/proposals/ChangeVotingRules.vue
```diff
@@ -87,10 +87,14 @@ export default ({
...
     ]),
     ...mapGetters([
+      'ourUsername',
       'groupShouldPropose',
       'groupProposalSettings',
       'groupSettings'
     ]),
+    disabled () {
+      return !this.groupShouldPropose && this.ourUsername !== this.groupSettings.groupCreator
+    },
```
taoeffect: Why was this added?

taoeffect (changes requested):
Nicely done @Silver-IT, I haven't finished reviewing everything yet, but I left another preliminary review and did some testing.

I noticed a few issues during testing:

### Issue 1: disabled menus

When `u2` joins the group this is what his menu looks like for creating proposals:

Screen Shot 2023-08-29 at 4 43 49 PM

It should look the same as on `master`:

Screen Shot 2023-08-29 at 5 19 53 PM

### Issue 2: Can't change the distribution date again

I created a group today (Aug 29) and the distribution was set to Sept 1 initially (if I remember correctly). As `u1` I changed it to be Aug 30th. Now I can no longer change it again:

Screen Shot 2023-08-29 at 5 32 14 PM

I should be able to change it as many times as I want until the first period starts.

frontend/model/state.js
Silver-IT: Oh, it was mistake, I guess.

frontend/model/contracts/group.js
Silver-IT: If the distributionDate is changed by proposal, `paymentsByPeriod` should be changed too because the keys of `paymentsByPeriod` are the distribution dates.
Also we shouldn't initialize the paymentPeriod by calling `initPaymentPeriod` function, because there could be users who have already entered their income details.

frontend/views/containers/proposals/ChangeVotingRules.vue
Silver-IT: So we don't allow a group member to update voting rules when group has less than 3 members and he is not group creator.

frontend/model/contracts/group.js
taoeffect: OK, that doesn't make much sense because if the user entered their income details, that will not modify or affect the values in `state.paymentsByPeriod`, instead it will affect their group profile.

What we should do instead is the following:

1. First we need to make sure that we are only updating the distribution date if it is safe to update it. It is only safe to update the distribution date if `meta.createdDate` is before the current distribution date. So you need to first create an `if` condition here to check for that.
2. Next, we can verify that there is only 1 group period currently in the `state.paymentsByPeriod`.
3. If so, we delete it / reset `state.paymentsByPeriod` to an empty object.
4. Next, we update the distribution date to the new distribution date. So `state.settings.distributionDate` is updated to the new date.
5. Finally, we call `initFetchPeriodPayments` exactly in the same way that it is called in the `process` function for `'gi.contracts/group'`

This will ensure the exact same steps as happened in the group constructor are followed for resetting the distribution date, and that it is done safely only at the appropriate time.

frontend/views/containers/proposals/ChangeVotingRules.vue
taoeffect: OK but this should be disabled at the menu item on the dashboard for bringing up the modal, not here, right?

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -117,16 +118,22 @@ export default ({
...
     ...mapGetters([
+      'groupSettings',
+      'ourUsername',
       'groupMembersCount',
       'groupShouldPropose',
       'groupProposalSettings'
     ]),
+    shouldPropose () {
+      return this.groupShouldPropose && !this.skipToPropose
+    },
```
taoeffect: What is this `skipToPropose` thing?

frontend/views/containers/proposals/DistributionDate.vue
```diff
@@ -0,0 +1,188 @@
...
+    },
+    distributionStarted () {
+      return this.groupDistributionStarted(new Date().toISOString())
+    },
+    skipToProposeDistributionDate () {
+      return !this.distributionStarted && this.ourUsername === this.groupSettings.groupCreator
+    },
+    disabled () {
+      return this.distributionStarted || (!this.groupShouldPropose && this.ourUsername !== this.groupSettings.groupCreator)
+    }
```
taoeffect: This is causing a DRY violation. We should disable bringing up the modal from the dropdown menu, not here.

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -92,12 +92,16 @@ export default ({
...
       'groupShouldPropose',
       'groupSettings',
       'groupMembersCount',
       'groupMincomeFormatted',
       'groupMincomeSymbolWithCode'
-    ])
+    ]),
+    disabled () {
+      return !this.groupShouldPropose && this.ourUsername !== this.groupSettings.groupCreator
+    }
```
taoeffect: Same issue here. Please revert related changes

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -7,7 +7,7 @@
     :a11yTitle='title'
   )
     template(slot='subtitle')
-      i18n(v-if='groupShouldPropose') New proposal
+      i18n(v-if='shouldPropose') New proposal
```
taoeffect, line 10: Not clear to me that these changes are needed either

taoeffect (changes requested):
Review ready!

frontend/views/containers/proposals/ChangeVotingRules.vue
Silver-IT: Users can open the modal even though we disable the menu item.
They can copy the URL (e.g, http://localhost:8000/app/dashboard?rule=percentage&modal=ChangeVotingRules) and open it directly in the browser. So I thought we should check the permission inside the modal and should do one of the following two things.
- Close the modal
- Disable the submission

frontend/model/contracts/group.js
Silver-IT: @taoeffect, Please correct me, if I am wrong.
I think we should set `haveNeedsSnapshot` using `getters.haveNeedsForThisPeriod(period)` after calling `initFetPeriodPayments`.

frontend/views/containers/proposals/ProposalTemplate.vue
Silver-IT: In the solution of the issue #1543, it describes that the group creator should not make propose to update distribution date before the first distribution started.
So an additional field is needed to make decision to propose or not for some proposals (Only DistributionDate for now, but not sure for later).

Silver-IT: Should we change the name then?
- `shouldImmediateUpdate`
- `shouldNotPropose`
- ...

frontend/views/containers/proposals/ProposalTemplate.vue
snowteamer: Maybe not needed, but doesn't seem to be harmful, and shorter so why not

frontend/model/contracts/group.js
snowteamer: > if the user entered their income details, that will not modify or affect the values in state.paymentsByPeriod
 
 I'm not 100% sure but I think the `haveNeedSnapshot` field of the payment period object gets updated when users update their income details

taoeffect: > I'm not 100% sure but I think the `haveNeedSnapshot` field of the payment period object gets updated when users update their income details

That's true, but AFAICT that code is not affected by this code.

frontend/model/contracts/group.js
```diff
@@ -1073,6 +1092,13 @@ sbp('chelonia/defineContract', {
           Vue.set(state.settings, key, data[key])
         }
 
+        if ('distributionDate' in data) {
+          Vue.set(state, 'paymentsByPeriod', {})
+          const curPeriodPayments = initFetchPeriodPayments({ contractID, meta, state, getters })
+          const period = Object.keys(getters.groupPeriodPayments)[0]
+          curPeriodPayments.haveNeedsSnapshot = getters.haveNeedsForThisPeriod(period)
+        }
```
taoeffect: @Silver-IT This is better, nice. 👍 

A few things are missing here from the review though:

>  It is only safe to update the distribution date if `meta.createdDate` is before the current distribution date. So you need to first create an `if` condition here to check for that.

 And:

> Next, we update the distribution date to the new distribution date. So `state.settings.distributionDate` is updated to the new date.

However, in the code above, the distribution date is always being updated. This is dangerous because as written this contract does not prevent the distribution date from being updated at any point in time. It must be updated only when conditions are correct (as mention the previous review). To fix this, I would move the entire `if` condition before the loop on line 1091 and do something like this:

```js
if ('distributionDate' in data) {
  if (meta.createdDate < state.settings.distributionDate) {
    // ...
  } else {
    delete data.distributionDate
  }
}
```

Finally, sorry for not answering your question earlier about `haveNeedsSnapshot`. As far as I can tell, that does not need to be updated because it is entirely unrelated to the distribution date. So lines 1098 and 1099 above should be deleted, and similarly, line 1097 can be rewritten so that `initFetchPeriodPayments` is called without capturing its return value.

frontend/model/state.js
```diff
@@ -751,8 +751,13 @@ sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
           //       period, then the distributionDate will get updated to the previous distribution date
           //       (incorrectly). That in turn will cause the Payments page to update and display TODOs
           //       before it should.
+          //       `distributionStarted` is different from getters.groupDistributionStarted
+          //       and it allows distributionDate can be updated
+          //       only from the 2nd distribution date when new distribution is started.
+          //       so this `distributionStarted` fixes the issue mentioned above
           const distributionDateInSettings = store.getters.groupSettings.distributionDate
-          if (oldPeriod && newPeriod && (newPeriod !== distributionDateInSettings)) {
+          const distributionStarted = reactiveDate.date > dateFromPeriodStamp(distributionDateInSettings)
```
taoeffect: I'm not convinced this is needed. What happens if this change is removed?

test/cypress/integration/group-paying.spec.js
```diff
@@ -106,13 +106,9 @@ describe('Group Payments', () => {
 
     // NOTE: TWO HEISENBUGS ARE IN THIS TEST! PLEASE LEAVE THESE COMMENTS FOR FUTURE
     //       REFERENCE IN CASE WE RUN INTO MORE!
-    // NOTE: when we had this command here, Cypress would sometimes fail because cy.window() would
-    //       for some reason return a promise that wouldn't resolve.
```
taoeffect, line 110: Why was this comment and the one below removed?

Please add it back unless there's a strong reason not to, in which case please help me understand.

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -37,10 +37,38 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
...
+      options['setting'] = L('voting rules')
+    }
   }
 
+  const settingChangeMessages = (options) => ({
+    [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to change the groups {setting}.', options),
+    [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to change the {setting} is expiring.', options),
+    [PROPOSAL_VARIANTS.ACCEPTED]: L('The groups {setting} changed.'),
+    [PROPOSAL_VARIANTS.REJECTED]: L('The group {setting} hasn\'t changed.'),
+    [PROPOSAL_VARIANTS.EXPIRED]: L('The group {setting} hasn\'t changed.')
```
taoeffect: It's generally preferred to use double-quotes whenever the string has `'` inside of it

taoeffect (changes requested):
@Silver-IT ready!

frontend/model/contracts/group.js
Silver-IT: > It is only safe to update the distribution date if meta.createdDate is before the current distribution date. So you need to first create an if condition here to check for that.

I think I did it in the `validate` function.
```
else if ('distributionDate' in data &&
    (getters.groupDistributionStarted(meta.createdDate) || Object.keys(getters.groupPeriodPayments).length > 1)) {
    throw new TypeError(L('Can\'t change distribution date because distribution period has already started.'))
}
```

frontend/model/state.js
Silver-IT: It's possible to update distributionDate before the distribution is started by calling `gi.contracts/group/updateSettings` function. Once the distributionDate is updated before the distribution is started, this `currentPaymentPeriod` watch also calls the `gi.actions/group/updateDistributionDate` function again and sets a wrong distributionDate.
**reactiveDate** and this watch should be allowed to update `distributionDate` only after the distribution is started.

test/cypress/integration/group-paying.spec.js
Silver-IT: I remember cypress was failed in this part and moving it back to the above made it work again.
Not sure why failed. But as you suggested, I will run the test again after moving it back to the below and if it doesn't fail, I will revert this change.

Silver-IT: Having tested several times, moving it back to the below makes Cypress failed.

frontend/views/containers/proposals/ProposalTemplate.vue
taoeffect: Oh, ok, yes, the name `skipToPropose` is very confusing in English (`skipProposal` would make more sense).

I would name this: `applyChangesWithoutProposal`, EDIT: but it seems you've renamed it to `shouldImmediateChange` which works too 👍

frontend/views/containers/proposals/Mincome.vue
```diff
@@ -106,7 +106,7 @@ export default ({
     validateMincome () {
       const mincomeAmount = normalizeCurrency(this.form.mincomeAmount)
       if (mincomeAmount === this.groupSettings.mincomeAmount) {
-        this.$refs.formMsg.danger(L('The new mincome should be different than the current one.'))
+        this.$refs.formMsg.danger(L('The new mincome should be different from the current one.'))
```
taoeffect: The previous phrasing was more correct IMO

frontend/model/state.js
taoeffect: Hmm, OK. One thing though, I think the comment is wrong. You could use `store.getters.groupDistributionStarted(reactiveDate.date)`, it would be the same result. Can you make that change to DRY things?

frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -120,15 +123,34 @@ export default ({
         },
         { type: 'header', name: 'Voting Systems' },
         // { type: 'item', id: 'change-disagreeing-number', name: 'Change disagreeing number', icon: 'vote-yea' },
-        { type: 'item', id: 'change-to-percentage-base', name: 'Change to percentage base', icon: 'vote-yea' },
+        {
+          type: 'item',
+          id: 'change-to-percentage-base',
+          name: 'Change to percentage base',
```
taoeffect: While we're boy-scoutting, three change requests:

1. We forgot to wrap all of these in `L` functions, please update this line and the lines below for each `name` entry to wrap the string in an `L` function and re-run `strings`
2. Also, can you change this specific string from `'Change to percentage base'` to: `'Change voting threshold'` ?
3. Also wrap using `L` function the `name` for the `header` (e.g. `'Voting Systems'` => `L('Voting Systems')`  --- all user-facing strings must be localized

taoeffect (changes requested):
Final review ready!

taoeffect (approved):
Approved!! Great job @Silver-IT!! 😄 🎉

PR #1644, opened 2023-06-06
Some bug fixes in dashboard

### LINKED ISSUE #1620: Invalid expire date of invite link
### Problem

When click `Add` button in dashboard page to get invitation link, we get the following modal which shows the constant expire date which is wrong.
image

### Solution

Fix it

### LINKED ISSUE #1621: Minor bugs in dashboard
### Problem

1. Can not select the text inside the notification even though it seems like that the clickable area is the `income details`.
image
2. Currently it is `Different than` instead of `different from`. I think it should be `different from`.
image

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1620 
- Closes #1621 

This PR fixes only the first issue of the #1621, since the second issue is fixed in the PR #1643.


SebinSong (approved):
Looks good! @Silver-IT

taoeffect (approved):
Nicely done @Silver-IT!

PR #1648, opened 2023-06-08
#1614 limit group size to Dunbar's number

### LINKED ISSUE #1614: Limit group membership to Dunbar's Number
### Problem

In an upcoming blog post on Sybil attacks we'll explain why Group Income limits groups to [Dunbar's Number](https://en.wikipedia.org/wiki/Dunbar's_number).

### Solution

When a user attempts to invite a new member via proposal after the group has 150 members, display an alert saying: `Group sizes are limited to Dunbar's Number to prevent fraud.`

EDIT: On second thought, we may want to allow people who really want to create large groups to be able to do so. However, we should warn them about the risks of fraud that can happen if they choose to do this. So for this issue, instead of preventing them from creating a group with more members, we can post a warning using `gi.ui/prompt` that asks:

> Groups over 150 members are at significant risk for fraud, [because it is difficult to verify everyone's identity](https://en.wikipedia.org/wiki/Dunbar's_number).
> 
> Are you sure that you want to add more members?
>
> Yes / Cancel

With the default being "Cancel".

### LINKED ISSUE #1651: A long user display name should be truncated in navigation menu
### Problem
A long user display name overflows out of the side menu and should be truncated when that happens.
(Found this issue while working on one of @Silver-IT 's change requests on [this PR](https://github.com/okTurtles/group-income/pull/1628))

### Solution
Fix it.

SebinSong:
closes #1614 
closes #1651

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -567,22 +567,31 @@ export default (sbp('sbp/selectors/register', {
     // if current size of the group is >= 150, display a warning prompt first before presenting the user with
     // 'AddMembers' proposal modal.
 
-    const { groupMembersCount } = sbp('state/vuex/getters')
-    const isGroupSizeLarge = groupMembersCount >= MAX_GROUP_MEMBER_COUNT
+    const enforceDunbar = true // Context for this hard-coded boolean variable: https://github.com/okTurtles/group-income/pull/1648#discussion_r1230389924
+    const { groupMembersCount, currentGroupState } = sbp('state/vuex/getters')
+    const memberInvitesCount = Object.values(currentGroupState.invites || {}).filter(invite => invite.creator !== INVITE_INITIAL_CREATOR).length
+    const isGroupSizeLarge = (groupMembersCount + memberInvitesCount) >= MAX_GROUP_MEMBER_COUNT
 
     if (isGroupSizeLarge) {
-      const promptOpts = {
-        heading: 'Large group size',
-        question: L(
-          "Groups over 150 members are at significant risk for fraud, {a_}because it is difficult to verify everyone's identity.{_a} Are you sure that you want to add more members?",
-          { a_: `<a class='link' href='${ALLOWED_URLS.WIKIPEDIA_DUMBARS_NUMBER}'' target='_blank'>`, _a: '</a>' }
-        ),
-        yesButton: L('Yes'),
-        noButton: L('Cancel')
+      const translationArgs = {
+        a_: `<a class='link' href='${ALLOWED_URLS.WIKIPEDIA_DUMBARS_NUMBER}' target='_blank'>`,
+        _a: '</a>'
       }
+      const promptConfig = enforceDunbar
+        ? {
+            heading: 'Large group size',
+            question: L("Group sizes are limited to {a_}Dunbar's Number{_a} to prevent fraud.", translationArgs),
+            yesButton: L('OK')
+          }
+        : {
+            heading: 'Large group size',
+            question: L("Groups over 150 members are at significant risk for fraud, {a_}because it is difficult to verify everyone's identity.{_a} Are you sure that you want to add more members?", translationArgs),
+            yesButton: L('Yes'),
+            noButton: L('Cancel')
+          }
 
-      const yesButtonSelected = await sbp('gi.ui/prompt', promptOpts)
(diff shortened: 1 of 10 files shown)


frontend/model/contracts/shared/constants.js
```diff
@@ -32,6 +32,7 @@ export const PAYMENTS_ARCHIVED = 'payments-archived'
 export const MAX_ARCHIVED_PERIODS = 100
 export const MAX_SAVED_PERIODS = 2
 export const MAX_HISTORY_PERIODS = 6
+export const MAX_GROUP_MEMBER_COUNT = 150 // Dunbar's number (https://en.wikipedia.org/wiki/Dunbar's_number)
```
SebinSong, line 35: we can test this PR by temporarily reducing this constant to a small number like 5 while running this branch locally.

SebinSong:
@taoeffect 
This is the message UI that says group size is limited to the Dunbar's number. wanted to add a wikipedia link you attached to the message as well, so that users can find out what that means in case they are curious. hope it looks good.

taoeffect (changes requested):
Excellent job @SebinSong! I realized from the conference that a change is needed. Please see the recent edit to the issue: https://github.com/okTurtles/group-income/issues/1614

frontend/controller/actions/group.js
```diff
@@ -561,6 +563,32 @@ export default (sbp('sbp/selectors/register', {
...
+
+    if (isGroupSizeLarge) {
+      const promptOpts = {
+        heading: 'Large group size',
+        question: L(
+          "Groups over 150 members are at significant risk for fraud, {a_}because it is difficult to verify everyone's identity.{_a} Are you sure that you want to add more members?",
+          { a_: `<a class='link' href='${ALLOWED_URLS.WIKIPEDIA_DUMBARS_NUMBER}'' target='_blank'>`, _a: '</a>' }
+        ),
+        yesButton: L('Yes'),
+        noButton: L('Cancel')
```
taoeffect: I don't remember for sure, but I think `yesButton` is the default right? If so, per the edited comment in the issue, we want the default to say `Cancel`, not `Yes`

frontend/controller/actions/group.js
```diff
@@ -561,6 +563,32 @@ export default (sbp('sbp/selectors/register', {
...
+      }
+
+      const yesButtonSelected = await sbp('gi.ui/prompt', promptOpts)
+      if (yesButtonSelected) {
+        sbp('okTurtles.events/emit', REPLACE_MODAL, 'AddMembers')
+      } else return false
```
taoeffect: @SebinSong So, @dotmacro and I are not in complete agreement about how we should handle this scenario. She thinks we should simply prevent users from creating groups larger than 150, rather than allowing them the option to, since a lot of our messaging (and future blog posts) are all about having groups be a maximum of 150 members to prevent Sybil attacks and foster a sense of community. I think she has a decent point, but I am not 100% convinced. However, because she feels more strongly on this issue than I do, I am willing to go along with her suggestion that we keep it locked down to 150 members.

However, I do not want to completely remove this code that you added. Instead, can you add a hard-coded boolean variable to the code, something like `const enforceDunbar = true`? If `enforceDunbar` is `true` (and by default it is), bring up a prompt that only has 1 button that says "OK", and do not allow users to create a proposal. If `enforceDunbar` is `false`, do what this PR is currently doing.

Sound OK? And this way we can change our mind later if needed.

taoeffect (changes requested):
See comment

frontend/controller/actions/group.js
SebinSong: @taoeffect yes, that sounds good.

taoeffect (changes requested):
Great work @SebinSong! Looks like this PR is ready to merge, except for this flow error I noticed when running `grunt dev`:

```
Error --------------------------------------------------------------------- frontend/controller/actions/group.js:572:103

Cannot get `invite.creator` because property `creator` is missing in mixed [1]. [incompatible-use]

   frontend/controller/actions/group.js:572:103
   572|     const memberInvitesCount = Object.values(currentGroupState.invites || {}).filter(invite => invite.creator !== INVITE_INITIAL_CREATOR).length
```

taoeffect (approved):
Nicely done @SebinSong! LGTM!

Issue #298, opened 2017-08-13
Pubsub should automatically keep in sync with state.contracts


taoeffect:
### Problem

From ["Notes" in #257 comment](https://github.com/okTurtles/group-income-simple/issues/257#issuecomment-308263188):

> In order to make the "revert vuex state" stuff work correctly, we need to tie the Pubsub to the vuex state by having the Pubsub maintain a list of websockets it has open, and have the vuex store also contain this same list in it under a key called "pubsub". Then have the Pubsub watch the vuex store for any modifications to that key. For example, if the vuex "pubsub" key has 1 item in it that the Pubsub list does not, then the Pubsub takes that to mean that it should open a websocket connection. If the vuex list does not have an item that the Pubsub does have, then the Pubsub takes that to mean it should unsubscribe/close the websocket.

### Solution

Per above pubsub keep its own array/map of rooms and monitor `state.contracts` for changes, making sure to join/leave as appropriate.

Might be worth introducing #295 with this.

PR #1649, opened 2023-06-08
#1632 Fix 'see original proposal' menu in Group-settings page

### LINKED ISSUE #1632: Non-functional "See original proposal" menu, also missing on mobile
### Problem

When a proposal to invite someone passes, for example for user "test", we see a "See Original Proposal" link on desktop in the invitation table in the group settings:

Screen Shot 2023-06-02 at 5 59 34 PM

However:

1. This menu doesn't work
2. And it's not visible on mobile:

Screen Shot 2023-06-02 at 5 59 25 PM

### Solution

Either make the menu work (by pulling up the archived proposals modal and scrolling to the specific row) and visible on mobile, or remove it. Better if you can get it to work though.

SebinSong:
closes #1632


frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -279,8 +287,21 @@ export default ({
...
-      console.log(inviteItem, 'TODO - See Original Proposal')
+    async handleSeeOriginal ({ invitee }) {
+      const key = `proposals/${this.ourUsername}/${this.currentGroupId}`
+      const archivedProposals = await sbp('gi.db/archive/load', key) || []
+      const proposalItemExists = archivedProposals.length > 0 || archivedProposals.some(entry => {
+        const { data } = entry[1]
+
+        return data.proposalType === PROPOSAL_INVITE_MEMBER &&
+          data.proposalData.member === invitee
+      })
```
Silver-IT: Since there is no validation in invitee `Full name`, I think checking existence using `inviteSecret` is more reasonable.
You can just add to check `inviteSecret` in line #298.

Silver-IT (changes requested):
@SebinSong, nice work. Just added one comment.

frontend/views/containers/group-settings/InvitationsTable.vue
SebinSong: @Silver-IT I'm not entirely sure what you meant by 'Since there is no validation in invitee Full name' but these lines check if the archived proposals of the app has the `PROPOSAL_INVITE_MEMBER` item **for a particular user.**
`inviteSecret` field exists even for `anyone` invite (as far as it seems) so I reckon checking the existence of that value is not useful in this case.

Silver-IT: What I mean is that we can make two proposals to add two member `Alex Jin`. So I don't think the `invitee full name` shouldn't be used as one of the identifiers of the proposal.

SebinSong: @Silver-IT Oh, now it's clear what you meant. yeah that's right and I think in that kind of case, additional check of whether the creator of the proposal is the current user would achieve our need here. what's your thought?

Silver-IT: One creator can add two members with same name, can't him?

SebinSong: @Silver-IT That's technically possible but why would anyone do that..? why does there has to be two different active invites for the same person?

Silver-IT: @taoeffect, what do you think?

SebinSong: @Silver-IT Btw, agreed  using `inviteSecret` is another good way to identify a proposal so made a change in the following commit here :)
https://github.com/okTurtles/group-income/pull/1649/commits/f8639e00a5a67e1ddff4813559d975c4608ae09a

Silver-IT (dismissed):
Nice work, @SebinSong. I totally approve this PR.

frontend/views/containers/group-settings/InvitationsTable.vue
Silver-IT: Thank you, @SebinSong.

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -278,8 +287,21 @@ export default ({
...
+        const { data, payload } = entry[1]
+
+        return data.proposalType === PROPOSAL_INVITE_MEMBER &&
+          payload.inviteSecret === inviteSecret
+      })
+
+      if (proposalItemExists) {
+        sbp('okTurtles.events/emit', OPEN_MODAL, 'PropositionsAllModal')
+      } else {
+        alert('Unable to find the original proposal.')
```
taoeffect: 1. All user-facing errors must be localized, so here we need the `L` function
2. What if the proposal is an existing open proposal? If it is, we should forward them to the proposals section on the dashboard instead of showing an error

taoeffect (changes requested):
Great work @SebinSong! Review finished!

frontend/views/containers/group-settings/InvitationsTable.vue
SebinSong: @taoeffect 
I believe an invite for a member only created after the proposal has been passed, right? That's why I took that case out of consideration.

taoeffect: @SebinSong oh, you're right 👍

taoeffect (approved):
Beautifully done @SebinSong! Tested & approved!

PR #1652, opened 2023-06-14
Some error fixes

### LINKED ISSUE #1637: Proposals from one group appear in another group sometimes
### Problem

While showing the app to a friend, who created his own group and invited me (resulting in me being in 2 groups), I saw in his group a cancelled proposal from our group:

Screenshot 2023-06-04 at 6 57 39 PM copy

The steps to reproduce are similar to #1636, although I'm not exactly sure how to reproduce consistently.

### Solution

Find and fix bug. Proposals from one group should never appear on another group.

### LINKED ISSUE #1647: Handle message related errors and display them to the user
### Problem

`DMMixin.js` has a lot of code that calls group actions but doesn't handle errors:

```js
    createPrivateDM (username: string) {
      // NOTE: username should be valid
      sbp('gi.actions/mailbox/createDirectMessage', {
        contractID: this.currentIdentityState.attributes.mailbox,
        data: {
          privacyLevel: CHATROOM_PRIVACY_LEVEL.PRIVATE,
          usernames: [username]
        }
      })
    },
    createGroupDM (usernames: string[]) {
      // NOTE: usernames.length should be gte 2
      sbp('gi.actions/mailbox/createDirectMessage', {
        contractID: this.currentIdentityState.attributes.mailbox,
        data: {
          privacyLevel: CHATROOM_PRIVACY_LEVEL.GROUP,
          usernames
        }
      })
    },
    async addMemberToGroupDM (chatRoomId: string, username: string) {
      // NOTE: chatRoomId, username should be valid
      const profile = this.ourContactProfiles[username]
      await sbp('gi.actions/chatroom/join', {
        contractID: chatRoomId,
        data: { username }
      })
      await sbp('gi.actions/mailbox/joinDirectMessage', {
        contractID: profile.mailbox,
        data: {
          privacyLevel: CHATROOM_PRIVACY_LEVEL.GROUP,
          contractID: chatRoomId
        }
      })
    },
    setDMVisibility (chatRoomId: string, hidden) {
      // NOTE: chatRoomId should be of valid direct message
      sbp('gi.actions/mailbox/setDirectMessageVisibility', {
        contractID: this.currentIdentityState.attributes.mailbox,
        data: { contractID: chatRoomId, hidden }
      })
    },
```

All `GIErrorUIRuntimeError` errors (and most other errors) in Group Income must be handled by displaying them to the user, but these are currently being ignored.  This will confuse users when errors happen, they will wonder why something isn't working and will have no idea what's going on.

### Solution

Properly handle all errors. Check the designs to see how errors should be handled. If there is no obvious design, then feel free to handle them by using either `v-error` + some place in the UI to show them, or use `alert()` and open an issue asking @leihla how they should be properly displayed (but in the meantime handle them anyway with `alert()`).

Silver-IT:
### Summary of Changes

- 1637


frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -170,6 +183,11 @@ export default ({
 
       this.openModal(modalNameMap[itemId], queries[itemId] || undefined)
     }
+  },
+  watch: {
+    currentGroupId () {
+      this.updateArchivedProposals()
```
SebinSong, line 198: Nice fix here that makes sure the different groups don't share archived proposals while switching the groups around.

SebinSong (dismissed):
I haven't found any issue in this PR. Great work. @Silver-IT

frontend/views/utils/allowedUrls.js
```diff
@@ -14,7 +14,7 @@ const ALLOWED_URLS: Object = Object.freeze(Object.fromEntries([
   ['DONATE_PAGE', 'https://groupincome.org/donate'],
   ['FAQ_PAGE', 'https://groupincome.org/faq'],
   ['COMMUNITY_PAGE', 'https://groupincome.org/community'],
-  ['WIKIPEDIA_DUMBARS_NUMBER', "https://en.wikipedia.org/wiki/Dunbar's_number"]
+  ['WIKIPEDIA_DUNBARS_NUMBER', "https://en.wikipedia.org/wiki/Dunbar's_number"]
```
SebinSong, line 17: Nice catch on this typo

SebinSong (dismissed):

frontend/model/contracts/group.js
```diff
@@ -831,6 +833,7 @@ sbp('chelonia/defineContract', {
 
             if (proposal) {
               Vue.set(proposal, 'status', STATUS_EXPIRED)
+              Vue.set(proposal, 'dateClosed', meta.createdDate)
```
taoeffect, line 836: Nice 👍

frontend/views/containers/chatroom/DMMixin.js
```diff
@@ -36,45 +37,69 @@ const DMMixin: Object = {
...
-        }
-      })
+      try {
+        sbp('gi.actions/mailbox/createDirectMessage', {
+          contractID: this.currentIdentityState.attributes.mailbox,
+          data: {
+            privacyLevel: CHATROOM_PRIVACY_LEVEL.PRIVATE,
+            usernames: [username]
+          }
+        })
```
taoeffect: This won't throw because `await` is not used. Same for all of the ones below.

frontend/views/containers/chatroom/DMMixin.js
```diff
@@ -36,45 +37,69 @@ const DMMixin: Object = {
...
+      } catch (err) {
+        const errMsg = L('Error during create direct message: {err}', { err: err.message })
```
taoeffect: You do not need to reconstruct the error message, it is already correctly generated by the `gi.actions`, simply call `alert` with `err.message`. Same for all of the ones below.

Please also test forcefully throwing these errors from within the `gi.actions` to verify that what the user sees makes sense. For example, if there is HTML included (e.g. a link to the troubleshooting page), then maybe it is better to use `gi.ui/prompt` to display the error instead of `alert` (and make sure it is rendered correctly).

frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -134,14 +136,26 @@ export default ({
...
-        })
-      }, 1000 * 60 * 60 * 24)) // 1 day
+      this.clearTimeouts()
+      for (const [hash, obj] of this.ephemeral.archivedProposals) {
+        const dateClosed = new Date(obj.dateClosed).getTime()
+        this.ephemeral.timeouts.push(setTimeout(() => {
+          this.ephemeral.archivedProposals = this.ephemeral.archivedProposals.filter(x => x[0] !== hash)
+        }, dateClosed - Date.now() + DAYS_MILLIS))
+      }
+    },
```
taoeffect: Here `hashWithProp` is no longer being used... is that correct? If so, should it be removed from the function parameters?

Is this line correct?

```js
this.ephemeral.archivedProposals = this.ephemeral.archivedProposals.filter(x => x[0] !== hash)
```

It used to check against `hashWithProp[0]`, but now it doesn't.

What is this condition doing instead? It looks like it might be removing all of the proposals...?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "group-income",
   "version": "0.0.1",
-  "contractsVersion": "0.1.4",
+  "contractsVersion": "0.1.6",
```
taoeffect: Why skip `0.1.5`?

EDIT: Oh I see. This PR depends on #1643 being merged first which updates to `0.1.5`

taoeffect (changes requested):
Nice work @Silver-IT!

Found a few issues + have some questions.

frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -164,12 +178,18 @@ export default ({
         'remove-member': { toRemove: true }
       }
 
-      if (itemId === 'add-new-member' && this.groupShouldPropose) {
+      const isWelcomeInviteExpired = this.currentWelcomeInvite.expires < Date.now()
+      if (itemId === 'add-new-member' && (this.groupShouldPropose || isWelcomeInviteExpired)) {
```
taoeffect, line 189: Note that your other PR adds code related to `groupShouldPropose`: https://github.com/okTurtles/group-income/pull/1643/files#diff-e12f80a0ba7af9f7795e718cfba169d8a5dee2e91542d13e203c242fff48dc88R104

Is `groupShouldPropose` ever checked without also checking the expiry of the invite? Maybe `groupShouldPropose` should be updated itself so that it accounts for the expiry of the invite so that everything works well with the changes in PR #1643 too?

frontend/views/containers/chatroom/DMMixin.js
Silver-IT: Nice catch. I will fix it.

PS: This seems to be related to the #1696.

frontend/views/containers/proposals/ProposalsWidget.vue
Silver-IT: > Here `hashWithProp` is no longer being used... is that correct? If so, should it be removed from the function parameters?

I didn't want to use that parameter, but I started thinking we should use that parameter to ignore the redundant operations. But one more parameter is needed. `groupId`. Nice catch, @taoeffect

> Is this line correct?

Yes

> What is this condition doing instead? It looks like it might be removing all of the proposals...?

Proposals which are archived in the last 24 hours should be removed automatically when 24 hours are passed from when they are archived. The timers are doing this.

frontend/views/containers/proposals/ProposalsWidget.vue
Silver-IT: `groupShouldPropose` only describes if there are 3 members in the group, and if users should propose to do something. And it could be used for all the proposals, so I don't think it should be updated.

Even though the user is not a group creator, he can add members if the initial invite link is expired (30 days by default) and there are less than 3 members in the group.

frontend/views/containers/proposals/ProposalsWidget.vue
```diff
@@ -134,14 +136,33 @@ export default ({
...
+          this.ephemeral.timeouts.push(setTimeout(() => {
+            this.ephemeral.archivedProposals = this.ephemeral.archivedProposals.filter(x => x[0] !== hash)
+          }, new Date(obj.dateClosed).getTime() - Date.now() + DAYS_MILLIS))
+        }
+      } else if (archPropData?.[0] === this.currentGroupId) {
+        this.ephemeral.archivedProposals.unshift([archPropData[1], archPropData[2]])
+        this.ephemeral.timeouts.push(setTimeout(() => {
+          this.ephemeral.archivedProposals = this.ephemeral.archivedProposals.filter(x => x[0] !== archPropData[1])
+        }, new Date(archPropData[2].dateClosed).getTime() - Date.now() + DAYS_MILLIS))
+      }
```
taoeffect, line 160: Please destructure the data in `archPropData` and assign it to clearly named variables to improve the code's readability

frontend/model/state.js
```diff
@@ -445,6 +450,11 @@ const getters = {
       amountTotal
     }
   },
+  currentWelcomeInvite (state, getters) {
+    const invites = getters.currentGroupState.invites
+    const secret = Object.keys(invites).find(invite => invites[invite].creator === INVITE_INITIAL_CREATOR)
+    return { ...invites[secret], secret }
+  },
```
taoeffect, line 457: @Silver-IT The way `invites` are handled on `e2e-protocol` is different. When you send the PR for this issue to `e2e-protocol`, please check how `e2e-protocol` uses the invites (because their location might be different).

Ask @corrideat if you have any questions.

taoeffect (changes requested):
Left some more review.

However, I think we need to split this PR up and also apply it to `e2e-protocol` instead of `master`.

@Silver-IT Can you please send 1PR per issue, and send the changes to `e2e-protocol`?

PR #1653, opened 2023-06-15
Use API_PORT and --port option instead of PORT_SHIFT

### LINKED ISSUE #1642: Minor change to how API_PORT and API_URL are set
### Problem

To set a different port we are currently using an envvar called `PORT_SHIFT` that works like so:

```js
const applyPortShift = (env) => {
  // TODO: implement automatic port selection when `PORT_SHIFT` is 'auto'.
  const API_PORT = 8000 + Number.parseInt(env.PORT_SHIFT || '0')
  const API_URL = 'http://127.0.0.1:' + API_PORT

  if (Number.isNaN(API_PORT) || API_PORT < 8000 || API_PORT > 65535) {
    throw new RangeError(`Invalid API_PORT value: ${API_PORT}.`)
  }
}

Object.assign(process.env, applyPortShift(process.env))
```

I'm not sure why I chose to go with `PORT_SHIFT` instead of simply specifying the port via `API_PORT` envvar.. we should be able to specify the port directly. However, the above code makes that impossible because `API_PORT` is always overwritten even if it's set.

### Solution

I think the solution is as simple as the following modification:

```js
  return { API_PORT, API_URL, ...env }
```

snowteamer:
Closes #1642

### Summary of changes:
- New `API_PORT` envar, allowing to specify the default backend port. If not defined, then port 8000 will be the default.
- New `--port` Grunt option, to be used in Grunt commands. Useful e.g. when running two or more server instances.
 Example: `grunt --port=8004` will try to launch a dev instance on port 8004, whereas the usual `grunt` command will rely on the `API_PORT` envar, or 8000 if not defined.
- Cypress tests have been updated to no longer refer to a hard-coded `http://localhost:8000 URL`.
  This allows to run them against a backend running on another port.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,27 +31,6 @@ const packageJSON = require('./package.json')
 // See https://esbuild.github.io/api/#define
 // =======================
 
-/**
- * Creates a modified copy of the given `process.env` object, according to its `PORT_SHIFT` variable.
- *
- * The `API_PORT` and `API_URL` variables will be updated.
- * TODO: make the protocol (http vs https) variable based on environment var.
- * @param {Object} env
- * @returns {Object}
- */
-const applyPortShift = (env) => {
-  // TODO: implement automatic port selection when `PORT_SHIFT` is 'auto'.
-  const API_PORT = 8000 + Number.parseInt(env.PORT_SHIFT || '0')
-  const API_URL = 'http://127.0.0.1:' + API_PORT
-
-  if (Number.isNaN(API_PORT) || API_PORT < 8000 || API_PORT > 65535) {
-    throw new RangeError(`Invalid API_PORT value: ${API_PORT}.`)
-  }
-  return { ...env, API_PORT, API_URL }
-}
-
-Object.assign(process.env, applyPortShift(process.env))
-
 process.env.GI_VERSION = `${packageJSON.version}@${new Date().toISOString()}`
 
 // Not loading babel-register here since it is quite a heavy import and is not always used.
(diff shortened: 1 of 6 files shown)


Gruntfile.js
```diff
@@ -66,6 +45,10 @@ const {
   EXPOSE_SBP = ''
 } = process.env
 
+if (!['development', 'production'].includes(NODE_ENV)) {
+  throw new TypeError(`Invalid NODE_ENV value: ${NODE_ENV}.`)
+}
```
Silver-IT, line 47: It's a nice checking for exceptions.

frontend/views/pages/DesignSystem.vue
```diff
@@ -1306,7 +1306,7 @@ page(
             pre
               | link-to-copy.c-link(link='string')
           td
-            link-to-copy.c-link(link='http://localhost:3000/app/join?groupId=21XWnNJvq9&secret=3763')
+            link-to-copy.c-link(link='/app/join?groupId=21XWnNJvq9&secret=3763')
```
Silver-IT, line 1309: Nice fix.

Silver-IT (dismissed):
Great work, @snowteamer. I totally approve this PR. :)

taoeffect (approved):
LGTM! Just needs merge conflict fix.

taoeffect (changes requested):
Marking as "changes requested" until merge conflict is fixed so that it correctly shows up in the list of PRs

test/cypress/integration/group-paying.spec.js
```diff
@@ -5,6 +5,7 @@ export function humanDate (datems, opts = { month: 'short', day: 'numeric' }) {
   return new Date(datems).toLocaleDateString(locale, opts)
 }
 
+const API_URL = Cypress.config('baseUrl')
```
taoeffect, line 8: Nice 👍

taoeffect (approved):
Approved! Great work @snowteamer!!

PR #1655, opened 2023-06-16
Fix issue 1629

### LINKED ISSUE #1629: Page doesn't refresh on new version
### Problem

Try the following:

1. Delete `data/groupincome.db` and run `GI_PERSIST=sqlite grunt dev`
2. Create a group and go to the chatroom, post a message
3. Stop the server with `^C` and make some changes to the chatroom contract, then run `grunt pin:0.1.5`
4. Run `GI_PERSIST=sqlite grunt dev` again

The open page you were logged in on will reconnect but won't refresh. It should refresh.

### Solution

Find and fix bug.

snowteamer:
Closes #1629

### Summary of changes:
- New `process.env.CONTRACTS_VERSION`  field available, taken from `package.json`'s `contractsVersion`.
- The server will now send both its current `GI_VERSION` and `CONTRACTS_VERSION` to the client on every (re)connection, including in development mode, allowing it to reload if necessary (fix issue #1629).
- Fixed a bug causing the server to not log ping rounds despite the option being enabled.
-

### DIFF
--- a/backend/pubsub.js
+++ b/backend/pubsub.js
@@ -107,7 +107,7 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
   // Setup a ping interval if required.
   if (server.options.pingInterval > 0) {
     server.pingIntervalID = setInterval(() => {
-      if (server.clients.length && server.options.logPingRounds) {
+      if (server.clients.size && server.options.logPingRounds) {
         log.debug('Pinging clients')
       }
       server.clients.forEach((client) => {
@@ -128,7 +128,7 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
 }
 
 const defaultOptions = {
-  logPingRounds: process.env.NODE_ENV === 'development' && !process.env.CI,
+  logPingRounds: process.env.NODE_ENV !== 'production' && !process.env.CI,
   logPongMessages: false,
   maxPayload: 6 * 1024 * 1024,
   pingInterval: 30000
--- a/frontend/controller/backend.js
+++ b/frontend/controller/backend.js
@@ -13,7 +13,7 @@ const languageFileMap = new Map([
   ['fr', 'french.json']
 ])
 
-sbp('okTurtles.events/on', NOTIFICATION_TYPE.UPDATE_AVAILABLE, (versionInfo) => {
+sbp('okTurtles.events/on', NOTIFICATION_TYPE.VERSION_INFO, (versionInfo) => {
   console.info('New Group Income version available:', versionInfo)
   // Prevent the client from trying to reconnect when the page starts unloading.
   sbp('okTurtles.data/get', PUBSUB_INSTANCE).destroy()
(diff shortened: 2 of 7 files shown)


backend/pubsub.js
```diff
@@ -107,7 +107,7 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
   // Setup a ping interval if required.
   if (server.options.pingInterval > 0) {
     server.pingIntervalID = setInterval(() => {
-      if (server.clients.length && server.options.logPingRounds) {
+      if (server.clients.size && server.options.logPingRounds) {
```
Silver-IT, line 110: Nice catch. `server.clients` is `Set` object which has `size` instead of `length`.

frontend/controller/backend.js
```diff
@@ -13,8 +13,8 @@ const languageFileMap = new Map([
   ['fr', 'french.json']
 ])
 
-sbp('okTurtles.events/on', NOTIFICATION_TYPE.APP_VERSION, (version) => {
-  console.info('New Group Income version available:', version)
+sbp('okTurtles.events/on', NOTIFICATION_TYPE.UPDATE_AVAILABLE, (versionInfo) => {
```
Silver-IT: I think it should be `NOTIFICATION_TYPE.VERSION_INFO` instead of `NOTIFICATION_TYPE.UPDATE_AVAILABLE`.

shared/domains/chelonia/chelonia.js
```diff
@@ -148,12 +148,14 @@ export default (sbp('sbp/selectors/register', {
...
           const ourVersion = process.env.GI_VERSION
-          const theirVersion = msg.data
+          const theirVersion = msg.data.GI_VERSION
 
-          if (ourVersion !== theirVersion) {
-            sbp('okTurtles.events/emit', NOTIFICATION_TYPE.APP_VERSION, theirVersion)
+          const ourContractsVersion = process.env.CONTRACTS_VERSION
+          const theirContractsVersion = msg.data.CONTRACTS_VERSION
+          if (ourVersion !== theirVersion || ourContractsVersion !== theirContractsVersion) {
+            sbp('okTurtles.events/emit', NOTIFICATION_TYPE.UPDATE_AVAILABLE, { ...msg.data })
```
Silver-IT: I also think this should be `VERSION_INFO` instead of `UPDATE_AVAILABLE`.

frontend/controller/backend.js
Silver-IT: Or should add `UPDATE_AVAILABLE` in `NOTIFICATION_TYPE`.

Silver-IT (changes requested):
Great work, @snowteamer. Left a few feedback.

shared/domains/chelonia/chelonia.js
snowteamer: Yeah there was an inconsistency here, thank you for noticing!

taoeffect (approved):
Great work @snowteamer! Tested & approved!

PR #1661, opened 2023-06-26
Implement favicon badge

### LINKED ISSUE #1379: Fix warning: "InstallTrigger is deprecated and will be removed in the future."
### Problem

A recently added dependency (favicon.js) causes this warning to appear in the console in Firefox:

> InstallTrigger is deprecated and will be removed in the future.

### Solution

Either fork the dependency and fix the problem in the fork (and submit a PR upstream to the original project), or find a better dependency for the functionality that we're using this for.

When selecting dependencies, please always prefer dependencies that have 0 dependencies of their own. If you cannot find such ones, choose the one that is well designed and implemented, and has the fewest dependencies.

And another alternative is to implement the functionality yourself entirely without resorting to a dependency.

Silver-IT:
### Summary of Changes

- Closes #1379

### DIFF
--- a/frontend/utils/faviconBadge.js
+++ b/frontend/utils/faviconBadge.js
@@ -1,3 +1,11 @@
+/*!
+ * FaviconBadge - A small library for manipulating the Favicon
+ * Modified from Tinycon.js by Tom Moor
+ * Original project: https://github.com/tommoor/tinycon
+ * Copyright (c) 2023 Alex Jin
+ * @license MIT Licensed
+ */
+
 let currentFavicon = null
 let originalFavicon = null
 let faviconImage
--- a/frontend/utils/faviconBadge.js
+++ b/frontend/utils/faviconBadge.js
@@ -9,28 +9,8 @@ const size = 16 * r
 const defaults = {
   width: 8,
   height: 8,
-  font: 10 * r + 'px arial',
-  color: '#ffffff',
   background: '#ff0000',
-  fallback: true,
-  crossOrigin: true,
-  abbreviate: true
-}
-
-const ua = (function () {
-  const agent = navigator.userAgent.toLowerCase()
-  // New function has access to 'agent' via closure
-  return function (browser) {
-    return agent.indexOf(browser) !== -1
-  }
-}())
-
-const browser = {
-  ie: ua('trident'),
-  chrome: ua('chrome'),
-  webkit: ua('chrome') || ua('safari'),
-  safari: ua('safari') && !ua('chrome'),
-  mozilla: ua('mozilla') && !ua('chrome') && !ua('safari')
+  crossOrigin: true
 }
 
 // private methods
@@ -91,14 +71,8 @@ const setFaviconTag = (url) => {
   }
 }
 
-const drawFavicon = (label, color) => {
-  // fallback to updating the browser title if unsupported
-  if (!(getCanvas() instanceof HTMLCanvasElement) || browser.ie || browser.safari || options.fallback === 'force') {
-    return updateTitle(label)
-  }
-
+const drawFavicon = (bubble) => {
   const context = getCanvas().getContext('2d')
-  color = color || '#000000'
   const src = getCurrentFavicon()
 
   faviconImage = document.createElement('img')
@@ -110,7 +84,7 @@ const drawFavicon = (label, color) => {
     context.drawImage(faviconImage, 0, 0, faviconImage?.width, faviconImage?.height, 0, 0, size, size)
 
     // draw bubble over the top
-    if ((label + '').length > 0) drawBubble(context, label, color)
+    if (bubble) drawBubble(context)
 
     // refresh tag in page
     refreshFavicon()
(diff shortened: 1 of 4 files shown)


Silver-IT (commented):
@taoeffect, @SebinSong, ready for your review.

frontend/utils/faviconBadge.js
```diff
@@ -0,0 +1,253 @@
...
+    for (const key in defaults) {
+      options[key] = custom[key] || defaults[key]
+    }
+  },
+  setImage: function (url: string) {
+    currentFavicon = url
+    refreshFavicon()
+  },
+  setBubble: function (label: string | number, color?: string) {
+    drawFavicon(String(label || ''), color)
```
SebinSong: is there a reason why `drawFavicon` has to be wrapped in `setBubble`? it looks like `drawFavicon` is only used in here. can this just be `FaviconBadge.drawFavicon(this.shouldSetBadge ? 1 : 0)` instead for better readability?

frontend/utils/faviconBadge.js
```diff
@@ -0,0 +1,253 @@
...
+  context.font = (browser.webkit ? 'bold ' : '') + options.font
+  context.fillStyle = options.background
+  context.strokeStyle = options.background
+  context.lineWidth = r
+
+  // bubble
+  // draw circle bubble
+  context.arc(right - arcRadius, bottom - arcRadius, arcRadius, 0, 2 * Math.PI)
+
+  // draw rectangle bubble
```
SebinSong: Can all these lines of code that are commented out be just removed so that `drawBubble()` is more readable?

frontend/utils/faviconBadge.js
```diff
@@ -0,0 +1,253 @@
...
+}
+
+const refreshFavicon = () => {
+  // check support
+  if (getCanvas() instanceof HTMLCanvasElement) {
+    setFaviconTag(getCanvas().toDataURL())
+  }
+}
+
+const abbreviateNumber = (label): string => {
```
SebinSong: As far as this code `FaviconBadge.setBubble(this.shouldSetBadge ? 1 : 0)` looks, currently the only possible value for `label` is either 1 or 0 and this `abbreviateNumber` function isn't doing anything. If there is no future case that this becomes useful, Can this  `abbreviateNumber` and the related logic be removed so that the file is kept concise and dry?

SebinSong:
Nice work @Silver-IT, left a few comments.

frontend/utils/faviconBadge.js
Silver-IT: Okay. I will remove all the code which are useless at the moment.

Silver-IT: Will remove all the lines out of the current requirement. KISS principle!

frontend/utils/faviconBadge.js
Silver-IT: I will remove this function since we are not using label at the moment and bubble is only the thing we need at the moment.

frontend/utils/faviconBadge.js
Silver-IT: In terms of workflow the function name should be `drawFavicon`, but for the user's point of view, it should be `setBubble`.
I like to keep the function name `setBubble` if it's not a big deal.

SebinSong (dismissed):
Looks good, @Silver-IT

frontend/utils/faviconBadge.js
```diff
@@ -0,0 +1,154 @@
+let currentFavicon = null
```
taoeffect, line 9: If this code is based on some other project (as it seems to be), we must include attribution, along with a copy of that project's license in this file.

taoeffect (changes requested):
Nice work @Silver-IT! Thanks for helping us remove dependencies 😄 

One comment!

frontend/utils/faviconBadge.js
Silver-IT: @taoeffect  I've used the library 'tinycon.js' as reference, but made many changes. Should I include the license in this case?

taoeffect: @Silver-IT Yes, the MIT license requires it: https://github.com/tommoor/tinycon/blob/master/LICENSE

Please also include a link back to the original project. You can also add a comment saying this is a modified version of tinycon.js, but it's important to always credit the original authors.

Silver-IT: @taoeffect  Can I add the comment below at the top of the file?

```
/*!
 * FaviconBadge - A small library for manipulating the Favicon
 * Modified from Tinycon.js by Tom Moor
 * Original project: https://github.com/tommoor/tinycon
 * Copyright (c) 2023 Alex Jin
 * @license MIT Licensed
 */
```

taoeffect: @Silver-IT Yep! That would work well too 😄 👍

taoeffect (approved):
Great work @Silver-IT!

PR #1668, opened 2023-07-14
Fix Vue error in groupProposalSettings() getter

snowteamer:
Closes #1662

PR #1673, opened 2023-07-24
Make payment periods a linked list

### LINKED ISSUE #1631: Store payment periods as a doubly linked list
### Problem

The way payment periods are stored at the moment is somewhat poorly designed.

It is based on the assumption that a payment period must be exactly 30 days long, and to move forward and backward through payment periods we add/subtract 30 days to a string to figure out what the next/previous period was.

Storing payment periods this way will break how we calculate streaks if:

- We change how long a payment period is
- We for some reason have to force the group into the next payment period (e.g. by calling `'gi.contracts/group/forceDistributionDate'`)

### Solution

It would be better if we stored payment periods as a doubly-linked list, with each payment period pointing to and referencing the previous and next payment period.

A payment period then:

- Is referenced by a UUID
- Contains within it the start date of the period (e.g. `startDate: "<ISOString>"`)
- Contains within it a pointer to the previous period (e.g. `prevPeriod: "<uuid>"`)
- Contains within it a pointer to the next period (e.g. `nextPeriod: "<uuid>"`, that is set once a new period is created)
- Contains its own UUID (e.g. `id: "<uuid>"`)

This fundamentally changes how we find and reference payment periods. For example, this code from `SupportHistory.vue` has to be changed:

```js
    periods () {
      for (let i = 0; i < MAX_HISTORY_PERIODS - 1; i++) {
        const period = this.periodBeforePeriod(periods[0])
        if (comparePeriodStamps(period, this.firstDistributionPeriod) < 0) break
        else periods.unshift(period)
      }
      return periods
    }
```

Similarly, anything using `paymentsByPeriod` needs to be updated. The way payments are stored in the archive needs to be updated too because it uses the period stamp in the key name:

```js
const archPaymentsKey = `payments/${username}/${period}/${contractID}`
```

And also this mysterious `payment.period` reference here in `RecordPayment.vue`:

```js
@click='openModal("PaymentDetail", { id: payment.hash, period: payment.period })'
```

Which I believe comes from this code in `'gi.contracts/group/archivePayments'`:

```js
newSentOrReceivedPayments[receivedOrSent].push({ hash, period, data, meta, amount: data.amount })
```

Of course, there are probably many other places that need to be updated.

snowteamer:
Closes #1631

### Summary of changes:

- Previously, the `periodStampGivenDate()` getter would return a timestamp for an "imaginary" payment period when the given date was too far in the future or in the past. It could do so by assuming the period length was a constant. Now since that assumption no longer holds, it will return `null` if no actual payment period data can be found for the given date, without trying to extrapolate in the future or the past.

- The eponymous function has been removed from `time.js`.


frontend/model/contracts/group.js
```diff
@@ -328,24 +338,18 @@ sbp('chelonia/defineContract', {
         })
       }
     },
-    periodBeforePeriod (state, getters) {
-      return (periodStamp: string) => {
-        const len = getters.groupSettings.distributionPeriodLength
-        return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), -len))
-      }
+    previousPaymentPeriodID (state, getters): string | void {
```
taoeffect: On Slack you said you wanted to avoid changing code in many places, and I agree that's a good goal, so that's why I'm confused by why you'd rename something that doesn't seem like it needs to be renamed?

frontend/model/contracts/group.js
```diff
@@ -536,13 +540,16 @@ sbp('chelonia/defineContract', {
...
         const initialState = merge({
+          chatRooms: {},
+          invites: {},
           payments: {},
           paymentsByPeriod: {},
-          thankYousFrom: {}, // { fromUser1: { toUser1: msg1, toUser2: msg2, ... }, fromUser2: {}, ...  }
-          invites: {},
+          profiles: {
+            [meta.username]: initGroupProfile(meta.identityContractID, meta.createdDate)
+          },
```
taoeffect: Similarly, here's a bunch of changes involving re-arranging the existing code that didn't need to be made... why?

frontend/model/contracts/group.js
snowteamer: This comment was rather about code logic than just names, but yeah I might revert the renaming. It's a change I made initially to help me remember things better, as `periodBeforePeriod` doesn't actually work with periods but period timestamps (which are also period ids)

frontend/model/contracts/group.js
snowteamer: I tend to find unsorted entries a bit distracting, so I couldn't resist the urge to sort them. The same goes for helper functions, but reordering them would have caused a big git diff unfortunately.
In particular this caused me trouble to remember where in `time.js` the `addTimeToDate()` function was, I had to Ctrl-F it several times

snowteamer: Of course I can revert this change for this PR

taoeffect: We can leave this if it's important to you, but, for the future, please send a minimal PR with only essential changes or clear improvements/simplifications

frontend/model/contracts/group.js
```diff
@@ -14,12 +14,13 @@ import {
...
 import { createPaymentInfo, paymentHashesFromPaymentPeriod } from './shared/functions.js'
 import { merge, deepEqualJSONType, omit, cloneDeep } from './shared/giLodash.js'
-import { addTimeToDate, dateToPeriodStamp, compareISOTimestamps, dateFromPeriodStamp, isPeriodStamp, comparePeriodStamps, periodStampGivenDate, dateIsWithinPeriod, DAYS_MILLIS } from './shared/time.js'
+import { addTimeToDate, dateToPeriodStamp, compareISOTimestamps, dateFromPeriodStamp, isPeriodStamp, comparePeriodStamps, dateIsWithinPeriod, DAYS_MILLIS } from './shared/time.js'
 import { unadjustedDistribution, adjustedDistribution } from './shared/distribution/distribution.js'
 import currencies, { saferFloat } from './shared/currencies.js'
 import { inviteType, chatRoomAttributesType } from './shared/types.js'
 import { arrayOf, mapOf, objectOf, objectMaybeOf, optional, string, number, boolean, object, unionOf, tupleOf } from '~/frontend/model/contracts/misc/flowTyper.js'
 
+// Just a shortcut for Vue.set() with a default initial value.
```
taoeffect, line 23: Nice comment 👍

frontend/model/contracts/group.js
```diff
@@ -313,38 +325,43 @@ sbp('chelonia/defineContract', {
       return getters.groupSettings.mincomeCurrency
     },
     periodStampGivenDate (state, getters) {
-      return (recentDate: string | Date) => {
-        if (typeof recentDate !== 'string') {
-          recentDate = recentDate.toISOString()
+      return (date: Date | string): string | null => {
+        const givenStamp = typeof date === 'string' ? date : date.toISOString()
+        // Should we throw instead?
+        if (!isPeriodStamp(givenStamp)) return null
```
taoeffect, line 331: Yes, please throw `TypeError('must be periodstamp')` (note: purposefully not localized because too obscure and developer-facing)

frontend/model/contracts/group.js
```diff
@@ -42,8 +43,16 @@ function initGroupProfile (contractID: string, joinedDate: string) {
...
-function initPaymentPeriod ({ getters }) {
+// Creates a PaymentPeriod object.
+function initPaymentPeriod ({ getters, period }) {
+  const len = getters.groupSettings.distributionPeriodLength
+  // For now we just add the current period length to the action timestamp,
+  // but this field must to be updated whenever the period length is changed while this period is active.
+  const nextPeriodID = dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(period), len))
   return {
+    // TODO: leave this undefined until the next PaymentPeriod object has been created.
+    nextPeriodID,
```
taoeffect, line 54: What is this TODO comment about?

frontend/model/contracts/group.js
```diff
@@ -107,9 +116,12 @@ function updateCurrentDistribution ({ contractID, meta, state, getters }) {
...
   const period = getters.periodStampGivenDate(meta.createdDate)
   const noPayments = Object.keys(curPeriodPayments.paymentsFrom).length === 0
-  // update distributionDate if we've passed into the next period
-  if (comparePeriodStamps(period, getters.groupSettings.distributionDate) > 0) {
+  const { distributionDate } = getters.groupSettings
+  // Update the distribution date, streaks, and previous/next pointers if we've passed into the next period.
+  if (comparePeriodStamps(period, distributionDate) > 0) {
     updateGroupStreaks({ state, getters })
+    state.paymentsByPeriod[distributionDate].nextPaymentPeriodID = period
+    curPeriodPayments.previousPaymentPeriodID = distributionDate
```
taoeffect, line 124: Why are these named `nextPaymentPeriodID` and `previousPaymentPeriodID` instead of `previousPeriodID` and `nextPeriodID` as used elsewhere in the code?

Looks like a bug to me - please remember to using `forceDistributionDate`!

frontend/model/contracts/group.js
```diff
@@ -313,38 +325,43 @@ sbp('chelonia/defineContract', {
...
+        let previousPeriodStamp = getters.periodBeforePeriod(currentPeriodStamp)
+        if (compareISOTimestamps(givenStamp, currentPeriodStamp) >= 0) {
+          // Find the first period stamp that's later than the given date.
+          while (nextPeriodStamp && compareISOTimestamps(givenStamp, nextPeriodStamp) >= 0) {
+            currentPeriodStamp = nextPeriodStamp
+            nextPeriodStamp = getters.periodAfterPeriod(nextPeriodStamp)
+          }
+          return currentPeriodStamp
+        } else {
+          // Iterate backwards to find a first previous period stamp that's lower than the given date.
```
taoeffect, line 345: `clearOldPayments` is going to clear out the payment periods available for iteration after 6 months (e.g. `MAX_SAVED_PERIODS`).

This code will break if a date is given outside of that range, and will return `null`.

Therefore this code in `PaymentDetail.vue` will break:

```js
        this.payment.periodstamp = this.periodStampGivenDate(this.payment.meta.createdDate)
```

Along with other code in the project that uses `periodStampGivenDate`.

You'll have to comb through the project for all uses of `periodStampGivenDate` and handle appropriately, perhaps creating a more accurate `periodStampGivenDate` async function in `time.js` that tries to use this version of `periodStampGivenDate`, and upon getting `null`, checks the archives for older payment periods. You'll need to look at `'gi.contracts/group/archivePayments'` for details.

Similarly, you'll need to do a project-wide search also for `periodBeforePeriod` and verify that all uses are safe and revert to checking the archives if necessary.

Note: you cannot use `async` stuff in getters, which must be synchronous and can only access data from `state`.

taoeffect (changes requested):
Nicely done @snowteamer, left a few comments/questions.

frontend/model/contracts/group.js
snowteamer: For now this refers to a period that doesn't exist yet, and things which don't exist yet are maybe not supposed to already have IDs. So this TODO is about setting the `nextPeriodID` only when the corresponding period object is actually created. But that's not a high priority

frontend/model/contracts/group.js
snowteamer: Yea most likely a bug. A test case in `group-paying.spec.js` is already using `giForceDistributionDateToNow`, which depends on `forceDistributionDate`, so I'm a bit surprised it didn't catch this. Could also have been caught with better typechecking.

frontend/model/contracts/group.js
snowteamer: Thank you for clarifying this, it was not obvious to me. So, `PaymentDetail.vue` and other locations need the ability to display information for payments belonging to older deleted periods?

taoeffect: @snowteamer Yes, the Payments page has Sent payments and Received payments that will show historical payments, and the ability to inspect them to show details about those old payments that are stored in IndexedDB under special keys. See also the file `PaymentsMixin.js`

frontend/model/contracts/group.js
```diff
@@ -313,38 +325,43 @@ sbp('chelonia/defineContract', {
...
+    periodBeforePeriod (state, getters): string | void {
+      return (periodStamp: string) => getters.groupPeriodPayments[periodStamp]?.previousPeriodID
     },
-    periodAfterPeriod (state, getters) {
-      return (periodStamp: string) => {
-        const len = getters.groupSettings.distributionPeriodLength
-        return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), len))
-      }
+    periodAfterPeriod (state, getters): string | void {
+      return (periodStamp: string) => getters.groupPeriodPayments[periodStamp]?.nextPeriodID
```
taoeffect, line 358: This needs to be modified because sometimes we are given dates in the future, for example, have a look at this code in `state.js`:

```js
setInterval(function () {
  // We want the getters to recalculate all of the payments within 1 minute of us entering a new period.
  const rememberedPeriodStamp = store.getters.periodStampGivenDate?.(reactiveDate.date)
  const currentPeriodStamp = store.getters.periodStampGivenDate?.(new Date())
  if (rememberedPeriodStamp !== currentPeriodStamp) {
    reactiveDate.date = new Date()
  }
}, 60 * 1000)
```

This code will break with the way `periodAfterPeriod` is currently implemented, but it would work if you re-introduced the code we have before. You might want to combine the two, e.g. if `nextPeriodID` doesn't exist, use the old code to generate the correct periodStamp using the `distributionPeriodLength`

taoeffect (changes requested):
Could you incorporate the changes from #1668 into this PR and close that other PR?

taoeffect:
Closing in favor of #1691

PR #1676, opened 2023-07-26
allow revoking 'Anyone' link only when group-size is >= 3

### LINKED ISSUE #1670: 'Anyone' invite link is accessible via 'Add members' menu even after it's revoked.
### Problem
[ Steps to reproduce ]

1. `user-1` signs up and creates a group `G1`
2. `user-2` signs up and gets invited to the group `G1`.
3. `G1`'s group-size is < 3 now. then, revoke the 'anyone' invite link of `G1` using `user-1`.

4. After it's revoked, click on 'propose adding a new member' menu below it.

5. The modal still has the revoked invite link in it.

### Solution
Discuss the expected behaviour of this case and make a fix.

SebinSong:
closes #1670

### DIFF
--- a/frontend/views/containers/group-settings/InvitationsTable.vue
+++ b/frontend/views/containers/group-settings/InvitationsTable.vue
@@ -281,7 +281,10 @@ export default ({
       }
     },
     showRevokeLinkMenu (inviteItem) {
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Nice work @SebinSong!

PR #1682, opened 2023-07-30
A couple of follow-up bug fixes for the Chat polls feature

### LINKED ISSUE #1674: Implement copying links to messages (and visiting messages by their links)
### Problem

Screen Shot 2023-07-25 at 11 18 03 AM

- Copying a message link doesn't work
- In PR #1628 - "Jump to poll" doesn't work 

### Solution

- Make it possible to jump to a specific message by visiting a link (so that loading in a new tab works)
- Make sure it also works to click on a link when the site is already loaded (e.g. so that "Jump to poll" works)

### LINKED ISSUE #1678: Poll should display its duration
### Problem
Current implementation of the poll doesn't show the information of how long the poll will last for, which is something users would be interested when looking at it.

### Solution

Add duration info to various state(e.g. `voted`, `to vote on` etc..) of the poll message component.

SebinSong:
closes #1674 
closes #1678

@taoeffect 
As the fix for the #1678, I added this badge to various message states that says the expiration date. Hope it looks good.


taoeffect (changes requested):
@SebinSong this is a beautiful PR and an amazing job getting all of these features working!! 👏 

The only issue I was able to find is related to how messages overlap when the vertical height of the screen is kinda small (found on Firefox Developer Edition).

These two screenshots show the overlap both when the message below the poll is hovered via the mouse and when it's not hovered:

Screen Shot 2023-08-01 at 10 31 54 AM
Screen Shot 2023-08-01 at 10 32 02 AM

SebinSong:
@taoeffect 
PR has been updated with the fix for the bug you raised, and thanks.
it turns out @Silver-IT had already created `scrollToMessage()` utility method which works great, so All I had to do was make the chat utilise it in the right places.

Pls let me know if you encounter any other issues.

taoeffect (approved):
@SebinSong - this was a well executed PR, and great job noticing and reusing the existing code! 👏👏🫡🥇🏆

PR #1683, opened 2023-08-01
#1680 add 'Create' button for DM

### LINKED ISSUE #1680: Creating DMs on mobile is confusing - needs button
### Problem

Mobile users are used to tapping things. It's not clear here how the DM group is supposed to be created:

Screen Shot 2023-07-28 at 10 15 37 AM

Normally there is a button somewhere that you tap to create the DM. The user might be confused and not realize they need to press the "return" key.

### Solution

On both mobile *AND* desktop, add a button, either inside of the input field (once at least 1 user has been entered into the input field), or outside of it just below (on mobile) or next to it (on desktop) that says `Create`.

SebinSong:
closes #1680

### DIFF
--- a/frontend/assets/style/components/_buttons.scss
+++ b/frontend/assets/style/components/_buttons.scss
@@ -236,6 +236,10 @@ button,
     justify-content: flex-start;
   }
 
+  &.is-end {
+    justify-content: flex-end;
+  }
+
   button,
   .button {
     &:not(:last-child) {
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Great job @SebinSong! Seems to work well. The one thing I'm not sure about is whether the button should be blue instead of green... but we can figure that out with @leihla when she's back. Approved!

PR #1685, opened 2023-08-04
#1005 DRY GroupPledgesGraph.vue

### LINKED ISSUE #1005: Rewrite GroupPledgesGraph to use getters
### Problem

The current implementation of `graphData` in `GroupLedgesGraph.vue` uses the `mincome-proportional.js` algorithm directly instead of using one or more of the getters: `groupIncomeDistribution `, `ourContributionSummary`, `ourPayments`, `ourPaymentsSummary`.

The reason is because it was created before those getters existed. However, this is causing a few issues:

1. It makes it harder to switch the underlying algorithm if we need to (because it would have to be switched in multiple places, here in the vue file, as well as in state.js)
2. It now might be violating DRY, and doing computation that might already be done in one or more of the getters of state.js
3. This is additional code that must be independently reasoned about, instead of us being safely assured that all of the relevant code for displaying payment distribution graphs is located in the getters

### Solution

Rewrite this code to use the minimal number of getters mentioned above from state.js, while preserving perfect functionality.

The end result should be more lines of code removed than added.

SebinSong:
closes #1005 

[ Details ]
- Found out that the group's `haveNeeds` calculation is executed manually there by going through each member's profile, so replaced it with the existing getter `haveNeedsForThisPeriod`
- the `haveNeed` entry for the currently active user must be computed dynamically in that Vue component, so unfortunately the rest of the logic there cannot be replaced. (e.g. using the `groupIncomeDistribution` getter in `state.js`), but did replace the part where it uses `mincomeProportional` function from `mincome-proportional.js` directly with `unadjustedDistribution` wrapper function in case it (`unadjustedDistribution`) will be extended or updated in the future.

Hope it sounds good.

### DIFF
--- a/frontend/views/containers/contributions/GroupPledgesGraph.vue
+++ b/frontend/views/containers/contributions/GroupPledgesGraph.vue
@@ -85,10 +85,10 @@ export default ({
       const ourIncomeAmount = doWeNeedIncome && this.amount < mincome && this.amount
       const haveNeeds = this.haveNeedsForThisPeriod(this.currentPaymentPeriod)
         .filter(entry => entry.name !== this.ourUsername)
-      let othersIncomeNeeded = haveNeeds.reduce(
+      const othersIncomeNeeded = haveNeeds.reduce(
         (accu, entry) => entry.haveNeed < 0 ? accu + (-1 * entry.haveNeed) : accu, 0
       )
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Beautifully done @SebinSong! Approved!!

PR #1688, opened 2023-08-15
#1686 - Fix a few bugs in  payment-detail modal

### LINKED ISSUE #1686: Bug with the Payment Page and Payment Details modal
### Problem

Scenario: after updating my income details on the live app, which involved switching status from haver to needer, I go to the `Payments` page to check my payment history. In the `Completed` tab, I have two completed payments I've made as a pledger.
- They now appear as "Sent by" the receiver, instead of "Sent to".
- In the first payment row, in the three-dots menu, clickling on `Payment details` doesn't reveal any information
  Instead it automatically switches the view to the `Received` tab, instead of staying in the `Completed` tab, and the following warning shows up in the console:
  
image

- In the other payment row, performing the same action does bring up the `Payment details` modal though, but it now offers a "Send Thanks" option, although I was the payment sender not the receiver:

image
The previous warning doesn't pop up again in that case.
- Then when I close the modal, I'm again automatically taken to the `Received` tab.
### Solution

Find and fix!

SebinSong:
closes #1686 

This PR contains bug-fixes for below items:

1. The payment table in `Completed` tab displays `Sent by` as a label when it's supposed to say `Sent to`
2. Upon closure of `PaymentDetail.vue` modal, the active tab gets always reset to the first tab, instead of remaining in the previously active tab.
3. In the `PaymentDetail` modal, `Send thank you` button is enabled even when I was the sender of the transaction.
4. In the first payment row, in the three-dots menu, clickling on Payment details doesn't reveal any information.

### DIFF
--- a/frontend/views/containers/payments/PaymentsMixin.js
+++ b/frontend/views/containers/payments/PaymentsMixin.js
@@ -41,7 +41,7 @@ const PaymentsMixin: Object = {
               for (const hash of paymentsFrom[fromUser][toUser]) {
                 if (hash in payments) {
                   const { data, meta } = payments[hash]
-                  paymentsInTypes[receivedOrSent].push({ hash, data, meta, amount: data.amount, username: toUser })
+                  paymentsInTypes[receivedOrSent].push({ hash, data, meta, amount: data.amount, username: toUser, period })
                 } else {
                   console.error(`getHistoricalPaymentsInTypes: couldn't find payment ${hash} for period ${period}!`)
                 }
@@ -89,6 +89,8 @@ const PaymentsMixin: Object = {
     async getHistoricalPaymentByHashAndPeriod (hash: string, period: string) {
       const paymentsKey = `payments/${this.ourUsername}/${period}/${this.currentGroupId}`
       const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
+
+      console.log(`@@ payments for the period [${period}]: `, payments)
       return payments[hash]
     },
     async getHaveNeedsSnapshotByPeriod (period: string) {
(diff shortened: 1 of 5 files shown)


SebinSong:
Am going to get back to fixing the test failure in 24 hrs

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -89,6 +89,8 @@ const PaymentsMixin: Object = {
     async getHistoricalPaymentByHashAndPeriod (hash: string, period: string) {
       const paymentsKey = `payments/${this.ourUsername}/${period}/${this.currentGroupId}`
       const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
+
+      console.log(`@@ payments for the period [${period}]: `, payments)
```
taoeffect: Think this debug logging needs to be removed before we can merge, right?

frontend/views/pages/Payments.vue
```diff
@@ -226,7 +226,14 @@ export default ({
         if (section && this.tabSections.includes(section)) {
           this.ephemeral.activeTab = section
         } else {
-          const defaultTab = this.tabSections[0]
+          const fromQuery = from?.query || {}
+          const isFromPaymentDetailModal = fromQuery.modal === 'PaymentDetail'
+          const defaultTab = isFromPaymentDetailModal
+            // When payment detail modal is closed, the payment table has to remain in the previously active tab.
+            // (context: https://github.com/okTurtles/group-income/issues/1686)
```
taoeffect, line 233: Thanks for adding this context in the comment 👍

taoeffect (changes requested):
Excellent work @SebinSong! As far as I can tell all of the issues @snowteamer found have been addressed. 👏

However, I found one more issue that we should take the time to fix now: old sent payments from previous periods have a "Cancel payment" button:

Screen Shot 2023-08-17 at 11 18 32 AM

Screen Shot 2023-08-17 at 11 24 36 AM

This doesn't make sense to show for payments that have been completed and are in an older payment period than the current one.

So can we hide the button and the menu item in such cases?

SebinSong:
@taoeffect Updated the PR again with the work for your feedback.

taoeffect (approved):
Fantastic work @SebinSong! Approved!

PR #1691, opened 2023-08-16
Support variable distribution period length

### LINKED ISSUE #1631: Store payment periods as a doubly linked list

### LINKED ISSUE #1662: Vue exception in console after leaving group
### Problem

When a user leaves a group, this exception appears in the console.

Screen Shot 2023-07-06 at 11 46 13 AM

### Solution

The solution might be as simple as modifying this code in the group contract:

```js
    groupProposalSettings (state, getters) {
      return (proposalType = PROPOSAL_GENERIC) => {
        return getters.groupSettings.proposals[proposalType]
      }
    },
```

To:

```js
    groupProposalSettings (state, getters) {
      return (proposalType = PROPOSAL_GENERIC) => {
        return getters.groupSettings?.proposals[proposalType]
      }
    },
```

snowteamer:
Closes #1631
Closes #1662


groupjscopy.js
snowteamer, line 1: Forgot to remove this file, will delete it as part of next actual commit

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
     groupMincomeCurrency (state, getters) {
       return getters.groupSettings.mincomeCurrency
     },
+    // Oldest period key first.
+    groupSortedPeriodKeys (state, getters) {
+      // The .sort() call might be only necessary in older browser which don't maintain object key ordering.
+      // A comparator function isn't required for now since our keys are ISO strings.
+      return Object.keys(getters.currentGroupState.paymentsByPeriod ?? {}).sort()
```
taoeffect: Nice comments 👍

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+    // Returns either the known period stamp for the given date,
+    // or the predicted one according to the period length.
+    // May return 'undefined', in which case the caller should check archived data.
     periodStampGivenDate (state, getters) {
       return (recentDate: string | Date) => {
-        if (typeof recentDate !== 'string') {
-          recentDate = recentDate.toISOString()
-        }
+        if (typeof recentDate !== 'string') recentDate = recentDate.toISOString()
+        if (!isPeriodStamp(recentDate)) throw new TypeError('must be date or isostring')
```
taoeffect: This check doesn't make sense, because `recentDate` is conceptually not a period stamp, it's a date, so please remove it.

You can convert a date to a periodstamp using `dateToPeriodStamp` if you need to.

The reason why we have these functions and must use them is because even though dates and period stamps happen to be the same thing, they are conceptually different. Treating them as the different types that they are and using the corresponding conversation functions allows us to change the format of a period stamp in the future if we need to.

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
         const { distributionDate, distributionPeriodLength } = getters.groupSettings
-
-        if (!distributionDate) return null
-
-        return periodStampGivenDate({
-          recentDate,
-          periodStart: distributionDate,
-          periodLength: distributionPeriodLength
-        })
+        if (!distributionDate) return
```
taoeffect: Is there a reason you changed the return type from `null` to `undefined` here? If not, can you revert it back?

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        const sortedPeriodKeys = getters.groupSortedPeriodKeys
+        // Maybe the distribution date has just been updated,
+        // but the corresponding period has not started or is not stored yet.
+        // So we try to use the distribution date first.
+        if (recentDate >= distributionDate) {
```
taoeffect: Here two dates are being compared, so this is OK I think. ISO Strings happen to be safe to compare this way, I think. But if they aren't then it's better to compare them after they're converted to `Date`s

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+          // Only extrapolate one period length in the future.
+          const extrapolatedDistributionDate = addTimeToDate(
+            distributionDate, distributionPeriodLength
+          ).toISOString()
+          if (recentDate >= extrapolatedDistributionDate) {
+            return dateToPeriodStamp(extrapolatedDistributionDate)
+          }
+          return dateToPeriodStamp(distributionDate)
```
taoeffect: This code should be replaced by the previous logic we had in the `periodStampGivenDate` function in `time.js`, so that it can extrapolate multiple periods out. There's no reason not to, and we have the code for that already.

I would recommend re-introducing that function in `time.js`, and passing in `sortedPeriodKeys` to it. That would keep this function here short & sweet while having the messy logic in `time.js`.

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        }
+        // For dates before the distribution date, we check the stored data first,
+        // as simply substracting the default period length won't give the previous period stamp
+        // if the distribution date was updated during that period.
+        if (!sortedPeriodKeys.length) {
+          // Looks like we're in the waiting period but haven't stored it yet.
+          const waitingPeriodStamp = dateToPeriodStamp(
+            addTimeToDate(distributionDate, -distributionPeriodLength)
+          )
+          return recentDate >= waitingPeriodStamp ? waitingPeriodStamp : undefined
```
taoeffect: This line is comparing a date with a period stamp. Conceptually, these are intended to be two different types so you can't use `>=` here. You have to convert `recentDate` to a periodstamp itself and then use `comparePeriodStamps` (or convert the `waitingPeriodStamp` to a date)

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        }
+        const oldestKnownStamp = sortedPeriodKeys[0]
+        const latestKnownStamp = sortedPeriodKeys[sortedPeriodKeys.length - 1]
+        if (recentDate < oldestKnownStamp) return
+        if (recentDate >= latestKnownStamp) {
+          const extrapolatedPeriodStamp = dateToPeriodStamp(
+            addTimeToDate(dateFromPeriodStamp(latestKnownStamp), distributionPeriodLength)
+          )
+          if (recentDate >= extrapolatedPeriodStamp) return extrapolatedPeriodStamp
```
taoeffect: I am confused, isn't `sortedPeriodKeys` supposed to contain all recent period stamps? If so, can't think `if` block can be deleted and replaced with the loop below?

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        // This should not happen
       }
     },
+    // May return 'undefined', in which case the caller should check archived data.
     periodBeforePeriod (state, getters) {
       return (periodStamp: string) => {
-        const len = getters.groupSettings.distributionPeriodLength
-        return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), -len))
+        if (!isPeriodStamp(periodStamp)) throw new TypeError('must be periodStamp')
+        const { distributionDate, distributionPeriodLength } = getters.groupSettings
```
taoeffect: Is it possible to somehow simplify and reuse some of this code in `periodBeforePeriod` within `periodStampGivenDate` (or vice versa)?

taoeffect (changes requested):
Nicely done @snowteamer!

Preliminary review ready for you! I didn't have time to do a full review this time and had to top short after reviewing `periodStampGivenDate`, but did my best to get what I had out to you.

frontend/model/contracts/group.js
snowteamer: Agreed, and in fact I was looking for an `isIsoString` function, but there is no such function in `time.js`, and adding one would have duplicated the code in `isPeriodStamp`. So I just used that

frontend/model/contracts/group.js
snowteamer: Having APIs wich return `null` rather than `undefined` is generally considered a bad idea AFAIK, and the change was easy to do

frontend/model/contracts/group.js
snowteamer: I think that's safe too, as long as we don't go past year 9999

frontend/model/contracts/group.js
taoeffect: > Having APIs wich return null rather than undefined is generally considered a bad idea AFAIK

🤔 Really? Why is that? I'm used to C and Java where returning `null` or `NULL` is common. It's JS that's weird with its two different "null" values...

frontend/model/contracts/group.js
snowteamer: Maybe you're right, but I haven't seen any use case where extrapolating more than one period in the future would be actually useful. I thought it didn't make much sense.

frontend/model/contracts/group.js
snowteamer: We could indeed get rid of the `if (recentDate < oldestKnownStamp) return` line by taking advantage of the fact that `sortedPeriodKeys[-1]` is always undefined. But that's a bit less readable I think, and likely won't play nice with typechecking when we move to TS.

As for the `if (recentDate >= latestKnownStamp` check, it isn't covered by the following loop.

frontend/model/contracts/group.js
snowteamer: IMO that has mainly to do with the handling of default function parameters.
For example if you have a `foo (arg = defaultValue) => { ... } function`, then `defaultValue` will be used if `arg` is undefined, but won't if `arg` is null. Then inside the function you'll likely have to check again whether `arg` is null.

This [link](https://www.split.io/blog/bias-against-null-javascript/) is also interesting

taoeffect: That's a great link! Thanks for sharing!

frontend/model/contracts/group.js
taoeffect: Say the user has their clock accidentally incorrectly set a year in the past. They will be browsing old payments they sent and using this getter to fetch the period in which the payment occurred. This getter will return the wrong value. Why should we return the wrong value when I've already written the code to return the correct value?

frontend/model/contracts/group.js
taoeffect: @snowteamer ?

For example, `periodStampGivenDate` iterates backwards in the periods, but it could be calling this getter instead to that, couldn't it?

And the same goes for `periodAfterPeriod`

frontend/model/contracts/shared/time.js
```diff
@@ -132,13 +132,17 @@ export function humanDate (
...
 export function isPeriodStamp (arg: string): boolean {
-  return /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(arg)
+  return isIsoString(arg)
 }
 
 export function isFullMonthstamp (arg: string): boolean {
   return /^\d{4}-(0[1-9]|1[0-2])$/.test(arg)
 }
 
+export function isIsoString (arg: string): boolean {
```
taoeffect: JS calls it an `ISOString` not an `IsoString`, so maybe we should change this to match?

frontend/model/contracts/group.js
```diff
@@ -1262,8 +1337,11 @@ sbp('chelonia/defineContract', {
       process ({ meta }, { state, getters }) {
         const period = getters.periodStampGivenDate(meta.createdDate)
         const current = getters.groupSettings?.distributionDate
-
-        if (current !== period) {
+        const inWaitingPeriod = !current || new Date().toISOString() < current
+        // Maybe we're updating the distribution date while in the waiting period.
+        if (inWaitingPeriod && meta.createdDate !== current) {
```
taoeffect: `new Date()` is never allowed inside of `process` functions (remember?)

Please check out PR #1643 for how to this, and also please merge `master` as that PR was recently merged.

taoeffect (changes requested):
Nice work @snowteamer! Preliminary review ready!

frontend/model/contracts/group.js
snowteamer: This block is only called when `recentDate >= distributionDate`, and I don't think `distributionDate` can depend on the user's clock.

Therefore in case the clock incorrectly thinks it's already several period lengths in the future, way past the distribution date, so that this block gets called, then there just won't be any matching payment to be displayed but I think that's correct.

If the user has their clock accidentally incorrectly set a year in the past, so that `recentDate` is one year lower than expected, then this block won't be called. The sorted periods will be searched instead, and I think the user should still be able to browse any stored payment period up to the distribution date.

frontend/model/contracts/shared/time.js
snowteamer: We could, and I've quickly considered it before writing this function, but I would advise slightly against it

- Even though remembering and respecting the correct letter-casing of the various names is not that hard, it's easier to just camel-case everything.
- Too many uppercase acronyms can quickly become ugly and add visual clutter, except maybe in suffix position. For example `JSONRPCCall` isn't that readable. Even with just one acronym per name like in `JSONData`, if that name is used several times in a block it will add more clutter than `jsonData`
- Uppercase acronyms will yield names easily confused with constructor names because of a starting uppercase letter (that's not the case here, but well in declarations like `let ISOString = '...Z'`. We likely don't want that).

snowteamer: @taoeffect 
> For example, periodStampGivenDate iterates backwards in the periods, but it could be calling this getter instead to that, couldn't it?

True, but please note that doing so might make removing `periodAfter/beforePeriod` in the future more difficult

frontend/model/contracts/group.js
```diff
@@ -348,37 +450,6 @@ sbp('chelonia/defineContract', {
...
-              console.warn(`paymentTotalFromUserToUser: super old payment shouldn't exist, ignoring! (curPeriod=${periodStamp})`, JSON.stringify(payment))
-              return a
-            }
-            exchangeRate *= periodPayments[paymentCreatedPeriodStamp].mincomeExchangeRate
-          }
-          return a + (amount * exchangeRate * mincomeExchangeRate)
-        }, 0)
-        return saferFloat(total)
-      }
-    },
```
taoeffect, line 381: This code is potentially useful. Even though we're not using it now, we might use it in the future.

If you really want to remove it for some reason, please don't remove all of it. What you can do instead is something like the following:

```
// paymentTotalFromUserToUser (state, getters) {
// // this code was removed in https://github.com/okTurtles/group-income/pull/1691
// // because it was unused. feel free to bring it back if needed.
// }
```

test/cypress/integration/group-paying.spec.js
```diff
@@ -87,6 +95,38 @@ describe('Group Payments', () => {
...
+      expect(periodAfterPeriod(onePeriodLengthAhead) === undefined).to.be.true
+      expect(periodAfterPeriod(onePeriodLengthBefore) === distributionDate).to.be.true
+
+      expect(periodBeforePeriod(distributionDate) === onePeriodLengthBefore).to.be.true
+      expect(periodBeforePeriod(onePeriodLengthAhead) === distributionDate).to.be.true
+      expect(periodBeforePeriod(onePeriodLengthBefore) === undefined).to.be.true
+      /* eslint-enable no-unused-expressions */
+    })
+    cy.giLogout()
+  })
```
taoeffect: The tests were just logged in as user1, then they logged out, and here it's logging in and logging out again. This is going to increase the test time unnecessarily, so can you move these checks into the previous test?

taoeffect (changes requested):
Minor change requests for tonight, still reviewing...

frontend/model/contracts/group.js
snowteamer: This code is still present in `historical/distribution-algo/real-time-algo/group.js`

Besides, we can always retreive older code via git (if we know where to find it)

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+          const extrapolatedPeriodStamp = dateToPeriodStamp(
+            addTimeToDate(dateFromPeriodStamp(latestKnownStamp), distributionPeriodLength)
+          )
+          if (recentDate >= extrapolatedPeriodStamp) return extrapolatedPeriodStamp
+          return latestKnownStamp
+        }
+        for (let i = 1; i < sortedPeriodKeys.length; i++) {
+          if (recentDate < sortedPeriodKeys) return sortedPeriodKeys[i - 1]
+        }
+        // This should not happen
```
Silver-IT: I think it should be `if (recentDate < sortedPeriodKeys[i]) return sortedPeriodKeys[i - 1]`.

frontend/model/contracts/group.js
Silver-IT: I think we don't need to make `extrapolatedPeriodstamp` here, and it should be 
```
if (recentDate >= latestKnownStamp) return latestKnownStamp
```

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        const distributionDateStamp = dateToPeriodStamp(distributionDate)
+        const sortedPeriodKeys = getters.groupSortedPeriodKeys
+        const latestKnownStamp = sortedPeriodKeys[sortedPeriodKeys.length - 1]
+        // Maybe the distribution date has just been updated,
+        // but the corresponding period has not started or is not stored yet.
+        // So we try to use the distribution date first.
+        if (periodStamp > distributionDateStamp) {
+          // Only extrapolate one period length in the future.
+          const onePeriodLenghtAhead = addTimeToDate(distributionDate, distributionPeriodLength)
+          if (periodStamp === dateToPeriodStamp(onePeriodLenghtAhead)) return distributionDate
```
Silver-IT: Typo here. `onePeriodLengthAhead`, not `onePeriodLenghtAhead`.

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -36,14 +36,20 @@ import { mapGetters } from 'vuex'
...
   components: {
     ProgressBar
   },
   methods: {
+    getDueDate () {
+      return addTimeToDate(this.getStartDate(), this.groupSettings.distributionPeriodLength)
+    },
+    getStartDate () {
+      return this.periodStampGivenDate(new Date())
+    },
```
Silver-IT: Is this `currentPaymentPeriod`?

Silver-IT (changes requested):
Impressive work, @snowteamer. I didn't fully review this PR yet, but I did a preliminary review.

frontend/model/contracts/group.js
taoeffect: @snowteamer yeah, but that's the thing, by the time someone needs this getter in the future, they won't know where to look to find it because they won't know that it existed. They might try to implement it on their own not knowing how it should be done (and that it was already done for them). So can you please add that comment reference?

frontend/model/contracts/group.js
taoeffect: DRY: replace `getters.currentGroupState.paymentsByPeriod ?? {}` with `getters.groupPeriodPayments` which does the same thing.

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -12,14 +12,75 @@ const PaymentsMixin: Object = {
...
+      return await this.getPeriodStampGivenDate(payment.date)
+    },
+
+    // Oldest key first.
+    async getSortedPeriodKeys () {
+      const key = `paymentsByPeriod/${this.ourUsername}/${this.currentGroupId}`
+      const archivedPaymentsByPeriod = await sbp('gi.db/archive/load', key) || {}
+      return [
+        ...Object.keys(archivedPaymentsByPeriod).sort(),
+        ...Object.keys(this.groupPeriodPayments).sort()
```
taoeffect: DRY: Replace `Object.keys(this.groupPeriodPayments).sort()` with `this.groupSortedPeriodKeys`

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -12,14 +12,75 @@ const PaymentsMixin: Object = {
...
   },
   methods: {
+    async getPeriodStampGivenPayment (payment) {
+      return await this.getPeriodStampGivenDate(payment.date)
+    },
+
+    // Oldest key first.
+    async getSortedPeriodKeys () {
+      const key = `paymentsByPeriod/${this.ourUsername}/${this.currentGroupId}`
+      const archivedPaymentsByPeriod = await sbp('gi.db/archive/load', key) || {}
```
taoeffect: This action is repeated 4 times in this file.

Let's do some boyscoutting here and DRY this by creating a method to fetch the archived payments by period. You can even call it `async archivedPaymentsByPeriod () { ... }`.

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -12,14 +12,75 @@ const PaymentsMixin: Object = {
...
+      'paymentHashesForPeriod'
     ])
   },
   methods: {
+    async getPeriodStampGivenPayment (payment) {
+      return await this.getPeriodStampGivenDate(payment.date)
+    },
+
+    // Oldest key first.
+    async getSortedPeriodKeys () {
```
taoeffect: To differentiate from `this.groupSortedPeriodKeys`, please rename to either `getAllSortedPeriodKeys` or `getHistoricalSortedPeriodKeys`

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -36,14 +36,20 @@ import { mapGetters } from 'vuex'
...
 
 export default ({
   name: 'MonthOverview',
   components: {
     ProgressBar
   },
   methods: {
+    getDueDate () {
+      return addTimeToDate(this.getStartDate(), this.groupSettings.distributionPeriodLength)
+    },
```
taoeffect: DRY: use `this.dueDateForPeriod(this.currentPaymentPeriod)` as is done elsewhere

frontend/model/contracts/group.js
taoeffect: Hmm, ok, that does seem to be true. Still, I don't see why we can't include a loop here that was already written, and therefore not need to worry about this returning an incorrect value.

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+        }
+        // For dates before the distribution date, we check the stored data first,
+        // as simply substracting the default period length won't give the previous period stamp
+        // if the distribution date was updated during that period.
+        if (!sortedPeriodKeys.length) {
+          // Looks like we're in the waiting period but haven't stored it yet.
+          const waitingPeriodStamp = dateToPeriodStamp(
+            addTimeToDate(distributionDate, -distributionPeriodLength)
+          )
+          return recentDate >= dateFromPeriodStamp(waitingPeriodStamp).toISOString() ? waitingPeriodStamp : undefined
```
taoeffect: Hmm... you can replace `dateFromPeriodStamp(waitingPeriodStamp).toISOString()` with `waitingPeriodStamp`

frontend/model/contracts/group.js
taoeffect: I agree with @Silver-IT 👍

frontend/model/contracts/group.js
snowteamer: Wow, can't believe no test caught this - thanks

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+    // May return 'undefined', in which case the caller should check archived data.
     periodBeforePeriod (state, getters) {
       return (periodStamp: string) => {
-        const len = getters.groupSettings.distributionPeriodLength
-        return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), -len))
+        if (!isPeriodStamp(periodStamp)) throw new TypeError('must be periodStamp')
+        const { distributionDate, distributionPeriodLength } = getters.groupSettings
+        if (!distributionDate) return
+        // This is not always the current period stamp.
+        const distributionDateStamp = dateToPeriodStamp(distributionDate)
```
taoeffect: I don't think you need to do this conversion, because `distributionDate` is not a `Date` object. It is in fact a period stamp that you can use directly.

frontend/views/containers/payments/MonthOverview.vue
snowteamer: It should be but `currentPaymentPeriod` is a bit special in that it's controlled by reactiveDate, which doesn't update in real time but in interval. So if I understand correctly it will sometimes be wrong relative to the actual current date, therefore I chose to not rely on it in other computed properties

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+          // but is one period length ahead of a known one.
+          // TODO: just return 'undefined' when sure it's always safe.
+          const maybePreviousStamp = dateToPeriodStamp(
+            addTimeToDate(dateFromPeriodStamp(periodStamp), -distributionPeriodLength)
+          )
+          return sortedPeriodKeys.includes(maybePreviousStamp) ? maybePreviousStamp : undefined
+        }
+        // If index is 0 then the caller will have to check the archive.
+        if (index === 0) return
+        return sortedPeriodKeys[index - 1]
```
taoeffect: I do not understand what is going on here or why this code is so complicated.

AFAICT this is what this function should be:

```js
    periodBeforePeriod (state, getters) {
      return (periodStamp: string) => {
        const { distributionPeriodLength } = getters.groupSettings
        const sortedPeriods = getters.groupSortedPeriodKeys
        for (const i = sortedPeriods.length - 1; i >= 0; --i) {
          const latestPeriod = sortedPeriods[i]
          if (periodStamp >= latestPeriod) return latestPeriod
        }
        return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), -distributionPeriodLength))
      }
    }
```

frontend/model/contracts/group.js
snowteamer: > I think we don't need to make extrapolatedPeriodstamp here

I'd be slightly reluctant to no longer do this, but if a majority thinks otherwise, then OK 👍

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,138 @@ sbp('chelonia/defineContract', {
...
+            addTimeToDate(distributionDate, -distributionPeriodLength)
+          )
+          return periodStamp === waitingPeriodStamp ? distributionDate : undefined
+        }
+        const index = sortedPeriodKeys.indexOf(periodStamp)
+        if (index === -1) return
+        // Maybe the given stamp is the last stored one but doesn't match the distribution date.
+        if (index === sortedPeriodKeys.length - 1) return dateToPeriodStamp(distributionDate)
+        // Now 'index + 1' is always a valid index.
+        return sortedPeriodKeys[index + 1]
```
taoeffect: Similarly, this seems like it can be simplified to:

```js
    periodAfterPeriod (state, getters) {
      return (periodStamp: string) => {
        const { distributionPeriodLength } = getters.groupSettings
        const sortedPeriods = getters.groupSortedPeriodKeys
        let latestPeriod
        for (const i = sortedPeriods.length - 1; i >= 0; --i) {
          latestPeriod = sortedPeriods[i]
          if (periodStamp >= latestPeriod) break
        }
        const onePeriodAfterLatest = dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(latestPeriod), distributionPeriodLength))
        if (periodStamp >= onePeriodAfterLatest) {
          return dateToPeriodStamp(addTimeToDate(dateFromPeriodStamp(periodStamp), distributionPeriodLength))
        }
        return latestPeriod
      }
    },
```

frontend/model/contracts/group.js
```diff
@@ -1286,8 +1361,11 @@ sbp('chelonia/defineContract', {
       process ({ meta }, { state, getters }) {
         const period = getters.periodStampGivenDate(meta.createdDate)
         const current = getters.groupSettings?.distributionDate
-
-        if (current !== period) {
+        const inWaitingPeriod = !current || new Date().toISOString() < current
```
taoeffect: Another `new Date()`!

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -12,14 +12,75 @@ const PaymentsMixin: Object = {
...
+      const key = `paymentsByPeriod/${this.ourUsername}/${this.currentGroupId}`
+      const archivedPaymentsByPeriod = await sbp('gi.db/archive/load', key) || {}
+      return [
+        ...Object.keys(archivedPaymentsByPeriod).sort(),
+        ...Object.keys(this.groupPeriodPayments).sort()
+      ]
+    },
+
+    // Note: 'recentDate' is a confusing name, as it can be in the future, or far in the past.
+    async getPeriodStampGivenDate (givenDate: string | Date) {
```
taoeffect: Rename to `historicalPeriodStampGivenDate`

taoeffect (changes requested):
Another preliminary review. Will wait until everything is addressed in previous reviews before proceeding with another review

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -111,6 +115,9 @@ export default ({
         console.error(e)
         alert(e.message)
       }
+    },
+    async updatePayment () {
+      this.payment.relativeTo = await this.getPeriodStampGivenDate(this.payment.date)
```
Silver-IT: I think since the `relativeTo` is reactive to the `payment` prop, we shouldn't set it only in the `created` lifecycle. Since the function `getPeriodStampGivenData` is async function, we could setup watch for `payment` prop.

```
data () {
    relativeTo: undefined
},
...
watch: {
    payment (to, from) {
        this.updatePaymentRelativeTo(to.date)
    }
},
methods: {
    async updatePaymentRelativeTo (paymentDate) {
        this.relativeTo = await this.getPeriodStampGivenData(paymentDate)
    }
}
```

frontend/views/containers/payments/PaymentRowSent.vue
```diff
@@ -106,6 +109,9 @@ export default ({
         console.error(e)
         alert(e.message)
       }
+    },
+    async updatePayment () {
+      this.payment.relativeTo = await this.getPeriodStampGivenDate(this.payment.date)
```
Silver-IT: Same as above.

Silver-IT (changes requested):
Finished my review. Nice work, @snowteamer

frontend/model/contracts/group.js
snowteamer: I thought `distributionDate` was an ISO date string, which happens to also be a period stamp but is conceptually a different thing (as you explained to me). It made sense to me since `distributionDate` comes from the group settings, whereas "period stamp" is a more internal type I think.

So anywhere I had the chance to compare a `distributionDate` value against a date directly, I didn't do any conversion.
I will likely now have to add a few of such conversions there if I start to consider `distributionDate` as being of the "period stamp" type.

frontend/model/contracts/group.js
snowteamer: Hmm I'm pretty sure I had fixed this one, using `meta.createdDate` instead. Seems like I somehow overwrote my own changes using git incorrectly 😭

frontend/views/containers/payments/PaymentsMixin.js
snowteamer: It can also return non-historical results, so I think it would be a bit confusing

frontend/model/contracts/group.js
snowteamer: This code is indeed a bit complicated, in part because of the `periodStamp === distributionDateStamp` case, so I've tried to explain it with a a lot of comments (but failed to make it clear enough)

> this is what this function should be:

If I'm not mistaken, this won't ever return `undefined` to defer searching to the archive code. And since every archived period could have a different length, simply substracting `distributionPeriodLength` is not correct in that case.
Also it doesn't do anything special for the aforementioned special case.

Would be great if that case was actually not special and we could just use a simple loop like yours, but for now the suggested code fails this assertion line 118 in group-paying.spec.js
```js
      expect(periodBeforePeriod(onePeriodLengthAhead) === distributionDate).to.be.true
```
(not 100% sure why but my guess it's because in that case there exists a stored object for the waiting period but not for the current distribution date yet)

frontend/model/contracts/group.js
taoeffect: @snowteamer I think we made a mistake in how we named that variable... it is in fact a period stamp... and it happens to be a date represented as an ISO string.

Maybe you can add a comment to the constructor to that effect? (Renaming it now is not a good idea since there are live groups running).

Also, it's a bit confusing with how we've implemented `dateToPeriodStamp`, which takes either a `Date` or a `string`:

```js
export function dateToPeriodStamp (date: string | Date): string {
  return new Date(date).toISOString()
}
```

And yet `dateFromPeriodStamp` only returns a `Date`:

```js
export function dateFromPeriodStamp (daystamp: string): Date {
  return new Date(daystamp)
}
```

But yeah, in the end, `distributionDate` should be treated as a period stamp :-\

frontend/model/contracts/group.js
taoeffect: OK, that's a good point regarding `distributionDate` - my code doesn't check for that because it assumes (incorrectly) that that is already stored in `sortedPeriods`, which you're absolutely right, it might not be.

Also, my code has a bug related to this check:

```js
if (periodStamp >= latestPeriod) return latestPeriod
```

I think that should be:

```js
if (periodStamp > latestPeriod) return latestPeriod
```

The code that's currently here in the PR needs to be changed for a few reasons:

1. As discussed above, we need to avoid passing `distributionDate` to `dateToPeriodStamp`
2. It needs to be simplified similarly to what I wrote above. Right now it's got some very questionable lines, like this:

```js
        if (periodStamp > distributionDateStamp) {
          // Only extrapolate one period length in the future.
          const onePeriodLenghtAhead = addTimeToDate(distributionDate, distributionPeriodLength)
          if (periodStamp === dateToPeriodStamp(onePeriodLenghtAhead)) return distributionDate
          else return
        }
```

This is very strange. The purpose of this function is to return the period before a given period stamp. That's not what the lines above do at all. Instead they seemingly randomly add a period length to `distributionDate` and compare `periodStamp` to that, and then bails if the check fails.

We need something conceptually equivalent to the code I wrote above that:

1. Takes into account `distributionDate`
2. Has a consistent mental model for how period stamps are created and iterated forward/backward

## Mental model

Our mental model is as follows:

```
<--- [ UNKNOWN PAST ] --- [ SORTED_KEYS ] ---- [ UNKOWN FUTURE ] --->
```

Here we see 3 section. There is a "known" timeline in the center, and an unknown timeline at each end.

If we are given a period stamp that is inside of `[ SORTED_KEYS ]`, here's what iterating it looks like:

```
[ A ] <-> [ B ] <-> [ C ]
            ^
            |
      periodStamp
```

If we want to go forward in time from `periodStamp` pointing at `[ B ]` stamp, then `periodAfterPeriod` points us to `[ C ]` and `periodBeforePeriod` points us to `[ A ]`.

If, however, we have this situation:

```
[ B ] <-> [ C ] <-> [ UNKNOWN FUTURE ]
            ^
            |
      periodStamp
```

Then:

- `periodAfterPeriod` adds `distributionPeriodLength` to `periodStamp` and returns the result
- `periodBeforePeriod` returns `[ B ]`

And, if we have this situation:

```
[ UNKNOWN PAST ] <-> [ A ] <-> [ B ]
                       ^
                       |
                   periodStamp
```

Then:

- `periodAfterPeriod` returns `[ B ]`
- `periodBeforePeriod` returns `undefined`

Note: that this is the **only scenario** where `undefined` can be returned.

Also note: in that case `historicalPeriodBeforePeriod` (please rename from currently called `getPeriodBeforePeriod`) **does the exact same logic as `periodBeforePeriod`**, except it performs the check on a version of `[ SORTED_KEYS ]` that includes historical payments from then archive.

In fact, because both `periodBeforePeriod` and `historicalPeriodBeforePeriod` contain line-by-line exactly the same logic, please create a shared function inside of `time.js` that does this logic. When `historicalPeriodBeforePeriod` and `periodBeforePeriod` call that function, they will include an array containing the `[ SORTED_KEYS ]`.

Finally, to take into account the `distributionDate`, simply add in the `distributionDate` to the `[ SORTED_KEYS ]` that you pass in to the internal `periodBeforePeriod` inside of `time.js`.

And similarly, there will be a corresponding shared `periodAfterPeriod` in `time.js` that is called from both `group.js` and `PaymentsMixin.js`.

Hope that's clear! Let me know if you have any questions.

frontend/views/containers/payments/PaymentsMixin.js
taoeffect: True, but the point is that it unlike `periodStampGivenDate` it *can* give historical payment periods, so that's why it needs to be named that way. See also the [mental mode](https://github.com/okTurtles/group-income/pull/1691#discussion_r1327635564) comment I just posted.

frontend/model/contracts/group.js
snowteamer: > This is very strange. The purpose of this function is to return the period before a given period stamp.

Exactly, so if the given date is not actually a past, current or future period stamp but just a random ISO string, my assumption was that `undefined` had to be returned.

If that assumption shoud be dropped so that these getters behave like `periodStampGivenDate`, e.g. `periodAfterPeriod` returns `B` for any date between `A` and `B` not just for `A`, then maybe they could as well be renamed for consistency:
- nextPeriodStampGivenDate
- periodStampGivenDate
- previousPeriodStampGivenDate
(I'd rather use -forDate than -givenDate). Otherwise `periodAfter/BeforeDate` also sound good to me.

Another assumption in this PR, is that `distributionPeriodLength` is now a variable setting whose value only holds for the current period. It can no longer be safely used to iterate between existing periods, but is just a hint to initially setup a default next distribution date.
Moreover, in the waiting period the distribution date i.e. the next period stamp can be updated without updating `distributionPeriodLength` accordingly

frontend/model/contracts/group.js
snowteamer: Ok, thanks for explaining! 😄

frontend/views/containers/payments/PaymentsMixin.js
snowteamer: I'm not fan of dropping the `get-` prefix here, as all other methods in this file are `get-` prefixed. It likely helps to distinguish them from computed properties. But I also think `getHistoricalPeriodStampGivenDate` would be a tad long...

snowteamer: Should all `PaymentsMixin.js` methods that can return both in-memory and historical results be named that way (some currently don't)?

frontend/model/contracts/group.js
snowteamer: Updated to handle future dates using the period length in a loop like before (re-using `periodStampGivenDate` in `time.js`)

frontend/views/containers/payments/PaymentsMixin.js
taoeffect: Yeah feel free to make it consistent and clear 👍 
(I haven't checked your latest changes as of this comment but will have a look now)

frontend/model/contracts/group.js
```diff
@@ -312,32 +312,53 @@ sbp('chelonia/defineContract', {
...
+    groupSortedPeriodKeys (state, getters) {
+      const { distributionDate, distributionPeriodLength } = getters.groupSettings
+      if (!distributionDate) return []
+      // The .sort() call might be only necessary in older browser which don't maintain object key ordering.
+      // A comparator function isn't required for now since our keys are ISO strings.
+      const keys = Object.keys(getters.groupPeriodPayments).sort()
+      // Append the waiting period stamp if necessary.
+      if (!keys.length && MAX_SAVED_PERIODS > 0) {
+        keys.push(dateToPeriodStamp(addTimeToDate(distributionDate, -distributionPeriodLength)))
+      }
```
taoeffect, line 333: Note: `!keys.length` should always be false because the contract constructor calls `initFetchPeriodPayments`, ensuring at least one entry

frontend/model/contracts/group.js
```diff
@@ -1286,8 +1276,11 @@ sbp('chelonia/defineContract', {
       process ({ meta }, { state, getters }) {
         const period = getters.periodStampGivenDate(meta.createdDate)
         const current = getters.groupSettings?.distributionDate
-
-        if (current !== period) {
+        const inWaitingPeriod = !current || meta.createdDate < current
```
taoeffect: I don't think this logic captures whether or not we're in the waiting period.

By "waiting period" I assume you mean in the period before the very first distribution has started. However, this logic doesn't check for that. We could be in in the 3rd period since the group started and this would still evaluate to true.

Whether or not we're in the first waiting period is determined really by the number of keys in `getter.groupPeriodPayments`. If there is only 1 key in there, and `meta.createdDate < current`, only then can we be considered to be `inWaitingPeriod`. And then you also don't need the redundant check of `meta.createdDate !== current` below:

frontend/model/contracts/shared/time.js
```diff
@@ -7,6 +7,48 @@ export const HOURS_MILLIS = 60 * MINS_MILLIS
 export const DAYS_MILLIS = 24 * HOURS_MILLIS
 export const MONTHS_MILLIS = 30 * DAYS_MILLIS
 
+const plusOnePeriodLength = (timestamp: string, periodLength: number): string => (
+  dateToPeriodStamp(addTimeToDate(timestamp, periodLength))
+)
+
+const minusOnePeriodLength = (timestamp: string, periodLength: number): string => (
+  dateToPeriodStamp(addTimeToDate(timestamp, -periodLength))
+)
```
taoeffect: These look useful, you could export these too...

frontend/model/contracts/shared/time.js
```diff
@@ -7,6 +7,48 @@ export const HOURS_MILLIS = 60 * MINS_MILLIS
...
+// - MUST be sorted in ascending order (lower timestamps first)
+// - MAY NOT contain duplicate elements
+export function periodStampsForDate (
+  date: Date | string,
+  { knownSortedStamps, periodLength }: { knownSortedStamps: string[], periodLength: number }
+): Object {
+  if (typeof date === 'string' && !isIsoString(date)) {
+    throw new TypeError('must be ISO string')
+  }
+  const timestamp = new Date(date).toISOString()
```
taoeffect: This is redundant when `date` is an iso string already. Perhaps this logic could be rewritten like so:

```js
const timestamp = typeof date === 'string' ? date : date.toISOString()
```

frontend/model/contracts/shared/time.js
```diff
@@ -7,6 +7,48 @@ export const HOURS_MILLIS = 60 * MINS_MILLIS
...
+)
+
+// @parameter knownSortedStamps:
+// - MUST be sorted in ascending order (lower timestamps first)
+// - MAY NOT contain duplicate elements
+export function periodStampsForDate (
+  date: Date | string,
+  { knownSortedStamps, periodLength }: { knownSortedStamps: string[], periodLength: number }
+): Object {
+  if (typeof date === 'string' && !isIsoString(date)) {
```
taoeffect: This can be rewritten to remove the redundant `typeof date === 'string'` check like so:

```js
if (!(isIsoString(date) || date instanceof Date)) {
```

(Note: for obsessive-compulsive "performance" reasons I've placed the check that's most likely to pass first, in this case the parameter is more likely to be an ISOString, so that check happens first to avoid triggering the `instanceof Date` check)

frontend/model/contracts/shared/time.js
```diff
@@ -7,6 +7,48 @@ export const HOURS_MILLIS = 60 * MINS_MILLIS
...
+  const timestamp = new Date(date).toISOString()
+  let previous, current, next
+  if (knownSortedStamps.length) {
+    const latest = knownSortedStamps[knownSortedStamps.length - 1]
+    if (timestamp >= latest) {
+      current = periodStampGivenDate({ recentDate: timestamp, periodStart: latest, periodLength })
+      next = plusOnePeriodLength(current, periodLength)
+      previous = current > latest
+        ? minusOnePeriodLength(current, periodLength)
+        : knownSortedStamps[knownSortedStamps.length - 2] ?? undefined
```
taoeffect: NOTE: After testing in node, deno, and 3 browsers, I've come to the conclusion that `?? undefined` is unnecessary as `undefined` will be returned in that case anyway.

If we are in this block there are 2 possibilities:

1. `timestamp = latest`
2. `timestamp > latest`

Following our [mental model](https://github.com/okTurtles/group-income/pull/1691#discussion_r1327635564), If `timestamp = latest`, then:

1. `current` should be `latest` (this should also be the value returned by `periodStampGivenDate` as is here)
2. `next` should be the value you're currently using with `plusOnePeriodLength`
3. `previous` should be `knownSortedStamps[knownSortedStamps.length - 2]`

If `timestamp > latest`, then:

1. `current` should be the value returned by `periodStampGivenDate` as is here (it could be either `latest` or 1 or more `periodLength`s in the future)
2. `next` should also be the value you're currently using here
3. `previous`: if `current = latest`, then `previous` would be either `knownSortedStamps[knownSortedStamps.length - 2]` (if that exists) or, it's implied that `current > latest`, and then `previous = latest`

frontend/model/contracts/shared/time.js
```diff
@@ -7,6 +7,48 @@ export const HOURS_MILLIS = 60 * MINS_MILLIS
...
+    } else {
+      for (let i = knownSortedStamps.length - 2; i >= 0; i--) {
+        if (timestamp >= knownSortedStamps[i]) {
+          current = knownSortedStamps[i]
+          // `i + 1` is always a valid index.
+          next = knownSortedStamps[i + 1]
+          previous = knownSortedStamps[i - 1] ?? undefined
```
taoeffect: - We don't need to `?? undefined` here (`undefined` is returned for us)
- Don't forget  to `break` out of the loop

frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -41,32 +39,18 @@ export default ({
...
-    }
+    ])
   },
   mounted () {
     this.updateHistory()
   },
   methods: {
     async updateHistory () {
-      this.history = await Promise.all(this.periods.map(async (period, i) => {
+      const periods = await this.getSortedPeriodKeys()
```
taoeffect: `this.getSortedPeriodKeys` doesn't exist!

(**Please** test code before pushing to make sure it works. I might not always be a good enough of a reviewer to catch stuff like this, but the computer will always catch it if you test these cases)

frontend/views/containers/payments/MonthOverview.vue
taoeffect: @snowteamer Alex is right, because Vue.js is not going to automatically update the UI when `new Date()` moves us into a new period. This is for 2 reasons:

1. This is a method, not a computed property
2. Even if it were a computed property, it still wouldn't work because computed properties are only updated in response to reactive changes. Since `new Date()` is not a reactive property itself, and is not updated as a result of a reactive Vue.js-related change (as `reactiveDate` is), the computed property doesn't get updated either because that code is never re-run

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -44,6 +44,12 @@ export default ({
     ProgressBar
   },
   methods: {
+    getDueDate () {
+      return this.dueDateForPeriod(this.currentPaymentPeriod)
+    },
```
taoeffect: This is supposed to be a computed property, not a method. Methods and computed properties are different. When computed properties change the UI gets updated automatically. A method cannot be bound to something as a computed property and won't result in UI updates when its return value changes. Check the Vue.js docs for more info.

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -71,6 +72,10 @@ export default ({
     PaymentNotReceivedTooltip,
     PaymentRow
   },
+  mixins: [PaymentsMixin],
+  data: {
+    relativeTo: null
+  },
```
taoeffect: Please check the Vue.js documentation for how to set `data` properly. [`data` cannot be an object](https://v2.vuejs.org/v2/guide/components#data-Must-Be-a-Function), it must be a function that returns an object.

taoeffect (changes requested):
Another preliminary review ready!

frontend/model/contracts/group.js
snowteamer: In fact, `initFetchPeriodPayments` calls `periodStampGivenDate` before doing anything else, which in turn calls `groupSortedPeriodKeys`

frontend/model/contracts/shared/time.js
snowteamer: Indeed - This time I wanted to keep the code simple, but actually I don't like creating `Date` objects when not necessary

frontend/model/contracts/shared/time.js
snowteamer: Yea  `?? undefined` is not really necessary, as non-existent array properties always return `undefined` already. It's there just for the sake of being explicit (and maybe for future TypeScript compatibility. I no longer remember, but I wouldn't be surprised if it choke on accessing out-of-bound indices without additional code like this). Initially I wrote `?? undefined` in many other places too, then deleted it (well, except here).

frontend/views/containers/contributions/SupportHistory.vue
snowteamer: Well I ran our test suite - I'm unfortunately no better at catching typos or old names, and need my tools to do it for me.😅

frontend/views/containers/payments/PaymentRowReceived.vue
snowteamer: Hmm, good idea, my Vue.js is visibly a bit rusty...

frontend/model/contracts/group.js
snowteamer: > We could be in in the 3rd period since the group started and this would still evaluate to true.

In a "normal" (not waiting) period, `meta.createdDate < current` should always be false, since the distribution date is when the current period started, not when the next period will start

Otherwise if I'm wrong, couldn't I just use the new `getters.groupDistributionStarted(meta.createdDate)` implemented by Alex?

Although I see it's not used alone somewhere else in this file, but combined with a check for the number of stored period keys:

          if (getters.groupDistributionStarted(meta.createdDate) || Object.keys(getters.groupPeriodPayments).length > 1)) {

(not sure why the getter isn't enough to ensure we're out of the waiting period)

frontend/model/contracts/shared/time.js
snowteamer: You're right, and I would even remove the `date instanceof Date` check if possible, since Flow annotations should already handle that.
Initially I added the `isIsoString` check only because string was are a superset of ISO strings

Edit: After further testing, it appears that `.toISOString()` throws when called on invalid date objects like new Date('foo'). So to avoid possible uncaught errors here we'll have to do a bit more validation or use a `try` block...

What about this:
```js
  // `isNaN` - not `Number.isNaN` - is used here to catch `Invalid Date` objects.
  if (!isIsoString(date) && !(date instanceof Date && !isNaN(date))) {
    throw new TypeError('must be ISO string or valid Date object')
  }
  const timestamp = typeof date === 'string' ? date : date.toISOString()
```
Or maybe we could restrict the API to just not accept Date objects at all?

snowteamer: However Flow complains about `if (!(isIsoString(date) || date instanceof Date)`, and changing the check order didn't do anything to it

frontend/model/contracts/group.js
taoeffect: Ah, I forgot that, that's a good point. 👍 

I investigated this further and surprisingly yes this code actually seems to reproduce the behavior of the code on `master`, which also initializes `state.paymentsByPeriod`  with the periodstamp of the period before the first `distributionDate` — the never-used "waiting period". Which is kind of surprising but yeah this code seems to do the same thing so 👍

frontend/model/contracts/group.js
taoeffect: > In a "normal" (not waiting) period, meta.createdDate < current should always be false, since the distribution date is when the current period started, not when the next period will start

That's probably true.

> couldn't I just use the new getters.groupDistributionStarted(meta.createdDate) implemented by Alex?

The two seem to be equivalent. You could do:

```js
const inWaitingPeriod = !getters.groupDistributionStarted(meta.createdDate)
```

That would probably be best from a DRY perspective.

> but combined with a check for the number of stored period keys:

Yeah I'm not sure what that's about. @Silver-IT couldn't you remove that `|| Object.keys(getters.groupPeriodPayments).length > 1` check? Although that might be there for a reason related to adjusting the distribution date that I don't remember.

Yeah, it could be that the distribution date was adjusted via a proposal to suddenly be in the past?

This is the code: https://github.com/okTurtles/group-income/pull/1643/files#diff-5c6306168381a1d6902fce7da1cdf9a6811e2ed4ac74d7d97bfff5145087b626R122

It seems like the proposal only allows for updating the distribution date to a date in future, starting with 1 day in the future.

However, what if by the time the proposal passes we are already ahead of that date? For example, what today is Sept. 1st, and we set the distribution date to Sept 30th, but then create a proposal to make it Sept. 2nd, but by the time the proposal passes it's already Sept 3rd?

Need to consider this.... 🤕

frontend/model/contracts/shared/time.js
taoeffect: I don't understand why you changed the condition. The one I used should work fine, and it also avoids both checks most of the time because it uses `||`

The flow annotations won't always save us (type info is lost when using SBP)

snowteamer: In this suggestion I used `&&` rather than` ||` to avoid an extra pair or parens, and `isNaN `to guard against `Invalid Date` values. These are quite uncommon but could cause the `toISOString()` call to throw

Otherwise it's equivalent to `if (!(isIsoString(date) || date instanceof Date))`

taoeffect: > used && rather than || to avoid an extra pair or parens

It's better to have the parens than to have an extra logical check imo

> and isNaN to guard against Invalid Date values.

I really don't think that's needed

> Otherwise it's equivalent to if (!(isIsoString(date) || date instanceof Date))

It's not... the `||` condition results in 1 check most of the time, whereas the modified `&&` results in 3 different check every time

frontend/model/contracts/group.js
```diff
@@ -1286,8 +1276,10 @@ sbp('chelonia/defineContract', {
       process ({ meta }, { state, getters }) {
         const period = getters.periodStampGivenDate(meta.createdDate)
         const current = getters.groupSettings?.distributionDate
-
-        if (current !== period) {
+        // Maybe we're updating the distribution date while in the waiting period.
+        if (!getters.groupDistributionStarted(meta.createdDate)) {
+          getters.groupSettings.distributionDate = meta.createdDate
+        } else if (period > current) {
```
taoeffect: I'm pretty sure this change needs to be reverted and the code left unmodified.

This action is only called from a single place in the app: in `state.js` when we enter a new payment period and the distribution date needs to be moved forward.

`state.js`, in turn, never calls this when we're in the waiting period:

```js
          const distributionStarted = store.getters.groupDistributionStarted(reactiveDate.date)
          if (oldPeriod && newPeriod && distributionStarted && (newPeriod !== distributionDateInSettings)) {
            sbp('gi.actions/group/updateDistributionDate', { contractID: store.state.currentGroupId })
          }
```

So this check doesn't make sense to add.

Finally, this sets the `distributionDate` to `meta.createdDate`. That also is incorrect, as the distribution date must be set to a periodstamp, not to any random date.

So I'm 99.9% certain this code needs to be reverted.

frontend/views/containers/contributions/SupportHistory.vue
```diff
@@ -82,7 +66,7 @@ export default ({
     }
   },
   watch: {
-    periods () {
+    currentPaymentPeriod () {
```
taoeffect, line 69: Nice update 👍

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -2,9 +2,9 @@
 .c-summary(data-test='monthOverview')
   i18n.c-summary-title.is-title-4(
     tag='h4'
-    data-test='thisMonth'
-    :args='{ month: humanDate(Date.now(), { month: "long" }) }'
-  ) {month} overview
+    data-test='monthOverviewTitle'
+    :args='{ start: humanDate(startDate), end: humanDate(dueDate) }'
```
taoeffect: Instead of calling `humanDate` here, it's better to call it inside of the computed properties `startDate` and `dueDate` where it's easier to customize it while keeping the markup clean and free of JS as much as possible.

frontend/views/containers/payments/PaymentDetail.vue
```diff
@@ -114,9 +112,11 @@ export default ({
         sbp('okTurtles.events/emit', SET_MODAL_QUERIES, 'PaymentDetail', { id })
       }
       if (payment) {
-        this.payment = cloneDeep(payment)
         // TODO: the payment augmentation duplication in Payment and PaymentRecord, and between todo/sent/received, needs to be resolved more thoroughly
-        this.payment.periodstamp = this.periodStampGivenDate(this.payment.meta.createdDate)
+        this.payment = {
+          ...cloneDeep(payment),
+          periodstamp: await this.historicalPeriodStampGivenDate(payment.meta.createdDate)
+        }
```
taoeffect: If you're going to combine lines 117 and 119 like this, you might as well combine line 120 into here too to be consistent.

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -21,7 +21,7 @@
       .cpr-date.has-text-1 {{ humanDate(payment.date) }}
 
     template(slot='cellRelativeTo')
-      .c-relative-to.has-text-1 {{ humanDate(periodStampGivenDate(payment.date)) }}
+      .c-relative-to.has-text-1 {{ humanDate(payment.relativeTo) }}
```
taoeffect: Why was this modification made? Is it to support `historicalPeriodStampGivenDate`? If so, see comment below

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -111,6 +117,14 @@ export default ({
...
         alert(e.message)
       }
+    },
+    async updatePaymentRelativeTo (paymentDate) {
+      this.relativeTo = await this.historicalPeriodStampGivenDate(paymentDate)
+    }
+  },
+  watch: {
+    payment (to, from) {
+      this.updatePaymentRelativeTo(to.date)
```
taoeffect: Do we really need a watcher for this? I don't think it's necessary because AFAIK a `PaymentRowReceived` never has its `:payment` binding changed after being created.

For such instances it's simpler to initialize the value in the `mounted` method (which gets automatically called). That will also allow us to remove `updatePaymentRelativeTo` and directly calling `historicalPeriodStampGivenDate` in `mounted`.

frontend/views/containers/payments/PaymentRowSent.vue
```diff
@@ -106,6 +114,14 @@ export default ({
...
         alert(e.message)
       }
+    },
+    async updatePaymentRelativeTo (paymentDate) {
+      this.relativeTo = await this.historicalPeriodStampGivenDate(paymentDate)
+    }
+  },
+  watch: {
+    payment (to, from) {
+      this.updatePaymentRelativeTo(to.date)
```
taoeffect: Same comment/concern

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -12,23 +12,70 @@ const PaymentsMixin: Object = {
...
+        periodLength: this.groupSettings.distributionPeriodLength
+      }).previous
+    },
+
+    async historicalPeriodAfterPeriod (periodStamp: string) {
+      return periodStampsForDate(periodStamp, {
+        knownSortedStamps: await this.getAllSortedPeriodKeys(),
+        periodLength: this.groupSettings.distributionPeriodLength
+      }).next
+    },
```
taoeffect, line 64: Beautiful 😍

taoeffect (changes requested):
Alright, now *this* seems to be the final review before I can start testing to verify everything works as it should.

frontend/views/containers/payments/PaymentRowSent.vue
snowteamer: Isn't there any way a PaymentRowSent can be updated, for example in case of partial payment update?

frontend/views/containers/payments/PaymentsMixin.js
snowteamer: I'm quite proud of this code, thank you! 😄

frontend/views/containers/payments/PaymentDetail.vue
snowteamer: Fully agree here (and also above), but sometimes it's not clear to me how much refactoring I'm allowed to do

frontend/views/containers/payments/PaymentRowSent.vue
taoeffect: @snowteamer I don't remember. However, even if that were the case it only updates the total sent, not the period it was sent in I don't think... Feel free to test it out manually with the two approaches and see if there's a difference. I bet there won't be. We can always restore the more complicated method if necessary.

frontend/model/contracts/group.js
snowteamer: Going to try reverting this, I think it was necessary at some point during the PR, but I'd rather get rid of it too

frontend/views/containers/payments/PaymentDetail.vue
snowteamer: OK - by the way, not sure whether `comparePeriodStamps` is necessary, since period stamps are not abstract IDs but always timestamps, which support relational operators out-of-the-box (unlike abstract IDs which may or may not be ordered)

frontend/views/containers/payments/PaymentRowSent.vue
snowteamer: Also is it safe to assign an async handler to `mounted`?

taoeffect: Yes I’m pretty sure you can write `async mounted () {`, there might even be some existing examples in the project (not at my computer at the moment)

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -54,14 +54,23 @@ export default ({
...
       'ourPaymentsSummary',
-      'ourPayments'
+      'ourPayments',
+      'periodStampGivenDate'
     ]),
     currency () {
       return currencies[this.groupSettings.mincomeCurrency].displayWithCurrency
     },
+    humanDueDate () {
+      return this.humanDate(this.dueDateForPeriod(this.currentPaymentPeriod))
```
taoeffect: You didn't write this code (and I'm not going to hold up the PR review over this), but just want to note that it doesn't make sense to turn `humanDate` into a method of this component - unless you want to use it in the markup (and I'm guessing that way back when that's what was being done and why it was added as a method).

frontend/views/containers/payments/PaymentDetail.vue
```diff
@@ -114,10 +112,13 @@ export default ({
         sbp('okTurtles.events/emit', SET_MODAL_QUERIES, 'PaymentDetail', { id })
       }
       if (payment) {
-        this.payment = cloneDeep(payment)
+        const periodstamp = await this.historicalPeriodStampGivenDate(payment.meta.createdDate)
```
taoeffect, line 115: Note: I'm not sure why this was saved to a variable that's never used... in the previous commit it was being inlined on the line below and that's fine. Won't hold up the PR though

frontend/views/containers/payments/PaymentRowReceived.vue
```diff
@@ -112,6 +118,9 @@ export default ({
         alert(e.message)
       }
     }
+  },
+  async mounted () {
+    this.relativeTo = await this.historicalPeriodStampGivenDate(this.payment.date)
```
taoeffect: In the future, please place `mounted` near the top of the component (after `data` and `props`, but before `methods`), as it is some of the first code to be executed

taoeffect (changes requested):
@snowteamer I thought I might be able to merge this PR but actually there are some serious errors still (testing still TBD).

### Flow errors

PRs must never introduce flow errors. The following errors are present:

```
Error -------------------------------------------------------------------- frontend/model/contracts/shared/time.js:25:47

Cannot get `Object.prototype.toString` because property `toString` [1] cannot be unbound from the context [2] where it
was defined. [method-unbinding]

   frontend/model/contracts/shared/time.js:25:47
    25|   if (!(isIsoString(date) || Object.prototype.toString.call(date) === '[object Date]')) {
                                                      ^^^^^^^^ [1]

References:
   /private/tmp/flow/flowlib_e73d01e77da1cc86_501/core.js:236:5
   236|     toString(): string;
            ^^^^^^^^^^^^^^^^^^ [2]
```

If you're certain this code is correct then you can use FlowFixMe comments to disable flow checking on that line (search the project for examples if you need).

### Contract files are incorrect

Running `grunt pin:0.1.8 --overwrite` shows that the generated contracts are in fact different from the changes in this PR, indicating that they're out of date. Whenever you modify contracts that command needs to be run before pushing a commit.

### Boyscoutting

I left some other comments earlier (e.g. see the comment about `mounted` placement, `this.humanDate` instead of `humanDate`), etc. that can now be addressed since these other changes are being requested.

While you're working on these changes I'll also be testing this PR and will get back to you with any other issues I find (if any).

frontend/views/containers/payments/PaymentDetail.vue
snowteamer: It's used to compute `isOldPayment`, so I had to save it to a new variable to avoid duplication

frontend/views/containers/payments/MonthOverview.vue
snowteamer: Makes sense, I should have reverted `humanDate` to be a plain import

taoeffect (changes requested):
So far the tests have gone pretty well!

I did run into one problem though:

Screenshot 2023-09-23 at 10 49 04 AM

To reproduce:

1. Create a $1000 group with u1 and invite u2, u3
2. u1 income details pledging $100, u2 making $800, u3 making $780
3. In console enter:

    ```js
    ```

4. Make 1 full payment, and 1 partial payment
5. Move to the next distribution period by running `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` again
6. As u1 check the completed payments

frontend/views/containers/payments/PaymentDetail.vue
taoeffect: Ah good point! I didn't see that! 😆

taoeffect (changes requested):
Almost perfect!

Don't forget to always run `grunt pin:0.1.8 --overwrite` before pushing changes that contain modifications to contracts code (directly or indirectly through includes).

Screenshot 2023-09-26 at 9 30 21 AM

### EDIT: found one more bug!

This time it appears under the RECEIVED tab for users receiving payments:

Screenshot 2023-09-26 at 9 50 04 AM

taoeffect (changes requested):
Two issues:

1. The contracts are still out of sync with the latest changes. Please run `grunt pin:0.1.8 --overwrite` each time any modifications are made to contracts before pushing a commit
2. The Record Payments modal displays 2 rows for some reason in the 2nd payment period:

Screenshot 2023-10-02 at 6 54 41 PM

To reproduce:

1. Create group with u1 and u2, and add income details for both
2. Run `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` to enter new period
3. u1 pays u2
4. Run `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` to enter 2nd period
5. u1 pays u2 — modal shows 2 rows for some reason

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -29,6 +29,10 @@ const PaymentsMixin: Object = {
       return await this.historicalPeriodStampGivenDate(payment.date)
     },
 
+    async getAllPaymentPeriods () {
+      return { ...await this.getHistoricalPaymentPeriods(), ...this.groupPeriodPayments }
+    },
```
taoeffect, line 34: No need to add this, please use `getAllSortedPeriodKeys` instead

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -69,44 +73,40 @@ const PaymentsMixin: Object = {
...
-        sent: cloneDeep(this.ourPayments?.sent || []),
-        received: cloneDeep(this.ourPayments?.received || []),
-        todo: cloneDeep(this.ourPayments?.todo || [])
-      }
-      const paymentsByPeriod = await this.getHistoricalPaymentPeriods()
+    async getAllPaymentsInTypes () {
+      const sent = []
+      const received = []
+      const todo = cloneDeep(this.ourPayments?.todo ?? [])
+      const paymentPeriods = await this.getAllPaymentPeriods()
```
taoeffect, line 80: Please use `getAllSortedPeriodKeys` instead

snowteamer: Not the same as `getAllSortedPeriodKeys` may also include the distribution date in case `getAllPaymentPeriods` doesn't

taoeffect: Ah! You're right

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -69,44 +73,40 @@ const PaymentsMixin: Object = {
...
-    async getHistoricalPaymentsInTypes () {
-      const paymentsInTypes = {
-        sent: cloneDeep(this.ourPayments?.sent || []),
-        received: cloneDeep(this.ourPayments?.received || []),
-        todo: cloneDeep(this.ourPayments?.todo || [])
-      }
-      const paymentsByPeriod = await this.getHistoricalPaymentPeriods()
+    async getAllPaymentsInTypes () {
+      const sent = []
+      const received = []
```
taoeffect, line 78: Why did this remove `cloneDeep(this.ourPayments?.sent || [])` and `cloneDeep(this.ourPayments?.received || [])`?

snowteamer: To avoid collecting the same entries twice since the current period is going to be processed anyway in the below loop. Another idea would be to keep the `cloneDeep` calls, but then exclude the current period from the loop.

taoeffect: I see..

taoeffect (approved):
WOOHOOOOOO!!!! 😄 🎉 

GREAT JOB @snowteamer !!!!!

This was a doozy of a PR but you got it working!! 😄 🥇 🏆

Issue #1640, opened 2023-06-05
File attachments in chat 


leihla:
### Problem

We need a simple UI for uploading file attachments and displaying them.

### Solution
I've designed two flows for send picture attachment or file attachments. Users should be able to attach a file either by clicking on the attachment icon on the message bar or dragging and dropping an attachment in the chat area.

**Send Image**
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=26191%3A296507&t=teMy1XvN3V9PR9oY-1

<img width="1533" alt="Screenshot 2023-06-05 at 16 12 43" src="https://github.com/okTurtles/group-income/assets/158178/536942c5-0292-4ea2-aa6d-27386d993329">
<img width="1520" alt="Screenshot 2023-06-05 at 16 12 50" src="https://github.com/okTurtles/group-income/assets/158178/250ddcf2-42fe-4238-a5ea-fb752679e5d2">
<img width="744" alt="Screenshot 2023-06-05 at 16 12 57" src="https://github.com/okTurtles/group-income/assets/158178/7e7c6280-ca50-4aa8-bb77-1d4887fbbedd">

**Send Attachment**
https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=26191%3A301188&t=teMy1XvN3V9PR9oY-1

<img width="1525" alt="Screenshot 2023-06-05 at 16 13 31" src="https://github.com/okTurtles/group-income/assets/158178/c566cee7-133c-460a-8122-54f10bdd501a">

Related to #1557

SebinSong:
# Dev TODO List

## Part 1. UI components that don't involve changing `chatroom.js` contracts

 - [x] **Task 1-1. Load and preview image attachment**
 Load image from the file-system and display the preview of the image and make it removable via 'x' button at the corner.
 
<img src='https://github.com/okTurtles/group-income/assets/17641213/f887fd5e-73b3-4fe8-b7d5-08feb3e634f0' width='480'>

<img src='https://github.com/okTurtles/group-income/assets/17641213/72c40596-3273-41eb-9c17-ae1ce74fd482' width='480'>



 - [x] **Task 1-2. Load and preview **non-image type** attachment**
Extend the **Task 1-1** to non-image type attachments.

[...]

PR #1694, opened 2023-08-18
File attachment to chat - 1st chunk

SebinSong:
partial work for #1640 

UI implementation for the file attachment in chat that don't involve changing `chatroom.js`.


frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -390,6 +426,36 @@ export default ({
...
+    openFileAttach (e) {
+      e.target.blur()
+      this.$refs.fileAttachmentInputEl.click()
+    },
+    fileAttachmentHandler (filesList) {
+      const targetFile = filesList[0]
+      const fileUrl = URL.createObjectURL(targetFile)
+      const fileSize = targetFile.size
+
+      if (fileSize > Math.pow(10, 9)) {
```
SebinSong: `Math.pow(10, 9)` here should be replaced with a constant whose name tells the purpose such as `FILE_ATTACHMENT_SIZE_LIMIT`. will make the change along with other upcoming change requests.

frontend/views/containers/chatroom/ChatAttachmentPreview.vue
```diff
@@ -0,0 +1,146 @@
...
+</template>
+
+<script>
+export default {
+  name: 'ChatAttachmentPreview',
+  props: {
+    url: String,
+    name: String,
+    attachType: {
+      // enum of ['image', 'non-image']
```
taoeffect: Again, great comments :+1:

frontend/views/containers/chatroom/ChatAttachmentPreview.vue
```diff
@@ -0,0 +1,146 @@
...
+      type: String,
+      default: 'image'
+    }
+  },
+  computed: {
+    fileExt () {
+      const splitted = this.name.split('.')
+      const ext = splitted[splitted.length - 1] || ''
+      return ext.toUpperCase()
+    }
```
taoeffect: What if the file has no file extension? 🤔

frontend/views/containers/chatroom/ChatFileTooLargeModal.vue
```diff
@@ -0,0 +1,39 @@
+<template lang='pug'>
+modal-template(
+  ref='modal'
+  :a11yTitle='L("Chat attachment too large modal")'
+)
+  template(slot='title')
+    i18n File too large
+
+  .c-content
+    i18n That file is too large and cannot be uploaded. The limit is 1 GB.
```
taoeffect: Could you please make `1 GB` a variable that's retrieved from a computed property?

E.g.

```pug
i18n(:args='{ sizeLimit }') That file is too large and cannot be uploaded. The limit is {sizeLimit}.
```

frontend/views/containers/chatroom/ChatFileTooLargeModal.vue
```diff
@@ -0,0 +1,39 @@
...
+
+  .buttons.is-centered
+    button.is-primary.c-dismiss-btn(
+      type='button'
+      @click='$refs.modal.close()'
+    )
+      i18n Ok
```
taoeffect: I think other OK's in the project are all caps, could you make this one all caps too for consistency?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -726,7 +743,31 @@ export default ({
...
+
+      if (e?.dataTransfer) {
+        // give user a correct feedback about what happens upon 'drop' action. (https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)
+        e.dataTransfer.dropEffect = 'copy'
+      }
+    },
+    dragEndHandler (e) {
+      e.preventDefault()
+      // handler function for 'dragleave', 'dragend', 'drop' events
+      console.log('@@ dragEndHandler !! : ', e.type, e.target)
```
taoeffect: No longer needed logging should be removed

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -104,6 +110,22 @@
...
+          )
+            button.is-icon.c-file-attachment-btn(
+              :aria-label='L("Attach file")'
+              @click='openFileAttach'
+            )
+              i.icon-paper-clip
+              input(
+                ref='fileAttachmentInputEl'
+                type='file'
+                accept='image/*,.pdf'
```
taoeffect: Is this string here what determines what file types we can and cannot upload?

If so, a few requests:

1. Could you bind it to a computed property?
2. Could you make sure users can upload commonly uploaded file types? (e.g. videos, archives (.tar.gz, .tgz, .zip, .bzip2, .7z, etc.), images, pdfs)
3. If a user tries to upload an unsupported file type, display a `gi.ui/prompt` or an `alert` that shows the supported filetypes as a comma-separated list of file extensions

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -374,11 +401,20 @@ export default ({
...
         return false
       }
 
-      this.$emit('send', this.$refs.textarea.value) // TODO remove first / last empty lines
+      let msgToSend = this.$refs.textarea.value || ''
+      if (this.ephemeral.attachment) {
+        // TODO: remove this block and implement file-attachment properly once it's implemented in the back-end.
+        msgToSend = msgToSend +
+          (msgToSend ? '\r\n' : '') +
+          `{ File Attached: ${this.ephemeral.attachment.name} } - Feature coming soon!`
```
taoeffect: Nice 👍

frontend/views/containers/chatroom/SendArea.vue
taoeffect: FYI, this value will come from the server via some server API, as the server dictates the maximum file size people can upload

taoeffect (changes requested):
Nicely done @SebinSong!

Left a few comments. Plus during my testing found an issue when dragging & dropping an image (after 1 file had already been attached via the paper-clip method):

Screen Shot 2023-08-18 at 7 08 30 PM
Screen Shot 2023-08-18 at 7 08 35 PM

It seems like multiple files are not yet supported. But still, it shouldn't error.

frontend/views/containers/chatroom/SendArea.vue
SebinSong: @taoeffect I will leave a comment saying that is a future TODO change once the backend for it is implemented.

frontend/views/containers/chatroom/ChatAttachmentPreview.vue
SebinSong: @taoeffect How about we put the detected `mime-type` there in that case and fall back to putting `UNKNOWN` as the placeholder in the case where the mime-type is not known either?

frontend/views/containers/chatroom/SendArea.vue
SebinSong: @taoeffect 
I took some time for researching on this topic but was not able to find one that seems to be frequently referenced by the dev community (Or, if you know any other useful reference, let me know) so decided to fall back to [this MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types).


Further down there is a table that lists a common file type and the extension. Do you see any file extension that you think must not be supported by our group-chat attachment for a security reason?
I can put together the rest and use the list as the supported types in our app.

frontend/views/containers/chatroom/ChatAttachmentPreview.vue
taoeffect: @SebinSong nah, feel free default to `UNKNOWN` immediately. OTOH, perhaps this should never happen if we force users to use whitelisted extensions per the [other comment](https://github.com/okTurtles/group-income/pull/1694#discussion_r1299000959)

frontend/views/containers/chatroom/SendArea.vue
taoeffect: @SebinSong Hmm. I do not see anything on there that we can't support, except maybe `.php`. However, as far as I know we will be saving these files under the content hash, not with their file extension, so even that should be fine. /cc @corrideat

frontend/views/containers/chatroom/ChatAttachmentPreview.vue
SebinSong: @taoeffect Ok, I can make the warning pop-up show as well when the user upload a file with no extension detected.

taoeffect: That sounds great 👍

frontend/views/containers/chatroom/SendArea.vue
SebinSong: @taoeffect 
This is the warning modal that the app will show when an unsupported file is attached. pls let me know if you want to tweak the message content.

SebinSong:
@taoeffect Updated the PR with additional work for your feedbacks.

frontend/model/contracts/shared/constants.js
```diff
@@ -123,4 +123,15 @@ export const POLL_STATUS = {
...
+  // reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+  '.avif', '.bmp', '.gif', '.ico', '.jpeg', '.jpg', '.png', '.svg', '.tif', '.tiff', '.webp', // image
+  '.aac', '.cda', '.mid', '.midi', '.mp3', '.oga', '.opus', '.wav', '.weba', // audio
+  '.avi', '.mp4', '.mpeg', '.ogv', '.ogx', '.ts', '.webm', '.3gp', '.3g2', // video
+  '.arc', '.bz', '.bz2', '.epub', '.gz', '.jar', '.rar', '.tar', '.zip', '.7z', // archive
+  '.csh', '.css', '.csv', '.doc', '.docx', '.htm', '.html', '.ics', '.js', '.json', '.jsonld', '.mjs',
+  '.odp', '.ods', '.odt', '.pdf', '.php', '.ppt', '.pptx', '.sh', '.txt', '.xhtml', '.xls', '.xlsx', '.xml', // common text/document
+  '.eot', '.otf', '.rtf', '.ttf', '.woff', '.woff2', // font
+  '.abw', '.azw', '.bin', '.csh', '.db', '.mpkg', '.vsd', '.xul' // miscellaneous
+]
```
taoeffect: I'm realizing that we don't have a `constants.js` file that's dedicated to the UI. The closest thing we have is `frontend/utils/events.js`. The problem is that this `shared/constants.js` file is specifically intended for constants that are shared between the UI and the contracts. Usually when this file is modified it forces the contracts to be updated (in this case that didn't happen because no contracts import `CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS`).

So if these constants are never being used in the contracts, and only in the UI, then can you create a file called `frontend/utils/constants.js` and move this variable to there?

taoeffect: Hmm,`.mov` is also missing, can you add that to the list of supported files?

taoeffect (changes requested):
Great work @SebinSong! One minor request and then I think this can be merged.

frontend/model/contracts/shared/constants.js
SebinSong: @taoeffect sure, I can create the `frontend/utils/constants.js` file and move this to there.

SebinSong:
@taoeffect Updated the PR with the work for multiple attachments and also for your change request.

taoeffect (changes requested):
@SebinSong Hmm, multiple file uploads isn't working in my tests...


https://github.com/okTurtles/group-income/assets/138706/687bd305-b998-4c6e-9178-8e2a98a86ec7

SebinSong:
@taoeffect The issue has been fixed. previously it only worked when multiple files are dragged into the chat area **at once**. but now I've made it possible to attach them **one by one** too.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -390,6 +435,64 @@ export default ({
...
+        const fileExt = getFileExtension(file.name)
+        const fileUrl = URL.createObjectURL(file)
+        const fileSize = file.size
+
+        if (fileSize > Math.pow(10, 9)) {
+          // TODO: update Math.pow(10, 9) above with the value delivered from the server once it's implemented there.
+          return sbp('okTurtles.events/emit', OPEN_MODAL, 'ChatFileAttachmentWarningModal', { type: 'large' })
+        } else if (!fileExt || !CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS.includes(fileExt)) {
+          // Give users a warning about unsupported file types
+          return sbp('okTurtles.events/emit', OPEN_MODAL, 'ChatFileAttachmentWarningModal', { type: 'unsupported' })
```
taoeffect, line 467: The file extension needs to be lowercased before it is compared. This way files like `file.JPG` are accepted too.

taoeffect (changes requested):
Excellent work @SebinSong!

I found one last issue and once that's fixed this looks good to merge to me!

taoeffect (approved):
Nicely done @SebinSong! 🫡

PR #1698, opened 2023-08-23
#1693 bug-fix for the blank User-settings modal in certain case

### LINKED ISSUE #1693: Blank page when bringing up user settings after showing payment details
### Problem

When bringing up the user's profile settings (by clicking the wheel icon in the lower left), after having previously brought up the PaymentDetails modal for a payment, you get a blank page:

Screen Shot 2023-08-17 at 11 34 21 AM

The problem is that this is the URL being loaded: http://localhost:3000/app/payments?section=PaymentRowTodo&modal=UserSettingsModal

The `section` isn't being cleared. If you switch away from the Payments page, it is cleared, and then the UserSettings does appear.

### Solution

I *think* the problem is that both the `UserSettingsModal` modal and the payments page use the URL query variable `section` to determine what to show.

Perhaps the solution is to make it so that the `UserSettingsModal` modal automatically switch to `section=my-account` if the current section cannot be found.

SebinSong:
closes #1693

### DIFF
--- a/frontend/views/components/tabs/TabWrapper.vue
+++ b/frontend/views/components/tabs/TabWrapper.vue
@@ -133,14 +133,22 @@ export default ({
           this.activeComponent = link.component
         }
       }
-      this.tabNav.forEach(item => {
-        item.links.forEach(link => {
-          switchTabIfMatch(link)
-        })
+      const allTabNavLinks = this.tabNav.reduce(
+        (allLinks, item) => [ ...allLinks, ...item.links ], []
+      )
+      const fallbackLink = allTabNavLinks.find(item => item.url === 'my-account')
+
+      allTabNavLinks.forEach(item => {
+        switchTabIfMatch(item)
       })
       this.subNav.forEach(navItem => {
         switchTabIfMatch(navItem)
       })
+
+      if (!this.activeComponent) {
+        // if still no matching link is found, fallback to 'my-account' tab.
+        this.tabClick(fallbackLink)
+      }
     }
   },
   beforeDestroy () {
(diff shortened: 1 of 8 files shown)


frontend/views/components/tabs/TabWrapper.vue
```diff
@@ -133,14 +133,22 @@ export default ({
...
         }
       }
-      this.tabNav.forEach(item => {
-        item.links.forEach(link => {
-          switchTabIfMatch(link)
-        })
+      const allTabNavLinks = this.tabNav.reduce(
+        (allLinks, item) => [...allLinks, ...item.links], []
+      )
+      const fallbackLink = allTabNavLinks.find(item => item.url === 'my-account')
```
SebinSong, line 141: `fallbackLink` here is the case where the specified `tab` query doesn't match any tab items in the last. (e.g. `?modal=UserSettings&tab=asdfsadf`) we need to manually direct it to the `my-account` tab.

SebinSong:
@taoeffect 
As you pointed out, the cause of the issue is that two different vue components(`UserSettingsModal.vue` and `Payments.vue`) use the same query param `section`.
Manually resetting the value to `my-account` in `UserSettingsModal` did work but then I spotted another issue where the value of `section` in the url instantly gets reverted back to the value used by `Payments.vue` even when User-settings modal is working completely fine.
Turns out that's because we are opening the modal on top of `Payment.vue` page in this particular case which has a watcher for `$route` in it that updates the `section` query every time it detects a change.
So I decided to go for the hard way which is renaming `section` query to something else(`tab`) in UserSettings modal.

frontend/views/components/tabs/TabWrapper.vue
taoeffect: Good comment,  can you please add it as an in-line comment to the code? (In the future, all such comments are best added directly to the code instead of Github, so that they'll be remembered.)

taoeffect (changes requested):
Looking & working really great @SebinSong! Just one change request.

taoeffect (approved):
Great work @SebinSong!

PR #1703, opened 2023-08-29
Fix manifest not found error upon contract file change

### LINKED ISSUE #1692: Manifest not found error upon contract file change in development
### Problem

A 404 Not Found error related to chelonia/private/loadManifest sometimes occurs after saving a contract file change in development, and the page is either blank or an alert saying "contract sync error" pops up.

#### Steps to reproduce:
- Run `grunt dev` without persistence enabled, i.e. `GI_PERSIST` should not be set
- Open the app in the browser
- Create an user account
- Save a change to e.g. `group.js`. The change must result in a new version of the file, and must not match any previous version of it, so that a new manifest is generated.
- When the app reloads, the page stays blank and the following error shows up:
image

Manually reloading a second time doesn't help, only rebuilding the app does.

### Solution

 `manifest.json` is automatically updated on contract file change, but maybe contracts are not re-deployed accordingly.

snowteamer:
Closes #1692

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,7 @@
 'use strict'
 
+if (process.env.CI) process.exit(1)
+
 // =======================
 // Entry point.
 //
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,7 +1,5 @@
 'use strict'
 
-if (process.env.CI) process.exit(1)
-
 // =======================
 // Entry point.
 //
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 
+if (process.env.CI) process.exit(1)
+
```
taoeffect: If this isn't a WIP this can be removed?

Gruntfile.js
```diff
@@ -136,11 +138,38 @@ module.exports = (grunt) => {
...
   }
 
+  async function redeployAndUpdateMainSrc (manifestDir) {
+    if (development) {
+      const { API_URL, GI_PERSIST } = process.env
+      const { dirname, filename } = databaseOptionBags[GI_PERSIST] ?? {}
+      let dest = './data'
+      if (GI_PERSIST === 'sqlite') dest = path.join(dirname, filename)
+      if (GI_PERSIST === 'fs') dest = dirname
+      if (!GI_PERSIST) dest = API_URL
```
taoeffect: AFAICT, this code can be simplified significantly.

1. We can start by setting `let dest = API_URL`. This eliminates the need for line 148
2. Next, since the variables `dirname` and `filename` are never used for anything except to set `dest`, we can further simplify by directly retrieving the full path from `databaseOptionBags` and having it there. This eliminates the need for lines 146 and 147

Gruntfile.js
```diff
@@ -136,11 +138,38 @@ module.exports = (grunt) => {
...
+      grunt.log.writeln(chalk.underline(`Running 'chel deploy to ${dest}'`))
+      const { stdout } = await execWithErrMsg(`./node_modules/.bin/chel deploy ${dest} ${manifestDir}/*.manifest.json`, 'error redeploying contracts')
+      console.log(stdout)
+      const r = /contracts\/([^.]+)\.(?:x|[\d.]+)\.manifest.*\/(.*)/g
+      const manifests = Object.fromEntries(Array.from(stdout.replace(/\\/g, '/').matchAll(r), x => [`gi.contracts/${x[1]}`, x[2]]))
+      await writeFile(manifestJSON, JSON.stringify({ manifests }, null, 2) + '\n')
```
taoeffect: This entire section appears to be repeated? If so then can you please place it in a function to DRY it with `deployAndUpdateMainSrc`? Or, alternatively (perhaps even better), merge the two functions into 1 function and pass in a parameter to distinguish behavior.

taoeffect (changes requested):
Review ready!

Gruntfile.js
snowteamer: Yea, that's the code duplication I was talking about on Slack

taoeffect (approved):
Beautifully done @snowteamer! 👏

PR #1708, opened 2023-09-02
Rethrow uncaught Vue errors in dev mode or CI

### LINKED ISSUE #684: Make Cypress tests fail on any console error
### Problem

During PR reviews there are often console errors that go unnoticed in the Travis build, because they are Vue.js related errors that are somewhat minor and don't affect the overall outcome.

It would be nice if these could be noticed automatically as part of `grunt test` and cause Travis to fail.

### Solution

- Handle all uncaught exceptions using [`Vue.config.errorHandler`](https://vuejs.org/v2/api/#errorHandler) or [`window.onerror`](https://blog.sentry.io/2016/01/04/client-javascript-reporting-window-onerror), whichever one is better (justify your decision)
- Make sure Travis fails if there's ever a Promise that is rejected without being properly handled
- Make sure Travis fails if there's ever an uncaught exception thrown

snowteamer:
Closes #684

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -47,7 +47,8 @@ console.info('NODE_ENV:', process.env.NODE_ENV)
 
 Vue.config.errorHandler = function (err, vm, info) {
   console.error(`uncaught Vue error in ${info}:`, err)
-  if (process.env.NODE_ENV === 'development' || process.env.CI) throw err
+  // Fix for https://github.com/okTurtles/group-income/issues/684
+  if (process.env.CI) throw err
 }
 
 async function startApp () {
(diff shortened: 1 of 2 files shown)


frontend/main.js
```diff
@@ -47,6 +47,7 @@ console.info('NODE_ENV:', process.env.NODE_ENV)
 
 Vue.config.errorHandler = function (err, vm, info) {
   console.error(`uncaught Vue error in ${info}:`, err)
+  if (process.env.NODE_ENV === 'development' || process.env.CI) throw err
```
taoeffect: In development mode I don't think we need to rethrow as it will not accomplish anything but will cause duplicate messages in the console.

Also, could you add a comment above this line linking to the issue or this PR?

taoeffect (changes requested):
Cool, thanks for the PR @snowteamer! One request

taoeffect (approved):

PR #1709, opened 2023-09-04
Merge PrivateDM and GroupDM

### LINKED ISSUE #1647: Handle message related errors and display them to the user

### LINKED ISSUE #1696: DMMixins.js does not handle errors
### Problem

Almost every function inside of `DMMixin.js` doesn't handle errors.

Example:

```js
    createPrivateDM (username: string) {
      // NOTE: username should be valid
      sbp('gi.actions/identity/createDirectMessage', {
        contractID: this.ourIdentityContractId,
        data: {
          privacyLevel: CHATROOM_PRIVACY_LEVEL.PRIVATE,
          usernames: [username]
        }
      })
    },
    createGroupDM (usernames: string[]) {
      // NOTE: usernames.length should be gte 2
      sbp('gi.actions/identity/createDirectMessage', {
        contractID: this.ourIdentityContractId,
        data: {
          privacyLevel: CHATROOM_PRIVACY_LEVEL.GROUP,
          usernames
        }
      })
    },
```

Neither of these functions handle errors that might happen when the action is sent.

Also, why are there two different types of DMs? /cc @Silver-IT 
I think all DMs need to be the same type and have the same `privacyLevel`: `CHATROOM_PRIVACY_LEVEL.PRIVATE`.

### Solution

Every single action that is called, always, must handle any errors.

Every user-facing error must be shown to users, either with an `alert()` or a `gi.ui/prompt`.

### LINKED ISSUE #1697: Remove "Private DMs", "Group DMs" everywhere and use "DirectMessages" instead
### Problem

The difference in the code between "Private DMs" and "Group DMs" is causing many problems both conceptually and otherwise with the `e2e-protocol` branch.

There should be no difference between these two. "Private DMs" **are** Group DMs with a group of 2.

### Solution

1. Remove all code related to "Private DMs" and replace with "Group DMs" only
2. All DMs must have a `privacyLevel` of `PRIVATE`, not `GROUP`.
3. The `ChatMembers.vue` component is named `GroupMembers`. This is wrong and causing confusion because there is already a component called `GroupMembers`. All components **must** have the same name as the file name. Fix this by renaming the component to `ChatMembers`.
4. Fix all UI issues that result.

### LINKED ISSUE #1700: Verify / ensure that this.latestEvents in ChatMain *always* has event from the correct chatroomId
### Problem

While debugging problems with `ChatMain.vue`, @corrideat and I discovered that right now it's not 100% clear that `this.latestEvents.push(..)` always pushes an event from the right chatroom.

`this.latestEvents.push(..)` is called in two places:

1. `listenChatRoomActions` - which is called (in his branch) from `EVENT_HANDLED`. There is a check to ensure that `contractID === this.summary.chatRoomId`, however, it isn't clear that this necessarily means that `this.latestEvents` has the events for `contractID`
2. `renderMoreMessages` - uses a mixture of `this.messageState.prevTo` and `this.currentChatRoomId`, but again it's not clear that `this.latestEvents` corresponds to those. Are we even 100% sure that `this.messageState.prevTo` and `this.currentChatRoomId` correspond to each other?

### Solution

1. Investigate all of these questions.
2. Either write a proof below that `this.latestEvents` always corresponds to the correct chatroom, or modify the code so that it's guaranteed to be the case.

Please note: if you are making any code changes, those code changes **must** happen on a branch off of [`e2e-protocol-ricardo`](https://github.com/okTurtles/group-income/tree/e2e-protocol-ricardo) — **NOT** `master`!

Silver-IT:
### Summary of Changes

- Closes #1697 
- Closes #1700 
- Closes #1647 
- Closes #1696


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -155,6 +155,7 @@ export default ({
         // NOTE: messagesInitiated describes if the messages are fully re-rendered
         //       according to this, we could display loading/skeleton component
         messagesInitiated: undefined,
+        renderingChatRoomId: null,
```
Silver-IT, line 176: Since the `this.currentChatRoomId` is a getter which can be changed anytime.
We can not use `this.currentChatRoomId` to point the current-rendering-chatRoomId because it takes some time to render the chatroom.
Since we initiate the chatroom state when we open or switch a chatroom, we can say the current-rendering-chatroom is the chatroom whose state is initiated for the last time.

frontend/model/state.js
```diff
@@ -524,78 +524,59 @@ const getters = {
...
+      ) {
+        const users = Object.keys(state[chatRoomId].users)
+        const partners = users
+          .filter(username => username !== getters.ourUsername)
+          .sort((p1, p2) => {
+            const p1JoinedDate = new Date(state[chatRoomId].users[p1].joinedDate).getTime()
+            const p2JoinedDate = new Date(state[chatRoomId].users[p2].joinedDate).getTime()
+            return p1JoinedDate - p2JoinedDate
+          })
+        const lastJoinedPartner = partners[partners.length - 1]
```
Silver-IT, line 560: `lastJoinedParter` is chatroom member who has joined the chatroom for the last time.
His profile picture can be used as the picture of the direct message with the badge of the number of partners.

frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -57,7 +57,7 @@ export default ({
       const displayName = this.userDisplayName(username)
 
       const notificationTemplates = {
-        onGroupDM: {
+        onDirectMessage: {
```
Silver-IT, line 63: This is not using at the moment since we don't make notification messages inside the direct messages.
Even though it's not needed now, I didn't remove this because we could use this in the future.

Silver-IT (commented):
@taoeffect, @corrideat. Seems ready for your review.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: Nice comment, but please add these comments in-line to the source instead of on Github so that they don't get lost

frontend/model/state.js
taoeffect: Can you add these and any other GH comments to the code?

shared/domains/chelonia/chelonia.js
```diff
@@ -177,7 +177,7 @@ const rawSignatureFnBuilder = (key) => {
   }
 }
 
-const signatureFnBuilder = function (this: any, signingContractID, signingKeyId) {
+const signatureFnBuilder = function (signingContractID: string, signingKeyId: string) {
```
taoeffect: This file must not be modified in this PR, so when re-creating the PR please make sure no files under `shared/domains/chelonia` are modified

taoeffect (changes requested):
Nicely done @Silver-IT. This is a preliminary review (I'm not finished reviewing this one yet).

Can you close and reopen this PR so that it's merging into `e2e-protocol` instead of `e2e-protocol-ricardo`?

In the future, please branch off of `e2e-protocol` instead of `e2e-protocol-ricardo` (if I said otherwise earlier, sorry, that was my mistake!). This will help us avoid conflicts with the current changes @corrideat is working on.

Also, please double check and make sure that none of the files under `shared/domains/chelonia` appear as being modified in the PR. If they are, you might have to re-create a branch off of `e2e-protocol` and re-apply your changes.

Silver-IT (commented):
@taoeffect, @corrideat.
This PR seems ready for your reviews.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "group-income",
   "version": "0.0.1",
-  "contractsVersion": "0.1.6",
+  "contractsVersion": "0.1.8",
```
taoeffect: Please set this to `0.2.0` per the reasons described here: https://github.com/okTurtles/group-income/pull/1722#discussion_r1323344309

frontend/controller/actions/identity.js
```diff
@@ -498,13 +499,16 @@ export default (sbp('sbp/selectors/register', {
     const rootState = sbp('state/vuex/state')
     const rootGetters = sbp('state/vuex/getters')
     const partnerProfiles = params.data.usernames.map(username => rootGetters.ourContactProfiles[username])
+    // NOTE: 'rootState.currentGroupId' could be changed while waiting for the sbp functions to be proceeded
+    //       So should save it as a constant variable 'currentGroupId', and use it which can't be changed
+    const currentGroupId = rootState.currentGroupId
```
taoeffect, line 504: Excellent catch @Silver-IT and **great** comment!! 😄 👍 🥇

frontend/model/contracts/chatroom.js
```diff
@@ -191,16 +186,14 @@ sbp('chelonia/defineContract', {
...
+          throw new Error(`Can not join the chatroom which ${username} is already part of`)
         }
 
         Vue.set(state.users, username, { joinedDate: meta.createdDate })
 
         const { type, privacyLevel } = state.attributes
-        const isPrivateDM = type === CHATROOM_TYPES.INDIVIDUAL && privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE
-        if (!state.onlyRenderMessage || isPrivateDM) {
+        const isDirectMessage = type === CHATROOM_TYPES.INDIVIDUAL && privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE
+        if (!state.onlyRenderMessage || isDirectMessage) {
```
taoeffect: Something doesn't make sense conceptually here.

1. It is not possible for DMs to have `CHATROOM_TYPES.GROUP`.
2. `CHATROOM_TYPES.INDIVIDUAL` is unclear, it should be renamed to `CHATROOM_TYPES.DIRECT_MESSAGE`
3. All `CHATROOM_TYPES.DIRECT_MESSAGE` have `CHATROOM_PRIVACY_LEVEL.PRIVATE`. It doesn't need to even be checked, because there is no such thing as a "group" DM or a "public" DM.
4. `CHATROOM_PRIVACY_LEVEL` is only relevant for `CHATROOM_TYPES.GROUP`

Can you update the code to reflect this?

You don't need to worry about backwards compatibility because we're going to be recreating all groups anyway.

frontend/model/contracts/chatroom.js
```diff
@@ -219,12 +212,13 @@ sbp('chelonia/defineContract', {
 
         emitMessageEvent({ contractID, hash })
         setReadUntilWhileJoining({ contractID, hash, createdDate: meta.createdDate })
+
         if (username === loggedIn.username) {
           const { type, privacyLevel } = state.attributes
-          const isPrivateDM = type === CHATROOM_TYPES.INDIVIDUAL && privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE
-          if (isPrivateDM) {
+          const isDirectMessage = type === CHATROOM_TYPES.INDIVIDUAL && privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE
```
taoeffect: Same here, it's a direct message if `type === CHATROOM_TYPES.DIRECT_MESSAGE`, no need to check the privacy level

frontend/model/contracts/identity.js
```diff
@@ -121,14 +123,14 @@ sbp('chelonia/defineContract', {
     'gi.contracts/identity/createDirectMessage': {
       validate: (data, { state, getters }) => {
         objectOf({
-          privacyLevel: unionOf(...Object.values(CHATROOM_PRIVACY_LEVEL).map(v => literalOf(v))),
+          groupContractID: string,
```
taoeffect: Can you add comments explaining the difference between `groupContractID` and `contractID`?

Also, shouldn't `groupContractID` be optional? DMs can be created without referencing a group.

Same goes for the lines below in `joinDirectMessage`

frontend/model/state.js
```diff
@@ -524,78 +524,62 @@ const getters = {
...
     }
-    return privateDMs
+    return currentGroupDirectMessages
   },
-  ourGroupDirectMessages (state, getters) {
-    const groupDMs = {}
-    for (const cID of Object.keys(getters.ourDirectMessages)) {
-      if (getters.ourDirectMessages[cID].privacyLevel === CHATROOM_PRIVACY_LEVEL.GROUP && state[cID]) {
-        groupDMs[cID] = getters.ourDirectMessages[cID]
+  ourGroupDirectMessageFromUsernames (state, getters) {
```
taoeffect, line 576: Can you add a comment above this explaining what this getter is for?

taoeffect (changes requested):
@Silver-IT preliminary review ready.

@corrideat This needs your review too!

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -422,10 +428,13 @@ export default ({
       }
     },
     async renderMoreMessages (shouldInitiate = true) {
+      // NOTE: 'this.renderingChatRoomId' can be changed while running this function
```
corrideat, line 463: @taoeffect @Silver-IT: This function looks like a good use case for an (async) generator (`async function * renderMoreMessages {}`). If not a generator, can you explain what this does and maybe also add a return type annotation? It's not immediately clear that it returns `undefined | boolean`, and even less clear what those mean or why?

frontend/model/contracts/identity.js
```diff
@@ -121,15 +123,22 @@ sbp('chelonia/defineContract', {
     'gi.contracts/identity/createDirectMessage': {
       validate: (data, { state, getters }) => {
         objectOf({
-          privacyLevel: unionOf(...Object.values(CHATROOM_PRIVACY_LEVEL).map(v => literalOf(v))),
-          contractID: string
+          // NOTE: 'groupContractID' is the contract ID of the group where the direct messages is created
+          //       it's optional parameter meaning the direct message could be created outside of the group
+          groupContractID: optional(string),
```
corrideat, line 125: I'm not sure about this `groupContractID`. It mostly seems unnecessary based on how I understand DMs to work, which isn't tied to a group.

frontend/model/contracts/identity.js
```diff
@@ -121,15 +123,22 @@ sbp('chelonia/defineContract', {
     'gi.contracts/identity/createDirectMessage': {
       validate: (data, { state, getters }) => {
         objectOf({
-          privacyLevel: unionOf(...Object.values(CHATROOM_PRIVACY_LEVEL).map(v => literalOf(v))),
```
corrideat, line 124: AFAIU, we could have 'public' DMs (?)

frontend/model/state.js
```diff
@@ -535,78 +534,65 @@ const getters = {
...
-    return chatRoomId => {
-      const contractID = chatRoomId || getters.currentChatRoomId
-      return getters.ourDirectMessages[contractID]?.privacyLevel === CHATROOM_PRIVACY_LEVEL.GROUP
-    }
-  },
-  isPrivateChatRoom (state, getters) {
-    return (chatRoomId: string) => {
-      const contractID = chatRoomId || getters.currentChatRoomId
-      return state[contractID]?.attributes?.privacyLevel === CHATROOM_PRIVACY_LEVEL.PRIVATE
+      return getters.isJoinedChatRoom(contractID) && !!getters.ourGroupDirectMessages[contractID]
```
corrideat: Can you explain (in a comment) the meaning of this? It's not immediately clear why either of these checks are needed or how they relate to DMs

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -624,11 +633,14 @@ export default ({
...
       if (this.ephemeral.messagesInitiated === undefined) {
         // NOTE: this infinite handler is being called once which should be ignored
-        // before calling the setInitMessages function
+        //       before calling the setInitMessages function
+        return
+      } else if (this.currentChatRoomId !== this.renderingChatRoomId) {
+        // NOTE: should render messages after the current chatroom state is initiated
         return
       }
       this.renderMoreMessages(!this.ephemeral.messagesInitiated).then(completed => {
```
corrideat, line 685: See comment above about `this.renderMoreMessages`. Also, this most probably needs a `.catch` part2

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: >  This function looks like a good use case for an (async) generator (async function * renderMoreMessages {})

I do not like generators at all (because they are confusing to reason about and also not available in most other languages), and I'm also not seeing how or why that would be helpful here.

This, however, I completely agree with:

> If not a generator, can you explain what this does and maybe also add a return type annotation? It's not immediately clear that it returns `undefined | boolean`, and even less clear what those mean or why?

Indeed, @Silver-IT it's not clear why this function is returning a boolean sometimes. The return value never seems to be used. So could you please simplify this function and remove everything that's not needed?

frontend/model/contracts/identity.js
taoeffect: @corrideat so this was added as a hint to help the UI keep track of DMs that were created in the context of a group. It's not really used for anything else except for that (helping visually organize DMs).

frontend/model/contracts/identity.js
taoeffect: There is no such thing as a public DM.

frontend/model/state.js
taoeffect: @corrideat This getter, `isDirectMessage`, is used in a variety of places to see if the chatroom is a direct message. He needs some boolean logic here to answer that question, right?

Although, I do agree that the first check (`getters.isJoinedChatRoom(contractID)`) might not be necessary because I *think* that check is already handled by `!!getters.ourGroupDirectMessages[contractID]`:

```js
      // NOTE: skip DMs whose chatroom contracts are not synced yet
      if (!state[chatRoomId] || !state[chatRoomId].users?.[getters.ourUsername]) {
        continue
      }
```

What do you think @Silver-IT?

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: Agreed, if there's an exception it should be logged with `console.error` in a `.catch`

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -397,7 +396,7 @@ export default ({
     },
     startMention (keyword, position) {
       const all = makeMentionFromUsername('').all.slice(1)
-      const availableMentions = this.isPrivateDirectMessage()
+      const availableMentions = this.users.length === 2
```
taoeffect: What is this `this.users.length === 2` check doing?

Should it be `>` ?

If not, can you add a comment explaining what's going on here?

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Yeah, the return value can be `boolean` or `undefined`, and they are used to change the state inside `'infiniteHandler'`.
What the return types mean is that
**true**: Loaded all the messages, and no more messages exists
**false**: Messages are loaded, but more messages exist
**undefined**: Loaded the wrong events from the server so those events should be ignoreed and no state changes are needed (like `messagesInitiated`). Since the `renderingChatRoomId` could be changed while processing this function, this function could do the redundant process. This normally happens when user switches channels very fast.

Should I create a new type instead of using `'boolean' or 'undefined'`?

frontend/model/state.js
Silver-IT: @taoeffect, that's correct. I will simplify this getter as you suggested.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: > Should I create a new type instead of using 'boolean' or 'undefined'?

No, can you copy/paste that explanation and put it into a comment right above the function definition for `renderMoreMessages`?

E.g.

```js
// This function returns a value that is used by `infiniteHandler` ... [continue explanation]
// [continued explanation of return values]
async renderMoreMessages (shouldInitiate = true) {
```

frontend/views/containers/chatroom/SendArea.vue
Silver-IT: Nice catch. It should be `<=2` here.
For better understanding, I will update it.
```
let availableMentions = this.users
// NOTE: '@all' mention should only be needed when the members are more than 3
if (this.users.length > 2) {
  availableMentions = [
    ...availableMentions,
    {
      username: all,
      displayName: all,
      picture: '/assets/images/horn.png'
    }
  ]
}
```

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Okay.

frontend/model/contracts/identity.js
```diff
@@ -62,7 +62,9 @@ sbp('chelonia/defineContract', {
       process ({ data }, { state }) {
         const initialState = merge({
           settings: {},
-          attributes: {},
+          attributes: {
+            allowDMInvite: true
+          },
```
taoeffect: What is this attribute? There is no comment...

If it's not needed it can & should be removed.

frontend/model/contracts/identity.js
```diff
@@ -121,15 +123,22 @@ sbp('chelonia/defineContract', {
...
-        })
+        const { groupContractID, contractID } = data
+        if (groupContractID) {
+          Vue.set(state.chatRooms, contractID, {
+            groupContractID,
+            visible: true // NOTE: this attr is used to hide/show direct message
+          })
+        } else {
+          // TODO: create a direct message outside of the group
+        }
```
taoeffect: Why are we splitting this out into an if/else?

Is there any issue with setting `state.chatRooms` in both cases? If not, let's do that. We can always modify it later if necessary. If there is a problem, then it should be explained in a comment so that we remember to address it in the future.

frontend/model/contracts/identity.js
```diff
@@ -169,18 +181,18 @@ sbp('chelonia/defineContract', {
...
       },
       process ({ data }, { state, getters }) {
-        Vue.set(state.chatRooms[data.contractID], 'hidden', data.hidden)
+        Vue.set(state.chatRooms[data.contractID], 'visible', data.visible)
       },
       sideEffect ({ data }) {
-        const { contractID, hidden } = data
-        hidden ? leaveChatRoom({ contractID }) : sbp('chelonia/contract/sync', contractID)
+        const { contractID, visible } = data
+        visible ? sbp('chelonia/contract/sync', contractID) : leaveChatRoom({ contractID })
```
taoeffect: Why are we leaving the chatroom if its visibility is set to hidden? That doesn't make sense... it should only affect the visibility in the UI, right?

(Do we even use the `visible` attribute in the app?)

frontend/model/state.js
```diff
@@ -535,78 +534,62 @@ const getters = {
...
+    const currentGroupDirectMessages = {}
+    for (const chatRoomId of Object.keys(getters.ourDirectMessages)) {
+      // NOTE: skip DMs whose chatroom contracts are not synced yet
+      if (!state[chatRoomId] || !state[chatRoomId].users?.[getters.ourUsername]) {
+        continue
+      }
+      // NOTE: get only visible DMs for the current group
+      if (getters.ourDirectMessages[chatRoomId].groupContractID === state.currentGroupId &&
+        getters.ourDirectMessages[chatRoomId].visible &&
+        state[chatRoomId]
```
taoeffect: You don't need to check `state[chatRoomId]`, that check was already done on line 541 above

frontend/model/state.js
```diff
@@ -535,78 +534,62 @@ const getters = {
...
-          contractID
+  ourGroupDirectMessages (state, getters) {
+    const currentGroupDirectMessages = {}
+    for (const chatRoomId of Object.keys(getters.ourDirectMessages)) {
+      // NOTE: skip DMs whose chatroom contracts are not synced yet
+      if (!state[chatRoomId] || !state[chatRoomId].users?.[getters.ourUsername]) {
+        continue
+      }
+      // NOTE: get only visible DMs for the current group
+      if (getters.ourDirectMessages[chatRoomId].groupContractID === state.currentGroupId &&
```
taoeffect: In this functionl both `state[chatRoomId]` and `getters.ourDirectMessages[chatRoomId]` are dereferenced a lot.

Whenever that sort of repetition happens, it's best to cache that value in a variable for both performance reasons and to DRY the code so that it's more readable.

So at the top of this loop I would do:

```js
for (const chatRoomId of Object.keys(getters.ourDirectMessages)) {
  const chatroomState = state[chatRoomId]
  const ourDMs = getters.ourDirectMessages[chatRoomId]
  // NOTE: skip DMs whose chatroom contracts are not synced yet
```

frontend/model/state.js
```diff
@@ -535,78 +534,62 @@ const getters = {
...
+  // NOTE: this getter is used to find the ID of the direct message in the current group
+  //       with the name[s] of partner[s]. Normally it's more useful to find direct message
+  //       by the partners instead of contractID
+  ourGroupDirectMessageFromUsernames (state, getters) {
+    return (partners) => { // NOTE: string | string[]
+      if (typeof partners === 'string') {
+        partners = [partners]
       }
+      return Object.keys(getters.ourGroupDirectMessages).find(chatRoomId => {
+        const cPartners = getters.ourGroupDirectMessages[chatRoomId].partners
```
taoeffect: Please cache `getters.ourGroupDirectMessages` at the top of this function to avoid calling it multiple times in the loop and to DRY things

taoeffect (changes requested):
Nicely done @Silver-IT! Looks like it's almost ready to merge (I think! Still need to test after you respond to this review)

frontend/model/contracts/identity.js
Silver-IT: > Why are we leaving the chatroom if its visibility is set to hidden?

As far as I checked the DM in Slack, I remember we don't get any sounds or notifications after we click that `X` button. And I wanted to free users from syncing the chatroom contract if they don't think they need the chatroom anymore.

I was going to implement the following feature which is in Slack.
image
Since we don't have design for this feature yet, I will remove all stuff related to this attribute visibility. Is that fine?

frontend/model/contracts/identity.js
Silver-IT: Since users can create DM and make someone else join the DM, I think all users have risk that others can control their workload.
We don't have any design for this, so this should be removed at the moment?
IMHO, it is necessary but if you want to remove it, I will.

frontend/model/contracts/identity.js
Silver-IT: To be honest, I have never thought before how to save the DMs from outside of any groups.
Just having thought about it, it could be fine to save the two types of DMs in `state.chatRooms`.
If the `groupContractID` is undefined, it means the DMs outside of the group. I will remove `if/else` statement.

frontend/model/contracts/identity.js
taoeffect: You don’t need to remove everything, just all of the logic related to it. The visibility attribute should not affect any behaviors like leaving or joining chatrooms. So logic related to that should be removed. The only thing in the contracts related to the visibility boolean attribute is turning it on and turning it off. That’s it. Nothing else. If we need it in the future, the UI will handle the rest. It must not affect contract logic.

Silver-IT: And also the notifications and the sounds should work only when the `visibile` attribute is `true`. Right?

taoeffect: No. They should always work. Remove all logic related to `visible`. Only leave logic to turn it on and off and remove everything else. We have not designed anything related to it, so please don’t make any assumptions with the logic.

Silver-IT: Got it.

frontend/model/contracts/identity.js
taoeffect: This is something that needs to be worked through in a separate issue/PR together with @corrideat, so while that’s a great observation, yes please remove it for now and open an issue so that we can tackle it later using cryptography

frontend/controller/actions/identity.js
```diff
@@ -498,14 +499,17 @@ export default (sbp('sbp/selectors/register', {
...
```
taoeffect, line 504: NOTE: right now it doesn't matter, but in the future when we allow the creation of DMs outside of a group it will - this value should be retrieved from `params.groupId`.

taoeffect (approved):
Approved! Great work @Silver-IT! This was a hard one but the code is now greatly simplified thanks to the changes you made! Great job!! 😄 👍

frontend/controller/actions/identity.js
Silver-IT: Noted. And I will add this comment in the PR #1730.

PR #1712, opened 2023-09-05
Fix Chat UI bugs

### LINKED ISSUE #1705: 2 Chat view UI bugs
### Problem

The video below shows two bugs:

1. When I scroll in the grey area just to the right of the chat UI (but **NOT** hovering over the right sidebar that shows the members), the chat UI scrolls up and down. It should not scroll up and down.
2. When I make the window height shorter by dragging from the bottom upward (while being scrolled all the way to the bottom), the chat UI hides the chats at the bottom. Instead, those bottom chats should stay fixed and visible as I pull upward (just like Slack and Discord behave).

https://github.com/okTurtles/group-income/assets/138706/235abfa4-7c5b-4dd9-9d71-22217955182f

### Solution

1. Fix it so that the chat UI doesn't scroll up and down in this situation
2. Whenever the chat view is scrolled all the way to the bottom, the behavior when the user makes the window smaller needs to be different than if they're scrolled up a bit. When the user is viewing the latest messages and making the window smaller, the latest messages must not be hidden. Try this out in Slack and Discord and see how they behave. They make sure that all of the messages get pushed upward, ensuring that the latest messages are not hidden. Do the same thing.

Silver-IT:
### Summary of Changes

- Closes #1705

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -613,7 +613,17 @@ export default ({
     resizeEventHandler () {
       const vh = window.innerHeight * 0.01
       document.documentElement.style.setProperty('--vh', `${vh}px`)
+
+      if (this.ephemeral.scrolledDistance < 40) {
+        // NOTE: 40px is the minimum height of a message
+        //       even though user scrolled up, if he scrolled less than 40px (one message)
+        //       should ignore the scroll position, and scroll to the bottom
+        this.debouncedJumpToLatest()
+      }
     },
+    debouncedJumpToLatest: debounce(function () {
+      this.updateScroll()
+    }, 300),
     infiniteHandler ($state) {
       this.ephemeral.infiniteLoading = $state
       if (this.ephemeral.messagesInitiated === undefined) {
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -627,8 +627,9 @@ export default ({
       }
     },
     throttledJumpToLatest: throttle(function (_this) {
+      // NOTE: 40ms makes the container scroll the 25 times a second which feels like animated
       _this.jumpToLatest('instant')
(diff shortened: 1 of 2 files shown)


taoeffect (changes requested):
The way this is implemented results in a delayed scroll down that's kind of weird.

In both Slack and Discord the messages "stick" to the bottom of the window constantly without any sort of delay. Can you reproduce that behavior?

SebinSong:
@taoeffect ,
As far as I know, unfortunately there is currently no css rule that governs scrollbar-related behaviour. 
I also actually took some time to have a close look at scroll-bar of Slack and it turns out the scrollbar it uses is a custom scrollbar UI that they've built themselves, not the default browser-generated one.



So we can assume there is clearly some javascript going on behind the scene that is related to their scrollbar.

So, In my opinion tracking&adjusting the scroll position within `resize` event handler attached to the viewport seems right approach here, like @Silver-IT did but I think there is a simpler logic that we can write by using [this method](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled) recommended in a MDN doc, which looks something like this.

```
resizeHandler () {
...
  const { scrollTop, scrollHeight, clientHeight } = this.$refs.conversation
        
  if ( scrollHeight - clientHeight - scrollTop < reasonable-value ) {
    this.$refs.conversation.scrollTop = scrollHeight
  }
}
```

EDIT:
`scrollHeight - clientHeight - scrollTop < reasonable-value` here is a check if the scrollbar is placed at the bottom of the area while viewport resizing is happening.

thnx.

Silver-IT:
Thanks for researching on this, @SebinSong.
The variable `this.ephemeral.scrolledDistance` is already being used so we don't need to take `scrollTop`, `scrollHeight`, `clientHeight` inside the `resizeEventHandler` function.

taoeffect (approved):
Nicely done @Silver-IT! It seems to work smoothly now. 😄 👍

PR #1713, opened 2023-09-05
Update chatroom privacy description when create chatroom

### LINKED ISSUE #1681: Update text explaining privacy settings when creating a chatroom
### Problem

When adding a group chatroom (default privacy selection), the modal shows 'All group members will be added to this channel.'. This is slightly incorrect, as joining a chatroom is a second, explicit option. What happens is that the channel will instead be visible to all group members.

### Solution

Update the text to say something that more closely reflects what happens, such as 'All group members will (have access to|be able to (join|see)) this channel.'

Silver-IT:
### Summary of Changes

- Closes #1681


taoeffect (approved):
👍

PR #1724, opened 2023-09-14
#1704 - Add 'Export to CSV' feature to payments table

### LINKED ISSUE #1704: Feature: Export payments history as CSV
### Problem

To make it convenient to use the app with accounting programs for whatever needs users may have, we need a way to export the data to a more usable format, like CSV, and we currently have no such feature.

### Solution

Add a link or button to the "Completed" and "Received" tabs that says "Export.", that when clicked saves a `.csv` file of all of the received or sent transactions. Remember that some users might have both tabs available (because they switched from being a pledger to a receiver or vice versa).

SebinSong:
closes #1704 

Pls note that in the generated CSV file, a blank space is used as the delimeter within a date string as `,` is used as a separator for columns in CSV itself.


SebinSong:
@taoeffect 

Just have updated the PR with additional design update (moving 'all periods' option into the dropdown), but I found one incorrect thing in the design though.

According to your issue description below,

> Add a link or button to the "Completed" and "Received" tabs that says "Export." ...

The requirement is adding the buttons to 'Completed' and 'Received' tabs but the figma design is showing TODO tab instead.

In fact, we actually already had a discussion on the position of the button [here](https://github.com/okTurtles/group-income/issues/1704#issuecomment-1718726726) and what I updated based on your answer there is more or less similar to what Leihla created in the figma which is below.

**(1)** If there is **enough** space for the button. - placing it **next to** other call-to-actions





**(2)** If there is **not enough** space for the button. - placing it **below** other call-to-actions




Are you happy with these positioning? If not, let me know your suggestion, will make another update.

SebinSong:
This is a screenshot for an example of a downloaded CSV file that I opened in MS Excel, btw.

taoeffect (changes requested):
> Are you happy with these positioning? If not, let me know your suggestion, will make another update.

Yes, the positioning seems good 👍 

I noticed a couple of issues:

### Changes to padding that need to be reverted

This PR modifies the vertical gap here, and those changes need to be reverted:

Screenshot 2023-09-26 at 9 58 31 AM

### Invalid Date for Export for Received Payments

Screenshot 2023-09-26 at 10 00 41 AM

This is curious - and perhaps we should wait until PR #1691 from @snowteamer is merged before merging this one, as it modifies a lot of code related to this.

SebinSong:
@taoeffect Resolved the extra padding issue and yeah sure, I can wait until his PR is completed and merged.

taoeffect:
After merging `master` in, can you also do some boyscoutting? I'm realizing I missed some details in my PR review of #1691.

For example, in `PaymentsMixin.js`, there's this function `getPaymentDetailsByPeriod` which has the following lines:

```js
        const paymentsKey = `payments/${this.ourUsername}/${period}/${this.currentGroupId}`
        const payments = await sbp('gi.db/archive/load', paymentsKey) || {}
```

However that is already done by `getHistoricalPaymentByHashAndPeriod`.

I would do the following: rename `getHistoricalPaymentByHashAndPeriod` to `getHistoricalPaymentByPeriod` and use it to DRY the code in `getPaymentDetailsByPeriod`.

Then in `PaymentDetails.vue`, update `initializeDetails` to use `getPaymentDetailsByPeriod` or `getHistoricalPaymentByPeriod`.

SebinSong:
@taoeffect 
**[My update](https://github.com/okTurtles/group-income/pull/1724/commits/1c28d1675e42203cae52930d66d82d0f36a6f516)** per your request contains below changes.

**[ 1 ]** Renamed `getAllPaymentPeriods` to `getAllPeriodPayments`
this method fetches all period payments **objects** of a group but original naming sounds like it is rather used to get all **period stamps**, not period payments **objects**, which I found confusing. so renamed it.

**[ 2 ]** Renamed `getHistoricalPaymentPeriods` to `getHistoricalPeriodPayments`
Same reason as `1.` above

**[ 3 ]** Renamed&updated `getHistoricalPaymentByHashAndPeriod` to `getHistoricalPaymentDetailsByPeriod` and then DRYed `getPaymentDetailsByPeriod` using it.

After spending more time on investigation, I realised your comment below

> I would do the following: rename getHistoricalPaymentByHashAndPeriod to getHistoricalPaymentByPeriod and use it to DRY the code in getPaymentDetailsByPeriod.

is actually a valid point..!
By updating it to give us all historical payment details rather than one details item, we can DRY the mixin. But I also renamed it to getHistorical**PaymentDetails**ByPeriod to make it clear it's used to fetch payment **details**.


Hope you agree with these too and let me know of your opinion if you have any better idea.

frontend/views/containers/payments/ExportPaymentsModal.vue
```diff
@@ -0,0 +1,188 @@
+<template lang='pug'>
+modal-template(ref='modal' :a11yTitle='modalTitle')
+  template(slot='title')
+    span {{ modalTitle }}
+
+  i18n.c-sub-title.has-text-1(:args='{ type: paymentType }') Export your {type} payment history to .csv
```
taoeffect: Minor request, for translators who just see this string by itself without comment or context, it would help to know what `{type}` might mean.

So could you rename this variable to `{sent_or_received}` ?

taoeffect (changes requested):
@SebinSong Fantastic work!

This PR is pretty much ready to merge based on my tests. I have only one minor request left for helping translators.

frontend/views/containers/payments/ExportPaymentsModal.vue
taoeffect: Actually, I'm realizing even this is wrong.

There are multiple problems with doing it this way:

1. `paymentType` isn't localized, so this is completely broken.
2. Even if `paymentType` were localized, it's still better to avoid inserting directly translated words or fragments into another string, but instead it is better to preserve the entire string and translate it as a whole. It's not always the case that if you correctly translate "sent" that that isolated translation will fit when inserted into the sentence "export your [sent] payment..." the direct translation of the word "sent" might need to be something different in that context.

So here I recommend using a computed property instead of `i18n`, and use `paymentType` to pick the full sentence:

```js
exportInstructions () {
  return this.paymentType === 'sent'
    ? L('Export your sent payment history to .csv')
    : L('Export your received payment history to .csv')
}
```

This is much safer and guarantees a correct translation.

taoeffect: It's a great thing you did this though!! 😄 

It's reminded me that I needed to update the Style Guide with instructions because this is a very common issue in the codebase: https://github.com/okTurtles/group-income/blob/master/docs/Style-Guide.md#user-facing-strings-guide

SebinSong: @taoeffect yeah, great catch on this one. it could've lead to an awkwardly translated sentence.

SebinSong:
@taoeffect Updated the PR again.
I actually spotted that same translation issue in the modal title so applied the same correction there too.

cheers,

taoeffect (approved):
Approved! Fantastic work @SebinSong!

PR #1728, opened 2023-09-18
Should not access group after being leaved

### LINKED ISSUE #1645: Error after leaving a group related to chatroom

### LINKED ISSUE #1716: Should not be accessible to the group unless group member tries

### LINKED ISSUE #1766: Simplify and remove `JOINING_GROUP_CHAT`
### Problem

I think we shouldn't use many global variables like `JOINING_GROUP`, `JOINING_GROUP_CHAT` if possible.
While I was investigating the joining group process, I have noticed that `JOINING_GROUP_CHAT` could be removed.

### Solution

There is a way to remove global variable `JOINING_GROUP_CHAT` in `controller/actions/group.js` and `model/contracts/group.js`. If we set `JOINING_GROUP` in the following way (just before and after syncing), we don't need to take care of `JOINING_GROUP_CHAT`.

```
// params.originatingContractID is set, it means that we're joining through an invite link,
const isJoiningGroup = !!params.originatingContractID
const preEffectBeforeSync = isJoiningGroup
  ? () => { sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true) }
  : () => {}
const postEffectAfterSync = isJoiningGroup
  ? () => { sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false) }
  : () => {}

console.log('@@@@@@@@ AT join for ' + params.contractID)
preEffectBeforeSync()
await sbp('chelonia/contract/sync', params.contractID)
postEffectAfterSync()
```

Silver-IT:
### Summary of Changes

- Closes #1716 
- Closes #1766 
- Closes #1645


taoeffect (commented):
@corrideat 

Could you test this PR and try it out?

The PR seems to fix the issue, but after the last step in #1716 I get this error:

Screenshot 2023-10-12 at 1 20 05 PM

Should I approve and merge anyway as-is?

frontend/controller/actions/group.js
```diff
@@ -325,7 +327,10 @@ export default (sbp('sbp/selectors/register', {
       console.log('@@@@@@@@ AT join for ' + params.contractID)
 
       await sbp('chelonia/contract/sync', params.contractID)
-      if (rootState.contracts[params.contractID]?.type !== 'gi.contracts/group') {
+      if (!rootState.contracts[params.contractID]) {
+        // NOTE: already kicked from the group by someone else
+        return
```
corrideat, line 334: This `return` looks like it'd fix the issue

frontend/controller/actions/group.js
```diff
@@ -316,7 +316,9 @@ export default (sbp('sbp/selectors/register', {
   // action if we haven't done so yet (because we were previously waiting for
   // the keys), or (c) already a member and ready to interact with the group.
   'gi.actions/group/join': async function (params: $Exact<ChelKeyRequestParams> & { options?: { skipUsableKeysCheck?: boolean; skipInviteAccept?: boolean } }) {
-    sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
+    if (!params.options?.skipInviteAccept) {
+      sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
```
corrideat: I need to test this, but I'd think it breaks waiting for an invite upon logging in (i.e., logging in before the keys have been received).

frontend/controller/actions/group.js
```diff
@@ -524,6 +527,7 @@ export default (sbp('sbp/selectors/register', {
       console.error('gi.actions/group/join failed!', e)
       throw new GIErrorUIRuntimeError(L('Failed to join the group: {codeError}', { codeError: e.message }))
     } finally {
+      sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
```
corrideat: This function returning does _not_ mean that the joining process is complete. It's complete when keys are received.

corrideat (commented):
I need to test this

corrideat:
I was able to reproduce, and I now need to text this PR. However, I expect it'd break in the following case:

1. U1 creates a group
2. U2 joins
3. U3 joins and immediately leaves
4. U1 and U2 close the browser
5. U3 joins again (is left at pending)
6. U3 logs out and closes the browser
7. (Optionally, U1 or U2 reopen GI)
8. U3 logs in again

Personally, I'd go with something like this:

```patch
diff --git a/frontend/controller/actions/group.js b/frontend/controller/actions/group.js
index bde4403e..65ecd0a6 100644
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -513,6 +513,15 @@ export default (sbp('sbp/selectors/register', {
         }
 
         sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
+      // We don't have the group secret keys, we're not waiting for a key
+      // share and we're marked as having left. This means we synced the contract
+      // and we've been kicked out
+      } else if (!hasSecretKeys && !sbp('chelonia/contract/isWaitingForKeyShare', state) && state.profiles[username].departedDate) {
+        sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
+        // DO SOMETHING HERE
+        // SOMETHING MEANS: Separate the logic for leaving, so that we can do it
+        // even if it's not an action we're processing in contracts/group.js
+
       // We have already sent a key request that hasn't been answered. We cannot
       // do much at this point, so we do nothing.
       // This could happen, for example, after logging in if we still haven't
```

frontend/controller/actions/group.js
Silver-IT: `'JOINING_GROUP-${contractID}'` is being used inside the sideEffect functions of group contracts.
It should be true only when joining group, not when try to log in or to see if the user has key.
That's why I added `if` conditional statement.

corrideat:
I've tested this PR, and it fails under simpler scenarios than I described. For example,

1. [BROWSER 1] u1 creates group
2. [BROWSER 2] u2 joins group
3. [BROWSER 1] u1 removes u2 from group
4. [BROWSER 2] u2 rejoins group, closes browser
5. [BROWSER **3**] u2 logs in

frontend/controller/actions/group.js
```diff
@@ -314,17 +314,29 @@ export default (sbp('sbp/selectors/register', {
...
-
+      const isJoiningGroup = !!params.originatingContractID
+      const preEffectBeforeSync = isJoiningGroup
+        ? () => { sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true) }
+        : () => {}
+      const postEffectAfterSync = isJoiningGroup
+        ? () => { sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false) }
+        : () => {}
+
+      preEffectBeforeSync()
```
SebinSong: @Silver-IT It seems `preEffectBeforeSync` and `postEffectAfterSync` here are only used once in this block of code, I think it would feel more readable if these are just
```js
if (isJoiningGroup) {
  sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, ...)
}
```
instead of making them as a function.

Silver-IT: Okay. I updated this way.
```
isJoiningGroup && sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
await sbp('chelonia/contract/sync', params.contractID)
```

Silver-IT (commented):
Ready for your reviews, @taoeffect, @corrideat.

taoeffect:
Closing in favor of recent changes in `e2e-protocol` and `e2e-protocol-ricardo`.

PR #1729, opened 2023-09-19
Set us up with Github CI

### LINKED ISSUE #974: Set us up with GitHub CI
### Problem

Currently we use Travis for CI, and it would be great if our PRs could also use GH's new built-in CI system which is available in their "Actions" feature.

Here's an example of a project and config that uses GH CI: https://github.com/vlang/v/blob/master/.github/workflows/ci.yml

One of the problems we've had with CI in general is that it's been a bit slow to run all our tests, and sometimes we have to re-run them. By adding support for GH CI we get resilience in case there's ever an issue with Travis. We can also check to see which one we like better.

Another advantage of GH CI is that it supports build artifacts, which can be useful, for example, for downloading Cypress videos (Travis also supports build artifacts, but it's more of a pain to set that up).

### Solution

Create the appropriate `.github/workflows/ci.yml` file. As an example, there's:

- https://github.com/marketplace/actions/setup-node-js-environment

And the template workflow for creating a CI GI action:

```yml
# This is a basic workflow to help you get started with Actions

name: CI

# Controls when the action will run. Triggers the workflow on push or pull request
# events but only for the master branch
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

# A workflow run is made up of one or more jobs that can run sequentially or in parallel
jobs:
  # This workflow contains a single job called "build"
  build:
    # The type of runner that the job will run on
    runs-on: ubuntu-latest

    # Steps represent a sequence of tasks that will be executed as part of the job
    steps:
      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
      - uses: actions/checkout@v2

      # Runs a single command using the runners shell
      - name: Run a one-line script
        run: echo Hello, world!

      # Runs a set of commands using the runners shell
      - name: Run a multi-line script
        run: |
          echo Add other actions to build,
          echo test, and deploy your project.

```

snowteamer:
Closes #974

### Summary of changes:
- New `ci.yml` file in the `.gitbub/workflows/` directory.

### Additional information:
- Tested on my fork:

image

- Details:

image

- Build logs:
[logs_46.zip](https://github.com/okTurtles/group-income/files/12656984/logs_46.zip)

- Not sure whether caching actually works and doesn't prevent things to work e.g. when doing a Cypress version upgrade

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,20 @@
+name: NodeJS with Grunt
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        cache: 'npm'
+    - run: npm ci
+    - run: npm test


taoeffect (approved):
Amazing, thanks @snowteamer!

PR #1745, opened 2023-09-28
E2e protocol ricardo

corrideat:


frontend/controller/actions/chatroom.js
```diff
@@ -121,7 +129,17 @@ export default (sbp('sbp/selectors/register', {
             foreignKey: cekOpts.foreignKey,
             meta: cekOpts.meta,
             data: cekOpts.data
-          }
+          },
+          encryptedOutgoingDataWithRawKey(CEK, {
```
taoeffect, line 134: Can you leave a TODO comment here to figure out how to avoid having the developer interact with `encryptedOutgoingDataWithRawKey`?

frontend/controller/actions/chatroom.js
```diff
@@ -174,8 +192,39 @@ export default (sbp('sbp/selectors/register', {
...
+    const CEKid = sbp('chelonia/contract/currentKeyIdByName', params.contractID, 'cek')
+    const userCSKid = sbp('chelonia/contract/currentKeyIdByName', userID, 'csk')
+
+    const state = rootState[params.contractID]
+
+    // Add the user's CSK to the contract
+    await sbp('chelonia/out/keyAdd', {
+      contractID: params.contractID,
+      contractName: 'gi.contracts/chatroom',
+      data: [encryptedOutgoingData(state, CEKid, {
```
taoeffect, line 217: Same request here re TODO comment

frontend/controller/actions/group.js
```diff
@@ -213,6 +220,18 @@ export default (sbp('sbp/selectors/register', {
...
+           encryptedOutgoingDataWithRawKey(CEK, {
+            foreignKey: `sp:${encodeURIComponent(userID)}?keyName=${encodeURIComponent('csk')}`,
+            id: userCSKid,
+            data: rootState[userID]._vm.authorizedKeys[userCSKid].data,
+            permissions: [GIMessage.OP_ACTION_ENCRYPTED + '#inner'],
+            allowedActions: '*',
+            purpose: ['sig'],
+            ringLevel: Number.MAX_SAFE_INTEGER,
+            name: `${userID}/${userCSKid}`
+          }) */
```
taoeffect: If this is no longer needed can it be removed?

frontend/controller/actions/group.js
```diff
@@ -281,11 +301,24 @@ export default (sbp('sbp/selectors/register', {
           }
         },
         signingKeyId: CSKid,
-        encryptionKeyId: CEKid
+        encryptionKeyId: CEKid,
+        innerSigningContractID: null
```
taoeffect, line 297: Does this need to be specified as `null` or can it be removed?

If it needs to be specified as `null` can you add a comment explaining why?

frontend/controller/actions/group.js
```diff
@@ -281,11 +301,24 @@ export default (sbp('sbp/selectors/register', {
...
+      await sbp('gi.actions/group/joinAndSwitch', {
+        originatingContractID: userID,
+        originatingContractName: 'gi.contracts/identity',
+        contractID: contractID,
+        contractName: 'gi.contracts/group',
+        signingKeyId: CSKid,
+        innerSigningKeyId: userCSKid,
+        encryptionKeyId: userCEKid,
+        options: {
+          skipUsableKeysCheck: true
```
taoeffect, line 313: Can you please add a comment explaining why this is being set to `true` here?

frontend/controller/actions/group.js
```diff
@@ -309,27 +342,28 @@ export default (sbp('sbp/selectors/register', {
...
+      keyIds.length && await sbp('gi.actions/out/shareVolatileKeys', {
         contractID: contractID,
         contractName: 'gi.contracts/group',
         subjectContractID: userID,
-        keyIds: [PEKid, ...revokedPEKids]
+        keyIds: keyIds
       })
 
+      */
+
```
taoeffect: If all this code is no longer needed should it be deleted?

frontend/model/contracts/group.js
```diff
@@ -552,9 +552,7 @@ sbp('chelonia/defineContract', {
             allowPublicChannels: false
           },
           streaks: initGroupStreaks(),
-          profiles: {
-            [meta.username]: initGroupProfile(meta.createdDate)
-          },
+          profiles: {},
```
taoeffect, line 555: Why was this removed?

shared/domains/chelonia/chelonia.js
```diff
@@ -567,6 +565,26 @@ export default (sbp('sbp/selectors/register', {
     // calling this will make pubsub unsubscribe for events on `contractID`
     sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, state.contracts)
   },
+  'chelonia/contract/disconnect': async function (contractID, contractIDToDisconnect) {
```
taoeffect, line 568: Is there a corresponding `'chelonia/contract/connect'`? If not, why not? It seems like there should be one for clarity.

frontend/model/contracts/group.js
corrideat: Because it's set when joining the group, which is now something that the group creator also does. (Also, because `meta.username` will be removed).

shared/domains/chelonia/chelonia.js
```diff
@@ -567,6 +565,26 @@ export default (sbp('sbp/selectors/register', {
...
+    const state = sbp(this.config.stateSelector)
+    const contractState = state[contractID]
+
+    const keyIds = Object.values(contractState._vm.authorizedKeys).filter((k) => {
+      // $FlowFixMe
+      return k._notAfterHeight == null && k.meta?.keyRequest?.contractID === contractIDToDisconnect
+    }).map(k => (k: any).id)
+
+    console.log('chelonia/contract/disconnect:  ', { keyIds })
```
taoeffect: Is this needed? Can it be removed? (We should start cleaning up the excess logging...)

shared/domains/chelonia/chelonia.js
```diff
@@ -567,6 +565,26 @@ export default (sbp('sbp/selectors/register', {
...
+
+    if (!keyIds.length) return
```
taoeffect, line 577: Is this a silent error? (All errors should be logged, exceptions thrown, etc.)

frontend/controller/actions/group.js
corrideat: This is related to how the group creator also joins normally (and it can be deleted).

shared/domains/chelonia/GIMessage.js
```diff
@@ -50,61 +62,143 @@ export type GIOpKeyUpdate = {
...
+            if (computedKeyId !== key.id) {
+              throw new Error(`Key ID mismatch. Expected to decrypt key ID ${key.id} but got ${computedKeyId}`)
+            }
+          })
+        }
+        if (key.meta?.keyRequest?.contractID) {
+          try {
+            key.meta.keyRequest.contractID = maybeEncryptedIncomingData(contractID, state, key.meta.keyRequest.contractID, height, additionalKeys, headJSON)?.valueOf()
+          } catch {
+            delete key.meta.keyRequest.contractID
```
taoeffect, line 110: Silent error? If an error shouldn't be logged, a comment should always explain why not.

Otherwise, errors should always be logged.

frontend/controller/actions/group.js
corrideat: Maybe it doesn't need to be `null` but it makes more sense, since the General chatroom is part of creating the group and not created by any specific user (technically, it's created by the group creator, but that's an implementation detail).

shared/domains/chelonia/chelonia.js
corrideat: There isn't by that name, but it'd be `/out/keyRequest'.

frontend/controller/actions/group.js
corrideat: It can

shared/domains/chelonia/GIMessage.js
```diff
@@ -50,61 +62,143 @@ export type GIOpKeyUpdate = {
...
+  // If the operation is GIMessage.OP_KEY_SHARE,
+  // extract encrypted data from keys?.[].meta?.private?.content
+  if (op === GIMessage.OP_KEY_SHARE) {
+    return maybeEncryptedIncomingData(contractID, state, (message: any), height, additionalKeys, headJSON, (message) => {
+      (message: any).keys?.forEach((key) => {
+        if (!key.meta?.private?.content) return
         const decryptionFn = message.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
-        const decryptionContract = message.foreignContractID ? message.foreignContractID : contractID
-        key.meta.private.content = decryptionFn(decryptionContract, state, key.meta.private.content, height, additionalKeys, (value) => {
+        const decryptionContract = String(message.foreignContractID ? message.foreignContractID : contractID)
```
taoeffect: Can't this be simplified to:

```js
message.foreignContractID || contractID
```

without the `String` wrapper?

corrideat: The `String` wrapper is for casting as it needs to be a string.

shared/domains/chelonia/chelonia.js
```diff
@@ -852,64 +893,67 @@ export default (sbp('sbp/selectors/register', {
...
+    const signingKeyId = findSuitableSecretKeyId(originatingState, [GIMessage.OP_KEY_ADD], ['sig'])
+    if (!signingKeyId) {
+      throw ChelErrorUnexpected(`Unable to send key request. Originating contract is missing a key with OP_KEY_ADD permission. contractID=${contractID} originatingContractID=${originatingContractID}`)
+    }
+    await sbp('chelonia/out/keyAdd', {
+      contractID: originatingContractID,
+      contractName: originatingContractName,
+      data: [{
+        id: keyRequestReplyKeyId,
+        name: 'krrk-' + keyRequestReplyKeyId,
```
taoeffect: Either needs to be done now or later, all internal key names (and reserved keys) should be prefixed with `#`

shared/domains/chelonia/crypto.js
```diff
@@ -255,7 +255,7 @@ export const verifySignature = (inKey: Key | string, data: string, signature: st
     throw new Error('Invalid signature')
   }
 }
-export const encrypt = (inKey: Key | string, data: string): string => {
+export const encrypt = (inKey: Key | string, data: string, ad?: string): string => {
```
taoeffect, line 264: What is `ad`? Can you please add comments here explaining function parameters, especially `ad` which is very unclear?

shared/domains/chelonia/chelonia.js
corrideat: Not needed other than for debugging purposes

shared/domains/chelonia/internals.js
```diff
@@ -159,6 +196,13 @@ export default (sbp('sbp/selectors/register', {
     saferEval({
       // pass in globals that we want access to by default in the sandbox
       // note: you can undefine these by setting them to undefined in exposedGlobals
+      ...(typeof self !== 'undefined' && {
+        self: {
+          crypto: {
+            getRandomValues: (v) => self.crypto.getRandomValues(v)
+          }
+        }
+      }),
```
taoeffect: Should this be exposed under `self`...? None of the other globals are. Maybe it should be directly included under `crypto`?

(And are we sure this is necessary here still?)

shared/domains/chelonia/internals.js
```diff
@@ -881,19 +930,55 @@ export default (sbp('sbp/selectors/register', {
...
+            }))
           },
-          signingKeyId
+          signingKey: deserializedResponseKey
         })
 
         // 4(i). Remove originating contract and update current contract with information
-        await sbp('chelonia/out/keyRequestResponse', { contractID, contractName, signingKeyId, data: { keyRequestHash: hash, success: true } })
+        const payload = { keyRequestHash: hash, success: true }
+        // TODO: Combine these two operations (KRR and KS) into one using OP_ATOMIC
```
taoeffect: Is this something that can be done now? Or should we do it in a separate PR?

shared/domains/chelonia/chelonia.js
corrideat: This isn't an error per se. If you call disconnect but there is nothing to disconnect, it returns (this ensures idempotency).

taoeffect (changes requested):
Great work @corrideat!! There are some flow errors, those need to be fixed before merging. Also see other comments

taoeffect (approved):
LGTM! Nice job @corrideat !!

PR #1747, opened 2023-10-02
Remove archived data when leave the group

### LINKED ISSUE #1737: [e2e-protocol] Proposals widget misbehavior when rejoining a group
### Problem

Screenshot 2023-09-22 at 10 10 06 AM

1. Create a group and join 5 users: u1, u2, u3, u4, u5
2. u1 creates proposal to remove u4
3. u2 and u3 vote to remove u4
4. u4 rejoin group using original invite link

Duplicate key errors and multiple proposals shown for u4 (when viewing group as u4 upon rejoining)

### Solution

Find and fix the bug.

Please branch off of `e2e-protocol` where this bug is, and send in a PR to that branch, not `master`.

Silver-IT:
### Summary of Changes

- Closes #1737


frontend/model/contracts/group.js
```diff
@@ -908,6 +908,11 @@ sbp('chelonia/defineContract', {
         const { username } = rootState.loggedIn
 
         if (data.member === username) {
+          // NOTE: should remove archived data from IndexedStorage
+          //       regarding the current group (proposals, payments)
+          sbp('gi.contracts/group/removeArchivedProposals', contractID)
+          sbp('gi.contracts/group/removeArchivedPayments', contractID)
+
```
taoeffect: 1. I think this should be below, after the check for `'JOINING_GROUP-'`
2. These are async methods that should be await'd on, and therefore the `sideEffect` should also be rewritten to be `async sideEffect`

frontend/model/contracts/group.js
```diff
@@ -1373,6 +1378,11 @@ sbp('chelonia/defineContract', {
       await sbp('gi.db/archive/save', key, proposals)
       sbp('okTurtles.events/emit', PROPOSAL_ARCHIVED, contractID, proposalHash, proposal)
     },
+    'gi.contracts/group/removeArchivedProposals': async function (contractID) {
+      const { username } = sbp('state/vuex/state').loggedIn
+      const key = `proposals/${username}/${contractID}`
+      await sbp('gi.db/archive/delete', key)
+    },
```
taoeffect: Please move this down below next to `'gi.contracts/group/removeArchivedPayments'` so that the file stays nicely organized

frontend/model/contracts/group.js
```diff
@@ -1434,6 +1444,13 @@ sbp('chelonia/defineContract', {
 
       sbp('okTurtles.events/emit', PAYMENTS_ARCHIVED, { paymentsByPeriod, payments })
     },
+    'gi.contracts/group/removeArchivedPayments': async function (contractID) {
+      const { username } = sbp('state/vuex/state').loggedIn
+      const archPaymentsByPeriodKey = `paymentsByPeriod/${username}/${contractID}`
+      const archSentOrReceivedPaymentsKey = `sentOrReceivedPayments/${username}/${contractID}`
+      await sbp('gi.db/archive/delete', archPaymentsByPeriodKey)
+      await sbp('gi.db/archive/delete', archSentOrReceivedPaymentsKey)
+    },
```
taoeffect: Nice, but you forgot `payments/${username}/${period}/${contractID}`

taoeffect (changes requested):
There are two ways to address issue #1737:

1. Is what this PR does.
2. Is by modifying `ProposalsWidget.vue`'s `onProposalArchived` method to not add duplicates to `this.ephemeral.archivedProposals`

After thinking on it for a while, I've decided that the approach taken by this PR is preferable, because if we are no longer part of a group, we should clean up locally stored data related to that group.

I found some issues and commented on them. Please note that this PR has merge conflicts with `e2e-protocol` due to recent merges, so please merge `e2e-protocol` to address those conflicts. Also, I'm also not sure why the chatroom 0.2.0 contracts are being updated by this PR, as this PR is only supposed to modify the `group.js` contracts. Perhaps after merging `e2e-protocol` and re-running `grunt pin:0.2.0 --overwrite` that should fix it.

Silver-IT:
@taoeffect, I've also spent some time to think about how this issue should be fixed. Of course, we can make it not add duplicated payments. But first of all, the duplicate payments shouldn't be exist.

And thank for your feedback. :+1:

frontend/model/contracts/group.js
Silver-IT: Oh, yes. Nice catch. 👍🏼

frontend/model/contracts/group.js
Silver-IT: Nice catch. :+1:

taoeffect (approved):
Nicely done @Silver-IT! LGTM!

EDIT: actually, this is still an issue. The same problem still happens.

PR #1750, opened 2023-10-04
#1742 Use 'submit-button' component for asynchronous actions

### LINKED ISSUE #1742: Disable buttons and add spinner to them
### Problem

In various parts of the app, there are buttons that perform operations that are asynchronous.

For example, when creating a DM in `e2e-protocol` branch, it takes a while to create the contracts and dismiss the modal. During this time the creation button should be disabled so that it cannot be clicked twice accidentally. However it is not:

Screenshot 2023-09-27 at 10 28 08 AM

### Solution

On a branch off of the `e2e-protocol` branch, go through the app and check every button that creates contracts or performs any sort of long-running operation.

Make sure that the button cannot be clicked twice while the operation is running, and that it displays a spinner like this:

Screenshot 2023-09-27 at 10 29 21 AM

Make sure the buttons are **immediately** disabled when they are clicked.

SebinSong:
closes #1742

### DIFF
--- a/frontend/views/components/UsersSelector.vue
+++ b/frontend/views/components/UsersSelector.vue
@@ -32,21 +32,20 @@ form.c-search-form(@submit.prevent='')
         )
 
   .buttons.is-end.c-button-container(v-if='usernames.length')
-    i18n.is-success.c-create-btn(
-      tag='button'
-      type='button'
-      @click='$emit("submit")'
-    ) Create
+    button-submit.is-success.c-create-btn(@click='submitHandler')
+      i18n Create
 </template>
 
 <script>
 import { mapGetters } from 'vuex'
 import AvatarUser from '@components/AvatarUser.vue'
+import ButtonSubmit from '@components/ButtonSubmit.vue'
 
 export default ({
   name: 'UsersSelector',
   components: {
-    AvatarUser
+    AvatarUser,
+    ButtonSubmit
   },
   props: {
     label: {
@@ -113,6 +112,9 @@ export default ({
       }
 
       this.value = this.$refs.input.textContent
+    },
+    async submitHandler (event) {
+      await this.$listeners.submit(event)
     }
   },
   watch: {
--- a/frontend/views/containers/chatroom/ChatMembersAllModal.vue
+++ b/frontend/views/containers/chatroom/ChatMembersAllModal.vue
@@ -95,20 +95,22 @@ modal-base-template.has-background(
                   .c-display-name(v-if='displayName' data-test='profileName') @{{ username }}
 
             .c-actions(v-if='isJoined')
-              i18n.button.is-outlined.is-small(
+              button-submit.button.is-outlined.is-small(
                 v-if='!joinedDate'
-                tag='button'
+                type='button'
                 @click.stop='addToChannel(username)'
                 :data-test='"addToChannel-" + username'
-                :args='LTags("span")'
-              ) Add {span_} to channel{_span}
+              )
+                i18n(:args='LTags("span")') Add {span_} to channel{_span}
+
               .has-text-success(v-else)
                 i.icon-check
                 i18n Added.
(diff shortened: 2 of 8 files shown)


frontend/views/components/UsersSelector.vue
```diff
@@ -32,21 +32,20 @@ form.c-search-form(@submit.prevent='')
         )
 
   .buttons.is-end.c-button-container(v-if='usernames.length')
-    i18n.is-success.c-create-btn(
-      tag='button'
-      type='button'
-      @click='$emit("submit")'
-    ) Create
+    button-submit.is-success.c-create-btn(@click='submitHandler')
+      i18n Create
```
SebinSong, line 36: In order to use a `button-submit` component, had to pass an async method as the click handler (otherwise it leads to a Vue console warning requiring it). That is why I created/passed an async submitHandler here instead of emitting a 'submit' event.

frontend/views/containers/chatroom/ChatMembersAllModal.vue
```diff
@@ -95,20 +95,22 @@ modal-base-template.has-background(
...
+              button-submit.button.is-outlined.is-small(
                 v-if='!joinedDate'
-                tag='button'
+                type='button'
                 @click.stop='addToChannel(username)'
                 :data-test='"addToChannel-" + username'
-                :args='LTags("span")'
-              ) Add {span_} to channel{_span}
+              )
+                i18n(:args='LTags("span")') Add {span_} to channel{_span}
```
SebinSong, line 104: wasn't quite sure about why span_` and `_span` are used here but just left it as is for now.

frontend/views/containers/user-settings/UserProfile.vue
```diff
@@ -69,6 +69,7 @@
           i18n.is-danger This action cannot be undone.
 
         .buttons
+          // TODO: once account-deletion is implemented, replace this with submit-button component to take care of the async process.
```
taoeffect: In the comment, can you rename `submit-button` to `ButtonSubmit.vue`? It makes it clearer, at least to me, which component is being discussed.

frontend/views/containers/payments/SendThankYouModal.vue
```diff
@@ -86,9 +86,9 @@ export default ({
         this.submit()
       }
     },
-    submit () {
+    async submit () {
       try {
-        sbp('gi.actions/group/sendPaymentThankYou', {
+        await sbp('gi.actions/group/sendPaymentThankYou', {
```
taoeffect, line 92: Good catch! Can you please also:

1. wrap the user-facing error message below with an `L` function
2. Add above line 102: `console.error('SendThankYouModal caught:', err)`

taoeffect (changes requested):
Nicely done @SebinSong! Left a few comments.

Also, when creating a DM and clicking the button here really quickly with the mouse:

Screenshot 2023-10-04 at 9 02 21 AM

I see these errors appear in the console:

Screenshot 2023-10-04 at 9 02 02 AM

frontend/views/containers/user-settings/UserProfile.vue
SebinSong: done!

frontend/views/containers/payments/SendThankYouModal.vue
SebinSong: done!

SebinSong:
@taoeffect Updated the PR again with the work for your feedbacks.

Regarding the console errors you pointed out, they are caused by Vue failing to anchor a `ref` to `ModalBaseTemplate.vue` because its root element is not an HTML element but a built-in Vue component (`<transition />`)
So I updated the `closeModal` method so that it emits `CLOSE_MODAL` event directly from within there, which clears those errors.

taoeffect (approved):
Great work!!!

PR #1753, opened 2023-10-06
Show user profile cards when user clicks name in payments

### LINKED ISSUE #1746: Show User Profile Cards when user clicks name in Payments
### Problem

This won't be an issue when we get better payment integration with Lightning #766, but at the moment it's a pain to have to switch to the dashboard to see a user's payment details.

Instead, we should be able to click on a user's profile name or profile icon in the Payments row to bring up the same profile card as can be seen in the dashboard user list.

Screenshot 2023-10-01 at 11 55 33 AM

### Solution

Implement:clicking on a user's profile name or profile icon in the Payments row should bring up their profile card with payment details.

Screenshot 2023-10-01 at 11 59 11 AM

Silver-IT:
### Summary of Changes

- Closes #1746

### DIFF
--- a/frontend/views/containers/payments/payment-row/PaymentRow.vue
+++ b/frontend/views/containers/payments/payment-row/PaymentRow.vue
@@ -6,8 +6,12 @@
       slot(name='cellUser')
       template(v-if='!$slots["cellUser"]')
         .c-user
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Brilliant! Love that it was so simple to add this! 😄

PR #1757, opened 2023-10-10
#1756 User marked as 'needing' even if their income is above the group's mincome

### LINKED ISSUE #1756: User marked as 'needing' even if their income is above the group's mincome
### Problem

Adjusting the group mincome doesn't adjust the member's status.

Steps to reproduce:

1. Create a group with a $1 mincome.
2. Add payment information by answering 'no' to 'Do you make at least $1 per month?' and entering $0.99 as income
3. Change the group's mincome to $0.01

#### What happens

1. The user is shown 'You need $-0.98. You will receive $0.' in the UI
2. Editing the payment information has 'no' as the answer to 'Do you make at least $0.01 per month?', with the income still being $0.99.

#### What should happen

1. The user is shown a different text, since they no longer need
2. Editing the payment information should have 'yes' as the answer to 'Do you make at least $0.01 per month?' (or maybe the payment information can be discarded)

### Solution

Review the way that the state is stored, processed or presented to display the correct information to the users in this case.

SebinSong:
closes #1756 

This is a fix for [this symptom](https://github.com/okTurtles/group-income/issues/1756#issuecomment-1752270090) in  `master` branch.

### DIFF
--- a/frontend/views/containers/proposals/ProposalTemplate.vue
+++ b/frontend/views/containers/proposals/ProposalTemplate.vue
@@ -199,7 +199,7 @@ export default ({
   },
   methods: {
     close () {
-      this.$refs.modal.close()
+      this.$refs.modal.unload()
     },
     next () {
       // TODO/BUG - we must clear formMsg (if visible) when changing steps.


taoeffect (approved):
Strangely enough, this seems to fix the issue! Great job @SebinSong!

PR #1758, opened 2023-10-10
E2e protocol ricardo

corrideat:


frontend/model/contracts/group.js
```diff
@@ -970,7 +970,11 @@ sbp('chelonia/defineContract', {
 
             // gi.contracts/group/removeOurselves will eventually trigger this
             // as well
-            sbp('gi.contracts/group/rotateKeys', contractID, state)
+            sbp('gi.contracts/group/rotateKeys', contractID, state).then(() => {
+              sbp('gi.contracts/group/revokeGroupKeyAndRotateOurPEK', contractID, false)
```
taoeffect: so that the `.catch` also catches exceptions on this inner SBP call, I think you want to use `return` here, right?

taoeffect (approved):
LGTM!!!!

PR #1759, opened 2023-10-11
#1751 - Dismissing a payment causes user to get banned

### LINKED ISSUE #1751: [Bug] Dismissing a payment causes user to get banned
### Problem

1. Set up a group with a mincome of 1000 and 3 users and add income details, with u1 pledging $200, u2 with $800 income (needing $200), and u3 with income $750 (needing $250).
2. Once all income details are entered, enter the following command into the console for u1:

    ```
    ```

3. Then click the "TODO" tab, and click the 3 dots to bring up the menu next to a TODO. Click "Dismiss this payment":

Screenshot 2023-10-04 at 9 32 15 AM

You should see the following errors in the console:

Screenshot 2023-10-04 at 9 31 29 AM

Additionally, u1 will be kicked from the group for sending a malformed message.

### Solution

There are a few issues going on here.

- For some reason in `PaymentRowTodo.vue`, it thinks the payment has a hash. In `cancelPayment` the `if (this.hash)` passes, even though it shouldn't. Why does the payment have a hash? These are TODOs, they haven't been sent yet... so it shouldn't have a hash.
- It's not clear to me why TODOs can be dismissed at all, and I do not think we've implemented support in the distribution algorithm for dismissing payments. It's also unclear to me the circumstances under which dismissing a TODO would make sense (cc @dotmacro @leihla), and such circumstances are not described in the [Figma designs](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=7781%3A4530&mode=design&t=TOKBVgW4JGfiOyV6-1). Even if there were a legitimate reason to dismiss TODOs, to get this working we would need to make changes to how we handle the output of the distribution algorithm, and that would be a very difficult endeavor in of itself.

So, at the very least, let's just not allow TODOs to be dismissed. Hide this menu item. If a payment TODO row has no menu items associated with it, the 3 dots should be hidden as well. I am also curious why the `PaymentRowTodo.vue` has a hash associated with it at all...

SebinSong:
closes #1751

### DIFF
--- a/frontend/views/containers/payments/PaymentRowTodo.vue
+++ b/frontend/views/containers/payments/PaymentRowTodo.vue
@@ -31,32 +31,20 @@
 
     template(slot='cellActions')
       .cpr-date(:class='payment.isLate ? "pill is-danger" : "has-text-1"') {{ humanDate(payment.date) }}
-      payment-actions-menu
-        menu-item(
-          tag='button'
-          item-id='message'
-          icon='times'
-          @click='cancelPayment'
-        )
-          i18n Dismiss this payment
 </template>
 
 <script>
-import sbp from '@sbp/sbp'
 import { mapGetters } from 'vuex'
 import { humanDate } from '@model/contracts/shared/time.js'
 import { MenuItem } from '@components/menu/index.js'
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
LGTM!

PR #1763, opened 2023-10-12
Make sure we sync any identity contracts we haven't synced upon login

### LINKED ISSUE #1684: Make sure we sync any identity contracts we haven't sync'd upon login
### Problem

In the sideEffect for `'gi.contracts/group/inviteAccept'`, it's possible that we fail to sync user contracts for some reason or another (a server could be down, etc.):

https://github.com/okTurtles/group-income/commit/0c1634b7343dc5d27fecc241506fcb70dfc50917#r123335604

### Solution

Upon logging in, check the users we have in our group, and use `'namespace/lookup'` to see if there are any identity contracts that we haven't sync'd, and sync those.

Silver-IT:
### Summary of Changes

- Closes #1684

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -316,9 +316,7 @@ export default (sbp('sbp/selectors/register', {
   // action if we haven't done so yet (because we were previously waiting for
   // the keys), or (c) already a member and ready to interact with the group.
   'gi.actions/group/join': async function (params: $Exact<ChelKeyRequestParams> & { options?: { skipUsableKeysCheck?: boolean; skipInviteAccept?: boolean } }) {
-    if (!params.options?.skipInviteAccept) {
-      sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
-    }
+    sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
     try {
       const rootState = sbp('state/vuex/state')
       const username = rootState.loggedIn.username
@@ -519,7 +517,10 @@ export default (sbp('sbp/selectors/register', {
           Object.keys(state.profiles)
             .map(username => sbp('namespace/lookup', username))
         )).filter(id => !rootState[id] && !sbp('chelonia/contract/isSyncing', id))
-        await sbp('chelonia/contract/sync', missingIDs)
+        if (missingIDs.length > 0) {
+          console.info('found unsynced identity contracts to sync:', missingIDs)
+          await sbp('chelonia/contract/sync', missingIDs)
+        }
 
         sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
       // We have already sent a key request that hasn't been answered. We cannot
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -514,14 +514,12 @@ export default (sbp('sbp/selectors/register', {
           })
         }
 
-        (async () => {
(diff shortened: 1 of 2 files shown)


frontend/controller/actions/identity.js
```diff
@@ -318,10 +318,7 @@ export default (sbp('sbp/selectors/register', {
       const additionalIdentityContractIDs = await Promise.all(chatRoomUsers.filter(username => {
         return getters.ourUsername !== username && !getters.ourContacts.includes(username)
       }).map(username => sbp('namespace/lookup', username)))
-
-      for (const identityContractID of additionalIdentityContractIDs) {
-        await sbp('chelonia/contract/sync', identityContractID)
-      }
```
taoeffect, line 324: This is a nice simplification 👍

frontend/controller/actions/group.js
```diff
@@ -316,7 +316,9 @@ export default (sbp('sbp/selectors/register', {
   // action if we haven't done so yet (because we were previously waiting for
   // the keys), or (c) already a member and ready to interact with the group.
   'gi.actions/group/join': async function (params: $Exact<ChelKeyRequestParams> & { options?: { skipUsableKeysCheck?: boolean; skipInviteAccept?: boolean } }) {
-    sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
+    if (!params.options?.skipInviteAccept) {
```
taoeffect: Why was this change added?

And I feel like this `skipInviteAccept` must have been removed for some reason, maybe @corrideat can also comment on that?

frontend/controller/actions/group.js
```diff
@@ -512,6 +514,13 @@ export default (sbp('sbp/selectors/register', {
           })
         }
 
+        // NOTE: sync identity contracts which are out of sync after joining group
+        const missingIDs = (await Promise.all(
+          Object.keys(state.profiles)
+            .map(username => sbp('namespace/lookup', username))
+        )).filter(id => !rootState[id] && !sbp('chelonia/contract/isSyncing', id))
+        await sbp('chelonia/contract/sync', missingIDs)
```
taoeffect: I recommend wrapping this in an `if` to check if `missingIDs.length > 0`, and then you can also add a line to: `console.info('found unsynced identity contracts to sync:', missingIDs)`

taoeffect (changes requested):
Nice work @Silver-IT!

Minor change requests + questions for you and @corrideat

frontend/controller/actions/group.js
Silver-IT: Got it.

frontend/controller/actions/group.js
Silver-IT: Actually, `JOINING_GROUP-params.contractID` was designed to be used in `sideEffect` functions only when joining group, not when logging in. It provide us the way to ignore removing group contract if we meet `removeOurselves` while joining group. `skipInviteAccept` is true only when login.

Having investigated the `gi.actions/group/join` function and login process, I have noticed that we sync the group contracts many times while logging in. And it's not easy to decide when should set this `JOINING_GROUP-params.contractID` true. If we remove the `if` conditional statement, there won't be any errors, but the contract would be removed and synced which is redundant activity.
At the moment, I am not sure when should I set, which conditional check would be correct, so I will remove this `if` conditional statement.

However, this `JOINING_GROUP-params.contractID` should be set `true` only when we sync the group contract to join.

taoeffect (approved):
Nicely done 👍

PR #1767, opened 2023-10-16
Removed skipInviteAccept and JOINING_GROUP_CHAT

### LINKED ISSUE #1766: Simplify and remove `JOINING_GROUP_CHAT`

Silver-IT:
### Summary of Changes

- Closes #1766


Silver-IT:
Will close this PR and add commit to [this PR](https://github.com/okTurtles/group-income/pull/1728).

PR #1768, opened 2023-10-16
Support retriable Chelonia actions

### LINKED ISSUE #1755: Implement a persistent action queue
### Problem

Some operations (like `'gi.actions/identity/shareNewPEK'`, and various actions that call `publishEvent`), might fail due to random chance (e.g. power going out somewhere, internet going down, etc.). But these are operations that we don't want to forget about - we want to retry them even if the computer/browser is restarted, and keep retrying them until they succeed.

### Solution

Create a new domain called `chelonia.persistentActions`.

Using it looks something like this:

```js
sbp('chelonia.persistentActions/enqueue', [
  // ex 1: use default options to save and invoke this selector
  ['gi.actions/identity/shareNewPEK', ...],
  // ex 2: enqueue an action using options
  {
    // the invocation to try
    invocation: ['gi.contracts/group/removeForeignKeys', contractID, ...],
    // the algorithm that determines how long to wait before recalling the selector
    retryAlgorithm: 'fibonnaci',
    // alternatively, how many seconds to wait between retrys
    retrySeconds: 30,
    // maximum number of tries, default: Inf
    maxAttempts: 100,
    // condition to check, if returns true, causes the invocation to be skipped
    skipCondition: ['server/hasEvent', hash],
    // selector to call each time the invocation fails
    errorInvocation: ['callMeOnFailure'],
    // selector to call when maxAttempts reached
    totalFailureInvocation: ['callMeWhenWeGiveUp']
  }
])
// => returns an array containing corresponding UUIDs for the invocations

// forces a retry of all existing persisted actions
sbp('chelonia.persistentActions/forceRetry')

// forces a retry specific action
sbp('chelonia.persistentActions/forceRetry', '<UUID>')

// return information about what actions are currently enqueued
sbp('chelonia.persistentActions/status')
// =>
// [
//   {
//     id: '<UUID>',
//     invocation: ['gi.contracts/group/removeForeignKeys', contractID],
//     nextRetry: '...string explaining when it will be next invoked...',
//     retryAttemptsSoFar: 5,
//     lastError: 'string containing err.message'
//   }
// ]]

// cancel a specific action by its UUID
sbp('chelonia.persistentActions/cancel', '<UUID>')
```

Uses `chelonia.db` to persist actions.

snowteamer:
Closes #1755

### Summary of changes:
- This PR implements the 'chelonia.persistentActions' SBP domain, according to requirements described in #1755


frontend/main.js
```diff
@@ -43,6 +44,7 @@ const { Vue, L } = Common
 
 console.info('GI_VERSION:', process.env.GI_VERSION)
 console.info('CONTRACTS_VERSION:', process.env.CONTRACTS_VERSION)
+console.info('LIGHTWEIGHT_CLIENT:', JSON.stringify(process.env.LIGHTWEIGHT_CLIENT))
```
taoeffect: I don't think stringify is needed here?

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,200 @@
...
+    status.lastError = ''
+    status.nextRetry = ''
+    status.pending = false
+    this.options.successInvocationSelector &&
+      await this.trySBP([this.options.successInvocationSelector, result])
+  }
+
+  trySBP (invocation: SbpInvocation | void): any {
+    try {
+      return invocation ? sbp(...invocation) : undefined
```
taoeffect: there is no `await` on the call to `sbp` so if an exception occurs on a selector that's async it won't be caught

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,200 @@
...
+      console.error(tag, error.message)
+    }
+  }
+}
+
+// SBP API
+
+sbp('sbp/selectors/register', {
+  'chelonia.persistentActions/_init' (): void {
+    sbp('okTurtles.events/on', LOGIN, (function () {
```
taoeffect: `LOGIN` is special to Group Income, it cannot be used in Chelonia.

taoeffect (changes requested):
Preliminary review - looking really solid so far...!

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,200 @@
...
+      this.actionsByID = Object.create(null)
+      this.databaseKey = `chelonia/persistentActions/${sbp('state/vuex/getters').ourIdentityContractId}`
```
taoeffect: Similarly, as this is a utility for Chelonia (and possibly in the future, an independent library), we cannot access `'state/vuex/getters'` either

shared/domains/chelonia/persistent-actions.js
snowteamer: In that case maybe I can make this module define a public `init` selector, which will be called from within Group Income on login.

shared/domains/chelonia/persistent-actions.js
snowteamer: Ditto

shared/domains/chelonia/persistent-actions.js
taoeffect: Most of this stuff can be left in `_init`. E.g. `this.actionsByID = Object.create(null)` etc. are all things that can be done inside of `/_init`.

The only thing that seems to need to be configured at runtime is the `databaseKey`, so similarly to `chelonia` itself you could add a `chelonia.persistentActions/configure` selector where that can be set.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,200 @@
...
+
+  async handleError (error: Error): Promise<void> {
+    const { options, status } = this
+    // Update relevant status fields before calling any optional selector.
+    status.failedAttemptsSoFar++
+    status.lastError = error.message
+    const anyAttemptLeft = options.maxAttempts > status.failedAttemptsSoFar
+    status.nextRetry = anyAttemptLeft && !status.cancelled
+      ? new Date(Date.now() + options.retrySeconds * 1e3).toISOString()
+      : ''
```
taoeffect: Hmm, it doesn't seem like the value of `status.nextRetry` is actually used when determining when to retry? 🤔

shared/domains/chelonia/persistent-actions.js
snowteamer: True but every call to `trySBP` in this file occurs within an `await` expression, so if the selector returns a promise it will be awaited.

shared/domains/chelonia/persistent-actions.js
snowteamer: That's because `setTimeout` isn't based on absolute time but on some relative number of milliseconds in the future. On the other hand, `status.nextRetry` is defined as an ISO timestamp, so it has to be absolute

taoeffect: Right, but if it's not really serving a purpose then why have it at all?

The `if` condition below could be changed from `if (status.nextRetry)` to `if (anyAttemptLeft && !status.cancelled)`

taoeffect (changes requested):
Looking really good @snowteamer!!

I'm wondering if we can also adjust the default options? Perhaps using the existing selector `'chelonia.persistentActions/configure'`?

frontend/main.js
```diff
@@ -240,13 +242,17 @@ async function startApp () {
...
         this.ephemeral.finishedLogin = 'yes'
 
         if (this.$store.state.currentGroupId) {
           this.initOrResetPeriodicNotifications()
           this.checkAndEmitOneTimeNotifications()
         }
+        const databaseKey = `chelonia/persistentActions/${sbp('state/vuex/getters').ourIdentityContractId}`
+        sbp('chelonia.persistentActions/configure', { databaseKey })
+        await sbp('chelonia.persistentActions/load')
+        await sbp('chelonia.persistentActions/retryAll')
```
taoeffect: I think `'chelonia.persistentActions/load'` should automatically start the retries

snowteamer: That would be fine, but on the other hand `/save` doesn't stop the pending retries

taoeffect: It shouldn't. The persistent action queue is something that is always active as long as it has actions inside of it while the app is running.

shared/domains/chelonia/persistent-actions.js
taoeffect: Yes, but you can still do that if you `await` here.

By not calling `await` here, it defeats the purpose of this function. You might as well remove `trySBP` (personally I would remove it and directly wrap calls that might fail in `try/catch` at their call site).

The `try/catch` here won't get triggered if there's an exception thrown here by `sbp` for a selector that is async. Instead there will be an unhandled promise error shown in the console. try/catch only works with `await` when dealing with async functions.

snowteamer: > I would remove it and directly wrap calls that might fail in `try/catch` at their call site).
We can do that, but that's going to require four `try/catch` blocks instead of one

shared/domains/chelonia/persistent-actions.js
taoeffect: Oh, it's part of the description in the issue, that I wrote... lol. 🤦‍♂️

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,203 @@
...
+    status.pending = false
+  }
+
+  async handleSuccess (result: any): Promise<void> {
+    const { status } = this
+    status.lastError = ''
+    status.nextRetry = ''
+    status.pending = false
+    this.options.successInvocationSelector &&
+      await this.trySBP([this.options.successInvocationSelector, result])
```
taoeffect: Can we refactor this to use events instead of `options.successInvocationSelector`?

I think it makes sense to add at least two events using `okTurtles.events`:

- `CHEL_PERSISTENT_ACTIONS_SUCCESS` - would be emitted above here
- `CHEL_PERSISTENT_ACTIONS_FAILURE` - would be emitted next to line 82 above, next to the `errorInvocation`, along with the `error` object

And if you think of any others that we should emit, feel free to suggest

taoeffect (changes requested):
I think this is almost ready to merge.

@snowteamer - I think you might have missed this request from the previous review:

> I'm wondering if we can also adjust the default options? Perhaps using the existing selector `'chelonia.persistentActions/configure'`?

This way we can override the default options for values like `maxAttempts` and `retrySeconds`.

Also added one more request about emitting events.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  async handleError (error: Error): Promise<void> {
+    const { id, options, status } = this
+    // Update relevant status fields before calling any optional code.
+    status.failedAttemptsSoFar++
+    status.lastError = error.message
+    const anyAttemptLeft = options.maxAttempts > status.failedAttemptsSoFar
+    if (!anyAttemptLeft) status.resolved = true
+    status.nextRetry = anyAttemptLeft && !status.resolved
```
Silver-IT, line 92: `!status.resolved` could be ignored, I think.
```
status.nextRetry = anyAttemptLeft ? new Date(Date.now() + options.retrySeconds * 1e3).toISOString() : ''
```

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+
+  // TODO: validation
+  'chelonia.persistentActions/configure' ({ databaseKey, options = {} }: { databaseKey: string; options: Object }): void {
+    if (!databaseKey) throw new TypeError(`${tag} 'databaseKey' is required`)
+    this.databaseKey = databaseKey
+    for (const key in options) {
+      if (key in defaultOptions) {
+        defaultOptions[key] = options[key]
+      } else {
+        throw new TypeError(`${tag} Unknown option: ${key}`)
```
Silver-IT, line 161: Is it necessary to throw an error? I think it's fine to call `console.error` instead, and go on.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+      }
+    }
+  },
+
+  'chelonia.persistentActions/enqueue' (...args): UUIDV4[] {
```
Silver-IT, line 166: I think it's preferable to declare possible types of `args` as a comment above this function.
BTW: Why is it necessary to accept two types of `arg` for `PersistantAction`?
```
[invocation, options]
{ invocation, ...options }
```
I think the second format isn't necessary. And also `invocation` could be added in `options` of `PersistantAction` instance.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+  },
+
+  // Loads and tries every stored persistent action under the configured database key.
+  async 'chelonia.persistentActions/load' (): Promise<void> {
+    const storedActions = JSON.parse((await sbp('chelonia/db/get', this.databaseKey)) ?? '[]')
+    for (const { id, invocation, options } of storedActions) {
+      this.actionsByID[id] = new PersistentAction(invocation, options)
+      // Use the stored ID instead of the autogenerated one.
+      // TODO: find a cleaner alternative.
+      this.actionsByID[id].id = id
```
Silver-IT, line 199: Why not to add `id` in the arguments of `PersistantAction` constructor?

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+// SBP API
+
+sbp('sbp/selectors/register', {
+  'chelonia.persistentActions/_init' (): void {
+    this.actionsByID = Object.create(null)
+    // Necessary for now as _init cannot be async. Becomes true when 'configure' has been called.
+    this.ready = false
```
Silver-IT: I can't find the place to set `this.ready` `true` in `configure` function. Did you miss it by mistake?

Silver-IT (changes requested):
Really impressive PR, @snowteamer. :+1::+1::+1:
Just left a few comments.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+
+  async handleError (error: Error): Promise<void> {
+    const { id, options, status } = this
+    // Update relevant status fields before calling any optional code.
+    status.failedAttemptsSoFar++
+    status.lastError = error.message
+    const anyAttemptLeft = options.maxAttempts > status.failedAttemptsSoFar
+    if (!anyAttemptLeft) status.resolved = true
+    status.nextRetry = anyAttemptLeft && !status.resolved
+      ? new Date(Date.now() + options.retrySeconds * 1e3).toISOString()
```
corrideat, line 93: Using `Date` is a bad idea for anything besides maybe logging. The reason is that the _date_ could change arbitrarily. The better alternative is to use a monotonic timer for this, such as `performance.now()`. I'm not sure exactly what this `nextRetry` field is used for, exactly, though.

The second issue is the, like for the `this.timer` assignment below, that `1e3` seems like a pretty low value for a retry

shared/domains/chelonia/persistent-actions.js
```diff
@@ -0,0 +1,219 @@
...
+    sbp('okTurtles.events/emit', PERSISTENT_ACTION_FAILURE, { error, id })
+    await this.trySBP(options.errorInvocation)
+    if (!anyAttemptLeft) {
+      sbp('okTurtles.events/emit', PERSISTENT_ACTION_TOTAL_FAILURE, { error, id })
+      await this.trySBP(options.totalFailureInvocation)
+    }
+    // Schedule a retry if appropriate.
+    if (status.nextRetry) {
+      // Note: there should be no older active timeout to clear.
+      this.timer = setTimeout(() => this.attempt(), this.options.retrySeconds * 1e3)
```
corrideat, line 106: 1e3 seems like a pretty low value

snowteamer: Ditto

shared/domains/chelonia/persistent-actions.js
snowteamer: The `nextRetry` field is actually used for logging as per the issue requirement for `'chelonia.persistentActions/status'`

> 1e3 seems like a pretty low value for a retry

 `options.retrySeconds` is a number of seconds, but we need milliseconds here, hence the `1e3` multiplier

shared/domains/chelonia/persistent-actions.js
snowteamer: Indeed, good catch. BTW I've since replaced it by a direct check for availability of `databaseKey`

shared/domains/chelonia/persistent-actions.js
snowteamer: That's a possible choice, but I felt like it was not that great either

shared/domains/chelonia/persistent-actions.js
snowteamer: Good point about the `args` comment.
> Why is it necessary to accept two types of arg for PersistantAction?

Here I just followed the issue requirements.

> And also invocation could be added in options of PersistantAction instance.

An invocation is required in any case, so I don't think `invocation` should go in `options` since it's not actually optional.

shared/domains/chelonia/persistent-actions.js
snowteamer: Maybe it's fine, but `console.error` won't help detect error automatically e.g. in future tests

shared/domains/chelonia/persistent-actions.js
snowteamer: `status.resolved` can be set independently by `.cancel()`, therefore checking for `anyAttemptLeft` was not enough

taoeffect (changes requested):
Nice work @snowteamer. Looks almost ready to merge. Seems like the only thing remaining are the logging out related changes we discussed today.

shared/domains/chelonia/persistent-actions.js
taoeffect: > Here I just followed the issue requirements.

Oh, it was a good thing @Silver-IT brought this up, because I missed this in my review.

This is **not** what the args look like according to the issue:

```
[invocation, options]
{ invocation, ...options }
```

Rather, they look like this:

```
[selector, arg1, arg2, arg3]
{ invocation, ...options }
```

So to answer @Silver-IT's question: the reason for allowing both types is because it makes the syntax when enqueing invocations simpler.

In most cases you can just pass in the sbpInvocation like so: `[selector, arg1, arg2, arg3]`

But when you want to adjust the defaults for a specific invocation, you use the second syntax.

~~I'm noticing that this code does not correctly handle the first syntax.~~

EDIT: oops, my mistake, the code seems to correctly handle this.

taoeffect (approved):
Great work @snowteamer! Approved & merged! 😄 🎉

PR #1770, opened 2023-10-18
#926 Create a PWA for the app

### LINKED ISSUE #926: PWA: create a progressive web app wrapper [$200 bounty]
### Problem

The user needs to get notifications about actions that happen in the app.

### Solution

The easiest way to implement this at the moment other than relying on emails or chat is to create an application that the user can install on their device and receive notifications

This ticket concern only the PWA implementation and not the notifications itself
Step 1: Create the PWA
Step 2: Test PWA with PWA Builder (https://www.pwabuilder.com/)
Step 3: Test app on different devices
Step 4: Test notifications to see if it work

Useful link: https://whatpwacando.today/

### Bounty

$200 bounty for a clean solution to this (paid in cryptocurrency).

SebinSong:
closes #926 

@taoeffect 
As briefly discussed via Slack chat, this PR is for achieving two things

**1.** Make the Group Income installable in both Desktop / Mobile devices (Have only tested in Chrome desktop & Android Chrome so far)
**2.** Ensure the app installed in **1.** receives a push notification via service worker.

I used [`web-push`](https://www.npmjs.com/package/web-push) npm package to test/implement push notification in the back-end side but I understand this can be replaced/updated in the future to better suit the project.

Another thing to note is, it appears that in order for a website to be installable as a PWA, it MUST be served in secure context(e.g. https:// or localhost. Please refer to [this doc](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#secure_context) for details)
So, pls make sure to run `grunt dev --tunnel` and uses **https** url it gives to test the PWA on mobile browsers.

Hope it all sounds good.
Thanks.


frontend/assets/pwa-manifest.json
```diff
@@ -0,0 +1,41 @@
+{
```
SebinSong: I wanted to prepare as many icon sizes as possible here (used [this online tool](https://www.iloveimg.com/resize-image/resize-png) to generate multiple dimension icons, in case you wonder), but let me know if we want to reduce the numbers here. (for example, [MDN](https://developer.mozilla.org/en-US/) website only has two icon entries 192x192 and 512x512 in its manifest.json)

frontend/main.js
```diff
@@ -206,6 +207,9 @@ async function startApp () {
     return
   }
 
+  // register service-worker
+  await sbp('service-workers/setup')
```
SebinSong, line 227: I placed this sbp call here only because it's before the app Vue instance is created but let me know if you think it has to be moved elsewhere.

backend/push.js
```diff
@@ -0,0 +1,19 @@
+const pushInstance = require('web-push')
+const pushSubscriptions: Map<string, any> = new Map()
+
+// VAPID public/private keys, generated via 'npx web-push generate-vapid-keys' command.
+const VAPID_PUBLIC_KEY = 'BPZrB5emRHd9w36EKEGuyF5rTZNf11qxp4Hr57Td0I3acZWt5jK8xLUhYpzgQ7P8OpH3kUuHSNsNxre4qr1vtiM'
+const VAPID_PRIVATE_KEY = 'jza_sGZNhBaT53r9VnQVf17ejZST9zv4nx8WX-nNKks'
+
+pushInstance.setVapidDetails(
+  'mailto:support@okTurtles.org', // it's an arbitrary email placeholder. should be a company email.
```
SebinSong: This is a dummy email address I put as a placeholder. has to be replaced with the official contact email of the company.

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,95 @@ sbp('sbp/selectors/register', {
...
+      if (existingSubscription) {
+        // If there is an existing subscription, no need to create a new one.
+        // But make sure server knows the subscription details too.
+        await fetch(`${API_URL}/push/subscribe`, { method: 'POST', body: JSON.stringify(existingSubscription.toJSON()) })
+        return
+      } else {
+        // Create a new push subscription
+
+        // 1. Get the public key from the server
+        const PUBLIC_VAPID_KEY = await fetch(`${API_URL}/push/publickey`)
```
corrideat: The response may or may not be what you expect (e.g., it could be a 404 page)

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,95 @@ sbp('sbp/selectors/register', {
     // TODO: move ahead with encryption stuff ignoring this service worker stuff for now
     // TODO: improve updating the sw: https://stackoverflow.com/a/49748437
     // NOTE: user should still be able to use app even if all the SW stuff fails
+    if (!('serviceWorker' in navigator)) { return }
```
corrideat, line 13: Wouldn't it be prudent to check for everything needed here? For example, `typeof Notification !== 'undefined'`

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,95 @@ sbp('sbp/selectors/register', {
...
+        return
+      } else {
+        // Create a new push subscription
+
+        // 1. Get the public key from the server
+        const PUBLIC_VAPID_KEY = await fetch(`${API_URL}/push/publickey`)
+          .then(handleFetchResult('text'))
+
+        // 2. Add a new subscription to pushManager using it.
+        const subscription = await registration.pushManager.subscribe({
```
corrideat: ditto

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,95 @@ sbp('sbp/selectors/register', {
...
+      )
+    } else {
+      console.error('No existing subscription found!')
+    }
   }
 })
+
+// helper method
+
+function urlBase64ToUint8Array (base64String) {
```
corrideat, line 157: Shorter version:

```
new Uint8Array(atob('AAAAAA'.replace(/-/g, '+').replace(/_/g, '/')).split('').map((c) => c.charCodeAt(0)))
```

A few notes:

1. I don't believe the padding is needed
2. I'd just all `atob` (as above). If you must refer to the global object, I'd use `self` or `globalThis` over `window`, as these are more generic.

frontend/index.html
```diff
@@ -6,6 +6,7 @@
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
     <link rel="stylesheet" href="/assets/css/main.css">
+    <link rel="manifest" href="/assets/pwa-manifest.json" />
```
corrideat: Nitpicking, but I believe the customary extension is `.webmanifest`, so that the right media type `application/manifest+json` can be applied.

backend/routes.js
```diff
@@ -369,3 +369,63 @@ route.POST('/zkpp/updatePasswordHash/{contract}', {
 
   return Boom.internal('internal error')
 })
+
+// PWA push notification
+
+route.GET('/push/publickey', {}, function () {
+  return VAPID_PUBLIC_KEY
```
corrideat: It'd be nice to add the headers for this, such as `content-type`, etc. I've looked and unfortunately there doesn't seem to be a standard URL for this, so this path is fine (otherwise, I'd have suggested using the standard path).

frontend/assets/pwa-manifest.json
corrideat: I can't remember now, but I believe 'just' 192×192 and 512×512 is fine for all cases where this might be used. _But_, you have various types of favicons that could be set (e.g., for Apple and Microsoft browsers / OSes). I suggest taking a look here: https://realfavicongenerator.net/

frontend/controller/service-worker.js
SebinSong: @corrideat Did you encounter the 404 type of page while testing? (I never have during development/testing) This block of code is wrapped with `try ... catch`, so that in any occasion where a failure happens in this series of steps it will just silently fails rather than interrupting in-app user flow.

frontend/controller/service-worker.js
SebinSong: Agreed.

frontend/controller/service-worker.js
SebinSong: @corrideat good explanation but I would rather leave it as is if they both eventually do the same job.
FYI, This method is originally from [here](https://gist.github.com/Klerith/80abd742d726dd587f4bd5d6a0ab26b6), which was referenced in multiple articles that feature the `web-push` npm package. The fact that it was mentioned in those popular articles made me decide it's safe to assume it's trusted by the Javascript community.

frontend/index.html
SebinSong: good to know!

frontend/controller/service-worker.js
SebinSong: `registration` here is an instance created by `navigator.serviceWorker.register()` in above line. So I believe checking `'serviceWorker' in navigator` earlier will suffice.

frontend/controller/service-worker.js
SebinSong: Re the point **2.**, At the moment this method is only used in browser context. So using `window` is fine.

frontend/assets/pwa-manifest.json
SebinSong: @corrideat That online tool seems pretty useful.

I'm not sure about preparing multiple icon types here though.
I haven't actually seen any docs or articles saying the icons should be prepared as multiple types yet and also, the live PWA websites I used as a reference (e.g. [MDN](https://developer.mozilla.org/en-US/), https://www.stuff.co.nz/ etc.) only have `.png` files specified for their icons. 

below is the `manifest.webmanifest` screenshot of [stuff.co.nz](https://www.stuff.co.nz/), which is New Zealand's largest news webzine.


I think it should be fine to have two `.png` (192x192, 512x512) and a `.svg`

backend/routes.js
SebinSong: @corrideat Good point. updated accordingly.

SebinSong:
@corrideat Thanks for your feedback. I've updated the PR for some of your comments that I agreed with too.

cc. @taoeffect

backend/push.js
```diff
@@ -0,0 +1,19 @@
+const pushInstance = require('web-push')
+const pushSubscriptions: Map<string, any> = new Map()
+
+// VAPID public/private keys, generated via 'npx web-push generate-vapid-keys' command.
+const VAPID_PUBLIC_KEY = 'BPZrB5emRHd9w36EKEGuyF5rTZNf11qxp4Hr57Td0I3acZWt5jK8xLUhYpzgQ7P8OpH3kUuHSNsNxre4qr1vtiM'
+const VAPID_PRIVATE_KEY = 'jza_sGZNhBaT53r9VnQVf17ejZST9zv4nx8WX-nNKks'
```
taoeffect: Since these shouldn't be hard-coded, can you add a TODO comment above this to pull these in from either environment variables or a file?

In fact they can come from both, with envvars taking precedence over a config file.

I would recommend envvars that look like this:

- `process.env.VAPID_PUBLIC_KEY`
- `process.env.VAPID_PRIVATE_KEY`
- `process.env.VAPID_EMAIL`

And the config file can be a file called `giconf.json` in the project root that's added to `.gitignore`.

backend/push.js
taoeffect: Same for this (from envvars first, then config file)

backend/push.js
```diff
@@ -0,0 +1,19 @@
+const pushInstance = require('web-push')
+const pushSubscriptions: Map<string, any> = new Map()
```
taoeffect: Note that our pubsub system that @snowteamer wrote already has this. Would it make sense to integrate the two together instead of maintaining two separate mappings?

backend/routes.js
```diff
@@ -369,3 +369,63 @@ route.POST('/zkpp/updatePasswordHash/{contract}', {
...
+    params: Joi.object({
+      subscriptionId: Joi.string().required()
+    })
+  }
+}, function (req, h) {
+  const { subscriptionId } = req.params
+
+  pushSubscriptions.delete(subscriptionId)
+  return { deletedId: subscriptionId }
+})
```
taoeffect: Do we really need to create new routes for subscribing/unsubscribing when the websocket API already has these?

frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -199,20 +200,36 @@ export default ({
...
+            await sbp('gi.actions/group/paymentUpdate', {
+              contractID: this.currentGroupId,
+              data: {
+                paymentHash: msg.hash(),
+                updatedProperties: {
+                  status: PAYMENT_COMPLETED
+                }
+              }
+            })
+          }
```
taoeffect: What is happening here? AFAIK these changes were already included in `e2e-protocol` or `master`, is that right?

These changes, as well as others above, seem unrelated to the push API, and already integrated somewhere.. I think with some git magic and re-branching you should be able to figure out how to make these disappear from this PR (or a separate PR that's re-sent)?

taoeffect (changes requested):
Excellent work @SebinSong - PR reviewed!

frontend/controller/service-worker.js
corrideat: I haven't, but it's still incorrect to use the response if it was unsuccessful (an example could be [504 Gateway Timeout](https://http.dev/504)). `fetch` doesn't throw if a response is received (unlike, e.g., axios), so the response code must be checked.

However, I see that this is handled in the .then part.

SebinSong:
@taoeffect Updated PR based on your comments.
There is a two comments of yours I could not resolve right now, which are related to 

- The pubsub system of the application
- Websocket API

My current back-end knowledge level is not deep enough to reply for these two yet.
(learning WebSocket from some online resources now btw)
So if you think these optimisation work is not urgent, can this be handled as a follow-up issue of this PR? I can work on them in near future once I learn enough in these topics.

frontend/views/containers/payments/MonthOverview.vue
```diff
@@ -19,8 +19,8 @@
         :secValue='item.hasPartials ? item.value + 0.5 : 0'
         :hasMarks='item.hasMarks'
       )
-      p(:class='{"has-text-success": item.max === item.value}')
-        i.icon-check.is-prefix(v-if='item.max === item.value')
+      p(:class='{ "has-text-success": item.max && (item.max === item.value) }')
+        i.icon-check.is-prefix(v-if='item.max && (item.max === item.value)')
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) here.

frontend/views/containers/payments/PaymentRowTodo.vue
```diff
@@ -31,32 +31,20 @@
...
     template(slot='cellActions')
       .cpr-date(:class='payment.isLate ? "pill is-danger" : "has-text-1"') {{ humanDate(payment.date) }}
-      payment-actions-menu
-        menu-item(
-          tag='button'
-          item-id='message'
-          icon='times'
-          @click='cancelPayment'
-        )
-          i18n Dismiss this payment
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) here.

frontend/views/containers/payments/PaymentRowTodo.vue
```diff
@@ -80,34 +68,11 @@ export default ({
...
-            }
-          })
-        } else {
-          alert(L("Cannot dismiss a payment that hasn't been sent yet."))
-        }
-      } catch (e) {
-        console.error(e)
-        alert(e.message)
-      }
-    },
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) in this file too.

frontend/views/containers/payments/PaymentsList.vue
```diff
@@ -154,6 +154,7 @@ export default ({
   &.c-is-todo {
     th.c-th-check-all {
       width: 1.125rem;
+      font-size: 14px; // font-size here has to be fixed. Otherwise, it leads to a UI bug(The check icon moving downwards)
```
taoeffect: And here too.

frontend/views/containers/contributions/TodoHistory.vue
```diff
@@ -57,14 +58,17 @@ export default ({
...
+          .filter(({ data }) => data.status === PAYMENT_NOT_RECEIVED).length
         const missedCount = getLen(lastAdjustedDistribution || {})
         this.history.unshift({
-          total: doneCount === 0 ? 0 : doneCount / (doneCount + missedCount),
+          total: doneCount === 0 ? 0 : (doneCount - markedAsNotReceivedCount) / (doneCount + missedCount),
           title: this.getPeriodFromStartToDueDate(period),
           tooltipContent: [
             L('Total: {total}', { total: doneCount + missedCount }),
-            L('Completed: {completed}', { completed: doneCount })
+            L('Completed: {completed}', { completed: doneCount - markedAsNotReceivedCount })
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) for this file.

frontend/views/components/graphs/Progress.vue
```diff
@@ -33,11 +33,14 @@ export default ({
...
-      return `${100 * this.value / this.max}%`
+      return !this.max
+        ? '0%' // When this.max 0, the calculation below becomes 'NaN%', so manually specifying 0% here.
+        : `${100 * this.value / this.max}%`
     },
     percentSoft () {
       if (!this.secValue) return false
-      return `${100 * this.secValue / this.max}%`
+
+      return !this.max ? '0%' : `${100 * this.secValue / this.max}%`
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) for this file.

frontend/model/state.js
```diff
@@ -430,17 +433,22 @@ const getters = {
...
+
     const paymentsTotal = ourAdjustedPayments.length + receivedOrSent.length
     const nonLateAdjusted = ourAdjustedPayments.filter((p) => !p.isLate)
-    const paymentsDone = paymentsTotal - nonLateAdjusted.length
+    const paymentsDone = paymentsTotal - nonLateAdjusted.length - markedAsNotReceived.length
     const hasPartials = ourAdjustedPayments.some(p => p.partial)
-    const amountDone = receivedOrSent.reduce((acc, payment) => acc + payment.amount, 0)
-    const amountLeft = ourAdjustedPayments.reduce((acc, payment) => acc + payment.amount, 0)
+    const amountDone = receivedOrSent.reduce(sumUpAmountReducer, 0) - markedAsNotReceivedTotal
+    const amountLeft = ourAdjustedPayments.reduce((acc, payment) => acc + payment.amount, 0) + markedAsNotReceivedTotal
```
taoeffect: I have the [same question as with `RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337) for this file.

taoeffect (changes requested):
I think I found a few more places in this PR that change stuff that I'm not sure should be changed, seems like the same thing that happened with [`RecordPayment.vue`](https://github.com/okTurtles/group-income/pull/1770#discussion_r1387017337)?

> My current back-end knowledge level is not deep enough to reply for these two yet.
(learning WebSocket from some online resources now btw)
So if you think these optimisation work is not urgent, can this be handled as a follow-up issue of this PR? I can work on them in near future once I learn enough in these topics.

Well, I don't think we need to create a separate issue out of it because it's all really part of the same design that we need to think about for solving the original issue.

We can definitely wait for your to evaluate and add more commits as needed to this PR.

Also, I don't think you need to study WebSockets too much, because @snowteamer has already done that and used the `ws` library to implementing some code on top of it. Mainly what needs to be studied is @snowteamer's code to see if it can be used with this PR.

frontend/views/containers/payments/RecordPayment.vue
SebinSong: I think all these unrelated changes are from `master` branch and looks like it happened in [this commit ](https://github.com/okTurtles/group-income/pull/1770/commits/834299de8021ac2e9c05cb18208d7fb69a468053) when what I wanted/should have done was merging the latest `e2e-protocol` branch instead.

backend/push.js
SebinSong: Hi @snowteamer, 
Can I borrow your opinion for what Greg suggested in the above comment please?
(It's my first time doing a task that involves changes in the back-end area and need some help)

`pushSubscriptions` map here stores push-subscription data that users send  as `{ subscriptionId` : `subscription-data }` format and the `pushInstance` above it uses these data to send push notification.

Currently all push notification related logics on the server side (e.g. user sending their subscription info to the server, server sending a push notification down to users upon a POST request etc.) are implemented as corresponding REST API (all `/push/*` routes in `routes.js`) but what @taoeffect suggests is to integrate these into the existing pubsub system.

I will have to take some more time to study/figure out how the pubsub works, but would you have some advice on how I can approach this task?

Thanks,

snowteamer: The current pubsub system is somewhat contract-specific (I could work on that if requested), so there would be a few caveats

I will try to summarize how to extend the default protocol. Both the client and server can receive and send JSON messages.

To add custom WebSocket message to the pubsub protocol, two steps are needed:

### Server-side
Custom message handlers can be registered by adding them as entries of the `messageHandlers` option in the `createServer` call in `server.js`.

For example, a `push-sub` message handler will be called when the server receives a `push-sub` message.
To implement it, you would have to update the following code in `server.js`:
https://github.com/okTurtles/group-income/blob/3200d2534e192a0c29b1e8901d1ba740b259dfa2/backend/server.js#L88-L95

Then it would become something like:
```js
sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
  messageHandlers: {
    [ PUSH_SUB ] (message: Message) {
      // Your handler code here.
    }
  },
  serverHandlers: {
    connection (socket: Object, request: Object) {
      const versionInfo = { GI_VERSION, CONTRACTS_VERSION }
      socket.send(createNotification(NOTIFICATION_TYPE.VERSION_INFO, versionInfo))
    }
  }
}))
```

A few notes about message handlers:
-  Their `this` binding refers to the connected `ws` socket object. See https://github.com/websockets/ws.
    That socket can be used e.g. to send messages back to the connected client. A few helpers like `createMessage`, `createNotification` and `createErrorResponse` are available to create such messages. However note that in that case, the client must have matching handlers. For example, if the server sends a `foo` message back to the client, the client must have a `foo` message handler.
    
### Messages
- Messages are ordinary JSON objects, albeit with a required read-only `type: string` field, but no `id` field like in some other implementations. I would add in a more generic system, but for un/subscribing to contracts a `contractID` field was enough.

### Client-side
In the browser, the pubsub client is created in `'chelonia/connect'`:
https://github.com/okTurtles/group-income/blob/3200d2534e192a0c29b1e8901d1ba740b259dfa2/shared/domains/chelonia/chelonia.js#L141
You can see it has a `messageHandlers` option like its server-side counterpart, so that you can register handlers for custom message types.

The client wraps a `socket: WebSocket` property, which can be used to send data.

Besides, the client API is quite minimalist and has not much beyond `pub`, `sub` and `unsub`.
- Currently, the client `pub` method is not implemented, as it was not required to publish contract events a.k.a. GI messages to the server (that was already done via POST requests).
- If the client has to send custom messages, this should be possible using `client.socket.send()`, or `this.socket.send()` from within message handlers. Helpers `createMessage` and `createRequest` are available.
- The default `sub` client method likely won't work to subscribe to anything else than GI contracts, since if you call `sub(subscriptionId)` then the current code will try to sync a non-existent `subscriptionId` contract upon reconnection, because of this code in the client socket `open` event handler:
    https://github.com/okTurtles/group-income/blob/3200d2534e192a0c29b1e8901d1ba740b259dfa2/shared/pubsub.js#L330-L334

SebinSong: @taoeffect Regarding your suggestion above about moving the push subscription map into the pubsub system, Did you have a specific place in your mind when you were mentioning that?

I was thinking that 
[here](https://github.com/okTurtles/group-income/blob/3200d2534e192a0c29b1e8901d1ba740b259dfa2/backend/pubsub.js#L94C1-L94C1) in `createServer()` of `backend/pubsub.js` (adding `server.pushSubscriptions = new Map()` there) 
would be ideal, because that way this map can be referenced easily in any server-side message handler via something like `this.server.pushSubscriptions`. Does it sound good to you too?

taoeffect: @SebinSong I'm not sure. When I've used pubsub systems before, they've worked really well with the concept of "rooms" - or in other words, identifiers/channels that you subscribe to, that are identified by arbitrary strings of the developer's choosing. Maybe we can use this opportunity to make the existing pubsub system we have more generic? Then all of the subscriptions can be done on the same map potentially.

SebinSong:
@taoeffect 
Just updated the PR with the work of replacing the `/push/*` REST APIs with the `pubsub` system equivalents.

Thanks again @snowteamer for writing up [this comment](https://github.com/okTurtles/group-income/pull/1770#discussion_r1398047350), which was so helpful that I came back to read multiple times while working on this task. (along with [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) for the client side, and [`ws`](https://github.com/websockets/ws/blob/HEAD/doc/ws.md) npm package for the server side)

I tried my best to use/extend the system in a way that I think is organised but that's the parts that needs review/advice from experts like you guys.

While reviewing, make sure to unregister the service-worker if there is already one installed from reviewing previously so that it is re-installed again.
Thanks,

backend/pubsub.js
```diff
@@ -91,6 +95,7 @@ export function createServer (httpServer: Object, options?: Object = {}): Object
   server.messageHandlers = { ...defaultMessageHandlers, ...options.messageHandlers }
   server.pingIntervalID = undefined
   server.subscribersByContractID = Object.create(null)
+  server.pushSubscriptions = new Map()
```
taoeffect: Is there a reason we're using a `Map` here? If not, can we switch to using `Object.create(null)` for consistency? My understanding of the difference between these two is that `Map` can use non-string keys, and `Object` has more functionality (methods) that can be used (like `Object.entries()`).

frontend/controller/service-worker.js
```diff
@@ -13,23 +16,120 @@ sbp('sbp/selectors/register', {
...
+      } else {
+        // Create a new push subscription
+
+        sbp('okTurtles.events/once', NOTIFICATION_TYPE.PUSH_ACTION, async ({ data }) => {
+          const PUBLIC_VAPID_KEY = data
+
+          // 1. Add a new subscription to pushManager using it.
+          const subscription = await registration.pushManager.subscribe({
+            userVisibleOnly: true,
+            applicationServerKey: urlBase64ToUint8Array(PUBLIC_VAPID_KEY)
```
taoeffect: [Here it says](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe#responding_to_user_gestures):

> going forward browsers will explicitly disallow notifications not triggered in response to a user gesture. Firefox is already doing this from version 72, for example.

So... is this the right place to call `subscribe`?

Although... [this example they give is in response to calling `.register`....](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/pushManager) 🤔

frontend/controller/service-worker.js
```diff
@@ -13,23 +16,120 @@ sbp('sbp/selectors/register', {
...
+
+    const pushSubscription = await swRegistration.pushManager.getSubscription()
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+
+    if (pushSubscription) {
+      pubsub.socket.send(createMessage(
+        NOTIFICATION_TYPE.PUSH_ACTION,
+        {
+          action: PUSH_SERVER_ACTION_TYPE.SEND_PUSH_NOTIFICATION,
+          payload: JSON.stringify({ ...payload, endpoint: pushSubscription.endpoint })
```
taoeffect, line 125: I am wondering about this paragraph:

> The endpoint takes the form of a custom URL pointing to a push server, which can be used to send a push message to the particular service worker instance that subscribed to the push service. For this reason, it is a good idea to keep your endpoint a secret, so others do not hijack it and abuse the push functionality.
> 
> — https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/endpoint

I haven't had a chance to study this in depth. Wonder if you or @corrideat have any insights here as far as whether there are any security issues in how we're using this.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -45,3 +49,32 @@ self.addEventListener('message', function (event) {
     console.error('[sw] unexpected data:', event.data)
   }
 })
+
+self.addEventListener('push', function (event) {
+  // PushEvent reference: https://developer.mozilla.org/en-US/docs/Web/API/PushEvent
+
+  if (!(self.Notification && self.Notification.permission === 'granted')) {
+    return
```
taoeffect, line 72: Think it would be a good idea to have `console.debug` here just to note that we got a notification but aren't displaying it.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -45,3 +49,32 @@ self.addEventListener('message', function (event) {
...
+  console.debug('[sw] push received: ', data)
+
+  self.registration.showNotification(
+    data.title,
+    {
+      body: data.body || '',
+      icon: '/assets/images/pwa-icons/group-income-icon-transparent.svg'
+    }
+  )
+})
```
taoeffect, line 85: What is the reason [`showNotification`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification) is being used here instead of `Notification` as used in the [example here](https://developer.mozilla.org/en-US/docs/Web/API/PushEvent#examples)?

frontend/main.js
taoeffect: Probably OK? 🤷 

However this reminds me about this comment you left:

> While reviewing, make sure to unregister the service-worker if there is already one installed from reviewing previously so that it is re-installed again.

Isn't there a way to automatically upgrade the service worker?

taoeffect (changes requested):
Great work @SebinSong! I'd have to say that you're becoming the expert here. :)

So I left we some questions and minor suggestions.

frontend/controller/service-worker.js
corrideat: I'm not entirely familiar with the implications of this URL (would need to see how this protocol works), although it seems, from a general POV, like URLs shouldn't be treated as secrets.

Until I review this, though, this is my preliminary assessment:

(1) Messages should be encrypted and signed. Providing this holds, the security risk from a 'public' URL should be limited to DoS at most.
(2) If the URL needs to be secret, we could derive it from a secret known to the client only (e.g., we could add a 'subscription secret' to the identity contract)

frontend/controller/serviceworkers/sw-primary.js
SebinSong: @taoeffect that was what I found from [this MDN PWA tutorial](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/js13kGames/Re-engageable_Notifications_Push) (down below in `service-worker.js` section).  But looking at the example link you attached, I think we can benefit more functionalities from using `Notification` instead (e.g. attaching the `click` event listener to the browser notification UI etc.) I will replace it with using `Notification` instead.

frontend/controller/service-worker.js
SebinSong: @taoeffect I agree it makes more sense if the push-subscription happens in response to user action, which in our case would be switching on this 'Browser notification' toggle button in the user-settings page.



If it's okay, I can work on this as a follow-up issue of this PR along with a couple of other Todos I have (e.g. re-generating a push-subscription in the front-end if the server fails to send a notification due to the front-end using an old `VAPID_PUBLIC_KEY`). Does it sound good?

taoeffect: > @taoeffect I agree it makes more sense if the push-subscription happens in response to user action, which in our case would be switching on this 'Browser notification' toggle button in the user-settings page.

Well, hold on. I was mainly just curious. Were you able to test this on FF as-is? (I plan to test it myself tomorrow). The thing is, Firefox is way past version 72, and if it works as-is that's something to be aware of it.

Regarding the idea of enabling it in response to something... I think we already prompt the user for permission to display notifications... maybe that's why it's working?

> If it's okay, I can work on this as a follow-up issue of this PR along with a couple of other Todos I have (e.g. re-generating a push-subscription in the front-end if the server fails to send a notification due to the front-end using an old VAPID_PUBLIC_KEY). Does it sound good?

Well, I agree we don't need to change anything yet, and **if** we do change anything, it can be done in a followup. I'm mainly curious / gathering info.

frontend/controller/serviceworkers/sw-primary.js
taoeffect: @SebinSong ok, yeah worth testing it out... let me know if it works that way or not. Maybe it's something special about being inside of a service worker? 🤔

taoeffect (changes requested):
So I tried testing this out today in Firefox Developer Edition using `grunt dev --tunnel` and ran into a few problems:

1. The browser doesn't prompt the user to show notifications. An error is shown in the console instead:

Screenshot 2023-11-30 at 9 03 08 AM

Note: we already had code in the project to properly bring up this prompt in response to certain user actions. Search the project for `requestNotificationPermission()`. Can this code be modified to use use that too? (Or some modified version of it if modifications are necessary?)

2. Even if permissions are given, there are still other errors that occur:

Screenshot 2023-11-30 at 9 07 20 AM

Screenshot 2023-11-30 at 9 07 27 AM

frontend/controller/serviceworkers/sw-primary.js
SebinSong: @taoeffect 
yeah, that's a method available on a service-worker registration instance.
(reference: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification)

SebinSong:
@taoeffect 
Thanks for review.
First of all, I wanted to get a clear evidence that they are firefox-specific issues and here are some proofs of it working on both Chrome desktop/mobile.

---
[ Chrome desktop ]



---

[ Chrome mobile on my android device ]

**1) recognised as 'installable'**




**2) Once the permission is given on the browser, receives the test notification on the device**




**3) Installed on the device and added to the home screen**



---

**[ Replies to your comment ]**

**1.**

> The browser doesn't prompt the user to show notifications. An error is shown in the console instead:

I did get the same error when accessing the app via the `https` url generated from the `grunt dev --tunnel` and it apparently is related to firefox blocking `Notification.requestPermission()` request that is not triggered by a user action, which is what you also mentioned in[ this comment](https://github.com/okTurtles/group-income/pull/1770#discussion_r1409775562) too.

reference: https://stackoverflow.com/questions/61475486/notification-requestpermission-throwing-error-on-mozilla

I have removed the block that asks users for notification permission from inside `sbp('service-workers/setup')` btw, and that will make sure we don't get this error.


**2.**

> "DOMexception: An attempt was made to use an object that is not, or is no longer, usable" error

Apparently that happened because `pubsub.socket.send()` method was called when the web socket instance is not yet in the state of `WebSocket.OPEN`. I made a fix that makes sure it is called after the web socket is opened.

After making fixes for both **1.** & **2.** (also giving the permission inside the app), I was able to get the test notification from the FF developer edition like the screenshot below.



The commit for both fixes is [here](https://github.com/okTurtles/group-income/pull/1770/commits/ac8842fea2ceec165c1cb0ddaf33c519fb522bca)

Thanks,

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,124 @@ sbp('sbp/selectors/register', {
...
         })
-      }, 1000)
+
+        if (pubsub.socket.readyState === WebSocket.OPEN) {
+          getVapidPublicKeyFromServer()
+        } else {
+          pubsub.socket.addEventListener('open', () => {
+            getVapidPublicKeyFromServer()
+          })
+        }
```
taoeffect: I think we were forced to write the code this way because of some quirks of how the browser works. This seems like non-intuitive behavior worth referencing in a comment with a link to more details.

frontend/controller/serviceworkers/sw-primary.js
taoeffect: Can you also log the `event` itself (e.g. move line 61 up to the top and log `data`?)

taoeffect (changes requested):
@SebinSong nice improvements. I'm still running into an issue.

https://github.com/okTurtles/group-income/assets/138706/028f19c6-46b1-4138-8fd6-236ffb97fdf2

This is the popup that should appear but doesn't:

Screenshot 2023-12-01 at 10 56 59 AM

Pointing to code here:

Screenshot 2023-12-01 at 10 56 23 AM

Screenshot 2023-12-01 at 10 56 28 AM

It doesn't because the request for notifications permissions is being done immediately when the page is loaded, instead of in response to a user action (as it is currently done on `master`).

If you can figure out a way to combine the changes here with the way things are done on `master` to request these permissions, then I think it will work better on Firefox.

Also left some comments below:

SebinSong:
@taoeffect 

**1.**

>  request for notifications permissions is being done immediately when the page is loaded, instead of in response to a user action 

This is actually not what the current logic does. if you see [my previous commit](https://github.com/okTurtles/group-income/pull/1770/commits/ac8842fea2ceec165c1cb0ddaf33c519fb522bca), the block that requests the user for a permission has been removed. Plus, as far as I know, it is not possible that "request for notifications permissions is being done immediately" is achieved via Javascript. it only can ask users to give the permission.

**2.**
From what I understood, what Firefox is banning is showing this **permission request popup itself** (triggered by `Notification.requestPermission()`) without it being resulted from some kind of user-action.



Please have a read this article that explains about this error relatively in details
https://pushpad.xyz/blog/the-notification-prompt-can-only-be-triggered-by-a-user-gesture-on-some-browsers

The same article references another article that suggests a solution that works across browsers, which is [use a double opt-in for notifications](https://pushpad.xyz/blog/the-double-opt-in-for-web-push-notifications/) (e.g. something like this modal in below screenshot)




**3.**

> "Uncaught DOMException : User denied permission to use Push API" error

From my investigation, it seems like FF sends this error if [`pushManager.subscribe()`](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe) method is called when no browser-notification permission is given yet.

So, I think the push-subscription logic now should be separated from service-worker registration part(currently it happens immediately followed by successful service-worker registration) and then should happen upon some kind of user-action (which is what is mentioned in **2.**)

---

**Q. So What would you like as the solution here?**
How about once a user signs up, the app displays a double opt-in modal and then it sets up push-subscription if user gives the permission(which is pretty much what [this article](https://pushpad.xyz/blog/the-double-opt-in-for-web-push-notifications/) says)?

taoeffect:
> Q. So What would you like as the solution here?
How about once a user signs up, the app displays a double opt-in modal and then it sets up push-subscription if user gives the permission(which is pretty much what [this article](https://pushpad.xyz/blog/the-double-opt-in-for-web-push-notifications/) says)?

I'm not a fan of "double opt-in", e.g. asking the user if it's OK to show them notifications once, and then a second time, because it feels like pestering users, and it gives more opportunity for them to click "no" - something we don't want them to do because it means they'll miss important notifications related to upcoming payments and that can hurt the group's ability to meet its mincome goals.

So, how about doing it where we're already doing it `master`, e.g. by updating `requestNotificationPermission()`?

SebinSong:
@taoeffect Sounds good, will let you know once the PR is updated with the work.

SebinSong:
@taoeffect Updated the PR with the work for `moving push-subscription setup login into requestNotificationPermission() method`.

**1.**
First of all, as I mentioned in Slack too, below issue happens in the latest `master` too, if it's served via the https URL generated via `grunt dev --tunnel`.



If we want to ask users for the browser notification permission the way we current does (calling `requestNotificationPermission` upon user's signing up or logging in), it seems there is no way to prevent this error from happening.

**2.**
Also, what I've observed is that this error in **1.** sometimes happens but then the other times it doesn't. (which was strange..)

And when that error happens, it leads to an immediate termination of `requestNotificationPermission` without `try ... catch` block catching it, but the browser still shows this pop-up dialog silently...! (which is strange too)



**3.**

When **2.** happens, `granted` / `denied` selection from the user cannot be picked up from `requestNotificationPermission` as it was lead to an immediate termination.
In this case, push-subscription setup logic in `requestNotificationPermission` method is handled the next time this method is executed, if the user has selected `allow` button of the pop-up dialog in **2.** (As you can check inside `requestNotificationPermission`, this method doesn't display the dialog pop-up anymore  once the permission has been granted, only executing the subsequent logics there)

Hope it all makes sense and sounds good.

Thnx,

frontend/model/contracts/shared/nativeNotification.js
```diff
@@ -16,6 +20,17 @@ export async function requestNotificationPermission (force: boolean = false): Pr
...
+
+  if (Notification.permission === 'granted' && notificationPayload) {
+    const isPushReady = await sbp('service-worker/check-push-subscription-ready')
+
+    if (isPushReady) {
+      sbp('service-worker/send-push', notificationPayload)
+    } else {
+      sbp('service-worker/setup-push-subscription', notificationPayload)
+    }
+  }
```
taoeffect: A few concerns here:

1. We need to call `'service-worker/setup-push-subscription'` regardless of whether or not there's a `notificationPayload`, right? But this code doesn't do that.
2. The `notificationPayload` seems a bit out of place to me. The only reason you're using it is so that you can test the notifications in a sort of "debug" mode, but in the final version we're going to remove most of those notifications. It would be better to send them separately if permissions were granted, instead of here.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -45,3 +49,33 @@ self.addEventListener('message', function (event) {
...
+
+self.addEventListener('pushsubscriptionchange', async function (event) {
+  // if a subscription is expired for some reason, re-subscribe it, so it doesn't lead to crashing the server in /push/send route.
+  // NOTE: Currently there is no specific way to validate if a push-subscription is valid. So it has to be handled in the front-end.
+  // (reference:https://pushpad.xyz/blog/web-push-how-to-check-if-a-push-endpoint-is-still-valid)
+  const subscription = await self.registration.pushManger.subscribe(event.oldSubscription.options)
+
+  // send the re-newed subscription details to the server
+  await fetch('/push/subscribe', { method: 'POST', body: JSON.stringify(subscription.toJSON()) })
```
taoeffect: This route no longer exists?

frontend/views/containers/access/LoginForm.vue
```diff
@@ -124,7 +124,13 @@ export default ({
...
         await this.postSubmit()
         this.$emit('submit-succeeded')
-        requestNotificationPermission()
+
+        requestNotificationPermission({
+          notificationPayload: {
+            title: 'Logged in',
+            body: 'Welcome again!'
+          }
+        })
```
taoeffect: Here and below we should separate these out, along with a comment, e.g.:

```js
const granted = requestNotificationPermission() === 'granted'
// TODO: remove in production - this is just for testing the notification
if (granted) { /* post notification */ }
```

Same for the code below in `SignupForm.vue`

taoeffect (changes requested):
Left a few more comments!

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,171 @@ sbp('sbp/selectors/register', {
...
-          key: 'foo',
-          value: 'bar'
-        })
-      }, 1000)
     } catch (e) {
-      // TODO: this
       console.error('error setting up service worker:', e)
     }
+  },
+  'service-worker/setup-push-subscription': async function (followingNotification = null) {
```
taoeffect: 2 questions:

1. Can `'service-worker/setup-push-subscription'` be called safely multiple times? I ask because `requestNotificationPermission`, where it's called, can be called multiple times.
2. `followingNotification` again, should probably be removed. It doesn't seem to be used anyway.

taoeffect (changes requested):
question + minor change request

shared/pubsub.js
```diff
@@ -91,7 +92,15 @@ export const REQUEST_TYPE = Object.freeze({
 
 export const RESPONSE_TYPE = Object.freeze({
   ERROR: 'error',
-  SUCCESS: 'success'
+  SUCCESS: 'success',
+  PUSH_ERROR: 'push-error'
```
snowteamer: Conceptually, push errors are still errors, so why was it necessary to add `PUSH_ERROR` rather than keep using `ERROR`?

shared/pubsub.js
```diff
@@ -91,7 +92,15 @@ export const REQUEST_TYPE = Object.freeze({
...
+})
+
+export const PUSH_SERVER_ACTION_TYPE = Object.freeze({
+  SEND_PUBLIC_KEY: 'send-public-key',
+  STORE_SUBSCRIPTION: 'store-subscription',
+  DELETE_SUBSCRIPTION: 'delete-subscription',
+  SEND_PUSH_NOTIFICATION: 'send-push-notification'
```
snowteamer, line 102: Since these message types are used to make the server do something, then maybe they could have been added as keys into the `REQUEST` type (preferably with a `PUSH_` prefix)?

frontend/controller/service-worker.js
```diff
@@ -1,6 +1,8 @@
 'use strict'
 
 import sbp from '@sbp/sbp'
+import { PUBSUB_INSTANCE } from '@controller/instance-keys.js'
+import { NOTIFICATION_TYPE, PUSH_SERVER_ACTION_TYPE, createMessage } from '~/shared/pubsub.js'
 
 // NOTE: this file is currently unused. I messed around with it just enough
```
snowteamer: Shouldn't this comment be removed?

frontend/controller/service-worker.js
```diff
@@ -13,23 +15,171 @@ sbp('sbp/selectors/register', {
...
+      return
+    }
+
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+    const existingSubscription = await registration.pushManager.getSubscription()
+
+    if (existingSubscription) {
+      // If there is an existing subscription, no need to create a new one.
+      // But make sure server knows the subscription details too.
+      pubsub.socket.send(createMessage(
```
snowteamer, line 59: Note that low-level `.socket.send()` calls do not benefit from automatic retriability upon reconnection. More specifically, the message won't be sent if `.socket.readyState !== WebSocket.OPEN`.
Not sure whether that's an actual problem though, I don't know enough about service workers.

snowteamer (commented):
Looks great so far! Just left a few pubsub-related comments

frontend/main.js
SebinSong: @taoeffect 
If you look at [this article](https://web.dev/articles/service-worker-lifecycle#updates) written about service-worker's lifecycle, the service-worker update is detected and executed automatically by the browser when the user visits the website. 

Also, there seems to be a method that we can use to manually do that when we want, which is [here](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update).

(Sorry for a late reply, needed to take some time for researching on this topic.)

taoeffect: It looks like there are a few APIs we might want to consider:

> You can take control of uncontrolled clients by calling `clients.claim()` within your service worker once it's activated.

> The waiting phase means you're only running one version of your site at once, but if you don't need that feature, you can make your new service worker activate sooner by calling `self.skipWaiting()`.

> If you expect the user to be using your site for a long time without reloading, you may want to call `update()` on an interval (such as hourly).

Perhaps for a subsequent PR

taoeffect: Created and assigned a new issue for that: https://github.com/okTurtles/group-income/issues/1791

frontend/controller/service-worker.js
SebinSong: @taoeffect 
1. Yes, `service-worker/setup-push-subscription` only takes the full steps of generating a new push subscription and sending/storing it to the server when there is no subscription created yet. For any subsequent calls, it only sends the existing subscription to the server to make sure the server keeps the subscription up-to-date on its side.

2. no prob.

shared/pubsub.js
SebinSong: @snowteamer Good point.

shared/pubsub.js
SebinSong: Right, but that would require `server.js` to have the corresponding message handlers for all of `REQUEST_TYPE.PUSH_` keys. Instead, I think it's cleaner for it to have one message handler for `REQUEST_TYPE.PUSH_ACTION` and then let this handler do appropriate jobs depending on which action it is (passed as `actionType` field in the payload)

Besides that, why not using these keys as `PUSH_ACTION.SOMETHING` instead of `PUSH_SOMETHING` if we want to make it clear they are push-notification related? :)

Btw,

> Since these message types are used to make the server do something, then maybe they could have been added as keys into the REQUEST type

This made me realise that it makes more sense to put `PUSH_ACTION` under `REQUEST_TYPE` instead of `NOTIFICATION_TYPE` which is where it currently is, so I moved it there.

frontend/controller/service-worker.js
SebinSong: Good catch

frontend/controller/service-worker.js
SebinSong: @snowteamer 
Currently where this sbp selector(`service-worker/setup-push-subscription`) is called is righter after the user's signup/login. So at least, we can guarantee this won't be executed when the `socket.readyState` is still `=== WebSocket.CONNECTING`.

But,

>  benefit from automatic retriability upon reconnection

sounds like an important stuff to have. If you give some explanations on what the **"non"** low-level usage of the pubsub system would be, I can make the modification in the follow-up PR while working on [this issue](https://github.com/okTurtles/group-income/issues/1791).

SebinSong:
@taoeffect 
Updated the PR again with yours and @snowteamer 's feedbacks.

backend/pubsub.js
```diff
@@ -20,13 +20,13 @@ import type {
...
 
-import type { JSONType } from '~/shared/types.js'
+import type { JSONType, JSONObject } from '~/shared/types.js'
 
 const { bold } = require('chalk')
 const WebSocket = require('ws')
 
 const { PING, PONG, PUB, SUB, UNSUB } = NOTIFICATION_TYPE
-const { ERROR, SUCCESS, PUSH_ERROR } = RESPONSE_TYPE
+const { ERROR, SUCCESS } = RESPONSE_TYPE
```
taoeffect, line 29: Just something to keep an eye on between your PR @SebinSong and @snowteamer's recent modifications to the pubsub in #1792 (e.g. `SUCCESS` => `OK`). If this PR is merged first @snowteamer will need to update this code as well.

SebinSong: @taoeffect This PR doesn't contain any change that uses `RESPONSE_TYPE.SUCCESS`, so it's safe.

taoeffect (approved):
Approved! Great work @SebinSong!!

frontend/controller/service-worker.js
SebinSong: @snowteamer A reminder for you for a potentially important question I left to you above, if this has gone unnoticed.

snowteamer: @SebinSong 
>  So at least, we can guarantee this won't be executed when the socket.readyState is still === WebSocket.CONNECTING

Sounds fine to me - maybe we need to check for `WebSocket.CONNECTING`?
Also do we need to call that selector again upon reconnection after a temporary disconnection?

> what the "non" low-level usage of the pubsub system would be

That would be using client methods like `sub()` or `pub()` instead of socket methods, but not sure whether there is a currently fitting one for this use case

SebinSong: @snowteamer Thanks for your opinion. I think the entire service-worker related workflow added by this PR needs to be re-written to use those `pub()`, `sub()` method in that case. I will work on that in some following weeks. (cc. @taoeffect)

snowteamer: @SebinSong Sorry I meant maybe checking for `WebSocket.CLOSING`, in case the connection would drop just after login.

MDN says:
>If you call send() when the connection is in the CLOSING or CLOSED states, the browser will silently discard the data.

If this ever happens we likely want to re-send the message later

PR #1772, opened 2023-10-18
#1761 - [Bug] Dismissing a payment and resending it doesn't clear it as being sent

### LINKED ISSUE #1761: [Bug] Dismissing a payment and resending it doesn't clear it as being sent
### Problem

Using container tabs add-on in Firefox Dev Edition:

1. [Tab1] u1 creates group w/$1k mincome and pledges $100
2. [Tab2] u2 joins group and has $800 income
3. [Tab1] In console: `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })`
4. [Tab1] u1 clicks "TODO" tab and sends payment to u2
5. [Tab2] u2 clicks "I did not receive this" in their received payments
6. [Tab1] u1 sends the payment again

After sending the payment a second time, it is not dismissed from the TODO, but remains there:

Screenshot 2023-10-11 at 9 21 16 AM

### Solution

Figure out what the problem is and fix it.

Users should be able to continue this sequence: saying they didn't receive it, resending it, saying they didn't receive it, resending it, etc. for as long as they want. But each time the payment is sent it should be placed back under the sent payments and removed from the TODOs of u1.

SebinSong:
closes #1761 

Currently `RecordPayment.vue` always creates new payment item regardless of whether it's an existing payment item with the status as `PAYMENT_NOT_RECEIVED` or a todo payment that hasn't been payed yet. Apparently this leads to duplicated payment items here and there in the UI. So I made a fix where the component checks if the payment item is `PAYMENT_NOT_RECEIVED` or not and then prevents it from creating a new payment item in group state if it is.
Hope it sounds good.

### DIFF
--- a/frontend/views/containers/payments/RecordPayment.vue
+++ b/frontend/views/containers/payments/RecordPayment.vue
@@ -136,11 +136,11 @@ export default ({
               hash: item.hash,
               data: item.data,
               meta: item.meta,
-              username: item.toUser,
-              displayName: this.userDisplayName(item.toUser),
+              username: item.data.toUser,
+              displayName: this.userDisplayName(item.data.toUser),
               date: item.meta.createdDate,
-              monthstamp: dateToMonthstamp(item.createdDate),
-              amount: item.amount
+              monthstamp: dateToMonthstamp(item.meta.createdDate),
+              amount: item.data.amount
             }
           : item
       })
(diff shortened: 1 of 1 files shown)


frontend/views/containers/payments/RecordPayment.vue
```diff
@@ -131,16 +131,16 @@ export default ({
     ]),
     paymentsList () {
       return this.todoItems.map(item => {
-        return item.status === PAYMENT_NOT_RECEIVED // if not received item, re-format the obj
```
SebinSong, line 134: Turns out it had always been resolved as `false` here, even when it's actually in the `PAYMENT_NOT_RECEIVED'`status. So made a correction to this logic as well while working on the task.

taoeffect (approved):
Great work @SebinSong!

Looks like it's working and addressed the issue! 🙂 👍

PR #1773, opened 2023-10-18
Implement a way to set default groupId

### LINKED ISSUE #1741: Set priority of groupId
### Problem

Since the process to join group is updated, user could have two types of groups (in their `state.contracts`).
One is the group which user has already successfully joined, and the other is the group which user is waiting the keys to join.
If the users try to login in a new browser or kicked from the current group by someone else, they should be redirected to anther group. So which group should be selected?
If a user has these two types of groups, the one of fully-joined groups should be set as the next `currentGroupId`.
But this logic is not implemented in the current code base.

### Solution

Implement the logic in the function `'gi.actions/identity/updateLoginStateUponLogin'`.
Especially, inside the following `if` statement.
```
if (!state.currentGroupId) {
  ...
}
```
This logic should also be implemented in the `sideEffect` of `'gi.contracts/group/removeMember'`.

Silver-IT:
### Summary of Changes

- Closes #1741


frontend/controller/actions/identity.js
```diff
@@ -326,7 +326,9 @@ export default (sbp('sbp/selectors/register', {
       // NOTE: users could notice that they leave the group by someone else when they log in
       if (!state.currentGroupId) {
         const { contracts } = state
-        const gId = Object.keys(contracts).find(cID => contracts[cID].type === 'gi.contracts/group')
+        const gId = Object.keys(contracts)
```
taoeffect, line 325: Since it's unclear what line 331 is doing, could you please add this comment above line 329?

```js
// grab the groupID of any group that we've successfully finished joining
```

frontend/model/contracts/group.js
```diff
@@ -918,8 +918,8 @@ sbp('chelonia/defineContract', {
           await sbp('gi.contracts/group/removeArchivedPayments', contractID)
 
           const groupIdToSwitch = Object.keys(contracts)
-            .find(cID => contracts[cID].type === 'gi.contracts/group' &&
-              cID !== contractID && rootState[cID].settings) || null
+            .filter(cID => contracts[cID].type === 'gi.contracts/group' && cID !== contractID)
+            .sort((cID1, cID2) => rootState[cID1].profiles?.[username] ? -1 : 1)[0] || null
```
taoeffect, line 923: Same here, above line 920.

```js
// grab the groupID of any group that we've successfully finished joining
```


Also, you do not need to add `|| null` at the end, `groupIdToSwitch` will be set to `undefined` if it's not there.

taoeffect (changes requested):
Reviewed!

frontend/model/contracts/group.js
Silver-IT: > Also, you do not need to add || null at the end, groupIdToSwitch will be set to undefined if it's not there.

Yeah, it'll be set to `undefined` if it's not there. But I can see that the type of `currentGroupId` is `string | null`. `null` value is used when there is no group as we can see in the `frontend/model/state.js`

taoeffect: > null value is used when there is no group as we can see in the frontend/model/state.js

Hmm, OK, nice catch 👍

package.json
```diff
@@ -1,183 +0,0 @@
-{
```
taoeffect: It seems for some reason this whole file has been deleted, can you please restore it?

(I've had this happen to me once too, I'm not sure what causes it)

taoeffect (changes requested):
Looking good @Silver-IT! Just one git-related issue:

taoeffect (approved):
LGTM! Nicely done!

PR #1775, opened 2023-10-19
Disable to change group settings unless group creator tries

### LINKED ISSUE #1707: Should block updating group setting unless group creator tries
### Problem

In the current version, general group members can open modal using the URL (e.g, http://localhost:8000/app/dashboard?modal=MincomeProposal), and update group settings when the group has less than 3 members.

### Solution

- Close the modal
- Disable the submission button
Among the two options, select one and make it work in the following modals (Mincome.vue, ChangeVotingRules.vue, DistributionDate.vue).

Silver-IT:
### Summary of Changes

- Closes #1707

### DIFF
--- a/frontend/views/containers/proposals/DistributionDate.vue
+++ b/frontend/views/containers/proposals/DistributionDate.vue
@@ -5,7 +5,6 @@
     :disabled='$v.form.$invalid || distributionStarted'
     :maxSteps='config.steps.length'
     :currentStep.sync='ephemeral.currentStep'
-    variant='changeDistributionDate'
     :shouldImmediateChange='shouldImmediateChangeDistributionDate'
     @submit='submit'
   )
--- a/frontend/views/containers/proposals/ProposalTemplate.vue
+++ b/frontend/views/containers/proposals/ProposalTemplate.vue
@@ -46,7 +46,7 @@
           :class='submitStyleNonProposal'
           v-if='!shouldPropose'
           @click='submit'
-          :disabled='disabled'
+          :disabled='disabled || !isGroupCreator'
           data-test='submitBtn'
         ) {{ submitTextNonProposal }}
 
@@ -118,7 +118,7 @@ export default ({
     },
     variant: {
       validator (value) {
-        return ['addMember', 'removeMember', 'changeDistributionDate'].indexOf(value) > -1
+        return ['addMember', 'removeMember'].indexOf(value) > -1
       }
     },
     shouldImmediateChange: Boolean
@@ -136,6 +136,9 @@ export default ({
     },
     proposalSettings () {
       return this.groupProposalSettings()
+    },
+    isGroupCreator () {
+      return this.ourUsername === this.groupSettings.groupCreator
     },
     isNextStep () {
       return this.currentStep <= this.maxSteps - 1
--- a/package-lock.json
+++ b/package-lock.json
@@ -5072,9 +5072,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001468",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001468.tgz",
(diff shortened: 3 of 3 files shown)


frontend/views/containers/proposals/DistributionDate.vue
```diff
@@ -5,7 +5,6 @@
     :disabled='$v.form.$invalid || distributionStarted'
     :maxSteps='config.steps.length'
     :currentStep.sync='ephemeral.currentStep'
-    variant='changeDistributionDate'
```
taoeffect, line 8: Why was this removed?

frontend/views/containers/proposals/ProposalTemplate.vue
```diff
@@ -46,7 +46,7 @@
           :class='submitStyleNonProposal'
           v-if='!shouldPropose'
           @click='submit'
-          :disabled='disabled'
+          :disabled='disabled || !isGroupCreator'
```
taoeffect, line 49: It's not clear to me why this condition is being added. Anyone should be able to create a proposal.

taoeffect (commented):
Some questions.

This issue is also very low priority.

frontend/views/containers/proposals/DistributionDate.vue
Silver-IT: It's not necessary. Nowhere using it.

frontend/views/containers/proposals/ProposalTemplate.vue
Silver-IT:
```
button-submit(
  key='change'
  :class='submitStyleNonProposal'
  v-if='!shouldPropose'
  @click='submit'
  :disabled='disabled || !isGroupCreator'
  data-test='submitBtn'
) {{ submitTextNonProposal }}
```

As the button title says, it's submit button without making proposal. This button is visible when there are less than 3 members in the group. So if he is not group creator, he shouldn't make group settings just like `mincome`, `distributionDate`, `votingRule`, ...
`Add Member`, and `Remove Member` have different user flow so we don't need to take care of them.

SebinSong:
@Silver-IT 
I agree that the right to make instant updates to those group-settings when the members are < 3 should only be given to the creator.
But it also occurred to me that it would be a better fix if the app somehow lets the non-creator users why they cannot make that action, instead of just disabling the submit button.

For example,
Either by showing a banner in the proposal modal saying only the group creator can make the instant update Or,
by showing a prompt pop-up(`sbp('gi.ui/prompt', ...)`) saying the same when the call-to-action is clicked.

What's your thought?

frontend/views/containers/proposals/ProposalTemplate.vue
taoeffect: > This button is visible when there are less than 3 members in the group

Where is this check done? `:disabled` is being set to `true` regardless of the value of `disabled` parameter, because of the `||`. So it will always be disabled if we are not the group creator.

SebinSong: @taoeffect 
This `v-if` line in the above does the check there. 
```js
button-submit(
  ...
  v-if='!shouldPropose'
  ...
)
```

Silver-IT:
> Either by showing a banner in the proposal modal saying only the group creator can make the instant update Or,
by showing a prompt pop-up(sbp('gi.ui/prompt', ...)) saying the same when the call-to-action is clicked. What's your thought?

@SebinSong. Yeah, that's good idea. However, with the normal userflow, non-creators can not access this modal. Please check `proposalOptions` function in `ProposalsWidget.vue`. What I did is to disable to change group settings when non-creator users would access modal by opening the specific URL. (Check Issue #1707) So I think we could ignore it. If most of the teammates, want to add it, I can add it for sure. However, thanks for your good opinion.

taoeffect (approved):
Nice - thanks for the explanations @Silver-IT! Tested & approved!

PR #1776, opened 2023-10-19
E2e protocol ricardo

corrideat:


frontend/controller/actions/identity.js
```diff
@@ -469,7 +428,51 @@ export default (sbp('sbp/selectors/register', {
...
+            // accepted an invite add added our profile)
+              if (state[cId]?.profiles?.[username]) {
+                sbp('gi.actions/group/updateLastLoggedIn', { contractID: cId }).catch(console.error)
+              }
+            })
+        }).catch((err) => {
+          console.error('Error during contract sync upon login', err)
+        }).finally(() => {
+          sbp('okTurtles.events/emit', LOGIN, { username, identityContractID })
+        })
```
taoeffect: This is different from what was there before, and therefore changes the behavior of the UI from how it worked before, possibly breaking the tests because it claims we're logged in when we're not.

frontend/model/contracts/group.js
```diff
@@ -1040,6 +1040,7 @@ sbp('chelonia/defineContract', {
         // TODO: per #257 this will ,have to be encompassed in a recoverable transaction
         // however per #610 that might be handled in handleEvent (?), or per #356 might not be needed
         if (meta.username === loggedIn.username) {
+          if (!sbp('okTurtles.data/get', 'JOINING_GROUP-' + contractID)) return
```
taoeffect: I could add a comment above this line explaining this check

taoeffect (approved):
LGTM!

PR #1777, opened 2023-10-20
#1774, #1771 - Bug-fixes in Payment.vue

### LINKED ISSUE #1771: [Bug] Check-box looks incorrect in payment-list table
### Problem

The checkbox for selecting all table items in the `Todo` tabs of `Payment.vue` doesn't look correct. The check icon should be centered, just like another checkbox below it.

### Solution
Figure out and fix.

### LINKED ISSUE #1774: Payment stats should be updated when payment is marked as not being received
### Problem

This is a followup to issue #1761 - please see that issue for details on steps to reproduce.

Basically, do #1761, and notice that when a payment is marked as being not received, the payment stats are not updated:

Screenshot 2023-10-18 at 11 49 56 AM

### Solution

Review and update any code as needed to consider these payments as not being completed, so that the payment stats are updated accordingly (a payment should not be marked as "1 of 1 received" if it was marked as not received, etc.)

SebinSong:
closes #1771 
closes #1774

### DIFF
--- a/frontend/views/components/graphs/Progress.vue
+++ b/frontend/views/components/graphs/Progress.vue
@@ -33,11 +33,12 @@ export default ({
   },
   computed: {
     percent () {
-      return `${100 * this.value / this.max}%`
+      return !this.max ? '0%' : `${100 * this.value / this.max}%`
     },
     percentSoft () {
       if (!this.secValue) return false
-      return `${100 * this.secValue / this.max}%`
+
+      return !this.max ? '0%' : `${100 * this.secValue / this.max}%`
     },
     marksStyle () {
       const color = this.percent === '100%'
--- a/frontend/views/components/graphs/Progress.vue
+++ b/frontend/views/components/graphs/Progress.vue
@@ -33,7 +33,9 @@ export default ({
   },
   computed: {
     percent () {
-      return !this.max ? '0%' : `${100 * this.value / this.max}%`
+      return !this.max
+        ? '0%' // When this.max 0, the calculation below becomes 'NaN%', so manually specifying 0% here.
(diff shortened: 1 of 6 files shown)


frontend/views/components/graphs/Progress.vue
```diff
@@ -33,11 +33,12 @@ export default ({
   },
   computed: {
     percent () {
-      return `${100 * this.value / this.max}%`
+      return !this.max ? '0%' : `${100 * this.value / this.max}%`
```
SebinSong: When `max` here is passed as `0`, the percent value here becomes `NaN%` and leads to a UI bug. This is a fix for that issue.

taoeffect: @SebinSong this would've been a great comment to add to the code... (￣ー￣；

taoeffect (changes requested):
Looking really good @SebinSong!

Two requests:

1. See comment about adding a comment inline in the code instead of on Github
2. I noticed that when the payments are reset because a payment was marked as not complete, the green checkboxes are not cleared:

Screenshot 2023-10-20 at 8 58 24 AM

Vs how they started out:

Screenshot 2023-10-20 at 8 58 55 AM

SebinSong:
@taoeffect Good catch on that one.
To fix that issue (**2.**), I had to revert some changes I've done previously and optimise them. Also, while working on it I realised the TODO history tab in the `/dashboard` page doesn't update accordingly for payment being marked as not-received. So made a fix there too.

taoeffect (approved):
Beautiful! Works perfectly! 😄  🎉

PR #1780, opened 2023-10-27
E2e protocol ricardo 20231025 login sync

corrideat:


Silver-IT:
Having checked this PR mainly focusing on the user flow, I could see several issues.

**Issue 1**
- Steps to reproduce
  - [Tab1] user1 signs up, and creates a group
  - [Tab2] user2 joins the group, and closes the tab
  - [Tab1] user1 kicks the u2 from the group
  - [Tab2] Opens 'http://localhost:8000'. Got 5 errors in the console
  - [Tab1] Got error in console too.
- Screenshot
  - Error in Tab1
image
  - 5 Errors in Tab2
image
image
image
image
- What should happen
  - No errors should be happening. And the important thing is that there MUST be no errors in Tab1.
  - `gi.actions/group/join` shouldn't be called. But it is called twice.
- DeadLock Output
  - [Tab1] {}
  - [Tab2] {}

Silver-IT:
**Issue 2**
- Steps to reproduce
  - [Tab1] user1 signs up, and creates a group
  - [Tab2] user2 joins the group, and closes the tab
  - [Tab1] user1 kicks the u2 from the group
  - [Tab3] user2 logs in, and got 5 errors.
- Screenshot
  - 5 Errors in Tab2
image
image
image
image
image
- What should happen
  - No errors should be happening.
  - `gi.actions/group/join` should be called ONLY once. But it is called twice.
- DeadLock Output
  - [Tab1] {}
  - [Tab2] {}

Silver-IT:
**Issue 3**
- Steps to reproduce
  - [Tab1] user1 signs up, and creates a group.
  - [Tab2] user2 joins the group.
  - [Tab1] user1 kicks the u2 from the group. Got 1 error in console.
  - [Tab2] user2 joins the group again with the initial invite link.
  - [Tab2] Wrong work flow to join. 6 Errors in console. Seems there are many redundant function calls.
  - [Tab1] Many logs in console than expected. Seems there are too many function calls. And duplicated notifications.
- Screenshot
  - Error in Tab1 (Step 3)
image
  - Errors in Tab2 (Step 5)
image
image
  - Duplicated notifications in Tab1 (Step 6)
image
- What should happen
  - No errors should be happening.
  - No duplicated notifications in Tab1
  - `gi.actions/group/join` should be called ONLY once. But it is called 4 times.
- DeadLock Output
  - [Tab1]
image
  - [Tab2] {}

Silver-IT:
**Issue 4**
- Steps to reproduce
  - [Tab1] user1 signs up, and creates a group.
  - [Tab2] user2 joins the group.
  - [Tab3] user3 joins the group, and closes the tab.
  - [Tab2] user2 creates a proposal to remove user3 from the group.
  - [Tab1] user1 approves the proposal, and kicks user3 from the group. Got an error in the notifications.
  - [Tab4] user3 tries to log in, and 8 errors BANG.
- Screenshot
  - Error in Tab1 (Step 5)
(user1 = alexjin, user2 = greg, user3 = andrea)
image
  - Errors in Tab4 (Step 6)
image
image
image
image
image
image
image
image
- What should happen
  - No errors should be happening.
  - `gi.actions/group/join` should be called ONLY once. But it is called twice in Tab4 (Step 6).
- DeadLock Output
  - [Tab1]
image
  - [Tab2] {}
  - [Tab4] {}

Silver-IT:
**Issue 5**
- Steps to reproduce
  - [Tab1] user1 signs up, and creates a group.
  - [Tab2] user2 joins the group.
  - [Tab3] user1 logs in.
  - [Tab4] user2 logs in.
  - [Tab1] user1 kicks user2 from the group.
  - [Tab1], [Tab2], [Tab4] Errors BANG.
- Screenshot
  - Error2 in Tab1 (Step 6)
(user1 = alexjin, user2 = greg)
image
  - Errors in Tab2 (Step 6)
image
  - Errors in Tab4 (Step 6)  
image
- What should happen
  - No errors should be happening.
- DeadLock Output
  - [Tab1] {}
  - [Tab2] {}
  - [Tab3] {}
  - [Tab4] {}

corrideat:
Thanks again, @Silver-IT!

As discussed:

1. I've made side-effects mostly work as they did earlier
2. I've added a check on `syncContractAndWatchKeys` to avoid syncing contracts that have been subscribed to

In addition, I've implemented fixes for all of the issues reported here, except issue 3. I'll continue working on this until issue 3 is resolved.

Regarding issue 1, there is the double sync, that I've tracked down to the login functionality. I'll work on addressing this as well (but it seems to be the case that implementing the `join|leaveGroup` actions would help solving it too).

@taoeffect: You'll probably be pleased to see that I've mostly removed `sync` from `recreateEvent`, something you weren't entirely happy with. There are some vestiges of it that remain, because _sometimes_ the `pubsub` functions don't work as expected and some events are never received. Once this is addressed in `pubsub`, the `syncContract` added to `publishEvent` can be removed.

corrideat:
@Silver-IT I've made another commit addressing issue 3. The solution so far is using `sbp('okTurtles.data/set', 'JOINING_CHATROOM-' + params.contractID, ?)` to avoid removing the chatroom contract when joining.

Besides that, two issues remain.

One is at _leaving_; specifically, when someone else leaves a chatroom we have left:

To write to a contract, we need to be subscribed to it. This is done in `encryptedAction`, but for contracts that call `/remove` in a side-effect, this is essentially pointless. This is an error.

The second issue is at _joining_. You'll see that the `/join` action calls `/sync`. This is 'unnecessary', as `encryptedAction` already does this. However, similarly to what happened above, the first `/sync` will be pointless because of the side-effect removing the contract, requiring a second call to sync that sets the `JOINING_CHATROOM-${cID}` key.

I think that we need to either adapt `encryptedAction` to take pre- and post-hooks (we could call them, e.g., prologue and epilogue) to handle this, or somehow re-think how `/remove` gets called (I think that one thing that would work is doing reference counting, but there may be other options too). These are not mutually exclusive, so we could very well do both.

Related to this, there is a third issue, which is a variant of issue 3:

* rejoin fails 
  - [tab1] u1: create group
  - [tab2] u2: join using invite link
  - [tab2] u2: close tab / navigate away
  - [tab1] u1: remove u2
  - [tab2] u2: open tab with invite link

To recap, apart from the issues mentioned, which are related to removing contracts, there is the double-sync issue I mentioned earlier.

Although double-syncs are a concern, I'll focus now on trying to address the actual issues that happen from missing contract states, which are errors.

Do you have a suggestion on how we could call `/remove` in an elegant way?

frontend/model/contracts/group.js
```diff
@@ -938,101 +985,17 @@ sbp('chelonia/defineContract', {
...
+        Object.keys(state.chatRooms).forEach((chatroomID) => {
+          removeGroupChatroomProfile(state, chatroomID, data.member)
+        })
+      },
+      sideEffect ({ data, meta, contractID }, { state, getters }) {
+        // Put this invocation at the end of a sync to ensure that leaving and
+        // re-joining works
+        sbp('chelonia/queueInvocation', contractID, () => sbp('gi.contracts/group/leaveGroup', { data, meta, contractID }, { state, getters })).catch(e => {
+          console.error(`[gi.contracts/group/removeMember/sideEffect] Error ${e.name} during queueInvocation for ${contractID}`, e)
         })
```
Silver-IT: We don't need to call `gi.contracts/group/leaveGroup` when the member has joined the group after this action.
So it would be necessary to check the group contract profile state in this invocation.
Please reference [my PR](https://github.com/okTurtles/group-income/pull/1728) for this.

```
sideEffect ({ data, meta, contractID }, { state, getters }) {
  const contracts = rootState.contracts || {}
  const { username } = rootState.loggedIn

  if (data.member === username) {
    // If this member is re-joining the group, ignore the rest
    // so the member doesn't remove themself again.
      if (rootState[contractID].profiles?.[username]?.status === PROFILE_STATUS.REMOVED) {
        // NOTE: should remove archived data from IndexedStorage
        //       regarding the current group (proposals, payments)
        await sbp('gi.contracts/group/removeArchivedProposals', contractID)
        await sbp('gi.contracts/group/removeArchivedPayments', contractID)
```

frontend/model/contracts/group.js
```diff
@@ -261,38 +261,79 @@ function updateGroupStreaks ({ state, getters }) {
...
-      signingKeyId: sbp('chelonia/contract/currentKeyIdByName', state, 'csk'),
-      innerSigningContractID: null
-    }),
+    ...extraParams,
     hooks: {
       preSendCheck: (_, state) => {
         // Avoid sending a duplicate action if the person has already
         // left or been removed from the chatroom
         return !!state?.users?.[member]
       }
```
Silver-IT, line 324: Inside `group` contract, it doesn't make sense to decide the next operation depending on the `chatroom` contract state.

Silver-IT (changes requested):
Nice work, @corrideat. I did a very partial review, and left a just two comments.
I can see many updates in contract parts (model/contracts, controller/actions) but I don't think most of them are needed. I have almost finished my PR #1728 and I want you to have even a brief look at it. I fixed most of the login issues with the smaller changes than this.

Also will check the 5 Issues (described above) again, and left comments here.

Silver-IT:
**Issue 1** (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784531132) This issue is not fixed yet.
image
This screenshot shows the console logs filtered by `gi.actions/group` on Tab2 in Step 5 .

**Problem: three errors and 2 redundant `gi.actions/group/join` function calls**

Silver-IT:
**Issue 2** (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784557201) This issue is not fixed yet.
image
This screenshot shows the console logs filtered by `gi.actions/group/join` on Tab3 in Step 4.

**Problem: 1 redundant `gi.actions/group/join` function call.**

Silver-IT:
Issue 3 (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784603749) seems not to be fixed yet.
image
image
image

**Problem: two errors in console**

shared/domains/chelonia/GIMessage.js
```diff
@@ -358,25 +358,24 @@ export class GIMessage {
...
-          return undefined
+        } else {
+          return data.data
         }
-      } else {
-        return data.data
       }
+    } catch {
+      // Signature or encryption error
+      return undefined
```
taoeffect: No `console.warn` is needed here?

shared/domains/chelonia/GIMessage.js
```diff
@@ -407,7 +406,7 @@ export class GIMessage {
     return `${desc}|${this.hash()} of ${this.contractID()}>`
   }
 
-  isFirstMessage (): boolean { return !this.head().previousHEAD }
+  isFirstMessage (): boolean { return !this.head().contractID }
```
taoeffect, line 411: Why was this changed?

frontend/views/pages/PendingApproval.vue
```diff
@@ -36,7 +37,7 @@ export default ({
     ...mapState(['currentGroupId']),
     ourGroupProfile () {
       if (!this.ephemeral.groupIdWhenMounted) return
-      return this.$store.state[this.ephemeral.groupIdWhenMounted]?.profiles?.[this.ourUsername]
+      return this.$store.state[this.ephemeral.groupIdWhenMounted]?.profiles?.[this.ourUsername]?.status === PROFILE_STATUS.ACTIVE
```
taoeffect, line 40: This change changes the meaning of this computed property.

The computed property is named `ourGroupProfile`, which implies it returns our group profile. But now this no longer happens, instead it returns a boolean.

So either the condition needs to be removed or the name of the computed property needs to be changed.

taoeffect (changes requested):
Preliminary review

frontend/common/errors.js
```diff
@@ -23,3 +23,14 @@ export class GIErrorUIRuntimeError extends Error {
...
+export class GIErrorMissingSigningKeyError extends Error {
+  constructor (...params: any[]) {
+    super(...params)
+    // this.name = this.constructor.name
+    this.name = 'GIErrorMissingSigningKeyError' // string literal so minifier doesn't overwrite
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor)
+    }
+  }
+}
```
taoeffect: I think this is a good time to copy over that utility function from Chelonia to define these using 1-liners...

shared/domains/chelonia/GIMessage.js
corrideat: Because I removed a bunch of the repetitive code in `chelonia.js` that was looking up the previous HEAD, which is now done by `publishEvent`. Since the previous HEAD could now momentarily be null, the check as it was would fail. Note that either condition is correct (if the previous HEAD were set).

shared/domains/chelonia/GIMessage.js
corrideat: This is doing the same as was it was doing earlier, more or less. There could be a `console.warn`, but that is already done elsewhere and it feels repetitive to warn multiple times.

frontend/views/pages/PendingApproval.vue
corrideat: Agreed.

frontend/common/errors.js
corrideat: Agreed.

shared/domains/chelonia/internals.js
```diff
@@ -292,14 +318,12 @@ export default (sbp('sbp/selectors/register', {
...
-          if (!newEntry) {
-            return
-          }
-          entry = newEntry
+
+        // TODO: The [pubsub] code seems to miss events that happened between
+        // a call to sync and the subscription time. This is a temporary measure
+        // to handle this until [pubsub] is updated.
+        if (entry.height() === lastAttemptedHeight) {
+          await sbp('okTurtles.eventQueue/queueEvent', entry.contractID(), ['chelonia/private/in/syncContract', entry.contractID()])
```
taoeffect: So I agree that we should address this in this PR... one way to go about it is to make sure the pubsub is connected before `'chelonia/contract/sync'` does the sync.

We might also want to verify that we are in fact connected to the pubsub server before allowing messages to be sent.

frontend/model/contracts/group.js
```diff
@@ -261,38 +261,79 @@ function updateGroupStreaks ({ state, getters }) {
...
+  console.log('jjjj removeGroupChatroomProfile', JSON.parse(JSON.stringify({ state, chatRoomID, member })))
+  Vue.set(state.chatRooms[chatRoomID], 'users',
+    Object.fromEntries(
+      Object.entries(state.chatRooms[chatRoomID].users)
+        .map(([memberKey, profile]) => {
+          console.log('jjjj removeGroupChatroomProfile ii', {
+            memberKey, member, profile, active: (profile: any)?.status === PROFILE_STATUS.ACTIVE
+          })
+          if (memberKey === member && (profile: any)?.status === PROFILE_STATUS.ACTIVE) {
+            return [memberKey, { status: PROFILE_STATUS.REMOVED }]
```
taoeffect: This line seems to be removing many profile fields... that seems wrong.

frontend/model/contracts/group.js
```diff
@@ -261,38 +261,79 @@ function updateGroupStreaks ({ state, getters }) {
...
+  // When a group is being left, we want to also leave chatrooms,
+  // including private chatrooms. Since the user issuing the action
+  // may not be a member of the chatroom, we use the group's CSK
+  // unconditionally in this situation, which should be a key in the
+  // chatroom (either the CSK or the groupKey)
+  if (leavingGroup) {
+    const signingKeyId = sbp('chelonia/contract/currentKeyIdByName', state, 'csk', true)
+
+    // If we don't have a CSK, it is because we've already been removed.
+    // Proceeding would cause an error
```
taoeffect, line 303: These are good comments 👍

frontend/model/contracts/group.js
```diff
@@ -261,38 +261,79 @@ function updateGroupStreaks ({ state, getters }) {
...
+    if (!signingKeyId) {
+      return
+    }
+
+    extraParams.signingKeyId = signingKeyId
+    extraParams.innerSigningContractID = null
```
taoeffect, line 312: Does `innerSigningContractID = null` need to be specified? If yes, could you add a comment explaining why?

frontend/model/contracts/group.js
corrideat: There isn't much in the profile _and_ this is for removed members. Would you prefer `{ ...profile, status: PROFILE_STATUS.REMOVED }` instead?

frontend/model/contracts/group.js
corrideat: Yes, it needs to be specified or else it'll be set to the identity CSK.

frontend/model/contracts/group.js
taoeffect: Yes, I think that would be much more robust.

frontend/model/contracts/group.js
```diff
@@ -261,38 +261,79 @@ function updateGroupStreaks ({ state, getters }) {
...
     hooks: {
       preSendCheck: (_, state) => {
         // Avoid sending a duplicate action if the person has already
         // left or been removed from the chatroom
         return !!state?.users?.[member]
       }
     }
+  }).catch((e) => {
+    console.log('jjj action', { e })
+    if (leavingGroup && e?.name === 'GIErrorUIRuntimeError' && e?.cause?.name === 'GIErrorMissingSigningKeyError') {
```
taoeffect, line 327: ~~This `.cause` thing doesn't seem to be a thing? `GIErrorMissingSigningKeyError` sets `.name` to `GIErrorMissingSigningKeyError`.~~

EDIT: nevermind, it's set in `actions/utils.js`

shared/domains/chelonia/utils.js
```diff
@@ -498,12 +485,13 @@ export const recreateEvent = async (entry: GIMessage, rootState: Object): Promis
 
   if (!newRawOpV) return
 
+  const newOp = opT === GIMessage.OP_ATOMIC && (newRawOpV: any).length === 1
+    ? (newRawOpV: any)[0]
+    : [opT, newRawOpV]
```
taoeffect, line 490: What's going on here with `OP_ATOMIC`? It seems like if it's an atomic op with 1 operation it's for some reason modifying the message that's being sent to remove the outer `OP_ATOMIC` wrapper... but that's wrong. Why is it modifying what the user wants?

taoeffect: What's going on here? Are you sure this is right? Is `? (newRawOpV: any)[0]` correct?

Why is this being done? Is it rewriting what the user is sending? We should probably keep the developer intent and not modify messages the developer is sending.

shared/domains/chelonia/internals.js
```diff
@@ -1339,64 +1362,66 @@ const handleEvent = {
...
+        get signingContractID () {
+          return getContractIDfromKeyId(contractID, signingKeyId, state)
+        },
+        innerSigningKeyId,
+        get innerSigningContractID () {
+          return getContractIDfromKeyId(contractID, innerSigningKeyId, state)
         }
-        return acc
       }
+      return Promise.resolve(sbp(`${manifestHash}/${action}/sideEffect`, mutation))
```
taoeffect: Why is this wrapping the result in `Promise.resolve`? What if an error occurred? What's the point of later checking `.then` and `Promise.allSettled` if they're guaranteed to settled because of the `Promise.resolve`?

Or is that not how it works? If so please add a comment explaining because this is confusing...

corrideat: Agreed

shared/domains/chelonia/GIMessage.js
taoeffect: Then a comment about that should be here...

shared/domains/chelonia/utils.js
corrideat: Reverted

Silver-IT:
- Still existing issues:
  - **Issue 1**(https://github.com/okTurtles/group-income/pull/1780#issuecomment-1797865986)
  - **Issue 2** (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1797876669)
  - **Issue 3** (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1797889651)
  - **Issue 4** (https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784699012)
  Still two function `gi.actions/group/join` calls

Silver-IT:
@corrideat, @taoeffect. I have just checked the test scenario.
### Still existing issues in terms of workflow
  - [**Issue 1**](https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784531132)
  Problem:
    - user2 has successfully logged in.
    - One error occurred.
image

- [**Issue 2**](https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784557201)
Problem:
  - Two errors occurred.
  - `gi.actions/group/join` is called twice. It should be called only once.
image
image
image

- [**Issue 3**](https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784603749)
Problem:
  - Error in [Tab2] after being kicked from the group. (Not before to try join)
  - Three `gi.actions/group/join` calls. Only two calls are needed.
image
image

- [**Issue 4**](https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784699012)
Problem: 
  - 4 errors in console in Tab4
  - `gi.actions/group/join` is called twice. It should be called only once.
image
image
- [**Issue 5**](https://github.com/okTurtles/group-income/pull/1780#issuecomment-1784711834)
Seems fine.

(user1 = alexjin, user2 = greg, user3 = andrea)


IMHO, I don't approve the many updates in `model/contracts` and `controllers/actions`. Especially, I can't approve the way to leave chatroom when leaving group.

shared/domains/chelonia/internals.js
```diff
@@ -926,8 +969,8 @@ export default (sbp('sbp/selectors/register', {
       // of waiting, we need to remove it ourselves
       const keyId = findKeyIdByName(contractState, keyName)
       if (!keyId) {
+        console.log('@@@@ pendingWatch --- pushed to keysToDelete', { contractID, externalContractID, keyName, state: cloneDeep(contractState) })
```
taoeffect, line 972: Is this logging still needed?

taoeffect (approved):
This is looking really solid now, thanks @Silver-IT for finding those issues and @corrideat for figuring out the fixes! I think it's ready to merge into `e2e-protocol` but with also double-check with @Silver-IT to see if he can confirm that the issues he found have been fixed.

PR #1786, opened 2023-11-18
#1781 - Add "Use admin permissions to remove immediately" checkbox

### LINKED ISSUE #1781: Add "Use admin permissions to remove immediately" checkbox
### Problem

To prevent groups from becoming stagnant and thereby negatively affect the distribution, we need a way to allow certain members to manage the group's members without requiring a vote. Similar to how Github allows admins to merge PRs without reviews on them by clicking a checkbox to override the review requirements, we need a checkbox on the modal for removing members that appears only for "admins". Since we don't have roles & permissions implemented yet (#202), for now this checkbox should only appear for the group creator.

### Solution

Screenshot 2023-11-08 at 9 52 12 AM

For the group creator only, show a checkbox here that says `Use admin permissions to remove immediately` (add appropriate padding above and below the checkbox). When checked, the "Next" button transforms into "Remove", and when "Remove" is clicked, the user is immediately removed.

**Please branch off of `e2e-protocol` and send in this PR to `e2e-protocol` instead of `master`**

SebinSong:
closes #1781


taoeffect (approved):
Seems to work! Great job @SebinSong! 😄 🎉

PR #1787, opened 2023-11-19
Make Cypress test passing for E2E protocol

### LINKED ISSUE #1719: Fix all tests on e2e-protocol branch
### Problem

The tests are not passing on `e2e-protocol`.

The main reason is because of how invites work now:

- When a user is invited to a group, someone else who is already a member of the group [must be logged in in order to automatically send them the key so that they can join.](https://okturtles.slack.com/archives/C0EH7P20Y/p1693416630420659)

### Solution

Update all tests to address this.

Please be aware: if 4 users are joining a group, we DO NOT want this:

1. u1 joins group, copies invite, logs out
2. u2 uses link to join group, u2 logs out, u1 logs in
3. u1 logs out, u3 uses link to join group, u3 logs out, u1 logs in
4. u1 logs out, u4 uses link to join group, u2 logs out
5. u1 logs in

We DO want this:

1. u1 joins group, copies invite, logs out
2. u2 uses invite link to join group, logs out
3. u3 uses invite link to join group, logs out
4. u4 uses invite link to join group, logs out
6. u1 logs in (sending keys to u2, u3, u4 all at once)

Please do not send 1PR fixing all of the tests. Instead, send 1PR per each test spec file that is fixed. Keep sending PRs for different test specs until they are all fixed.

Test only locally - do not use Travis. The `e2e-protocol` branch is designed to prevent tests from running on Travis, please keep it that way.

Silver-IT:
### Summary of Changes

- Closes #1719


test/cypress/support/commands.js
```diff
@@ -336,6 +365,62 @@ Cypress.Commands.add('giAcceptGroupInvite', (invitationLink, {
...
+    cy.wait(1000)
+
+    cy.giLogout()
+  }
+
+  cy.giLogin(existingMemberUsername, { bypassUI })
+  // TODO: should remove the following cy.wait()
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000)
+  cy.giLogout()
```
taoeffect: @Silver-IT what's going on with these waits?

Do you need something from @corrideat to remove them?

test/cypress/integration/group-chat-direct-message.spec.js
```diff
@@ -233,32 +215,37 @@ describe('Create/Join direct messages and orders of direct message channels', ()
...
+      cy.getByDT('addToChannel-' + user2).click()
     })
-    cy.getByDT('closeModal').click()
 
-    cy.getByDT('channelName').should('contain', `${user1}, ${user4}, ${user5}`)
-    cy.getByDT('conversationWrapper').find('.c-message').should('have.length', 4) // 3 means user1, user3, user4, user5
+    cy.getByDT('channelName').should('contain', `${user4}, ${user1}, ${user2}`)
+    cy.giWaitUntilMessagesLoaded(false)
+    // NOTE: no notification messages in DM
+    cy.getByDT('conversationWrapper').find('.c-message').should('have.length', 0)
```
taoeffect, line 248: Can you explain why these tests have been changed so significantly, including the conditions that are being tested for?

test/cypress/integration/group-chat-scrolling.spec.js
```diff
@@ -85,23 +85,17 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
...
     me = user2
     cy.giRedirectToGroupChat()
 
-    cy.getByDT('channelName').should('contain', CHATROOM_GENERAL_NAME)
-    cy.giCheckIfJoinedChatroom(CHATROOM_GENERAL_NAME, me)
-
-    cy.getByDT('channelsList').find('ul>li:first-child').within(() => {
-      cy.get('[data-test]').should('contain', CHATROOM_GENERAL_NAME)
-    })
-
```
taoeffect, line 104: Why were these conditions removed?

test/cypress/integration/group-proposals.spec.js
```diff
@@ -276,8 +291,13 @@ describe('Proposals - Add members', () => {
 
   it('user4 registers and then joins the group through their unique invitation link', () => {
     cy.giSignup(`user4-${userId}`, { bypassUI: true })
+    // NOTE: need to wait until to save user settings to indexedStorage
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000)
```
taoeffect: Ideally this would be removed too. Can using `sbp` for listening to events help? Similar to how you did it in `giLogin`?

test/cypress/plugins/index.js
```diff
@@ -11,16 +11,10 @@
...
-// module.exports = (on) => {
-//   // Make cypress share the same babelrc config as the rest of the project
-//   // https://github.com/cypress-io/cypress-browserify-preprocessor#modifying-default-options
-//   const options = browserify.defaultOptions
-//   options.browserifyOptions.transform[1][1].babelrc = true
-
-//   on('file:preprocessor', browserify(options))
-// }
+const options = browserify.defaultOptions
+options.browserifyOptions.transform[1][1].babelrc = true
```
taoeffect: Why was this uncommented if it wasn't needed before?

test/cypress/support/commands.js
```diff
@@ -133,18 +164,28 @@ Cypress.Commands.add('giCreateGroup', (name, {
...
+              ruleName,
+              ruleThreshold
+            }
+          })
+
+          sbp('okTurtles.events/on', JOINED_GROUP, async ({ contractID }) => {
+            if (contractID === message.contractID()) {
+              await sbp('controller/router').push({ path: '/dashboard' }).catch(e => {})
+              resolve()
+            }
```
taoeffect: We probably want to remove the event listener just before calling `resolve()` so that it doesn't get called again

test/cypress/support/commands.js
```diff
@@ -264,64 +305,52 @@ Cypress.Commands.add('giAcceptGroupInvite', (invitationLink, {
...
+    // TODO: should remove the following cy.wait()
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000)
+    cy.giLogout()
+
+    cy.giLogin(existingMemberUsername, { bypassUI })
+    // TODO: should remove the following cy.wait()
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(2000)
+    cy.giLogout()
```
taoeffect: Indeed, all `cy.wait`s need to be removed

taoeffect (changes requested):
Thanks for the PR @Silver-IT!

Some comments & questions ready for you!

test/cypress/support/commands.js
Silver-IT: Yeah, nice catch.

test/cypress/plugins/index.js
Silver-IT: Actually, I didn't remove the comment, just moved it below. Please check the line 26, 27.
However, I can make it as before so that we have less changes for this solution.

test/cypress/integration/group-chat-scrolling.spec.js
Silver-IT: That's because we do this in `giAcceptGroupInvite`.
When it's the first time login, we check if they've fully synced general chatroom. It is defined in `giLogin` command especially line 123~125 in commands.js.

test/cypress/plugins/index.js
taoeffect: @Silver-IT 

> Actually, I didn't remove the comment, just moved it below

What do you mean...? The code has been uncommented is what I meant. It's now using browserify when I think it didn't need to use it?

test/cypress/integration/group-chat-direct-message.spec.js
Silver-IT: Actually, in this test scenario (group-chat-direct-message), it has the similar scenario as the previous version.
Changes are for
- Optimizing the workflow to save time
- Removing `user5` because I found that we don't need 5 members for the test. The more users, the more time takes.

test/cypress/integration/group-proposals.spec.js
Silver-IT: It was to wait for saving vuex to indexedStorage. I will do this in another way using `sbp`.

test/cypress/support/commands.js
Silver-IT: Actually
- The first one (`cy.wait(1000)`) is to wait for `KEY_REQUEST` event is published
- The second one (`cy.wait(2000)`) is to wait for `KEY_SHARE` event is published after getting received the `KEY_REQUEST` event

I will think about it more to remove this `cy.wait`. If you have good solution, please kindly share with me.

cc: @taoeffect, @corrideat

test/cypress/plugins/index.js
Silver-IT: You mean
- `browserify` is not needed now?
- These changes should be reverted again?

EDITED: Having checked without this change, I can see it works well.

taoeffect: I am saying that before we had `browserify` disabled here by having it commented out, right? So if it was working fine without `browserify` enabled, then there's no need to enable it because it will slow things down.

Silver-IT: > I am saying that before we had browserify disabled here by having it commented out, right?

Yes

> So if it was working fine without browserify enabled, then there's no need to enable it because it will slow things down.

I agree with you. Thanks

test/cypress/support/commands.js
```diff
@@ -336,6 +367,62 @@ Cypress.Commands.add('giAcceptGroupInvite', (invitationLink, {
   }
 })
 
+Cypress.Commands.add('giAcceptUsersGroupInvite', (invitationLink, {
```
taoeffect: Why are there two different commands that essentially do the same exact thing?

Can you see if you can DRY `giAcceptGroupInvite` and `giAcceptUsersGroupInvite` by combining them together?

test/cypress/support/commands.js
taoeffect: @Silver-IT In Cypress, we have 2 options:

1. Wait for a change in the UI
2. Wait for an event to be emitted using `sbp`

So you'll need to pick one of these.

If there is no change in the UI, and there is no event, you can create the event yourself.

test/cypress/support/commands.js
Silver-IT: Okay. I will combine them together.

test/cypress/support/commands.js
Silver-IT: Yeah, I know we have those two options. But in this case, there is no changes in the UI, and no events emitted.
Do you think it's fine to make events every time we publish the `KEY_REQUEST`, and `KEY_SHARE` events?

taoeffect: @Silver-IT I wouldn't post an event each time anything sends a KEY_REQUEST/KEY_SHARE, but rather in the `gi.actions` selector that you're waiting to finish, at the end of it, emit an event. You can even name it `CYPRESS_FINISHED_KEY_SHARE_EVENT` so that it's clear what it's being used for

test/cypress/support/commands.js
Silver-IT: @taoeffect, I tried to combine them together, and started to think that we should keep both of them with the reasons below;
- We should change argument names (`username` -> `usernames`, `displayName` -> `displayNames`), and it drives many changes in all the test scenario files
- We have three arguments (`isLoggedIn`, `shouldLogoutAfter`, `inviteCreator`) which are only for `giAcceptGroupInvite` function which should be used for single-user joining, not for multi-user joining. If we combine two functions together, these three arguments should be checked when it is multi-user joining, and that could be messy I guess.
- `existingMemberUsername` could be optional in single-user joining when he joins the group by the second time or more.

taoeffect: @Silver-IT OK if you think we should keep them separate then we can. In that case can you please add a comment above each command explaining what it does and how it's different? Also, maybe rename one of the commands to make it more clear how it's different from the other.

frontend/views/pages/Join.vue
```diff
@@ -158,22 +159,13 @@ export default ({
...
-          contractName: 'gi.contracts/identity',
-          data: {
-            groupContractID: groupId,
-            inviteSecret: secret
-          }
-        }).then(() => {
-          return sbp('gi.actions/group/switch', groupId)
-        })
+        await sbp('gi.actions/group/joinWithInviteSecret', groupId, secret)
+        // this.pageStatus = 'WELCOME'
```
taoeffect, line 168: @corrideat can you review these?

frontend/views/containers/navigation/GroupsList.vue
```diff
@@ -79,10 +79,10 @@ export default ({
       if (hash !== this.currentGroupId) {
         sbp('gi.actions/group/switch', hash)
       }
-      if (!emptyGroupPicture && path !== '/dashboard') {
-        this.$router.push(({ path: '/dashboard' }))
-      } else if (emptyGroupPicture && path !== '/pending-approval') {
+      if (emptyGroupPicture && path !== '/pending-approval') {
         this.$router.push(({ path: '/pending-approval' }))
+      } else if (!emptyGroupPicture && path === '/pending-approval') {
+        this.$router.push(({ path: '/dashboard' }))
```
taoeffect, line 85: @corrideat plz review

frontend/controller/actions/group.js
```diff
@@ -492,6 +505,20 @@ export default (sbp('sbp/selectors/register', {
     // after joining, we can set the current group
     sbp('gi.actions/group/switch', params.contractID)
   },
+  'gi.actions/group/joinWithInviteSecret': async function (groupId: string, secret: string) {
```
taoeffect, line 473: @corrideat plz review this file

frontend/model/contracts/group.js
```diff
@@ -1339,6 +1339,15 @@ sbp('chelonia/defineContract', {
...
         }
+      },
+      sideEffect ({ contractID }, { state }) {
+        if (Object.keys(state.chatRooms).length === 1) {
+          // NOTE: only general chatroom exists, meaning group has just been created
+          sbp('state/vuex/commit', 'setCurrentChatRoomId', {
+            groupId: contractID,
+            chatRoomId: state.generalChatRoomId
+          })
+        }
```
taoeffect, line 1362: @corrideat plz review

taoeffect (changes requested):
Great job @Silver-IT ! Enjoy your well deserved vacation!

I'm leaving a review here for @corrideat to pick up.

@corrideat plz see comments + see if you can revert the changes to `package.json` and `package-lock.json`

Silver-IT:
Thank you, @taoeffect, @corrideat.

test/cypress/support/commands.js
```diff
@@ -529,7 +529,7 @@ Cypress.Commands.add('giForceDistributionDateToNow', () => {
         contractID: sbp('state/vuex/state').currentGroupId,
         hooks: {
           // Setup a hook to resolve the promise when the action has been processed locally.
-          onprocessed: resolve()
+          onprocessed: () => resolve()
```
taoeffect, line 532: can also do `onprocessed: resolve` I think

taoeffect (approved):
Although there seems to be a rare deadlock in `group-member-removal.spec.js` still I'm going to merge this and we can figure out the problem in another PR.

https://github.com/okTurtles/group-income/assets/138706/87e18628-46a1-4181-a73f-28be8b4e9581

corrideat (approved):

frontend/controller/actions/group.js
```diff
@@ -361,6 +366,11 @@ export default (sbp('sbp/selectors/register', {
           console.error(`[gi.actions/group/join] Error while sending key request for ${params.contractID}:`, e)
           throw e
         })
+
+        if (params.blockOriginatingContract !== false) {
+          await keyRequestPromise
+        }
+
```
taoeffect, line 373: If you recall, `gi.actions/group/join` used to take some other parameter that we fought hard to get rid of in order to simplify the developer-facing API for joining a group — to avoid forcing developers to think about internal details.

Is it possible to do the same with this `blockOriginatingContract` parameter? Is there a way to get rid of it?

corrideat: That's needed to avoid a deadlock depending on which queue is being used. If it weren't there, the `identity/joinGroup` side-effect would result in a deadlock. On the other hand, usually you want to `await` to have feedback on when the operation completed.

taoeffect: So there's clearly a reason for it... but the question was whether or not it's possible to do things some other way so that this parameter doesn't need to be passed in?

corrideat: Right, I guess it's always _possible_, but not sure how _feasible_ it is. Paraphrasing the reason, without this parameter:

* Either we don't block on the originating contract (identity contract), which sometimes you want to do, if it's possible as this way you get feedback
* Or we block on the originating contract, in which we can't await on this when it's called from the identity side-effect

Removing this involves one of these tradeoffs, or perhaps rewriting `join` in some way. Perhaps the option of *not* awaiting on it from the side-effect, possibly along with a `preSendCheck` hook might work, but I don't think this reduces complexity, just merely hides it (like, you still need to know that calling / not calling it is appropriate and whether or not `await` can be used and why)

taoeffect: Ok, this might have to be something I return to after my break. 🔖 bookmarking for later

PR #1795, opened 2023-12-12
#1790 - Remove the extra space in "Add to channel" button

### LINKED ISSUE #1790: "Extra space" in "Add to channel" button
### Problem

Recently in the `1719-fix-all-tests-on-e2e-protocol-branch` I noticed that the `Add to channel` buttons had an extra space after "Add":

Screenshot 2023-12-11 at 9 59 44 AM

There actually was an extra space, but after removing it the problem is still there! I thought it was because a newline was being inserted, but that's actually not the problem (that's just how Firefox renders the markup to show 1 tag per line).

It turns out that this is a problem caused by this `margin-left` inline css:

Screenshot 2023-12-11 at 10 07 14 AM

That seems to be needed for spacing after the ✅  when a users is added to the channel, but it's not needed anywhere else.

### Solution

Create a branch off of `e2e-protocol` and fix this in both `ChatMembersAllModal.vue` and `GroupMembersAllModal.vue`.

SebinSong:
closes #1790 

From my investigation, it seems like the left-margin is only needed when the `span` wrapping a text is placed right next to an icon, such as for the case of `added` above and also for the case of `removed` (although no screenshot's attached for it).
So made a fix accordingly.

### DIFF
--- a/frontend/views/containers/chatroom/ChatMembersAllModal.vue
+++ b/frontend/views/containers/chatroom/ChatMembersAllModal.vue
@@ -441,11 +441,13 @@ export default ({
 
 ::v-deep .c-actions {
   span {
-    margin-left: 0.3rem;
-
     @include phone {
       display: none;
     }
+  }
+
+  i + span {
+    margin-left: 0.3rem;
   }
 
   .c-action-undo {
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
👏

PR #1796, opened 2023-12-13
Fix for a cypress failure related to push-subscription in Electron

SebinSong:
Fix for below issue in Cypress when it's run with Electron89.

Apparently the root cause is that below method is not supported in Electron (refer to [this article](https://medium.com/@MatthieuLemoine/my-journey-to-bring-web-push-support-to-node-and-electron-ce70eea1c0b0))
`pushManager.subscribe()`

I made a fix where the error is caught and silently logged out to the console instead of it becoming an uncaught Promise error.


taoeffect (approved):
Thanks @SebinSong!

PR #1799, opened 2023-12-16
#1791 - Immediately install new service worker on refresh

### LINKED ISSUE #1791: Immediately install new service worker on refresh
### Problem

Have a look at [this comment](https://github.com/okTurtles/group-income/pull/1770#discussion_r1423023047).

It turns out that [the way that service workers update](https://web.dev/articles/service-worker-lifecycle) is a bit convoluted and non-optimal by default.

The article says:

> After it's successfully installed, the updated service worker delays activating until the existing service worker is no longer controlling clients. This state is called "waiting", and it's how the browser ensures that only one version of your service worker is running at a time.

However, that last sentence is somewhat non-sensical / wrong, because it's contradicted later down the page here:

> The waiting phase means you're only running one version of your site at once, but if you don't need that feature, you can make your new service worker activate sooner by calling `self.skipWaiting()`.
> 
> This causes your service worker to kick out the current active worker and activate itself as soon as it enters the waiting phase (or immediately if it's already in the waiting phase). It doesn't cause your worker to skip installing, just waiting.

Since calling `self.skipWaiting()` kicks out the old service worker, there is still only "one version" of the service worker running at a time. So it sounds like we can safely call `self.skipWaiting()` all the time.

### Solution

Double-check what I'm saying above, and if it's right:

1. Call `self.skipWaiting()`
2. Do a manual call to `update()` on page refresh
3. Call `update()` once per hour
4. And we may also want to call `clients.claim()`

SebinSong:
close #1791 

1. `self.skipWaiting()` is already called in `install` event listener of the service-worker, which allows the website to use a new service-worker immediately.
2. `clients.claim()` method is currently called in the right place according to [this article](https://web.dev/articles/service-worker-lifecycle#clientsclaim), which is in `activate` event listener.


frontend/controller/service-worker.js
```diff
@@ -62,17 +62,31 @@ sbp('sbp/selectors/register', {
...
+      const readyState = pubsub.socket.readyState
+      const sendMsg = () => pubsub.socket.send(createMessage(
+        REQUEST_TYPE.PUSH_ACTION,
+        msgPayload
+      ))
+
+      if (readyState === WebSocket.CLOSED || readyState === WebSocket.CLOSING) {
+        pubsub.socket.addEventListener('open', function openHandler () {
+          sendMsg()
+          pubsub.socket.removeEventLister('open', openHandler)
```
snowteamer: This `open` handler is never going to be called, since a web socket cannot be reused once closed. That's why the client uses a new socket for every connection, and automatically removes and re-adds its socket event handlers.

Such custom handlers can be registered under the 'handlers' option in `createClient`, where the client is created (you already know the `messageHandlers` option which is very similar).

SebinSong: @snowteamer
>  a web socket cannot be reused once closed

 That's a good point that I missed while working on the previous commit. Will make another update that resolves this.

SebinSong: @snowteamer I realised that there is already a sbp event `PUBSUB_RECONNECTION_SUCCEEDED` that you created, which is emitted when the WebSocket is re-connected after losing a network connection. This seems to be useful in this case so decided to resolve the mentioned issue using it. ([here](https://github.com/okTurtles/group-income/pull/1799/commits/0327c551132cb4db8deeb72754d5cf538db37a9b))

frontend/controller/service-worker.js
```diff
@@ -62,28 +62,30 @@ sbp('sbp/selectors/register', {
...
+      // if it is, attach a event listener to PUBSUB_RECONNECTION_SUCCEEDED sbp event.
       // (context: https://github.com/okTurtles/group-income/pull/1770#discussion_r1439005731)
 
       const readyState = pubsub.socket.readyState
-      const sendMsg = () => pubsub.socket.send(createMessage(
-        REQUEST_TYPE.PUSH_ACTION,
-        msgPayload
-      ))
+      const sendMsg = (socket) => {
+        (socket || pubsub.socket).send(createMessage(
```
snowteamer: Better to just use `pubsub.socket` here I think. as the `socket` argument is going to refer to a custom event object instead of a socket

SebinSong: @snowteamer that's right. updated.

frontend/controller/service-worker.js
```diff
@@ -17,6 +18,15 @@ sbp('sbp/selectors/register', {
...
         swRegistration.active?.postMessage({ type: 'store-client-id' })
+
+        // if an active service-worker exists, checks for the updates immediately first and then repeats it every 1hr
+        await swRegistration.update()
+
+        const recursiveUpdate = async () => {
+          await sbp('service-worker/update')
+          setTimeout(recursiveUpdate, HOURS_MILLIS)
+        }
+        setTimeout(recursiveUpdate, HOURS_MILLIS)
```
taoeffect: This could probably be simplified to 1 line:

```js
```

frontend/controller/service-worker.js
```diff
@@ -52,17 +62,33 @@ sbp('sbp/selectors/register', {
 
     const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
     const existingSubscription = await registration.pushManager.getSubscription()
+    const messageToPushServerIfSocketConnected = (msgPayload) => {
+      // make sure the websocket client is not in the state of CLOSING, CLOSED before sending a message.
+      // if it is, attach a event listener to PUBSUB_RECONNECTION_SUCCEEDED sbp event.
+      // (context: https://github.com/okTurtles/group-income/pull/1770#discussion_r1439005731)
+
+      const readyState = pubsub.socket.readyState
+      const sendMsg = (socket) => {
```
taoeffect: Since `socket` is unused can probably be removed from parameters

taoeffect (changes requested):
Minor code cleanup requests

taoeffect (approved):

PR #1800, opened 2023-12-20
#1793 Dark theme status bar color in the PWA 

### LINKED ISSUE #1793: PWA darkmode issue
### Problem

While testing out the PWA PR that was [recently merged](https://github.com/okTurtles/group-income/pull/1770), I noticed an issue related to saving the PWA to the home screen on mobile in dark mode. The top part of the screen is white when it should be the same color as the rest of the background:

### Solution

Make sure the background matches in both dark-mode and light-mode.

SebinSong:
closes #1793 

The `theme_color` field in the `manifest.json` file is responsible for colouring the status bar but the thing is, the value needs to be updated dynamically either according to the browser's system preference or the user's selection in the app.

So I chose to use `<meta name='theme-color' >` instead, so that we can dynamically update the value from inside the app.

(Mainly referenced [this article](https://css-tricks.com/meta-theme-color-and-trickery/) for the task.)

### DIFF
--- a/frontend/assets/pwa-manifest.webmanifest
+++ b/frontend/assets/pwa-manifest.webmanifest
@@ -23,6 +23,5 @@
       "type": "image/svg+xml"
     }
   ],
-  "background_color": "#EDEDED",
-  "theme_color": "#EDEDED"
+  "background_color": "#EDEDED"
 }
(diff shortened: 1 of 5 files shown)


taoeffect (approved):
Approved!! Great work @SebinSong! 😄

PR #1802, opened 2023-12-30
Fix issue #1789

### LINKED ISSUE #1789: Incorrect date in notification
### Problem

nov0

Notification is dated "Nov 0", which isn't a correct date. Notification was received on 26 November and there's clearly enough space for a second number character, so doesn't appear to be an issue of truncation/overflow/etc.

This is on Firefox.

### Solution

1. Determine what's causing the issue.
2. Fix it.

snowteamer:
Closes #1789

This quickfix PR addresses an issue where `timeSince()` would replace `2` (\x32) characters instead of plain white spaces (\x20) by non-breaking spaces in its output, causing incorrect date display e.g. in notifications.


frontend/model/contracts/shared/time.js
```diff
@@ -224,7 +224,8 @@ export function timeSince (datems: number, dateNow: number = Date.now()): string
 
   if (interval >= DAYS_MILLIS * 2) {
     // Make sure to replace any ordinary space character by a non-breaking one.
-    return humanDate(datems).replace(/\x32/g, '\xa0')
+    // TODO: use .replaceAll when migrating to TS.
+    return humanDate(datems).replace(/\x20/g, '\xa0')
```
taoeffect: We can't we use `.replaceAll` now? 🤔

taoeffect (commented):
Question

frontend/model/contracts/shared/time.js
snowteamer: Just because our Flow version would complain about it being a non-existent string property. Or maybe we could add a `$FlowFixMe` comment to use it now anyway

test/cypress/integration/agroup-contributions.spec.js
```diff
@@ -0,0 +1,565 @@
+const userId = Math.floor(Math.random() * 10000)
+const groupName = 'Dreamers'
+const usersDisplayName = {
+  1: 'Greg',
+  2: 'Margarida',
+  3: 'Pierre',
+  4: 'Sandrina'
+}
+
```
taoeffect: This file seems to have been duplicated and added on accident? Can it be removed?

taoeffect (changes requested):

test/cypress/integration/agroup-contributions.spec.js
snowteamer: Yea sure - renaming a specific test file this way is a method I use sometimes to have it run first

taoeffect (approved):
Nice job @snowteamer!

EDIT: see comment: https://github.com/okTurtles/group-income/issues/1789#issuecomment-1887707789 - think this issue is still open

PR #1805, opened 2023-12-31
#1665 Typing indicator in the chatroom

### LINKED ISSUE #1665: Typing indicator
### Problem

It's often useful to know when someone is typing in the same chatroom as you.

### Solution

~~Use the server's key-value store to store this information on a key that maps to the chatroom contract, e.g. `<contractID>/typing`~~, and then use websockets to receive updates.

~~This issue depends on #1656 being closed first.~~ EDIT: no it doesn't. See commentary below. This issue can be closed simply by using the pubsub system.

SebinSong:
closes #1665 

**[ Video record ]**

https://github.com/okTurtles/group-income/assets/17641213/b880f4ee-b6aa-4265-b743-79cf3cf71d3e

**::: NOTE :::**
1. The UI of the typing indicator was inspired from that of Slack. If we want to update this to something else to better suit our chatroom design in the future, I can work on it then.

2. This is not demonstrated in the video above but the typing indicator sentence is `{user1}, {user2} are typing` when multiple users are typing at the same time.

3. How it works in summary is, the chatroom uses the pubsub system to publish/subscribe to the `PUBSUB_EVENT_TYPE.CHATROOM_USER_TYPING` event. Whenever the users navigate around different chatrooms that they are part of, they are unsubscribed from the previous pubsub channel(identified by the chatroom contract ID) and subscribed to the pubsub channel of the current chatroom, so that the Chatroom page (`GroupChat.vue`) stops listening for an event for the other chatrooms that it's not showing.

Hope it all makes sense and sounds good.

**cc.** @taoeffect


taoeffect (changes requested):
This is looking really great @SebinSong, awesome!

As I'm still on vacation I won't be able to dive deep into the code right now, but just from a quick look at the recording I have one request which is, would it be possible to implement this in a way such that the typing indicator doesn't cause a vertical "jump" in the chat text? Perhaps you could use a `<div>` as a slight horizontal strip that appears and covers some of text instead of shifting it up?

SebinSong:
@taoeffect 

I have moved the indicator down below the text input area, so that the existence of it doesn't lead to increasing the height of the `SendArea.vue` component.

SebinSong:
@snowteamer This PR uses the changes made in [this recently merged PR](https://github.com/okTurtles/group-income/pull/1792) of yours, so would appreciate your review on this PR as well when you have time.

frontend/controller/actions/chatroom.js
```diff
@@ -183,6 +183,18 @@ export default (sbp('sbp/selectors/register', {
...
     }))
   },
+  'gi.actions/chatroom/pubsub/sub': (chatroomContractID: string) => {
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+    pubsub.sub(chatroomContractID, true)
+  },
+  'gi.actions/chatroom/pubsub/unsub': (chatroomContractID: string) => {
+    const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
+    pubsub.unsub(chatroomContractID, true)
+  },
```
taoeffect: So... we can't do this. The pubsub system is used to listen for events for contracts. Have a look at `'chelonia/connect'`, `'chelonia/pubsub/update'`, and related code in `chelonia.js`.

Only that code is allowed to manage subscribing/unsubscribing from the contractIDs. If any other code starts managing that, it will break the app because the app will miss important new events on the contract chains and will fail to process updates to the group/chatroom state.

Another important point about the code in `'chelonia/connect'`: that code contains some `messageHandlers`, that currently look like this:

```js
    this.pubsub = createClient(pubsubURL, {
      ...this.config.connectionOptions,
      messageHandlers: {
        [NOTIFICATION_TYPE.ENTRY] (msg) {
          // We MUST use 'chelonia/private/in/enqueueHandleEvent' to ensure handleEvent()
          // is called AFTER any currently-running calls to 'chelonia/contract/sync'
          // to prevent gi.db from throwing "bad previousHEAD" errors.
          // Calling via SBP also makes it simple to implement 'test/backend.js'
          const { contractID } = GIMessage.deserializeHEAD(msg.data)
          sbp('chelonia/private/in/enqueueHandleEvent', contractID, msg.data)
        },
        [NOTIFICATION_TYPE.VERSION_INFO] (msg) {
          const theirVersion = msg.data.GI_VERSION

          const ourContractsVersion = process.env.CONTRACTS_VERSION
          const theirContractsVersion = msg.data.CONTRACTS_VERSION
          if (ourVersion !== theirVersion || ourContractsVersion !== theirContractsVersion) {
            sbp('okTurtles.events/emit', NOTIFICATION_TYPE.VERSION_INFO, { ...msg.data })
          }
        },
        [REQUEST_TYPE.PUSH_ACTION] (msg) {
          sbp('okTurtles.events/emit', REQUEST_TYPE.PUSH_ACTION, { data: msg.data })
        }
      }
```

We cannot really modify this code to add new handlers here. In fact, these existing handlers need to be modified (and I'll create a separate issue) so that they're not Group Income specific (as Chelonia is a generic framework).

Handlers for different types of messages could be passed in via a parameter to `'chelonia/connect'`, or via `'chelonia/configure'`.

Anyways, just mentioning that in case you need it.

Remember that with @snowteamer's changes you can listen to arbitrary rooms, not just contractIDs. But you might not need to use that I don't think... you could probably just publish a special type of message on the existing pubsub for the contractID, just with a different message type so that it's not classified as an `ENTRY` message and processed.

taoeffect (changes requested):
Great work @SebinSong! Preliminary review here.

One thing I noticed also was that the amount of time that the typing indicator is shown is imo a bit short. After they stop typing the fact that they were typing should remain visible for at least 30 seconds *if there is text in their send field*. If they deleted the text in their send field, the typing indicator should immediately disappear (perhaps a include a special type of message to "clear" the fact that they were typing).

Also, on FFDev edition, I noticed that there's no padding/margin above `u1 is typing`:

Screenshot 2024-01-09 at 10 40 22 AM

frontend/controller/actions/chatroom.js
SebinSong: @taoeffect Thanks for the detailed explanation about the system.

> Only that code is allowed to manage subscribing/unsubscribing from the contractIDs.

> Remember that with @snowteamer's changes you can listen to arbitrary rooms, not just contractIDs. 

it sounds like I must avoid using the chatroom contractID as a channel name, and in this case prefixing it to something like `CHATROOM_UI-${contractID}` can be an alternative solution. Let me make updates for this.

SebinSong:
@taoeffect 

> One thing I noticed also was that the amount of time that the typing indicator is shown is imo a bit short. After they stop typing the fact that they were typing should remain visible for at least 30 seconds if there is text in their send field. If they deleted the text in their send field, the typing indicator should immediately disappear (perhaps a include a special type of message to "clear" the fact that they were typing).

It sounds like another type of event for user having stopped typing is needed in that case. :)
will make an update accordingly.

frontend/controller/actions/chatroom.js
taoeffect: @SebinSong well, I'm not 100% sure. You could use the existing subscriptions that already established for you and publish different types of events on them. That's one simple option. The other option is to manage a new related subscription, which seems more complicated...

SebinSong: @taoeffect Right. updated accordingly.

SebinSong:
@taoeffect Updated the PR again with the work for your reviews above.

shared/pubsub.js
```diff
@@ -412,9 +417,22 @@ const defaultMessageHandlers = {
...
+        break
+      }
+      case PUBSUB_EVENT_TYPE.CHATROOM_USER_STOP_TYPING: {
+        sbp('okTurtles.events/emit', PUBSUB_EVENT_TYPE.CHATROOM_USER_STOP_TYPING, { username: data.username })
+        break
+      }
+      default: {
+        console.log(`[pubsub] Received data from channel ${channelID}:`, data)
+      }
+    }
```
taoeffect: Excellent, this PR is starting to shape up nicely.

There is one important thing to keep in mind though: these handlers need to be part of Group Income, and this code in `shared/pubsub.js` is going to be part of Chelonia. The two must be kept separate to make it easier to split out Chelonia in the future.

This is a great opportunity to improve the framework though. We made an oversight: we forgot to handle setting custom `messageHandlers` in `'chelonia/connect'`.

Currently, the code in there looks contains this:

```js
    this.pubsub = createClient(pubsubURL, {
      ...this.config.connectionOptions,
      messageHandlers: {
        [NOTIFICATION_TYPE.ENTRY] (msg) {
          // We MUST use 'chelonia/private/in/enqueueHandleEvent' to ensure handleEvent()
          // is called AFTER any currently-running calls to 'chelonia/contract/sync'
          // to prevent gi.db from throwing "bad previousHEAD" errors.
          // Calling via SBP also makes it simple to implement 'test/backend.js'
          const { contractID } = GIMessage.deserializeHEAD(msg.data)
          sbp('chelonia/private/in/enqueueHandleEvent', contractID, msg.data)
        },
        [NOTIFICATION_TYPE.VERSION_INFO] (msg) {
          const theirVersion = msg.data.GI_VERSION

          const ourContractsVersion = process.env.CONTRACTS_VERSION
          const theirContractsVersion = msg.data.CONTRACTS_VERSION
          if (ourVersion !== theirVersion || ourContractsVersion !== theirContractsVersion) {
            sbp('okTurtles.events/emit', NOTIFICATION_TYPE.VERSION_INFO, { ...msg.data })
          }
        },
        [REQUEST_TYPE.PUSH_ACTION] (msg) {
          sbp('okTurtles.events/emit', REQUEST_TYPE.PUSH_ACTION, { data: msg.data })
        }
      }
    })
```

From here we need to do several things here to prepare Chelonia for release:

1. Move `NOTIFICATION_TYPE.VERSION_INFO` and `REQUEST_TYPE.PUSH_ACTION` handlers out of there, and instead include them as part of an object that gets merged into `messageHandlers`, and pass that object in via the `'chelonia/connect'` parameter.
2. Undo your changes here to `NOTIFICATION_TYPE.PUB`, and move that code as part of the parameter that gets passed in to `'chelonia/connect'` (with the other 2 handlers)

So in the end you'll have something that looks something like this:

```js
  'chelonia/connect': function ({ messageHandlers } = {}): Object {
```

And then that gets merged below:

```js
    this.pubsub = createClient(pubsubURL, {
      ...this.config.connectionOptions,
      messageHandlers: {
        ...messageHandlers,
        [NOTIFICATION_TYPE.ENTRY] (msg) {
          // We MUST use 'chelonia/private/in/enqueueHandleEvent' to ensure handleEvent()
          // is called AFTER any currently-running calls to 'chelonia/contract/sync'
          // to prevent gi.db from throwing "bad previousHEAD" errors.
          // Calling via SBP also makes it simple to implement 'test/backend.js'
          const { contractID } = GIMessage.deserializeHEAD(msg.data)
          sbp('chelonia/private/in/enqueueHandleEvent', contractID, msg.data)
        }
```

And then update `main.js` with the handlers.

backend/pubsub.js
```diff
@@ -273,6 +273,7 @@ const defaultMessageHandlers = {
       ))
       return
     }
+
```
taoeffect: Stray whitespace modification

taoeffect (changes requested):

SebinSong:
@taoeffect 
Again, thnx for the helpful detailed explanations.
Updates for the change request is done in this commit: https://github.com/okTurtles/group-income/pull/1805/commits/1f161fbe1266a22c995694c5acb4552122dec024

shared/pubsub.js
```diff
@@ -412,11 +412,6 @@ const defaultMessageHandlers = {
     }, client.options.pingTimeout)
   },
 
-  [NOTIFICATION_TYPE.PUB] ({ channelID, data }) {
-    console.log(`[pubsub] Received data from channel ${channelID}:`, data)
-    // No need to reply.
-  },
-
```
taoeffect: I think it's actually a good idea to have this default handler here so that people know it's a built-in one that can be overwritten.

taoeffect (approved):
Great work @SebinSong! Approved! 😄 🎉

Issue #1558, opened 2023-04-26
Markdown in chat


taoeffect:
### Problem

We have no way of styling chat.

### Solution

Support basic markdown rendering.

This will obviously need a new dependency. The usual rules for adding new dependencies apply:

- Find the dependency that accomplishes what we need to do with the absolute minimum amount of additional code
- Find the dependency that is extremely well written and self-contained
- Make sure to pick a dependency that has few bugs
- Make sure it has few if any other dependencies

Silver-IT:
I think `markdown-it` library seems good. https://www.npmjs.com/package/markdown-it
It has only 5 dependencies and only a few issues remaining.

SebinSong:
It's implementing all the text styling in blue boxes in below screenshot

<img src='https://github.com/okTurtles/group-income/assets/17641213/1fc1a460-3d09-410c-ac96-74fca63cae70' width='460' >

SebinSong:
@taoeffect 
If you want to track the commit history for my weekly Todos, the feature branch:
[sebin/task/##1558-markdown-in-chat](https://github.com/okTurtles/group-income/commits/sebin/task/%23%231558-markdown-in-chat/)

SebinSong:
[ DEV MEMO ]

Possibly helpful resources:
- https://github.com/hl037/vue-contenteditable

SebinSong:
@taoeffect If you would like to track the commit history for my weekly Todos, the feature branch is:
[sebin/task/##1558-markdown-in-chat](https://github.com/okTurtles/group-income/commits/sebin/task/%23%231558-markdown-in-chat/)

SebinSong:
[ DEV MEMO ]
Possibly helpful resources:
- https://github.com/hl037/vue-contenteditable

taoeffect:
Closed by @SebinSong in #1815!

PR #1815, opened 2024-01-18
#1558 - markdown in chat

SebinSong:
work on #1558

First of all, I chose to use [`marked`](https://www.npmjs.com/package/marked) instead of `markdown-it` as the dependency because,

1. It is a 0 dependency package.
2. great popularity in the dev community (800M+ weekly downloads in npm)
3. has a great documentation and demo (https://marked.js.org)

---

This PR is the work for allowing below list of styled markdowns to be displayed in the chat message.
**1.** Code inline
**2.** Code block
**3.** Italic
**4.** Bold
**5.** Unordered list
**6.** Strikethrough
**7.** Link tag

[ screenshots ]

The colours of them are arbitrarily chosen by me (e.g. `<code>abc</code>` is in orange color) but let me know if we want to update them to different ones.

I've also worked on adding this various call-to-action buttons for some of the styles, which can make the menu start to look a bit crowded now, but let me know if we don't want this buttons.

cc. @taoeffect


marked-test.js
```diff
@@ -0,0 +1,29 @@
+// !!! TODO: REMOVE THIS FILE AFTER DEVELOPEMENT !!!
```
taoeffect: Should this file be removed now?

taoeffect (changes requested):
Fantastic work @SebinSong! 🥇 

This is looking really solid.

Here are the only issues I could find:

#### 1. spacing on adjacent styled words

Notice that whenever adjacent words have styling on them, it looks like they're part of the same word:

Screenshot 2024-01-18 at 8 46 11 AM

Compare that with the spacing that Slack adds between adjacent words with styling.

#### 2. not using real markdown

This is my fault, as I forgot to say that we shouldn't use Slack's syntax for styling.

Slack doesn't actually use real markdown, it uses its own modified version of markdown. In most markdown renderers that I know, and in the official specs,\*asdf\* gets rendered as *asdf*, not **asdf**.

Let's avoid propagating Slack's confusion any further, and conform to the official markdown spec.

#### 3. fix combined styling issues

In a lot of the examples i show in the above screenshot, whenever styling is combined with other styling, the results are wrong.

For example, ***bold italic*** didn't get rendered properly. `~*_strikethrough_*~` didn't get rendered properly (note: with the fixed markdown, it would be `~**_strikethrough_**~`).

Similarly, need to make sure that styling with links renders properly.

#### 4. Move the non-styling buttons to the right

Now that we've got more buttons, I think it makes sense to organize them a bit better. Since styling is a more commonly used activity than some of the special buttons that we already have, we can move those special buttons over to the right:

Screenshot 2024-01-18 at 8 59 58 AM

I would also preserve a tiny bit of margin between the "send" button and the special buttons so that the send button is not accidentally tapped on mobile devices.

SebinSong:
@taoeffect 
Just updated the PR with the work for your feedbacks. let me attach the demo for a couple of your feedback items too.

image

taoeffect (changes requested):
Fantastic work @SebinSong! 💪 

I think this will be the final change request before merging:

1. Can you avoid changing the font color for *italic* and ***bold italic*** text?
2. Can you use the existing blue color that we use for links for links?
3. I think I found an issue with blockquotes:

Screenshot 2024-01-23 at 11 07 47 AM

There was supposed to be an empty line in the blockquote:

Screenshot 2024-01-23 at 11 08 06 AM

E.g.:

> test
> test
> test
> 
> test

Not:

> test
> test
> test
> test

SebinSong:
@taoeffect updated PR according to your feedbacks

taoeffect (approved):
🎉 🎉 🎉 

Approved!! 😄

PR #1819, opened 2024-01-24
#1803 offline lazy load issue

### LINKED ISSUE #1803: Issue with offline handling for loading lazy loaded views
### Problem

*Note: the screenshot for this issue is taken from Firefox Developer Edition (FDE). I've noticed that in the Network tab when you set "Throttling" to "Offline", the behavior is different between FDE and Chrome. In FDE the white "offline" banner doesn't appear when you set Throttling to Offline, because it seems there's a bug in Firefox that doesn't disconnect the web socket in that mode, whereas Chrome correctly handles this. Please test both browsers as you're debugging this issue.*

Screenshot 2023-12-30 at 2 19 12 PM

When setting the browser to "Offline" mode via the Network Throttling tab, and then clicking an unloaded page (e.g. the Chat view), you see this screen. However, when going back online, and then switching to the Dashboard and back to the unloaded Chat view, this error doesn't disappear, and no attempt to re-load the resource is made.

### Solution

Try and reload the page when we switch back to it, in case we're back online.

SebinSong:
closes #1803

display a 'retry' button on the 500 server error page which reloads the page only if the browser detects the network connection.

https://github.com/okTurtles/group-income/assets/17641213/11835860-fb96-4d33-a581-4583fb2c36b1

### DIFF
--- a/frontend/views/containers/loading-error/ErrorPage.vue
+++ b/frontend/views/containers/loading-error/ErrorPage.vue
@@ -3,6 +3,11 @@
   h1.is-title-1.p-title 500
   i18n.is-title-3.c-title(tag='h2') Server error
   i18n.c-text(tag='p') We are having some trouble. Please try again later.
+  i18n.is-outlined.is-small.c-retry-btn(
+    v-if='ephemeral.isOnline'
+    tag='button'
+    @click='onRetryClick'
+  ) Retry
 
 .p-wrapper(v-else)
   h1.is-title-1.p-title 404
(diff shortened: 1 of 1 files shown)


taoeffect (commented):
Nicely done @SebinSong!

There seems to be an issue with Firefox Developer Edition though, using the "Throttling" toggle, and it didn't seem to trigger the "offline" event:

Screenshot 2024-01-25 at 8 47 27 AM

Therefore the reload button didn't appear. Do you know whether Firefox just doesn't emit the event when that menu is set to `Offline`? Or is it that it emits some other event?

SebinSong:
@taoeffect yes, I think that `offline` throttling option in firefox is broken. So instead of using the devtool, I tested it with actually disabling/re-enabling the wifi connection on my laptop.

https://github.com/okTurtles/group-income/assets/17641213/4fea09a4-51a7-4a9a-8fce-63321ca97a62

taoeffect (approved):
Great work!

PR #1821, opened 2024-01-25
#1818 Cropped off menu in chat

### LINKED ISSUE #1818: 'Edit message' menu in the chatroom is cropped off sometimes
### Problem
In the menu revealed by hovering over a message UI component in the chatroom, sometimes the 'Edit' menu item is cropped off/hidden and makes it impossible to switch to the edit mode.

### Solution
Find out the cause of the issue and fix it.

SebinSong:
closes #1818


taoeffect (changes requested):
Hmm, this doesn't pass the linter (see `grunt dev`).

Also, I wasn't able to really test this because I don't know how to reproduce it in `e2e-protocol`... in what situation did it fail before?

SebinSong:
@taoeffect Fixed that linter error. and here is how you reproduce the issue. (you have to reduce the screen width to the tablet range)

https://github.com/okTurtles/group-income/assets/17641213/663b0f1f-dcfe-418c-93e8-342f98780e38

taoeffect (approved):
👏

PR #1824, opened 2024-01-28
Use CYPRESS_RECORD_KEY secret in ci.yml

### LINKED ISSUE #1804: Upload build artifacts to Cypress Dashboard for failed Github Actions
### Problem

Here's an example of a failed build: https://github.com/okTurtles/group-income/actions/runs/7363656062/job/20043319096?pr=1802

When a build fails with Travis, we have it set to send the built artifacts to https://dashboard.cypress.io/ , e.g.: https://cloud.cypress.io/projects/q6whky/runs/4e36a813-3058-48aa-8bd9-8d2dd93ba830/test-results/a08c2575-a751-4d89-b02a-cf1df74f9977/video

This doesn't happen with failed Github Actions.

### Solution

We'd like to do the same with Github Actions so that we can see what happened. Figure out how to do this. If you need me to set some setting and don't have permissions, reach out on Slack and let me know what needs to be done.

snowteamer:
Closes #1804

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: NodeJS with Grunt
 
+env:
+  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+
 on:
   push:
     branches: [ "master" ]


taoeffect (approved):
😄 🎉 

Let's see if this works!!

PR #1825, opened 2024-01-28
Add initial loading-screen to index.html

### LINKED ISSUE #1822: Create initial app loading screen
### Problem

In PR #1521, to address #1765, @corrideat added this `v-if` to the router in `frontend/index.html`:

```html
<router-view v-if="ephemeral.ready" class="l-page"></router-view>
```

This causes a blank white page to appear on initial page load for a bit.

### Solution

Create a `v-else` and set it to a page that shows a simple loading page to indicate that the app is loading.

Create a branch off of `master` (by the time you see this issue Sebin, `e2e-protocol` should be merged)

SebinSong:
closes #1822 

@taoeffect 
I thought I would try using the existing screen in the app, which is `Loading.vue` with adding a GI logo in it like below screenshot. The logo will only be exposed there when the component is used as the initial loading screen btw.

If you have a more interesting idea for it though, let me know. happy to work on that.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -377,10 +377,9 @@ async function startApp () {
       }).finally(() => {
         this.ephemeral.ready = true
 
+        // remove the loading animation screen
         const loadingScreenEl = document.querySelector('#main-loading-screen')
-        if (loadingScreenEl) {
-          loadingScreenEl.classList.add('is-removed')
-        }
+        loadingScreenEl && loadingScreenEl.remove()
       })
     },
     computed: {
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -29,7 +29,6 @@ import BannerGeneral from './views/components/banners/BannerGeneral.vue'
 import Navigation from './views/containers/navigation/Navigation.vue'
 import AppStyles from './views/components/AppStyles.vue'
 import Modal from './views/components/modal/Modal.vue'
-import Loading from '@components/Loading.vue'
 import ALLOWED_URLS from '@view-utils/allowedUrls.js'
 import './views/utils/avatar.js'
 import './views/utils/ui.js'
@@ -252,8 +251,7 @@ async function startApp () {
       BackgroundSounds,
       BannerGeneral,
       Navigation,
-      Modal,
-      Loading
+      Modal
     },
     data () {
       return {
@@ -378,6 +376,11 @@ async function startApp () {
         console.error(`[main] caught ${e?.name} while fetching settings or handling a login error: ${e?.message || e}`, e)
       }).finally(() => {
         this.ephemeral.ready = true
+
+        const loadingScreenEl = document.querySelector('#main-loading-screen')
+        if (loadingScreenEl) {
+          loadingScreenEl.classList.add('is-removed')
+        }
       })
     },
     computed: {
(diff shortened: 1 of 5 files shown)


taoeffect (commented):
Hmm, I'm not seeing the loading screen... 🤔 

Notice the blank white page after refresh is clicked (this in FFDev):

https://github.com/okTurtles/group-income/assets/138706/de6ec06f-c9ac-439f-bbc4-60fe92f0f4dc

SebinSong:
@taoeffect I see.
I just looked into it again and that blank white screen shows regardless of the value of `ephemeral.ready`. In fact, it is when even before the `Vue` application is initialised. So if we are aiming to replace it, apparently it has to be handled outside of the Vue context. (e.g. adding raw html/css to the `index.html`) does it sound good?

taoeffect:
I'll reply on Slack

SebinSong:
@taoeffect 
here is the screenshot for the proof.

image

frontend/main.js
```diff
@@ -374,7 +375,12 @@ async function startApp () {
       }).catch(e => {
         console.error(`[main] caught ${e?.name} while fetching settings or handling a login error: ${e?.message || e}`, e)
       }).finally(() => {
-        Vue.set(this.ephemeral, 'ready', true)
+        this.ephemeral.ready = true
+
+        const loadingScreenEl = document.querySelector('#main-loading-screen')
+        if (loadingScreenEl) {
+          loadingScreenEl.classList.add('is-removed')
+        }
```
snowteamer: This will minimize the element so that it becomes a single pixel wide. Why couldn't it be removed instead, or its display set to `none`? I think a clarifying comment would be useful here

SebinSong: @snowteamer I think it makes sense to remove it completely instead of hiding it if there is no other cases that the app needs to display this loader screen.

SebinSong:
@taoeffect 
I spent some time re-ordering the existing operation steps in `startApp()` in `main.js` so that Vue app initialisation happens faster. that did reduce little bit of loading time but what I noticed instead during the process was that, the browser is still waiting a long list of javascript files for the most time when that white blank screen is showing (will attach a video recording for this in Slack).
So I decided that the loading animation has to be an element that is part of `index.html` instead of being rendered as a Vue component. I tried to make it as minimal as I can in favor of this opinion of yours in Slack.

>  In SPAs, index.html is treated as a blank loader usually. In other words, it should be small and uncluttered.

Hope that looks good for you too.

taoeffect (changes requested):
Interesting approach @SebinSong!

When I loaded a previously loaded session (after starting FFDev), I was greeted with the following:

Screenshot 2024-02-01 at 5 37 28 PM

So there are 2 bugs here:

1. Vue correctly complains about `<loading>` component not being defined as part of the `components` of `App.vue` in `main.js`
2. The loading screen covers this prompt that is supposed to appear when quitting `grunt dev` and restarting it while there was an existing user+group:

Screenshot 2024-02-01 at 5 41 52 PM

To reproduce, create a group, then restart `grunt dev` and refresh the page after `grunt dev` fully loads.

taoeffect (changes requested):
Great work @SebinSong - seems like all of the issue were fixed!

Only thing is, some debug `console.log` statements were still left in the PR. Once those are removed I can merge!

SebinSong:
@taoeffect my bad on all those `console.log` :)
Updated again.

taoeffect (approved):
Nicely done @SebinSong!

With some more testing I noticed an issue I didn't notice before.

If you perform a refresh it always redirects to the dashboard. For example, if you are viewing the general chatroom and refresh, it takes you to the dashboard still.

However, this seems to be an issue on `master` too, not just this PR, so I'll go ahead and approve and then open a new issue.

PR #1844, opened 2024-02-12
Remove pepper, read secret from file and minimise the number of secrets

corrideat:


data/secret_zkpp_parameters.json
```diff
@@ -0,0 +1 @@
+{"secret":"<REPLACE THIS WITH A LONG RANDOM STRING>"}
```
taoeffect: Some issues here:

1. We are perfectly capable of generating "a long random string" for the user, so we should do this for them.
2. The `data/` folder is not supposed to have anything in it under version control. So this file needs to be deleted.
3. Have a look at [how the `chel serve` command will be implemented](https://github.com/okTurtles/chel/issues/29). This should be a key in the database, and the database can be in-memory, on-disk as a file, or on-disk as a record in the sqlite database. So again, this file doesn't belong here.

backend/zkppSalt.js
```diff
@@ -1,27 +1,57 @@
...
 import nacl from 'tweetnacl'
 import sbp from '@sbp/sbp'
 import { boxKeyPair, encryptContractSalt, hashStringArray, hashRawStringArray, hash, parseRegisterSalt, randomNonce, computeCAndHc, base64urlToBase64, base64ToBase64url } from '~/shared/zkpp.js'
+import { readFileSync } from 'fs'
+
+// Input keying material used to derive various secret keys used in this protocol
+let IKM: string
+
+try {
+  const secretZkppParams = readFileSync('data/secret_zkpp_parameters.json', 'utf-8')
```
taoeffect: See comment below. This needs to come from the database.

shared/domains/chelonia/db.js
```diff
@@ -17,6 +17,7 @@ const getLogHead = (contractID: string): string => `${headPrefix}${contractID}`
 type HEADInfo = { HEAD: string; height: number }
 
 export const checkKey = (key: string): void => {
+  if (key.startsWith('private/')) return
```
taoeffect: Is it also still allowing private keys to be read from the database by anyone using the `/file` API..?

backend/zkppSalt.js
```diff
@@ -1,27 +1,57 @@
...
+    throw new Error('ZKPP secret is too short. It must be at least 16 characters long.')
+  }
+  if (parsedSecretZkppParams.secret === '<REPLACE THIS WITH A LONG RANDOM STRING>') {
+    if (process.env.NODE_ENV === 'production') {
+      throw new Error('Please change the ZKPP from its default value')
+    }
+    console.warn('Using the default value for the ZKPP secret. This is fine during development, but you must give it a different value in production mode.')
+    console.warn('If you are setting up this server for the first time, you can create a file at `data/secret_zkpp_parameters.json` with the following freshly-generated contents:')
+    console.warn(JSON.stringify({ secret: randomBytes(33).toString('base64') }))
+  }
```
taoeffect: All of these checks can be removed by managing this secret ourselves.

backend/zkppSalt.js
```diff
@@ -1,27 +1,57 @@
-import { blake32Hash } from '~/shared/functions.js'
-import { timingSafeEqual } from 'crypto'
+import { timingSafeEqual, randomBytes } from 'crypto'
 import nacl from 'tweetnacl'
 import sbp from '@sbp/sbp'
 import { boxKeyPair, encryptContractSalt, hashStringArray, hashRawStringArray, hash, parseRegisterSalt, randomNonce, computeCAndHc, base64urlToBase64, base64ToBase64url } from '~/shared/zkpp.js'
+import { readFileSync } from 'fs'
+
+// Input keying material used to derive various secret keys used in this protocol
+let IKM: string
```
taoeffect: Could you add more comments here to comment on whether or not this secret is relevant to data migration from one server to another?

taoeffect (changes requested):
Review ready!

data/secret_zkpp_parameters.json
corrideat: > We are perfectly capable of generating "a long random string" for the user, so we should do this for them.

Not really, because this parameter is set during server configuration and should not be changed afterwards without good reason to do so (if it is, passwords will stop working). So, we can generate this as part of an installation step, but not at runtime.

taoeffect: > So, we can generate this as part of an installation step, but not at runtime.

Yes, especially if this is such a sensitive value. We definitely need to be in charge of it fully.

So we can automatically create it during first run if it's not set, and prevent it from being overwritten from then on.

You can create a new prefix for this: `private/immutable/<key>`

backend/routes.js
```diff
@@ -280,46 +293,48 @@ route.POST('/zkpp/register/{contract}', {
...
     const ip = req.info.remoteAddress
-    console.error('Error at POST /zkpp/{contract}: ' + e.message, { ip })
+    console.error('Error at POST /zkpp/{contractID}: ' + e.message, { ip })
   }
 
   return Boom.internal('internal error')
 })
 
-route.GET('/zkpp/{contract}/auth_hash', {
+route.GET('/zkpp/{contractID}/auth_hash', {
```
taoeffect, line 319: Nice job on renaming these from `contract` to `contractID` 👍

backend/zkppSalt.js
```diff
@@ -1,27 +1,69 @@
...
+//      for the old instance.
+//   2. For other migrations (when the server administrator changes):
+//      a. Instances could implement a method to transfer salt records from one
+//         server to the other, either in an encrypted or unencrypted form.
+//         Since the actual encryption key used for `_private_rid_` entries is
+//         different for each contract ID, either key is possible. However, this
+//         is not implemented.
+//      b. Alternatively, migration can be done without migrating password salt
+//         records. This requires user interaction to create new salt records
+//         on the new server.
```
taoeffect, line 48: Really well written and clear comment 👏, well done, thanks @corrideat!

backend/zkppSalt.js
```diff
@@ -1,27 +1,69 @@
...
-const recordPepper = 'pepper' // TODO: get rid of `recordPepper`, just use `private/rid/${contractID}`
 // used to encrypt salts in database
-const recordMasterKey = 'masterKey' // TODO: store in file that's got root privs (after dropping privs)
+const recordSecret = IkmPromise.then(IKM => Buffer.from(hashStringArray('private/recordSecret', IKM)).toString('base64'))
 // corresponds to the key for the keyed Hash function in "Log in / session establishment"
-const challengeSecret = 'secret' // TODO: generate randomly and store in DB under private prefix
+const challengeSecret = IkmPromise.then(IKM => Buffer.from(hashStringArray('private/challengeSecret', IKM)).toString('base64'))
 // corresponds to a component of s in Step 3 of "Salt registration"
-const registrationSecret = 'secret' // TODO: generate randomly and store in DB under private prefix
+const registrationSecret = IkmPromise.then(IKM => Buffer.from(hashStringArray('private/registrationSecret', IKM)).toString('base64'))
```
taoeffect: These values can be set in a `initZkpp()` function that gets called from `database.js` (see bottom of that file)

taoeffect (approved):
Great work!

PR #1849, opened 2024-02-13
Fix error to close CreatePoll modal by clicking sidebar

### LINKED ISSUE #1843: Update z-index of modal compared to the sidebar
### Problem

We have `frontend/assets/style/_variables.scss` which defines the SCSS variables we are using in the whole frontend styles.
I have one thing weird here that `$zindex-modal` and `$zindex-sidebar` is same.
If these two things are same, which one would be on top of the other? Sidebar? Modal?
It depends on the how early it is rendered. So if modal is rendered before the sidebar, the sidebar would be on top of the modal.
And if the sidebar is rendered earlier, than the modal would be on top of the modal.
It happens while we create a poll. Clicking the sidebar doesn't not close the `CreatePoll` modal.

### Solution

Update `$zindex-modal` or `zindex-sidebar` so that `$zindex-modal` would be bigger than `$zindex-sidebar`.

Silver-IT:
### Summary of Changes

- Closes #1843

### DIFF
--- a/frontend/views/containers/chatroom/CreatePoll.vue
+++ b/frontend/views/containers/chatroom/CreatePoll.vue
@@ -273,7 +273,7 @@ export default {
   height: 0;
   pointer-events: none;
   overflow: hidden;
-  z-index: $zindex-modal;
+  z-index: $zindex-modal + 1;
 
   &.is-active {
     pointer-events: initial;
--- a/frontend/views/containers/chatroom/CreatePoll.vue
+++ b/frontend/views/containers/chatroom/CreatePoll.vue
@@ -274,7 +274,7 @@ export default {
   pointer-events: none;
   overflow: hidden;
   // NOTE: z-index should be bigger than the one of sidebar
-  //       because CreatePool modal is rendered before the sidebar
+  //       because CreatePoll modal is rendered before the sidebar
   // REF:  https://github.com/okTurtles/group-income/issues/1843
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/CreatePoll.vue
```diff
@@ -273,7 +273,10 @@ export default {
   height: 0;
   pointer-events: none;
   overflow: hidden;
-  z-index: $zindex-modal;
+  // NOTE: z-index should be bigger than the one of sidebar
+  //       because CreatePool modal is rendered before the sidebar
+  // REF:  https://github.com/okTurtles/group-income/issues/1843
+  z-index: $zindex-modal + 1;
```
Silver-IT, line 279: Not sure if it's better to use `$zindex-sidebar + 1` but it's what we discussed in [here](https://github.com/okTurtles/group-income/issues/1843#issuecomment-1939568088).

Silver-IT (commented):
Ready for the review!

SebinSong (dismissed):
LGTM!

frontend/views/containers/chatroom/CreatePoll.vue
SebinSong: A minor typo in the comment here btw:

`CreatePool` -> `CreatePoll`

taoeffect (approved):

PR #1850, opened 2024-02-14
Signature verification

corrideat:


backend/database.js
```diff
@@ -233,7 +235,7 @@ export default async () => {
       if (!persistence || !await sbp('chelonia/db/get', key)) {
         const value = await readFile(path.join(dataFolder, key), 'utf8')
         // Load only contract source files and contract manifests.
-        if (value.startsWith(CONTRACT_MANIFEST_MAGIC) || value.startsWith(CONTRACT_SOURCE_MAGIC)) {
+        if (value.startsWith(CONTRACT_MANIFEST_MAGIC) || value.slice(0, 32).includes(CONTRACT_SOURCE_MAGIC)) {
```
taoeffect: per call, plz revert this and related changes

.gitignore
```diff
@@ -3,6 +3,8 @@
 /dist/
 /dist-dashboard/
 *?rollup-plugin-vue=script.js
+key.json
+key.pub.json
```
taoeffect, line 7: Nice 👍

Gruntfile.js
```diff
@@ -111,9 +111,18 @@ module.exports = (grunt) => {
...
   async function generateManifests (dir, version) {
     if (development) {
+      const keyFile = process.env.KEY_FILE || 'key.json'
+      const pubKeyFile = process.env.PUB_KEY_FILE || 'key.pub.json'
+      if (fs.existsSync(keyFile)) {
+        grunt.log.writeln(chalk.underline(`Key file ${keyFile} exists, using that.`))
+      } else {
+        grunt.log.writeln(chalk.underline(`\nRunning 'chel keygen --pubout ${pubKeyFile} --out ${keyFile}'`))
+        const { stdout } = await execWithErrMsg(`chel keygen --pubout ${pubKeyFile} --out ${keyFile}`)
+        console.log(stdout)
```
taoeffect, line 121: Isn't this creating a situation where we're signing contracts with any random key?

Don't we want to restrict signing to a specific key?

Or is this fine because it will make it easier for new volunteers to submit PRs ... in which case we can re-sign the contracts for them?

backend/database.js
```diff
@@ -212,7 +212,7 @@ export default async () => {
   if (persistence !== 'fs' || options.fs.dirname !== './data') {
     // Remember to keep these values up-to-date.
     const HASH_LENGTH = 52
-    const CONTRACT_MANIFEST_MAGIC = '{"head":{"manifestVersion"'
+    const CONTRACT_MANIFEST_MAGIC = '{"head":"{\\"manifestVersion\\"'
```
taoeffect, line 215: Great catch! 👍

shared/domains/chelonia/internals.js
```diff
@@ -169,14 +169,71 @@ export default (sbp('sbp/selectors/register', {
...
+      }
+
+      if (!has(contractSigningKeys, manifest.signature.keyId)) {
+        throw new Error(`Invalid or missing signature in manifest ${manifestHash} (named ${contractName}). It claims to be signed with a key with ID ${manifest.signature.keyId}, which is not listed in its 'signingKeys' field.`)
+      }
+      verifySignature(contractSigningKeys[manifest.signature.keyId], manifest.body + manifest.head, manifest.signature.value)
+      console.info(`[chelonia] successful signature verification for ${manifestHash} (named ${contractName}) using ${manifest.signature.keyId}. The following key IDs will now be trusted for this contract name`, Object.keys(contractSigningKeys))
+      signatureValidated = true
+      rootState.contractSiginingKeys[contractNameLookupKey] = contractSigningKeys
+    }
```
taoeffect: All of this new code, is it possible to refactor it to better organize things and keep this function from getting out of control? Perhaps move some of this code to function(s)?

shared/domains/chelonia/internals.js
```diff
@@ -169,14 +169,71 @@ export default (sbp('sbp/selectors/register', {
   'chelonia/private/queueEvent': function (queueName, invocation) {
     return sbp('okTurtles.eventQueue/queueEvent', queueName, ['chelonia/private/invoke', this._instance, invocation])
   },
-  'chelonia/private/loadManifest': async function (manifestHash: string) {
+  'chelonia/private/loadManifest': async function (contractName: string, manifestHash: string) {
+    if (!contractName || typeof contractName !== 'string') {
```
taoeffect: This check can be simplified to `typeof contractName !== 'string'` by itself

Gruntfile.js
corrideat: > Isn't this creating a situation where we're signing contracts with any random key?

Correct.

> Don't we want to restrict signing to a specific key?

Also correct.

> Or is this fine because it will make it easier for new volunteers to submit PRs

Also correct.

> in which case we can re-sign the contracts for them?

And also correct.

Note:
- This is run only during development (so, the key used is of little significance)
- During deployment, a specific trusted key should be used
- And ideally there'll be some review involved, even if the signing itself is automated somehow. Volunteers shouldn't ever have to sign a contract for deployment.

shared/domains/chelonia/internals.js
corrideat: It cannot. `typeof '' === 'string'`

taoeffect: I think it's ok... it's not really a bug IMO. Will leave this decision to you though as it doesn't matter much

shared/domains/chelonia/internals.js
```diff
@@ -169,14 +169,71 @@ export default (sbp('sbp/selectors/register', {
...
     if (this.manifestToContract[manifestHash]) {
       console.warn('[chelonia]: already loaded manifest', manifestHash)
       return
     }
     const manifestURL = `${this.config.connectionURL}/file/${manifestHash}`
-    const manifest = await fetch(manifestURL, { signal: this.abortController.signal }).then(handleFetchResult('json'))
+    const manifestSource = await fetch(manifestURL, { signal: this.abortController.signal }).then(handleFetchResult('text'))
+    const manifestHashOurs = createCID(manifestSource)
+    if (manifestHashOurs !== manifestHash) {
+      throw new Error(`bad hash ${manifestHashOurs} for manifst '${manifestHash}'! Should be: ${manifestHash}`)
```
taoeffect: Typo: `manifst` => `manifest`

taoeffect: Recommend simplifying Error message to:

> `expected manifest hash ${manifestHash}. Got: ${manifestHashOurs}`

shared/domains/chelonia/internals.js
```diff
@@ -169,14 +169,71 @@ export default (sbp('sbp/selectors/register', {
...
+    }
+    const manifest = JSON.parse(manifestSource)
+    if (!has(manifest, 'signature') || typeof manifest.signature.keyId !== 'string' || typeof manifest.signature.value !== 'string') {
+      throw new Error(`Invalid or missing signature field for manifest ${manifestHash} (named ${contractName})`)
+    }
+    const rootState = sbp(this.config.stateSelector)
+    if (!has(rootState, 'contractSiginingKeys')) {
+      this.config.reactiveSet(rootState, 'contractSiginingKeys', Object.create(null))
+    }
+    const contractNameLookupKey = `name:${contractName}`
```
taoeffect, line 191: Why not just use `contractName` directly?

corrideat: Because it's a user-defined key, so it's potentially insecure (e.g., if the name happens to be `__proto__`, `prototype` or something like that defined later.

shared/domains/chelonia/internals.js
corrideat: In that case, I'll leave it as is... I think an empty contract name will break things with the selectors (imagine `<manifestHash>//my.action.selector` (?)), or at least be very confusing, so it shouldn't be allowed.

shared/domains/chelonia/internals.js
corrideat: It is. I'll make put it elsewhere

shared/domains/chelonia/internals.js
corrideat: I'll fix this

shared/domains/chelonia/internals.js
taoeffect: I don't think that's a problem. This key is coming from a trusted source (the developer's own source code), so it's fine.

If you can't trust the developer's code then it doesn't make sense to do anything at all, as there's no protection.

shared/domains/chelonia/internals.js
```diff
@@ -169,14 +169,90 @@ export default (sbp('sbp/selectors/register', {
...
     if (this.manifestToContract[manifestHash]) {
       console.warn('[chelonia]: already loaded manifest', manifestHash)
       return
     }
     const manifestURL = `${this.config.connectionURL}/file/${manifestHash}`
-    const manifest = await fetch(manifestURL, { signal: this.abortController.signal }).then(handleFetchResult('json'))
-    const body = JSON.parse(manifest.body)
+    const manifestSource = await fetch(manifestURL, { signal: this.abortController.signal }).then(handleFetchResult('text'))
+    const manifestHashOurs = createCID(manifestSource)
+    if (manifestHashOurs !== manifestHash) {
```
taoeffect, line 251: Nicely spotted!! 👏

taoeffect (approved):
Excellent work!

PR #1854, opened 2024-02-20
#1571 - Create vuex module for chat

### LINKED ISSUE #1571: Create Vuex module for chat related settings
### Problem

The `state.js` file is quite large still and we are slowly improving its organization. We've already moved a lot of state & mutations & getters into two separate vuex modules: `'~/frontend/model/notifications/vuexModule.js'` & `'~/frontend/model/settings/vuexModule.js'`, but the chat related settings are still in `state.js`

### Solution

Similar to how the notifications and settings vuex modules work, create a `model/chatroom/vuexModule.js` file & folder and move the chatroom related state/mutations/getters into that and update all dependent files.

SebinSong:
closes #1571 

The work was done based on the docs here:  https://v3.vuex.vuejs.org/guide/modules.html


frontend/model/state.js
```diff
@@ -72,22 +70,23 @@ sbp('sbp/selectors/register', {
...
       }
-      if (state.chatRoomUnread[chatRoomId].others) {
-        state.chatRoomUnread[chatRoomId].others.forEach(o => {
-          state.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
+      if (chatroomState.chatRoomUnread[chatRoomId].others) {
+        chatroomState.chatRoomUnread[chatRoomId].others.forEach(o => {
+          chatroomState.chatRoomUnread[chatRoomId].messages.push(Object.assign({ type: MESSAGE_TYPES.INTERACTIVE }, o))
         })
-        Vue.delete(state.chatRoomUnread[chatRoomId], 'others')
+        Vue.delete(chatroomState.chatRoomUnread[chatRoomId], 'others')
```
taoeffect: I think we can delete this whole section now since the groups will need to be recreated anyway.

taoeffect (dismissed):
Nice! I couldn't find any issues... but this is a significant change so I'm going to wait on either @Silver-IT and/or @corrideat for another approval before merging.

frontend/model/state.js
Silver-IT: Yeah, correct. As we can see in the comment above, we could clear the `state/vuex/postUpgradeVerification` as we have already published version 0.2.0.

Silver-IT (dismissed):
I don't find any issues too.

corrideat (dismissed):

taoeffect (approved):

PR #1871, opened 2024-02-27
Server signature validation

corrideat:


backend/routes.js
```diff
@@ -41,13 +41,13 @@ route.POST('/event', {
...
+    const { contractID, hash } = GIMessage.deserializeHEAD(request.payload)
     try {
-      await sbp('backend/server/handleEntry', entry)
+      await sbp('backend/server/handleEntry', contractID, request.payload)
     } catch (err) {
-      if (err.name === 'ChelErrorDBBadPreviousHEAD') {
-        console.error(chalk.bold.yellow('ChelErrorDBBadPreviousHEAD'), err)
-        const HEADinfo = await sbp('chelonia/db/latestHEADinfo', entry.contractID()) ?? { HEAD: null, height: 0 }
+      if (err.name === 'ChelErrorDBBadPreviousHEAD' || err.name === 'ChelErrorAlreadyProcessed') {
+        console.error(chalk.bold.yellow(err.name), err)
```
taoeffect: `err` should be logged as first parameter now b/c of Pino

backend/routes.js
```diff
@@ -41,13 +41,13 @@ route.POST('/event', {
...
+    const deserializedHEAD = GIMessage.deserializeHEAD(request.payload)
     try {
-      await sbp('backend/server/handleEntry', entry)
+      await sbp('backend/server/handleEntry', deserializedHEAD, request.payload)
     } catch (err) {
-      if (err.name === 'ChelErrorDBBadPreviousHEAD') {
-        console.error(err, chalk.bold.yellow('ChelErrorDBBadPreviousHEAD'))
-        const HEADinfo = await sbp('chelonia/db/latestHEADinfo', entry.contractID()) ?? { HEAD: null, height: 0 }
+      if (err.name === 'ChelErrorDBBadPreviousHEAD' || err.name === 'ChelErrorAlreadyProcessed') {
+        console.error(err, chalk.bold.yellow(err.name), err)
```
taoeffect: Extraneous `, err)` at the end

frontend/main.js
```diff
@@ -160,11 +165,16 @@ async function startApp () {
...
+        // We re-throw the error to indicate to Chelonia not to proceed
+        throw e
       },
       sideEffectError: (e: Error, message: GIMessage) => {
         sbp('gi.ui/seriousErrorBanner', e)
         sbp('okTurtles.data/set', 'sideEffectError', message.hash())
         errorNotification('sideEffect', e, message)
+        // We used to revert the state and rethrow the error here, but we no longer do that
+        // see this issue for why: https://github.com/okTurtles/group-income/issues/1544
+        // NO 'throw e' here
```
taoeffect, line 177: Nice comment! 👍

shared/domains/chelonia/internals.js
```diff
@@ -1631,9 +1688,13 @@ export default (sbp('sbp/selectors/register', {
...
-        this.config.hooks.processError?.(e, message, getMsgMeta(message, contractID, state))
-        // special error that prevents the head from being updated, effectively killing the contract
-        if (e.name === 'ChelErrorUnrecoverable') throw e
+        try {
+          if (!this.config.hooks.processError) throw e
+          this.config.hooks.processError(e, message, getMsgMeta(message, contractID, state))
+        } catch (e) {
+          // special error that prevents the head from being updated, effectively killing the contract
+          if (e.name === 'ChelErrorUnrecoverable') throw e
+        }
```
taoeffect: This change, and the related `handleEventError` change below, has significantly complicated the code and is now also forcing developers to decide in their hooks how Chelonia's internals should behave. This is a clearly less simple, and seems to add unnecessary complexity to the codebase. Could these changes be reverted?

frontend/model/captureLogs.js
```diff
@@ -61,7 +61,16 @@ function captureLogEntry (type, ...args) {
...
+          JSON.stringify(arg, (_, v) => {
+            if (v instanceof Error) {
+              return {
+                name: v.name,
+                message: v.message,
+                stack: v.stack
+              }
+            }
+            return v
+          })
```
taoeffect, line 73: Why did you do it this way instead of sticking with the previous `?:` method?

frontend/views/components/Avatar.vue
```diff
@@ -31,7 +31,7 @@ export default ({
   },
   mounted () {
     console.log(`Avatar under ${this.$parent.$vnode.tag} blobURL:`, this.blobURL, 'src:', this.src)
-    if (typeof this.src === 'object') {
+    if (this.src && typeof this.src === 'object') {
```
taoeffect, line 35: Why was this change made?

corrideat: This fixes a defect in the previous files PR, since `typeof null === 'object'`.

shared/domains/chelonia/encryptedData.js
```diff
@@ -295,6 +295,7 @@ export const isRawEncryptedData = (data: any): boolean => {
 
 export const unwrapMaybeEncryptedData = (data: any): { encryptionKeyId: string | null, data: any } | void => {
   if (isEncryptedData(data)) {
+    if (process.env.BUILD !== 'web') return
```
taoeffect, line 301: Where is this envvar set? (probably could use a comment mentioning that too)

frontend/model/captureLogs.js
corrideat: This improves the logs by adding more information about errors, and it also captures errors anywhere, not just as an argument. When debugging, I needed to see the message and error name as well as the stack.

backend/routes.js
corrideat: Fixed

shared/domains/chelonia/internals.js
corrideat: I think they could since it seems like it's not needed after all.

shared/domains/chelonia/encryptedData.js
corrideat: It's set in Gruntfile.js

frontend/views/components/Avatar.vue
taoeffect: Great catch!! Could you add a single comment for this above line 34? I feel like a lot of people would get tripped up and confused by this JavaScript quirk.

shared/domains/chelonia/internals.js
```diff
@@ -1295,31 +1299,99 @@ export default (sbp('sbp/selectors/register', {
...
+      return acc
+    }, [Number.POSITIVE_INFINITY, '', []])
+
+    if (keyUpdateArgs.length !== 0) {
+      const contractName = contractState._vm.type
+
+      // This is safe to do without await because it's sending an operation
+      // Using await could deadlock when retrying to send the message
+      sbp('chelonia/out/keyUpdate', { contractID, contractName, data: keyUpdateArgs, signingKeyId: keyUpdateSigningKeyId }).catch(e => {
+        console.error(`[chelonia/private/deleteOrRotateRevokedKeys] Error sending OP_KEY_DEL for ${contractID}`)
```
taoeffect: Should log the `e.message` as part of the string

shared/domains/chelonia/internals.js
```diff
@@ -1335,15 +1407,15 @@ export default (sbp('sbp/selectors/register', {
...
 
-    // This is safe to do without await because it's sending an operation
-    // Using await could deadlock when retrying to send the message
-    sbp('chelonia/out/keyDel', { contractID, contractName, data: keyIds, signingKeyId }).catch(e => {
-      console.error(`[chelonia/private/deleteRevokedKeys] Error sending OP_KEY_DEL for ${contractID}`)
-    })
+      // This is safe to do without await because it's sending an operation
+      // Using await could deadlock when retrying to send the message
+      sbp('chelonia/out/keyDel', { contractID, contractName, data: keyIdsToDelete, signingKeyId: keyDelSigningKeyId }).catch(e => {
+        console.error(`[chelonia/private/deleteRevokedKeys] Error sending OP_KEY_DEL for ${contractID}`)
```
taoeffect: Same here

shared/domains/chelonia/internals.js
```diff
@@ -1677,39 +1757,67 @@ export default (sbp('sbp/selectors/register', {
...
-        }
-        if (!eventsToReinjest.includes(hash)) {
-          console.warn(`[chelonia] WARN bad previousHEAD for ${message.description()}, will attempt to re-sync contract to reinjest message`)
-          eventsToReinjest.push(hash)
-          reprocessDebounced(contractID)
-          return false // ignore the error for now
-        } else {
-          console.error(`[chelonia] ERROR already attempted to reinjest ${message.description()}, will not attempt again!`)
-        }
+      throw new ChelErrorAlreadyProcessed(`Message ${hash} with height ${height} in contract ${contractID} has already been processed. Current height: ${state.contracts[contractID]?.height}.`)
```
taoeffect: DRY: `${state.contracts[contractID]?.height}` => `${latestProcessedHeight}`

shared/domains/chelonia/internals.js
```diff
@@ -1677,39 +1757,67 @@ export default (sbp('sbp/selectors/register', {
...
+    }
+    // If the message is from the future, add it to eventsToReingest
+    if ((latestProcessedHeight + 1) < height) {
+      if (!this.config.reingestEvents) {
+        throw new ChelErrorDBBadPreviousHEAD(`Unexpected message ${hash} with height ${height} in contract ${contractID}: height is too high. Current height: ${state.contracts[contractID]?.height}.`)
```
taoeffect: DRY `${state.contracts[contractID]?.height}` here too

shared/domains/chelonia/signedData.js
```diff
@@ -227,19 +227,14 @@ export const signedIncomingData = (contractID: string, state: ?Object, data: any
...
-        }
-        verifySignedValue = verifySignatureData.call(state || rootStateFn()[contractID], height, data, additionalData)
-        if (mapperFn) verifySignedValue[1] = mapperFn(verifySignedValue[1])
-        return verifySignedValue[1]
-      }
-    : () => {
-        const signedValue = JSON.parse(data._signedData[0])
-        return mapperFn ? mapperFn(signedValue) : signedValue
-      }
+  const verifySignedValueFn = () => {
```
taoeffect, line 229: Should the comment `// TODO: Temporary until the server can validate signatures` be removed now?

backend/server.js
```diff
@@ -119,6 +119,15 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
...
 ;(async function () {
+  await initDB()
+  await sbp('chelonia/configure', {
+    acceptAllMessages: true,
+    skipActionProcessing: true,
+    skipSideEffects: true,
+    reingestEvents: false,
+    strictProcessing: true,
+    strictOrdering: true
+  })
```
taoeffect: Consider adding a variable called `server: true` that inside of `'chelonia/configure'` sets the other variables accordingly, and could also be checked inside of chelonia where something must be done always on the server.

backend/server.js
```diff
@@ -61,17 +62,19 @@ hapi.ext({
...
   },
-  'backend/server/handleEntry': async function (entry: GIMessage) {
-    sbp('okTurtles.data/get', PUBSUB_INSTANCE).channels.add(entry.contractID())
-    await sbp('chelonia/db/addEntry', entry)
-    await sbp('backend/server/broadcastEntry', entry)
+  'backend/server/handleEntry': async function (deserializedHEAD: Object, entry: string) {
+    sbp('okTurtles.data/get', PUBSUB_INSTANCE).channels.add(deserializedHEAD.contractID)
+    await sbp('chelonia/private/in/enqueueHandleEvent', deserializedHEAD.contractID, entry)
+    // Persist the Chelonia state after processing a message
+    await sbp('chelonia/db/set', '_private_cheloniaState', JSON.stringify(sbp('chelonia/private/state')))
```
taoeffect: As this value will get very large, we don't want to write a massive amount of data to disk each and every time we get a new message. So instead, we need to split up the state under multiple keys, one per contract.

EDIT: technically, we already have this key, it's called `head=<contractID>`. We could use this same key for this purpose.

taoeffect (changes requested):
EDIT: wrong PR

backend/server.js
```diff
@@ -62,6 +62,45 @@ hapi.ext({
...
+    // is used when starting up the server to know which keys to fetch.
+    // In the future, consider having a multi-level index, since the index can
+    // get pretty large.
+    if (contractID === deserializedHEAD.hash) {
+      // We want to ensure that the index is updated atomically (i.e., if there
+      // are multiple new contracts, all of them should be added), so a queue
+      // is needed for the load & store operation.
+      await sbp('okTurtles.eventQueue/queueEvent', 'update-contract-indices', async () => {
+        const currentIndex = await sbp('chelonia/db/get', '_private_cheloniaState_index')
+        const updatedIndex = `${currentIndex ? `${currentIndex}\x00` : ''}${contractID}`
```
taoeffect, line 115: What is this `${currentIndex}\x00` weirdness?

(New rule added to Style Guide: [all weirdness must be either eliminated or commented](https://github.com/okTurtles/group-income/blob/master/docs/Style-Guide.md#when-to-write-comments))

backend/server.js
```diff
@@ -62,6 +62,45 @@ hapi.ext({
...
+      return
+    }
+    // If the current HEAD is not what we expect, don't save (the state could
+    // have been updated by a later message). This ensures that we save the
+    // latest state and also reduces the number of write operations
+    if (cheloniaState.contracts[contractID].HEAD === deserializedHEAD.hash) {
+      // Extract the parts of the state relevant to this contract
+      const state = {
+        state: cheloniaState[contractID],
+        contractState: cheloniaState.contracts[contractID]
```
taoeffect: Is `contractState` the right name for this attribute? It sounds like ... the contract state. But this isn't the contract state, it's internal Chelonia metadata about the state isn't it?

Relatedly, `state` sounds a bit vague, maybe that can be renamed to something like `contractVMState`?

backend/server.js
```diff
@@ -125,9 +165,21 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
...
-  if (savedState) {
-    Object.assign(sbp('chelonia/private/state'), JSON.parse(savedState))
+  // First, get the contract index
+  const savedStateIndex = await sbp('chelonia/db/get', '_private_cheloniaState_index')
+  if (savedStateIndex) {
+    // Now, we contract the contract state by reading each contract state
+    // partition
+    const recoveredState = Object.create(null)
+    recoveredState.contracts = Object.create(null)
+    await Promise.all(savedStateIndex.split('\x00').map(async (contractID) => {
```
taoeffect, line 191: It seems we're using this "index" to figure out what contractIDs we have?

Another potential approach would be to update the database API functions to return that info. For example, with FS database, you can do the equivalent of `ls head=*` and that'll return all the contractIDs. For SQlite you can do `SELECT * FROM Data WHERE key LIKE 'head=%'` (or similar), and for in-memory can probably iterate the keys.

backend/server.js
corrideat: I wasn't sure what an appropriate name would be, since they are collectively part of the chelonia state.

backend/server.js
corrideat: It doesn't look weird to me, but I can add a comment. It's just adding to the current index (so the result is `${currentIndex}${separator}${contractID}`, with `separator='\x00'`). Since `currentIndex` could be empty or undefined, and we don't want `undefined${separator}${contractID}`, there's a ternary condition.

backend/server.js
corrideat: I thought that the model was a basic KV store, in which case listing all available keys may not be possible, or if possible, may not be desirable.

As you see in your suggestions:

* For SQLite maybe this would be OK-ish, but still you're reading *all* keys and then filtering them. I'm saying this is OK-ish because maybe internally SQLite can be told to make an index.
* For FS and memory, you have to enumerate *all* keys, process then and then filter them. Doing this is memory and resource intensive, especially if there are many keys, and even more so if the proportion of keys to get out of the total number of keys is low (which would be expected in this case).

taoeffect: Ah, that's a great point. Resolving this.

backend/server.js
taoeffect: Makes sense. Coming from a C background, seeing `\x00` is especially confusing since that would normally terminate the string 😅

corrideat: I see. I come from a C background too, but `\x00` only terminates a string if you're using the various `str*` functions, which normally you shouldn't use. In any case, I chose `\x00` because it's less magic than other values and also it's probably fair to say that contract IDs can't contain it (in a previous iteration of this change, I was using `,${JSON.stringify(cID)}`), but I figured that this would be simpler.

backend/auth.js
```diff
@@ -13,7 +13,7 @@ exports.plugin = {
       return {
         authenticate: function (request, h) {
           const { authorization } = request.headers
-          if (!authorization) h.unauthenticated(Boom.unauthorized('Missing authorization'))
+          if (!authorization) return h.unauthenticated(Boom.unauthorized('Missing authorization'))
```
taoeffect, line 16: Interesting, nice job finding a bug in this old code!

backend/routes.js
```diff
@@ -41,23 +41,29 @@ route.POST('/event', {
...
         Object.assign(r.output.headers, {
           'shelter-headinfo-head': HEADinfo.HEAD,
           'shelter-headinfo-height': HEADinfo.height
         })
         return r
+      } else if (err.name === 'ChelErrorSignatureError') {
+        console.error(err, chalk.bold.yellow(err.name))
+        return Boom.badData('Invalid signature')
+      } else if (err.name === 'ChelErrorSignatureKeyUnauthorized') {
+        console.error(err, chalk.bold.yellow(err.name))
```
taoeffect: Please DRY the repetition of these 3 `console.error(err, chalk.bold.yellow(err.name))` statements into a single one at the start of the `catch` block

shared/domains/chelonia/internals.js
```diff
@@ -1630,6 +1702,9 @@ export default (sbp('sbp/selectors/register', {
         }
         // we revert any changes to the contract state that occurred, ignoring this mutation
         console.warn(`[chelonia] Error processing ${message.description()}: ${message.serialize()}. Any side effects will be skipped!`)
+        if (this.config.strictProcessing) {
+          throw e
+        }
```
taoeffect, line 1707: I'm not sure this is correct.

Users can write contracts that throw errors for whatever reason (intentionally or unintentionally) due to a bug in the contract code, or on purpose. In such cases the codebase has been designed to ignore the mutation, but accept and store the message.

If the server throws however, then what that means is that there are situations where the chain could potentially become unwriteable for an extended period of time. I'm not sure we want to allow this...

What are your thoughts?

EDIT: nevermind, good points:

- The server isn't processing contract code anyway
- Maybe it's a good thing for messages that trigger exceptions to be blocked

backend/server.js
```diff
@@ -61,17 +62,62 @@ hapi.ext({
...
+      const state = {
+        contractState: cheloniaState[contractID],
+        cheloniaContractInfo: cheloniaState.contracts[contractID]
+      }
+      // Save the state under a 'contract partition' key, so that updating a
+      // contract doesn't require saving the entire state.
+      // Although it's not important for the server right now, this will fail to
+      // persist changes the state for other contracts (for example, this happens
+      // when watching keys)
+      await sbp('chelonia/db/set', '_private_cheloniaState_cp_' + contractID, JSON.stringify(state))
```
taoeffect: One more question: the request from before was to use `head=` to store this info, to avoid having to do a few extra writes to the database per contract even update.

This PR doesn't seem to do that, what is the reason why?

Note that `head=` is still being used in addition to the code here, and so the latestHEADInfo is being stored twice...

taoeffect: This comment is a bit unclear...

```js
      // Although it's not important for the server right now, this will fail to
      // persist changes the state for other contracts (for example, this happens
      // when watching keys)
```

What is this saying?

Of course changes for other contracts are not saved here (but they should be saved under their own key), right?

Also, I do not know what "when watching keys" refers to. Can this could be cleared up?

Regarding the key, is `cp` needed?

```js
'_private_cheloniaState_cp_' + contractID
```

How about?

```js
'_private_cheloniaState_' + contractID
```

shared/domains/chelonia/utils.js
```diff
@@ -278,18 +279,16 @@ export const keyAdditionProcessor = function (hash: string, keys: (GIKey | Encry
...
-          decryptedKey = key.meta.private.content.valueOf()
-          decryptedKeys.push([key.id, decryptedKey])
-          storeSecretKey(key, decryptedKey)
-        } catch (e) {
-          console.warn(`Secret key decryption error '${e.message || e}':`, e)
-          // Ricardo feels this is an ambiguous situation, however if we rethrow it will
-          // render the contract unusable because it will undo all our changes to the state,
-          // and it's possible that an error here shouldn't necessarily break the entire
-          // contract. For example, in some situations we might read a contract as
-          // read-only and not have the key to write to it.
```
taoeffect, line 291: Did this situation get cleared up or is this comment still needed?

corrideat: After refactoring, there isn't any exception to re-throw, so I don't think this is relevant.

backend/server.js
corrideat: Done. We can't really use `head=` because it's updated at a different time and it holds different information.

taoeffect (approved):
Amazing work @corrideat! 🎉

shared/domains/chelonia/utils.js
```diff
@@ -278,18 +279,16 @@ export const keyAdditionProcessor = function (hash: string, keys: (GIKey | Encry
...
+        const decryptedKeyResult = unwrapMaybeEncryptedData(key.meta.private.content)
+        // Unable to decrypt
+        if (!decryptedKeyResult) continue
+        // Data aren't encrypted
+        if (decryptedKeyResult.encryptionKeyId == null) {
+          throw new Error('Expected encrypted data but got unencrypted data for key with ID: ' + key.id)
         }
+        decryptedKey = decryptedKeyResult.data
+        decryptedKeys.push([key.id, decryptedKey])
+        storeSecretKey(key, decryptedKey)
```
Silver-IT, line 291: This could be the reason of the following error.

https://okturtles.slack.com/archives/C0EH7P20Y/p1710209299493659

PR #1875, opened 2024-03-01
Check that the username matches the registered name

corrideat:
Missing: namespace registration happening along contract registration. This can probably be done with a header when publishing the event, but I'm holding off on that until #1871 is merged, because it'll likely conflict.

Once the above is done, this PR will address following issues:

* Restrict contract creation to identity contracts
* Restrict name registration to real identity contracts

### DIFF
--- a/frontend/controller/namespace.js
+++ b/frontend/controller/namespace.js
@@ -18,16 +18,24 @@ sbp('sbp/selectors/register', {
       return result
     })
   },
-  'namespace/lookup': (name: string) => {
-    // TODO: should `name` be encodeURI'd?
+  'namespace/lookupCached': (name: string) => {
     const cache = sbp('state/vuex/state').namespaceLookups
     if (name in cache) {
       // Wrapping in a Promise to return a consistent type across all execution
       // paths (next return is a Promise)
       // This way we can call .then() on the result
-      return Promise.resolve(cache[name])
+      return cache[name]
     }
-    return fetch(`${sbp('okTurtles.data/get', 'API_URL')}/name/${name}`).then((r: Object) => {
+  },
+  'namespace/lookup': (name: string) => {
+    const cached = sbp('namespace/lookupCached', name)
+    if (cached) {
+      // Wrapping in a Promise to return a consistent type across all execution
+      // paths (next return is a Promise)
+      // This way we can call .then() on the result
+      return Promise.resolve(cached)
+    }
+    return fetch(`${sbp('okTurtles.data/get', 'API_URL')}/name/${encodeURIComponent(name)}`).then((r: Object) => {
       if (!r.ok) {
         console.warn(`namespace/lookup: ${r.status} for ${name}`)
         if (r.status !== 404) {
@@ -37,6 +45,7 @@ sbp('sbp/selectors/register', {
       }
       return r['text']()
     }).then(value => {
+      const cache = sbp('state/vuex/state').namespaceLookups
       if (value !== null) {
         Vue.set(cache, name, value)
       }
--- a/frontend/controller/namespace.js
+++ b/frontend/controller/namespace.js
@@ -27,13 +27,15 @@ sbp('sbp/selectors/register', {
       return cache[name]
     }
   },
-  'namespace/lookup': (name: string) => {
-    const cached = sbp('namespace/lookupCached', name)
-    if (cached) {
-      // Wrapping in a Promise to return a consistent type across all execution
-      // paths (next return is a Promise)
-      // This way we can call .then() on the result
-      return Promise.resolve(cached)
+  'namespace/lookup': (name: string, { skipCache }: { skipCache: boolean } = { skipCache: false }) => {
+    if (!skipCache) {
+      const cached = sbp('namespace/lookupCached', name)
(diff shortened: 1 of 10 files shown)


frontend/controller/namespace.js
```diff
@@ -18,16 +18,24 @@ sbp('sbp/selectors/register', {
       return result
     })
   },
-  'namespace/lookup': (name: string) => {
-    // TODO: should `name` be encodeURI'd?
+  'namespace/lookupCached': (name: string) => {
```
taoeffect, line 24: Part of the idea behind selectors is that they can be expected to be called from multiple different places potentially.

Is there any possibility that this selector will be called by something other than `'namespace/lookup'`? If not, consider either integrating this code back directly into `'namespace/lookup'`, or turning it into a private function in this file that gets called by `'namespace/lookup'`

frontend/model/contracts/identity.js
```diff
@@ -31,16 +64,12 @@ sbp('chelonia/defineContract', {
...
-              manifestCid: string,
-              downloadParams: optional(object)
-            }))
-          })
+          attributes: attributesType
         })(data)
         const { username } = data.attributes
+        if (!username) {
+          throw new TypeError('A username is required')
+        }
```
taoeffect, line 96: Below this line are the same checks that are now in `validateUsername` - please DRY

frontend/model/contracts/identity.js
```diff
@@ -70,18 +99,44 @@ sbp('chelonia/defineContract', {
         for (const key in initialState) {
           Vue.set(state, key, initialState[key])
         }
+      },
+      async sideEffect ({ contractID, data }) {
+        // Lookup and save the username so that we can verify that it matches
+        const lookupResult = await sbp('namespace/lookup', data.attributes.username)
+        if (lookupResult !== contractID) {
+          console.error(`Mismatched username. The lookup result was ${lookupResult} instead of ${contractID}`)
```
taoeffect: Seems like a big deal, should an exception be thrown to alert the user?

frontend/model/state.js
```diff
@@ -468,7 +474,12 @@ const getters = {
       .forEach(contractID => {
         const attributes = state[contractID].attributes
         if (attributes) { // NOTE: this is for fixing the error while syncing the identity contracts
-          profiles[attributes.username] = { ...attributes, contractID }
+          const username = checkedUsername(state, attributes.username, contractID)
```
taoeffect, line 477: This is kinda odd, that if the check doesn't pass we're still using it, it's just `undefined` now... Is that right?

taoeffect (changes requested):
Nice! Review ready!

frontend/controller/namespace.js
corrideat: Well, the difference is that the new version only checks the local cache, which is synchronous. This can be useful when you need a response immediately (for example, in `state.js`, although there I'm getting the value directly)

frontend/model/contracts/identity.js
corrideat: I think alerts should be used sparingly, and probably not here. This check is expected to fail if the username ever changes, so in reality we'd need to check for the latest username before notifying the user, and then I'd use a notification instead of throwing.

frontend/model/state.js
corrideat: At that point, the username can't be trusted. However, the Vuex state needs to be synchronous.

shared/domains/chelonia/internals.js
```diff
@@ -488,6 +488,7 @@ export default (sbp('sbp/selectors/register', {
           method: 'POST',
           body: entry.serialize(),
           headers: {
+            ...headers,
```
taoeffect, line 491: The line below that reads:

> 'Authorization': 'gi TODO - signature - if needed here - goes here'

Maybe now is a good time to remove it? 🤔 

We have signatures working now, and they've not using headers

test/backend.test.js
```diff
@@ -253,7 +254,7 @@ describe('Full walkthrough', function () {
       users.bob.decryptedValue().data.attributes.email.should.equal('bob@okturtles.com')
     })
 
-    it('Should register Alice and Bob in the namespace', async function () {
+    it.skip('Should register Alice and Bob in the namespace', async function () {
```
taoeffect, line 262: Perhaps this test should be removed entirely now?

frontend/model/contracts/identity.js
```diff
@@ -14,6 +14,63 @@ import {
...
+    if (!has(rootState, contractID)) return
+
+    const username = rootState[contractID].attributes.username
+    if (sbp('namespace/lookupCached', username) !== contractID) {
+      sbp('gi.notifications/emit', 'WARNING', {
+        contractID,
+        message: L('Unable to confirm that the username {username} belongs to this identity contract', { username })
+      })
+    }
+  })
```
taoeffect, line 71: Nicely done 👍

taoeffect (commented):
Nice 👍 

Left a couple of comments

taoeffect (approved):
Nice work @corrideat 👍

PR #1876, opened 2024-03-04
Limit events after

corrideat:
This is a local copy of #1740, to be worked on and updated.


taoeffect (changes requested):
I forgot to mention something important: we want to make sure that the server can never get easily bogged down by a single request.

Let's say there's a chain with a million events. We don't want the server to respond with all million events because of a call to `eventsAfter`.

So there needs to be a server-side hard limit called a `MAX_EVENTS_BATCH_SIZE` (or something to that effect) that limits the number of events that can be returned from any given call (to say, 500 events).

And then the client is then responsible for noticing if not all events that it requested were returned, and if so, make another call to the backend as needed, until it gets all the events it was asking for.

EDIT: also, separately, there should be a separate client-side request limit on the events to be able to restrict the events even further. E.g. clients should be able to say they only want 20 events.

backend/database.js
```diff
@@ -37,38 +39,53 @@ if (!fs.existsSync(dataFolder)) {
...
+    // $FlowFixMe[prop-missing]
+    stream.headers = {
+      'shelter-headinfo-head': latestHEADinfo.HEAD,
+      'shelter-headinfo-height': latestHEADinfo.height
+    }
+    return stream
   },
-  'backend/db/streamEntriesBefore': async function (before: string, limit: number): Promise<*> {
+  'backend/db/streamEntriesBefore': async function (before: string, requestedLimit: number): Promise<*> {
+    let limit = Math.min(requestedLimit, process.env.MAX_EVENTS_BATCH_SIZE ?? Number.POSITIVE_INFINITY)
```
taoeffect: We need to set `MAX_EVENTS_BATCH_SIZE` as a constant in the code, let's set it to 500

taoeffect (changes requested):
Left a comment + another request:

- Can we modify `syncContract` so that it only loads `MAX_EVENTS_BATCH_SIZE = 500` events at a time as it's processing the events? (And calls `eventsAfter` repeatedly to process the whole chain)?

backend/database.js
```diff
@@ -37,38 +39,53 @@ if (!fs.existsSync(dataFolder)) {
   fs.mkdirSync(dataFolder, { mode: 0o750 })
 }
 
+// Streams stored contract log entries since the given entry hash (inclusive!).
 sbp('sbp/selectors/register', {
-  'backend/db/streamEntriesAfter': async function (contractID: string, hash: string): Promise<*> {
+  'backend/db/streamEntriesAfter': async function (contractID: string, hash: string, requestedLimit: number = MAX_EVENTS_AFTER): Promise<*> {
+    const limit = Math.min(requestedLimit, process.env.MAX_EVENTS_BATCH_SIZE ?? 500)
```
taoeffect: It seems to me like `MAX_EVENTS_AFTER` isn't really relevant anymore in the face of this `Math.min` and `process.env.MAX_EVENTS_BATCH_SIZE ?? 500`, as that will always be less than `Infinity`.

shared/domains/chelonia/chelonia.js
```diff
@@ -21,6 +21,8 @@ import './internals.js'
 import './files.js'
 import { findForeignKeysByContractID, findKeyIdByName, findRevokedKeyIdsByName, findSuitableSecretKeyId, getContractIDfromKeyId } from './utils.js'
 
+const MAX_EVENTS_AFTER = Number.parseInt(process.env.MAX_EVENTS_AFTER, 10) || Infinity
```
taoeffect: 1. Can we use the same `MAX_EVENTS_BATCH_SIZE` here?
2. `10` seems rather small... maybe `500`?

shared/domains/chelonia/chelonia.js
```diff
@@ -829,30 +831,199 @@ export default (sbp('sbp/selectors/register', {
...
-      .then(handleFetchResult('json'))
-    if (Array.isArray(events)) {
-      return events.reverse().map(b64ToStr)
+  'chelonia/out/eventsAfter': function (contractID: string, since: string, limit?: number) {
+    const fetchEventsStreamReader = async () => {
+      requestLimit = Math.min(limit ?? MAX_EVENTS_AFTER, remainingEvents)
+      const eventsResponse = await fetch(`${this.config.connectionURL}/eventsAfter/${contractID}/${since}/${requestLimit}`, { signal })
+      if (!eventsResponse.ok) throw new Error('Unexpected status code')
+      if (!eventsResponse.body) throw new Error('Missing body')
+      latestHEAD = eventsResponse.headers.get('shelter-headinfo-head')
```
taoeffect: Should we check the type on this also?

shared/domains/chelonia/chelonia.js
```diff
@@ -829,30 +831,199 @@ export default (sbp('sbp/selectors/register', {
...
+      requestCount++
+      // $FlowFixMe[incompatible-use]
+      return eventsResponse.body.getReader()
     }
+    const signal = this.abortController.signal
+    let requestCount = 0
+    let remainingEvents = limit ?? Number.POSITIVE_INFINITY
```
taoeffect: This is using `limit`, should it be using `requestLimit` instead? By using `limit` or Infinity, isn't it not accurately representing the number of remaining events?

backend/database.js
corrideat: `MAX_EVENTS_AFTER` is a client-side constraint, while `MAX_EVENTS_BATCH_SIZE` is a server-side constraint.

shared/domains/chelonia/chelonia.js
corrideat: `10` is _not_ a default value, it's the radix (base).

shared/domains/chelonia/chelonia.js
```diff
@@ -829,30 +831,199 @@ export default (sbp('sbp/selectors/register', {
...
+                break
+              }
+              let enqueued = false
+              try {
+                // Extract the current event's value and trim whitespace
+                const eventValue = buffer.slice(0, nextIdx).trim()
+                if (eventValue) {
+                  // Check if the event limit is reached; if so, throw an error
+                  if (count === requestLimit) {
+                    controller.error(new Error('Received too many events'))
```
taoeffect: Should this be a special error that's treated differently?

shared/domains/chelonia/chelonia.js
taoeffect: Oh, ok, then ignore the 2nd question 😄

shared/domains/chelonia/chelonia.js
corrideat: We could, but if it's not set or incorrect, it should fail later here:

```
if (remainingEvents === 0 || since === latestHEAD) {
```

shared/domains/chelonia/chelonia.js
corrideat: No, because this is remaining events for the entire stream, not just this request. If a limit was set when calling the function, that's the max number of events returned (whether that's one request or multiple). OTOH, if the limit is not set, effectively we're asking for _all_ events, which is unbounded, hence `Infinity`.

shared/domains/chelonia/chelonia.js
corrideat: Special how?

backend/database.js
corrideat: Yes, we can remove it from the server.

shared/domains/chelonia/chelonia.js
taoeffect: I'm not sure what happens when an error is thrown, hopefully the whole state we've already built up by processing the events doesn't get discarded. If it does get discarded, then detecting a special error could be a way to prevent it from being discarded. (I'm guessing the state isn't thrown out though, so my comment is probably unnecessary)

backend/routes.js
```diff
@@ -93,6 +93,27 @@ route.GET('/eventsAfter/{contractID}/{since}', {}, async function (request, h) {
   }
 })
 
+route.GET('/events/{contractID}/height/{height}', {
```
taoeffect: Can you please rename this API to something clearer, for example:

```
route.GET('/eventHashAtHeight/{contractID}/{height}', ...
```

shared/domains/chelonia/internals.js
```diff
@@ -1124,26 +1126,29 @@ export default (sbp('sbp/selectors/register', {
       if (latest !== recent) {
         console.debug(`[chelonia] Synchronizing Contract ${contractID}: our recent was ${recent || 'undefined'} but the latest is ${latest}`)
         // TODO: fetch events from localStorage instead of server if we have them
-        const events = await sbp('chelonia/out/eventsAfter', contractID, recent || contractID)
+        const eventsStream = sbp('chelonia/out/eventsAfter', contractID, recent || contractID)
```
taoeffect: I think the code for `'chelonia/latestContractState'` also needs to be updated to use this this new method, right?

taoeffect (approved):
Well @corrideat, I'm not sure what to say except that, this is a rather bad-ass PR. 😎 

"Bad-ass" as in very good.

👏

PR #1877, opened 2024-03-05
Improve behavior of sending messages with attachments

### LINKED ISSUE #1845: Improve behavior of sending messages with large attachments
### Problem

Let's say a user is on a slow connection and wants to share a photo with a friend.

They take this photo, attach it to a message in the chat, and hit send.

We want to make sure the UI is not locked up while waiting for the image to upload, and we want them to also be able to continue sending other messages in the meantime. We need to update both the code and the UI for handling this gracefully.

### Solution

We're going to discuss a scenario involving **message1**: a message with a large image attachment, and **message2**: a followup message without any attachments.

1. When the user sends **message1**, we need to begin the image upload process in the background, and visually communicate to the user that the image is being uploaded by displaying it in a greyed out way with a progress bar superimposed over the image in the chat, along with grayed out text that accompanies the image.
2. If the user types in the additional **message2** and hits send,  it should be sent out immediately while the other image is still uploading. It will also appear in grey after **message1**. When the user's client receives back **message2** (before **message1** has finished uploading), it needs to move **message1** _below **message2**_! This way, the user visually sees that even though **message1** was sent first, it hasn't actually made it to the other users, and will appear to them after **message2** in the timeline. This way the chat timeline will always show everyone the same order of messages.
3. Once the file attachment for **message1** finishes uploading, *then* it can be actually sent to the server, referencing the file attachment so that if it's an image or a video, it displays inline in the viewer.

See also companion issue #1846 for implementing file upload in chat.

### LINKED ISSUE #1846: Implement uploading file attachments in chat
### Problem

We can attach files to the chat, but they aren't actually uploaded.

### Solution

1. Upload the file attachment.
2. After it's uploaded, send the message, and reference the file attachment so that it's displayed inline along with the message.

See also companion issue #1845 which gives more detail about how this should be done.

### LINKED ISSUE #1873: Chat page crashed after file attachment
### Problem

**Steps to reproduce**
- user creates a group, and opens the chat page
- user clicks the `Attachment` button, and attach an image
- user sees the image in the SendArea, and tries to drag it
image
- the chat page is crashed

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1845 
- Closes #1846
- Closes #1873

### DIFF
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -2,10 +2,10 @@
 
 import {
   objectOf, objectMaybeOf, arrayOf, unionOf,
-  string, optional, number, mapOf, literalOf
+  object, string, optional, number, mapOf, literalOf
 } from '~/frontend/model/contracts/misc/flowTyper.js'
 import {
-  CHATROOM_TYPES, CHATROOM_PRIVACY_LEVEL, ATTACHMENT_TYPES,
+  CHATROOM_TYPES, CHATROOM_PRIVACY_LEVEL,
   MESSAGE_TYPES, MESSAGE_NOTIFICATIONS, PROPOSAL_VARIANTS
 } from './constants.js'
 
@@ -46,16 +46,10 @@ export const messageType: any = objectMaybeOf({
   }),
   attachments: arrayOf(objectOf({
     name: string,
-    extension: string,
     mimeType: string,
-    attachmentId: string,
-    attachType: unionOf(...Object.values(ATTACHMENT_TYPES).map(v => literalOf(v))),
     downloadData: objectOf({
       manifestCid: string,
-      downloadParams: objectOf({
-        IKM: string,
-        rs: number
-      })
+      downloadParams: optional(object)
     })
   })),
   replyingMessage: objectOf({
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -44,6 +44,12 @@ export const messageType: any = objectMaybeOf({
     type: unionOf(...Object.values(MESSAGE_NOTIFICATIONS).map(v => literalOf(v))),
     params: mapOf(string, string) // { username }
   }),
+  attachments: arrayOf(objectOf({
+    attachType: string,
+    extension: string,
+    name: string,
+    url: string
+  })),
   replyingMessage: objectOf({
     hash: string, // scroll to the original message and highlight
     text: string // display text(if too long, truncate)
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -5,7 +5,7 @@ import {
   string, optional, number, mapOf, literalOf
 } from '~/frontend/model/contracts/misc/flowTyper.js'
 import {
-  CHATROOM_TYPES, CHATROOM_PRIVACY_LEVEL,
+  CHATROOM_TYPES, CHATROOM_PRIVACY_LEVEL, ATTACHMENT_TYPES,
   MESSAGE_TYPES, MESSAGE_NOTIFICATIONS, PROPOSAL_VARIANTS
 } from './constants.js'
 
@@ -47,16 +47,16 @@ export const messageType: any = objectMaybeOf({
   attachments: arrayOf(objectOf({
     name: string,
     extension: string,
-    memeType: string,
+    mimeType: string,
     attachmentId: string,
-    attachType: string,
-    downloadData: {
+    attachType: unionOf(...Object.values(ATTACHMENT_TYPES).map(v => literalOf(v))),
+    downloadData: objectOf({
       manifestCid: string,
-      downloadParams: {
+      downloadParams: objectOf({
         IKM: string,
         rs: number
-      }
-    }
+      })
+    })
   })),
   replyingMessage: objectOf({
     hash: string, // scroll to the original message and highlight
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -45,10 +45,18 @@ export const messageType: any = objectMaybeOf({
     params: mapOf(string, string) // { username }
   }),
   attachments: arrayOf(objectOf({
-    attachType: string,
-    extension: string,
     name: string,
-    url: string
+    extension: string,
+    memeType: string,
+    attachmentId: string,
+    attachType: string,
+    downloadData: {
+      manifestCid: string,
+      downloadParams: {
+        IKM: string,
+        rs: number
+      }
+    }
   })),
   replyingMessage: objectOf({
     hash: string, // scroll to the original message and highlight
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -9,7 +9,7 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
   'aac', 'cda', 'mid', 'midi', 'mp3', 'oga', 'opus', 'wav', 'weba', // audio
   'avi', 'mov', 'mp4', 'mpeg', 'ogv', 'ogx', 'ts', 'webm', '3gp', '3g2', // video
   'arc', 'bz', 'bz2', 'epub', 'gz', 'jar', 'rar', 'tar', 'zip', '7z', // archive
-  'csh', 'css', 'csv', 'doc', 'docx', 'htm', 'html', 'ics', 'js', 'json', 'jsonld', 'md', 'mjs',
+  'csh', 'css', 'csv', 'doc', 'docx', 'htm', 'html', 'ics', 'js', 'json', 'jsonld', 'md', 'mjs', 'vue', 'ts', 'scss',
   'odp', 'ods', 'odt', 'pdf', 'php', 'ppt', 'pptx', 'sh', 'txt', 'xhtml', 'xls', 'xlsx', 'xml', // common text/document
   'eot', 'otf', 'rtf', 'ttf', 'woff', 'woff2', // font
   'abw', 'azw', 'bin', 'csh', 'db', 'mpkg', 'vsd', 'xul' // miscellaneous
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -14,3 +14,7 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
   '.eot', '.otf', '.rtf', '.ttf', '.woff', '.woff2', // font
   '.abw', '.azw', '.bin', '.csh', '.db', '.mpkg', '.vsd', '.xul' // miscellaneous
 ]
+
+// NOTE: Below value was obtained from the '413 Payload Too Large' server error
+//       meaning if this limit is updated on the server-side, an update is required here too.
+export const CHAT_ATTACHMENT_SIZE_LIMIT = 6291456 // in byte.
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -5,16 +5,17 @@
 
 export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
   // reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
-  '.avif', '.bmp', '.gif', '.ico', '.jpeg', '.jpg', '.png', '.svg', '.tif', '.tiff', '.webp', // image
-  '.aac', '.cda', '.mid', '.midi', '.mp3', '.oga', '.opus', '.wav', '.weba', // audio
-  '.avi', '.mov', '.mp4', '.mpeg', '.ogv', '.ogx', '.ts', '.webm', '.3gp', '.3g2', // video
-  '.arc', '.bz', '.bz2', '.epub', '.gz', '.jar', '.rar', '.tar', '.zip', '.7z', // archive
-  '.csh', '.css', '.csv', '.doc', '.docx', '.htm', '.html', '.ics', '.js', '.json', '.jsonld', '.md', '.mjs',
-  '.odp', '.ods', '.odt', '.pdf', '.php', '.ppt', '.pptx', '.sh', '.txt', '.xhtml', '.xls', '.xlsx', '.xml', // common text/document
-  '.eot', '.otf', '.rtf', '.ttf', '.woff', '.woff2', // font
-  '.abw', '.azw', '.bin', '.csh', '.db', '.mpkg', '.vsd', '.xul' // miscellaneous
+  'avif', 'bmp', 'gif', 'ico', 'jpeg', 'jpg', 'png', 'svg', 'tif', 'tiff', 'webp', // image
+  'aac', 'cda', 'mid', 'midi', 'mp3', 'oga', 'opus', 'wav', 'weba', // audio
+  'avi', 'mov', 'mp4', 'mpeg', 'ogv', 'ogx', 'ts', 'webm', '3gp', '3g2', // video
+  'arc', 'bz', 'bz2', 'epub', 'gz', 'jar', 'rar', 'tar', 'zip', '7z', // archive
+  'csh', 'css', 'csv', 'doc', 'docx', 'htm', 'html', 'ics', 'js', 'json', 'jsonld', 'md', 'mjs',
+  'odp', 'ods', 'odt', 'pdf', 'php', 'ppt', 'pptx', 'sh', 'txt', 'xhtml', 'xls', 'xlsx', 'xml', // common text/document
+  'eot', 'otf', 'rtf', 'ttf', 'woff', 'woff2', // font
+  'abw', 'azw', 'bin', 'csh', 'db', 'mpkg', 'vsd', 'xul' // miscellaneous
 ]
 
 // NOTE: Below value was obtained from the '413 Payload Too Large' server error
 //       meaning if this limit is updated on the server-side, an update is required here too.
+// TODO: fetch this value from a server API
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 6291456 // in byte.
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -118,7 +118,6 @@ import { mapGetters } from 'vuex'
 import { Vue } from '@common/common.js'
 import Avatar from '@components/Avatar.vue'
 import InfiniteLoading from 'vue-infinite-loading'
-import Loading from '@components/Loading.vue'
 import Message from './Message.vue'
 import MessageInteractive from './MessageInteractive.vue'
 import MessageNotification from './MessageNotification.vue'
@@ -212,7 +211,6 @@ export default ({
     ConversationGreetings,
     Emoticons,
     InfiniteLoading,
-    Loading,
     Message,
     MessageInteractive,
     MessageNotification,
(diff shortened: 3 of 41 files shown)


Silver-IT (commented):
Review ready!!!

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -41,7 +41,7 @@
         )
 
         p.c-text(v-else-if='text')
-          template(v-for='(objText, index) in textObjects')
+          template(v-for='objText in textObjects')
```
SebinSong, line 44: `key` attribute here (for`v-for`) seems to be missing.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -600,9 +597,9 @@ export default ({
         const lastDotIndex = name.lastIndexOf('.')
         return lastDotIndex === -1 ? '' : name.substring(lastDotIndex).toLowerCase()
       }
-      const attachmentsExist = Boolean(this.ephemeral.attachment.length)
+      const attachmentsExist = Boolean(this.ephemeral.attachments.length)
```
SebinSong: (This is something I missed in my own work but), `attachmentsExist` here can be replaced with the computed prop `this.hasAttachments` defined above.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -600,9 +597,9 @@ export default ({
...
       const list = appendItems && attachmentsExist
-        ? [...this.ephemeral.attachment]
+        ? [...this.ephemeral.attachments]
         : []
 
       if (attachmentsExist) {
```
SebinSong: `this.hasAttachments` can be used here too

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
```diff
@@ -1,46 +1,162 @@
...
+      const blob = await sbp('chelonia/fileDownload', attachment.downloadData)
+      const url = URL.createObjectURL(blob)
+      return { blob, url }
+    },
+    async downloadAttachment (attachment) {
+      if (!attachment.downloadData) { return }
+
+      // reference: https://blog.logrocket.com/programmatically-downloading-files-browser/
+      const { name, attachmentId } = attachment
+      this.isPreparingDownload = [...this.isPreparingDownload, attachmentId]
```
SebinSong: The initial intention of me creating `this.isPreparingDownload` array here is for below two possible uses

- To give visual indicator(blurring or darkening etc..) that  the attachment item that user has just clicked on is being prepared for downloading
- To prevent from calling `sbp('chelonia/fileDownload')` multiple times on the same item while the blob is loading from the server.

If you think this is not needed in your work on graceful handling of large size message, pls go ahead and remove this list.

SebinSong:
Great work @Silver-IT!

your update on `ChatAttachmentPreview.vue` (preparing the attachment download in the background as soon as the message loads) seems to have improved my original work there.

I just tried this PR locally and attaching some images to the chat



and it leads to this type validation error in the console.



Could you make additional update for this fix?

frontend/utils/constants.js
```diff
@@ -14,3 +14,7 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
   '.eot', '.otf', '.rtf', '.ttf', '.woff', '.woff2', // font
   '.abw', '.azw', '.bin', '.csh', '.db', '.mpkg', '.vsd', '.xul' // miscellaneous
 ]
+
+// NOTE: Below value was obtained from the '413 Payload Too Large' server error
+//       meaning if this limit is updated on the server-side, an update is required here too.
+export const CHAT_ATTACHMENT_SIZE_LIMIT = 6291456 // in byte.
```
taoeffect: Can you please add a TODO comment along these lines?

```js
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -386,59 +386,127 @@ export default ({
...
-          // messages as pending in the UI
-          prepublish: (message) => {
+        data = { ...data, replyingMessage }
+      }
+
+      const sendMessage = (beforePrePublish) => {
+        const prepublish = (message) => {
+          if (!this.checkEventSourceConsistency(contractID)) return
+
+          if (beforePrePublish && typeof beforePrePublish === 'function') {
```
taoeffect: OK to simplify this check to just `typeof beforePrePublish === 'function'`

But even better I think would be to replace the entire `if` block with:

```js
beforePrePublish?.()
```

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -615,38 +612,41 @@ export default ({
...
-        if (fileSize > Math.pow(10, 9)) {
-          // TODO: update Math.pow(10, 9) above with the value delivered from the server once it's implemented there.
+        if (fileSize > CHAT_ATTACHMENT_SIZE_LIMIT) {
           return sbp('okTurtles.events/emit', OPEN_MODAL, 'ChatFileAttachmentWarningModal', { type: 'large' })
         } else if (!fileExt || !CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS.includes(fileExt)) {
           // Give users a warning about unsupported file types
           return sbp('okTurtles.events/emit', OPEN_MODAL, 'ChatFileAttachmentWarningModal', { type: 'unsupported' })
         }
 
+        // !@#
```
taoeffect: What is this comment?

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -615,38 +612,41 @@ export default ({
...
         list.push({
           url: fileUrl,
           name: file.name,
           extension: fileExt,
-          attachType: file.type.match('image/') ? 'image' : 'non-image'
+          mimeType: file.type || '',
+          attachmentId: randomHexString(15),
```
taoeffect: @corrideat thoughts on this `attachmentId`? Is it fine? Is it long enough? Should it be something else?

frontend/views/containers/chatroom/file-attachment/ChatFileAttachmentWarningModal.vue
```diff
@@ -32,7 +32,7 @@ export default {
         'large': {
           title: L('File too large'),
           // TODO: replace '1 GB' below with a value delivered from the server once implemented.
-          content: L('That file is too large and cannot be uploaded. The limit is {sizeLimit}', { sizeLimit: L('1 GB') })
+          content: L('That file is too large and cannot be uploaded. The limit is {sizeLimit}', { sizeLimit: L('6.29 MB') })
```
taoeffect: A few issues here:

1. 6.29 MB is a constant that's specified right here. It should be retrieved from somewhere else.
2. If you're trying to translate "MB", then translate it like so: `L('{sizeInMegabytes} MB', { sizeInMegabytes })`
3. But don't do that. It makes no sense to have two L functions, one inside the other. Instead, do it like this: 

```js
L('That file is too large and cannot be uploaded. The limit is {sizeLimit} MB', { sizeLimit })
```

taoeffect (changes requested):
Nice work @Silver-IT! My review is ready for you, see comments below.

Also, while testing, I encountered two additional issues:

### PDF attachments don't work

Screenshot 2024-03-07 at 7 30 58 PM

### Image upload doesn't work

Trying to upload a JPG:

Screenshot 2024-03-07 at 7 31 26 PM

frontend/model/contracts/shared/types.js
```diff
@@ -44,6 +44,20 @@ export const messageType: any = objectMaybeOf({
...
+    attachmentId: string,
+    attachType: unionOf(...Object.values(ATTACHMENT_TYPES).map(v => literalOf(v))),
+    downloadData: objectOf({
+      manifestCid: string,
+      downloadParams: objectOf({
+        IKM: string,
+        rs: number
+      })
+    })
+  })),
```
taoeffect: It seems like a few of these fields aren't necessary here and should be removed:

1. `extension` - this is derived from `name`, so to DRY things it should be removed
2. `attachmentId` - as far as I can tell, this was used for local ephemeral data for Vuejs, and shouldn't be stored in the chain. There is also already `manifestCid` which is what should be used to uniquely identify the file
3. `attachType` - this is derived from `mimeType` and therefore to DRY things should be removed

Finally there is `downloadParams`, according to @corrideat the type of this object can change depending on the ciphers used, so its type should be `optional(object)`.

frontend/views/containers/chatroom/SendArea.vue
taoeffect: I had a conversation with @corrideat about this and he says that it seems to be fine here as it's just used for identifying file previews in the UI as ephemeral data. However it shouldn't be included in the chain ([see this comment](https://github.com/okTurtles/group-income/pull/1877#discussion_r1515958005)), and to uniquely identify files, the `manifestCid` is what should be used.

frontend/views/containers/chatroom/file-attachment/ChatFileAttachmentWarningModal.vue
SebinSong: > It should be retrived from somewhere else

I believe `CHAT_ATTACHMENT_SIZE_LIMIT` in `utils/constants.js` can be used for now for this.

frontend/views/containers/chatroom/MessageBase.vue
Silver-IT: I only removed unused parameter `index` here. And also we could not use `:key` attribute here since the content is simple. But it's fine to add it too.

Silver-IT:
```
[plugin vue] <template> cannot be keyed. Place the key on real elements instead.
```

I tried to add `key` attribute to the template element and got this error.

taoeffect: Try a `<div>`?

Silver-IT (commented):
Final review ready!

taoeffect (changes requested):
Nicely done @Silver-IT.

I found a few additional issues:

### Two issues related to hover behavior

Screenshot 2024-03-08 at 7 24 26 PM

1. There is a button to delete the file. This problematic for two reasons:
  1. The button should not be shown to users who do not own the file.
  2. The button shouldn't be shown at all right now because deleting files isn't implemented here.
2. Hovering over file turns the cursor into a hand icon, indicating that an action can be performed when clicking the file/image. However, clicking does nothing. Therefore, either an action should be performed, or the cursor should remain as a cursor and not a hand.

### Testing large vertical image and PNGs

I tried to test the behavior of uploading a large vertical image to see if max-height was properly implemented, but it didn't work at all.

Screenshot 2024-03-08 at 7 28 17 PM

I've attached the image I tried to upload here: [Cowspiracy-Infographic.png.zip](https://github.com/okTurtles/group-income/files/14536964/Cowspiracy-Infographic.png.zip) (I've zipped it up to ensure that Github doesn't convert it to some other format. Please unzip the zip file before testing)

In the console I see this error, but the image isn't corrupt...

Screenshot 2024-03-08 at 7 34 45 PM

Incidentally, I was unable to upload any PNG images.

Silver-IT:
Thanks for the review, @taoeffect.
- For the`Delete` button, yes, we need to hide it if the current user is not message sender. And I recommend to make alert `'TODO - Delete Attachment'`, and do not hide it for the message senders.
- For the uploading large portrait images or PNG files, it worked well in Google Chrome. Having tested in Firefox, uploading PNG files was working but the example file `Cowspiracy-Infographic.png` (large portrait image) was not working. I will fix it soon.

Silver-IT:
Having tested more about the the error message "Image corrupt or truncated", the main reason is because the file is damaged. This issue is what @SebinSong already reported [here](https://okturtles.slack.com/archives/G01Q9LZLV0S/p1709618617368679).

I guess it's related to the browser because Google Chrome doesn't raise that issue but Firefox and Safari do.
However, I will investigate more.

taoeffect:
@Silver-IT 

> Having tested more about the the error message "Image corrupt or truncated", the main reason is because the file is damaged. This issue is what @SebinSong already reported [here](https://okturtles.slack.com/archives/G01Q9LZLV0S/p1709618617368679).

Sorry, I can't tell what exactly in Sebin's video shows the same behavior as here. In Sebin's video the images seem to upload with no problem.

Also, I don't believe the problem is that the image is corrupted, for a few reasons:

1. The image opens fine in an image viewer
2. I tried other PNGs and got the same error. They all open fine in image viewers too. It's unlikely that they're also corrupted.

Silver-IT:
@taoeffect, it could seem to be different, but I think it's same in basic. It depends on how much the image is damaged while uploading. Also I can say that not all the images are damaged while uploading. Some are damaged, and some are not.

I attached the three screenshots below. Look at the second image which I got in Google Chrome browser. Two images are uploaded in Firefox browser, and they are however opens fine in Google Chrome. That's why I can say it depends on how much the image is damaged.

image

taoeffect:
@Silver-IT If the images upload fine in Google Chrome, and then display fine in Firefox, it clearly means the images are not damaged.

This must be some other problem. Please search online to see what the problem could possibly be. Try asking Claude 3. Or maybe @SebinSong or @corrideat can help?

corrideat:
@Silver-IT I've written a small function for testing uploads / downloads, and it seems like the `fileUpload` / `fileDownload` selectors are working fine.

```
async function test() { const x = new Uint8Array((1<<20) + ((Math.random()*512) | 0)); x.forEach((_, i) => x[i] = Math.random()*256|0); console.log('Uploading'); const y = await sbp('chelonia/fileUpload', new Blob([x]), { cipher: 'aes256gcm' }); console.log('Uploaded'); return JSON.stringify(Array.from(new Uint8Array(await ((await sbp('chelonia/fileDownload', y)).arrayBuffer())))) === JSON.stringify(Array.from(x)); }
```

corrideat:
If I had to guess, I'd think that it's got to do with the `fetch` used to convert back from an object URL to a blob, although I could be wrong. I'd try the following:

1. Keep a reference to a `Blob` (or `File`, which inherits from `Blob`) around until the upload is complete.
2. Use that reference directly
3. Also, this allows you to immediately revoke the object URL after the preview loads, which can be good for performance, because it allows for the object to be garbage collected.

Silver-IT:
@corrideat, thanks for your kind help. I also think that `fileUpload` / `fileDownload` works fine. If it is wrong, it must not be working in Google Chrome browser either.

I also tested keeping the Blob reference until the upload is finished and even more. That Blob (which is used for uploading, and not the one which is created by `fileDownload`) seems to keep the right content because it doesn't shows the wrong image in the chat page and also I can download the correct image using its ObjectURL.

At the moment, I couldn't guess the which one is making this kinda weird problem.

corrideat:
>  thanks for your kind help. I also think that fileUpload / fileDownload works fine. If it is wrong, it must not be working in Google Chrome browser either.

It _could_ be an issue that's only visible in Firefox. Besides being different implementations, the code is also different. For example, when uploading Chrome may use a stream and Firefox will not. In reality, using a stream in Chrome requires HTTP 2, so unless you've done something to enable it, the code should be the same (although, if you want to be sure, you can disable the check here: https://github.com/okTurtles/group-income/blob/64e8a2b0892cddc6898b10cf16ce272f8fa86d37/shared/domains/chelonia/files.js#L52C3-L52C41, which will make Chrome and Firefox use the same logic). I don't think that this will help, but you can try.

The next thing you can do is upload a file that gets corrupted, download it, and then inspect the original and the downloaded version side-by-side. That could give some information about what's happening. Since the issue happens consistently when showing a file, my guess is that something goes wrong during the _upload_.

taoeffect:
Maybe somehow the image isn't being drawn or displayed incorrectly?

I ask because @corrideat [claims](https://github.com/okTurtles/group-income/pull/1877#issuecomment-1988597746) that the files are being uploaded and downloaded correctly on Firefox (the same byte sequence is being received...)

SebinSong:
@taoeffect , @Silver-IT 
My first guess when I ran into this issue was the same as below comment of @corrideat 

> If I had to guess, I'd think that it's got to do with the fetch used to convert back from an object URL to a blob

I've just done some research and it seems like instead of using `URL.createObjectURL` for the browser to convert/download `Blob` we receive via `sbp('chelonia/fileDownload')`, we can also use `FileReader.readAsDataURL()` API too, even though apparently the solution will look more verbose.

Let me try this in my local dev and share the result of how it goes.

SebinSong:
@taoeffect @Silver-IT
I just created this method that uses `FileReader()` API and tried downloading using it but got the same issue.
Now it feels like the cause of the issue isn't related to the way front-end handles `blob`.

```js
async getDateURLOfAttachment (attachment) {
    const blob = await sbp('chelonia/fileDownload', attachment.downloadData)
    const reader = new FileReader()

      reader.onload = () => { resolve(reader.result) }
      reader.readAsDataURL(blob)
    })
  }
```

This is the before / after of the image I had an issue with, when trying in Firefox Dev Edition.

**[ before ]**


**[ after ]**

Silver-IT:
Having tested this issue more in different browsers, I can say it works fine in Google Chrome and Brave browser. And in Opera browser, I can not even upload any files, and in other browsers like Firefox and Safari, the file is damaged.

Based on my poor understanding of this kind of area, I am not sure but my opinion is that the problem is related to this one.
> It could be an issue that's only visible in Firefox. Besides being different implementations, the code is also different. For example, when uploading Chrome may use a stream and Firefox will not. In reality, using a stream in Chrome requires HTTP 2, so unless you've done something to enable it, the code should be the same (although, if you want to be sure, you can disable the check here: https://github.com/okTurtles/group-income/blob/64e8a2b0892cddc6898b10cf16ce272f8fa86d37/shared/domains/chelonia/files.js#L52C3-L52C41, which will make Chrome and Firefox use the same logic). I don't think that this will help, but you can try.

cc: @taoeffect, @corrideat, @SebinSong

corrideat:
@Silver-IT: I've managed to find the issue. It was an error in the encryption function, that resulted in sometimes (depending on how the input was chunked) skipping over some data. This has been fixed in https://github.com/Exact-Realty/ts-rfc8188/pull/5 and I've also pushed a new commit that updates the package and addresses this issue.

taoeffect (changes requested):
Nice work @Silver-IT & @corrideat!

@Silver-IT I tested this to see if #1845 is closed, and with throttling enabled we can see that the issue isn't closed:

https://github.com/okTurtles/group-income/assets/138706/41579a6d-d7c4-427f-a368-c5c03ed2220a

For whatever reason, the messages sent after the file upload starts don't get sent immediately. They need to be sent immediately instead of waiting for the file upload to finish or fail.

Also, as a separate unrelated issue (but something to not forget), if we are on a cellular connection (or a slow connection), we don't want to autodownload the files. Instead, a download button overlay should appear over the file, and only if it's clicked should the file preview be downloaded. But we can save that for a separate issue and PR.

Silver-IT:
@taoeffect, I think the delay is because of that throttle works for all the network transactions. Not only for file uploading.
Also based on the record video, I updated the code to handle error which is caught when uploading attachment could be failed.

taoeffect (approved):
Approved!

Great work @Silver-IT, @SebinSong and @corrideat!

It looks like the issue with messages not appearing when they should when throttling is turned on is a bug in Firefox, not in this code, as it works correctly in Brave with throttling enabled.

SebinSong (approved):
Great work @Silver-IT !

PR #1882, opened 2024-03-10
Two minor errors while logging out and creating group

### LINKED ISSUE #1880: Minor errors regarding file downloading and navigation
### Problem

While logging out, I can see this error in console. I could because of updated workflow of downloading files.
image

While creating a new group after signing up, I got this error.
image

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1880

### DIFF
--- a/frontend/views/components/Avatar.vue
+++ b/frontend/views/components/Avatar.vue
@@ -81,7 +81,9 @@ export default ({
   },
   watch: {
     src (to) {
-      if (typeof to === 'object') {
+      // NOTE: src could be null while logging out
+      //       since typeof null === 'object', we should check if it's falsy
+      if (to && typeof to === 'object') {
         this.downloadFile(to).catch((e) => {
           console.error('[Avatar.vue] Error in downloadFile', e)
         })
--- a/frontend/views/pages/Home.vue
+++ b/frontend/views/pages/Home.vue
@@ -61,11 +61,11 @@ main.c-splash(data-test='homeLogo' v-if='!$store.state.currentGroupId')
 <script>
 import sbp from '@sbp/sbp'
 import { mapGetters, mapState } from 'vuex'
-import VueRouter from 'vue-router'
 import { OPEN_MODAL } from '@utils/events.js'
 import BannerSimple from '@components/banners/BannerSimple.vue'
 import SvgCreateGroup from '@svgs/create-group.svg'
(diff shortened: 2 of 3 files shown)


frontend/views/components/Avatar.vue
```diff
@@ -82,6 +82,8 @@ export default ({
   },
   watch: {
     src (to) {
+      // NOTE: src could be null while logging out
+      //       since typeof null === 'object', we should check if it's falsy
       if (to && typeof to === 'object') {
```
Silver-IT, line 87: It seems like someone already fixed this issue. After merging the latest master branch, I could see it's already fixed. So I have just added comment. The previous code was like below.

**Previous**
```
if (typeof to === 'object')
```

frontend/views/pages/Home.vue
```diff
@@ -96,20 +97,20 @@ export default ({
...
+      // takes user to 'Home.vue' with the '$route.query.next' being set to the initial url.
+      // In this particular condition, the app needs to immediately redirect user to '$route.query.next'
+      // so that the user stays in the same page after the browser refresh.
+      // (Related GH issue: https://github.com/okTurtles/group-income/issues/1830)
+      const path = this.$route.query.next ?? (this.ourGroupProfile ? '/dashboard' : '/pending-approval')
+      this.$router.push({ path }).catch(e => {
+        const { isNavigationFailure, NavigationFailureType } = VueRouter
+        if (isNavigationFailure(e, NavigationFailureType.cancelled)) {
+          console.log(`[Home.vue] Navigation to ${path} cancelled, and another navigation took place instead`)
+        }
```
Silver-IT: Even though, VueRouter causes error and the navigation is cancelled, we could ignore it because another navigation is already done. So we don't need to do anything and just do log the message saying that the previous navigation is cancelled.

Silver-IT (commented):
Ready for the review!

frontend/views/pages/Home.vue
taoeffect: This is OK, and nice job on adding more context to the error message, however, now there is one case where no message will be logged when it should. Please also log the `else` condition.

taoeffect (changes requested):
Good work @Silver-IT, I left one minor comment

taoeffect (approved):
Well done @Silver-IT 👍

PR #1886, opened 2024-03-14
Swap UserID for Username while editing text

### LINKED ISSUE #1867: Fix identityContractIDs showing up in edit field mentions
### Problem

When a chat message contains @mentions, and it is edited, the userIDs are shown instead of the corresponding usernames:

Screenshot 2024-02-26 at 9 22 50 AM
Screenshot 2024-02-26 at 9 22 56 AM

#### Background

In a recent PR #1829 - @corrideat converted the entire codebase from being based on usernames to userIDs. The reason for this is because they are more stable. It is possible that during a migration of a user from one server to another, or some other situation, that user might get a different username. However, their identityContractID will remain stable.

To ensure that in the UI everything is updated correctly and no code breaks, we are relying on translating userIDs to username any time we want to display a username. New getters like `usernameFromID` and `userDisplayNameFromID` have been created for this purpose and are now widely used throughout the app.

#### Solution

Perform this translation when displaying the edit field, so that the user sees usernames, but when the message is saved, it is still using contractIDs.

Please pay careful attention to how this translation is already being done currently.

Silver-IT:
### Summary of Changes

- Closes #1867

### DIFF
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -34,7 +34,7 @@
             ) {{ objReplyMessage.text }}
         send-area(
           v-if='isEditing'
-          :defaultText='text'
+          :defaultText='swapUserIDForUsername(text)'
           :isEditing='true'
           @send='onMessageEdited'
           @cancelEdit='cancelEdit'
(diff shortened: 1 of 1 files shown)


corrideat (approved):
This looks fine to me. Perhaps for a different PR, I'd consider:

1) splitting messages into parts (e.g., text and mentions) so that regular expressions are not needed
2) diplaying mentions in the send / edit areas as rich text instead of plain text.

taoeffect (approved):
Seems to be working well, great job @Silver-IT!

PR #1888, opened 2024-03-14
Fix #1861

### LINKED ISSUE #1861: Log errors from outgoing messages during processMessage differently
### Problem

Messages are now passed through the process function before being sent, to see if they would result in any errors. If they do, they are not sent.

However, this means that we will still see them logged as errors in the console when they shouldn't be (because nothing has actually been changed with the state, and no messages have been sent).

### Solution

When `publishEvent` is called *from a sideEffect*, it should not log any errors from `processMessage` as errors, instead it should log them as warnings, and make it clear that these are errors that would have happened had we sent the message.

So something like:

```js
console.warn(`[publishEvent/processMessage] message would have resulted in error: ${e.message}`, e)
```

However, if the error happens as a result of a user action, then it should probably be reported as an error.

So one way this can be done is to pass in a special option to `publishEvent` in its parameters, like `userInitiated: false` (by default, `userInitiated` would be set to true). If `userInitiated: false`, then log as a warning, if `userInitiated: true` (the default), log as an error.

corrideat:


frontend/model/contracts/chatroom.js
```diff
@@ -235,17 +235,17 @@ sbp('chelonia/defineContract', {
...
             if (!rootGetters.ourContactProfilesById[memberID]) {
               sbp('chelonia/contract/sync', memberID).catch((e) => {
-                console.error(`Error while syncing new memberID's contract ${memberID}`, e)
+                console.warn(`Error while syncing new memberID's contract ${memberID}`, e)
               })
             }
           }
         }).catch((e) => {
-          console.error('[gi.contracts/chatroom/join/sideEffect] Error at sideEffect', e?.message || e)
+          console.warn('[gi.contracts/chatroom/join/sideEffect] Error at sideEffect', e?.message || e)
```
taoeffect: There are a lot of errors that got converted to warnings that are unrelated to calls to publishEvent in this PR.

Ideally we keep most of the errors as errors, and underfit - in other words, it'd be better to have more errors logged as errors in the console, than missing errors we might want to be aware of.

Ideally this PR only changes a couple of lines related to publishEvent that are regularly logged as errors that happen due to expected circumstances.

taoeffect (changes requested):
Review ready!

corrideat:
Instead of using `setTimeout` / `setInterval`, I'd refactor this as follows

```js
const foo = () => {
  sbp('chelonia/contract/wait', contractIDs, () => {
    if (currentUserHasChanged) return; // alternatively, you could use an event listener
    if (contractIDsInQueue) { // the check that you're doing
      foo()
      return
    }
    // The rest of the logic follows here
  }).catch(/* ... */);
}
```

I think this would be safe (or safer) while avoiding unnecessary checking. It also checks for the user changing, which is needed because you don't want to remove contracts in that case. Maybe you can also give up after a certain number of attempts.

taoeffect (approved):
👍

Issue #1881, opened 2024-03-08
Reuse existing file attachments


taoeffect:
### Problem

After PR #1877 is merged, please see if you can modify those changes so that uploading the same file multiple times to the chat doesn't result in the same file getting uploaded multiple times.

### Solution

- Implement a "forward" feature, that lets the user forward an existing message to others, by including the decryption key along with the forwarded message so that the file doesn't need to be re-uploaded.
- If possible, detect that the file had already been uploaded before (by using an HTTP `HEAD` check on the file's `manifestCid`), and reuse the existing information instead of reuploading the file.

corrideat:
> Implement a "forward" feature, that lets the user forward an existing message to others, by including the decryption key along with the forwarded message so that the file doesn't need to be re-uploaded.

If this is the ultimate goal, it can already be implemented _today_ and it does _not_ require re-uploading files. A 'forward' feature would presumably work similarly to reply, except that it'd copy the message contents. Since files are _not_ stored in the message itself (only a reference to them is, in the form of download parameters), forwarding a message that includes files doesn't require any re-uploading.

> If possible, detect that the file had already been uploaded before (by using an HTTP HEAD check on the file's manifestCid), and reuse the existing information instead of reuploading the file.

This is difficult or even undesirable to do because of several factors.

1. When a(n encrypted) file is uploaded, it's preprocessed using random data to derive the file's encryption key and a nonce. These need to be random (at the very least, the nonce must not be re-used for any given encryption key). Because of this randomness, uploading an identical file repeatedly will result in entirely different content.
2. There are also various degrees of freedom in the manifest itself. For one, it can contain arbitrary data, so you can't ensure that the manifest itself is unique. Since the file can contain arbitrary chunks, each chunk will also have a different CID.
3. Using a header or similar is somewhat incompatible with the purpose of using streams, because you need to keep the entire data in memory (or process the content twice on the client). This is because you don't know what the hash will be until after reading the file (and encrypting it, if you're using encryption). This said, it could have some usefulness to maybe just upload the data and have the server do this check (i.e., if a chunk already exists, it ignores it).

Now, these are a few ways this could work, but all seem rather limited.

One is making the 'random' data deterministic. However, the only thing that makes sense for this approach is to derive it from the file itself (plus the encryption key that'll be used in a chatroom). Doing this poses two challenges:

1. Using the file as input again requires either loading the entire file in memory or processing it twice client-side.
[...]

PR #1892, opened 2024-03-21
Delete an attachment in a message

### LINKED ISSUE #1879: Implement deleting attachment of message
### Problem

There are two features for the attachments in the message. One is `Download` and the other is `Delete`.
Download feature is done, but `Delete` feature is not yet.

### Solution

Create a `gi.contracts/chatroom/deleteAttachment` function and implement it.

Silver-IT:
### Summary of Changes

- Closes #1879 
- Fix error that the **Message** component is rendered twice every time user sends a message
- Fix error in setting `currentChatRoomIDs`
- Fix error in setting `isNew` position
- Updated the behavior when user click `Cancel` button in `UploadAttachments` modal


taoeffect (commented):
Nice work @Silver-IT. However, I think this PR needs to wait until @corrideat is finished implementing the accounting logic, please see this comment: https://github.com/okTurtles/group-income/issues/1879#issuecomment-2016369139

corrideat (commented):
This is a general comment based on the implementation in #1898.

In order to delete a file from the server, there is a new selector, `chelonia/fileDelete`, and there've been some changes to how `chelonia/fileUpload` works.

Now, when you call `sbp('chelonia/fileUpload')`, you get an object that has two properties, `download` and `delete`. `download` is what you got previously, which is the argument to `fileDownload`. `delete` is a token that is used for `chelonia/fileDelete`, so for example you call `sbp('chelonia/fileDelete', { manifestCid}, { token: (await sbp('chelonia/fileUpload', ...)).delete })`.

In addition, `chelonia/fileDelete` can, instead of getting a `token`, get a `billableContractID` (set to the group creator), which is so that the group creator can delete files.

Some of the changes needed are, therefore, storing the `delete` token in the uploader's identity contract, so that they are then able to call `fileDelete` when deleting a file, and, maybe, also implementing the UI for the group creator, using their contractID instead of a token.

Silver-IT:
Good to know that, @corrideat. Nice work. So the required updates is that

- If a user calls the `chelonia/fileUpload`, he should also save the `token` to the identity contract, which must be useful when delete the file whenever needed. In the identity contract, those tokens could be stored in map of this format `{ [manifestCid]: token }`.
- When user calls the `chelonia/fileDelete`, he should also call an identity contract function in order to remove the `token` which was saved while file uploading.
- The `billableContractID` is only used whenever the group creator tries to delete attachments. Since the UI design is not ready, so we can skip this at the moment. One question here. What about when user tries to update his avatar? We should also remove the previous image, right? Then, the `billableContractID` must be their identity contractID, right?

corrideat:
> If a user calls the chelonia/fileUpload, he should also save the token to the identity contract, which must be useful when delete the file whenever needed. In the identity contract, those tokens could be stored in map of this format { [manifestCid]: token }.

Right. It doesn't have to be their identity contract, but I guess it's the place that makes most sense.

> When user calls the chelonia/fileDelete, he should also call an identity contract function in order to remove the token which was saved while file uploading.

Probably a good idea to keep the size of the state object small, yes.

> The billableContractID is only used whenever the group creator tries to delete attachments. Since the UI design is not ready, so we can skip this at the moment. One question here. What about when user tries to update his avatar? We should also remove the previous image, right? Then, the billableContractID must be their identity contractID, right?

Right, I think you can skip this depending on what the priorities are. However, @taoeffect wants to have some test cases for deleting files so bear that mind.

taoeffect (changes requested):
This PR can now really delete attachments using @corrideat's latest changes that have been merged into `master`. Please give that a try and add a single Cypress test case to verify it works

Silver-IT (commented):
Review Ready!

frontend/model/contracts/chatroom.js
```diff
@@ -426,7 +438,7 @@ sbp('chelonia/defineContract', {
           state.messages.push(createMessage({ meta, data, hash, height, state, pending, innerSigningContractID }))
         } else if (direction !== 'outgoing') {
           // If an existing message is found, it's no longer pending.
-          delete existingMsg.pending
+          existingMsg.pending = false
```
corrideat: This doesn't look very important, but why set it to `false`? Although a boolean probably doesn't cost much, it's still slightly more information to keep in memory.

frontend/model/contracts/chatroom.js
```diff
@@ -114,6 +114,18 @@ function messageReceivePostEffect ({
   shouldSoundMessage && sbp('okTurtles.events/emit', MESSAGE_RECEIVE)
 }
 
+function deleteEncryptedFilesByToken (manifestCids: string | string[], innerSigningContractID: string) {
+  if (typeof manifestCids === 'string') {
+    manifestCids = [manifestCids]
+  }
+
+  if (Array.isArray(manifestCids) && manifestCids.length) {
+    sbp('gi.actions/identity/removeFiles', {
```
corrideat: This could reject, and that rejection isn't handled.

frontend/model/contracts/chatroom.js
```diff
@@ -551,6 +568,7 @@ sbp('chelonia/defineContract', {
         }
 
         if (me === innerSigningContractID) {
+          deleteEncryptedFilesByToken(data.manifestCids, innerSigningContractID)
```
corrideat: I'm commenting here, although the 'error' will happen at the action in the identity contract. This logic isn't necessarily correct, because the person sending the message isn't necessarily the one that uploaded the files (see #1881). So, although this particular line here is _probably_ fine from what I can see (other than being in the `process` function), throwing/logging an error when a token isn't found isn't ideal. Also, remember that even if the assumption of `uploader = owner` holds, `process` (and `sideEffect`) can be called multiple times.

frontend/model/contracts/chatroom.js
```diff
@@ -562,12 +580,39 @@ sbp('chelonia/defineContract', {
...
+              return attachment.downloadData.manifestCid !== data.manifestCid
+            })
+            Vue.set(state.messages[msgIndex], 'attachments', newAttachments)
+          }
+        }
+      },
+      sideEffect ({ data, contractID, hash, meta, innerSigningContractID }) {
+        const me = sbp('state/vuex/state').loggedIn.identityContractID
+        if (innerSigningContractID === me) {
+          deleteEncryptedFilesByToken(data.manifestCid, innerSigningContractID)
```
corrideat: Same comment about treating a missing token as an error.

test/cypress/fixtures/test.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Name": "Alex Jin",
+  "Birthday": "02.20.1992"
```
corrideat: Do we really need PII for test data?

corrideat (changes requested):

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+    if (me !== identityContractID) {
+      return
+    }
+
+    try {
+      const tokensMap = {}
+      for (const manifestCid of manifestCids) {
+        const token = rootGetters.currentIdentityState.fileDeleteTokens[manifestCid]
+        if (!token) {
+          console.error(`Missing delete token for file with manifestCid ${manifestCid}`)
```
corrideat: See https://github.com/okTurtles/group-income/pull/1892/files#r1557216751

frontend/model/contracts/chatroom.js
Silver-IT: To keep the consistency. When the user receives the message sent by another, the `pending` would be `false`.
```
if (!existingMsg) {
  // If no existing message, simply add it to the messages array.
  const pending = direction === 'outgoing'
  state.messages.push(createMessage({ meta, data, hash, height, state, pending, innerSigningContractID }))
} else if (direction !== 'outgoing') {
  // If an existing message is found, it's no longer pending.
  existingMsg.pending = false
}
```

frontend/model/contracts/chatroom.js
Silver-IT: Rejection can't be handled here. Already handled the errors in `gi.actions/identity/removeFiles` function.

test/cypress/fixtures/test.json
Silver-IT: Of course, not. I forgot it. Thanks for your catch.

frontend/controller/actions/identity.js
Silver-IT: Then, what about using `console.warn`? And updated the logic according to your feedback. Please check it out.
```
'gi.actions/identity/removeFiles': async ({ manifestCids, identityContractID }: {
  manifestCids: string[], identityContractID: string, billableContractID?: string
}) => {
  const me = sbp('state/vuex/state').loggedIn.identityContractID

  if (me !== identityContractID) {
    return
  }

  try {
    const tokensMap = {}
    for (const manifestCid of manifestCids) {
      const token = rootGetters.currentIdentityState.fileDeleteTokens[manifestCid]
      if (!token) {
        console.warn(`Missing delete token for file with manifestCid ${manifestCid}`)
      } else {
        tokensMap[manifestCid] = { token }
      }
    }

    const removableManifestCids = Object.keys(tokensMap)
    if (removableManifestCids.length) {
      await sbp('chelonia/fileDelete', removableManifestCids, tokensMap)
      await sbp('gi.actions/identity/removeFileDeleteToken', {
        contractID: identityContractID,
        data: { manifestCids: removableManifestCids }
      })
    }
  } catch (err) {
    console.warn(err?.message || err)
  }
}
```

frontend/model/contracts/chatroom.js
taoeffect: I do agree with @corrideat here as both approaches result in the same value for `!existingMsg.pending` and `v-if='existingMsg.pending'`, but the `delete` approach uses slightly less memory

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+        tokensMap[manifestCid] = { token }
+      }
+
+      await sbp('chelonia/fileDelete', manifestCids, tokensMap)
+      await sbp('gi.actions/identity/removeFileDeleteToken', {
+        contractID: identityContractID,
+        data: { manifestCids }
+      })
+    } catch (err) {
+      console.error(err?.message || err)
```
taoeffect: "Naked" calls to `console.error/warn` logs are generally something we want to avoid.

In other words, we always want to include a string as the first parameter that describes where the error is coming from (and possibly why). Without this, it will be less clear in the console logs where the error is coming from.

So something like this would be preferable:

```js
console.error("'gi.actions/identity/removeFiles' error:", err?.message || err)
```

Also, all of these actions must behave consistently: this means throwing an exception when the error occurs. Both `'gi.actions/identity/uploadFiles'` and `'gi.actions/identity/removeFiles'` fail to do this.

Please look at how the other actions behave, using `GIErrorUIRuntimeError` to create a localized user-facing string that can be shown to the user if an error occurs.

frontend/model/contracts/chatroom.js
taoeffect: > Already handled the errors in gi.actions/identity/removeFiles function.

The exception does need to be handled here too. Please see [my comment here about that.](https://github.com/okTurtles/group-income/pull/1892#discussion_r1557878667)

frontend/model/contracts/chatroom.js
taoeffect: @corrideat:

> other than being in the `process` function

This is in the `sideEffect` AFAICT...

> `process` (and `sideEffect`) can be called multiple times.

Can you elaborate on how `sideEffect` can be called multiple times? Do you mean because every user in the chatroom will call it at least once? (This is gated by the `if` check...)

@Silver-IT:

Note that per my [other comment](https://github.com/okTurtles/group-income/pull/1892#discussion_r1557878667), `'gi.actions/identity/removeFiles'` will need to throw an exception upon an error. So that means that you either need a `try/catch` here, or inside of `deleteEncryptedFilesByToken` in order to catch and either log or display the error to the user (perhaps via a notification).

corrideat: >> other than being in the process function

Please disregard that, I updated my comment but not that part

> Can you elaborate on how sideEffect can be called multiple times?

`sideEffect` can be called any number of times, for every time a contract gets re-synced from scratch (this goes for `process` and `validate` too, which can be called even more times, in part because of outgoing messages). Hence, expecting it to be called once and reporting errors on successive runs is something to avoid, unless the error is severe enough to merit a (likely) false positive.

frontend/controller/actions/identity.js
corrideat: I think using `console.warn` is fine here, but also see @taoeffect's comment on error handling. _Ideally_ failing to delete a file _would_ be treated as an error, but only when the user is the one initiating the action. One way to do this could be to do the deletion as part of the `action` instead of as part of the contract (OTOH, doing it in the action makes it more fault tolerant).

frontend/model/contracts/chatroom.js
taoeffect: > Hence, expecting it to be called once and reporting errors on successive runs is something to avoid, unless the error is severe enough to merit a (likely) false positive.

Good point 👍 

So the error can be just logged (not shown to the user), but it still needs to be caught either here or inside of `deleteEncryptedFilesByToken`

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
   ...encryptedAction('gi.actions/identity/joinDirectMessage', L('Failed to join a direct message.')),
   ...encryptedAction('gi.actions/identity/joinGroup', L('Failed to join a group.')),
   ...encryptedAction('gi.actions/identity/leaveGroup', L('Failed to leave a group.')),
-  ...encryptedAction('gi.actions/identity/setDirectMessageVisibility', L('Failed to set direct message visibility.'))
+  ...encryptedAction('gi.actions/identity/setDirectMessageVisibility', L('Failed to set direct message visibility.')),
+  'gi.actions/identity/uploadFiles': async ({ attachments, billableContractID }: {
+    attachments: Object, billableContractID: string
```
taoeffect: You are using `attachments.map`, suggesting this should be an `Array<Object>` instead of `Object`

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+      const attachmentsData = await Promise.all(attachments.map(async (attachment) => {
+        const { mimeType, url } = attachment
+        // url here is an instance of URL.createObjectURL(), which needs to be converted to a 'Blob'
+        const attachmentBlob = await objectURLtoBlob(url)
+        const response = await sbp('chelonia/fileUpload', attachmentBlob, {
+          type: mimeType, cipher: 'aes256gcm'
+        }, { billableContractID })
+        const { delete: token, download: downloadData } = response
+        return {
+          downloadData: { ...omit(attachment, ['url']), downloadData },
```
taoeffect: Does `downloadData.downloadData` make sense to do?

And `...omit(attachment, ['url'])` doesn't seem to be used at all?

So I think this can be simplified to:

```js
        return {
          downloadData,
          deleteData: { token }
        }
```

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+      }))
+
+      await sbp('gi.actions/identity/saveFileDeleteToken', {
+        contractID: rootGetters.ourIdentityContractId,
+        data: { tokensByManifestCid }
+      })
+
+      return attachmentsData.map(({ downloadData }) => downloadData)
+    } catch (err) {
+      console.error('Error during uploading files', err)
```
taoeffect: Remember to re-throw using a localized `GIErrorUIRuntimeError`

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+    }
+  },
+  'gi.actions/identity/removeFiles': async ({ manifestCids, identityContractID }: {
+    manifestCids: string[], identityContractID: string, billableContractID?: string
+  }) => {
+    const rootGetters = sbp('state/vuex/getters')
+    const me = sbp('state/vuex/state').loggedIn.identityContractID
+
+    if (me !== identityContractID) {
+      return
```
taoeffect: Probably a good idea to issue a `console.warn` here before returning

EDIT: although remember files of other users can be deleted by the group admin too, and see also my comment below about throwing exceptions whenever actions fail.

frontend/controller/actions/identity.js
```diff
@@ -701,5 +701,72 @@ export default (sbp('sbp/selectors/register', {
...
+    }
+
+    try {
+      const tokensMap = {}
+      for (const manifestCid of manifestCids) {
+        const token = rootGetters.currentIdentityState.fileDeleteTokens[manifestCid]
+        if (!token) {
+          console.error(`Missing delete token for file with manifestCid ${manifestCid}`)
+          return
```
taoeffect: In order for the caller to know that the call to `removeFiles` failed, we need to throw an exception.

For all actions the recommended exception is a `GIErrorUIRuntimeError` with a localized string that explains in a human-readable way what went wrong.

This gives the app the ability to explain to the user why one of their actions couldn't be done.

test/cypress/fixtures/test.json
taoeffect: I'm clicking "Unresolve" on this to not forget about it (since the updates haven't yet been pushed)

frontend/model/contracts/chatroom.js
```diff
@@ -114,6 +114,18 @@ function messageReceivePostEffect ({
   shouldSoundMessage && sbp('okTurtles.events/emit', MESSAGE_RECEIVE)
 }
 
+function deleteEncryptedFilesByToken (manifestCids: string | string[], innerSigningContractID: string) {
+  if (typeof manifestCids === 'string') {
+    manifestCids = [manifestCids]
+  }
+
+  if (Array.isArray(manifestCids) && manifestCids.length) {
```
taoeffect: With the way the function signature is written, this must be an array at this point.

However, in reality this check is necessary because this function is called with `undefined` sometimes. That violates the function signature types though, and also doesn't make much sense to call this function with `undefined`.

So I would do the check for `if (manifestCids)` in the caller, not here, but where this function is called from (e.g. the contracts).

Then here it is safe to assume we have an array.

taoeffect (changes requested):
Review ready!

frontend/model/contracts/chatroom.js
```diff
@@ -562,12 +580,39 @@ sbp('chelonia/defineContract', {
...
+      validate: actionRequireInnerSignature(objectOf({
+        hash: string,
+        manifestCid: string
+      })),
+      process ({ data, innerSigningContractID }, { state }) {
+        if (!state.onlyRenderMessage) {
+          return
+        }
+        const msgIndex = findMessageIdx(data.hash, state.messages)
+        if (msgIndex >= 0 && innerSigningContractID === state.messages[msgIndex].from) {
```
taoeffect: @corrideat correct me if I'm wrong but I believe this could/should also include a check for being the groupCreator?

corrideat: That's right, although in that case instead of a token the `billableContractID` attribute will be used

frontend/controller/actions/identity.js
Silver-IT: At the moment, we don't have any user flow for the group creators to delete attachments.

Or, should we simply make the `Delete` button visible for the attachments when group creator hovers on them? And one question related to this is that should it be possible for the group creators to delete messages sent by another? (When the message has only attachments and no text, deleting those attachments is same to the deleting message itself)

My personal opinion is that we should consider this later when we have the design for the user flow describing how the group creator could delete attachments. (And exact answer for only attachments or even messages?)

frontend/controller/actions/identity.js
Silver-IT: According to the @corrideat comment [here](https://github.com/okTurtles/group-income/pull/1892#discussion_r1557216751) and your comment [here](https://github.com/okTurtles/group-income/pull/1892#discussion_r1557926007), we can simply log warnings and continue process, can't we?

frontend/model/contracts/chatroom.js
Silver-IT: We don't have any user flow for the group creators to delete attachments. So I would like to avoid this now.

frontend/controller/actions/identity.js
Silver-IT: > And ...omit(attachment, ['url']) doesn't seem to be used at all?

Those are needed. So I will update this way.
```
return {
  attributes: omit(attachment, ['url']),
  downloadData,
  deleteData: { token }
}
```

frontend/controller/actions/identity.js
taoeffect: > should we simply make the Delete button visible for the attachments when group creator hovers on them? And one question related to this is that should it be possible for the group creators to delete messages sent by another?

Yes to both questions

Silver-IT: I will update the PR to make it possible for the group creators to delete attachments with the same user flow for the sender to delete them.
And for the second problem (It's actually a new workflow which makes me surprise), I will create another issue and implement it in another PR. Does that sounds good?

frontend/controller/actions/identity.js
taoeffect: @Silver-IT no, both of those comments are about behavior inside of contracts (after an action is sent, while it is being processed), whereas this is about the behavior of user-initiated actions (before they are successfully sent).

frontend/controller/actions/identity.js
taoeffect: Yep!

frontend/controller/actions/identity.js
Silver-IT: `gi.actions/identity/uploadFiles` is the API that user can initiate. But, `gi.actions/identity/removeFiles` is not something that user can initiate. It's something that could be called in the `sideEffect` of `gi.contracts/chatroom/deleteMessage` and `gi.contracts/chatroom/deleteAttachments`.

taoeffect: Ah, I see, in that case my comments do not apply to removeFiles, and if it makes sense to move that into a method inside of the contract, that might be a better place for it

Silver-IT: Actually, At first it was inside the contracts. But If we do that, we need to allow two sbp functions because they are being called inside the contract. That's why I made `gi.actions/identity/removeFiles`.
```
'chelonia/fileDelete'
'gi.actions/identity/removeFileDeleteToken'
```
 And another reason is to keep consistency. We have the similar function `gi.actions/identity/uploadFiles`, and I thought it's good to have `gi.actions/identityremoveFiles` which is much related to the former.

taoeffect: Fair enough 👍
Maybe then just add a comment above it to mention that that action is intended to be called from a contract side effect, and not directly by the user

taoeffect (dismissed):
Nicely done @Silver-IT! 👏

corrideat (dismissed):

frontend/model/contracts/chatroom.js
```diff
@@ -114,6 +114,15 @@ function messageReceivePostEffect ({
   shouldSoundMessage && sbp('okTurtles.events/emit', MESSAGE_RECEIVE)
 }
 
+async function deleteEncryptedFiles (manifestCids: string | string[], option: Object) {
+  if (Object.values(option).reduce((a, c) => a || c, false)) {
+    if (typeof manifestCids === 'string') {
```
corrideat: Perhaps not critical, but I'd use `!Array.isArray(manifestCids)` in this case because it'll break more consistently if manifestCids is not a string.

Also, I've pretty much given up at this point on this, but, in JS, `typeof x === 'string'` doesn't work for Object-wrapped strings (e.g., strings from `new String('foo')` or `Object('foo')`). This is mostly academic as such strings are rare and the 'proper' check for this is ugly and doesn't play nice with TypeScript. However, this is in part why if I can avoid `typeof`, as in this case, I do.

Silver-IT: Reasonable. I will update it now.

taoeffect (approved):
Excellent work @Silver-IT!

PR #1894, opened 2024-03-22
Retry message when uploading attachments failed

### LINKED ISSUE #1890: Add "retry" button instead of deleting messages that failed to send
### Problem

If a message fails to send, PR #1877 currently deletes it.

This can be a problem for users who wrote a long message and don't want to retype it.

### Solution

Leave it grayed out and add a button or a link that says "resend message" to the message to try again.

- When the user clicks on it, make the link disappear to show that an attempt has been made to resend it
- If the message succeeds, make the message no longer grey and remove the "resend message" link
- If the message fails to send again, show the link again

Silver-IT:
### Summary of Changes

- Closes #1890


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -454,6 +454,7 @@ export default ({
             beforeRequest
           }
         }).catch((e) => {
+          // TODO: add Retry button on the right side
```
taoeffect: I'm confused by this comment... isn't the point of this PR to implement this? 🤔 

Issue #1890 is about showing retry when sending the message fails.

-----

Separately, on line 411 above (Github won't let comment on that line since it wasn't modified) we have this:

```js
data = { ...data, replyingMessage }
```

That is the same as the much more straightforward and faster:

```js
data.replyingMessage = replyingMessage
```

Could update that code as part of a little boyscouting?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -508,19 +509,20 @@ export default ({
...
+              // NOTE: remove temporary message which is created before uploading attachments
+              if (temporaryMessage) {
+                const msgIndex = findMessageIdx(temporaryMessage.hash, this.messageState.contract.messages)
+                this.messageState.contract.messages.splice(msgIndex, 1)
+              }
+            }
             sendMessage(removeTemporaryMessage)
           } else {
-            removeTemporaryMessage()
+            temporaryMessage.pending = false
```
taoeffect: This line doesn't make sense to me. The message should still be pending if the files failed to upload (see description of message ordering behavior in #1845).

taoeffect (changes requested):
Nice work @Silver-IT!

I have left a few comments and also have a question: I didn't see where the "Retry" link was added in this PR.

Also, can you provide a series of steps for how to test this PR? For example, maybe a line of code to insert somewhere to fake a failed message delivery.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: > I'm confused by this comment... isn't the point of this PR to implement this?

We make two calls to the backend when creating a new message. The first call is for `fetchServerTime` and second one is to publish event. When we create a message without any attachments, the pending message would be created after `fetchServerTime` is successfully executed, and before the event is actually published. If the user is offline, the first call (`fetchServerTime`) would be firstly failed, so no pending message would be created.
So, I thought it would be difficult to have the situation when users would have pending message which is failed to be published. So I ignored it here. But added TODO comment.

Silver-IT:
> Also, can you provide a series of steps for how to test this PR? For example, maybe a line of code to insert somewhere to fake a failed message delivery.

You can see `uploadAttachments` asynchronous function inside the `handleSendMessage` in ChatMain.vue. It returns the boolean value which tells the attachments are successfully uploaded or not. You can return `false` somehow, in order to simulate when the uploading attachments are failed.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: @Silver-IT

1. If we are offline, ~~the send button should be disabled~~ (EDIT: better to display an alert when they try to send the message instead saying, `Can't send message when offline, please connect to the Internet`), so that the user can't send any messages (please feel free to update this PR to do that).
2. If the message fails to send, the retry button should appear. This `.catch` doesn't do that. It needs to do that, as that is part of Issue #1890 - if the message fails to send for whatever reason, the retry button must appear

Silver-IT: Okay.

taoeffect:
> You can return `false` somehow, in order to simulate when the uploading attachments are failed.

I tested this, and it didn't work...

Screenshot 2024-03-25 at 6 21 57 PM

Message sits there in grey:

Screenshot 2024-03-25 at 6 21 52 PM

taoeffect (changes requested):
Nice work @Silver-IT! I was able to test this now with your suggestion! 🙏 

Here's what I found:

1. If the file files to upload, we need a message in smaller font with a text-link that says something like, "Message failed to send. [Resend message](#)". I would make the text bold also so that it can be clearly seen by the user. The way that it currently uses a button on hover is too confusing for users (most of whom won't realize that button exists, and those who do see it, won't know what it's for).
2. There's a weird error that I get when following the following steps:

### Steps to reproduce error

1. Send a message with attachments that fails (see [patch below](https://github.com/okTurtles/group-income/pull/1894#issuecomment-2017944840)). Do not do anything else (leave it in gray, don't resend it)
2. Save the ChatMain.vue file to cause the page to refresh

When the page refreshes, the following error is visible in the console, and the messages in the chat fail to load:

Screenshot 2024-03-25 at 8 04 22 PM

Silver-IT:
> There's a weird error that I get when following the following steps:

I can not reproduce your error even though I followed the steps you provided. I tried several times, but couldn't get that error. Could this be that weird error which I faced, but you and @corrideat didn't get?

EDITED: I mean, it could be different in its appearance, but could be of the same reason.

taoeffect:
@Silver-IT here's the diff from my changes to get this to happen (since it's random when the file fails, you could change the code to `randomIntFromRange(1, 1)`):

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a793c8e14..3f25cce09 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -135,7 +135,7 @@ import {
 } from '@model/contracts/shared/constants.js'
 import { findMessageIdx, createMessage } from '@model/contracts/shared/functions.js'
 import { proximityDate, MINS_MILLIS } from '@model/contracts/shared/time.js'
-import { cloneDeep, debounce, throttle } from '@model/contracts/shared/giLodash.js'
+import { cloneDeep, debounce, throttle, delay } from '@model/contracts/shared/giLodash.js'
 import { EVENT_HANDLED } from '~/shared/domains/chelonia/events.js'
 import { objectURLtoBlob } from '@utils/image.js'
 
@@ -464,6 +464,10 @@ export default ({
       }
       const uploadAttachments = async () => {
         try {
+          if (!isNaN(0)) {
+            await delay(1000)
+            return false
+          }
           const attachmentsToSend = await Promise.all(attachments.map(async (attachment) => {
             const { mimeType, url, name } = attachment
             // url here is an instance of URL.createObjectURL(), which needs to be converted to a 'Blob'
```

> Could this be that weird error which I faced, but you and @corrideat didn't get?

I don't remember, but no I don't think so..

Silver-IT:
> If the file files to upload, we need a message in smaller font with a text-link that says something like, "Message failed to send. Resend message". I would make the text bold also so that it can be clearly seen by the user. The way that it currently uses a button on hover is too confusing for users (most of whom won't realize that button exists, and those who do see it, won't know what it's for).

I will update the PR based on your feedback.

> I don't remember, but no I don't think so..

I mean [this](https://okturtles.slack.com/archives/C02QF71TK1Q/p1710924068406959).

taoeffect:
> I mean [this](https://okturtles.slack.com/archives/C02QF71TK1Q/p1710924068406959).

The error seems rather different, I don't think it's the same.

taoeffect:
> I mean [this](https://okturtles.slack.com/archives/C02QF71TK1Q/p1710924068406959).

The error seems rather different, I don't think it's the same.

taoeffect (changes requested):
Screenshot 2024-03-26 at 6 29 36 PM

A few requests for the styling:

1. Could you make the link color a regular blue link color?
2. Could you remove the italics? (leave it as bold though)
3. Could you please add an underline so that it's clear to mobile users that it's a clickable link?

Silver-IT (commented):
@taoeffect, Ready for the final review!

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -394,24 +395,26 @@ export default ({
...
-        // If not replying to a message, use original data; otherwise, append
-        // replyingMessage to data.
-        data = { ...data, replyingMessage }
+        // NOTE: If not replying to a message, use original data; otherwise, append replyingMessage to data.
+        data.replyingMessage = replyingMessage
+        // NOTE: for the messages with only images, the text should be updated with file name
+        if (!replyingMessage.text) {
+          const msg = this.messages.find(m => (m.hash === replyingMessage.hash))
+          const hasMoreThanTwoFiles = msg.attachments.length > 1
+          data.replyingMessage.text = msg.attachments[0].name + (hasMoreThanTwoFiles ? ' and more' : '')
```
taoeffect: Whenever constructing user-facing strings, always construct complete sentences using either `L` function or `i18n` component

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -454,8 +458,14 @@ export default ({
             beforeRequest
           }
         }).catch((e) => {
-          console.error(`Error while publishing message for ${contractID}`, e)
-          alert(e?.message || e)
+          if (e.cause?.name === 'ChelErrorFetchServerTimeFailed') {
+            alert('Can\'t send message when offline, please connect to the Internet')
```
taoeffect: - Missing `L` function
- Recommend using double-quotes to avoid having to escape single-quote

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Nice catch. I will cover it with `L`. However, I remember we are using double-quotes only when designing vue components. There are examples, we are using single quotes in js files. Should I update it anyway?

```
...
throw new Error('Username can\'t be deleted')
...
[PROPOSAL_GENERIC]: {
      [PROPOSAL_VARIANTS.CREATED]: L('{from} created a proposal. "{title}"', options),
      [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} is expiring. "{title}"', options),
      [PROPOSAL_VARIANTS.ACCEPTED]: L('{from}\'s proposal is accepted. "{title}"', options),
      [PROPOSAL_VARIANTS.REJECTED]: L('{from}\'s proposal is rejected. "{title}"', options),
      [PROPOSAL_VARIANTS.EXPIRED]: L('{from}\'s proposal is rejected. "{title}"', options)
}
...
L('{span_}1 member{_span} hasn\'t {strong_} logged in past {days} days or more {_strong}', argsMap['haventLoggedIn'])
```

taoeffect: It is fine (preferable) to use double-quotes here in this situation because no double-quotes are being shown to the user, and so in this case we can use double-quotes to avoid having to escape the single-quote that's being shown to the user.

*EDIT: in the other example you gave, we're showing double-quotes to the user in the user-facing string, so we use single-quotes on the outside to avoid having to escape the double-quotes on the inside.

taoeffect (approved):
Nicely done @Silver-IT! Brave! 😄 👏

PR #1898, opened 2024-03-27
Resource accounting implementation

corrideat:


backend/routes.js
```diff
@@ -39,25 +39,68 @@ route.POST('/event', {
...
           }
         }
       }
+      // Store size information
+      const sizeKey = `_private_size_${deserializedHEAD.contractID}`
+      // Use a queue to ensure atomic updates
+      await sbp('okTurtles.eventQueue/queueEvent', sizeKey, async () => {
+        // Size is stored as a hex value
+        const existingSize = parseInt(await sbp('chelonia/db/get', sizeKey, 16)) || 0
+        await sbp('chelonia/db/set', sizeKey, (existingSize + Buffer.byteLength(request.payload)).toString(16))
```
taoeffect: Why is the size stored as a hex value? Seems simpler and more natural (for humans) for it to be stored as a regular decimal?

corrideat: Sure. This will likely not be seen by humans directly and hex is slightly more efficient to parse (that was the reason)

backend/auth.js
```diff
@@ -2,44 +2,55 @@
...
 
-import { verify, b64ToStr } from '~/shared/functions.js'
-
+import { verifyShelterAuthorizationHeader } from '~/shared/domains/chelonia/utils.js'
 const Boom = require('@hapi/boom')
 
 exports.plugin = {
   name: 'gi-auth',
   register: function (server: Object, opts: Object) {
+    server.auth.scheme('gi-bearer', function (server, options) {
```
taoeffect: Perhaps it is time to drop the `gi-` prefix and start using either `chel-` or `shelter-`? (Same for below)

backend/routes.js
```diff
@@ -36,28 +46,55 @@ const route = new Proxy({}, {
...
   try {
     console.debug('/event handler')
     const deserializedHEAD = GIMessage.deserializeHEAD(request.payload)
     try {
+      if (!manifestRegex.test(deserializedHEAD.head.manifest)) {
+        return Boom.badData('Invalid manifest')
+      }
+      const credentials = request.auth.credentials
+      // Only allow identity contracts to be created without attribution
+      if (!credentials?.billableContractID && deserializedHEAD.contractID === deserializedHEAD.hash) {
```
taoeffect: Maybe this is a good time to define `.isFirstMessage()` for the object returned by `deserializeHEAD`? I think that would be clearer than this.

backend/routes.js
```diff
@@ -253,11 +297,29 @@ route.POST('/file', {
...
+    if (await sbp('chelonia/db/get', manifestHash)) {
+      throw new Error(`Manifest ${manifestHash} already exists`)
+    }
+    // Ensure that the chunks do not exist
+    await Promise.all(chunks.map(async ([cid]) => {
+      const exists = !!(await sbp('chelonia/db/get', cid))
+      if (exists) {
+        throw new Error(`Chunk ${cid} already exists`)
+      }
+    }))
```
taoeffect, line 318: It's not clear to me why we're making sure these chunks and everything else don't exist... if they do exist, is that a big problem? If so, could you add a comment explaining why?

backend/routes.js
```diff
@@ -287,6 +355,91 @@ route.GET('/file/{hash}', {
   return h.response(blobOrString).etag(hash)
 })
 
+route.POST('/deleteFile/{hash}', {
```
taoeffect, line 358: Can you add a comment above this line something to the effect of, "allow file deletion, and allow either the bearer of the deletion token or the file owner to delete it" ?

backend/routes.js
```diff
@@ -287,6 +355,94 @@ route.GET('/file/{hash}', {
...
+    if (!Array.isArray(manifest.chunks) || !manifest.chunks.length) return Boom.badData('missing chunks')
+    // Delete all chunks
+    await Promise.all(manifest.chunks.map(([, cid]) => sbp('chelonia/db/delete', cid)))
+  } catch (e) {
+    console.warn(e, `Error parsing manifest for ${hash}. It's probably not a file manifest.`)
+    return Boom.notFound()
+  }
+  await sbp('chelonia/db/delete', hash)
+  await sbp('chelonia/db/delete', `_private_owner_${hash}`)
+  await sbp('chelonia/db/delete', `_private_size_${hash}`)
```
taoeffect, line 432: Should these also be done atomically in a queue? If not, can you add a comment explaining why there's an exception for them?

backend/routes.js
```diff
@@ -333,64 +489,95 @@ route.GET('/', {}, function (req, h) {
...
+        r: Joi.string().required(),
+        // `s` is an opaque (to the client) value that was earlier returned by
+        // the server
+        s: Joi.string().required(),
+        // `sig` is an opaque (to the client) value returned by the server
+        // to validate the request (ensuring that (`r`, `s`) come from a
+        // previous request
         sig: Joi.string().required(),
+        // `Eh` is the  Eh = E_{S_A + S_C}(h), where S_A and S_C are salts and
+        //                                     h = H\_{S_A}(P)
```
taoeffect, line 520: Nice!! Thanks for updating these! 😄

backend/routes.js
```diff
@@ -333,64 +489,95 @@ route.GET('/', {}, function (req, h) {
...
+    if (req.params['name'].startsWith('_private')) return Boom.notFound()
+    console.error({ name: req.params.name, x: 'foo' })
+    const contractID = await sbp('backend/db/lookupName', req.params['name'])
+    if (contractID !== credentials.billableContractID) {
+      // This ensures that only the owner of the contract can set a salt for it,
+      // closing a small window of opportunity(*) during which an attacker could
+      // potentially lock out a new user from their account by registering a
+      // different salt.
+      // (*) This is right between the moment an OP_CONTRACT is sent and the
+      // time this endpoint is called, which should follow almost immediately after.
```
taoeffect, line 540: Nice comment! 👍

frontend/controller/actions/group.js
```diff
@@ -169,6 +176,20 @@ export default (sbp('sbp/selectors/register', {
...
+            purpose: ['sak'],
+            ringLevel: 0,
+            permissions: [],
+            allowedActions: [],
+            meta: {
+              private: {
+                content: SAKs
+              }
+            },
+            data: SAKp
```
taoeffect, line 192: Shouldn't we make sure that the `#sak` has proper permissions before the server allows it to be registered?

I.e. verify `ringLevel: 0` and `permissions: []`?

corrideat: Possibly, although this isn't the place to check that.

shared/domains/chelonia/time-sync.js
```diff
@@ -0,0 +1,95 @@
...
+  const newMonotonicBase = performance.now()
+  // Now, ask the server for the time
+  const time = await fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
+  const requestTimeElapsed = performance.now()
+  if (requestTimeElapsed - newMonotonicBase > 1000) {
+    throw new Error('Error fetching server time: request took too long')
+  }
+  // If the request didn't succeed, report it
+  if (!time.ok) throw new Error('Error fetching server time')
+  const serverTime = (new Date(await time.text())).valueOf()
```
taoeffect, line 31: If it takes 200 ms for our request to get to the server, and 200ms to get the response, by the time we get the response from the server, the time we receive from the server will be 200ms old. So we should add 200ms to the time received from the server.

We can estimate this time by doing `(requestTimeElapsed - newMonotonicBase) / 2`

shared/domains/chelonia/time-sync.js
```diff
@@ -0,0 +1,95 @@
+import sbp from '@sbp/sbp'
+
+let wallBase = Date.now()
+let monotonicBase = performance.now()
```
taoeffect: Could you add comments here explaining what `wallBase` and `monotonicBase` mean? These terms might be confusing to people (as they are confusing to me)

shared/domains/chelonia/time-sync.js
```diff
@@ -0,0 +1,95 @@
...
+
+    // Watchdog to ensure our time doesn't drift. Periodically check for
+    // differences between the elapsed wall time and the elapsed monotonic
+    // time
+    watchdog = setInterval(() => {
+      const wallNow = Date.now()
+      const monotonicNow = performance.now()
+      const difference = Math.abs(Math.abs((wallNow - wallLast)) - Math.abs((monotonicNow - monotonicLast)))
+      // Tolerate up to a 10ms difference
+      if (difference > 10) {
```
taoeffect, line 82: 10ms seems rather small... how often would this get called?

shared/domains/chelonia/time-sync.js
```diff
@@ -0,0 +1,95 @@
...
+      clearTimeout(resyncTimeout)
+      watchdog = undefined
+      resyncTimeout = undefined
+    }
+  },
+  'chelonia/time': function () {
+    const monotonicNow = performance.now()
+    const wallNow = wallBase - monotonicBase + monotonicNow
+    return (wallNow / 1e3 | 0)
+  }
```
taoeffect, line 111: Can you add a comment above this selector explaining what this function does?

shared/domains/chelonia/time-sync.js
```diff
@@ -0,0 +1,95 @@
...
+  if (!time.ok) throw new Error('Error fetching server time')
+  const serverTime = (new Date(await time.text())).valueOf()
+  // If the value could not be parsed, report that as well
+  if (Number.isNaN(serverTime)) throw new Error('Unable to parse server time')
+  wallBase = serverTime
+  monotonicBase = newMonotonicBase
+}
+
+export default (sbp('sbp/selectors/register', {
+  'chelonia/private/startClockSync': function () {
```
taoeffect, line 42: How difficult would it be to adapt this code to support multiple different servers? For federation it would be necessary to be sync'd to each server on which we are monitoring/interacting with contracts.

shared/domains/chelonia/time-sync.js
corrideat: This is called locally and the difference should typically be `0`, unless the local date has been adjusted for some reason.

shared/domains/chelonia/time-sync.js
corrideat: For this particular code, probably not so much. The issue is that currently there's only a single connectionURL. The changes needed here would depend on how multiple server URLs are handled.

taoeffect (approved):
Nicely done @corrideat!! Very impressive!!! LGTM 💪

Issue #1619, opened 2023-05-29
Design for Global Dashboard


taoeffect:
### Problem

We need a place to:

- Handle DMs between users outside of groups
- Display important news related to Group Income (e.g. feature updates, etc.)
- Possibly host global widgets (for example, possible leaderboards to encourage competition between groups to perform well)

### Solution

Design a global dashboard that has the capability of handling all of this stuff, or at the very least DMs + news.

SebinSong:
**[ Figma link ]** : https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=26396-326779&mode=design&t=LHJ3EeI9yZVurepS-0

SebinSong:
Hi @taoeffect , just started researching/working on this issue. One instant question is,

**Q. is this Global dashboard expected to show only when there is a multiple groups for the user? Or do we show this page even when the user has only one group?**

Also, to make the development&testing feel smoother, I'm thinking of splitting this issue into 3 PRs, which are 
1) News&updates -> 2) Direct messages -> 3) Dashboard

which is 'smaller' -> 'larger' order. Does it sound good?

taoeffect:
> Q. is this Global dashboard expected to show only when there is a multiple groups for the user? Or do we show this page even when the user has only one group?

I'm thinking it probably makes sense to always show it.

> which is 'smaller' -> 'larger' order. Does it sound good?

Well, I'm a bit confused because it's not clear to me what the difference between "News & Updates" and "Dashboard" is. Doesn't implementing the "News & Updates" require implementing the scaffolding for the Dashboard as well?

SebinSong:
@taoeffect 
I might have misunderstood the [figma](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=26396-326779&mode=design&t=LHJ3EeI9yZVurepS-0). so are we only caring about this red-box but not the skeleton-like designs above it?

![image](https://github.com/okTurtles/group-income/assets/17641213/a2b3a6e9-c23a-4b1d-921f-03464d63cf42)

taoeffect:
@SebinSong The two designs above it don't have the right names. Sorry that it's not clear! What I meant by the "Global Dashboard" is this entire page, and all of the sub pages within it, so the general layout of that page.

SebinSong:
@taoeffect No worries on that (since there is no designer at the company atm).
[...]

PR #1899, opened 2024-03-28
PR for the Step 1 of #1619

SebinSong:
Working on the step 1 in [this comment](https://github.com/okTurtles/group-income/issues/1619#issuecomment-2011034723) in the issue #1619 , which is 

> Step 1) Entire layout of the Global dashboard (also making sure it looks good in all screen sizes) + adding 'News & Updates' sub-page.

image

### DIFF
--- a/frontend/controller/utils/misc.js
+++ b/frontend/controller/utils/misc.js
@@ -1,8 +1,23 @@
 'use strict'
 
+import { L } from '../../common/common.js'
+
 export function handleFetchResult (type: string): ((r: any) => any) {
   return function (r: Object) {
     if (!r.ok) throw new Error(`${r.status}: ${r.statusText}`)
     return r[type]()
+  }
+}
+
+export const GLOBAL_DASHBOARD_SETTINGS: {[string]: Object } = {
+  'news-and-updates': {
+    title: L('News & Updates'),
+    routeTo: '/global-dashboard/news-and-updates',
+    icon: 'newspaper'
+  },
+  'direct-messages': {
+    title: L('Direct Messages'),
+    routeTo: '/global-dashboard/direct-messages',
+    icon: 'comment'
   }
 }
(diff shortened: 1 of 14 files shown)


frontend/controller/utils/misc.js
```diff
@@ -1,5 +1,20 @@
...
+    title: L('News & Updates'),
+    routeTo: '/global-dashboard/news-and-updates',
+    icon: 'newspaper'
+  },
+  'direct-messages': {
+    title: L('Direct Messages'),
+    routeTo: '/global-dashboard/direct-messages',
+    icon: 'comment'
+  }
+}
```
taoeffect: To me this feels a bit out of place. Perhaps it would make more sense to store this route-related information in `frontend/controller/router.js`?

taoeffect (commented):
This is looking and working great @SebinSong 👏 

I can either merge as-is, or wait for one minor change request, up to you (see comment).

frontend/controller/utils/misc.js
SebinSong: @taoeffect sure!

SebinSong: @taoeffect I tried to move it to `router.js` and for some reason the entire app started to crash after this change is made, producing below seemingly unrelated error message. (it looks like Vue initialisation process itself fails with this change)

image

Couldn't really find the clear reason for this during my couple of hours of investigation, but it **seems** to be related to some sort of 'a circular reference' structure created like below as a result of this movement.

1) `GlobalDashboard.vue` imports something from `router.js`
2) `router.js` has to import `GlobalDashboard.vue` back again, to register it as a page component for a route.
3) something gets messed up during Vue initialisation process.

So I decided to move it back to the `misc.js`.

taoeffect: @SebinSong in that case, can you move it into `GlobalDashboard.vue`? That seems more relevant than `misc.js`

SebinSong: @taoeffect Updated!

taoeffect (approved):
Perfect! 👏

PR #1902, opened 2024-04-01
Fix error in setting unread position of chatroom

### LINKED ISSUE #1893: Incorrect unread position in chat pages
### Problem

I am not sure from when it doesn't work, but the unread position doesn't work correctly in chat pages.
image
As we can see in the above screenshot, I (Alex Jin) sent the last message, but there is a `New` above it which says that I haven't read the last message yet.

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1893

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -891,24 +891,23 @@ export default ({
 
         const isMessageAddedOrDeleted = (message: GIMessage) => {
           const allowedActionType = [GIMessage.OP_ACTION_ENCRYPTED, GIMessage.OP_ACTION_UNENCRYPTED]
-          let action = null
-          if (message.opType() === GIMessage.OP_ATOMIC) {
-            const opTypes = value.map(([type, value]) => type)
-            for (let i = 0; i < opTypes.length; i++) {
-              if (allowedActionType.includes(opTypes[i])) {
-                // NOTE: choose the first action with allowed opType
-                //       which also would be the only child action of OP_ATOMIC message
-                const v = value[i][1].valueOf()
-                action = !Object.getPrototypeOf(v)?._isSignedData ? v.action : v?.valueOf()?.action
-                break
-              }
+          const getAllowedMessageAction = (opType, opValue) => {
+            if (opType === GIMessage.OP_ATOMIC) {
+              const actions = opValue
+                .map(([t, v]) => getAllowedMessageAction(t, v.valueOf().valueOf()))
+                .filter(Boolean)
+              // TODO: Now we return the first action of list
+              //       because there is only one allowedAction in OP_ATOMIC message now.
+              //       But later we need to consider several child actions for A OP_ATOMIC message
+              return actions[0]
+            } else if (allowedActionType.includes(opType)) {
+              return opValue.action
+            } else {
+              return undefined
             }
-          } else if (allowedActionType.includes(message.opType())) {
-            action = value.action
-          } else {
-            return {}
           }
 
+          const action = getAllowedMessageAction(message.opType(), value)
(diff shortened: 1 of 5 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -891,9 +890,25 @@ export default ({
...
+          const allowedActionType = [GIMessage.OP_ACTION_ENCRYPTED, GIMessage.OP_ACTION_UNENCRYPTED]
+          let action = null
+          if (message.opType() === GIMessage.OP_ATOMIC) {
+            const opTypes = value.map(([type, value]) => type)
+            for (let i = 0; i < opTypes.length; i++) {
+              if (allowedActionType.includes(opTypes[i])) {
+                // NOTE: choose the first action with allowed opType
+                //       which also would be the only child action of OP_ATOMIC message
+                const v = value[i][1].valueOf()
+                action = !Object.getPrototypeOf(v)?._isSignedData ? v.action : v?.valueOf()?.action
```
corrideat: The `_isSignedData` isn't a stable API and is also weird. I'd change that to `.valueOf().valueOf()`, which always works, and add a comment explaining why.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -891,9 +890,25 @@ export default ({
         if (!value) throw new Error('Unable to decrypt message')
 
         const isMessageAddedOrDeleted = (message: GIMessage) => {
-          if (![GIMessage.OP_ACTION_ENCRYPTED, GIMessage.OP_ACTION_UNENCRYPTED].includes(message.opType())) return {}
+          const allowedActionType = [GIMessage.OP_ACTION_ENCRYPTED, GIMessage.OP_ACTION_UNENCRYPTED]
+          let action = null
+          if (message.opType() === GIMessage.OP_ATOMIC) {
```
corrideat: Good catch with `OP_ATOMIC`. However, there could technically be multiple messages in OP_ATOMIC, so I think it'd be better to call the parent function recursively with all actions.

corrideat (changes requested):

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: For the future `OP_ATOMIC` messages,
- Could it have more than two child messages of `OP_ACTION_UNENCRYPTED` and `OP_ACTION_ENCRYPTED`?
- Could it create two messages? Then, there could be a serious problem to set message hash in `createMessage` function. Because can not create two messages with one hash.
- Could it delete messages and add messages at once?

corrideat: > Could it have more than two child messages of OP_ACTION_UNENCRYPTED and OP_ACTION_ENCRYPTED?

Yes, you _could_ have any number of actions. (We don't generate such messages in GI as far as I know, but they are _possible_)

> Could it create two messages? Then, there could be a serious problem to set message hash in createMessage function. Because can not create two messages with one hash.

I don't follow entirely, but I don't think it'd be an issue with `createMessage` because it doesn't create OP_ATOMIC with multiple messages. However, in general, many 'messages' can have the same hash if part of OP_ATOMIC. Maybe the hash could be `${hash}-${index}`?

> Could it delete messages and add messages at once?

From the protocol POV, OP_ATOMIC is just a wrapper with an array of opcodes, and it can contain any number of operations and opcodes, except OP_ATOMIC itself.

corrideat:
@Silver-IT I'm approving this now because it's better than what was there before, but I still think that children of `OP_ATOMIC` must be handled.

corrideat (approved):

taoeffect (approved):
LGTM

PR #1904, opened 2024-04-03
Improve the performance of Chat page

### LINKED ISSUE #1897: Remove this.$forceUpdate as much as possible in Chat pages
### Problem

**Background**
`this.$forceUpdate()` forces the component to be re-rendered, and we use it many times in `ChatMain.vue` and in other pages/components too. The one of the main reasons why we used this function is make the component reactive to the asynchronous response such as `sbp` calls. Furthermore, in `ChatMain.vue`, more events are loaded and all the messages are forced to be re-rendered whenever user scrolls up. This definitely slows down the performance, and gives users the bad feeling.
**What to do**
`this.$forceUpdate` should be replaced to another as much as possible in order to remove redundant re-rendering and increase performance.

### Solution

Instead of using `this.$forceUpdate`, use `:key` or `:ref` to re-render only necessary components

Silver-IT:
### Summary of Changes

- Closes #1897 
- Fix error to display text in notification (When there is mentions inside the text)

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -738,7 +738,6 @@ export default ({
 
           for (;;) {
             const { done, value: event } = await newEventsStreamReader.read()
-            console.error({ done, event })
             if (done) break
 
             const state = this.messageState.contract
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -661,6 +661,16 @@ export default ({
         console.error(`Error while adding emotion for ${contractID}`, e)
       })
     },
+    generateNewChatRoomState () {
+      return {
+        settings: cloneDeep(this.chatRoomSettings),
+        attributes: cloneDeep(this.chatRoomAttributes),
+        users: cloneDeep(this.chatRoomMembers),
+        _vm: cloneDeep(this.currentChatVm),
+        messages: [],
+        onlyRenderMessage: true // NOTE: DO NOT RENAME THIS OR CHATROOM WOULD BREAK
+      }
+    },
     initializeState () {
       // NOTE: this state is rendered using the chatroom contract functions
       // so should be CAREFUL of updating the fields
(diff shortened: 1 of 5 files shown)


Silver-IT (commented):
Ready for the review!

taoeffect (changes requested):
Nice work @Silver-IT! It does seem a bit smoother. 😎 👍 

While reviewing this PR, I found a few issues, and while they might not be related to this PR specifically, I'm wondering if you can still address them in this PR?

### 1. Scrolling behavior broken

When new messages are added to a chatroom, and someone was scrolled all the way to the bottom, they do not continue scrolling downward to see the new messages as they should:

https://github.com/okTurtles/group-income/assets/138706/06116699-558b-407e-b5fd-d85b5adc3181

### 2. Extraneous error logging

There is a rogue `console.error`:

Screenshot 2024-04-03 at 6 42 36 PM

Resulting in console errors like this depending on how the chatroom views are accessed:

Screenshot 2024-04-03 at 6 43 52 PM
Screenshot 2024-04-03 at 6 42 31 PM

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -794,23 +795,17 @@ export default ({
...
-            const state = this.messageState.contract
-            const newState = await sbp('chelonia/in/processMessage', event, state)
-
-            if (!this.checkEventSourceConsistency(contractID)) return
-
-            Vue.set(this.messageState, 'contract', newState)
+        if (this.latestEvents.length > 0) {
+          let state = this.generateNewChatRoomState()
+          for (const event of this.latestEvents) {
+            state = await sbp('chelonia/in/processMessage', event, state)
```
taoeffect, line 804: @corrideat could you review these changes?

Silver-IT:
> Scrolling behavior broken

I remember this is what we discussed very long ago. That scroll works wrong only when using Multi Container Firefox addon. When you open two browsers(not using Multi Container Addon), and test, it would work fine as expected.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: This seems to be fine, although I'd keep the check for contract ID changes so that processing messages can be stopped if no longer needed

taoeffect (approved):
Nice!

PR #1906, opened 2024-04-04
Set skeleton size of preview image attachments

### LINKED ISSUE #1901: Set skeleton size of image attachments
### Problem

When we try to attach images in the chatroom, we can see the skeleton for the image while we are downloading the images from the server, and saving them as Blob in the browser. The size(width, height) of the skeleton should be same to the original image, because it also affects to the scroll position of the chat page.

### Solution

Save the dimension(width,.height) inside the message, and display skeleton based on it.

Silver-IT:
### Summary of Changes

- Closes #1901

### DIFF
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -47,6 +47,10 @@ export const messageType: any = objectMaybeOf({
   attachments: arrayOf(objectOf({
     name: string,
     mimeType: string,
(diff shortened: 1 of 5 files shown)


Silver-IT (commented):
Ready for the review!

taoeffect (approved):
Beautiful - nicely done @Silver-IT!

Also great that you noticed that this would be a good idea to do in the first place! 👏

PR #1907, opened 2024-04-04
Bug-fix for a DM related issue

### LINKED ISSUE #1905: A bug regarding the Chatroom DM creation
### Problem
When a user is part of multiple groups, and attempts to create a DM. A couple of issues happens. Let's say there is two groups a user (user-1 here) is part of,

**Group A:** user-1, user-a
**Group B:** user-1, user-b

switching to one group, I hit the new direct messages button and a modal for user list is displayed. This modal is supposed to only show the members of the currently active group but it displays users in both groups of user-1, which in this case is 'user-a' and 'user-b'. this crashes the DM and leads to user-b receiving this cryptic error as an in-app notification.

### Solution
Make sure the user-addition modal only displays the users of the currently active group.

SebinSong:
closes #1905

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -485,14 +485,12 @@ const getters = {
     return profiles
   },
   ourContactProfilesById (state, getters) {
-    const currentGroupProfileIds = Object.keys(getters.currentGroupState.profiles || {})
     const profiles = {}
     Object.keys(state.contracts)
       .filter(contractID => state.contracts[contractID].type === 'gi.contracts/identity')
       .forEach(contractID => {
         const attributes = state[contractID].attributes
-        if (attributes && // NOTE: this is for fixing the error while syncing the identity contracts
-          currentGroupProfileIds.includes(contractID)) {
+        if (attributes) { // NOTE: this is for fixing the error while syncing the identity contracts
           const username = checkedUsername(state, attributes.username, contractID)
           profiles[contractID] = {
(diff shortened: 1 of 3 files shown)


frontend/model/state.js
```diff
@@ -485,12 +485,14 @@ const getters = {
     return profiles
   },
   ourContactProfilesById (state, getters) {
```
SebinSong: this getter is named **our**ContactProfilesById, but the most important logic is missing inside it, which is only filtering the users that are members of the currently active group. So added the logic in as the fix.

taoeffect (commented):
Hmm, I'm not sure about this...

`ourContactProfilesById` sounds like it returns all contacts... regardless of group.

Perhaps instead of applying the fix there, one of two options could be done:

- A new getter called `currentGroupContactProfilesById` is created, that calls `ourContactProfilesById` and does the filtering based on the results of that (and then use that new getter where appropriate)
- The filtering is done inside of the DM modal

Thoughts?

corrideat:
I think this is mostly fine, except that that getter could also be used for notifications (which could be for different groups). Because of this, I'd either confirm that it doesn't affect anything else or add a new getter with this filter on.

SebinSong:
@taoeffect @corrideat Okay, will create a new getter with this filter in `state.js`

taoeffect (approved):
Beautiful!

PR #1910, opened 2024-04-05
#1884 make more uses of 'bypassUI: true' / speed up the cypress test as much as we can

### LINKED ISSUE #1884: Speed up the tests by making more use of `bypassUI: true`
### Problem

While reviewing PR #1883, I checked to see how many function calls were using `bypassUI: true`, and was surprised to see that many were not using it.

### Solution

Go through each of the commands that supports `bypassUI: true` in `commands.js`, search the tests for them (e.g. search for `giSwitchUser`, `giLogin` etc), and make it so that all the `.spec.js` files are using `bypassUI: true` except for up to two tests.

**Make sure you leave the test in `group-chat.spec.js` [on line 210 as it is](https://github.com/okTurtles/group-income/pull/1883/files#diff-13e9388468b0e4452c95a09cd93db8efe64588871bb2471c6983f8e4dc2444e9R210), as that is an important test to detect a specific test @corrideat had to make `bypassUI: false` to catch a specific bug in #1883**

SebinSong:
closes #1884

### DIFF
--- a/test/cypress/integration/group-chat-scrolling.spec.js
+++ b/test/cypress/integration/group-chat-scrolling.spec.js
@@ -64,7 +64,7 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
 
   it(`user1 creates '${groupName}' group and joins "${CHATROOM_GENERAL_NAME}" channel by default and sends 15 messages`, () => {
     cy.visit('/')
-    cy.giSignup(user1)
+    cy.giSignup(user1, { bypassUI: true })
     me = user1
 
     cy.giCreateGroup(groupName, { bypassUI: true })
--- a/test/cypress/integration/group-chat-scrolling.spec.js
+++ b/test/cypress/integration/group-chat-scrolling.spec.js
@@ -30,7 +30,7 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
 
   function sendMessage (message) {
     cy.getByDT('messageInputWrapper').within(() => {
-      cy.get('textarea').clear().type(`${message}{enter}`)
+      cy.get('textarea').clear().type(`${message}{enter}`, { delay: 0 })
       cy.get('textarea').should('be.empty')
     })
     cy.getByDT('conversationWrapper').within(() => {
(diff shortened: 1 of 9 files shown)


test/cypress/integration/group-chat-scrolling.spec.js
```diff
@@ -30,7 +30,7 @@ describe('Send/edit/remove messages & add/remove emoticons inside group chat', (
 
   function sendMessage (message) {
     cy.getByDT('messageInputWrapper').within(() => {
-      cy.get('textarea').clear().type(`${message}{enter}`)
+      cy.get('textarea').clear().type(`${message}{enter}`, { delay: 0 })
```
SebinSong, line 33: `delay: 0` is an option to give no delay between key-strokes here ([reference](https://docs.cypress.io/api/commands/type#Arguments)).  `sendMessage` here is used for an action that is repeated many times (e.g. `user1 sends 15 messages` etc), so thought I would want to make it a little faster.

test/cypress/support/commands.js
```diff
@@ -605,27 +608,44 @@ Cypress.Commands.add('randomPaymentMethodInIncomeDetails', () => {
   })
 })
 
-Cypress.Commands.add('giAddNewChatroom', (
-  name, description = '', isPrivate = false
-) => {
+Cypress.Commands.add('giAddNewChatroom', ({
+  name, description = '', isPrivate = false, bypassUI = false
+}) => {
```
SebinSong, line 613: I noticed `giAddNewChatroom` command is used many times in succession in `group-chat.spec.js` (e.g. `user1 creates several channels and logout` etc..) but it didn't have `bypassUI: true` option implemented yet. So I thought I would implement&use it here to make that test-suite take shorter time.

taoeffect (approved):
@SebinSong - this is awesome!! 😄 

Look - almost a 2 minute improvement! 🎉 

**Old:**

Screenshot 2024-04-05 at 4 57 50 PM

**New:**

Screenshot 2024-04-05 at 4 57 56 PM

PR #1911, opened 2024-04-05
Improvements

### LINKED ISSUE #1908: Ability to specify the type of key being requested to OP_KEY_REQUEST
### Problem

Currently `OP_KEY_REQUEST` is treated as a general key request that contracts must know how to respond to by assuming the keys that are being requested. It would be nice if we could specify the specific keys that we want to support ambiguous situations.

### Solution

Add a `request` parameter that represents a string describing which keys are being requested.

For this issue, it's sufficient to simply support `'*'` initially, and later on figure out details on what this string should look like otherwise.

corrideat:
* Ability to specify type of key requested in `OP_KEY_REQUEST` (fixes #1908)
* Prevent download secret parameters from being logged by SBP

### DIFF
--- a/frontend/views/components/Avatar.vue
+++ b/frontend/views/components/Avatar.vue
@@ -69,7 +69,7 @@ export default ({
         return
       }
       try {
-        const blob = await sbp('chelonia/fileDownload', src)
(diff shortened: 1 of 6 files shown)


taoeffect (approved):
Nicely done!

PR #1912, opened 2024-04-05
1656 kv store

### LINKED ISSUE #1656: Move recent user activity (login dates) in server-side key-value store instead of in contracts
### Problem

The group contract is currently being filled with `'gi.contracts/group/updateLastLoggedIn'` messages, which really are not that important and don't need to be kept in the contract chain, as that slows down contract syncing and wastes space unnecessarily.

Similar information (that is constantly updated) should be stored outside of the contract chain - directly in the key-value store.

### Solution

Remove `'gi.contracts/group/updateLastLoggedIn'` and instead set this information directly on the key-value store. Make sure to encrypt it using the latest group key, and make sure that only group members can write to these keys.

Create an API to use for this purpose - to create a secure key-value store that only authorized users can write to. Make sure that keys with the `private` prefix cannot be accessed / read via this API or via the `/file` API (or any other APIs that return the contents of the database).

EDIT: e.g. `_private_kvstore_{contractID}_{key}`

corrideat:
Fixes #1656

### DIFF
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -469,9 +469,30 @@ route.POST('/kv/{contractID}/{key}', {
     return Boom.unauthorized(null, 'shelter')
   }
 
-  const existingSize = Buffer.from(await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`) ?? '').byteLength
+  const existing = await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`)
+
+  // Some protection against accidental overwriting by implementing the if-match
+  // header
+  if (request.headers['if-match']) {
+    if (!existing) {
+      return Boom.preconditionFailed()
+    }
+    const expectedEtag = request.headers['if-match']
+    if (expectedEtag === '*') {
+      // pass through
+    } else {
+      // "Quote" string
+      const cid = JSON.stringify(createCID(existing))
+      if (!expectedEtag.split(',').map(v => v.trim()).includes(cid)) {
+        return Boom.preconditionFailed()
+      }
+    }
+  }
+
+  const existingSize = existing ? Buffer.from(existing).byteLength : 0
   await sbp('chelonia/db/set', `_private_kv_${contractID}_${key}`, request.payload)
   await sbp('backend/server/updateSize', contractID, request.payload.byteLength - existingSize)
+  await sbp('backend/server/broadcastKV', contractID, key, request.payload)
 
   return h.response().code(204)
 })
@@ -494,7 +515,12 @@ route.GET('/kv/{contractID}/{key}', {
     return Boom.unauthorized(null, 'shelter')
   }
 
-  return await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`)
+  const result = await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`)
+  if (!result) {
+    return Boom.notFound()
+  }
+
+  return h.response(result).etag(createCID(result))
 })
 
 // SPA routes
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -447,6 +447,56 @@ route.POST('/deleteFile/{hash}', {
   return h.response()
 })
 
+route.POST('/kv/{contractID}/{key}', {
+  auth: {
+    strategies: ['chel-shelter'],
+    mode: 'required'
+  },
+  payload: {
+    parse: false,
+    maxBytes: 6 * MEGABYTE, // TODO: make this a configurable setting
+    timeout: 10 * SECOND // TODO: make this a configurable setting
+  }
+}, async function (request, h) {
+  const { contractID, key } = request.params
+  console.debug(`POST /kv/${contractID}/${key}`)
+
+  if (key.startsWith('_private')) {
+    return Boom.notFound()
+  }
+
+  if (!ctEq(request.auth.credentials.billableContractID, contractID)) {
+    return Boom.unauthorized(null, 'shelter')
+  }
+
+  const existingSize = Buffer.from(await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`) ?? '').byteLength
+  await sbp('chelonia/db/set', `_private_kv_${contractID}_${key}`, request.payload)
+  await sbp('backend/server/updateSize', contractID, request.payload.byteLength - existingSize)
+
+  return h.response().code(204)
+})
+
+route.GET('/kv/{contractID}/{key}', {
+  auth: {
+    strategies: ['chel-shelter'],
+    mode: 'required'
+  },
+  cache: { otherwise: 'no-store' }
+}, async function (request, h) {
+  const { contractID, key } = request.params
+  console.debug(`GET /kv/${contractID}/${key}`)
+
+  if (key.startsWith('_private')) {
+    return Boom.notFound()
+  }
+
+  if (!ctEq(request.auth.credentials.billableContractID, contractID)) {
+    return Boom.unauthorized(null, 'shelter')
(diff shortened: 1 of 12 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -1247,7 +1247,79 @@ export default (sbp('sbp/selectors/register', {
...
+    })
+    this.pubsub.pub(contractID, serializedData)
+  },
+  'chelonia/kv/set': async function (contractID: string, key: string, data: Object, {
+    contractName,
+    innerSigningKeyId,
+    encryptionKeyId,
+    signingKeyId
+  }: {
+    contractName: string,
```
taoeffect: If possible, since we're already passing in the `contractID`, to avoid a potential mismatch, can you remove `contractName` as a parameter and fetch it based on the `contractID` that's passed in?

backend/routes.js
```diff
@@ -447,6 +447,99 @@ route.POST('/deleteFile/{hash}', {
...
+  }
+
+  if (!ctEq(request.auth.credentials.billableContractID, contractID)) {
+    return Boom.unauthorized(null, 'shelter')
+  }
+
+  const existing = await sbp('chelonia/db/get', `_private_kv_${contractID}_${key}`)
+
+  // Some protection against accidental overwriting by implementing the if-match
+  // header
```
taoeffect, line 472: A bit more explanation here would be helpful, e.g. the specific scenario that this logic helps with

backend/routes.js
```diff
@@ -447,6 +447,99 @@ route.POST('/deleteFile/{hash}', {
...
+      if (!expectedEtag.split(',').map(v => v.trim()).includes(cid)) {
+        return Boom.preconditionFailed()
+      }
+    }
+  }
+
+  try {
+    const serializedData = JSON.parse(request.payload.toString())
+    const { contracts } = sbp('chelonia/rootState')
+    // Check that the height is the latest value
```
taoeffect: Could you add a comment here explaining why this height check is being done? It's not completely clear, but I'm guessing it's to make sure that we're using the latest keys for the signature

backend/routes.js
```diff
@@ -447,6 +447,99 @@ route.POST('/deleteFile/{hash}', {
...
+      serializedData,
+      meta: key
+    })
+  } catch (e) {
+    return Boom.badData()
+  }
+
+  const existingSize = existing ? Buffer.from(existing).byteLength : 0
+  await sbp('chelonia/db/set', `_private_kv_${contractID}_${key}`, request.payload)
+  await sbp('backend/server/updateSize', contractID, request.payload.byteLength - existingSize)
```
taoeffect, line 527: Why isn't this just passing in `request.payload.byteLength`?

backend/server.js
```diff
@@ -155,7 +163,7 @@ sbp('sbp/selectors/register', {
   },
   'backend/server/updateSize': async function (resourceID: string, size: number) {
     const sizeKey = `_private_size_${resourceID}`
-    if (!(size >= 0)) {
+    if (!Number.isSafeInteger(size)) {
```
taoeffect, line 166: ~~Probably still want to make sure the value is positive~~

EDIT: erm, no... actually, I'm confused by this line below:

```js
await sbp('chelonia/db/set', sizeKey, (existingSize + size).toString(10))
```

It seems to imply the size can only increase. But if the KV sets a smaller value, it should decrease...

frontend/controller/actions/group.js
```diff
@@ -995,6 +995,29 @@ export default (sbp('sbp/selectors/register', {
...
+      const now = new Date().toISOString()
+
+      // Wait for any pending operations (e.g., sync) to finish
+      await sbp('chelonia/queueInvocation', contractID, async () => {
+        const current = await sbp('chelonia/kv/get', contractID, 'lastLoggedIn')?.data || {}
+        current[userID] = now
+        await sbp('chelonia/kv/set', contractID, 'lastLoggedIn', current, {
+          encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
+          signingKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk')
+        })
```
taoeffect, line 1026: Missing `catch`

EDIT: I forgot that the `try` block on line 999 should catch this....

frontend/controller/actions/group.js
```diff
@@ -995,6 +995,29 @@ export default (sbp('sbp/selectors/register', {
...
+      const userID = rootState.loggedIn?.identityContractID
+
+      if (!userID) {
+        throw new Error('Unable to update last logged in without an active session')
+      }
+      const now = new Date().toISOString()
+
+      // Wait for any pending operations (e.g., sync) to finish
+      await sbp('chelonia/queueInvocation', contractID, async () => {
+        const current = await sbp('chelonia/kv/get', contractID, 'lastLoggedIn')?.data || {}
```
taoeffect, line 1021: also missing `catch`

shared/domains/chelonia/chelonia.js
```diff
@@ -546,22 +546,36 @@ export default (sbp('sbp/selectors/register', {
...
                       contractID: msg.channelID,
-                      contractName: rootState.contracts[msg.channelID].type,
                       serializedData: msg.data
                     }))
                   } catch (e) {
                     console.error(`[chelonia] Error processing pub event for ${msg.channelID}`, e)
                   }
                 }
-              : v
+              : k === NOTIFICATION_TYPE.KV
```
taoeffect: This is rather difficult to read IMO. Can we use something like a `switch` statement maybe instead? (or a function map?)

shared/pubsub.js
```diff
@@ -417,6 +422,11 @@ const defaultMessageHandlers = {
     // No need to reply.
   },
 
+  [NOTIFICATION_TYPE.KV] ({ channelID, key, data }) {
+    console.log(`[pubsub] Received KV update from channel ${channelID} ${key}:`, data)
```
taoeffect, line 426: This should probably be `console.debug`

frontend/controller/actions/group.js
corrideat: It doesn't have `catch` on purpose, as proceeding would be an error

frontend/controller/actions/group.js
corrideat: Also doesn't have `catch` on purpose, as an error here means updating the last logged in info (the caller's intent) failed.

frontend/controller/actions/group.js
taoeffect: I forgot that the `try` block on line 999 should catch this....

shared/pubsub.js
corrideat: Maybe, but then `console.log` is used for these types of logs elsewhere (see `main.js:228` and `pubsub.js:421`)

backend/routes.js
```diff
@@ -447,6 +447,118 @@ route.POST('/deleteFile/{hash}', {
...
+
+route.GET('/kv/{contractID}/{key}', {
+  auth: {
+    strategies: ['chel-shelter'],
+    mode: 'required'
+  },
+  cache: { otherwise: 'no-store' }
+}, async function (request, h) {
+  const { contractID, key } = request.params
+  console.debug(`GET /kv/${contractID}/${key}`)
```
taoeffect: note to self: double-check if this gets logged twice (once by Hapi, and once here)

taoeffect (approved):
Fantastic work @corrideat ❗ 😄 🎉

Silver-IT (commented):
Nice work!

PR #1921, opened 2024-04-13
Have grunt enable git-fsck in local git config

### LINKED ISSUE #359: Add fsckObjects = true to a project .gitconfig file so that everyone has it
### Problem

Git, [by default](https://groups.google.com/forum/#!topic/binary-transparency/f-BI4o8HZW0), doesn't actually verify the sha-based integrity of files (which is crazy, especially given that sha1 is already broken, at the very least it should attempt to verify it).

See that thread, scary:

> interestingly, these settings cause a failure when cloning
> https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware
> as noted in https://bugs.debian.org/743227
>
> If anywhere was a place to look for evidence of binary tampering, i'd
> expect repos of firmware blobs to be a spot to look. so, yikes.

### Solution

Everyone should have this in their `~/.gitconfig`:

```
# for $deity's sake, check that anything we're getting is complete and sane on a regular basis
# https://groups.google.com/forum/#!topic/binary-transparency/f-BI4o8HZW0
[transfer]
fsckobjects = true
[fetch]
fsckobjects = true
[receive]
fsckObjects = true
```

We can ensure this (I think), by having a `.gitconfig` in this project repo, or by running `git config` commands via `grunt`.

Also, I'll create a separate issue to get rid of NPM completely and use git submodules instead, as that might be worth doing.

snowteamer:
Close #359

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,8 @@
+# For $deity's sake, check that anything we're getting is complete and sane on a regular basis.
+# https://groups.google.com/forum/#!topic/binary-transparency/f-BI4o8HZW0
+[transfer]
+    fsckObjects = true
+[fetch]
+    fsckoOjects = true
+[receive]
+    fsckObjects = true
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Thank you @snowteamer!

PR #1925, opened 2024-04-15
File deletion

### LINKED ISSUE #1920: File deletion isn't fully implemented
### Problem

While testing #1892 I discovered that deleting a file attachment doesn't actually delete it.

It turns out this is for two reasons:

1. We don't overwrite `'chelonia/db/delete'` at all in `backend/database.js`
2. We completely forgot to implement the `deleteData` function for `./database-${persistence}.js` files

### Solution

1. Implement `deleteData` for both `fs` and `sqlite` persistence modes
2. Add code to overwrite `'chelonia/db/delete'` near line 160 in `backend/database.js`

corrideat:
Fixes #1920

### DIFF
--- a/backend/database.js
+++ b/backend/database.js
@@ -122,7 +122,7 @@ export default async () => {
   // - load and initialize the selected storage backend
   // - then overwrite 'chelonia/db/get' and '-set' to use it with an LRU cache
   if (persistence) {
-    const { initStorage, readData, writeData } = await import(`./database-${persistence}.js`)
+    const { initStorage, readData, writeData, deleteData } = await import(`./database-${persistence}.js`)
 
     await initStorage(options[persistence])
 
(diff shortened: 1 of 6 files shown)


backend/database.js
```diff
@@ -156,6 +156,17 @@ export default async () => {
...
         await writeData(key, value)
         cache.set(key, value)
+      },
+      'chelonia/db/delete': async function (key: string): Promise<void> {
+        checkKey(key)
+        if (key.startsWith('_private_immutable_')) {
+          const existingValue = await readData(key)
+          if (existingValue !== undefined) {
+            throw new Error('Cannot delete already set immutable key')
+          }
```
taoeffect: This check doesn't make sense, as it implies `_private_immutable` keys would be deleted if they don't exist (at which point, there's no point in deleting them).

Also, to be consistent, if any check is performed, it should be for `_private_immutable` instead of `_private_immutable_`

taoeffect (changes requested):
Excellent PR @corrideat! Only one minor issue found

taoeffect (approved):
🥇

PR #1928, opened 2024-04-15
Allow group creator to delete messages in any group chatroom

### LINKED ISSUE #1919: Allow group creator to delete messages in any group chatroom
### Problem

According to the discussion [here](https://github.com/okTurtles/group-income/pull/1892#discussion_r1558539645), a new feature is needed to be implemented. That is to allow group creators to delete messages in any group chatrooms.

### Solution

Implement it

Silver-IT:
### Summary of Changes

- Close #1919


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -630,58 +631,58 @@ export default ({
...
+      const originalText = message.text || (message.attachments?.[0].name || '')
+      const text = convertToMarkdown(swapUserIDForUsername(originalText))
+      const promptConfig = {
+        heading: L('Delete message'),
+        question: L('Are you sure you want to delete this message permanently?{textPreview}', {
+          textPreview: `<p><span class="custom-markdown-content">${text}</span></p>`
+        }),
+        primaryButton: L('Yes'),
+        secondaryButton: L('Cancel')
+      }
```
taoeffect: I have a few requests here:

1. I don't think we need to show a re-rendering of the message here. We can simply ask if they're sure if they want to delete the message permanently.
2. I think the function `convertToMarkdown` is named in a rather confusing way. This is something I missed from my previous PR review of @SebinSong's PR. It sounds like it's taking HTML and converting it to Markdown. But that's not what it's doing. Instead, it's taking Markdown and converting it to HTML (the opposite). So could you please refactor/rename this function to something like, `renderMarkdown`?
3. Please use two different strings here, depending on the situation:
  - No attachments: `Are you sure you want to delete this message permanently?`
  - Attachments: `Are you sure you want to delete this message and its file attachments permanently?`

taoeffect (changes requested):
Nice work @Silver-IT! One minor request

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: Okay, Will do.

Silver-IT (commented):
Final review ready, @taoeffect!

Actually, I have one thing in my mind about the updates of this PR.
Since we don't check the group creator in the `validate` of `deleteMessage` function, if someone will try to run `sbp('gi.actions/group/deleteMessage')` in the console, he can easily delete any messages.
Of course, in the production server, we will not import `sbp` to the `window` object.
```
if (process.env.NODE_ENV === 'development' || window.Cypress || process.env.EXPOSE_SBP === 'true') {
  // In development mode this makes the SBP API available in the devtools console.
  window.sbp = sbp
}
```

However, should we check this in the `validate` anyway?

taoeffect:
@Silver-IT I think you can in the `process` function throw an error if the `innerSigningContractID` is not either the `groupCreatorID` or the identityContractID of the message creator, but @corrideat can confirm.

corrideat:
@Silver-IT It's not true that anyone can delete messages, if you add the right check. In the past, you'd have used `meta.username` (which could be forged). Now, that doesn't exist but you can check the `innerSigningContractID` (which cannot be forged). So, you could check whether the `innerSigningContractID` matches the author or the creator.

Silver-IT:
@corrideat, if the `sbp` is imported to the `window` object, it's true that anyone can delete messages. And maybe some other actions too.
Of course, `innerSigningContractID` can not be forged, but it isn't used to check if the user is the group creator. That's because there is no way to to access the global state and to get `groupCreatorID` in process function. I don't think we should send `groupCreatorID` in data, because if someone wants to delete the other's message using `sbp` function in console, he can also set that data and the `groupCreatorID` too.

taoeffect:
> if the sbp is imported to the window object, it's true that anyone can delete messages. And maybe some other actions too.

This isn't true. `sbp` cannot be used to delete messages. Everyone else in the group will reject invalid messages.

> Of course, innerSigningContractID can not be forged, but it isn't used to check if the user is the group creator. That's because there is no way to to access the global state and to get groupCreatorID in process function. I don't think we should send groupCreatorID in data, because if someone wants to delete the other's message using sbp function in console, he can also set that data and the groupCreatorID too.

The `groupCreatorID` is stored under the group settings when the group is created, and it is already used in the group contract for authentication. It should be used in this PR as well to perform the authentication. Please update the PR so that the `innerSigningContractID` is checked against both the `groupCreatorID` and the person who originally created the message.

EDIT: this conversation inspired two new security related issues:

- https://github.com/okTurtles/group-income/issues/1956
- https://github.com/okTurtles/group-income/issues/1957

Silver-IT:
> Please update the PR so that the `innerSigningContractID` is checked against both the `groupCreatorID` and the person who originally created the message.

I think this should be done in the `validate` function where we can access the global state and `groupCreatorID`.

Silver-IT:
> This isn't true. sbp cannot be used to delete messages. Everyone else in the group will reject invalid messages.

It's true that anyone in the chatroom (not group) can delete any messages in the group channel, no matter they are group creator or not. I have already tested it. Running the following format of function call **in the console** will remove a message.

```
sbp('gi.actions/chatroom/deleteMessage', {
  contractID: 'z9brRu3VUrJhFa2yyy4XPF7Gat26VXNpg3gBP5j5SBGHGpAVjkNe',
  data: {
    hash: 'z9brRu3VRaS6bVnFWyaUoywChf3RWxXDuWAPFPpeEkbvBb3qqf4g',
    manifestCids: [],
    messageSender: 'z9brRu3VRkxTNVgntSrPicVG7vcvYmpjEwSL8HUqnfggAJD6gFjJ'
  }
})
```

taoeffect:
Ah, I meant hypothetically, if the function were properly implemented to check the innerSigningContractID permissions 😅

frontend/model/contracts/chatroom.js
```diff
@@ -521,13 +529,28 @@ sbp('chelonia/defineContract', {
...
+        })(data)
+
+        const rootGetters = sbp('state/vuex/getters')
+
+        if (innerSigningContractID !== data.messageSender) {
+          if (rootGetters.isDirectMessage(contractID)) {
+            throw new TypeError(L('Only the person who sent the message can delete it.'))
+          } else {
+            const groupID = rootGetters.groupIdFromChatRoomId(contractID)
+            if (sbp('state/vuex/state')[groupID]?.settings.groupCreatorID !== innerSigningContractID) {
```
taoeffect: Please merge `master` and update all uses of `groupCreatorID`, which is now `groupOwnerID` and lives above `.settings`

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -121,7 +121,8 @@ export default ({
       }
     },
     type: String,
-    isMsgSender: Boolean
+    isMsgSender: Boolean,
+    isGroupCreator: Boolean
```
taoeffect, line 125: Instead of passing this in via a parameter, maybe a computed property can be used instead to check `getters.ourIdentityContractId === getters.currentGroupOwnerID`?

taoeffect (changes requested):
Nice work @Silver-IT! 👏 

Looks like there's a merge conflict. Also added a couple of comments

frontend/views/containers/chatroom/MessageActions.vue
Silver-IT: I would like to use the `isGroupCreator` property instead of adding a new computed variable in the `MessgeActions` component. The reason is because we are using `isGroupCreator` property in the `ChatAttachmentPreview` component either, so we should also make a computed variable in `ChatAttachmentPreview` component if we go this way.

Additionally, using a computed variable instead of prop requires to add 4 getters(`isDirectMessage`, `currentGroupOwnerID`, `ourIdentityContractId`, `currentChatRoomId`) which seems to be more untidy.

frontend/common/translations.js
```diff
@@ -133,13 +133,14 @@ export default function L (
...
   let url = `/app/dashboard?modal=UserSettingsModal&tab=application-logs&errorMsg=${encodeURI(error.message)}`
+  const target = !sbp('state/vuex/state').loggedIn ? 'target="_blank"' : ''
   if (!sbp('state/vuex/state').loggedIn) {
     url = 'https://github.com/okTurtles/group-income/issues'
   }
   return {
     reportError: L('"{errorMsg}". You can {a_}report the error{_a}.', {
       errorMsg: error.message,
-      'a_': `<a class="link" target="_blank" href="${url}">`,
+      'a_': `<a class="link" ${target} href="${url}">`,
```
taoeffect: I have the same change request here as I did in the other PR: https://github.com/okTurtles/group-income/pull/1965#discussion_r1585236100

Could these lines please be reverted?

taoeffect (changes requested):
Nice work @Silver-IT! Just one change request (same as the other PR)

frontend/common/translations.js
Silver-IT: Yeah. I am sure that it was my mistake to make this change in this PR.

taoeffect (approved):
LGTM!

PR #1929, opened 2024-04-16
Remove pushClientErrorHandler from pubsub.js

SebinSong:
This is a small PR for removing `pushClientErrorHandler` block written for a pubsub error with `REQUEST_TYPE.PUSH_ACTION` message type.

Currently this block does nothing but generating a console error message, which is what `console.warn()` in L456 already does. 
For the context, the thinking behind creating the `pushClientErrorHandler` logic  was **in case** a certain error might need some kind of post action in response to it. For now, every  push-related error will just be silently ignored so that it doesn't affect users using the app.

### DIFF
--- a/shared/pubsub.js
+++ b/shared/pubsub.js
@@ -453,14 +453,7 @@ const defaultMessageHandlers = {
       }
       case REQUEST_TYPE.PUSH_ACTION: {
         const { actionType, message } = data
-        const errorHandler = actionType ? pushClientErrorHandler[actionType] : null
-
         console.warn(`[pubsub] Received ERROR for PUSH_ACTION request with the action type '${actionType}' and the following message: ${message}`)
-        if (errorHandler) {
-          errorHandler()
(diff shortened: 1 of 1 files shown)


shared/pubsub.js
```diff
@@ -453,14 +453,7 @@ const defaultMessageHandlers = {
       }
       case REQUEST_TYPE.PUSH_ACTION: {
         const { actionType, message } = data
-        const errorHandler = actionType ? pushClientErrorHandler[actionType] : null
-
         console.warn(`[pubsub] Received ERROR for PUSH_ACTION request with the action type '${actionType}' and the following message: ${message}`)
```
SebinSong, line 456: @taoeffect I was about to add `console.warn` below as we discussed, but then realised there is already one up here(L 456). So decided to just remove the duplicate part. (hope it sounds good!)

taoeffect (approved):

PR #1933, opened 2024-04-16
#1891 - remove allowed file-extension list from the app

### LINKED ISSUE #1891: Reevaluating file extension handling: Allow all extensions
### Problem

In PR #1877, there's a list of 'common' extensions defined in `CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS`. This list is used to decide which files to allow, and files without any of those extensions will fail with a message that reads 'Unsupported file type'.

I'm not sure what the purpose of such a list was. I could understand if it's being used to decide how to process a file (in this case, I'd only use it as a hint rather than as an absolute), but in this case it's only used for the purpose of allowing or blocking the files that can be submitted.

Furthermore, the list includes some files that are not common at all outside of (web)dev circles, like `php`, `sh`, `js`, `vue` or `xul` while excluding others that _are_ common, like `flac`, `dwf` (AutoCAD), `key` (Keynote), etc. The point is that such lists can't possibly be comprehensive enough without being too long to the point that they lose meaning.

If the purpose was to prevent 'dangerous' files, note that some of those extensions can also be malicious. For example, `js` files in Windows can do anything an executable can do and are a common way of distributing malware (the same applies for `js` files in general on all platforms, although someone using, e.g., `node`, presumably knows what they're doing). Similarly, `pdf` files can contain scripts and can be malicious. The same goes for `jar` files (can contain arbitrary Java code), `doc` (can contain macros), etc. So, the list also fails at protecting users from potentially dangerous file types.

### Solution

I think the list should be removed entirely, or at least not used to decide the file types that are allowed. Some alternatives are:

* Removing the list and allowing all file types
* Removing the list and allowing all file types. Optionally flagging some potentially dangerous extensions in the UI or changing their extension when downloading (for example, `.exe` becomes `.exe.download`). I don't like this idea too much, but it's maybe a compromise.

In cases where the extension is used to determine a file type (which, like I said, I'd use it only as a hint), keeping a list internally is fine (for example, to treat images or videos especially). This refers only to the part about blocking files without a pre-approved extensions.

SebinSong:
closes #1891

Along with the requirement described in #1891 , preventing directories from being attached to the chat.
When a directory is attached, it will just be silently ignored.


frontend/views/containers/chatroom/file-attachment/ChatFileAttachmentWarningModal.vue
```diff
@@ -4,10 +4,10 @@ modal-template(
   :a11yTitle='L("Chat attachment too large modal")'
 )
   template(slot='title')
-    span {{ warningMessages.title }}
+    i18n File too large
 
   .c-content
-    span {{ warningMessages.content }}
+    i18n(:args='{ sizeLimit: config.sizeLimit }') That file is too large and cannot be uploaded. The limit is {sizeLimit} MB
```
taoeffect, line 10: Think this will show `6291456 MB`, which I think is wrong 😆

taoeffect (changes requested):
Nicely done @SebinSong! One minor issue!

frontend/views/containers/chatroom/file-attachment/ChatFileAttachmentWarningModal.vue
SebinSong: @taoeffect Right, good catch. Fixed it.

taoeffect (changes requested):
Nice work @SebinSong!

I tested this and the "file too large" dialog appears ... usually.

When I dragged and dropped a 337MB `.mkv` file onto it, the "too large" dialog didn't appear. It did appear though for a 420MB `.mp4` file... why doesn't it appear for the `.mkv` file?

SebinSong:
@taoeffect 

the cause is that there is no MIME type returned when the browsers (both firefox, chrome) reads the attached `.mkv` file and that makes the chatroom decide it is a directory. From my research, it appears that browsers returns an empty string for the mime-type of some uncommon file types.
([This ](https://bugzilla.mozilla.org/show_bug.cgi?id=1451786) is a relevant firefox issue reported several years ago if you are curious)





So I updated the directory-checking logic to take this into account.([here](https://github.com/okTurtles/group-income/pull/1933/commits/4a7ac73e4f2f6eb356fc21fbe3ff9f2fab546b90))
Now the chatroom correct detects `.mkv` as a file

taoeffect (approved):
Interesting - thanks for the background on why that didn't work!

Nicely done @SebinSong! 👏

PR #1939, opened 2024-04-18
Add two grunt commands

### LINKED ISSUE #10: Add  `grunt deploy` and/or `grunt serve` (or  similar)
## Problem

We'll at some point need to be able to deploy GI to a running server.

## Solution

**Background:**

- Right now we have `grunt dev` as the only way of serving the app.
- We also have an [open issue for `chel serve`](https://github.com/okTurtles/chel/issues/29) which is the ideal way that this issue would be closed
- A temporary alternative to `chel serve` would be implementing `grunt serve` - the idea of which would be to serve an optimized production build of the app.

I think whether we do `chel serve` or `grunt serve` first, we need a command to generate a build of the app that is production optimized. Currently we only generate a development build.

So a possible route forward would be to:

- Implement a `grunt deploy` command that uses `esbuild` to create a production bundle of the app
- Then implement `grunt serve <folder>` that just runs the server on port 8000 by default (and not `browsersync`). Make sure to take into account the `contracts` folder (perhaps `grunt deploy` would copy the `contracts` folder into a directory containing the groupincome bundle).
  - Note: `chel serve` is essentially the same thing, but more difficult to implement because the `chel` command doesn't have the Chelonia server in it (that's still here in Group Income under the `backend/` folder)

---------

### Old Commentary/notes

See:
- http://vuejs.org/guide/application.html#Deploying_for_Production
- https://github.com/hughsk/envify
- **excellent:** https://gist.github.com/rauchg/8199de60db48026a6670620a1c33b700#production-optimizations
- Related issue: #41

EDIT: Another possibility is making it possible install a `groupincome` command via `npm install -g` and running that.

The point is our current approach of running `grunt dev` is not appropriate for deploying to a server, and we need something simpler and without the overhead of `grunt dev`.

Silver-IT:
### Summary of Changes

- Closes #10


backend/database.js
```diff
@@ -18,12 +18,13 @@ const production = process.env.NODE_ENV === 'production'
 const persistence = process.env.GI_PERSIST || (production ? 'fs' : undefined)
 
 // Default database options. Other values may be used e.g. in tests.
+const dbRootPath = process.env.DB_ROOT_PATH || ''
 const options = {
   fs: {
-    dirname: './data'
+    dirname: `.${dbRootPath}/data`
```
taoeffect: `.${dbRootPath}/data` is not going to give users the directory they expect.

Typically users do something like:

```
DB_PATH=/path/to/db GI_PERSIST=fs grunt dev
```

But with this command that will result in the app looking in `./path/to/db/data`

Completely unexpected, in a different location entirely.

So if you're going to add an environment variable, it would be better to do something like:

```js
const dbRootPath = process.env.DB_PATH || './data'
// ...
dirname: dbRootPath
```

corrideat: I'm not a fan of the `.${dbRootPath}` thing and would replace it with `${dbRootPath}`.

frontend/main.js
```diff
@@ -201,12 +201,16 @@ async function startApp () {
...
           const ourVersion = process.env.GI_VERSION
           const theirVersion = msg.data.GI_VERSION
 
           const ourContractsVersion = process.env.CONTRACTS_VERSION
           const theirContractsVersion = msg.data.CONTRACTS_VERSION
-          if (ourVersion !== theirVersion || ourContractsVersion !== theirContractsVersion) {
+
+          const isContractVersionDiff = ourContractsVersion !== theirContractsVersion
+          const isGIVersionDiff = ourVersion !== theirVersion
+          if (isContractVersionDiff || (!isProduction && isGIVersionDiff)) {
```
corrideat: Why is the production environment a factor here?

Gruntfile.js
```diff
@@ -564,11 +630,10 @@ module.exports = (grunt) => {
     // and then we build the main bundle since it depends on manifests.json
     await Promise.all([buildContracts.run(), buildContractsSlim.run()])
       .then(() => {
-        return genManifestsAndDeploy(distContracts, packageJSON.contractsVersion)
-      })
-      .then(() => {
-        return Promise.all([buildMain.run(), buildServiceWorkers.run()])
+        const dest = !this.flags.deploy ? undefined : `./${distDir}/data`
```
corrideat: Why not use `${distDir}/data` directly? It works the same for relative paths, but makes it easier to use absoulute paths.

Gruntfile.js
```diff
@@ -522,9 +570,27 @@ module.exports = (grunt) => {
     grunt.task.run(`_pin:${version}`)
   })
 
+  grunt.registerTask('deploy', function () {
+    if (!production) {
+      console.log(chalk.yellow('The command has some requirements in setting environment variables.\nNODE_ENV=production'))
```
corrideat: Is this necessary? I can think of some situations where running the debug build would be useful (it's fine to give a warning IMO, just not the `process.exit(1)` that follows)

Gruntfile.js
```diff
@@ -404,6 +394,64 @@ module.exports = (grunt) => {
 
   let child = null
 
+  grunt.registerTask('copy:frontend', function () {
+    grunt.task.run(['copy:htmlFiles', 'copy:assets', 'copy:strings'])
+  })
+
+  grunt.registerTask('copy:contracts', async function () {
```
taoeffect: Instead of

```js
grunt.task.run(['copy:htmlFiles', 'copy:assets', 'copy:strings'])
```

You can simply do:

```js
grunt.task.run(['copy'])
```

To run them all.

However, because you are overloading `copy` in `registerTask` as well, you cannot do this.

So I recommend the following changes:

1. Instead of calls to `copy:frontend`, just call `copy`, and delete the `copy:frontend` task
2. Rename `copy:contracts` to ~~`copyContracts`~~ (EDIT: see comment below regarding `copyAndMoveContracts`), since it is a special case that you only call sometimes

Gruntfile.js
taoeffect: I would also shorten the message to just:

```js
console.warn(chalk.yellow('Please run with NODE_ENV=production'))
```

Same for `serve` below

Gruntfile.js
```diff
@@ -522,9 +570,27 @@ module.exports = (grunt) => {
...
+      'exec:gitconfig', 'checkDependencies', 'exec:eslint', 'exec:flow',
+      'exec:puglint', 'exec:stylelint', 'clean:dist',
+      'copy:frontend', 'esbuild:deploy', 'copy:contracts'
+    ])
+  })
+  grunt.registerTask('serve', function () {
+    if (!production) {
+      console.log(chalk.yellow('The command has some requirements in setting environment variables.\nNODE_ENV=production'))
+      process.exit(1)
+    }
```
taoeffect: See above comment about shortened message, and also I agree with @corrideat that it is fine to print a warning and proceed instead of existing.

Gruntfile.js
```diff
@@ -522,9 +570,27 @@ module.exports = (grunt) => {
...
+  grunt.registerTask('deploy', function () {
+    if (!production) {
+      console.log(chalk.yellow('The command has some requirements in setting environment variables.\nNODE_ENV=production'))
+      process.exit(1)
+    }
+    grunt.task.run([
+      'exec:gitconfig', 'checkDependencies', 'exec:eslint', 'exec:flow',
+      'exec:puglint', 'exec:stylelint', 'clean:dist',
+      'copy:frontend', 'esbuild:deploy', 'copy:contracts'
+    ])
```
taoeffect: 1. We don't need `exec:gitconfig` here (it's sufficient to have it in `dev` alone)
2. DRY: instead of repeating most of these steps, call `build:deploy` and inside of the existing `build` task handle the `deploy` flag appropriately

taoeffect (changes requested):
Preliminary review

Gruntfile.js
```diff
@@ -404,6 +394,64 @@ module.exports = (grunt) => {
...
+    for (const version of versions) {
+      await cp(`contracts/${version}`, `${distContracts}/${version}`, { recursive: true })
+    }
+
+    done()
+  })
+
+  grunt.registerTask('chelDeploy', function () {
+    const tasks = !production ? ['exec:chelDevDeploy'] : ['exec:chelProdDeploy']
+    grunt.task.run(tasks)
```
taoeffect: We can reduce the number of characters here a bit:

```js
grunt.task.run([production ? 'exec:chelProdDeploy' : 'exec:chelDevDeploy'])
```

Gruntfile.js
taoeffect: Actually, I see that this is a destructive operation anyway, so it shouldn't be called "copy". On line 411 there's a call to delete the contracts being copied, so perhaps a better name for this task would be `copyAndMoveContracts`, since some of the contracts get deleted.

Gruntfile.js
```diff
@@ -404,6 +394,64 @@ module.exports = (grunt) => {
...
+    })
+    child.on('exit', (c) => {
+      if (c !== 0) {
+        grunt.log.error(`child exited with error code: ${c}`.bold)
+        // ^C can cause c to be null, which is an OK error.
+        process.exit(c || 0)
+      }
+    })
+    done()
+  })
```
taoeffect: The `'backend:launch'` task is more lightweight, as it avoids spinning off a new nodejs process...

Is there a reason this `run` task was created instead of reusing the existing `'backend:launch'` task?

Gruntfile.js
```diff
@@ -455,7 +503,7 @@ module.exports = (grunt) => {
     const esbuild = this.flags.watch ? 'esbuild:watch' : 'esbuild'
 
     if (!grunt.option('skipbuild')) {
-      grunt.task.run(['exec:eslint', 'exec:flow', 'exec:puglint', 'exec:stylelint', 'clean', 'copy', esbuild])
+      grunt.task.run(['exec:eslint', 'exec:flow', 'exec:puglint', 'exec:stylelint', 'clean:dist', 'copy:frontend', esbuild])
```
taoeffect: As mentioned above, I think these changes can be reverted for both `'clean'` and `'copy'`

frontend/main.js
taoeffect: Agreed, I have the same question

Silver-IT: I am not sure why we need `GI_VERSION`, but we are currently using it for version management. As we can see the definition of `GI_VERSION` below (it's in Gruntfile.js), it uses the current date.
```
const GI_VERSION = packageJSON.version + (NODE_ENV === 'production' ? `@${new Date().toISOString()}` : '')
```
So the `GI_VERSION` would be different each time of calling `grunt deploy`, `grunt serve`. Because of the difference in two `GI_VERSION`s, the `NOTIFICATION_TYPE.VERSION_INFO` event is called, and the page is refreshing infinitely.
```
sbp('okTurtles.events/on', NOTIFICATION_TYPE.VERSION_INFO, (versionInfo) => {
  console.info('New Group Income version available:', versionInfo)
  // Prevent the client from trying to reconnect when the page starts unloading.
  sbp('okTurtles.data/get', PUBSUB_INSTANCE).destroy()
  // TODO: allow the user to manually reload the page later.
  window.location.reload()
})
```

Gruntfile.js
Silver-IT: Understand.

backend/database.js
Silver-IT: Understood. It's a nice solution.

Gruntfile.js
Silver-IT: Actually, I tried to use `backend:launch`, and got an error saying `can not find ./backend/index.js`. Now, I've fixed that error and I am going to use `backend:launch`.

Gruntfile.js
Silver-IT: Nice catch!

Gruntfile.js
```diff
@@ -452,7 +477,7 @@ module.exports = (grunt) => {
   })
 
   grunt.registerTask('build', function () {
-    const esbuild = this.flags.watch ? 'esbuild:watch' : 'esbuild'
+    const esbuild = this.flags.deploy ? 'esbuild:deploy' : (this.flags.watch ? 'esbuild:watch' : 'esbuild')
```
taoeffect: It appears this `esbuild:deploy` flag is not used anywhere, so it seems like both `esbuild:deploy` and `build:deploy` flags can be removed?

frontend/main.js
taoeffect: I see. Could you please then check, instead of `'production'` for `'development'` (e.g. `process.env.NODE_ENV === 'development'`)?

And then update the `if` condition plus add an explanatory comment:

```js
// We only compare GI_VERSION in development mode so that the page auto-refreshes if `grunt dev` is re-run
// This check cannot be done in production mode as it would lead to an infinite page refresh bug
// when using `grunt deploy` with `grunt serve`
if (isContractVersionDiff || (isDevelopment && isGIVersionDiff)) {
```

taoeffect (changes requested):
Looking really solid @Silver-IT! Just two change requests left, and then once these are applied I will go ahead and test the PR and see how everything works.

frontend/main.js
Silver-IT: Sure.

Gruntfile.js
Silver-IT: Nice catch!

taoeffect (approved):
Approved! Great work @Silver-IT !! 👏 👏

PR #1955, opened 2024-04-19
Reference counting

### LINKED ISSUE #828: Removing a member from a group - detect what identity contracts should be deleted
### Problem

When removing a member from the group (#560), their related identity contracts should also be removed from the state, based on this condition:

- If we are the member being removed, we need to remove the contracts of all the other members of the group.
- If we are another member of the group, we need to remove the contract of the member being removed. 

**However,** we can only remove an identity contract if we don't have any other type of connections with this member (ex: another group in common or a chat message).

> Imagine if a friend of ours gets kicked out of a group, but we have an open DM chat with them. We need to be subscribed to their identity contract still if the chat is outside of the group. Otherwise, we won't be able to see their profile picture or their displayName. Basically, if we're interacting with someone at all in any way, we need to keep their identity contract in the state.

### Solution

Implement an algorithm that detects this relationship between identity contracts and use it to conditionally remove an identity contract after a member is removed from a group.

### LINKED ISSUE #1249: Contract system: add retain/release concept (reference counting)

corrideat:
Fixes #1249
Fixes #828

### DIFF
--- a/frontend/controller/actions/chatroom.js
+++ b/frontend/controller/actions/chatroom.js
@@ -147,18 +147,7 @@ export default (sbp('sbp/selectors/register', {
               }
             },
             data: SAKp
-          },
-          // TODO: Find a way to have this wrapping be done by Chelonia directly
-          encryptedOutgoingDataWithRawKey(CEK, {
-            foreignKey: `sp:${encodeURIComponent(userID)}?keyName=${encodeURIComponent('csk')}`,
-            id: userCSKid,
-            data: rootState[userID]._vm.authorizedKeys[userCSKid].data,
-            permissions: [GIMessage.OP_ACTION_ENCRYPTED + '#inner'],
-            allowedActions: '*',
-            purpose: ['sig'],
-            ringLevel: Number.MAX_SAFE_INTEGER,
-            name: `${userID}/${userCSKid}`
-          })
+          }
         ],
         data: {
           ...params.data,
@@ -230,18 +219,7 @@ export default (sbp('sbp/selectors/register', {
       throw new Error(`Unable to send gi.actions/chatroom/join on ${params.contractID} because user ID contract ${userID} is missing`)
     }
 
-    const isCurrentUserJoining = rootState.loggedIn.identityContractID === userID
-
-    if (isCurrentUserJoining) {
-      // Cancel remove when sending this (join) action. This is because if we're
-      // trying to join a chatroom that we've previously left, it'll be removed
-      // by its side-effects. Calling 'chelonia/contract/cancelRemove' clears
-      // the pendingRemove flag in the contract, preventing it from being
-      // removed (which is the intent here, as we're re-joining)
-      sbp('chelonia/contract/cancelRemove', params.contractID)
-    }
-
-    const CEKid = sbp('chelonia/contract/currentKeyIdByName', params.contractID, 'cek')
+    const CEKid = params.encryptionKeyId || sbp('chelonia/contract/currentKeyIdByName', params.contractID, 'cek')
 
     const userCSKid = sbp('chelonia/contract/currentKeyIdByName', userID, 'csk')
     return await sbp('chelonia/out/atomic', {
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -280,13 +280,10 @@ export default (sbp('sbp/selectors/register', {
   // action if we haven't done so yet (because we were previously waiting for
   // the keys), or (c) already a member and ready to interact with the group.
   'gi.actions/group/join': async function (params: $Exact<ChelKeyRequestParams>) {
-    // We want to process any current events first, before setting
-    // JOINING_GROUP, so that we process leave actions and don't interfere
-    // with the leaving process (otherwise, the side-effects will prevent
-    // us from fully leaving).
+    // We want to process any current events first, so that we process leave
+    // actions and don't interfere with the leaving process (otherwise, the
+    // side-effects could prevent us from fully leaving).
     await sbp('chelonia/contract/wait', [params.originatingContractID, params.contractID])
-    // TODO: See if we can remove the need for okTurtles.data/set
-    sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, true)
     try {
       const { loggedIn } = sbp('state/vuex/state')
       if (!loggedIn) throw new Error('[gi.actions/group/join] Not logged in')
@@ -295,16 +292,9 @@ export default (sbp('sbp/selectors/register', {
 
       // When syncing the group contract, the contract might call /remove on
       // itself if we had previously joined and left the group. By using
-      // deferredRemove we ensure that it's not deleted until we've finished
+      // ephemeral we ensure that it's not deleted until we've finished
       // trying to join.
-      // When we are re-joinining, this function will call /cancelRemove to
-      // ensure that the contract isn't removed, and if we're not re-joining,
-      // the call at the end to /remove with removeIfPending will remove the
-      // group contract.
-      // Part of this same functionality is what `sbp('okTurtles.data/set', 'JOINING_GROUP-'`
-      // does. In a later improvement, we could remove that and handle re-joinining
-      // within the contract state.
-      await sbp('chelonia/contract/sync', params.contractID, { deferredRemove: true })
+      await sbp('chelonia/contract/retain', params.contractID, { ephemeral: true })
       const rootState = sbp('state/vuex/state')
       if (!rootState.contracts[params.contractID]) {
         console.warn('[gi.actions/group/join] The group contract was removed after sync. If this happened during logging in, this likely means that we left the group on a different session.', { contractID: params.contractID })
@@ -408,9 +398,6 @@ export default (sbp('sbp/selectors/register', {
           console.error(`[gi.actions/group/join] Error while sending key request for ${params.contractID}:`, e?.message || e, e)
           throw e
         })
-        // While we're waiting for keys, we should not remove the group contract
-        // (if had previously left) because we're re-joining
-        sbp('chelonia/contract/cancelRemove', params.contractID)
 
         // Nothing left to do until the keys are received
 
@@ -476,7 +463,6 @@ export default (sbp('sbp/selectors/register', {
           }
         }
 
-        sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
         sbp('okTurtles.events/emit', JOINED_GROUP, { contractID: params.contractID })
       // We don't have the secret keys and we're not waiting for OP_KEY_SHARE
       // This means that we've been removed from the group
@@ -485,27 +471,20 @@ export default (sbp('sbp/selectors/register', {
         // do much at this point, so we do nothing.
         // This could happen, for example, after logging in if we still haven't
         // received a response to the key request.
-        sbp('okTurtles.data/set', 'JOINING_GROUP-' + params.contractID, false)
       } else if (pendingKeyShares) {
         console.info('Requested to join group but already waiting for OP_KEY_SHARE. contractID=' + params.contractID)
-        // If we're waiting on keys to be shared with us, we need to cancel any
-        // pending removal while we're waiting
-        sbp('chelonia/contract/cancelRemove', params.contractID)
       } else {
         console.error('Requested to join group but the state appears invalid. This should be unreachable. contractID=' + params.contractID, { sendKeyRequest, hasSecretKeys, pendingKeyShares })
       }
     } catch (e) {
       console.error('gi.actions/group/join failed!', e)
-      // If an error occurred, the join process is incomplete and we should not
-      // remove the group contract.
-      sbp('chelonia/contract/cancelRemove', params.contractID)
       throw new GIErrorUIRuntimeError(L('Failed to join the group: {codeError}', { codeError: e.message }))
     } finally {
       // If we called join but it didn't result in any actions being sent, we
       // may have left the group. In this case, we execute any pending /remove
       // actions on the contract. This will have no side-effects if /remove on
       // the group contract hasn't been called.
-      await sbp('chelonia/contract/remove', params.contractID, { removeIfPending: true })
+      await sbp('chelonia/contract/release', params.contractID, { ephemeral: true })
     }
   },
   'gi.actions/group/joinAndSwitch': async function (params: $Exact<ChelKeyRequestParams>) {
@@ -527,16 +506,29 @@ export default (sbp('sbp/selectors/register', {
     // We can avoid this by waiting on both contracts, especially the group
     // contract.
     await sbp('chelonia/contract/wait', [groupId, identityContractID])
-    await sbp('gi.actions/identity/joinGroup', {
-      contractID: identityContractID,
-      contractName: 'gi.contracts/identity',
-      data: {
-        groupContractID: groupId,
-        inviteSecret: secret
-      }
-    }).then(() => {
-      return sbp('gi.actions/group/switch', groupId)
-    })
+    try {
+      // Similarly, because not all events may have been processed, including
+      // side-effects, the group contract may be released too early (when
+      // re-joining a group, there could be an action for leaving the
+      // group that is pending processing). The following pattern (ephemeral
+      // retain + ephemeral release) ensures that we won't unsubscribe to it
+      // until we know what the next step is.
+      // Because the retain is paired with a release, at worst this will have
+      // no effect on the lifetime of the group contract.
+      await sbp('chelonia/contract/retain', groupId, { ephemeral: true })
+      await sbp('gi.actions/identity/joinGroup', {
+        contractID: identityContractID,
+        contractName: 'gi.contracts/identity',
+        data: {
+          groupContractID: groupId,
+          inviteSecret: secret
+        }
+      }).then(() => {
+        return sbp('gi.actions/group/switch', groupId)
+      })
+    } finally {
+      await sbp('chelonia/contract/release', groupId, { ephemeral: true })
+    }
   },
   'gi.actions/group/switch': function (groupId) {
     sbp('state/vuex/commit', 'setCurrentGroupId', groupId)
@@ -928,69 +920,6 @@ export default (sbp('sbp/selectors/register', {
     } else {
       sbp('okTurtles.events/emit', OPEN_MODAL, 'AddMembers')
     }
-  },
-  'gi.actions/group/removeUselessIdentityContracts': function ({ contractID, possiblyUselessContractIDs }) {
-    // NOTE: should remove the identity contracts which we don't need to sync anymore
-    //       for users who don't have any common groups, and any common DMs
-    const rootState = sbp('state/vuex/state')
-    const rootGetters = sbp('state/vuex/getters')
-    const me = rootGetters.ourIdentityContractId
-
-    const removeIdentityContracts = () => {
-      const chatRoomIDs = Object.keys(rootGetters.ourDirectMessages)
-      const groupIDs = Object.keys(rootState.contracts)
-        .filter(gID => rootState.contracts[gID].type === 'gi.contracts/group' && gID !== contractID)
-
-      if (!chatRoomIDs.length && !groupIDs.length) {
-        sbp('chelonia/contract/remove', possiblyUselessContractIDs).catch(e => {
-          console.error(`[gi.actions/group/removeUselessIdentityContracts]: ${e.name} thrown by /remove useless identity contracts:`, e)
-        })
-        return
-      }
-
-      const waitFor = [...chatRoomIDs, ...groupIDs]
-      sbp('chelonia/contract/wait', waitFor).then(() => {
-        if (rootGetters.ourIdentityContractId && rootGetters.ourIdentityContractId !== me) {
-          return
-        }
-
-        const pending = Object.entries(sbp('okTurtles.eventQueue/queuedInvocations'))
-          .filter(([q]) => waitFor.includes(q))
-          .flatMap(([, list]) => list)
-
-        if (pending.length) {
-          return removeIdentityContracts()
-        }
-
-        const identityContractsMapToKeepSyncing = {}
-        // NOTE: contracts for the members from another groups should not be removed
-        groupIDs.forEach(gID => {
-          for (const [iID] of Object.entries((rootState[gID].profiles || {}))) {
-            identityContractsMapToKeepSyncing[iID] = true
-          }
-        })
-
-        // NOTE: contracts for the members from direct messages should not be removed
-        for (const chatRoomId of chatRoomIDs) {
-          const chatRoomState = rootState[chatRoomId]
-
-          if (!chatRoomState || !chatRoomState.members?.[me]) {
-            continue
-          }
-
-          Object.keys(chatRoomState.members).forEach(memberID => {
-            identityContractsMapToKeepSyncing[memberID] = true
-          })
-        }
-
-        const identityContractsToRemove = possiblyUselessContractIDs.filter(cID => !identityContractsMapToKeepSyncing[cID])
-        sbp('chelonia/contract/remove', identityContractsToRemove).catch(e => {
-          console.error(`[gi.actions/group/removeUselessIdentityContracts]: ${e.name} thrown by /remove useless identity contracts:`, e)
-        })
-      })
-    }
-
-    removeIdentityContracts()
   },
   ...encryptedAction('gi.actions/group/leaveChatRoom', L('Failed to leave chat channel.')),
   ...encryptedAction('gi.actions/group/deleteChatRoom', L('Failed to delete chat channel.')),
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -332,8 +332,6 @@ export default (sbp('sbp/selectors/register', {
       const loginAttributes = { identityContractID, encryptionParams, username }
 
       // If username was not provided, retrieve it from the state
-      // TODO: This is temporary until username is no longer needed internally
-      // in contracts
       if (!loginAttributes.username) {
         loginAttributes.username = Object.entries(state.namespaceLookups)
           .find(([k, v]) => v === identityContractID)
@@ -360,9 +358,19 @@ export default (sbp('sbp/selectors/register', {
       }
 
       //            loading the website instead of stalling out.
-      // See the TODO note in startApp (main.js) for why this is not awaited
       try {
-        await sbp('chelonia/contract/sync', identityContractID, { force: true })
+        if (!state) {
+          // Make sure we don't unsubscribe from our own identity contract
+          // Note that this should be done _after_ calling
+          // `chelonia/storeSecretKeys`: If the following line results in
+          // syncing the identity contract and fetching events, the secret keys
+          // for processing them will not be available otherwise.
+          await sbp('chelonia/contract/retain', identityContractID)
+        } else {
+          // If there is a state, we've already retained the identity contract
+          // but might need to fetch the latest events
+          await sbp('chelonia/contract/sync', identityContractID, { force: true })
+        }
       } catch (err) {
         sbp('okTurtles.events/emit', LOGIN_ERROR, { username, identityContractID, error: err })
         const errMessage = err?.message || String(err)
--- a/frontend/model/contracts/chatroom.js
+++ b/frontend/model/contracts/chatroom.js
@@ -5,6 +5,7 @@
 import { L, Vue } from '@common/common.js'
 import sbp from '@sbp/sbp'
 import { objectOf, optional, string, arrayOf, actionRequireInnerSignature } from '~/frontend/model/contracts/misc/flowTyper.js'
+import { ChelErrorGenerator } from '~/shared/domains/chelonia/errors.js'
 import { findForeignKeysByContractID, findKeyIdByName } from '~/shared/domains/chelonia/utils.js'
 import {
   CHATROOM_ACTIONS_PER_PAGE,
@@ -28,6 +29,9 @@ import {
 import { cloneDeep, merge } from './shared/giLodash.js'
 import { makeNotification } from './shared/nativeNotification.js'
 import { chatRoomAttributesType, messageType } from './shared/types.js'
+
+export const GIChatroomAlreadyMemberError: typeof Error = ChelErrorGenerator('GIChatroomAlreadyMemberError')
+export const GIChatroomNotMemberError: typeof Error = ChelErrorGenerator('GIChatroomNotMemberError')
 
 function createNotificationData (
   notificationType: string,
@@ -176,7 +180,12 @@ sbp('chelonia/defineContract', {
         }
       },
       sideEffect ({ contractID }) {
-        Vue.set(sbp('state/vuex/state')?.chatroom?.chatRoomUnread, contractID, {
+        const chatroomUnread = sbp('state/vuex/state').chatroom?.chatRoomUnread
+        if (!chatroomUnread) {
+          console.warn('[gi.contracts/chatroom] rootState.chatroom?.chatRoomUnread is not an object')
+          return
+        }
+        Vue.set(chatroomUnread, contractID, {
           readUntil: undefined,
           messages: []
         })
@@ -197,7 +206,7 @@ sbp('chelonia/defineContract', {
           if (!state.members) {
             Vue.set(state, 'members', {})
           } else if (state.members[memberID]) {
-            throw new Error(`Can not join the chatroom which ${memberID} is already part of`)
+            throw new GIChatroomAlreadyMemberError(`Can not join the chatroom which ${memberID} is already part of`)
           }
 
           Vue.set(state.members, memberID, { joinedDate: meta.createdDate })
@@ -233,13 +242,6 @@ sbp('chelonia/defineContract', {
           setReadUntilWhileJoining({ contractID, hash, createdDate: meta.createdDate })
 
           if (memberID === loggedIn.identityContractID) {
-            // The join process is now complete, so we can remove this key if
-            // it was set. This key is used to prevent us from calling `/remove`
-            // when the join process is incomplete. See the comment on group.js
-            // (joinChatRoom sideEffect) for a more detailed explanation of
-            // what this does.
-            sbp('okTurtles.data/delete', `JOINING_CHATROOM-${contractID}-${memberID}`)
-
             if (state.attributes.type === CHATROOM_TYPES.DIRECT_MESSAGE) {
             // NOTE: To ignore scroll to the message of this hash
             //       since we don't create notification when join the direct message
@@ -327,7 +329,7 @@ sbp('chelonia/defineContract', {
             throw new Error('Missing members state')
           }
           if (!state.members[memberID]) {
-            throw new Error(`Can not leave the chatroom ${contractID} which ${memberID} is not part of`)
+            throw new GIChatroomNotMemberError(`Can not leave the chatroom ${contractID} which ${memberID} is not part of`)
           }
 
           Vue.delete(state.members, memberID)
@@ -365,18 +367,9 @@ sbp('chelonia/defineContract', {
           }
 
           if (memberID === rootState.loggedIn.identityContractID) {
-            // If we're in the process of joining this chatroom, don't call
-            // /remove, as we're still waiting to be added to the chatroom.
-            // See group.js (joinChatRoom sideEffect) for a more detailed
-            // explanation of why we need this
-            if (!sbp('okTurtles.data/get', `JOINING_CHATROOM-${contractID}-${memberID}`)) {
-            // NOTE: make sure *not* to await on this, since that can cause
-            //       a potential deadlock. See same warning in sideEffect for
-            //       'gi.contracts/group/removeMember'
-              sbp('chelonia/contract/remove', contractID).catch(e => {
-                console.error(`[gi.contracts/chatroom/leave/sideEffect] (${contractID}): remove threw ${e.name}:`, e)
-              })
-            }
+            leaveChatRoom({ contractID }).catch((e) => {
+              console.error(`[gi.contracts/chatroom/leave/sideEffect] Error for ${contractID}`, e)
+            })
           } else {
             setReadUntilWhileJoining({ contractID, hash, createdDate: meta.createdDate })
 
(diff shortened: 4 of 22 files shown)


frontend/controller/actions/chatroom.js
```diff
@@ -147,18 +147,7 @@ export default (sbp('sbp/selectors/register', {
...
-          encryptedOutgoingDataWithRawKey(CEK, {
-            foreignKey: `sp:${encodeURIComponent(userID)}?keyName=${encodeURIComponent('csk')}`,
-            id: userCSKid,
-            data: rootState[userID]._vm.authorizedKeys[userCSKid].data,
-            permissions: [GIMessage.OP_ACTION_ENCRYPTED + '#inner'],
-            allowedActions: '*',
-            purpose: ['sig'],
-            ringLevel: Number.MAX_SAFE_INTEGER,
-            name: `${userID}/${userCSKid}`
-          })
```
taoeffect, line 161: Just curious... could you comment here (on GH) why this was removed?

frontend/controller/actions/group.js
```diff
@@ -295,16 +292,9 @@ export default (sbp('sbp/selectors/register', {
...
       // trying to join.
-      // When we are re-joinining, this function will call /cancelRemove to
-      // ensure that the contract isn't removed, and if we're not re-joining,
-      // the call at the end to /remove with removeIfPending will remove the
-      // group contract.
-      // Part of this same functionality is what `sbp('okTurtles.data/set', 'JOINING_GROUP-'`
-      // does. In a later improvement, we could remove that and handle re-joinining
-      // within the contract state.
-      await sbp('chelonia/contract/sync', params.contractID, { deferredRemove: true })
+      await sbp('chelonia/contract/retain', params.contractID, { ephemeral: true })
```
taoeffect, line 297: OK... but where do we call `retain` in a persistent way for joining the group?

Don't we need to do that somewhere? And if not, why not?

(Also a request for an explanation regarding this in the code itself via comments)

frontend/controller/actions/group.js
```diff
@@ -929,69 +913,6 @@ export default (sbp('sbp/selectors/register', {
...
-
-        const identityContractsToRemove = possiblyUselessContractIDs.filter(cID => !identityContractsMapToKeepSyncing[cID])
-        sbp('chelonia/contract/remove', identityContractsToRemove).catch(e => {
-          console.error(`[gi.actions/group/removeUselessIdentityContracts]: ${e.name} thrown by /remove useless identity contracts:`, e)
-        })
-      })
-    }
-
-    removeIdentityContracts()
-  },
```
taoeffect, line 994: 😍

frontend/controller/actions/identity.js
```diff
@@ -343,6 +343,11 @@ export default (sbp('sbp/selectors/register', {
       sbp('state/vuex/commit', 'login', loginAttributes)
       await sbp('chelonia/storeSecretKeys', () => transientSecretKeys)
 
+      if (!state) {
+        // Make sure we don't unsubscribe from our own identity contract
+        await sbp('chelonia/contract/retain', identityContractID)
+      }
```
taoeffect: A few questions here:

1. Could this be moved to an `else` branch of the same `if (state)` check above? (e.g. on line 329?
2. On line 368 it says someting about the sync not being awaited... but it is awaited. Is this comment something we forgot to clean up related to #1765 ? (Can it be boyscoutted now?)

frontend/controller/actions/chatroom.js
corrideat: This is already done when joining a chatroom (using `OP_KEY_ADD` inside `OP_ATOMIC`). Having this here too makes it harder to maintain this code, and in this particular case the encryption key used (the `cek` is the wrong key to use because it causes signature errors). I think it's easier to remove this than to maintain two versions.

frontend/controller/actions/group.js
corrideat: The persistent retain is called by the corresponding action in the identity contract. That is, the source of truth for group membership is the identity contract.

> Don't we need to do that somewhere?

Correct. As mentioned, it's the identity contract doing it.

> And if not, why not?

n/a

frontend/controller/actions/identity.js
corrideat: > 1. Could this be moved to an else branch of the same if (state) check above? (e.g. on line 329?

It needs to happen _after_ `await sbp('chelonia/storeSecretKeys', () => transientSecretKeys)`. If some secret keys are missing, since this is possibly calling `sync`, it could result in an invalid state.

> On line 368 it says someting about the sync not being awaited... but it is awaited.

Right.

> Is this comment something we forgot to clean up related to #1765 ? (Can it be boyscoutted now?)

I think you're right. However, IIRC, the no-await comment there referred to the group contracts, not the identity contract. In either case, I'll go ahead and remove that comment.

taoeffect: > It needs to happen after await sbp('chelonia/storeSecretKeys', () => transientSecretKeys). If some secret keys are missing, since this is possibly calling sync, it could result in an invalid state.

I'd add that as a comment just so that nobody else accidentally moves it there without realizing

frontend/model/contracts/chatroom.js
```diff
@@ -176,7 +176,9 @@ sbp('chelonia/defineContract', {
         }
       },
       sideEffect ({ contractID }) {
-        Vue.set(sbp('state/vuex/state')?.chatroom?.chatRoomUnread, contractID, {
+        const chatroomUnread = sbp('state/vuex/state').chatroom?.chatRoomUnread
+        if (!chatroomUnread) return
```
taoeffect: I'm rather confused by this...

This `chatroomModule` is supposed to be set/defined. It's part of the initial state for Vuex chatroom module.

And if, for some extraordinary reason (that should definitely generate a `console.warn` at the very least), it is not set, then shouldn't we just proceed with setting it instead of returning immediately?

corrideat: This was causing some issues during testing, although note that the previous version of it was expecting it not to be set sometimes too:

```
sbp('state/vuex/state')?.chatroom?
```

Using `Vue.set(undefined, ...)` is an error, hence the changes.

As for the second part:

> And if, for some extraordinary reason (that should definitely generate a console.warn at the very least),

IMO, this is the wrong place to do this, as it's the application's responsibility to set up `chatroom`. Again IMO, we shouldn't set it here and just treat it as a non-critical error as it not being set doesn't break the contract state.

frontend/model/contracts/chatroom.js
```diff
@@ -410,6 +402,9 @@ sbp('chelonia/defineContract', {
         // NOTE: make sure *not* to await on this, since that can cause
         //       a potential deadlock. See same warning in sideEffect for
         //       'gi.contracts/group/removeMember'
+        leaveChatRoom({ contractID }).catch((e) => {
+          console.error(`[gi.contracts/chatroom/delete/sideEffect] Error for ${contractID}`, e)
+        })
```
taoeffect, line 405: I'm a little confused here as to why this sidEffect is written in a different way from the sideEffect of `chatroom/leave`.

In `chatroom/leave` sideEffect we push something to the queue to ensure the sideEffect runs after the chatroom has synced, yet here instead we do something completely different. Instead of doing something after the sync is done, we check if we're currently syncing and abort the sideEffect if so.

Doesn't this open up the possibility that the chatroom isn't left? For example, if person A deletes the chatroom while person B is logged out, and then person B logs in, won't person B still be subscribed to the chatroom?

frontend/model/contracts/chatroom.js
```diff
@@ -365,18 +366,9 @@ sbp('chelonia/defineContract', {
...
-            // NOTE: make sure *not* to await on this, since that can cause
-            //       a potential deadlock. See same warning in sideEffect for
-            //       'gi.contracts/group/removeMember'
-              sbp('chelonia/contract/remove', contractID).catch(e => {
-                console.error(`[gi.contracts/chatroom/leave/sideEffect] (${contractID}): remove threw ${e.name}:`, e)
-              })
-            }
+            leaveChatRoom({ contractID }).catch((e) => {
+              console.error(`[gi.contracts/chatroom/leave/sideEffect] Error for ${contractID}`, e)
+            })
```
taoeffect, line 372: Before, we would call `/remove` here. No instead we call `leaveChatRoom` which doesn't call `/remove` or `/release`.

Is this right? Where is `/release` called for leaving the chatroom? (Can an explanation be added as a comment here?)

frontend/model/contracts/chatroom.js
corrideat: > In chatroom/leave sideEffect we push something to the queue to ensure the sideEffect runs after the chatroom has synced, yet here instead we do something completely different. Instead of doing something after the sync is done, we check if we're currently syncing and abort the sideEffect if so.

The reason to wait in `chatroom/leave` is that someone could leave and then rejoin. In that case, the queue is used to ensure that leaving doesn't happen, as leave + join means that we're a member.

OTOH, the `delete` action is meant to be final.

> For example, if person A deletes the chatroom while person B is logged out, and then person B logs in, won't person B still be subscribed to the chatroom?

If that happens, person B presumably won't have the latest state, which includes deleting the chatroom. When person B comes back online, the side-effect will be executed (because they're subscribed to the contract) and they'll then leave.

frontend/model/contracts/chatroom.js
corrideat: `release` is called by the contract that manages chatrooms (i.e., the group or the identity contract). This is the same contract that calls `retain`, which helps keep the reference count.

In general, `remove` should never be called directly outside of chelonia.

frontend/model/contracts/group.js
```diff
@@ -1409,7 +1418,7 @@ sbp('chelonia/defineContract', {
       sideEffect ({ meta, data, contractID, innerSigningContractID }, { state }) {
         const rootState = sbp('state/vuex/state')
         const memberID = data.memberID || innerSigningContractID
-        if (innerSigningContractID === rootState.loggedIn.identityContractID && !sbp('okTurtles.data/get', 'JOINING_GROUP-' + contractID)) {
```
taoeffect, line 1412: Can you explain here on GH (and maybe in a comment) why this `'JOINING_GROUP-' + contractID` check is no longer necessary?

frontend/model/contracts/group.js
```diff
@@ -1587,59 +1576,6 @@ sbp('chelonia/defineContract', {
...
-          ).sort(cID =>
-          // prefer successfully joined groups
-            rootState[cID]?.profiles?.[identityContractID] ? -1 : 1
-          )[0] || null
-        sbp('state/vuex/commit', 'setCurrentChatRoomId', {})
-        sbp('state/vuex/commit', 'setCurrentGroupId', groupIdToSwitch)
-      }
-
-      // Remove last logged in information
-      Vue.delete(rootState.lastLoggedIn, contractID)
```
taoeffect, line 1611: Is this all now done somewhere else? Or is it no longer necessary because...?

frontend/model/contracts/group.js
corrideat: The short answer is that it's because that key no longer exists, so any such checks are dead code (always true or always false, depending on whether it's for presence or absence).

The longer answer is that this was here because leaving a group / chatroom would call `/remove`, which needed to be prevented. Now that reference counting is implemented, if the `release` / `retain` calls are correctly paired, leaving a group that is in the process of being re-joined shouldn't be an issue (as this can be handled with ephemeral references --while joining--, and persistent references --after joining--).

frontend/model/contracts/identity.js
```diff
@@ -300,7 +309,43 @@ sbp('chelonia/defineContract', {
             })
           }
 
-          // TODO disconnect, key rotations (PEK), etc.
+          sbp('chelonia/contract/release', data.groupContractID).catch((e) => {
```
corrideat, line 312: https://github.com/okTurtles/group-income/pull/1955/files/ccdfb7599dc4240066204d943cb5ff044aebd3d0#diff-eaaa3e4a455bba1ad6131a29aec3a915aa1f656241bc956c556e341812bbd42eR312-R327

frontend/model/contracts/group.js
corrideat: https://github.com/okTurtles/group-income/pull/1955/files/ccdfb7599dc4240066204d943cb5ff044aebd3d0#diff-eaaa3e4a455bba1ad6131a29aec3a915aa1f656241bc956c556e341812bbd42eR312-R327

corrideat: This has been moved to the identity contract, which keeps track of the groups one is member of.

frontend/model/contracts/group.js
```diff
@@ -1810,16 +1746,26 @@ sbp('chelonia/defineContract', {
...
+              postpublish: () => {
+                sbp('chelonia/contract/retain', chatRoomId)
+              },
               onprocessed: () => {
                 sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupId: contractID, chatRoomId })
               }
             }
           }
         }).catch(e => {
+          if (e.name === 'GIErrorUIRuntimeError' && e.cause?.message?.endsWith('is already part of')) {
```
taoeffect: Generally speaking, `GIErrorUIRuntimeError`s are intended to be displayed to the user, i.e. their cause is a user-facing string.

At the moment, this string isn't being translated (wrapped in an `L` function). But it probably should be. And if/when that happens, you won't be able to do this comparison, because you'll be checking against a translation.

So it's probably better/safer to check some sort of other parameter that is set on the error, like a `.type` or something, that is set to some internal non-userfacing value.

corrideat: > At the moment, this string isn't being translated (wrapped in an L function). But it probably should be. And if/when that happens, you won't be able to do this comparison, because you'll be checking against a translation.

I'm not checking the `message` of the `GIErrorUIRuntimeError`, but rather the `message` of the `cause` of the `GIErrorUIRuntimeError`, which isn't wrapped in an `L` function. The string that is user facing is made by `LError`, that right now uses the original error string verbatim.

I could change this to use, e.g., the `name` attribute, but I'm not so sure how good of an idea it's to translate internal errors, as that just complicates debugging.

frontend/model/contracts/group.js
```diff
@@ -1889,27 +1825,33 @@ sbp('chelonia/defineContract', {
...
+        }).catch(e => {
+          console.warn(`[gi.contracts/group/leaveGroup] ${e.name} thrown by gi.contracts/identity/leaveGroup ${identityContractID} for ${contractID}:`, e)
         })
       } else {
         const myProfile = getters.groupProfile(identityContractID)
 
+        // Do _not_ release memberID, because doing so will remove their profile
+        // from our view
+        // // sbp('chelonia/contract/release', memberID)
+
```
taoeffect: Maybe a comment can be added here as a `// TODO` (and a new issue) to mention that we could simply save in a local cache their username + profile picture and load that in the UI instead of remaining subscribed to their contract.

shared/domains/chelonia/chelonia.js
```diff
@@ -21,6 +21,10 @@ import './files.js'
 import './time-sync.js'
 import { buildShelterAuthorizationHeader, eventsAfter, findForeignKeysByContractID, findKeyIdByName, findRevokedKeyIdsByName, findSuitableSecretKeyId, getContractIDfromKeyId } from './utils.js'
 
+const clearObject = (o: Object) => {
+  Object.keys(o).forEach((k) => delete o[k])
```
taoeffect: Might want to use `reactiveDel` here to be safe

shared/domains/chelonia/chelonia.js
```diff
@@ -792,48 +793,106 @@ export default (sbp('sbp/selectors/register', {
...
+      return (
+        // Check persistent references
+        (!has(rootState.contracts, id) || !has(rootState.contracts[id], 'references')) &&
+        // Check ephemeral references
+        !has(this.ephemeralReferenceCount, id)) &&
+        // Check foreign keys (i.e., that no keys are being watched)
+        (!has(rootState, id) || !has(rootState[id], '_volatile') || !has(rootState[id]._volatile, 'watch') || rootState[id]._volatile.watch.length === 0 || rootState[id]._volatile.watch.filter(([, cID]) => this.subscriptionSet.has(cID)).length === 0)
+    })
+    if (idsToRemove.length === 0) return Promise.resolve()
+    return sbp('chelonia/contract/remove', idsToRemove)
```
taoeffect: Could probably be rewritten without two `return`s and a `Promise.resolve()` like so:

```js
if (idsToRemove.length > 0) {
  return sbp('chelonia/contract/remove', idsToRemove)
}
```

Or:

```js
return idsToRemove.length > 0 && sbp('chelonia/contract/remove', idsToRemove)
```

shared/domains/chelonia/internals.js
```diff
@@ -408,9 +408,26 @@ export default (sbp('sbp/selectors/register', {
...
+      Object.keys(state[contractID])
+        .filter((k) => k !== '_volatile')
+        .forEach((k) => this.config.reactiveDel(state[contractID], k))
+      if (state[contractID]._volatile) {
+        Object.keys(state[contractID]._volatile)
+          .filter((k) => k !== 'watch')
+          .forEach((k) => this.config.reactiveDel(state[contractID]._volatile, k))
+      }
+    } else {
+      this.config.reactiveDel(this.ephemeralReferenceCount, contractID)
```
taoeffect: This I think you said can be done with `delete` directly?

shared/domains/chelonia/chelonia.js
corrideat: It should return a `Promise`, which your suggestions don't do, unless the function is marked as `async`.

shared/domains/chelonia/internals.js
corrideat: That's right, I'll fix this.

shared/domains/chelonia/chelonia.js
```diff
@@ -792,48 +793,106 @@ export default (sbp('sbp/selectors/register', {
...
+        if (!has(this.ephemeralReferenceCount, id)) {
+          this.ephemeralReferenceCount[id] = 1
+        } else {
+          this.ephemeralReferenceCount[id] = this.ephemeralReferenceCount[id] + 1
+        }
+      })
+    }
+    return sbp('chelonia/contract/sync', listOfIds)
+  },
+  'chelonia/contract/release': function (contractIDs: string | string[], params?: { ephemeral?: boolean, try?: boolean }): Promise<*> {
```
taoeffect, line 853: Could you add a brief comment on what `try` means above the definition of this selector?

frontend/model/contracts/group.js
corrideat: Using the `name` attribute now

shared/domains/chelonia/chelonia.js
taoeffect: Hmm. OK, I think you're right. I'm used to `async` functions allowing this.

Perhaps, then:

```js
return idsToRemove.length > 0 ? sbp('chelonia/contract/remove', idsToRemove) : Promise.resolve()
```

corrideat: Sure, if you think that's easier to read.

shared/domains/chelonia/internals.js
```diff
@@ -958,7 +975,13 @@ export default (sbp('sbp/selectors/register', {
                 const rootState = sbp(config.stateSelector)
                 if (Array.isArray(rootState[foreignContract]?._volatile?.watch)) {
                   // Stop watching events for this key
-                  rootState[foreignContract]._volatile.watch = rootState[foreignContract]._volatile.watch.filter(([name, cID]) => name !== foreignKeyName || cID !== contractID)
+                  const oldWatch = rootState[foreignContract]._volatile.watch
+                  rootState[foreignContract]._volatile.watch = oldWatch.filter(([name, cID]) => name !== foreignKeyName || cID !== contractID)
+                  if (oldWatch.length !== rootState[foreignContract]._volatile.watch.length) {
+                    sbp('chelonia/contract/release', foreignContract, { try: true }).catch(e => {
```
taoeffect, line 984: Maybe a comment above this explaining why `try: true` here?

frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -309,7 +309,7 @@ export default ({
         if (
           this.currentGroupState._vm.invites?.[inviteKeyId]?.status === INVITE_STATUS.VALID &&
           this.currentGroupState._vm.authorizedKeys?.[inviteKeyId] &&
-          this.currentGroupState._vm.authorizedKeys[inviteKeyId]._notAfterHeight != null &&
+          this.currentGroupState._vm.authorizedKeys[inviteKeyId]._notAfterHeight == null &&
```
taoeffect, line 312: Oof, seems like an important bug catch. Surprised tests didn't catch this

corrideat: Indeed. Well, this breaks the tests on this branch, but for some reason not on master. The previous line is wrong anyhow, as `_notAfterHeight != null` means that the invite has since been revoked.

shared/domains/chelonia/chelonia.js
```diff
@@ -792,48 +797,113 @@ export default (sbp('sbp/selectors/register', {
...
+      })
+    }
+    return sbp('chelonia/contract/sync', listOfIds)
+  },
+  // the `try` parameter does not affect (ephemeral or persistent) reference
+  // counts, but rather removes a contract if the reference count is zero
+  // and the contract isn't being monitored for foreign keys. This parameter
+  // is meant mostly for internal chelonia use, so that removing or releasing
+  // a contract can also remove other contracts that this first contract
+  // was monitoring.
```
taoeffect, line 852: Could you further elaborate how it is possible for a contract to find itself in the state where the retain count is 0 and yet it hasn't been removed?

Is it for contracts that are sync'd without calling `retain` on them? (If such a thing is possible, note that this isn't how Objective-C works, where any time an object is created, it begins with a retain count of `1`). Perhaps some thought should be given to the possibility of setting retain counts to `1` automatically when a contract is first sync'd?

shared/domains/chelonia/chelonia.js
```diff
@@ -763,14 +771,11 @@ export default (sbp('sbp/selectors/register', {
   },
   // 'chelonia/contract' - selectors related to injecting remote data and monitoring contracts
   // TODO: add an optional parameter to "retain" the contract (see #828)
-  'chelonia/contract/sync': function (contractIDs: string | string[], params?: { force?: boolean, deferredRemove?: boolean }): Promise<*> {
+  'chelonia/contract/sync': function (contractIDs: string | string[], params?: { force?: boolean, resync?: boolean }): Promise<*> {
```
taoeffect, line 766: If a contract hasn't been sync'd before, the retain count should probably be set to `1`.

shared/domains/chelonia/chelonia.js
corrideat: > Could you further elaborate how it is possible for a contract to find itself in the state where the retain count is 0 and yet it hasn't been removed?

As it's mentioned in the proposal, there are _three_ separate counts. One of them is internal to Chelonia, namely the when contracts are being monitored for foreign keys. Because this counter isn't really stored as an integer (it's calculated dynamically), and it's managed by Chelonia, you can think of `try: true` as maybe decrementing that count and then removing the contract.

As you point out, the second reason why the count could be 0 (excluding foreign keys) is if `sync` is called directly. While maybe there are legitimate reasons for doing this if one looks hard enough, contract developers should _not_ call `sync` on contracts that haven't been `retain`ed.

> Perhaps some thought should be given to the possibility of setting retain counts to 1 automatically when a contract is first sync'd?

I think this could be confusing and tricky to implement.

Confusing because you should pair all your calls to `retain` with `release`. If sync magically changes the count value, it could be confusing.

On the other hand, `sync` on a non-`retain`ed contract _could_ throw an error, which is probably the correct thing to do. This is tricky because Chelonia does this when subscribing to a contract to monitor foreign keys.

shared/domains/chelonia/chelonia.js
taoeffect: Based on discussion, the idea would be to add an extra parameter called `_foreign` (with an underscore to indicate it's a internal parameter, not to be used by developers). If it is set, then the persistent retain count is set to `0`, otherwise, if we are subscribing to the contract for the first time, it's set to `1`

taoeffect: Based on discussion, the idea would be to add an extra parameter called something like `_ignoreRetainCount` (with an underscore to indicate it's a internal parameter, not to be used by developers).

- If it is not set, then the persistent retain count is set to `1` if we are subscribing to the contract for the first time.
- If it is set, the retain count is left unmodified to avoid screwing up foreign contracts

taoeffect: Update: I'm admitting defeat here, and am reverting my position. My original attempt was to try to get rid of `try` in `release`. But I couldn't do that because of arguments @corrideat made related to recursive foreign contracts.

So then I thought maybe at least we can get `sync` to behave more like `alloc] init]` in ObjC, but looking at the changes requested here, this would be strictly a complication of the code, not a simplification.

So therefore, we're going to leave the approach as-is, and state in the documentation that `contract/sync` should never be called without first calling `retain`, and that `retain` should be used in place of `contract/sync` when syncing a contract for the first time.

shared/domains/chelonia/chelonia.js
```diff
@@ -21,6 +21,14 @@ import './files.js'
 import './time-sync.js'
 import { buildShelterAuthorizationHeader, eventsAfter, findForeignKeysByContractID, findKeyIdByName, findRevokedKeyIdsByName, findSuitableSecretKeyId, getContractIDfromKeyId } from './utils.js'
 
+const clearObject = (o: Object) => {
+  Object.keys(o).forEach((k) => delete o[k])
+}
+
+const reactiveClearObject = (o: Object, fn: (o: Object, k: PropertyKey) => any) => {
+  Object.keys(o).forEach((k) => fn(o, k))
+}
```
taoeffect: These seem like functions that would work as utilities in `utils.js`

frontend/controller/actions/identity.js
```diff
@@ -343,6 +343,15 @@ export default (sbp('sbp/selectors/register', {
       sbp('state/vuex/commit', 'login', loginAttributes)
       await sbp('chelonia/storeSecretKeys', () => transientSecretKeys)
 
+      if (!state) {
```
taoeffect: Question about the comment above in line 335 (I can only comment on changed lines):

```js
      // TODO: This is temporary until username is no longer needed internally
      // in contracts
```

Is this still needed now that you went through and changed everything from usernames to identityContractIDs?

frontend/controller/actions/identity.js
```diff
@@ -343,6 +343,15 @@ export default (sbp('sbp/selectors/register', {
...
+        // Make sure we don't unsubscribe from our own identity contract
+        // Note that this should be done _after_ calling
+        // `chelonia/storeSecretKeys`: If the following line results in
+        // syncing the identity contract and fetching events, the secret keys
+        // for processing them will not be available otherwise.
+        await sbp('chelonia/contract/retain', identityContractID)
```
taoeffect: - Does it really make sense to call retain on the identity contract here?
- Is it possible for the retain count to be incremented too high this way?
- Should this code maybe be moved inside of a `try` block, like right after line 373?

frontend/model/contracts/chatroom.js
taoeffect: OK, well, a reminder then to output a `console.warn` line

frontend/model/contracts/group.js
```diff
@@ -1154,6 +1160,8 @@ sbp('chelonia/defineContract', {
                     }
                   }
                 }).catch((e) => {
+                  // If already joined, ignore this error
+                  if (e?.name === 'GIErrorUIRuntimeError' && e.cause?.name === 'GIGroupAlreadyJoinedError') return
```
taoeffect, line 1164: Where in the code do we throw a `GIErrorUIRuntimeError` with a `cause` of `GIGroupAlreadyJoinedError`? 🤔

corrideat: This error class (`GIGroupAlreadyJoinedError `) is new, along with a few others, and is thrown in the group contract. I added this after your comment of not relying on the error message.

frontend/controller/actions/identity.js
corrideat: That seems to be setting `username` if it's not been provided. I think doing this is fine but maybe the comment itself isn't (I don't remember the reason for it, but I had the intention of removing `username` from almost everywhere)

frontend/controller/actions/identity.js
corrideat: > Does it really make sense to call retain on the identity contract here?

Yes. There is no state at that point, so the identity contract must be retained.

> Is it possible for the retain count to be incremented too high this way?

It's not, as this only happens when there's no state

> Should this code maybe be moved inside of a try block, like right after line 373?

Sure.

frontend/model/contracts/group.js
corrideat: The `cause` itself is set in `frontend/controller/actions/utils.js:244` (`throw new GIErrorUIRuntimeError(userFacingErrStr, { cause: e })`), where `e` may have been thrown by the contract.

corrideat:
I've now incorporated all of the feedback (see the last two commits). If there are no further changes requested, this should now be ready to be merged.

frontend/model/contracts/chatroom.js
taoeffect: > When person B comes back online, the side-effect will be executed (because they're subscribed to the contract) and they'll then leave.

Why would the sideEffect be re-executed if they're currently syncing the contract when they get this message? See above this code, we have these lines:

```js
        if (sbp('chelonia/contract/isSyncing', contractID)) {
          return
        }
```

Won't that prevent `leaveChatRoom` from getting called?

corrideat: You're right. I'm not sure why those lines are there and for my response to make sense they should be removed.

frontend/controller/actions/group.js
```diff
@@ -527,16 +506,21 @@ export default (sbp('sbp/selectors/register', {
...
+        data: {
+          groupContractID: groupId,
+          inviteSecret: secret
+        }
+      }).then(() => {
+        return sbp('gi.actions/group/switch', groupId)
+      })
+    } finally {
+      await sbp('chelonia/contract/release', groupId, { ephemeral: true })
+    }
```
taoeffect: Why are we calling retain and release on the `groupId` here?

Shouldn't we just wait for the sideEffect of `joinGroup` to get us to join the group?

corrideat: This replaces the old `JOINING_GROUP` globals. We can't (just) wait for the side-effect because we may be re-joining, and the old leave action may remove the group. This `retain`/`release` ensures that the group is kept around until we know what to do next.

corrideat: (Compare this to the process for joining a chatroom, which uses the same pattern and for the same reason)

frontend/model/contracts/group.js
```diff
@@ -1810,16 +1747,26 @@ sbp('chelonia/defineContract', {
...
             }
           }
         }).catch(e => {
+          if (e.name === 'GIErrorUIRuntimeError' && e.cause?.name === 'GIChatroomAlreadyMemberError') {
+            if (actorID === memberID) {
+              // Increase reference count if we've already joined
+              sbp('chelonia/contract/retain', chatRoomId)
+            }
+            return
+          }
```
taoeffect: In what sort of situation would we want to increase the retain count here again?

corrideat: When re-syncing contracts (e.g., from a new device), as the reference count would be 0 and the action would fail to send with `GIChatroomAlreadyMemberError`.

frontend/controller/actions/group.js
taoeffect: OK, then two questions:

1. What happens if this block gets executed before the old leave action is processed? Won't it have zero effect in that case?
2. If your logic is sound, could you please add an explanatory comment to the code explaining all of this since this seems unintuitive.

corrideat: > What happens if this block gets executed before the old leave action is processed? Won't it have zero effect in that case?

It won't be executed in its entirety before the old leave action is processed because sending the `joinGroup` action (or any action) requires being up to date with the contract. However, if it somehow happened, it'd have 0 effect as you say, which is still correct.

frontend/model/contracts/group.js
taoeffect: > as the reference count would be 0 and the action would fail to send with GIChatroomAlreadyMemberError.

Which action would fail to send?

This is inside of a `.catch`, the action has already failed?

corrideat: The `gi.actions/chatroom/join` action, of course. And since it failed, that's why this is in the `.catch` block, to see why it failed.

taoeffect: So, we are trying to join a chatroom that we are already a part of. If we are already a part of it, why would the reference count be `0`?

corrideat: Because this is _a new device_ with an _empty state_. The reference count is 0 because retain hasn't been called yet, which this is doing.

frontend/model/contracts/group.js
```diff
@@ -1871,7 +1808,7 @@ sbp('chelonia/defineContract', {
           // Now, let's see if we had a key request that's been answered
           const sentKeyShares = sbp('chelonia/contract/successfulKeySharesByContractID', state, identityContractID)
           // We received a key share after the last time we left
-          if (sentKeyShares?.[identityContractID]?.[0].height > state.profiles[member].departedHeight) {
+          if (sentKeyShares?.[identityContractID]?.[0].height > state.profiles[memberID].departedHeight) {
```
taoeffect, line 1820: This is part of a whole bunch of new code that was added (although in the diff it doesn't look that way because it's being added by the removal of multiline comments `/* .. */`).

So it's a significant change. Why did things work correctly before without all of this code? Are we sure this code works the way it's supposed to? Doesn't it seem a bit odd for the sideEffect of a `removeMember` action to call `join`? Shouldn't this be done someplace else? And what if there are multiple joins/rejoins, could this code result in duplicate joins being called?

frontend/model/contracts/group.js
taoeffect: I see.... OK. This too, is super unintuitive. Although this code already has a comment, could you please expand on it to make it very clear about this specific situation, to mention that this is the situation where this a new device syncing from scratch to a channel we've already joined on another one?

frontend/model/contracts/group.js
corrideat: This is part of a previous attempt at removing `JOINING_GROUP`. Before things worked correctly without because `JOINING_GROUP` was checking for these cases.

> Doesn't it seem a bit odd for the sideEffect of a removeMember action to call join?

It might seem odd, but this is handling re-joining. In particular, this is addressing the case where the last action is to leave, but we have since requested to re-join (which is a two-step process). I'm not sure if there's another place that's more suitable, as this is part of the contract logic in terms of what should be the next steps.

> And what if there are multiple joins/rejoins, could this code result in duplicate joins being called?

Since this is at the end of the queue, what matters is that the last action we sent was to leave, but we haven't yet sent `inviteAccept`.

taoeffect: Why can't we do this near the code that resulted in these key shares being in the contract?

And what specific situation is this (e.g. logging in on a new device... or..?)

Could all of this be elaborated on further in the code comment above?

corrideat: > Why can't we do this near the code that resulted in these key shares being in the contract?

Because we can't (currently) run custom code when receiving `OP_KEY_SHARE` or `OP_KEY_REQUEST`.

> And what specific situation is this (e.g. logging in on a new device... or..?)

This is logging in, not necessarily from a new device, but from an older state.

> Could all of this be elaborated on further in the code comment above?

Sure.

shared/domains/chelonia/chelonia.js
```diff
@@ -792,48 +789,113 @@ export default (sbp('sbp/selectors/register', {
...
+      })
+    }
+    return sbp('chelonia/contract/sync', listOfIds)
+  },
+  // the `try` parameter does not affect (ephemeral or persistent) reference
+  // counts, but rather removes a contract if the reference count is zero
+  // and the contract isn't being monitored for foreign keys. This parameter
+  // is meant mostly for internal chelonia use, so that removing or releasing
+  // a contract can also remove other contracts that this first contract
+  // was monitoring.
```
taoeffect, line 852: Could you add the following elaboration on the `try` parameter that explains why it exists:

> The `try` parameter is an internal parameter (not intended for use by developers), that helps Chelonia remove foreign contracts that are no longer being watched.
> The reason why we have it, and cannot simply use reference counting to do this automatically, is because we have two different types of reference counting (persistent and ephemeral). And therefore, if, for example, there was a foreign contract that was monitored for an ephemeral reason at one point, and a persistent reason at a different point, then because ephemeral reasons don't get saved like persistent reasons, it would be difficult to know when to unsubscribe from it.

TBD: tomorrow come up with a concrete scenario that proves that the above is a true statement.

corrideat:
```ts
sbp('chelonia/contract/retain', 'A', { ephemeral: true })
// Will also call sbp('chelonia/private/in/syncContractAndWatchKeys', 'B')
// Now 'A' has ephemeral count set to 1 and no non-ephemeral count
// Say that 'B' also has ephemeral count set to 1 and no non-ephemeral count
// (because we use `ephemeral` from the call to 'A')

sbp('chelonia/contract/retain', 'A', { ephemeral: false })
// Now 'A' has ephemeral count set to 1 and non-ephemeral count set to 1
// What do we do with 'B'?

sbp('chelonia/contract/release', 'A', { ephemeral: true })
// Now 'A' has no ephemeral count set and non-ephemeral count set to 1
// What do we do with 'B'?
// If the page is refreshed, 'B' won't have any reference count and will be
// released, which is wrong.

sbp('chelonia/contract/release', 'A', { ephemeral: false })
// Now we remove 'A'
// How do we know whether to use `{ ephemeral: true }`? Using the wrong one will
// result in an exception.

```

taoeffect (approved):
This was not an easy task, and you pulled it off and did a great job!

PR #1958, opened 2024-04-19
#1948 - Wrong overlay UI bug on brave

### LINKED ISSUE #1948: Dragging a person's profile icon in the chat brings up the file upload overlay forever
### Problem

In Brave, I dragged a person's profile icon and the file upload overlay appeared:

Screenshot 2024-04-18 at 12 27 54 PM

What's especially annoying is that it remained visible. Either it forces you to upload a file from the file picker, or you're stuck with this overlay.

### Solution

Nothing should happen when you drag someone's profile icon. The file should not be uploaded, and the file upload overlay shouldn't appear either.

SebinSong:
closes #1948

### DIFF
--- a/frontend/views/components/Avatar.vue
+++ b/frontend/views/components/Avatar.vue
@@ -8,6 +8,7 @@
     @error='revokeObjectURL'
     ref='img'
     v-on='$listeners'
+    draggable='false'
   )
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Nice! 👍

Issue #1945, opened 2024-04-18
Clicking "report the error" in the emergency banner doesn't work


taoeffect:
### Problem

![tmp-1713455351100_720](https://github.com/okTurtles/group-income/assets/138706/94632500-b192-47c7-8472-56b76baf0677)

In the PWA on Android/GrapheneOS, I noticed that clicking the "report the error" link in #1944 didn't actually take me to the `UserSettingsModal`.

Instead it opened up a new browser window that I think took me to the dashboard.

This is wrong, it shouldn't open up a new browser window, and it should take me to the `UserSettingsModal`.

### Solution

This code needs to be fixed:

```js
export function LError (error: Error): {|reportError: any|} {
  let url = `/app/dashboard?modal=UserSettingsModal&tab=application-logs&errorMsg=${encodeURI(error.message)}`
  if (!sbp('state/vuex/state').loggedIn) {
    url = 'https://github.com/okTurtles/group-income/issues'
  }
  return {
    reportError: L('"{errorMsg}". You can {a_}report the error{_a}.', {
      errorMsg: error.message,
      'a_': `<a class="link" target="_blank" href="${url}">`,
      '_a': '</a>'
    })
  }
}
```

At minimum, `target="_blank"` shouldn't be used with the modal. But it might require more than that to fix this (potentially even setting up JS logic...).

taoeffect:
Note: I have the same problem clicking "Report the error" here: #2018

Silver-IT:
Update the component `BannerGeneral` and `BannerScoped` just like how we handle onsite redirects inside MessageBase in PR #2005.

This is the comment to me. 😉

Silver-IT:
This issue seems like to be very much similar to the Issue #1949, so I am going to resolve this issue inside the PR #2005.

PR #1965, opened 2024-04-26
Save recent messages in chatroom contract

### LINKED ISSUE #1960: Unexpected behaviour in creating a new channel
### Problem

**Steps to reproduce**
- [Tab1] `user1` creates a group, and redirects to the General chatroom
- [Tab2] `user1` logs in a group, and redirects to the General chatroom
- [Tab1] `user1` creates a channel
- [Tab2] `user1` redirects to the new channel (should not be redirected)

I think this is related to the updated workflow of the how to sync chatroom, process events, ...
Also, I would like to say that it doesn't happen in the **always**, it happens **sometimes**. (I am creating this issue after I did many number of testing, and found 3 times of happening.) But there is really problem.

https://github.com/okTurtles/group-income/assets/57976479/70af0afe-95e4-4fee-95e1-025801a07b2b

### Solution

Fix it

### LINKED ISSUE #1961: Incorrect position of isNew indicator in Chat page
### Problem

**1. When entering the channel for the first time**
- [Tab1] `user1` creates a group
- [Tab2] `user2` joins the group and redirects to the General chatroom
- [Tab1] `user1` redirects to the General chatroom and creates a DM between `user2`
- [Tab1] `user1` sends messages to `user2` thru DM
- [Tab2] `user2` opens the DM between `user1`, but there is no `isNew` indicator even though all the messages are new

**2. Anytime**
I can not say the exact steps to reproduce at the moment, but the position of `isNew` indicator sets wrong when we switch channels. Try to test more and find out the exact steps to reproduce, so that we can trim possible reasons of this error.

Additionally, I recommend to use `height` instead of `createdDate` when we manage the `readUntil` message. `createdDate` was used before we add `height` to the message, but using `height` seems to be better now, I think.

### Solution

Fix them

### LINKED ISSUE #1962: Save the recent messages in the chatroom contract state
### Problem

In order to solve the issue described [here](https://github.com/okTurtles/group-income/issues/1619#issuecomment-2067529936), we need to save the latest message(or messages) in the chatroom contract state. Now, we don't save any messages in the chatroom contract in order to reduce the memory usage, but instead we are save the recent events in the local storage for each chatroom.
I think we should not use the local storage for the recent events, and should save the recent messages in the chatroom contract state. Additionally, it will be more faster when we open any chatroom too.

### Solution

Implement this

Silver-IT:
### Summary of Changes

- Closes #1962 
- Closes #1960 
- Closes #1961


frontend/model/contracts/chatroom.js
```diff
@@ -123,6 +124,17 @@ async function deleteEncryptedFiles (manifestCids: string | string[], option: Ob
   }
 }
 
+function cutMessagesIfOverflow (state) {
+  // NOTE: 'shouldSaveMessages' attribute is not original attribute
+  //       and it is set in Chat page only to save all messages
+  if (state.shouldSaveMessages) {
```
corrideat: I see why this is now a different attribute, but I'd pick a different name that better reflects what this does. `shouldSaveMessages`, apart from saving messages, also disables some validation.

frontend/model/contracts/chatroom.js
```diff
@@ -123,6 +124,17 @@ async function deleteEncryptedFiles (manifestCids: string | string[], option: Ob
   }
 }
 
+function cutMessagesIfOverflow (state) {
```
corrideat: I'd refactor this function to work more like a circular list rather than pushing and then shifting.

corrideat (commented):
Notice that there'll likely be quite a few conflicts with master now that #1955 has been merged.

frontend/controller/actions/group.js
```diff
@@ -678,12 +678,18 @@ export default (sbp('sbp/selectors/register', {
...
-          sbp('state/vuex/commit', 'setCurrentChatRoomId', { chatRoomId: chatRoomID, groupId: msg.contractID() })
+          const fnEventHandled = (cID, message) => {
+            if (cID === chatRoomID) {
+              if (sbp('state/vuex/getters').isJoinedChatRoom(chatRoomID)) {
+                sbp('state/vuex/commit', 'setCurrentChatRoomId', { chatRoomID, groupID: msg.contractID() })
+                sbp('okTurtles.events/off', EVENT_HANDLED, fnEventHandled)
+              }
+            }
+          }
+          sbp('okTurtles.events/on', EVENT_HANDLED, fnEventHandled)
```
taoeffect, line 697: The `onprocessed` hook is already designed to be called on `EVENT_HANDLED` after the message is received back and processed.

So this unnecessary handler here needs to be removed.

frontend/common/translations.js
```diff
@@ -133,13 +133,14 @@ export default function L (
...
   let url = `/app/dashboard?modal=UserSettingsModal&tab=application-logs&errorMsg=${encodeURI(error.message)}`
+  const target = !sbp('state/vuex/state').loggedIn ? 'target="_blank"' : ''
   if (!sbp('state/vuex/state').loggedIn) {
     url = 'https://github.com/okTurtles/group-income/issues'
   }
   return {
     reportError: L('"{errorMsg}". You can {a_}report the error{_a}.', {
       errorMsg: error.message,
-      'a_': `<a class="link" target="_blank" href="${url}">`,
+      'a_': `<a class="link" ${target} href="${url}">`,
```
taoeffect: I don't think this change does anything useful, should be undone.

There is a related issue #1945, but this change doesn't address that.

frontend/model/contracts/chatroom.js
taoeffect: > but I'd pick a different name that better reflects what this does

@corrideat do you have any suggestions?

frontend/model/contracts/chatroom.js
taoeffect: Personally I see no issue with this approach.

frontend/model/contracts/chatroom.js
```diff
@@ -222,11 +231,11 @@ sbp('chelonia/defineContract', {
           notificationType,
           notificationType === MESSAGE_NOTIFICATIONS.ADD_MEMBER ? { memberID, actorID: innerSigningContractID } : { memberID }
         )
-        const newMessage = createMessage({ meta, hash, height, data: notificationData, state, innerSigningContractID })
-        state.messages.push(newMessage)
+        state.messages.push(createMessage({ meta, hash, height, state, data: notificationData, innerSigningContractID }))
+
+        cutMessagesIfOverflow(state)
```
taoeffect: I think this should be DRY'd by doing something like:

```js
addMessage(state, createMessage({ meta, hash, height, state, data: notificationData, innerSigningContractID }))
```

This way it's 1-line instead of 2 lines repeated everywhere. In other words, rename `cutMessagesIfOverflow` to `addMessage` and do the pushing and checking in there.

taoeffect (changes requested):
Nice work @Silver-IT! Review ready!

frontend/model/contracts/chatroom.js
```diff
@@ -213,14 +213,20 @@ sbp('chelonia/defineContract', {
...
+            Vue.set(state, 'members', {})
+          }
           if (state.members[memberID]) {
             throw new GIChatroomAlreadyMemberError(`Can not join the chatroom which ${memberID} is already part of`)
           }
         }
 
         Vue.set(state.members, memberID, { joinedDate: meta.createdDate })
 
+        if (!state.attributes) return
```
corrideat, line 233: This patch solves the issue of the action failing to process, but you may want to add a comment explaining why this is needed (and perhaps I'd also use `has(state, 'attributes')` instead).

frontend/model/contracts/chatroom.js
```diff
@@ -213,14 +213,20 @@ sbp('chelonia/defineContract', {
         if (!memberID) {
           throw new Error('The new member must be given either explicitly or implcitly with an inner signature')
         }
+
         if (!state.shouldSaveMessages) {
```
corrideat: Still unsure about the name for this field, but I also can't think of a much better name. What about keeping the old one? The issue with this name is that:

* It saves all messages, but the name implies that messages aren't saved if not set. This is incorrect, as some messages are still saved (hence this PR)
* Besides saving messages, it skips some validation, which is needed for rendering, but which means that it shouldn't be used in other contexts.

If you don't like the previous name, what about something like `renderingContext`, along with a comment explaining what it does and why the rendering code needs this different logic?

frontend/controller/actions/group.js
Silver-IT: Here, we are not waiting for the current event (`gi.contracts/group/joinChatRoom`) to be finished. The function `fnEventHandled` is to wait for `gi.contracts/chatroom/join` event to be finished, and the user is fully joined chatroom.

frontend/model/contracts/chatroom.js
Silver-IT: Good idea!

frontend/model/contracts/chatroom.js
Silver-IT: What do you think, @taoeffect? If you are fine with the `renderingContext`, let's go with that.

taoeffect: @Silver-IT I actually like `renderingContext` a lot! It seems like a great name for this 👍

frontend/controller/actions/group.js
taoeffect: This seems like a very strange place to do this then.

Wouldn't it make more sense to do this in the `sideEffect` of `gi.contracts/chatroom/join`?

Silver-IT: I don't think we should do that way because it will cause the issue #1960.

taoeffect: I see, in that case, since this looks kind of confusing and weird, could you please add a comment? Something along these lines (right above line 683):

```js
// joinChatRoom sideEffect will trigger a call to 'gi.actions/chatroom/join', we want
// to wait for that action to be received and processed, and then switch the UI to the
// new chatroom. We do this here instead of in the sideEffect for chatroom/join to
// avoid causing the UI to change in other open tabs/windows, as per bug:
// https://github.com/okTurtles/group-income/issues/1960
```

taoeffect (changes requested):
Left one request for a comment, and also ran into an odd bug:

Screenshot 2024-05-07 at 5 29 38 PM

1. [Tab1] u1 creates group
2. [Tab2] u2 joins group
3. [Tab2] u2 creates a new group chatroom
4. [Tab1] u1 views the new group chatroom and joins it
5. [Tab1] u1 posts a message linking back to `#general`
6. [Tab2] u2 clicks on `#general` chatroom link

A `NavigationDuplicated` error appears, even though we are not in the general chatroom.

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -317,7 +318,7 @@ export default ({
       this.$router.push({
         name: 'GroupChatConversation',
         params: { chatRoomID: obj.chatroomId }
-      })
+      }).catch(logExceptNavigationDuplicated)
```
taoeffect: While this might hide the error, I don't think it will fix the issue, which is that we're supposed to switch to the #general chatroom and it doesn't

Silver-IT: Oh, do you mean it doesn't redirect to the `#general` chatroom? On my side, it redirects well. I will check several times more again.

Silver-IT: > A NavigationDuplicated error appears, even though we are not in the general chatroom.

I can see the `#general` chatroom link has the link to the `#new-chatroom`. Two chatrooms IDs are same; one is from the error message and the other is from the current tab link.

But I can't reproduce this error actually.

taoeffect: Huh. OK I can't reproduce it now anymore either. Weird. 🤷‍♂️

Silver-IT: @taoeffect, I think this error is already fixed with [this commit](https://github.com/okTurtles/group-income/pull/1965/commits/4c5ba994c1d83abff88d42563d54c71de7f63455). Can you please test it again and keep me updated?

taoeffect: I mentioned above that I can't reproduce anymore

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -752,94 +736,82 @@ export default ({
...
+          if (entryHeight > 0) {
+            // NOTE: considering the KeyAdd GI event
+            //       authorizedKeys should not exist before _notBeforeHeight
+            Object.keys(state._vm.authorizedKeys).forEach(keyId => {
+              if (state._vm.authorizedKeys[keyId]._notBeforeHeight >= entryHeight) {
+                delete state._vm.authorizedKeys[keyId]
+              }
+            })
+          }
+
```
taoeffect: @corrideat could you review these and related changes to `ChatMain.vue` and give either your approval or a change request?

corrideat: @Silver-IT Can you explain your reasoning here? I don't think you should be doing this at all, for three reasons:

1. The `_notBeforeHeight` and `_notAfterHeight` attributes are checked (or, at least, should be; if they aren't, it's a bug) by Chelonia everywhere where they matter. This means that having 'extra' keys should not have any effect.
2. The goal is for contract developers not to have to mess with Chelonia internals, for the most part. Doing manual key management like this is at odds with that, and it also makes it difficult to maintain (for example, in the future we might change how the `not*Height` thing works to support adding and removing keys with different permissions throughout their lifetime) if there are future changes in Chelonia.
3. The work that I'm currently doing (moving Chelonia into a service worker) will more strictly separate Chelonia from the UI. While it's not clear yet that this particular addition won't work with the new architecture, it certainly goes against the principle of keeping things separate.

Besides the reasons above, I'm not sure why you're doing this. If you could explain, I could help see if there's another and better way to go about it.

corrideat: Other than that, there's an off-by-one error. If you want to remove keys in the future, the check should be `_notBeforeHeight > entryHeight`. As the name says, _not before_, the range is inclusive. So, a key with `_notBeforeHeight` of `x` is valid for height `x`, but _not before_ that. So, it's not valid for heights `[0, x-1]`. ETA: If the reason is what I commented below, I think this is correct, but at least a comment is warranted, since it looks wrong.

corrideat: I think I see it now. Are you trying to remove keys from the future so that key operations for those keys do get processed? If so, that makes some sense, but doesn't it still work without it? ]

Silver-IT: > I think I see it now. Are you trying to remove keys from the future so that key operations for those keys do get processed?

Yeah, since those keys would be added again when the future events are processed. I described this in the comment, but if it doesn't make sense, we should update it.

> If so, that makes some sense, but doesn't it still work without it?

I remember it made errors (I don't remember the error name now), and that's why I made additional work to remove those keys.

taoeffect: @Silver-IT could you please share more info about what errors you encountered?

I very much agree with @corrideat that this code shouldn't be here because we don't want developers to have to mess around with Chelonia internals in order to implement their app.

So if you can share the steps plus any errors that you encounter, it would help us figure out how to properly address this.

Silver-IT: > if you can share the steps plus any errors that you encounter, it would help us figure out how to properly address this.

Steps to reproduce
- [Tab1] user1 creates group, and create a new channel `channel-1`
- [Tab1] user1 sends more than 60 messages. (when I was testing, I sent 70 messages)
- [Tab2] user2 joins the group, and joins the channel `channel-1`
- [Tab2] user2 switches to general chatroom, and switches back to `channel-1`
- [Tab2] user2 scrolls up so that infinite scroll would work
- [Tab2] user2 would get the error in the console
image

This error happens inside the `keysToMap` function in `chelonia/internals.js`.

corrideat: How was this not an issue in the past? In either case, this is something that could be fixed in Chelonia.

Option A) Create a selector that does pretty much what you're doing here (actually, I've already defined a new selector (not in master) to get the contract state for a contract, it could also take a height parameter and remove keys.
Option B) Change the validation code that's failing to take _notBefore into account. If `_notBefore > thisMsgHeight`, then treat the key as not existing.

taoeffect: I've gone ahead and asked @corrideat to just post a commit directly to this PR with the fix, since he's responsible for the Chelonia internals at this point.

corrideat: @Silver-IT Take a look at https://github.com/okTurtles/group-income/pull/1993, implementing option A.

You can incorporate those changes into this PR using the web UI or in the terminal by using `git merge` or `git cherry-pick`.

frontend/model/contracts/group.js
```diff
@@ -342,7 +342,19 @@ const leaveChatRoomAction = (state, chatRoomID, memberID, actorID, leavingGroup)
...
-    ...extraParams
+    ...extraParams,
+    hooks: {
+      onprocessed: () => {
+        const rootState = sbp('state/vuex/state')
+        if (memberID === rootState.loggedIn.identityContractID) {
+          sbp('chelonia/queueInvocation', chatRoomID, () => {
+            sbp('chelonia/contract/release', chatRoomID).catch(e => {
+              console.error(`[leaveChatRoomAction] Error releasing chatroom ${chatRoomID}`, e)
+            })
```
taoeffect, line 356: Could you add a comment above line 350 to explain why this is being pushed to a queue?

Silver-IT: Okay, sure.

taoeffect (changes requested):
@Silver-IT this is looking really solid! 💪 

I did my best to test the new changes and I wasn't able to find any console errors! 🎉 

The only issue I found is that when creating a chatroom with a *lot* of messages in it, and you scroll to the bottom, then switch channels, switch back, and slowly scroll upward, once the spinner appears and the old messages load, it instantly jumps you to the very top of the chatroom, instead of leaving you where you are (but with more messages to scroll up to):

https://github.com/okTurtles/group-income/assets/138706/252aeeba-bba4-4ba0-9a82-a31261ad4316

Silver-IT:
@taoeffect, I know this is happening from the first time when we added infinite loading feature. Having tested several times on my side, this only happens when it loads the first message, in other words when no more infinite loading is needed. And it seems like it's the issue from the `infinite loading` vuejs library. Also it might be an expected behaviour, since it only happens when the infinite loading is finished, and when no more loading is needed.

Also, I thought this couldn't be a big problem, because as we keep using the chatroom, it's a very much rare case that scroll up to the top of the message list. And even though, the users scroll up to the top and encounter this issue, they would not have the bad feeling just like me because it seems like to say `No more messages now. This is the beginning`.

taoeffect (approved):
I see, OK, thanks for that clarification @Silver-IT ! I tested and verified what you said 👍 

Great work on this PR!! 😄

PR #1976, opened 2024-04-29
#1947 - Channel links in chat

### LINKED ISSUE #1947: Channel links in chat
### Problem

Screenshot 2024-04-18 at 11 49 25 AM

We need the ability to link to channels by name from within a chat channel.

### Solution

1. When a user types `#` in the chat area, similar to how they type `@`, we need a popup to appear showing all the channels names that the *currently logged in user has access to inside of the current group* (both public, group, and private channels)
2. When the message is rendered, the channel should appear in blue and clickable (unless it's a private chatroom that the reader doesn't have access to, see note below)

Example:

Screenshot 2024-04-18 at 12 01 13 PM

#### Special note on private chatrooms

When a private chatroom is linked to, the user who is sending it can see the link, and so can anyone who is already a member of the room:

Screenshot 2024-04-18 at 11 56 59 AM

But anyone outside of that room sees something else:

screen_shot_2024-04-18_at_11 53 01_am_360

Note that the color of the link should be different (grey), and also that it cannot be clicked, and also that both the blue and grey links use a different symbol: a 🔒 icon instead of a `#`

SebinSong:
closes #1947 

implementing various Channel-Link related UI elements which are in below.

**1] Active channel links & Auto-completion menu for channel mentions**

**2] A private channel is not clickable to someone who has not been invited**

### DIFF
--- a/frontend/views/containers/chatroom/CreateNewChannelModal.vue
+++ b/frontend/views/containers/chatroom/CreateNewChannelModal.vue
@@ -15,7 +15,7 @@
           :class='{ error: $v.form.name.$error }'
           v-model='form.name'
           data-test='createChannelName'
-          @input='onNameInput'
+          @input='debounceField("name")'
           @blur='updateField("name")'
           v-error:name=''
         )
@@ -203,20 +203,6 @@ export default ({
     },
     handlePrivacyLevel (e) {
       this.form.privacy = e.target.value
-    },
-    onNameInput (e) {
-      const newVal = e.target.value
-      const currVal = this.form.name
-
-      if (newVal.length) {
-        if (newVal === `${currVal} `) {
-          this.form.name = `${currVal}-`
-        } else {
-          this.form.name = newVal.toLowerCase()
-        }
-      }
-
-      this.debounceField('name')
     }
   },
   validations: {
@@ -243,6 +229,17 @@ export default ({
       privacy: {
         [L('This field is required')]: function (value) {
           return !this.isPublicChannelCreateAllowed || !!value
+        }
+      }
+    }
+  },
+  watch: {
+    'form.name' (newVal, oldVal) {
+      if (newVal.length) {
+        if (newVal === `${oldVal} `) {
+          this.form.name = `${oldVal}-`
+        } else {
+          this.form.name = newVal.toLowerCase()
         }
       }
     }
--- a/frontend/views/containers/chatroom/CreateNewChannelModal.vue
+++ b/frontend/views/containers/chatroom/CreateNewChannelModal.vue
@@ -232,6 +232,13 @@ export default ({
         }
       }
     }
+  },
+  watch: {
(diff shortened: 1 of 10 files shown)


frontend/views/containers/chatroom/CreateNewChannelModal.vue
```diff
@@ -232,6 +246,13 @@ export default ({
         }
       }
     }
+  },
+  watch: {
+    'form.name' (newVal, oldVal) {
+      if (newVal.length && newVal === `${oldVal} `) {
+        this.form.name = `${oldVal}-`
+      }
+    }
```
taoeffect: This same substitution is needed here too? 🤔 

Do you happen to know why? If so, could you add a comment explaining why it's also necessary here?

taoeffect (dismissed):
@SebinSong You did an amazing job on this PR!

I'm really quite impressed with this PR. It's a rather large PR, yet perfectly addresses the issue, and no matter how hard I tried, I couldn't find any bug in it.

I'm especially impressed with how you even got it to work where, when someone posts a link to a private chatroom and someone hasn't joined, it is grey, but then when they are added to the room, it becomes blue and clickable. 👏 

I could merge it as-is right now.

Or, if you'd like to make a couple of small changes to it now (instead of in a separate PR), I can wait.

One change is a request for a comment, and the other is related to a boyscoutting issue related to the "Channel Description" button when viewing a channel you haven't yet joined:

Screenshot 2024-04-30 at 9 43 19 AM

This button currently does nothing. I think it wouldn't be very professional to release the app with it. I also don't know what it's even supposed to do, so I updated [the Figma designs](https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=23766%3A267514&mode=design&t=lRcnO92ItfsTXBBo-1) to make it just say "Details", same as Slack.

However, I think we don't really need this button right now. So another simple change this PR could do would be to just hide this button entirely. What do you think, is that something that can be done for this PR?

frontend/views/containers/chatroom/CreateNewChannelModal.vue
SebinSong: @taoeffect No, This is something I forgot to delete actually. good catch.

SebinSong:
thanks @taoeffect. I've removed that 'channel description' button.

taoeffect (changes requested):
Perhaps removing some of that code wasn't the right thing to do...

Screenshot 2024-04-30 at 6 08 23 PM

Maybe the other code needed to be removed?

Or if both are needed, can you add a comment explaining why? (If you know)

SebinSong:
@taoeffect
Updated the PR with the fix.
Turns out the watcher was needed. the name field is bound to the component state via `v-model` so putting things in an `input` event handler didn't work.

taoeffect (approved):
Beautifully done @SebinSong !

PR #1978, opened 2024-04-30
Moved groupCreatorID out of group settings

### LINKED ISSUE #1956: Security: move `groupCreatorID` out of group settings
### Problem

Currently the `groupCreatorID` is stored in the group's `state.settings`.

This is a major problem because it is used to perform various permissions checks, and the settings can be overwritten by anyone using `'gi.contracts/group/updateSettings'`.

### Solution

Move the `groupCreatorID` out of the settings and rename it to `groupOwnerID`. Place it on the same object that stores the `.settings` key.

Update all contracts as necessary.

Although we could do this in a way that doesn't involve having to recreate the groups, don't worry about writing any migration code because we have to recreate the groups one last time anyway once #1955 is merged.

Silver-IT:
### Summary of Changes

- Closes #1956

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -202,7 +202,6 @@ export default (sbp('sbp/selectors/register', {
           settings: {
             // authorizations: [contracts.CanModifyAuths.dummyAuth()], // TODO: this
             groupName: name,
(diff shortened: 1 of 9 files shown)


taoeffect (approved):
LGTM!

PR #1979, opened 2024-04-30
Bugfix for the issue #1954

### LINKED ISSUE #1954: Can't create poll on mobile or search emoji picker
### Problem

On mobile (Android/GrapheneOS) using Brave's PWA version of the app, I can't seem to create a poll. When the modal appears, the moment I tap the "Ask a question!" input field, the entire modal disappears:

https://github.com/okTurtles/group-income/assets/138706/5a4b41e8-75f6-45ee-a178-f70db9324f50

The same thing happens when I bring up the emoji picker and tap the "Search" field, making it impossible to search for a specific emoji.

### Solution

Find and fix the bug!

SebinSong:
closes #1954 

The cause of the issue in both `CreatePoll.vue` and `Emoticons.vue` components was that there is resize handler that closes the pop-up when it detects viewport changes. But then what happens in touch devices is, when the input field is clicked, the keyboard pad is pulled out which triggers viewport size changes.

The fix I made here is to update the logic to take this into account.

### DIFF
--- a/frontend/views/containers/chatroom/CreatePoll.vue
+++ b/frontend/views/containers/chatroom/CreatePoll.vue
@@ -232,6 +232,14 @@ export default {
         this.$v.form.$reset()
         this.close()
       })
+    },
+    resizeHandler () {
+      if (window.matchMedia('(hover: hover)').matches) {
+        // This is a fix for the issue #1954(https://github.com/okTurtles/group-income/issues/1954)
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Appears to work well, and excellent job on adding that great comment linking back to the issue! 👏 😎 👍

PR #1981, opened 2024-05-01
Groundwork to move Chelonia into a service worker

corrideat:


shared/serdes/index.js
```diff
@@ -0,0 +1,126 @@
...
+// Note that it's paramount that the _receiving_ side call `deserializer.register`
+// with all of the possible types that are supported.
+// For how to implement `serialize` and `deserialize` support for custom types,
+// see the example below for a class `X`.
+
+// Internal function to mark the result of 'serializer' as internal, so that it
+// doesn't get accidentally reprocessed.
+const rawResult = (obj: Object) => {
+  Object.defineProperty(obj, raw, { value: true })
+  return obj
```
taoeffect, line 21: Could you:

- Add a comment explaining why `Object.defineProperty` is used instead of `obj[raw] = true`?
- I think this can be shortened to a 1-liner if you put the `return` in front of the `Object` on line 20

shared/serdes/index.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+    return value
+  })
+}
+
+// Internal lookup table for registered deserializers
+const deserializerTable = Object.create(null)
+
+// The `deserializer` function reconstructs data on the receiving side
+export const deserializer = (data: any) => {
```
taoeffect, line 95: What is passed in? I see `JSON.stringify` is called on `data`. Normally a deserializer takes in a serialized value, which in the case of JSON is a string. However, the fact that `JSON.stringify` is being used here indicates that this is not a JSON string. If so, then it sounds like `data` is the result of calling `JSON.parse` elsewhere on the serialized JSON string that was passed over the message port..? Or am I mistaken? Anyway, this all should be clarified and explaining in a comment.

shared/serdes/index.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// The `serializer` function prepares data before sending it as a message
+export const serializer = (data: any) => {
+  const verbatim: any[] = []
+  return JSON.parse(JSON.stringify(data, (_key: string, value: any) => {
+    if (value && value[raw]) return value
+    if (value === undefined) return rawResult(['_', '_'])
+    if (!value) return value
+    if (Array.isArray(value) && value[0] === '_') return rawResult(['_', '_', ...value])
```
taoeffect, line 33: Can you add a comment explaining what this is checking for?

frontend/controller/actions/group.js
```diff
@@ -680,27 +660,11 @@ export default (sbp('sbp/selectors/register', {
...
-          const fnEventHandled = (cID, message) => {
-            if (cID === chatRoomID) {
-              if (sbp('state/vuex/getters').isJoinedChatRoom(chatRoomID)) {
-                sbp('state/vuex/commit', 'setCurrentChatRoomId', { chatRoomID, groupID: msg.contractID() })
-                sbp('okTurtles.events/off', EVENT_HANDLED, fnEventHandled)
-              }
-            }
-          }
-          sbp('okTurtles.events/on', EVENT_HANDLED, fnEventHandled)
-        },
```
taoeffect, line 699: Why was this removed?

frontend/controller/app/group.js
```diff
@@ -0,0 +1,42 @@
...
+    // to wait for that action to be received and processed, and then switch the UI to the
+    // new chatroom. We do this here instead of in the sideEffect for chatroom/join to
+    // avoid causing the UI to change in other open tabs/windows, as per bug:
+    // https://github.com/okTurtles/group-income/issues/1960
+    if (sbp('state/vuex/getters').isJoinedChatRoom(chatRoomID)) {
+      sbp('state/vuex/commit', 'setCurrentChatRoomId', { chatRoomID, groupID: params.contractID })
+    }
+
+    return chatRoomID
+  }
```
taoeffect: Why was this separated out like this?

I see no explanation for the separation of `actions` and `app` folders.

Could you maybe create a `README.md` inside of `controller/` that explains this?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        const result = await sbp('gi.ui/prompt', promptOptions)
+        if (!result) {
+          return sbp('gi.app/identity/logout')
+        } else {
+          throw err
+        }
+      }
+      */
```
taoeffect, line 193: This was important and commonly used code that comes into play frequently during development (anytime the server is restarted).

Why is it commented out and how will this be handled instead?

EDIT: oh I see, is the code below supposed to replace it?

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -7,6 +7,216 @@
...
+        return
+      }
+      errorNotification('process', e, message)
+    },
+    sideEffectError: (e: Error, message: GIMessage) => {
+      sbp('gi.ui/seriousErrorBanner', e)
+      sbp('okTurtles.data/set', 'sideEffectError', message.hash())
+      errorNotification('sideEffect', e, message)
+    }
+  } */
```
taoeffect, line 177: How are these going to be handled?

frontend/model/database.js
```diff
@@ -34,14 +165,17 @@ class EmptyValue extends Error {}
 export const SETTING_CURRENT_USER = '@settings/currentUser'
 
 sbp('sbp/selectors/register', {
+  'gi.db/ready': function () {
+    return localforage.ready()
+  },
   'gi.db/settings/save': function (user: string, value: any): Promise<*> {
-    return appSettings.setItem(user, value)
+    return appSettings.setItem('u' + user, value)
```
taoeffect, line 179: What's the reason for this `u` prefix? (Can you add a comment?)

frontend/model/state.js
```diff
@@ -531,7 +542,260 @@ const getters = {
...
+  },
+  groupCreatedDate (state, getters) {
+    return getters.groupProfile(getters.currentGroupOwnerID).joinedDate
+  },
+  groupMincomeAmount (state, getters) {
+    return getters.groupSettings.mincomeAmount
+  },
+  groupMincomeCurrency (state, getters) {
+    return getters.groupSettings.mincomeCurrency
+  },
```
taoeffect, line 605: As far as I can tell, these and related getters were copied here from the group contract.

1. Why?
2. Why were they not deleted from group.js? (DRY)

Getters are useful in contracts. If you are moving so many of them out of contracts then we need to come up with a total replacement for them and rewrite a lot of the code in the group.js contract.

Or we need to come up with a way to make them work the way they did before (or in similar fashion) without violating DRY.

corrideat: > As far as I can tell, these and related getters were copied here from the group contract.

Correct.

 1. Because that seemed like the easiest thing to do with the fewest number of changes required (an alternative would have been to move them to a different file which is then imported from both places).
 2. Because they're needed there. In particular, after this change, contract getters and root getters are no longer the same and cannot call each other (at least, not synchronously).

> Getters are useful in contracts. If you are moving so many of them out of contracts then we need to come up with a total replacement for them and rewrite a lot of the code in the group.js contract.

That's correct. I think we should probably get rid of the concept of getters in contracts altogether, as that's a Vue.js concept anyhow. However, a shorter term alternative is to import them from a different shared file, to avoid source duplication (there'll still be duplication in the build artifacts).

frontend/model/database.js
corrideat: It's to disambiguate the type of data that's being stored. I can add a comment.

frontend/controller/actions/group.js
corrideat: Because that deals with the Vuex state and doesn't belong into actions.

shared/serdes/index.js
corrideat: You're maybe getting confused because of the calls to `JSON.stringify` and `JSON.parse`. Those are incidental and an implementation detail. Both `serialize` and `deserialize` use `JSON.parse(JSON.stringify(data, a), b)`. The 'magic' is in what `a` and `b` do, and `JSON` happens to be used because it traverses objects and is relatively fast.

shared/serdes/index.js
corrideat: > this can be shortened to a 1-liner if

I don't think that contributes to readability, because it's not self-evident that `Object.defineProperty(a, b)` returns `a`.

frontend/controller/app/group.js
corrideat: I'm not sure if this is an actual question or just a request to create a `README.md` explaining it, but the `app` name was your idea. The reason for the separation is that `app` runs on the... app (as in, in the browser window context) and actions run in the service worker context.

frontend/controller/serviceworkers/sw-primary.js
corrideat: This is still in progress and I don't have an answer at this time. The problematic part here is the notifications. Having contracts run in a SW will require re-thinking how notifications work.

frontend/model/state.js
taoeffect: I definitely agree we would need to at the very least fix the DRY issue in the source files.

> I think we should probably get rid of the concept of getters in contracts altogether, as that's a Vue.js concept anyhow.

It's a very useful concept for building Vue.js GUIs. We would need to keep it just for that, as otherwise it would require rewriting a bunch of the frontend code as well, and we don't want to do that not just because we don't have time for it, but because we want to support Vue.js and similar libraries that use the concept of getters.

So what is your proposal...?

frontend/controller/app/group.js
taoeffect: >  just a request to create a README.md explaining it

👍

frontend/model/state.js
corrideat: > It's a very useful concept for building Vue.js GUIs

Precisely my point. It's useful for **Vue.js**. It's not generally useful outside of Vue.js ('it' refers to the 'getters' API, not to the idea of a function that returns a value), because it's not a universal concept (e.g., Redux doesn't have them) and even if it were a common enough concept, it's currently only useful (in the sense of having 'getters' be part of a contract definition) because the API is made to be exactly what Vue.js uses, which may be different for other frameworks.

I'm not saying to get rid of them on Vue.js, Vue.js can still use them. What I'm saying is that Vue.js doesn't have access to them in contracts and so the whole idea that getters can be shared is just not feasible. So, for _contracts_, there'd be nothing to replace them. However, since contracts can run arbitrary code, they can define their own getters and for applications that use Vue.js, they can use the same getter signature that's used in Vue.js.

corrideat: To expand on the not feasible part:

I mean that it cannot be literally the same function reference, as it's done now on `master`, where `sbp('state/vuex/getters').groupCreatedDate` will use some code that's defined in the group contract. This is because the UI is running on the main thread and the group contract is running in a SW (potentially in yet another context after sandboxing). What _can_ be done, as I suggested, is moving the group getters definitions to a different file, and importing that from the app and the contract.

taoeffect:
Closing in favor of #2357

Issue #1190, opened 2022-03-25
Automate testing updating an existing production install


taoeffect:
### Problem

Currently we don't test two things:

1. The production build (which might involve a different bundler #1182)
2. Updating an existing install

### Solution

We should create a special `data/` folder with an existing group with users (e.g. `test/data/`), tell the backend to load it as part of a production build, and test it using Cypress with the new code to verify that everything appears as it should with no errors in the console.

PR #1986, opened 2024-05-08
Fix proposal issues

### LINKED ISSUE #1971: Proposal notification using incorrect icon
### Problem

The notification for a recently closed proposal is using a random user's avatar rather than any appropriate proposal icon. (The choice of user may not be truly random, but regardless, using a user icon is incorrect.)

### Solution

Wherever an icon/avatar is needed for a proposal, the icon used should be the appropriate proposal icon. 
See proposal iconography on Figma: https://www.figma.com/file/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?type=design&node-id=1876%3A17778&mode=design&t=KWfuzXQrEmv1KfXw-1

### LINKED ISSUE #1972: Proposal notifications not working properly
### Problem

A proposal was opened, voted on, and closed. However
1. there was no red badge notification on the browser tab indicating there were new notifications to check
2. there is an alert about the proposal in my notifications drawer, but it is marked as read even though I did not click on it or mark it as read
3. there is no evidence of said proposal on the dashboard (see issue #1969)

### Solution

1. Ensure that new proposals trigger the red badge notification.
2. Ensure that notifications are not marked as read unless the user marks them as read. (Perhaps the initial proposal notification is incorrectly marked as read once voting closes?)

Silver-IT:
### Summary of Changes

- Closes #1971 
- Closes #1972


Silver-IT (commented):
Review ready!

frontend/model/contracts/group.js
```diff
@@ -840,12 +840,16 @@ sbp('chelonia/defineContract', {
...
+            sbp('chelonia/queueInvocation', contractID, () => {
+              sbp('chelonia/queueInvocation', innerSigningContractID, () => {
+                sbp('gi.notifications/emit', 'PAYMENT_RECEIVED', {
+                  createdDate: meta.createdDate,
+                  groupID: contractID,
+                  creatorID: innerSigningContractID,
+                  paymentHash: data.paymentHash,
+                  amount: getters.withGroupCurrency(payment.data.amount)
+                })
+              })
```
taoeffect: Could you please add a comment above line 843 explaining why this is being wrapped in two queues?

Perhaps even better would be to try this a bit by turning this into a contract method that's called, and adding the comment there.

E.g. something like:

```js
sbp('gi.contracts/group/emitAfterSyncing', [contractID, innerSigningContractID], 'PAYMENT_RECEIVED', {
  createdDate: meta.createdDate,
  groupID: contractID,
  creatorID: innerSigningContractID,
  paymentHash: data.paymentHash,
  amount: getters.withGroupCurrency(payment.data.amount)
})
```

Then it would use `'chelonia/contract/wait'` on the passed in contractIDs instead of `'chelonia/queueInvocation'`.

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -73,14 +51,14 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
     [PROPOSAL_INVITE_MEMBER]: {
       [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to add new members to the group.', options),
       [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to add new members is expiring.', options),
-      [PROPOSAL_VARIANTS.ACCEPTED]: L('Added members to this group: {to}', options),
+      [PROPOSAL_VARIANTS.ACCEPTED]: L('Added members to this group: {group}', options),
```
taoeffect: I'm confused by this message.

Shouldn't this say who was added instead of the group name?

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -73,14 +51,14 @@ const interactiveMessage = (proposal, baseOptions = {}) => {
...
       [PROPOSAL_VARIANTS.REJECTED]: L('No members were added.'),
       [PROPOSAL_VARIANTS.EXPIRED]: L('No members were added.')
     },
     [PROPOSAL_REMOVE_MEMBER]: {
       [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to remove members from the group.', options),
       [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to remove members is expiring.', options),
-      [PROPOSAL_VARIANTS.ACCEPTED]: L('Left {title}', options),
+      [PROPOSAL_VARIANTS.ACCEPTED]: L('Left {member}', options),
```
taoeffect: Shouldn't this say something like, `{member} was removed from the group.` ?

frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -43,7 +43,7 @@ li.c-item-wrapper(data-test='proposalItem')
         banner-scoped(ref='voteMsg' data-test='voteMsg')
         p.c-sendLink(v-if='invitationLink' data-test='sendLink')
           i18n(
-            :args='{ user: proposal.data.proposalData.memberName}'
+            :args='{ user: proposal.data.proposalData.memberID}'
```
taoeffect: `memberID` is a variable name that is used to show the member's identity contractID.

But here we want a either a username or a display name.

frontend/views/containers/proposals/ProposalVoteOptions.vue
```diff
@@ -107,7 +107,7 @@ export default ({
         if (oneVoteToPass(proposalHash)) {
           if (this.type === PROPOSAL_INVITE_MEMBER) {
             passPayload = await createInvite({
-              invitee: this.proposal.data.proposalData.memberName,
+              invitee: this.proposal.data.proposalData.memberID,
```
taoeffect: Same isue here, I'm pretty sure `memberID` isn't the right name for the variable

taoeffect (changes requested):
Nice work @Silver-IT!

Preliminary review ready!

frontend/model/contracts/group.js
Silver-IT: 1. That's a good idea to make a new contract function like `emitAfterSyncing`. Then, I think I should add a comment at the top of the function.
2. About this approach (that uses several `chelonia/queueInvoation`), I had a discussion with @corrideat in [here](https://okturtles.slack.com/archives/C02QF71TK1Q/p1714988249312109?thread_ts=1714724867.814369&cid=C02QF71TK1Q). If we make a new chelonia function like `chelonia/contract/waitInSequence`, I think I can use that function. Should I make the new chelonia function in this PR?

frontend/views/containers/proposals/ProposalItem.vue
Silver-IT: Nice catch!

frontend/views/containers/chatroom/MessageInteractive.vue
Silver-IT:
```
[PROPOSAL_INVITE_MEMBER]: {
      [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to add {member} to the group.', options),
      [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to add {member} is expiring.', options),
      [PROPOSAL_VARIANTS.ACCEPTED]: L('Proposal from {from} to add {member} is accepted.', options),
      [PROPOSAL_VARIANTS.REJECTED]: L('Proposal from {from} to add {member} is rejected.', options),
      [PROPOSAL_VARIANTS.EXPIRED]: L('Proposal from {from} to add {member} is expired.', options)
},
```

These are more precise, aren't they?

I would like to remind that we are not make any messages of type `MESSAGE_TYPES.INTERACTIVE`, but only one in the function `gi.actions/group/notifyExpiringProposals` when the proposal is expiring.

frontend/views/containers/chatroom/MessageInteractive.vue
Silver-IT:
```
[PROPOSAL_REMOVE_MEMBER]: {
      [PROPOSAL_VARIANTS.CREATED]: L('{from} wants to remove {member} from the group.', options),
      [PROPOSAL_VARIANTS.EXPIRING]: L('Proposal from {from} to remove {member} is expiring.', options),
      [PROPOSAL_VARIANTS.ACCEPTED]: L('Proposal from {from} to remove {member} is accepted.', options),
      [PROPOSAL_VARIANTS.REJECTED]: L('Proposal from {from} to add {member} is rejected.', options),
      [PROPOSAL_VARIANTS.EXPIRED]: L('Proposal from {from} to add {member} is expired.', options)
}
```

Then, are you fine with these messages?

frontend/model/contracts/group.js
taoeffect: > About this approach (that uses several chelonia/queueInvoation), I had a discussion with @corrideat in [here](https://okturtles.slack.com/archives/C02QF71TK1Q/p1714988249312109?thread_ts=1714724867.814369&cid=C02QF71TK1Q). If we make a new chelonia function like chelonia/contract/waitInSequence, I think I can use that function. Should I make the new chelonia function in this PR?

Well, I saw that discussion, but I don't think it's necessary to use `chelonia/queueInvoation` to emit the event. `chelonia/queueInvoation` is used instead of `wait` when you want to make sure that an event is added to the event queue and that no other asynchronous events can happen until that event is done executing.

In this case it doesn't matter because we're just calling `'gi.notifications/emit'`, which has no impact on either of the event queues. So it's not necessary to create `chelonia/contract/waitInSequence`, you can just use `/wait` to wait until both contract queues are done, and then emit the event.

frontend/views/containers/chatroom/MessageInteractive.vue
taoeffect: > These are more precise, aren't they?

Yes, these seem much better 👍

frontend/views/containers/chatroom/MessageInteractive.vue
taoeffect: Yes 👍

frontend/model/contracts/group.js
Silver-IT: > `chelonia/queueInvoation` is used instead of wait when you want to make sure that an event is added to the event queue and that no other asynchronous events can happen until that event is done executing. In this case it doesn't matter because we're just calling 'gi.notifications/emit', which has no impact on either of the event queues. 

We need it. That's because according to the updated workflow of joining group, we start syncing the identity contracts for the other group members after we fully finished the syncing group contracts. That's because the following codes are in the `sbp('chelonia/queueInvocation', contractID, () => {})` in the `/inviteAccept/sideEffect #1190`.
```
sbp('chelonia/contract/retain', profileIds).catch((e) => {
  console.error('Error while syncing other members\' contracts at inviteAccept', e)
})
```
But based on the previous codebase, we emit the notifications while we are syncing the group contracts at the time of we are ignorance of the group member's identity contracts. If you don't understand what I mean, you could try the following steps in **master** branch.
- [Tab3] user1 logs in the group, and has the weird notification with identity contract ID, instead of username

I think the best solution to handle notification text is that we should not generate the **text** at the time of emitting notification. The notifications should be saved as an **object format**, and they should be rendered in the `NotificationList.vue` or any other necessary places. That's because if we generate the notification text at the time of its creation, they are not updated to the user's profile, exactly the displayName at the moment. So we could have the following notifications for the same user `alexjin`.
```
The group has a new member. Say hi to alexjin!
Alex Jin proposed to add a member to the group. Vote now!
Alex created a proposal. Vote now!
SilverIT  proposed to change the group voting system. Vote now!
```
I am planning to create an issue for this issue this week.

taoeffect: @Silver-IT ok, however, if that is what happens when using `/wait`, it will also happen when using `/queueInvocation`, because the timing will be the same. The only difference between using `/wait` with a `.then` (or an `await`), and `/queueInvocation`, is that `/queueInvocation` will have the function run on the queue instead of outside of it. Let me try to visualize this with a diagram:

```
Using /wait with groupID and identityContractID:

[--- groupID finishes syncing ---][---- identityContractID finishes syncing ---][-- some other event --]
                                                                                     [-- emit notification --]


Using /queueInvocation inside of identityContractID inside of groupID

[--- groupID finishes syncing ---][---- identityContractID finishes syncing ---][-- emit notification --][-- some other event --]
```

As you can see, `[-- emit notification --]` will happen after both contracts have finished syncing in both cases, and the only difference is that when using `/wait`, the notification might be emitted in the middle of some other event being processed on one of the queues.

> I am planning to create an issue for this issue this week.

Could you create and use the `'gi.contracts/group/emitAfterSyncing'` method in the meantime? It's fine to address any other issues in a future PR.

Silver-IT: > if that is what happens when using /wait, it will also happen when using /queueInvocation, because the timing will be the same.

@taoeffect, I don't think so. As you can see, `chelonia/contract/wait` uses `Promise.all` and all the `chelonia/queueInvocation` would run **in parallel**, in other words **at the same time**.
```
'chelonia/contract/wait': function (contractIDs?: string | string[]): Promise<*> {
  const listOfIds = contractIDs
    ? (typeof contractIDs === 'string' ? [contractIDs] : contractIDs)
    : Object.keys(sbp(this.config.stateSelector).contracts)
  return Promise.all(listOfIds.flatMap(cID => {
    return sbp('chelonia/queueInvocation', cID, ['chelonia/private/noop'])
  }))
}
```
But, what I did is that `sbp('chelonia/queueInvocation', innerSigningContractID, () => {})` would run **after** the group contractID is **fully synced** because it is inside the queued invocation.

taoeffect: OK, I see what you mean. Then, feel free then to either implement `chelonia/contract/waitInSequence`, or implement it directly inside of `'gi.contracts/group/emitAfterSyncing'`, e.g. something like:

```js
for (const id of contractIDs) {
  await sbp('chelonia/queueInvocation', id, ['chelonia/private/noop'])
}
// ... emit notification
```

taoeffect (approved):
Nicely done @Silver-IT!

PR #1990, opened 2024-05-09
#1935 - Give mobile users a way to know where a link goes

### LINKED ISSUE #1935: Give mobile users a way to know where a link goes
### Problem

On Desktop, when someone shares a link in the chat, users can know where it goes without clicking it by hovering the mouse over it. On mobile, there is no such feature, and there is no way to know where the link goes. This can be a potential security risk for users.

### Solution

One way that Slack attempts to address this is by letting users long-press a link. When they do this, a panel appears with more information about the link and various options:

However, when a link is long-pressed in Group Income, this panel appears instead:

This is the panel for interacting with a message.

What we need is to be able to identify when a link within a message is long-pressed, and display a different panel, similar to the Slack one, instead.

We need to show at least:

- What the full link is
- The ability to open the link
- The ability to copy the link

SebinSong:
closes #1935 

Building below overlay UI component and implement each menus in them.

**[Dark theme]**

**[Light theme]**


frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -328,6 +330,16 @@ export default ({
         name: 'GroupChatConversation',
         params: { chatRoomId: obj.chatroomId }
       })
+    },
+    longPressHandler (e) {
+      console.log('!@# asfasadf')
```
taoeffect: This kind of logging needs to be removed before merging

frontend/views/containers/chatroom/TouchLinkHelper.vue
```diff
@@ -0,0 +1,230 @@
+<template lang='pug'>
+.c-touch-link-helper(:class='{ "is-active": ephemeral.isActive }')
+  .c-overlay-background(@click='close')
```
taoeffect, line 3: We do have [`vue-clickaway`](https://github.com/simplesmiler/vue-clickaway), but the overlay makes more sense because it dims the background 👍

taoeffect (changes requested):
Beautifully designed @SebinSong! 🤩 

Everything seems to work well in my tests. Just some stray logging that needs to be removed

taoeffect (approved):
LGTM!

frontend/views/containers/chatroom/TouchLinkHelper.vue
SebinSong: Yes and Plus, it also allows us to dismiss the component when any point on the background is pressed.

PR #1993, opened 2024-05-13
Move key deletion key to chelonia

corrideat:
This PR implements a new selector, `chelonia/contract/state` and moves the logic to exclude future keys into this selector, so that `ChatMain.vue` doesn't need to do that.

Note that it was a bit flaky getting this to work and I had to use `this.renderingChatRoomId` instead of the more intuitive `currentChatRoomId` (which seems to be what was currently being used).

Note also that `ChatMain.vue` will need to be refactored to a large degree (not because of this PR, but because of moving Chelonia into a service worker). The main change will be that everything using `chelonia/` will need to be asynchronous (including this new selector).


Silver-IT (approved):
LGTM!

PR #1995, opened 2024-05-13
#1987 - Improve Channel name validation

### LINKED ISSUE #1987: Improve validation for channel chatroom names
### Problem

The validation that's currently done to ensure no spaces are used in chatroom names isn't "real" and can be bypassed fairly easily:

Screenshot 2024-05-08 at 8 31 15 AM

You can bypass it using autocomplete suggestions, or by highlighting the `-` and changing it to a space, and then you can create the chatroom.

### Solution

These improvements are necessary to the chatroom:

1. The substitution must also be performed on "blur", i.e. when focus is switched away from the chatroom name field
2. The substitution must be performed when the "Create channel" button is clicked
3. **Real** validation must be performed by the contract action on the channel name to ensure that even if the UI is bypassed, it is still impossible to create invalid chatroom names.

SebinSong:
closes #1987 

1. updated the logic in the watcher so it works when **copy/paste** and **blur** happen too.
2. Add name-validation logic to `gi.contracts/group/addChatRoom` for when the UI is somehow by-passed.
3.  As a result of **2.**, All chatroom names in the cypress test had to be updated too.

@taoeffect As you pointed out in the issue description, I was going to add validation logic for when the 'Create' button is clicked but actually covering `blur` seemed to suffice. (Wasn't able to come up with any scenario where the name is incorrect by the time the 'Create' button is clicked, because the auto-correction on `blur` already takes care of them before we get to that point.)

But let me know if you find something else to cover during your review.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -1358,10 +1358,26 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/group/addChatRoom': {
       // The #General chatroom is added without an inner signature
-      validate: objectOf({
-        chatRoomID: string,
-        attributes: chatRoomAttributesType
-      }),
+      validate: (data) => {
+        objectOf({
+          chatRoomID: string,
+          attributes: chatRoomAttributesType
+        })(data)
+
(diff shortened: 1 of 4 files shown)


taoeffect (approved):
Excellent work @SebinSong! 👏

Issue #1934, opened 2024-04-17
Improve behavior of chat formatting buttons on mobile


taoeffect:
#### Problem 1

When I tried messing around with the bold/italic/etc buttons on mobile, they didn't work very well:

![1000031407](https://github.com/okTurtles/group-income/assets/138706/57a00a1d-45da-45aa-99d5-af48117ca7ef)

When I highlighted a word that was right above, say, the "Bold" button, and then tapped the bold button, the word wouldn't be made bold. It made me think the buttons didn't work.

I realized later that the buttons do work, but Android/GrapheneOS thinks I'm trying to change my selection because of how close my finger is to the highlighted text when I tap the Bold button, preventing the button from actually being tapped.

#### Problem 2

Problem 2 is that even if the button does get tapped successfully, it deselects the highlighted text. This isn't that big of a problem, but it's a bit annoying, because it means that if I change my mind and want to unbold the text, or turn it into a link, I can't simply tap another button, I have to re-highlight the text first. Other apps, like Slack, don't behave this way, they keep the text highlighted.

#### Problem 3

On iOS, selecting a word causes the built-in action buttons to appear below the selected text instead of above it, hiding the formatting buttons:

![file](https://github.com/okTurtles/group-income/assets/138706/e324738e-633e-4d97-aa61-093be23fd722)

#### Problem 1

I recommend solving Problem 1 by increasing the amount of space between the input field and the formatting toolbar. Perhaps add an extra 3 or 4 pixels of margin. Whatever is the minimum amount of pixels needed to get it to work somewhat reliably.

#### Problem 2

For Problem 2, see if there's a way to keep the current selection highlighted after formatting has been applied.

#### Problem 3

See if there's a CSS selector or something to tell iOS to show the menu above the highlighted text instead of below it.

The problems can be solved in the same PR or separate PRs.

PR #1999, opened 2024-05-18
Resolving two issues in #1934

SebinSong:
work on **problem 1. & 2.** of the issue #1934 

@taoeffect 
for the **problem 1.** , I actually went a bit ambitious and increased the buttom padding to `1.25rem` but please let me know if you feel it's too much.

And as a fix for the **problem 2.** , the target text segment now remains selected after pressing the format buttons.

---

For **problem 3.** there which is an issue related iOS's built-in call-out menu, there seems to be no CSS rule to control behaviour of it nor disable/enable it. (`-webkit-touch-callout` rule doesn't work for this case)
I will keep researching on this issue with iOS simulator on my Mac and send another PR once I figure out the solution.

Thanks,

### DIFF
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -90,7 +90,7 @@ export default ({
   },
   methods: {
     show () {
-      if (!this.manual) this.isActive = true
+      if (!this.manual && !this.deactivated) this.isActive = true
     },
     hide () {
       if (!this.manual) this.isActive = false
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -85,6 +85,7 @@
             v-if='ephemeral.showButtons'
             direction='top'
             :text='L("Add reaction")'
+            :deactivated='ephemeral.isPhone'
           )
             button.is-icon(
               :aria-label='L("Add reaction")'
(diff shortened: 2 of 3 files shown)


taoeffect (changes requested):
Nicely done @SebinSong! This seems much improved.

One issue I noticed is that when toggling the formatting, the formatting icon remains blue instead of toggling on and off.

E.g. highlight some text, press the italic icon, and press it again. It will remain highlighted in blue instead of toggling itself off and on.

Note: I tested this on my phone using `grunt dev --tunnel`

SebinSong:
@taoeffect 
Whether the format button is highlighted or not doesn't mean that it's toggled on/off. It's that the button remains focused after clicking the button. I will make a fix where it blurs after the text transformation (just like how those buttons in text editors of the Github work)

taoeffect (changes requested):
tmp-1716172239537

@SebinSong please give this a try on your end using `grunt dev --tunnel`. I tried the latest changes on my phone and there were still a few problems:

- The latest changes didn't seem to do anything, the blue icon remained blue after being tapped
- It's still really hard to tap the icon while text is highlighted when the lighted text is right above the icon. Sometimes Android still gets confused doesn't realize you're trying to press the icon and instead thinks you're messing with the selection
- Additionally, the white hover bubble appears above the button, e.g. in this case "Strikethrough", covering and hiding the text that's being modified so that the user doesn't really have any idea what's happening / how the text is being modified. Perhaps the best way to deal with this is to just not show it on mobile.

taoeffect (approved):
Nicely done!

PR #2005, opened 2024-05-22
Implement message link inside message

### LINKED ISSUE #1945: Clicking "report the error" in the emergency banner doesn't work

### LINKED ISSUE #1949: Linking to messages doesn't work
### Problem

1. Copy a link to a message in chat (say, from one of the first messages near the top)
2. Paste the link in the chat and hit send
3. Click the link

Expected result:

- Chat view should jump to the linked message and highlight it

Actual result:

- A new window appears, and we aren't even taken to the message

### Solution

Fix it so that we get the expected result. No new windows should open. We should be immediately taken to that message, or shown an error, according to the following rules:

- If the message is from a different chatroom that we are a part of, we should also be taken to that message
- If the message is from a public or group chatroom that we are not a part of, we should be switched to that chatroom and shown that message, but *not* joined to that chatroom
- If the message is from a *private* chatroom that we are not part of, clicking the link should present a dialog (via `gi.ui/prompt`) that says the `Sorry, this message is from a private chatroom that you are not part of.`, with a single button that says `OK`

Silver-IT:
### Summary of Changes

- Closes #1949
- Closes #1945


taoeffect (changes requested):
Nice, it seems to work really well on Desktop, but on mobile it doesn't work at all.

Please try with `grunt dev --tunnel` and then with a cellphone + Desktop.

On desktop you'll have URLs that look like this: http://localhost:3000/....

You'll need to replace that with the URL that localtunnel gives you, e.g. https://gi5209.loca.lt/... in order to test on mobile.

Silver-IT:
> Nice, it seems to work really well on Desktop, but on mobile it doesn't work at all.

Having tested in mobile, it works fine to me. Both in browser and PWA. @SebinSong, can you please try to review this PR? Especially, test how message link works in mobile.

SebinSong:
@Silver-IT I will review this PR once I have time this week.

> Especially, test how message link works in mobile.

you can test it yourself using `grunt dev --tunnel` btw.

EDIT: I just realised you mentioned you already did. ;)

Silver-IT:
> I will review this PR once I have time this week.

Thanks, @SebinSong. When you test this PR, do not make mistake in creating message links.
In order to test message link in mobile, you should create link in mobile.
And in order to test it in desktop, you should create link in desktop browser.

That's because hostnames would be different; one would be `http://localhost:8000` (Of course, when you do not use tunnel to test this PR in your desktop), and the other would be like `https://gi5209.local.lt`.

cc: @taoeffect

SebinSong:
@taoeffect 

> That's because hostnames would be different; one would be http://localhost:8000 (Of course, when you do not use tunnel to test this PR in your desktop), and the other would be like https://gi5209.local.lt.

That's too obvious... but thnx for a reminder.

Silver-IT:
I will mark this PR as draft, and I am going to fix the Issue #1945 inside this PR. That issue is very much similar to the current issue #1949.

Silver-IT:
Need to consider the Issue #2041 too. This is the comment to me.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -783,7 +783,11 @@ export default ({
         if (events.length) {
           // NOTE: if 'messageHashToScroll' was not there in the messages of the contract state
           //       we need to retrieve more events, and render to scroll to that message
-          this.updateScroll(messageHashToScroll, Boolean(mhash))
+          this.updateScroll(messageHashToScroll, Boolean(mhash)).then(() => {
+            if (mhash) {
+              this.$router.replace({ query: {} })
+            }
+          })
```
Silver-IT: It's to remove `mhash` in queries after scroll to and highlight the message with `mhash`. Since we use only `mhash` as query in the Chat page, so just made query empty.

frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
```diff
@@ -21,7 +23,33 @@ const RenderMessageWithMarkdown: any = {
...
+                href && sbp('okTurtles.events/emit', OPEN_TOUCH_LINK_HELPER, href)
+                e?.preventDefault()
+              }
+            }
+          },
+          hasChildren
+            ? entry.children.map(child => recursiveCall(child))
+            : undefined
+        )
+      } else if (entry.tagName) {
```
Silver-IT: This is to find in-app links which are created by `makeInAppLinkElement` function and create elements in real.

frontend/views/pages/GroupChat.vue
```diff
@@ -186,12 +200,25 @@ export default ({
...
+              alert(L('Sorry, this message is from a private chatroom that you are not part of.'))
+              this.$router.go(-1)
+            }
+          } else {
+            this.updateCurrentChatRoomID(chatRoomID)
+          }
+        } else if (mhash) {
+          this.$refs.chatMain?.scrollToMessage(mhash).then(() => {
+            this.$router.replace({ query: {} })
+          })
```
Silver-IT: When user clicks the link to the message of the current chatroom, it is handled in here.

frontend/views/utils/markdown-utils.js
```diff
@@ -1,4 +1,16 @@
...
+import { TextObjectType } from '~/frontend/utils/constants.js'
+
+export const makeInAppLinkElement = (data?: Object): {
+  prefix: string, suffix: string
+} => {
+  return {
+    prefix: `<router route='${JSON.stringify(data ?? {})}'>`,
+    suffix: '</router>'
+  }
+}
```
Silver-IT: To create a temporary string for in-app link element, we can use this function.

frontend/views/utils/markdown-utils.js
```diff
@@ -127,6 +150,35 @@ export function injectOrStripLink (
   }
 }
 
+export const filterOutInAppLinksFromSafeHTML = (textInSafeHTML: string): Array<Object> => {
```
Silver-IT: This function finds the temporary string for in-app links, which is generated by `makeInAppLinkElement`, from the text and returns an array of two types of Object; one is for InAppLink, and the other is for original text.

Silver-IT (commented):
Ready for the review!

frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
SebinSong: @Silver-IT , it looks like this additional block has some duplicate codes with the `else if (entry.tagName) { ... }` below.
I would DRY this by including the logic in there like,

```js
if (entry.tagName) {
  const hasChildren = Array.isArray(entry.children)
  const isRouter = entry.tagName === 'ROUTER'

  const elName =  isRouter ? 'a' : entry.tagName.toLowerCase()
  const opts = isRouter
    ? {
          // options you created above
       }
    :  {
          // default options
        }
    
  ....
  
   return createElement(
     elName,
     opts,
     hasChildren
        ? entry.children.map(child => recursiveCall(child))
        : undefined
   )
}
```

SebinSong:
Hi @Silver-IT , Just roughly read through this PR and (correct me if I'm wrong) what some of logics here does is apparently, it intercepts the link markdown and 

1. check if hostname in the `href` is the same as the app.
2. if so, turns this into `<router> ... </router>`
3. But then, this again is turned back into `<a>..</a>` tag later on down the line.

could you give some explanation on why adding **step 2. & 3.** here was necessary?

Cuz below block added in this PR feels redundant (Or it's that I don't have a full understanding on the context of the issue.)

```js
    on: {
      click: (e) => {
        route && this.$router.push(route)
        e?.preventDefault()
      },
      touchhold: (e) => {
        href && sbp('okTurtles.events/emit', OPEN_TOUCH_LINK_HELPER, href)
        e?.preventDefault()
      }
    }
```

`$router.push(...)` can be achieved via  `<a href='app-hostname/path'></a>` too and `sbp('okTurtles.events/emit', OPEN_TOUCH_LINK_HELPER, href)` here is already handled by the ancestor component which is `MessageBase.vue` ([here](https://github.com/okTurtles/group-income/blob/0bc1ba7f485d25a2ec45a740c07315ea5c87400d/frontend/views/containers/chatroom/MessageBase.vue#L204)).

Silver-IT:
> Could you give some explanation on why adding step 2. & 3. here was necessary?

@SebinSong, that's very nice catch. I've just realised those two steps seems not be necessary. Actually, the logic was implemented before you finished improving the way to render messages which contains markdown and mentions. But now, when your code is already implemented, I don't think those two steps are needed any more. 👍

And also I think we can use the `RenderMessageWithMarkdown` component for `BannerGeneral` and `BannerScoped` component either. Do you think it's fine too?

SebinSong:
@Silver-IT I just tried your fix in both desktop and in my mobile phone and it works well for both.

> And also I think we can use the RenderMessageWithMarkdown component for BannerGeneral and BannerScoped component either. Do you think it's fine too?

And yes, for anywhere it can be useful and also pls feel free to enhance the component to suit your need.

SebinSong (dismissed):
Great work @Silver-IT

taoeffect (changes requested):
Nice work @Silver-IT!

I tested this and it seems to work almost perfectly.

A few remaining issues:

### On mobile, clicking "report the error" shows wrong view

Screenshot 2024-06-14 at 1 18 31 PM

### Flow errors

```
Error ------------------------------ frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:40:27

Cannot assign object literal to `routerOptions.route` because property `route` is missing in object literal [1].
[prop-missing]

   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:40:27
   40|             routerOptions.route = { path, query }
                                 ^^^^^

References:
   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:30:31
   30|         const routerOptions = { isInAppRouter: false }
                                     ^^^^^^^^^^^^^^^^^^^^^^^^ [1]


Error ------------------------------ frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:41:27

Cannot assign `this.$router.resolve(...).href` to `routerOptions.href` because property `href` is missing in object
literal [1]. [prop-missing]

   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:41:27
   41|             routerOptions.href = this.$router.resolve(routerOptions.route).href
                                 ^^^^

References:
   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:30:31
   30|         const routerOptions = { isInAppRouter: false }
                                     ^^^^^^^^^^^^^^^^^^^^^^^^ [1]



Found 2 errors
The Flow check failed!
```

taoeffect (changes requested):
Almost ready to merge, still a flow error:

```
Error ------------------------------------------------------------------ frontend/controller/actions/identity.js:801:102

Cannot call `deleteResult.map` because property `map` is missing in possibly uninitialized variable [1].
[incompatible-use]

   frontend/controller/actions/identity.js:801:102
   801|       console.error('[gi.actions/identity/removeFiles] Some CIDs could not be deleted', deleteResult.map((r, i) => r.status === 'rejected' && toDelete[i]).filter(Boolean))
                                                                                                             ^^^

References:
   frontend/controller/actions/identity.js:766:9
   766|     let deleteResult, toDelete
                ^^^^^^^^^^^^ [1]



Found 1 error
The Flow check failed!
```

taoeffect (approved):
Ah, that flow error is from `master`, and @SebinSong fixes it in #2090

PR #2011, opened 2024-05-24
Implement pinned messages

### LINKED ISSUE #1963: Design for Pinned Messages in Chat
### Problem

We need to be able to pin important messages in chatrooms as well as in DMs.

### Solution

Create a design, similar to the below, (and later on, implement it).

Screenshot 2024-04-22 at 10 01 01 AM

The "Pinned" menu should appear only if there are pinned messages.

A "Pin message to chatroom" menu item should be added to in the dropdown menu for every message.

Silver-IT:
### Summary of Changes

- Closes #1963


taoeffect (changes requested):
Great work @Silver-IT!

I have not yet had a chance to review the code, but while testing I found a couple of issues:

### 1. positioning of pinned message button

Could you move the pinned message button to the same line as the other info (in front of the other channel info buttons)?

Screenshot 2024-06-11 at 9 41 26 AM

### 2. Error upon pinning message in 2nd tab

1. Create a group and join a user in tab2
2. in both tabs, post messages
3. [tab1] pin u2's message

in tab2, error, and the message doesn't appear pinned:

Screenshot 2024-06-11 at 9 42 24 AM

frontend/main.js
```diff
@@ -430,7 +430,7 @@ async function startApp () {
         }
       },
       isInCypress () {
-        return !!window.Cypress
+        return !!window.Cypress && window.sbp === sbp
```
taoeffect: I think the first condition is sufficient, and adding this second condition may cause problems.

There are other situations where we might want to set `window.sbp = sbp` and not be in Cypress. We already do this in some such situations, see `main.js`:

```js
  if (process.env.NODE_ENV === 'development' || window.Cypress || process.env.EXPOSE_SBP === 'true') {
    // In development mode this makes the SBP API available in the devtools console.
    window.sbp = sbp
  }
```

frontend/model/contracts/chatroom.js
```diff
@@ -530,19 +550,28 @@ sbp('chelonia/defineContract', {
...
         }
+
+        const msgIndex = findMessageIdx(hash, state.messages)
+        if (msgIndex >= 0) {
+          fnDeleteAttachment(state.messages[msgIndex])
+        }
+        const pinnedMsgIndex = findMessageIdx(hash, state.pinnedMessages)
+        if (pinnedMsgIndex >= 0) {
+          fnDeleteAttachment(state.pinnedMessages[pinnedMsgIndex])
+        }
```
taoeffect: Hmm, so the code is duplicating the pinned messages completely, even the attachments?

Is it possible to rewrite the code so that the pinned messages do not duplicate the data, but instead only reference the messages in `state.messages` by hash instead? This way we can DRY the code, save on memory,  and not worry about updating data in two different places.

As it is right now, the code has to have all of these changes to add these `fn*` functions all over the place to perform the same operation twice. Can we please instead add the message hashes to the pinned messages list?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -334,6 +335,13 @@ export default ({
         return this.currentUserAttr.id === this.currentGroupOwnerID
       }
       return false
+    },
+    isInCypress () {
+      // NOTE: By only checking the window.Cypress, it's difficult to say that the app is running in Cypress
+      //       because anyone can access the window object.
+      //       in Cypress mode, sbp is exposed to window object (see main.js) so window.sbp can be compared
+      //       with our sbp object to verify if the app is in Cypress mode in real
+      return !!window.Cypress && window.sbp === sbp
```
taoeffect: (See the same comment elsewhere about this. Also, might be a good thing to DRY using mixins)

taoeffect (changes requested):
Nice work @Silver-IT!

Preliminary review (round 1) done!

In addition to the comments I posted, I found some other issues while testing:

### 1. Missing "Unpin message" from menu item

As there is a "Pin message" menu item, there should be a corresponding "Unpin message" menu item that appears on pinned messages:

Screenshot 2024-06-12 at 10 14 39 AM

### 2. Pinned messages don't appear in DMs

The pinned messages menu should appear in DMs as well:

Screenshot 2024-06-12 at 10 17 13 AM

frontend/main.js
Silver-IT: But the problem is that `window.Cypress` is something that anyone can set through console.
```
window.Cypress = true
```
So I think we need to verify if it's real.

taoeffect: They can completely change the code on the frontend and there is nothing we can do to prevent that. The way we've designed the app to come to "consensus" among peers is how we deal with compromised clients. So the existing check we have is fine

Silver-IT:
> As there is a "Pin message" menu item, there should be a corresponding "Unpin message" menu item that appears on pinned messages:

I can't find the unpin icon in FontAwesome. Can I use the same icon(thumbtack) for unpin button?
image

Or this undo icon?
image

PS: I can see that Discord has same icon (thumbtack) for both `Pin` and `Unpin` button.

frontend/model/contracts/chatroom.js
Silver-IT: The reason why I made `fn*` function and called it twice for `state.messages` and `state.pinnedMessages` is that we don't save all messages in `state.messages`.
When users open channel, they should see `N pinned` in the chat page and also by clicking it, they should see all the pinned messages.

We can also save only message hashes. If so, it's fine to display `N pinned`, but when user clicks it, he needs to render all the events from the first pinned message to the latest event. And I think this is bad in terms of performance. That's why I saved the all pinned messages, instead of their hashes.

frontend/main.js
Silver-IT: Okay, since it's for fixing Cypress errors (Issue #2058, and #2059), I will update it in the PR #2063.

frontend/model/contracts/chatroom.js
taoeffect: I see, ok that make sense.

In that case, since the code is so similar, do you think we can DRY it a bit further? E.g. something like this maybe?

```js
        [state.messages, state.pinnedMessages].forEach(messageArray => {
          const msgIndex = findMessageIdx(hash, messageArray)
          if (msgIndex >= 0) {
            fnDeleteAttachment(messageArray[msgIndex])
          }
        })
```

BTW, that was suggested by an LLM. I used [Cursor](https://www.cursor.com/) and highlighted the code, and asked it to DRY it so that `fnDeleteAttachment` was only called once. 😄

frontend/model/contracts/chatroom.js
```diff
@@ -381,7 +385,7 @@ sbp('chelonia/defineContract', {
           addMessage(state, createMessage({ meta, data, hash, height, state, pending, innerSigningContractID }))
         } else if (direction !== 'outgoing') {
           // If an existing message is found, it's no longer pending.
-          delete existingMsg.pending
+          Vue.delete(existingMsg, 'pending')
```
taoeffect, line 388: Nice catch!

taoeffect (changes requested):
Nice improvements, this PR is shaping up 👍 

I noticed this PR modifies the code for voting on polls and noticed that it introduces some bugs:

Screenshot 2024-06-17 at 11 36 32 AM

For example, above u1 tries to vote on a poll u2 created, and even though they clicked "Submit" the poll doesn't switch to the show the results, and instead remains on the choice screen. That's why in the chat it shows u1 voting multiple times because they just click "Submit" again and again...

Please test all modified code thoroughly, and especially code for which we do not have tests yet.

frontend/views/containers/chatroom/PinnedMessages.vue
```diff
@@ -0,0 +1,338 @@
...
+  },
+  created () {
+    this.matchMediaDesktop = window.matchMedia('screen and (min-width: 1200px)')
+    this.matchMediaDesktop.onchange = (e) => {
+      this.ephemeral.isDesktopScreen = e.matches
+    }
+    this.ephemeral.isDesktopScreen = this.matchMediaDesktop.matches
+
+    window.addEventListener('resize', this.resizeHandler)
+    document.addEventListener('keydown', this.trapFocus)
```
taoeffect: When pressing the <kbd>TAB</kbd> key while entering the name of a poll to jump to the first option, this line results in this error:

Screenshot 2024-06-18 at 9 05 49 AM

taoeffect (changes requested):
Nice! Seems to be almost ready to merge. I found one more bug:

frontend/views/pages/DesignSystem.vue
```diff
@@ -149,7 +149,7 @@ page(
     section.card
       h2.is-title-2.card-header Icons
       p
-        | This is only the currently used icons (at _icons.scss). A complete list can be find here:&nbsp;
+        | This is only the currently used icons at (_icons.scss). A complete list can be find here:&nbsp;
```
taoeffect: This change should be reverted as it makes the sentence make less sense

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -621,6 +623,36 @@ export default ({
...
+      const contractID = this.renderingChatRoomId
+
+      const promptConfig = {
+        heading: L('Remove pinned message'),
+        question: L('Are you sure you want to remove this pinned message?'),
+        primaryButton: L('Yes'),
+        secondaryButton: L('Cancel')
+      }
+
+      const primaryButtonSelected = await sbp('gi.ui/prompt', promptConfig)
```
taoeffect, line 645: We do not need this prompt, it's better to just unpin it immediately just like Slack does

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -61,47 +61,65 @@ menu-parent(ref='menu')
       menu-item.hide-desktop.is-icon-small(
         tag='button'
         v-if='isEditable'
-        @click='action("editMessage")'
+        @click='action("editMessage", $event)'
```
taoeffect: What is this change doing? Is it possible to do it by doing `@click.stop=...` instead?

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
```diff
@@ -176,6 +176,11 @@ export default {
 
         aTag.setAttribute('href', url)
         aTag.setAttribute('download', attachment.name)
+
+        aTag.addEventListener('click', function (event) {
+          event.stopPropagation()
+        })
+
```
taoeffect: What was this added for?

taoeffect (changes requested):
Minor change requests and questions

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: I just followed the way how Slack works.
image

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
Silver-IT: This is to keep showing the `PinnedMessages` dialog when users try to download attachments inside the dialog.
image

frontend/views/containers/chatroom/MessageActions.vue
Silver-IT: Yeah, right!!!

taoeffect (approved):
Approved! Amazing work @Silver-IT !!

PR #2013, opened 2024-05-24
fix dbPath

taoeffect:
Now uses `./data` in both production and development (unless otherwise specified by `process.env.DB_PATH`)


Silver-IT (approved):
LGTM! 👍🏼

PR #2016, opened 2024-05-25
Resolve Problem 3. of #1988

### LINKED ISSUE #1988: Broken styles inside Markdown
### Problem

Writing the following messages would not be rendered correctly, and it looked like its below screenshot.

**[Problem 1.]** Problem with mentions inside the markdown

```
First sentence @andrea 
Second sentence @greg 
Third @alexjin sentence
Fourth Sentence @alexjin @greg @andrea 





Fifth sentence @all @all Sixth sentence @alexjin 
```
image

---

**[Problem 2.]** Problem with multiple line breaks
```
First sentence





Second sentence
```
image

---

**[Problem 3.]** Problem with mentions inside the bullets
```
- Drink water: Sometimes thirst can be mistaken for hunger. @alexjin Drinking a glass of water can help you feel full and reduce your appetite.
- Have a small, healthy snack: Choose something from the list we discussed or another healthy option to tide you over until mealtime. @greg - Plan your meal: If you know what you'll be having for your next meal, it can help to distract you from immediate hunger pangs. @andrea 
```
image

---

### Solution

Fix it

SebinSong:
closes #1988 

Proof of the fix working:

Figuring out the fix for the problem 3 actually required some ground-breaking change to how `MessageBase.vue` currently renders the message content(details in [here](https://github.com/okTurtles/group-income/issues/1988#issuecomment-2113659964)), as the current way is not compatible with when the text content contains markdown in it.

So I had to create another component(`RenderMessageWithMarkdown.js`) which uses Vue's[ render function](https://v2.vuejs.org/v2/guide/render-function).

The way it works in summary is,

**1. it firstly convert the markdown into the raw html string** 
(e.g.) `- asdfasdf sadfasdf _asdf_ asdfasdf`) -> `<ul><li>asdfasdf sadfasdf <em>asdf</em> asdfasdf</il></ul>`

**2. it turns this raw html into a recursive DOM object tree structure, which looks like below example.**
(e.g.) `<ul><li>asdfasdf sadfasdf <em>asdf</em> asdfasdf</il></ul>` ->

```js
{
  tagName: 'ul',
  children: [
    {
       tagName: 'li',
       children: [
          { tagName: null, text: 'asdfasdf sadfasdf' },
          { tagName: 'em', text: 'asdf' },
          { tagName: null,  text: 'asdfasdf' }
       ]
    }
  ]
}
```
**3. `RenderMessageWithMarkdown.js` renders this object tree as the equivalent recursive structure of `createElement` call.**
(e.g.)
```js
createElement('ul',
  { ... },
  [
    createElement('li', 
      [ createElement(...), ... ]
    )
  ]
)
```

This way, we can properly handle the complex html DOM structure produced as a result of converting the markdown string. Hope it looks good.

cc. @taoeffect @Silver-IT 

---

There is a minor issue where if a mention is the only child of any formatted styles like below, the mention block is treated as a pure text, which is incorrect.

This is actually related to the **[lookAhead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookahead_assertion)/lookBehind assertions** used in below regular expression and needs an in-depth investigation. So I will handle this issue as a follow-up work of this PR.
```js
new RegExp(`(?<=\\s|^)(${allMention}|${this.possibleMentions.join('|')})(?=[^\\w\\d]|$)`)
```


frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
```diff
@@ -0,0 +1,236 @@
+<template lang='pug'>
```
SebinSong, line 1: some of the existing mention related logics(e.g. splitting the text by possible mentions, using `ProfileCard.vue` in certain cases etc.) are DRYed into this component now, as this needs to be used in multiple places.

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -18,57 +18,38 @@
...
-              i(:class='"icon-" + objText.icon')
-              span {{ objText.text }}
+        template(v-if='replyingMessage')
+          render-message-with-markdown(
+            v-if='shouldRenderMarkdown'
+            :isReplyingMessage='true'
+            :text='replyingMessage'
+            @click='$emit("reply-message-clicked")'
+          )
+          render-message-text.c-replying(
```
Silver-IT, line 28: I think we can simplify these codes using `:is` attribute just like how we do in `ChatMain.vue`.
```
...

component(
  :is='messageType(message)'

...

computed () {
  messageType (message) {
    return {
      [MESSAGE_TYPES.NOTIFICATION]: 'message-notification',
      [MESSAGE_TYPES.INTERACTIVE]: 'message-interactive',
      [MESSAGE_TYPES.TEXT]: 'message',
      [MESSAGE_TYPES.POLL]: 'message-poll'
    }[message.type]
  }

...

```

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -18,57 +18,38 @@
...
+        template(v-else-if='text')
+          render-message-with-markdown.c-text(
+            v-if='shouldRenderMarkdown'
+            :text='text'
+            :edited='edited'
+          )
+          render-message-text.c-text(
+            v-else
+            :text='text'
+            :edited='edited'
```
Silver-IT, line 51: And this too.

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -18,57 +18,38 @@
...
+            :isReplyingMessage='true'
+            :text='replyingMessage'
+            @click='$emit("reply-message-clicked")'
+          )
+          render-message-text.c-replying(
+            v-else
+            :isReplyingMessage='true'
+            :text='replyingMessage'
+            @click='$emit("reply-message-clicked")'
+          )
```
Silver-IT, line 33: I think we can simplify this code using `:is` attribute just like how we do in `ChatMain.vue`.

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -18,57 +18,38 @@
...
+          render-message-with-markdown.c-text(
+            v-if='shouldRenderMarkdown'
+            :text='text'
+            :edited='edited'
+          )
+          render-message-text.c-text(
+            v-else
+            :text='text'
+            :edited='edited'
+          )
```
Silver-IT, line 52: And this too.

frontend/views/pages/GroupChat.vue
```diff
@@ -117,6 +117,7 @@ import ChatMembers from '@containers/chatroom/ChatMembers.vue'
 import { OPEN_MODAL } from '@utils/events.js'
 import { MenuParent, MenuTrigger, MenuContent, MenuItem, MenuHeader } from '@components/menu/index.js'
 import { CHATROOM_PRIVACY_LEVEL } from '@model/contracts/shared/constants.js'
+import '../containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js'
```
Silver-IT, line 120: Is this importing necessary?

Silver-IT (changes requested):
@SebinSong, very nice work! Just left a few change requests.

And I have found one issue in editing the text. I tried editing [the text of Problem 3](https://github.com/okTurtles/group-income/issues/1988#:~:text=%5BProblem%203.%5D%20Problem%20with%20mentions%20inside%20the%20bullets), and got the following result.
image
I think the `(edited)` indicator should be in the same line.

frontend/views/pages/GroupChat.vue
Silver-IT: Is this importing necessary?

frontend/views/containers/chatroom/MessageBase.vue
SebinSong: @Silver-IT Good point, but let me modify the logic that way once the number of options becomes `>2` in the future. Currently only two components are used here and I feel like using `v-if/v-else` pair would just suffice.  Plus, I think it's more readable.

frontend/views/pages/GroupChat.vue
SebinSong: Yes. as far as I know the component written with render function cannot be exported as SFC, so the registration of the component (executing `Vue.component(...)`) has to be executed somewhere in the app. This could've been done anywhere (e.g. `main.js`) but this component is only relevant in the chat area of the app. So I exported the file here instead.

SebinSong:
@Silver-IT 

> I think the (edited) indicator should be in the same line.

I think that's because the content of your particular message ends with a block element (`<ul> ... <li>...<li> </ul>`). If it does end with an inline element, the `(edited)` text will look inline too like below. it actually kind of feels more noticeable to me that way with the list content above.



The markdown conversion outcome is sometimes relatively complex structure like a multi-nested DOM structures. In this case, Somehow injecting that text to the end of last text content would overcomplicate the component. So I would suggest we treat that as a non-critical thing and leave it as is.

frontend/views/pages/GroupChat.vue
SebinSong: I actually changed `RenderMessageWithMarkdown` as an exportable object and have registered it as a child component of `MessageBase.vue` in the latest update.

Silver-IT:
> So I would suggest we treat that as a non-critical thing and leave it as is.

Okay, then it's fine with me either.

frontend/views/containers/chatroom/MessageBase.vue
Silver-IT: Even though, we use two components, I personally prefer to use `:is` attribute because we can use it in two places; one for original message and one for replying message. Of course, it's not a big issue, and just shared my opinion.

I am going to approve this PR, since nothing wrongs I found. But I would like to keep this conversation, and to know how @taoeffect thinks about it.

Silver-IT (dismissed):
Excellent work, @SebinSong!

frontend/views/containers/chatroom/MessageBase.vue
SebinSong: > Of course, it's not a big issue, and just shared my opinion.

Yeah, I think it's not a big of a deal now either way and I still like to keep it as `v-if/v-else`, as they look more straightforward to read. But if you really feel like, you can go ahead and create an issue for it and work on adding `<component :is='...'>` as a follow-up work of this PR.

taoeffect: I think it makes sense to use `:is=` when there is a `type` variable that can be used to key an object (as shown in the code example @Silver-IT put). Since in this example there are only 2 components, and because there doesn't seem to be a `type` available to key the components, I think it's fine the way it is. But if a `type` is added somewhere as an attribute that can be used to key an object map, then sure `:is` would also make sense.

taoeffect (changes requested):
Nice work @SebinSong!

Here's a test:

---------------------

@u2 hey 

- THis is 
   - a @u1 to @u2 @u1 message from the great world

```
@how about this @u1 

- dsf
```

1. @u1 and @u2 
2. @u2 @u1 **@u1**

     ```
     asdfsadf
     ```

---------------------

Here's what that looks like in this PR:

Screenshot 2024-05-27 at 8 15 41 PM

You can see what it looks like here when I click edit on the message:

Screenshot 2024-05-27 at 8 15 36 PM

Probably not right, right?

SebinSong:
@taoeffect 
Just updated the PR again with additional fixes. Below is the proof of the fix working.

frontend/views/utils/markdown-utils.js
```diff
@@ -14,24 +19,21 @@ marked.use({
...
-  // remove unecessary line-breaks from the converted markdown outcome.
-  converted = converted.replace(/^\s+|\s+$/g, '')
+  // STEP 1. Handle multiple line-breaks
+  // markedjs with the gfm(Github Flavored Markdown) style always collapses multiple line-breaks into one
+  // so we need some custom logic to handle it manually.
+  // (Reference issue here: https://github.com/markedjs/marked/issues/190)
+  str = str.replace(/\n(?=\n)/g, '\n<br>')
+    .replace(/<br>\n(\s*)(\d+\.|-|```)/g, '\n\n$1$2')
+    .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n')
+  console.log('!@# str converted: ', str)
```
taoeffect: Don't forget to remove extraneous console.log debug statements :)

taoeffect (changes requested):
Beautiful @SebinSong!

Just one extraneous `console.log` debug statement that needs to be removed before we can merge :)

frontend/views/utils/markdown-utils.js
SebinSong: done

taoeffect (approved):
Amazing work @SebinSong!

This PR and your solution was really impressive!! 💪

U1:
I can confirm that I get notified.  🙃   

    (by github, mentioning my username @u1 above)

PR #2023, opened 2024-05-27
#2012 - Add "Copy message text" menu to the chat message toast pop-up

### LINKED ISSUE #2012: Add "Copy message text" to press & hold chat menu on mobile
### Problem

There is no way to copy the text someone wrote in a chatroom at the moment.

When you press & hold it doesn't highlight text, but brings up a menu instead, and in that menu there's no way to copy the text.

### Solution

Add a menu item called "Copy message text" to copy the contents of the message.

SebinSong:
closes #2012 

Adding 'Copy message text' menu to the chat message's toast menu.

### DIFF
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -100,6 +100,8 @@
     v-if='!isEditing'
     :variant='variant'
     :type='type'
+    :text='text'
+    :messageHash='messageHash'
     :isMsgSender='isMsgSender'
     :isGroupCreator='isGroupCreator'
     ref='messageAction'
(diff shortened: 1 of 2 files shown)


frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -236,14 +237,6 @@ export default ({
     reply () {
       this.$emit('reply')
     },
-    copyMessageLink () {
```
SebinSong, line 197: This existing 'copy' action can rather be handled inside `MessageActions.vue` directly (it's not necessary to keep/execute the logic in the parent component, which requires emitting a custom-event from the child). So I thought I would DRY this logic while at this task.

taoeffect (approved):
Seems to work well!

PR #2025, opened 2024-05-29
Added idle check when make sounds

### LINKED ISSUE #2022: Prevent DM "ding" notification sound when DM is open, active and in the foreground
### Problem

It's annoying to constantly hear a sound when someone you're chatting with in an open DM messages you.

### Solution

Don't play the sound if the DM is open and in the foreground AND the user has recently interacted with the UI.

It's OK to play the sound if the user hasn't moved the mouse in a while and hasn't pressed any keys either (say, for at least 2 minutes).

Silver-IT:
### Summary of Changes

- Closes #2022

### DIFF
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -535,7 +535,7 @@ const getters = {
   isAppIdle (state) {
     // NOTE: idleVue.isIdle is created by the IdleVue plugin
     //       https://www.npmjs.com/package/idle-vue
-    return state.idleVue.isIdle
+    return Boolean(state.idleVue?.isIdle)
   }
 }
 
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -531,11 +531,6 @@ const getters = {
         const nameB = getters.ourContactProfilesByUsername[usernameB].displayName || usernameB
         return nameA.normalize().toUpperCase() > nameB.normalize().toUpperCase() ? 1 : -1
       })
-  },
-  isAppIdle (state) {
-    // NOTE: idleVue.isIdle is created by the IdleVue plugin
-    //       https://www.npmjs.com/package/idle-vue
-    return Boolean(state.idleVue?.isIdle)
   }
 }
 
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -531,6 +531,9 @@ const getters = {
         const nameB = getters.ourContactProfilesByUsername[usernameB].displayName || usernameB
         return nameA.normalize().toUpperCase() > nameB.normalize().toUpperCase() ? 1 : -1
       })
+  },
+  isAppIdle (state) {
+    return state.idleVue.isIdle
   }
 }
 
--- a/frontend/model/state.js
+++ b/frontend/model/state.js
@@ -533,6 +533,8 @@ const getters = {
       })
   },
   isAppIdle (state) {
+    // NOTE: idleVue.isIdle is created by the IdleVue plugin
+    //       https://www.npmjs.com/package/idle-vue
     return state.idleVue.isIdle
   }
 }
--- a/frontend/views/components/sounds/Background.vue
+++ b/frontend/views/components/sounds/Background.vue
@@ -15,13 +15,16 @@ export default ({
     sbp('okTurtles.events/on', MESSAGE_SEND, this.playMessageSend)
   },
   methods: {
+    shouldPlay () {
+      return document.hidden || this.isAppIdle
+    },
     playMessageReceive () {
-      if (this.isAppIdle) {
+      if (this.shouldPlay()) {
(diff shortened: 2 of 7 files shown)


Silver-IT (commented):
@taoeffect, review ready!

frontend/model/state.js
```diff
@@ -531,6 +531,11 @@ const getters = {
         const nameB = getters.ourContactProfilesByUsername[usernameB].displayName || usernameB
         return nameA.normalize().toUpperCase() > nameB.normalize().toUpperCase() ? 1 : -1
       })
+  },
+  isAppIdle (state) {
+    // NOTE: idleVue.isIdle is created by the IdleVue plugin
+    //       https://www.npmjs.com/package/idle-vue
+    return Boolean(state.idleVue?.isIdle)
```
taoeffect: Based on the documentation from the plugin [here](https://github.com/soixantecircuits/idle-vue/tree/master?tab=readme-ov-file#example---mainjs-1), it indicates that `isAppIdle` is already accessible on all vue objects.

In other words, you don't need to create this getter, and you don't need to import it in `Background.vue`, you should be able to directly access `this.isAppIdle`.

Also, as it appears to be a boolean variable already, there's no need to wrap it in `Boolean`.

frontend/views/components/sounds/Background.vue
```diff
@@ -6,20 +6,28 @@
...
-      this.$refs.msgReceive.play()
+      if (this.isAppIdle) {
+        this.$refs.msgReceive.play()
+      }
     },
     playMessageSend () {
-      this.$refs.msgSend.play()
+      if (this.isAppIdle) {
+        this.$refs.msgSend.play()
+      }
```
taoeffect: Note that per the original issue it's OK to play the sound if the app is in the background and the user has recently moved the mouse.

So it might be worth using `idle-js` directly instead of this wrapper because that gives you access to `onHide` and `onShow`.

taoeffect (changes requested):
Nice! Review ready!

frontend/model/state.js
Silver-IT: Yeah, right. `this.isAppIdle` works fine as they described.
I thought it won't work because I couldn't find `isAppIdle` in the vuex state(`sbp('state/vuex/state')`).

Nice catch!

frontend/views/components/sounds/Background.vue
Silver-IT: Using `document.hidden`, I updated PR to make sound when the app is in background.

taoeffect (changes requested):
Nicely done @Silver-IT! Cool library you found! 😎 

So I tested this PR using Firefox Developer Edition plus a private Brave window.

1. [FFDev] u1 creates group and copies invite link, remains on Dashboard
2. [Brave] u2 joins group and switches to Chat, sends message: `@u1 hello!`

Problems:

1. There is no red indicator in u1's navbar that should tell them that they have an unread mention in the chat (it does appear though for DMs)
2. There is no sound notification (note: see below)

**IF** I hide the Firefox window, or switch to a different virtual desktop when sending this message, the sound does appear.

Also, If I send a DM from Brave to Firefox, and Firefox is on the Dashboard, there is no sound, but there definitely should be. Likewise, if I'm on the dashboard, and someone @mentions me in the chatroom, I should probably hear a sound, right?

So it sounds like we need a bit more logic here, because the `idle-vue` plugin by itself isn't enough.

Also, the red dot should appear on this nav menu item when there is a regular chatroom mention:

Screenshot 2024-05-30 at 9 46 58 AM

Currently, it *does* appear correctly on the browser tab:

Screenshot 2024-05-30 at 9 47 32 AM

Silver-IT:
> There is no red indicator in u1's navbar that should tell them that they have an unread mention in the chat (it does appear though for DMs)

This is because there is a mistake in using variable name. I have already fixed it in PR #1998.
image

> There is no sound notification (note: see below)

I guess that's because the current window is not `hidden` (it is active) and 2 mins are not passed yet from the last activity.

> Also, If I send a DM from Brave to Firefox, and Firefox is on the Dashboard, there is no sound, but there definitely should be.

I guess also it's because 2 mins are not passed from the last activity. Meaning `this.isAppIdle = false`.

> Likewise, if I'm on the dashboard, and someone @mentions me in the chatroom, I should probably hear a sound, right?

Yes. Will hear sound when there is no activity for at least 2 mins. Or when the window is hidden.

Silver-IT:
Plus, I have created `isAppIdle` computed variable in `Background.vue` again because sometimes `this.isAppIdle` (which `idle-vue` plugin provides) returns `undefined` even though the `this.$store.state.idleVue.isIdle` has the correct value.

taoeffect (approved):
Nicely done!

PR #2027, opened 2024-05-30
Update KV store set

corrideat:
This updates the `chelonia/kv/set` function to automatically retry in case there's a conflict. With this, you can remove the `requireToSuccess` function and calls to it, with no further changes required for having the same functionality (well, other than that this change by default retries thrice as opposed to ten times).

This PR introduces two new options to `chelonia/kv/set`, `maxAttempts` and `onconflict`. You can use `maxAttempts` to control the number of attempts made in case there's a conflict.

Example:

`sbp('chelonia/kv/set', contractID, key, value, { ...keys, maxAttempts: 10 })`

`onconflict`, on the other hand, is a callback function (`(contractID, key, data) => Promise<Object>`) used to replace the data that's written to the KV in case there's a conflict. You can (and should!) use this to manually resolve conflicts that may arise. If you don't specify it, the orignal `data` value will be used (thus overwriting the KV).

Example:

```
let value = 0
sbp('chelonia/kv/set', contractID, key, value, { ...keys, onconflict: async (_, __, value) => { return ++value } })
```

That'll write to `key` with the number of attempts made.

If, as part of conflict resolution, you need to use the currently stored value in the KV, you need to fetch it yourself.

### DIFF
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -4,7 +4,7 @@ import '@sbp/okturtles.eventqueue'
 import '@sbp/okturtles.events'
 import sbp from '@sbp/sbp'
 import { handleFetchResult } from '~/frontend/controller/utils/misc.js'
-import { cloneDeep, difference, has, intersection, merge, randomHexString } from '~/frontend/model/contracts/shared/giLodash.js'
+import { cloneDeep, delay, difference, has, intersection, merge, randomHexString, randomIntFromRange } from '~/frontend/model/contracts/shared/giLodash.js'
 import { NOTIFICATION_TYPE, createClient } from '~/shared/pubsub.js'
 import type { GIKey, GIOpActionUnencrypted, GIOpContract, GIOpKeyAdd, GIOpKeyDel, GIOpKeyRequest, GIOpKeyRequestSeen, GIOpKeyShare, GIOpKeyUpdate } from './GIMessage.js'
 import type { Key } from './crypto.js'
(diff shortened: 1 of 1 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -1377,30 +1377,48 @@ export default (sbp('sbp/selectors/register', {
...
+        method: 'POST',
+        body: JSON.stringify(serializedData),
+        signal: this.abortController.signal
+      })
+      if (!response.ok) {
+        if (response.status === 409) {
+          if (--maxAttempts <= 0) {
+            throw new Error('Conflict setting KV value')
+          }
+          await new Promise(resolve => setTimeout(resolve, randomIntFromRange(0, 1500)))
```
taoeffect: Since we're using `randomIntFromRange` here from `giLodash`, we might as well import `delay` instead to shorten this code (as it's exactly for this purpose)

shared/domains/chelonia/chelonia.js
```diff
@@ -1377,30 +1377,48 @@ export default (sbp('sbp/selectors/register', {
...
+          if (typeof onconflict === 'function') {
+            data = await onconflict(contractID, key, data)
+          }
+          continue
+        }
+        throw new Error('Invalid response status: ' + response.status)
```
taoeffect: Can this say: `'kv/set invalid response status: ' + ...`?

taoeffect (approved):
Nicely done, thanks @corrideat!

@Silver-IT This PR means we can remove the `requireToSuccess` function because it implements that functionality inside of `'chelonia/kv/set'`.

Importantly, it also provides a new `onconflict` hook that can be used to handle conflict resolution. If you know that a conflict might mean modifying the old value (instead of overwriting it), then you can pass in the `onconflict` handler and call `'chelonia/kv/get'` inside of it to fetch the current value that's in the KV store, and then modify it accordingly.

PR #2036, opened 2024-06-04
#2020 - UI bug related to auto-scroll on 'Focus' in mobile browsers.

### LINKED ISSUE #2020: UI bug when creating group in PWA
### Problem

After already being part of an existing group, when I went to create another group in the PWA on mobile (by clicking on the "+ Add a group" button in the left navbar), the modal appears cut off and poorly positioned:

tmp-1716602937777

### Solution

Fix.

SebinSong:
closes #2020 

---

@taoeffect 

This issue actually happens in all mobile browsers, which is where the native keyboard panel is pulled out when an input field focuses.

What I noticed is that the same doesn't happen in the following group creation steps despite them all calling `inputEl.focus()` in their `mounted()` hook too. 
The difference in the first step is, The native keyboard is pulled out [1] almost at the same time as when the Modal pops out [2] and this requires browser to handle the layout change (by [2]) while the viewport size is decreasing (by [1]) and it somehow leads to this overzealous behaviour.

The best solution I could come up with was to giving a bit of delay to calling `focus()` on the input element, so that browser can fully render modal first and then deals with the viewport change caused by the focus. Below is the improvement I see.

**[Before fix]**

**[After fix]**

### DIFF
--- a/frontend/views/components/group-creation-steps/GroupName.vue
+++ b/frontend/views/components/group-creation-steps/GroupName.vue
@@ -62,6 +62,8 @@ export default ({
   },
   mounted () {
     window.setTimeout(() => {
+      // An arbitrary delay here is to fix the issue #2020.
+      // See the PR description here for the details: https://github.com/okTurtles/group-income/pull/2036#issue-2332505747
       this.$refs.name.focus()
     }, 300)
 
--- a/frontend/views/components/group-creation-steps/GroupName.vue
+++ b/frontend/views/components/group-creation-steps/GroupName.vue
@@ -61,7 +61,10 @@ export default ({
     sbp('okTurtles.events/on', AVATAR_EDITED, this.updateGroupPictureByEditor)
   },
   mounted () {
-    this.$refs.name.focus()
+    window.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


frontend/views/components/group-creation-steps/GroupName.vue
```diff
@@ -61,7 +61,10 @@ export default ({
     sbp('okTurtles.events/on', AVATAR_EDITED, this.updateGroupPictureByEditor)
   },
   mounted () {
-    this.$refs.name.focus()
+    window.setTimeout(() => {
+      this.$refs.name.focus()
+    }, 300)
+
```
taoeffect: In general, we want to add code comments to code that someone might find strange, odd, confusing, or unclear. This is a great example, as someone reading this would have absolutely no idea why this delay is here, and the answer is quite unintuitive. So my only request for this PR would be to add a comment referencing the issue this is closing and explaining why this is here.

taoeffect (commented):
Nicely done @SebinSong! Just a request for comment

taoeffect (approved):
Excellent!

PR #2050, opened 2024-06-06
Cancel message editing by pressing ESC

### LINKED ISSUE #2047: Implement pressing "Escape" to cancel editing a message
### Problem

When editing a message, you cannot press the escape key to cancel editing the message. This is annoying.

### Solution

Make it so that when the user presses the escape key it's the same as the "Cancel" button being pressed.

Silver-IT:
### Summary of Changes

- Closes #2047


Silver-IT (commented):
Ready for review!

taoeffect (approved):
Great work @Silver-IT !!

PR #2055, opened 2024-06-10
#2051 - Improve the UI for already joined group

### LINKED ISSUE #2051: Improve "Cannot join already joined group" UX
### Problem

If user is logged in and clicks an invite link for a group they're already a member of, they're shown a full-screen error message.  

1. The large text indicates that the link itself is invalid (which is confusing because the link _is_ valid) and the smaller text that explains the actual issue is not very user-friendly.
2. The included `href` is neither clickable nor human-readable (to the average user). [In this instance, copy-pasting the URL also does not "report the error" but rather redirects to the group dashboard.]

alreadyJoined

### Solution

1. Possible solutions include:
    1. When user is already a member, change "This invite is not valid" to "You are already a member". If we're showing the full error text on the modal, then the error message more user-friendly by showing the group name instead of '${groupContractID}`. (I assume if we know user is already a group member, then the group name is accessible.)
    2. Alternatively, automatically redirect the user to that group's dashboard. Optional: include alert indicating user was already a group member
 2. If included, links in the error message should be clickable, with the anchor text solving the "human-readable" problem.

SebinSong:
closes #2051

adding below screen to `Join.vue` page component.
(NOTE: 'A Group' here is the group name btw)

---

@taoeffect, @dotmacro  let me know if you would like to change the copies here.

### DIFF
--- a/frontend/views/pages/Join.vue
+++ b/frontend/views/pages/Join.vue
@@ -33,9 +33,9 @@ div
 
     .c-joined(v-else-if='isStatus("JOINED")')
       svg-create-group.c-svg
-      i18n.is-title-1(tag='h1' data-test='pageTitle') You are already a member.
-      i18n.has-text-1(tag='p' data-test='helperText') Cannot join already joined group.
-      i18n.c-goHome(tag='button' @click='goToDashboard') Go to dashboard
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Well done! 👏

PR #2063, opened 2024-06-12
Fix Cypress failures

### LINKED ISSUE #2058: Broken Cypress test related to login modal (heisenbug)
### Problem

https://cloud.cypress.io/projects/q6whky/runs/2370/test-results/0417b59b-4b92-4529-95bc-8228fffc0f88?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github

Screenshot 2024-06-11 at 10 03 13 AM

From: https://cloud.cypress.io/projects/q6whky/runs/2370/test-results/0417b59b-4b92-4529-95bc-8228fffc0f88/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github

### Solution

It seems like maybe the Cypress tests are written incorrectly, or there's some bug in the vuejs code that causes the modal to not really be closed.

### LINKED ISSUE #2059: Broken Cypress test related to ChatMain (heisenbug)
### Problem

Screenshot 2024-06-11 at 10 39 03 AM

https://cloud.cypress.io/projects/q6whky/runs/2371/test-results/876ef43a-5199-42fb-8c62-6119b97a295b?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github

### Solution

Fix.

Silver-IT:
### Summary of Changes

- Closes #2058
- Closes #2059


frontend/views/utils/misc.js
```diff
@@ -23,3 +24,15 @@ export const showNavMixin = {
...
+  computed: {
+    isInCypress (): boolean {
+      // NOTE: By only checking the window.Cypress, it's difficult to say that the app is running in Cypress
+      //       because anyone can access the window object.
+      //       in Cypress mode, sbp is exposed to window object (see main.js) so window.sbp can be compared
+      //       with our sbp object to verify if the app is in Cypress mode in real
+      return !!window.Cypress && window.sbp === sbp
+    }
+  }
+}
```
Silver-IT: Since anyone can access `window` object and set value in console, I think `!!window.Cypress` is not enough.
```
window.Cypress = true
```

Silver-IT (commented):
Ready for the review!

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -946,8 +948,22 @@ export default ({
...
+            //       This normally happens in Cypress, when user logs out just after sending a message.
+            const curMessages = newContractState.messages || []
+            if (addedOrDeleted === 'ADDED' && curMessages.length) {
+              const lastMessage = curMessages[curMessages.length - 1]
+              this.updateReadUntilMessageHash({
+                messageHash: lastMessage.hash,
+                createdHeight: lastMessage.height
+              })
+            }
+          }
```
taoeffect: Even with the comment, it's not clear to me why this code is here.

- What issue is this fixing, and which Cypress test would be failing without it?
- Is it possible to fix this in some other way, perhaps by modifying the Cypress tests themselves and having them wait on something?

taoeffect (commented):
Thanks @Silver-IT! I left a question

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: > What issue is this fixing, and which Cypress test would be failing without it?

This was a very picky situation which kept making Cypress to be failed. Let me explain.
Once we receive the message event(e.g., `'gi.contracts/chatroom/addMessage'`), we process the event in this function and the `this.messageState` is updated. So the computed value `this.messages` would be updated.
After that, we save the `readUntil` message to the KV store by calling `this.updateReadUntilMessageHash` function.

Now, let's think about Cypress test scenario to logout as soon as to send a message. So the workflow is like the following.

- User sends a message
- Cypress waits for the message to be added (`pending` now)
- Cypress waits for the message status to be `sent` (This is when the event is received, processed and `this.messageState` is updated)
- (Cypress decides the `Sending a message` is finished, actually is.) Call cy.giLogout()

So the Cypress didn't wait for the `this.updateReadUntilMessageHash` to be called, and also for the invocations from the queue `UNREAD_MESSAGES_QUEUE`. The invocations from `UNREAD_MESSAGES_QUEUE` run the following sbp function, so it would break Cypress whenever the invocations were called after logout.
```
return sbp('chelonia/kv/set', ourIdentityContractId, KV_KEYS.UNREAD_MESSAGES, data, {
  encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', ourIdentityContractId, 'cek'),
  signingKeyId: sbp('chelonia/contract/currentKeyIdByName', ourIdentityContractId, 'csk'),
})
```
Adding `await sbp('okTurtles.eventQueue/queueEvent', UNREAD_MESSAGES_QUEUE, () => {})` in the `gi.actions/identity/logout` function, couldn't solve the problem because a new invocation could be added after we call the `gi.actions/identity/logout` function. When the user's actions are very quick, only in Cypress. Look at the following orders of function calls

- Update this.messageState in ChatMain.vue
- Call `gi.actions/identity/logout` function
- Call `await sbp('okTurtles.eventQueue/queueEvent', UNREAD_MESSAGES_QUEUE, () => {})`
- Call `this.updateReadUntilMessageHash` and add new invocation to `UNREAD_MESSAGES_QUEUE`

> Is it possible to fix this in some other way, perhaps by modifying the Cypress tests themselves and having them wait on something?

I did also tried to find another way to wait for this in Cypress, but couldn't find alternative solution which doesn't use `cy.wait()`.

Silver-IT: I have just had an inspiration that add exception handler in `fetchChatRoomUnreadMessages` and `saveChatRoomUnreadMessages` function.

**EDITED**: I just realised that the exception handler couldn't solve the problem too. Because it's necessary to wait for the `UNREAD_MESSAGES_QUEUE` invocations to be finished before user logs out, and those kinds of exceptions (which would be related to the identityContractID) should not be thrown. So far, I couldn't find the alternative solution.

frontend/controller/actions/identity.js
```diff
@@ -782,18 +782,22 @@ export default (sbp('sbp/selectors/register', {
   },
   'gi.actions/identity/fetchChatRoomUnreadMessages': async () => {
     const { ourIdentityContractId } = sbp('state/vuex/getters')
-    return (await sbp('chelonia/kv/get', ourIdentityContractId, KV_KEYS.UNREAD_MESSAGES))?.data || {}
+    if (ourIdentityContractId) {
+      return (await sbp('chelonia/kv/get', ourIdentityContractId, KV_KEYS.UNREAD_MESSAGES))?.data || {}
+    }
```
taoeffect: This action can now return either `undefined` or `{}` or `.data`.... did you mean to implement it this way and do all calling sites handle these different values correctly?

Silver-IT: Oh, yeah. I made mistake in here. Nice catch. Thanks

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: @taoeffect, I will re-write the test scenario and remove the picky cases.

taoeffect (approved):

PR #2073, opened 2024-06-14
Closes #2072 - capture tunnel logs

### LINKED ISSUE #2072: Capture dev tunnel logs on server
### Problem

When testing via `grunt dev --tunnel` a lot of things seem to go wrong and it's difficult on a phone to know what happened.

Although it's possible to capture logs via a USB cable in some situations, this is inconvenient and not always possible.

### Solution

Send the logs directly to the server and have the server log them whenever we're in development mode and the page is loaded over the tunnel.

taoeffect:
Captures phone logs to server console run accessing via `grunt dev --tunnel`.

Adds new `'appLogs/logServer'` selector that can be called manually if necessary before the log capture begins (when login is called). Generally speaking you won't need to call this manually and the logs will appear automatically on the backend, color coded.

### DIFF
--- a/frontend/model/captureLogs.js
+++ b/frontend/model/captureLogs.js
@@ -191,19 +191,17 @@ sbp('sbp/selectors/register', {
   'appLogs/save' () { getLogger()?.save() },
   'appLogs/pauseCapture': captureLogsPause,
   'appLogs/startCapture': captureLogsStart,
-  'appLogs/logServer': process.env.NODE_ENV !== 'development'
+  'appLogs/logServer': process.env.NODE_ENV !== 'development' || !window.location.href.startsWith('https://gi')
     ? noop
     : function (level, string) {
-      if (window.location.href.startsWith('https://gi')) {
-        fetch(`${sbp('okTurtles.data/get', 'API_URL')}/log`, {
-          method: 'POST',
-          headers: {
-            'Content-Type': 'application/json'
-          },
-          body: JSON.stringify({ level, value: string })
-        }).catch(e => {
-          originalConsole.error(`[captureLogs] '${e.message}' attempting to log [${level}] to server:`, string)
-        })
-      }
+      fetch(`${sbp('okTurtles.data/get', 'API_URL')}/log`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({ level, value: string })
+      }).catch(e => {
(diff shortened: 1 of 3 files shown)


frontend/model/captureLogs.js
```diff
@@ -189,5 +190,20 @@ sbp('sbp/selectors/register', {
   'appLogs/get' () { return getLogger()?.entries?.toArray() ?? [] },
   'appLogs/save' () { getLogger()?.save() },
   'appLogs/pauseCapture': captureLogsPause,
-  'appLogs/startCapture': captureLogsStart
+  'appLogs/startCapture': captureLogsStart,
+  'appLogs/logServer': process.env.NODE_ENV !== 'development' || !window.location.href.startsWith('https://gi')
+    ? noop
+    : function (level, stringifyMe) {
+      if (level === 'debug') return // comment out to send much more log info
+      const string = JSON.stringify(stringifyMe)
```
Silver-IT: I would like to change name from `string` to `value` since `string` is the name of type.

frontend/model/captureLogs.js
```diff
@@ -189,5 +190,20 @@ sbp('sbp/selectors/register', {
   'appLogs/get' () { return getLogger()?.entries?.toArray() ?? [] },
   'appLogs/save' () { getLogger()?.save() },
   'appLogs/pauseCapture': captureLogsPause,
-  'appLogs/startCapture': captureLogsStart
+  'appLogs/startCapture': captureLogsStart,
+  'appLogs/logServer': process.env.NODE_ENV !== 'development' || !window.location.href.startsWith('https://gi')
```
Silver-IT, line 196: What does `'https://gi'` mean?

Silver-IT (changes requested):
Nice work, @taoeffect! It works well.

frontend/model/captureLogs.js
SebinSong: Seems like `L:194` here is for specifically filtering the app run by `grunt dev --tunnel`. I noticed every external https URL created by it always starts with `https://gi...`.

SebinSong (dismissed):
LGTM

frontend/model/captureLogs.js
Silver-IT: Yeah, I remember. 👍

Silver-IT (approved):
LGTM! Nice work! 👍

Issue #2066, opened 2024-06-14
Testing site broken (private chatroom not syncing across devices)


taoeffect:
### Problem

1. While desktop computer was asleep, I created a private chatroom on my phone via the PWA
2. Posted a few messages to it
3. Woke up computer and did not see the chatroom listed. No errors in console. Refreshing page doesn't fix it! However, opening a private window and re-logging in there does show the private chatroom.

Phone PWA:

![tmp-1718337318871](https://github.com/okTurtles/group-income/assets/138706/8b6a3c2f-09dd-4626-82c7-90f2defe4d97)

Desktop:

![Screenshot 2024-06-13 at 8 54 54 PM](https://github.com/okTurtles/group-income/assets/138706/9b334861-7ff7-43ff-be35-bee4d4cb59b3)

### Solution

Find and fix the problem.

corrideat:
Are you able to reproduce this, or do you have a copy of the state or logs?

taoeffect:
Re-opening this as I'm not convinced it's fixed and making it depend on #2110 being closed.

Issue #2064, opened 2024-06-13
Testing website is broken (users not joining properly)


taoeffect:
### Problem

User appears to join group but their name isn't shown in the chat and their user doesn't appear in the user list.

![Screenshot 2024-06-13 at 11 15 44 AM](https://github.com/okTurtles/group-income/assets/138706/6be12168-aca9-4366-a4fb-1b83c7d2b9ae)

![Screenshot 2024-06-13 at 11 15 49 AM](https://github.com/okTurtles/group-income/assets/138706/dc44fc7c-dec6-433f-a2d8-7d9d4a522f6d)

**But - if page is refreshed, then it shows:**

![Screenshot 2024-06-13 at 11 20 57 AM](https://github.com/okTurtles/group-income/assets/138706/c4673eb4-bddf-4c6a-812c-0a3f508567ba)

And everything is fine.

### Solution

Find and fix the problem.

Database is attached.

- Username: `greg`
- Password: `asdfasdf`

[groupincome.testing-2024-06-13.db.zip](https://github.com/user-attachments/files/15827022/groupincome.testing-2024-06-13.db.zip)

taoeffect:
Note that one of the errors I'm seeing on the backend is:

```
[11:19:57.845] ERROR (77): GET /eventsAfter/undefined/0 contractID undefined doesn't exist!                                                                                                                                                                                                                                  
    err: {                                                                                                                                                                                                                                                                                                                   
      "type": "Error",                                                                                                                                                                                                                                                                                                       
      "message": "contractID undefined doesn't exist!",                                                                                                                                                                                                                                                                      
      "stack":                                                                                                                                                                                                                                                                                                               
          Error: contractID undefined doesn't exist!                                                                                                                                                                                                                                                                         
              at Object.backend/db/streamEntriesAfter (/home/unpriv/gi/backend/database.js:47:18)                                                                                                                                                                                                                            
              at /home/unpriv/gi/backend/routes.js:128:20                                                                                                                                                                                                                                                                    
              at exports.Manager.execute (/home/unpriv/gi/node_modules/@hapi/hapi/lib/toolkit.js:60:28)                                                                                                                                                                                                                      
              at Object.internals.handler (/home/unpriv/gi/node_modules/@hapi/hapi/lib/handler.js:46:20)                                                                                                                                                                                                                     
              at exports.execute (/home/unpriv/gi/node_modules/@hapi/hapi/lib/handler.js:31:20)                                                                                                                                                                                                                              
[...]

PR #2083, opened 2024-06-17
Various bugfixes, related to #2064 and #2066

corrideat:
* eventsAfter: Reject early if the contractID is not set
* Fix instances causing such unset contractIDs
* Fix group members list to include members with only a contractID
* Don't set an empty contract state when receiving OP_KEY_SHARE
* Files: laxer validation when removing to process individual files separately

### DIFF
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -761,23 +761,44 @@ export default (sbp('sbp/selectors/register', {
   }) => {
     const rootGetters = sbp('state/vuex/getters')
     const { identityContractID } = sbp('state/vuex/state').loggedIn
-    const { shouldDeleteFile, shouldDeleteToken } = option
+    const { shouldDeleteFile, shouldDeleteToken, throwIfMissingToken } = option
+    let deleteResult, toDelete
 
     if (shouldDeleteFile) {
       const credentials = Object.fromEntries(manifestCids.map(cid => {
+        // It could be that the file was already deleted, if we no longer have
+        // a delete token. In this case, omit those CIDs.
+        if (!throwIfMissingToken && shouldDeleteToken && !rootGetters.currentIdentityState.fileDeleteTokens[cid]) {
+          console.info('[gi.actions/identity/removeFiles] Skipping file as token is missing', cid)
+          return [cid, null]
+        };
         const credential = shouldDeleteToken
           ? { token: rootGetters.currentIdentityState.fileDeleteTokens[cid] }
           : { billableContractID: identityContractID }
         return [cid, credential]
       }))
-      await sbp('chelonia/fileDelete', manifestCids, credentials)
+      toDelete = !throwIfMissingToken ? manifestCids.filter((cid) => !!credentials[cid]) : manifestCids
+      deleteResult = await sbp('chelonia/fileDelete', toDelete, credentials)
+    } else {
+      toDelete = manifestCids
     }
 
     if (shouldDeleteToken) {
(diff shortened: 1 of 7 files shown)


frontend/controller/actions/identity.js
```diff
@@ -770,15 +771,23 @@ export default (sbp('sbp/selectors/register', {
...
+          manifestCids: manifestCids.filter((_, i) => {
+            return deleteResult[i].status === 'fulfilled'
+          })
+        }
       })
     }
+
+    if (deleteResult.some(r => r.status === 'rejected')) {
+      throw new Error('[gi.actions/identity/removeFiles] Some CIDs could not be deleted')
+    }
```
taoeffect: ~~Note that throwing in a sideEffect will show the user a very serious unrecoverable red error global banner ... are you sure you want to throw? Maybe log the error instead?~~

EDIT: nevermind, I see that there's a `.catch` on it below

frontend/model/contracts/chatroom.js
```diff
@@ -511,7 +511,7 @@ sbp('chelonia/defineContract', {
             shouldDeleteToken: me === data.messageSender
           }
           deleteEncryptedFiles(data.manifestCids, option).catch(e => {
-            console.error(`[gi.contracts/chatroom/deleteMessage/sideEffect] (${contractID}):`, e)
+            console.warn(`[gi.contracts/chatroom/deleteMessage/sideEffect] (${contractID}):`, e)
```
taoeffect: Is there a way to keep this as an error (as it's important for users to know if the file failed to delete), by modifying the code above to detect if a file has already been deleted (and not throwing if that's the case?)

frontend/controller/actions/identity.js
```diff
@@ -770,15 +771,25 @@ export default (sbp('sbp/selectors/register', {
...
+            ? manifestCids.filter((_, i) => {
+              return deleteResult[i].status === 'fulfilled'
+            })
+            : manifestCids
+        }
       })
     }
+
+    if (deleteResult?.some(r => r.status === 'rejected')) {
+      throw new Error('[gi.actions/identity/removeFiles] Some CIDs could not be deleted')
```
taoeffect: Should probably also log the ones that are missing

frontend/controller/actions/identity.js
```diff
@@ -761,24 +761,44 @@ export default (sbp('sbp/selectors/register', {
   }) => {
     const rootGetters = sbp('state/vuex/getters')
     const { identityContractID } = sbp('state/vuex/state').loggedIn
-    const { shouldDeleteFile, shouldDeleteToken } = option
+    const { shouldDeleteFile, shouldDeleteToken, maybeDeleted } = option
```
taoeffect: `maybeDeleted` => `throwIfMissingToken`

frontend/controller/actions/identity.js
```diff
@@ -761,24 +761,44 @@ export default (sbp('sbp/selectors/register', {
...
+    let deleteResult, toDelete
 
     if (shouldDeleteFile) {
       const credentials = Object.fromEntries(manifestCids.map(cid => {
+        // It could be that the file was already deleted, if we no longer have
+        // a delete token. In this case, omit those CIDs.
+        if (maybeDeleted && !rootGetters.currentIdentityState.fileDeleteTokens[cid]) {
+          return [cid, null]
```
taoeffect: Here it would need to log a warning

taoeffect (approved):

PR #2084, opened 2024-06-18
#2077 - Warn user about char limit in group description

### LINKED ISSUE #2077: Warn user about char limit in group description
### Problem

The group description clearly has a character limit, but there's no indication of it. There's no "x/n" or "x chars remaining" countdown, no "up to x chars" tip/tooltip on the screen, and no info/error alert to tell the user why typing isn't working. 

groupDescriptionCharLimit  

We do warn users about the char limit for usernames:

usernameCharLimit

### Solution

Indicate the user has reached the char limit the same way we do with usernames: an error with the text "A group description cannot exceed 500 characters."

SebinSong:
closes #2077 

Add below character length indication UI to both group-creation Modal and also to `/group-settings` page.

**-> Group-Creation step**

**-> `GroupSettings.vue`**

---
@taoeffect @dotmacro 

Normally, specifying maxlength HTML attribute to the `textarea` element will prevent user from entering more characters, just like **[this](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength#constraint_validation)** MDN doc says and how FireFox Dev edition behaves regarding it. But in case there is any browser out there where this is not how `maxlength` attribute behaves, I've added the in-app validation logic to both form so they show this error message when the description becomes longer than 500 characters.

Thanks,

### DIFF
--- a/frontend/model/contracts/shared/constants.js
+++ b/frontend/model/contracts/shared/constants.js
@@ -13,6 +13,9 @@ export const PROFILE_STATUS = {
   PENDING: 'pending', // shortly after being approved to join the group
   REMOVED: 'removed'
 }
+export const GROUP_DESCRIPTION_MAX_CHAR = 500
+
+// group-proposal related
 
 export const PROPOSAL_RESULT = 'proposal-result'
 export const PROPOSAL_INVITE_MEMBER = 'invite-member'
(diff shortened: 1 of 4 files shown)


taoeffect (approved):
Amazing work @SebinSong! Beautifully done!

Issue #2070, opened 2024-06-14
Replace `alert` in login that use LError with `gi.ui/prompt`


taoeffect:
### Problem

In #2069 I got this error:

![tmp-1718390238728](https://github.com/okTurtles/group-income/assets/138706/212eeb34-f95c-4179-b25f-8f13df428940)

This is clearly a problematic prompt as it shows HTML to the user instead of rendering it:

```js
      console.error('gi.actions/identity/login failed!', e)
      const humanErr = L('Failed to login: {reportError}', LError(e))
      alert(humanErr)
```

EDIT: I just noticed this is a more widespread problem in the codebase, as this use of `alert` seems to exist in the contracts too:

![tmp-1718396956352](https://github.com/okTurtles/group-income/assets/138706/01a43304-58c4-43a7-8f7e-29bef4ab1452)

**IMPORTANT: the use of `alert` by itself isn't a problem, but when the message contains HTML - `gi.ui/prompt` must be used instead because only that can render it.**

### Solution

1. Replace the call to `alert` with `gi.ui/prompt`
2. Test by throwing an error during login and verify that the link is rendered correctly

**Make 100% sure to test every `alert` that's replaced with `gi.ui/prompt` to verify it displays correctly.**

If called from contracts, `gi.ui/prompt` will need to be whitelisted in `main.js` under `allowedSelectors`.

dotmacro:
> I just noticed this is a more widespread problem in the codebase

Is the screenshot in #2051 another example?

taoeffect:
> Is the screenshot in https://github.com/okTurtles/group-income/issues/2051 another example?

Yeah that could be another thing to boy-scout for this issue 👍

PR #2091, opened 2024-06-18
#2070 replace `Alert()` with the in-app prompt dialog

### LINKED ISSUE #2096: Poor handling of error when waiting to join group
### Problem

If for some reason an exception is thrown inside of `gi.actions/group/join`, and this handler gets called:

```js
    } catch (e) {
      console.error('gi.actions/group/join failed!', e)
      throw new GIErrorUIRuntimeError(L('Failed to join the group: {codeError}', { codeError: e.message }))
```

Then the user is still stuck on the `PendingApproval.vue` page, and there is no indication whatsoever that an error occurred. This is a problem because the correct action that the user needs to take is to refresh the page.

I encountered this while testing `grunt dev --tunnel` and an error `502 Bad Gateway` was thrown by the tunnel. This is the error that appeared in the console on the mobile device:

```
[error] ["[gi.actions/group/join] Error while sending key request for z9brRu3VUtFRxWvL1uuMkc9q89999VC5B1PnXMf7KdKKs46E1BZL:",{"name":"SyntaxError","message":"Unexpected token 'B', \"Bad Gateway\" is not valid JSON","stack":"SyntaxError: Unexpected token 'B', \"Bad Gateway\" is not valid JSON"}]
[error] ["[gi.actions/group/join] Error during join (inside CONTRACT_HAS_RECEIVED_KEYS event handler)",{"name":"GIErrorUIRuntimeError","message":"Failed to join the group: Unexpected token 'B', \"Bad Gateway\" is not valid JSON","stack":"GIErrorUIRuntimeError: Failed to join the group: Unexpected token 'B', \"Bad Gateway\" is not valid JSON\n    at gi.actions/group/join (https://gi2131.loca.lt/assets/js/main.js:4013:13)"}]
[error] ["gi.actions/group/join failed!",{"name":"SyntaxError","message":"Unexpected token 'B', \"Bad Gateway\" is not valid JSON","stack":"SyntaxError: Unexpected token 'B', \"Bad Gateway\" is not valid JSON"}]
[error] ["[chelonia] failed to sync z9brRu3VTGqRArbqGu4XqZ4TkJVrXn1e8ckH4etgvwe6cyg3Nvz3:",{"name":"Error","message":"502: ","stack":"Error: 502: \n    at https://gi2131.loca.lt/assets/js/chunk-PF6LA6BE-cached.js:622:13\n    at async chelonia/private/in/syncContract (https://gi2131.loca.lt/assets/js/chunk-PF6LA6BE-cached.js:1613:34)\n    at async okTurtles.eventQueue/queueEvent (https://gi2131.loca.lt/assets/js/chunk-PF6LA6BE-cached.js:192:20)\n    at async Promise.all (index 0)"}]
```

### Solution

If an exception is caught here, we need to let the user know on the `PendingApproval.vue` page that an error occurred (and what the `e.message` was). And also display a button to refresh the page, similar to how we have the `Retry` button in the `ErrorPage.vue` page.

SebinSong:
work on #2070
closes #2096

Below are the screenshots of some places where I repalced the `alert()` instances.

prompt-task-1
prompt-task-2
prompt-task-3
prompt-task-4
prompt-tesk-5

---
@taoeffect 
The `report the error` link tag currently doesn't navigate to `Application log` page and goes to the `/dashboard` instead. This can be investigated/worked on separately from this issue later on.


taoeffect (commented):
> The report the error link tag currently doesn't navigate to Application log page and goes to the /dashboard instead. This can be investigated/worked on separately from this issue later on.

Why not fix it in this PR? FYI, similar problems relating to that link were fixed in #2005 that was just merged

SebinSong:
@taoeffect 
> FYI, similar problems relating to that link were fixed in https://github.com/okTurtles/group-income/pull/2005 that was just merged

I've tried it again after merging this but it doesn't fix the issue either.

> Why not fix it in this PR? 

I actually already spent a few hours investigating this problem and the root cause is that, [this block](https://github.com/okTurtles/group-income/blob/8e032da1c2862595ba9ad1be2cd5ccdaf2a02424/frontend/controller/router.js#L182C1-L189C4) of navigation guard logic triggers before [this block](https://github.com/okTurtles/group-income/blob/8e032da1c2862595ba9ad1be2cd5ccdaf2a02424/frontend/main.js#L404C7-L409C11) of silent login in `main.js` is complete. 
To give more details, that navigation guard logic is there to prevent users from accessing any modal meant to be used in post-auth area if they are not signed in. Our current silent login logic actually involves fetching user-data from the DB and storing them to the Vuex store, which takes a bit of time. So that nav guard logic actually should either be registered after some delay or moved to somewhere else.

I couldn't come up with the best solution for it straight away and thought it deserves some more time alone. it's a typical little code updates but with longer effort kind of issue I think.

SebinSong:
@taoeffect Just updated this PR for the fix of #2096 too

SebinSong:
@taoeffect some comment re [above](https://github.com/okTurtles/group-income/pull/2091#issuecomment-2187544497) :
I did not implement the requirement as part of the `/pending-approval` page because it was not that page that captures and console-log out the error. So I thought the in-app prompt UI we have would be a more suitable solution here and used it. Cheers,

frontend/model/contracts/group.js
```diff
@@ -1154,15 +1154,18 @@ sbp('chelonia/defineContract', {
...
-                  })()
+                  console.error('Error while joining the #General chatroom', e)
+                  const errMsg = L("Couldn't join the #{chatroomName} in the group. An error occurred: {error}", { chatroomName: CHATROOM_GENERAL_NAME, error: e?.message || e })
+                  const promptOptions = {
+                    heading: L('Error while joining a chatroom'),
+                    question: errMsg,
+                    primaryButton: L('Close')
+                  }
+
+                  await sbp('gi.ui/prompt', promptOptions)
```
corrideat: If you look at how it was previously, it was written in a non-blocking way, which I believe is no longer the case. One reason not to await here is that now you've made an error handler that could potentially throw. Ideally, error handlers should _not_ result in an error.

I'd make it non-blocking again because ideally contracts should not have code that depends on user interation, even when it doesn't block message processing. In this case, I think you can resolve it by removing the `await` and adding a `.catch()` handler (this is because not using a catch handler would result in an unhandled promise rejection, if the promise throws).

frontend/controller/actions/identity.js
```diff
@@ -408,8 +409,10 @@ export default (sbp('sbp/selectors/register', {
           return sbp('okTurtles.eventQueue/queueEvent', `login:${identityContractID ?? '(null)'}`, ['chelonia/contract/sync', ids, { force: true }])
         }))
       } catch (err) {
-        alert(L('Sync error during login: {msg}', { msg: err?.message || 'unknown error' }))
         console.error('Error during contract sync upon login (syncing all contractIDs)', err)
+
+        const humanErr = L('Sync error during login: {msg}', { msg: err?.message || 'unknown error' })
+        throw new GIErrorUIRuntimeError(humanErr)
```
corrideat: In cases where you `throw`, you should use the `cause` option including the original error (see example below). This allows error handlers access to the original error, which can be helpful for error handling and debugging.

```
throw new GIErrorUIRuntimeError(humanErr, { cause: err })
```

corrideat (dismissed):
Overall, I think this looks fine and is an improvement over `alert()`. There are two relatively small issues that I think should be addressed. I'm marking this as approved because I don't think either of the things I pointed out are blockers, but you should consider addressing my points.

frontend/controller/actions/group.js
```diff
@@ -482,7 +482,16 @@ export default (sbp('sbp/selectors/register', {
...
+      const primaryClicked = await sbp('gi.ui/prompt', {
+        heading: L('Failed to join the group'),
+        question: L('Error details:{br_}{err}', { err: e.message, ...LTags() }),
+        primaryButton: L('Refresh')
+      })
+
+      if (primaryClicked) {
+        const url = (window.location.href).split('?')[0]
+        window.open(url, '_self')
+      }
```
taoeffect: In this PR, I think it's best to preserve the behavior of `master` as much as we can, as it will save both of us headaches on having to think through the consequences of changing the behavior.

Before there was no call to refresh the page, and no there shouldn't be either. It's good to change the behavior to have it show the error, but let's limit it to that for now unless there's a strong reason to do more.

taoeffect (changes requested):
Fantastic work @SebinSong! One minor change request made.

frontend/model/contracts/group.js
SebinSong: @corrideat fair point. I think we can just remove `async / await` here, because there is no required action followed by clicking on that 'Close' button on the prompt.

frontend/controller/actions/group.js
SebinSong: @taoeffect Will do so. I added the page refresh logic here as it was part of the suggested solution of the issue #2096, BTW.

SebinSong:
@taoeffect 
Updated the PR again according to the feedbacks

frontend/controller/actions/group.js
taoeffect: Hmm, but isn't #2096 about updating the PendingApproval page? 🤔

taoeffect (approved):
Nice job @SebinSong !

frontend/controller/actions/group.js
SebinSong: @taoeffect 
Yes, I started off looking at the `PendingApproval.vue` but then realised `/pending-approval` page is merely the place redirected to as a follow-up step of an actual async action where the error is thrown, meaning `PendingApproval.vue` cannot catch that error.

For example,
`await sbp('gi.actions/group/joinWithInviteSecret')` [here](https://github.com/okTurtles/group-income/blob/41dc1f1d8ca0fea71f65a837d1ec717f522657e7/frontend/views/pages/Join.vue#L207C1-L209) leads to a chain of several `sbp` calls which involves `sbp('gi.actions/group/join')` ([here](https://github.com/okTurtles/group-income/blob/41dc1f1d8ca0fea71f65a837d1ec717f522657e7/frontend/model/contracts/identity.js#L257) in `identity.js` contract) down the line. that's actually where the target error mostly occurs. So I decided to add the prompt logic there instead.

So even after we land `/pending-approval` page, there is still a chain of async operations (which started before the page navigation) going on behind the scene. so the error thrown as part of that looks like it occurred within the pending approval page. This is what it appeared from my investigation.

PR #2095, opened 2024-06-19
Improve header style

### LINKED ISSUE #2067: Fix the style of direct message with many members
### Problem

I have just noticed that the style is broken when the DM title is overflow with it'
image
For the greeting message, we should consider the number of users and could display `and more...`.
`You, user1 and more...`

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #2067

### DIFF
--- a/frontend/views/components/Page.vue
+++ b/frontend/views/components/Page.vue
@@ -159,6 +159,7 @@ $pagePaddingDesktop: 5.5rem;
     top: 0;
     background: $general_2;
     z-index: 3;
+    padding: 0 5.75rem;
   }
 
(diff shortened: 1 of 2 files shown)


Silver-IT (commented):
Ready for the review!

I didn't work on the stuff to display `and more...` for the following reasons
- Noticed Slack doesn't display `and more` inside the greeting message
- We have 3 same titles; greeting message, channel name, document title. And I am not sure how to update the others too.

taoeffect (approved):
Thanks @Silver-IT !

PR #2101, opened 2024-06-20
Fix error to pin message in the older version of contracts

Silver-IT:
### Summary of Changes

- Fix error to pin messages in the older version of chatroom contracts which doesn't have `pinnedMessages` field in it's state


Silver-IT (commented):
Review ready!

taoeffect (approved):

PR #2109, opened 2024-06-22
Enable source maps in production

### LINKED ISSUE #2045: Souce maps missing from release version
### Problem

The release version doesn't seem to include source maps. The lack of source maps makes it difficult to identify and resolve problems as it obscures the source of errors in the generated logs, and it also makes it difficult to attach a debugger to the running application.

### Solution

Modify the build steps to ensure that source maps are being generated and included along release assets. Then, verify that source maps are working by manual inspection (for example, purposely generating an error and inspecting the logged information).

snowteamer:
Closes #2045

### Summary of changes
- Enable source maps in default esbuild options when `NODE_ENV=production`.
- Bump `package-lock.json` version field to match `package.json`'s one (i.e. v.0.5.4).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -239,7 +239,7 @@ module.exports = (grunt) => {
       minifySyntax: production,
       minifyWhitespace: production,
       outdir: distJS,
-      sourcemap: development,
+      sourcemap: true,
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Approved! Thanks @snowteamer !

Issue #2103, opened 2024-06-21
Multiple issues signing up and joining a new account on testing site


taoeffect:
### Problem

Multiple different errors encountered when joining a new user to the existing group we have:

#### Error 1

**UPDATE:** ✅ Fixed in #2105

```
[pubsub] Connection re-established
main.js:66 VERSION_INFO received: {ourVersion: '0.5.3', theirVersion: '0.5.3', ourContractsVersion: '0.5.3', theirContractsVersion: '0.5.3'}
chunk-VWGY47Z6-cached.js:22 
        
        
       GET https://testing.groupincome.org/kv/z9brRu3VN2ftwkQsqJRf9HC2LLRf5KoCygAVeR2DqzRKBT6h3zpM/unreadMessages 404 (Not Found)
chelonia/kv/get @ chunk-VWGY47Z6-cached.js:22
u @ chunk-SSANYCJJ-cached.js:1
gi.actions/identity/fetchChatRoomUnreadMessages @ main.js:66
u @ chunk-SSANYCJJ-cached.js:1
(anonymous) @ main.js:66
okTurtles.eventQueue/queueEvent @ chunk-VWGY47Z6-cached.js:1
u @ chunk-SSANYCJJ-cached.js:1
gi.actions/identity/loadChatRoomUnreadMessages @ main.js:66
u @ chunk-SSANYCJJ-cached.js:1
gi.actions/identity/login @ main.js:66
await in gi.actions/identity/login (async)
u @ chunk-SSANYCJJ-cached.js:1
(anonymous) @ main.js:156
Promise.then (async)
mounted @ main.js:156
Nr @ chunk-SNGBA5T7-cached.js:1
De @ chunk-SNGBA5T7-cached.js:1
$u @ chunk-SNGBA5T7-cached.js:1
J.$mount @ chunk-SNGBA5T7-cached.js:3
J.$mount @ chunk-SNGBA5T7-cached.js:5
tl @ main.js:156
await in tl (async)
(anonymous) @ main.js:156
main.js:66 [chelonia] verifying signature for z9brRu3VJAmRnpebQd2mZCkYKzdDyBHfmEZqgLhchYbj7MehvJrx with an existing key
[...]

PR #2116, opened 2024-06-24
Fix manifest loading

corrideat:


taoeffect (dismissed):
Nice catch 👍

shared/domains/chelonia/internals.js
```diff
@@ -1293,53 +1296,69 @@ export default (sbp('sbp/selectors/register', {
...
+    try {
+      // Since we require a contract state for the duration of this operation,
+      // we call retain on this contract ephemerally and release it when we're
+      // done. This prevent the contract form being removed while this function
+      // is executing. We need to do reference counting manually here because
+      // we can't block on the contract queue
+      if (!this.ephemeralReferenceCount[contractID]) {
+        this.ephemeralReferenceCount[contractID] = 1
+      } else {
+        this.ephemeralReferenceCount[contractID] = this.ephemeralReferenceCount[contractID] + 1
```
taoeffect: Could you modify this to simply say `+= 1`?

Alternatively, these lines can be collapsed like so:

```js
this.ephemeralReferenceCount[contractID] = (this.ephemeralReferenceCount[contractID] || 0) + 1
```

shared/domains/chelonia/internals.js
```diff
@@ -1293,53 +1296,69 @@ export default (sbp('sbp/selectors/register', {
...
-    // subscribed to the contract; if not, we call sync.
-    if (!this.subscriptionSet.has(contractID)) {
-      await sbp('chelonia/private/in/syncContract', contractID)
-    }
+    try {
+      // Since we require a contract state for the duration of this operation,
+      // we call retain on this contract ephemerally and release it when we're
+      // done. This prevent the contract form being removed while this function
+      // is executing. We need to do reference counting manually here because
+      // we can't block on the contract queue
```
taoeffect: If release operations are done on the same queue as this selector, then we wouldn't need to do this. Can you check if this is the case? If not, can you update the comment to explain why it's not the case?

shared/domains/chelonia/internals.js
```diff
@@ -1293,53 +1296,69 @@ export default (sbp('sbp/selectors/register', {
...
 
-      sbp('chelonia/private/queueEvent', externalContractID, ['chelonia/private/deleteOrRotateRevokedKeys', externalContractID]).catch((e) => {
-        console.error(`Error at deleteOrRotateRevokedKeys for contractID ${contractID} and externalContractID ${externalContractID}`, e)
+        sbp('chelonia/private/queueEvent', externalContractID, ['chelonia/private/deleteOrRotateRevokedKeys', externalContractID]).catch((e) => {
+          console.error(`Error at deleteOrRotateRevokedKeys for contractID ${contractID} and externalContractID ${externalContractID}`, e)
+        })
+      }
+    } finally {
+      sbp('chelonia/contract/release', contractID, { ephemeral: true }).catch((e) => {
+        console.warn('[chelonia/private/in/syncContractAndWatchKeys] Error calling release', contractID, e)
```
taoeffect: This should be `console.error`

taoeffect (approved):
very nice

(note: this will close #2103 because of the commit message, but not all issues have been fixed. Specifically: Error 5 may still likely be an issue)

PR #2121, opened 2024-06-25
#2120 - PWA missing an icon on iOS

### LINKED ISSUE #2120: PWA is missing icon
### Problem

The PWA seems to be missing an icon, which results in the OS assigning it a default one.

ETA: This seems to affect iOS. It doesn't seem to be an issue on Android.

### Solution

Create or update the manifest file to include an icon. Go through a test service (e.g., <https://www.webpagetest.org/>) to also find other potential related issues and fix those.

See also: https://realfavicongenerator.net/

SebinSong:
closes #2120 

@corrideat the app does have[ a manifest file](https://github.com/okTurtles/group-income/blob/master/frontend/assets/pwa-manifest.webmanifest) here if you don't remember, BTW.

I made two changes in this PR.

**1.** adding `<link rel='apple-touch-icon'>` attribute([reference](https://webhint.io/docs/user-guide/hints/hint-apple-touch-icons/)) to `index.html`

MDN, which I reference the most when it comes to PWA related topics, has that property with `192x192` size image. So I've done that same here.

**2.** [This article](https://pwa-book.awwwards.com/chapter-3) written about PWA recommends having `270x270` and `180x180` icon sizes too. So I've added them.

---
`grunt dev --tunnel` keeping failing and returns 404 on the iOS simulator on my Mac. So I actually wasn't able to fully verify if it works for the **mobile in `https://` context**. 
But at least I observed this difference in **desktop Safari** on my Mac after making the change **1.** above,

**[Before adding `apple-touch-icon`]**

before-apple-touch-icon

**[After adding `apple-touch-icon`]**

after-apple-touch-icon

I will try testing it again with better internet connection later on and make further updates if the fix doesn't work. Cheers,


SebinSong:
@taoeffect 
I can see the fix working on my iOS simulator (Safari browser):

**[ `grunt dev --tunnel` ] on this PR**



---
**[ Our testing website ]**

taoeffect (approved):
Nice work @SebinSong !

PR #2123, opened 2024-06-26
Only update readUntilMessageHash for non-pending messages

### LINKED ISSUE #2119: Chatrooms not loading on testing site
### Problem

Switching to some chatrooms results in an error and the messages do not load:

Screenshot 2024-06-25 at 11 59 04 AM

Screenshot 2024-06-25 at 11 58 55 AM

Screenshot 2024-06-25 at 12 04 23 PM

Sometimes it does load, and sometimes it doesn't. Usually it doesn't..?

This seems to happen specifically with private chatrooms.

### Solution

Find out what the problem is and fix it.

corrideat:
Fix #2119

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -181,7 +181,7 @@ const onChatScroll = function () {
     if (offsetTop + height <= curScrollBottom) {
       const bottomMessageCreatedHeight = msg.height
       const latestMessageCreatedHeight = this.currentChatRoomReadUntil?.createdHeight
-      if (!latestMessageCreatedHeight || latestMessageCreatedHeight <= bottomMessageCreatedHeight) {
+      if ((!latestMessageCreatedHeight || latestMessageCreatedHeight <= bottomMessageCreatedHeight) && !msg.pending) {
         this.updateReadUntilMessageHash({
           messageHash: msg.hash,
           createdHeight: msg.height
@@ -998,7 +998,7 @@ export default ({
               if (!fromOurselves && isScrollable) {
                 this.updateScroll()
               } else {
-                const msg = this.messages[this.messages.length - 1]
+                const msg = this.messages[this.messages.length - 1].filter(m => !m.pending)
                 this.updateReadUntilMessageHash({
                   messageHash: msg.hash,
                   createdHeight: msg.height
@@ -1047,7 +1047,7 @@ export default ({
             $state.complete()
             if (!this.$refs.conversation ||
             this.$refs.conversation.scrollHeight === this.$refs.conversation.clientHeight) {
-              const msg = this.messages[this.messages.length - 1]
+              const msg = this.messages[this.messages.length - 1].filter(m => !m.pending)
               if (msg) {
                 this.updateReadUntilMessageHash({
                   messageHash: msg.hash,
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -998,11 +998,13 @@ export default ({
               if (!fromOurselves && isScrollable) {
                 this.updateScroll()
               } else {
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Nicely done @corrideat !

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -181,7 +181,7 @@ const onChatScroll = function () {
     if (offsetTop + height <= curScrollBottom) {
       const bottomMessageCreatedHeight = msg.height
       const latestMessageCreatedHeight = this.currentChatRoomReadUntil?.createdHeight
-      if (!latestMessageCreatedHeight || latestMessageCreatedHeight <= bottomMessageCreatedHeight) {
+      if ((!latestMessageCreatedHeight || latestMessageCreatedHeight <= bottomMessageCreatedHeight) && !msg.pending) {
```
Silver-IT, line 184: This is not needed. msg.pending is already checked above in the line 177.
```
if (msg.pending || msg.hasFailed) continue
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -998,11 +998,13 @@ export default ({
               if (!fromOurselves && isScrollable) {
                 this.updateScroll()
               } else {
-                const msg = this.messages[this.messages.length - 1]
-                this.updateReadUntilMessageHash({
-                  messageHash: msg.hash,
-                  createdHeight: msg.height
-                })
+                const msg = this.messages.filter(m => !m.pending).slice(-1)
```
Silver-IT, line 1001: I think this returns array.
```
const msg = this.messages.filter(m => !m.pending).slice(-1)[0]

```

And we don't need to filter here `.filter(m => !m.pending)` because the last message can't be pending here.
```
const msg = this.messages[this.messages.length - 1]
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1047,7 +1049,7 @@ export default ({
             $state.complete()
             if (!this.$refs.conversation ||
             this.$refs.conversation.scrollHeight === this.$refs.conversation.clientHeight) {
-              const msg = this.messages[this.messages.length - 1]
+              const msg = this.messages.filter(m => !m.pending).slice(-1)
```
Silver-IT, line 1052: Also here, slice(-1) returns array.
And why is it needed to filter by `pending` here? Why do you think the last message could be pending here in this block?

Silver-IT (commented):
Nice work, @corrideat! This PR is already merged but I have some questions here.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: This sounds right, it's already checked. I added the check to all places that were updating the last unread count.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: > I think this returns array.

Yes, good catch. I should have used  `.pop()` instead.

>  because the last message can't be pending here.

Maybe, I need to double check that. But if you're right, it can be simplified to what it was.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: > Also here, slice(-1) returns array.

Correct, that should be fixed.

> Why do you think the last message could be pending

I'd need to double check, as for the other similar question. But the fact was that some messages that were pending were being added to the last unread, which was then causing issues loading the chatroom. Hence, at least at one location the messages were pending.

frontend/views/containers/chatroom/ChatMain.vue
Silver-IT: > Yes, good catch. I should have used .pop() instead.

`.pop()` can't be solution too. The length of `this.messages` will be changed.

PR #2129, opened 2024-06-28
Store notifications in server side KV store

### LINKED ISSUE #1936: Store notifications in server-side KV store
### Problem

Currently, the read status of notifications is per-device.

Screenshot 2024-04-17 at 8 41 39 AM

So when notifications are dismissed or cleared, they must be dismissed or cleared again in another browser or different device, or if someone logs out and logs back in again.

This is annoying.

### Solution

PR #1912 recently introduced a server-side authenticated and encrypted key-value store that can be used to solve this problem.

That PR used it to update `'gi.actions/group/updateLastLoggedIn'` to store the `lastLoggedIn` information for for the group contract.

The key-value store is scoped for a specific contract. In PR #1912, the `lastLoggedIn` information is stored under the group contractID so that there's a single place to look for the latest information about everyone's `lastLoggedIn` information.

For this issue, you'll need to scope the notifications key under the user's identityContractID and store all of the notifications under that key.

Very important: as over time there can be a lot of notifications, make sure that only the N most recent notifications are stored. I believe we already do this (see `MAX_COUNT` in `frontend/model/notifications/storageConstants.js` and `applyStorageRules` in `frontend/model/notifications/utils.js`).

Also: remember to make use of the KV-store's pubsub feature so that when a notification is marked as read or removed on one device, it instantly happens on any other open device. Likewise, clients should re-read the notifications key to get the latest value upon page refresh/logging in *and* re-establishing a connection to the server.

Silver-IT:
### Summary of Changes

- Closes #1936


frontend/controller/actions/identity-kv.js
```diff
@@ -176,8 +186,92 @@ export default (sbp('sbp/selectors/register', {
...
+    //     .map(notification => notification.hash)
+    //     .reduce((acc, hash) => ({ ...acc, [hash]: notificationStatus[hash] }), {})
+    // }
+
+    // const updatedOnConflict = async (...args) => {
+    //   if (typeof onconflict === 'function') {
+    //     return enforceStorageRules(await onconflict(...args))
+    //   }
+    //   return null
+    // }
```
Silver-IT: To comment these out was my intentional to keep in mind that we should save only notifications which are applied the storage rules. (by `applyStorageRules` function)

But, I don't think we should do this in here, where we save notification status in KV store, with the following reasons.
- Each device could have different notifications, and storage-rule-applied notifications could differ from one to another.
- I don't think to save all of them will cause a memory problem since they are very simple object.

@taoeffect, if you still want to apply storage rules in saving notification status in KV store, I can restore this code.
And if you agree with me, I will remove them and add this comment instead.

Silver-IT (commented):
Ready for the review!

frontend/model/state.js
```diff
@@ -74,8 +74,8 @@ sbp('sbp/selectors/register', {
     // Note: Update this function when renaming a Vuex module, or implementing a new one,
     // or adding new settings to the initialState above
     // Example:
-    // if (!state.notifications) {
-    //   state.notifications = []
+    // if (!state.preferences) {
+    //   state.preferences = {}
```
taoeffect, line 78: Please revert unnecessary change (also confusing since we have the same lines uncommented below)

frontend/model/notifications/vuexModule.js
```diff
@@ -66,62 +72,52 @@ const getters = {
...
+        : getters.notifications.filter(item => !item.read && item.groupID === groupID)
     )
   }
 }
 
 const mutations = {
   // Seems necessary because the red badge would not clear upon signing up a new user in Cypress via the bypassUI mechanism.
   logout (state) {
-    state.splice(0, state.length)
+    Object.assign(state, cloneDeep(defaultState))
```
taoeffect, line 86: Probably want to loop over all keys and delete them first, because if keys change between versions this won't necessarily reset the state

frontend/model/contracts/group.js
```diff
@@ -345,8 +345,8 @@ const leaveChatRoomAction = (state, chatRoomID, memberID, actorID, leavingGroup)
     ...extraParams,
     hooks: {
       onprocessed: () => {
-        const rootState = sbp('state/vuex/state')
-        if (memberID === rootState.loggedIn.identityContractID) {
+        const { ourIdentityContractId } = sbp('state/vuex/getters')
+        if (memberID === ourIdentityContractId) {
```
taoeffect: These changes will cause a lot of conflicts and problems for the work that @corrideat is doing with moving Chelonia into the service worker.

Could you please revert all of the non-essential changes in this file? Ricardo requests that changes related to new uses of root getters be avoided if possible.

frontend/controller/actions/identity-kv.js
taoeffect: I think we should figure out a way to have the best of all worlds.

In other words, I think you are correct that we need to be able to allow different machines to have different notifications, while at the same time we should try to avoid keeping a record of all notification statuses forever in the kv-store. Even though you are right that they are a small amount of data, over a long enough period of time this could potentially cause problems.

So let's apply different storage rules to statuses stored in the KV store.

`initNotificationStatus`, instead of being an object, can be a function that returns an object, and that will allow us to store a `createdDate` field on it.

We can use that field to remove all notification statuses that are older than the oldest notifications possible that are stored in the Vuex state.

Based on the rules of `applyStorageRules` we can apply the following rule here:

- Delete all read notifications that are older than `MAX_AGE_READ` (see `storageConstants.js`)

This will allow us to safely store all notifications from all devices on the server while keeping their number reasonably limited.

We should also add a comment here in the code to explain this reasoning.

taoeffect (changes requested):
Great work @Silver-IT!! Review ready!

frontend/model/notifications/utils.js
```diff
@@ -121,3 +120,14 @@ export function isOlder (notification: Notification): boolean {
 export function maxAge (notification: Notification): number {
   return notification.read ? MAX_AGE_READ : MAX_AGE_UNREAD
 }
+
+export function makeNotificationHash (notification: Object): string {
+  const excludingKeys = ['read', 'hash', 'body']
```
taoeffect: ### Problem 1 - same hash for different notifications

Hmm, by excluding the `body` key, is it possible that two different notifications might get hashed the same way? Looking at the `templates.js`, it seems that it is possible.

But, I agree that the `body` should be excluded because it could change (e.g. when someone changes their display name, and then logs back in on a new device).

So what's needed is an identifier that is **both** unique to the notification, and also consistent across devices (meaning it doesn't matter when the notification is sent or on what device).

When calling `gi.notifications/emit`, we can pass in a variable called `uniqueConsistentID`.

- For notifications that are generated inside of contract sideEffects, we can use the `hash` parameter that is passed in to the `sideEffect`. This guarantees that the same unique consistent identifier is used for these notifications. E.g. do:

    ```js
    uniqueConsistentID: `MEMBER_ADDED.${hash}`
    ```

- For notifications that are generated outside of sideEffects, e.g. the ones created by the periodic notifications system inside of `mainNotificationsMixin.js`, you will have to construct a unique consistent identifier yourself. One example is given below:

    ```js
          emit ({ rootState, rootGetters }) {
            const period = rootGetters.groupSettings.distributionDate
            const groupID = /* ... code here to fetch corresponding groupID ... */
            sbp('gi.notifications/emit', 'NEAR_DISTRIBUTION_END', {
              // WARNING: see "Problem 2" below for why createdDate can cause problems
              createdDate: new Date().toISOString(),
              groupID: rootState.currentGroupId,
              period,
              uniqueConsistentID: `NEAR_DISTRIBUTION_END.${groupID}.${period}`
            })
          },
    ```

### Problem 2 - different hash for the same notifications

As shown in the example above for `NEAR_DISTRIBUTION_END`, some notifications include dynamic data in them that changes depending on when the notification is posted.

This means that if you log in on another device, the same notification will have a different hash. In this case, the offending variable is `createdDate`.

So, **both problems can be solved by ONLY using `uniqueConsistentID`.**

In fact, by requiring all notifications to have a `uniqueConsistentID`, we can remove the `makeNotificationHash` function entirely and only rely on the `uniqueConsistentID` parameter to act as the key.

frontend/model/state.js
Silver-IT: Actually, I updated this because the previous comment was not correct.
```
if (!state.notifications) {
  state.notifications = []
}
```

But in this PR, the `state.notifications` is no longer an array, and it's an object. That's why I changed the comment. Or do you want me to change it to the following?
```
if (!state.notifications) {
  state.notifications = { items: [], status: {} }
}
```

frontend/model/notifications/vuexModule.js
Silver-IT: Nice catch. Thanks

frontend/model/contracts/group.js
Silver-IT: If you want, I will revert all my changes regarding `ourIdentityContractID` getter. I think as long as we have `getter` that returns the identity contract ID of the current user, we should not try to get it from the `state`. That's because once we need to change something related to this, it would be fine to update only the `getter`, not all the codes using `state`.

taoeffect: @Silver-IT I agree with you that it is an improvement, and normally I would say it's OK, but for @corrideat's sake, let's please revert it. He has made changes to how this is done in this PR and the changes here will cause a lot of problems for him.

Silver-IT: Okay. I will revert it.

frontend/model/notifications/utils.js
Silver-IT: > by excluding the body key, is it possible that two different notifications might get hashed the same way?

Simply, the `body` is the outcome of translation. 😁 So we should not include it.

##
For the `uniqueConsistentID`, it's a good idea. And I don't think we don't need to add `type` in the `uniqueConsistentID` because the `notification` object has its type`.
```
const notification = {
  ...template,
  hash: '',
  avatarUserID: template.avatarUserID || sbp('state/vuex/getters').ourIdentityContractId,
  // Sets 'groupID' if this notification only pertains to a certain group.
  ...(template.scope === 'group' ? { groupID: data.groupID } : {}),
  // Store integer timestamps rather than ISO strings here to make age comparisons easier.
  type
}
notification.hash = makeNotificationHash(notification)
```
And for the notifications created inside the contract sideEffect, I think there are no problems with the current function `makeNotificationHash` because it uses `meta.createdDate` as `createdDate` of notification itself.
But for the other notifications created outside of contract sideEffect, their `createdDate` could differ from one device to another even though they are same notifications. So, I think we should consider only this one.

Silver-IT: However, I totally agree with your idea that we should use `uniqueConsistentID`.
I thought you wanted to only add that field and still using `makeNotificationHash` function. But if we add `uniqueConsistentID`, we don't need `makeNotificationHash` function anymore. 😁👍

frontend/controller/actions/identity-kv.js
Silver-IT: 👍

frontend/model/notifications/utils.js
Silver-IT: @taoeffect, while I was working on this, I noticed that we don't need `uniqueConsistentID` and we can still have identifiable data to make hash using `makeNotificationHash`. The only reason why we need `uniqueConsistentID` was that we set `createdDate` inside the `notificationData` in the place outside of contract sideEffect.
```
sbp('gi.notifications/emit', 'NEAR_DISTRIBUTION_END', {
  createdDate: new Date().toISOString(),
  groupID: rootState.currentGroupId,
  period: rootGetters.groupSettings.distributionDate
})
```
But we don't need to set the `createdDate` here because we have a way to use the default value `Date.now()`.
```
const notification = {
  ...template,
  hash: makeNotificationHash({ ...data, type }),
  avatarUserID: template.avatarUserID || sbp('state/vuex/getters').ourIdentityContractId,
  // Sets 'groupID' if this notification only pertains to a certain group.
  ...(template.scope === 'group' ? { groupID: data.groupID } : {}),
  // Store integer timestamps rather than ISO strings here to make age comparisons easier.
  type
}
```
This means that we can easily using `makeNotificationHash` by removing `createdDate` in the place where we pass `notificationData`.

And to be honest, I really prefer to use `makeNotificationHash` because to make ID (`uniqueConsistentID`) manually is something like tricky temporary solution when we don't have the better solution. We also create hash for GIMessages using the identifiable value.
```
hash = createCID(value)
```

Silver-IT (commented):
Ready for the review!

frontend/model/notifications/utils.js
taoeffect: Removing `createdDate` from the hashed data only addresses **Problem 2** in the comment I wrote above.

**Problem 1** is still an issue. So I think we need to introduce `uniqueConsistentID`, unless you have some other idea.

Silver-IT: I don't think there is a problem (Problem 1), because notificationData contains `createdDate`, which is `meta.createdDate`, when the we create notifications in the contract sideEffect.

Silver-IT: Notification is created by it's type and it's data.
```
'gi.notifications/emit' (type: string, data: NotificationData) {
  const template: NotificationTemplate = templates[type](data)

  if (template.scope === 'group' && !data.groupID) {
    throw new TypeError('Incomplete notification data: `data.groupID` is required.')
  }

  // Creates the notification object in a single step.
  const notification = {
    ...template,
    hash: makeNotificationHash({ ...data, type }),
    avatarUserID: template.avatarUserID || sbp('state/vuex/getters').ourIdentityContractId,
    // Sets 'groupID' if this notification only pertains to a certain group.
    ...(template.scope === 'group' ? { groupID: data.groupID } : {}),
    // Store integer timestamps rather than ISO strings here to make age comparisons easier.
    type
  }
  sbp('gi.actions/identity/kv/addNotificationStatus', notification)
  sbp('state/vuex/commit', keys.ADD_NOTIFICATION, notification)
}
```
For the notifications which are created in contracts sideEffect, their type and data are same in any devices.

taoeffect: > I don't think there is a problem (Problem 1), because notificationData contains createdDate, which is meta.createdDate, when the we create notifications in the contract sideEffect.

There is a problem. It is possible for notifications to contain the same data but be different notifications.

For example, `PAYMENT_THANKYOU_SENT` has this:

```js
  PAYMENT_THANKYOU_SENT (data: { creatorID: string, fromMemberID: string, toMemberID: string }) {
    return {
      avatarUserID: data.fromMemberID,
      body: L('{strong_}{name}{_strong} sent you a {strong_}thank you note{_strong} for your contribution.', {
        name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.fromMemberID}`,
        ...LTags('strong')
      }),
      creatorID: data.fromMemberID,
      icon: '',
      level: 'info',
      linkTo: `/payments?modal=ThankYouNoteModal&from=${data.fromMemberID}&to=${data.toMemberID}`,
      scope: 'group'
    }
  },
```

This notification will be sent multiple times, and it will have the same exact hash each time it is sent between these two members.

EDIT: There may be many other such notifications as there is nothing preventing the developer from creating such notifications. OTOH, with `uniqueConsistentID` the problem is explicitly considered and acknowledged.

Silver-IT: We need to focus on the data here, where we call the `gi.notifications/emit` function.
```
sbp('gi.contracts/group/emitNotificationsAfterSyncing', [contractID, innerSigningContractID], [{
  notificationName: 'PAYMENT_THANKYOU_SENT',
  notificationData: {
    createdDate: meta.createdDate,
    groupID: contractID,
    fromMemberID: innerSigningContractID,
    toMemberID: data.toMemberID
  }
}])
```

taoeffect: This is going to generate the same hash every time:

```js
        sbp('gi.notifications/emit', 'NEW_DISTRIBUTION_PERIOD', {
          groupID: rootState.currentGroupId,
          creatorID: rootGetters.ourIdentityContractId,
          memberType: rootGetters.ourGroupProfile.incomeDetailsType === 'pledgeAmount' ? 'pledger' : 'receiver'
        })
```

This is going to continue to be a problem for the reasons I mentioned in the original post above.

Anyone creating these notifications is going to eventually accidentally create notifications that result in the same hash (Problem 1).

Instead of ignoring this problem and manually fixing it anyway when it comes up, I think it's much better to explicitly acknowledge the problem and require all notifications to define `uniqueConsistentID` manually. You need to do it anyway with the data. By defining `uniqueConsistentID` explicitly we inform all developers about this issue and force them to think about it.

frontend/model/state.js
taoeffect: @Silver-IT sure, feel free to change it to that

frontend/model/notifications/utils.js
```diff
@@ -121,3 +120,11 @@ export function isOlder (notification: Notification): boolean {
 export function maxAge (notification: Notification): number {
   return notification.read ? MAX_AGE_READ : MAX_AGE_UNREAD
 }
+
+export function makeNotificationHash (notification: Object): string {
+  const sorted = Object.keys(notification).sort().reduce((acc, curKey) => {
+    acc.push([curKey, notification[curKey]])
+    return acc
+  }, [])
```
taoeffect: Please create an algorithm that recursively checks whether the value (`notification[curKey]`) is an object, and if so performs the same reduction on the object (to an array). Some notification data might include objects, as well as nested objects, and if we're going to properly create the hash for the outer object, we should do the same for any inner objects.

taoeffect (changes requested):
Great work @Silver-IT! Only a few remaining issues. Also the previous issue about the same hash [has not been addressed](https://github.com/okTurtles/group-income/pull/2129#discussion_r1661741183).

frontend/model/contracts/shared/giLodash.js
```diff
@@ -163,6 +163,28 @@ export function deepEqualJSONType (a: any, b: any): boolean {
   return true
 }
 
+export function sortAny (unsorted: any): any {
+  const sort = (unsorted) => {
+    if (!unsorted || typeof unsorted !== 'object') {
+      return unsorted
+    }
+    if (Array.isArray(unsorted)) {
+      return unsorted.map(v => sort(v)).sort()
```
taoeffect: The last `.sort()` at the end here isn't needed, so can be removed to speed up the algorithm

frontend/model/contracts/shared/giLodash.js
```diff
@@ -163,6 +163,28 @@ export function deepEqualJSONType (a: any, b: any): boolean {
   return true
 }
 
+export function sortAny (unsorted: any): any {
```
taoeffect: I would rename this to `hashableRepresentation`, as that is what we're trying to do.

frontend/model/contracts/shared/giLodash.js
```diff
@@ -163,6 +163,28 @@ export function deepEqualJSONType (a: any, b: any): boolean {
...
+  const sort = (unsorted) => {
```
taoeffect: I don't think this inner nested `sort` function is needed. You can call `hashableRepresentation` recursively

frontend/model/contracts/shared/giLodash.js
```diff
@@ -163,6 +163,28 @@ export function deepEqualJSONType (a: any, b: any): boolean {
...
+    }
+    if (Array.isArray(unsorted)) {
+      return unsorted.map(v => sort(v)).sort()
+    } else {
+      return Object.keys(unsorted).sort().reduce((acc, curKey) => {
+        if (unsorted[curKey] && typeof unsorted[curKey] === 'object') {
+          acc.push([curKey, sort(unsorted[curKey])])
+        } else {
+          acc.push([curKey, unsorted[curKey]])
+        }
```
taoeffect: As these `typeof` checks are already done at the start of `hashableRepresentation`, you should be able to replace these 5 lines with 1 line:

```js
acc.push([curKey, hashableRepresentation(unsorted[curKey])])
```

frontend/main.js
```diff
@@ -99,7 +99,7 @@ async function startApp () {
 
   // Used in 'chelonia/configure' hooks to emit an error notification.
   function errorNotification (activity: string, error: Error, message: GIMessage) {
-    sbp('gi.notifications/emit', 'CHELONIA_ERROR', { activity, error, message })
+    sbp('gi.notifications/emit', 'CHELONIA_ERROR', { createdDate: new Date().toISOString(), activity, error, message })
```
taoeffect, line 102: FYI, this notification now runs afoul of `Problem 2 - different hash for the same notifications` because it uses `new Date().toISOString()`.

Meaning: the same notification on different devices will have different hashes.

taoeffect (changes requested):
Review ready!

frontend/model/notifications/utils.js
taoeffect: If you really don't want to use `uniqueConsistentID` you will have to manually perform the same checks on the `data` that's passed in to `gi.notifications/emit`.

Also, in this case please add a comment above the definition of `gi.notifications/emit` explaining all of this.

My recommendation remains to prefer `uniqueConsistentID` over `makeNotificationHash`, simply because using `uniqueConsistentID` forces the developer to think about this problem, whereas with `makeNotificationHash` the developer is allowed to ignore it, but then passes on that burden to the reviewer who has to double-check their work and also not miss any mistakes the developer made in either running afoul of **Problem 1 - same hash for different notifications** or **Problem 2 - different hash for the same notifications**.

Either way, please document both problems 1 and 2 in a comment above `gi.notifications/emit`. 🙏

taoeffect (changes requested):
Also a request for comment above `gi.notifications/emit`

frontend/model/contracts/shared/giLodash.js
Silver-IT: I think it's needed because `unsorted` could be an array just like the following.
```
[
  1,
  [2, 4, 1],
  { b: 2, a: 1 },
  undefined,
  0,
  null,
  'any string'
]
```

Silver-IT: Aha, I got what you mean. Yeah, I don't think it's needed. Nice catch!!!

frontend/model/contracts/shared/giLodash.js
Silver-IT: Yeah, right. Nice catch. Thanks

frontend/main.js
Silver-IT: This was my intention to make all the errors independent because
- Same error could occur several times
- Even though it seems to be same, the reason could be different from one to another, from device to device

frontend/model/notifications/utils.js
Silver-IT: Yeah, thanks.

taoeffect (approved):
Fantastic work @Silver-IT!!! 🎉

PR #2131, opened 2024-06-28
Fix #2128

### LINKED ISSUE #2128: Failure to re-subscribe to chatroom channel after update
### Problem

After deploying v0.5.6 to the testing site, the tab I had open in Brave reloaded and refreshed, and showed that for some reason a whole bunch of pubsub channels couldn't be joined:

https://github.com/okTurtles/group-income/assets/138706/3c598aff-a1e4-4f4b-8055-9ba185e99fdf

Screenshot 2024-06-27 at 12 14 30 PM

The channel I was in at the moment (`#new-private`) was one of those channels.

On my phone I logged in and posted messages to that channel:

tmp-1719515424058

You can see that the bug #2061 got triggered again too.

But these messages did not appear on desktop! Meaning it really had stopped receiving updates and was no longer subscribed!

So I refreshed the page then to see what would happen.

After refreshing the page, I saw that these `pubsub` errors were still going on:

Screenshot 2024-06-27 at 12 23 34 PM

So in the screenshot above, you can see that the `#general` chatroom has been not been subscribed to, and that posting a message to it leaves it as pending, per #2061.

In the screenshot below, you can see that this really is the `#general` chatroom, as it has the right users in it:

Screenshot 2024-06-27 at 12 27 23 PM

Posting another message reproduces the behavior of #2061 with the previous message shown as no longer pending:

Screenshot 2024-06-27 at 12 27 57 PM

Also, when debugging and running sync with force:true on the #general chatroom to fetch new messages, when they got fetched the chatroom didn't scroll to the bottom as it's supposed to.

```js
sbp('chelonia/contract/sync', chatroomID, { force: true })
```

### Solution

Fix this failure to subscribe problem, and #2061 as well. They might be the same issue, or related issues.

One issue is that on the server we aren't adding all known `contractID`s to the pubsub `channels`.

The only place they're added is here:

```js
  'backend/server/handleEntry': async function (deserializedHEAD: Object, entry: string) {
    const contractID = deserializedHEAD.contractID
    sbp('okTurtles.data/get', PUBSUB_INSTANCE).channels.add(contractID)
```

We need to add all `contractID`s on boot  to the pubsub `channels`.

corrideat:


backend/server.js
```diff
@@ -241,12 +244,18 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
     // partition
```
snowteamer, line 244: Should I read this as `we create...` ?

backend/server.js
```diff
@@ -241,12 +244,18 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
...
     recoveredState.contracts = Object.create(null)
+    const channels = sbp('okTurtles.data/get', PUBSUB_INSTANCE).channels
     await Promise.all(savedStateIndex.split('\x00').map(async (contractID) => {
       const cpSerialized = await sbp('chelonia/db/get', `_private_cheloniaState_${contractID}`)
-      if (!cpSerialized) return
+      if (!cpSerialized) {
+        console.warn('[chelonia] Contract in index but missing state', contractID)
+        return
+      }
       const cp = JSON.parse(cpSerialized)
```
snowteamer, line 254: What does `cp` mean in `cpSerialized` and `cp`? From the above comment I would guess `contractPartition`, but I'm not sure, and without the comment I would have no clue.

Is the concept of contract partition explained somewhere? `git grep partition` only yields this one occurence in that comment.

taoeffect (approved):

snowteamer (commented):
LGTM!

backend/server.js
corrideat: Yes, it means contract partition and no, it's not explained elsewhere, but it refers to how the state is partitioned by contract (hence, contract partition). The reason to do this is performance.

backend/server.js
corrideat: Indeed. Let me fix that.

PR #2134, opened 2024-06-28
Fix #2133

### LINKED ISSUE #2133: 'Error: Cannot set existing unrevoked key' when joining group
### Problem

@Silver-IT found an error when joining group:

image

Screenshot 2024-06-28 at 9 42 58 AM

Note that this `alert` no longer appears, probably because of the changes that were just merged in from #2091. But the error still does appear in the console. @corrideat said that it should appear to the user too...

**Steps to reproduce**

1. [Tab1] user1 creates a group, copies the invitation link, and logs out
2. [Tab2] user2 tries to join group with the invitation link, gets the pending page, and closes the tab
3. [Tab3] user3 tries to join group with the invitation link, gets the pending page, and closes the tab
4. [Tab1] user1 logs in
5. [Tab4] user2 logs in and gets the following error

### Solution

I'm not sure what the solution is, but I think this series of steps should not result in any error.

corrideat:


taoeffect (approved):
Nice job @corrideat! Interesting changes - especially with the new queue for `'gi.actions/group/join'`, but it seems like that apparently might be necessary for some situations, per your comment. Nicely done with the comments 👍

PR #2137, opened 2024-06-28
Backport SW API changes without service worker

corrideat:


shared/domains/chelonia/chelonia.js
```diff
@@ -645,11 +654,12 @@ export default (sbp('sbp/selectors/register', {
         [`${contract.manifest}/${action}/process`]: (message: Object, state: Object) => {
           const { meta, data, contractID } = message
           // TODO: optimize so that you're creating a proxy object only when needed
-          const gProxy = gettersProxy(state, contract.getters)
-          state = state || contract.state(contractID)
-          contract.metadata.validate(meta, { state, ...gProxy, contractID })
-          contract.actions[action].validate(data, { state, ...gProxy, meta, message, contractID })
-          contract.actions[action].process(message, { state, ...gProxy })
+          const stateCopy = cloneDeep(state || contract.state(contractID))
```
taoeffect: Why is this change being made? Ultimately `state` gets modified on line 662 anyway... so why this clone/copy?

shared/domains/chelonia/utils.js
```diff
@@ -790,3 +791,14 @@ export const clearObject = (o: Object) => {
...
 }
+
+export const checkCanBeGarbageCollected = function (id: string) {
+  const rootState = sbp(this.config.stateSelector)
+  return (
+    // Check persistent references
+    (!has(rootState.contracts, id) || !has(rootState.contracts[id], 'references')) &&
+    // Check ephemeral references
+    !has(this.ephemeralReferenceCount, id)) &&
+    // Check foreign keys (i.e., that no keys are being watched)
```
taoeffect: Please clarify a bit more in this comment: are we checking whether this `id` is watching other contracts? Or are we checking whether other contracts are watching this `id`? Or both?

shared/serdes/index.js
```diff
@@ -0,0 +1,155 @@
...
+// This file is at attempt at addressing the lack of support for
+// custom objects in `structuredClone`. See <https://github.com/whatwg/html/issues/7428>.
+// We need this so that certain custom objects can be shared using a
+// `MessagePort`. To use this functionality, the `serializer` function must be
+// called on the _sending_ side and the `deserializer` function must be called
+// on the _receiving_ side.
+// Note that it's paramount that the _receiving_ side call `deserializer.register`
+// with all of the possible types that are supported.
+// For how to implement `serialize` and `deserialize` support for custom types,
+// see the example below for a class `X`.
```
taoeffect, line 15: Where is this example? I do not see it. It should probably be placed here near the top

test/cypress/support/index.js
```diff
@@ -18,9 +18,32 @@ import './commands.js'
...
   cy.clearLocalStorage()
-  indexedDB.deleteDatabase('Group Income')
+  if (typeof indexedDB === 'object') {
+    cy.wrap(indexedDB.databases().then((db) => {
+      console.log('[cypress] Registered DBs', db)
+      return Promise.all(db.map(({ name }) => indexedDB.deleteDatabase(name)))
+    }
+    ))
+  }
+  cy.wrap(undefined).then(() => console.log('[cypress] Finished `before`'))
```
taoeffect, line 48: What is this `cy.wrap(undefined)` thing and why is it here?

(Weird stuff should be either unweirded or commented)

frontend/main.js
```diff
@@ -175,6 +240,10 @@ async function startApp () {
     }
   })
 
+  sbp('okTurtles.events/on', EVENT_HANDLED, (contractID, message) => {
+    sbp('okTurtles.events/emit', SHELTER_EVENT_HANDLED, contractID, message)
+  })
+
```
taoeffect: Weird code: why do this instead of just listening for `EVENT_HANDLED`?

If there's a good reason, it needs to be explained in a comment.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -57,6 +57,18 @@ self.addEventListener('message', function (event) {
...
+      case 'ping':
+        event.source.postMessage({ type: 'pong' })
+        break
+      case 'shutdown':
+        self.registration.unregister()
+          .then(function () {
+            return self.clients.matchAll()
+          })
+          .then(function (clients) {
+            clients.forEach(client => client.navigate(client.url))
```
taoeffect, line 71: Could you add a comment here explaining what this is doing?

frontend/controller/service-worker.js
```diff
@@ -22,6 +22,12 @@ sbp('sbp/selectors/register', {
         // if an active service-worker exists, checks for the updates immediately first and then repeats it every 1hr
         await swRegistration.update()
         setInterval(() => sbp('service-worker/update'), HOURS_MILLIS)
+
+        // Keep the service worker alive while the window is open
+        // The default idle timeout on Chrome and Firefox is 30 seconds. We send
+        // a ping message every 5 seconds to ensure that the worker remains
+        // active.
+        setInterval(() => navigator.serviceWorker.controller?.postMessage({ type: 'ping' }), 5000)
```
taoeffect: Could you further add a comment explaining why this is a good thing to do and whether or not there are any downsides to doing this?

frontend/main.js
```diff
@@ -103,19 +105,105 @@ async function startApp () {
...
+        // prefer successfully joined groups
+          sbp('state/vuex/state')[cID]?.profiles?.[groupContractID] ? -1 : 1
+        )[0] || null
+      sbp('state/vuex/commit', 'setCurrentChatRoomId', {})
+      sbp('state/vuex/commit', 'setCurrentGroupId', { contractID: groupIdToSwitch })
+      if (currentGroupId === groupContractID) {
+        sbp('controller/router').push({ path: '/' }).catch(() => {})
+      }
+    }
+  })
```
taoeffect: `main.js` is not the right place for this sort of code. This belongs either in `controller/actions/chatroom.js` or in `model/contracts/chatroom.js`, or maybe in `ChatMain.vue`, or somewhere else that is context appropriate. Same for the rest of the code.

`frontend/main.js` is too cluttered as it is. Any modifications to it should remove code and simplify it.

frontend/main.js
```diff
@@ -175,6 +264,68 @@ async function startApp () {
...
+      const state = cheloniaState[x]
+      const contractState = cheloniaState.contracts[x]
+      if (contractState) {
+        Vue.set(vuexState.contracts, x, cloneDeep(contractState))
+      }
+      if (state) {
+        Vue.set(vuexState, x, cloneDeep(state))
+      }
+    })
+  })
```
taoeffect: Please see if this code can be rewritten to be more generic (to support redux etc) and moved into some chelonia namespace that's imported. We cannot expect developers to write this.

frontend/main.js
```diff
@@ -270,7 +421,14 @@ async function startApp () {
...
   // register service-worker
-  await sbp('service-workers/setup')
+  /* await Promise.race(
+    [sbp('service-workers/setup'),
+      new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error('Timed out setting up service worker'))
+        }, 15e3)
+      })]
+  ) */
```
taoeffect: Why is this commented out?

shared/domains/chelonia/chelonia.js
corrideat: This is done to emulate the sandbox environment, where contracts don't have direct access to modify the state. I made this change to test that things didn't break or to fix it as appropriate (remember, when I started working on this it was the SW + sandboxing). It's probably fine to remove this, other than that contracts need to be re-tested when sandboxing is implemented. On a similar note, the same thing applies to the proxy (I didn't change it, but it might need to be done differently).

corrideat: TL;DR: Can be reverted.

shared/domains/chelonia/utils.js
corrideat: I only moved this code from where it was, but I can clarify.

test/cypress/support/index.js
corrideat: I can re-test or add a comment. It seemed to be needed for cypress to actually wait for the console.log instead of printing it before being actually done (i.e., forces the call into a queue).

frontend/main.js
corrideat: Because this is a new event I added that Cypress listens for. The commit I made today also adds more code to this particular handler (as we discussed; it deals with keeping Chelonia and Vuex state in sync) that I'd hoped I wouldn't need for this PR.

frontend/main.js
corrideat: It shouldn't be, but as I mentioned I needed to do this so that I could run the tests (weird issue starting the service worker)

frontend/main.js
corrideat: I think it can be made more generic, at least as generic as chelonia is with `reactiveSet`, etc. It already is pretty generic (save for the `Vue` parts), it's just needing a way to call / include it. The parts to move are these, the saveCheloniaState part and maybe some of the login handler.

shared/domains/chelonia/chelonia.js
taoeffect: OK, yeah I would revert this and save it for when we get to the sandbox

test/cypress/support/index.js
taoeffect: Yeah that makes sense, still a comment is needed. And maybe check if `cy.then()` works directly instead of needing to wrap something (or if there's a similar/related API that doesn't require `.wrap()`)

frontend/main.js
taoeffect: I don't think that answers my question as far as why Cypress can't listen directly for `EVENT_HANDLED`...

frontend/controller/actions/group-kv.js
```diff
@@ -5,8 +5,8 @@ import { KV_QUEUE } from '~/frontend/utils/events.js'
 
 export default (sbp('sbp/selectors/register', {
   'gi.actions/group/kv/updateLastLoggedIn': async ({ contractID }: { contractID: string }) => {
-    const { ourIdentityContractId } = sbp('state/vuex/getters')
-    if (!ourIdentityContractId) {
+    const identityContractID = sbp('chelonia/rootState').loggedIn?.identityContractID
```
taoeffect, line 8: I assume this change is because for some reason we're not allowed to use root getters in the service worker? 🤔 

(Maybe there's a way to enable that...?)

frontend/main.js
corrideat: Because Cypress needs to listen for when the Vuex state changed, which happens on SHELTER_EVENT_HANDLED.

frontend/controller/actions/group-kv.js
corrideat: Correct. There isn't a way to enable that without code duplication (of the bundle kind, not of the source kind). However, global getters are a bad idea for contracts because they make it difficult or impossible to enforce containment, which is a requirement for any meaningful sandboxing, or to audit what contracts do. Furthermore, getters will likely need to be asynchronous if contracts are to be sandboxed at a later time.

Getters are fine to use and have in the application itself, but I wouldn't say they are well-suited to the service worker environment, where they just add complexity.

In a similar vein, `sbp('chelonia/rootState')` is also not well suited for the SW environment, for the same reasons. However, this could later be replaced (in contracts) with code that represents the running environment. For example, when a contract is loaded, Chelonia could give it certain 'globals' with state (e.g., login information) that contracts can access and would work similarly to accessing the root state / getters today.

frontend/controller/actions/group.js
```diff
@@ -970,7 +911,36 @@ export default (sbp('sbp/selectors/register', {
...
+    }
+
+    return sendMessage({
+      ...params,
+      data: {
+        ...data,
+        passPayload
+      }
+    })
+  }),
```
taoeffect, line 947: Why was this moved here?

corrideat: Because none of this belongs into a `.vue` file. This is something that affects Chelonia contract state and the pass payload should be determined by an action. (Ideally, it'd be determined by a contract, but that's harder to do at the moment because an invite needs to be created)

frontend/views/containers/access/LoginModal.vue
```diff
@@ -27,7 +27,7 @@ export default ({
   methods: {
     submit () {
       this.$refs.modal.close()
-      this.$router.push({ query: this.$route.query, path: this.$route.query.next ?? '/dashboard' }).catch(() => {})
+      this.$router.push({ query: this.$route.query, path: this.$route.query.next ?? '/' }).catch(() => {})
```
taoeffect, line 30: Why was this changed?

shared/domains/chelonia/localSelectors.js
```diff
@@ -0,0 +1,82 @@
...
+
+      const oldContracts = Object.keys(vuexState.contracts)
+      const oldContractsToRemove = oldContracts.filter(x => !subscriptionSet.includes(x))
+      const newContracts = subscriptionSet.filter(x => !oldContracts.includes(x))
+
+      oldContractsToRemove.forEach(contractID => {
+        reactiveDel(vuexState.contracts, contractID)
+        reactiveDel(vuexState, contractID)
+      })
+      newContracts.forEach(contractID => {
```
taoeffect: Minor nit: `forEach` is less efficient than a `for-of` loop. I agree it's easier to read, but we're already copying state so every microsecond counts 😄

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
...
+    if (!cheloniaState) return
+    const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
+    if (!identityContractID) return
+    Object.assign(sbp('chelonia/rootState'), cheloniaState)
+  })
+
+  const saveChelonia = () => sbp('okTurtles.eventQueue/queueEvent', 'CHELONIA_STATE', () => {
+    return sbp('gi.db/settings/save', 'CHELONIA_STATE', sbp('chelonia/rootState'))
+  })
+  const saveCheloniaDebounced = debounce(saveChelonia, 200)
```
taoeffect: There is already a `debouncedSave` in `state.js`... how does this relate to it, and perhaps maybe this code should be moved to `state.js`?

frontend/main.js
```diff
@@ -175,6 +205,13 @@ async function startApp () {
     }
   })
 
+  sbp('chelonia/externalStateSetup', { reactiveSet: Vue.set, reactiveDel: Vue.delete })
```
taoeffect: Shouldn't this be done before we setup the web socket via `chelonia/configure`? Just to be safe I would move it there.

frontend/main.js
corrideat: This is for Chelonia state specifically. With this PR, the state is broken down into Vuex state (that's `state.js`) and Chelonia state (that will later be in the SW).

This affects logging in, for example. When logging in, cheloniaState is saved in the SW and then Vuex state is broadcasted to all tabs.

frontend/views/containers/access/LoginModal.vue
corrideat: I think both should work, but `/` always has the same number of redirects. That's why I changed it here.

frontend/main.js
corrideat: Not necessarily, because there's no state to sync yet. But it could be done.

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
...
+        saveCheloniaDebounced()
+      }
+    },
+    reactiveDel: (o: Object, k: string) => {
+      // Simulate SW environment
+      if (has(o, k)) {
+        delete o[k]
+        saveCheloniaDebounced()
+      }
+    },
```
taoeffect: So these are simulating sending messages back to the browser tabs with the updated state?

If so (or if otherwise), could you add a comment above line 126 explaining this?

frontend/main.js
taoeffect: I would move it there just to be safe

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+          // in
+          await saveChelonia()
+          sbp('okTurtles.events/emit', LOGIN_COMPLETE, { identityContractID })
+        } catch (e) {
+          sbp('okTurtles.events/emit', LOGIN_ERROR, { identityContractID, error: e })
+        }
       })
       sbp('okTurtles.events/on', LOGOUT, () => {
+        const state = sbp('state/vuex/state')
+        if (!state.loggedIn) return
```
taoeffect: Any way to move these additions someplace outside of `main.js`?

frontend/main.js
corrideat: These would go into the SW. They're not simulating messages, they're just saving Chelonia state.

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
     // Since a runtime error just occured, we likely want to persist app logs to local storage now.
     sbp('appLogs/save')
   }
+
+  await sbp('gi.db/settings/load', 'CHELONIA_STATE').then(async (cheloniaState) => {
+    // TODO: PLACEHOLDER TO SIMULATE CHELONIA IN A SW
+    if (!cheloniaState) return
+    const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
+    if (!identityContractID) return
```
taoeffect, line 149: What does loading the current userID have to do with loading the chelonia state? Could you please explain this in a comment?

corrideat: There isn't (should not be) any meaningful Chelonia state unless there's a logged in user. I.e., opening the page should not restore any state unless there's an active session

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
...
+    return sbp('gi.db/settings/save', 'CHELONIA_STATE', sbp('chelonia/rootState'))
+  })
+  const saveCheloniaDebounced = debounce(saveChelonia, 200)
+
   await sbp('chelonia/configure', {
     connectionURL: sbp('okTurtles.data/get', 'API_URL'),
-    stateSelector: 'state/vuex/state',
-    reactiveSet: Vue.set,
-    reactiveDel: Vue.delete,
+    // stateSelector: 'state/vuex/state',
```
taoeffect: If the state selector is no longer needed could you add a comment explaining why?

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
...
+    reactiveSet: (o: Object, k: string, v: string) => {
+      // Simulate SW environment
+      if (o[k] !== v) {
+        o[k] = v
+        saveCheloniaDebounced()
+      }
+    },
+    reactiveDel: (o: Object, k: string) => {
+      // Simulate SW environment
+      if (has(o, k)) {
```
taoeffect, line 176: I think this `if (has..` thing can be removed. No need for an extra branch here

frontend/main.js
```diff
@@ -103,19 +106,45 @@ async function startApp () {
...
+
   await sbp('chelonia/configure', {
     connectionURL: sbp('okTurtles.data/get', 'API_URL'),
-    stateSelector: 'state/vuex/state',
-    reactiveSet: Vue.set,
-    reactiveDel: Vue.delete,
+    // stateSelector: 'state/vuex/state',
+    reactiveSet: (o: Object, k: string, v: string) => {
+      // Simulate SW environment
+      if (o[k] !== v) {
```
taoeffect, line 170: I think this `if (o[k]..` thing can be removed, no need for an extra branch.

frontend/main.js
```diff
@@ -175,6 +205,13 @@ async function startApp () {
     }
   })
 
+  sbp('chelonia/externalStateSetup', { reactiveSet: Vue.set, reactiveDel: Vue.delete })
+
+  /* sbp('okTurtles.events/on', NAMESPACE_REGISTRATION, ({ name, value }) => {
+    const cache = sbp('state/vuex/state').namespaceLookups
+    Vue.set(cache, name, value)
+  }) */
```
taoeffect, line 258: Is this code important? If so, it needs to be uncommented and placed in the right file. If not, should be cleaned up / removed.

frontend/main.js
corrideat: That's checking whether the value actually changed. It should be much cheaper to do than to serialise and save the state.

frontend/main.js
corrideat: Same as for `reactiveSet`. This is for performance reasons.

taoeffect: If I had to guess, I'd bet that in most cases the call to `reactiveDel` deletes an actual key-value mapping. If that's true, then in most cases this check is unnecessary and therefore serves to slow things down unnecessarily.

frontend/main.js
taoeffect: If I had to guess, I'd bet that in most cases the call to `reactiveSet` does result in an actual update to the key-value mapping. If that's true, then in most cases this check is unnecessary and therefore serves to slow things down unnecessarily.

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+        ]).catch(e => {
+          console.error('Logout event: error deleting Chelonia state')
+        })
+      })
+      sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams, state }) => {
+        try {
+          const vuexState = sbp('state/vuex/state')
+          if (vuexState.loggedIn && vuexState.loggedIn.identityContractID !== identityContractID) {
+            throw new Error('Received login event but there already is an active session')
+          }
```
taoeffect: What's going on here? Definitely needs a comment explaining how this could happen, what this situation means, etc.

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+          const cheloniaState = cloneDeep(await sbp('chelonia/rootState'))
+          if (state) {
+          // TODO Do this in a cleaner way
+          // Exclude contracts from the state
+            Object.keys(state).forEach(k => {
+              if (k.startsWith('z9br')) {
+                delete state[k]
```
taoeffect: Should this be `Vue.delete`?

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+      })
+      sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams, state }) => {
+        try {
+          const vuexState = sbp('state/vuex/state')
+          if (vuexState.loggedIn && vuexState.loggedIn.identityContractID !== identityContractID) {
+            throw new Error('Received login event but there already is an active session')
+          }
+          const cheloniaState = cloneDeep(await sbp('chelonia/rootState'))
+          if (state) {
+          // TODO Do this in a cleaner way
```
taoeffect: Instead of checking `k.startsWith('z9br')`, you could probably check `Object.keys(state.contracts)`, which should give you the same values.

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+          // Exclude contracts from the state
+            Object.keys(state).forEach(k => {
+              if (k.startsWith('z9br')) {
+                delete state[k]
+              }
+            })
+            Object.keys(cheloniaState.contracts).forEach(k => {
+              if (cheloniaState[k]) {
+                state[k] = cheloniaState[k]
```
taoeffect: `Vue.set`?

frontend/main.js
```diff
@@ -321,34 +365,126 @@ async function startApp () {
...
+          console.error('Logout event: error deleting Chelonia state')
+        })
+      })
+      sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams, state }) => {
+        try {
+          const vuexState = sbp('state/vuex/state')
+          if (vuexState.loggedIn && vuexState.loggedIn.identityContractID !== identityContractID) {
+            throw new Error('Received login event but there already is an active session')
+          }
+          const cheloniaState = cloneDeep(await sbp('chelonia/rootState'))
```
taoeffect: Above you grab `const vuexState = sbp('state/vuex/state')`. And here you grab `await sbp('chelonia/rootState')`.

Are these different? If so, please add a comment explaining the difference. If not, then one of these can be removed.

frontend/main.js
corrideat: It's not needed because it's used to replace the state.

frontend/main.js
corrideat: It shouldn't happen, hence the error.

frontend/main.js
corrideat: It should not

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+      delete state.cheloniaState
+
+      return { encryptionParams, state, cheloniaState }
+    } else {
+      return { encryptionParams, state, cheloniaState: null }
+    }
+  } else {
+    const state = await sbp('gi.db/settings/load', identityContractID)
+
+    return { encryptionParams: null, state, cheloniaState: null }
```
taoeffect: Could you add a comment explaining why sometimes the `cheloniaState` is stored in the settings and sometimes it's not?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+      try {
+        const loginCompletePromise = new Promise((resolve, reject) => {
+          const loginCompleteHandler = ({ identityContractID: id }) => {
+            sbp('okTurtles.events/off', LOGIN_ERROR, loginErrorHandler)
+            if (id === identityContractID) {
+              // We need to save the state to ensure it's possible to refresh
+              // the page
+              resolve(sbp('state/vuex/save'))
+            } else {
+              reject(new Error('Unexpected identity contract ID'))
```
taoeffect: The error message should say what the mismatch was. E.g.:

```js
new Error(`Identity contract ID mismatch during login: ${identityContractID} != ${id}`)
```

And better yet, it should try to retrieve the corresponding username (if it's available) from the state and show that to the user too.

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+            }
+          }
+          const loginErrorHandler = ({ identityContractID: id, error }) => {
+            sbp('okTurtles.events/off', LOGIN_COMPLETE, loginCompleteHandler)
+            if (id === identityContractID) {
+              reject(error)
+            } else {
+              reject(new Error('Unexpected identity contract ID'))
```
taoeffect: The error message should say what the mismatch was. E.g.:

```js
new Error(`Identity contract ID mismatch during login: ${identityContractID} != ${id}`)
```

And better yet, it should try to retrieve the corresponding username (if it's available) from the state and show that to the user too.

frontend/controller/actions/README.md
```diff
@@ -0,0 +1,12 @@
...
+The `gi.actions` domain is used for operations that
+_execute in the same context as Chelonia_, which may be a _browsing context_
+(e.g., a tab window), but it could also be something like a service worker.
+**Operations that affect the overall application state, regardless of the UI state in a specific tab or window genrally belong under `gi.actions`**.
+
+The operations under this domain, `gi.actions`, will usually be called from
+contracts, other selectors under `gi.actions` and other selectors under `gi.app`.
+
+For an explanation about `gi.app` and a worked example of where a selector
+belongs, see the [`README.md` file there](../app/README.md).
```
taoeffect: Will calls to the selectors in this folder be forwarded to the service worker using the SBP star-selector?

If so, could you please add a paragraph here explaining that the entire `gi.actions/*` domain is forwarded?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+      let loginCompleteHandler, loginErrorHandler
+
+      try {
+        const loginCompletePromise = new Promise((resolve, reject) => {
+          const loginCompleteHandler = ({ identityContractID: id }) => {
+            sbp('okTurtles.events/off', LOGIN_ERROR, loginErrorHandler)
+            if (id === identityContractID) {
+              // We need to save the state to ensure it's possible to refresh
+              // the page
+              resolve(sbp('state/vuex/save'))
```
taoeffect: Passing in `sbp('state/vuex/save')` to the resolved promise doesn't seem to do anything... so either a `// TODO` comment is needed or this needs to be rethought.

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+  if (password) {
+    const stateKeyEncryptionKeyFn = (stateEncryptionKeyId, salt) => {
+      return deriveKeyFromPassword(CURVE25519XSALSA20POLY1305, password, salt + stateEncryptionKeyId)
+    }
+
+    const { encryptionParams, value: state } = await sbp('gi.db/settings/loadEncrypted', identityContractID, stateKeyEncryptionKeyFn)
+
+    if (state) {
+      const cheloniaState = state.cheloniaState
+      delete state.cheloniaState
```
taoeffect, line 29: Where in this project is it clearly explained/defined what is inside of `cheloniaState`?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+const loadState = async (identityContractID: string, password: ?string) => {
+  if (password) {
+    const stateKeyEncryptionKeyFn = (stateEncryptionKeyId, salt) => {
+      return deriveKeyFromPassword(CURVE25519XSALSA20POLY1305, password, salt + stateEncryptionKeyId)
+    }
+
+    const { encryptionParams, value: state } = await sbp('gi.db/settings/loadEncrypted', identityContractID, stateKeyEncryptionKeyFn)
+
+    if (state) {
+      const cheloniaState = state.cheloniaState
```
taoeffect, line 28: Does `cheloniaState` have a copy of all of the contract states?

Does `state` also have a copy of all of the contract states?

frontend/controller/app/identity.js
corrideat: Yes, this does something. `state/vuex/save` is an async function, so it'll make the promise continue to wait until that resolves.

frontend/controller/app/identity.js
corrideat: Yes and yes.

shared/domains/chelonia/chelonia.js
```diff
@@ -343,30 +353,36 @@ export default (sbp('sbp/selectors/register', {
...
+    this.config.reactiveSet(rootState, 'contracts', Object.create(null))
     clearObject(this.ephemeralReferenceCount)
     this.pending.splice(0)
     clearObject(this.currentSyncs)
     clearObject(this.postSyncOperations)
     clearObject(this.sideEffectStacks)
     this.subscriptionSet.clear()
     sbp('chelonia/clearTransientSecretKeys')
-    sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, this.subscriptionSet)
+    sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, Array.from(this.subscriptionSet))
```
taoeffect, line 373: I think it might make more sense to do this at the very end of this function, no? After the `newState` is set.

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,252 @@
...
+              reject(new Error('Unexpected identity contract ID'))
+            }
+          }
+
+          sbp('okTurtles.events/once', LOGIN_COMPLETE, loginCompleteHandler)
+          sbp('okTurtles.events/once', LOGIN_ERROR, loginErrorHandler)
+        })
+
+        if (password) {
+          await sbp('gi.actions/identity/login', { identityContractID, encryptionParams, cheloniaState, state, transientSecretKeys: transientSecretKeys.map(k => new Secret(serializeKey(k, true))) })
```
taoeffect, line 316: What are these two different states that are being sent to the service worker, `state` and `cheloniaState`? Could you please add a comment here explaining their contents?

shared/domains/chelonia/chelonia.js
corrideat: Not really, because the subscription set is empty.

frontend/controller/app/identity.js
corrideat: (However, only the `cheloniaState` copy is to be used)

frontend/controller/actions/identity.js
```diff
@@ -359,41 +283,13 @@ export default (sbp('sbp/selectors/register', {
...
-        }
-      }
-
       await sbp('gi.actions/identity/kv/load')
+      // NOTE: update chatRoomUnreadMessages to the latest one we do this here
+      //       just after the identity contract is synced because
+      //       while syncing the chatroom contract it could be necessary to update chatRoomUnreadMessages
+      await sbp('gi.actions/identity/kv/loadChatRoomUnreadMessages')
+      // NOTE: load users preferences config which is saved in KV store
+      await sbp('gi.actions/identity/kv/loadPreferences')
```
taoeffect: These selectors are already called by line 286 above so should be removed here

corrideat: That's right.

frontend/controller/app/identity.js
corrideat: Done

frontend/controller/app/identity.js
corrideat: The various states are now better explained.

frontend/controller/app/identity.js
corrideat: It wasn't, but I've now added comments explaining it.

frontend/controller/actions/README.md
corrideat: Done

frontend/controller/app/identity.js
corrideat: Updated

frontend/controller/app/identity.js
taoeffect: Could you add a comment stating that somewhere (maybe here) and also just some of the backstory on why this is the case? (This is similar to my other comment above (or below, depending on where you're reading this on Github)

frontend/controller/app/identity.js
corrideat: Done

frontend/controller/app/identity.js
taoeffect: OK, could you add a comment stating that and also explaining why you're doing this?

frontend/main.js
corrideat: Yes, they are.

frontend/main.js
corrideat: Added comment above  it

frontend/main.js
corrideat: > in most cases 
Serialising and saving is likely several orders of magnitude slower than this check. So, it's probably worth it even if 99.9% of the time it's true.

frontend/main.js
corrideat: Explained

frontend/main.js
corrideat: Done

shared/domains/chelonia/chelonia.js
corrideat: Done

shared/serdes/index.js
corrideat: Added it back.

test/cypress/support/index.js
corrideat: Documented it.

frontend/controller/serviceworkers/sw-primary.js
corrideat: Done

frontend/main.js
corrideat: Moved.

frontend/main.js
corrideat: Uncommented

shared/domains/chelonia/localSelectors.js
corrideat: Done

frontend/main.js
corrideat: Moved

frontend/main.js
```diff
@@ -405,9 +485,31 @@ async function startApp () {
...
+          return index === -1 ? contractSyncPriorityList.length : index
+        }
+        await sbp('chelonia/contract/sync', identityContractID, { force: true })
+        const contractIDs = Object.keys(cheloniaState.contracts)
+        await Promise.all(Object.entries(contractIDs).sort(([a], [b]) => {
+          // Sync contracts in order based on type
+          return getContractSyncPriority(a) - getContractSyncPriority(b)
+        }).map(([, ids]) => {
+          return sbp('okTurtles.eventQueue/queueEvent', `appStart:${identityContractID ?? '(null)'}`, ['chelonia/contract/sync', ids, { force: true }])
+        }))
```
taoeffect: All of this code appears to be a DRY violation from duplication in `'gi.actions/identity/login'`, is it possible to consolidate it to one place?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,376 @@
...
+  //   or new Vuex state: any Chelonia-specific state will be set directly from
+  //   `cheloniaState` and any exisiting contract state in `state` or `vuexState`
+  //   will be discarded.
+  try {
+    const vuexState = sbp('state/vuex/state')
+    if (vuexState.loggedIn && vuexState.loggedIn.identityContractID !== identityContractID) {
+      // This shouldn't happen. It means that we received a LOGIN event but
+      // there's an active session for a different user. If this happens, it
+      // means that there's buggy login logic that should be reported and fixed
+      throw new Error('Received login event but there already is an active session')
```
taoeffect: It's important for us to catch such buggy logic if it exists, but we'll miss it in the logs because there's no corresponding `console.error` line above line 65 here. Could you please add it and also log the two different conflicting sessions (and any other useful info)?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,376 @@
...
+    }
+    const cheloniaState = cloneDeep(await sbp('chelonia/rootState'))
+    // If `state` is set, process it and replace Vuex state with it
+    if (state) {
+      // Exclude contracts from the state
+      if (state.contracts) {
+        Object.keys(state.contracts).forEach(k => {
+          // Vue.delete not needed as the entire object will replace the state
+          delete state[k]
```
taoeffect, line 76: ~~Should this be `delete vuexState[k]`?~~

EDIT: nm...

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,376 @@
...
+    // managed separately
+    return { encryptionParams: null, state, cheloniaState: null }
+  }
+}
+
+sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams, state }) => {
+  // This function restores the state after a successful login. It's complex
+  // because it deals with _three_ different states.
+  // * `state` (input parameter): (optional). This is the _saved_ Vuex state
+  //   that we'll use to replace the current Vuex state (`vuexState`).
```
taoeffect: Could you further add commentary here on when `state` is passed in and when it isn't?

frontend/controller/app/identity.js
corrideat: No, I mean what's written.

frontend/controller/app/identity.js
corrideat: There already is a comment, right in the part you quoted.

taoeffect: Are you referring to this?

```js
              // We need to save the state to ensure it's possible to refresh
              // the page
```

If so, that comment doesn't help me understanding why `sbp('state/vuex/save')` is being called here.

The line below (314) does nothing with this value:

```js
await loginCompletePromise
```

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,376 @@
...
+    // removed, as it is only used for recovering the state when logging in
+    sbp('gi.db/settings/deleteEncrypted', identityContractID).catch(e => {
+      console.error('Error deleting encrypted settings after login')
+    })
+
+    // TODO: [SW] It may make more sense to load persistent actions in
+    // actions in the SW instead of on each tab
+    const databaseKey = `chelonia/persistentActions/${identityContractID}`
+    sbp('chelonia.persistentActions/configure', { databaseKey })
+    await sbp('chelonia.persistentActions/load')
```
taoeffect, line 142: Agreed, feel free to move this to `actions/identity.js`

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,376 @@
...
+      if (!state.loggedIn) return
+
+      const cheloniaState = await sbp('gi.actions/identity/logout')
+
+      const { encryptionParams } = state.loggedIn
+      if (encryptionParams) {
+        // If we're logging out, save the current Chelonia state under the
+        // `.cheloniaState1 key. This will be used later when logging in
+        // to restore both the Vuex and Chelonia states
+        state.cheloniaState = cheloniaState
```
taoeffect: Typo here: `.cheloniaState1 key.`

frontend/controller/app/identity.js
corrideat: save the state => sbp('state/vuex/save')
The line there does something with that value, namely waiting.

frontend/main.js
```diff
@@ -103,19 +106,60 @@ async function startApp () {
...
+  // Load Chelonia state (this needs to be done in the SW when Chelonia is
+  // running there)
+  // We only load Chelonia state when SETTING_CURRENT_USER is set because,
+  // currently, Chelonia only has meaningful persistent state when there's
+  // an active session. If there's no logged in user, it means that there's
+  // no state to restore.
+  await sbp('gi.db/settings/load', 'CHELONIA_STATE').then(async (cheloniaState) => {
+    if (!cheloniaState) return
+    const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
+    if (!identityContractID) return
```
taoeffect: Sorry, I still don't get this... What about all of the code in `gi.app/identity/login` (and related selectors and events)? Isn't that supposed to be loading the Chelonia state? What then is this code all about?

Could you please explain in a more elaborate comment?

corrideat: What you are referring to is loading Chelonia state during _login_. This is loading Chelonia state when there already is an active session.

taoeffect: But `gi.app/identity/login` is called on every page load, including when there is an active session, is it not?

shared/domains/chelonia/localSelectors.js
```diff
@@ -0,0 +1,83 @@
...
+      const oldContracts = Object.keys(vuexState.contracts)
+      const oldContractsToRemove = oldContracts.filter(x => !subscriptionSet.includes(x))
+      const newContracts = subscriptionSet.filter(x => !oldContracts.includes(x))
+
+      oldContractsToRemove.forEach(contractID => {
+        reactiveDel(vuexState.contracts, contractID)
+        reactiveDel(vuexState, contractID)
+      })
+      for (let i = 0; i < newContracts.length; i++) {
+        const contractID = newContracts[i]
```
taoeffect: Nit: for-of loop is easier to write and read and just as fast (pretty sure)

shared/domains/chelonia/localSelectors.js
```diff
@@ -0,0 +1,83 @@
...
+  // 3. Each tab calls this selector once to set up event listeners on EVENT_HANDLED
+  //    and CONTRACTS_MODIFIED, which will keep each tab's state updated every
+  //    time Chelonia handles an event.
+  'chelonia/externalStateSetup': ({ reactiveSet, reactiveDel } = {
+    reactiveSet: Reflect.set.bind(Reflect),
+    reactiveDel: Reflect.deleteProperty.bind(Reflect)
+  }) => {
+    sbp('okTurtles.events/on', EVENT_HANDLED, async (contractID, message) => {
+      const { contractState, cheloniaState } = await sbp('chelonia/contract/fullState', contractID)
+      const vuexState = sbp('state/vuex/state')
```
taoeffect: This seems Group Income specific, yet it is being done inside of the Chelonia library.

Is ether a way to either do this in an app-agnostic way, or to move this elsewhere?

shared/domains/chelonia/localSelectors.js
```diff
@@ -0,0 +1,83 @@
...
+      } else {
+        reactiveDel(vuexState, contractID)
+      }
+
+      // This EVENT_HANDLED_READY event lets the current context (e.g., tab)
+      // know that an event has been processed _and_ committed to the state
+      // (as opposed to EVENT_HANDLED, which means the event was processed by
+      // _Chelonia_ but state changes may not be reflected in the current tab
+      // yet).
+      sbp('okTurtles.events/emit', EVENT_HANDLED_READY, contractID, message)
```
taoeffect, line 53: Excellent comment! 👍

frontend/main.js
corrideat: Yes, but it doesn't load Chelonia state in that case

shared/domains/chelonia/chelonia.js
```diff
@@ -798,6 +829,14 @@ export default (sbp('sbp/selectors/register', {
...
 
       return sbp('chelonia/private/queueEvent', contractID, () => {
+        // This allows us to double-check that the contract is meant to be
+        // removed, as circumstances could have changed from the time remove
+        // was called and this function is executed. For example, `/release`
+        // makes a synchronous check, but processing of other events since
+        // require this to be re-checked (in this case, for reference counts).
+        if (confirmRemovalCallback && !confirmRemovalCallback(contractID)) {
+          return
+        }
```
taoeffect, line 846: This `confirmRemovalCallback` doesn't seem to be used. Do you have a concrete plan to use it soon? If not then maybe it should be removed until it's needed?

corrideat: What do you mean it's not used? It's used by release

shared/domains/chelonia/localSelectors.js
corrideat: Thanks

shared/domains/chelonia/localSelectors.js
corrideat: Made more generic

frontend/main.js
corrideat: Expanded comment

frontend/controller/app/identity.js
corrideat: Fixed

frontend/controller/app/identity.js
corrideat: It doesn't need to be moved now

frontend/main.js
taoeffect: Can this code be moved into `controller/app/identity.js` then?

corrideat: No, it cannot. See where the comment starts, it says this needs to be in the service worker. Chelonia state is part of the service worker and there is a single one. This state cannot be set every time there is a page refresh.

taoeffect: OK, on a somewhat related note, I'm starting to think that this logic is important and common enough that it would need to be done by every Chelonia-based app, right?

I've created an issue related to this: https://github.com/okTurtles/group-income/issues/2162

Maybe we can get this done for this PR too.

corrideat: It is common logic in the sense that restoring the state may be needed, but the specifics of how it's done depend on the application. For example, here it's checking `SETTING_CURRENT_USER` which would not always exist. On the server, the state is partitioned by contracts, which is not relevant here.

Also, I don't see the connection between this and #2162. Using `chelonia.db` or `chelonia/db` would not address this because it's the state for Chelonia itself that's being loaded. In the client, when the lightweight client option is enabled, `chelonia.db` is mostly a no-op, so it wouldn't work here.

frontend/main.js
```diff
@@ -103,19 +107,73 @@ async function startApp () {
...
+  //   - Completely fresh session (no saved state): fresh state (not loaded from
+  //     anywhere).
+  //   - Fresh session with saved state: /login logic, CHELONIA_STATE is
+  //     decrypyted from the saved state and loaded. This is also saved in
+  //     CHELONIA_STATE so that refreshing the page works.
+  await sbp('gi.db/settings/load', 'CHELONIA_STATE').then(async (cheloniaState) => {
+    if (!cheloniaState) return
+    const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
+    if (!identityContractID) return
+    Object.assign(sbp('chelonia/rootState'), cheloniaState)
```
taoeffect: Reminder (to not forget) to update this to `chelonia/reset`

shared/domains/chelonia/chelonia.js
taoeffect: Oops, you're right. Sorry for some reason I thought `confirmRemovalCallback` was a named parameter and couldn't find where it was used.

frontend/controller/app/identity.js
taoeffect: OK, sorry, the problem here was that my brain was being momentarily dyslexic or something, I think. It was so used to reading `'state/vuex/state'`, that I didn't see this was a totally different selector called `'state/vuex/save'`. I thought you were passing in `'state/vuex/state'`.

frontend/controller/actions/utils.js
```diff
@@ -313,3 +305,48 @@ export async function createInvite ({ quantity = 1, creatorID, expires, invitee
...
+      if (!contractIDs[type]) {
+        contractIDs[type] = []
+      }
+      contractIDs[type].push(id)
+    })
+  }
+  return contractIDs
+}
+
+export async function syncContractsInOrder (identityContractID: string | null, queuePrefix: string, groupedContractIDs: Object): Promise<any> {
```
taoeffect: This `queuePrefix` seems entirely unnecessary. Can it be removed?

corrideat: Why is it unnecessary? It's used to select the queue name, as this can be called when starting the app or when logging in (with a password).

frontend/controller/actions/chatroom.js
```diff
@@ -178,11 +178,11 @@ export default (sbp('sbp/selectors/register', {
     const originatingContractID = state.attributes.groupContractID ? state.attributes.groupContractID : contractID
 
     // $FlowFixMe
-    return Promise.all(Object.keys(state.members).map((pContractID) => {
-      const CEKid = findKeyIdByName(rootState[pContractID], 'cek')
+    return Promise.all(Object.keys(state.members).map(async (pContractID) => {
+      const CEKid = await sbp('chelonia/contract/currentKeyIdByName', pContractID, 'cek')
```
taoeffect, line 182: Why is this calling `await`?

corrideat: Because of sandboxing in the future. All `chelonia/` calls may be async.

shared/domains/chelonia/localSelectors.js
```diff
@@ -0,0 +1,86 @@
...
+      for (const contractID of newContracts) {
+        const { contractState, cheloniaState } = states[contractID]
+        if (cheloniaState) {
+          reactiveSet(vuexState.contracts, contractID, cloneDeep(cheloniaState))
+        }
+        if (contractState) {
+          reactiveSet(vuexState, contractID, cloneDeep(contractState))
+        }
+      }
+    })
```
taoeffect: Should there be a corresponding `CONTRACTS_MODIFIED_READY`?

corrideat: There could be. It's not being used currently as there's nothing listening for CONTRACTS_MODIFIED in the application.

frontend/controller/actions/utils.js
taoeffect: I see no reason for it to be called on an event queue. What is this protecting from?

`'chelonia/contract/sync'` already uses an event queue.

frontend/controller/actions/identity-kv.js
```diff
@@ -152,26 +153,27 @@ export default (sbp('sbp/selectors/register', {
...
+    return sbp('chelonia/kv/set', identityContractID, KV_KEYS.PREFERENCES, data, {
+      encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
+      signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
       onconflict
     })
   },
   'gi.actions/identity/kv/loadPreferences': () => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const preferences = await sbp('gi.actions/identity/kv/fetchPreferences')
+      // TODO: Can't use state/vuex/commit
```
taoeffect, line 179: Should this be taken care of now?

frontend/controller/actions/utils.js
corrideat: sync might use an event queue, but this is more about initialisation. In particular, the `login:` queue is used to prevent multiple login (logout) events being handled out of order. It's probably not that important for the app init phase though.

frontend/controller/actions/identity.js
```diff
@@ -531,34 +368,35 @@ export default (sbp('sbp/selectors/register', {
...
+        return cheloniaState
       })
     } catch (e) {
       console.error(`${e.name} during logout: ${e.message}`, e)
     }
     // Clear the file cache when logging out to preserve privacy
     sbp('gi.db/filesCache/clear').catch((e) => { console.error('Error clearing file cache', e) })
-    sbp('state/vuex/reset')
     sbp('okTurtles.events/emit', LOGOUT)
-    sbp('appLogs/pauseCapture', { wipeOut: true }) // clear stored logs to prevent someone else accessing sensitve data
```
taoeffect, line 552: This seems to have been deleted not moved anywhere. Why?

frontend/controller/actions/utils.js
taoeffect: How can there be multiple login events at the same time?

frontend/controller/actions/index.js
```diff
@@ -44,7 +44,8 @@ sbp('sbp/selectors/register', {
         throw new Error('Missing CEK; unable to proceed sharing keys')
       }
 
-      const secretKeys = sbp('state/vuex/state')['secretKeys']
+      // TODO: Use 'chelonia/haveSecretKey'
+      const secretKeys = await sbp('chelonia/rootState')['secretKeys']
```
taoeffect, line 48: Should this be done now?

frontend/controller/app/identity.js
```diff
@@ -0,0 +1,379 @@
...
+    sbp('okTurtles.events/emit', LOGIN_ERROR, { identityContractID, error: e })
+  }
+})
+
+sbp('okTurtles.events/on', LOGOUT, (a) => {
+  // TODO: [SW] It may make more sense to load persistent actions in
+  // actions in the SW instead of on each tab
+  sbp('chelonia.persistentActions/unload')
+})
+
```
taoeffect: We are not using persistent actions at all right now, so I think all persistent action related code should be commented out (including the imports, to reduce bundle size).

frontend/main.js
```diff
@@ -272,7 +339,14 @@ async function startApp () {
   }
 
   // register service-worker
-  await sbp('service-workers/setup')
+  await Promise.race(
+    [sbp('service-workers/setup'),
+      new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error('Timed out setting up service worker'))
```
taoeffect, line 361: This should probably be a user-facing message that's wrapped in an `L` function and displayed to the user in an `alert`

frontend/main.js
```diff
@@ -321,34 +395,54 @@ async function startApp () {
...
+        Promise.all([
+          sbp('chelonia/reset'),
+          sbp('gi.db/settings/delete', 'CHELONIA_STATE')
+        ]).catch(e => {
+          console.error('Logout event: error deleting Chelonia state')
+        })
+      })
+      sbp('okTurtles.events/on', LOGIN_COMPLETE, (a) => {
+        const state = sbp('state/vuex/state')
+        if (!state.loggedIn) return
```
taoeffect: Should probably add a `console.warn` message here because this is quite unexpected

frontend/controller/actions/utils.js
corrideat: sbp('gi.app/identity/login', ...)
sbp('gi.app/identity/login', ...)
sbp('gi.app/identity/login', ...)

frontend/controller/actions/identity-kv.js
corrideat: It's an involved change

frontend/main.js
```diff
@@ -321,34 +395,54 @@ async function startApp () {
...
       })
       sbp('okTurtles.events/once', LOGIN_ERROR, () => {
         // Remove the loading animation that sits on top of the Vue app, so that users can properly interact with the app for a follow-up action.
         this.removeLoadingAnimation()
       })
-      sbp('okTurtles.events/on', SWITCH_GROUP, () => {
+      sbp('okTurtles.events/on', SWITCH_GROUP, ({ contractID, isNewlyCreated }) => {
         this.initOrResetPeriodicNotifications()
         this.checkAndEmitOneTimeNotifications()
+        sbp('state/vuex/commit', 'setCurrentGroupId', { contractID, isNewlyCreated })
```
taoeffect: This seems like it should be done inside of `'gi.app/group/switch'`, not here

frontend/controller/actions/utils.js
taoeffect: 1. That does not answer my question, which is about how that can happen.
2. It's unclear to me why there are two separate queues for this, `appStart` and `login`, when you could just have one queue inside of `syncContractsInOrder` (no need for parameter)
3. It's not clear why even a single queue would be needed, since `'chelonia/contract/sync'` already uses a queue

So this seems like unnecessary complexity that should be removed.

frontend/model/contracts/chatroom.js
```diff
@@ -129,26 +130,7 @@ sbp('chelonia/defineContract', {
       }
     }
   },
-  getters: {
-    currentChatRoomState (state) {
-      return state
-    },
```
taoeffect: This is being removed and not re-added anywhere. That's a problem because it is still referenced elsewhere in the project.

frontend/model/contracts/chatroom.js
```diff
@@ -129,26 +130,7 @@ sbp('chelonia/defineContract', {
...
-      return getters.currentChatRoomState.members || {}
-    },
-    chatRoomRecentMessages (state, getters) {
-      return getters.currentChatRoomState.messages || []
-    },
-    chatRoomPinnedMessages (state, getters) {
-      return (getters.currentChatRoomState.pinnedMessages || []).sort((a, b) => a.height < b.height ? 1 : -1)
-    }
-  },
+  getters: {},
```
taoeffect: Why were these getters removed?

frontend/model/contracts/group.js
```diff
@@ -1418,14 +1404,33 @@ sbp('chelonia/defineContract', {
...
             chatRoomID: state.generalChatRoomId
           })
         }
+        // If it's the #General chatroom being added, add ourselves to it
+        if (data.chatRoomID === state.generalChatRoomId) {
+          sbp('chelonia/queueInvocation', contractID, () => {
+            const { identityContractID } = sbp('state/vuex/state').loggedIn
+            if (
+              state.profiles?.[identityContractID]?.status === PROFILE_STATUS.ACTIVE &&
+                      state.chatRooms?.[contractID]?.members[identityContractID]?.status !== PROFILE_STATUS.ACTIVE
```
taoeffect: Ideally, there should be no extra indentation here on line 1421. If standardjs cmplains, try putting && on line 1421 at the start.

frontend/model/contracts/chatroom.js
corrideat: Because they've been moved to the chatroom module.

frontend/model/contracts/chatroom.js
corrideat: It's not referenced elsewhere

frontend/views/containers/group-settings/GroupLeaveModal.vue
```diff
@@ -108,9 +108,9 @@ export default ({
       if (this.$v.form.$invalid) { return }
       try {
         const groupContractID = this.currentGroupId
-        await sbp('gi.actions/identity/leaveGroup', {
-          contractID: this.ourIdentityContractId,
-          data: { groupContractID }
+        await sbp('gi.actions/group/removeMember', {
+          contractID: groupContractID,
+          data: {}
```
taoeffect, line 113: Curious, why was this change made? It seems like a significant logic change that is unrelated to the other thematic changes of this PR

frontend/model/contracts/chatroom.js
taoeffect: I see that, but that doesn't explain why that was done.

frontend/model/contracts/chatroom.js
```diff
@@ -396,15 +376,15 @@ sbp('chelonia/defineContract', {
...
           (newMessage.text.includes(mentions.me) || newMessage.text.includes(mentions.all))
 
-        messageReceivePostEffect({
+        await messageReceivePostEffect({
           contractID,
           messageHash: newMessage.hash,
           height: newMessage.height,
           text: newMessage.text,
-          isDMOrMention: isMentionedMe || getters.chatRoomAttributes.type === CHATROOM_TYPES.DIRECT_MESSAGE,
+          isDMOrMention: isMentionedMe || state.settings.type === CHATROOM_TYPES.DIRECT_MESSAGE,
```
taoeffect: The type is stored in `state.attributes` not `state.settings`. This issue is repeated in several places here.

While fixing this, can you please restore the use of the getters that was here since we can now import them from `frontend/model/contracts/shared/getters/chatroom.js`?

frontend/model/contracts/group.js
```diff
@@ -1742,7 +1763,8 @@ sbp('chelonia/defineContract', {
       //   1) automatically switch that user to a 'pledging' member with 0 contribution,
       //   2) pop out the prompt message notifying them of this automatic change,
       //   3) and send 'MINCOME_CHANGED' notification.
-      const myProfile = sbp('state/vuex/getters').ourGroupProfile
+      const identityContractID = sbp('state/vuex/state').loggedIn.identityContractID
+      const myProfile = sbp('chelonia/rootState')[contractID].profiles[identityContractID]
```
taoeffect: How is `'state/vuex/state'` and `'chelonia/rootState'` different here? Can a comment be added to explain this?

frontend/model/contracts/identity.js
```diff
@@ -74,17 +75,7 @@ const checkUsernameConsistency = async (contractID: string, username: string) =>
...
-      return state
-    },
-    loginState (state, getters) {
-      return getters.currentIdentityState.loginState
-    },
-    ourDirectMessages (state, getters) {
-      return getters.currentIdentityState.chatRooms || {}
-    }
-  },
+  getters: {},
```
taoeffect: Plz restore 🙏  (via `frontend/model/contracts/shared/getters/identity.js`)

frontend/model/contracts/group.js
```diff
@@ -1441,6 +1446,15 @@ sbp('chelonia/defineContract', {
           ['gi.contracts/group/releaseDeletedChatRoom', state.chatRooms[data.chatRoomID].members, data.chatRoomID]
         )
         Vue.delete(state.chatRooms, data.chatRoomID)
+      },
+      sideEffect ({ data, contractID, innerSigningContractID }) {
+        sbp('okTurtles.events/emit', DELETED_CHATROOM, { groupContractID: contractID, chatRoomID: data.chatRoomID })
+        const { identityContractID } = sbp('state/vuex/state').loggedIn
```
taoeffect, line 1244: The `DELETED_CHATROOM` listener is defined to also take an `identityContractID`:

```js
{ identityContractID, groupContractID, chatRoomID }
```

So should this include `identityContractID: innerSigningContractID` here?

frontend/model/contracts/shared/getters/group.js
```diff
@@ -0,0 +1,256 @@
...
+} from '../constants.js'
+import currencies from '../currencies.js'
+import { createPaymentInfo, paymentHashesFromPaymentPeriod } from '../functions.js'
+import { PAYMENT_COMPLETED } from '../payments/index.js'
+import { addTimeToDate, dateFromPeriodStamp, dateIsWithinPeriod, dateToPeriodStamp, periodStampsForDate } from '../time.js'
+
+export default ({
+  currentGroupLastLoggedIn () {
+    return {}
+  },
```
taoeffect: This is not a shared getter, note that its definition is different in state.js:

```js
  currentGroupLastLoggedIn (state) {
    return state.lastLoggedIn[state.currentGroupId] || {}
  },
```

So it belongs directly in the contract.

corrideat: fixed

frontend/model/contracts/group.js
corrideat: ~~It's optional, but, yes, I think it could be added.~~

frontend/model/contracts/identity.js
corrideat: done

frontend/views/containers/group-settings/GroupLeaveModal.vue
corrideat: gi.actions/group/removeMember will eventually call `leaveGroup` as contracts are kept in sync. It could have been that this was breaking something too, but I don't remember.

taoeffect:
So the [latest Cypress failure](https://cloud.cypress.io/projects/q6whky/runs/2569/test-results/c362a8e9-7d59-4d3d-9836-af5590dc5c4f/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github) is related to the one of the changes made today, and it could be related to the changes with removing the event queue for login (we could try simply adding a `login-logout` queue).

However the tests did pass with `Feedback #6` commit. Another possibility is the change related to [this change related to `removeMember` on leaving a group](https://github.com/okTurtles/group-income/pull/2137#discussion_r1667736283).

Or something else.

frontend/controller/app/identity.js
```diff
@@ -113,7 +113,7 @@ sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams,
     }
 
     if (encryptionParams) {
-      sbp('state/vuex/commit', 'login', { identityContractID, encryptionParams })
+      sbp('state/vuex/commit', 'APP-LOGIN', { identityContractID, encryptionParams })
```
taoeffect: 👆

frontend/model/contracts/chatroom.js
corrideat: Resolved as this change was reverted

corrideat:
> So the [latest Cypress failure](https://cloud.cypress.io/projects/q6whky/runs/2569/test-results/c362a8e9-7d59-4d3d-9836-af5590dc5c4f/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github) is related to the one of the changes made today, and it could be related to the changes with removing the event queue for login (we could try simply adding a `login-logout` queue).
> 
> However the tests did pass with `Feedback #6` commit. Another possibility is the change related to [this change related to `removeMember` on leaving a group](https://github.com/okTurtles/group-income/pull/2137#discussion_r1667736283).
> 
> Or something else.

This had to do with `/process` now being async. Fixed by making proposals also be async when needed.

frontend/controller/actions/group.js
```diff
@@ -456,6 +450,9 @@ export default (sbp('sbp/selectors/register', {
             await sbp('chelonia/contract/wait', params.contractID)
             await sbp('gi.actions/group/inviteAccept', {
               ...omit(params, ['options', 'action', 'hooks', 'encryptionKeyId', 'signingKeyId']),
+              data: {
+                reference: rootState[userID].groups[params.contractID].hash
+              },
```
taoeffect: Hmm.. Is this a new feature? Or .... why wasn't this already here?

corrideat: New feature, kind of. This is there to make leaving and joining groups work correctly by adding a 'reference'. This way, when, e.g., you leave a group, you only leave if the reference value matches. This prevents an old 'leave' action from making you leave a group.

frontend/controller/actions/identity-kv.js
```diff
@@ -152,26 +153,27 @@ export default (sbp('sbp/selectors/register', {
   },
   // Preferences
   'gi.actions/identity/kv/fetchPreferences': async () => {
-    const { ourIdentityContractId } = sbp('state/vuex/getters')
-    if (!ourIdentityContractId) {
+    const identityContractID = sbp('chelonia/rootState').loggedIn?.identityContractID
+    if (!identityContractID) {
```
taoeffect, line 160: [Elsewhere](https://github.com/okTurtles/group-income/pull/2137#discussion_r1667739696) in the contracts you do:

```js
sbp('state/vuex/state').loggedIn.identityContractID
```

So... why here are you using `sbp('chelonia/rootState')`? Alternatively, why [there](https://github.com/okTurtles/group-income/pull/2137#discussion_r1667739696) are you using `sbp('state/vuex/state')`?

corrideat: Maybe I should be consistent and use `chelonia/rootState`. The reason for this change is as follows:

If contracts are going to be asynchronous, using `chelonia/rootState` will be async (and also, it probably shouldn't be allowed, because it makes sandboxing ineffective). `state/vuex/state` likely won't exist in contracts (when using a SW) and should not be used either.

What I meant to do is to make these instances easy to find so that, in the future, Chelonia can pass globals to contracts when they are loaded. Those globals would be synchronous and would include environment information (e.g., the currently logged in user). By always using sbp('xxx').loggedIn, I ensure those uses are simple to find and replace.

taoeffect: > Maybe I should be consistent and use `chelonia/rootState`

Yes please, consistency is always appreciated 🙏

frontend/controller/actions/group.js
taoeffect: Could you add a comment explaining this somewhere?

frontend/model/contracts/group.js
taoeffect: Based on our conversation about this, could you please add a comment above line 1766 to say something like this:

```js
// using 'state/vuex/state' here instead of 'chelonia/rootState' to fetch the identityContractID because although
// both 'chelonia/rootState' and 'state/vuex/state' are the same at the moment, in the future they will not be, and
// all calls to 'state/vuex/state' will need to be replaced in the future with something else. Using 'state/vuex/state'
// here now makes it easier to find these calls in the future.
```

EDIT: that comment is not quite right, because they are not the same. the Chelonia state is copied to the vuex state, but not the other way around. The `loggedIn` attribute in the chelonia root state is actually set separately, it's not exactly copied from the vuex state, it's set in `'gi.actions/identity/_private/login'`

frontend/controller/actions/group.js
```diff
@@ -238,30 +235,27 @@ export default (sbp('sbp/selectors/register', {
...
+          data: {
+            groupContractID: contractID,
+            inviteSecret: serializeKey(CSK, true),
+            creatorID: true
+          }
+        })
+      })
 
-      return message
+      return message.contractID()
```
taoeffect, line 253: Why was this change made?

frontend/model/contracts/group.js
corrideat: It should not be used when deleting a chatroom. If it's set, it'll skip the handler unless the user deleting it is the same as the currently logged in user.

taoeffect: Then could you please add a comment saying that the `identityContractID` is intentionally left out?

frontend/controller/actions/group.js
corrideat: To avoid having to copy GIMessage across boundaries (a potentially expensive operation)

frontend/controller/actions/identity.js
```diff
@@ -806,6 +643,12 @@ export default (sbp('sbp/selectors/register', {
       throw new Error('Some CIDs could not be deleted')
     }
   },
+  'gi.actions/identity/login': (...params) => {
+    return sbp('okTurtles.eventQueue/queueEvent', 'ACTIONS-LOGIN', ['gi.actions/identity/_private/login', ...params])
+  },
+  'gi.actions/identity/logout': (...params) => {
+    return sbp('okTurtles.eventQueue/queueEvent', 'ACTIONS-LOGIN', ['gi.actions/identity/_private/logout', ...params])
+  },
```
taoeffect, line 651: Can we replace these with anonymous functions pushed onto the event queue?

corrideat: I think it's better this way for logging, as otherwise you don't know when the anonymous function is executed

taoeffect (approved):
Incredible work @corrideat!

PR #2139, opened 2024-06-29
#2130 - Chat rendering issue re handling '<' and '>' character

### LINKED ISSUE #2130: Chat rendering issues
### Problem

Sending:

```
test<test
```

Renders as:

```
test
```

### Solution

Fix. It might have to do with `v-safe-html` and/or `dompurify` or the markdown code.

SebinSong:
closes #2130 

[ proof of fix ]

@taoeffect (cc. @corrideat )
That was a bug caused by caveats in how `dompurify` and `parseFromString` method of the DOMParser API ([reference](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString)) interpret `<'`,`>` and also `&lt;` and `&gt;` characters. So had to make updates to two places to take them into account.

### DIFF
--- a/frontend/views/containers/chatroom/chat-mentions/chat-mentions-utils.js
+++ b/frontend/views/containers/chatroom/chat-mentions/chat-mentions-utils.js
@@ -11,8 +11,13 @@ export function htmlStringToDomObjectTree (htmlString: any): Array<DomObject> {
   // Refernce 2. (Converting html to virtual DOM): https://medium.com/@fulit103/converting-html-to-a-virtual-dom-in-javascript-3a6db0f563b1)
 
   const parser = new DOMParser()
+
+  // Below is a bug-fix for the issue #2130 (https://github.com/okTurtles/group-income/issues/2130)
+  // DOMParser.parseFromString() has some caveats re how it interprets &lt; and &gt;
+  // so manually wrap them with <span> tags
   htmlString = htmlString.replaceAll('&lt;', '<span>&lt;</span>')
     .replaceAll('&gt;', '<span>&gt;</span>')
+
   const doc = parser.parseFromString(htmlString, 'text/html')
   const rootNode = doc.body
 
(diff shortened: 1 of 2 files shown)


SebinSong:
@taoeffect 
Another verification in more complex context:

frontend/views/containers/chatroom/chat-mentions/chat-mentions-utils.js
```diff
@@ -5,12 +5,19 @@ export type DomObject = {
   children?: Array<DomObject>
 }
 
-export function htmlStringToDomObjectTree (htmlString: string): Array<DomObject> {
+export function htmlStringToDomObjectTree (htmlString: any): Array<DomObject> {
```
taoeffect: Why was this updated to `any`? What besides `string` can it be? If it can be one of two types, then maybe it should be changed to a union of two types?

SebinSong: @taoeffect 
I wanted to fix (or avoid) this Flow error.

image

Which actually doesn't make sense as `replaceAll` is [a valid method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) of the string type.

taoeffect: @SebinSong if flow is being dumb then you can add a comment above the broken line like this:

```js
// $FlowFixMe[prop-missing]
```

It's better to get the type information in the parameters right and add a comment like this than use the wrong type information because it could mislead other developers looking at this code into thinking they can pass anything to the function.

taoeffect (changes requested):
One minor change request relating to Flow parameter typing

taoeffect (approved):

frontend/views/containers/chatroom/chat-mentions/chat-mentions-utils.js
snowteamer: @SebinSong 
> Which actually doesn't make sense as replaceAll is [a valid method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) of the string type.

This is true but wasn't the case when this older Flow version (0.154.0) was released

PR #2142, opened 2024-07-01
Fix regressions introduced in PR #2123

corrideat:


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -181,7 +181,8 @@ const onChatScroll = function () {
     if (offsetTop + height <= curScrollBottom) {
       const bottomMessageCreatedHeight = msg.height
       const latestMessageCreatedHeight = this.currentChatRoomReadUntil?.createdHeight
-      if ((!latestMessageCreatedHeight || latestMessageCreatedHeight <= bottomMessageCreatedHeight) && !msg.pending) {
```
corrideat, line 184: Reverted change as this is not needed

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -998,7 +999,9 @@ export default ({
               if (!fromOurselves && isScrollable) {
                 this.updateScroll()
               } else {
-                const msg = this.messages.filter(m => !m.pending).slice(-1)
+                // If there are any temporary messages that do not exist in the
+                // contract, they should not used for updateReadUntilMessageHash
+                const msg = this.messages.filter(m => !m.pending && !m.hasFailed).pop()
```
corrideat, line 1004: `.pop()` is safe to use because it's done on the result of `.filter()`, which is a new array. Since this is looking at `this.messages` in general, there could be pending or failed messages there that need to be excluded.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1049,7 +1052,9 @@ export default ({
             $state.complete()
             if (!this.$refs.conversation ||
             this.$refs.conversation.scrollHeight === this.$refs.conversation.clientHeight) {
-              const msg = this.messages.filter(m => !m.pending).slice(-1)
+              // updateReadUntilMessageHash should only use messages that exist
+              // in the contract
+              const msg = this.messages.filter(m => !m.pending && !m.hasFailed).pop()
```
corrideat, line 1057: As above, `.pop()` doesn't mutate `this.messages` because `.filter()` returns a new array. I think this check here might not be needed as this is mostly processing new events, which should all have incoming direction. However, I decided to keep it to be safe (as the chatroom breaks badly if non-existing messages are added) as this is still operating `this.messages`, which could have previous contents.

taoeffect (dismissed):
LGTM!

Silver-IT (approved):
Nice work, @corrideat!

PR #2147, opened 2024-07-03
Notify proposal state changes in general chatroom

### LINKED ISSUE #2138: Notify all types of proposals in general chatroom
### Problem

We need to notify all types of proposals in general chatroom as we discussed [here](https://github.com/okTurtles/group-income/pull/2046#pullrequestreview-2145965649).

### Solution

Implement it

Silver-IT:
### Summary of Changes

- Closes #2138


frontend/model/contracts/shared/types.js
```diff
@@ -36,14 +35,7 @@ export const chatRoomAttributesType: any = objectOf({
...
   type: unionOf(...Object.values(MESSAGE_TYPES).map(v => literalOf(v))),
   text: string,
-  proposal: objectMaybeOf({
-    proposalId: string,
-    proposalType: string,
-    expires_date_ms: number,
-    createdDate: string,
-    creatorID: string,
-    variant: unionOf([STATUS_EXPIRING].map(v => literalOf(v))) // NOTE: only expiring proposals could be notified at the moment
-  }),
```
taoeffect: I think having this type information is important, as it can help catch bugs in the code.

frontend/views/containers/chatroom/MessageInteractive.vue
```diff
@@ -129,17 +136,17 @@ export default ({
   computed: {
     ...mapGetters(['userDisplayNameFromID']),
     interactiveMessage () {
-      const { variant, creator } = this.proposal
-      const baseOptions = { from: this.userDisplayNameFromID(creator) }
+      const { status, creatorID } = this.proposal
+      const baseOptions = { from: `@${creatorID}` }
```
taoeffect: Should you use that special variable/constant you created instead of `@` literal here?

taoeffect (changes requested):
Nice work @Silver-IT !  Preliminary review.. see comments

Silver-IT (commented):
Ready for the review!
image

Silver-IT:
While I keep fixing several issues found in this PR, I have noticed that we should not create message (to post proposal state changes in general chatroom) inside the group contract sideEffect functions.
That's because the group contracts could be synced in several devices of several users and it will create several messages in general chatroom. Of course, we do the same way when creating notifications and leaving chatroom in the sideEffect of group contracts, but every notification has its `hash` which we use it for removing duplication and every chatroom has `members` which we use it for removing duplicated `/leave` action. But in `messages`, it doesn't.
So, I think I should create messages in the group actions.

Additionally, while I am investing the several issues in this PR
- Calling `processMessage` before we publish event should be removed. I have noticed that that `processMessage` doesn't only call `process` function, but also sideEffect functions which is added by `pushSideEffect` inside `process` function. And those sideEffect funtions will add invocations in several queues and it will do unexpected behavior.
- The way to update contract A by create actions in the sideEffect of contract B has potential risk. We should consider that there could be multiple open sessions for several users, that even a user can have multiple open sessions, and that the sideEffect functions could be called in different time (this means the state of contract A could be different at the time of calling several sideEffect functions of contract B).

frontend/model/contracts/group.js
```diff
@@ -100,20 +100,22 @@ function clearOldPayments ({ contractID, state, getters }) {
...
   }
 
-  sbp('gi.contracts/group/pushSideEffect', contractID,
-    ['gi.contracts/group/archivePayments', contractID, archivingPayments]
-  )
+  if (direction === 'incoming') {
+    sbp('gi.contracts/group/pushSideEffect', contractID,
+      ['gi.contracts/group/archivePayments', contractID, archivingPayments]
+    )
+  }
```
taoeffect: Instead of making all of these changes, it would be simpler to clear the sideEffect stack for outgoing processings

Silver-IT: Yeah, I think so too. This is a temporary solution and I wanted @corrideat to solve this after he would come back.
But, what I really want is to remove calling `processMessage` before publishEvent.

taoeffect: @Silver-IT Actually.... we cannot clear the sideEffect stack... because things are asynchronous and the process function is acting on the `publish:<contractID>` queue, and incoming messages are processed on a different queue `<contractID>`, clearing the sideEffect stack in publishEvent could break the app (if a contract is being processed at the same time a different event is being published).

So this comment that @corrideat made is wrong:

```js
          // Process message to ensure that it is valid. Should this thow,
          // we propagate the error.
          // Because of this, 'chelonia/private/in/processMessage' SHOULD NOT
          // change the global Chelonia state and it MUST NOT call any
          // side-effects or change the global state in a way that affects
          // the meaning of any future messages or successive invocations.
          await sbp('chelonia/private/in/processMessage', entry, cloneDeep(state || {}))
```

In fact, because contracts could push to the sideEffect stack, they do in fact "change the global Chelonia state".

I agree that it would be better if events weren't processed when they were outgoing. However, @corrideat seems to insist it's necessary. When he is back I'd like to go over this with him one more time.

In the meantime it seems that your changes are unfortunately necessary.

taoeffect: OTOH, maybe I'm mistaken, I missed this part above:

```js
        const newEntry = await sbp('chelonia/private/queueEvent', contractID, async () => {
```

This makes it so that the `processMessage` is called on the same queue as incoming events that are being handled. This means that maybe it is possible to clear the `this.sideEffectStacks[contractID]` stack...

taoeffect: I've submitted a PR to clear the sideEffect stack before messages are processed, so sideEffects that are pushed by publishing outgoing messages shouldn't affect the processing of other incoming messages: https://github.com/okTurtles/group-income/pull/2205

Silver-IT: Thanks!!! :+1:

frontend/model/contracts/chatroom.js
```diff
@@ -287,6 +287,7 @@ sbp('chelonia/defineContract', {
         state.members[memberID].leftHeight = height
         delete state.members[memberID]
 
+        if (!state.attributes) return
```
Silver-IT, line 293: These changes in this file is to avoid errors that happens when the group private chatroom is being synced by the group members who are not part of this private chatroom.

frontend/model/contracts/group.js
```diff
@@ -210,7 +209,7 @@ function isActionNewerThanUserJoinedDate (height: number, userProfile: ?Object):
   if (!userProfile) {
     return false
   }
-  return userProfile.joinedHeight < height
+  return userProfile.status === PROFILE_STATUS.ACTIVE && userProfile.joinedHeight < height
```
Silver-IT, line 218: Since we are not remove profile in the state.profiles when the group member leaves group, we should check the profile status in order to some errors not to be happening unexpectedly. Actually, it was happening.

Silver-IT (commented):
Finally, ready for the review!

frontend/model/contracts/chatroom.js
taoeffect: This check doesn't make sense, because `state.members` and `data.memberID` have already been accessed by this point, and if for some reason this chatroom was being synced by members who are not part of it, then those would have already resulted in errors.

So this line should be removed, because the real bug would be why is someone syncing this room when they are not part of it? That's what would need to be fixed.

frontend/model/contracts/chatroom.js
```diff
@@ -341,6 +342,7 @@ sbp('chelonia/defineContract', {
         }
       }),
       process ({ meta }, { state }) {
+        if (!state.attributes) return
```
taoeffect, line 348: This same goes for this check and the others below.

frontend/model/contracts/chatroom.js
```diff
@@ -680,13 +690,7 @@ sbp('chelonia/defineContract', {
         message: object
       })),
       process ({ data, innerSigningContractID }, { state }) {
-        // TODO: remove the below 'if' statement when no older version of contracts are being used
-        if (!state.pinnedMessages) {
-          // NOTE: this is temporary solution for the older version of contracts
-          //       that doesn't have 'pinnedMessages' field which is created in its constructor
-          state.pinnedMessages = []
-        }
-
```
taoeffect, line 691: Nice job remembering this!

frontend/controller/actions/identity-kv.js
```diff
@@ -28,10 +28,13 @@ export default (sbp('sbp/selectors/register', {
...
-      encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
-      signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
-      onconflict
+
+    return sbp('chelonia/queueInvocation', identityContractID, () => {
+      return sbp('chelonia/kv/set', identityContractID, KV_KEYS.UNREAD_MESSAGES, data, {
+        encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
+        signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
+        onconflict
+      })
```
taoeffect: Why is this being placed on the invocation queue? If this is not necessary it must be reverted.

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+        onprocessed: (message) => {
+          (async () => {
+            const proposalId = message.hash()
+            const state = await sbp('chelonia/contract/state', contractID)
+            const proposal = state.proposals[proposalId]
+            const proposalToSend = { ...extractProposalData(proposal, { proposalId }), status: STATUS_OPEN }
+
+            await sbp('gi.actions/group/notifyProposalStateInGeneralChatRoom', { groupID: contractID, proposal: proposalToSend })
+          })()
+        }
```
taoeffect: You don't need to create an async function inside of `onprocessed`. You can write this like so:

```js
onprocessed: async (message) => {
  // ...
```

This also should be rewritten:

```js
const proposalToSend = { ...extractProposalData(proposal, { proposalId }), status: STATUS_OPEN }
```

Like this:

```js
const proposalToSend = extractProposalData(proposal, { proposalId, status: STATUS_OPEN })
```

Finally, you should probably wrap the contents of the event handler in a try/catch block and `console.error` if there's an error.

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+          contractID,
+          invitee: proposal.data.proposalData.memberName,
+          creatorID: proposal.creatorID,
+          expires: state.settings.inviteExpiryProposal
+        })
       }
+
+      proposalToSend = { ...extractProposalData(proposal, { proposalId: proposalHash }), status: STATUS_PASSED }
+    } else if (willBeFailed && isVoteAgainst) {
+      proposalToSend = { ...extractProposalData(proposal, { proposalId: proposalHash }), status: STATUS_FAILED }
```
taoeffect: Please rewrite these. Should be:

```js
proposalToSend = extractProposalData(proposal, { proposalId: proposalHash, status: STATUS_PASSED })
```

Same for `STATUS_FAILED`

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
 
     return sendMessage({
       ...params,
-      data: {
-        ...data,
-        passPayload
+      data: { ...data, passPayload },
+      hooks: {
+        ...params?.hooks,
+        postpublish: (...args) => {
```
taoeffect: Please be consistent. Either `'gi.actions/group/proposal'` should use `postpublish`, or this should use `onprocessed`.

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+        }
       }
     })
   }),
-  ...encryptedAction('gi.actions/group/proposalCancel', L('Failed to cancel proposal.')),
+  ...encryptedAction('gi.actions/group/proposalCancel', L('Failed to cancel proposal.'), async function (sendMessage, params) {
+    const { contractID, data } = params
+    const state = await sbp('chelonia/contract/state', contractID)
+    const proposal = state.proposals[data.proposalHash]
+    const proposalToSend = { ...extractProposalData(proposal, { proposalId: data.proposalHash }), status: STATUS_CANCELLED }
```
taoeffect: Please simplify

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+
+    await sendMessage(params)
+
+    await sbp('gi.actions/group/notifyProposalStateInGeneralChatRoom', { groupID: contractID, proposal: proposalToSend })
+  }),
```
taoeffect: Please be consistent. Either this should be rewritten to use `onprocessed` (or `postpublish`), or the other code should be rewritten to avoid their use.

This does seem to be the simpler approach, should I vote for this method.

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+  ...encryptedAction('gi.actions/group/markProposalsExpired', L('Failed to mark proposals expired.'), async function (sendMessage, params) {
+    const { contractID, data } = params
+    const state = await sbp('chelonia/contract/state', contractID)
+    const proposal = state.proposals[data.proposalHash]
+    const proposalToSend = { ...extractProposalData(proposal, { proposalId: data.proposalHash }), status: STATUS_EXPIRED }
```
taoeffect: Please simplify

frontend/controller/actions/group.js
```diff
@@ -914,41 +912,92 @@ export default (sbp('sbp/selectors/register', {
...
+
+    await sendMessage(params)
+
+    await sbp('gi.actions/group/notifyProposalStateInGeneralChatRoom', { groupID: contractID, proposal: proposalToSend })
```
taoeffect: Needs to be consistent.

frontend/model/contracts/group.js
```diff
@@ -1234,10 +1233,10 @@ sbp('chelonia/defineContract', {
...
       }),
       process ({ contractID, data }, { state }) {
-        sbp('gi.contracts/group/pushSideEffect', contractID,
-          ['gi.contracts/group/releaseDeletedChatRoom', state.chatRooms[data.chatRoomID].members, data.chatRoomID]
-        )
-        delete state.chatRooms[data.chatRoomID]
+        const { chatRoomID } = data
+        const members = { ...state.chatRooms[chatRoomID].members }
+        delete state.chatRooms[chatRoomID]
+        sbp('gi.contracts/group/pushSideEffect', contractID, ['gi.contracts/group/releaseDeletedChatRoom', chatRoomID, members])
```
taoeffect: There is no need to make these changes, please revert.

taoeffect (changes requested):
Nice work @Silver-IT, some changes requested.

frontend/model/contracts/chatroom.js
Silver-IT: I don't think `state.members` would be `undefined` here, because `state.members` is already initiated inside `/join/process` when the contract is synced by someone who is not part of. But since the `attributes` is not initiated in any places, actually it was making errors and also notifications of type `CHELONIA_ERROR`.

In fact, this code and workflow are what @corrideat made and I think it's to handle the following situation.
- [Tab3] user3 joins the group
- [Tab2] user2 creates a private chatroom and invites user3
- [Tab2] user2 creates a proposal to remove user3
- [Tab1] user1 approves the proposal and kicks user3 from the group (and also kicks user3 from all the group chatrooms including the private ones)
In this case, the actorID is the user1 and he doesn't have the signing key of private chatroom. In this case, I can see that the user1 users the group contract key to kick user3 from the private chatroom.
In this case, the chatroom constructor can not be synced and user1 can only sync the actions from when user2 joins the group. So `state.members`, `state.attributes`, `state.messages`, ... are all `undefined` and we should make error handlers manually.

I am not sure I have 100% understood his code, but above is based on my understanding.

taoeffect: >  So state.members, state.attributes, state.messages, ... are all undefined and we should make error handlers manually.

You just said it yourself, if `state.members` is undefined then the code above online 287 would throw as well.

frontend/controller/actions/identity-kv.js
Silver-IT: This is something that I have already discussed in Slack. Discussion is in [here](https://okturtles.slack.com/archives/C0EH7P20Y/p1720133431952109?thread_ts=1720025185.746849&cid=C0EH7P20Y)

taoeffect: OK, then can you please a comment before line 32 explaining that this is to fix a Cypress heisenbug and include that link?

frontend/controller/actions/group.js
Silver-IT:
```
export type GIActionParams = {
  action: string;
  contractID: string;
  data: Object;
  options?: Object; // these are options for the action wrapper
  signingKeyId: ?string;
  encryptionKeyId: ?string;
  originatingContractID: ?string;
  signingContractID: ?string;
  innerSigningContractID: ?string;
  innerSigningKeyId: ?string;
  hooks?: {
    preSendCheck?: null | (Object, Object) => boolean;
    prepublishContract?: null | (Object) => void;
    prepublish?: null | (Object) => void;
    postpublish?: null | (Object) => void;
    onprocessed?: null | (Object) => void;
  };
  publishOptions?: { maxAttempts: number };
  returnInvocation?: boolean;
}
```
Since the `onprocessed` hook function is not defined as to return `Promise`, I couldn't do that way.
Also, since `onprocessed`, `prepublish`, `postpublish` are all defined to return `void`, not `Promise` in chelonia, if I change this `GIActionParams`, I also need to change many types too; `ChelRegParams`, `ChelActionParams`, `ChelKeyAddParams`, ...
However, I can see that the in the usage of those hooks, we assumed they would be async functions. Yeah, it's confusion. See the code below. 
```
await hooks?.postpublish?.(entry)
```
That's why I made that way. However, I will add `status` value inside.

taoeffect: If we do `await hooks?.postpublish?.(entry)` then it is correct to define them as returning a promise.

So feel free to update `(Object) => void` to `(Object) => Promise<void>;`

frontend/controller/actions/group.js
Silver-IT: The reason why I used `onprocessed` hook for `gi.actions/group/proposal` is that at the time of calling this function, the proposal doesn't exist in state. Because we are creating proposal in `gi.actions/group/proposal/process`.

And the reason why I used `postpublish` here, is that I wanted to call `gi.actions/group/notifyProposalStateInGeneralChatRoom` function just after the event is published, and before processed. The reason is that let's imagine the proposal is to remove a user from the group.
We should create two messages in the general chatroom.
- One is to notify that the proposal is closed by approval
- The other is to say the user has left the group

Which one do you think should be created before? Yes, the first message (which tells that the proposal is closed by approval). That's why I used `postpublish` here instead of `onprocessed`.

And I already made a question [here](https://okturtles.slack.com/archives/C0EH7P20Y/p1721189410310829), because sometimes (especially only in Cypress) the second one is created before the first one.

frontend/controller/actions/group.js
Silver-IT: Okay, I will remove `postpublish` in the other function.

frontend/controller/actions/group.js
taoeffect: What you are saying doesn't make any sense to me. In both cases, `notifyProposalStateInGeneralChatRoom` will get sent and processed by clients after `proposalVote`.

frontend/model/contracts/group.js
Silver-IT: Please see the `'gi.contracts/group/releaseDeletedChatRoom'` function definition below.
```
'gi.contracts/group/releaseDeletedChatRoom': (contractID, members) => {
  const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
  if (identityContractID && members[identityContractID]?.status === PROFILE_STATUS.ACTIVE) {
    sbp('chelonia/contract/release', contractID).catch(e => {
      console.error('[gi.contracts/group/releaseDeletedChatRoom] Error', e)
    })
  }
}
```

The first parameter is contractID, and the second parameter is members. But the previous code was wrong in the order of passing params. First was `members`, and the second was `chatRoomID`. And since the `members` is reference to the `state.chatRooms[data.chatRoomID].members` and we delete the `state.chatRooms[data.chatRoomID]` just after calling `pushSideEffect`, it could be `undefined`. That's why I created another variable `members`.

taoeffect: You are right that the order of the parameters needs to be fixed, but the rest of it can remain the same.

> And since the members is reference to the state.chatRooms[data.chatRoomID].members and we delete the state.chatRooms[data.chatRoomID] just after calling pushSideEffect, it could be undefined. That's why I created another variable members.

I'm pretty sure that isn't true and the only bug here was the order of the parameters.

Even though we do call `delete state.chatRooms[data.chatRoomID]`, that doesn't matter. the `members` will be saved as it is referenced and stored by the array that's passed to `pushSideEffect`. Try it and see.

frontend/model/contracts/chatroom.js
Silver-IT: `state.members` can only be `undefined` in `/join/process`. And we manually initiate it in the line 178.
```
'gi.contracts/chatroom/join': {
  validate: actionRequireInnerSignature(objectOf({
    memberID: optional(string) // user id of joining memberID
  })),
  process ({ data, meta, hash, height, contractID, innerSigningContractID }, { state }) {
    const memberID = data.memberID || innerSigningContractID

    if (!memberID) {
      throw new Error('The new member must be given either explicitly or implcitly with an inner signature')
    }

    if (!state.renderingContext) {
      if (!state.members) {
        state['members'] = {}
      }
        throw new GIChatroomAlreadyMemberError(`Can not join the chatroom which ${memberID} is already part of`)
      }
    }

...
```

frontend/controller/actions/group.js
Silver-IT: Okay. I will update all the types of GIActionParams and types of Chelonia too.

frontend/controller/actions/group.js
Silver-IT: `notifyProposalStateInGeneralChatRoom` is sent after `proposalVote` is published. But not sure before or after the `proposalVote` is processed.

frontend/model/contracts/group.js
Silver-IT: Okay.

frontend/controller/actions/group.js
taoeffect: So what?  I agree but I don't see how it matters.

Silver-IT: Please see this [Cypress failure](https://cloud.cypress.io/projects/q6whky/runs/2689/test-results/c9fc0180-0ff5-4107-8ec6-2f8879a4cf65?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github).
I thought I fixed this error in Cypress test case, but it still happens. But this shows what I mean. The order of two messages is random so the Cypress fails.

I want to make sure that the first message would always be created before the second one.

test/cypress/integration/group-chat.spec.js
```diff
@@ -498,7 +498,7 @@ describe('Group Chat Basic Features (Create & Join & Leave & Close)', () => {
     cy.giLogout()
   })
 
-  it(`user2 joins the ${groupName1} group and ${CHATROOM_GENERAL_NAME} again and logout`, () => {
+  it(`user2 joins the ${groupName1} group and ${CHATROOM_GENERAL_NAME} channel again and logout`, () => {
```
taoeffect, line 504: On my local test run the last test in `group-chat.spec.js` (`user2 joins the ${groupName1} group and ${CHATROOM_GENERAL_NAME} channel again and logout`) failed:

https://github.com/user-attachments/assets/136109de-d25c-4cc1-8ae8-08cb0b74f2b5

This is failing here inside of `cy.giAcceptGroupInvite`:

```js
  if (existingMemberUsername) {
    // NOTE: checking 'data-groupId' is for waiting until joining process would be finished
    cy.getByDT('pendingApprovalTitle').invoke('attr', 'data-groupId').should('eq', groupId)
    // NOTE: should wait until KEY_REQUEST event is published
    cy.giKeyRequestedGroupIDs(groupId)
    cy.giEmptyInvocationQueue() // <---------remove

    cy.giLogout()
```

And then in:

```js
Cypress.Commands.add('giLogout', ({ hasNoGroup = false } = {}) => {
  cy.giEmptyInvocationQueue()

  if (hasNoGroup) {
    cy.window().its('sbp').then(async sbp => await sbp('gi.app/identity/logout'))
  } else {
    cy.getByDT('settingsBtn').click() // <------ failing here
```

There are multiple problems here:

1. `cy.giEmptyInvocationQueue()` is being called an extra time before `cy.giLogout()` (which also calls it) - so that unnecessary call in `giAcceptGroupInvite` on line `commands.js:498` should be removed.
2. And the main problem: for some reason we were given the key to join the group and so the approval screen is visible here and the settings button to log out cannot be clicked, and also there is no need for it be clicked because we're already in the group

It might be because `user2` was already a member of the group before, and so was somehow able join again because the keys were not properly rotated when he left in `'user2 leaves the group by himself'`.

However, running the tests a second time everything passed, so this is a challenging and important Heisenbug that needs to be fixed. I will open an issue for this.

cc @corrideat

taoeffect (commented):
Leaving this comment here so that it can be referenced from another issue

test/cypress/integration/group-chat.spec.js
Silver-IT: Yeah, that makes sense. I remember I have already encountered the second issue several times before. Also it doesn't only happen in `group-chat.spec.js`, and I remember it was happening in any other test cases too.

Since you created an issue for this heisenbug, we can check it out and work on it later.

frontend/model/contracts/chatroom.js
taoeffect: I tested your theory above while removing these checks and did get an error:

```
[chelonia] ERROR 'TypeError' in processMutation for <op_a|z9brRu3VK2pPouEujA3sxDkiRsb1Zz8MsGaxw222wgeTxTe8CoT5 of z9brRu3VViuiGEcbSsroBWZ3GiqNUjNxvaRmaCye5R62q1sx15CZ>: can't access property "type", state.attributes is undefined TypeError: can't access property "type", state.attributes is undefined
```

So it seems like you're right.

Could you please copy/paste the steps you mentioned up above and add a comment with them to the code so that we remember why this is being done?

Also, it seems one check is still missing. After following the steps above I got this error:

```
[gi.contracts/chatroom/leave/sideEffect] Error at sideEffect can't access property "privacyLevel", state.attributes is undefined
```

Screenshot 2024-07-17 at 7 38 21 PM

taoeffect (changes requested):
Request for comment plus one more bug fix request

frontend/controller/actions/group.js
```diff
@@ -888,17 +886,9 @@ export default (sbp('sbp/selectors/register', {
...
-        ...params?.hooks,
-        postpublish: (...args) => {
-          sbp('okTurtles.events/emit', ACCEPTED_GROUP, { contractID: params.contractID })
-          params.hooks?.postpublish?.(...args)
-        }
-      }
-    })
+  ...encryptedAction('gi.actions/group/inviteAccept', L('Failed to accept invite.'), async function (sendMessage, params) {
+    await sendMessage(params)
+    sbp('okTurtles.events/emit', ACCEPTED_GROUP, { contractID: params.contractID })
```
taoeffect: Please return the result of `await sendMessage(params)` before the end of this function

frontend/controller/actions/group.js
```diff
@@ -914,41 +904,87 @@ export default (sbp('sbp/selectors/register', {
   ...encryptedAction('gi.actions/group/paymentUpdate', L('Failed to update payment.')),
   ...encryptedAction('gi.actions/group/sendPaymentThankYou', L('Failed to send a payment thank you note.')),
   ...encryptedAction('gi.actions/group/groupProfileUpdate', L('Failed to update group profile.')),
-  ...encryptedAction('gi.actions/group/proposal', L('Failed to create proposal.')),
+  ...encryptedAction('gi.actions/group/proposal', L('Failed to create proposal.'), async (sendMessage, params) => {
+    const { contractID } = params
+    await sendMessage({
```
taoeffect: Please `return` the result of `sendMessage` here

frontend/controller/actions/group.js
```diff
@@ -914,41 +904,87 @@ export default (sbp('sbp/selectors/register', {
...
     }
 
-    return sendMessage({
-      ...params,
-      data: {
-        ...data,
-        passPayload
-      }
-    })
+    await sendMessage({ ...params, data: { ...data, passPayload } })
```
taoeffect: Please save and return the result of `sendMessage` at the end of this function

frontend/controller/actions/group.js
```diff
@@ -914,41 +904,87 @@ export default (sbp('sbp/selectors/register', {
...
+
+    await sbp('gi.actions/group/notifyProposalStateInGeneralChatRoom', { groupID: contractID, proposal: proposalToSend })
+  }),
+  ...encryptedAction('gi.actions/group/markProposalsExpired', L('Failed to mark proposals expired.'), async function (sendMessage, params) {
+    const { contractID, data } = params
+    const state = await sbp('chelonia/contract/state', contractID)
+    const proposal = state.proposals[data.proposalHash]
+    const proposalToSend = extractProposalData(proposal, { proposalId: data.proposalHash, status: STATUS_EXPIRED })
+
+    await sendMessage(params)
```
taoeffect: Please save and return the result of `sendMessage` at the end of the function

frontend/controller/actions/identity-kv.js
```diff
@@ -28,10 +28,17 @@ export default (sbp('sbp/selectors/register', {
...
     }
-    return sbp('chelonia/kv/set', identityContractID, KV_KEYS.UNREAD_MESSAGES, data, {
-      encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
-      signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
-      onconflict
+
+    // NOTE: add the function `chelonia/kv/set` in identityContractID invocation queue
+    //       in order to remove conflict error while calling it
+    //       which uses fields of the identity contract state including height, cek, csk
+    //       https://okturtles.slack.com/archives/C0EH7P20Y/p1720053305870019?thread_ts=1720025185.746849&cid=C0EH7P20Y
```
taoeffect: Please mention this comment that this is to address a Heisenbug that can occur

taoeffect (changes requested):
A few minor change requests, but also please see this new Cypress failure that this PR is creating: https://cloud.cypress.io/projects/q6whky/runs/2699/test-results/85bbdf50-dca9-4e35-ab5d-ede8d5f4ab3e/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github

It looks like a heisenbug that depends on the order of messages in the chat perhaps. Or on the scroll position of the chatroom maybe?

Either way, it needs to be fixed because it seems quite frequent.

Silver-IT:
I have already added a comment in `gi.contracts/chatroom/join/process`.
```
// NOTE: this patch solves the issue of the action failing to process.
//       when the contract was not fully synced because some encryption keys are missing.
//       this normally happens when the user (not a member of PRIVATE chatroom)
//       is trying to sync the contract.
//       this comment works same for another checks like `if (!state.members)` of above and below codes
if (!state.attributes) return
```

Silver-IT:
> It looks like a heisenbug that depends on the order of messages in the chat perhaps. Or on the scroll position of the chatroom maybe?

It depends on the order of messages, not the scroll position.

Silver-IT:
> Either way, it needs to be fixed because it seems quite frequent.

Can I create an issue for this and work on it later?
That's because
- it only happens in Cypress (which works very fast than usual)
- I don't have solution at the moment
- We need to merge this PR ASAP

taoeffect:
> Can I create an issue for this and work on it later?

I agree we need to merge this PR asap, but we also need the tests to be regularly passing and this PR introduces serious and regular Heisenbugs. We already have heisenbugs. Please give it at least one more day of trying. You can try to change the tests so that they are not written in such a way that they depend on the order of messages so much.

test/cypress/integration/group-chat.spec.js
taoeffect: I'd like to leave this unresolved just so that when the issue links to it this conversation can be seen.

taoeffect (approved):
Nice work @Silver-IT!

PR #2154, opened 2024-07-04
Fix issue with ephemeral reference count

corrideat:


frontend/model/contracts/shared/functions.js
```diff
@@ -168,7 +168,7 @@ export function leaveChatRoom (contractID: string) {
     sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupID: rootState.currentGroupId })
   }
 
-  sbp('gi.actions/identity/kv/deleteChatRoomUnreadMessages', { contractID })
+  await sbp('gi.actions/identity/kv/deleteChatRoomUnreadMessages', { contractID })
```
Silver-IT: Why do we need to wait this KV action? It's to save unreadMessages in identityContractID space, not chatRoomID space.

Silver-IT (changes requested):
Nice work, @corrideat!

I can see the changes of this PR is to wait `gi.actions/identity/kv/deleteChatRoomUnreadMessages`, and I made one question.

frontend/model/contracts/shared/functions.js
corrideat: In that case, you're right that waiting shouldn't be needed.

Silver-IT (dismissed):
Nice work, @corrideat!

taoeffect (approved):
Probably OK

PR #2155, opened 2024-07-04
#2143 remove channel deletion message / Fix the bug re mentioning a deleted channel

### LINKED ISSUE #2143: Privacy: do not log what channels are deleted
### Problem

Whenever someone deletes a chatroom, the fact that they deleted it is logged to `#general`:

Screenshot 2024-07-01 at 10 11 35 AM

This fact is visible even to everyone, even for private chatrooms.

### Solution

If a chatroom is private, this message should not be logged.

SebinSong:
closes #2143 

@taoeffect 
As I mentioned in the issue too, made a bug-fix for mentioning a deleted channel too, like below.

It will say, 'Unknown chatroom' instead as in the screenshot. But please let me know if you want to present it otherwise.

---
As part of the bug-fix I had to modify how the app convert/interpret the channel mention.
It currently `#chatroom-1` -> `#chatroomId` before storing the message text in the contract. 
But later if it wants to convert it back to `#chatroom-1` but then chatroom has been deleted, The app can't find the chatroom data with that ID and interprets this `#chatroomId` as **a plain text** instead as a result.

To fix this, I updated it to do `#chatroom-1` -> **`#:chatID:chatroomId`** instead.
This way, if the app can't find the data with the `chatroomId` it still knows this text was a channel mention and present it as an unknown chatroom.

Hope it sounds good.

### DIFF
--- a/frontend/assets/style/_icons.scss
+++ b/frontend/assets/style/_icons.scss
@@ -5,6 +5,7 @@ $icons: (
   angle-left: "\f104",
   angle-down: "\f107",
   arrow-right: "\f061",
+  ban: "\f05e",
   bars: "\f0c9",
   bell: "\f0f3",
   bold: "\f032",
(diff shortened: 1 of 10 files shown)


taoeffect (approved):
Fantastic work @SebinSong!!!

PR #2156, opened 2024-07-05
Set reasonable character limits for payment info

### LINKED ISSUE #2112: Set reasonable character limits for payment info
### Problem

The text boxes on the payment info screen seem to accept an unlimited number of characters. (For reference, I gave up after 524,288 chars in 97,967 separate words.) Among other issues, this pushes the top of the unscrollable user info card off the top of the screen, hiding the profile photo, username, and any payment info that's higher in the list. 

paymentInfoNoCharLimit

### Solution

Limit the input to 250 characters. 
Indicate the user has reached the char limit the same way we do with usernames and group descriptions: an error with the text "Payment info cannot exceed 250 characters."

Silver-IT:
### Summary of Changes

- Closes #2112

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -1330,10 +1330,12 @@ sbp('chelonia/defineContract', {
           )
         })(data)
 
-        for (const paymentMethod of data.paymentMethods) {
-          const { value } = paymentMethod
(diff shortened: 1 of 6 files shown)


taoeffect (dismissed):
Approved! Nicely done!

taoeffect (approved):

PR #2157, opened 2024-07-05
Set reasonable character limit for non-monetary contributions

### LINKED ISSUE #2113: Set reasonable character limits for contributions
### Problem

The text boxes on the contribution screen seem to accept an unlimited number of characters. (For reference, I gave up after 30k chars.) Among other issues, this can make the contributions screen unnecessarily long, and make it difficult for users to see any contributions that are listed between/under lengthy contribution descriptions. This will affect all members, as the lengthy descriptions will show up under either "Receiving" or "Giving".

contributions-noCharLimits

### Solution

- Indicate the user has reached the char limit the same way we do with usernames and group descriptions: an error with the text "Contribution descriptions cannot exceed N characters."
- *Enhancement:* Allow a portion of long descriptions to be hidden under a "more..." toggle.

Silver-IT:
### Summary of Changes

- Closes #2113


taoeffect (approved):
Great work @Silver-IT! Approved!

PR #2161, opened 2024-07-05
#2160 - hide private-chat name to someone who is not part of

### LINKED ISSUE #2160: Users who are not part of a private chatroom can see the name of a private chatroom when it's mentioned
### Problem

When a private chatroom called `#test-private` is mentioned in `#general`, a user who is part of that chatroom sees this it in blue like this:

Screenshot 2024-07-05 at 10 24 17 AM

So far so good. But when a user who is not a member of that private chatroom sees it, they see it like this:

Screenshot 2024-07-05 at 10 24 34 AM

They can still see the name `#test-private`.

### Solution

The name that should be shown to users who are not part of the room should be `private channel`.

SebinSong:
closes #2160


taoeffect (approved):
Fantastic!

PR #2175, opened 2024-07-09
#2140 - Set reasonable character limit for user bio

### LINKED ISSUE #2140: Set reasonable character limit for user bio
### Problem

The text box for the user bio seems to accept an unlimited number of characters. (For reference, I gave up after ~500k characters.) Among other issues, this pushes the top of the unscrollable user info card off the top of the screen, hiding the profile photo and username. (Fortunately, payment info is unaffected because that's currently always at the bottom.) 

bio-noCharLimit

### Solution

Determine a reasonable character limit. (500?)
Limit the input to _n_ characters. 
Indicate the user has reached the char limit the same way we do elsewhere: an error with the text "Bio cannot exceed _n_ characters."

SebinSong:
closes #2140 

While at it, I actually added & DRYed the same UI to some other places too. (e.g. below modal)


taoeffect (changes requested):
Great work @SebinSong!

In most projects these sorts of checks are done on both the frontend (which can be bypassed easily) and the backend (which is hard to bypass). Our equivalent of the "backend" are the contracts. Similar to how other PRs handled this, could you update the contract validation functions to check for these limits too?

SebinSong:
Updated

frontend/model/contracts/group.js
```diff
@@ -1361,10 +1369,16 @@ sbp('chelonia/defineContract', {
...
-        chatRoomID: string,
-        description: string
-      })),
+      validate: (data, props) => {
+        actionRequireActiveMember(objectOf({
+          chatRoomID: string,
+          description: string
+        }))(data, props)
+
+        if (data?.description.length > CHATROOM_DESCRIPTION_LIMITS_IN_CHARS) {
```
taoeffect, line 1378: This `data?` is unnecessary because `data` shouldn't be `null` (the previous check will throw if it is null).

taoeffect (approved):
Approved! Fantastic work @SebinSong!

PR #2176, opened 2024-07-09
#2152 - Remove vote notification messages

### LINKED ISSUE #2010: Poll Notifications could be created no matter what hideVoters is
### Problem

While I was investigating the chatroom contract function `gi.contracts/chatroom/voteOnPoll`, I found a wrong workflow to create poll notification.
```
const msgIndex = findMessageIdx(data.hash, state.messages)
if (msgIndex >= 0) {
  ...

  if (pollData.hideVoters) { return }
}

// create a notification for vote (or vote changes)
```
Since the `msgIndex` could be `-1`, `if (pollData.hideVoters) { return }` does not always run. So the notification could be created no matter what the `pollData.hideVoters` is.

### Solution

Fix it

### LINKED ISSUE #2152: Remove "Voted on ____ Jump to poll" messages
### Problem

I just posted a poll and voted in it, but then I changed my mind about having the poll show who voted, so I deleted it with the intention of re-posting a poll with votes hidden. However, my vote was still shown:

Screenshot 2024-07-04 at 9 05 02 AM

### Solution

Remove these messages about who voted on what, they aren't needed and mess with the user experience.

SebinSong:
closes #2152 
closes #2010

### DIFF
--- a/frontend/model/contracts/chatroom.js
+++ b/frontend/model/contracts/chatroom.js
@@ -606,8 +606,6 @@ sbp('chelonia/defineContract', {
         votesAsString: string
       })),
       process ({ data, meta, hash, height, innerSigningContractID }, { state }) {
-        let shouldHideVoters = false
-
         const fnVoteOnPoll = (message) => {
           const myVotes = data.votes
           const pollData = message.pollData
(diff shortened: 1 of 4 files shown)


taoeffect (approved):
Great work @SebinSong! Excellent simplifications!

PR #2178, opened 2024-07-10
Fix error while rejoining group

### LINKED ISSUE #2172: Fix unstable workflow while joining/leaving group
### Problem

[Steps to reproduce] ([ref](https://okturtles.slack.com/archives/C02QF71TK1Q/p1720494359539889))
- [Tab2] user2 leaves the group by himself (doesn't logout)
- [Tab2] user2 opens the invitation link to join the group again
- [Tab2] user2 gets the following errors
image
image
image

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #2172

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -1424,7 +1424,8 @@ sbp('chelonia/defineContract', {
     'gi.contracts/group/_cleanup': ({ contractID, state }) => {
       // NOTE: should remove archived data from IndexedStorage
       //       regarding the current group (proposals, payments)
-      const possiblyUselessContractIDs = Object.keys(state.profiles || {})
+      const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
+      const possiblyUselessContractIDs = Object.keys(state.profiles || {}).filter(cID => cID !== identityContractID)
       sbp('chelonia/contract/release', possiblyUselessContractIDs).catch(e =>
         console.error('[gi.contracts/group/leaveGroup] Error calling release on all members', e)
       )
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -967,7 +967,7 @@ sbp('chelonia/defineContract', {
             }
 
             // subscribe to founder's IdentityContract & everyone else's
-            const profileIds = Object.keys(profiles)
+            const profileIds = Object.keys(profiles).filter(cID => cID !== userID)
             if (profileIds.length !== 0) {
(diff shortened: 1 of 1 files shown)


corrideat (commented):
Maybe this fixes the issue, but the identity contract has a `/retain` on the identity contract when logging in[1], to prevent it from being removed.

In short, while this may fix the unsubscribing issue, it likely doesn't solve the reference count issue.

Also, if you make this change, this other line [2] needs to be adjusted

[1] https://github.com/okTurtles/group-income/blob/aba2c13d28ecc6d0578a132da2eaf834e3353595/frontend/controller/actions/identity.js#L242
[2] https://github.com/okTurtles/group-income/blob/aba2c13d28ecc6d0578a132da2eaf834e3353595/frontend/model/contracts/group.js#L972

Silver-IT:
> The identity contract has a /retain on the identity contract when logging in[1], to prevent it from being removed.

Actually, that couldn't prevent it from being removed. I attached screenshot that shows only `retain` and `release` for the identity contract ID. The number (or undefined) shows the reference and ephemeral count at the end of each function.
image
```
'chelonia/contract/retain': async function (contractIDs: string | string[], params?: { ephemeral?: boolean}): Promise<*> {
  if (listOfIds.length === 0) return Promise.resolve()
  if (!params?.ephemeral) {
    listOfIds.forEach((id) => {
      if (!has(rootState.contracts, id)) {
        this.config.reactiveSet(rootState.contracts, id, Object.create(null))
      }
      this.config.reactiveSet(rootState.contracts[id], 'references', (rootState.contracts[id].references ?? 0) + 1)
      console.log('----- RETAIN [Reference] -----', id, rootState.contracts[id].references)
    })
  } else {
    listOfIds.forEach((id) => {
      if (!has(this.ephemeralReferenceCount, id)) {
        this.ephemeralReferenceCount[id] = 1
      } else {
        this.ephemeralReferenceCount[id] = this.ephemeralReferenceCount[id] + 1
      }
      console.log('----- RETAIN [Ephemeral] -----', id, this.ephemeralReferenceCount[id])
    })
  }
  return await sbp('chelonia/contract/sync', listOfIds)
}
```
So whenever the RELEASE is called and the `reference` (or `ephemeral`) count is `undefined`, the identity contract is removed. There are 3 times in the above screenshot. This is what I captured while rejoining the group of course with the previous code.

> If you make this change, this other line [2] needs to be adjusted

I will check it more.

taoeffect (approved):
Nice job @Silver-IT !

PR #2181, opened 2024-07-10
#2179 - bugfix for identityContract syncing failure not being notified

### LINKED ISSUE #2179: [Bug] 'Loading events from server..' banner doesn't disappear in the main page.
### Problem

when `sbp('chelonia/contract/sync', identityContractID ...)` in `main.js` fails, This banner doesn't go away.

image

### Solution
Make sure the `this.ephemeral.syncs` in `main.js` correctly filters out that failed `identityContractID`.

SebinSong:
closes #2179 

When this issue happens in `main.js` as a result of below async operation,
https://github.com/okTurtles/group-income/blob/1e959f059e07ca1d2278eb76fc176e6a71bebb1d/frontend/main.js#L532

It (the root component) needs to be notified of it so it can filters out that failed `identityContractID` from the `this.ephemeral.syncs` array.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -48,7 +48,7 @@ import './utils/touchInteractions.js'
 import { showNavMixin } from './views/utils/misc.js'
 import './views/utils/vStyle.js'
 
-const { Vue, L, LTags, LError } = Common
+const { Vue, L, LError } = Common
 
 console.info('GI_VERSION:', process.env.GI_VERSION)
 console.info('CONTRACTS_VERSION:', process.env.CONTRACTS_VERSION)
@@ -540,9 +540,7 @@ async function startApp () {
 
         sbp('gi.ui/prompt', {
           heading: L('Failed to login'),
-          question: e.message
-            ? L('Error details:{br_}{err}', { err: e.message, ...LTags() })
-            : L('Error details: {reportError}', ...LError(e)),
+          question: L('Error details: {reportError}', LError(e)),
           primaryButton: L('Close')
         })
       }).finally(() => {
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -48,7 +48,7 @@ import './utils/touchInteractions.js'
 import { showNavMixin } from './views/utils/misc.js'
 import './views/utils/vStyle.js'
 
-const { Vue, L } = Common
+const { Vue, L, LTags, LError } = Common
 
 console.info('GI_VERSION:', process.env.GI_VERSION)
 console.info('CONTRACTS_VERSION:', process.env.CONTRACTS_VERSION)
@@ -534,12 +534,17 @@ async function startApp () {
         await syncContractsInOrder(identityContractID, contractIDs)
 
         if (this.ephemeral.finishedLogin === 'yes') return
-        return sbp('gi.app/identity/login', { identityContractID }).catch((e) => {
-          console.error(`[main] caught ${e?.name} while logging in: ${e?.message || e}`, e)
-          console.warn(`It looks like the local user '${identityContractID}' does not exist anymore on the server 😱 If this is unexpected, contact us at https://gitter.im/okTurtles/group-income`)
-        })
+        return sbp('gi.app/identity/login', { identityContractID })
       }).catch(e => {
         console.error(`[main] caught ${e?.name} while fetching settings or handling a login error: ${e?.message || e}`, e)
+
+        sbp('gi.ui/prompt', {
+          heading: L('Failed to login'),
+          question: e.message
+            ? L('Error details:{br_}{err}', { err: e.message, ...LTags() })
+            : L('Error details: {reportError}', ...LError(e)),
+          primaryButton: L('Close')
+        })
       }).finally(() => {
         this.ephemeral.ready = true
         this.removeLoadingAnimation()
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -10,7 +10,7 @@ import type { GIKey, GIOpActionUnencrypted, GIOpContract, GIOpKeyAdd, GIOpKeyDel
 import type { Key } from './crypto.js'
 import { EDWARDS25519SHA512BATCH, deserializeKey, keyId, keygen, serializeKey } from './crypto.js'
 import { ChelErrorUnexpected, ChelErrorUnrecoverable } from './errors.js'
-import { CONTRACTS_MODIFIED, CONTRACT_REGISTERED, CONTRACT_IS_SYNCING } from './events.js'
+import { CONTRACTS_MODIFIED, CONTRACT_REGISTERED } from './events.js'
 // TODO: rename this to ChelMessage
 import { GIMessage } from './GIMessage.js'
 import type { Secret } from './Secret.js'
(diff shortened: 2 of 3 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -812,6 +812,7 @@ export default (sbp('sbp/selectors/register', {
         'chelonia/private/in/syncContract', contractID, params
       ]).catch((err) => {
         console.error(`[chelonia] failed to sync ${contractID}:`, err)
+        sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, false)
```
taoeffect: The proper way to fix this is by modifying `internals.js`, which is supposed to call `sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, false)`, and in fact already does in the `finally` block of that function. But the problem is that we don't get to the `finally` call, because the exception is thrown earlier at the call to

```js
const { HEAD: latestHEAD } = await sbp('chelonia/out/latestHEADInfo', contractID)
```

Which is before the `try` block even begins. So that is what needs to be fixed.

taoeffect (changes requested):
Nice! PR Review ready!

shared/domains/chelonia/chelonia.js
taoeffect: As I've been working on this as well, I was able to get it work by modifying `internals.js` like so to move the `try` up:

```js
  'chelonia/private/in/syncContract': async function (contractID: string, params?: { force?: boolean, resync?: boolean }) {
    try {
```

However, as you correctly pointed out on Slack, there is the additional problem that we should deal with, which is restoring the functionality of how it was working before when suddenly the identity contract has disappeared. We should forget about it, but right now we keep remembering:

```js
      sbp('gi.db/settings/load', SETTING_CURRENT_USER).then(async (identityContractID) => {
        // This loads CHELONIA_STATE when _not_ running as a service worker
        const cheloniaState = await sbp('gi.db/settings/load', 'CHELONIA_STATE')
        if (!cheloniaState || !identityContractID) return
        if (cheloniaState.loggedIn?.identityContractID !== identityContractID) return
        await sbp('chelonia/contract/sync', identityContractID, { force: true }) // this is throwing!
```

That `/sync` throws and whereas before we'd forget about the identity contract, now we are remembering it.

This logic is also questionable:

```js
        return sbp('gi.app/identity/login', { identityContractID }).catch((e) => {
          console.error(`[main] caught ${e?.name} while logging in: ${e?.message || e}`, e)
          console.warn(`It looks like the local user '${identityContractID}' does not exist anymore on the server 😱 If this is unexpected, contact us at https://gitter.im/okTurtles/group-income`)
        })
```

In that situation we would want to probably show a `gi.ui/prompt` to the user and to show the `e.message` that was caught.

So if you'd like to boyscout all of this in this PR as well, feel free!

taoeffect: > That `/sync` throws and whereas before we'd forget about the identity contract, now we are remembering it.

It is probably OK to remember it between sessions, as it will be overwritten when we sign up with a new account (or login to another account that does work). Refining this can be done for another PR.

But this is probably still a good idea:

> In that situation we would want to probably show a `gi.ui/prompt` to the user and to show the e.message that was caught.

shared/domains/chelonia/internals.js
```diff
@@ -1144,6 +1146,12 @@ export default (sbp('sbp/selectors/register', {
       this.config.reactiveSet(state[contractID], '_volatile', currentVolatileState)
     }
     const { HEAD: latestHEAD } = await sbp('chelonia/out/latestHEADInfo', contractID)
+      .catch(e => {
+        console.error(`[chelonia] chelonia/out/latestHEADInfo error: ${e.message || e}`, e)
+        emitContractSyncStop()
+        throw e
+      })
```
taoeffect: There's a simpler solution: if we move the `try` block near the top of this function (just after `const state = ...`, then that's all that's needed and the rest of the function can be unmodified.

taoeffect (changes requested):
Minor simplification request

shared/domains/chelonia/internals.js
SebinSong: right. done it.

shared/domains/chelonia/chelonia.js
SebinSong: added the error prompt for it.

image

The thing I noticed is that this catch block never catches any log-in related errors. (probably because `'okTurtles.eventQueue/queueEvent'` is not promise-based.)
```js
return sbp('gi.app/identity/login', { identityContractID }).catch((e) => {
          console.error(`[main] caught ${e?.name} while logging in: ${e?.message || e}`, e)
          console.warn(`It looks like the local user '${identityContractID}' does not exist anymore on the server 😱 If this is unexpected, contact us at https://gitter.im/okTurtles/group-income`)
```

So I placed the logic another catch block below it.

shared/domains/chelonia/internals.js
```diff
@@ -1143,21 +1143,23 @@ export default (sbp('sbp/selectors/register', {
...
-    if (!isSubcribed) {
-      const entry = this.pending.find((entry) => entry?.contractID === contractID)
-      // we're syncing a contract for the first time, make sure to add to pending
-      // so that handleEvents knows to expect events from this contract
-      if (!entry) {
-        this.pending.push({ contractID })
-      }
-    }
-    this.postSyncOperations[contractID] = this.postSyncOperations[contractID] ?? Object.create(null)
+
```
taoeffect, line 1132: To be safe, could you move the `try` just above line 1132?

The reason I ask is because while we're making this change, it's a good idea to move it up because some of the lines above might throw. Specifically, `this.config.reactiveSet` could theoretically throw, and so could `sbp('okTurtles.events/emit', CONTRACT_IS_SYNCING, contractID, true)`

frontend/main.js
```diff
@@ -534,12 +534,17 @@ async function startApp () {
...
-        })
+        return sbp('gi.app/identity/login', { identityContractID })
       }).catch(e => {
         console.error(`[main] caught ${e?.name} while fetching settings or handling a login error: ${e?.message || e}`, e)
+
+        sbp('gi.ui/prompt', {
+          heading: L('Failed to login'),
+          question: e.message
+            ? L('Error details:{br_}{err}', { err: e.message, ...LTags() })
+            : L('Error details: {reportError}', ...LError(e)),
```
taoeffect: Hmm, this is a bit confusing, because it calls `LError` if `e.message` is not defined, but `LError` is written in a way that expects `e.message` to be defined:

```js
  return {
      errorMsg: error.message,
      'a_': `<a class="link" target="_blank" href="${url}">`,
      '_a': '</a>'
    })
  }
```

frontend/main.js
```diff
@@ -534,12 +534,17 @@ async function startApp () {
         await syncContractsInOrder(identityContractID, contractIDs)
 
         if (this.ephemeral.finishedLogin === 'yes') return
-        return sbp('gi.app/identity/login', { identityContractID }).catch((e) => {
-          console.error(`[main] caught ${e?.name} while logging in: ${e?.message || e}`, e)
-          console.warn(`It looks like the local user '${identityContractID}' does not exist anymore on the server 😱 If this is unexpected, contact us at https://gitter.im/okTurtles/group-income`)
```
taoeffect, line 539: Nice job removing this outdated chatroom that we no longer use :)

taoeffect (changes requested):
Review ready!

taoeffect (approved):
Great work @SebinSong! 👏

PR #2184, opened 2024-07-11
Fix error in creating duplicated notifications while rejoining

### LINKED ISSUE #1838: Duplicated notifications after rejoining
### Problem

**Steps to reproduce**
- [Tab1] `u1` signs up, and creates a group
- [Tab2] `u2` joins the group
- [Tab3] `u3` joins the group
- [Tab1] `u1` creates two proposals; one to change mincome, and the other is generic
- [Tab3] `u3` leaves the group, and rejoins the group
- [Tab3] `u3` checks the duplicated notifications
image

**What should be happening**
- [Tab3] There should be no duplicated notifications

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #1838

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -553,8 +553,7 @@ async function startApp () {
       ...mapState(['contracts']),
       ourUnreadMessagesCount () {
         return Object.keys(this.ourUnreadMessages)
-          // TODO: need to remove the '|| []' after we release 0.2.*
-          .map(cId => (this.ourUnreadMessages[cId].unreadMessages || []).length)
+          .map(cId => (this.ourUnreadMessages[cId].unreadMessages).length)
           .reduce((a, b) => a + b, 0)
       },
       shouldSetBadge () {
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -199,7 +199,7 @@ function memberLeaves ({ memberID, dateLeft, heightLeft }, { contractID, meta, s
   state._volatile.pendingKeyRevocations[CEKid] = true
 }
 
-function isActionOlderThanUser (contractID: string, height: number, userProfile: ?Object): boolean {
+function isActionOlderThanUser (height: number, userProfile: ?Object): boolean {
   // A util function that checks if an action (or event) in a group occurred after a particular user joined a group.
   // This is used mostly for checking if a notification should be sent for that user or not.
   // e.g.) user-2 who joined a group later than user-1 (who is the creator of the group) doesn't need to receive
@@ -721,7 +721,7 @@ sbp('chelonia/defineContract', {
 
         const myProfile = getters.groupProfile(loggedIn.identityContractID)
 
-        if (isActionOlderThanUser(contractID, height, myProfile)) {
+        if (isActionOlderThanUser(height, myProfile)) {
           sbp('gi.contracts/group/emitNotificationsAfterSyncing', [contractID, innerSigningContractID], [{
             notificationName: 'NEW_PROPOSAL',
             notificationData: {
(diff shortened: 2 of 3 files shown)


frontend/main.js
```diff
@@ -553,8 +553,7 @@ async function startApp () {
       ...mapState(['contracts']),
       ourUnreadMessagesCount () {
         return Object.keys(this.ourUnreadMessages)
-          // TODO: need to remove the '|| []' after we release 0.2.*
-          .map(cId => (this.ourUnreadMessages[cId].unreadMessages || []).length)
```
taoeffect, line 557: Great catch!

frontend/model/contracts/group.js
```diff
@@ -199,7 +199,7 @@ function memberLeaves ({ memberID, dateLeft, heightLeft }, { contractID, meta, s
   state._volatile.pendingKeyRevocations[CEKid] = true
 }
 
-function isActionOlderThanUser (contractID: string, height: number, userProfile: ?Object): boolean {
```
taoeffect, line 202: Also a nice catch 👍 

One problem though: this function is named incorrectly. It asks if the action is older than the user.

So if the action happened at block 15, and the user joined at block 20, then it should return true (because the action is "older than" when the user joined).

However, the condition below will do:

```
20 < 15 ===> false
```

So this function is named in the opposite way of how it should be.

Either the function needs to be renamed to `isActionNewerThanUserJoinedDate`, or the condition below needs to be reversed.

I recommend renaming the function, because that won't require changing the code that uses the function.

frontend/model/contracts/group.js
```diff
@@ -830,7 +836,11 @@ sbp('chelonia/defineContract', {
           })
         }
 
-        sbp('gi.contracts/group/emitNotificationsAfterSyncing', contractID, notifications)
+        const { loggedIn } = sbp('state/vuex/state')
```
taoeffect, line 821: Since `'gi.contracts/group/emitNotifications'` does nothing significant anymore, it can be removed, and instead of calling it like this, `notifications` array can be deleted.

Then the loop above can be rewritten, something like this:

```js
if (isActionNewerThanUserJoinedDate(height, myProfile)) {
  for (const proposalId of data.proposalIds) {
    sbp('gi.notifications/emit', 'PROPOSAL_EXPIRING', {
      // ...
    })
  }
}
```

frontend/model/contracts/group.js
```diff
@@ -1823,12 +1831,7 @@ sbp('chelonia/defineContract', {
         console.warn(`removeForeignKeys: ${e.name} error thrown:`, e)
       })
     },
-    'gi.contracts/group/emitNotificationsAfterSyncing': async (contractIDs, notifications) => {
-      const listOfIds = typeof contractIDs === 'string' ? [contractIDs] : contractIDs
-      for (const id of listOfIds) {
-        await sbp('chelonia/contract/wait', id)
-      }
-
+    'gi.contracts/group/emitNotifications': (notifications) => {
```
taoeffect: This method no longer does anything useful and therefore can be deleted

frontend/model/contracts/group.js
```diff
@@ -982,21 +1004,7 @@ sbp('chelonia/defineContract', {
...
-                sbp('gi.contracts/group/emitNotificationsAfterSyncing', [], [{
-                  notificationName: 'MEMBER_ADDED',
-                  notificationData: {
-                    createdDate: meta.createdDate,
-                    groupID: contractID,
-                    memberID: innerSigningContractID
-                  }
-                }])
-              }
-            }).catch((e) => {
```
taoeffect: I do not see a good reason to move this code above. Can it be reverted?

taoeffect (changes requested):
Great work @Silver-IT! Review ready!

frontend/model/contracts/group.js
Silver-IT: I also think renaming is better. Nice catch!

frontend/model/contracts/group.js
Silver-IT: Actually, the reason was to emit notification inside the invocation queue of syncing contracts. But since I blocked the duplicated notifications in notifications Vuex mutation, this could be reverted.

taoeffect (approved):
Great work @Silver-IT 👏

PR #2186, opened 2024-07-11
Fix errors in switching chatroom when user leaves group or chatroom

### LINKED ISSUE #2185: Switching chatroom doesn't work when the user leaves it
### Problem

**Steps to reproduce**
- [Tab1] user1 creates a new channel and invites user2
- [Tab2] switches to the new channel
- [Tab1] user2 kicks user2 from the new channel
- [Tab2] user2 should be switched to the general chatroom, but nothing happens

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #2185


frontend/model/contracts/group.js
```diff
@@ -362,9 +362,13 @@ const leaveChatRoomAction = async (groupID, state, chatRoomID, memberID, actorID
     // Instead, in the future contracts will have an 'environment', provided
     // by Chelonia, which will include global / environment / ambient
     // information they need.
-    if (memberID === sbp('state/vuex/state').loggedIn.identityContractID) {
-      sbp('okTurtles.events/emit', LEFT_CHATROOM, { identityContractID: memberID, groupContractID: groupID, chatRoomID })
-    }
+
+    // NOTE: comment out the code below and run it just inside the /leaveChatRoom/sideEffect
+    //       this code shouldn't be here because group members who are leaving can not reach to this code
+    //       when they are kicked from someone else
```
taoeffect: This should say *kicked **by** someone else*

frontend/views/containers/chatroom/ChatMixin.js
```diff
@@ -141,6 +142,22 @@ const ChatMixin: Object = {
...
         this.redirectChat(to)
       }
+    },
+    'ourGroupDirectMessages': {
+      immediate: true,
+      handler (to, from) {
+        // NOTE: whenever any group members leave the group, and the ourGroupDirectMessage is changed
+        //       currentChatRoomId needs to be considered to needs to be changed
+        //       if the currentChatRoomId (if it's DM) is no longer group direct message
+        if (Object.keys(this.ourDirectMessages).includes(this.currentChatRoomId)) {
```
taoeffect: Should this be `Object.keys(from)`?

taoeffect (changes requested):
A minor comment improvement and a question

frontend/views/containers/chatroom/ChatMixin.js
Silver-IT: No. It is `this.ourDirectMessages`, not `this.ourGroupDirectMessages`. Since there could be confusion, I will update some related codes.

frontend/model/chatroom/vuexModule.js
```diff
@@ -107,10 +107,14 @@ const getters = {
...
   },
-  isDirectMessage (state, getters) {
+  isGroupDirectMessage (state, getters) {
     // NOTE: identity contract could not be synced at the time of calling this getter
     return chatRoomID => !!getters.ourGroupDirectMessages[chatRoomID || getters.currentChatRoomId]
   },
+  isDirectMessage (state, getters) {
+    // NOTE: identity contract could not be synced at the time of calling this getter
+    return chatRoomID => !!getters.ourDirectMessages[chatRoomID || getters.currentChatRoomId]
+  },
```
taoeffect, line 117: This is based on getter `ourDirectMessages` which I'm noticing is defined twice (DRY violation) in both `state.js` and `shared/getters/identity.js`.

Could you please delete the copy of it in `state.js`?

taoeffect (changes requested):
Great work @Silver-IT!

I left a comment [here](https://github.com/okTurtles/group-income/pull/2186#discussion_r1676476927) for a minor change request.

I also found a serious bug:

I decided to test leaving a chatroom voluntarily and being kicked out of the chatroom to see if both worked.

- To do this, I created a group with two users, u1 and u2.
- I had each user create two chatrooms, so 4 chatrooms in total.
- Then, each user added the other user to the chatrooms they created.
- Next, for each of the chatrooms each user created, I kicked the other user out of one of the chatrooms, and had the other user leave voluntarily.
- Finally, in separate session tabs, I had both users log in (so `u1` is logged in to `[Tab1]` and `[Tab3]` and `u2` is logged in to `[Tab2]` and `[Tab4]`

I started to record toward the end of this, and you can see the result:

Screenshot 2024-07-12 at 3 06 50 PM

When the users left the chartroom voluntarily, this resulted in an error when they logged back in on a new fresh session tab.

```
[chelonia/contract/release] Invalid negative reference count for z9brRu3VKg2ifBpQHXqLZyCa3aqw5Kv397h2XECS2com8ug8dQM4 [captureLogs.js:36:30](http://localhost:3000/frontend/model/captureLogs.js)
[leaveChatRoomAction] Error releasing chatroom z9brRu3VKg2ifBpQHXqLZyCa3aqw5Kv397h2XECS2com8ug8dQM4 Error: Invalid negative reference count
```

https://github.com/user-attachments/assets/48b8a3b0-e85f-4f3e-90b0-9444ad2c42f7

frontend/model/chatroom/vuexModule.js
Silver-IT: Nice catch! I will remove it

frontend/model/contracts/group.js
```diff
@@ -348,24 +348,21 @@ const leaveChatRoomAction = async (groupID, state, chatRoomID, memberID, actorID
...
+    //       to find these calls in the future.
+    //       `chelonia/rootState` contains that same information but it's a bad
+    //       choice because:
+    //          1. Once sandboxing is implemented, it may need to be an async call
+    //          2. Generally, contracts should _not_ access the root state because
+    //             it makes it difficult or impossible to contain them (meaning there
+    //             would be no point in sandboxing)
+    //       Instead, in the future contracts will have an 'environment', provided
+    //       by Chelonia, which will include global / environment / ambient
+    //       information they need.
```
taoeffect: There are multiple problems here.

1. The code for which this was a comment was moved. Now this is a `.then` handler that does nothing and the comment here is not for anything.
2. The code below with `.catch` is also nonsensical.

So to fix this, please move this comment to where you moved `sbp('state/vuex/state').loggedIn.identityContractID`, because it must stay with it. In this case, above line 1265.

Then remove the `.then` that does nothing, and replace the two `.catch` with one:

```js
  }).catch((e) => {
    if (
      leavingGroup &&
      (e?.name === 'ChelErrorSignatureKeyNotFound' || (
        e?.name === 'GIErrorUIRuntimeError' &&
        (
          ['ChelErrorSignatureKeyNotFound', 'GIErrorMissingSigningKeyError'].includes(e?.cause?.name) ||
          e?.cause?.name === 'GIChatroomNotMemberError'
        )
      ))
    ) {
      // This is fine; it just means we were removed by someone else
      return
    }
    console.warn('[gi.contracts/group] Error sending chatroom leave action', e)
  })
```

taoeffect (changes requested):
Great work @Silver-IT! I tested it and the previous error is gone now!

One comment left.

frontend/model/contracts/chatroom.js
```diff
@@ -163,9 +163,7 @@ sbp('chelonia/defineContract', {
       }
     },
     'gi.contracts/chatroom/join': {
-      validate: actionRequireInnerSignature(objectOf({
-        memberID: optional(string) // user id of joining memberID
-      })),
```
taoeffect: Why was `actionRequireInnerSignature` removed from here?

frontend/model/contracts/chatroom.js
```diff
@@ -265,9 +263,7 @@ sbp('chelonia/defineContract', {
       }
     },
     'gi.contracts/chatroom/leave': {
-      validate: objectOf({
-        memberID: optional(string) // member to be removed
-      }),
```
taoeffect: In the future, I ask that you please avoid making these sorts of modifications to the code, as it adds more work for me for no reason and also increases the likelihood of merge conflicts

taoeffect (commented):
Some questions...

frontend/model/contracts/chatroom.js
Silver-IT: Oh, nice catch. I should revert it.

frontend/model/contracts/chatroom.js
Silver-IT: I will revert this change too since I am going to make a commit to resolve your previous feedback.

taoeffect (approved):
Great work @Silver-IT ! Approved!

PR #2187, opened 2024-07-11
#2173 - Add prompt for updating password

### LINKED ISSUE #2173: Disable updating password.
### Problem

Updating the user's password isn't implemented right now (as it's somewhat complicated and involves rotating keys).

Screenshot 2024-07-09 at 1 49 41 PM

### Solution

For now please make it so that when "Update Password" is clicked, a `gi.ui/prompt` is shown that says:

> Sorry, this feature has not been implemented yet. Please check back later.

With a single button: "OK"

SebinSong:
closes #2173

### DIFF
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -1,7 +1,7 @@
 <template lang='pug'>
   div
-    component(:is='content' ref='content' v-bind='childData')
-    component(:is='subcontent[subcontent.length-1]' ref='subcontent' v-bind='childData')
+    component(:is='content' ref='content' v-bind='childData[content]')
+    component(v-if='activeSubContent' :is='activeSubContent' ref='subcontent' v-bind='childData[activeSubContent]')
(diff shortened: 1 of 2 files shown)


frontend/views/components/modal/Modal.vue
```diff
@@ -1,7 +1,7 @@
 <template lang='pug'>
```
SebinSong, line 1: All changes in this file is irrelevant to the #2173 but a fix for below intermittent bug I encountered in `Home.vue`

image

This happens because when the prompt pops up for any login-error but then `LoginModal.vue` is immediately queried too in the `Home.vue`. `childData` for error pop-up is lost in this case, which leads to this blank modal.
Updated `Modal.vue` so that it stores the `childData` as a map of `{ 'modal-name': data }` instead.

taoeffect (approved):
Great work @SebinSong!

You also seem to have fixed another issue that I just opened! Awesome! 😄 🎉

PR #2197, opened 2024-07-15
#2193 modals dismissed in wrong orders

### LINKED ISSUE #2193: Problem dismissing modals in the wrong order
### Problem

https://github.com/user-attachments/assets/98a97269-1cd7-4368-abe0-d681d0fd7efc

A recent PR updated how the modal system works. Here's an example of a problem that still exists. You can reproduce this with the following steps:

1. Run `grunt dev` and create a group
2. Visit the dashboard
3. Stop `grunt dev`
4. Run `grunt dev` again and wait for the server to be fully running
5. Refresh the page you had open

### Solution

Find and fix the problem, making sure to test thoroughly the existing modals in the group, especially any modals-in-modals, e.g. when the group settings or user settings are up and then a modal is brought up inside of one of those.

In this situation the modal at the top should be dismissed first.

SebinSong:
closes #2193 

https://github.com/user-attachments/assets/39caa20a-1d41-4b22-b75d-d9141256137f

### DIFF
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -86,19 +86,20 @@ export default ({
       }
     },
     updateUrl () {
-      if (this.content) {
+      const subContentLen = this.subcontent.length
+      if (this.content || subContentLen > 0) {
         const contentQueries = this.queries[this.content] || {}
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Great work @SebinSong! It seems to work.

PR #2204, opened 2024-07-15
#2201 - Incorrect poll expiration

### LINKED ISSUE #2201: Polls automatically close far too soon
### Problem

On the `testing` server, 1-_day_ polls are closing within approximately 1 _minute_.

pollClosed

It does not seem to make any difference what kind of poll is created or whether anyone votes. 

### Solution

Ensure polls are open for the correct duration.

SebinSong:
closes #2201 

cc. @dotmacro

### DIFF
--- a/frontend/views/containers/chatroom/MessagePoll.vue
+++ b/frontend/views/containers/chatroom/MessagePoll.vue
@@ -128,7 +128,7 @@ export default ({
         // if the poll has been expired, mark it 'closed' immediately.
         markPollClosed()
       } else if (Math.abs(timeDiff) < DAYS_MILLIS) {
-        // if the poll is expiring soon(at least within 24 hrs), periodically check & mark it 'closed'.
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/MessagePoll.vue
```diff
@@ -116,7 +116,7 @@ export default ({
       }
 
       const checkAndSetTimer = () => {
-        if (Date.now() - this.pollData.expires_date_ms) {
+        if (Date.now() - this.pollData.expires_date_ms >= 0) {
```
SebinSong, line 119: When poll is not expired yet, this calculation evaluates to `-` value which is `truthy`. So adding `>= 0` for the bug-fix.

And this issue could only be observed for `1 Day` duration poll because, this periodic check only gets executed if the expiration happens in 24 hrs.

taoeffect (approved):

PR #2207, opened 2024-07-16
#2144 - Updating user avatar while group settings is open updates both avatars

### LINKED ISSUE #2144: Updating user avatar while group settings is open updates both avatars
### Problem

Updating user avatar while group settings is open updates both avatars.

Steps to reproduce:

1. Open group settings. 
2. Open user settings.
3. Change user avatar.
4. Close user settings.
5. Tada! User and group now have same avatar, from step 3.

With screenshots:

Step 1. Open group settings
avatar—gsettingsOpenDiffAvatars

Step 2. Open user settings
avatar—usettingsOpenDiffAvatars

Step 3. User has changed their avatar
avatar—usettingsOpenChangedAvatar

Step 5. User and group now have same avatar
avatar—gsettingsOpenSameAvatars

### Solution

Ensure that the correct avatar, and only the correct avatar, is updated when the user updates either the group avatar or their user avatar. 
NOTE: While it may be technically possible to do in the other direction (i.e. updating both avatars via group settings) that's not currently feasible in the GUI because the user settings is displayed as full-screen modal (i.e. there's no way to access group settings while user settings is open). The solution should make sure it's not technically possible to do in either case, regardless of how a modal is displayed.

SebinSong:
closes #2144 (great catch @dotmacro, btw. This is something I implemented a while ago.)

`AvatarUpload.vue` subscribes to `AVATAR_EDITED` for when the user is done with editing the image. But then, when two components that have `AvatarUpload.vue` are both in the screen, they all end up listening to this event and updating the avatar at the same time.
As a fix, I've placed `avatarType` prop in the event payload as an identifier, so that the component knows whether the event is meant for it or not.

### DIFF
--- a/frontend/views/components/AvatarUpload.vue
+++ b/frontend/views/components/AvatarUpload.vue
@@ -34,6 +34,10 @@ export default ({
     sbpParams: {
       type: Object,
       required: true
+    },
+    avatarType: {
+      type: String,
+      validator: v => ['user', 'group'].includes(v) // feel free to extend this list.
     }
   },
   components: {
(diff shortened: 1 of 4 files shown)


taoeffect (approved):
Great work @SebinSong !

PR #2209, opened 2024-07-16
improvements to grunt pin

taoeffect:
`grunt pin` now doesn't need to run `genManifestsAndDeploy` a second time and it has better error handling in the event that `contracts/` directory doesn't exist.


SebinSong (approved):
Looks good.

PR #2211, opened 2024-07-17
#2196 - Set reasonable character limit for group name

### LINKED ISSUE #2196: Set reasonable character limit for group name
### Problem

The text box for group name seem to accept an unlimited number of characters. (For reference, I gave up after 5500 chars.) 
Note: This did not seem to cause any issues with the UI; the group name was centered or truncated as appropriate everywhere I saw. 

(apologies for missing this issue in earlier testing!)

### Solution

- Indicate the user has reached the char limit the same way we do elsewhere

SebinSong:
closes #2196

### DIFF
--- a/frontend/model/contracts/chatroom.js
+++ b/frontend/model/contracts/chatroom.js
@@ -13,6 +13,7 @@ import {
   CHATROOM_DESCRIPTION_LIMITS_IN_CHARS,
   CHATROOM_MAX_MESSAGES,
   CHATROOM_NAME_LIMITS_IN_CHARS,
+  CHATROOM_MAX_MESSAGE_LEN,
   CHATROOM_PRIVACY_LEVEL,
   CHATROOM_TYPES,
   MESSAGE_NOTIFICATIONS,
@@ -399,7 +400,7 @@ sbp('chelonia/defineContract', {
       validate: actionRequireInnerSignature(objectOf({
         hash: stringMax(MAX_HASH_LEN, 'hash'),
         createdHeight: number,
-        text: string
+        text: stringMax(CHATROOM_MAX_MESSAGE_LEN, 'text')
       })),
       process ({ data, meta }, { state }) {
         const { hash, text } = data
@@ -461,7 +462,7 @@ sbp('chelonia/defineContract', {
           hash: stringMax(MAX_HASH_LEN, 'hash'),
           // NOTE: manifestCids of the attachments which belong to the message
           //       if the message is deleted, those attachments should be deleted too
-          manifestCids: arrayOf(string),
+          manifestCids: arrayOf(stringMax(MAX_HASH_LEN, 'manifestCids')),
           messageSender: stringMax(MAX_HASH_LEN, 'messageSender')
         })(data)
 
--- a/frontend/model/contracts/chatroom.js
+++ b/frontend/model/contracts/chatroom.js
@@ -4,10 +4,11 @@
 
 import { L } from '@common/common.js'
 import sbp from '@sbp/sbp'
-import { actionRequireInnerSignature, arrayOf, number, object, objectOf, optional, string } from '~/frontend/model/contracts/misc/flowTyper.js'
+import { actionRequireInnerSignature, arrayOf, number, object, objectOf, optional, string, stringMax } from '~/frontend/model/contracts/misc/flowTyper.js'
 import { ChelErrorGenerator } from '~/shared/domains/chelonia/errors.js'
 import { findForeignKeysByContractID, findKeyIdByName } from '~/shared/domains/chelonia/utils.js'
 import {
+  MAX_HASH_LEN,
   CHATROOM_ACTIONS_PER_PAGE,
   CHATROOM_DESCRIPTION_LIMITS_IN_CHARS,
   CHATROOM_MAX_MESSAGES,
@@ -236,7 +237,7 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/chatroom/rename': {
       validate: actionRequireInnerSignature((data, { state, message: { innerSigningContractID } }) => {
-        objectOf({ name: string })(data)
+        objectOf({ name: stringMax(CHATROOM_NAME_LIMITS_IN_CHARS, 'name') })(data)
 
         if (state.attributes.creatorID !== innerSigningContractID) {
           throw new TypeError(L('Only the channel creator can rename.'))
@@ -252,7 +253,7 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/chatroom/changeDescription': {
       validate: actionRequireInnerSignature((data, { state, message: { innerSigningContractID } }) => {
-        objectOf({ description: string })(data)
+        objectOf({ description: stringMax(CHATROOM_DESCRIPTION_LIMITS_IN_CHARS, 'description') })(data)
 
         if (state.attributes.creatorID !== innerSigningContractID) {
           throw new TypeError(L('Only the channel creator can change description.'))
@@ -396,7 +397,7 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/chatroom/editMessage': {
       validate: actionRequireInnerSignature(objectOf({
-        hash: string,
+        hash: stringMax(MAX_HASH_LEN, 'hash'),
         createdHeight: number,
         text: string
       })),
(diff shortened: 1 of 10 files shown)


frontend/model/contracts/group.js
```diff
@@ -456,29 +458,36 @@ sbp('chelonia/defineContract', {
...
           })
-        })
-      }),
+        })(data)
+
+        const groupName = data.settings.groupName
+        if (!stringMax(GROUP_NAME_MAX_CHAR)(groupName)) {
+          throw new TypeError(`A group name cannot exceed ${GROUP_NAME_MAX_CHAR} characters.`)
+        }
+      },
```
taoeffect: This could be restored to how it was and made way more secure if you implement `stringMax()` validation function (#902).

Then there's no need to create a custom validation function, and every one of `string` parameters here could be validated.

frontend/model/contracts/shared/validators.js
```diff
@@ -14,3 +14,7 @@ export const decimals = (digits: number): ((value: number) => boolean) => (value
 export const noUppercase = (value: string): boolean => value.toLowerCase() === value
 
 export const noWhitespace = (value: string): boolean => /^\S+$/.test(value)
+
+export function stringMax (maxChar: number): ((value: string) => boolean) {
+  return (value: string): boolean => value.length <= maxChar
+}
```
taoeffect: Please move this to `flowTyper.js` and implement it in the same way as the other validation functions in there are implemented.

taoeffect (changes requested):
Nice work @SebinSong. Sorry I wasn't clear enough about how `stringMax` is supposed to work. I added more details in this review.

SebinSong:
@taoeffect Updated the PR again.

frontend/model/contracts/misc/flowTyper.js
```diff
@@ -341,6 +341,18 @@ export const string = (
...
+
+  return function stringMax (value, _scope = '') {
+    string(value, _scope)
+    if (value.length <= numChar) return value
+    throw new TypeError(
+      key ? `[stringMax] string type '${key}' cannot exceed ${numChar} characters` : `[stringMax] cannot exceed ${numChar} characters`
+    )
+  }
+}
+
```
SebinSong, line 355: Let me know if this can be simplified and improved.

frontend/model/contracts/group.js
```diff
@@ -459,7 +460,7 @@ sbp('chelonia/defineContract', {
       validate: objectMaybeOf({
         settings: objectMaybeOf({
           // TODO: add 'groupPubkey'
-          groupName: string,
+          groupName: stringMax(GROUP_NAME_MAX_CHAR, 'groupName'),
```
SebinSong, line 468: I thought passing `key` (`'groupName'` above)here and displaying it in the error message would help with faster debugging. (But let me know if it's not a good thing to do)

frontend/model/contracts/misc/flowTyper.js
SebinSong: Let me know if this can be simplified and improved.

frontend/model/contracts/group.js
taoeffect: Seems like a good idea 👍

frontend/model/contracts/misc/flowTyper.js
taoeffect: @SebinSong you did a really good job here. It's almost perfect. I worked with Claude to see if some improvements could be made, and here's what "we" came up with:

```js
export const stringMax = (numChar: number, key: string = ''): TypeValidator<string> => {
  if (!isNumber(numChar)) {
    throw new Error('param for stringMax must be number')
  }

  function stringMax (value, _scope = '') {
    const strValue = string(value, _scope)
    if (strValue.length <= numChar) return strValue
    throw validatorError(
      stringMax,
      value,
      _scope,
      key
        ? `string type '${key}' cannot exceed ${numChar} characters`
        : `cannot exceed ${numChar} characters`
    )
  }
  stringMax.type = () => `string(max: ${numChar})`
  return stringMax
}
```

This version is very similar to yours, but it retains greater consistency with the rest of the validation functions by reusing the return value of `string` validation function, using `validationError` instead of `TypeError`, and defining a `type`.

frontend/model/contracts/group.js
taoeffect, line 468: I'd like to improve the security of the rest of the `string` values.

Let's take this opportunity to replace all `string` types in this file with `stringMax` types.

We can define a `MAX_HASH_LEN = 300` and use that for `manifestCid` and `txid`.

For `memo`, we can use `MAX_MEMO_LEN = 4096`.

`sharedValues` please use `GROUP_DESCRIPTION_MAX_CHAR` to match the UI limit.

`mincomeCurrency` you can use `10`.

`chatRoomID`, `toMemberID` and `proposalHash`, use `MAX_HASH_LEN`, and same for `proposalIds` (an array of `MAX_HASH_LEN` strings).

For `reason` the same as `GROUP_DESCRIPTION_MAX_CHAR` seems fine.

For the chatroom `name`, `CHATROOM_NAME_LIMITS_IN_CHARS`, and for chatroom `description`, `CHATROOM_DESCRIPTION_LIMITS_IN_CHARS`.

I think that's all of them. Please use your best judgement for anything I missed, and feel free to do the same for all of the validations in `chatroom.js` and `identity.js`.

The old manual-style validations in custom validation functions can be removed, and if there's nothing left inside of the custom validation function, the validator should be restored to using `objectOf`/`objectMaybeOf` directly.

taoeffect (changes requested):
Fantastic work @SebinSong!

Review ready. :)

SebinSong:
@taoeffect Updated again.

frontend/model/contracts/chatroom.js
```diff
@@ -457,11 +458,11 @@ sbp('chelonia/defineContract', {
     'gi.contracts/chatroom/deleteMessage': {
       validate: actionRequireInnerSignature((data, { state, message: { innerSigningContractID }, contractID }) => {
         objectOf({
-          hash: string,
+          hash: stringMax(MAX_HASH_LEN, 'hash'),
           // NOTE: manifestCids of the attachments which belong to the message
           //       if the message is deleted, those attachments should be deleted too
           manifestCids: arrayOf(string),
-          messageSender: string
+          messageSender: stringMax(MAX_HASH_LEN, 'messageSender')
```
taoeffect, line 466: Just above this is `arrayOf(string)`. Can you modify that to be `arrayOf(stringMax(MAX_HASH_LEN, 'manifestCids'))`?

frontend/model/contracts/chatroom.js
```diff
@@ -396,7 +397,7 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/chatroom/editMessage': {
       validate: actionRequireInnerSignature(objectOf({
-        hash: string,
+        hash: stringMax(MAX_HASH_LEN, 'hash'),
```
taoeffect, line 401: Just below this is `text: string`, can you modify that to be `text: stringMax(MAX_MESSAGE_LEN, 'text')`?

Where `MAX_MESSAGE_LEN` is a new constant of `20000`?

This will also require updating `addMessage` and the UI to add the same hard limit to both the edit field and the submit message field in the UI.

frontend/model/contracts/identity.js
```diff
@@ -18,10 +22,11 @@ import {
 } from './shared/validators.js'
 
 const attributesType = objectMaybeOf({
-  username: string,
+  username: stringMax(IDENTITY_USERNAME_MAX_CHARS, 'username'),
```
taoeffect, line 27: Just below this, email should use `IDENTITY_EMAIL_MAX_CHARS` that's equal to `320`

frontend/model/contracts/identity.js
```diff
@@ -18,10 +22,11 @@ import {
...
   email: string,
+  bio: optional(stringMax(IDENTITY_BIO_MAX_CHARS, 'bio')),
   picture: unionOf(string, objectOf({
-    manifestCid: string,
+    manifestCid: stringMax(MAX_HASH_LEN, 'manifestCid'),
```
taoeffect, line 31: Just above this is a `string` that I assume is a URL. Feel free to set that to `2048` characters.

taoeffect (changes requested):
Great work @SebinSong! I found a few more ones we missed.

frontend/model/contracts/chatroom.js
SebinSong: @taoeffect can this be worked on as a separate issue?

taoeffect: Yeah 👍

taoeffect: Although we can still add that limit here, and later do the UI

SebinSong: > Although we can still add that limit here, and later do the UI

Sounds good.

taoeffect (approved):
Great work @SebinSong !!

PR #2221, opened 2024-07-18
Fix error in making notification body

### LINKED ISSUE #2189: Missing identity contract of group members who are already left from the group
### Problem

While trying to log in in a new device, we don't sync the identity contracts of group members who are already left from the group.
That is the problem because we are using their profiles in several places; notifications, proposals, chatroom, ...
image

### Solution

Need to sync the identity contracts who have ever joined the group

Silver-IT:
### Summary of Changes

- Closes #2189


frontend/model/notifications/templates.js
```diff
@@ -207,6 +207,11 @@ export default ({
       name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${creatorID}`
     }
 
+    if (options.memberID) {
+      // NOTE: update member only when the type is  PROPOSAL_REMOVE_MEMBER
```
taoeffect: This comment doesn't make sense.

There is no check for `PROPOSAL_REMOVE_MEMBER`. Please either update the comment or update the code so that it fits the comment.

taoeffect (changes requested):
Nice work @Silver-IT! One issue found:

taoeffect (approved):

PR #2222, opened 2024-07-18
Remove alert in creating private chatroom

### LINKED ISSUE #2212: Error message in creating private chatroom
### Problem

Even though all works fine, the error message is displayed.
image

### Solution

Fix it

Silver-IT:
### Summary of Changes

- Closes #2212


frontend/controller/actions/group.js
```diff
@@ -651,23 +652,37 @@ export default (sbp('sbp/selectors/register', {
...
+
+    const switchChannelAfterJoined = (contractID: string) => {
+      if (contractID === chatRoomID) {
+        if (rootState[chatRoomID]?.members?.[identityContractID]) {
+          sbp('okTurtles.events/emit', JOINED_CHATROOM, { identityContractID, groupContractID, chatRoomID })
+          sbp('okTurtles.events/off', EVENT_HANDLED, switchChannelAfterJoined)
+        }
+      }
+    }
+    sbp('okTurtles.events/on', EVENT_HANDLED, switchChannelAfterJoined)
```
taoeffect, line 684: This handler needs to be set up before calling `sendMessage`, not after, to make sure it is ready when the message comes back, otherwise it is possible the message will be processed before the handler is setup.

taoeffect (changes requested):
Good work @Silver-IT! One change request:

frontend/controller/actions/group.js
Silver-IT: Nice catch!

taoeffect (approved):

PR #2224, opened 2024-07-18
#2218 - Set upper limit for group pledge amount

### LINKED ISSUE #2218: Set safe char limit for pledge, need, mincome
### Problem

On `testing`, there is no upper limit for pledge, and entering too big of a pledge effectively breaks the app. 

The entry appears to be saved (i.e. no error indicating otherwise) and the user is returned to the dashboard. However, nothing else seems to work again -- no buttons, menus, or links respond to clicks. When refreshed, the Dashboard is blank and the right menu is missing.

User pledges a lot
pledgeInfinite

User refreshes dashboard because nothing is responding to clicks, dashboard goes blank
pledgeInfinite-refreshLosesDashboard

pledge-outOfMemory

 

### Solution

Fix.

SebinSong:
closes #2218


frontend/model/contracts/misc/flowTyper.js
```diff
@@ -332,6 +332,26 @@ export const number = (
   : TypeValidator<number>
 )
 
+export const numberRange = (from: number, to: number, key: string = ''): TypeValidator<number> => {
+  if (!isNumber(from) || !isNumber(to)) { throw new TypeError('Params for numberRange must be numbers') }
+  if (from >= to) { throw new TypeError('Params "to" should be bigger than "from"') }
+
+  function numberRange (value, _scope = '') {
+    number(value, _scope)
```
taoeffect, line 340: Please re-use the return value of `number` (like `stringMax` does)

frontend/model/contracts/shared/constants.js
```diff
@@ -24,7 +24,8 @@ export const GROUP_DESCRIPTION_MAX_CHAR = 500
 export const GROUP_PAYMENT_METHOD_MAX_CHAR = 250
 export const GROUP_NON_MONETARY_CONTRIBUTION_MAX_CHAR = 150
 export const GROUP_CURRENCY_MAX_CHAR = 10
-
+export const GROUP_MAX_PLEDGE_AMOUNT = 1000000000
+export const GROUP_MAX_PLEDGE_AMOUNT_STRING = '1,000,000,000$'
```
taoeffect: The `$` should be in front. But nevermind that, this constant shouldn't be here at all, please use the getter `withGroupCurrency` instead.

taoeffect (changes requested):
Great work @SebinSong! A couple of change requests:

taoeffect (commented):
The way this PR is implemented, the error doesn't appear immediately, but only when you click "Save":

Screenshot 2024-07-18 at 4 01 17 PM

Is there a way to get it to be validated immediately as well, like when letters are added?

Screenshot 2024-07-18 at 4 00 45 PM

frontend/views/containers/access/SignupForm.vue
```diff
@@ -156,7 +156,7 @@ export default ({
         }
       } catch (e) {
         console.error('Signup.vue submit() error:', e)
-        this.$refs.formMsg.danger(e.message)
+        this.$refs.formMsg?.danger(e.message)
```
SebinSong, line 159: boy-scouting the console error occurring during sign-up.

taoeffect (approved):
Fantastic!

PR #2228, opened 2024-07-19
#2192 Remove email from signup

### LINKED ISSUE #2192: Remove Email field from signup form
### Problem

The email field will reduce the number of signups.

We do not currently use it or need it.

If we ever need it, we can always ask for it later.

### Solution

Remove it.

SebinSong:
closes #2192 

As the safest and fastest way, and in case it becomes required in the future again. I didn't remove the email implementation from the `identity.js` contract but rather just made it optional. But let me know if they are removed too.


taoeffect (approved):
Nice job!

Issue #2226, opened 2024-07-18
Heisenbug in group-chat.spec.js (more persistent one)


taoeffect:
### Problem

Yet another (more persistent) heisenbug in group-chat.spec.js:

- https://cloud.cypress.io/projects/q6whky/runs/2721/test-results/745b61bf-dbcd-4d94-a1ea-4aa09bd3de5f/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github
- https://cloud.cypress.io/projects/q6whky/runs/2734/test-results/f0bc02ea-edb9-4958-96e6-9524de1a5ede/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github

`Group Chat Basic Features (Create & Join & Leave & Close)...and removes user3 and logout`

![Screenshot 2024-07-18 at 3 47 40 PM](https://github.com/user-attachments/assets/66e9fcc8-baff-4ce4-9fff-8d6820c6759c)

### Solution

Right now there are a bunch of hacks in that file that use `cy.wait` and they are not even solving the problem. This needs to be properly fixed and all of the `cy.wait`s removed.

PR #2229, opened 2024-07-19
#2226 - fix the group-chat heisenbug

SebinSong:
work on #2226

### DIFF
--- a/test/cypress/integration/group-chat.spec.js
+++ b/test/cypress/integration/group-chat.spec.js
@@ -51,26 +51,29 @@ describe('Group Chat Basic Features (Create & Join & Leave & Close)', () => {
       // considering it sync the chatroom contract from the beginning
     } else {
       const message = selfLeave ? `Left ${channelName}` : `Kicked a member from ${channelName}: ${leaver}`
+      const messageSelectors = {
+        last: 'div.c-message:last-child',
+        secondList: 'div.c-message:nth-last-child(2)'
+      }
+      const assertKickerAndMessageContent = (msgSelector) => {
+        cy.get(`${msgSelector} .c-who > span:first-child`).should('contain', kicker)
+        cy.get(`${msgSelector} .c-notification`).should('contain', message)
+      }
 
-      let isLastElement = true
       if (byProposal) {
         // NOTE: when the member is kicked from the from by proposal
         //       two messages will be created in general chatroom; INTERACTIVE, and NOTIFICATION
         //       INTERACTIVE message should be created before the NOTIFICATION message
         //       but sometimes (only in Cypress) NOTIFICATION message could be created earlier
         //       this block is to handle that heisenbug
-        cy.wait(1000) // eslint-disable-line cypress/no-unnecessary-waiting
-        cy.get('div.c-message:last-child').invoke('attr', 'class').then(classNames => {
-          isLastElement = classNames.includes('is-type-notification')
+        cy.get(messageSelectors.last).invoke('attr', 'class').then(classNames => {
+          const isLastMsgTypeNotification = classNames.includes('is-type-notification')
+          assertKickerAndMessageContent(
+            isLastMsgTypeNotification ? messageSelectors.last : messageSelectors.secondLast
+          )
         })
-      }
-
-      if (isLastElement) {
-        cy.get('div.c-message:last-child .c-who > span:first-child').should('contain', kicker)
-        cy.get('div.c-message:last-child .c-notification').should('contain', message)
       } else {
-        cy.get('div.c-message:nth-last-child(2) .c-who > span:first-child').should('contain', kicker)
-        cy.get('div.c-message:nth-last-child(2) .c-notification').should('contain', message)
+        assertKickerAndMessageContent(messageSelectors.last)
       }
     }
   }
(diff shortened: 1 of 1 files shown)


Silver-IT (changes requested):
Nice work, @SebinSong.

But I don't think this PR fixes the real heisenbug. Even though, the test passed three times in a row, I can't see any updates which fixes the reason of this heisenbug.

The main reason is because INTERACTIVE message is sometimes created after the NOTIFICATION message. Of course, I think it doesn't happen while we are testing, but only happens in Cypress. So the order of two messages (INTERACTIVE, NOTIFICATION) is random and Cypress can not pick which element to assert with.

The previous code of `cy.wait(1000)` was to wait until two messages are all created. But if we remove this code, the last message could be something that we don't know. Neither the NOTIFICATION, nor the INTERACTIVE message. Because the cypress test code could be reached before those two messages are created.

SebinSong:
@Silver-IT 

>  I can't see any updates which fixes the reason of this heisenbug.

I actually did man. It kinda feels like you left a comment straight away without reading the code..

> The main reason is because INTERACTIVE message is created after the NOTIFICATION message. Of course, I think it doesn't happen while we are testing, but only happens in Cypress. So the order of two messages (INTERACTIVE, NOTIFICATION) is random and Cypress can not pick which element to assert with.

I read your inline comment in the code detailing what causes the heisenbug so made below updates based on that, which seems like fixing the bug.

```js
cy.get(messageSelectors.last).invoke('attr', 'class').then(classNames => {
    const isLastMsgTypeNotification = classNames.includes('is-type-notification')
    assertKickerAndMessageContent(
      isLastMsgTypeNotification ? messageSelectors.last : messageSelectors.secondLast
    )
  })
```

Silver-IT:
> The previous code of cy.wait(1000) was to wait until two messages are all created. But if we remove this code, the last message could be something that we don't know. Neither the NOTIFICATION, nor the INTERACTIVE message. Because the cypress test code could be reached before those two messages are created.

@SebinSong, what about this then?

SebinSong:
@Silver-IT I removed that `cy.wait(1000)` cuz it didn't seem to be needed.

Silver-IT:
I can see your updates are same as the previous one. With only one difference which is that you created and used a new function `assertKickerAndMessageContent`.

SebinSong:
@Silver-IT
I called that assertion within the `then()` block is the difference. (when the check for whether it's the last one or the second last one is needed)

SebinSong:
@Silver-IT I thought you did a good job of checking if it's the last message or the second last one via the `classNames`. My updates are only enhancement on top of that (along with removing `cy.wait()`). So, calm down bro..

Silver-IT:
Okay, I will try to test some more with your changes in [this branch](https://github.com/okTurtles/group-income/tree/2226-heisenbug-in-group-chatspecjs-more-persistent-one).

test/cypress/integration/group-chat.spec.js
```diff
@@ -51,26 +51,29 @@ describe('Group Chat Basic Features (Create & Join & Leave & Close)', () => {
...
-      let isLastElement = true
       if (byProposal) {
         // NOTE: when the member is kicked from the from by proposal
         //       two messages will be created in general chatroom; INTERACTIVE, and NOTIFICATION
         //       INTERACTIVE message should be created before the NOTIFICATION message
         //       but sometimes (only in Cypress) NOTIFICATION message could be created earlier
-        //       this block is to handle that heisenbug
-        cy.wait(1000) // eslint-disable-line cypress/no-unnecessary-waiting
-        cy.get('div.c-message:last-child').invoke('attr', 'class').then(classNames => {
-          isLastElement = classNames.includes('is-type-notification')
```
SebinSong, line 64: It looked to me that the intention here was, some actions needed to happen based on the check result of this `isLastElement`. but the fact that the check is happening within the `then()` block whereas the "some actions" are placed outside that was something looking strange and I felt creating the bug. so, made this updates.

(Leaving this PR for other tasks anyways)

SebinSong:
@Silver-IT btw, let's keep `assertKickerAndMessageContent` function I created. I think it simplifies the logic.

Silver-IT:
@SebinSong Yeah, I agree on keeping that function which simplifies the code nicely. 👍

Silver-IT:
I think this PR reduces the percentage for the heisenbug to happen. (But it still happens actually. [here](https://cloud.cypress.io/projects/q6whky/runs/2750/test-results?utm_source=github&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D))

> Why it reduces the percentage?

I think it's because it asserts inside the `then` clause so that the two `last messages` (one to invoke class names, and one to assert with) are same, even though they are not in real. This also makes sense only when the `NOTIFICATION` message is created before the `INTERACTIVE` message and the test case is reached after the `NOTIFICATION` message is created.

> Then, when it still happens?

When we pick the last message and invoke class of it **before all the messages are loaded**. This was the reason why I used `cy.wait(1000)`.  Shortly, `isLastMsgTypeNotification` is not always right value.

SebinSong:
Seems like there is another PR created for the same issue. Closed.

PR #2242, opened 2024-07-20
#2233 - Custom PWA installation in Home.vue

### LINKED ISSUE #2233: Tell mobile users to install the PWA "app"
### Problem

When people visit the site on mobile, the browser doesn't tell them that they can install it as a PWA.

### Solution

On the homepage (where the login and signup buttons are), below them we should add an inset info thing that either tells them to use the browser's features to install the PWA, or better yet, has a button they can click that will trigger trigger the installation.

- https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Trigger_install_prompt

E.g. something here:

Screenshot 2024-07-19 at 10 08 45 AM

SebinSong:
closes #2233 

@taoeffect 
First of all, this implementation is based on [this MDC article](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Trigger_install_prompt), which states below warning:

> Warning: The technique described here depends on the [beforeinstallprompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event) event, which is non-standard and currently only implemented in Chromium-based browsers.

Both Safari desktop/iOS don't support([reference](https://caniuse.com/?search=beforeinstallprompt)) / fire **'beforeinstallprompt'** event, so the logic won't apply to them. (Meaning the promo banner will be hidden on those browsers)
 
Below are some proof screenshots that it worked on my Android chrome when tried with `grunt dev --tunnel`

 [ Chrome Android ]
 
KakaoTalk_Photo_2024-07-20-17-59-24 002

KakaoTalk_Photo_2024-07-20-17-59-23 001

 [ Chrome desktop ]
 
 
image

### DIFF
--- a/frontend/views/pages/Home.vue
+++ b/frontend/views/pages/Home.vue
@@ -51,12 +51,15 @@ main.c-splash(data-test='homeLogo' v-if='!currentGroupId')
         data-test='joinGroup'
       ) Join a Group
 
-  footer.c-footer(v-if='!isLoggedIn')
-    banner-simple.hide-hoverable-device.hide-tablet.c-pwa-promo(severity='info')
+  footer.c-footer(v-if='!isLoggedIn && ephemeral.showPwaPromo')
+    banner-simple.c-pwa-promo(severity='info')
       template(#header='')
         i18n Install this web app on your device for better usability.
 
-      i18n.is-success.is-small.c-install-btn(tag='button' type='button' @click='onInstallClick') Install
+      button-submit.is-success.is-small.c-install-btn(
+        type='button'
+        @click='onInstallClick'
+      ) {{ L('Install') }}
 
   //- TODO: conditionally show this depending on environment variable
   //- footer.c-footer(v-if='!isLoggedIn')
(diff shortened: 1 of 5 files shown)


taoeffect (changes requested):
Awesome work @SebinSong!

Oddly enough I wasn't able to get this to appear in Brave, although that's good (can you make sure it doesn't appear on Desktop browsers?).

Some change requests/adjustments:

- Could you move the box downward so that it's about 30 px from the bottom of the screen?
- Could you adjust the colors so that instead of blue on a dark background, it's more of a different shade of grey when in dark mode?
- Could you also make sure it looks good in light mode too?

I'll test this on mobile when I have time (preparing a talk right now)

SebinSong:
@taoeffect Updated the PR again

**[ Light ]**

light

**[ Dark ]**

dark

frontend/views/pages/Home.vue
```diff
@@ -50,6 +50,19 @@ main.c-splash(data-test='homeLogo' v-if='!currentGroupId')
         @click='openModal("GroupJoinModal")'
         data-test='joinGroup'
       ) Join a Group
+
+  banner-simple.hide-hoverable-device.hide-tablet.c-pwa-promo(
```
SebinSong, line 54: `.hide-hoverable-device.hide-tablet` ensures that this banner does not show in desktop.

taoeffect (approved):
Nice work @SebinSong!

PR #2246, opened 2024-07-23
Fix error that invite link doesn't work inside chat

### LINKED ISSUE #2240: Invite links not working reliably
### Problem

Invite links don't seem to be working properly 100% of the time.

If the user is already in at least one group, the invite link seems to only redirect them to the dashboard of the first group in their group list. (See comments below).

### Solution

Fix.

Silver-IT:
### Summary of Changes

- Closes #2240

### DIFF
--- a/frontend/model/contracts/shared/voting/proposals.js
+++ b/frontend/model/contracts/shared/voting/proposals.js
@@ -44,7 +44,7 @@ export function notifyAndArchiveProposal ({ state, proposalHash, proposal, contr
 export function buildInvitationUrl (groupId: string, groupName: string, inviteSecret: string, creatorID?: string): string {
   const rootGetters = sbp('state/vuex/getters')
   const creatorUsername = creatorID && rootGetters.usernameFromID(creatorID)
-  return `${location.origin}/app/join?${(new URLSearchParams({
+  return `${location.origin}/app/join#${(new URLSearchParams({
       groupId: groupId,
       groupName: groupName,
       secret: inviteSecret,
--- a/frontend/model/contracts/shared/voting/proposals.js
+++ b/frontend/model/contracts/shared/voting/proposals.js
@@ -44,7 +44,7 @@ export function notifyAndArchiveProposal ({ state, proposalHash, proposal, contr
 export function buildInvitationUrl (groupId: string, groupName: string, inviteSecret: string, creatorID?: string): string {
   const rootGetters = sbp('state/vuex/getters')
   const creatorUsername = creatorID && rootGetters.usernameFromID(creatorID)
-  return `${location.origin}/app/join#?${(new URLSearchParams({
+  return `${location.origin}/app/join?${(new URLSearchParams({
       groupId: groupId,
       groupName: groupName,
       secret: inviteSecret,
--- a/frontend/model/contracts/shared/voting/proposals.js
+++ b/frontend/model/contracts/shared/voting/proposals.js
@@ -44,7 +44,7 @@ export function notifyAndArchiveProposal ({ state, proposalHash, proposal, contr
 export function buildInvitationUrl (groupId: string, groupName: string, inviteSecret: string, creatorID?: string): string {
   const rootGetters = sbp('state/vuex/getters')
   const creatorUsername = creatorID && rootGetters.usernameFromID(creatorID)
-  return `${location.origin}/app/join#${(new URLSearchParams({
+  return `${location.origin}/app/join#?${(new URLSearchParams({
       groupId: groupId,
       groupName: groupName,
       secret: inviteSecret,
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -971,6 +971,14 @@ export default (sbp('sbp/selectors/register', {
     const response = await sendMessage({ ...params, data: { ...data, passPayload } })
 
     if (proposalToSend) {
+      // NOTE: sometimes 'notifyProposalStateInGeneralChatRoom' function could be called
+      //       after the 'proposalVote' event is finished its processing (published, received, called process/sideEffect)
+      //       it's not a big problem unless the proposal to remove someone from the group is accepted by the current vote.
+      //       it's when the two messages will be created in general chatroom; one is to tells the proposal is approved
+      //       and the other is to tell the member is left general chatroom.
+      //       the order of two functions that will be called means the order of the two messages in general chatroom.
+      //       so the order of messages isn't always same and that could cause the heisenbug below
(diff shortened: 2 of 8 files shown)


Silver-IT (commented):
Ready for the review!

frontend/model/contracts/shared/voting/proposals.js
```diff
@@ -44,7 +44,7 @@ export function notifyAndArchiveProposal ({ state, proposalHash, proposal, contr
 export function buildInvitationUrl (groupId: string, groupName: string, inviteSecret: string, creatorID?: string): string {
   const rootGetters = sbp('state/vuex/getters')
   const creatorUsername = creatorID && rootGetters.usernameFromID(creatorID)
-  return `${location.origin}/app/join#?${(new URLSearchParams({
+  return `${location.origin}/app/join?${(new URLSearchParams({
```
taoeffect: @corrideat Need your review on this

Silver-IT: Actually, the hash symbol `#` was not needed since we saved all the parameters inside query which lies after `?`.

corrideat: That's wrong, the `#` is needed, see <https://github.com/okTurtles/group-income/commit/f8cac6630b098bf693585db3fd23a60b5a9977a7#diff-ab2099fc20649c2e08f92b0ca5b846b5073fc274c7d740b8e1baec7bcd26b92aR98>

corrideat: (also, if for some reason it works without a `#`, that's a bug: the query must not contain secrets)

taoeffect: @Silver-IT This PR needs to be closed while keeping this URL using a `#`.

The reason we're using a hash here is because the hash prevents the server from learning about what groups the user is joining, and also prevents the invite secret from being shared with the server.

taoeffect (changes requested):
Review ready!

frontend/model/contracts/shared/voting/proposals.js
Silver-IT: > That's wrong, the # is needed.

Why needed?
##
@corrideat, what do you mean by the following code? Why could it be the reason that we should keep using `#`?
I think it's just to get a query (groupId, secret, ...) and nothing else.
```
this.ephemeral.hash = new URLSearchParams(this.$route.hash.slice(1))
```

I used `query` instead of `hash`, and also updated `Join.vue`.
```
this.ephemeral.query = Object.assign({}, this.$route.query)
```
##
> The reason we're using a hash here is because the hash prevents the server from learning about what groups the user is joining, and also prevents the invite secret from being shared with the server.

@taoeffect, what's wrong if we use `query` instead of `hash`? And the format `/join#?...` doesn't make sense. I can't understand the intention of the code lies after `?`. Is it `hash`? or `query`? Why `#` and `?` are there all together?

Silver-IT: > if for some reason it works without a #, that's a bug

Why is it bug? I have updated all the related codes, and works fine all time.

> the query must not contain secrets

The invite link is created, there is a `secret` inside. Why is it fine to contain `secret` in `hash`, and why not in `query`?

taoeffect: > Why is it bug? I have updated all the related codes, and works fine all time.
> The invite link is created, there is a secret inside. Why is it fine to contain secret in hash, and why not in query?

I answered both of these questions in my reply above.

> The reason we're using a hash here is because the hash prevents the server from learning about what groups the user is joining, and also >>>> **prevents the invite secret from being shared with the server.** <<<

If the invite secret is shared with the server then it's not so secret anymore and the server could use it to join the group.

Regarding:

> Why # and ? are there all together?

I'm not sure, perhaps it's to make it easier for using with `URLSearchParams`?

Silver-IT: > If the invite secret is shared with the server then it's not so secret anymore and the server could use it to join the group.

So do you mean the first one is good in terms of secret and the second one is bad?
```
http://localhost:8000/app/join#?groupId=z9brRu3VXVfvXAHFxjuzphoq8mSsPt4Y8fVREeM5g9mPoSVEhFUG&groupName=Dreamers&secret=%5B%22edwards25519sha512batch%22%2Cnull%2C%224nGfJcEZjoVk7MOnPAZlKUBJeQbQthkK2YPAb%2BTXJwxpEIZfXdqJpTuQssNu0MrZUDbGsIrn5bOJdIODk5zRKw%3D%3D%22%5D
```
```
http://localhost:8000/app/join?groupId=z9brRu3VXVfvXAHFxjuzphoq8mSsPt4Y8fVREeM5g9mPoSVEhFUG&groupName=Dreamers&secret=%5B%22edwards25519sha512batch%22%2Cnull%2C%224nGfJcEZjoVk7MOnPAZlKUBJeQbQthkK2YPAb%2BTXJwxpEIZfXdqJpTuQssNu0MrZUDbGsIrn5bOJdIODk5zRKw%3D%3D%22%5D
```

And it's just the format change of invite link. There is no more changes. I can't understand what you mean by secret problem. What is the different on the server side when the user uses these two formats?

taoeffect: > So do you mean the first one is good in terms of secret and the second one is bad?

Yes.

> What is the different on the server side when the user uses these two formats?

The first one sends all of that information to the server and the second one doesn't.

Silver-IT: Now, I think I understood what you mean, @taoeffect. So we should use `hash` instead of `query` for the security reason.

taoeffect: Yep! 👍

Silver-IT (commented):
Ready for the review!

taoeffect (changes requested):
Nice! Thanks for the fixing the hash issue!

One issue:

```
Error ------------------------------ frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:42:35

Cannot assign `url.hash` to `routerOptions.route['hash']` because property `hash` is missing in object literal [1].
[prop-missing]

   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:42:35
   42|               routerOptions.route['hash'] = url.hash
                                         ^^^^^^

References:
   frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js:40:35
   40|             routerOptions.route = { path, query }
                                         ^^^^^^^^^^^^^^^ [1]



Found 1 error
The Flow check failed!
```

taoeffect (approved):
Excellent work @Silver-IT !

Issue #2243, opened 2024-07-21
Wide code snippets in chat don't display correctly


taoeffect:
### Problem

![Screenshot 2024-07-20 at 6 28 34 PM](https://github.com/user-attachments/assets/acb5cbd8-0b0c-4a7e-80bb-0a184d799fcd)

Notice how it's going off and to the right past the scrollbar in the chatroom.

### Solution

- The code block should be inside of a div that has a maximum width that expands to the side of the chat container.
- When text goes beyond that width, a horizontal scrollbar should appear at the bottom of the codeblock (don't wrap the text, in other words)

Silver-IT:
Also I think we need limit on the number of reactions or to wrap them to fix the overflow of reactions.
![image](https://github.com/user-attachments/assets/d08251cb-40ad-49fe-8565-8d44c3099250)

Silver-IT:
Also for the pinned messages too.
![image](https://github.com/user-attachments/assets/212266a0-3a77-473b-9394-b8f9c143e443)

EDITED: Fine to skip this error in this issue. Since it's added to another issue #2250

SebinSong:
@Silver-IT go ahead and open an issue / work on them yourself.

(EDIT: I actually fixed it myself)

Silver-IT:
Okay, I added the issues in #2250.

SebinSong:
@Silver-IT I actually fixed it as part of the ongoing PR. So, no worries. (And good catch 👍)

SebinSong:
Issue resolved in #2247 . closing.

PR #2247, opened 2024-07-23
#2243 - Fix markdown <code> element issue

SebinSong:
#2243 

[ Fix screenshot ]

### DIFF
--- a/frontend/views/containers/chatroom/MessageReactions.vue
+++ b/frontend/views/containers/chatroom/MessageReactions.vue
@@ -82,6 +82,7 @@ export default ({
 .c-emoticons-list {
   padding-left: 3rem;
   display: flex;
+  flex-wrap: wrap;
 
   &.for-type-poll {
     @include phone {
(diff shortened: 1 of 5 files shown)


taoeffect:
Question, the issue solution section says:

> When text goes beyond that width, a horizontal scrollbar should appear at the bottom of the codeblock (don't wrap the text, in other words)

But the screenshot shows the text being wrapped. The reason why I wrote add a horizontal scrollbar is because wrapping text sometimes makes it hard to read source code on a small screen and be able to tell what line is what. Do you think you could update this to use the horizontal scroll bar?

SebinSong:
@taoeffect Updated again with the fix

frontend/views/containers/chatroom/MessageReactions.vue
```diff
@@ -82,6 +82,7 @@ export default ({
 .c-emoticons-list {
   padding-left: 3rem;
   display: flex;
+  flex-wrap: wrap;
```
SebinSong, line 85: This is a fix for the overflowing emoji issues @Silver-IT  found.

taoeffect (approved):
Looking really solid - great work @SebinSong !

PR #2248, opened 2024-07-23
Allow users to force notifications to be marked as read

### LINKED ISSUE #2235: Notifications unread status not being sync
### Problem

On Desktop on the GIG group, all notifications are read, but on my phone on the PWA, these are not only unread, but clicking "Mark as read" doesn't do anything:

Screenshot_20240719-154053

Screenshot_20240719-154051

EDIT: what's odd is that this works on the testing website...

EDIT2: OK, after @Silver-IT joined the group, suddenly the notifications looked like this:

tmp-1721429497316

In other words:

- The two notifications that I had clicked "Mark as read" on PWA that wouldn't mark themselves as read, suddenly became read.
- The new notification came in as unread
- And this time I was *was* able to click "Mark as read" and it did sync

### Solution

Fix.

Silver-IT:
### Summary of Changes

- Closes #2235

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -434,6 +434,8 @@ async function startApp () {
         }
         this.ephemeral.finishedLogin = 'yes'
 
+        sbp('gi.actions/identity/kv/load')
+
         if (this.$store.state.currentGroupId) {
           this.initOrResetPeriodicNotifications()
           this.checkAndEmitOneTimeNotifications()
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -27,7 +27,7 @@ import './controller/service-worker.js'
 import manifests from './model/contracts/manifests.json'
 import { SETTING_CURRENT_USER } from './model/database.js'
 import store from './model/state.js'
-import { CHATROOM_USER_STOP_TYPING, CHATROOM_USER_TYPING, LOGIN_COMPLETE, LOGIN_ERROR, LOGOUT, SWITCH_GROUP, THEME_CHANGE } from './utils/events.js'
+import { CHATROOM_USER_STOP_TYPING, CHATROOM_USER_TYPING, LOGIN_COMPLETE, LOGIN_ERROR, LOGOUT, ONLINE, SWITCH_GROUP, THEME_CHANGE } from './utils/events.js'
 import AppStyles from './views/components/AppStyles.vue'
 import BannerGeneral from './views/components/banners/BannerGeneral.vue'
 import Modal from './views/components/modal/Modal.vue'
@@ -492,6 +492,7 @@ async function startApp () {
           },
           'reconnection-succeeded' () {
             sbp('gi.ui/clearBanner')
+            sbp('okTurtles.events/emit', ONLINE)
           },
           'subscription-succeeded' (event) {
             const { channelID } = event.detail
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -434,7 +434,9 @@ async function startApp () {
         }
         this.ephemeral.finishedLogin = 'yes'
 
-        sbp('gi.actions/identity/kv/load')
+        sbp('gi.actions/identity/kv/load').catch(e => {
+          console.error("Error from 'gi.actions/identity/kv/load' during login:", e)
+        })
 
         if (this.$store.state.currentGroupId) {
           this.initOrResetPeriodicNotifications()
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -483,6 +483,8 @@ async function startApp () {
           },
           online () {
             sbp('gi.ui/clearBanner')
+            sbp('okTurtles.events/emit', ONLINE)
+            console.info('back online!')
           },
           'reconnection-attempt' () {
             sbp('gi.ui/showBanner', L('Trying to reconnect...'), 'wifi')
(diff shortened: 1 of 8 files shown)


taoeffect (commented):
Question: I don't see how this PR will help sync the existing status of the notification back to the client. This seems to be modifying the action of marking all notifications as read, but where is the code for grabbing the latest value of a notification's read status? Is that what should be fixed?

Silver-IT:
> Question: I don't see how this PR will help sync the existing status of the notification back to the client. This seems to be modifying the action of marking all notifications as read, but where is the code for grabbing the latest value of a notification's read status? Is that what should be fixed?

Actually, the syncing notifications status is already implemented. I guess the reason of the difference status between two devices is because that there could be a network issue in one device. So it could fail to listen the KV event and to update notification status on this device.
Actually, the syncing contracts could also be failed when there is a network issue. I also faced this kind of problem while I was testing on my side. Syncing an identity contract was failed, and can not display username (or displayname) only for one member. I investigated this issue and noticed that it was because that his identity contract was not fully synced.

 So what I did in this PR is to allow users to force updating the notifications read status. When the user clicks `Mark all as read`, `'gi.actions/identity/kv/markNotificationStatusRead'` function will be executed even though the values of KV store are already updated. Actually, this function call won't make any changes on KV store side, but an event will be emitted on backend side which will update the notification status of all the devices where the event is being listened.


Actually, there could be two choices.
One is to allow users to force updating notifications status and to get updated to the latest KV store values
And the other is to call `'gi.actions/identity/kv/loadNotificationStatus'` function every time users open the notification card.

I thought the first one is better in terms of performance, so I worked on it.

@taoeffect, do you think there is a real bug in listening KV event? If not, which option do you prefer?

taoeffect:
> And the other is to call 'gi.actions/identity/kv/loadNotificationStatus' function every time users open the notification card.

I don't see how that would help, because this is already called on login by `'gi.actions/identity/kv/load'`, and yet the notification status isn't updated on page refresh. So can you investigate why that is?

Silver-IT:
> I don't see how that would help, because this is already called on login by 'gi.actions/identity/kv/load', and yet the notification status isn't updated on page refresh.

That makes sense.

> So can you investigate why that is?

As I have already said in [here](https://github.com/okTurtles/group-income/issues/2235#issuecomment-2242709024), I couldn't have any chance to encounter this issue on my side. I tried many times, but never encountered. So, it is difficult to investigate.

taoeffect:
> As I have already said in https://github.com/okTurtles/group-income/issues/2235#issuecomment-2242709024, I couldn't have any chance to encounter this issue on my side. I tried many times, but never encountered. So, it is difficult to investigate.

Here's something you can try:

- use https://localhost.run/
- device 1 (desktop) invites device 2 (phone) to group
- add 3rd user and create proposal
- device 1 has its internet connection turned off
- device 2 marks notification as read
- device 1 comes back online

Then do that again if the problem doesn't reproduce, but switch device 1 and device 2 (e.g. if device 1 was desktop, make it phone).

Silver-IT:
@taoeffect, with your recommended workflow, I can see the the second device doesn't get updated with the notifications status even though refresh the browser. After some investigation, I have noticed that we should call `sbp('gi.actions/identity/kv/load')` in other place in `main.js`. Please have a look at the updates.

frontend/main.js
```diff
@@ -434,6 +434,8 @@ async function startApp () {
         }
         this.ephemeral.finishedLogin = 'yes'
 
+        sbp('gi.actions/identity/kv/load')
```
taoeffect: All errors need to handled some way so that users can send us useful error logs if there are problems.

So here we need something like:

```js
.catch(e => {
  console.error("Error from 'gi.actions/identity/kv/load' during login:", e)
})
```

taoeffect (changes requested):
Great work @Silver-IT !!!

Minor request for error handling please:

taoeffect (changes requested):
Actually, one more change request: we want to call `sbp('gi.actions/identity/kv/load')` when we come back online, so could you please do the following:

1. Add an event to `events.js` called `ONLINE`
2. emit this event in `main.js` in the `online` handler **EDIT: and the `'reconnection-succeeded'` handler too**
3. listen for this event in `identity-kv.js` and call `sbp('gi.actions/identity/kv/load')` in the event listener (please copy the same `.catch` handler, but modify the error message to make it clear that the error happened in the online handler, instead of during login)

taoeffect:
When I run `grunt test` locally on this branch I get this heisenbug error as well: #2257

So this PR seems to result in two common heisenbugs:

- #2256 on Github
- #2257 locally on my machine

frontend/main.js
```diff
@@ -488,6 +492,7 @@ async function startApp () {
           },
           'reconnection-succeeded' () {
             sbp('gi.ui/clearBanner')
+            sbp('okTurtles.events/emit', ONLINE)
```
taoeffect, line 497: Also needs to be done above for `online` handler

taoeffect (changes requested):
Great work @Silver-IT ! (And @taoeffect). To deal with the heisenbug I've commented out the problematic line for now and updated issue #2256.

taoeffect (approved):
Great work @Silver-IT ! (And @taoeffect). To deal with the heisenbug I've commented out the problematic line for now and updated issue #2256.

PR #2253, opened 2024-07-25
#2244 - Drag and Drop file upload doesn't work in Safari

### LINKED ISSUE #2244: File drag & drop in chat doesn't work on Safari
### Problem

I found out at last night's demo that dragging & dropping files onto the chat window doesn't work in Safari. 😅 

### Solution

Please fix.

SebinSong:
closes #2244 

[ video - images are uploaded correctly via drag&drop now in Safari. Also, silently discarding a folder works properly too in there]
https://github.com/user-attachments/assets/b5038233-9968-4367-bd5f-e34beed9f821

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -2,7 +2,6 @@
 .c-chat-main(
   v-if='summary.chatRoomID'
   :class='{ "is-dnd-active": dndState && dndState.isActive }'
-  @dragstart='dragStartHandler'
   @dragover='dragStartHandler'
 )
   drag-active-overlay(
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1087,17 +1086,15 @@ export default ({
...
     // Handlers for file-upload via drag & drop action
     dragStartHandler (e) {
+      e.preventDefault()
       // handler function for 'dragstart', 'dragover' events
-      const items = Array.from(e?.dataTransfer?.items) || []
-
-      if (items.some(entry => entry.kind === 'file')) { // check if the detected content is something attachable to the chat.
-        if (!this.dndState.isActive) {
-          this.dndState.isActive = true
-        }
```
SebinSong, line 1096: (If my memory is correct) This check was added in an attempt to prevent a folder from being attached to the chat. But the `filtering out directories` logic ended up  being moved to `SendArea.vue` ([here](https://github.com/okTurtles/group-income/blob/cb73265fd9712e1f89da5cab68e0234d65412427/frontend/views/containers/chatroom/SendArea.vue#L694-L703)) later on instead, so this is not needed here anymore.

The cause of the issue was that the `e?.dataTransfer?.items` of `dragover` event here gets populated in chrome / FF but does not in Safari, btw.

taoeffect (approved):
Seems to work well! I tested Safari, Firefox, and Brave - all work! Great job!

PR #2254, opened 2024-07-25
Fix styling issue of more options menu

### LINKED ISSUE #2234: Ensure "more options" menu in chat is visible when opened
### Problem

If a chat message is too close to the top of the visible chat history, the "more options" menu in chat is hidden (in whole or in part) when the user clicks on it.

chatMoreOptionsHidden

### Solution

The "more options" menu should be visible when opened. If it won't be visible if opened above `⋯`, then it should be opened below `⋯` instead.

Silver-IT:
### Summary of Changes

- Closes #2234

### DIFF
--- a/frontend/views/containers/chatroom/MessageActions.vue
+++ b/frontend/views/containers/chatroom/MessageActions.vue
@@ -46,10 +46,13 @@ menu-parent.c-message-menu(ref='menu')
 
     menu-trigger.is-icon-small(
       :aria-label='L("More options")'
+      @trigger='moreOptionsTriggered'
     )
       i.icon-ellipsis-h
 
-  menu-content.c-responsive-menu
+  menu-content.c-responsive-menu(
+    :class='{ "is-to-down": isToDown }'
+  )
     ul
       menu-item.hide-desktop.is-icon-small(
         tag='button'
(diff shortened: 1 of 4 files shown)


Silver-IT (commented):
Ready for the review!

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -130,6 +133,27 @@ import { MenuParent, MenuTrigger, MenuContent, MenuItem } from '@components/menu
...
+  }, {
+    name: 'Unpin from channel',
+    conditionToShow: this.isAlreadyPinned
+  }, {
+    name: 'Delete message',
+    conditionToShow: this.isDeletable
+  }]
+
+  return moreOptions.filter(option => option.conditionToShow)
+}
```
taoeffect: This is a cool idea 👍 

One thing though: remember all user-facing strings should be wrapped in either `L` function or `i18n`. Here the `L` function would be needed.

taoeffect (changes requested):
Nice work @Silver-IT!

I'm not understanding why there is duplication of the menu items in the markup (HTML) and in the code (`getAvailableMoreOptions`).

Also found one issue.

SebinSong:
Great fix @Silver-IT .

calculating available space is tall enough for menu is a good idea and seems to work well too.
One thing I observed is, depending on the viewport width, the height of each menu item becomes different. (e.g. in below layout, the menu is `54px` tall whereas the calculation is based on the assumption that it's always `36px`). Could you enhance the calculation to take this into account?

image

frontend/views/containers/chatroom/MessageActions.vue
SebinSong: @Silver-IT Agree this is a tidy solution. Since all `this.something` here look like the component props, I think this list can be stored as a computed prop instead, so that `conditionToShow` here are evaluated one time and cached.

Silver-IT: The function `getAvailableMoreOptions` was not used in HTML, so I didn't wrapped name with `L`. 
But I will make computed value `moreOptions` (instead of `getAvailableMoreOptions`) and will wrap `name` with `L` and use it in HTML and `moreOptionsTriggered` function too.
That will fix the DRY too.

SebinSong (dismissed):
Great work @Silver-IT

taoeffect (approved):
Nice job 👍

PR #2263, opened 2024-07-27
fix help and feedback page link

taoeffect:
The new site doesn't have a `/community` page link (should probably add one, but later)


Silver-IT (approved):
LGTM!

PR #2268, opened 2024-07-29
#2164 - Keep payments info when switching between pledger and receiver

### LINKED ISSUE #2164: Save payment info
### Problem

User's payment info is lost when changing from below mincome to pledging. 

1. User does not make mincome, enters payment info.
paymentsBelow

2. User now makes mincome, adjusts income details accordingly. Payment info is removed from UI. 
paymentsAbove

3. User does not make mincome, adjusts income details accordingly. User needs to re-fill in all payment info.
paymentsBelow2

### Solution

Store payment info so user does not have to manually re-enter each time they switch income details to above/below mincome. This will provide better UX (and also help with payments matching, if that's added to the app).

SebinSong:
closes #2164

### DIFF
--- a/frontend/views/containers/contributions/IncomeDetails.vue
+++ b/frontend/views/containers/contributions/IncomeDetails.vue
@@ -174,9 +174,9 @@ export default ({
         return
       }
 
-      const paymentMethods = []
+      let paymentMethodsUpdates = null
 
-      if (this.needsIncome) {
+      if (this.needsIncome && this.$refs.paymentMethods.checkHasUpdates()) {
         this.$refs.paymentMethods.$v.form.$touch()
 
         // Find the methods that have some info filled...
@@ -204,8 +204,9 @@ export default ({
           return
         }
 
+        paymentMethodsUpdates = []
         for (const method of filledMethods) {
-          paymentMethods.push(method)
+          paymentMethodsUpdates.push(method)
         }
       }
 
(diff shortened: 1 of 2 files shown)


frontend/views/containers/contributions/IncomeDetails.vue
```diff
@@ -216,7 +216,11 @@ export default ({
           data: {
             incomeDetailsType,
             [incomeDetailsType]: normalizeCurrency(this.form.amount),
-            paymentMethods
```
SebinSong, line 219: This was the cause of the issue. when the user switches back to `pledger`, `paymentMethods` in its group profile was being reset to `[]`. updated the logic so that this field is not included in the payload unless necessary.

SebinSong:
@taoeffect This PR is ready besides the heisenbug with the group-chat test suite. (which passes when I run the test locally)

taoeffect (changes requested):
Nice work @SebinSong. Found possibly two issues:

1. The pie chart is broken:

Screenshot 2024-08-05 at 11 28 59 AM

2. Did you test whether the changes persist across distribution periods?

I.e. if in period 1 someone is receiving, and period 2 they switch to pledging, and in period 3 they switch back to receiving, did you check whether their data from period 1 is still there?

You can use this code to force a new distribution period in the console:

```js
```

SebinSong:
@taoeffect 

> 1. The pie chart is broken

image

This is screenshot where I tried the same scenario (Assuming the group's mincome is `1000` and the only member of the group is a receiver with `500` as the income) in `master`. The pie chart looks the same there.

> 2. Did you test whether the changes persist across distribution periods?

Yes it works in that scenario too.

taoeffect (approved):
My mistake @SebinSong!

This seems to work really well, great work!

PR #2272, opened 2024-07-30
#2217 - Add UI limit for chat message

### LINKED ISSUE #2217: Add UI limit to chatroom max message length
### Problem

In [this PR](https://github.com/okTurtles/group-income/pull/2211#discussion_r1681921289) a message max limit was added to the contracts, but not the UI.

See: https://github.com/okTurtles/group-income/pull/2211#discussion_r1681921289

### Solution

Update the UI so that messages greater than that limit cannot be typed. It's OK to just cut the message off and not let anymore characters be typed, but if you want to make it fancier with some counter like `n/20000` it's fine too as long as it looks good.

SebinSong:
closes #2217 

Currently `CHATROOM_MAX_MESSAGE_LEN = 20000` and I've added `maxlength` attribute for it (which makes sure string beyond that limit gets cut off), but not that `n/20000` indicator UI because I thought it will be rare that the user types messages longer than that and having it won't be that useful compared to other places.
But let me know if we need it in place here too.


SebinSong:
`stringMax(CHATROOM_MAX_MESSAGE_LEN)` validator in `chatroom.js` contract has already been added before. So no need to take care here.

taoeffect (changes requested):
Nice work @SebinSong!

One issue I found: the UI gets kind of broken in Firefox when I past a very large file. I couldn't even check if the text is being cut off because I'm not able to scroll the page or the input box when it's like this:

Screenshot 2024-08-02 at 1 46 37 PM

Can you maybe add a height limit to the input box itself, and then add a scrollbar that appears so that the contents can be scrolled? (Please verify it also works well on mobile too when a very large amount of text is copy/pasted)

taoeffect (changes requested):
Nice work @SebinSong, I think this is a good height for mobile:

Screenshot 2024-08-03 at 11 32 14 AM

For desktop, I would make it slightly taller.

However, the most important issue is that still I cannot scroll the input box. Can you? When I try to scroll using the scroll wheel it doesn't work. I would also maybe make a vertical scrollbar visible on the righthand side when the text overflows past the max height.

SebinSong:
@taoeffect 
Increased max-height for desktop size screen to `12.25rem` and also fixed the hidden scrollbar bug.

taoeffect (approved):
Excellent!

PR #2280, opened 2024-08-03
rate limit signups

taoeffect:
Thanks so much to @SGrondin's amazing Bottleneck library we can now rate limit signups easily!

### DIFF
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -10,6 +10,29 @@ import path from 'path'
 import chalk from 'chalk'
 import './database.js'
 import { registrationKey, register, getChallenge, getContractSalt, updateContractSalt } from './zkppSalt.js'
+import Bottleneck from 'bottleneck'
+
+const limiterPerMinute = new Bottleneck.Group({
+  strategy: Bottleneck.strategy.LEAK,
+  highWater: 0,
+  reservoir: 1,
+  reservoirRefreshInterval: 60 * 1000,
+  reservoirRefreshAmount: 1
+})
+const limiterPerHour = new Bottleneck.Group({
+  strategy: Bottleneck.strategy.LEAK,
+  highWater: 0,
+  reservoir: 10,
+  reservoirRefreshInterval: 60 * 60 * 1000,
+  reservoirRefreshAmount: 10
+})
+const limiterPerDay = new Bottleneck.Group({
+  strategy: Bottleneck.strategy.LEAK,
+  highWater: 0,
+  reservoir: 50,
+  reservoirRefreshInterval: 24 * 60 * 60 * 1000,
+  reservoirRefreshAmount: 50
+})
 
 // Constant-time equal
 const ctEq = (expected: string, actual: string) => {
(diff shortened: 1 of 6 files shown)


SebinSong (dismissed):

backend/routes.js
```diff
@@ -10,6 +10,29 @@ import path from 'path'
 import chalk from 'chalk'
 import './database.js'
 import { registrationKey, register, getChallenge, getContractSalt, updateContractSalt } from './zkppSalt.js'
+import Bottleneck from 'bottleneck'
+
+const limiterPerMinute = new Bottleneck.Group({
+  strategy: Bottleneck.strategy.LEAK,
```
corrideat, line 24: I don't think any of these values should be hardcoded (although defaults are fine). See how configuration is done, e.g., for the ZKPP keys.

backend/routes.js
```diff
@@ -69,6 +92,17 @@ route.POST('/event', {
         const parsedManifest = JSON.parse(manifest)
         const { name } = JSON.parse(parsedManifest.body)
         if (name !== 'gi.contracts/identity') return Boom.unauthorized('This contract type requires ownership information', 'shelter')
+        // rate limit signups in production
```
corrideat, line 105: Commenting here because it's actually a line above not modified by this PR.

```js
const ip = request.headers['x-real-ip'] || request.info.remoteAddress
```

That line needs to have a comment saying that it's a REQUIREMENT that x-real-ip be a trusted header and that if is setting up a server it MUST be set to a trusted value that CANNOT be overridden by users. If that assumption doesn't hold, this PR is useless and potentially dangerous.

backend/routes.js
```diff
@@ -69,6 +92,17 @@ route.POST('/event', {
...
+        if (process.env.NODE_ENV === 'production') {
```
corrideat: I think that it should be possible to disable this functionality altogether to allow for this to be done at a lower level (yes, it needs to know about contract types, but it's possible to do this with things like OpenResty or commercial WAF offerings such as Cloudflare)

corrideat (changes requested):
See my comment for concrete changes. In addition, I have the following comment:

I think that the documentation (which now doesn't exist, so next best thing is adding a comment somewhere in this file) should explain the limitations of this approach:

1. Limiting by IP addresses is maybe the simplest way of doing rate-limiting, but it has serious limitations:
  a. Because of CGNAT, multiple users (potentially millions, but more likely a few hundred) can be sharing a single IP address. This makes this PR break things for legitimate users (if they're all using Group Income).
  b. On the flip side, it doesn't really stop attacks, because attackers that are rate-limited can easily get a new IP address.
  c. Because of the previous point, this PR can be used to cause a DoS by filling up the list of IP addresses.
  d. **At the very least**, the key used should be the /64 block for IPv6 addresses. Not doing this makes this PR pointless and exposes the server to very easy DoS.
2. Because this state is local to the server, it doesn't work under load balancing scenarios.

backend/routes.js
```diff
@@ -69,6 +92,17 @@ route.POST('/event', {
...
+          try {
+            await limiterPerMinute.key(ip).schedule(() => Promise.resolve())
```
corrideat, line 110: See my comment about why `ip` cannot be used as a key if the address is IPv6.

taoeffect: Something to possibly deal with in the future. On a call Ricardo asked an AI for the code to get this ipv6 key, and this is what he said the AI said:

```js
function getFirstIPv6_64(ipv6Address) {
  // Expand the address if it uses compressed notation
  const expandedAddress = expandIPv6Address(ipv6Address);
  
  // Split the address into its 8 groups
  const groups = expandedAddress.split(':');
  
  // Take the first 4 groups (64 bits) and join them
  return groups.slice(0, 4).join(':');
}

function expandIPv6Address(address) {
  // If the address doesn't contain ::, it's already expanded
  if (!address.includes('::')) {
    return address;
  }

  // Split the address into parts before and after ::
  const [left, right] = address.split('::');
  const leftGroups = left ? left.split(':') : [];
  const rightGroups = right ? right.split(':') : [];

  // Calculate how many zero groups we need to insert
  const missingGroups = 8 - leftGroups.length - rightGroups.length;

  // Create the expanded address
  return [
    ...leftGroups,
    ...Array(missingGroups).fill('0000'),
    ...rightGroups
  ].map(group => group.padStart(4, '0')).join(':');
```

backend/routes.js
```diff
@@ -12,26 +12,34 @@ import './database.js'
 import { registrationKey, register, getChallenge, getContractSalt, updateContractSalt } from './zkppSalt.js'
 import Bottleneck from 'bottleneck'
 
+const MEGABYTE = 1048576 // TODO: add settings for these
```
corrideat, line 15: I'd use `1<<20` or `2**20` which are much more readable and less error-prone

corrideat (dismissed):

corrideat (approved):

PR #2292, opened 2024-08-08
#2245 - Make non-monetary contribution mandatory in IncomeDetails.vue for non-pledgers

### LINKED ISSUE #2245: Require non-monetary contributions for those making less than mincome
### Problem

If people who are receiving monetary income do not add non-monetary pledges, then this can be both bad for the monetary receivers and pledgers as the pledgers may start to feel taken advantage of, and the receivers may start to feel like they aren't contributing anything.

### Solution

Avoid this problem by requiring at least 1 non-monetary pledge in the income details screen for monetary receivers.

This is both a design and implementation issue. @dotmacro said she'd work on the design.

SebinSong:
closes #2245 

Adding below block to `IncomeDetails.vue`

@taoeffect 
Please let me know if you would like to change the validation copy.

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -1061,6 +1061,7 @@ sbp('chelonia/defineContract', {
           with: stringMax(GROUP_NON_MONETARY_CONTRIBUTION_MAX_CHAR, 'with')
         }),
         nonMonetaryRemove: stringMax(GROUP_NON_MONETARY_CONTRIBUTION_MAX_CHAR, 'nonMonetaryRemove'),
+        nonMonetaryReplace: arrayOf(stringMax(GROUP_NON_MONETARY_CONTRIBUTION_MAX_CHAR)),
         paymentMethods: arrayOf(
           objectOf({
             name: stringMax(GROUP_NAME_MAX_CHAR),
@@ -1082,6 +1083,9 @@ sbp('chelonia/defineContract', {
               break
             case 'nonMonetaryEdit':
               nonMonetary.splice(nonMonetary.indexOf(value.replace), 1, value.with)
+              break
+            case 'nonMonetaryReplace':
+              groupProfile.nonMonetaryContributions = cloneDeep(value)
               break
             default:
               groupProfile[key] = value
--- a/frontend/views/containers/contributions/NonMonetaryPledges.vue
+++ b/frontend/views/containers/contributions/NonMonetaryPledges.vue
@@ -1,7 +1,7 @@
 <template lang='pug'>
 fieldset(
   data-test='nonMonetaryPledges'
-  v-error:pledges=''
+  v-error:pledges='{ attrs: { "data-test": "badPledges" } }'
 )
   legend.has-text-bold.c-legend
     i18n.is-title-4 Non-monetary pledge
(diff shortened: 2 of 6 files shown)


frontend/model/contracts/group.js
```diff
@@ -1083,6 +1084,9 @@ sbp('chelonia/defineContract', {
             case 'nonMonetaryEdit':
               nonMonetary.splice(nonMonetary.indexOf(value.replace), 1, value.with)
               break
+            case 'nonMonetaryReplace':
+              groupProfile.nonMonetaryContributions = cloneDeep(value)
+              break
```
SebinSong, line 1089: Adding this feature requires a way to update `groupProfile.nonMonetaryContributions` array as a whole instead of element by element. So adding `nonMonetaryReplace` here.

frontend/views/containers/contributions/NonMonetaryPledges.vue
```diff
@@ -0,0 +1,170 @@
...
+    }
+  },
+  validations () {
+    return {
+      form: {
+        pledges: {
+          [L('At least one non-monetary pledge is required')]: (value) => {
+            return this.optional ||
+              (value?.length && value.some(entry => entry.value))
+          },
```
SebinSong: No need to validate this component when non-monetary contribution is optional. So implementing a custom validator here.

frontend/views/containers/contributions/NonMonetaryPledges.vue
```diff
@@ -0,0 +1,174 @@
...
+    optional: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data () {
+    return {
+      form: {
+        pledges: [
+          { id: randomHexString(10), value: '' }
```
taoeffect, line 68: Is this `id` thing necessary? If all it's used for is is the `:key` in the `v-for`, then can't the position (`index`) in the array be used instead?

frontend/views/containers/contributions/NonMonetaryPledges.vue
```diff
@@ -0,0 +1,174 @@
+<template lang='pug'>
+fieldset(
+  data-test='nonMonetaryPledges'
+  v-error:pledges='{ attrs: { "data-test": "badPledges" } }'
+)
+  legend.has-text-bold.c-legend
+    i18n.is-title-4 Non-monetary pledge
+    i18n.c-optional(v-if='optional') (optional)
```
taoeffect, line 8: Note: typically I'd request that these be made two complete sentence using a computed property, however, I see that in the UI they are visually separate elements, so this is OK.

taoeffect (commented):
Fantastic work @SebinSong!

This is just a comment / question for you before approval.

frontend/views/containers/contributions/NonMonetaryPledges.vue
SebinSong: @taoeffect Can we keep this as is? We have a feature to add/remove an array item here and in this case, having unique id like this can help prevent duplicated keys from being generated accidentally.

taoeffect: yeah, I was was just trying to understand why it was necessary

taoeffect (approved):
Great work!

PR #2293, opened 2024-08-08
#2283 - Chat should not replace symbols in <code>

### LINKED ISSUE #2283: Chat should not replace symbols in <code> (aka <pre>) tags
### Problem

Screenshot 2024-08-05 at 10 36 24 AM

We can't paste code that has a `>` symbol in it into code blocks, so code like this doesn't work:

```js
() => foo
```

### Solution

Not sure.

SebinSong:
closes #2283 

[Proof of the fix]

### DIFF
--- a/frontend/views/utils/markdown-utils.js
+++ b/frontend/views/utils/markdown-utils.js
@@ -40,14 +40,13 @@ export function renderMarkdown (str: string): any {
       entryText = entryText.replace(/</g, '&lt;').replace(/>/g, '&gt;')
 
       entryText = entryText.replace(/\n(?=\n)/g, '\n<br>')
-        .replace(/<br>\n(\s*)(\d+\.|-|```)/g, '\n\n$1$2') // custom-handling the case where <br> is directly followed by the start of block-code (```)
+        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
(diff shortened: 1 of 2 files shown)


frontend/views/utils/markdown-utils.js
```diff
@@ -33,11 +33,21 @@ export function renderMarkdown (str: string): any {
...
-    .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
+      entryText = entryText.replace(/\n(?=\n)/g, '\n<br>')
+        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
+        .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
+
+      entry.text = entryText
+    }
+  })
+
+  str = combineMarkdownSegmentListIntoString(strSplitByCodeMarkdown)
```
SebinSong, line 50: The cause of the issue was that,
These additional fixes & adjustments related to white-spaces, line-breaks and `<` and `>` did not need to be applied to the contents of `<code>...</code>`.
So a logic added here where it selectively apply these adjustments to the text contents outside of `<code>...</code>`.

taoeffect (approved):
Fantastic!

Issue #2271, opened 2024-07-29
Clean up and simplify login flow


taoeffect:
### Problem

The way logging in right now works is extraordinarily confusing and complicated.

There's events being thrown here and there, event handlers across multiple files, and the whole process involves lots of code in lots of places across `main.js`, `controller/actions/identity.js`, and `controller/app/identity.js`.

### Solution

- Clean up the login flow, especially in `main.js`.
- Cease the use of string literals like `'CHELONIA_STATE'` and be consistent with the rest of the code (define it as a constant somewhere)
- Shouldn't `sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, Array.from(this.subscriptionSet))` be done at the end of `'chelonia/reset'` (after `newState` is applied)?
- Remove the use of `_private` selectors, e.g. `'gi.app/identity/_private/login'` etc. Either they are not necessary at all, or they are proven necessary, in which case an anonymous function inside the original selector is preferred.

corrideat:
> Shouldn't sbp('okTurtles.events/emit', CONTRACTS_MODIFIED, Array.from(this.subscriptionSet)) be done at the end of 'chelonia/reset' (after newState is applied)?

I don't think it makes a difference as that entire block is synchronous. However, the answer would be no, not necessarily. `CONTRACTS_MODIFIED` should be emitted every time the `subscriptionSet` changes.

> Remove the use of _private selectors, e.g. 'gi.app/identity/_private/login' etc. Either they are not necessary at all, or they are proven necessary, in which case an anonymous function inside the original selector is preferred.

They may look ugly, but I think they're necessary. Besides keeping things in sync, they also help debug issues, as the relevant call is logged this way.

Even if the selector is removed (which would remove some debug information), I'd use a named constant and not `sbp('queue', 'foo', () => {})`, because that'd unnecessarily use indentation (but this is mostly an aesthetic concern).

corrideat:
> Clean up the login flow, especially in main.js.

> Cease the use of string literals like 'CHELONIA_STATE' and be consistent with the rest of the code (define it as a constant somewhere)

Those seem reasonable

corrideat:
> Clean up the login flow, especially in main.js.

I've been inspecting the code, and took a look especially at `main.js`. Although `main.js` is relatively complex, and more so than it should be, I'm not convinced that the situation is _that bad_.

[...]

Issue #2108, opened 2024-06-21
Additional contract upgrade / migration mechanism


taoeffect:
### Problem

There are several situations when we may want a way to perform explicit contract upgrade/migration:

- New state state is added to the constructor
- Changes to how contracts are retained in sideEffects

#### New state state is added to the constructor

For example, see this PR: https://github.com/okTurtles/group-income/pull/2101/files

Because `pinnedMessages` was added to the `chatroom.js` state in the constructor, it caused a problem when attempting to pin messages and an existing deployed site that didn't have that variable defined.

#### Changes to how contracts are retained in sideEffects

#2105 is an example of changing the behavior of how contracts are retained: https://github.com/okTurtles/group-income/pull/2105/files#diff-65d789b5ed021c8e1547a855fc7347654957fa7198761ad2c55d689dc1a0c95aL1757

Here's a more explicit example of the problem:

![Screenshot 2024-06-21 at 11 31 58 AM](https://github.com/okTurtles/group-income/assets/138706/79dad4da-d182-4ae1-8f4a-43e8ea3dfd91)

### Solution

Implement a function that takes current version and previous version and uses that to conditionally perform upgrades.

This would be implemented directly on the contract itself, and would be called each time a contract instance is loaded with a new manifest.

```js
upgrade ({ previousVersion }, { state ) {
  if (versionCompare(previousVersion, '0.5.1') < 0) {
    state.pinnedMessages = []
  }
}
```

Example from call related to retain/release:

![Screenshot 2024-06-21 at 11 38 15 AM](https://github.com/okTurtles/group-income/assets/138706/07708ff0-4b7b-4cb9-ba47-8073e4f9274a)

Issue #1957, opened 2024-04-19
Server-side and client-side mechanism to prevent contract version downgrade attack


taoeffect:
### Problem

Issue #1956 made me realize that there's a potential way attackers can exploit old vulnerabilities.

For example, to exploit the issue in #1956 even after it is "patched" would involve sending an action that references an old contract version/manifest. This will force clients to run old vulnerable code, even though the latest patched version is available.

### Solution

1. Give the server the ability to read the `manifest.json` file of any apps that the server supports, and only accept actions that uses one of those manifest IDs (EDIT: this might be difficult to do while supporting federation...). Return a special error if an old version is used to send an action, so that clients know to upgrade their app if necessary.
2. Have clients do the same. Note: this check should only apply to new messages, not old ones (as otherwise we won't be able to process old events in the contract chain). It's not entirely clear to me how we can distinguish between "old" and "new" events, but perhaps the simplest way is to not allow version downgrades. Once an action with a new version of that contract is seen, messages from older versions shouldn't be processed. Since the server also enforces the version, it shouldn't be an issue if an old version of a running client tries to send a message - that message should fail to be sent with a special error that tells the client it needs to upgrade its app in order to send new messages.

PR #2294, opened 2024-08-11
Fix #2277

### LINKED ISSUE #2277: Invalid negative reference count when deleting a chatroom
### Problem

1. Create a chatroom
2. Delete it

Screenshot 2024-08-01 at 12 00 02 PM

### Solution

Don't know how this snuck in...

corrideat:
- Fix reference count errors
- Implement re-syncing contracts with foreign keys on `OP_KEY_SHARE` on foreign contract


corrideat:
~~TODO: Add comments explaining changes made~~

frontend/model/contracts/group.js
```diff
@@ -1220,14 +1220,19 @@ sbp('chelonia/defineContract', {
     'gi.contracts/group/leaveChatRoom': {
       validate: actionRequireActiveMember(objectOf({
         chatRoomID: stringMax(MAX_HASH_LEN, 'chatRoomID'),
-        memberID: optional(stringMax(MAX_HASH_LEN), 'memberID')
+        memberID: optional(stringMax(MAX_HASH_LEN), 'memberID'),
+        // `reference` is the height used in the corresponding join action
+        reference: numberRange(1, Number.MAX_SAFE_INTEGER)
```
taoeffect: The term `reference` sounds rather vague, like it could be anything.

But it seems like it must be a very specific type of value: the last known `joinedHeight` value as seen from the perspective of someone (else?) in the group. So two change requests:

1. An updated comment here to better explain what this value is, e.g. "last known `joinedHeight` value as seen from the perspective of someone else (or ourselves) in the group"
2. An updated variable name, e.g. `lastKnownJoinedHeight` or `recentJoinedHeight`

frontend/model/contracts/group.js
```diff
@@ -1267,6 +1280,47 @@ sbp('chelonia/defineContract', {
...
+              skippedRetain
+            ) {
+              return
+            }
+
+            // Release chatroom if we're sure `retain` was called
+            sbp('chelonia/contract/release', data.chatRoomID).catch(e => {
+              console.error(`[gi.contracts/group/leaveChatRoom/sideEffect] Error releasing chatroom ${data.chatRoomID}`, e)
+            })
+          })
```
taoeffect: Did you say that it might be possible to simplify this code by adding a new feature to `okTurtles.eventQueue`? If so, let's do that.

frontend/model/contracts/group.js
```diff
@@ -1412,8 +1466,7 @@ sbp('chelonia/defineContract', {
   methods: {
     'gi.contracts/group/_cleanup': ({ contractID, state }) => {
       // unsubscribe from other group members identity contract
-      const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
-      const possiblyUselessContractIDs = Object.keys(state.profiles || {}).filter(cID => cID !== identityContractID)
```
taoeffect, line 1420: Why was this change made?

I remember it used to be this way, and `identityContractID` was added to prevent some bug.

taoeffect (changes requested):
Preliminary review

shared/domains/chelonia/internals.js
```diff
@@ -774,14 +773,64 @@ export default (sbp('sbp/selectors/register', {
...
+                targetState._vm.authorizedKeys[keyId]._notBeforeHeight >= newestEncryptionKeyHeight
+              ) {
+                return contractIDs
+              }
+              return []
+            })))
+            // Mark these contracts as dirty
+            contractIdsToUpdate.forEach((contractID) => {
+              const targetState = cheloniaState[contractID]
+              if (!targetState) return
```
taoeffect, line 810: Hmm, under what circumstance would this happen, and does it make sense to return instead of creating the state?

frontend/model/contracts/identity.js
```diff
@@ -202,13 +202,13 @@ sbp('chelonia/defineContract', {
...
         const { groupContractID, inviteSecret } = data
-        if (has(state.groups, groupContractID)) {
+        if (has(state.groups, groupContractID) && state.groups[groupContractID].active) {
           throw new Error(`Cannot join already joined group ${groupContractID}`)
         }
 
         const inviteSecretId = await sbp('chelonia/crypto/keyId', new Secret(inviteSecret))
 
-        state.groups[groupContractID] = { hash, inviteSecretId }
+        state.groups[groupContractID] = { hash, inviteSecretId, active: true }
```
taoeffect: This is a rather huge / significant change.

What is this `active` thing, why is it being added, and doesn't this have the potential to break all existing groups that have been created and joined and do not have this property?

taoeffect (changes requested):
Moar comments!

frontend/model/contracts/group.js
corrideat: >  was added to prevent some bug.

That bug should be fixed instead.

> Why was this change made?

If the reference count is correct, this code is simpler. There is no need not to remove ourselves because there should still be a positive reference count for our own identity contract. OTOH, if the reference count is wrong, that bug should be fixed. This PR is supposed to correct those errors.

shared/domains/chelonia/internals.js
corrideat: I think this shouldn't normally happen (if it does, it likely indicates stale references). However, there is no 'state' to create. If the contract is synced again, it'll have the correct state as it's being synced from scratch.

frontend/model/contracts/identity.js
corrideat: For correctness, `state.groups[groupContractID]` is no longer deleted. This is needed for the leave action to work correctly, as the `reference` property needs to be available, which it won't be if the state is removed.

> doesn't this have the potential to break all existing groups that have been created and joined and do not have this property?

Yes, but I'd say all of the changes in this PR are breaking. Thanks for pointing that out.

frontend/model/contracts/group.js
corrideat: I don't have specific suggestions right now for the changes needed to `eventQueue`. I can work on a specific proposal.

frontend/model/contracts/group.js
taoeffect: Are we sure that the reference count on our own contractID then is in fact correct in this case?

EDIT: maybe a comment could be added here to acknowledge this scenario? E.g. something like:

```js
// NOTE: it is safe to decrease the reference count on our own identityContractID
// because it has an extra retain in {location}
```

frontend/model/contracts/identity.js
taoeffect: See:

- `'state/vuex/postUpgradeVerification'` in `state.js`
- https://github.com/okTurtles/group-income/issues/2108

corrideat: I think the easiest change right now is to make this a negative property (e.g., `inactive` / `left`) so that it doesn't break existing groups.

`'state/vuex/postUpgradeVerification'` seems inappropriate to solve this because:

  * It deals with Vuex state instead of contract state
  * It is executed in the browsing context. This would need to run in the service worker.
  * It is entirely unaware of contract versions. Doing this correctly needs awareness of contract versions.

Implementing #2108 would address this to some extent. However, I think that the preferred solution to this would require a solid in-contract upgrade mechanism. Looking at how contracts are handled now, I see several deficiencies:

1. It don't think versions are checked at all, which means that someone could downgrade a contract. This should probably be prevented.
2. Because contracts can be downgraded, they need to be forward and backwards compatible.

taoeffect: > It don't think versions are checked at all, which means that someone could downgrade a contract. This should probably be prevented.

Yep there's #1957 to address exactly that.

> Because contracts can be downgraded, they need to be forward and backwards compatible.

It's fine to assume contracts will never be downgraded in the sense that contracts should never be downgraded and we will never downgrade them. If someone hacks the app to do this they are responsible for any app breakage, not us.

frontend/model/contracts/group.js
corrideat: How much of an issue is this code? I originally thought that changing the `eventQueue` could work, but upon closer examination, I think that that alone can't fix it.

I've already described the situation in [Slack](https://okturtles.slack.com/archives/C02QF71TK1Q/p1723240199009389), but the gist of this is that either:

  1. We need to increment / decrement the reference count without using any event queue, which would simplify the code but result in potentially unnecessary contract syncs (for example, you join and leave a group; the `retain` will result in syncing that group only for the `release` causing it to be removed, and the situation is worsened for multiple join / leaves, and also by chatrooms that are synced as a side effect of groups), or
  2. We keep on using an event queue to alleviate this concern. However, for the reference count to be correct, we need to know that `retain` was called before calling `release`. This needs to be somewhere in the state, and since side-effects can't change the contract state, I'm using `okTurtles.data`.

corrideat: If you have any questions about what the specific issue is, please let me know, because understanding it well is crucial to discussing a solution.

corrideat: I think an alternative solution is something I discussed earlier (which would, however, complicate reference counting).

The solution in this case would be that references, instead of being a number, which needs to be managed very carefully, are an identifier. This way, you have an explicit 'reason' for being subscribed, which can be removed. This is less error prone, but perhaps sounds more complicated (or not).

In this particular case, since the reference already tracks the reason, we can call /release unconditionally, and it doesn't matter if /retain wasn't called. We could either make it not throw, or we could ignore the error.

Needless to say, this would be a breaking change.

frontend/model/contracts/identity.js
corrideat: See new commit addressing this by using a negated property.

frontend/controller/app/group.js
```diff
@@ -28,9 +28,11 @@ sbp('okTurtles.events/on', LEFT_GROUP, ({ identityContractID, groupContractID })
   // grab the groupID of any group that we're a part of
   const currentGroupId = rootState.currentGroupId
   if (!currentGroupId || currentGroupId === groupContractID) {
-    const groupIdToSwitch = Object.keys(state.groups)
+    const groupIdToSwitch = Object.entries(state.groups)
+      // $FlowFixMe[incompatible-use]
+      .map(([cID, { hasLeft }]) => !hasLeft && cID)
       .filter(cID =>
-        cID !== groupContractID
+        cID && cID !== groupContractID
```
taoeffect, line 35: Is this `cID &&` business necessary? You already do `&& cID` in the `.map`...

frontend/model/contracts/chatroom.js
```diff
@@ -287,7 +287,6 @@ sbp('chelonia/defineContract', {
           }
         }
 
-        state.members[memberID].leftHeight = height
```
taoeffect, line 290: Was this not used for anything?

taoeffect (commented):
Some questions

frontend/model/contracts/chatroom.js
corrideat: See the line below that makes this line pointless

frontend/controller/app/group.js
corrideat: It is, because it could be `null` / `undefined` / `false`, &c.

corrideat: Because of the `cID !== groupContactID` part it could be fine without (if we're sure `groupContractID` is never falsy), but I think explicit is good because it lowers the cognitive load to understand exactly what's going on.

frontend/model/contracts/group.js
```diff
@@ -1220,14 +1220,19 @@ sbp('chelonia/defineContract', {
     'gi.contracts/group/leaveChatRoom': {
       validate: actionRequireActiveMember(objectOf({
         chatRoomID: stringMax(MAX_HASH_LEN, 'chatRoomID'),
-        memberID: optional(stringMax(MAX_HASH_LEN), 'memberID')
+        memberID: optional(stringMax(MAX_HASH_LEN), 'memberID'),
+        // `joinedHeight` is the height used in the corresponding join action
+        joinedHeight: numberRange(1, Number.MAX_SAFE_INTEGER)
```
taoeffect, line 1217: So... in this action we can expect `joinedHeight` to be in `data`, but below where it says:

```js
state.chatRooms[data.chatRoomID].members[memberID].joinedHeight !== data.joinedHeight
```

(And similar)

Is this a problem for existing clients who did not set `joinedHeight` because they used an older version of `'gi.contracts/group/joinChatRoom'`?

frontend/model/contracts/group.js
```diff
@@ -1267,6 +1280,47 @@ sbp('chelonia/defineContract', {
...
+            // Has `retain` been skipped?
+            // Note: we check for 'skipped' instead of 'called' because
+            // side-effects cannot mutate state. The 'skipped' case can be
+            // checked without persistent storage, while the 'called' case
+            // would require some kind of persistent state (since the reference
+            // count is itself persisted)
+            const skippedRetain = sbp('okTurtles.data/get', `gi.contracts/group/chatroom-skipped-${contractID}-${data.chatRoomID}-${data.joinedHeight}`)
+            // There's no use of this transient value, so it's deleted to
+            // preserve memory
+            sbp('okTurtles.data/delete', `gi.contracts/group/chatroom-skipped-${contractID}-${data.chatRoomID}-${data.joinedHeight}`)
```
taoeffect: I feel like this `gi.contracts/group/chatroom-skipped-${contractID}-${data.chatRoomID}-${data.joinedHeight}` value should be saved to a const instead of re-typed here just to be safe and also DRY things

frontend/model/contracts/group.js
```diff
@@ -1267,6 +1280,47 @@ sbp('chelonia/defineContract', {
...
+
+            const { identityContractID } = sbp('state/vuex/state').loggedIn
```
taoeffect: You already grabbed this value above... why are you re-grabbing it? Either shouldn't be done, or comment needs to justify/explain

EDIT: I see a comment below, but I would move that above here and elaborate further. e.g. this seems like a Cypress thing

frontend/model/contracts/identity.js
```diff
@@ -298,10 +296,19 @@ sbp('chelonia/defineContract', {
...
-          const { groupContractID } = data
+          const { groupContractID, reference } = data
+
+          // For an explanation of skippedRetain, see the chatroom contract.
+          // TL;DR: This ensures that the reference count is correct when a
+          // group has been joined and then left and the state is being
+          // synced from scratch.
+          const skippedRetain = sbp('okTurtles.data/get', `gi.contracts/identity/group-skipped-${groupContractID}-${reference}`)
+          // No use for this value beyond this point.
+          sbp('okTurtles.data/delete', `gi.contracts/identity/group-skipped-${groupContractID}-${reference}`)
```
taoeffect: Same here I'd DRY `gi.contracts/identity/group-skipped-${groupContractID}-${reference}`

frontend/model/contracts/group.js
corrideat: Hm, I think it'd indeed be an issue. A way to possibly fix this could be to test for presence of this attribute (also, it won't be in `data` for old contracts). Those contracts may have issues with reference counts without this information.

frontend/model/contracts/group.js
corrideat: Sure

frontend/model/contracts/group.js
corrideat: >  this seems like a Cypress thing

This is not a Cypress thing (other than the fact that users will likely not be that fast).

frontend/model/contracts/identity.js
corrideat: Sure

frontend/model/contracts/group.js
corrideat: Another way would be to implement contract upgrades. In this case, we'd set joinedHeight to the height of the update.

taoeffect:
Based on conversion, another consideration is that this PR changes how the identity contract is retained, and so it's possible that old versions of the app could remove their own identity contract when they are leaving a room. This can be mitigated by checking if we're upgrading an old versions of the app to this new version of the app (whatever version number we give it), and then issuing an extra `retain` on the identity contract in, for example, `postUpgradeVerification`.

Alternatively, in `postUpgradeVerification` we can check if `joinedHeight` isn't defined and if so issue an action to the contract to perform the upgrade and in the side effect issue the retain.

Yet a third option may be reverting the changes in this PR and keeping the old behavior of retain/release on the user's identityContractID, this would avoid the need to implement a sideEffect on the upgrade action that's sent, and seems like the simplest option.

corrideat:
As I continue to work on this PR, I see some additional issues with correctness and the way we do state management.

We've had this conversation on Slack already about why I think the model of _only process can change the state_ is limiting in the context of reference counting. I've found another situation that is even harder to handle correctly.

Let's say that you are leaving a group. When you leave a group, the group contract reference is decremented, and the leave action also decrements the reference count on the group members. I think I may have got that more or less working.

However, let's say that now the group contract is removed (by calling `chelonia/contract/remove`). In this case, the group contract should undo all of the changes it made to references, but doing so is rather difficult to do without knowing exactly which how the reference count was affected. One potential solution to this is making the remove method async, which would make removeImmediately also async, but I'm still unsure this solves the issue of correctly changing reference counts, because this needs to be done in a way that is stateless.

For example, let's say you're a member of the group, then you leave and then the group contract is removed.

In this case, by the time the leave `process` function is called, you're no longer an active member, and the state looks the same to both the `sideEffect` for the leave action and the remove function, which means that they both have to rely on one of them having reduced the reference count. I think in this case, it makes sense of it to be done in the leave action.

On the other hand, if the group contract is removed without first having left the group, it needs to look at the state and decrease the reference count for chatrooms and members. For this to work, the remove function would need to push something on the queue. If this isn't done, those references will be left dangling.

I'll be looking further into how to keep a correct reference count in all of these edge cases.

corrideat:
To update on this issue: I'm working on a fuzzer to properly design and test a long term solution.

frontend/model/contracts/shared/functions.js
```diff
@@ -265,3 +261,43 @@ export function swapMentionIDForDisplayname (
     .map(t => regEx.test(t) ? swap(t) : t)
     .join('')
 }
+
+export const referenceTally = (selector: string): Object => {
```
taoeffect, line 334: Could you add a comment here explaining here what this is?

corrideat: Sure. Preliminarily, so that it doesn't block your review, this is a temporary / ephemeral reference count that is kept in memory until the 'real' release or retain operation is called.

So, if you join / leave / join / leave, and all of those actions are processed at once, this ensures that the reference count isn't affected, as those operations cancel out.

This is similar to what we were doing previously, except that this works and the other thing didn't. Before, the count managed by side-effects, which tried to guess based on the state.

The main difference is that, before, side effects pushed to the end of the queue each. Now, a single function is pushed to the end of the queue, no matter the number of actions.

corrideat: Comment added.

corrideat:
I've fixed a bug that came from wrongly-done refactoring. Now the tests are passing again (12 times in a row, see b1b9c9c, a00ed0d and 9d0de13).

I think this is once again ready for review.

frontend/model/contracts/shared/functions.js
```diff
@@ -265,3 +261,113 @@ export function swapMentionIDForDisplayname (
...
+        if (count != null) return
+        sbp('chelonia/queueInvocation', parentContractID, () => {
+          const count = sbp('okTurtles.data/get', key)
+          sbp('okTurtles.data/delete', key)
+          if (count && count !== Math.sign(count)) {
+            console.warn(`[${selector}] Unexpected value`, parentContractID, childContractID, count)
+            if (process.env.CI) {
+              Promise.reject(new Error(`Unexpected value ${parentContractID} ${childContractID}: ${count}`))
+            }
+          }
```
taoeffect: Isn't this `count !== Math.sign(count)` a bit presumptive...?

It sounds like it's expecting `referenceTally` to be called only in situations where there are "guaranteed matching pairs of increments/decrements" or something to that effect.

Is that really always going to be the case wherever this function is ever used?

EDIT: I see - ok that makes sense. In that case can you add the `${selector}` to the error message in the `Error` object? It might make it easier to track down on failure.

taoeffect (commented):
Question

frontend/model/state.js
```diff
@@ -29,6 +29,7 @@ const initialState = {
   contracts: {}, // contractIDs => { type:string, HEAD:string, height:number } (for contracts we've successfully subscribed to)
   loggedIn: false, // false | { username: string, identityContractID: string }
   namespaceLookups: Object.create(null), // { [username]: sbp('namespace/lookup') }
+  reverseNamespaceLookups: Object.create(null), // { [contractID]: username }
```
taoeffect, line 32: I think down below in `postUpgradeVerification` you are missing adding this `reverseNamespaceLookups`

frontend/controller/namespace.js
```diff
@@ -9,6 +9,10 @@ sbp('sbp/selectors/register', {
     const cache = sbp('state/vuex/state').namespaceLookups
     return cache?.[name] ?? null
   },
+  'namespace/lookupReverseCached': (id: string) => {
+    const cache = sbp('state/vuex/state').reverseNamespaceLookups
+    return cache?.[id] ?? null
```
taoeffect: After implementing `postUpgradeVerification` should be able to rewrite this as a 1-liner:

```js
return sbp('state/vuex/state').reverseNamespaceLookups[id]
```

frontend/controller/app/identity.js
```diff
@@ -109,7 +109,8 @@ sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams,
       if (cheloniaState.namespaceLookups) {
         Vue.set(state, 'namespaceLookups', cheloniaState.namespaceLookups)
       }
-    // End exclude contracts
+      // End exclude contracts
+      sbp('state/vuex/postUpgradeVerification', state)
```
taoeffect, line 113: Yikes. Scary this was missed by both of us.

Yet another reason all of this must be cleaned up (#2271).

frontend/model/contracts/group.js
```diff
@@ -301,13 +301,13 @@ function updateGroupStreaks ({ state, getters }) {
...
-const removeGroupChatroomProfile = (state, chatRoomID, member) => {
-  state.chatRooms[chatRoomID]['members'] =
+const removeGroupChatroomProfile = (state, chatRoomID, member, ourselvesLeaving) => {
+  state.chatRooms[chatRoomID].members =
     Object.fromEntries(
       Object.entries(state.chatRooms[chatRoomID].members)
         .map(([memberKey, profile]) => {
           if (memberKey === member && (profile: any)?.status === PROFILE_STATUS.ACTIVE) {
-            return [memberKey, { ...profile, status: PROFILE_STATUS.REMOVED }]
+            return [memberKey, { ...profile, status: ourselvesLeaving ? PROFILE_STATUS.LEFT_GROUP : PROFILE_STATUS.REMOVED }]
```
taoeffect: Why is this function written in such a bizarre way?

Wouldn't a straightforward implementation be a significantly more efficient O(1) 1-liner?

```js
state.chatRooms[chatRoomID].members[member].status = ourselvesLeaving ? PROFILE_STATUS.LEFT_GROUP : PROFILE_STATUS.REMOVED
```

Also, did you check all related code that checks the status for whether or not it handles `LEFT_GROUP` appropriately?

For example, below on line 322 we have a check for `status !== PROFILE_STATUS.REMOVED`, but what if the status is `LEFT_GROUP`?

Likewise, on 1262 we have another check:

```js
state.chatRooms?.[data.chatRoomID]?.members[memberID]?.status === PROFILE_STATUS.REMOVED
```

It's worth asking here: do we need this distinction between `REMOVED` and `LEFT_GROUP`? Or could we just have a single variable here to act as an opposite to `ACTIVE`? The logic could be greatly simplified in that case and these kinds of bugs would be less likely to occur.

taoeffect (changes requested):
Along with other comments (see above) left a few more:

frontend/controller/namespace.js
corrideat: This function is essentially a copy-paste of the function above, handling the possibility of `reverseNamespaceLookups` (or namespaceLookups) not being defined. That was because the side-effects in the identity contract call the `lookupCached` selector, and I thought that an incorrect state should not cause the side-effect to fail. The `??null` part also coerces `undefined` into null, so that the function returns only two types instead of three (`string`, `undefined`, `null`).

frontend/model/state.js
corrideat: That's right

frontend/controller/namespace.js
taoeffect: OK, I can get with `?? null` but not `cache?.[id]`. That should just be `cache[id]`.

frontend/model/contracts/shared/functions.js
corrideat: What about changing the error message to be more descriptive, like `Unexpected reference tally value`? I could add the selector to the error message, but if you look closely, that is already logged above, so I think it should be easy to find.

Regardless, I'll add a comment explaining why an exception is thrown.

taoeffect: Yeah `Unexpected reference tally value` sounds good 👍

frontend/model/contracts/group.js
corrideat: > Wouldn't a straightforward implementation be a significantly more efficient O(1) 1-liner?

Probably. I think the idea was not to mutate data, but yeah, your suggestion is simpler and more performant.

> did you check all related code that checks the status for whether or not it handles LEFT_GROUP appropriately?

Yes. Most of that code is fine as is because this will only be set on our own profile. However, this change prevents us from sending our own leave actions to chatrooms when we are leaving a group (instead, someone else will need to send them). I think this was the intended change, but I'm not sure it's still relevant given the latest changes.

> It's worth asking here: do we need this distinction between REMOVED and LEFT_GROUP?

As noted, the change was part of an earlier solution, and might no longer be needed. I'll try that out.

frontend/controller/actions/group.js
```diff
@@ -1030,7 +1030,7 @@ export default (sbp('sbp/selectors/register', {
   ...encryptedAction('gi.actions/group/updateSettings', L('Failed to update group settings.')),
   ...encryptedAction('gi.actions/group/updateAllVotingRules', (params, e) => L('Failed to update voting rules. {codeError}', { codeError: e.message })),
   ...encryptedAction('gi.actions/group/updateDistributionDate', L('Failed to update group distribution date.')),
-  ...encryptedAction('gi.actions/group/upgradeFrom1.0.6', L('Failed to upgrade from version 1.0.6')),
+  ...encryptedAction('gi.actions/group/upgradeFrom1.0.7', L('Failed to upgrade from version 1.0.6')),
```
taoeffect: Oops, forgot to bump the error message too 👆

taoeffect (approved):
This was a tough one — amazing work @corrideat! 💪

PR #2297, opened 2024-08-12
#2269 - Only show history if it exists

### LINKED ISSUE #2269: Dashboard: only show history if it exists
### Problem
Note: This is very similar to issue #2163 

On `testing`, as soon as the group is created, the dashboard "support %" and "todo %" tabs both correctly indicate that there is no history yet. (This was not tested on `live`.)

noHistory-testingCorrect

However, on `live` and `testing`, as soon as the first distribution period starts, the dashboard "support %" and "todo %" tabs both show 0% for the period preceding the first distribution, even though that period didn't exist -- meaning no users were instructed to send payments, nor could they have marked payments as sent in the app. 

noHistory--testingIncorrectSupport
noHistory-testingIncorrectTodo

### Solution

Only show history when history exists. The first distribution period should be the first period tracked on the `Support`, `TODO`, and `Activity` tabs.

SebinSong:
closes #2269 

Updated the filtering logic so that it takes into account group-creation date as well.

### DIFF
--- a/frontend/views/containers/payments/PaymentsMixin.js
+++ b/frontend/views/containers/payments/PaymentsMixin.js
@@ -15,6 +15,7 @@ const PaymentsMixin: Object = {
       'dueDateForPeriod',
       'groupPeriodPayments',
       'groupSettings',
+      'groupCreatedDate',
       'groupSortedPeriodKeys',
       'ourIdentityContractId',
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
This LGTM!

PR #2303, opened 2024-08-16
Clarify use of sync. Closes #2298

### LINKED ISSUE #2298: Fix confused and outdated 'chelonia/contract/sync' situation
### Problem

Our [`Information-Flow.md` documentation is outdated](https://github.com/okTurtles/group-income/blob/master/docs/Information-Flow.md#subscribing-to-a-contract) for how to sync a contract.

It currently says that developers should use `'chelonia/contract/sync'`, but that's wrong, as they're supposed to use `'cheloina/contract/retain'` along with `'chelonia/contract/release'`.

A separate problem is this code inside of `'chelonia/contract/sync'`:

```js
      if (!forcedSync && this.subscriptionSet.has(contractID)) {
        if (!rootState[contractID]?._volatile?.dirty) {
          return sbp('chelonia/private/queueEvent', contractID, ['chelonia/private/noop'])
        }
      }
```

It's unclear to me what this code is doing, and it should be clearly commented.

### Solution

1. Search the Group Income side of the app and replace all calls to `'chelonia/contract/sync'` with `'chelonia/contract/retain'` (and `release`).
2. Then anywhere the concept of resyncing a currently sync'd and retained contract is needed, either call a new selector (e.g. `'chelonia/contract/forceSync'`), or change the behavior of `'chelonia/contract/sync'` so that it does force syncing. Be careful if you pick the "change behavior" route, since if any old versions of contracts that call that expecting the old behavior to be there, then the behavior shouldn't be changed (since those are deployed for life now). My preference would be to change the behavior for `'chelonia/contract/sync'`, so that `forceSync` is `true` by default, and tell developers to call this on contracts upon logging in (or coming back online). Go through the app and search for places where we need to explicitly set `forceSync: false`.
3. Update the [Information Flow section](https://github.com/okTurtles/group-income/blob/master/docs/Information-Flow.md#subscribing-to-a-contract) with the latest correct guidance (use `retain` most of the time, and `sync` upon login).

corrideat:
Fix #2298


frontend/controller/actions/identity.js
```diff
@@ -170,35 +170,39 @@ export default (sbp('sbp/selectors/register', {
...
-              headers: {
-                'authorization': await sbp('chelonia/shelterAuthorizationHeader', message.contractID()),
-                'content-type': 'application/x-www-form-urlencoded'
-              },
-              body: new URLSearchParams({
-                'r': r,
-                's': s,
-                'sig': sig,
-                'Eh': Eh
+            await sbp('chelonia/contract/retain', message.contractID(), { ephemeral: true })
```
taoeffect, line 174: Shouldn't this be inside of the `try` block so that it's guaranteed to be paired with the `finally`?

frontend/controller/actions/utils.js
```diff
@@ -310,7 +310,8 @@ export function groupContractsByType (contracts: Object): Object {
   const contractIDs = Object.create(null)
   if (contracts) {
     // $FlowFixMe[incompatible-use]
-    Object.entries(contracts).forEach(([id, { type }]) => {
+    Object.entries(contracts).forEach(([id, { references, type }]) => {
+      if (!references) return
```
taoeffect, line 324: Could you add a comment here explaining under what circumstances `references` might not exist?

frontend/controller/app/identity.js
```diff
@@ -324,7 +324,7 @@ export default (sbp('sbp/selectors/register', {
         sbp('okTurtles.events/off', LOGIN_ERROR, loginErrorHandler)
 
         const errMessage = e?.message || String(e)
-        console.error('Error during login contract sync', e)
+        console.error('[gi.app/identity] Error during login contract sync', e)
```
taoeffect: Nice 👍

frontend/controller/actions/index.js
```diff
@@ -208,10 +208,5 @@ sbp('sbp/selectors/register', {
         }
       })
     }
-
-    // TODO: Temporary sync until the server does signature validation
-    // This prevents us from sending messages signed with just-revoked keys
-    // Once the server enforces signatures, this can be removed
-    await sbp('chelonia/contract/sync', contractID, { force: true })
```
taoeffect, line 215: Great catch!

taoeffect (changes requested):
Great work @corrideat! Some questions

frontend/controller/actions/utils.js
corrideat: Sure. Short story: for ephemeral references and 'internal' references (i.e., foreign keys)

frontend/controller/actions/identity.js
corrideat: Perhaps counterintuitively, no.

```js
const retain = () => { throw new Error('Retain failed') }
const release = () => { throw new Error('Release should be unreachable in this example') }

console.error('Test 1')
try {
  (() => {
    try {
      retain()
      // ...
    } finally {
      release()
    }
  })()
} catch (e) { console.error(e) }

console.error('Test 2')
try {
  (() => {
    retain()
    try {
      // ...
    } finally {
      release()
    }
  })()
} catch (e) { console.error(e) }
```

This gives:

```
Test 1
Error: Release should be unreachable in this example
    at release (<anonymous>:2:31)
    at <anonymous>:11:7
    at <anonymous>:13:5

Test 2
Error: Retain failed
    at retain (<anonymous>:1:30)
    at <anonymous>:19:5
    at <anonymous>:25:5
```

The correct behaviour is the second one, meaning that `release` should only be called if `retain` succeeded.

corrideat: In Java, this is called _`try` with resources_. See <https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html> (since this `try` syntax isn't valid JavaScript, see the _Prior to Java SE 7_ example).

corrideat:
Updated!

frontend/controller/actions/utils.js
```diff
@@ -309,8 +309,15 @@ export async function createInvite ({ contractID, quantity = 1, creatorID, expir
...
+    // Note: `references` holds non-ephemeral references (i.e., explicit
+    // calls to `retain` without `{ ephemeral: true }`). These are the contracts
+    // that we want to restore.
+    // Apart from non-ephemeral references, `references` may not be set for
+    // contracts being 'watched' for foreign keys.
     // $FlowFixMe[incompatible-use]
-    Object.entries(contracts).forEach(([id, { type }]) => {
+    Object.entries(contracts).forEach(([id, { references, type }]) => {
+      // If the contract wasn't explicitly retained, skip it
+      if (!references) return
```
taoeffect: This is definitely an improvement. 👍 

Can you also add an explanation for *why* we are skipping syncing foreign-key contracts?

taoeffect (changes requested):
Since the build is failing and we need to re-run it anyway, can you push a commit with an elaborated comment?

frontend/views/pages/PendingApproval.vue
```diff
@@ -54,15 +55,35 @@ export default ({
...
+        if (reset) return
+        sbp('chelonia/contract/release', this.ephemeral.groupIdWhenMounted, { ephemeral: true }).catch(e => {
+          console.error('[PendingApproval.vue]: Error releasing contract', e)
+        })
+      })
+    }
+    sbp('okTurtle.events/on', CHELONIA_RESET, listener)
+  },
+  beforeDestroy () {
+    this.ephemeral.ondestroy?.()
```
taoeffect: What is all of this new logic and why has it been added here?

I'm not sure this is a good idea... did you test this with multiple groups? It's possible that `beforeDestroy` could be called as a result of the group switcher switching to a different group. If that happens it would break the joining attempt on the new group that we're trying to join.

taoeffect (commented):
Concern:

frontend/views/pages/PendingApproval.vue
corrideat: This isn't really new logic, rather redefining the old logic that called `/sync`, which is unsafe to call.

> It's possible that beforeDestroy could be called as a result of the group switcher switching to a different group

I don't see how that'd break anything. The `onDestroy` is there to call `release`, which is the correct thing to do after calling `retain`. If the view is destroyed, there's no point in keeping the group around (if it shouldn't be released, the reference count should indicate that). Note that before this only called `sync` and it also didn't affect the reference count.

> did you test this with multiple groups?

No, I'm not sure how that'd be done as joining a group would open a new URL?

taoeffect: > No, I'm not sure how that'd be done as joining a group would open a new URL?

You'd create and join group1 as u1. Then u2 (in a separate container tab or browser) would send you an invite link to group2, but before u1 visits it, u2 would log out. Then u1 visits the link to join u2's group2 and get stuck on this pending page. Then they'd switch back to their own group. Then u2 would log back in.

> I don't see how that'd break anything. The onDestroy is there to call release, which is the correct thing to do after calling retain. If the view is destroyed, there's no point in keeping the group around (if it shouldn't be released, the reference count should indicate that). Note that before this only called sync and it also didn't affect the reference count.

I see. Could you add a comment here as well then to explain where the "real" (persistent) retain happens on this group and why we're calling ephemeral retain here instead of `sync`?

corrideat: > You'd create and join group1 as u1. [...]

I did this and nothing broke, as expected, since this is merely a UI concern.

> Could you add a comment here as well then

Done. Also added a comment to the effect that `vue` files should not be managing references.

taoeffect (approved):
Approved! Fantastic work @corrideat!

PR #2306, opened 2024-08-17
Fix #2125. Add message sender when available

### LINKED ISSUE #2125: Ability to get contractID of person who sent a GIMessage (innerSigningContractID() function)
### Problem

In PR #2046 there is the desire to display who sent a GIMessage that resulted in an error:

Screenshot 2024-06-27 at 10 49 28 AM

It seems like there's no simple way to get this information from a `GIMessage`.

### Solution

Implement `innerSigningContractID()` on `GIMessage` that returns this info if available as a `string`, otherwise `null`.

Update the code [here](https://github.com/okTurtles/group-income/pull/2046#discussion_r1657560017) (see screenshot above if link doesn't work) to restore the `contractID` information for that notification so that we can know what user sent the failing message.

corrideat:


frontend/model/notifications/templates.js
```diff
@@ -24,14 +25,22 @@ export default ({
...
+    let who
+
+    if (message.innerSigningKeyId()) {
+      const innerSigningContractID = findContractIDByForeignKeyId(state[message.contractID()], message.innerSigningKeyId())
+      who = innerSigningContractID && `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${innerSigningContractID}`
+    }
 
     return {
-      body: L("{errName} during {activity} for '{action}' to '{contract}': '{errMsg}'", {
+      body: L(who ? "{errName} during {activity} for '{action}' from {b_}{who}{_b} to '{contract}': '{errMsg}'" : "{errName} during {activity} for '{action}' to '{contract}': '{errMsg}'", {
```
taoeffect: The `L` function is not allowed to have a ternary operator inside of it. It is only allowed to have a **string literal** and (optionally) an object for the arguments inside of the string literal.

A ternary operator can be used like so though: `who ? L(...) : L(...)`

taoeffect (changes requested):
Nice work @corrideat ! One minor issue found.

taoeffect (changes requested):
Nice work @corrideat! Think I found a problem: to test, updated `group.js` contract code for changing chatroom description like so:


```js
    'gi.contracts/group/changeChatRoomDescription': {
      validate: actionRequireActiveMember(objectOf({
        chatRoomID: stringMax(MAX_HASH_LEN, 'chatRoomID'),
        description: stringMax(CHATROOM_DESCRIPTION_LIMITS_IN_CHARS, 'description')
      })),
      process ({ data, direction }, { state }) {
        state.chatRooms[data.chatRoomID]['description'] = data.description
        if (direction === 'incoming') {
          throw new Error('chatroom description error')
        }
      }
    },
```

Then I made the chatroom description edit using **u1**.

This was seen by u1:

Screenshot 2024-09-15 at 1 07 54 PM

This was seen by u2, notice how it doesn't say "u1" but it should:

Screenshot 2024-09-15 at 1 07 45 PM

frontend/model/contracts/group.js
```diff
@@ -1339,7 +1339,10 @@ sbp('chelonia/defineContract', {
         chatRoomID: stringMax(MAX_HASH_LEN, 'chatRoomID'),
         description: stringMax(CHATROOM_DESCRIPTION_LIMITS_IN_CHARS, 'description')
       })),
-      process ({ data }, { state }) {
+      process ({ data, direction }, { state }) {
+        if (direction === 'incoming') {
+          throw new Error('chatroom description error')
+        }
```
taoeffect: This should be removed.

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -222,6 +222,26 @@ const periodicNotificationEntries = [
...
+            if (cID !== contractID) {
+              console.error(`[periodic notification] Mismatched username. The lookup result was ${cID} instead of ${contractID}`)
+            }
+          }).catch((e) => {
+            console.error('[periodic notification] Error looking up username', username, 'for', contractID, e)
+          })
+        })
+      },
+      shouldClearStateKey: () => true
+    }
```
taoeffect: Instead of doing this periodically, can we do it a single time on page refresh? (E.g., after we've logged in?)

And did you double-check that that condition is what we need for this? 🧐

taoeffect (changes requested):
Minor change requests

frontend/model/contracts/group.js
corrideat: Yeah, that certainly shouldn't be there. Good catch.

taoeffect (approved):
LGTM! Great job @corrideat !

PR #2315, opened 2024-08-22
Compound PR for two simple UI updates #2310 / #2200

### LINKED ISSUE #2200: Voting % should not go to 0
### Problem

During group creation, the user can select "0" as the voting percentage.

zeroPercentCreation

However, because "there should always be at least 2 `yes` votes" (which is correct and good), the 0% threshold will never be honored. (By contrast, it's possible for 2 "yes" votes to be >1% of a group.) 

zeroPercent-groupsettings
zeroPercentProposal

### Solution

Change lower value to 1, so slider goes from 1-100.

### LINKED ISSUE #2310: Remove "Privacy & Safety" user setting
### Problem

"TODO" is shown:

Screenshot 2024-08-20 at 9 27 02 AM

### Solution

We actually don't have a good idea of what should go here right now. The app has privacy & safety built in to everything already, so for now just hide this menu item.

SebinSong:
closes #2310 
closes #2200

### DIFF
--- a/frontend/views/components/VotingRulesInput.vue
+++ b/frontend/views/components/VotingRulesInput.vue
@@ -47,7 +47,7 @@ export default ({
     config: {
       [RULE_PERCENTAGE]: {
         sliderLabel: L('What percentage of members need to agree to pass a proposal?'),
-        sliderMin: 0,
+        sliderMin: 1,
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Nice!

PR #2316, opened 2024-08-22
#2158 - Fix the broken 'Reduced-motion' mode

### LINKED ISSUE #2158: "Reduced Motion" setting does nothing
### Problem

Screenshot 2024-07-05 at 10 13 05 AM

This setting doesn't seem to affect the transition animations when showing/hiding notifications and other modals.

### Solution

Respect this setting and disable all transitional animations.

SebinSong:
closes #2158 

it appears that this feature was broken when the vuex-state for app-settings was moved to a separate module. So fixed it.

While at it, created a `.force-motion` css class for any part of the app where the animation still needs to happen regardless of the reduced-motion setting.
(eg. this highlight animation for `move-to-specific-message` feature in the chat-room.)

Cheers,

### DIFF
--- a/frontend/assets/style/_transitions.scss
+++ b/frontend/assets/style/_transitions.scss
@@ -1,5 +1,14 @@
 @import "_variables";
 
+@mixin reduce-motion {
+  *:not(.force-motion),
+  *:not(.force-motion)::after,
+  *:not(.force-motion)::before {
+    transition: none !important;
+    animation: none !important;
+  }
+}
+
 .fade-enter-active,
 .fade-leave-active {
   transition: opacity 0.5s cubic-bezier(0.165, 0.84, 0.44, 1);
@@ -161,21 +170,11 @@
 }
 
 @media (prefers-reduced-motion: reduce) {
-  *,
-  *::after,
-  *::before {
-    transition: none !important;
-    animation: none !important;
-  }
+  @include reduce-motion;
 }
 
 .js-reducedMotion {
-  *,
-  *::after,
-  *::before {
-    transition: none !important;
-    animation: none !important;
-  }
+  @include reduce-motion;
 }
 
 @keyframes focused {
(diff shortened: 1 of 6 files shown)


frontend/views/components/Tooltip.vue
```diff
@@ -15,7 +15,7 @@ span.c-twrapper(
 
   .c-tooltip(
     :style='styles'
-    :class='{"has-text-center": isTextCenter, "is-active": isActive, manual, "is-dark-theme": $store.getters.isDarkTheme}'
+    :class='{"has-text-center": isTextCenter, "is-active": isActive, manual, "is-dark-theme": $store.getters.isDarkTheme,  "in-reduced-motion": isReducedMotionMode }'
```
SebinSong, line 18: Tooltip needs a special treatment here because, when it becomes active, it's appended to `<body>` not the root Vue component (`<div id='#app'>`)

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -1,5 +1,5 @@
 <template lang='pug'>
-.c-message(
+.c-message.force-motion(
```
SebinSong, line 2: Message highlight animation still needs to happen regardless of `reduced-motion` setting.

frontend/assets/style/_transitions.scss
```diff
@@ -1,5 +1,14 @@
 @import "_variables";
 
+@mixin reduce-motion {
+  *:not(.force-motion),
+  *:not(.force-motion)::after,
+  *:not(.force-motion)::before {
```
SebinSong, line 6: I noticed some animations happen in the `::after`, `::before` pseudo elements. eg) checkbox, toggle-button.
So took that into account here too.

taoeffect (approved):
Fantastic work @SebinSong! Tested and it works well. 👍

PR #2317, opened 2024-08-24
Simplify and explain login flow

corrideat:


taoeffect (commented):
Question: is this PR finished or is there more work to do?

I ask because as far as the jumping around between events related to login that happens when following the logic of `actions/identity.js` and `app/identity.js`, that seems unchained. Is there no way to simplify that part / make it clearer?

corrideat:
> Question: is this PR finished or is there more work to do?

As mentioned on Slack, the only thing missing (other than feedback) is completing the `LOGIN_FLOW.md` file.

> I ask because as far as the jumping around between events related to login that happens when following the logic of actions/identity.js and app/identity.js, that seems unchained. Is there no way to simplify that part / make it clearer?

I'm going to assume you mean _unchanged_ other than _unchained_. Even then, I'm still unsure what you're asking specifically.

If you're talking about _where_ the event listeners are defined, they are in those files as per your request when those event listeners were defined. There's no particular reason why they couldn't be in a different location.

If you're talking about the events themselves, that's the result of the service worker and the app being different contexts entirely and of both needing to know about the currently logged in user.

taoeffect:
> I'm going to assume you mean unchanged other than unchained.

Yes, that was a typo.

This PR does a great job of simplifying `main.js`. As-is, it's an improvement over what we have.

What I'm wondering is, is it possible in any way at all, to simplify the logic that's inside of `gi.app/identity/login` and `gi.actions/identity/login` (and related code)? Is this something you can take a closer look at? If nothing can be done to simplify it, then I suppose we'll have to live with it as-is, but if something can be done to simplify it, it would be awesome and also this would probably be the time to do it.

corrideat:
> gi.app/identity/login and gi.actions/identity/login

I've taken a look and it looks pretty simple to me already (both about 100 lines with relatively straightforward logic). I'll add some brief additional comments to the uncommented sections.

corrideat:
I consider this PR complete now unless there are additional requests. I highly recommend taking the time to read through `LOGIN_FLOW.md`. It's an extensive file that presents in detail the motivations behind everything, documents where things happen and why and walks though specific scenarios.

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+JSON, the data types supported are predefined, and, importantly, functions and
+custom object types are _not_ supported. The `serdes` library in
+`shared/domains/chelonia` is a collection of functions to augment the
+capabilities of the structured clone algorithm.
+
+#### Specifics about Service Workers
+
+Unlike other types of web workers, there can be a instance of single service
+worker for any given page. A single service worker can be shared across multiple
+browsing contexts (e.g., if the user opens many tabs with Group Income, they
```
taoeffect: How about instead of "Unlike other types of web workers, there can be a instance of single service worker for any given page"

Something like: "Unlike other types of web workers, there can only be a single service worker per website."

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+events are other events that are conveyed to let individual tabs know that the
+current user session changed. In general terms, it can be said that Group Income
+(the app) is event-driven as a consequence of using a service worker architecture,
+and that presentation depends on correctly sending and processing events.
+
+## Login process
+
+Since logging in and out occurs in the browsing context as a result of user
+interaction, it's special in the sense that it's a 'shared state' event that
+originates in the browsing context. This makes the process be harder to picture
```
taoeffect: Typo: "This makes the process be harder"

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+are not prompted to log in again when they have already done so in the past.
+
+### Login events
+
+#### Overview
+
+  * `LOGIN`: The `LOGIN` event represents an active login intent, which is to
+   say, it represents that there is an active login process, which may either
+   fail or succeed. This event can originate in the service worker or in a local
+   browsing context.
```
taoeffect: Could you elaborate a bit on this? Why would it originate in the service worker? Why would it originate in the browsing context? Under what circumstances?

EDIT: oh, I see, there's more details down below. (Maybe mention that here?)

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+would be thrown directly for the caller to handler. The reason for this is that,
+while all tabs need to know that a user recently logged in, only the caller
+needs to know when it logging in fails (for example, to display feedback in the
+login form)
+
+###### `app/identity.js`
+
+Exceptionally, the `LOGIN` event can also be emitted locally in `app/identity.js`.
+This happens when _there already is an active session_ and
+`'gi.actions/identity/login'`` is not called (i.e., when opening a new tab of
```
taoeffect: Typo: `'gi.actions/identity/login'`` (extra backtick)

LOGIN_FLOW.md
corrideat: Indeed. That sentence needs correcting because it looks like some editing went wrong. However,

> worker per website

That's inaccurate, depending on how you define website. You can also have multiple service workers defined (e.g., during upgrades), but at most a single one can be active.

corrideat: This is how they say a similar thing on MDN:

> A single service worker can control many pages. Each time a page within your scope is loaded, the service worker is installed against that page and operates on it.

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+
+The `LOGIN` event is emitted in the service worker after calling
+`'gi.actions/identity/login'`. Its purpose is to let all open tabs that a new
+active session exists.
+
+Note that there is no corresponding `LOGIN_ERROR` emitted here, because an error
+would be thrown directly for the caller to handler. The reason for this is that,
+while all tabs need to know that a user recently logged in, only the caller
+needs to know when it logging in fails (for example, to display feedback in the
+login form)
```
taoeffect, line 217: The way this is written makes it sound like if the `LOGIN` event is emitted in the service worker, it is *not* emitted in the browsing context. Is that right? If so, it doesn't make much sense because down below for the `app` section it says:

> The purpose of
emitting this event is to trigger the various handlers that set up the app state
based on an existing session.

Wouldn't these handlers still need to get triggered when logging in the for the first time?

Can this section be updated to address this?

LOGIN_FLOW.md
```diff
@@ -0,0 +1,355 @@
...
+`'gi.actions/identity/login'`` is not called (i.e., when opening a new tab of
+Group Income on a device where the user has already logged in). The purpose of
+emitting this event is to trigger the various handlers that set up the app state
+based on an existing session.
+
+##### Handlers
+
+###### `app/identity.js`
+
+The handler in `app/identity.js` will react to a login event and load a copy of
```
taoeffect: Should this read "will react to a `LOGIN` event" (instead of "login")?

LOGIN_FLOW.md
corrideat: That's right. `LOGIN` can happen in the SW (when using a password) _(x)or_ in the browsing context (when an active session exists).

LOGIN_FLOW.md
corrideat: Right

LOGIN_FLOW.md
corrideat: Note: the reason the event is used in the second case is for (my) convenience as the logic is already there. Technically, it can be done without an event since it's all local. However, doing it _without_ using an event would make the logic slightly more convoluted as an `if` and a function definition would be needed.

taoeffect (changes requested):
Excellent work on `LOGIN_FLOW.md`! Left some requests for changes & typo fixes.

corrideat:
Thanks! Updated as requested.

taoeffect (approved):
Alright @corrideat - this PR is quite something. I tested it, and it seems to work. The tests pass. Let's cross our fingers nothing sneaky broke. 🤞

PR #2319, opened 2024-08-25
DRY kv/set

### LINKED ISSUE #2289: DRY: 'chelonia/kv/set' should always use 'chelonia/queueInvocation'
### Problem

The frontend code for calling `'chelonia/queueInvocation'` is littered with calls to `'chelonia/queueInvocation'`, e.g.:

```js
    return sbp('chelonia/queueInvocation', identityContractID, () => {
      return sbp('chelonia/kv/set', identityContractID, KV_KEYS.UNREAD_MESSAGES, data, {
        encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
        signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
        onconflict
      })
    })
```

```js
    return sbp('chelonia/queueInvocation', identityContractID, () => {
      return sbp('chelonia/kv/set', identityContractID, KV_KEYS.PREFERENCES, data, {
        encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'cek'),
        signingKeyId: sbp('chelonia/contract/currentKeyIdByName', identityContractID, 'csk'),
        onconflict
      })
    })
```

Etc.

This:

- Violates DRY
- But more importantly, is a serious developer-side UX bug. Developers should not be forced to know to do this.

EDIT: similar changes may need the done for `/get`, please [see this message on Slack](https://okturtles.slack.com/archives/C0EH7P20Y/p1722966450525149) about an error.

### Solution

Go through the entire codebase and remove all wrapped `'chelonia/kv/set'` calls. Instead, have `'chelonia/kv/set'` itself call `'chelonia/queueInvocation'`.

corrideat:
Closes #2289


frontend/controller/actions/index.js
```diff
@@ -213,5 +213,14 @@ sbp('sbp/selectors/register', {
...
     await sbp('chelonia/contract/sync', contractID, { force: true })
+  },
+  'gi.actions/kv/set': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+    return sbp('chelonia/queueInvocation', contractID, () => {
+      return sbp('chelonia/kv/set', contractID, key, data, {
+        encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, encryptionKeyName),
+        signingKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, signingKeyName),
+        onconflict
+      })
+    })
```
taoeffect: Since we're creating a framework for other developers too, and this seems like it would be a very common pattern, could you create a `chelonia` namespaced selector for this instead of using `gi.actions`?

taoeffect (changes requested):
Nice work @corrideat!

One slight change request:

frontend/controller/actions/index.js
corrideat: What should this be called? `'chelonia/kv/set'` is already taken.

taoeffect: `'chelonia/kv/queuedSet'`?

shared/domains/chelonia/chelonia-utils.js
```diff
@@ -0,0 +1,13 @@
...
+export default (sbp('sbp/selectors/register', {
+  'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+    return sbp('chelonia/queueInvocation', contractID, () => {
+      return sbp('chelonia/kv/set', contractID, key, data, {
+        encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, encryptionKeyName),
+        signingKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, signingKeyName),
+        onconflict
+      })
+    })
+  }
```
taoeffect, line 19: Request:

- could be this placed next to the the existing selector `'chelonia/kv/set'` so that it's easier to see?
- could you please add a comment above it explaining why usually calling this version is preferred and what this queueing accomplishes?

taoeffect (changes requested):
Minor organizational and comment-related change request

shared/domains/chelonia/chelonia-utils.js
corrideat: > could be this placed next to the the existing selector 'chelonia/kv/set' so that it's easier to see?

I think `chelonia.js` is already cluttered as it is. The reason for putting this in a different file is also that IMHO this belongs in a possibly optional library as it makes some assumptions about which keys are defined (i.e., the csk and the cek) which isn't done in Chelonia currently.

When encryptedAction is moved into Chelonia, it would also be placed in this new file, because it also is a 'suggested' way of doing a common action, but not something that applies to all situations.

> Could you please add a comment above it explaining why usually calling this version is preferred

Sure, although calling this version isn't 'preferred', just less much verbose in a common situation. I can explain that.

taoeffect: Sure, if you want to leave it in that file, then please add a comment above `'chelonia/kv/set'` to tell people to look at `'chelonia/kv/queuedSet'` which is preferable (and where to find that).

taoeffect (approved):
Beautiful!

PR #2320, opened 2024-08-25
#2076 - Fix the group placeholder icon issue

### LINKED ISSUE #2076: placeholder group icon should be an initialism
### Problem

When there's a symbol in the group name, the placeholder/default group icon will (often) use the first character after a symbol. For example, with the group name "dotmacro's test", the placeholder group icon is "DS", but should be "DT".  

groupIcon

"dotmacro test group" and "dotmacro t'st group" are both properly given "DT". "\*\*DOTMACRO\*\* \*\*TEST\*\*" is also properly given "DT". 

However, "\*\*DOTMACRO'S\*\* \*\*TEST\*\*" is given this nifty one:

groupIconOverlap

### Solution

Fix so placeholder/default group icon uses first characters of words.

SebinSong:
closes #2076

### DIFF
--- a/frontend/views/components/group-creation-steps/GroupName.vue
+++ b/frontend/views/components/group-creation-steps/GroupName.vue
@@ -110,8 +110,8 @@ export default ({
   },
   computed: {
     groupInitials () {
-      const intialsCombined = str => str.split(/\s+/) // 1. Split the groupname by white-space
-        .map(segment => segment.match(/\b\w/)?.[0] || '') // 2. Extract intial from each segment
+      const intialsCombined = str => str.split(/\s+/) // 1. Split the groupname by white-space.
+        .map(segment => segment.match(/\b\w/)?.[0] || '') // 2. Extract intial from each segment.
         .join('').toUpperCase() // 3. Combine them & uppercase it.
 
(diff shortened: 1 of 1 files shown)


frontend/views/components/group-creation-steps/GroupName.vue
```diff
@@ -110,8 +110,12 @@ export default ({
   },
   computed: {
     groupInitials () {
+      const intialsCombined = str => str.split(/\s+/) // 1. Split the groupname by white-space
+        .map(segment => segment.match(/\b\w/)?.[0] || '') // 2. Extract intial from each segment
+        .join('').toUpperCase() // 3. Combine them & uppercase it.
+
       return this.group.groupName
-        ? this.group.groupName.match(/\b(\w)/g)?.join('').slice(0, 2).toUpperCase() ?? ''
```
SebinSong, line 114: To fix the issue,

`match(/\b(\w)/g)` here needed to be updated to -> 

1. split the name by white-space
2. extract **one** initial letter from **each** segment.

taoeffect (approved):
Nicely done @SebinSong! 👏

PR #2324, opened 2024-09-01
Fix #2309: Add badge on mobile if there is an unread message

### LINKED ISSUE #2309: New designs for chat on mobile version of the app
### Problem

Right now when a user sends a DM to someone on the mobile version of the app, it's not obvious that this happened.

### Solution

Part of the problem is that there's no push notification (#1456), but also there could be a visual indicator somehow that a user received a DM.

Perhaps we could add a red dot to the "i" icon here:

Screenshot 2024-08-20 at 9 30 53 AM

pieer:


taoeffect (changes requested):
Cool, sorta works, however the indicator doesn't always make sense.

For example, it appears on the dashboard when the user is mentioned in chat:

Screenshot 2024-09-06 at 2 54 45 PM

But pressing it doesn't help at all in this situation:

Screenshot 2024-09-06 at 2 55 52 PM

We need a better design here I think.

corrideat:
> We need a better design here I think.

Another issue is that the badge appears to be floating around, so it's not immediately obvious that it's part of the button.

pieer:
I've updated the logic:
If we are anywhere in the app but the chat group, we show a notification in the left menu to indicate a new message.
If we are in the group chat page, we have on mobile, an indication on the right panel to indicate new message.

I've updated the icon design to have a background attached to the notification, so the badge is not floating anymore, but attached to the button.

If there is already notification for something else, the main menu show that notification. When the user open the menu he can either look at the notification but also can see that there is a new message in the chat

taoeffect (approved):
Seems to be working, nice work @pieer!

PR #2333, opened 2024-09-06
Fixes #2311.

### LINKED ISSUE #2311: Cannot mark notifications as read
### Problem

Screenshot 2024-08-20 at 9 42 35 AM

In ~~1.0.7~~ 1.0.6, both on desktop and on mobile, no matter whether I click "Mark all as read" or click on the notification itself, the notification does not get marked as read.

In fact this happens on `testing` too, where I have 16 unread notifications:

Screenshot 2024-08-20 at 9 47 29 AM

### Solution

Fix!

corrideat:
This implements the following changes:

  * Increase KV notification status memory to 2 months
  * Set Vuex notification memory to 6 months
  * Automatically mark notifications as read after the notification status memory window (e.g. after 2 months)
  * Purge notifications from Vuex store that are older than 6 months

### DIFF
--- a/frontend/model/notifications/utils.js
+++ b/frontend/model/notifications/utils.js
@@ -20,6 +20,9 @@ export function age (notification: Notification): number {
 /*
  * Creates a copy of the given notification list, possibly sorted and/or truncated to
  * make it suitable for offline storage.
+ * The second argument, `status` contains separarately-stored parameters for that
+ * notification hash to reflect how notifications are used in the app
+ * (specifically, the `read` parameter is stored there).
  *
  * Algorithm steps taken on the copy:
  *
@@ -40,9 +43,9 @@ export function age (notification: Notification): number {
  *   5f. discard unread notifications, older ones first.
  * 6. Return the remaining notifications.
  */
-export function applyStorageRules (notifications: Notification[]): Notification[] {
+export function applyStorageRules (notifications: Notification[], status: { [string]: Notification } = {}): Notification[] {
   // Apply the MAX_AGE constraint by selecting items that have not yet expired.
-  let items = notifications.filter(item => !isExpired(item))
+  let items = notifications.filter(item => !isExpired({ ...item, ...status[item.hash] }))
 
   if (items.length > MAX_COUNT) {
     // Sort the list by descending priority order.
--- a/frontend/model/notifications/vuexModule.js
+++ b/frontend/model/notifications/vuexModule.js
@@ -1,11 +1,12 @@
 'use strict'
 
 import { Vue } from '@common/common.js'
-import type { Notification } from './types.flow.js'
-import './selectors.js'
-import { compareOnTimestamp, isNew, isOlder } from './utils.js'
-import * as keys from './mutationKeys.js'
 import { cloneDeep } from '~/frontend/model/contracts/shared/giLodash.js'
+import * as keys from './mutationKeys.js'
+import './selectors.js'
+import { MAX_AGE_READ, MAX_AGE_UNREAD } from './storageConstants.js'
+import type { Notification } from './types.flow.js'
+import { age, compareOnTimestamp, isNew, isOlder } from './utils.js'
 
 const defaultState = {
   items: [], status: {}
@@ -13,7 +14,18 @@ const defaultState = {
 
 const getters = {
   notifications (state, getters, rootState) {
-    return state.items.map(item => ({ ...item, ...state.status[item.hash] }))
+    return state.items.map(item => {
+      const notification = { ...item, ...state.status[item.hash] }
+      // Notifications older than MAX_AGE_UNREAD are discarded
+      if (age(notification) > MAX_AGE_UNREAD) {
+        return null
+      } else if (!notification.read && age(notification) > MAX_AGE_READ) {
+        // Unread notifications older than MAX_AGE_READ are automatically
+        // marked as read
+        notification.read = true
+      }
+      return notification
+    }).filter(Boolean)
   },
   // Notifications relevant to the current group only.
   currentGroupNotifications (state, getters, rootState) {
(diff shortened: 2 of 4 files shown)


frontend/model/notifications/utils.js
```diff
@@ -107,6 +110,7 @@ export function compareOnPriority (notificationA: Notification, notificationB: N
 }
 
 export function isExpired (notification: Notification): boolean {
+  console.error('@@@isExpired', notification, age(notification) > maxAge(notification), age(notification), maxAge(notification), 'read', notification.read)
```
taoeffect: Stray debug logging.

taoeffect (changes requested):

frontend/model/notifications/utils.js
corrideat: Thanks!

frontend/model/notifications/vuexModule.js
```diff
@@ -1,19 +1,31 @@
...
 }
 
 const getters = {
   notifications (state, getters, rootState) {
-    return state.items.map(item => ({ ...item, ...state.status[item.hash] }))
+    return state.items.map(item => {
+      const notification = { ...item, ...state.status[item.hash] }
+      // Notifications older than MAX_AGE_UNREAD are discarded
+      if (age(notification) > MAX_AGE_UNREAD) {
+        return null
```
taoeffect, line 21: OK, they are discarded in the getter, but where are they discarded from the state?

frontend/model/notifications/vuexModule.js
```diff
@@ -1,19 +1,31 @@
...
+      } else if (!notification.read && age(notification) > MAX_AGE_READ) {
+        // Unread notifications older than MAX_AGE_READ are automatically
+        // marked as read
+        notification.read = true
```
taoeffect, line 25: Same question here - should these be marked as read in the state as well, not just the getter, or perhaps instead of in the getter?

taoeffect (commented):
Questions

frontend/model/notifications/vuexModule.js
corrideat: They're discarded from the state when the state is saved (see `state/vuex/save`)

frontend/model/notifications/vuexModule.js
corrideat: I don't think so, because the state (here meaning `state.status`) is handled separately by incoming KV events, and those notifications will not be saved in the state. Modifying this state makes no sense as it'll be overwritten anyhow by said events. Instead, by modifying the getter, those notifications will be removed.

frontend/model/notifications/vuexModule.js
taoeffect: I genuinely do not see where that happens. Can you point it out?

frontend/model/notifications/vuexModule.js
corrideat: BTW, I think it makes sense to try out these changes apart from discussing them in the abstract. If you haven't done so, I suggest modifying (locally) `storageConstants.js` to set the READ and UNREAD values to a low value (e.g., I used 1 min and 5 min, resp.) and see what the state looks like using, e.g., the Vue extension or calling `sbp('state/vuex/state')`. ETA: If you do this, because changes happen 'externally' (i.e., current time is not part of the state and Vuex doesn't know about it), you may need to trigger something that changes the state (e.g., sending a message, changing a setting, etc.)

frontend/model/notifications/vuexModule.js
corrideat: The line is literally part of this commit (although the change in question is related but actually a bugfix for a different issue).

https://github.com/okTurtles/group-income/pull/2333/files#diff-170fa1f8952aa138da06e2d0a0e0039fffda8cdf4168989b4886eceaf66d502cR126

As you can see, each time the state is saved, the notifications are filtered.

taoeffect: Then my logic comprehension might be broken. I saw that line and that line doesn't appear to do be related to these changes here as it is accessing the notifications through the `state` (which is unfiltered) unlike this getter. What am I missing?

corrideat: > What am I missing?

You're missing it here:

>  (which is unfiltered)

The call to `applyStorageRules` does the filtering. Hence, the only the `getter` needs to be updated.

taoeffect: OK, thanks, I think I see what you mean, it's this line:

```js
let items = notifications.filter(item => !isExpired({ ...item, ...status[item.hash] }))
```

And then the `status[item.hash]` supposedly has the correct value to get `maxAge` to return `MAX_AGE_UNREAD`, is my guess. Although per my [other question](https://github.com/okTurtles/group-income/pull/2333#discussion_r1747839448) I'm not super clear on how that would have the correct value, but in my tests this seemed to work so I'll go ahead and approve once the stray debug logging is removed. If there are problems we can return to this later.

corrideat: Well, I'm not sure what's confusing to you, but I can comment on what was somewhat confusing to me.

There are two different things going on: saving or mutating the UI and display concerns. In addition, there notification state is split up into the `items` and `status` properties. `items` is generated when a notification happens, and `status` mostly (only?) holds the read / unread status.

The 'saving' part is mostly addressed in this PR by adjusting the constants. That was mostly already working (the change I pointed out in `state.js` didn't entirely work because it looks at the `read` property, but it wasn't gonna find it because it's usually undefined in `items`; the change makes it also look into `status`). In short, saving already filters things out based on age.

This PR addresses the UI part. This has two goals:

1. Since the `read` state for notifications that are too old (i.e., between the two windows) will _not_ be saved, it sets `read` to true for display purposes.
2. It also filters out notifications that are too old. This is for consistency, but it's not that important since saving the state takes care of removing these old notifications.

taoeffect (approved):
Great work @corrideat !

PR #2335, opened 2024-09-08
#2296 - Fix the broken search functionality in the `/payments` page

### LINKED ISSUE #2296: Implement search functionality for payments
### Problem

On `testing` and `live`, the "Completed" tab on the "Payments" view shows a search box and a loupe icon. The user can type text into the search box, but there is no way to submit the query -- neither clicking the icon nor hitting the enter key does anything. Search is not implemented. 

search

### Solution

Implement this search function. (In the meantime, hide the non-functional search box and loope from the UI.)

SebinSong:
closes #2296


frontend/views/pages/Payments.vue
```diff
@@ -458,7 +458,7 @@ export default ({
 
       const matchesMethodFilter = methodFilterVal === 'all' || acceptedMethods.includes(methodFilterVal)
       const matchesSearchQuery = searchQuery === '' ||
-        `${amount}${username.toUpperCase()}${displayName.toUpperCase()}`.indexOf(searchQuery.toUpperCase()) !== -1
+        `${amount}${displayName.toUpperCase()}`.indexOf(searchQuery.toUpperCase()) !== -1
```
SebinSong, line 461: it appears that `username` was replaced with `displayName` following it at some point but `username` hasn't been removed, causing this run-time error.

image

taoeffect (approved):
Nice work @SebinSong! I think we should also search notes in addition to the name and amount, but that wasn't the issue's task so I'll approve and merge!

PR #2343, opened 2024-09-10
#2002 allow DM to yourself

### LINKED ISSUE #2002: Allow creation of DM with yourself
### Problem

It is very useful to be able to create a DM with yourself to share and remember notes with yourself.

Right now we can't do this because the DM creation modal doesn't show your username in it.

### Solution

- Allow creating a DM to yourself.
- If a DM is created between you and some other people, treat it the same way as we do right now with creating a DM with those other people (just ignore the fact that you're also listed in the DM - you are obviously are going to be part of the conversation).

SebinSong:
closes #2002


frontend/views/components/UsersSelector.vue
```diff
@@ -117,9 +115,6 @@ export default ({
   watch: {
     value () {
       this.$emit('change', this.value)
-    },
-    usernames () {
-      this.clear()
```
SebinSong, line 122: Apparently `usernames` here was meant to be removed in [this previous commit](https://github.com/okTurtles/group-income/commit/f8cac6630b098bf693585db3fd23a60b5a9977a7) but went unnoticed, causing a run-time error. So removed it here.

SebinSong:
@taoeffect 
Let me know if you think we need to create a test-step for creating a DM to myself  in `group-chat-direct-message.spec.js` too. I can either work on it as part of this PR or as a follow-up issue.

frontend/controller/actions/identity.js
```diff
@@ -478,7 +478,9 @@ export default (sbp('sbp/selectors/register', {
     const rootState = sbp('state/vuex/state')
     // TODO: Can't use rootGetters
     const rootGetters = sbp('state/vuex/getters')
-    const partnerIDs = params.data.memberIDs.map(memberID => rootGetters.ourContactProfilesById[memberID].contractID)
+    const partnerIDs = params.data.memberIDs
+      .filter(memberID => memberID !== rootGetters.ourIdentityContractId)
+      .map(memberID => rootGetters.ourContactProfilesById[memberID].contractID)
```
taoeffect, line 483: Why is this change being made? I don't quite understand the reason for filtering out our own contract even if it's intentionally specified.

taoeffect (commented):
Nice work @SebinSong, preliminary / partial review and one question:

taoeffect (changes requested):
I don't know if this is an issue with this PR or something that went unnoticed.

1. I created a group and invited u2 and u3.
2. u2 created a DM with u1
3. u1 then added u3  to that DM:

Screenshot 2024-09-10 at 6 17 10 PM

And now for u1 there's a DM with u2 and u3, but the user icon is grey:

Screenshot 2024-09-10 at 6 16 57 PM

frontend/controller/actions/identity.js
SebinSong: `partnerIDs` here is used to call sbp(`gi.actions/chatroom/join`) further down the logic [2], which is called by default for the DM creator [1].

```js
    await sbp('gi.actions/chatroom/join', { ... }) [1]

    for (const partnerID of partnerIDs) {
      await sbp('gi.actions/chatroom/join', { ... }) [2]
    }
```

Now that we expose the current user in this 'create new DM' modal list, there is cases where the members contain the current user too. This case needed to be taken account here.

me-in-other

SebinSong:
@taoeffect Fixed the typo that caused that greyed-out photo issue. it does show now.

taoeffect (approved):
Excellent job @SebinSong! 👏

PR #2344, opened 2024-09-12
Limit logs to 500 lines.

### LINKED ISSUE #2339: Limit Cypress test failure logs to 500 lines
### Problem

Sometimes it seems like the Cypress Dashboard video fails to upload because the amount of data we're sending to Cypress is too much. Most likely this has less to do with the video and more to do with the logs (or at least the logs are something we can control).

### Solution

Modify `output-logs.js`:

```js
Cypress.on('fail', (error) => {
  error.stack += '\nConsole Logs:\n========================\n'
  error.stack += '[' + logs + ' null]'
  // clear logs after fail so we dont see duplicate logs
  logs = ''
  // still need to throw the error so tests wont be marked as a pass
  throw error
})
```

Modify that line `error.stack += '[' + logs + ' null]'` to include only up to the last 500 lines from that array.

corrideat:
Fix #2339


taoeffect (approved):
Nice!

PR #2346, opened 2024-09-13
#2345 - Broken username in the in-app notification after the user has been kicked out

### LINKED ISSUE #2345: Username of a removed member looks wrong in the in-app notifications.
### Problem
the username of a member who has been removed looks like below in an in-app notification content.

**[ Steps to reproduce ]**
1. Sign up for user 1, 2 and 3.
2. User 1 creates a group and invite user 2 and 3.
3. Kick out the user 3.
4. Check any existing in-app notification that mentions user 3, now the identity contract ID is not properly converted to the display name. 

### Solution
Fix it

SebinSong:
closes #2345

### DIFF
--- a/frontend/model/contracts/shared/functions.js
+++ b/frontend/model/contracts/shared/functions.js
@@ -225,13 +225,13 @@ export function swapMentionIDForDisplayname (
   }
 ): string {
   const {
-    ourContactProfilesById,
     getChatroomNameById,
     usernameFromID,
     userDisplayNameFromID
   } = sbp('state/vuex/getters')
+  const { reverseNamespaceLookups } = sbp('state/vuex/state')
   const possibleMentions = [
-    ...Object.keys(ourContactProfilesById).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
+    ...Object.keys(reverseNamespaceLookups).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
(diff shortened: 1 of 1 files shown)


frontend/model/contracts/shared/functions.js
```diff
@@ -225,13 +225,13 @@ export function swapMentionIDForDisplayname (
...
   const {
-    ourContactProfilesById,
     getChatroomNameById,
     usernameFromID,
     userDisplayNameFromID
   } = sbp('state/vuex/getters')
+  const { reverseNamespaceLookups } = sbp('state/vuex/state')
   const possibleMentions = [
-    ...Object.keys(ourContactProfilesById).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
+    ...Object.keys(reverseNamespaceLookups).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
```
SebinSong, line 234: I actually don't know the exact context that this `reverseNamespaceLookups` field was created for but apparently this field contains user-ids for both current members and members who have been kicked out. (as opposed to `ourContractProfilesById` only having current members of the group) So making an update here.

taoeffect: @SebinSong this was introduced in this PR: https://github.com/okTurtles/group-income/pull/2294

Nice job realizing it would be useful for this issue!

taoeffect (approved):
Approved!

PR #2350, opened 2024-09-15
Fix #2329. Remove unused GIMessage attributes.

### LINKED ISSUE #2329: Remove `originatingContractHeight` from GIMessage head
### Problem

The value `originatingContractHeight` is in the head but it's not used from there, but instead is included as part of the encrypted payload, see discussion:

https://github.com/okTurtles/shelterprotocol.net/issues/5

### Solution

Remove `originatingContractHeight` from the head.

corrideat:
Fix #2329


taoeffect (approved):
Nice!

PR #2353, opened 2024-09-16
Use proper validator functions

### LINKED ISSUE #2352: Broken validators
### Problem

In `'gi.contracts/group/updateSettings'`, I mistakenly thought that these custom validation functions in `objectMaybeOf` worked:

```js
    'gi.contracts/group/updateSettings': {
      // OPTIMIZE: Make this custom validation function
      // reusable accross other future validators
      validate: actionRequireActiveMember((data, { getters, meta, message: { innerSigningContractID } }) => {
        objectMaybeOf({
          groupName: stringMax(GROUP_NAME_MAX_CHAR, 'groupName'),
          groupPicture: x => typeof x === 'string',
          sharedValues: stringMax(GROUP_DESCRIPTION_MAX_CHAR, 'sharedValues'),
          mincomeAmount: x => typeof x === 'number' && x > 0,
          mincomeCurrency: stringMax(GROUP_CURRENCY_MAX_CHAR, 'mincomeCurrency'),
          distributionDate: x => typeof x === 'string',
          allowPublicChannels: x => typeof x === 'boolean'
        })(data)
```

But they do not.

The problem is that the custom validators do not throw an exception.

Similar problem in `'gi.contracts/group/groupProfileUpdate'` too.

EDIT: similarly, there is another big problem that validators from outside of `flowTyper.js`, like `isIsoString` and `isPeriodStamp` are used with `objectMaybeOf`.

### Solution

- Replace `x => typeof x === 'string'` with the `string` validation function, and do the same for `x => typeof x === 'boolean'` (except use `boolean` function validator).
- Then for `mincomeAmount`, use the new `numberRange` validator
- In `'gi.contracts/group/groupProfileUpdate'`: `incomeDetailsType` and `incomeAmount` have a similar problem
- Create a `validatorFrom` function inside of `flowTyper.js` that takes a function that returns a boolean and throws a `validatorError` if that function returns false, and then wrap functions like `isPeriodStamp` with it (when they're called as part of a validation with `objectMaybeOf`)

corrideat:
Fix #2352

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -5,7 +5,7 @@
 import { Errors, L } from '@common/common.js'
 import sbp from '@sbp/sbp'
 import { DELETED_CHATROOM, JOINED_GROUP, LEFT_CHATROOM } from '@utils/events.js'
-import { actionRequireInnerSignature, arrayOf, boolean, number, numberRange, object, objectMaybeOf, objectOf, optional, string, stringMax, tupleOf, unionOf } from '~/frontend/model/contracts/misc/flowTyper.js'
+import { actionRequireInnerSignature, arrayOf, boolean, number, numberRange, object, objectMaybeOf, objectOf, optional, string, stringMax, tupleOf, validatorFrom, unionOf } from '~/frontend/model/contracts/misc/flowTyper.js'
 import { REMOVE_NOTIFICATION } from '~/frontend/model/notifications/mutationKeys.js'
 import { ChelErrorGenerator } from '~/shared/domains/chelonia/errors.js'
 import { findForeignKeysByContractID, findKeyIdByName } from '~/shared/domains/chelonia/utils.js'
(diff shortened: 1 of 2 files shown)


frontend/model/contracts/group.js
```diff
@@ -1038,7 +1038,7 @@ sbp('chelonia/defineContract', {
     'gi.contracts/group/groupProfileUpdate': {
       validate: actionRequireActiveMember(objectMaybeOf({
         incomeDetailsType: x => ['incomeAmount', 'pledgeAmount'].includes(x),
-        incomeAmount: x => typeof x === 'number' && x >= 0,
+        incomeAmount: numberRange(0, Number.MAX_VALUE),
```
taoeffect, line 1044: `incomeDetailsType` needs a `validatorFrom` wrapper

taoeffect (changes requested):
Great work @corrideat ! Minor issue found:

taoeffect (approved):
LGTM!

PR #2355, opened 2024-09-17
#2337 - Fix the overly tall chatroom height when a message is in edit mode

### LINKED ISSUE #2337: Editing messages on mobile looks wrong
### Problem

The edit field doesn't take up the space that it should. It looks really messed up:

a

### Solution

Fix!

SebinSong:
closes #2337 

edit-scroll-issue

From my investigation there is an invisible element `.c-send-mask` that is used to adjust/calculate the height of `<textarea />` element (that has `.c-send-textarea` css class) while the message is in edit mode. The cause of the issue was that this invisible `.c-send-mask` becomes way too tall if the message content is large.

So, made a fix for this.

### DIFF
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -37,7 +37,7 @@
             @click='$emit("reply-message-clicked")'
           )
 
-        send-area(
+        send-area.c-edit-send-area(
           v-if='isEditing'
           :defaultText='swapMentionIDForDisplayname(text)'
           :isEditing='true'
@@ -378,5 +378,9 @@ export default ({
 
 .c-disappeared {
   animation: disappeared 0.5s linear;
+}
+
+.c-edit-send-area {
+  padding: 0 0 1rem;
 }
 </style>
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -973,6 +973,11 @@ export default ({
     height: auto;
     white-space: pre-line;
     min-height: 0;
+    max-height: 9rem;
+
+    @include tablet {
+      max-height: 12.25rem;
+    }
   }
 
   &-btn {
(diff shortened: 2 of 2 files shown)


taoeffect (changes requested):
Nice! I found a few issues and have a request for another minor improvement from my testing.

First, I wasn't able to test this on Firefox Developer Edition in the responsive mode because it wouldn't let me edit the message. There was no way to "click" the message to bring up the edit menu. So that's Bug 1.

In Brave, I was able to test, however there were two issues:

1. When clicking "Edit" the viewport jumped in an unnatural way that required re-scrolling in order to get to a comfortable place where to edit the message. See first screenshot below:

a

2. The edit viewport itself was still kinda unnecessarily small. See 2nd screenshot below:

b

SebinSong:
1.
> First, I wasn't able to test this on Firefox Developer Edition in the responsive mode because it wouldn't let me edit the message. There was no way to "click" the message to bring up the edit menu. So that's Bug 1.

That's a bug in FF devtool rather than a bug of app itself. The touch behaviour in the emulators of other browsers work well (Brave desktop, Chrome). When that buggy touch behaviour happens in FF, I sometimes just turn this button off to check something. (Also, I feel Chrome has been the best tool when it comes to the mobile emulator)



2.
> When clicking "Edit" the viewport jumped in an unnatural way that required re-scrolling in order to get to a comfortable place where to edit the message. See first screenshot below:

Investigated this too but, I think this is something that I could only observe in Brave mobile-browser. In Chrome Android and Samsung Internet browser on my phone, they all gets scrolled to a correct position.

[Samsung Internet]


[Chrome Android]



3.
> The edit viewport itself was still kinda unnecessarily small.

Increased the max-height value from `9rem` to `13.75rem`. let me know if it needs to be taller.

SebinSong:
@taoeffect as discussed via Slack, removed the side-padding in the edit mode, so the text-area element becomes wider.

taoeffect (approved):
Approved! Nice work @SebinSong !

PR #2359, opened 2024-09-20
No document deps

corrideat:
This PR is part of the sequence for porting Chelonia and contracts to run in a service worker. This PR removes all references to `document` in contracts and the service worker.


taoeffect (changes requested):
Thanks @corrideat, I'll have a look possibly tomorrow or over the weekend.

Could you please revert some of the changes made to the orders of imports so that this PR is a minimal set of changes? I would typically let it slide, but in general these types of changes should be avoided, and also now `'@sbp/sbp'` is no longer the first import, and it's a project-wide code pattern to make it first. It would also greatly speed up my ability to review the PR for the import order to stay as much the same as possible.

taoeffect (changes requested):
OK, looks good as far as logic goes.

Some cleanup requested:

- See previous review - please minimize movement of imports
- ~~Please keep the files `translations.js` and `i18n.js` together in the same folder as they are part of the same "library". Perhaps under `common/localization`~~ - nm

taoeffect (approved):
Awesome!!

PR #2360, opened 2024-09-22
Various minor fixes

corrideat:
These are some minor changes that I've made while working on the SW.

### DIFF
--- a/frontend/controller/app/chatroom.js
+++ b/frontend/controller/app/chatroom.js
@@ -17,6 +17,11 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
     let attemptCount = 0
     // Sometimes, the state may not be ready (it needs to be copied from the SW
     // to Vuex). In this case, we try again after a short delay.
+    // The specific issue is that the browsing-side state is updated in response
+    // to the EVENT_HANDLED event. Although that event is correctly emitted
+    // prior to JOINED_CHATROOM, processing might take slightly longer, causing
+    // rootState[chatRoomID]?.members?.[identityContractID] to be briefly
+    // undefined.
     // TODO: Figure out a better way of doing this that doesn't require a timeout
     const setCurrentChatRoomId = () => {
       if (!rootState[chatRoomID]?.members?.[identityContractID]) {
--- a/frontend/controller/app/chatroom.js
+++ b/frontend/controller/app/chatroom.js
@@ -14,7 +14,22 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
   const rootState = sbp('state/vuex/state')
   if (rootState.loggedIn?.identityContractID !== identityContractID) return
   if (!rootState.chatroom.currentChatRoomIDs[groupContractID] || rootState.chatroom.pendingChatRoomIDs[groupContractID] === chatRoomID) {
-    sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupID: groupContractID, chatRoomID })
+    let attemptCount = 0
+    // Sometimes, the state may not be ready (it needs to be copied from the SW
+    // to Vuex). In this case, we try again after a short delay.
+    // TODO: Figure out a better way of doing this that doesn't require a timeout
+    const setCurrentChatRoomId = () => {
+      if (!rootState[chatRoomID]?.members?.[identityContractID]) {
+        if (++attemptCount > 5) {
+          console.warn('[JOINED_CHATROOM] Given up on setCurrentChatRoomId after 5 attempts', { identityContractID, groupContractID, chatRoomID })
+          return
+        }
+        setTimeout(setCurrentChatRoomId, 5 + 5 * attemptCount)
+      } else {
+        sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupID: groupContractID, chatRoomID })
+      }
+    }
+    setCurrentChatRoomId()
   }
 })
 sbp('okTurtles.events/on', LEFT_CHATROOM, switchCurrentChatRoomHandler)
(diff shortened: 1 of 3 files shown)


test/cypress/support/index.js
```diff
@@ -61,8 +58,8 @@ afterEach(function () {
 
 // Prevent errors when English is not the current OS locale language.
 Cypress.on('window:before:load', window => {
-  Object.defineProperty(window.navigator, 'language', { value: 'en-US-POSIX' })
-  Object.defineProperty(window.navigator, 'languages', { value: ['en-US-POSIX', 'en-US', 'en'] })
+  Object.defineProperty(window.navigator, 'language', { value: 'en-US-POSIX', configurable: true })
+  Object.defineProperty(window.navigator, 'languages', { value: ['en-US-POSIX', 'en-US', 'en'], configurable: true })
```
taoeffect, line 68: Can you add a comment explaining why these are marked as `configurable: true`?

frontend/controller/service-worker.js
```diff
@@ -29,31 +29,39 @@ sbp('sbp/selectors/register', {
     // TODO: move ahead with encryption stuff ignoring this service worker stuff for now
     // TODO: improve updating the sw: https://stackoverflow.com/a/49748437
     // NOTE: user should still be able to use app even if all the SW stuff fails
-    if (!('serviceWorker' in navigator)) { return }
+    if (!('serviceWorker' in navigator)) {
+      throw new Error('Service worker APIs missing')
```
corrideat, line 33: This throws now because that way the caller can know if the call succeeded or not.

frontend/controller/service-worker.js
```diff
@@ -29,31 +29,39 @@ sbp('sbp/selectors/register', {
...
+    }
 
     try {
       const swRegistration = await navigator.serviceWorker.register('/assets/js/sw-primary.js', { scope: '/' })
 
-      if (swRegistration) {
```
corrideat, line 37: swRegistration is always truthy AFAIK, so I removed this check as it's redundant

frontend/controller/service-worker.js
```diff
@@ -29,31 +29,39 @@ sbp('sbp/selectors/register', {
...
-        // a ping message every 5 seconds to ensure that the worker remains
-        // active.
-        // The downside of this is that there are messges going back and forth
-        // between the service worker and each tab, the number of which is
-        // proportional to the number of tabs open.
-        // The upside of this is that the service worker remains active while
-        // there are open tabs, which makes it faster and smoother to interact
-        // with contracts than if the service worker had to be restarted.
-        setInterval(() => navigator.serviceWorker.controller?.postMessage({ type: 'ping' }), 5000)
+      } else {
```
corrideat, line 49: This makes sure the `store-client-id` message is sent if the SW is not yet ready

frontend/controller/app/chatroom.js
```diff
@@ -14,7 +14,22 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
   const rootState = sbp('state/vuex/state')
   if (rootState.loggedIn?.identityContractID !== identityContractID) return
   if (!rootState.chatroom.currentChatRoomIDs[groupContractID] || rootState.chatroom.pendingChatRoomIDs[groupContractID] === chatRoomID) {
-    sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupID: groupContractID, chatRoomID })
+    let attemptCount = 0
+    // Sometimes, the state may not be ready (it needs to be copied from the SW
+    // to Vuex). In this case, we try again after a short delay.
+    // TODO: Figure out a better way of doing this that doesn't require a timeout
```
taoeffect, line 25: Can you expand this comment so that someone who is revisiting this code in 1 month understands what the issues are in order to implement this TODO? i.e. answering the question of why the `JOINED_CHATROOM` event can't simply be called after the state is copied over, and the relationship between `EVENT_HANDLED` and this `JOINED_CHATROOM` event?

frontend/controller/service-worker.js
```diff
@@ -29,31 +29,39 @@ sbp('sbp/selectors/register', {
...
+        const handler = () => {
+          navigator.serviceWorker.removeEventListener('controllerchange', handler, false)
+          navigator.serviceWorker.controller.postMessage({ type: 'store-client-id' })
+        }
+        navigator.serviceWorker.addEventListener('controllerchange', handler, false)
```
taoeffect, line 54: Once this event fires, it seems like the handler can only be called a maximum of 1 time. Does that mean the service worker can be changed only once..? Or can it be changed multiple times and therefore we shouldn't call `removeEventListener` in the handler?

(Whatever your answer, can you add a comment inside of the `handler` explaining this?

corrideat: The service worker can change any number of times (but presumably then the ID will already be stored). I was just making changes here to ensure that the message is sent. However, I need to review how this ID ends up being used when working on notifications.

taoeffect (changes requested):
Two requests for comment

taoeffect (approved):
LGTM

frontend/controller/app/chatroom.js
```diff
@@ -14,7 +14,27 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
...
+        if (++attemptCount > 5) {
+          console.warn('[JOINED_CHATROOM] Given up on setCurrentChatRoomId after 5 attempts', { identityContractID, groupContractID, chatRoomID })
+          return
+        }
+        setTimeout(setCurrentChatRoomId, 5 + 5 * attemptCount)
+      } else {
+        sbp('state/vuex/commit', 'setCurrentChatRoomId', { groupID: groupContractID, chatRoomID })
+      }
+    }
+    setCurrentChatRoomId()
```
taoeffect, line 37: Think I missed something here: the `rootState` inside of `setCurrentChatRoomId` may be a stale reference. You need to refetch it don't you?

corrideat: Yes, technically you're right and I can fix that. However, I don't think it's a real issue. The rootState would only be replaced if the user logs out, at which point the stale reference doesn't matter.

PR #2361, opened 2024-09-22
#2356 - Fix the chat markdown bug related to a line-break

### LINKED ISSUE #2356: Minor markdown formatting bug in chat
### Problem

This markdown:

```md
Hey @sebin here's the `data/` folder - the contents of which you can copy into your own local `data/` folder.

Then run with: `GI_PERSIST=sqlite grunt dev`

And login with:

- User: `u1`
- Password:


Note that this user has joined two groups. The group that's most useful is the one called `Turtles (4)` as that's the one that has a single user in it and the invite link expired.
```

Renders like this:

Screenshot 2024-09-19 at 10 03 13 AM

Notice how the last paragraph is indented to be lined up with the bulleted list, but it shouldn't be.

I had to use two empty newlines because using 1 empty newline is even worse:

```md
- Password:

Note that this user has joined two groups. The group that's most useful is the one called `Turtles (4)` as that's the one that has a single user in it and the invite link expired.
```

It renders like this:

Screenshot 2024-09-19 at 10 07 54 AM

When it should render like this:

- Password:

Note that this user has joined two groups. The group that's most useful is the one called `Turtles (4)` as that's the one that has a single user in it and the invite link expired.

### Solution

Fix these issues.

SebinSong:
closes #2356 

Screenshot for the fix.

### DIFF
--- a/frontend/views/utils/markdown-utils.js
+++ b/frontend/views/utils/markdown-utils.js
@@ -40,14 +40,15 @@ export function renderMarkdown (str: string): any {
       entryText = entryText.replace(/</g, '&lt;').replace(/>/g, '&gt;')
 
       entryText = entryText.replace(/\n(?=\n)/g, '\n<br>')
-        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
-        .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
+        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // [1] custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
+        .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // [2] this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
 
       entry.text = entryText
     }
   })
 
   str = combineMarkdownSegmentListIntoString(strSplitByCodeMarkdown)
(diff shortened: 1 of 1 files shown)


frontend/views/utils/markdown-utils.js
```diff
@@ -40,14 +40,15 @@ export function renderMarkdown (str: string): any {
       entryText = entryText.replace(/</g, '&lt;').replace(/>/g, '&gt;')
 
       entryText = entryText.replace(/\n(?=\n)/g, '\n<br>')
-        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
-        .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
+        .replace(/<br>\n(\s*)(\d+\.|-)/g, '\n\n$1$2') // [1] custom-handling the case where <br> is directly followed by the start of ordered/unordered lists
+        .replace(/(\d+\.|-)(\s.+)\n<br>/g, '$1$2\n\n') // [2] this is a custom-logic added so that the end of ordered/un-ordered lists are correctly detected by markedjs.
```
SebinSong, line 44: As inline-commented in the same file, `markedjs` with `gfm` style has a default behaviour where it collapses multiple consecutive line-breaks into one which is not what we want for the app. `[1]` and `[2]` here are the custom logics added to work around this. But [2] here didn't capture an edge-case where **the last-child of a list ends with `<code>` element**.

(The reason for this edge case is that, we split the whole text by markdown code elements first before applying these `[1]` and `[2]` because no content wrapped within `<code>...</code>` should be converted. eg. we already made a fix before for a bug where something like **a<=b** within `<code>` element looks incorrect.)

So added a fix in `L51` below to take this into account.

taoeffect (approved):
Great work @SebinSong! 👏

Issue #2334, opened 2024-09-08
Emoji picker issues & enhancements


taoeffect:
### Problem

1. External resources: it seems like emojis are loaded from https://unpkg.com/emoji-datasource-apple@5.0.1/img/apple/sheets-256/64.png. No external resources should be used.
2. Lack of fallback: After 'reacting', the reaction may or may not display (well, it'll display but show a placeholder character), based on the availability of glyphs. (EDIT: this seems like a minor issue)
3. It doesn't contain all of the emojis, some are missing (I don't remember which ones exactly, but remember using one on Slack that wasn't available on in the GI one)
4. It doesn't look good on Dark Mode

Related issue: It doesn't scroll down (issue #2601)

### Solution

Either find a way to use the existing library while addressing the above issues or maybe pick a different one. Remember: the fewer dependencies it has the better. Ideally, zero dependencies on other libraries.

We should never be loading anything from any other website than the site hosting GI.

Fixing this issue might also fix #3099

corrideat:
> zero dependencies on other libraries

I agree, but I think in this case a dependency would help at least for categorising things. Then, using a picker library isn't too far off.

When it comes to loading resources, ideally one could use the system font (no dependencies, smaller payload to deliver). The issue is that this could have some inconsistencies issues across clients (e.g., different degrees of completeness, different rendering, etc.). For example, (some) Android and iOS have different rendering for the gender-neutral pictograms. Another emblematic example is the PISTOL (or WATER PISTOL) emoji, aka U+1F52B. See <https://commons.wikimedia.org/wiki/Category:U%2B1F52B> for examples of different renderings, which is bound to cause misunderstandings.

corrideat:
It seems like the remote resource is defined at `frontend/assets/style/components/_emoji-mart.scss`.  Replacing the URL there fixes issue 1.

> It doesn't contain all of the emojis, some are missing (I don't remember which ones exactly, but remember using one on Slack that wasn't available on in the GI one)

As with many dependencies, the package version used isn't current. The emoji image version referenced is 5.0.1, while the current version is 15.0.1. This could explain some of the missing elements.

PR #2362, opened 2024-09-23
Add missing security headers

### LINKED ISSUE #2291: Missing Security Headers
### Problem

The site `groupincome.app` is missing some important security headers that could mitigate vulnerabilities client-side. This can be verified using <https://securityheaders.com>.

  * `Content-Security-Policy`: This is _mainly_ designed to prevent XSS, although it can also be important for privacy (for example, to block loading images from a third party domain). Mostly, it defines origins that are allowed to serve content (plus some additional settings, such as blocking loading the page in a frame)
  * `X-Content-Type-Options`: AFAIK `nosniff` is the only permissible (and useful) value. This header instructs the user agent to honour the content-type provided the server, an instruction user agents follow in most circumstances. This is useful to prevent, among other things, stored XSS.
  * `Referrer-Policy`: This is mostly a privacy-impacting header for users. It tells the user agent under which circumstances to send a `referrer` header and what information it should contain. Setting it can be useful to prevent destination servers of links in GI from learning the exact URL (which can contain, e.g., a group) the user came from.
  * `Permissions-Policy`: This is also mostly a privacy-impacting defence-in-depth header. It includes a list of permissions that the website is allowed to request. This way, even if some malicious script is loaded, it cannot access sensitive APIs, such as geolocation.
  * `Cross-Origin-Opener-Policy`: This is a security-sensitive header that defines whether the site can share a browsing context group with other sites (e.g., via `opener`).
  * `X-Frame-Options`: This one is actually set but included here for completeness. It defines whether the site can be loaded in a frame (and is overridden by CSP).
  * `Cross-Origin-Embedder-Policy`: Blocks resources not explicitly allowed by either CORS or CORP.
  * `Cross-Origin-Resource-Policy`: Defines who can load the resource. IMHO, not as useful as all the other headers here, but useful at least to prevent hotlinking.

#### A note on HTTPS and server trust

Regardless of the values of these headers, they are set by the server, which means that their effectiveness depends on trusting the server _and that the response reached the client unmodified_.

##### Trusting the server

I think that it's acceptable to have some minimal degree of trust on the server. It'd be a useful exercise to also have a mechanism to relatively easily audit that all artifacts delivered (scripts especially) haven't been tampered with.

Besides the audit mechanism discussed, another (complimentary) option is to use the service worker to control the flow of information from the server to the app. The service worker can verify all subsequent server responses, effectively turning the trust model to TOFU.

##### Trusting that the response arrived unmodified

This is a tricker requirement. For the most part, HTTPS provides protection (save for the remote possibility of MitM).

###### Feasibility of MitM

Although Certificate Transparency [1] is not currently required for valid certificates, Chrome requires [2][3] CT as well as Safari [4] require certificate transparency, making it _de facto_ required. AFAIU, this is not an online check, but issuing such proofs of inclusion will require collusion of parties, making it a difficult attack to carry out, since it's practically difficult and detection is likely to be deleterious for all parties involved.

This attack, however unlikely, is also mitigated in a TOFU model using service workers.

[1] <https://www.rfc-editor.org/rfc/rfc6962>
[2] <https://googlechrome.github.io/CertificateTransparency/ct_policy.html> "Since Chrome 94, Chrome clients will also attempt to obtain updated CT Log lists from the Component Updater infrastructure several times per day. On a successful update of the CT Log list, Chrome will update the start of its 70 day CT enforcement window to the freshest log_list_timestamp."
[3] <https://github.com/GoogleChrome/CertificateTransparency/commit/95694b20950c91b6d72ec1d35fff7a60e7bac242> "Chrome requires all publicly-trusted TLS certificates issued after April 30, 2018 to support CT in order to be recognized as valid. This site provides details on what is required. Any questions should be directed to the ct-policy@chromium.org list."
[4] <https://support.apple.com/en-us/103214>

### Solution

  * `Content-Security-Policy`: `default-src 'none'; script-src 'unsafe-eval'; script-src-elem 'self'; script-src-attr 'none'; style-src 'self'; style-src-elem 'self'; style-src-attr 'none'; img-src 'self' blob:; font-src 'self'; connect-src 'self'; media-src 'self'; prefetch-src 'self'; worker-src 'self'; frame-ancestors 'none'; form-action 'self'; upgrade-insecure-requests; manifest-src 'self'` (this is the most likely to break things now; it may need changing when sandboxing or federation are implemented)
  * `X-Content-Type-Options`: `nosniff`
  * `Referrer-Policy`: `strict-origin-when-cross-origin` (send only the origin as `referer`). Alternatively, `no-referrer` can be used to block sending `referer` in all circumstances, which is the most privacy-preserving option.
  * `Permissions-Policy`: `accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(self), display-capture=(), document-domain=(), encrypted-media=(), execution-while-not-rendered=(self), execution-while-out-of-viewport=(self), fullscreen=(self), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(self), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(self), xr-spatial-tracking=(), clipboard-read=(), clipboard-write=(self), gamepad=(), speaker-selection=()` (self-explanatory, list generated using <https://www.permissionspolicy.com/>)
  * `Cross-Origin-Opener-Policy`: `same-origin` (don't share context with potentially untrusted sites)
  * `X-Frame-Options`: `DENY` (block loading the page in a frame)
  * `Cross-Origin-Embedder-Policy`: `require-corp` (blocks embedding cross-origin resources that don't explicitly allow it).
  * `Cross-Origin-Resource-Policy`: `same-origin` (block hotlinking)

snowteamer:
Fix #2291 

Emoji sheet, audio asset loading and file upload in chat manually tested on:
- Firefox Developer Edition v132.b09 
- Chrome 130

Note that the favicon badge gets blocked in Chrome, but maybe it's acceptable, so as to avoid allowing `data:` URLs?
image


backend/server.js
```diff
@@ -24,6 +24,23 @@ import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
...
+  'X-Content-Type-Options': 'nosniff',
+  'Referrer-Policy': 'strict-origin-when-cross-origin',
+  // List generated using https://www.permissionspolicy.com.
+  'Permissions-Policy': 'accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(self), display-capture=(), document-domain=(), encrypted-media=(), execution-while-not-rendered=(self), execution-while-out-of-viewport=(self), fullscreen=(self), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(self), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(self), xr-spatial-tracking=(), clipboard-read=(), clipboard-write=(self), gamepad=(), speaker-selection=()',
+  // Don't share context with potentially untrusted sites.
+  'Cross-Origin-Opener-Policy': 'same-origin',
+  // Block loading the page in a frame.
+  'X-Frame-Options': 'deny',
```
corrideat: This should be `DENY` (uppercase)

backend/server.js
```diff
@@ -24,6 +24,23 @@ import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
 
 const { CONTRACTS_VERSION, GI_VERSION } = process.env
 
+const securityHeaders = {
+  // This is the most likely to break things now; it may need changing when sandboxing or federation is implemented.
+  'Content-Security-Policy': "default-src 'none'; script-src 'unsafe-eval'; script-src-elem 'self'; script-src-attr 'none'; style-src 'self'; style-src-elem 'self'; style-src-attr 'none'; img-src 'self' blob:; font-src 'self'; connect-src 'self'; media-src 'self'; prefetch-src 'self'; worker-src 'self'; frame-ancestors 'none'; form-action 'self'; upgrade-insecure-requests; manifest-src 'self'",
```
corrideat: This will cause issues with the emoji images, since those are loaded externally. See #2334. Either `img-src` will need to be expanded, or the sheet be embedded locally (preferred solution).

corrideat (changes requested):

backend/server.js
snowteamer: Is using `img-src 'self' blob: https://unpkg.com` OK for now? Otherwise, installing and redistributing emoji sheets via `dist/assets` seems a bit tricky regarding e.g. licensing issues.

corrideat: In terms of solving the CSP issue, yes (although you need to check manually, it could interfere with Cross-Origin-Embedder-Policy as well).

The preferred (to me) solution would be embedding the sheet. There could be some licensing concerns, but I'm not sure those concerns are alleviated by hotlinking. This is best left as a question to a lawyer or to the copyright holder. (Sidenote: fonts designs _generally_ aren't under copyright in the _US_, but I'm guessing emoji designs likely are, and even if they weren't, there's other places to consider).

corrideat: An alternative solution is to drop the Apple emojis and use an emoji set with a free license. E.g., https://github.com/mozilla/fxemoji

corrideat:
Running the app locally and opening `http://localhost:3000` on Firefox doesn't seem to work due to CSP.

image

taoeffect (changes requested):
Please test this PR on as many browsers as you have available locally and let us know which browsers you tested so that I can test on the remaining ones that you don't have access to (e.g. Safari).

taoeffect (changes requested):
Nice work @snowteamer, it's coming along, still a few more browsers we need to test on.

BTW, if you'd like to test on [BrowserStack](https://www.browserstack.com/), I'll send you credentials.

The site doesn't load on Safari on macOS:

Screenshot 2024-12-09 at 9 13 07 AM

And the other error your noticed in Chrome related to the favicon not being set also needs to be fixed, as we need the red dot to be visible when DMs are sent.

snowteamer:
@taoeffect
I could reproduce the app not loading in BrowserStack Safari, which unfortunately didn't give any error message.
In the end I could work around the issue by dropping the `upgrade-insecure-requests` directive, which is related to HTTP requests

backend/server.js
```diff
@@ -108,6 +108,27 @@ const creditsWorker = process.env.CHELONIA_ARCHIVE_MODE || !CREDITS_WORKER_TASK_
...
+  // Block embedding cross-origin resources that don't explicitly allow it. Disabled because it blocks emoji sheet rendering.
+  // 'Cross-Origin-Embedder-Policy': 'require-corp',
+  // Don't share context with potentially untrusted sites.
+  'Cross-Origin-Opener-Policy': 'same-origin',
+  // Block hotlinking.
+  'Cross-Origin-Resource-Policy': 'same-origin',
+  // List generated using https://www.permissionspolicy.com.
+  'Permissions-Policy': 'accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(self), display-capture=(), document-domain=(), encrypted-media=(), execution-while-not-rendered=(self), execution-while-out-of-viewport=(self), fullscreen=(self), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(self), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(self), xr-spatial-tracking=(), clipboard-read=(), clipboard-write=(self), gamepad=(), speaker-selection=()',
```
taoeffect, line 123: I think we might not want these here because the Chelonia server will be running arbitrary apps. Do you agree @corrideat ?

backend/server.js
```diff
@@ -108,6 +108,27 @@ const creditsWorker = process.env.CHELONIA_ARCHIVE_MODE || !CREDITS_WORKER_TASK_
 
 const { CONTRACTS_VERSION, GI_VERSION } = process.env
 
+const securityHeaders = {
+  // This is the most likely to break things now; it may need changing when sandboxing or federation is implemented.
+  // - Not setting media-src since Background.vue dynamically adds sound assets.
+  // - The localhost:3000 entries are for BrowserSync support.
+  'Content-Security-Policy': "child-src 'none'; connect-src 'self' blob: http://localhost:3001 ws://localhost:3001; font-src 'self'; form-action 'self'; frame-ancestors 'none'; frame-src 'none'; img-src 'self' https://unpkg.com blob: data:; manifest-src 'self'; object-src 'none'; script-src 'unsafe-eval'; script-src-attr 'none'; script-src-elem 'self' 'unsafe-inline'; style-src 'self'; style-src-attr 'none'; style-src-elem 'self' 'unsafe-inline'; worker-src 'self'",
```
taoeffect, line 115: Will these negatively affect loading the app over a tunnel like ngrok?

taoeffect:
> Note that the favicon badge gets blocked in Chrome, but maybe it's acceptable, so as to avoid allowing data: URLs?

I think we want to be able to update the favicon to indicate unread messages.

PR #2363, opened 2024-09-27
Refactoring removing global getters from contracts

corrideat:


frontend/controller/actions/group.js
```diff
@@ -41,9 +42,20 @@ import { CONTRACT_HAS_RECEIVED_KEYS, EVENT_HANDLED } from '~/shared/domains/chel
...
 import type { GIActionParams } from './types.js'
+import { REMOVE_NOTIFICATION } from '~/frontend/model/notifications/mutationKeys.js'
 import { createInvite, encryptedAction } from './utils.js'
 import { extractProposalData } from '@model/notifications/utils.js'
 
+sbp('okTurtles.events/on', LEFT_GROUP, ({ identityContractID, groupContractID }) => {
+  const rootState = sbp('state/vuex/state')
+  if (rootState.loggedIn?.identityContractID !== identityContractID) return
+  // NOTE: remove all notifications whose scope is in this group
+  // TODO: FIND ANOTHER WAY OF DOING THIS WITHOUT ROOTGETTERS
```
taoeffect, line 53: Does this comment still make sense now?

frontend/model/contracts/chatroom.js
```diff
@@ -154,6 +100,7 @@ sbp('chelonia/defineContract', {
             maxDescriptionLength: CHATROOM_DESCRIPTION_LIMITS_IN_CHARS
           },
           attributes: {
+            adminIDs: [],
```
taoeffect, line 103: If this PR requires recreating groups, then it is not necessary to do anything about this.

However, if the plan is to support existing groups, then this must have corresponding code in `postUpgradeVerification`

frontend/controller/actions/group.js
corrideat: Yes, until I figure it out. Like mentioned, the main issue is that root anything shouldn't be used in contracts (for sandboxing among other things). However, actions (which will run in the SW) ideally shouldn't need Vuex to be defined. For documentation purposes of what's been discussed, the options seem to be:

1. Refactoring so that getters are not used
2. A Vuex-lite module in the SW that defines used getters
3. Importing Vuex and all getters

frontend/model/contracts/chatroom.js
corrideat: Thanks for bringing this up. Although this PR in itself shouldn't require recreating groups, it'll probably be required at some point in the future ~~, as references to global getters will fail. However, another option would be replacing the only accessor to `adminIDs` with an optional chain (`adminID?.includes`)~~

corrideat: Or actually, scratch that. No change is needed for the same intended effect (delete will fail with the wrong error message if a non-admin does it). I'd also avoid having to define an action that sets the admin, lest it be used by someone who isn't an admin, at least until there's a supported way of doing this. So, let's consider this a breaking change.

corrideat: Options then are:

1. Leave this as is. Admin using this action will get a generic JS error.
2. Use optional chain. Admin won't be able to use this action but will get proper error message.
3. Add update action (downside is that anyone can set themselves as admin). Maybe we can go for this, and then remove this action once groups are re-created.

I don't think this PR is breaking in itself (although that I haven't tried) other than for this `adminIDs` thing. However, it's not a forward-compatible change (meaning that a _future_ version of Chelonia won't support _current & past_ contracts, since at that point `sbp('state/vuex/getters)` will not be defined).

taoeffect, line 103: We need to specify when creating the chatroom who the admin is.

taoeffect: OK, how about a write-once action that sets the chatroom admin if the admin isn't already defined.

This way, once the action is used, it cannot be used again to modify the admin.

frontend/model/state.js
```diff
@@ -92,26 +92,61 @@ sbp('sbp/selectors/register', {
...
-          return upgradeRequired || (member.status === PROFILE_STATUS.ACTIVE && member.joinedHeight == null)
+        }).reduce((upgradeRequired: string | boolean, member: Object) => {
+          return upgradeRequired || (member.status === PROFILE_STATUS.ACTIVE && member.joinedHeight == null && groupID)
         }, false)
+      }).forEach((contractID) => {
+        if (!contractID) return
+        sbp('gi.actions/group/upgradeFrom1.0.7', { contractID }).catch(e => {
+          console.error('[state/vuex/postUpgradeVerification] Error during gi.actions/group/upgradeFrom1.0.7', contractID, e)
+        })
       })
```
taoeffect: Why was this code rewritten?

frontend/model/state.js
```diff
@@ -92,26 +92,61 @@ sbp('sbp/selectors/register', {
...
+        Object.entries(state[ourIdentityContractId].groups).map(([groupID, { hasLeft }]: [string, Object]) => {
+          if (hasLeft || !state[groupID]?.chatRooms || !state[groupID].groupOwnerID) return []
+          // $FlowFixMe[incompatible-use]
+          return Object.entries((state[groupID].chatRooms: { [string]: Object })).flatMap(([chatroomID, { members }]) => {
+            return members[ourIdentityContractId]?.status === PROFILE_STATUS.ACTIVE && !Array.isArray(state[chatroomID]?.attributes?.adminIDs) && [chatroomID, state[groupID].groupOwnerID]
+          })
+        }).forEach(([contractID, groupOwnerID]) => {
+          if (!contractID) return
+          sbp('gi.actions/chatroom/upgradeFrom1.0.8', { contractID, data: groupOwnerID }).catch(e => {
+            console.error('[state/vuex/postUpgradeVerification] Error during gi.actions/chatroom/upgradeFrom1.0.8', contractID, e)
```
taoeffect: Shouldn't the chatroom `data` for the `adminIDs` include both the group owner and whoever created the chatroom?

frontend/model/state.js
corrideat: Because it didn't actually work. Two issues were: missing `return Object.values` and missing `contractID` when calling the action.

taoeffect: Great. 🤦‍♂️ 

Well, nice catch!

frontend/model/state.js
corrideat: Maybe, but that's not how it presently works. Either way, that could probably be done without using adminIDs, but adding a check for the creator.

frontend/model/state.js
corrideat: You can test it out yourself (old and new versions) by changing the check (e.g., joinedHeight  to joinedHeightX)

frontend/model/contracts/shared/functions.js
```diff
@@ -84,7 +83,7 @@ export function getProposalDetails (proposal: Object): Object {
     options['member'] = proposalData.memberName
   } else if (proposalType === PROPOSAL_REMOVE_MEMBER) {
     options['memberID'] = proposalData.memberID
-    options['member'] = sbp('state/vuex/getters').userDisplayNameFromID(proposalData.memberID)
+    // options['member'] = sbp('state/vuex/getters').userDisplayNameFromID(proposalData.memberID)
```
taoeffect: Could you either replace this with, or add a comment above this line, explaining where this is now taken care of? (i.e. file & function?)

frontend/model/state.js
corrideat: What made me question if it worked was the missing contractID. I don't know why neither of us caught that!

frontend/model/contracts/shared/functions.js
corrideat: The intent was actually to remove this line. I can add a comment, but the reason to remove this is that contracts can't access external information.

frontend/controller/actions/chatroom.js
```diff
@@ -3,13 +3,52 @@ import sbp from '@sbp/sbp'
...
+  const isMentionedMe = (!!newMessage || data.type === MESSAGE_TYPES.TEXT) && msgData.text &&
+  (msgData.text.includes(mentions.me) || msgData.text.includes(mentions.all))
+
+  if (!newMessage) {
+    const isAlreadyAdded = !!getters
+      .chatRoomUnreadMessages(contractID).find(m => m.messageHash === data.hash)
+
+    if (!isMentionedMe) {
+      sbp('gi.actions/identity/kv/removeChatRoomUnreadMessage', { contractID, messageHash: data.hash })
+    }
```
taoeffect: Before, in `editMessage`, this would get called only when `isAlreadyAdded` were true... now the logic seems to have slightly changed.

Is this intentional? (if so, why?)

frontend/controller/actions/group.js
taoeffect: I like (2).... there can be a shared filed that's imported in both the browser and sw that defines these getters.

frontend/controller/actions/chatroom.js
corrideat: You're right. It shouldn't make much of a difference, but I can amend that. I tried to simplify the logic and maybe went too far.

frontend/model/state.js
taoeffect: Did you test this code?

It still seems wrong, as it's reducing to boolean inside of the map and yet expecting a `contractID` in the `forEach`

corrideat: I did test it, although it's a complex piece of code so I encourage you to test it too so that we ensure it works.

Back to your question, I see why it might seem that way, but notice the `&& groupID` at the end. So, it'll reduce to `false | string`. If you'd prefer, I can make that less terse.

taoeffect: Yeah, not just less terse, but easier to follow 👍

corrideat: Updated it

frontend/controller/actions/group.js
```diff
@@ -606,6 +618,10 @@ export default (sbp('sbp/selectors/register', {
 
     const message = await sbp('gi.actions/chatroom/create', {
       data: {
+        attributes: {
+          ...params.data?.attributes,
+          adminIDs: [contractState.groupOwnerID]
+        },
```
taoeffect: By placing `adminIDs` after `...params` it is now impossible to call `'gi.actions/group/addChatRoom'` and specify `adminIDs`. Perhaps the order should be reversed?

frontend/controller/actions/chatroom.js
```diff
@@ -3,13 +3,52 @@ import sbp from '@sbp/sbp'
...
 import type { GIRegParams } from './types.js'
 import { encryptedAction, encryptedNotification } from './utils.js'
+import { makeMentionFromUserID } from '@model/chatroom/utils.js'
+import messageReceivePostEffect from '@model/notifications/messageReceivePostEffect.js'
+
+sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
+  contractID,
+  data,
+  innerSigningContractID,
+  newMessage
```
taoeffect, line 22: Please add a comment here to explain that if `newMessage` is undefined, it means an existing message is being edited.

frontend/views/containers/group-settings/InvitationLinkModal.vue
```diff
@@ -28,10 +28,10 @@ import sbp from '@sbp/sbp'
 import { mapGetters } from 'vuex'
 import ModalTemplate from '@components/modal/ModalTemplate.vue'
 import LinkToCopy from '@components/LinkToCopy.vue'
-import { buildInvitationUrl } from '@model/contracts/shared/voting/proposals.js'
 import { humanDate } from '@model/contracts/shared/time.js'
 import { REPLACE_MODAL } from '@utils/events.js'
 import SvgBrokenLink from '@svgs/broken-link.svg'
+import { buildInvitationUrl } from '@view-utils/buildInvitationUrl'
```
taoeffect: Please always include the file extension in the import

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -122,12 +122,12 @@ import PageSection from '@components/PageSection.vue'
 import Tooltip from '@components/Tooltip.vue'
 import SvgInvitation from '@svgs/invitation.svg'
 import LinkToCopy from '@components/LinkToCopy.vue'
-import { buildInvitationUrl } from '@model/contracts/shared/voting/proposals.js'
 import { INVITE_STATUS } from '~/shared/domains/chelonia/constants.js'
 import { INVITE_INITIAL_CREATOR } from '@model/contracts/shared/constants.js'
 import { OPEN_MODAL } from '@utils/events.js'
 import { mapGetters, mapState } from 'vuex'
 import { L } from '@common/common.js'
+import { buildInvitationUrl } from '@view-utils/buildInvitationUrl'
```
taoeffect: Same here

frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -81,8 +81,8 @@ import {
 } from '@model/contracts/shared/constants.js'
 import currencies from '@model/contracts/shared/currencies.js'
 import { humanDate } from '@model/contracts/shared/time.js'
-import { buildInvitationUrl } from '@model/contracts/shared/voting/proposals.js'
 import { RULE_DISAGREEMENT, RULE_PERCENTAGE, VOTE_AGAINST, VOTE_FOR, getPercentFromDecimal } from '@model/contracts/shared/voting/rules.js'
+import { buildInvitationUrl } from '@view-utils/buildInvitationUrl'
```
taoeffect: And here

frontend/model/state.js
```diff
@@ -92,26 +92,65 @@ sbp('sbp/selectors/register', {
...
+        // DM chatrooms
+        return Object.keys((state[ourIdentityContractId].chatRooms: { [string]: Object })).map((chatroomID) => {
+          return state[chatroomID]?.members[ourIdentityContractId] && !Array.isArray(state[chatroomID].attributes?.adminIDs) && chatroomID
+        }).forEach((contractID) => {
+          if (!contractID) return
+          sbp('gi.actions/chatroom/upgradeFrom1.0.8', { contractID }).catch(e => {
+            console.error('[state/vuex/postUpgradeVerification] Error during gi.actions/chatroom/upgradeFrom1.0.8', contractID, e)
+          })
+        })
+      }
```
taoeffect: I ran this logic through Claude Sonnet 3.5 and it suggested this:

```js
(() => {
  // Upgrade from version 1.0.8 to a newer version
  // The new chatroom contracts have an admin IDs list
  // This code checks if the attribute is missing, and if so, issues the
  // corresponding upgrade action.
  const ourIdentityContractId = state.loggedIn?.identityContractID
  if (!ourIdentityContractId || !state[ourIdentityContractId]) return

  const upgradeAction = async (contractID, data = null) => {
    try {
      await sbp('gi.actions/chatroom/upgradeFrom1.0.8', { contractID, data })
    } catch (e) {
      console.error(`[state/vuex/postUpgradeVerification] Error during gi.actions/chatroom/upgradeFrom1.0.8 for ${contractID}:`, e)
    }
  }

  const needsUpgrade = (chatroomID) => !Array.isArray(state[chatroomID]?.attributes?.adminIDs)

  // Group chatrooms
  if (state[ourIdentityContractId].groups) {
    Object.entries(state[ourIdentityContractId].groups).forEach(([groupID, { hasLeft }]) => {
      if (hasLeft || !state[groupID]?.chatRooms || !state[groupID].groupOwnerID) return

      Object.entries(state[groupID].chatRooms).forEach(([chatroomID, { members }]) => {
        if (members[ourIdentityContractId]?.status === PROFILE_STATUS.ACTIVE && needsUpgrade(chatroomID)) {
          upgradeAction(chatroomID, state[groupID].groupOwnerID)
        }
      })
    })
  }

  // DM chatrooms
  if (state[ourIdentityContractId].chatRooms) {
    Object.keys(state[ourIdentityContractId].chatRooms).forEach((chatroomID) => {
      if (state[chatroomID]?.members[ourIdentityContractId] && needsUpgrade(chatroomID)) {
        upgradeAction(chatroomID)
      }
    })
  }
})()
```

Is there anything to that or is it nonsense?

taoeffect (changes requested):
Think this PR review is finished.... hopefully didn't miss anything.

frontend/controller/actions/group.js
corrideat: That was on purpose, as I though the admin should always be the group owner.

frontend/model/state.js
corrideat: I don't see anything obviously wrong with it at a glance, but I'm not sure what the point of this is. Was it for it to be clearer / better organised? If so, I think the `upgradeAction` and `needsUpgrade` functions make the intent clearer and is slightly less repetitive.

frontend/controller/actions/group.js
taoeffect: @corrideat you can do that but then it must be done in a way where it's added to any existing admins, instead of overwriting them

frontend/model/state.js
taoeffect: > Was it for it to be clearer / better organised?

Yes, and DRYs it a bit.

corrideat: I've taken some of the suggestions.

frontend/controller/actions/group.js
corrideat: Put admin first so that it can be overridden.

taoeffect (approved):
LGTM

PR #2365, opened 2024-09-29
Notifications without Vuex

corrideat:


frontend/model/contracts/group.js
```diff
@@ -1686,12 +1685,6 @@ sbp('chelonia/defineContract', {
       }
 
       if (memberID === identityContractID) {
-        // NOTE: remove all notifications whose scope is in this group
-        // TODO: FIND ANOTHER WAY OF DOING THIS WITHOUT ROOTGETTERS
-        for (const notification of sbp('state/vuex/getters').notificationsByGroup(contractID)) {
```
corrideat: This logic was moved to the `LEFT_GROUP` event handler.

frontend/controller/actions/identity-kv.js
```diff
@@ -251,6 +251,7 @@ export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/kv/loadNotificationStatus': () => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const status = await sbp('gi.actions/identity/kv/fetchNotificationStatus')
+      // TODO: Do this without Vuex
```
corrideat: This is temporary. It should be fixed before this PR is merged.

frontend/views/containers/notifications/NotificationModal.vue
```diff
@@ -49,7 +49,7 @@ export default {
       })
     },
     markAllNotificationsAsRead () {
-      sbp('gi.notifications/markAllAsRead')
+      sbp('gi.notifications/markAllAsRead', this.currentGroupId)
```
corrideat, line 52: This is needed because the SW doesn't know the current group ID.

frontend/views/containers/notifications/NotificationCard.vue
```diff
@@ -56,7 +56,7 @@ export default {
       this.toggleTooltip()
     },
     markAllNotificationsAsRead () {
-      sbp('gi.notifications/markAllAsRead')
+      sbp('gi.notifications/markAllAsRead', this.currentGroupId)
```
corrideat, line 59: This is needed because the SW doesn't know the current group ID.

frontend/model/notifications/selectors.js
```diff
@@ -39,17 +39,41 @@ sbp('sbp/selectors/register', {
...
+    if (!rootState.notifications) return
+    const hashes = rootState.notifications.items.filter(item => {
+      return !item.read && (!groupID || !item.groupID || item.groupID === groupID)
+    }).map(item => item.hash)
     sbp('gi.actions/identity/kv/markNotificationStatusRead', hashes)
+  },
+  'gi.notifications/remove' (hashes: string | string[]) {
+    if (!Array.isArray(hashes)) hashes = [hashes]
+    const rootState = sbp('chelonia/rootState')
+    if (rootState.notifications) return
```
taoeffect: This seems like a bug?

corrideat: Do you mean that it should be `if (!rootState.notifications) return`? If so, you're right. Good catch.

If you mean that `rootState.notifications` would always be defined, that's not true. Calling `chelonia/reset` or just starting up the app would clear it.

If you mean something else, please explain (you did find a bug regardless).

taoeffect: Yes, I meant the condition should be negated

corrideat:
@taoeffect Is there anything left in this PR? Is it ready to be merged?

frontend/model/contracts/group.js
taoeffect: Please add a comment then here explaining that.

frontend/model/notifications/selectors.js
```diff
@@ -39,17 +39,53 @@ sbp('sbp/selectors/register', {
...
     }
+    const rootState = sbp('chelonia/rootState')
+    if (!rootState.notifications) {
+      rootState.notifications = { items: [], status: {} }
+    }
+    if (rootState.notifications.items.some(item => item.hash === notification.hash)) {
+      // We cannot throw here, as this code might be called from within a contract side effect.
+      return console.error('This notification is already in the store.')
+    }
+    const index = rootState.notifications.items.findLastIndex(item => item < notification.timestamp)
```
taoeffect: `item` cannot be compared to a timestamp like this

frontend/controller/actions/identity-kv.js
```diff
@@ -251,7 +251,7 @@ export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/kv/loadNotificationStatus': () => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const status = await sbp('gi.actions/identity/kv/fetchNotificationStatus')
-      sbp('state/vuex/commit', 'setNotificationStatus', status)
+      sbp('gi.notifications/setNotificationStatus', status)
```
taoeffect, line 254: This no longer seems to perform any modifications to the Vuex state. Is that intentional?

taoeffect (changes requested):
Prelim review

frontend/controller/actions/identity-kv.js
corrideat: Yes, that's very much intentional, since there's no Vuex in the SW. However, this will result in an event that ultimately sets the Vuex state.

frontend/model/notifications/selectors.js
corrideat: Good catch

frontend/views/containers/notifications/NotificationCard.vue
corrideat: See this addition: `sbp('okTurtles.events/emit', NOTIFICATION_STATUS_LOADED, status)`

frontend/controller/actions/identity-kv.js
corrideat: See this addition: `sbp('okTurtles.events/emit', NOTIFICATION_STATUS_LOADED, status)`

frontend/model/notifications/selectors.js
```diff
@@ -39,17 +39,53 @@ sbp('sbp/selectors/register', {
       timestamp: data.createdDate ? new Date(data.createdDate).getTime() : Date.now(),
       type
     }
+    const rootState = sbp('chelonia/rootState')
```
taoeffect, line 42: (Commenting not on this line but the file as a whole)

This file seems misplaced to me. It is not concerned with state management (model), but rather performing actions. So it should be somewhere under `controller/`.

Another thing that's unclear to me is where exactly under `controller/` it should be. We should be able to post notifications when there are no open tabs right? So this should probably be moved under `controller/actions/notifications/selectors.js`

taoeffect (changes requested):
For some reason when trying to create a generic proposal in this PR using `GI_PERSIST=sqlite grunt dev` (EDIT: this happened with `grunt dev` too) I got this error:

Screenshot 2024-10-03 at 1 08 03 PM

corrideat:
I see! That particular error seems to be entirely unrelated to this PR and also exists in `master`. Nevertheless, I've fixed it now in the latest commit (wrong import).

taoeffect (approved):

PR #2366, opened 2024-09-30
Corrideat/task/#2341 anyone can join invite issue

corrideat:


frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -237,22 +238,16 @@ export default ({
...
-      const days = Math.floor(timeLeft / MIL_DAY)
-      remainder = timeLeft % MIL_DAY
-      const hours = Math.floor(remainder / MIL_HR)
-      remainder = remainder % MIL_HR
-      const minutes = Math.ceil(remainder / MIL_MIN)
+      if (expiryTime == null) return
+      const { years, months, days, hours, minutes } = timeLeft(expiryTime)
+      // In the cases when displaying years/months, count the remainer hours/mins as +1 day eg) 3days 15hrs 25mins -> 4days.
+      if (years) return L('{years}yr {months}mth {days}d left', { years, months, days })
+      if (months) return L('{months}mth {days}d left', { months, days })
```
taoeffect: Maybe `y` for your, `mo` for month?

corrideat:
@greg Looks like this should be ready!

frontend/model/contracts/shared/time.js
```diff
@@ -257,3 +259,39 @@ export function cycleAtDate (atDate: string | Date): number {
...
   return partialCycles
 }
+
+export function timeLeft (expiryTime: number) {
+  const now = new Date()
+  const expiry = new Date(expiryTime)
+
+  if (expiry < now) {
+    return {}
+  }
```
taoeffect: Maybe it should return the same type?

```js
return { years: 0, months: 0, days: 0, hours: 0, minutes: 0 }
```

Or..

```js
return { expired: true, years: 0, months: 0, days: 0, hours: 0, minutes: 0 }
```

frontend/model/contracts/shared/time.js
```diff
@@ -257,3 +259,39 @@ export function cycleAtDate (atDate: string | Date): number {
...
+    const lastMonth = new Date(expiry.getFullYear(), expiry.getMonth(), 0)
+    days += lastMonth.getDate()
+    months--
+  }
+  if (months < 0) {
+    months += 12
+    years--
+  }
+
+  return { years, months, days, hours, minutes }
```
taoeffect: Similarly, for consistency of return object shape:

```js
return { expired: false, years, months, days, hours, minutes }
```

taoeffect (commented):
Prelim

frontend/model/contracts/shared/time.js
corrideat: The `expired` looks like a good idea, but I'm not sure about setting the other fields to nonsensical values (nothing other than `null`, `undefined` or `NaN` makes sense, IMO). With TS (and likely Flow too) you can tell the two shapes apart `{ expired: false, year: number, ... } | { expired: false }` and at worst rely on type guard functions.

corrideat: The other values that would make sense would be (negative?) numbers indicating how long ago expiry happened, but this sounds like an unneeded complication since we don't use those values.

taoeffect: If the name of the function is `timeLeft`, it should return the time left IMO. In this case, either `0` or a negative value along with `{ expired: true, ... }` makes sense to me. I agree unnecessary complexity is a bad thing, so we don't need to calculate negative values, returning `0` should be fine.

corrideat: Yeah, but I don't get the point of defining a value that will not be used. In fact, the reason for returning an empty object was to simplify code (to support destructuring and to avoid needing a try block), but setting `year` and all other fields to 0 (or any other value, except negative numbers that say when it expired) would provide exactly nil useful information

corrideat: I mean, if you feel strongly about this, it's a trivial change, but I don't really see any situations where it'd be useful (the `expired` field OTOH, would potentially be useful)

taoeffect: Because it is useful to developers who are using this API. Say you are working on the frontend and want to display the amount of time left on an invite. You can use these values and bind them to the UI, to show the user that the invite has "0 seconds left", and the color it in red. Or say you want to destructure the return value using the same logic and avoid having an if/else based on whether or not the `years` value is defined. Or say you want to, in the future, using TypeScript, define a custom type that represents the return value of this function. Again, that would be simpler by keeping the same object shape.

corrideat: > Or say you want to, in the future, using TypeScript, define a custom type that represents the return value of this function

Like mentioned, that's not an issue at all. I even provided an example. In fact, I think it's encouraged in TypeScript to use objects with different shapes rather than try to fit everything into a one-size-fits-all type.

> You can use these values and bind them to the UI, to show the user that the invite has "0 seconds left"

We don't return seconds, but even then you shouldn't be displaying '0' seconds, you should be checking if `expired` is true or not.

In fact, the logic would be more complicated because you'd need something like:

```
if (years === 0 && months === 0 && hours === 0 && minutes === 0) {
return '0 but in red'
} else {
return `{years}:{months}:{hours}:{minutes} left`
}
```

as opposed to 

```
if (expired) {
return '0 but in red'
} else {
return `{years}:{months}:{hours}:{minutes} left`
}
```

> Or say you want to destructure the return value using the same logic and avoid having an if/else based on whether or not the years value is defined

Sure, but you shouldn't be doing this to begin with (checking if `year` is defined).

taoeffect: > In fact, the logic would be more complicated because you'd need something like:

That's not what I had in mind... rather I was thinking something like this:

```pug
i18n(:class="{ red: expired }" :args="{ minutes }") {minutes} left
```

I don't see a downside to returning `return { expired: true, years: 0, months: 0, days: 0, hours: 0, minutes: 0 }`, but I do see potential negative unintended consequences of not doing so.

Also, remember that as these functions are shared with the UI, the should not be modified once they are created because the UI doesn't version-control them. So a more flexible API is preferable IMO to less flexible one.

corrideat: I'll go ahead and return 0 to settle this, but I still don't think it's a good idea, because it's GIGO at that point.

> That's not what I had in mind... rather I was thinking something like this:

Sure, that makes sense, but it's not how this function was intended to be used (it was intended to be used for unexpired values).

> I do see potential negative unintended consequences of not doing so.

I do, because relying on specific values means that, e.g., if in the future we decided to show how long ago a value expired (changing this function) then those functions will be displaying the wrong values. The current design is based on [making invalid states unrepresentable](https://hugotunius.se/2020/05/16/making-invalid-state-unrepresentable.html).

> remember that as these functions are shared with the UI, the should not be modified once they are created because the UI doesn't version-control them.

Although I don't like breaking APIs and I generally agree, I don't see the issue here. As (1) this function is meant to be used by the UI only and (2) the UI is itself version controlled by virtue of whatever VCS we use for GI. Presumably, if we were to change this function to return an array, the UI could be updated simultaneously to use an array. Also (3) even if a contract uses this for some reason, those values don't directly affect the UI. And (4), I don't see how this is 'flexible', it's just defining some 'sensible' constant value that doesn't contain any useful information, which being relied upon will make updating _harder_, not _easier_.

As mentioned, I'll make the requested change as it's not the end of the world at the end of the day, but I only see downsides and no upsides.

taoeffect (changes requested):
```
Error -------------------------------------------------------------------- frontend/model/contracts/shared/time.js:177:4

Cannot expect string as the return type of function because string [1] is incompatible with implicitly-returned
undefined. [incompatible-return]

   177| ): string {
           ^^^^^^ [1]


Error ------------------------------------------------------------------- frontend/model/contracts/shared/time.js:263:46

Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type
annotation at function return: [signature-verification-failure]

   263| export function timeLeft (expiryTime: number) {
                                                     
```

taoeffect (changes requested):
A few weird errors after visiting `http://localhost:3000/app/group-settings` using `GI_PERSIST=sqlite grunt dev` with the data I sent to you on GIG DM.

First there was this:

Screenshot 2024-10-04 at 11 58 07 AM

Not sure if that's a big deal or not.

But then 3 different invites were created, when there should only be one:

Screenshot 2024-10-04 at 11 58 33 AM

And they are showing `0/60`, when they should show `0/150`.

It also took a while for the new invites to appear...

corrideat:
1. The lastLoggedIn appears to be unrelated
2. 3 different invites, I'll see what you're doing.
3. And they are showing 0/60, when they should show 0/150. Not sure where you're getting the 150 number (maybe you mean 180?), but it's correct that _each_ has 60, or whatever the number is.

taoeffect:
> And they are showing 0/60, when they should show 0/150. Not sure where you're getting the 150 number (maybe you mean 180?), but it's correct that each has 60, or whatever the number is.

No I mean 150, please see this review comment for details:

https://github.com/okTurtles/group-income/pull/2351#pullrequestreview-2316303623

corrideat:
I see. Well, that count logic looks needlessly complicated and confusing.

Besides that, I see now what you mean with the various invite links popping up. They shouldn't. It seems like there's three things to fix:

1. Invite number
2. Avoid calling the action when it shouldn't be
3. Make it so that calling the action is a no-op if the invite hasn't expired.

taoeffect:
Agreed on most counts except this:

> 3. Make it so that calling the action is a no-op if the invite hasn't expired.

It needs to be updated even for non-expired invites. I.e. if someone just now created a group with an anyone-can-join invite of `0/60`, it needs to be replaced with one that's `0/150`. (and `1/60` needs to be replaced with `1/150`, etc.), and also one that doesn't expire.

frontend/model/state.js
```diff
@@ -157,6 +157,28 @@ sbp('sbp/selectors/register', {
...
+        if (hasLeft || !groupState?.invites) return undefined
+        const now = Date.now()
+        const usedInvites = Object.keys(groupState.invites)
+          .filter(invite => groupState.invites[invite].creatorID === INVITE_INITIAL_CREATOR)
+          .reduce((acc, cv) => acc +
+        (!(groupState._vm.invites[cv].expires >= now)
+          ? groupState._vm.invites[cv].initialQuantity
+          : ((groupState._vm.invites[cv].initialQuantity - groupState._vm.invites[cv].quantity)
+            ) || 0), 0)
+        return (usedInvites < 150) ? groupID : undefined
```
taoeffect: Everywhere we use this value `150`, could you replace it with `MAX_GROUP_MEMBER_COUNT`?

corrideat: Sure

corrideat:
I've noticed an issue that `chelonia/time` returns a value in seconds instead of milliseconds (as JS APIs do). I need to double check that usage of `chelonia/time` takes this into account.

frontend/controller/actions/group.js
```diff
@@ -180,8 +182,11 @@ export default (sbp('sbp/selectors/register', {
             ringLevel: Number.MAX_SAFE_INTEGER,
             permissions: [GIMessage.OP_KEY_REQUEST],
             meta: {
-              quantity: 60,
-              expires: Date.now() + DAYS_MILLIS * INVITE_EXPIRES_IN_DAYS.ON_BOARDING,
+              quantity: MAX_GROUP_MEMBER_COUNT,
+              ...(INVITE_EXPIRES_IN_DAYS.ON_BOARDING && {
+                expires:
+                await sbp('chelonia/time') + DAYS_MILLIS * INVITE_EXPIRES_IN_DAYS.ON_BOARDING
```
taoeffect: Before we had `Date.now()` (unix timestamp in millis) here, and now we have what I think is unix timestamp in seconds - so a mismatch

frontend/controller/actions/group.js
```diff
@@ -907,6 +912,72 @@ export default (sbp('sbp/selectors/register', {
...
+    const state = await sbp('chelonia/contract/state', contractID)
+    const now = await sbp('chelonia/time')
+
+    // For used invites, we add all 'anyone can join' invites and do the following:
+    //  1. For non-expired invites, we consider all of the invites
+    //       (initialQuantity) as used
+    //  2. For expired invites, we consider the quantity actually used
+    //  3. We add up 1. and 2. above
+    // Then, we take the difference and, if the number is less than
+    // MAX_GROUP_MEMBER_COUNT, we create a new invite.
```
taoeffect: This comment doesn't make sense to me for the following reasons:

> we add all 'anyone can join' invites

This is plural - but there is only one anyone-can-join invite when this code is run. This code should never be run more than once - by postUpgradeVerification, to fix the current situation we're in. And in the current situation, there can only ever be a single anyone-can-join invite.

> 1. For non-expired invites, we consider all of the invites (initialQuantity) as used as used
> 2. For expired invites, we consider the quantity actually used

We should not be touching or caring about expired (or non-expired) invites that are not the anyone-can-join invite. Those should be left alone completely.

And for the anyone-can-join invite, whether or not it is expired is irrelevant. We want to replace the *single* anyone-can-join invite with a different, single, anyone-can-join invite (and pretend like the original one never existed in the first place).

Then we take the number of used invites, and subtract that from `MAX_GROUP_MEMBER_COUNT`, and that is the amount we have remaining.

frontend/model/contracts/shared/time.js
```diff
@@ -184,7 +185,8 @@ export function humanDate (
     : ((navigator.languages: any): string[]) ?? navigator.language ?? fallback
   // NOTE: `.toLocaleDateString()` automatically takes local timezone differences into account.
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
-  return new Date(date).toLocaleDateString(locale, options)
+  const dateObj = new Date(date)
+  if (!isNaN(dateObj.valueOf())) return dateObj.toLocaleDateString(locale, options)
```
taoeffect, line 191: I'm not a fan of this change. Why is this being done?

Is it possible to instead of making the change here, handling any potential issues at the callsite?

frontend/model/state.js
```diff
@@ -157,6 +157,28 @@ sbp('sbp/selectors/register', {
           upgradeAction(contractID)
         })
       }
+    })();
+    (() => {
+      // Update expired invites
+      // If fewer than MAX_GROUP_MEMBER_COUNT 'anyone can join' have been used,
+      // create a new 'anyone can join' link up to MAX_GROUP_MEMBER_COUNT invites
```
taoeffect, line 257: Can this comment be updated to add that we're not just creating a new invite, but we're replacing the old one?

corrideat: That'd be inaccurate. This is creating a new invite, not replacing the old one.

frontend/model/contracts/shared/time.js
corrideat: Because otherwise it'll return 'Invalid Date', which most (all?) of the time isn't what one wants.

frontend/model/state.js
```diff
@@ -157,6 +157,28 @@ sbp('sbp/selectors/register', {
...
+      const ourIdentityContractId = state.loggedIn?.identityContractID
+      if (!ourIdentityContractId || !state[ourIdentityContractId]?.groups) return
+      Object.entries(state[ourIdentityContractId].groups).map(([groupID, { hasLeft }]: [string, Object]) => {
+        const groupState = state[groupID]
+        if (hasLeft || !groupState?.invites) return undefined
+        const now = Date.now()
+        const usedInvites = Object.keys(groupState.invites)
+          .filter(invite => groupState.invites[invite].creatorID === INVITE_INITIAL_CREATOR)
+          .reduce((acc, cv) => acc +
+        (!(groupState._vm.invites[cv].expires >= now)
```
taoeffect: All anyone-can-join invites need to be updated, regardless of whether or not they're expired.

frontend/model/state.js
taoeffect: I don't think it's inaccurate. Yes, we're creating a new one, but the purpose is to replace the old one, and this isn't mentioned in the comment.

frontend/controller/actions/group.js
corrideat: > We should not be touching or caring about expired (or non-expired) invites that are not the anyone-can-join invite. Those should be left alone completely.

That is exactly what this is doing.

> This is plural - but there is only one anyone-can-join invite when this code is run. 

Right now, there can be multiple anyone-can-join invites.

frontend/model/contracts/shared/time.js
taoeffect: You didn't answer my question tho...

> Is it possible to instead of making the change here, handling any potential issues at the callsite?

frontend/model/state.js
corrideat: They're not updated, a new invite is created. This logic takes into account how many people to allow using them, as explained in the comments.

frontend/controller/actions/group.js
taoeffect: > That is exactly what this is doing.

But the comment doesn't make this clear.

> Right now, there can be multiple anyone-can-join invites.

Please elaborate how this is possible when the app only creates one?

frontend/model/contracts/shared/time.js
corrideat: > Is it possible to instead of making the change here, handling any potential issues at the callsite?

Not really, because you need to run `new Date()` to know if the value will be invalid or not.

frontend/controller/actions/group.js
corrideat: The app creates multiple.

For example, if you created a group with 60 invites (1 of which was used and then expired), now you'll have 2 invite links, one expired, and one with 149 slots that doesn't expire.

For example, if you created a group with 60 invites (1 of which was used and hasn't yet expired), now you'll have 2 invite links, one non-expired with 59 remaining, and one with 91 slots that doesn't expire.

frontend/model/state.js
corrideat: https://github.com/okTurtles/group-income/pull/2366#discussion_r1790653189

frontend/controller/actions/group.js
taoeffect: > For example, if you created a group with 60 invites (1 of which was used and then expired), now you'll have 2 invite links, one expired, and one with 149 slots that doesn't expire.

But this code is for getting rid of the 60-max invite and replacing it with the 150-max invite, isn't it? At least that's what it's supposed to be doing.

> For example, if you created a group with 60 invites (1 of which was used and hasn't yet expired), now you'll have 2 invite links, one non-expired with 59 remaining, and one with 91 slots that doesn't expire.

That's not what this PR is supposed to do. The PR is supposed to get rid of the 60-max invite and replace it with a 150-max invite (subtracting however many from the 60-max were used).

So no, you shouldn't have 2 invite links. And in this example, a new non-expiring anyone-can-join invite should be created with 149 invites remaining.

corrideat: I see, but doing it the way you suggest will break invite links that were already shared.

taoeffect: That's OK, they can be re-shared, and we can remember to announce that when we release the update.

corrideat: In any case, the math here is still (almost) correct because you could have multiple invite links when using, e.g., an alternate client, so you need to add all of them up.

taoeffect: In fact, if you'd like to update this PR with a new entry for the `NewsAndUpdates.vue` file, please feel free to do so and add this text:

```js
  {
    createdAt: new Date('2024-10-29'),
    title: '1.1.0: Anyone-can-join Invites Updated! Please read!',
    content: "Version 1.1.0 introduces many new features and bug fixes, and among them is replacing the old anyone-can-join invite links with new ones that will let your group grow to Dunbar's Number.\n\n" +
      '**The old links are expired, please use the new ones!**'
  },
```

frontend/views/containers/dashboard/GroupMembers.vue
```diff
@@ -8,8 +8,7 @@
       @click='headerButtonAction'
     )
       i.icon-plus.is-prefix
-      i18n(v-if='action === "addMember"') Add
-      i18n(v-else) New
```
taoeffect, line 12: Why is this logic changed? 🤔

frontend/views/containers/payments/PaymentsMixin.js
```diff
@@ -187,7 +187,7 @@ const PaymentsMixin: Object = {
     // Returns a human-readable description of the time interval identified by a given period stamp.
     getPeriodFromStartToDueDate (period, knownPeriods) {
       const dueDate = this.dueDateForPeriod(period, knownPeriods)
-      return `${humanDate(dateFromPeriodStamp(period))} - ${humanDate(dateFromPeriodStamp(dueDate))}`
+      return `${humanDate(dateFromPeriodStamp(period)) || ''} - ${humanDate(dateFromPeriodStamp(dueDate)) || ''}`
```
taoeffect: Can this be reverted and `humanDate` fixed to not return `undefined` but `''`?

frontend/views/containers/dashboard/GroupMembers.vue
corrideat: @SebinSong I believe you made this change, please clarify if I'm mistaken. I believe this related to fixing the issue with groups being unable to add new members if the invite link is expired or used up and there are no members.

corrideat:
@taoeffect This PR has been updated based on your feedback. Now, all non-expired invite links are removed and consolidated into one.

frontend/views/containers/dashboard/GroupMembers.vue
SebinSong: please check this comment:
https://github.com/okTurtles/group-income/pull/2351#discussion_r1760569165

The `action` prop has only has `addMember` as the option and no other value. So I just removed the prop. (Looks like there used to be other options in earlier days)

frontend/controller/actions/group.js
```diff
@@ -907,6 +912,72 @@ export default (sbp('sbp/selectors/register', {
...
+    await Promise.all(activeInvites.map(inviteKeyId => sbp('gi.actions/group/inviteRevoke', { contractID, data: { inviteKeyId } })))
+
+    await sbp('chelonia/out/keyAdd', {
+      contractName: 'gi.contracts/group',
+      contractID,
+      data: [ik],
+      signingKeyId: CSKid
+    })
+
+    await sbp('gi.actions/group/invite', { contractID, data: { inviteKeyId, creatorID: 'invite-initial-creator' } })
```
taoeffect: Please use `INVITE_INITIAL_CREATOR` instead of `'invite-initial-creator'` here

frontend/views/utils/time.js
```diff
@@ -0,0 +1,35 @@
+export function timeLeft (expiryTime: number): { expired: boolean, years: number, months: number, days: number, hours: number, minutes: number } {
```
taoeffect, line 1: We already have a `time.js` file, instead of creating a new file of the same name just for this function, could you move it to the existing one?

taoeffect (changes requested):
Nice improvements @corrideat!

I left a few comments (see above), and also tried testing it again.

I wasn't successful though with the test data that I sent you. Logged in as `u1` and switched to `Turtles (4)` group.

Old invite still there, showing wrong expiry too:

Screenshot 2024-10-09 at 12 10 05 PM

Using that invite shows it's expired.

This PR needs to also update the UI to say that this invite doesn't expire (if it doesn't have an expiry).

frontend/views/utils/time.js
corrideat: The file you're referring to is in `model/contracts/shared`, and this function isn't meant to be used by contracts.

taoeffect: I suppose that's true, although I don't see why it couldn't be used in contracts in the future. but sure, for now we can keep it here if you really prefer.

corrideat: It _could_ be used by contracts, but I don't think it _should_ be used. It adds unnecessary complexity to contracts and contracts shouldn't need 'human' dates anyhow, which is a presentation concern (for example, no math should be done by contracts on human dates, and we're relying on a specific calendar; in the future, things could be displayed in the Chinese or Islamic calendars, for example, in which case the values would differ).

corrideat:
@taoeffect I think this should be ready now

frontend/model/contracts/shared/time.js
```diff
@@ -207,11 +213,11 @@ export function isShortMonthstamp (arg: string): boolean {
   return /^(0[1-9]|1[0-2])$/.test(arg)
 }
 
-export function monthName (monthstamp: string): string {
+export function monthName (monthstamp: string): ?string {
```
taoeffect: These no longer need to be modified right? `?string` => `string`

(Same below)

frontend/model/state.js
```diff
@@ -22,6 +23,39 @@ import identityGetters from './contracts/shared/getters/identity.js'
...
+  // Add event listeners for CONTRACTS_MODIFIED, LOGOUT, LOGIN_COMPLETE and
+  // LOGIN_ERROR
+  sbp('okTurtles.events/on', CONTRACTS_MODIFIED, modifiedHandled)
+  sbp('okTurtles.events/once', LOGOUT, logoutHandler)
+  sbp('okTurtles.events/once', LOGIN_COMPLETE, loginCompleteHandler)
+  sbp('okTurtles.events/once', LOGIN_ERROR, loginErrorHandler)
+
+  // Call the update function in the next tick
+  setTimeout(updateFn, 0)
+}
```
taoeffect: So we're just going to keep calling the postUpgrade functions every time the contracts are modified?

That seems ugly.

1. Is there someway to make it so that after the upgrade gets run, it's the `modifiedHandled` is no longer called?
2. This seems like a potentially useful utility function. Can we import it from some other location instead of here? (Is there a way to make this available as a Chelonia dev utility)?

taoeffect: > Is there someway to make it so that after the upgrade gets run, it's the modifiedHandled is no longer called?

One idea that occurred to me, is that `updateFn` can pass in a function that will unregister the listener that can be called once a successful upgrade is done.

corrideat: > Is there someway to make it so that after the upgrade gets run, it's the modifiedHandled is no longer called?

Well, no, because that's the entire point of this change not to have to refresh the page. The `contractIDHints` argument is there to allow returning early.

frontend/model/state.js
```diff
@@ -90,7 +124,13 @@ sbp('sbp/selectors/register', {
...
       Vue.set(state, 'reverseNamespaceLookups', Object.fromEntries(Object.entries(state.namespaceLookups).map(([k, v]: [string, string]) => [v, k])))
     }
-    (() => {
+    contractUpdate((contractIDHints: ?string[]) => {
+      const state = sbp('state/vuex/state')
+      if (Array.isArray(contractIDHints)) {
+        if (!contractIDHints.reduce((acc, contractID) => {
+          return (acc || state.contracts[contractID]?.type === 'gi.contracts/group')
+        }, false)) return
+      }
```
taoeffect: Seeing a lot of DRY violations with this chunk of code. Seems ugly too.

Is there a way that, instead of this blob, we can pass in some parameter to `contractUpdate` when calling it that results in the same sort of logic being run to fix the DRY violation and remove this block from here?

frontend/model/state.js
```diff
@@ -157,7 +207,37 @@ sbp('sbp/selectors/register', {
...
+        const hasBeenUpdated = Object.keys(groupState.invites).some(inviteId => {
+          return groupState.invites[inviteId].creatorID === INVITE_INITIAL_CREATOR &&
+          groupState._vm.invites[inviteId].expires == null
+        })
+        if (hasBeenUpdated) return undefined
+        const usedInvites = Object.keys(groupState.invites)
+          .filter(invite => groupState.invites[invite].creatorID === INVITE_INITIAL_CREATOR)
+          .reduce((acc, cv) => acc +
+        ((groupState._vm.invites[cv].initialQuantity - groupState._vm.invites[cv].quantity) || 0), 0)
+        return (usedInvites < MAX_GROUP_MEMBER_COUNT) ? groupID : undefined
```
taoeffect: This incomprehensibly dense blob seems to be a DRY violation with the same / similar logic in `fixAnyoneCanJoinLink`?

frontend/model/state.js
```diff
@@ -499,8 +579,8 @@ const getters = {
   },
   currentWelcomeInvite (state, getters) {
     const invites = getters.currentGroupState.invites
-    const inviteId = Object.keys(invites).find(invite => invites[invite].creatorID === INVITE_INITIAL_CREATOR)
-    const expires = getters.currentGroupState._vm.authorizedKeys[inviteId].meta.expires
+    const inviteId = Object.keys(invites).find(invite => invites[invite].creatorID === INVITE_INITIAL_CREATOR && !(getters.currentGroupState._vm.invites[invite].expires >= Date.now()) && !(getters.currentGroupState._vm.invites[invite].quantity <= 0) && getters.currentGroupState._vm.invites[invite].status === INVITE_STATUS.VALID)
```
taoeffect: When lines of code grow beyond a certain point with logic that starts to break mere mortals brains, can you please, instead of inlining it like this, create a utility function that is called instead?

It will help improve code readability and potentially get rid of DRY violations too.

frontend/model/state.js
taoeffect: But some of these functions only need to be run once, right? No need to call them every single time a contract is added or removed.

Did you see this?

> One idea that occurred to me, is that `updateFn` can pass in a function that will unregister the listener that can be called once a successful upgrade is done.

frontend/model/state.js
corrideat: It's similar logic but it needs to be there to know whether `fixAnyoneCanJoinLink` should be called or not. Otherwise, we'd call `fixAnyoneCanJoinLink` on every group every time a tab is opened or a new group is subscribed.

taoeffect (changes requested):
OK, think I've finished the review, sorry it took so long!

frontend/model/state.js
corrideat: Sure

frontend/model/state.js
corrideat: It could be in `contractUpdate`, although that could be limiting. It could also be a utility function that takes `(state, contractIDHints, type)`

frontend/model/state.js
corrideat: > This seems like a potentially useful utility function. Can we import it from some other location instead of here? (Is there a way to make this available as a Chelonia dev utility)?

Useful for what, other than upgrades? Like, I'm all for moving it, but I don't see it as being particularly useful as a generic function outside it being used here.

> Is there a way to make this available as a Chelonia dev utility

One issue with that is that all of the events other than `CONTRACTS_MODIFIED` are GI events, not Chelonia events.

taoeffect: > Useful for what, other than upgrades? Like, I'm all for moving it, but I don't see it as being particularly useful as a generic function outside it being used here.

Useful for anyone writing Chelonia-based apps. This is one of 2 ways we've got for doing upgrades (and the only way that exists currently). Seems like every dev would need it.

> One issue with that is that all of the events other than CONTRACTS_MODIFIED are GI events, not Chelonia events.

Honestly, with how complicated logging in and logging out is, I think we will be forced to move that code into Chelonia.

corrideat: > with how complicated logging in and logging out is, I think we will be forced to move that code into Chelonia.

I'm not sure that feasible nor a good idea, and it's not really that logging in or out is complicated: it is, in a way, but that's because there are things that need to be done in specific places (the SW and the app), and because of specific things related to saving and loading the state, which can also need encryption and decryption.

In this case, the `LOGIN_COMPLETE` and `LOGIN_ERROR` events are also only on the application / browsing-context side (implementing this function in the actions / SW side wouldn't need these events, _maybe_ it'd need `LOGIN`, but you asked to use the existing selector we have).

Moving logging in and out into Chelonia means that certain types of contracts need to be defined, rather than having a generic system.

In another type of application, you only really need to listen for the `CONTRACTS_MODIFIED` modified event.

corrideat: Can we leave this without major changes and improve it later? Like I said, I'm happy to put it in a different file (in fact, I think it doesn't belong in this already complicated file). My question is more about the 'moving it to Chelonia' part, as this is feeling like an ever-growing PR.

frontend/model/state.js
```diff
@@ -90,7 +158,12 @@ sbp('sbp/selectors/register', {
       // $FlowFixMe[incompatible-call]
       Vue.set(state, 'reverseNamespaceLookups', Object.fromEntries(Object.entries(state.namespaceLookups).map(([k, v]: [string, string]) => [v, k])))
     }
-    (() => {
+    contractUpdate(state, (state: Object, contractIDHints: ?string[]) => {
+      if (Array.isArray(contractIDHints)) {
+        if (!contractIDHints.reduce((acc, contractID) => {
+          return (acc || state.contracts[contractID]?.type === 'gi.contracts/group')
+        }, false)) return
+      }
```
taoeffect: Wait.. this is still here...?

corrideat: Good catch, refactoring gone wrong

frontend/model/state.js
```diff
@@ -22,6 +24,57 @@ import identityGetters from './contracts/shared/getters/identity.js'
...
+  // Wrapper for the update function. This performs a common check, namely that
+  // the contract is of a certain type, which helps return early
+  const wrappedUpdateFn = contractType
+  // The following disable is because eslint gets confused with 'Object'
+  // eslint-disable-next-line no-use-before-define
+    ? (state: Object, contractIDHints: ?string[]) => {
+        if (Array.isArray(contractIDHints)) {
+          if (!contractIDHints.reduce((acc, contractID) => {
+            return (acc || state.contracts[contractID]?.type === contractType)
+          }, false)) return
```
taoeffect: This code seems to me, unnecessarily confusing. Could you please replace this with something simpler, e.g.:

```js
if (!contractIDHints.some(contractID => state.contracts[contractID]?.type === contractType)) {
  return
}
```

frontend/model/state.js
```diff
@@ -22,6 +24,57 @@ import identityGetters from './contracts/shared/getters/identity.js'
 import notificationModule from '~/frontend/model/notifications/vuexModule.js'
 import settingsModule from '~/frontend/model/settings/vuexModule.js'
 import chatroomModule from '~/frontend/model/chatroom/vuexModule.js'
+import { CHELONIA_RESET, CONTRACTS_MODIFIED } from '../../shared/domains/chelonia/events.js'
+
+// Wrapper function for performing contract upgrades and migrations
+// TODO: Consider moving this function into a different file
+const contractUpdate = (initialState: Object, updateFn: (state: Object, contractIDHints: ?string[]) => any, contractType: ?string) => {
```
taoeffect, line 31: Could you please move `contractType` to be the 2nd or 1st parameter?

frontend/model/state.js
```diff
@@ -22,6 +24,57 @@ import identityGetters from './contracts/shared/getters/identity.js'
...
+            return (acc || state.contracts[contractID]?.type === contractType)
+          }, false)) return
+        }
+        updateFn(state, contractIDHints)
+      }
+    : updateFn
+
+  const resetHandler = () => {
+    sbp('okTurtles.events/off', CONTRACTS_MODIFIED, modifiedHandler)
+  }
```
taoeffect, line 49: Since this is 1 line and it is only used once with a `/once` listener, it can be inlined as an anonymous function below.

frontend/model/state.js
```diff
@@ -22,6 +24,57 @@ import identityGetters from './contracts/shared/getters/identity.js'
...
+  sbp('okTurtles.events/on', CONTRACTS_MODIFIED, modifiedHandler)
+  // Receiving `CHELONIA_RESET` means that a new session has started. To prevent
+  // memory leaks and duplicate handlers, this event will remove the
+  // `CONTRACTS_MODIFIED` handler.
+  sbp('okTurtles.events/once', CHELONIA_RESET, resetHandler)
+
+  // Call the update function in the next tick
+  // We want this (in addition to `CONTRACTS_MODIFIED`) because this way we
+  // can update contracts that already exist, e.g., upon login with a saved
+  // state
```
taoeffect, line 72: This is a good comment 👍

frontend/model/state.js
```diff
@@ -101,6 +169,7 @@ sbp('sbp/selectors/register', {
       if (!ourIdentityContractId || !state[ourIdentityContractId]?.groups) return
       Object.entries(state[ourIdentityContractId].groups).map(([groupID, { hasLeft }]: [string, Object]) => {
         if (hasLeft || !state[groupID]?.chatRooms) return undefined
+        if (Array.isArray(contractIDHints) && !contractIDHints.includes(groupID)) return undefined
```
taoeffect: Isn't the point of the `wrappedUpdateFn` to handle this?

If this line is needed, can you add a 1-line comment above it explaining why `wrappedUpdateFn` is insufficient?

frontend/model/state.js
```diff
@@ -90,7 +158,7 @@ sbp('sbp/selectors/register', {
       // $FlowFixMe[incompatible-call]
       Vue.set(state, 'reverseNamespaceLookups', Object.fromEntries(Object.entries(state.namespaceLookups).map(([k, v]: [string, string]) => [v, k])))
     }
-    (() => {
+    contractUpdate(state, (state: Object, contractIDHints: ?string[]) => {
```
taoeffect, line 164: I asked Claude if it could make the code from line 170 to 182 more readable, and here's what it suggested:

```js
Object.entries(state[ourIdentityContractId].groups)
  .filter(([groupID, { hasLeft }]) => {
    return !hasLeft && 
           state[groupID]?.chatRooms && 
           (!Array.isArray(contractIDHints) || contractIDHints.includes(groupID))
  })
  .map(([groupID]) => {
    // $FlowFixMe[incompatible-use]
    const chatRooms = state[groupID].chatRooms
    const needsUpgrade = Object.values(chatRooms)
      .flatMap(({ members }) => Object.values(members))
      .some(member => 
        member.status === PROFILE_STATUS.ACTIVE && member.joinedHeight == null
      )

    return needsUpgrade ? groupID : null
  })
  .filter(Boolean)
```

If this accomplishes the same thing, do you think it's safe to replace it with this?

frontend/model/state.js
```diff
@@ -51,6 +104,21 @@ const checkedUsername = (state: Object, username: string, userID: string) => {
...
+// of those could have exipred, we need a for loop
+const anyoneCanJoinInviteId = (invites: Object, getters: Object): ?string =>
+  Object.keys(invites).find(invite =>
+    // First, we want 'anyone can join' invites
+    invites[invite].creatorID === INVITE_INITIAL_CREATOR &&
+    // and that haven't been revoked
+    getters.currentGroupState._vm.invites[invite].status === INVITE_STATUS.VALID &&
+    // and that haven't expired (using negative logic because expires could be
+    // undefined for non expiring-invites)
+    !(getters.currentGroupState._vm.invites[invite].expires >= Date.now()) &&
```
taoeffect: Copying Claude's critique here (which I think I agree with):

> The logic for checking if an invite has expired is inverted. The condition `!(getters.currentGroupState._vm.invites[invite].expires >= Date.now())` is intended to check if the invite has not expired, but it will actually return true if the invite has expired. This will cause the function to incorrectly filter out valid invites that have not expired.

frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -237,22 +238,18 @@ export default ({
...
-      let remainder
-
-      const days = Math.floor(timeLeft / MIL_DAY)
-      remainder = timeLeft % MIL_DAY
-      const hours = Math.floor(remainder / MIL_HR)
-      remainder = remainder % MIL_HR
-      const minutes = Math.ceil(remainder / MIL_MIN)
+      if (expiryTime == null) return L("Doesn't expire")
+      const { expired, years, months, days, hours, minutes } = timeLeft(expiryTime)
+      if (expired) L('Expired')
```
taoeffect: Missing `return` here

Gruntfile.js
```diff
@@ -223,7 +224,8 @@ module.exports = (grunt) => {
         'process.env.MAX_EVENTS_AFTER': `'${MAX_EVENTS_AFTER}'`,
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
         'process.env.EXPOSE_SBP': `'${EXPOSE_SBP}'`,
-        'process.env.ENABLE_UNSAFE_NULL_CRYPTO': `'${ENABLE_UNSAFE_NULL_CRYPTO}'`
+        'process.env.ENABLE_UNSAFE_NULL_CRYPTO': `'${ENABLE_UNSAFE_NULL_CRYPTO}'`,
+        'process.env.UNSAFE_TRUST_ALL_MANIFEST_SIGNING_KEYS': UNSAFE_TRUST_ALL_MANIFEST_SIGNING_KEYS
```
taoeffect: Inconsistent handling of this environment variable, could you please wrap it the same way as the others?

taoeffect (changes requested):
Another round of review ready!

Gruntfile.js
corrideat: but this is a boolean, not a string

frontend/model/state.js
corrideat: <del>
Claude is wrong.

`!(5 >= 5)` => false
`!(3 >= 5)` => true
`!(NaN >= 5)` => true

The non-inverted version is `expires < Date.now()`, which is the same as what's here `(expires >= Date.now())`, except when dealing with `NaN`.
<del>

taoeffect: Can you explain more?

If the expires is in the future from now, it means the invite hasn't expired yet. And yet, in that case, you flip the true to a false

frontend/model/state.js
corrideat: I'll add a comment. It's not sufficient because `wrappedUpdateFn` is filtering contracts by type, while this is filtering out contracts (groups) that haven't been recently added.

frontend/model/state.js
corrideat: Thanks

frontend/model/state.js
corrideat: I thought that you preferred named functions over inlined ones (and it's also more consistent in this case)

frontend/model/state.js
corrideat: It wouldn't work that well because it's meant to be an optional parameter, as you may want to update contracts regardless of type.

frontend/model/state.js
corrideat: Sure

frontend/views/containers/group-settings/InvitationsTable.vue
corrideat: Indeed

frontend/model/state.js
corrideat: Maybe, I'd need to see if they do the same thing or not. So far, it seems like it might do the same thing.

Gruntfile.js
taoeffect: So is ENABLE_UNSAFE_NULL_CRYPTO, I don't think it matters

frontend/model/state.js
corrideat: I see. I think you're right, and that it should be `<` instead of `>=`.

Background: it was meant to be the inverse of this:

```
const isWelcomeInviteExpired = this.currentWelcomeInvite.expires < Date.now()
```

I inverted the logic and then inverted it inside too, which I shouldn't have done.

Gruntfile.js
corrideat: It does matter (although I can see the argument for consistency). See:

```js
if (process.env.ENABLE_UNSAFE_NULL_CRYPTO === 'true' && ...
```

If you make it a string, then it's a string. If you don't quote it, it's whatever the type of that expression is.

taoeffect: I see. Well, can we make it consistent anyway, for consistency's sake?

frontend/model/state.js
taoeffect: I do, usually, it especially helps when there's a large or confusing chunk of logic to simplify, or to DRY things. In this case it's simple enough to inline and there's no DRY violation. However, it's not a big to leave as-is too of you prefer.

corrideat: Left as is for consistency

taoeffect (changes requested):
I tried upgrading with our group data and got a lot of these errors. Will send you data so that you can try yourself.

Screenshot 2024-10-15 at 11 58 50 AM

corrideat:
Thanks for the data! I've fixed a few issues with the upgrade code, and now you shouldn't see that error. Where it happens, it's harmless, so I've also made it into a warning.

frontend/model/state.js
```diff
@@ -99,36 +170,52 @@ sbp('sbp/selectors/register', {
...
       const upgradeAction = async (contractID: string, data?: Object) => {
         try {
           await sbp('gi.actions/chatroom/upgradeFrom1.0.8', { contractID, data })
         } catch (e) {
+          // If the action failed because the upgrade has already happened, we
+          // can safely ignore the error
+          if (e.message?.includes('Upgrade can only be done once')) {
+            console.warn(`[state/vuex/postUpgradeVerification] Error during gi.actions/chatroom/upgradeFrom1.0.8 for ${contractID}:`, e)
+            return
+          }
```
taoeffect, line 218: Why issue the action at all just to capture this error? Is it possible to instead avoid sending the action?

corrideat: I don't think so because new events could have been received between the time the check was made and the time the action is sent (i.e., no matter how much or how often you check, there can still be a new event that does this update).

taoeffect (approved):
Approved! Amazing work @corrideat - this was a tough PR!

PR #2367, opened 2024-10-02
#2347 - Build Image viewer in the app

### LINKED ISSUE #2347: Picture viewer
### Problem

We don't have a way to zoom the pictures to view them more closely.

### Solution

Create a picture viewer so that when a picture attachment clicked (or tapped) then it is shown in a view in more detail similar to how other apps do it.

On mobile it should be very simple - simply make the picture full screen and make it possible to pinch to zoom in and out.

SebinSong:
closes #2347 

**[ Screenshot - Desktop ]**

<br>
<br>

**[ Screenshot - Mobile ]**

On mobile browser, zooming in/out can be done both by using the slider at the bottom and via pinch zoom gestures. (tested on my Android and worked well)

<br>
<br>

This viewer is triggered by clicking on two places in the chatroom.


frontend/views/containers/chatroom/image-viewer/ImageViwerModal.vue
```diff
@@ -0,0 +1,233 @@
...
+import { mapGetters } from 'vuex'
+import sbp from '@sbp/sbp'
+import trapFocus from '@utils/trapFocus.js'
+import { CLOSE_MODAL } from '@utils/events.js'
+import AvatarUser from '@components/AvatarUser.vue'
+import PreviewImageArea from './PreviewImageArea.vue'
+
+export default {
+  // NOTE: gave this component a generic name in case this is used outside the chatroom area. (eg. instead of 'ChatImageViewer' etc.)
+  name: 'ImageViewerModal',
```
taoeffect: Looks like the filename has a typo in it :)

SebinSong: good catch. fixed.

taoeffect (changes requested):
Fantastic work @SebinSong!!

I tested it and it seems to work almost perfectly!

I found a few issues:

1. On Firefox Desktop when dragging an image:

https://github.com/user-attachments/assets/dbbb2838-50de-4378-b24f-001e32a5fbf6

As you can see, sometimes when you drag it too far it will get "stuck" and from that point on you won't be able to keep dragging it. 

2. On mobile, pinching in and zooming is a bit "stuttery" and not smooth.
3. On mobile, pinching in and zooming does not zoom into the origin point of where you are pinching in to. Instead it zooms from the center. Try uploading a large picture and zooming in - it will zoom using the origin (the center of the image) as the zoom point. This makes it difficult to zoom in on say the top-right of an image, because you have to zoom, and then scroll. You can't just zoom to the point you want to immediately.


EDIT: also a question: how difficult would it be to have the viewer cycle between photos using the ◀️▶️ arrow keys on desktop if the message contains multiple photo attachments? Is this something that you think can easily be done in this PR or should it be saved for a separate issue?

SebinSong:
@taoeffect 

> how difficult would it be to have the viewer cycle between photos using the ◀️▶️ arrow keys on desktop if the message contains multiple photo attachments? Is this something that you think can easily be done in this PR or should it be saved for a separate issue?

I think it's something achievable for sure but would love to work on it as a separate PR.

frontend/views/containers/chatroom/image-viewer/PreviewImageArea.vue
```diff
@@ -0,0 +1,375 @@
+<template lang="pug">
+.c-image-view-area(
+  @wheel.prevent.stop='wheelEventHandler'
+)
+  img.c-preview-image(ref='previewImg'
```
corrideat, line 7: For accessibility, `img` elements need an `alt` attribute

SebinSong:
@taoeffect 
> sometimes when you drag it too far it will get "stuck" and from that point on you won't be able to keep dragging it.

How far it can be draggable is calculated by below block,

```js
    movableDistances () {
      // calculates how much vertical/horizontal spaces are available to move image around in.

      if (!this.isImageMovable) {
        return { x: 0, y: 0 }
      } else {
        const percentDiff = (this.ephemeral.currentZoom - this.config.zoomMin) / 100
        return {
          x: this.config.imgData.naturalWidth * percentDiff / 2,
          y: this.config.imgData.naturalHeight * percentDiff / 2
        }
      }
    }
```

and this is actually based on my observation in the Slack image-viewer where as long as the edge of the image hits the viewer's edge, it apparently decides that those corner/edge areas are visible and stops to move.
For example, in the screenshot below,  I can't move further down beyond the below point.

image

frontend/views/containers/chatroom/image-viewer/PreviewImageArea.vue
SebinSong: good catch

SebinSong:
@taoeffect The PR has been updated with the work for your feedback.

SebinSong:
@taoeffect 
regarding

> On mobile, pinching in and zooming is a bit "stuttery" and not smooth.

I decreased the throttling value for `pointermove` event handler for this which I felt at least improved how smoothly it reacts to my touch actions. Hope this improvement happens to your devices too.

frontend/views/containers/chatroom/image-viewer/PreviewImageArea.vue
```diff
@@ -0,0 +1,457 @@
...
+    :class='{ "is-movable": isImageMovable }'
+    :src='imgSrc'
+    :width='config.imgData.naturalWidth'
+    :height='config.imgData.naturalHeight'
+    :style='previewImgStyles'
+    :alt='name'
```
taoeffect: Is "name" a good `alt` for this? Seems a bit unclear to me.

Maybe "Image preview"? And probably should be wrapped in an `L` function for localization.

SebinSong: @taoeffect OK, will use `L('Image preview')` then.

SebinSong:
BTW,
since merging the latest `master` ,`group-chat-message.spec.js` intermittently fails with below error.

image

taoeffect (changes requested):
Nice @SebinSong! Zooming now works well on mobile AFAICT, great job! (What issue did you encounter?)

Dragging also works well now!

On Desktop though, I should be able to zoom in using the scroll wheel, and that doesn't work. Is that something you can fix?

SebinSong:
@taoeffect The PR is ready for review again.

taoeffect (approved):
Incredible work @SebinSong!!! 😄  🎉 🎉 🎉 🎉 🎉 🎉 

✅ Tested and verified on desktop and mobile!

SebinSong:
Thanks @taoeffect ,
Implementing zooming in on a certain point was a real challenge here. I can rest in peace now.

PR #2370, opened 2024-10-04
#2369 chat paste bug

### LINKED ISSUE #2369: Send button remains disabled when pasting text in chat on mobile
### Problem

a

On both iOS and Android, pasting text into the chat text field on mobile doesn't make the send button become active. It remains disabled. You have to type something to make it turn blue.

### Solution

Fix.

SebinSong:
closes #2369 

The textarea element needed a 'paste' event handler.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -148,10 +148,10 @@ async function startApp () {
         }, 8e3)
       })]
   ).catch(e => {
-    console.error('[main] Error setting up service worker', e)
-    alert(L('Error while setting up service worker'))
-    window.location.reload() // try again, sometimes it fixes it
-    throw e
+    // console.error('[main] Error setting up service worker', e)
+    // alert(L('Error while setting up service worker'))
+    // window.location.reload() // try again, sometimes it fixes it
+    // throw e
   })
 
   /* eslint-disable no-new */
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -148,10 +148,10 @@ async function startApp () {
         }, 8e3)
       })]
   ).catch(e => {
-    // console.error('[main] Error setting up service worker', e)
-    // alert(L('Error while setting up service worker'))
-    // window.location.reload() // try again, sometimes it fixes it
-    // throw e
+    console.error('[main] Error setting up service worker', e)
+    alert(L('Error while setting up service worker'))
+    window.location.reload() // try again, sometimes it fixes it
+    throw e
   })
 
   /* eslint-disable no-new */
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -555,7 +555,7 @@ export default ({
       }
     },
     handlePaste (e) {
-      // fix for the edge-case related to 'paste' action when nothing has been typed.
+      // fix for the edge-case related to 'paste' action when nothing has been typed
       // (reference: https://github.com/okTurtles/group-income/issues/2369)
       const currVal = this.$refs.textarea.value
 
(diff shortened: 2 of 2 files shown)


corrideat:
This works! However, I wonder if using the `input` event might be a more generic approach. This seems to work for me: `@input='()=>this.updateTextArea()'`.

SebinSong:
@corrideat
If you look more closely in `SendArea.vue`, capturing what user has typed are all done via either calling `this.$refs.textarea.value` or through `keydown` event handler (which is already calling `this.updateTextarea`). so your solution will lead to calling it twice unnecessarily.

(Don't know if you write Vue code too but) It's more typical/generic approach to use `v-model` (which inherently uses `input` event) in a vue component.

But I'm sure the front-end author(s) of this component didn't use neither of them for a reason:
`SendArea.vue` currently has various UI/UX related logics triggered by user typing action (eg. auto-sizing the textarea element height if the content is multiple lines etc..) and they decided that using `keydown` event is more suitable to account for all of them.

So, adding `input` event handler is not only duplicate but also will very likely lead to creating conflicts with these existing logics. dismissing the review.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -553,6 +554,11 @@ export default ({
         this.updateMentionKeyword()
       }
     },
+    handlePaste (e) {
+      const pastedText = e.clipboardData.getData('text')
+      this.$refs.textarea.value = pastedText
```
taoeffect: This will replace all existing text with the pasted text, overwriting the value.

Instead of that, we want the pasted value to be inserted where the cursor is like normal.

taoeffect (changes requested):
Thanks @SebinSong! There's a critical bug here:

frontend/views/containers/chatroom/SendArea.vue
corrideat: My suggestion of using the `input` event doesn't need this `handlePaste` handler (it needs a much simpler handler, as shown in my example). However, I didn't thoroughly test it.

>  and they decided that using keydown event is more suitable to account for all of them.

I did see the duplicate logic for triggering the update (actually, it could _possibly_ be duplicate, if you paste without using keys it won't be a duplicate), however it doesn't seem like a big issue considering that it's not expensive to call it.

As for conflicts, I didn't really see anything that would conflict.

SebinSong:
@taoeffect Updated the fix again!

frontend/views/containers/chatroom/SendArea.vue
SebinSong: It does not make sense to add `input` when there is `keyodwn` already in place. It has to be either one of them. They are two events that look similar but behave differently in certain cases in my experience. For that reason it's a type of change I would be very cautious to do at this point (Considering this code has been this way for a long while). Also, as I said above, I think other front-end devs chose to use `keydown` instead of `input` for a reason.

frontend/main.js
```diff
@@ -148,10 +148,10 @@ async function startApp () {
...
       })]
   ).catch(e => {
-    console.error('[main] Error setting up service worker', e)
-    alert(L('Error while setting up service worker'))
-    window.location.reload() // try again, sometimes it fixes it
-    throw e
+    // console.error('[main] Error setting up service worker', e)
+    // alert(L('Error while setting up service worker'))
+    // window.location.reload() // try again, sometimes it fixes it
+    // throw e
```
taoeffect: Why have these been commented out?

taoeffect (changes requested):

frontend/main.js
SebinSong: forgot to uncomment it again after development. Uncommented them.

FYI, I commented out because this block leads to `reload -> error -> reload ...` infinite cycle, when using below link on my phone for investigation.

taoeffect (dismissed):
Nice work @SebinSong! 👏

taoeffect (approved):

PR #2375, opened 2024-10-12
Remember whether the welcome screen was seen

corrideat:


frontend/views/components/GroupWelcome.vue
```diff
@@ -52,6 +54,14 @@ export default ({
     toDashboard () {
       if (this.isButtonClicked) return
       this.isButtonClicked = true
+      const groupContractID = this.currentGroupId
```
corrideat, line 57: I'm wondering whether this should be here, or if instead it should be done on the `/dashboard` page on `onMounted`. Maybe it doesn't matter.

corrideat: Pros of doing it here: it's only done once (for users). Pros of doing it in the `/dashboard` page: we don't need to change the `commands.js` in the tests. Cons of doing it in the `/dashboard` page: need to do this check every time.

frontend/model/contracts/identity.js
```diff
@@ -283,8 +283,7 @@ sbp('chelonia/defineContract', {
           throw new Error(`Cannot leave group ${groupContractID} because the reference hash does not match the latest`)
         }
 
-        state.groups[groupContractID].hasLeft = true
-        delete state.groups[groupContractID].inviteSecret
+        state.groups[groupContractID] = { hash: reference, hasLeft: true }
```
taoeffect, line 292: There might be other attributes on `groups` that you're overwriting here that shouldn't be overwritten.

taoeffect (changes requested):
Nice work @corrideat! Review ready!

frontend/model/contracts/identity.js
corrideat: There aren't. This is when leaving a group, so we should keep the attributes there at a minimum.

taoeffect: It's true that there aren't at the moment, but in the future if more attributes are added then developers must remember to update this code to make sure nothing breaks. Whereas the way it worked before didn't need as much attention.

corrideat: > but in the future if more attributes are added then developers must remember to update this code to make sure nothing breaks

Nothing should break because these are chatroom that we're not part of. Note that attributes are _already_ overwritten when _joining_ a chatroom. The only reason we keep these values around at all (before we deleted the whole thing) is for accounting purposes, so that re-joining works correctly.

> then developers must remember to update

I think this is a good thing. You shouldn't need attributes from chatroom you're not a part of. The only exception I can think of is for a future feature that allows you to 'block' a chatroom (to prevent re-joining), in which case it makes sense to be explicit.

corrideat: Either way, making things as they were before still means that developers need to pay attention to this. If a `foo` attribute were to be added, which is not used after having left, it needs to have a corresponding `delete` here.

taoeffect: ok that last comment makes sense... can you add a comment here explaining that?

frontend/views/pages/Home.vue
```diff
@@ -91,10 +91,12 @@ export default ({
...
   computed: {
     ...mapGetters([
-      'ourProfileActive'
+      'ourProfileActive',
+      'currentIdentityState'
     ]),
     ...mapState([
-      'currentGroupId'
+      'currentGroupId',
+      ''
```
taoeffect: This `''` seems like a bug

test/cypress/support/commands.js
```diff
@@ -347,7 +354,14 @@ Cypress.Commands.add('giCreateGroup', (name, {
...
         if (router.history.current.path === '/dashboard') return
-        return router.push({ path: '/dashboard' })
+        const state = sbp('state/vuex/state')
+        return sbp('gi.actions/identity/setGroupAttributes', {
+          contractID: state.loggedIn.identityContractID,
+          data: {
+            groupContractID: state.currentGroupId,
+            attributes: { seenWelcomeScreen: true }
+          }
+        }).then(() => router.push({ path: '/dashboard' }))
```
taoeffect: Is there a way this can be DRY'd? (Repeated above)

corrideat: I'm not entirely sure. It could maybe be a Cypress command.

taoeffect (changes requested):
Review ready!

frontend/controller/router.js
```diff
@@ -33,7 +33,14 @@ Vue.use(Router)
  */
 const homeGuard = {
   guard: (to, from) => !!store.state.currentGroupId,
-  redirect: (to, from) => ({ path: store.getters.ourProfileActive ? '/dashboard' : '/pending-approval' })
+  redirect: (to, from) => ({
+    path:
+      // If we haven't accepted the invite OR we haven't clicked 'Awesome' on
+      // the welcome screen, redirect to the '/pending-approval' page
+      (store.getters.ourProfileActive && store.getters.currentIdentityState?.groups?.[store.state.currentGroupId]?.seenWelcomeScreen)
```
taoeffect: Could you simplify this last part please? You're already using getters, it should be fairly easy to add a new one so that this just reads `store.getters.seenWelcomeScreen`

frontend/views/pages/Home.vue
```diff
@@ -155,7 +156,9 @@ export default ({
   },
   watch: {
     currentGroupId (to) {
-      this.ephemeral.ourProfileActive = this.ourProfileActive
+      // Redirect to `/pending-approval` if our profile isn't active or the
+      // welcome screen hasn't been approved
+      this.ephemeral.ourProfileActive = this.ourProfileActive && this.currentIdentityState?.groups?.[this.currentGroupId]?.seenWelcomeScreen
```
taoeffect: Then `this.seenWelcomeScreen` can be used here too

taoeffect (changes requested):
Should be final set of change requests

taoeffect (approved):
Nice!

PR #2377, opened 2024-10-14
#2342 - `bypassUI: true` option in `group-large.spec.js` and etc.

### LINKED ISSUE #2342: add `bypassUI: true` option for `cy.giAddRandomIncome()` in `group-large.spec.js`
### Problem

in `group-large.spec.js`, there is a test step where a group with a large number of members (currently **12**) is created. This test involves a repetitive step where many members add their income-details (which is `cy.giAddRandomIncome()`) but this is currently done in a manual way (by launching the `IncomeDetails` modal, typing the info, and clicking submit button etc..)
By implementing `bypassUI: true` option for this command, the test duration can be reduced.

### Solution
Implement it

SebinSong:
closes #2342 

Currently `group-large.spec.js` test-suite has relatively high number of repeated action such as, 'A group with **12 members** shows correctly the pledging month overview widget' etc. The goal of this PR is to reduce the duration of this test by implementing `bypassUI: true` options for those repeated actions, which are

`cy.giAddRandomIncome()`
`cy.giLogout()`

This lead to below improvement when tested locally on my Mac OS.

**[Tested in `master`]**

before-update

<br><br>

**[After update]**

after-update

<br><br>
while at it, `cy.giLogout({ bypassUI: true })` has been added in other `*.spec.js` files.


SebinSong:
it appears that adding `cy.giLoyout({ bypassUI: true })` to various test-suites leads to reducing the total test duration by around over `15 secs` which is not bad.

**[Before]**

logout-before

<br><br>

**[After]**

logout-after

<br>

Another attempt:

test/cypress/integration/group-large.spec.js
```diff
@@ -16,18 +16,15 @@ describe('Large group', () => {
...
   })
 
   it(`A group with ${groupLength} members shows correctly the pledging month overview widget`, () => {
     const indexes = Array.from({ length: groupLength - 1 }, (_, i) => i + 2)
     const usernames = indexes.map(i => `user${i}-${userId}`)
     const actionsBeforeLogout = indexes.map(i => {
       return () => {
-        if (i > 3) {
-          cy.getByDT('contributionsLink').click()
-        }
```
taoeffect, line 29: Why was this logic removed? 🤔

taoeffect (commented):
Very cool, nicely done @SebinSong!

And I see you remembered to save at least some `cy.giLogout()` tests - nice!

One question...

test/cypress/integration/group-large.spec.js
SebinSong: Now that we use `cy.giAddRandomIncome({ bypassUI: true })` right below it, it's not necessary to navigate to the `/contributions` page at this point in the flow.

taoeffect (approved):
Great work @SebinSong! 👏

PR #2383, opened 2024-10-20
#2188 - Fix payment table UI bug that happens in certain desktop viewport range

### LINKED ISSUE #2188: [UI Bug] Menu button overflows in the payment page.
### Problem
Found below UI bug while running cypress test. (caused when the username is very long)

image

### Solution
Fix it.

SebinSong:
closes #2188 

Apparently 'Showing ellipsis when the user name is too long' has been resolved since the issue was created, but there is still an outstanding bug where some table header elements (`<th>..</th>`) become double lines with not enough margins in between them. 

Made a fix where it makes sure the headers remain single-line in narrow screen and add horizontal scroll-bar to prevent overflowing elements from happening. Hope it sounds good.

**[Before]**

payment-table-issue

<br>

**[After]**

payment-table-fix

### DIFF
--- a/frontend/views/containers/payments/PaymentsList.vue
+++ b/frontend/views/containers/payments/PaymentsList.vue
@@ -149,6 +149,34 @@ export default ({
     @include phone {
       text-align: right;
     }
+
+    @include desktop {
+      min-width: 4.5rem;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Nice!

PR #2388, opened 2024-10-21
#2373 - Bug: Proposal does not expire

### LINKED ISSUE #2373: Close proposals when they expire
### Problem

On `testing`, proposals are not closed when they expire. This was tested with a proposal that closed more than two months ago.

expiredProposal

Even though the proposal has expired, the "Change vote" link remains, and works.

expiredProposal--canVote

Even though the proposal has expired, the user can change their vote.

expiredProposal--changedVote

### Solution

When proposals expire, they should be closed. This means that:
- on expirey, proposals should automatically close with `STATUS_FAILED` or `STATUS_PASSED`
- proposals with `STATUS_FAILED` or `STATUS_PASSED` should not be able to be voted on by any member

SebinSong:
closes #2373 

I can see the app now handles proposal expiration correctly.

<br>

---

It appears that `gi.actions/group/markProposalsExpired` used to only take one proposal ID as the param(`data.proposalHash`) and then was changed to take an array of IDs(`data.proposalIds`) later. But the relevant part in `frontend/controller/actions/group.js` file has not been updated to take this into account, which lead to causing below runtime error:

<br>

So made a fix accordingly.

### DIFF
--- a/frontend/controller/actions/group.js
+++ b/frontend/controller/actions/group.js
@@ -1106,13 +1106,15 @@ export default (sbp('sbp/selectors/register', {
   ...encryptedAction('gi.actions/group/markProposalsExpired', L('Failed to mark proposals expired.'), async function (sendMessage, params) {
     const { contractID, data } = params
     const state = await sbp('chelonia/contract/state', contractID)
-    const proposal = state.proposals[data.proposalHash]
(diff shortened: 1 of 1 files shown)


taoeffect:
Note to self when testing this: be sure to test with existing data from server to verify that old messages won't be broken by this PR.

taoeffect (approved):
LGTM

Issue #2386, opened 2024-10-21
Direct Message notifications broken on `master`


taoeffect:
### Problem

It seems somehow direct message notifications are broken on `master` now.

- They do not result in a sound
- They do not show the unread red badge number
- They do not send a browser notification either

Steps to reproduce:

1. [Container Tab 1] u1 creates group, copies invite link
2. [Container Tab 2] u2 joins group
3. [Container Tab 2] u2 sends u1 a DM

There should be a DM sound and the red badge should appear showing a `1` next to `u2`'s name in the chat view showing the DM list.

### Solution

1. Fix the bug.
2. Add a single test with this scenario to prevent us from missing it again.

taoeffect:
Closed by @SebinSong in #2389

PR #2389, opened 2024-10-22
Fix the broken DM notification

SebinSong:
resolve **solution 1.** of #2386 

I can see the notification works properly now.

notification works

### DIFF
--- a/frontend/controller/actions/chatroom.js
+++ b/frontend/controller/actions/chatroom.js
@@ -44,7 +44,7 @@ sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
     messageHash: msgData.hash,
     height: msgData.height,
     text: msgData.text,
-    isDMOrMention: isMentionedMe || targetChatroomState.attributes.type === CHATROOM_TYPES.DIRECT_MESSAGE,
+    isDMOrMention: isMentionedMe || targetChatroomState?.attributes.type === CHATROOM_TYPES.DIRECT_MESSAGE,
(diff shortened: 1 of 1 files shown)


frontend/controller/actions/chatroom.js
```diff
@@ -42,7 +44,7 @@ sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
     messageHash: msgData.hash,
     height: msgData.height,
     text: msgData.text,
-    isDMOrMention: isMentionedMe || getters.chatRoomAttributes.type === CHATROOM_TYPES.DIRECT_MESSAGE,
+    isDMOrMention: isMentionedMe || targetChatroomState.attributes.type === CHATROOM_TYPES.DIRECT_MESSAGE,
```
SebinSong: This was the cause of the issue.
`getters.chatRoomAttributes` getter here is for a currently active chatroom which means the DM notification only works when the receivers of the DM have the chatroom open in their browsers and this is what I observed too.

Made a fix here where it checks if the target chatroom is DM or not, regardless of whether the chatroom is active or not.

SebinSong:
@taoeffect I realized that there is a requirement to update the DM test-suite for this bug too. Can I maybe work on it as a separate PR later?

SebinSong:
@taoeffect The PR is ready for review.

I can work on updating(or creating) a test item to check for this bug separately from this PR later.

taoeffect (approved):
Seems to have done it!

PR #2394, opened 2024-10-27
#2392 - No need to send a notification when the chatroom is open

### LINKED ISSUE #2392: Notifications displayed when someone sends you a DM while you have the DM open
### Problem

So when using the `ssh -R 80:localhost:8000 nokey@localhost.run` tunnel with `grunt dev` to test on mobile, I opened up a DM with myself from Desktop to Mobile, and any time I sent a DM to the phone while I had the phone open and showing the DM page, I would get a notification and the phone would buzz:

tmp-1729712704525

### Solution

We already have code not to make a sound when the DM is active. We should likewise not post a notification in this instance.

SebinSong:
closes #2392 

It appears the fix works well when tested with multiple containers of Firefox Dev Edition.

### DIFF
--- a/frontend/controller/actions/chatroom.js
+++ b/frontend/controller/actions/chatroom.js
@@ -24,6 +24,9 @@ sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
   const getters = sbp('state/vuex/getters')
   const rootState = sbp('state/vuex/state')
   const targetChatroomState = rootState[contractID]
+  const currentRoute = sbp('controller/router').history.current || ''
+  const isTargetChatroomCurrentlyActive = currentRoute.path.includes('/group-chat') &&
+    getters.currentChatRoomId === contractID
   const mentions = makeMentionFromUserID(getters.ourIdentityContractId)
   const msgData = newMessage || data
   const isMentionedMe = (!!newMessage || data.type === MESSAGE_TYPES.TEXT) && msgData.text &&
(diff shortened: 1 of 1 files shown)


frontend/controller/actions/chatroom.js
```diff
@@ -24,6 +24,9 @@ sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
   const getters = sbp('state/vuex/getters')
   const rootState = sbp('state/vuex/state')
   const targetChatroomState = rootState[contractID]
+  const currentRoute = sbp('controller/router').history.current || ''
+  const isTargetChatroomCurrentlyActive = currentRoute.path.includes('/group-chat') &&
+    getters.currentChatRoomId === contractID // when the target chatroom is currently open/active on the browser, No need to send a notification.
```
taoeffect, line 29: There are two potential issues we need to consider here:

1. Should we send a notification anyway when the user is idle but on the page? (We currently use `idleVue` to check for this, search the project for it)
2. Are we able to, when using the service worker approach, tell what page we're on in the actions (and whether we're idle or not)? This question is for @corrideat

taoeffect: I'm going to approve this after creating an issue to deal with this in a different manner later.

taoeffect (approved):
LGTM!

Approving this and to deal with the service-worker and `idleVue` considerations I created issue #2397 and assigned it to @corrideat!

taoeffect (commented):
Ooops... I think I approved too quickly.

`messageReceivePostEffect` is supposed to always be called, because it calls `'gi.actions/identity/kv/addChatRoomUnreadMessage'` which is always supposed to be called (as it's not notification related).

PR #2400, opened 2024-10-28
#2387 Two emoji picker issues

### LINKED ISSUE #2387: Two Emoji picker issues
### Problem

1. When bringing up the emoji picker, the search field in the emoji picker should be immediately selected
2. When clicking on an emoji in the emoji picker to insert an emoji into the text, the emoji should be inserted where the cursor was, and not at the very end of the message.

### Solution

Fix.

SebinSong:
closes #2387


taoeffect (approved):
Amazing job @SebinSong!

Tested on both Desktop and Mobile and it worked in both places first go around!

PR #2402, opened 2024-10-29
#2393 - Add `ButtonSubmit.vue` for 'Send message' button

### LINKED ISSUE #2393: Multiple DMs to the same user can be created by accident
### Problem

This can happen especially over a slow connection, like when testing on mobile with the `ssh -R 80:localhost:8000 nokey@localhost.run` service.

Screenshot 2024-10-23 at 12 43 34 PM

What happens is you click on a user's icon and then click the "Send message" button. It takes a while to create the new DM contract, and in that time you can press the button again, multiple times, and each time it's pressed a new DM contract to that same user will be created.

### Solution

Disable this "Send message" button until the DM contract is finished creating.

SebinSong:
closes #2393 

Apparently, the tooltip is dismissed immediately after the 'Send message' button when `this.createDirectMessage(this.contractID)` triggered by it is a series of async operation.
https://github.com/okTurtles/group-income/blob/83f11f1ed18c5d3740088cabd97f86e93ff9617e/frontend/views/components/ProfileCard.vue#L171

So replaced the button with `ButtonSubmit.vue` as the visual indicator so that it gets dismissed after the DM is created.

Hope it sounds good and let me know if there is addition al thing needed along with this.

### DIFF
--- a/frontend/views/components/ProfileCard.vue
+++ b/frontend/views/components/ProfileCard.vue
@@ -63,11 +63,12 @@ tooltip(
         ) Add payment information
 
       .buttons(v-if='!isSelf')
-        i18n.button.is-outlined.is-small(
-          tag='button'
-          @click='sendMessage'
+        button-submit.is-outlined.is-small(
+          type='button'
           data-test='buttonSendMessage'
-        ) Send message
+          @click='sendMessage'
+        )
+          i18n Send message
 
         i18n.button.is-outlined.is-small(
           v-if='groupShouldPropose || isGroupCreator'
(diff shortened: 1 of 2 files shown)


frontend/views/components/ProfileCard.vue
```diff
@@ -178,6 +181,9 @@ export default ({
       }
       this.toggleTooltip()
     }
+  },
+  beforeDestroy () {
+    console.log('!@# destroyed!!')
```
taoeffect: Forgotten debug logging

taoeffect (approved):
Nice! Seems to work well!

PR #2405, opened 2024-10-29
Two bug-fixes

### LINKED ISSUE #2396: Member can leave contributions blank
### Problem

The `IncomeDetails` modal on`testing` accepts ` ` as a valid non-monetary contribution even though space character(s) is not valid for any other required fields. In the screenshot below, this is the blank non-monetary contribution.

On the `Contributions` view, ` ` seems to trigger the proper alerts, as shown in the screenshot, where the user has entered one space char in the field.

contributionSpace

### Solution

When the non-monetary contribution field contains only space characters, show the same alert as would be shown if the field were blank. _Don't_ accept only space chars as a valid contribution or add ` ` to the list of non-monetary contributions.

### LINKED ISSUE #2404: Bug in the text in the "Add new members" modal
### Problem

It says 60 passes, we forgot to update it to say 150 passes:

Screenshot 2024-10-29 at 10 57 40 AM

### Solution

Update it to use the `MAX_GROUP_MEMBER_COUNT` variable here.

SebinSong:
closes #2396 
closes #2404

### DIFF
--- a/frontend/views/containers/group-settings/InvitationLinkModal.vue
+++ b/frontend/views/containers/group-settings/InvitationLinkModal.vue
@@ -6,7 +6,7 @@ modal-template(ref='modal' :a11yTitle='L("Add new members")')
   .c-container
     template(v-if='link')
       i18n.is-title-4(tag='h3') Share this link to grant access to your group.
-      i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 150 free passes!
+      i18n.has-text-1(tag='p' :args='{ count: anyoneLinkMax }') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy {count} free passes!
       link-to-copy.c-link(:link='link')
       i18n.has-text-1(v-if='expireDate' tag='p' :args='{ expireDate }') This invite link expires on {expireDate}.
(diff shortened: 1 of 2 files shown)


frontend/views/containers/group-settings/InvitationLinkModal.vue
```diff
@@ -6,7 +6,7 @@ modal-template(ref='modal' :a11yTitle='L("Add new members")')
   .c-container
     template(v-if='link')
       i18n.is-title-4(tag='h3') Share this link to grant access to your group.
-      i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 60 free passes!
+      i18n.has-text-1(tag='p') After the onboarding period has ended, everyone will be asked to vote on whether or not a new member should be added. But for now, enjoy 150 free passes!
```
taoeffect: Could you make this a variable so that we don't have to re-translate it next time it changes?

See issue for recommended variable value.

taoeffect (changes requested):
Thx! Minor issue

frontend/views/containers/group-settings/InvitationLinkModal.vue
SebinSong: I jumped too early to making a fix and missed that part in the issue. updated.

taoeffect (approved):
Nice!

PR #2413, opened 2024-11-06
#2412 `<Enter>` key issue on mobile

### LINKED ISSUE #2412: <Enter> / <Return> key should create a new line instead of sending message
### Problem

While on Desktop you can press <kbd>Shift+Enter</kbd> to create a new line, on mobile that doesn't seem possible.

### Solution

On mobile only, the enter key should create a new line. The user should be forced to tap the send button (just like on Slack) to send the message.

SebinSong:
closes #2412 

I can write multi-line messages from mobile now.

### DIFF
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -397,7 +397,7 @@ export default ({
     sbp('okTurtles.events/off', CHATROOM_USER_TYPING, this.onUserTyping)
     sbp('okTurtles.events/off', CHATROOM_USER_STOP_TYPING, this.onUserStopTyping)
 
-    this.mediaIsPhone.onchange = null
+    this.mediaIsPhone.onchange = null // change handler needs to be destoryed to prevent memory leak.
     this.ephemeral.staleObjectURLs.forEach(url => {
       URL.revokeObjectURL(url)
     })
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -65,7 +65,7 @@
       :style='textareaStyles'
       :maxlength='config.messageMaxChar'
       @blur='textAreaBlur'
-      @keydown.enter.exact.prevent='handleKeyDownEnter'
+      @keydown.enter.exact='handleKeyDownEnter'
       @keydown.tab.exact='handleKeyDownTab'
       @keydown.ctrl='isNextLine'
       @keydown='handleKeydown'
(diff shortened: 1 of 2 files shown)


corrideat:
LGTM.

I'm not so happy about the mobile detection, but then I can't really find a better way of doing this. What I'd do is maybe add a setting or a UI component (like a checkbox or dropdown) to explicitly select the <kbd>Enter</kbd> behaviour (tri-state: 'auto', <kbd>⇧ Shift</kbd>+<kbd>Enter</kbd> for new line, <kbd>Enter</kbd> for new line) but that may be beyond the scope of this issue.

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -376,9 +376,9 @@ export default ({
   },
   created () {
     // TODO #492 create a global Vue Responsive just for media queries.
-    const mediaIsPhone = window.matchMedia('(hover: none) and (pointer: coarse)')
-    this.ephemeral.isPhone = mediaIsPhone.matches
-    mediaIsPhone.onchange = (e) => { this.ephemeral.isPhone = e.matches }
+    this.mediaIsPhone = window.matchMedia('(hover: none) and (pointer: coarse)')
```
taoeffect, line 380: This seems to be creating a variable on the Vue object on-the-fly, which, as far as I know, isn't recommended by Vue.js. So correct me if I'm wrong, but I think this `mediaIsPhone` should be defined in the object returned by the `data()` section.

taoeffect (changes requested):
Nice work @SebinSong!

I found one minor concern re Vue.js behavior (see comment), and a bug on mobile: try typing some text, and then pressing <kbd>Enter</kbd> many times on mobile. What will happen is that the text field will get taller up to a point, and the original text you typed will get moved way up. However, you won't be able to scroll the text field to edit the original text you type. On Android at least, whenever I tried to scroll the textfield, it would instead scroll the background chat.

frontend/views/containers/chatroom/SendArea.vue
SebinSong: Fixed it.

SebinSong:
@taoeffect 

>  a bug on mobile: try typing some text, and then pressing Enter many times on mobile. What will happen is that the text field will get taller up to a point, and the original text you typed will get moved way up. However, you won't be able to scroll the text field to edit the original text you type.

I've been trying to figure this out for a while thinking it's a bug too, but it appears that this is a default behaviour by the browsers **on mobile**. Also, **this only happens when the keyboard panel is revealed**. When you click outside the `<textarea />` to fold the keyboard back and try scrolling again it does work.

For example, you can confirm this symptom happens in any `<textarea />` elements in [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) too. (I've tried this on a web-app of my other company in NZ too and confirmed the same there.)

But we can still use below handle that is attached to the cursor to move up/down which allows us to push the text content back down to a position you want. I guess this is how we are supposed to scroll the text content on textarea while the keyboard panel is up on mobile.



Thanks,

taoeffect (approved):
Nice!

PR #2419, opened 2024-11-11
#2391 - Multiple images in the viewer

### LINKED ISSUE #2391: Implement cycling between multiple images in the image viewer
### Problem
As proposed in [this comment](https://github.com/okTurtles/group-income/pull/2367#pullrequestreview-2344456580) of PR #2367 , make it possible to  cycle between photos using the arrow keys on desktop if the message contains multiple photo attachments, similarly to what Slack does for the same case (below).

image

### Solution
Implement it.

SebinSong:
closes #2391 

**[Desktop]**
It also allows navigation via arrows in the keyboard too.

image

**[Mobile]**

### DIFF
--- a/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
+++ b/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
@@ -213,6 +213,18 @@ export default {
       }
     },
     openImageViewer (objectURL, data) {
+      const allImageAttachments = this.attachmentList.filter(entry => this.fileType(entry) === 'image')
+        .map((entry, index) => {
+          return {
+            name: entry.name,
+            ownerID: this.ownerID,
+            imgUrl: entry.url || this.objectURLList[index] || '',
+            createdAt: this.createdAt || new Date()
+          }
+        })
+
+      console.log('!@# allImageAttachments: ', allImageAttachments)
+
       if (objectURL) {
         sbp(
           'okTurtles.events/emit', OPEN_MODAL, 'ImageViewerModal',
--- a/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
+++ b/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
@@ -94,6 +94,7 @@ import { MESSAGE_VARIANTS } from '@model/contracts/shared/constants.js'
 import { getFileExtension, getFileType } from '@view-utils/filters.js'
 import { Secret } from '~/shared/domains/chelonia/Secret.js'
 import { OPEN_MODAL } from '@utils/events.js'
+import { randomHexString } from '@model/contracts/shared/giLodash.js'
 
 export default {
   name: 'ChatAttachmentPreview',
@@ -221,7 +222,8 @@ export default {
             name: entry.name,
             ownerID: this.ownerID,
             imgUrl: entry.url || this.objectURLList[index] || '',
-            createdAt: this.createdAt || new Date()
+            createdAt: this.createdAt || new Date(),
+            id: randomHexString(12)
           }
         })
       const initialIndex = allImageAttachments.findIndex(attachment => attachment.imgUrl === objectURL) || 0
--- a/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
+++ b/frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
@@ -226,14 +226,14 @@ export default {
             id: randomHexString(12)
           }
         })
-      const initialIndex = allImageAttachments.findIndex(attachment => attachment.imgUrl === objectURL) || 0
+      const initialIndex = allImageAttachments.findIndex(attachment => attachment.imgUrl === objectURL)
 
       sbp(
         'okTurtles.events/emit', OPEN_MODAL, 'ImageViewerModal',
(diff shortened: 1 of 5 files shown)


frontend/views/containers/chatroom/image-viewer/PreviewImageArea.vue
```diff
@@ -5,7 +5,7 @@
   @mouseup='mouseUpHandler'
 )
   img.c-preview-image(ref='previewImg'
-    :class='{ "is-movable": isImageMovable }'
+    :class='{ "is-movable": isImageMovable, "is-hidden": !ephemeral.imgInitDone }'
```
SebinSong, line 8: Discovered a bug where the image kind of flickers  when it finishes the load/initialization steps(happens because image dimensions changes from it's natural size to the initial zoom value). So made a fix here.

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
```diff
@@ -212,21 +213,29 @@ export default {
...
+      if (!objectURL) { return }
+
+      const allImageAttachments = this.attachmentList.filter(entry => this.fileType(entry) === 'image')
+        .map((entry, index) => {
+          return {
+            name: entry.name,
+            ownerID: this.ownerID,
+            imgUrl: entry.url || this.objectURLList[index] || '',
+            createdAt: this.createdAt || new Date(),
+            id: randomHexString(12)
```
corrideat, line 226: Instead of a random ID, wouldn't it be better to use a deterministic ID in this case? For example, I think the URLs would be both deterministic and unique (or it could be combined with the message ID).

SebinSong: Valid suggestion. But I think it's safe to use the random string here because the `id` here is merely used for differentiating multiple images rendered in the DOM.

corrideat: Yeah, I think it's safe also as used here, but I'm partial to deterministic IDs because they're always safe and can also be used, e.g., in testing.

frontend/views/containers/chatroom/image-viewer/PreviewImageArea.vue
taoeffect: Awesome! I noticed this behavior too but thought it was Android's fault.

frontend/views/containers/chatroom/image-viewer/ImageViewerModal.vue
```diff
@@ -73,36 +80,81 @@ export default {
...
   },
   created () {
-    if (!this.metaData) {
+    if (!Array.isArray(this.images)) {
       this.$nextTick(() => this.close())
+    } else {
+      this.ephemeral.currentIndex = this.initialIndex
+    }
+
+    this.touchMatchMedia = window.matchMedia('(hover: none) and (pointer: coarse)')
```
taoeffect, line 114: I think this is the same issue as on the other PR with not having this in `data ()`...

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
```diff
@@ -212,21 +213,29 @@ export default {
...
           }
-        )
-      }
+        })
+      const initialIndex = allImageAttachments.findIndex(attachment => attachment.imgUrl === objectURL) || 0
```
taoeffect: `findIndex` returns `-1` when an image isn't found, not falsy. This `|| 0` wouldn't get triggered on `-1`.

frontend/views/utils/pointerEventsMixins.js
```diff
@@ -100,16 +100,16 @@ const pointerEventsMixinFactory = (opts: any = mixinGeneratorDefaultOpts): any =
...
+          // leading to an abrupt change in image scale. so it needs to ignore this value.
+          const changeFactorTooLarge = currPinchAction === this.pointer.prevPinchAction &&
             this.pointer.prevDistChangeFactor > 0 &&
-            (distChangeFactorCalc > this.pointer.prevDistChangeFactor * 3)
-            ? this.pointer.prevDistChangeFactor
-            : distChangeFactorCalc
+            distChangeFactorCalc > this.pointer.prevDistChangeFactor * 3
 
-          const args = { changeFactor: changeFactorToUse, center }
+          if (changeFactorTooLarge) { return }
```
taoeffect, line 108: For some reason Claude thinks line 128 below is an issue and says:

> Failure to Update State on Large Change Factor

I'm not sure what to make of this comment. Feel free to ignore if nothing.

taoeffect (changes requested):
Review ready!

frontend/views/utils/pointerEventsMixins.js
SebinSong: Yes, this line was added intentionally (to remove a sudden jump in zoom value when pinch action was performed quickly and stopped) and added a `NOTE:` comment for this as well in line `L102`

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
SebinSong: Good catch. Fixed.

taoeffect (approved):
Nice!

PR #2420, opened 2024-11-11
#2418 - Show error UI for mincome maximum

### LINKED ISSUE #2418: Show applicable mincome limits error on mincome step in group creation
### Problem

When creating a group, the user can enter a mincome that's larger than the app can handle. The error isn't shown on the "3. Minimum Income" step. Instead, the user sees a cryptic error on the unrelated "4. Voting system" step, and only after clicking "Create Group". 

creation-mincomeLimit

### Solution

Show a user-friendly error on the "3. Mincome Income" step, where the user can easily understand and correct the error.

SebinSong:
closes #2418 

Please let me know if we want to change the error sentence.

### DIFF
--- a/frontend/views/containers/group-settings/GroupCreationModal.vue
+++ b/frontend/views/containers/group-settings/GroupCreationModal.vue
@@ -56,7 +56,8 @@ import proposals from '@model/contracts/shared/voting/proposals.js'
 import {
   PROPOSAL_GENERIC,
   GROUP_NAME_MAX_CHAR,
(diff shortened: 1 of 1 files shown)


corrideat:
LGTM in terms of fixing the issue, although not a fan of adding arbitrary limits like these (OTOH, the limit needs to exist regardless, even if it were the full range of numbers allowed by JS).

corrideat (approved):

taoeffect (approved):
LGTM

PR #2421, opened 2024-11-13
#2414 - "See proposal" link should always lead to proposal 

### LINKED ISSUE #2414: "See proposal" link should always lead to proposal
### Problem

On `testing`, once a proposal is archived, the "See proposal" link no longer directly links to the proposal. The link leads to the dashboard, but the proposal isn't immediately visible because it has been archived. 

https://github.com/user-attachments/assets/0921256b-8da3-430f-a030-f846ecf56495

### Solution

The "See proposal" link should always send the user directly to the view where they can see the proposal in question.

SebinSong:
closes #2414 

I realised there is a same issue in this in-app notification UI too, so made a fix there as well.

### DIFF
--- a/frontend/views/containers/proposals/PropositionsAllModal.vue
+++ b/frontend/views/containers/proposals/PropositionsAllModal.vue
@@ -66,11 +66,11 @@ export default ({
       const targetId = this.$route.query?.targetProposal || ''
 
       if (targetId && this.ephemeral.proposals.some(entry => entry[0] === targetId)) {
-        setTimeout(() => {
+        this.$nextTick(() => {
           const targetEl = this.$refs.pList.querySelector(`[data-proposal-hash="${targetId}"]`)
 
           targetEl && targetEl.scrollIntoView({ block: 'center' })
-        }, 150)
+        })
       }
     }
   },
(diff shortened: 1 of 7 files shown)


frontend/views/containers/proposals/PropositionsAllModal.vue
```diff
@@ -55,10 +55,23 @@ export default ({
...
+    },
+    checkTargetAndScroll () {
+      const targetId = this.$route.query?.targetProposal || ''
+
+      if (targetId && this.ephemeral.proposals.some(entry => entry[0] === targetId)) {
+        setTimeout(() => {
+          const targetEl = this.$refs.pList.querySelector(`[data-proposal-hash="${targetId}"]`)
+
+          targetEl && targetEl.scrollIntoView({ block: 'center' })
+        }, 150)
```
corrideat: Is there a way to do this without a timeout?

SebinSong: The delay is actually not necessary here. So replaced it with `nextTick()` instead.

taoeffect (approved):
Fantastic work on this as well @SebinSong!!

PR #2428, opened 2024-11-19
#2411 - Compress large images attached in the chat

### LINKED ISSUE #2411: Compress images before uploading them
### Problem

Images taken by modern phones can be huge, ranging from 5MB to 10MB or even more.

This causes these problems:

- Server running out of space
- Images taking a very long time to load on slower connections

There is no reason for us to upload these massive images in their entirety to the server when they're attached.

### Solution

Similar to how Signal works when uploading an image, find a way to compress the image by reducing its size so that it remains under a specific file size while keeping the quality of the image high.

For example, you could compress images so that they're no larger than 400KB.

SebinSong:
closes #2411

<br>

---
**[NOTE]**

- As @taoeffect suggested in the issue, `400KB` will be the size value to determine compression is needed.

- The app uses HTML canvas API ([toBlob()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob), [drawImage()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage)) to compress the image on the browser, which is a popular client-side image-compression [npm package](https://www.npmjs.com/package/compressorjs) internally uses as well.

- The compression outcome will be one of two most popular images formats on the web that support lossy-compression (`image/webp` or `image/jpeg`). According to tests I've done (pls feel free to check out [this codepen](https://codepen.io/freenaturalsoul/pen/vYoMWWv?editors=0100) I created for testing for yourself too), **converting to 'webp' format apparently has a better compression efficiency than 'jpeg'** (meaning it drops the file size below **400KB** without sacrificing too much quality). So if the browser supports `webp` format, the outcome will always be `webp` and otherwise `jpeg`.

### DIFF
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -587,13 +587,13 @@ export default (sbp('sbp/selectors/register', {
     const { identityContractID } = sbp('state/vuex/state').loggedIn
     try {
       const attachmentsData = await Promise.all(attachments.map(async (attachment) => {
-        const { url, needsIamgeCompression } = attachment
+        const { url, needsImageCompression } = attachment
         // url here is an instance of URL.createObjectURL(), which needs to be converted to a 'Blob'
-        const attachmentBlob = needsIamgeCompression
+        const attachmentBlob = needsImageCompression
           ? await compressImage(url)
           : await objectURLtoBlob(url)
 
-        if (needsIamgeCompression) {
+        if (needsImageCompression) {
           // Update the attachment details to reflect the compressed image.
           const fileNameWithoutExtension = attachment.name.split('.').slice(0, -1).join('.')
           const extension = attachmentBlob.type.split('/')[1]
@@ -607,7 +607,7 @@ export default (sbp('sbp/selectors/register', {
         }, { billableContractID })
         const { delete: token, download: downloadData } = response
         return {
-          attributes: omit(attachment, ['url', 'needsIamgeCompression']),
+          attributes: omit(attachment, ['url', 'needsImageCompression']),
           downloadData,
           deleteData: { token }
         }
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -9,7 +9,7 @@ import {
 import { cloneDeep, has, omit } from '@model/contracts/shared/giLodash.js'
 import { SETTING_CHELONIA_STATE } from '@model/database.js'
 import sbp from '@sbp/sbp'
-import { imageUpload, objectURLtoBlob } from '@utils/image.js'
+import { imageUpload, objectURLtoBlob, compressImage } from '@utils/image.js'
 import { SETTING_CURRENT_USER } from '~/frontend/model/database.js'
 import { KV_QUEUE, LOGIN, LOGOUT } from '~/frontend/utils/events.js'
 import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
@@ -587,15 +587,27 @@ export default (sbp('sbp/selectors/register', {
     const { identityContractID } = sbp('state/vuex/state').loggedIn
     try {
       const attachmentsData = await Promise.all(attachments.map(async (attachment) => {
-        const { mimeType, url } = attachment
+        const { url, needsIamgeCompression } = attachment
         // url here is an instance of URL.createObjectURL(), which needs to be converted to a 'Blob'
-        const attachmentBlob = await objectURLtoBlob(url)
+        const attachmentBlob = needsIamgeCompression
+          ? await compressImage(url)
+          : await objectURLtoBlob(url)
+
+        if (needsIamgeCompression) {
+          // Update the attachment details to reflect the compressed image.
+          const fileNameWithoutExtension = attachment.name.split('.').slice(0, -1).join('.')
+          const extension = attachmentBlob.type.split('/')[1]
+
+          attachment.mimeType = attachmentBlob.type
+          attachment.name = `${fileNameWithoutExtension}.${extension}`
+        }
         const response = await sbp('chelonia/fileUpload', attachmentBlob, {
-          type: mimeType, cipher: 'aes256gcm'
+          type: attachment.mimeType,
+          cipher: 'aes256gcm'
         }, { billableContractID })
         const { delete: token, download: downloadData } = response
         return {
-          attributes: omit(attachment, ['url']),
+          attributes: omit(attachment, ['url', 'needsIamgeCompression']),
           downloadData,
           deleteData: { token }
         }
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -20,6 +20,7 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
 // TODO: fetch this value from a server API
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
+export const IMAGE_ATTACHMENT_MAX_SIZE = 400000 // 400KB
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -18,9 +18,10 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
 // NOTE: Below value was obtained from the '413 Payload Too Large' server error
 //       meaning if this limit is updated on the server-side, an update is required here too.
 // TODO: fetch this value from a server API
+export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
-export const IMAGE_ATTACHMENT_MAX_SIZE = 400000 // 400KB
+export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/utils/image.js
+++ b/frontend/utils/image.js
@@ -104,7 +104,7 @@ export async function compressImage (imgUrl: string, sourceMimeType?: string): P
     const sizeDiff = blob.size - IMAGE_ATTACHMENT_MAX_SIZE
 
     if (sizeDiff <= 0 || // if the compressed image is already smaller than the max size, return the compressed image.
-      quality <= 0.4) { // Do not sacrifice the image quality too much.
+      quality <= 0.3) { // Do not sacrifice the image quality too much.
       return blob
     } else {
       // if the size difference is greater than 100KB, reduce the next compression factors by 10%, otherwise 5%.
--- a/frontend/utils/image.js
+++ b/frontend/utils/image.js
@@ -35,12 +35,12 @@ export const imageUpload = async (imageFile: File, params: ?Object): Promise<Obj
 export function supportsWebP (): Promise<boolean> {
   // Uses a very small webP image to check if the browser supports 'image/webp' format.
   // (reference: https://developers.google.com/speed/webp/faq#in_your_own_javascript)
-  const verySmallWebP = 'data:image/webp;base64,UklGRhIAAABXRUJQVlA4WAoAAAAQAAAAMwAAQUxQSAwAAAAwAQCdASoEAAQAAVAfCWkAQUwAAAABABgAAgAAAAAABAAAAAAAAAA'
+  const verySmallWebP = 'data:image/webp;base64,UklGRiIAAABXRUJQVlA4IBYAAAAwAQCdASoBAAEADsD+JaQAA3AAAAAA'
   const img = new Image()
 
   return new Promise(resolve => {
-    img.onload = () => resolve(img.height > 0)
-    img.onerror = () => resolve(false)
+    img.onload = () => { resolve(img.height > 0) }
+    img.onerror = (e) => { resolve(false) }
     img.src = verySmallWebP
   })
 }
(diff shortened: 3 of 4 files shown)


frontend/utils/image.js
```diff
@@ -28,3 +29,88 @@ export const imageUpload = async (imageFile: File, params: ?Object): Promise<Obj
...
+    const blob = await generateImageBlobByCanvas({
+      sourceImage,
+      resizingFactor,
+      quality,
+      compressToType
+    })
+    const sizeDiff = blob.size - IMAGE_ATTACHMENT_MAX_SIZE
+
+    if (sizeDiff <= 0 || // if the compressed image is already smaller than the max size, return the compressed image.
+      quality <= 0.3) { // Do not sacrifice the image quality too much.
```
SebinSong, line 119: From the multiple tests I've done, when the quality factor value passed to `canvas.toBlob()` becomes `<= 0.3`, the outcome image starts to lose the quality to an unpleasant level. `0.3` here is a value I chose subjectively, but feel free to try compression test yourself using [this codepen](https://codepen.io/freenaturalsoul/pen/vYoMWWv) I created for this task and let me know if you would like a change.

taoeffect:
> So if the browser supports webp format, the outcome will always be webp and otherwise jpeg.

Question: what browsers don't support webp?

And if there are browsers that we want to support that don't support webp, wouldn't that mean that we shouldn't use it at all (since some users wouldn't be able to view the images)?

frontend/controller/actions/identity.js
```diff
@@ -587,15 +587,27 @@ export default (sbp('sbp/selectors/register', {
     const { identityContractID } = sbp('state/vuex/state').loggedIn
     try {
       const attachmentsData = await Promise.all(attachments.map(async (attachment) => {
-        const { mimeType, url } = attachment
+        const { url, needsIamgeCompression } = attachment
         // url here is an instance of URL.createObjectURL(), which needs to be converted to a 'Blob'
-        const attachmentBlob = await objectURLtoBlob(url)
+        const attachmentBlob = needsIamgeCompression
```
corrideat: typo in `Image`

frontend/utils/image.js
```diff
@@ -28,3 +29,88 @@ export const imageUpload = async (imageFile: File, params: ?Object): Promise<Obj
...
+    canvasEl.toBlob(blob => {
+      resolve(blob)
+    }, compressToType, quality)
+  })
+}
+
+export async function compressImage (imgUrl: string, sourceMimeType?: string): Promise<any> {
+  // Takes a source image url and generate a blob of the compressed image.
+
+  // According to the testing result, 0.8 is a good starting point for both resizingFactor and quality for .jpeg and .webp.
```
corrideat: which testing?

corrideat: IMO, this shouldn't be a 'resizing factor' anyhow, but it should resize images to some arbitrary maximum size. E.g., not the real suggestion, the max size could be 1024×768, and images are scaled not to exceed this. The dimensions would come from how the images are displayed in the chat.

corrideat:
I've seen this new PR and I have a few general comments.

First off, and this is a subjective point, WebP is a Google-controlled format and I wouldn't use it without fallback for this very reason, favouring formats that are real standards like JPEG, JPEG XL, AVIF or HEIC (although HEIC is patent-encumbered and therefore might not be a good choice) (*). Furthermore, JPEG is more widely supported than WebP. If there's a potential for users to want to save the images, I'd stick to JPEG because anecdotally users aren't familiar with WebP and may not know what to do with them or may not be as widely supported. For example, on Windows JPEG opens by default on Paint and WebP opens by default on Edge.

Secondly, although WebP sometimes (often even) has superior compression than JPEG, I suspect most of the impressive gains here aren't because of WebP but rather because of re-encoding the image. Although I haven't tried this, I strongly suspect that doing the same thing but with JPEG will yield similar results (albeit the JPEG files will be on average slightly larger, slightly meaning about 30% larger, more or less).

Thirdly, I would avoid doing these operations without providing a way to 'opt out' because re-encoding images is a lossy process that may be unwanted. WhatsApp, for example, compresses images to sizes that display well in mobile, but provides a way to opt out by sending the image as a 'file'.

Fourthly, responding to Greg:

> Question: what browsers don't support webp?

All major browsers (Chrome, Edge, Firefox, Safari) and their derivatives support WebP by default, although this can be configured.

> And if there are browsers that we want to support that don't support webp, wouldn't that mean that we shouldn't use it at all (since some users wouldn't be able to view the images)?

This is a good point. I think the testing for WebP support is misplaced because, while it may be necessary, it can't test the viewing side.

Based on this, I'd suggest the following changes:

1. Dropping WebP (admittedly, this point has subjective reasons)
2. Dropping the resizing factor as well in favour of a max size that displays well in the range of devices GI is expected to run. See https://gs.statcounter.com/screen-resolution-stats/ for common screen sizes.
3. Provide some way to opt out of post processing images when it'd be undesirable, since this reduces quality and destroys metadata.

(*) Of these, JPEG and AVIF are the only realistic choices (other than WebP) in terms of widespread support. AVIF should compress as well as WebP on average, but efficient AVIF encoding is _extremely_ slow from my experience. Moreover, AVIF has the same 'format weirdness' that WebP has, as do all of the newer formats, so in reality it's JPEG or WebP.

corrideat:
And a last comment:

The 'long time to load' issue can be addressed by resizing images to different sizes, so that devices with small screens (which are also more likely to have a slower connection) load images faster. However, this would have the effect of requiring _more_ server storage than a one-size-fits-all solution.

frontend/utils/image.js
SebinSong: Just like I wrote in the desc, check this codepen I created for this task : https://codepen.io/freenaturalsoul/pen/vYoMWWv?editors=0100
I tried compressiing various large size images there. try it there.

> this shouldn't be a 'resizing factor' anyhow, but it should resize images to some arbitrary maximum size.

It doesn't seem like you are suggesting this without knowing what the goal of the issue..? The goal of the issue Greg wrote is to reduce the image **file-size** down below a certain value (400KB). Sacrificing the quality and the size here is inevitable, at least when using `Canvas.toBlob()`. 

According to tests I've done, only changing the `quality` argument of that canvas API without reducing the image dimension along with it have lead to the outcomes getting blurry in earlier stages. That's why I chose an approach to decrease the quality factor and the dimension at the same time. Plus, that is something [this article](https://img.ly/blog/how-to-compress-an-image-before-uploading-it-in-javascript/) suggested too.

SebinSong:
@taoeffect 

> Question: what browsers don't support webp?

That idea was based on this can-I-use result:
https://caniuse.com/webp

I saw `96.81%` there (Also, I myself has not been used this image format commonly in my life either) and so I went ahead to search to see if the world has a javascript logic to check the browser support for it. and then I ran into this article written by Google: https://developers.google.com/speed/webp/faq#in_your_own_javascript

The same article has [Which web browsers natively support WebP?](https://developers.google.com/speed/webp/faq#which_web_browsers_natively_support_webp) section there too and it appears the support level is quite good though.
So I can remove the logic to check the support if you think it's not necessary.

SebinSong:
@corrideat 

> First off, and this is a subjective point, WebP is a Google-controlled format and I wouldn't use it without fallback.

This is a subjective opinion. So not taking it unless Greg wants to drop it.


> 2. Dropping the resizing factor

Like I explained in a comment above, reducing the `quality` argument to `canvas.toBlob` and image dimension at the same time have made the image outcome less blurry, when I tested (via [this codepen](https://codepen.io/freenaturalsoul/pen/vYoMWWv) I created. Actually countless times with various different image formats.)
Also, this is an approach [this article](https://img.ly/blog/how-to-compress-an-image-before-uploading-it-in-javascript/) suggests too.


> Although **I haven't tried this**, I strongly suspect that doing the same thing but with JPEG will yield similar results.

Try using [canvas.toBlob(imageEl, type, quality)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) with specifying `type` to either `image/jpeg` or `image/webp` and then compare the results. you will see that passing `image/webp` achieves similar level of file-size compression with higher value of `quality` argument.


> 3. Provide some way to opt out of post processing images when it'd be undesirable.

I think this is something out of scope of the issue #2411 . You can discuss with Greg after this PR about adding this feature to the app and create an issue and work on it yourself.

frontend/controller/actions/identity.js
SebinSong: Good catch. fixed it.

corrideat:
> Like I explained in a comment above, reducing the quality argument to canvas.toBlob and image dimension at the same time have made the image outcome less blurry,

I didn't say not to drop the image size. What I said, however, is resizing the image to a fixed size. The reason for this is that most of the image size savings come from reducing the image resolution regardless of image format, it is more consistent and the image will appear complete. For example, if I upload a 40000px by 30000px image, unless I really intend the image to be displayed in its entirety (see point 3 about opting out), likely it'll display well at 2000px by 1500px in any reasonable screen that people use. This is what chat apps already do.

> will see that passing image/webp achieves similar level of file-size compression with higher value of quality argument.

Yes, that is expected. WebP is supposed to compress better (I have some counterexamples where it doesn't), which ultimately depends on a lot of factors, including the specific encoder used. There's a Google study on this in fact, 
 which found exactly this: <https://developers.google.com/speed/webp/docs/webp_study>.

SebinSong:
@corrideat
Thanks for clarification. Now I understand what you meant. Then, I feel we would def need to make sure the image's physical size doesn't get too small and too large either. I will go with 1024×768 you suggested first and see what Greg thinks.

SebinSong:
Updated the PR again [here](https://github.com/okTurtles/group-income/commit/db4564d79b83fd5d1926a472758128d226ba5db7) with Ricardo's feedback:
- Do not reduce the image dimension unless its physical size is too large.

<br>

So if the given image's physical size is beyond width 1024px and height 768px, the compressed image's dimension will be reduced to those max values. Otherwise, it doesn't change the dimension at all and only uses `quality` parameter to `canvas.toBlob()` for compression. (Just tested a few images after this update and it seems to work well too)

Let me know if we would like to change `1024X768` to something else here.

frontend/utils/image.js
```diff
@@ -28,3 +29,99 @@ export const imageUpload = async (imageFile: File, params: ?Object): Promise<Obj
...
+  return new Promise((resolve) => {
+    canvasEl.toBlob(blob => {
+      resolve(blob)
+    }, compressToType, quality)
+  })
+}
+
+function getResizingFactor (sourceImage) {
+  // If image's physical size is greater than the max dimension, resize the image to the max dimension.
+  const imageMaxDimension = { width: 1024, height: 768 }
```
taoeffect: Let's double this to 2048 x 1536

corrideat: Note that doubling it quadruples the image size. 

`2048*1536 / (1<<20) === 3`

I.e., an uncompressed image of that size would take up 3 MiB. Of course, this would generally be less with JPEG or WebP compression, but it'll come with some quality loss.

ETA: This comment is meant just as an observation. Also, the uncompressed size is actually 9 MiB, not 3 MiB as originally stated. The reason is that there are three channels (RGB) instead of just one.

corrideat: According to Adobe, JPEG typically has a 10:1 compression ratio over BMP (BMP is essentially uncompressed):

> Nonetheless, JPEG files often achieve compression of 10:1 with no significant decrease in quality. (<https://www.adobe.com/creativecloud/file-types/image/comparison/bmp-vs-jpeg.html>)

Assuming that WebP is about 30% more efficient and a ratio of 13:1, this is about 0.70 MiB. So, a `2048*1536` will likely have reduced quality to reach the 400 KiB goal.

SebinSong:
@taoeffect 
Updated the maximum image dimension to `2048 x 1536` as you requested. For a reminder, it means:

a) if the image file size is > `400kb` and physical size is **larger than** `2048 x 1536` -> compresses the image by reducing the dimension to the max size and using `quality` factor to `canvas.toBlob()` method.

b) if image file size is > `400kb` but physical size is **less than** `2048 x 1536` -> compresses the image only using the quality factor to `canvas.toBlob()` method.

---

I was worried increasing this max dimension might lead to noticeable low quality of the compression outcome. But so far, I haven't seen any visibly bad quality from the compressed images. So this adjustment seems okay.

frontend/utils/constants.js
```diff
@@ -20,6 +20,7 @@ export const CHAT_ATTACHMENT_SUPPORTED_EXTENSIONS = [
 // TODO: fetch this value from a server API
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
+export const IMAGE_ATTACHMENT_MAX_SIZE = 400000 // 400KB
```
taoeffect: Can you create a new entry for `KILOBYTE` just below `MEGABYTE`?

```js
export const KILOBYTE = 1 << 10
```

Then:

```js
export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE
```

This is a slightly larger number than 400000

frontend/utils/image.js
```diff
@@ -28,3 +29,99 @@ export const imageUpload = async (imageFile: File, params: ?Object): Promise<Obj
...
+      compressToType
+    })
+    const sizeDiff = blob.size - IMAGE_ATTACHMENT_MAX_SIZE
+
+    if (sizeDiff <= 0 || // if the compressed image is already smaller than the max size, return the compressed image.
+      quality <= 0.3) { // Do not sacrifice the image quality too much.
+      return blob
+    } else {
+      // if the size difference is greater than 100KB, reduce the next compression factors by 10%, otherwise 5%.
+      const minusFactor = sizeDiff > 100 * 1000 ? 0.1 : 0.05
```
taoeffect: And this can be edited to `100 * KILOBYTE` once that constant is defined.

taoeffect (changes requested):
Fantastic work @SebinSong! Two minor changes related to defining `KILOBYTE` and this should be good to merge 👍

frontend/utils/constants.js
SebinSong: Updated.

taoeffect (approved):
Fantastic work @SebinSong! 👏

PR #2431, opened 2024-11-25
#2425 - Improve navigation behaviour

### LINKED ISSUE #2425: Improve mobile navigation behavior
### Problem

On mobile, when the user selects the current view from the lefthand nav menu, nothing happens. It's not completely clear whether the app is unresponsive, or how to close the nav menu. 

image

### Solution

When the user selects a view from the nav menu, close the menu and (re)direct the user to the selected view, even if it's the existing view.

SebinSong:
closes #2425 

I noticed this issue happens in both Mobile/Desktop. I believe it makes sense to fold the menu regardless whether it navigates to a different page or the current page. So made that change as the fix. But please let me know if we don't want this.

### DIFF
--- a/frontend/views/containers/navigation/Navigation.vue
+++ b/frontend/views/containers/navigation/Navigation.vue
@@ -37,7 +37,7 @@ nav.c-navigation(
         notification-bell(v-if='!notApprovedToGroupYet' data-test='notificationBell')
 
       .c-navigation-body(
-        @click.self='enableTimeTravel'
+        @click.stop='onMenuItemsClick'
         v-if='!notApprovedToGroupYet && groupsByName.length'
       )
         .c-navigation-body-top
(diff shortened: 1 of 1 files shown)


frontend/views/containers/navigation/Navigation.vue
```diff
@@ -187,14 +186,12 @@ export default ({
     openModal (mode) {
       sbp('okTurtles.events/emit', OPEN_MODAL, mode)
     },
-    enableTimeTravel (evt) {
-      if (evt.shiftKey && process.env.NODE_ENV !== 'production') {
-        console.debug('enable time travel!')
-        this.ephemeral.timeTravelComponentName = 'TimeTravel'
-      }
-    },
```
SebinSong, line 195: I noticed this logic related to time-travel is no longer used anywhere in the app. So just removed it.

corrideat (approved):

taoeffect (approved):
LGTM

Issue #2429, opened 2024-11-20
Files should show how big they are


taoeffect:
### Problem

Currently when viewing a file attachment it doesn't tell you how big the attachment is:

![Screenshot 2024-11-20 at 11 53 08 AM](https://github.com/user-attachments/assets/4923e2eb-fe54-4cf5-ae43-99c6d03aad44)

### Solution

Display file size info (e.g. "7KB", "10MB", etc.) somewhere here:

![Screenshot 2024-11-20 at 11 53 08 AM](https://github.com/user-attachments/assets/5379ace6-deea-4624-a838-ca42dfca0506)

SebinSong:
NOTE to myself:
Once PR #2428 is merged, the original file sizes of images change if they are compressed. This needs to be taken into account and be worked on as another PR.

taoeffect:
@SebinSong The compression should happen before any file is uploaded, not afterward. Once files are uploaded their size shouldn't change... so I'm confused by your comment.

SebinSong:
@taoeffect I mean, the attachment `size` data should be updated accordingly... not compression itself. (Of course before uploading.)

SebinSong:
closed by #2435

PR #2435, opened 2024-11-27
#2429 -  File attachment in chat should show the size

SebinSong:
works on #2429 
There are currently two UIs for files attached to a message:

**[case 1] General one** - size is displayed next to the file format.

<br>

**[case 2] Image preview** - size is displayed as part of the tooltip text when hovered over the download button.

But please let me know if you have different ideas for them.

### DIFF
--- a/frontend/model/contracts/shared/types.js
+++ b/frontend/model/contracts/shared/types.js
@@ -2,7 +2,7 @@
 
 import {
   objectOf, objectMaybeOf, arrayOf, unionOf, boolean,
-  object, string, stringMax, optional, number, mapOf, literalOf
+  object, string, stringMax, optional, number, mapOf, literalOf, numberRange
 } from '~/frontend/model/contracts/misc/flowTyper.js'
 import {
   CHATROOM_TYPES,
@@ -19,6 +19,7 @@ import {
   CHATROOM_NAME_LIMITS_IN_CHARS,
   CHATROOM_DESCRIPTION_LIMITS_IN_CHARS
 } from './constants.js'
+import { CHAT_ATTACHMENT_SIZE_LIMIT } from '~/frontend/utils/constants.js'
 
 // group.js related
 
@@ -67,6 +68,7 @@ export const messageType: any = objectMaybeOf({
   attachments: arrayOf(objectOf({
     name: string,
     mimeType: string,
+    size: optional(numberRange(1, CHAT_ATTACHMENT_SIZE_LIMIT)),
     dimension: optional(objectOf({
       width: number,
       height: number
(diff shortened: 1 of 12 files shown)


frontend/model/contracts/shared/types.js
```diff
@@ -67,6 +68,7 @@ export const messageType: any = objectMaybeOf({
   attachments: arrayOf(objectOf({
     name: string,
     mimeType: string,
+    size: optional(numberRange(1, CHAT_ATTACHMENT_SIZE_LIMIT)),
```
taoeffect: Some observations:

- Not all messages have attachments, but all attachments have a file size, right? So I think maybe it makes sense to make the `arrayOf` on line 68 `optional`, and make this size here mandatory.
- Since `CHAT_ATTACHMENT_SIZE_LIMIT` could change in the future and even be an admin configurable value, let's make this value `Number.MAX_SAFE_INTEGER` instead.

SebinSong: Right. updated.

frontend/views/utils/filters.js
```diff
@@ -15,6 +15,17 @@ export const getFileType = (
   return mimeType.match('image/') ? 'image' : 'non-image'
 }
 
+export const formatBytesDecimal = (bytes: number, decimals: number = 2): string => {
+  if (bytes === 0) { return '0 Bytes' }
```
taoeffect: I would add some other checks here:

```js
if (bytes < 0 || !Number.isFinite(bytes)) return 'Invalid size'
```

frontend/views/utils/filters.js
```diff
@@ -15,6 +15,17 @@ export const getFileType = (
...
+
+  const k = 1000 // Decimal base
```
taoeffect: I think this should be 1024, as 1024 bytes = 1KB

frontend/model/contracts/shared/types.js
```diff
@@ -64,18 +64,21 @@ export const messageType: any = objectMaybeOf({
...
-    downloadData: objectOf({
-      manifestCid: string,
-      downloadParams: optional(object)
-    })
-  })),
+  attachments: optional(
+    arrayOf(objectOf({
+      name: string,
+      mimeType: string,
+      size: optional(numberRange(1, Number.MAX_SAFE_INTEGER)),
```
taoeffect: Can you check if we always provide this size info now? If so, this should probably no longer be optional.

taoeffect (changes requested):
Review ready!

SebinSong:
Updated the PR again

taoeffect (approved):
Excellent work @SebinSong!

Issue #2434, opened 2024-11-25
Ensure info/tooltips show in visible area


dotmacro:
### Problem

On mobile, when the user taps the "n members" link on the dashboard, the tooltip is not fully visible on the screen. 
![image](https://github.com/user-attachments/assets/949b0540-1121-47fd-bfec-05ef134bfef2)

### Solution

Ensure info/tooltip is fully visible on screen.

Vayras:
@dotmacro I'm a new contributor and I wish to contribute towards this issue

taoeffect:
@Vayras I've gone ahead and assigned this issue to you. Feel free to take it on!

taoeffect:
@dotmacro you shouldn't be seeing contractIDs for usernames. Can you open an issue for that too?

dotmacro:
> @dotmacro you shouldn't be seeing contractIDs for usernames. Can you open an issue for that too?

Investigated, opened #2437

taoeffect:
Closed by @Vayras in https://github.com/okTurtles/group-income/pull/2436

PR #2436, opened 2024-11-27
fixed tooltip position on mobile

Vayras:
issue: #2434 

fixed the tooltip position on mobile 
Screenshot 2024-11-27 at 11 52 08 AM

### DIFF
--- a/frontend/views/containers/dashboard/SentenceWithMemberTooltip.vue
+++ b/frontend/views/containers/dashboard/SentenceWithMemberTooltip.vue
@@ -4,7 +4,7 @@ component(
   :key='members.length'
   :opacity='members.length === 0 ? 0 : 0.95'
   triggerElementSelector='.t-trigger'
-  direction='bottom-center'
+  direction='bottom-middle'
 )
   // The reason for using <component /> tag here instead of <tooltip /> and specifying 'key' attr is,
   // to fix the bug where the link between the tooltip content(template(slot='tooltip') below) and the trigger target element(.t-trigger)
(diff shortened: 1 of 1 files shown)


taoeffect (changes requested):
Awesome - thanks for your first PR @Vayras!

I found a bug - please be sure to thoroughly test all PRs and check the console for errors, in this case there's an issue with `bottom-center` not being a valid `Tooltip` direction:

Screenshot 2024-11-27 at 12 37 28 PM

I recommend also testing with dummy data, you can temporarily modify `GroupMembersActivity.vue` to use dummy data to test the display.

Vayras:
@taoeffect thanks for the review I will fix this 🚀

Vayras:
@taoeffect I have fixed the direction property can you please review this again for me? Thanks

taoeffect (changes requested):
This change has the same problem.

corrideat:
@Vayras If you take a look at `Tooltip.vue`, these are the allowed attributes:

```js
    direction: {
      type: String,
      validator: (value) => ['bottom', 'bottom-left', 'bottom-right', 'right', 'left', 'top', 'top-left'].includes(value),
      default: 'bottom'
    },
```

If you need other values, you need to modify `Tooltip.vue`. However, your PR 'works' because the default value of `bottom` will actually be centred at the bottom.

Vayras:
@corrideat thanks for the info I looked at the file and found an attribute "bottom-middle" which is as you said the default attribute as well so I'm using that
Screenshot 2024-12-02 at 12 07 43 AM

taoeffect:
@Vayras What file are you seeing that in? I don't see that in any file in the project.

Vayras:
@corrideat Thanks for the heads up I think I was referring to the wrong tooltip file, fixed that in the latest commit

taoeffect (approved):
Seems to work in my tests, great job @Vayras!

PR #2443, opened 2024-12-01
Two bug-fixes for forced-color mode

### LINKED ISSUE #2385: Insufficient contrast image viewer zoom slider in forced colors mode
### Problem

Insufficient contrast / invisible zoom slider on image viewer when using [forced colors mode](https://drafts.csswg.org/css-color-adjust-1/#forced-colors-mode).

image

### Solution

Detect forced colors and make the slider visible

### LINKED ISSUE #2440: Modal close button disappears in forced-color mode
### Problem

Modal close button is there but not visible when forced-color mode is enabled.

### Solution
Fix it.

SebinSong:
closes #2385 
closes #2440 

**[Fix 1]**

<br>
<br>

**[Fix 2]**

### DIFF
--- a/frontend/assets/style/_mixins.scss
+++ b/frontend/assets/style/_mixins.scss
@@ -55,6 +55,12 @@ $payment-table-desktop-bp: 1290px;
   }
 }
 
+@mixin forced-color-mode {
+  @media (forced-colors: active) {
+    @content;
+  }
+}
+
 %unselectable {
   -webkit-touch-callout: none;
   -webkit-user-select: none;
--- a/frontend/assets/style/_mixins.scss
+++ b/frontend/assets/style/_mixins.scss
@@ -55,7 +55,7 @@ $payment-table-desktop-bp: 1290px;
   }
 }
 
-@mixin forced-color-mode {
+@mixin if-forced-color-mode {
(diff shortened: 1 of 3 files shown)


corrideat:
This looks good, but I have an issue with this part:

> I observed most elements that involve user interactions are rendered with white color when 'forced-color' mode is turned on. So made fixes using $white sass variable we have in the source-code.

That's not entirely accurate. See <https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors>. In particular (emphasis mine):

> **Note: User agents choose system colors based on native element semantics**, not on added ARIA roles. As an example, adding role="button" to a div will not cause an element's color to be forced to ButtonText

> There are two cases where the user agent does not force the values for the above properties — when a forced-color-adjust value of none is applied to an element, **or when a system color is specified by the author.**

So, instead of using `$white` or any arbitrary colour, I'd use a [system colour](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color) instead.

SebinSong:
@corrideat Those articles are informative actually. Updated the `$white` to system-color.

taoeffect (approved):
Nicely done @SebinSong!

corrideat (approved):

PR #2446, opened 2024-12-06
Password changes

### LINKED ISSUE #1923: Implement updating / changing password
### Problem

Password changes aren't supported.

### Solution

Now that the encrypted & authenticated key-value store has been implemented, we need to do one of two things to be able to rotate passwords:

1. Store old IEKs in the KV store and use them to decrypt old messages when syncing contracts
2. Store state snapshots / checkpoints (#354), which we need to do anyway... but it's a potentially difficult thing to do

corrideat:
Closes #1923


frontend/controller/actions/identity.js
```diff
@@ -680,6 +681,142 @@ export default (sbp('sbp/selectors/register', {
...
     return sbp('okTurtles.eventQueue/queueEvent', 'ACTIONS-LOGIN', ['gi.actions/identity/_private/logout', ...params])
   },
+  'gi.actions/identity/changePassword': async ({
+    identityContractID,
+    username,
+    oldIPK,
+    oldIEK,
+    newIPK: IPK,
+    newIEK: IEK,
+    newSAK: SAK,
```
taoeffect: Are these flowtype annotations...? I thought they weren't allowed like this and had to be done after the `}` as in `}: { newIPK: <type>`?

frontend/controller/actions/identity.js
```diff
@@ -680,6 +681,142 @@ export default (sbp('sbp/selectors/register', {
...
+
+    // Key IDs
+    const oldIPKid = keyId(oldIPK)
+    const oldIEKid = keyId(oldIEK)
+    const IPKid = keyId(IPK)
+    const IEKid = keyId(IEK)
+    const CSKid = keyId(CSK)
+    const CEKid = keyId(CEK)
+    const PEKid = keyId(PEK)
+    const SAKid = keyId(SAK)
```
taoeffect: Where is `SAK`, `IPK`, `IEK` etc. defined...?

taoeffect (commented):
Questions...

frontend/controller/actions/identity.js
corrideat: No, they're destructuring names.

```js
const foo = { bar: 1 }
const { bar: baz } = foo
console.log(baz) // 1
console.log(typeof bar) // undefined
```

frontend/controller/actions/identity.js
corrideat: Above (as arguments, see note on restructuring)

corrideat: Note: the shape of arguments for this function may not be finalised yet.

frontend/controller/actions/identity.js
taoeffect: 🤦 

Thanks!

frontend/controller/actions/identity.js
```diff
@@ -252,6 +321,38 @@ export default (sbp('sbp/selectors/register', {
       await sbp('chelonia/reset', { ...cheloniaState, loggedIn: { identityContractID } })
       await sbp('chelonia/storeSecretKeys', new Secret(transientSecretKeys))
 
+      if (oldKeysAnchorCid) {
+        try {
+          // Fetch the old keys from the server
+          const result = await fetch(`${sbp('okTurtles.data/get', 'API_URL')}/file/${oldKeysAnchorCid}`).then(handleFetchResult('json'))
```
corrideat: Move this block into a separate function and handle `OP_ATOMIC` as well.

shared/domains/chelonia/Secret.js
```diff
@@ -4,17 +4,17 @@ import { serdesDeserializeSymbol, serdesSerializeSymbol, serdesTagSymbol } from
 
 /* Wrapper class for secrets, which identifies them as such and prevents them
 from being logged */
-export class Secret<T> {
-  _content: T
 
+const wm = new WeakMap()
```
taoeffect, line 12: What is the reason for introducing this `WeakMap` class and related changes?

Could you please explain here in a comment?

frontend/controller/app/identity.js
```diff
@@ -434,5 +470,32 @@ export default (sbp('sbp/selectors/register', {
   },
   'gi.app/identity/logout': (...params) => {
     return sbp('okTurtles.eventQueue/queueEvent', 'APP-LOGIN', ['gi.app/identity/_private/logout', ...params])
+  },
+  'gi.app/identity/changePassword': async (woldPassword: Secret<string>, wnewPassword: Secret<string>) => {
```
taoeffect: `woldPassword` => `wOldPassword`
`wnewPassword` => `wNewPassword`

frontend/controller/actions/identity.js
```diff
@@ -689,6 +790,133 @@ export default (sbp('sbp/selectors/register', {
...
+      signingKeyId: oldIPKid,
+      publishOptions: {
+        headers: {
+          'shelter-name': username,
+          'shelter-salt-update-token': updateToken
+        }
+      }
+      /* hooks: {
+        preSendCheck
+      } */
```
taoeffect: If this isn't needed, can it be removed?

If it's needed, can you add a comment for why it's here? (Or uncomment?)

shared/domains/chelonia/Secret.js
corrideat: The reason is that this approach is a more foolproof way of avoiding logging `Secret` objects, since the data won't be in the object itself. Before, I didn't use `WeakMap`, but since we're now using that for circular reference checking, I decided to also use it here.

shared/zkpp.js
```diff
@@ -114,3 +135,26 @@ export const buildRegisterSaltRequest = async (publicKey: string, secretKey: Uin
 
   return [contractSalt, base64ToBase64url(Buffer.concat([nonce, encryptedHashedPasswordBuf]).toString('base64'))]
 }
+
+export const buildUpdateSaltRequestEa = async (password: string, c: Uint8Array): Promise<[string, string]> => {
```
taoeffect: What is `Ea`? Could you either explain that in a comment here, or update the name of the function to spell it out a bit more?

Same for `c` (the parameter)

shared/zkpp.js
```diff
@@ -114,3 +135,26 @@ export const buildRegisterSaltRequest = async (publicKey: string, secretKey: Uin
...
+  // Derive S_A and S_C as follows:
+  //   -> S_T -< BASE64(SHA-512(SHA-512(T) + SHA-512(c))[0..18]) with T being
```
taoeffect, line 147: Please explain in the comment or by updating the variable names how `T` maps to the input parameters here. I'm assuming `c` in this comment refers to `c` in the parameter.

shared/zkpp.js
```diff
@@ -114,3 +135,26 @@ export const buildRegisterSaltRequest = async (publicKey: string, secretKey: Uin
...
+  //     `AUTHSALT` or `CONTRACTSALT`
+  // This way, we ensure both the server and the client contribute to the
+  //   salts' entropy.
+  // When sending the encrypted data, the encrypted information would be
+  // `hashedPassword`, which needs to be verified server-side to verify
+  // it matches p and would be used to derive S_A and S_C.
+  const authSalt = Buffer.from(hashStringArray('AUTHSALT', c)).slice(0, 18).toString('base64')
+  const contractSalt = Buffer.from(hashStringArray('CONTRACTSALT', c)).slice(0, 18).toString('base64')
```
taoeffect: Since `'AUTHSALT'` and `'CONTRACTSALT'` string literals are used twice in this file, does it make sense to define them as constants at the top of the file to avoid the risk of typos and to make it clear that these values must be the same in both places? (Same for `'SU'` and `'CS'`)

shared/zkpp.js
```diff
@@ -114,3 +135,26 @@ export const buildRegisterSaltRequest = async (publicKey: string, secretKey: Uin
 
   return [contractSalt, base64ToBase64url(Buffer.concat([nonce, encryptedHashedPasswordBuf]).toString('base64'))]
 }
+
+export const buildUpdateSaltRequestEa = async (password: string, c: Uint8Array): Promise<[string, string]> => {
+  // Derive S_A and S_C as follows:
+  //   -> S_T -< BASE64(SHA-512(SHA-512(T) + SHA-512(c))[0..18]) with T being
+  //     `AUTHSALT` or `CONTRACTSALT`
+  // This way, we ensure both the server and the client contribute to the
+  //   salts' entropy.
```
taoeffect: Why do we care that the server contributes to the salt's entropy? Can this be explained as well here?

shared/domains/chelonia/encryptedData.js
```diff
@@ -277,6 +277,51 @@ export const encryptedIncomingForeignData = <T>(contractID: string, _0: any, dat
...
+  return wrapper({
+    get encryptionKeyId () {
+      return encryptedDataKeyId(data)
+    },
+    get serialize () {
+      return () => data
+    },
+    get toString () {
+      return () => JSON.stringify(this.serialize())
+    },
```
taoeffect, line 325: I realize I didn't ask this when you sent in the original PR that introduced all of these concepts (or at least, I don't remember asking), but why is it that `serialize` returns the input encrypted `data` instead of the decrypted version of it? Maybe that can be explained in a comment somewhere?

taoeffect (changes requested):
Preliminary review ready!

Also, could you please add one test to `signup-and-login.spec.js` to test the changing of the password? I.e. it can do the following:

1. After logging in, change the password.
2. Log out
3. Verify that logging back in with the old password doesn't work
4. Verify that logging back in with the new password does work

shared/zkpp.js
corrideat: Having more sources of entropy is better because it makes the result more random and unpredictable.

shared/zkpp.js
corrideat: `T` doesn't map at all to the input parameters.

shared/domains/chelonia/encryptedData.js
corrideat: Comments added at the top of the file. `serialize` returns encrypted data because `encryptedData` is meant as a container for data. It just seems logical to me. Also, this is called from `GIMessage.serialize()`. To get the plaintext, you'd have to call `valueOf()`

frontend/controller/actions/identity.js
corrideat: I've added `hooks` as a parameter.

corrideat:
I believe this is ready to merge, for the most part. ~~However, when you test it, you'll notice that the validation error messages seem to be default ones, like `minLength`. I can add the appropriate messages if you tell me what they should be.~~

backend/routes.js
```diff
@@ -116,7 +116,24 @@ route.POST('/event', {
           }
         }
       }
+      const saltUpdateToken = request.headers['shelter-salt-update-token']
```
corrideat, line 119: This implements the atomic update using a token.

backend/routes.js
```diff
@@ -821,7 +838,7 @@ route.GET('/zkpp/{name}/contract_hash', {
   return Boom.internal('internal error')
 })
 
-route.POST('/zkpp/updatePasswordHash/{name}', {
+route.POST('/zkpp/{name}/updatePasswordHash', {
```
corrideat, line 841: Renamed for consistency

backend/zkppSalt.js
```diff
@@ -60,10 +63,23 @@ export const initZkpp = async () => {
   recordSecret = Buffer.from(hashStringArray('private/recordSecret', IKM)).toString('base64')
   challengeSecret = Buffer.from(hashStringArray('private/challengeSecret', IKM)).toString('base64')
   registrationSecret = Buffer.from(hashStringArray('private/registrationSecret', IKM)).toString('base64')
+  hashUpdateSecret = Buffer.from(hashStringArray('private/hashUpdateSecret', IKM)).toString('base64')
```
corrideat, line 66: Used for encrypting the update token.

backend/zkppSalt.js
```diff
@@ -288,21 +310,50 @@ export const updateContractSalt = async (contract: string, r: string, s: string,
...
+  )
+
+  const [timestamp, hashedPassword, authSalt, contractSalt] = JSON.parse(decryptedToken)
+
+  if (timestamp < (Date.now() - 180e3)) {
+    throw new Error('ZKPP token expired')
+  }
+
+  return (cid: ?string) => {
+    return setZkppSaltRecord(contract, hashedPassword, authSalt, contractSalt, cid)
```
corrideat, line 357: Two-step process to allow validating the token first before making the change.

frontend/views/containers/user-settings/UserProfile.vue
```diff
@@ -73,6 +73,7 @@
         i18n.is-title-3(tag='h3' class='card-header') Delete account
         p
           i18n Deleting your account will erase all your data, and remove you from the groups you belong to.
+          | {{ ' ' }}
```
corrideat, line 76: This fixes a display issue (there was no space)

shared/domains/chelonia/encryptedData.js
```diff
@@ -8,10 +8,20 @@ import { isRawSignedData, signedIncomingData } from './signedData.js'
 const rootStateFn = () => sbp('chelonia/rootState')
 
 export interface EncryptedData<T> {
+  // The ID of the encryption key used
```
corrideat, line 11: Explanation you requested added here.

backend/zkppSalt.js
```diff
@@ -288,21 +310,50 @@ export const updateContractSalt = async (contract: string, r: string, s: string,
...
 
-    await setZkppSaltRecord(contract, hashedPassword, authSalt, contractSalt)
+    const token = encryptSaltUpdate(
+      hashUpdateSecret,
+      recordId,
+      JSON.stringify([Date.now(), hashedPassword, authSalt, contractSalt])
+    )
 
-    return true
+    return encryptContractSalt(c, JSON.stringify([oldContractSalt, token]))
```
corrideat, line 330: Instead of returning `true`, we return the token for requesting an update. We also return the `oldContractSalt` because it's helpful for avoiding an extra request, as we need it to decrypt the old (current) IEK.

backend/zkppSalt.js
```diff
@@ -85,17 +101,23 @@ const getZkppSaltRecord = async (contractID: string) => {
     try {
       const recordObj = JSON.parse(recordString)
 
-      if (!Array.isArray(recordObj) || recordObj.length !== 3 || !recordObj.reduce((acc, cv) => acc && typeof cv === 'string', true)) {
+      if (
+        !Array.isArray(recordObj) ||
+        (recordObj.length !== 3 && recordObj.length !== 4) ||
```
corrideat, line 106: This is because before the record was a triplet, and now it can be a triplet or a 4-tuple (with the 4th entry being a CID, if there's been a password update)

backend/zkppSalt.js
```diff
@@ -251,10 +273,10 @@ export const getContractSalt = async (contract: string, r: string, s: string, si
     throw new Error('getContractSalt: Bad challenge')
   }
 
-  return encryptContractSalt(c, contractSalt)
+  return encryptContractSalt(c, JSON.stringify([contractSalt, cid]))
```
corrideat, line 276: We now return the contractSalt and the CID.

frontend/controller/actions/identity.js
```diff
@@ -9,22 +9,173 @@ import {
...
+  sbp('gi.actions/identity/logout').catch(e => {
+    console.error('Error while automatically logging out', e)
+  })
+})
+
 export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/create': async function ({
-    IPK,
-    IEK,
+    IPK: wIPK,
```
corrideat, line 177: These changes aren't strictly related to this PR, rather something that slipped through the cracks in the past. This wraps secrets in `Secret` objects.

frontend/controller/app/identity.js
```diff
@@ -187,7 +187,41 @@ export default (sbp('sbp/selectors/register', {
       'hc': Buffer.from(hc).toString('base64').replace(/\//g, '_').replace(/\+/g, '-').replace(/=*$/, '')
     })).toString()}`).then(handleFetchResult('text'))
 
-    return decryptContractSalt(c, contractHash)
+    // [contractSalt, cid]
+    return JSON.parse(decryptContractSalt(c, contractHash))
```
corrideat, line 191: This is because of the API change, now instead of just `contractSalt`, it could include a CID.

frontend/controller/app/identity.js
```diff
@@ -220,9 +254,8 @@ export default (sbp('sbp/selectors/register', {
     // next create the identity contract itself
     try {
       const userID = await sbp('gi.actions/identity/create', {
-        // TODO: Wrap IPK and IEK in "Secret"
-        IPK: serializeKey(IPK, true),
-        IEK: serializeKey(IEK, true),
+        IPK: new Secret(serializeKey(IPK, true)),
```
corrideat, line 257: Implement the old TODO.

frontend/controller/app/identity.js
```diff
@@ -434,5 +476,32 @@ export default (sbp('sbp/selectors/register', {
...
+    const getters = sbp('state/vuex/getters')
+
+    const { identityContractID } = state.loggedIn
+    const username = getters.usernameFromID(identityContractID)
+    const oldPassword = wOldPassword.valueOf()
+    const newPassword = wNewPassword.valueOf()
+
+    const [newContractSalt, oldContractSalt, updateToken] = await sbp('gi.app/identity/updateSaltRequest', username, wOldPassword, wNewPassword)
+
+    const oldIPK = await deriveKeyFromPassword(EDWARDS25519SHA512BATCH, oldPassword, oldContractSalt)
```
corrideat, line 492: We do this here to avoid ever sending the password to the SW.

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -14,7 +14,6 @@ modal-template(class='is-centered is-left-aligned' back-on-mobile=true ref='moda
       :value='form'
       :$v='$v'
       @enter='changePassword'
-      @input='(password) => { newPassword = password }'
```
corrideat, line 17: These seemed to be wrong.

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -40,34 +38,40 @@ modal-template(class='is-centered is-left-aligned' back-on-mobile=true ref='moda
...
       ) Cancel
 
-      i18n.is-success(
-        tag='button'
+      button-submit(
         @click='changePassword'
-        :disabled='$v.form.$invalid'
-      ) Change password
-
-  template(slot='errors') {{ form.response }}
```
corrideat, line 62: This didn't work, and the slot doesn't exist. Replaced it with `banner-scoped`.

taoeffect (changes requested):
Really impressive work @corrideat!

It seems to work well in my tests in almost every browser. However, there seems to be an issue with Safari:

Steps:

1. Create account `u1` in Firefox Dev Edition, with group
2. Open Safari, attempt to sign in as `u1`

Results:

Screenshot 2024-12-15 at 5 02 03 PM

Server log shows:

```
[17:00:50.538] DEBUG (14007): 127.0.0.1: GET /assets/js/undefined/name/u1 --> 404
[17:01:15.697] DEBUG (14007): [pubsub] Pinging clients
[17:01:45.698] DEBUG (14007): [pubsub] Pinging clients
[17:02:15.510] DEBUG (14007): 127.0.0.1: GET /assets/js/undefined/name/u1 --> 404
[17:02:15.699] DEBUG (14007): [pubsub] Pinging clients
```

frontend/views/containers/access/PasswordModal.vue
```diff
@@ -119,13 +114,22 @@ export default ({
...
+            new Secret(this.form.newPassword)
+          )
+          this.closeModal()
+        } catch (error) {
+          console.error('[PasswordModal.vue]', error)
+          this.$refs.formMsg.danger(L('Invalid password'))
+        }
+      })().finally(() => {
+        this.processing = false
+      })
```
taoeffect, line 132: Nice job preventing accidental double-clicks!

backend/zkppSalt.js
taoeffect: Are these strings like `'private/hashUpdateSecret'` something that should be put in `zkppConstants.js` too?

corrideat: Probably not, since they're only used here for initialisation purposes.

corrideat: They also have no particular significance with regard to the protocol.

frontend/setupChelonia.js
```diff
@@ -61,9 +61,6 @@ const setupChelonia = async (): Promise<*> => {
     }
   })
 
-  // this is to ensure compatibility between frontend and test/backend.test.js
-  sbp('okTurtles.data/set', 'API_URL', self.location.origin)
-
```
taoeffect, line 66: Is this change intentional?

corrideat: Yes. This was moved to `sw-primary.js`

corrideat: That ensures both that it's called earlier and that it's not in `setupChelonia.js` (to avoid it being called twice if it was called by `main.js`.

backend/routes.js
```diff
@@ -116,7 +116,24 @@ route.POST('/event', {
...
+        if (namedContractID !== deserializedHEAD.contractID) {
+          throw new Error('Mismatched contract ID and name')
+        }
+        updateSalts = await redeemSaltUpdateToken(name, saltUpdateToken)
+      }
       await sbp('backend/server/handleEntry', deserializedHEAD, request.payload)
+      // If it's a salt update, do it now after handling the message. This way
+      // we make it less likely that someone will end up locked out from their
+      // identity contract.
+      await updateSalts?.(deserializedHEAD.hash)
```
taoeffect, line 136: Claude:

> There's a potential race condition between validating the token and applying the update. If multiple requests come in with the same token, the validation could succeed for both but only one should be allowed to update. The token validation and update should be atomic.

Is he full of it?

corrideat: Potentially it could happen, however, not really, because the token only applies to one contract and we can't write to the same contract simultaneously (or shouldn't be able to), so `handleEntry` will throw.

taoeffect (approved):
Amazing!

PR #2357, opened 2024-09-19
Chelonia in SW

corrideat:
@taoeffect adds: we should add `VAPID_EMAIL` now to our host server

PR #2454, opened 2024-12-11
#2452 - Unecessary kv error in console

### LINKED ISSUE #2452: Unnecessary key-value errors in console when app starts logged out
### Problem

Screenshot 2024-12-09 at 10 32 04 AM

`'gi.actions/identity/kv/load'` and possibly also `'gi.actions/group/kv/load'` trigger these errors when starting up a fresh session with no logged in user.

Also, there's a typo in error message in `group-kv.js` that makes it say `'gi.actions/identity/kv/load'` instead of `'gi.actions/group/kv/load'`:

```js
sbp('okTurtles.events/on', ONLINE, () => {
  sbp('gi.actions/group/kv/load').catch(e => {
    console.error("Error from 'gi.actions/identity/kv/load' after reestablished connection:", e)
  })
})
```

### Solution

Do not log any error to the console if we are not logged in.

Could be as simple as checking whether or not we're logged in inside of the `ONLINE` handler and not calling `load` if we're not logged in.

SebinSong:
closes #2452

### DIFF
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -6,7 +6,7 @@ import { KV_QUEUE, ONLINE } from '~/frontend/utils/events.js'
 sbp('okTurtles.events/on', ONLINE, async () => {
   try {
     const identityContractID = (await sbp('chelonia/rootState')).loggedIn?.identityContractID
-    if (identityContractID) {
+    if (identityContractID) { // check if the user is logged in before loading the data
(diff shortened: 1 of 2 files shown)


SebinSong:
@taoeffect I think this issue has been resolved by https://github.com/okTurtles/group-income/pull/2357 and we can discard this PR.

taoeffect:
I think you're right @SebinSong!

PR #2456, opened 2024-12-13
#2448 - Fix the invisible menu issue

### LINKED ISSUE #2448: Ensure menus expand in visible area
### Problem

"Create proposal v" menu does not always expand in visible area.

Steps to reproduce:
1. Scroll to bottom of Dashboard (i.e. bottom of "Proposals" section).
2. Tap "Create proposal v" to open menu
3. See menu expand, but not fully onscreen (depending on height of "Proposals" section) **<-- the problem**
4. Optionally, scroll to bottom of screen to see full list of proposal options
5. Tap "Create proposal ^" to hide menu
6. Menu disappears and bottom of "Proposals" section is properly spaced from bottom of screen (via shrink & auto-scrolling if necessary)

Short "Proposals" section means full menu of proposal types isn't visible
image

Sufficiently long "Proposals" section means full menu of proposal types is visible, no problem
image

### Solution

1. Ensure all proposal types are visible when menu is opened. Expanding the page and auto-scrolling so the full menu is visible (i.e. the opposite of functionality in step 6 above) would be a smooth way to do this.
2. Optional: Align menu with right (preferred, current) or left (fallback if necessary) of "Create proposal v" button so that the menu is also fully visible left-to-right.

SebinSong:
closes #2448 

[Fix screenshot]

Also, made it auto-scroll to the menu when it's open.

### DIFF
--- a/frontend/views/components/ButtonDropdownMenu.vue
+++ b/frontend/views/components/ButtonDropdownMenu.vue
@@ -4,7 +4,7 @@ menu-parent.c-menu(v-on='listeners')
     | {{ buttonText }}
     i.icon-angle-down.is-suffix
 
-  menu-content.c-menu-content
+  menu-content.c-menu-content(:class='"bound-to-" + boundEdge')
     ul
       template(v-if='options')
         component(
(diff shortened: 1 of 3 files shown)


taoeffect (approved):
Seems to work well!

PR #2458, opened 2024-12-14
#2457 - Fix the broken blockquotes in the chat markdown

### LINKED ISSUE #2457: Chat doesn't render blockquotes in chat anymore
### Problem

Here's a bit of markdown:

```
> test

test
```

Here's how it's rendered in the chat:

Screenshot 2024-12-13 at 3 54 34 PM

And here's how it's supposed to render:

> test

test

### Solution

Update the markdown renderer to properly restore rendering quote strings. (I think this used to work? What happened?)

SebinSong:
closes #2457 

[ Fix screenshot ]

Made sure this fix does not revert the fix for #2130 too.

### DIFF
--- a/frontend/assets/style/components/_custom-markdowns.scss
+++ b/frontend/assets/style/components/_custom-markdowns.scss
@@ -115,6 +115,7 @@
     position: relative;
     padding-left: 1rem;
     white-space: pre-line;
+    margin: 0.25rem 0;
 
     &::before {
       content: "";
@@ -126,6 +127,11 @@
       bottom: 0;
       border-radius: 8px;
       background-color: $general_0;
+    }
+
+    p:has(+ p, + pre),
+    pre:has(+ p, + pre) {
+      margin-bottom: 1rem;
     }
   }
 
--- a/frontend/views/utils/markdown-utils.js
+++ b/frontend/views/utils/markdown-utils.js
@@ -37,12 +37,13 @@ export function renderMarkdown (str: string): any {
   strSplitByCodeMarkdown.forEach((entry, index) => {
     if (entry.type === 'plain' && strSplitByCodeMarkdown[index - 1]?.text !== '```') {
       let entryText = entry.text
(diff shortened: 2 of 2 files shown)


taoeffect (changes requested):
Nice work @SebinSong!

Found one issue related to empty paragraph breaks in blockquotes.

Markdown:

```md
> Test

And

> `another () => test` of the
> 
> ```
> emergency () => system
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one

() => test
```

Group Income:

Screenshot 2024-12-15 at 4 46 11 PM

Github:

> Test

And

> `another () => test` of the
> 
> ```
> emergency () => system
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one

() => test

SebinSong:
@taoeffect Improved the PR with additional fix.

taoeffect (changes requested):
EDIT: I can also save this for a separate issue to fix if you'd prefer!

Interesting... it seems the logic for @mentions and channel name mentions is a bit wonky too, as sometimes it works, and sometimes it doesn't:

Markdown:

```md
> @u1 Test #general 

And

> `another () => test` of the
> 
> ```
> emergency () => system
> and
> 
> break
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one @u1 #general 

() => test
```

Result:

Screenshot 2024-12-16 at 10 26 03 AM


Github (checking to see if it handles @u1...):

> @u1 Test #general 

And

> `another () => test` of the
> 
> ```
> emergency () => system
> and
> 
> break
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one @u1 #general 

() => test

taoeffect (approved):
I'm actually going to merge this now and open a separate issue for that separate bug I found, since this PR does close the original issue. Great work! 👍

PR #2459, opened 2024-12-15
Basic notification support

### LINKED ISSUE #2455: Implement push notifications for notifications in `templates.js`
### Problem

We need users to be notified when the distribution period is almost up, when a new proposal is opened (or closed), etc.

These notifications are enumerated in the file `templates.js`, and they are called from various places. An in-app notification is created, but a push notification isn't shown to the user.

### Solution

Make it so that push notifications are shown for these notifications.

Some of these notifications are triggered as a result of side effect functions, while others are on a timer of some sort (like `NEAR_DISTRIBUTION_END`).

For the ones that are on a timer, you'll probably need to have the server send a periodic push notification to wake up the service worker periodically (e.g. twice a day or something like that), to then check to see if any of these notifications are worth showing.

corrideat:
Closes #2455


corrideat:
So far:

> 1. Augment templates.js with fields specific to native notifications (I think at least title and body, possibly also icon).

TBD after discussion.

> 2. Modify gi.notifications/emit or define an event handler for NOTIFICATION_EMITTED calling makeNotification.

Using `NOTIFICATION_EMITTED`.

> 3. Solve the currentGroupId issue for group-scoped native notifications (relevant for when the native notification is clicked)

Solved for existing windows, not solved yet when opening a new window.

> 4. Implement periodic notifications by either waking up the service worker at regular intervals or through on-demand wake up requests.

Not done.

In addition, this remains:

1. If not augmenting `templates.js` with body, find a way to strip the HTML tags
2. Replace `@contractID` with `@username` or `display name` as in-app notifications do.

frontend/model/notifications/templates.js
```diff
@@ -93,9 +108,13 @@ export default ({
   },
   INCOME_DETAILS_OLD (data: { months: number, lastUpdatedDate: string }) {
     return {
+      title: L('Update income details'),
       body: L("You haven't updated your income details in more than {months} months. Would you like to review them now?", {
         months: Math.floor(data.months) // Avoid displaying decimals
       }),
+      plaintextBody: L("You haven't updated your income details in more than {months} months. Would you like to review them now?", {
+        months: Math.floor(data.months) // Avoid displaying decimals
+      }),
```
taoeffect, line 117: In the case where there's no difference, can we leave it `undefined` and default to `body`?

frontend/model/notifications/templates.js
```diff
@@ -107,49 +126,67 @@ export default ({
...
     return {
       avatarUserID: data.memberID,
+      title: rootState[data.groupID]?.settings?.groupName || L('Member added'),
       body: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
         name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.memberID}`,
         ...LTags('strong')
       }),
+      plaintextBody: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
+        name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.memberID}`
+      }),
```
taoeffect: `{strong}` still here

corrideat: Yes, that's intentional to avoid having to translate the strings twice. Note that there's no `LTags('strong')` and that the `L` function replaces these missing references with `''`. In effect, this is eliding the `{strong_}` and `{_strong}`.

taoeffect: Ah, makes senes.

> the L function replaces these missing references with ''

I forgot it was able to do that!

frontend/model/notifications/templates.js
```diff
@@ -241,10 +307,18 @@ export default ({
       name: !isCreator ? `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${creatorID}` : ''
     }
 
+    const plaintextArgs = {
+      ...options,
+      ...LTags('strong'),
```
taoeffect: I thought plaintext didn't have markup?

corrideat: Right, that shouldn't be there. Good catch. Also, I think I'll replace the @ with the actual display name

frontend/controller/service-worker.js
```diff
@@ -104,9 +104,16 @@ sbp('sbp/selectors/register', {
...
+              if (data.groupID) {
+                sbp('state/vuex/commit', 'setCurrentGroupId', { contractID: data.groupID })
+              }
               sbp('controller/router').push({ path: data.path }).catch(console.warn)
               break
             }
+            case 'sbp': {
+              sbp(...deserializer(event.data.data))
+              break
+            }
```
taoeffect, line 139: Can you add a comment here explaining what this addition is for?


Also, a reminder to update `push.js` to address:

```js
      if (data.contractType === 'gi.contracts/chatroom') {
        // TODO: Text for this notification
        return makeNotification({ title: '@@@err', body: e.message })
      }
```

EDIT: regarding what the notification should say for platforms that don't implement PWAs correctly and prevent access to storage (iOS):

- For `gi.contracts/chatroom`: `New chatroom message. An iOS bug prevents us from saying what it is.`
- For `gi.contracts/group`: `New group activity. An iOS bug prevents us from saying what it is.`

corrideat: > Can you add a comment here explaining what this addition is for?

What do you mean by that? What needs to be documented here? I think that what it does is pretty obvious (execute SBP invocations from the SW to the browser). The purpose is perhaps less clear (it's currently used by the 'notificationclick' handler) for notifications that have an 'sbpInvocation' instead of a path.

Can you clarify what needs to be documented and why this `case` is special over all the others here?

taoeffect: We need to explain why it's here, what it's used for.

> The purpose is perhaps less clear (it's currently used by the 'notificationclick' handler) for notifications that have an 'sbpInvocation' instead of a path.

You can just copy/paste this. 👆

frontend/views/containers/notifications/NotificationCard.vue
```diff
@@ -149,7 +149,7 @@ export default {
 }
 
 .c-body {
-  height: 18rem;
+  height: 30rem;
```
corrideat: To be reverted

test/cypress/integration/group-paying.spec.js
```diff
@@ -475,11 +475,14 @@ describe('Group Payments', () => {
     ])
 
     cy.log('user3 receives a notification for the payment and clicking on it opens a "Payment details" modal.')
+    cy.getByDT('DOESNOTEXIST')
```
corrideat: To be removed

test/cypress/integration/group-paying.spec.js
```diff
@@ -475,11 +475,14 @@ describe('Group Payments', () => {
...
     openNotificationCard({
       notificationsCount: 5,
       messageToAssert: `user1-${userId} sent you a $250 mincome contribution. Review and send a thank you note.`
     })
 
+    cy.getByDT('DOESNOTEXIST')
```
corrideat: To be removed

frontend/model/notifications/periodicNotifications.js
```diff
@@ -38,59 +34,142 @@ const validateNotificationData = objectOf({
...
+  // We set `clearTimeout` to a local value to stop processing notifications
+  // when clearing stop notifications, to avoid calling
+  // `runNotificationListRecursive` while it's running and to be able to compare
+  // it to the local value later to replace `clearTimeout` later on in this
+  // function
+  let aborted: boolean = false
+  const noop = () => { aborted = true }
+  ephemeralNotificationState.clearTimeout = noop
+
+  // If there are any queued invocations, wait until they're done
```
taoeffect, line 88: Can you add a comment explaining why you're waiting for queued invocations to clear?

frontend/model/notifications/mainNotificationsMixin.js
```diff
@@ -262,16 +106,7 @@ const periodicNotificationEntries = [
```
SebinSong, line 106: (Unless it was intentional) Is there a reason that this notification entry was not moved to `mainPeriodicNotificationEntries.js` along with others?

frontend/model/notifications/periodicNotifications.js
```diff
@@ -13,16 +12,13 @@ export const PERIODIC_NOTIFICATION_TYPE = {
...
-const every15MinTimeout = { notifications: [], state: {}, delay: 15 * MINS_MILLIS }
-const every30MinTimeout = { notifications: [], state: {}, delay: 30 * MINS_MILLIS }
+const ephemeralNotificationState: { notifications: any[], partition: Object, clearTimeout?: Function } = { notifications: [], partition: Object.create(null) }
 
-const typeToObjectMap = {
-  [PERIODIC_NOTIFICATION_TYPE.MIN1]: every1MinTimeout,
-  [PERIODIC_NOTIFICATION_TYPE.MIN5]: every5MinTimeout,
-  [PERIODIC_NOTIFICATION_TYPE.MIN15]: every15MinTimeout,
-  [PERIODIC_NOTIFICATION_TYPE.MIN30]: every30MinTimeout
+const delayToObjectMap = {
```
SebinSong: The variable name here would be more clear if it was `typeToDelayMap` or something like that. (as the field value here no longer refers to an object)

frontend/model/notifications/periodicNotifications.js
```diff
@@ -38,59 +34,142 @@ const validateNotificationData = objectOf({
...
+    const delay = entry.delay
+    if ((Date.now() - lastRun) >= delay) {
+      try {
+        if (!firedMap[stateKey] && callWithStates(entry.emitCondition, stateKey)) {
+          await callWithStates(entry.emit, stateKey)
+          firedMap[stateKey] = true
+        }
 
-    if (firedMap[entry.stateKey] && callWithStates(entry.shouldClearStateKey)) {
-      Vue.delete(firedMap, entry.stateKey)
```
SebinSong, line 54: I see multiple places where `Vue.delete` and `Vue.set` are replaced with pure Javascript logics here. I think it is fine as `alreadyFired` and `lastRun` maps are currently only used in this file and _[1] there is no place in the app to re-render in response to these state updates_, as far as I know.

For the same reason stated in [1] above, I think the entire `periodicNotificationAlreadyFiredMap` object can be moved out of Vuex store too but it's not necessary.

frontend/model/notifications/periodicNotifications.js
```diff
@@ -13,16 +12,13 @@ export const PERIODIC_NOTIFICATION_TYPE = {
   MIN30: '30MIN'
 }
 
-const every1MinTimeout = { notifications: [], state: {}, delay: MINS_MILLIS }
-const every5MinTimeout = { notifications: [], state: {}, delay: 5 * MINS_MILLIS }
-const every15MinTimeout = { notifications: [], state: {}, delay: 15 * MINS_MILLIS }
-const every30MinTimeout = { notifications: [], state: {}, delay: 30 * MINS_MILLIS }
+const ephemeralNotificationState: { notifications: any[], partition: Object, clearTimeout?: Function } = { notifications: [], partition: Object.create(null) }
```
SebinSong, line 27: This looks like a great DRYing here:
- multiple `state: {}` are now merged into `partition: { stateKey: {}, ... }`.
- multiple `notifications: []` are now merged into `ephemeralNotificationState.notifications: [...]`

SebinSong (dismissed):
@corrideat Reviewed the parts related to the periodic notifications logic in this PR. They look good to me besides a couple of minor suggestions.

test/cypress/integration/group-paying.spec.js
```diff
@@ -475,11 +475,14 @@ describe('Group Payments', () => {
     ])
 
     cy.log('user3 receives a notification for the payment and clicking on it opens a "Payment details" modal.')
+    // cy.getByDT('DOESNOTEXIST')
```
taoeffect: If the PR is no longer a draft, then this commented-out debug stuff should be removed (along with the one below)

frontend/model/notifications/mainNotificationsMixin.js
corrideat: It was intentional. Since we added it to fix an issue that we don't fully understand and seemed to affect the UI, I left it there to run in the UI.

frontend/model/notifications/periodicNotifications.js
corrideat: I agree with you that it's not used elsewhere in the UI AFAIK. The reason to remove `Vue.set` and friends is so that it can run in the SW without having to pull in Vue there as a dependency. However, I think it's fine to leave it in the state (as is) so that we can restore the state alongside the rest of it.

frontend/model/notifications/periodicNotifications.js
corrideat: Awesome

frontend/model/notifications/periodicNotifications.js
corrideat: Done

frontend/model/notifications/templates.js
```diff
@@ -43,10 +46,23 @@ export default ({
...
     }
 
+    const LplaintextParams = {
+      errName: error.name,
+      activity,
+      action: action ?? opType,
+      contract: state.contracts[contractID]?.type ?? contractID,
+      who: plaintextWho,
+      errMsg: error.message ?? '?'
+    }
```
taoeffect: A lot of repetition here between `Lparams` and `LplaintextParams`, can it be DRY'd please using the spread operator?

frontend/model/state.js
```diff
@@ -82,10 +82,13 @@ const initialState = {
...
   reverseNamespaceLookups: Object.create(null), // { [contractID]: username }
-  periodicNotificationAlreadyFiredMap: {}, // { notificationKey: boolean },
   contractSigningKeys: Object.create(null),
   lastLoggedIn: {}, // Group last logged in information
-  preferences: {}
+  preferences: {},
+  periodicNotificationAlreadyFiredMap: {
+    alreadyFired: Object.create(null), // { notificationKey: boolean },
+    lastRun: Object.create(null) // { notificationKey: number },
+  }
```
taoeffect, line 91: I think it's a general rule that whenever this initial dataset is modified it requires a corresponding section under `postUpgradeVerification` for older clients that did not start out with this data.

frontend/model/notifications/periodicNotifications.js
```diff
@@ -13,16 +12,13 @@ export const PERIODIC_NOTIFICATION_TYPE = {
...
```
taoeffect: This is such a long long, and on small screens it's kind of hard to read. Could you split this up using the brackets as guides, e.g.:

```js
const ephemeralNotificationState: {
  notifications: any[], partition: Object, clearTimeout?: Function
} = {
  notifications: [], partition: Object.create(null)
}
```

frontend/model/state.js
corrideat: I'm not sure it's needed in this particular case, since starting periodic notifications will reset this object.

frontend/model/notifications/periodicNotifications.js
```diff
@@ -1,7 +1,6 @@
 'use strict'
 
 import sbp from '@sbp/sbp'
-import Vue from 'vue'
```
taoeffect, line 4: Could you add a comment near the top of the file explaining:

- That this file is used in both the SW and the Browser UI
- That we do not import Vue for two reasons: (1) it's bulk that's not needed in the SW, and (2) what @SebinSong [said](https://github.com/okTurtles/group-income/pull/2459#discussion_r1902262722):

> I see multiple places where Vue.delete and Vue.set are replaced with pure Javascript logics here. I think it is fine as alreadyFired and lastRun maps are currently only used in this file and [1] there is no place in the app to re-render in response to these state updates, as far as I know.

frontend/model/notifications/periodicNotifications.js
```diff
@@ -38,59 +34,147 @@ const validateNotificationData = objectOf({
...
+  if (ephemeralNotificationState.clearTimeout) return
+  // Prevent accidentally calling `runNotificationListRecursive` while it's
+  // still running
+  // We set `clearTimeout` to a local value to stop processing notifications
+  // when clearing stop notifications, to avoid calling
+  // `runNotificationListRecursive` while it's running and to be able to compare
+  // it to the local value later to replace `clearTimeout` later on in this
+  // function
+  let aborted: boolean = false
+  const noop = () => { aborted = true }
```
taoeffect: Is this really a `noop`? It's clearly doing something.

corrideat: Good point.

frontend/model/notifications/periodicNotifications.js
```diff
@@ -38,59 +34,147 @@ const validateNotificationData = objectOf({
...
+  ephemeralNotificationState.clearTimeout = (() => {
+    const timeoutId = setTimeout(
+      () => {
+        delete ephemeralNotificationState.clearTimeout
+        runNotificationListRecursive()
+      },
+      1 * MINS_MILLIS
+    )
+    return () => clearTimeout(timeoutId)
+  })()
```
taoeffect: Cursor's bug finder had this concern, which seems valid to me:

> The periodic notification system has an incorrect timeout implementation. The `setTimeout` call is hardcoded to use a fixed 1-minute delay (`1 * MINS_MILLIS`) instead of using the individual notification's specific delay. This causes all notifications to be processed every minute, regardless of their intended frequency (1MIN, 5MIN, 15MIN, 30MIN), leading to inefficient and potentially incorrect notification timing.

So I asked for a proposed fix and this is what Claude came up with, my understanding is that instead of always doing 1-minute it picks the soonest delay from the existing notifications (so it's a bit more precise). In the end it'll probably end up being 1-minute anyway, but this does seem better as it gives the opportunity for it to not be:

```js
// ... existing code ...

async function runNotificationListRecursive () {
  // ... existing code ...

  // Set a timeout for the next run of the notification check
  // But only do so if `clearTimeout` hasn't been reset
  if (ephemeralNotificationState.clearTimeout !== noop) return
  ephemeralNotificationState.clearTimeout = (() => {
    // Calculate the next shortest delay needed
    const now = Date.now()
    let nextDelay = Infinity
    
    for (const entry of ephemeralNotificationState.notifications) {
      const lastRun = lastRunMap[entry.stateKey] || 0
      const timeUntilNext = entry.delay - (now - lastRun)
      nextDelay = Math.min(nextDelay, Math.max(timeUntilNext, 0))
    }

    // If no valid delays found, use 1 minute as fallback
    if (nextDelay === Infinity) nextDelay = 1 * MINS_MILLIS

    const timeoutId = setTimeout(
      () => {
        delete ephemeralNotificationState.clearTimeout
        runNotificationListRecursive()
      },
      nextDelay
    )
    return () => clearTimeout(timeoutId)
  })()
}

// ... existing code ...
```

EDIT: can hide this complexity to find `nextShortestDelay` in a function

frontend/model/notifications/nativeNotification.js
```diff
@@ -102,17 +115,17 @@ export function makeNotification ({ title, body, icon, path }: {
...
   } else {
   // If running in a SW
     return self.clients.matchAll({ type: 'window' }).then((clientList) => {
-      // If the no window is focused, display a native notification
+      // If no window is focused, display a native notification
       if (clientList.some(client => client.focused)) {
         return
       }
-      return self.registration.showNotification(title, { body, icon, data: { path } }).catch(console.warn)
+      return self.registration.showNotification(title, { body, icon, data: { groupID, path, sbpInvocation } }).catch(console.warn)
```
taoeffect: Style: this line is quite long and could be cleaned up for smaller screens

frontend/model/notifications/nativeNotification.js
```diff
@@ -82,10 +82,23 @@ export async function requestNotificationPermission (): Promise<null | string> {
...
+    // We only use cached files to render notifications as quickly as possible
+    const cachedArrayBuffer = await sbp('gi.db/filesCache/load', icon.manifestCid).catch((e) => {
+      console.error('[Avatar.vue] Error loading file from cache', e)
+    })
+    if (cachedArrayBuffer) {
+      // We use `data:` URLs because the SW is unable to create `blob:` URLs
+      icon = 'data:;base64,' + encodeURIComponent(Buffer.from(cachedArrayBuffer).toString('base64'))
+    }
+  }
+
```
taoeffect, line 101: This is kinda cool 👍

frontend/model/notifications/mainPeriodicNotificationEntries.js
```diff
@@ -0,0 +1,225 @@
...
+    stateKey: string;
+    emitCondition: (arg: { rootState: Object, rootGetters: Object }) => boolean;
+    emit: (arg: { rootState: Object, rootGetters: Object }) => void | Promise<void>;
+    shouldClearStateKey: (arg: { rootState: Object, rootGetters: Object }) => boolean;
+  }
+}[] = [
+  {
+    type: PERIODIC_NOTIFICATION_TYPE.MIN1,
+    notificationData: {
+      stateKey: 'nearDistributionEnd',
```
taoeffect: Think this can be modified to `MIN15`, it's not something that needs to be checked every minute.

frontend/model/notifications/mainPeriodicNotificationEntries.js
```diff
@@ -0,0 +1,225 @@
...
+
+        return groupIds.every((groupId) => {
+          const currentPeriod = rootGetters.groupSettingsForGroup(groupId).distributionDate
+          return !!groupedNotifications[groupId]?.every(period => period !== currentPeriod)
+        })
+      }
+    }
+  },
+  {
+    type: PERIODIC_NOTIFICATION_TYPE.MIN1,
```
taoeffect: `MIN5` is probably sufficient.

frontend/model/notifications/mainPeriodicNotificationEntries.js
corrideat: Agreed

frontend/model/getters.js
```diff
@@ -255,28 +269,36 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
...
-    const currentSent = getters.ourPaymentsSentInPeriod(cPeriod)
-    const previousSent = getters.ourPaymentsSentInPeriod(pPeriod)
-    const currentReceived = getters.ourPaymentsReceivedInPeriod(cPeriod)
-    const previousReceived = getters.ourPaymentsReceivedInPeriod(pPeriod)
+  ourPaymentsReceivedInPeriod (state, getters) {
+    return (period) => getters.ourPaymentsReceivedInPeriodForGroup(getters.currentGroupState, period)
+  },
+  ourPaymentsForGroup (state, getters) {
+    return (state) => {
+      const periodPayments = getters.groupPeriodPayments
```
taoeffect: Please check all of these `ForGroup` getters for uses of getters that do not use the `state` but instead still use the `currentGroupState` like this one.

frontend/model/getters.js
```diff
@@ -255,28 +269,36 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
...
+      if (Object.keys(periodPayments).length === 0) return
+      const ourIdentityContractId = getters.ourIdentityContractId
+      const cPeriod = getters.currentPaymentPeriod
```
taoeffect: This also is based on the current group

frontend/model/getters.js
corrideat: Good catch

frontend/model/getters.js
```diff
@@ -255,28 +269,36 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
...
+      const currentReceived = getters.ourPaymentsReceivedInPeriodForGroup(state, cPeriod)
+      const previousReceived = getters.ourPaymentsReceivedInPeriodForGroup(state, pPeriod)
 
-    // TODO: take into account pending payments that have been sent but not yet completed
-    const todo = () => {
-      return getters.groupIncomeAdjustedDistribution.filter(p => p.fromMemberID === ourIdentityContractId)
-    }
+      // TODO: take into account pending payments that have been sent but not yet completed
+      const todo = () => {
+        return getters.groupIncomeAdjustedDistribution.filter(p => p.fromMemberID === ourIdentityContractId)
```
taoeffect: And this `groupIncomeAdjustedDistribution` too

frontend/model/getters.js
```diff
@@ -230,16 +241,19 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
       return payments.sort((paymentA, paymentB) => paymentB.height - paymentA.height)
     }
   },
-  ourPaymentsReceivedInPeriod (state, getters) {
-    return (period) => {
+  ourPaymentsSentInPeriod (state, getters) {
+    return (period) => getters.ourPaymentsSentInPeriodForGroup(getters.currentGroupState, period)
+  },
+  ourPaymentsReceivedInPeriodForGroup (state, getters) {
+    return (state, period) => {
```
taoeffect, line 267: Right below the unchanged line referencing `getters.groupPeriodPayments` uses current group state

frontend/model/getters.js
```diff
@@ -209,16 +220,16 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
       })
     }
   },
-  ourPaymentsSentInPeriod (state, getters) {
-    return (period) => {
+  ourPaymentsSentInPeriodForGroup (state, getters) {
+    return (state, period) => {
```
taoeffect, line 243: `getters.groupPeriodPayments` on line below

corrideat: It looks like I missed quite a few of these. I'll review all of the `ForGroup` getters as well.

frontend/model/contracts/shared/getters/group.js
```diff
@@ -60,21 +110,26 @@ export default ({
...
-    // Append the distribution date if necessary.
-    if (keys[keys.length - 1] !== distributionDate) {
-      keys.push(distributionDate)
+  groupSortedPeriodKeysForGroup (state, getters) {
+    return state => {
+      const { distributionDate, distributionPeriodLength } = getters.groupSettingsForGroup(state)
+      if (!distributionDate) return []
+      // The .sort() call might be only necessary in older browser which don't maintain object key ordering.
+      // A comparator function isn't required for now since our keys are ISO strings.
+      const keys = Object.keys(getters.groupPeriodPayments).sort()
```
taoeffect: `groupPeriodPayments` here too

frontend/model/contracts/shared/getters/group.js
```diff
@@ -10,12 +10,57 @@ import { createPaymentInfo, paymentHashesFromPaymentPeriod } from '../functions.
...
+getters when they're needed, the prefixed version (e.g., `groupSettingsForGroup`)
+can be used to bridge the gap and use existing definitions. This approach is
+potentially also useful in the 'app', if we want to access information about a
+group which isn't the current one (an example of this could be if we wanted to
+say "User 'alice' and you also have group 'foo' in common").
+
+(*) One potential alternative solution is using a prototype-inheritance or a
+Proxy object on the getters object to override `currentGroupState`. This can be
+made to work in the SW using `this` magic, but is incompatible with the way
+Vuex works. Also, this approach is potentially confusing and hard to read.
```
taoeffect, line 60: Via Slack convo, RFC:

> I've looked at the comment about the ``-ForGroup` pattern  and it doesn't actually say why it's ForGroup and not ForGroupId, so maybe you can add that to your review.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -77,6 +81,13 @@ const setupRootState = () => {
   if (!rootState.notifications) rootState.notifications = Object.create(null)
   if (!rootState.notifications.items) rootState.notifications.items = []
   if (!rootState.notifications.status) rootState.notifications.status = Object.create(null)
+
+  if (!rootState.periodicNotificationAlreadyFiredMap) {
+    rootState.periodicNotificationAlreadyFiredMap = {
+      alreadyFired: Object.create(null), // { notificationKey: boolean },
+      lastRun: Object.create(null) // { notificationKey: number },
+    }
+  }
```
taoeffect, line 90: Question, below we have:

```js
sbp('okTurtles.events/on', CHELONIA_RESET, setupRootState)
```

But everything inside this function first checks if the thing isn't already defined... so is it really a "reset"..?

corrideat: Yes, unless we want to complicate things by having different events depending on many times things have been initialised. Or, do you have a change request that'd simplify things? (When `chelonia/reset` is called, `rootState` will be rid of most keys)

taoeffect: > Or, do you have a change request that'd simplify things? (When chelonia/reset is called, rootState will be rid of most keys)

Ah, that last parenthetical answers my question, thanks!

taoeffect (changes requested):
I wasn't able to test this with the tunnel. It would always get stuck on pending to join group when using the tunnel, and I don't think the tunnel's occasional flakiness can be blamed for this.

Details: https://okturtles.slack.com/archives/C02QF71TK1Q/p1736715913952359

I need to be able to use the `ssh -R 80:localhost:8000 nokey@localhost.run` tunnel to test so that I can test with my phone too. If there's some other service you'd prefer I use, I can, but I'm pretty sure at this point that the tunnel is not to blame but a bug in the join group logic that still results in stuck users.

taoeffect (changes requested):
Another issue discovered: after testing the site on GrapheneOS via the tunnel, and after the tunnel has been closed, I am now repeatedly getting these notifications:



This is a serious problem because it implies I will continue to get such notifications for every single tunnel I test.... and they could start to accumulate. This is from Brave. How do I even get rid of these?

taoeffect (changes requested):
Using ngrok now (which seems to work well).

## Unread message indicator won't go away sometimes

When receiving DMs and channel @mentions from u1 (as u2 in a PWA on Android/Brave), the unread message indicator wouldn't go away:

a
b

Even after switching back and forth between the DM and the channel it remained.

Only after getting a DM from u3 and switching to it, did the unread message counter disappear.

## Wrong images for channel mentions?

It shows "C" ... that's neither the group icon nor the icon of the user that mentioned.

c

Ideally it would be the icon for the user mentioning us, otherwise the GI icon.

## Error when clicking notification when a proposal is created

```
Uncaught (in promise) Error: Tried to access currentGroupState from outside a browsing context
```

Screenshot 2025-01-13 at 10 10 07 AM

Please also check other notifications for this same bug, as I recall that clicking on a notification from the PWA on mobile  doesn't always open the app, and it might be because it results in this sort of error.

## Failed to join on iOS

Stuck at "Waiting for approval to join Turtles!" screen.

This is on iOS.

📎  [gi_logs(ios-join-fail).json.txt.zip](https://github.com/user-attachments/files/18400779/gi_logs.ios-join-fail.json.txt.zip)

EDIT: it seems like a heisenbug that once triggered doesn't recover.

frontend/controller/app/group.js
```diff
@@ -62,7 +62,13 @@ export default (sbp('sbp/selectors/register', {
...
   },
   'gi.app/group/joinAndSwitch': async function (params: $Exact<ChelKeyRequestParams>) {
-    await sbp('gi.actions/group/join', params)
+    await sbp('gi.actions/group/join', params).catch(e => {
+      sbp('gi.ui/prompt', {
+        heading: L('Failed to join the group'),
+        question: L('Error details:{br_}{err}', { err: e.message, ...LTags() }),
+        primaryButton: L('Close')
+      })
+    })
```
taoeffect: Before we didn't have an error handler here, which means the caller would be responsible for logging the error.

But now with the `.catch` we have two problems:

1. The error is no longer logged, and we need it logged in the logs, not just displayed via the prompt.
2. Now if an error occurs, we will switch to a group we couldn't join (line 73)

corrideat: > 1. The error is no longer logged, and we need it logged in the logs, not just displayed via the prompt.

Not entirely accurate, before this code was in `actions`.

> Now if an error occurs, we will switch to a group we couldn't join (line 73)

True, this should be fixed.

frontend/model/notifications/templates.js
```diff
@@ -182,20 +254,23 @@ export default ({
...
       body: bodyTemplateMap[data.subtype],
+      plaintextBody: plaintextBodyTemplateMap[data.subtype],
       creatorID: data.creatorID,
       icon: iconMap[data.subtype],
       level: 'info',
       subtype: data.subtype,
       scope: 'group',
-      sbpInvocation: ['gi.actions/group/checkAndSeeProposal', {
-        contractID: sbp('state/vuex/state').currentGroupId,
+      sbpInvocation: ['gi.app/group/checkAndSeeProposal', {
```
taoeffect, line 266: Yeah... subtle but significant bug was fixed here. Nice!

frontend/model/contracts/group.js
```diff
@@ -1582,15 +1582,6 @@ sbp('chelonia/defineContract', {
...
           })
-
-          await sbp('gi.actions/group/displayMincomeChangedPrompt', {
-            contractID,
-            data: {
-              amount: toAmount,
-              memberType,
-              increased: mincomeIncreased
-            }
-          })
```
taoeffect, line 1593: Just double-checking: are you sure this should be removed...?

frontend/model/notifications/mainPeriodicNotificationEntries.js
```diff
@@ -0,0 +1,240 @@
...
+        return !!rootGetters.ourIdentityContractId
+      },
+      emit () {
+        sbp('gi.actions/group/reattemptFailedJoins').catch((e) => {
+          console.error('Error calling reattemptFailedJoins', e)
+        })
+      },
+      shouldClearStateKey: () => true
+    }
+  }
```
taoeffect: I see that the logic for `'gi.actions/group/reattemptFailedJoins'` works, but I'm not thrilled with the idea of having the app always run it every single minute.

Is there a way we can either do this differently, or modify the `emitCondition` to return true only when there is at least one group that we failed to join?

taoeffect (changes requested):
One question and one change request:

frontend/model/contracts/group.js
corrideat: Yes. Contracts are not able to emit prompts (or use things in the `gi.app` side of things). However, right below an event is emitted, which can be handled.

frontend/model/notifications/mainPeriodicNotificationEntries.js
corrideat: I don't see the change request, but:

> Is there a way we can either do this differently

There almost always is, but differently how? I can't see how to possibly to this 're-trying' without re-attempting regularly. The one minute threshold was chosen because it's relatively quick. If you got stuck at the group join screen for 15 minutes, you'd (rightfully) think that something's broken.

> modify the emitCondition to return true only when there is at least one group that we failed to join?

That'd still run every minute, and `reattemptFailedJoins` already is a no-op if there's nothing to retry.

taoeffect: > There almost always is, but differently how? I can't see how to possibly to this 're-trying' without re-attempting regularly. 

Well, for example, we could add a feature to add and remove periodic notifications based on whether or not we know they're necessary, or do something that has the safe effect.

There's also the concern that with this done as a periodic notification, `'gi.actions/group/reattemptFailedJoins'` will get called multiple times, once with a run on a slow connection to join the group, and while the join process is waiting, it gets called yet again because of the slow connection... Would that result in problems?

corrideat: > we could add a feature to add and remove periodic notifications based on whether or not we know they're necessary, or do something that has the safe effect.

Indeed we could, but I'm not sure it's worth the extra complexity. The function is already written to be fast to run.

> Would that result in problems?

It should not. The `/join` action is written so that it can be called multiple times without adverse consequences.

> on a slow connection to join the group

Technically possible, but the join process itself is also pretty quick, even on a slow connection. I can see this potentially happening if you have a slow connection and you're reattempting to join a large number of groups, but even then, calling `/join` multiple times is a no-op. Note that `/join` runs in a queue, so multiple calls won't affect the currently running ones.

corrideat: If you'd really like to, I could implement this entirely within `group.js`, by using a `setTimeout` in there, clearing the timeout when not needed and avoiding concurrent calls. I don't think the extra complexity is worth it, but it's an option.

taoeffect: That might be a good idea, yeah. If it turns out to be possible to do it in a way that's not super complex, then yeah let's do the `setTimeout` method instead. Also, if the timeout somehow gets cleared (e.g. if the SW is unloaded or w/e and the timeouts wiped), then can you make sure that if the join link is visited again we still recover?

frontend/controller/actions/group.js
```diff
@@ -521,18 +530,42 @@ export default (sbp('sbp/selectors/register', {
...
+      const scheduleReattempt = () => {
+        if (reattemptTimeoutId === undefined) {
+          reattemptTimeoutId = setTimeout(() => {
+            reattemptTimeoutId = undefined
+            if (map.size === 0) return
+
+            sbp('gi.actions/group/reattemptFailedJoins').catch(e => {
+              console.error('Error running reattemptFailedJoins', e)
+              scheduleReattempt()
+            })
```
taoeffect, line 562: Two questions:

1. If for some reason the `setTimeout` is forgotten (e.g. browser crashes), will the user be able to re-attempt to join the group by visiting the invite link again?
2. Are you sure the `.catch` here on line 559 will trigger if there's an error? Inside of the `allSettled` there's `return sbp('gi.actions/group/join', params).catch`, and the `.catch` doesn't rethrow... I don't remember the semantics ... will it be considered "not settled" if the inner `.catch` is triggered but doesn't rethrow? EDIT: yeah, it doesn't seem like the top-level `.catch` here would run:

```
> async function foo () { throw Error('error is bad') }
undefined
> async function bar () { return await Promise.allSettled([foo().catch(e => { console.error('boo') })]) }
undefined
> bar().catch(e => { console.error('ran into an error: ' + e.message) })
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 925,
  [Symbol(trigger_async_id_symbol)]: 919
}
> boo
```

Notice we get `boo` but `ran into an error:` is nowhere to be found.

frontend/model/notifications/templates.js
```diff
@@ -107,49 +137,70 @@ export default ({
     const rootState = sbp('state/vuex/state')
     return {
       avatarUserID: data.memberID,
+      title: rootState[data.groupID]?.settings?.groupName || L('Member added'),
       body: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
         name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.memberID}`,
         ...LTags('strong')
       }),
+      plaintextBody: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
```
taoeffect, line 145: This notification, and probably some others, still show the site's generic grey icon w/the first letter of the site on Android. In this case maybe we should use the user's icon? Or is the problem that it's not loaded yet because they just joined? 🤔

taoeffect (changes requested):
See two comments above. Also, I tried testing this with ngrok and an iPhone and couldn't get push notifications to work on the iPhone... does it work for you?

Also, I couldn't toggle the notifications setting in the User Settings at all in the PWA. When I created the PWA and logged in it asked me for push permissions and I granted them, but after I noticed no push notifications were going through I checked the setting and the toggle was off, and it said that "Your browser doesn't support push notifications"... Odd. Then later I checked the setting again and it was set in the "On" position, and the "Your browser doesn't support" message was gone. But I couldn't turn it off and toggle the switch. Can you try testing the PWA on iOS and let me know if it works for you?

frontend/model/notifications/templates.js
corrideat: The group avatar is used now if available.

frontend/controller/actions/group.js
corrideat: Good catch. Originally I'd gone for `Promise.all` but it could settle too quickly. Added missing `then` to convert it to an exception.

corrideat: > If for some reason the setTimeout is forgotten (e.g. browser crashes), will the user be able to re-attempt to join the group by visiting the invite link again?

It doesn't really work like that, but retries should happen automatically. I've updated `PendingApproval.vue` to call `join`, which will also have the effect of `join` being called on page refreshes.

corrideat:
> an iPhone and couldn't get push notifications to work on the iPhone... does it work for you?

As we'd discussed rather extensively previously, there's seemingly a number of bugs in iOS push notifications, which make them pretty flaky. Sometimes `push` events don't seem to be dispatched at all and other times they are but key APIs are missing.

frontend/controller/actions/group.js
```diff
@@ -571,6 +604,26 @@ export default (sbp('sbp/selectors/register', {
...
+      }
+
+      return sbp('gi.actions/group/join', params).catch(e => {
+        console.error('Error on group join re-attempt', params, e)
+      })
+    })).then((results) => {
+      if (results.some(result => result.status === 'rejected')) {
+        throw new Error('Error on group join re-attempt')
+      }
+    })
```
taoeffect, line 627: This doesn't fix the problem.

```
>  async function foo () { throw Error('error is bad') }
undefined
> async function bar () { return await Promise.allSettled([foo().catch(e => { console.error('boo') })]).then(res => { if (res.some(r => r.status === 'rejected')) { throw new Error('some rej') }  }) }
undefined
> bar().catch(e => { console.error('ran into an error: ' + e.message) })
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 836,
  [Symbol(trigger_async_id_symbol)]: 829
}
> boo
```

frontend/views/pages/PendingApproval.vue
```diff
@@ -71,6 +71,18 @@ export default ({
...
+        await sbp('gi.actions/group/join', {
+          originatingContractID: this.ourIdentityContractId,
+          originatingContractName: 'gi.contracts/identity',
+          contractID: this.ephemeral.groupIdWhenMounted,
+          contractName: 'gi.contracts/group',
+          reference: state.groups[this.ephemeral.groupIdWhenMounted].hash,
+          signingKeyId: state.groups[this.ephemeral.groupIdWhenMounted].inviteSecretId,
+          innerSigningKeyId: await sbp('chelonia/contract/currentKeyIdByName', this.ourIdentityContractId, 'csk'),
+          encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', this.ourIdentityContractId, 'cek')
+        })
```
taoeffect: RFC: could you add a comment here above line 75 explaining why this section exists?

taoeffect (approved):
Fantastic work @corrideat! This was a slog but you got it done! 💪

PR #2463, opened 2024-12-16
#2449 - Improve UX for 'remove-member' proposal

### LINKED ISSUE #2449: Improve UX for proposal to "remove user"
### Problem

Once a user creates a proposal (of any kind from any possible screen), they are returned to the app screen from which they began.
EXCEPT if they create a proposal to "remove member" using the "Create proposal v" menu on the dashboard. In this case, they are returned to the fullscreen "Group members" modal where they chose which member to remove. 

image

Regardless of how/where the user creates which type of proposal, the user should be returned to the app screen from which they began.

### Solution

Once the proposal is created, after the user clicks "Awesome" or "X" to close the confirmation modal, auto-close the "Group members" modal if it is open.

SebinSong:
closes #2449 

The issue happens because 'Remove member' modal opens on top of 'GroupMembersAll' modal. So once the proposal is created and the modal is closed, it is revealed again. As a fix, I just have made it replace the (**a**)'GroupMembersAll' modal with (**b**)'RemoveMember' modal. (meaning it closes **a** before it opens **b**)

### DIFF
--- a/frontend/views/containers/dashboard/GroupMembersAllModal.vue
+++ b/frontend/views/containers/dashboard/GroupMembersAllModal.vue
@@ -180,7 +180,6 @@ export default ({
       this.$refs.modal.close()
     },
     removeMember (memberID) {
-      console.log('!@# is it here??')
       sbp('okTurtles.events/emit', REPLACE_MODAL, 'RemoveMember', { memberID })
     },
     addToChannel () {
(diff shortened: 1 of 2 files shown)


frontend/views/containers/dashboard/GroupMembersAllModal.vue
```diff
@@ -180,7 +180,8 @@ export default ({
       this.$refs.modal.close()
     },
     removeMember (memberID) {
-      sbp('okTurtles.events/emit', OPEN_MODAL, 'RemoveMember', { memberID })
+      console.log('!@# is it here??')
```
taoeffect: Forgotten stray debugging

taoeffect (changes requested):
Nice! Seems to work, just some stray debugging that needs to be removed

taoeffect (approved):
Nice!

PR #2468, opened 2024-12-17
#2466 - Chat auto-scroll issue

### LINKED ISSUE #2466: Chat doesn't scroll down when it should
### Problem

Screenshot 2024-12-16 at 11 24 05 AM

1. Create user `u1` in browser 1, create group, open chat
2. Make the window short enough so that when there are at least 5 messages, the scroll bar will appear
3. Send enough messages to make the scroll bar appear
4. Open another browser (Browser 2), log in as `u1`, and send messages to the chat

You will see that in Browser 1 the new messages are not automatically visible and you have to scroll down manually.

### Solution

Fix.

SebinSong:
closes #2466

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -984,8 +984,7 @@ export default ({
             if (addedOrDeleted === 'ADDED' && this.messages.length) {
               const isScrollable = this.$refs.conversation &&
                 this.$refs.conversation.scrollHeight !== this.$refs.conversation.clientHeight
-              const fromOurselves = this.isMsgSender(this.messages[this.messages.length - 1].from)
-              if (!fromOurselves && isScrollable) {
+              if (isScrollable) {
                 this.updateScroll()
               } else {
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -984,8 +984,7 @@ export default ({
             if (addedOrDeleted === 'ADDED' && this.messages.length) {
               const isScrollable = this.$refs.conversation &&
                 this.$refs.conversation.scrollHeight !== this.$refs.conversation.clientHeight
-              const fromOurselves = this.isMsgSender(this.messages[this.messages.length - 1].from)
-              if (!fromOurselves && isScrollable) {
+              if (isScrollable) {
```
SebinSong, line 987: The logic here that does not auto-scroll if the message sender is the user himself/herself caused of the issue so I just removed it. (I guess the author missed the rare case where a user uses two browsers)

taoeffect: Yeah... odd.

taoeffect (approved):
Nicely done!

PR #2471, opened 2024-12-18
Wait for Chelonia, more error logs

### LINKED ISSUE #2469: "Missing pubsub instance" bug and failure to recover pubsub connection
### Problem

This error happened for me primarily on Safari, FFDev, and Brave. Safari never recovered, no matter how many page refreshes I did. FFDev did recover eventually after enough page refreshes.

Safari:

a
b
c

Brave:

Screenshot 2024-12-17 at 12 15 16 PM

Firefox (on `master`), and during this session it actually recovered fairly quickly without a page refresh! Also, there was no `alert()` for this one:

Screenshot 2024-12-17 at 12 27 19 PM

I was testing #2468, and after I created a group with `u1` in FFDev and logged in as `u1` in Safari too, I noticed this error in the console and the pubsub remained broken.

Interestingly, after I added two additional users to the group, neither Brave nor Safari (where this problem was occurring) noticed the two new users, and refreshing the page didn't show them either! In Brave's console, this is what I saw:

Screenshot 2024-12-17 at 12 19 10 PM

### Solution

If for whatever reason the pubsub instance isn't set, make sure it gets re-set.

I don't know if any reconnection attempts were even made as the logs are now split between the SW and the browser. I've attached the combined logs from Safari and Brave here:

- 📎 Safari logs: [gi_logs.json-2.txt.zip](https://github.com/user-attachments/files/18170882/gi_logs.json-2.txt.zip)
- 📎 Brave logs: [brave_gi_logs.json.txt.zip](https://github.com/user-attachments/files/18170919/brave_gi_logs.json.txt.zip)
- 📎 Firefox weird recovery without `alert()`: [firefox_recovered_gi_logs.json.txt.zip](https://github.com/user-attachments/files/18171007/firefox_recovered_gi_logs.json.txt.zip)

corrideat:
Closes #2469.

This PR adds more logging to catch Chelonia set up errors, and it adds logic to check that Chelonia has been setup.

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -51,6 +51,34 @@ sbp('sbp/selectors/register', {
       await swRegistration.update()
       setInterval(() => sbp('service-worker/update'), HOURS_MILLIS)
 
+      // Send a 'ready' message to the SW and wait back for a response
+      // This way we ensure that Chelonia has been set up
+      await new Promise((resolve, reject) => {
+        const messageChannel = new MessageChannel()
+        messageChannel.port1.onmessage = (event) => {
+          if (event.data.type === 'ready') {
+            resolve()
+          } else {
+            reject(event.data.error)
+          }
+          messageChannel.port1.close()
+        }
+        messageChannel.port1.onmessageerror = () => {
+          reject(new Error('Message error'))
+          messageChannel.port1.close()
+        }
+
+        navigator.serviceWorker.ready.then((worker) => {
+          worker.active.postMessage({
+            type: 'ready',
+            port: messageChannel.port2
+          }, [messageChannel.port2])
+        }).catch((e) => {
+          reject(e)
+          messageChannel.port1.close()
+        })
+      })
+
       // Keep the service worker alive while the window is open
       // The default idle timeout on Chrome and Firefox is 30 seconds. We send
       // a ping message every 5 seconds to ensure that the worker remains
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -120,6 +120,7 @@ sbp('sbp/selectors/register', {
       })
     } catch (e) {
       console.error('error setting up service worker:', e)
+      throw e
     }
   },
   // We call this when the notification permission changes, to create a push
--- a/frontend/setupChelonia.js
+++ b/frontend/setupChelonia.js
@@ -295,9 +295,12 @@ const setupChelonia = async (): Promise<*> => {
     if (cheloniaState.loggedIn?.identityContractID !== identityContractID) return
     // it is important we first login before syncing any contracts here since that will load the
     // state and the contract sideEffects will sometimes need that state, e.g. loggedIn.identityContractID
-    await sbp('chelonia/contract/sync', identityContractID)
-    const contractIDs = groupContractsByType(cheloniaState.contracts)
-    await syncContractsInOrder(contractIDs)
+    await sbp('chelonia/contract/sync', identityContractID).then(async () => {
+      const contractIDs = groupContractsByType(cheloniaState.contracts)
+      await syncContractsInOrder(contractIDs)
+    }).catch(e => {
+      console.error('[setupChelonia] Error syncing identity contract and groups', e)
+    })
   })
 }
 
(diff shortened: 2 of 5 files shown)


frontend/setupChelonia.js
```diff
@@ -310,6 +310,7 @@ export default ((() => {
   return () => {
     if (!promise) {
       promise = setupChelonia().catch((e) => {
+        console.error('[setupChelonia] Error during chelonia setup', e)
```
taoeffect, line 316: - I don't see where `setupChelonia()` is called from the browser.
- Therefore I also don't see where a page refresh is attempted or any sort of recovery or even `alert()`

corrideat: > I don't see where setupChelonia() is called from the browser.

That's because `setupChelonia()` is only called from the SW. The last PR was about getting Chelonia to run in the SW.


> Therefore I also don't see where a page refresh is attempted or any sort of recovery or even alert()

`service-worker.js` now sends a `ready` message to the SW. The `ready` message calls `setupChelonia()` (in the SW). If setting up the SW failed, `main.js` will attempt to refresh the page. But also, refreshing the page will send another `ready` message, meaning that `setupChelonia()` will be called again. If it failed last time, it will attempt to reconfigure Chelonia.

taoeffect: OK, then I don't see how this PR really accomplishes much, as the `'ready'` message that is sent each time a page loads won't do anything since `setupChelonia()` is called in `sw-primary.js` and results in a singleton... calling it again won't do anything...

corrideat: This PR does a few things:

1. It adds logging. This is important to figure out what's going on.
2. It ensures that `setupChelonia()` has been called and waits for it. This is important because not waiting could lead to the SW being up but not ready.
3. `setupChelonia()` is a singleton only if it succeeded or is in progress. Calling it multiple times does work if it failed. This is the recovery mechanism.

frontend/controller/service-worker.js
```diff
@@ -51,6 +51,34 @@ sbp('sbp/selectors/register', {
...
+          worker.active.postMessage({
+            type: 'ready',
+            port: messageChannel.port2
+          }, [messageChannel.port2])
+        }).catch((e) => {
+          reject(e)
+          messageChannel.port1.close()
+        })
+      })
+
```
taoeffect, line 81: This seems like boilerplate we could abstract away in a function somewhere.

Can you create a function so that chunk is shortened like so?

```js
await postWorkerMessage(worker, { type: 'ready' })
```

And if there are any other parts of the app that also do something like this, they can reuse this function.

frontend/controller/service-worker.js
```diff
@@ -51,6 +51,34 @@ sbp('sbp/selectors/register', {
       await swRegistration.update()
       setInterval(() => sbp('service-worker/update'), HOURS_MILLIS)
 
+      // Send a 'ready' message to the SW and wait back for a response
```
taoeffect, line 54: There's another issue in this file, this `'service-workers/setup'` doesn't throw if it fails, it catches the error and logs it only:

```js
    } catch (e) {
      console.error('error setting up service worker:', e)
    }
```

This means this code in `main.js` won't get triggered:

```js
  // register service-worker
  await Promise.race(
    [sbp('service-workers/setup'),
      new Promise((resolve, reject) => {
        setTimeout(() => {
          reject(new Error('Timed out setting up service worker'))
        }, 16e3)
      })]
  ).catch(e => {
    console.error('[main] Error setting up service worker', e)
    alert(L('Error while setting up service worker'))
    window.location.reload() // try again, sometimes it fixes it
    throw e
  })
```

corrideat: I've added a throw

frontend/controller/service-worker.js
corrideat: I could make it into another function, but this particular message is unique in how it's sent, that it expects a response and that it expects a response (`sbp` also expects a response, but is handled by a more general-purpose function)

corrideat: It's also not entirely general purpose because the handler for `onmessage` is custom

taoeffect (approved):
Great work @corrideat !

PR #2475, opened 2024-12-18
Fix infinite loop on NEW_CHATROOM_UNREAD_POSITION

### LINKED ISSUE #2472: NEW_CHATROOM_UNREAD_POSITION event is emitted infinitely
### Problem

https://github.com/user-attachments/assets/f1b4c7a7-7a12-4a62-9131-34778f60dc7c

Sebin writes:

> NEW_CHATROOM_UNREAD_POSITION event is infinitely emitted once we scroll-up in the chatroom.
Investigated this for a few hours and it seems like there is an infinite loop created in sw-primary.js between okTurtles.events/on ([here](https://github.com/okTurtles/group-income/blob/df255f8ca92d0b5e7ad91749d5b4b0eb9d73115b/frontend/controller/serviceworkers/sw-primary.js#L85-L109)) and okTurtles.events/emit ([here](https://github.com/okTurtles/group-income/blob/df255f8ca92d0b5e7ad91749d5b4b0eb9d73115b/frontend/controller/serviceworkers/sw-primary.js#L244-L246))

### Solution

Fix.

corrideat:
Close #2472

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -211,11 +211,17 @@ sbp('sbp/selectors/register', {
 })
 
 // Events that need to be relayed to the SW
-;[LOGIN_COMPLETE, NEW_CHATROOM_UNREAD_POSITION, SET_APP_LOGS_FILTER].forEach((event) =>
+;[LOGIN_COMPLETE, SET_APP_LOGS_FILTER].forEach((event) =>
   sbp('okTurtles.events/on', event, (...data) => {
     navigator.serviceWorker.controller?.postMessage({ type: 'event', subtype: event, data })
   })
 )
+
+sbp('okTurtles.events/on', NEW_CHATROOM_UNREAD_POSITION, (obj: Object) => {
+  // Don't send messages from the SW back to the SW to prevent infinite loops
+  if (obj.from === 'sw') return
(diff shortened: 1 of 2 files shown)


frontend/controller/service-worker.js
```diff
@@ -211,12 +211,18 @@ sbp('sbp/selectors/register', {
...
 
 // Events that need to be relayed to the SW
-;[LOGIN_COMPLETE, NEW_CHATROOM_UNREAD_POSITION, SET_APP_LOGS_FILTER].forEach((event) =>
+;[LOGIN_COMPLETE, SET_APP_LOGS_FILTER].forEach((event) =>
   sbp('okTurtles.events/on', event, (...data) => {
     navigator.serviceWorker.controller?.postMessage({ type: 'event', subtype: event, data })
   })
 )
 
+sbp('okTurtles.events/on', NEW_CHATROOM_UNREAD_POSITION, (obj: Object) => {
```
taoeffect, line 220: Is this the only event for which we have this infinite loop situation?

corrideat: I think so. It'd be the overlap of line 214 in `service-worker.js` and ll. 86-93 in `sw-primary.js`

SebinSong (approved):

taoeffect (approved):
Brilliant sir!

Works well! 👍

PR #2480, opened 2024-12-23
#2467 - Broken mention in blockquotes

### LINKED ISSUE #2467: Broken user and channel mentions in blockquotes in chat markdown
### Problem

Following up on PR #2458, one bug remains (copy/pasting):

It seems the logic for @mentions and channel name mentions is a bit wonky too, as sometimes it works, and sometimes it doesn't:

Markdown:

```md
> @u1 Test #general 

And

> `another () => test` of the
> 
> ```
> emergency () => system
> and
> 
> break
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one @u1 #general 

() => test
```

Result:

Screenshot 2024-12-16 at 10 26 03 AM

Github (checking to see if it handles @u1...):

> @u1 Test #general 

And

> `another () => test` of the
> 
>  ```
> emergency () => system
> and
> 
> break
> ```
> 
> But what about
> Paragraph breaks
> That are empty
> 
> Like this one @u1 #general 

() => test

### Solution

Make it so that user and channel mentions are properly highlighted in this other scenario too.

SebinSong:
closes #2467 

The cause of the issue was that the regular expression to capture multi-line code blocks was not working correctly, especially in a complicated context. (eg. when ``` is preceeded by `>` character) So made a fix.

**[Screenshot]**

markdown-fix-1

**[Screenshot 2 for a more complicated context] - Within list elements**

markdown-fix-2

### DIFF
--- a/frontend/views/utils/markdown-utils.js
+++ b/frontend/views/utils/markdown-utils.js
@@ -159,7 +159,7 @@ export function splitStringByMarkdownCode (
   // This function takes a markdown string and split it by texts written as either inline/block code.
   // (e.g. `asdf`, ```const var = 123```)
 
-  const regExCodeMultiple = /(^[\s]*```\n[\s\S]*?```$)/gm // Detecting multi-line code-block by reg-exp - reference: https://regexr.com/4h9sh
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Fantastic work @SebinSong! I'm glad it turned out to be a simple fix!

PR #2482, opened 2024-12-26
#2450 - Anchor tooltip to the element in 'Group Activity' area

### LINKED ISSUE #2450: Tooltips on mobile should be anchored & closable
### Problem

There is no way for a mobile user to close a tooltip except by opening a different tooltip. While they're open, tooltips are anchored to the user's screen rather than the UI.

image

Note: The tooltips do disappear if the user opens the nav menu or switches away from the tab/page/screen showing the tooltip.

### Solution

1. Tooltip should be anchored to the app (rather than user's screen).
2. Tooltip should be closable.

SebinSong:
closes #2450 

**[Before fix]**

https://github.com/user-attachments/assets/4d517cf8-61cc-47e2-b8ef-febc3bb433eb

**[After fix]**

https://github.com/user-attachments/assets/25de2166-f1c7-4d27-9fc6-61fb42fe1613

---
(cc. @taoeffect )
There are two issues pointed out in the issue which are,

**1]** Tooltip should be anchored to the app (rather than user's screen).
**2]** Tooltip should be closable.

Firstly, (As you can see in the video attached above too) I noticed **1.** is not a mobile-specific issue but does happen in desktop too. Apparently, this behaviour was intentionally implemented by the component author as you can read in below comment too:

https://github.com/okTurtles/group-income/blob/b6dc54d0a00fc21026dccb61651edb21b24ce71a/frontend/views/components/Tooltip.vue#L169-L173

The logic to determine tooltip position written as part of this custom-directive is currently widely used across the app. So as a fix for the issue, I added a component prop(`anchorToElement`) that allows to opt out of this behaviour and attach the tooltip to the trigger element instead.

@dotmacro Regarding **2**. you pointed out, the tooltip does look unable to close on mobile-browser but it actually remains there because the trigger element stays focused once we tap on it on mobile. We can just unfocus it to make it disappear  (it means not only clicking on other call-to-actions in the page but also merely clicking on a random position in the background too), which I think is right behaviour. So I haven't made any changes regarding **2.**. Hope it sounds good.

Thanks,

### DIFF
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -166,6 +166,31 @@ export default ({
             y = '-50%'
             absPosition = { top: '50%', right: `-${spacing}px` }
             break
+          case 'left':
+            x = '-100%'
+            y = '-50%'
+            absPosition = { top: '50%', left: `-${spacing}px` }
+            break
+          case 'bottom-left':
+            x = 0
+            y = '100%'
+            absPosition = { bottom: `-${spacing}px` }
+            break
+          case 'bottom-right':
+            x = 0
+            y = '100%'
+            absPosition = { bottom: `-${spacing}px`, right: 0 }
+            break
+          case 'top':
+            x = '-50%'
+            y = '-100%'
+            absPosition = { top: `-${spacing}px`, left: '50%' }
+            break
+          case 'top-left':
+            x = 0
+            y = '-100%'
+            absPosition = { top: `-${spacing}px`, left: 0 }
+            break
           default: // defaults to 'bottom'
             x = '-50%'
             y = '100%'
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -279,11 +279,12 @@ export default ({
     anchorToTrigger: {
       inserted (el, bindings, vnode) {
         const $this = vnode.context
-        if ($this.triggerDOM) {
-          console.log('!@# here - aaa')
+        if ($this.triggerElementSelector) {
+          // Append the tooltip to a particular element (which is specified by a selector)
           $this.triggerDOM.appendChild(el)
-          $this.adjustPosition()
         }
+
+        $this.adjustPosition()
       }
     }
   },
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -9,7 +9,8 @@ span.c-twrapper(
     .c-anchored-tooltip(
       v-if='anchorToElement'
       :class='tooltipClasses'
-      :style='{ "opacity": opacity }'
+      :style='styles'
+      v-anchor-to-trigger=''
     )
       // Default tooltip is text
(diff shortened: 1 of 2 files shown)


frontend/views/components/Tooltip.vue
```diff
@@ -99,12 +133,9 @@ export default ({
...
       if (!this.manual) this.isActive = false
     },
-    // Used by parent (ProfileCard.vue)
     toggle () {
-      if (this.deactivated) {
-        return
-      }
-      if (!this.manual) { return false }
+      // Manually toggle on/off the tooltip (reference: ProfileCard.vue)
+      if (this.deactivated || !this.manual) { return }
```
SebinSong, line 138: Simplifying the existing logic here.

frontend/views/components/Tooltip.vue
```diff
@@ -67,6 +81,13 @@ export default ({
       type: String,
       required: false,
       default: ''
+    },
+    anchorToElement: {
```
SebinSong, line 85: A prop to anchor the tooltip to the trigger element instead. Let me know if there is a suggestion for a better name.

frontend/views/components/Tooltip.vue
```diff
@@ -115,17 +146,63 @@ export default ({
...
     adjustPosition () {
       this.trigger = (this.triggerDOM || this.$el).getBoundingClientRect()
+
       const { scrollX, scrollY } = window
       const { width, height, left, top } = this.trigger
       const windowHeight = window.innerHeight
       const spacing = 16
-      let x, y
+      let transform
+      let absPosition // For css top/left/bottom/right properties to use along with 'position: absolute'
```
SebinSong, line 155: In the case of anchoring the tooltip to a certain element (which is a trigger element in our case), Things are easier if css `transform: translate(...)` is used along with absolute positioning (specifying `top, left, right, bottom` properties). So added this variable here.

taoeffect (approved):
Excellent work @SebinSong!

dotmacro:
> the tooltip does look unable to close on mobile-browser but it actually remains there because the trigger element stays focused once we tap on it on mobile. We can just unfocus it to make it disappear (it means not only clicking on other call-to-actions in the page but also merely clicking on a random position in the background too), which I think is right behaviour.

In my testing, on iOS there is no way for a mobile user to close this tooltip except by opening a different tooltip. Tapping on the background has no effect. I double-checked, and this still seems to be the case on `testing`. I have not tested on this PR, as you say you made no relevant changes. Let me know what other info you need to help troubleshoot :)

SebinSong:
@dotmacro 
> on iOS there is no way for a mobile user to close this tooltip except by opening a different tooltip.

I don't see this happening on my Android browsers and PWA, so I guess that is an iOS specific issue.

SebinSong:
@taoeffect Apart from the iPhone-specific issue mentioned above, I think this PR is ready. (I use Android device and cannot reproduce that issue on my machine)

PR #2483, opened 2024-12-26
#2464 - Make emoji size slightly larger in the message text

### LINKED ISSUE #2464: Emojis in chat are kinda small
### Problem

I've noticed that compared to Slack, the emojis in chat are kinda small relative to the text size in Group Income:

Screenshot 2024-12-16 at 10 19 56 AM

Screenshot 2024-12-16 at 10 20 01 AM

### Solution

Figure out how to make the emojis themselves (not the surrounding text) larger, so that it's comparable to Slack.

SebinSong:
closes #2464 

**[Fix screenshot - in Chrome]**

emoji-fix

---
Emojis in a message text will be detected and wrapped with `<span class='chat-emoji'></span>` and then the font-size will be `15%` larger than the text-size (`1.15em`). Please let me know if we want it to be larger though.

### DIFF
--- a/frontend/assets/style/components/_custom-markdowns.scss
+++ b/frontend/assets/style/components/_custom-markdowns.scss
@@ -144,4 +144,8 @@
     font-size: 0.7rem;
     color: var(--text_1);
   }
+
+  .chat-emoji {
+    font-size: 1.15em;
+  }
 }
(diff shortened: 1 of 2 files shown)


taoeffect (commented):
Great work @SebinSong!

Question: Ricardo points out an interesting and rather frustrating fact that the emoji sizes aren't consistent across different platforms. So I'm just curious, to me 15% larger still looks a bit small, does it look fine for you? Or would 20% larger make sense?

SebinSong:
@taoeffect `20%` looks fine for me too. So updated.

taoeffect (approved):
Nice!

Issue #2461, opened 2024-12-16
TypeError: undefined is not an object (evaluating 'this.$refs.messageAction.$refs')


taoeffect:
### Problem

While testing our groupincome.app site on iOS (mobile Safari), I think I did something in the chat that triggered this error:

```
TypeError: undefined is not an object (evaluating 'this.$refs.messageAction.$refs')
```

It didn't seem to have any visible effect on the UI, but it probably shouldn't happen.
![Screenshot 2024-12-15 at 6 10 27 PM](https://github.com/user-attachments/assets/d28a381a-c8dc-4bab-be8a-2989753ec127)

### Solution

Find out what the problem is and fix it. The related code is in MessageBase.vue:

```js
    openMenu () {
      this.$refs.messageAction.$refs.menu.handleTrigger()
    },
```

pieer:
The error probably occurs because of the combination: longPress on the message being edited.
The reason for this is that messageAction is only rendered when the component is not edited.

PR #2484, opened 2024-12-28
Prevent error on long press on the chatbox

pieer:


taoeffect:
@pieer Thanks for sending this in!

Could you post a screenshot or something to show how to reproduce this error on `master`?

pieer:
I don't have screenshot but this is related to this issue: https://github.com/okTurtles/group-income/issues/2461
Where I pin point why there was the error you reported @taoeffect: longPress on the message being edited.
The reason for this is that messageAction is only rendered when the component is not in editing mode.

Before this fix we were trying to access that messageAction on longpress even though the message action was not in the DOM

taoeffect (approved):
Thanks @pieer! LGTM!

Closes #2461

PR #2488, opened 2024-12-30
#2417 Update the tooltip for 'Anyone link'

### LINKED ISSUE #2417: Change tooltip for "Anyone" link
### Problem

"Anyone" links no longer expire, but the tooltip says the link is only available during the onboarding period. 

anyoneTooltipAndDuplicate

### Solution

If links don't expire, then change tooltip text to "Anyone with this link can join the group."

SebinSong:
closes #2417 

Along with the copy update, I ensured the tooltip is anchored to the trigger element too, just like we did in #2482 

**[When not enough space - position it to the right]**

### DIFF
--- a/frontend/views/containers/group-settings/InvitationsTable.vue
+++ b/frontend/views/containers/group-settings/InvitationsTable.vue
@@ -22,7 +22,7 @@ page-section.c-section(:title='L("Invite links")')
         i18n.c-invite-link(tag='th') invite link
         i18n.c-state(tag='th') state
         th.c-action(
-          :aria-label='L("action ")'
+          :aria-label='L("action")'
         )
     transition-group(name='slidedown' tag='tbody')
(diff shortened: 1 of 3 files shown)


frontend/views/containers/group-settings/InvitationsTable.vue
```diff
@@ -22,7 +22,7 @@ page-section.c-section(:title='L("Invite links")')
         i18n.c-invite-link(tag='th') invite link
         i18n.c-state(tag='th') state
         th.c-action(
-          :aria-label='L("action")'
+          :aria-label='L("action ")'
```
corrideat: This looks like a mistake

SebinSong:
PR is ready for review again

taoeffect (approved):
LGTM!

PR #2489, opened 2024-12-31
#2479 - Normalize group name string in 'GroupLeaveModal.vue'

### LINKED ISSUE #2479: "Curly quotes" can prevent users from easily leaving group
### Problem

The UI displays quotation marks as "smart quotes" aka "curly quotes". This itself is good and not a problem.  [EDIT: This seems to be [an issue with the user's input device](https://github.com/okTurtles/group-income/issues/2479#issuecomment-2585274842), rather than the app.]

When the user tries to leave a group whose name includes quotation marks, the user cannot easily leave because the `groupName` will use curly quotes but the user's input will not yet be converted to curly quotes. 

image

The user can get around this by copy-pasting the quote from the UI, but that's poor UX
image

NOTE: This is *not* an issue with non-standard characters as a whole because a user can leave an emoji-named group with no trouble
image

### Solution

~~For the "Leave Group" modal, revert any curly quotes to neutral aka typewriter aka ASCII quotes.~~
~~If there is any other similar auto-processing of the characters in `groupName`, revert it for the "Leave Group" modal.~~
Ensure user is able to type a group name that will match `groupName` so they can leave the group.

SebinSong:
closes #2479 

**[Fix screenshot]**

fix-for-string-comparison

---
As a fix for the issue, I created `normalizeString` function in `giLodash.js` where two conversions happen:

1]  `'` and `’` are both considered `'`
2] [Canonical equivalence normalization](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#canonical_equivalence_normalization)

### DIFF
--- a/frontend/model/contracts/shared/giLodash.js
+++ b/frontend/model/contracts/shared/giLodash.js
@@ -78,6 +78,15 @@ export function randomHexString (length: number): string {
   return Array.from(randomBytes(length), byte => (byte % 16).toString(16)).join('')
 }
 
+export function normalizeString (str: string): string {
+  return str
+    // [1]. Normalize strings by replacing both apostrophes and single quotes with a standard character (reference issue: https://github.com/okTurtles/group-income/issues/2479)
+    .replace(/['’]/g, "'")
+    // [2]. Normalize the string based on 'Canonical equivalence'. eg) 'Amélie' !== 'Amélie' even when they are visually identical because their unicode sequences are different.
+    //      (reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#canonical_equivalence_normalization)
+    .normalize('NFC')
+}
+
 export function randomIntFromRange (min: number, max: number): number {
   return Math.floor(Math.random() * (max - min + 1) + min)
 }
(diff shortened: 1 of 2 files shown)


frontend/model/contracts/shared/giLodash.js
```diff
@@ -78,6 +78,15 @@ export function randomHexString (length: number): string {
   return Array.from(randomBytes(length), byte => (byte % 16).toString(16)).join('')
 }
 
+export function normalizeString (str: string): string {
+  return str
+    // [1]. Normalize strings by replacing both apostrophes and single quotes with a standard character (reference issue: https://github.com/okTurtles/group-income/issues/2479)
+    .replace(/['’]/g, "'")
```
corrideat, line 84: I think that this will not catch all possible cases. See https://github.com/okTurtles/group-income/issues/2479#issuecomment-2585274842 for a more general solution.

SebinSong: @corrideat  Feel free to work on the issue yourself. I have no objection to it. (I'm actually busy working on another issue and I heard Greg said you are looking for issue to work on) 

just re-assign yourself to that issue and let me know. I will discard this PR. 

Edit: just did close it actually. Go ahead and take it over.

PR #2493, opened 2025-01-03
Improvements to time syncing

### LINKED ISSUE #2485: Group inactivity still shows the wrong inactive members (lastLoggedIn bug)
### Problem

I just had a chat with Sebin yesterday in the GIG group, and yet he is listed as not having logged in "the past 14 days or more".

### Solution

Find and fix this bug for good.

Previous attempts to fix this:

- #1997
- #2288

corrideat:
Closes #2485.

### DIFF
--- a/shared/domains/chelonia/time-sync.js
+++ b/shared/domains/chelonia/time-sync.js
@@ -38,6 +38,8 @@ const syncServerTime = async function () {
   wallBase =
     serverTime +
     (requestTimeElapsed - startTime) / 2 +
+    // Also take into account the time elapsed between `requestTimeElapsed`
+    // and this line (which should be very little)
     (newMonotonicBase - requestTimeElapsed)
   monotonicBase = newMonotonicBase
 }
--- a/shared/domains/chelonia/time-sync.js
+++ b/shared/domains/chelonia/time-sync.js
@@ -19,11 +19,11 @@ let watchdog
 
 const syncServerTime = async function () {
   // Get our current monotonic time
-  const newMonotonicBase = performance.now()
+  const startTime = performance.now()
   // Now, ask the server for the time
   const time = await fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
   const requestTimeElapsed = performance.now()
-  if (requestTimeElapsed - newMonotonicBase > 1000) {
+  if (requestTimeElapsed - startTime > 8000) {
     throw new Error('Error fetching server time: request took too long')
   }
   // If the request didn't succeed, report it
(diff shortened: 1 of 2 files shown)


shared/domains/chelonia/time-sync.js
```diff
@@ -34,12 +34,19 @@ const syncServerTime = async function () {
   // Adjust `wallBase` based on the elapsed request time. We can't know
   // how long it took for the server to respond, but we can estimate that it's
   // about half the time from the moment we made the request.
-  wallBase = serverTime + (requestTimeElapsed - newMonotonicBase) / 2
+  const newMonotonicBase = performance.now()
+  wallBase =
+    serverTime +
+    (requestTimeElapsed - startTime) / 2 +
+    (newMonotonicBase - requestTimeElapsed)
```
taoeffect, line 43: Can you update the comment to explain what this new term is?

taoeffect (changes requested):
Left one comment. Also consider, is there any useful `console.info` (or higher level) debugging that should be added anywhere to help diagnose the issue?

corrideat:
> is there any useful console.info (or higher level) debugging that should be added anywhere to help diagnose the issue?

I don't think so, but I also think this issue is resolved. If you want to test these changes locally, you're gonna need two devices (or a VM). You should carry out these steps on `master` to reproduce the issue and on this branch.

I recommend using Firefox on the device that will be suspended (device B) and running the server on device A. The reason to use Firefox is that you can follow these [instructions to test locally without HTTPS](https://stackoverflow.com/questions/61867895/can-firefox-76-be-forced-to-consider-localhost-subdomains-as-secure-context-wit/74000953#74000953).

1. Change `LAST_LOGGED_IN_THROTTLE_WINDOW` in `constants.js` to a small value, like 2 minutes. Also, change `lastLoggedIn` to `type: PERIODIC_NOTIFICATION_TYPE.MIN1` in `mainNotificationMixin.js`. This is not strictly necessary but will make testing _a lot_ faster.
2. Register on device A and create a group. Add income details (not necessary, but you can see the activity tab in the UI).
3. Join the group on device B as a new user. Also add income details.
4. Close the app in device A.
5. Put device B to sleep for at least the time you set in `LAST_LOGGED_IN_THROTTLE_WINDOW`. You may need to do this a few times.
6. On master, you should eventually miss the entry for the user created in device A. On this branch, you shouldn't miss any entries.

taoeffect (commented):
Nice! Congrats @corrideat !

taoeffect (approved):

PR #2494, opened 2025-01-05
Changes to contract CIDs (requires updating the chel command)

### LINKED ISSUE #2115: Investigate: potential ways of preventing malicious contracts from being loaded
### Problem

It is possible for someone to upload both contracts and contract manifests as files to the server, and then, for example, using either a malicious invite link (that references a groupid that uses these manifests), or possibly by registering their username with a modified identity contract, to get malicious contracts to load on the client.

### Solution

Look into ways to restrict contract manifests from being uploaded by users, so that only the developers are apps are allowed to define contract manifests for their apps.

corrideat:
This will probably fail until the `chel` command is updated. See <https://github.com/okTurtles/chel/compare/1927-design-implement-contract-deletion-op_contract_delete-1>.


corrideat:
https://github.com/multiformats/multicodec/pull/369 merged

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "group-income",
   "version": "1.1.0",
-  "contractsVersion": "1.1.0",
+  "contractsVersion": "1.1.1",
```
taoeffect: This should be undone.

shared/domains/chelonia/utils.js
```diff
@@ -848,3 +848,9 @@ export const collectEventStream = async (s: ReadableStream): Promise<any[]> => {
   }
   return r
 }
+
+export const CONTRACT_MANIFEST_REGEX: RegExp = /^zL7mM9d4Xb4T[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const CONTRACT_SOURCE_REGEX: RegExp = /^zLAeVmpcc88g[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const CONTRACT_DATA_REGEX: RegExp = /^zLDXeQ2AgfCu[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const FILE_MANIFEST_REGEX: RegExp = /^zLGQo2DimCH8{44}$/
+export const FILE_CHUNK_REGEX: RegExp = /^zLKHweRGqjMM{44}$/
```
taoeffect: Why are the last two missing `[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]`?

corrideat: Bad copy & paste.

Gruntfile.js
```diff
@@ -261,11 +261,14 @@ module.exports = (grunt) => {
...
       'define', 'bundle'
     ]),
-    // format: 'esm',
+    // Format must be 'iife' because we don't want 'import' in the output
     format: 'iife',
-    // banner: {
+    banner: {
+      // This banner makes contracts easy to detect and harmless in the event
+      // of accidental execution
+      js: 'for(;;)"use shelter";'
```
taoeffect: Please update this comment to say that the reason this is being added is to prevent people from uploading malicious contracts to our servers and then loading them elsewhere (e.g. via XSS or whatever).

EDIT: however, this isn't enforced server side, so it seems there's no point to it.

IMO, unless it's enforced server-side, we shouldn't have this as it's unnecessary complexity that accomplishes nothing. And if we enforce it server-side, then we can't support other languages — something we want.

So please could you remove this banner and related code?

shared/domains/chelonia/utils.js
```diff
@@ -848,3 +848,9 @@ export const collectEventStream = async (s: ReadableStream): Promise<any[]> => {
   }
   return r
 }
+
+export const CONTRACT_MANIFEST_REGEX: RegExp = /^zL7mM9d4Xb4T[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const CONTRACT_SOURCE_REGEX: RegExp = /^zLAeVmpcc88g[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const CONTRACT_DATA_REGEX: RegExp = /^zLDXeQ2AgfCu[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const FILE_MANIFEST_REGEX: RegExp = /^zLGQo2DimCH8[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
+export const FILE_CHUNK_REGEX: RegExp = /^zLKHweRGqjMM[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/
```
taoeffect: I asked Claude if this could be simplified:

Looking at your regex: `/^zLKHweRGqjMM[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{44}$/`

I can simplify it in a few ways:

1. The character class in square brackets appears to be Base58 encoding (Bitcoin style, without 0, O, I, and l). You can replace it with a named character class if your regex engine supports it.

2. The fixed prefix "zLKHweRGqjMM" followed by exactly 44 Base58 characters suggests this is validating some kind of specific identifier or hash format.

Here's a simplified version:

```js
/^zLKHweRGqjMM[1-9A-HJ-NP-Za-km-z]{44}$/
```

This uses character ranges instead of listing every character, making it more concise while maintaining the same functionality.

If your regex engine supports it, you could also define a Base58 character set variable and use that for better readability:

```js
const base58Chars = '[1-9A-HJ-NP-Za-km-z]';
const regex = new RegExp(`^zLKHweRGqjMM${base58Chars}{44}$`);
```

Either approach will give you equivalent matching behavior to your original regex.

Gruntfile.js
```diff
@@ -261,11 +261,11 @@ module.exports = (grunt) => {
     ...pick(clone(esbuildOptionBags.default), [
       'define', 'bundle'
     ]),
-    // format: 'esm',
+    // Format must be 'iife' because we don't want 'import' in the output
     format: 'iife',
-    // banner: {
+    banner: {
```
taoeffect: This can be left commented out then as it was.

backend/routes.js
```diff
@@ -267,7 +275,10 @@ route.POST('/name', {
 route.GET('/name/{name}', {}, async function (request, h) {
   const { name } = request.params
   try {
-    return await sbp('backend/db/lookupName', name)
+    const lookupResult = await sbp('backend/db/lookupName', name)
+    return lookupResult
+      ? h.response(lookupResult).type('text/plain')
+      : Boom.notFound()
```
taoeffect: Could you return `notFoundNoCache(h)` instead of `Boom.notFound()`?

backend/routes.js
```diff
@@ -341,9 +357,9 @@ if (process.env.NODE_ENV === 'development') {
       const { hash, data } = request.payload
       if (!hash) return Boom.badRequest('missing hash')
       if (!data) return Boom.badRequest('missing data')
-      const ourHash = createCID(data)
-      if (ourHash !== hash) {
-        console.error(`hash(${hash}) != ourHash(${ourHash})`)
+      const ourHashes = Object.values(multicodes).map((multicode) => createCID(data, ((multicode: any): number)))
+      if (ourHashes.includes(hash)) {
+        console.error(`hash(${hash}) != ourHash(${ourHashes.join(',')})`)
```
taoeffect: This is a very odd chunk of code here...

First off, I think the condition is backwards, should be `!ourHashes.includes(hash)`. And then still the error message would need to be updated to specify that the hash isn't in that list (as opposed to not being equal to the list).

However, I don't see a good reason to do it this way... can you update the code to extract the prefix and then based on the prefix decide which multicode to use? That way we don't need to try them all - we know immediately which one we need to use and the error message doesn't need to be modified.

backend/database.js
```diff
@@ -109,9 +110,9 @@ sbp('sbp/selectors/register', {
     await sbp('chelonia.db/set', namespaceKey(name), value)
     return { name, value }
   },
-  'backend/db/lookupName': async function (name: string): Promise<string | Error> {
+  'backend/db/lookupName': async function (name: string): Promise<string> {
     const value = await sbp('chelonia.db/get', namespaceKey(name))
-    return value || Boom.notFound()
+    return value
```
taoeffect, line 109: If you're going to modify this you need to make sure to modify every place that calls `'backend/db/registerName'`, e.g. now the logic in `'backend/db/registerName'` needs to be updated.

backend/database.js
```diff
@@ -196,7 +199,10 @@ export default async () => {
       if (!persistence || !await sbp('chelonia.db/get', key)) {
         const value = await readFile(path.join(dataFolder, key), 'utf8')
         // Load only contract source files and contract manifests.
-        if (value.startsWith(CONTRACT_MANIFEST_MAGIC) || value.startsWith(CONTRACT_SOURCE_MAGIC)) {
+        if (
+          CONTRACT_MANIFEST_REGEX.test(key) ||
+          CONTRACT_SOURCE_REGEX.test(key)
+        ) {
```
taoeffect: Isn't this check now redundant since it's already done above in the `.filter`?

backend/routes.js
```diff
@@ -195,7 +197,13 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {}, async function (requ
...
   const ip = request.headers['x-real-ip'] || request.info.remoteAddress
   try {
-    if (contractID.startsWith('_private') || since.startsWith('_private')) {
+    if (
+      !contractID ||
+      !CONTRACT_DATA_REGEX.test(contractID) ||
+      contractID.startsWith('_private') ||
+      !/^[0-9]+$/.test(since) ||
+      (limit && !/^[0-9]+$/.test(limit))
+    ) {
```
taoeffect, line 206: Could you update the `notFound` below to be `notFoundNoCache(h)`?

backend/routes.js
```diff
@@ -279,7 +290,12 @@ route.GET('/latestHEADinfo/{contractID}', {
 }, async function (request, h) {
   const { contractID } = request.params
   try {
-    if (contractID.startsWith('_private')) return Boom.notFound()
+    if (
+      !contractID ||
+      !CONTRACT_DATA_REGEX.test(contractID) ||
+      contractID.startsWith('_private')
+    ) return Boom.notFound()
```
taoeffect, line 297: => `notFoundNoCache(h)`

backend/routes.js
```diff
@@ -293,7 +309,7 @@ route.GET('/latestHEADinfo/{contractID}', {
 })
 
 route.GET('/time', {}, function (request, h) {
-  return new Date().toISOString()
+  return h.response(new Date().toISOString()).type('text/plain')
```
taoeffect: Can you add a `no-store` header here too?

backend/database.js
```diff
@@ -173,9 +174,8 @@ export default async () => {
   // TODO: Update this to only run when persistence is disabled when `chel deploy` can target SQLite.
   if (persistence !== 'fs' || options.fs.dirname !== dbRootPath) {
     // Remember to keep these values up-to-date.
-    const HASH_LENGTH = 52
-    const CONTRACT_MANIFEST_MAGIC = '{"head":"{\\"manifestVersion\\"'
-    const CONTRACT_SOURCE_MAGIC = '"use strict";'
+    const HASH_LENGTH = 56
```
taoeffect, line 171: Interesting that the hash length changed... are you sure it's 56 then for all of the ones we're checking for here?

backend/routes.js
```diff
@@ -467,16 +483,39 @@ route.POST('/file', {
...
+    type = 'application/vnd.shelter.contractmanifest+json'
+  } else if (CONTRACT_SOURCE_REGEX.test(hash)) {
+    type = 'application/vnd.shelter.contracttext'
+  } else if (FILE_MANIFEST_REGEX.test(hash)) {
+    type = 'application/vnd.shelter.filemanifest+json'
+  } else if (FILE_CHUNK_REGEX.test(hash)) {
+    type = 'application/vnd.shelter.filechunk+octet-stream'
+  } else if (hash.startsWith('name=')) {
+    type = 'text/plain'
+  }
```
taoeffect: Not saying you need to change this code, but have you considered something like this instead?

```js
if (hash.startsWith('name=') {
  type = 'text/plain'
} else {
  type = {
    [multicodes.SHELTER_CONTRACT_MANIFEST]: 'application/vnd.shelter.contractmanifest+json',
    [multicodes.SHELTER_CONTRACT_TEXT]: 'application/vnd.shelter.contracttext',
    // ...
  }[multicodes.extract(hash)] || 'application/octet-stream'
}
```

backend/routes.js
```diff
@@ -467,16 +483,39 @@ route.POST('/file', {
...
+
+  return h
+    .response(blobOrString)
+    .etag(hash)
     .header('Cache-Control', 'public,max-age=31536000,immutable')
+    // CSP to disable everything -- this only affects direct navigation to the
+    // `/file` URL.
+    .header('content-security-policy', "default-src 'none'; frame-ancestors 'none'; form-action 'none'; upgrade-insecure-requests; sandbox")
+    .header('x-content-type-options', 'nosniff')
```
taoeffect: Could you expand the comment a bit to explain the purpose of these headers?

backend/routes.js
```diff
@@ -666,7 +709,7 @@ route.GET('/kv/{contractID}/{key}', {
 }, async function (request, h) {
   const { contractID, key } = request.params
 
-  if (key.startsWith('_private')) {
+  if (!CONTRACT_DATA_REGEX.test(contractID) || !key || key.includes('\x00') || key.startsWith('_private')) {
```
taoeffect: Many of these improved checks have been added to this file (not just here), and I'm thinking: since these represent requests that should never be made in the first place, could we maybe return a different error than "not found"? E.g. bad request?

taoeffect (changes requested):
Fantastic work @corrideat! PR review ready!

backend/database.js
corrideat: Yes, you can also verify this by looking at the hashes generated. The reason the length changed is that the content id is a varint, and the new values are longer than the previous one (0x00 I believe, which is encoded as a single octet)

shared/domains/chelonia/utils.js
corrideat: The reason for being explicit with character ranges is because character ranges can be ambiguous depending on the locale (I don't think this necessarily applies to JavaScript, but it certainly applies to regular expressions in general). For example,   `ñ` will be matched by `[a-z]` in the `C.UTF-8` locale but not in the `en_US.UTF-8` locale. 

However, maybe we don't need these regular expressions at all if we parse the CID.

backend/routes.js
```diff
@@ -467,16 +487,48 @@ route.POST('/file', {
...
+      case multicodes.SHELTER_CONTRACT_DATA:
+        type = 'application/vnd.shelter.contractdata+json'
+        break
+      case multicodes.SHELTER_FILE_MANIFEST:
+        type = 'application/vnd.shelter.filemanifest+json'
+        break
+      case multicodes.SHELTER_FILE_CHUNK:
+        type = 'application/vnd.shelter.filechunk+octet-stream'
+        break
+    }
```
taoeffect: Could this be converted to using an object/hash table approach? e.g.

```js
  type = {
    [multicodes.SHELTER_CONTRACT_MANIFEST]: 'application/vnd.shelter.contractmanifest+json',
    [multicodes.SHELTER_CONTRACT_TEXT]: 'application/vnd.shelter.contracttext',
    // ...
  }[multicodes.extract(hash)] || 'application/octet-stream'
```

shared/functions.js
```diff
@@ -7,7 +7,48 @@ import { blake2b256 } from '~/shared/multiformats/blake2b.js'
...
+
+  return parsed
+}
+
+export const maybeParseCID = (cid: string) => {
+  try {
+    return parseCID(cid)
+  } catch (e) {
+    // We ignore errors
+    console.error(e, '@@@XXXX')
```
taoeffect: This error message needs to be fixed...

backend/routes.js
corrideat: I prefer the existing approach, unless the table is defined outside of this function. However, since the values are numbers, doing it this way requires (implicitly) converting a number to a string twice (when the table is defined and then for looking the value up)

backend/routes.js
corrideat: Hm, not really sure. I think that a 404 is a valid response, and so is 400, depending on perspective. A 404 is valid because 'technically' there could be a response but we know there won't be / there shouldn't be.

A 400 is also valid. The spec says: 'The request could not be understood by the server due to malformed syntax.', and we could say that the path is itself invalid and there couldn't possibly be a response.

Let me know if I should change this to 400.

taoeffect: yes, 400 (or something similar) would be better than (not found) IMO

backend/routes.js
taoeffect: I'm really not a fan of these switch statements. They're not necessary and they are slower and visually longer. Fundamentally what you're doing is mapping one value to another `O(1)`, and that's what this map object is doing, whereas a switch statement is going to go one at a time doing comparisons `O(n/2)` - it's slower, visually longer, and just not as conceptually elegant.

If it would make you feel better to save this map near the top of the file (or in some other file) so that you can reference it in multiple places, feel free to do so.

backend/routes.js
corrideat: Made this change. However, personally I prefer a 404 in these less clear cases as it divulges slightly less information to clients.

backend/database.js
corrideat: I believe I already did this

backend/routes.js
corrideat: You're right about the condition, seems to be wrong.

backend/routes.js
taoeffect: yeah but ... it's not like we're hiding anything important.

backend/database.js
taoeffect: AFAICT that code hasn't been touched....

corrideat: No, but I think it needed to. Updated now.

taoeffect (dismissed):
Approved! Great work @corrideat!

shared/domains/chelonia/SPMessage.js
```diff
@@ -426,15 +424,15 @@ export class SPMessage {
 
   head (): Object { return this._head }
 
-  message (): SPOpOpValue { return this._message }
+  message (): SPOpValue { return this._message }
 
-  op (): SPOp { return [this.head().op, this.message()] }
+  op (): SP { return [this.head().op, this.message()] }
```
taoeffect: 👆

taoeffect (approved):
Nice!

PR #2495, opened 2025-01-06
Contract deletion

### LINKED ISSUE #1927: Design & Implement Contract Deletion (OP_CONTRACT_DELETE?)
### Problem

Currently we have no way to delete a contract once it's been created.

This means unnecessary space will be used on the server.

### Solution

Implement `OP_CONTRACT_DELETE` and use it when, for example, deleting chatrooms.

This opcode will go through and delete every message in a contract and all state (except attachments) associated with it.

Clients trying to sync the contract again should get [`410 Gone`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) errors and assume the contract has been deleted. Clients syncing contracts that never existed should return `404 not  found`.

Note: in the case of deleted chatrooms, the attachments in that chatroom must be handled separately. Deleting a contract doesn't delete the attachments inside of it, so those must be deleted separately.

EDIT: note [this comment](https://github.com/okTurtles/group-income/issues/1913#issuecomment-2059304322) that mentions that attachments could be deleted as well.

corrideat:
Close #1927


frontend/setupChelonia.js
```diff
@@ -12,6 +12,52 @@ import manifests from './model/contracts/manifests.json'
...
+  const { cheloniaState, contractState } = sbp('chelonia/contract/fullState', contractID)
+  if (!cheloniaState) return
+
+  await sbp('chelonia/contract/remove', contractID)
+
+  const rootGetters = sbp('state/vuex/getters')
+
+  switch (cheloniaState.type) {
+    case 'gi.contracts/chatroom': {
+      // TODO: Leave chatroom on our identity contract or our group, if we've
```
corrideat: Not having a chatroom->parent relationship makes this difficult to implement (efficiently, at least). We also need to add a new action to the `identity` contract, to delete chatrooms (currently, it's append-only).

corrideat:
Some useful console invocations for testing this PR:

```js
// Delete the current identity contract (and everything it owns)
await sbp('gi.app/identity/delete', sbp('state/vuex/state').reverseNamespaceLookups[sbp('state/vuex/state').loggedIn.identityContractID], sbp('state/vuex/state').loggedIn.identityContractID, prompt('Password'))

// Delete the current group (and everything it owns)
await sbp('chelonia/out/deleteContract', sbp('state/vuex/state').currentGroupId, {[sbp('state/vuex/state').currentGroupId]: { billableContractID: sbp('state/vuex/state').loggedIn.identityContractID } })


```

backend/server.js
```diff
@@ -363,6 +474,9 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
       })
     }))
   }
+  sbp('chelonia.persistentActions/load').catch(e => {
+    console.error(e, 'Error loading persistent actions')
+  })
```
taoeffect, line 497: I think if the server fails to load these perhaps we should shut down with this error instead of proceeding so that the sysadmin knows something is critically wrong.

backend/server.js
```diff
@@ -213,6 +215,112 @@ sbp('sbp/selectors/register', {
...
+    return sbp('chelonia/queueInvocation', cid, async () => {
+      const owner = await sbp('chelonia.db/get', `_private_owner_${cid}`)
+      const rawManifest = await sbp('chelonia.db/get', cid)
+      if (rawManifest === '') throw new BackendErrorGone()
+      if (!rawManifest) throw new BackendErrorNotFound()
+
+      const resourcesKey = `_private_resources_${cid}`
+      const resources = await sbp('chelonia.db/get', resourcesKey)
+      if (resources) {
+        await Promise.allSettled(resources.split('\x00').map(async (resourceCid) => {
```
taoeffect: Claude: For large contracts with many resources, consider batching the deletion operations to avoid overwhelming the system.

backend/errors.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+
+import { ChelErrorGenerator } from '~/shared/domains/chelonia/errors.js'
+
+export const BackendErrorNotFound: typeof Error = ChelErrorGenerator('BackendErrorNotFound')
+export const BackendErrorGone: typeof Error = ChelErrorGenerator('BackendErrorGone')
+export const BackendErrorBadData: typeof Error = ChelErrorGenerator('BackendErrorBadData')
```
taoeffect, line 7: Claude is unhappy with these as they seem to be redundant and require a bunch of if/else clauses that merely convert to booms.

If you can't throw booms directly then can you create a generic class to wrap them that would get rid of the need for this type of code?

```js
    switch (e.name) {
      case 'BackendErrorNotFound':
        return Boom.notFound()
      case 'BackendErrorGone':
        return Boom.resourceGone()
      case 'BackendErrorBadData':
        return Boom.badData(e.message)
      default:
        console.error(e, 'Error during deletion')
        return Boom.internal(e.message ?? 'internal error')
    }
```

Note also [`Boom.boomify`](https://hapi.dev/module/boom/api/?v=10.0.1#boomifyerr-options)

taoeffect (changes requested):
Preliminary review ready

backend/errors.js
corrideat: Not really a fan of `Boom`, I think that it pollutes code by adding presentation layer logic (i.e., the error to return) to control logic and is also HAPI-specific. But, I could use the `switch` proposed.

backend/server.js
corrideat: Isn't this what this already does? See the `chelonia.persistentActions/enqueue` below?

backend/server.js
corrideat: Hmm, possibly. Right now, we're only using persistent actions for contract deletion, which I'd not necessarily consider 'critical' (in the sense that it's possible for the server to operate normally without this). We could shut the server down, though.

I think that if we go the shutdown route, some other changes may be needed to do that cleanly (currently, the server would start, possibly start serving clients and _then_ shutdown).

backend/errors.js
corrideat: On second thought, seeing where these are used (i.e., things that aren't in Chelonia), I think `Boom` _could_ be fine, but with the caveat that this makes the server code more dependent on the HAPI ecosystem. Let me know how you prefer to tackle this (i.e., use `Boom` or use a `switch`)

taoeffect: That's a good point you bring up regarding tying us to Hapi. We definitely don't want to be tied to Hapi.

Is there some way you could stick with the Chel errors but without the switch? I.e. would it be possible to somehow in one line coerce / convert these to Boom errors or the equivalent of whatever Hapi natively uses (without Boom)? Perhaps some sort of utility function?

I see at least two places in the code that perform this `switch` conversion, and it's ugly and a DRY violation.

backend/server.js
taoeffect: Hmm, maybe. Right now I'm a bit tired to look very closely at this (can review in more detail later), but one thing that stands out (that I've noticed in other parts of the code too, e.g. in `server.js` near the bottom), is this `Promise.allSettled(resources.split('\x00').map` pattern, which I think is potentially dangerous and a DoS vector.

It creates a potentially extremely large amount of functions and promises in memory that all need to happen at the same time. If this contract has an extreme amount of resources, it could really hurt the memory usage on the server.

So is there a way to for example, do say batches of 100 at a time?

corrideat: A change like this (i.e., batching) is possible, but I think it should be benchmarked. It's not true that everything needs to be executed at the same time (these are likely scheduled by the engine). It _is_ true that a function needs to be allocated, but the overhead of that should be quite minimal.

> If this contract has an extreme amount of resources, it could really hurt the memory usage on the server.

To an extent, although the bigger issue here is the `.split()` part. Still, we'd be talking about a really large number of resources, and we could limit that (e.g., a contract could be limited to no more than, e.g., 100k resources). Remember that the index itself needs to be stored in memory.

All in all, I really don't think this is a significant issue, and, if it is, the solution is probably in splitting the index into smaller subindices, which adds complexity.

backend/errors.js
corrideat: I've moved them to a function now.

shared/domains/chelonia/persistent-actions.js
```diff
@@ -231,7 +242,7 @@ sbp('sbp/selectors/register', {
   'chelonia.persistentActions/unload' (): void {
     for (const id in this.actionsByID) {
       // Clear the action's timeout, but don't cancel it so that it can later resumed.
-      this.actionsByID[id].timer && clearTimeout(this.actionsByID[id].timer)
+      this.actionsByID[id][timer] && clearTimeout(this.actionsByID[id].timer)
```
taoeffect: Did you mean `clearTimeout(this.actionsByID[id][timer])`?

corrideat: I did

backend/database.js
```diff
@@ -213,3 +216,101 @@ export default async () => {
...
+      // If the current index exists, check if the value is already present
+      if (currentIndex) {
+        // Check for existing value to avoid duplicates
+        if (
+          // Check if the value is at the end
+          currentIndex.endsWith('\x00' + value) ||
+          // Check if the value is at the start
+          currentIndex.startsWith(value + '\x00') ||
+          // Check if the current index is exactly the value
+          currentIndex === value
```
taoeffect, line 269: How about using a normal JSON array instead of this `\x00` string manipulation stuff?

It would also be possible to provide more robust duplicate prevention, and it would make it possible to integrate more tightly with databases that support native JSON manipulation functions.

backend/database.js
```diff
@@ -213,3 +216,101 @@ export default async () => {
...
+        await sbp('chelonia.db/delete', key)
+        return
+      }
+
+      // Find the index of the value in the existing entries
+      const entryIndex = existingEntries.indexOf('\x00' + value + '\x00')
+      if (entryIndex === -1) return
+
+      // Create an updated index by removing the value
+      const updatedIndex = existingEntries.slice(0, entryIndex) + existingEntries.slice(entryIndex + value.length + 1)
```
taoeffect, line 306: Are you sure this `+ 1` should be there...?

```
> a = 'hellomothermynameis'
'hellomothermynameis'
> i = a.indexOf('mother')
5
> a.slice(0, i) + a.slice(i + 'mother'.length + 1)
'helloynameis'    # <---- no 'ynameis' not 'mynameis'
> a.slice(i + 'mother'.length)
'mynameis'
```

taoeffect, line 328: By removing both the leading `\x00` and the ending `\x00` I think this code would be putting the values before and after this value next to each other without a separator.

backend/database.js
corrideat: This doesn't require parsing or escaping the way JSON does, and this index can be read or written to using a streaming approach, which JSON cannot. It doesn't require allocating an array in memory (unfortunately, it does require allocating a string in memory, sort of because of how the DB interface is written).

> It would also be possible to provide more robust duplicate prevention

That's also possible now, but doing so on every operation is expensive.

> it would make it possible to integrate more tightly with databases that support native JSON manipulation functions

True, although such DBs (e.g., pgSQL) also have no trouble with this approach. That said, a DB like pgSQL likely would benefit a lot from having these indices as tables (not JSON)

backend/database.js
corrideat: But this doesn't remove both separators.

```
(() => {
      const existingEntries = 'a\x00b\x00c'
      const value = 'b'


      const entryIndex = existingEntries.indexOf('\x00' + value + '\x00')
      if (entryIndex === -1) return

      // Create an updated index by removing the value
      const updatedIndex = existingEntries.slice(0, entryIndex) + existingEntries.slice(entryIndex + value.length + 1)

      console.error('updatedIndex', JSON.stringify(updatedIndex))
      // Output: "a\u0000c"
})()
```

frontend/setupChelonia.js
```diff
@@ -12,6 +12,43 @@ import manifests from './model/contracts/manifests.json'
...
+  const handler = type && sbp('sbp/selectors/fn', `${type}/_ondeleted`)
+
+  const currentIdentityState = sbp('state/vuex/getters').currentIdentityState
+  // Delete redudant file deletion tokens. If a contract has been deleted, so
+  // have been the files it contained, and therefore we no longer need to
+  // hold on to their file deletion tokens.
+  if (currentIdentityState.fileDeleteTokens) {
+    // $FlowFixMe[incompatible-use]
+    const manifestCids = Object.entries(currentIdentityState.fileDeleteTokens).filter(([, { billingContractID }]) => {
+      return billingContractID === contractID
```
taoeffect: In most places you use `billableContractID`, not `billingContractID`... why the change here?

corrideat: Seems like an oversight.

corrideat: Seems like it's also set as `billableContractId` in the contract:

```js
          state.fileDeleteTokens[manifestCid] = {
            billableContractID: data.billableContractID,
            token
          }
```

shared/domains/chelonia/chelonia.js
```diff
@@ -1198,6 +1200,58 @@ export default (sbp('sbp/selectors/register', {
...
+      throw new TypeError('A contract ID must be provided')
+    }
+    if (!Array.isArray(contractID)) contractID = [contractID]
+    return await Promise.allSettled(contractID.map(async (cid) => {
+      const hasCredential = has(credentials, cid)
+      const hasToken = has(credentials[cid], 'token') && credentials[cid].token
+      const hasBillableContractID = has(credentials[cid], 'billableContractID') && credentials[cid].billableContractID
+      if (!hasCredential || (!hasToken && hasToken === hasBillableContractID)) {
+        throw new TypeError(`Either a token or a billable contract ID must be provided for ${cid}`)
+      }
```
taoeffect: Seems more straightforward to write:

```js
if (!hasCredential || (!hasToken && !hasBillableContractID)) { … }
```

backend/server.js
```diff
@@ -213,6 +216,111 @@ sbp('sbp/selectors/register', {
...
+  async 'backend/deleteFile' (cid: string): Promise<void> {
+    const owner = await sbp('chelonia.db/get', `_private_owner_${cid}`)
+    const rawManifest = await sbp('chelonia.db/get', cid)
+    if (rawManifest === '') throw new BackendErrorGone()
+    if (!rawManifest) throw new BackendErrorNotFound()
+
+    try {
+      const manifest = JSON.parse(rawManifest)
+      if (!manifest || typeof manifest !== 'object') throw new BackendErrorBadData('manifest format is invalid')
+      if (manifest.version !== '1.0.0') throw BackendErrorBadData('unsupported manifest version')
```
taoeffect: Missing `new` after `throw` on line 229

shared/domains/chelonia/chelonia.js
corrideat: That's not the relly same, as the intent is that this should be either-or (i.e., XOR). However, I think that `(!hasCredential || hasToken === hasBillableContractID)` would work.

backend/routes.js
```diff
@@ -236,6 +254,18 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {}, async function (requ
...
+  auth: {
+    strategies: ['chel-shelter'],
+    mode: 'required'
+  }
+}, async function (request, h) {
+  const billableContractID = request.auth.credentials.billableContractID
+  const resources = (await sbp('chelonia.db/get', `_private_resources_${billableContractID}`))?.split('\x00')
+
+  return resources || []
+})
```
taoeffect, line 272: Could you add a comment above line 257 explaining who is allowed to call this and an example of what this route is useful for?

While trying to understand `chel-shelter` authorization I noticed that `verifyShelterAuthorizationHeader` has:

```
 if (Math.abs(parseInt(timestamp) - (Date.now() / 1e3 | 0)) > 60) 
```

So it seems like `master` hasn't been merged here or that was missed.

corrideat: `master` hadn't been merged. An error like that would very likely result failing tests (due to being unable to create contracts).

backend/routes.js
```diff
@@ -594,40 +629,78 @@ route.POST('/deleteFile/{hash}', {
...
-  if (!rawManifest) return Boom.notFound()
   try {
-    const manifest = JSON.parse(rawManifest)
-    if (!manifest || typeof manifest !== 'object') return Boom.badData('manifest format is invalid')
-    if (manifest.version !== '1.0.0') return Boom.badData('unsupported manifest version')
-    if (!Array.isArray(manifest.chunks) || !manifest.chunks.length) return Boom.badData('missing chunks')
-    // Delete all chunks
-    await Promise.all(manifest.chunks.map(([, cid]) => sbp('chelonia.db/delete', cid)))
+    await sbp('backend/deleteFile', hash)
+    return h.response()
```
taoeffect, line 642: Above these lines there are a whole bunch of naked calls to `'chelonia.db/get'`. Should those be surrounded with `try/catch` as well?

corrideat: I don't see why; those will result in `500` errors, mostly. The purpose of this `try / catch` is mostly to remap errors

backend/routes.js
```diff
@@ -594,40 +629,78 @@ route.POST('/deleteFile/{hash}', {
...
+        } else {
+          break
+        }
+      // Prevent an infinite loop
+      } while (count < 128)
+      // Check that the user making the request is the ultimate owner (i.e.,
+      // that they have permission to delete this file)
+      if (!ctEq(request.auth.credentials.billableContractID, ultimateOwner)) {
+        return Boom.unauthorized('Invalid token', 'bearer')
+      }
```
taoeffect: Both `chel-shelter` and `chel-bearer` return this `'bearer'` value for `Boom.unauthorized`. Did you mean that here, or should it say `'shelter'` or something to that effect?

backend/routes.js
```diff
@@ -594,40 +629,78 @@ route.POST('/deleteFile/{hash}', {
...
+    default:
+      return Boom.unauthorized('Missing or invalid auth strategy')
+  }
 
-  return h.response()
+  // Authentication passed, now proceed to delete the contract and its associated
+  // keys
+  try {
+    const [id] = sbp('chelonia.persistentActions/enqueue', ['backend/deleteContract', hash])
+    return h.response({ id }).code(202)
```
taoeffect: Can you add a comment here explaining why we're passing this `id` value back to the user? (Why do they need it?)

backend/routes.js
```diff
@@ -715,6 +788,7 @@ route.POST('/kv/{contractID}/{key}', {
   const existingSize = existing ? Buffer.from(existing).byteLength : 0
   await sbp('chelonia.db/set', `_private_kv_${contractID}_${key}`, request.payload)
   await sbp('backend/server/updateSize', contractID, request.payload.byteLength - existingSize)
+  await appendToIndexFactory(`_private_kvIdx_${contractID}`)(key)
```
taoeffect, line 806: Seems a bit wasteful to create a factory function on every call to updating a key.. I thought the point of a factory function was to save it for later use on-demand? Can this be cached somewhere so that we don't re-create it on each call?

backend/routes.js
corrideat: It should probably say `shelter`.

backend/routes.js
corrideat: IIRC, you asked for this. It's so that later users can track the progress of deletes.

backend/routes.js
corrideat: I agree in principle, but this is the best approach I could find. The reason they can't be cached is that they have a dynamic component (the `contractID`)

corrideat: Maybe I can export a dynamic and a factory version of those functions?

taoeffect: nah it's fine

backend/server.js
```diff
@@ -213,6 +216,111 @@ sbp('sbp/selectors/register', {
...
+
+          if (parsed.code === multicodes.SHELTER_CONTRACT_DATA) {
+            return sbp('chelonia.persistentActions/enqueue', ['backend/deleteContract', resourceCid])
+          } else if (parsed.code === multicodes.SHELTER_FILE_MANIFEST) {
+            return sbp('chelonia.persistentActions/enqueue', ['backend/deleteFile', resourceCid])
+          }
+
+          return undefined
+        }))
+      }
```
taoeffect: Could you add some comments here explaining what this block is doing?

I assume it's deleting resources related to this contract, e.g. file attachments in a chatroom, and associated contracts (e.g. group => chatroom contracts)

backend/server.js
```diff
@@ -213,6 +216,111 @@ sbp('sbp/selectors/register', {
...
+        }
+        await sbp('chelonia/db/deleteLatestHEADinfo', cid)
+      }
+
+      const kvIndexKey = `_private_kvIdx_${cid}`
+      const kvKeys = await sbp('chelonia.db/get', kvIndexKey)
+      if (kvKeys) {
+        await kvKeys.split('\x00').map((key) => {
+          return sbp('chelonia.db/delete', `_private_kv_${cid}_${key}`)
+        })
```
taoeffect: Think this is missing a surrounding `Promise.all` or `Promise.allSettled`?

backend/server.js
```diff
@@ -213,6 +216,111 @@ sbp('sbp/selectors/register', {
...
+        for (let i = latestHEADinfo.height; i > 0; i--) {
+          const eventKey = `_private_hidx=${cid}#${i}`
+          const event = await sbp('chelonia.db/get', eventKey)
+          if (event) {
+            await sbp('chelonia.db/delete', event)
+            await sbp('chelonia.db/delete', eventKey)
+          }
+        }
+        await sbp('chelonia/db/deleteLatestHEADinfo', cid)
+      }
```
taoeffect, line 305: Similarly, this block begins a new logical section / purpose, so it'd be good to add at least a 1-line comment here to the effect of:

```js
// Next, loop through all the events in the contract and delete them, starting with the most recent ones
```

Likewise, for the section below, a comment to the effect of:

```js
// Then, delete all KV-store values associated with this contract
```

backend/server.js
corrideat: Indeed.

backend/routes.js
taoeffect: It still says `bearer` tho...

shared/domains/chelonia/persistent-actions.js
```diff
@@ -142,12 +151,11 @@ sbp('sbp/selectors/register', {
...
   // Cancels a specific action by its ID.
   // The action won't be retried again, but an async action cannot be aborted if its promise is stil attempting.
-  'chelonia.persistentActions/cancel' (id: UUIDV4): void {
+  async 'chelonia.persistentActions/cancel' (id: UUIDV4): Promise<void> {
     if (id in this.actionsByID) {
       this.actionsByID[id].cancel()
       delete this.actionsByID[id]
-      // Likely no need to await this call.
-      sbp('chelonia.persistentActions/save')
+      return await sbp('chelonia.persistentActions/save')
```
taoeffect, line 162: Something to be aware of: the way this is written with the `delete` renders `this.status.resolved = true` kind of meaningless - in other words if an action is cancelled it won't appear in `'chelonia.persistentActions/status'`. That may be fine, but just pointing this out. Might want to add a comment here to that effect too.

shared/domains/chelonia/chelonia.js
```diff
@@ -1198,6 +1200,58 @@ export default (sbp('sbp/selectors/register', {
...
+      headers: new Headers([
+        [
+          'authorization',
+          buildShelterAuthorizationHeader.call(this, contractID)
+        ]
+      ])
+    })
+    if (!response.ok) {
+      throw new Error(`Unable to fetch own resources for ${contractID}`)
+    }
```
taoeffect: Could you:

- include a copy of the HTTP statuscode and status message in this error? E.g. `Error fetching resources for ${contractID}: ${statusText}`
- also console.error log it?

shared/domains/chelonia/chelonia.js
```diff
@@ -1198,6 +1200,58 @@ export default (sbp('sbp/selectors/register', {
...
+
+    return response.json()
+  },
+  'chelonia/out/deleteContract': async function (contractID: string | string[], credentials: { [contractID: string]: { token: ?string, billableContractID: ?string } } = {}) {
```
taoeffect: This line is way too long and unreadable. Please break it up like so:

```js
  'chelonia/out/deleteContract': async function (
    contractID: string | string[],
    credentials: { [contractID: string]: { token: ?string, billableContractID: ?string } } = {}
  ) {
```

shared/domains/chelonia/chelonia.js
```diff
@@ -1198,6 +1200,58 @@ export default (sbp('sbp/selectors/register', {
...
+              // $FlowFixMe[incompatible-type]
+              ? `bearer ${(credentials[cid].token: any).valueOf()}`
+              // $FlowFixMe[incompatible-type]
+              // $FlowFixMe[incompatible-call]
+              : buildShelterAuthorizationHeader.call(this, credentials[cid].billableContractID)]
+        ])
+      })
+      if (!response.ok) {
+        throw new Error(`Unable to delete contract ${cid}`)
+      }
```
taoeffect: Same sort of request here:

- `Error deleting contract ${contractID}: ${statusText}`
- a console.error with the status code and text

frontend/views/containers/user-settings/AccountRemovalModal.vue
```diff
@@ -0,0 +1,173 @@
...
+        i18n(
+          :args='LTags("strong")'
+        ) This action {strong_}cannot be undone{_strong}.
+
+      banner-simple.c-banner(severity='general' v-if='!ownResources')
+        i18n Loading
+      banner-simple.c-banner(severity='warning' v-else-if='Array.isArray(ownResources) && ownResources.length')
+        i18n(tag='p') This action will also delete the following groups:
+        ul.c-list
+          li.c-item(v-for='cid in ownResources') {{cid}}
```
taoeffect: Please rename `cid` to `groupName` since that's what it is

frontend/setupChelonia.js
```diff
@@ -141,8 +179,15 @@ const setupChelonia = async (): Promise<*> => {
...
     },
     hooks: {
-      syncContractError: (e: Error, contractID: string) => {
-        if (['ChelErrorUnrecoverable', 'ChelErrorForkedChain'].includes(e?.name)) {
+      syncContractError: (e, contractID) => {
+        if (!e) return
+        if (e.name === 'ChelErrorUnexpectedHttpResponseCode' && e.message.startsWith('410:')) {
+          console.info('[syncContractError] Contract ID ' + contractID + ' has been deleted')
+          handleDeletedContract(contractID).catch(e => {
+            console.error('Error handling contract deletion', e)
```
taoeffect: To help us debug problems, console error messages should be unique. This `'Error handling contract deletion'` message appears identically in 3 different places. Could you please make each version of it unique so that when reviewing logs we can identify where it came from?

frontend/setupChelonia.js
```diff
@@ -12,6 +12,43 @@ import manifests from './model/contracts/manifests.json'
...
+  if (currentIdentityState.fileDeleteTokens) {
+    // $FlowFixMe[incompatible-use]
+    const manifestCids = Object.entries(currentIdentityState.fileDeleteTokens).filter(([, { billableContractID }]) => {
+      return billableContractID === contractID
+    }).map(([cid]) => cid)
+    await sbp('gi.actions/identity/removeFiles', {
+      manifestCids,
+      option: {
+        shouldDeleteToken: true
+      }
```
taoeffect, line 37: This seems like it will result in a call to `'chelonia/fileDelete'` which will try to delete the file on the server, but per the comment, those files should already be deleted, so this call shouldn't be made.

Perhaps either an additional option should be added here to skip that or ... what?

frontend/setupChelonia.js
```diff
@@ -12,6 +12,43 @@ import manifests from './model/contracts/manifests.json'
...
+    }).catch(e => {
+      console.error('[handleDeletedContract] Error deleting saved tokens for deleted contract', contractID, e)
+    })
+  }
+
+  if (typeof handler === 'function') {
+    await handler(contractID, contractState).catch(e => {
+      console.error('Error handling deletion of contract', contractID)
+    })
```
taoeffect: We need some clarification in a comment somewhere, perhaps here, or perhaps elsewhere, to explain why exactly this is safe to do like this...

The specific concern that I have is that when a group is deleted, it will result in these handlers being called on clients in an unpredictable order. In other words, perhaps the identity contract's `_ondeleted` will be called first, or last, or somewhere in between, to the `chatroom` and `group` `_ondeleted` calls... how do you know that no matter what order they're called in the same result will occur?

shared/domains/chelonia/persistent-actions.js
corrideat: I agree, although I'm not sure what the motivation was when this was originally written. I think that it could be left as-is, and in the future there could be a periodic persistent action to prune completed actions (I'm thinking when period reporting is implemented).

frontend/setupChelonia.js
corrideat: > Perhaps either an additional option should be added here to skip that or ... what?

That's already done. Calling `chelonia/fileDelete` requires `shouldDeleteFile`

frontend/setupChelonia.js
corrideat: That's a valid concern, but AFAICT the way it's written / used it's not really an issue, other than redundancy. For example, the `_ondeleted` on a chatroom will attempt to remove it from the parent group or identity contract. If the parent is what happened to also be deleted, this action will fail, but it's of little consequence as everything will be removed in the end.

backend/routes.js
corrideat: I'd updated it in `deleteFile`. Updated.

frontend/setupChelonia.js
```diff
@@ -12,6 +12,50 @@ import manifests from './model/contracts/manifests.json'
...
+  }
+
+  // Note: when multiple contracts are deleted, the order in which the `handler`
+  // is called isn't guaranteed. The `_ondeleted` handlers should be written in
+  // such a way that the order in which they're called is of no consequence.
+  // For example, a group might use `_ondeleted` to remove it from its associated
+  // identity contract, since this is safe. If the identity contract is also being
+  // removed, this is at worst redudant, but still save, since removal of the
+  // identity contract also deletes the same information.
```
taoeffect: `but still save` => `but still safe`

frontend/model/contracts/identity.js
```diff
@@ -399,6 +399,28 @@ sbp('chelonia/defineContract', {
...
+      },
+      sideEffect ({ data, contractID, innerSigningContractID }) {
+        sbp('okTurtles.events/emit', DELETED_CHATROOM, { chatRoomID: data.contractID })
+        const { identityContractID } = sbp('state/vuex/state').loggedIn
+        if (identityContractID === innerSigningContractID) {
+          sbp('gi.actions/chatroom/delete', { contractID: data.contractID, data: {} }).catch(e => {
+            console.log(`Error sending chatroom removal action for ${data.chatRoomID}`, e)
+          })
+        }
+      }
```
taoeffect: I don't think we want to necessarily emit `DELETED_CHATROOM` and call `'gi.actions/chatroom/delete'` every single time this message is processed. We could have already done that, in which case I don't think it would make sense to do so.

So I'd recommend wrapping all of this interior stuff in a queueInvocation on the identity contractID, and then only run the sideEffect code if the chatroom hasn't been deleted.

corrideat: This is essentially a copy of the delete code for groups.

>  I'd recommend wrapping all of this interior stuff in a queueInvocation on the identity contractID, and then only run the sideEffect code if the chatroom hasn't been deleted.

Not really sure what you mean.

> we want to necessarily emit DELETED_CHATROOM and call 'gi.actions/chatroom/delete' every single time this message is processed. 

Why not? I think this is exactly what we should do, because if we're processing this it means it hasn't been seen before. Therefore, code that's received the `JOINED_CHATROOM` event could be expecting a `DELETED_CHATROOM` event. Then, calling `gi.actions/chatroom/delete` is the only way to make sure that the action is sent. Due to how the state is processed, we have no way of knowing if this was called, or if it was successful, in the past.

taoeffect: > Not really sure what you mean.

This is a common pattern throughout the codebase with which you are familiar with, so how can it be that you're not sure what I mean?

Here's an example, written by you:

```js
            const state = await sbp('chelonia/contract/state', contractID)
            // In order to send the leaveChatRoom action, we need to be an active
            // group member that's a chatroom member.
            // In addition, we skip sending an action if there there have been
            // other join or leave events by checking the `joinedHeight` value.
            if (
              state?.profiles?.[innerSigningContractID]?.status === PROFILE_STATUS.ACTIVE &&
              state.chatRooms?.[data.chatRoomID]?.members[memberID]?.status === PROFILE_STATUS.REMOVED &&
              state.chatRooms[data.chatRoomID].members[memberID].joinedHeight === data.joinedHeight
            ) {
              await leaveChatRoomAction(contractID, state, data.chatRoomID, memberID, innerSigningContractID)
            }
          }).catch((e) => {
            console.error(`[gi.contracts/group/leaveChatRoom/sideEffect] Error for ${contractID}`, { contractID, data, error: e })
          })
```

> Why not? I think this is exactly what we should do, because if we're processing this it means it hasn't been seen before. Therefore, code that's received the JOINED_CHATROOM event could be expecting a DELETED_CHATROOM event. Then, calling gi.actions/chatroom/delete is the only way to make sure that the action is sent. Due to how the state is processed, we have no way of knowing if this was called, or if it was successful, in the past.

Again - I see no reason to do this every single time we login. We are perfectly capable of logging in, waiting for the contract to finish syncing, and then checking if this code is necessary.

corrideat: Thanks for clarifying.

I don't really see how the `queueInvocation` pattern is useful here. That's done for delaying when the code there happens, and I don't see how delaying anything helps here. Deleting, unlike joining or leaving, is intended to be final. I also don't see how we'd know "if the chatroom hasn't been deleted" without attempting to delete it, which is what this code does. I don't see how delaying the `/delete` action using `queueInvocation` or something else expedites the process of finding out if something has been deleted.

> Again - I see no reason to do this every single time we login

This isn't 'every single time we login', it's only when logging in on a new device, or when the local state is cleared. I've explained the reasons for why this needs to be done (or at the very least, why it's a very good idea to do so), namely that the event is useful for code trying to keep track of chatrooms and that we can't know if the `delete` action was called or not.

corrideat: And once again, this is pretty much the same as `'gi.contracts/group/deleteChatRoom'`, which also doesn't use a queue because it's safe to do it immediately, unlike the similar code in `'gi.contracts/group/leaveChatRoom'`, which does use `queueInvocation` because one can re-join after leaving.

taoeffect: OK. If you're certain this is safe, then could you please add a comment here in the sideEffect explaining:

1. why it's safe to call `'gi.actions/chatroom/delete'` multiple times even when the contract doesn't exist because it's already been deleted
2. what exactly will happen when the SPMessage for `'gi.actions/chatroom/delete'` gets sent to a chatroom that doesn't exist.
3. why you've decided to avoid using `queueInvocation` and checking if the contract exists before sending this message.

Also, this:

> `console.log(`Error sending chatroom removal action for ${data.chatRoomID}`, e)`

Should be a `console.warn`. And it sounds like we'll be seeing a lot of those.

frontend/model/contracts/identity.js
```diff
@@ -359,14 +355,18 @@ sbp('chelonia/defineContract', {
     },
     'gi.contracts/identity/saveFileDeleteToken': {
       validate: objectOf({
+        billableContractID: stringMax(MAX_HASH_LEN, 'manifestCid'),
```
taoeffect: This should say `'billableContractID'` instead of `'manifestCid'`

frontend/model/contracts/chatroom.js
```diff
@@ -301,6 +301,18 @@ sbp('chelonia/defineContract', {
         //       a potential deadlock. See same warning in sideEffect for
         //       'gi.contracts/group/removeMember'
         await leaveChatRoom(contractID, state)
+        const me = sbp('state/vuex/state').loggedIn.identityContractID
```
taoeffect, line 304: I'm looking at the implementation of `leaveChatRoom`, and I'm not really seeing any code to leave the chatroom...

```js
export async function leaveChatRoom (contractID: string, state: Object) {
  if (await sbp('chelonia/contract/isSyncing', contractID, { firstSync: true })) {
    return
  }

  sbp('gi.actions/identity/kv/deleteChatRoomUnreadMessages', { contractID }).catch((e) => {
    console.error('[leaveChatroom] Error at deleteChatRoomUnreadMessages ', contractID, e)
  })
  sbp('okTurtles.events/emit', NEW_CHATROOM_UNREAD_POSITION, { chatRoomID: contractID })
  // NOTE: The contract that keeps track of chatrooms should now call `/release`
  // This would be the group contract (for group chatrooms) or the identity
  // contract (for DMs).
}
```

What happened here? 🤔 

Is this really all it's supposed to do?

Did it used to do something else?

And if this is all it does (and is supposed to do), maybe we should rename this function from `leaveChatRoom` to something like `removeChatRoomUnreadMessages`?

frontend/model/contracts/chatroom.js
```diff
@@ -301,6 +301,18 @@ sbp('chelonia/defineContract', {
...
+        if (me === state.attributes.creatorID || state.attributes.adminIDs.includes(me)) {
```
taoeffect, line 305: Are you sure we should be checking `state.attributes.adminIDs`?

Just because someone's an admin, doesn't mean they're paying for the contract, right? So `billableContractID` might not apply in that case

frontend/controller/actions/utils.js
```diff
@@ -372,7 +372,14 @@ export async function syncContractsInOrder (groupedContractIDs: Object): Promise
             await sbp('chelonia/contract/sync', contractID)
           } catch (e) {
             console.error(`syncContractsInOrder: failed to sync ${type}(${contractID}):`, e)
-            failedSyncs.push(`${type}(…${contractID.slice(-5)}) failed sync with '${e.message}'`)
+            if (e.name === 'ChelErrorUnexpectedHttpResponseCode' && e.message.startsWith('410:')) {
+              console.info('[syncContractsInOrder] Contract ID ' + contractID + ' has been deleted')
+              sbp('chelonia/contract/remove', contractID).catch(e => {
+                console.error('[syncContractsInOrder] Error handling contract deletion', e)
+              })
```
taoeffect: Will this call to `'chelonia/contract/remove'` here conflict with the same call to `'chelonia/contract/remove'` in `handleDeletedContract`?

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -551,11 +551,12 @@ sbp('okTurtles.events/on', NEW_CHATROOM_UNREAD_POSITION, ({ chatRoomID, messageH
 })
 
 sbp('okTurtles.events/on', NOTIFICATION_EMITTED, (notification) => {
+  const rootState = sbp('state/vuex/state')
   const rootGetters = sbp('state/vuex/getters')
   const icon = notification.avatarUserID && rootGetters.ourContactProfilesById[notification.avatarUserID]?.picture
     ? rootGetters.ourContactProfilesById[notification.avatarUserID].picture
     : notification.groupID
-      ? rootGetters.groupSettingsForGroup(notification.groupID).groupPicture
+      ? rootGetters.groupSettingsForGroup(rootState[notification.groupID]).groupPicture
```
taoeffect: Nice catch 👍

frontend/controller/app/identity.js
```diff
@@ -562,5 +562,30 @@ export default (sbp('sbp/selectors/register', {
...
+  'gi.app/identity/delete': async (contractID: string, wPassword: Secret<string>) => {
+    const password = wPassword?.valueOf()
+    const transientSecretKeys = []
+    let oldKeysAnchorCid
+
+    // If we're creating a new session, here we derive the IEK. This key (not
+    // the password) will be passed to the service worker.
+    if (password) {
+      try {
+        const [salt, cid] = await sbp('gi.app/identity/retrieveSalt', contractID, wPassword)
```
taoeffect, line 579: What is this `cid` thing?

In the selector it's `oldContractSalt`. A salt isn't a "cid" (contractID).

1. Can you rename it from `cid` to something else so that it's not confused with "contractID"?
2. Could you please update the `'gi.app/identity/retrieveSalt'` code to very clearly comment and explain what `oldContractSalt` is and how it's different from `contractSalt`?
3. Could you add a comment here explaining why it's needed here?

frontend/controller/actions/identity.js
```diff
@@ -960,6 +969,43 @@ export default (sbp('sbp/selectors/register', {
...
     sbp('chelonia/clearTransientSecretKeys', [oldIEKid, oldIPKid, IEKid, IPKid])
   },
+  'gi.actions/identity/delete': async ({
+    contractID,
+    transientSecretKeys,
+    oldKeysAnchorCid
+  }: {
+    contractID: string,
+    transientSecretKeys: Secret<string[]>,
+    oldKeysAnchorCid: string
```
taoeffect, line 981: Similar request here to rename `oldKeysAnchorCid` and also add a comment below above line 991 explaining what it is and what that code is doing.

frontend/controller/actions/identity.js
```diff
@@ -960,6 +969,43 @@ export default (sbp('sbp/selectors/register', {
...
+    await sbp('chelonia/out/deleteContract', contractID, {
+      [contractID]: { token: new Secret(token.valueOf()) }
+    })
+  },
+  'gi.actions/identity/_ondeleted': async (contractID: string, state: Object) => {
+    const ourIdentityContractId = sbp('state/vuex/getters').ourIdentityContractId
+
+    if (contractID === ourIdentityContractId) {
+      await sbp('gi.actions/identity/logout')
+    }
```
taoeffect: You said this code has been written in a way that it doesn't matter in what order the `_ondeleted` selectors are called.

Are you sure about this? Could you add a comment then if so explaining why it's safe to call `logout` and that doing so won't result in any conflicts with any other pending calls to `_ondeleted`?

frontend/controller/app/identity.js
corrideat: > What is this cid thing?

This is a CID / hash. https://gitlab.okturtles.org/okturtles/group-income-simple/-/wikis/E2E-Protocol/Password-Change.md

> 1. Can you rename it from cid to something else so that it's not confused with "contractID"?

Not sure, it's a CID

frontend/controller/actions/identity.js
corrideat: Again, this is a CID. I think the name is pretty descriptive in this case saying exactly what it is.

frontend/controller/app/identity.js
taoeffect: Oops, I didn't notice where `'gi.app/identity/retrieveSalt'` ended and my eyes went to the end of `'gi.app/identity/updateSaltRequest'`.


OK, if it's a CID, then can you explain here in a comment why we're grabbing it and whether or not it's different from `contractID` (which we already have)? If it's different, can you explain why, and if it's not different, can you explain why we're grabbing it when we already have `contractID`?

frontend/controller/actions/identity.js
taoeffect: Ah ok, then just:

> add a comment below above line 991 explaining what it is and what that code is doing.

frontend/controller/actions/identity.js
corrideat: What else are we supposed to do if our identity contract has been deleted? In any case, the calls to `_ondeleted` are for handling state, and there is no more state when our identity contract is deleted.

frontend/controller/actions/utils.js
corrideat: No, because calls to `chelonia/contract/remove` don't conflict with one another.

frontend/model/contracts/chatroom.js
corrideat: Agreed, but I don't see many options. What do you suggest we do? We have no way of knowing who the owner is, as it's not in the contract. The way it's written assumes that the owner will be either the creator or admin, so that they can issue the deletion command. I added the check to reduce the number of calls that will likely fail.

frontend/controller/actions/identity.js
taoeffect: If our identity contract has been deleted, we probably also want to still clear out the old state for chatrooms and groups still, right? How can we do that if we're logged out?

frontend/model/contracts/chatroom.js
corrideat: Yes, I think this is all it's supposed to do. At some point, I believe this was in `group.js` and had some more logic. Now, it's a function to carry out cleanup tasks related to leaving a chatroom (which, actually, we could probably implement outside of contracts using an event handler)

frontend/model/contracts/chatroom.js
taoeffect: If it's safe to do this then multiple times (and by running those code more frequently you're trying to guarantee at least someone who is the owner will successfully execute this code), then at the very least a comment is needed here explaining this situation. Explaining why some of the calls will fail and why that's expected and being done this way.

frontend/controller/actions/identity.js
```diff
@@ -232,6 +235,7 @@ export default (sbp('sbp/selectors/register', {
           ...publishOptions,
           headers: {
             ...publishOptions?.headers,
+            'shelter-deletion-token': deletionToken,
```
taoeffect: Would it make more sense, instead of sending the raw deletion token here, to send a hash of it instead, and then reveal the value to the server when we actually want to use it for deleting things?

frontend/controller/actions/chatroom.js
```diff
@@ -247,6 +247,47 @@ export default (sbp('sbp/selectors/register', {
...
+        if (!groupState.chatRooms[contractID].deletedDate) {
+          // If the chatroom hasn't been 'deleted' in the group, attempt to do
+          // so now.
+          await sbp('gi.actions/group/deleteChatRoom', {
+            contractID: cID,
+            data: { chatRoomID: contractID }
+          }).catch(e => {
+            console.warn(`[handleDeletedContract] ${e.name} thrown by gi.actions/group/deleteChatRoom ${cID} for ${contractID}:`, e)
+          })
+        }
```
taoeffect, line 283: One of the goals if you recall was to reuse the chatroom contract in other apps, including apps that might not have a group contract. Does this code break that requirement? If so, can it be written in a way that would still work with other apps that do not interact with the group contract?

frontend/controller/actions/identity.js
corrideat: Do you have some scenario in mind? I mean, maybe, but it complicates things and I'm not exactly sure what this is supposed to address.

We discussed this implementation at some point. The gist of it is that files have a deletion token that is server-generated. For contracts, we don't have that (we could, but we don't), and the reason to have it be client-generated in this case is to save a few roundtrips (the token is stored in the contract).

If we do the hash approach, then we need to handle that especially (compared to files) and I don't see what function it'd serve, since the server is capable of deleting data without knowing the token.

frontend/controller/actions/chatroom.js
corrideat: This isn't part of the contract so I don't see how the requirement is broken.

frontend/controller/actions/identity.js
corrideat: If our identity contract has been deleted, we can no longer use the app or log in, so clearing out the old state doesn't seem to be useful.

frontend/model/contracts/chatroom.js
corrideat: Sure. Should I remove the check and let 'anyone' attempt to delete it? Or leave it as-is, assuming that the owner will at least be an admin or creator?

frontend/controller/actions/identity.js
taoeffect: I suppose if this happens over an insecure connection it could be sniffed and used to cause DoS via deletion

frontend/controller/actions/identity.js
taoeffect: I guess we can leave as is, but maybe a comment explaining why it's safe to do so would be helpful

frontend/controller/actions/identity.js
corrideat: Maybe don't use insecure connections? We send this in the clear for files (server to client)

corrideat: If you feel strongly about this, we can maybe also make it a client-generated value for files and provide a hash

taoeffect: Sure, but why risk leaving it like this? It should be a minor change to eliminate this potential.

taoeffect: > If you feel strongly about this, we can maybe also make it a client-generated value for files and provide a hash

Yeah I don't have time right now to look at how files work, maybe we can look at it on the call tomorrow. If it's also a minor change then go for it

corrideat: Everything is a hash now. I don't think the insecure connection thing is that much of an issue, but I guess a more probable scenario is the server DB being leaked. This would at least make it harder to use that to start deleting things.

backend/database.js
```diff
@@ -152,6 +155,9 @@ export default async () => {
           }
         }
         await writeData(key, value)
+        Object.keys(prefixHandlers).forEach(prefix => {
+          cache.delete(prefix + key)
+        })
```
taoeffect: I do not understand what this is or why it's here...

1. Are you certain this is correct?
2. If so, could you please explain why it's correct and what it's doing in a comment above?

corrideat: See https://github.com/okTurtles/group-income/commit/b62d291d4b0fe211bcc49a422401b39c968611c0

taoeffect (changes requested):
## Testing

When the admin leaves the group, it should tell them that it will delete the group.

Also, when leaving a group as u1 with two members (u1 & u2), service worker logs show:

Screenshot 2025-03-24 at 12 05 53 PM

So that needs to be fixed, but also it should have warned me that this action would delete the group as well (since I was the admin).

Also, when clicking the `Delete account` button:

Screenshot 2025-03-24 at 12 22 46 PM

The way these buttons are supposed to work is that when they're performing a long running action they turn into a spinner that cannot be clicked, indicating that something is happening. See for example how the Sign Up button works.

However, this button did not turn into a spinner, and it appeared as if nothing was happening when I clicked it (even though stuff was happening). This needs to be fixed.

taoeffect:
## `GIChatroomAlreadyMemberError` error

Screenshot 2025-03-24 at 12 26 12 PM

`GIChatroomAlreadyMemberError` in the following situation:

1. u1 creates a group and invites u2
2. u1 deletes account
3. u2 creates their own group
4. then, in browser 2 (Brave), u2 logs in

taoeffect:
## `Invalid negative reference count` when deleting group

- u2, as the sole member in a group, leaves the group and deletes it using the leave group modal

SW console shows:

Screenshot 2025-03-24 at 12 31 25 PM

corrideat:
> The way these buttons are supposed to work is that when they're performing a long running action they turn into a spinner that cannot be clicked, indicating that something is happening. See for example how the Sign Up button works.

> However, this button did not turn into a spinner, and it appeared as if nothing was happening when I clicked it (even though stuff was happening). This needs to be fixed.

You're right, but AFAICT this _is_ implemented correctly (verified by adding an artificial delay). The spinner isn't visible because of how fast calling `/delete` is.

However, this isn't implemented correctly for group deletion. I'll fix it there.

shared/domains/chelonia/internals.js
```diff
@@ -1297,6 +1303,9 @@ export default (sbp('sbp/selectors/register', {
   },
   'chelonia/private/in/syncContract': async function (contractID: string, params?: { force?: boolean, resync?: boolean }) {
     const state = sbp(this.config.stateSelector)
+    if (state.contracts[contractID] === null) {
+      throw new Error('Cannot sync permanently deleted contract ' + contractID)
+    }
```
taoeffect: Won't this affect other code that relied on the previous `410` error messages?

corrideat: I'd think they're on different codepaths, but sure, I can prefix the error with `410: `

corrideat: (The reason they're different codepaths is that the `410:` is what triggers the permanent removal0

corrideat:
Regarding the other errors:

1. `GIChatroomAlreadyMemberError error`. As you can see, this happens on an _outgoing_ action and is part of the usual cross-sync process (i.e., keeping state consistent between contracts) and also unrelated to this PR.
2. Invalid negative reference count when deleting group: addressed by flagging deleted contracts to ignore these errors.
3. Failed to leave group warning: This seems like a bug (membersCount should be 2 in that case, I believe), but now that same action (admin leaving) will remove the group.

shared/domains/chelonia/internals.js
```diff
@@ -1297,6 +1303,9 @@ export default (sbp('sbp/selectors/register', {
   },
   'chelonia/private/in/syncContract': async function (contractID: string, params?: { force?: boolean, resync?: boolean }) {
     const state = sbp(this.config.stateSelector)
+    if (state.contracts[contractID] === null) {
+      throw new Error('410: Cannot sync permanently deleted contract ' + contractID)
```
taoeffect: The condition that's checked elsewhere is:

```js
if (e.name === 'ChelErrorUnexpectedHttpResponseCode' && e.message.startsWith('410:'))
```

I think the term for what's being done above with the manual `410:` insertion is "code smell".

This is really a flaky and not-future-proof way of doing this. If in 15 years someone is modifying this code or does something to remove the 410 because it's weird and out of place (say they decide to show the error message to the user and decide "what's this 410 thing doing here?" and remove it)... things could break.

The right thing to do is not a string comparison against `e.message` but a specific error type for this, against which you can use `e.name`. Then it's much clearer to the developer that this error actually has special meaning and may be used to trigger special code paths elsewhere.

So, instead of throwing `Error` here could you please throw a special error type?

Also, the check elsewhere is also weird:

```js
if (e.name === 'ChelErrorUnexpectedHttpResponseCode' && e.message.startsWith('410:'))
```

Note that it's still incompatible with the thrown error here because `e.name = 'Error'`, and, ideally this needs to be just checking just `e.name` instead of both `e.name` and `e.message`.

taoeffect:
Flow errors:

```
Error ---------------------------------------------------------------------- frontend/controller/actions/utils.js:330:24

Property `references` is missing in mixed [1]. [incompatible-use]

   330|       .forEach(([id, { references, type }]) => {
                               ^^^^^^^^^^


Error ---------------------------------------------------------------------- frontend/controller/actions/utils.js:330:36

Property `type` is missing in mixed [1]. [incompatible-use]

   330|       .forEach(([id, { references, type }]) => {
                                           ^^^^
```

taoeffect:
1. u1 created a group
2. invited u2
3. u2 joined and then left the group
4. u2 rejoined
5. u1 then left the group and deleted it

In u1's regular console:

Screenshot 2025-03-26 at 7 19 42 PM
Screenshot 2025-03-26 at 7 19 49 PM

Then several odd errors in the SW console, note especially the TypeError and the /retain error that also has a similar warning (plus some odd messageHandler errors):

Screenshot 2025-03-26 at 7 20 23 PM

taoeffect:
1. I created two groups with u1 and invited u2 to both of them.
2. Deleted u1's account

The good news is that everything seems to be working as it should.

I'm not sure about some of these errors though.

- We agreed that the `[messageHandler]` error needs to be a `console.info` on Slack
- Some of these other errors are related to retain counts - if you say to ignore them I can ignore them. If you see a way to either prevent them from being logged or making them warnings we should probably do that instead.

u2 browser console:

Screenshot 2025-03-27 at 10 21 28 AM

u2 SW console:

Screenshot 2025-03-27 at 10 22 26 AM

taoeffect:
> [handleDeletedContract] Error deleting saved tokens for deleted contract zLDXeQ2AgfCuVc6d7o1bVqVC1AUpLQ5zGxpy7vWs5uVdMLL5D1PDge6Y Error: Invalid negative ephemeral reference count

Screenshot 2025-03-27 at 11 54 58 AM

After the previous steps in the above comment, I created a group using u2 in Safari (in other words, u2 is signed in both in Safari and Firefox Dev). Then I created a user u3 (in FFDev container) and joined using the invite link u2's group. This is what I saw in u3's console ☝️

backend/routes.js
```diff
@@ -594,40 +637,80 @@ route.POST('/deleteFile/{hash}', {
...
+})
+
+route.POST('/deleteContract/{hash}', {
+  auth: {
+    // Allow file deletion, and allow either the bearer of the deletion token or
+    // the file owner to delete it
+    strategies: ['chel-shelter', 'chel-bearer'],
+    mode: 'required'
+  }
+}, async function (request, h) {
```
taoeffect, line 655: When an identity contract is being deleted, could you please `console.info` log:

1. The IP address
2. Username
3. ContractID

frontend/model/getters.js
```diff
@@ -515,7 +515,15 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
       })
   },
   seenWelcomeScreen (state, getters) {
-    return getters.ourProfileActive && getters.currentIdentityState?.groups?.[state.currentGroupId]?.seenWelcomeScreen
+    return (
+      (getters.currentIdentityState?.groups?.[state.currentGroupId] &&
+      getters.currentIdentityState?.groups?.[state.currentGroupId].hasLeft
```
taoeffect: Seems like this can be simplified to remove the `&&` with just:

```js
getters.currentIdentityState?.groups?.[state.currentGroupId]?.hasLeft
```

shared/domains/chelonia/chelonia.js
```diff
@@ -1198,6 +1225,69 @@ export default (sbp('sbp/selectors/register', {
...
+              ? `bearer ${(credentials[cid].token: any).valueOf()}`
+              // $FlowFixMe[incompatible-type]
+              // $FlowFixMe[incompatible-call]
+              : buildShelterAuthorizationHeader.call(this, credentials[cid].billableContractID)]
+        ])
+      })
+      if (!response.ok) {
+        if (response.status === 404 || response.status === 410) {
+          console.warn('Contract appears to have been deleted already', cid, response.status)
+          return
```
taoeffect, line 1287: Why are you logging `Contract appears to have been deleted already` for `404`?

corrideat: The response in this case will be 404 (but also, 404 does mean something maybe was deleted)

backend/routes.js
corrideat: Done now with reverse lookup

frontend/model/getters.js
corrideat: Indeed, it can. This is one of the few locations where we're looking at the positive value.

backend/routes.js
```diff
@@ -594,40 +637,86 @@ route.POST('/deleteFile/{hash}', {
...
+  // keys
+  try {
+    const [id] = sbp('chelonia.persistentActions/enqueue', ['backend/deleteContract', hash])
+    if (username) {
+      const ip = request.headers['x-real-ip'] || request.info.remoteAddress
+      console.info({ contractID: hash, username, ip, taskId: id }, 'Scheduled deletion on named contract')
+    }
+    // We return the queue ID to allow users to track progress
+    // TODO: Tracking progress not yet implemented
+    console.info('Scheduled task for ')
```
taoeffect: This `console.info` line seems to not really log anything...

backend/routes.js
```diff
@@ -594,40 +637,86 @@ route.POST('/deleteFile/{hash}', {
...
+})
+
+route.POST('/deleteContract/{hash}', {
+  auth: {
+    // Allow file deletion, and allow either the bearer of the deletion token or
+    // the file owner to delete it
+    strategies: ['chel-shelter', 'chel-bearer'],
+    mode: 'required'
+  }
+}, async function (request, h) {
```
taoeffect, line 655: Question: as this is a long-running process, should we immediately mark this contract as "disabled" or "don't touch" or whatever so that, for example, if someone tries to do a lookup on someone's identity contract (e.g. grabs their username and then fetches their contract and starts loading it) while this contract is being deleted, they are instead immediately presented with 410? Perhaps we should immediately mark the name=>CID lookup as dead?

Or do you think it's fine to leave as-is?

backend/server.js
taoeffect: and maybe we could shut down by default, and allow the server to run without shutting down if a flag like `--ignore-errors` is passed in?

backend/server.js
```diff
@@ -202,17 +207,138 @@ sbp('sbp/selectors/register', {
...
+      await sbp('chelonia.db/delete', kvIndexKey)
+
+      await sbp('chelonia.db/get', `_private_cid2name_${cid}`).then((name) => {
+        if (!name) return
+        return Promise.all([
+          sbp('chelonia.db/delete', `_private_cid2name_${cid}`),
+          appendToOrphanedNamesIndex(name)
+        ])
+      })
```
taoeffect, line 323: Question: what about the `name=` mapping? Should anything happen to that? (e.g. be set to `''`?)

taoeffect:
I don't know if this critique from Gemini 2.5 Pro is relevant, but including it here in case it is:

Screenshot 2025-03-27 at 8 51 24 PM

backend/server.js
corrideat: It could, but I don't see the need to do so.

backend/routes.js
corrideat: I think this is fine to leave as is. The contract will effectively become read-only soon (as it will be deleted after its descendants are scheduled to be deleted too) and then eventually deleted.

taoeffect (changes requested):
So, effectively I think this is what happened (although I did some stuff before this; let me know if you can't reproduce):

1. u3 created a group
2. u2 joined
3. u2 joined again in a different browser and deleted their account

For u3, while on the dashboard, u2 is still visible:

Screenshot 2025-03-28 at 10 22 36 AM

Switching to chat:

Screenshot 2025-03-28 at 10 23 03 AM

Switching back to Dashboard, now profile icon is missing, but u2 still there:

Screenshot 2025-03-28 at 10 23 07 AM

taoeffect:
While testing a setup saw this in u1's sw console:

Screenshot 2025-03-28 at 10 57 58 AM

What happened was that u3 was creating private chatrooms and public chatrooms and DMs and adding u1 to all of them.

The problem happens on this line:

Screenshot 2025-03-28 at 11 03 54 AM

taoeffect:
After a user leaves the group, pulling up the in-app notifications results in this error in the browser (**not** SW) console:

Screenshot 2025-03-28 at 11 08 06 AM

taoeffect:
Two more issues found:

## `Delete account` button becomes available when it shouldn't

The `Delete account` button becomes briefly available before the modal is closed. This could momentarily allow a user to click it again when they shouldn't be allowed. The Signup modal and Create Group buttons don't have this issue:

https://github.com/user-attachments/assets/5f2099e2-55fd-4dba-af9a-647958d9b11a

EDIT: Please verify this problem doesn't happen with the `Leave Group` and `Delete Group` buttons either.

## 3+ party DMs are deleted (Maybe DMs should be kept?)

I've tested and verified that public chatrooms, private chatrooms, and 2+ party DMs are deleted properly.

Does it make sense though to delete DMs when another person is still in them with you? I.e. if u1, u2, and u3 have a DM together, and u3 deletes their account, does it make sense to delete the entire conversation? I don't think so... Actually, I'm not sure it even makes sense to delete the conversation history when it's only 2 people in it initially too. It should probably work the same way as private chatrooms right? Those are handled correctly now btw. The person is just shown as having left the chatroom, but the conversation history is preserved.

EDIT: This seems to not be an issue — they're only deleted when they're owned by the user that deleted their account... might be something we could return to and re-evaluate later.

taoeffect:
I'm not sure this is an issue, but pointing it out anyway as it's an error:

u2 deleted their account (after creating a bunch of public & private chatrooms, as well as DMs), then in u2's SW console:

Screenshot 2025-03-28 at 11 50 00 AM

taoeffect:
In groups with users who have deleted their accounts, going to create a DM from a user that still exists in the group (and existed before the other user deleted their account), shows the old users in the list of users when creating a DM:

Screenshot 2025-03-28 at 11 54 12 AM
Screenshot 2025-03-28 at 11 53 52 AM

taoeffect (approved):
Heroic work @corrideat!!! 🥳 🎉 🥇 🏆

PR #2499, opened 2025-01-08
#2498 - Fix the link bug in the chat

### LINKED ISSUE #2498: When two links are put in the chat next to one another, there's no space between them
### Problem

```
https://okturtles.org https://groupincome.org
```

Renders as:

Screenshot 2025-01-07 at 5 24 27 PM

### Solution

Fix.

SebinSong:
closes #2498

### DIFF
--- a/frontend/assets/style/components/_custom-markdowns.scss
+++ b/frontend/assets/style/components/_custom-markdowns.scss
@@ -54,8 +54,8 @@
   }
 
   em {
-    margin-left: 1px;
-    margin-right: 1px;
+    margin-left: 2px;
+    margin-right: 2px;
   }
 
   ul,
@@ -137,6 +137,11 @@
 
   .link {
     border-bottom-color: currentColor;
+    margin-left: 2px;
+
+    &:has(+ *) {
+      margin-right: 2px;
+    }
   }
 
   .c-edited {
--- a/frontend/assets/style/components/_custom-markdowns.scss
+++ b/frontend/assets/style/components/_custom-markdowns.scss
@@ -54,8 +54,8 @@
   }
 
   em {
-    margin-left: 2px;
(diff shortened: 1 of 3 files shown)


frontend/assets/style/components/_custom-markdowns.scss
```diff
@@ -137,6 +137,11 @@
 
   .link {
     border-bottom-color: currentColor;
+    margin-left: 2px;
+
+    &:has(+ *) {
+      margin-right: 2px;
```
corrideat: I'm not sure using `margin` is the right solution here. Using `content: ' ';` would provide a space visually. However, no CSS solution will give a space when copying text, so probably the space should be part of the DOM. ETA: I've just tested this, and copying the text doesn't contain a space.

frontend/views/containers/chatroom/chat-mentions/chat-mentions-utils.js
```diff
@@ -87,9 +88,14 @@ function createRecursiveDomObjects (element: any): DomObject {
       nodeObj.children.push(createRecursiveDomObjects(child))
     }
 
-    nodeObj.children = nodeObj.children.filter(
-      child => child.tagName || (child.text || '').trim().length
```
SebinSong, line 91: @taoeffect Had more thorough investigation for the original issue and found out `trim()` here was the cause of the issue. As far as I remember, this was added to remove unnecessary `\n` created from `[DOMParser: parseFromString()](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString) API. So updated the logic here.

SebinSong:
@taoeffect Updated the PR with the better fix

taoeffect (approved):
Nicely done @SebinSong !!

PR #2504, opened 2025-01-12
Corrideat/task/#2479 string comparison issue

### LINKED ISSUE #2479: "Curly quotes" can prevent users from easily leaving group

corrideat:
Closes #2479

### DIFF
--- a/frontend/model/contracts/shared/giLodash.js
+++ b/frontend/model/contracts/shared/giLodash.js
@@ -78,6 +78,15 @@ export function randomHexString (length: number): string {
   return Array.from(randomBytes(length), byte => (byte % 16).toString(16)).join('')
 }
 
+export function normalizeString (str: string): string {
+  return str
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
Nice!

PR #2505, opened 2025-01-13
#2502 - Payment method max-char

### LINKED ISSUE #2502: Support BOLT12 Lightning invoices
### Problem

Bolt12 Lightning reusable payment request (BOLT12) invoices can be very long, for example at least 335 characters in my testing.

The problem is that the max characters that we allow for "Lightning" is 250 characters.

### Solution

I don't quite know what the maximum length is, so for now make it 1024 characters.

SebinSong:
closes #2502


taoeffect (approved):
Thx!

Issue #202, opened 2017-03-22
Roles & permissions


taoeffect:
### Problem

Some users will likely need to perform special functions for the group. It's likely that some of these functions should be restricted only to those users with the assigned role to perform them.

### Solution

Design a generic role + permissions system, where each user can be assigned a role that comes with certain permissions.

A "role" is a bundle of permissions.

A permission is a string that is internally interpreted by both the UI and the contracts as being allowed to perform certain actions.

Possibly useful stuff:

- https://github.com/leonardovilarinho/vue-acl

Silver-IT:
@taoeffect, @leihla. Seems like we need design for this.

SebinSong:
## Development Steps

**Figma link for the hypothetical UIs:** [here](https://www.figma.com/design/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=27350-54034&t=1gasEw69Na8Z0lET-0)

### Step 1. - Build 'Roles & Permissions' UI with dummy data ✅ 
Building **'Roles & Permissions'** section in **'Group Settings'** page, which will look something like below.
(NOTE: below is just an example design and the actual UI might look different from it.)

<img src='https://github.com/user-attachments/assets/b1d39393-3524-48b3-b320-9c1591ef3229' width='480'>

### Step 2. - Build 'Add Permissions' modal / Implement `group.js` action for it.  ✅ 
- Build 'Add Permissions' modal UIs
- Create a group contract action for adding permissions for a group member
- Display actual roles & permission details of group members in `GroupSettings.vue` (This table will only be exposed for a member who has been granted `view permissions` permission item.)

<img width="530" src="https://github.com/user-attachments/assets/e4791c25-b730-4d26-be27-724a2ee19d70" />

### Step 3. Build UIs / Implement 'Update Permissions' and 'Remove Permissions' ✅ 

- Build UIs / Implement front-end logics for 'Update Permissions' functionality

<img width="530"  src="https://github.com/user-attachments/assets/ada71330-3612-4a8b-8408-8cb0b1206215" />

- Build UIs / Implement front-end logics for 'Remove Permissions' functionality

<img width="530" src="https://github.com/user-attachments/assets/8411ca56-e3d9-4994-8f12-fad2dcfc8440" />

### Step 4. Conditionally enable features / functionalities based on permissions granted
- Eg. `Delete chatroom` functionality is only possible via members with `delete-channel` permission etc.

### Step 5. Write Cypress test suites for roles & permissions
[...]

PR #2507, opened 2025-01-14
#202 Step 1 - Add `Roles and permissions` section UI on GroupSettings page

SebinSong:
Step 1 of #202 Todo list: [here](https://github.com/okTurtles/group-income/issues/202#issuecomment-2567190831)

NOTE: This PR is UI implementation only. The data presented in the table are fake.

[ Desktop ]

[ Mobile ]


frontend/views/pages/GroupSettings.vue
```diff
@@ -80,6 +80,8 @@ page.c-page
           data-test='saveBtn'
         ) {{ L('Save changes') }}
 
+  roles-and-permissions
+
```
taoeffect, line 84: Can we have this section displayed only in development mode since it doesn't work and hide it in production?

taoeffect (changes requested):
Nice work @SebinSong! Preliminary review ready (I haven't had a chance to do an in-depth review today)

frontend/views/pages/GroupSettings.vue
SebinSong: Sure, updated in [here](https://github.com/okTurtles/group-income/pull/2507/commits/4eaec1e0432e4c0a3dec6b90f3d2b8cb7dada70a)

SebinSong:
Just made some additional updates:

- Better table structure for mobile size screen. (Combine 'user' and 'role' columns for smaller space)
- No need to expose 'See all' button when there is only one item to show.

frontend/model/contracts/shared/constants.js
```diff
@@ -59,6 +59,45 @@ export const STREAK_MISSED_PAYMENTS = 1
...
+  ADMIN: 'admin',
+  MODERATOR: 'moderator',
+  MODERATOR_DELEGATOR: 'moderator-delegator',
+  CUSTOM: 'custom'
+}
+
+export const GROUP_PERMISSIONS = {
+  VIEW_PERMISSIONS: 'view-permissions',
+  REMOVE_PERMISSIONS: 'remove-permissions',
+  DELEGATE_PERMISSIONS: 'delegate-permissions',
```
taoeffect: What is the idea as far as the difference between `REMOVE_PERMISSIONS` and `DELEGATE_PERMISSIONS`?

I think if you have the ability to give permissions (`DELEGATE_PERMISSIONS`), it should imply the ability to take them away (`REMOVE_PERMISSIONS`), what do you think?

frontend/model/contracts/shared/constants.js
```diff
@@ -59,6 +59,45 @@ export const STREAK_MISSED_PAYMENTS = 1
...
+  ],
+  MODERATOR_DELEGATOR: [
+    GP.VIEW_PERMISSIONS,
+    GP.REMOVE_MEMBER,
+    GP.REVOKE_INVITE,
+    GP.DELETE_CHANNEL
+  ],
+  MODERATOR: [
+    GP.VIEW_PERMISSIONS
+  ]
```
taoeffect: `MODERATOR` AND `MODERATOR_DELEGATOR` should have the same exact permissions, except that `MODERATOR_DELEGATOR` should have `GP.DELEGATE_PERMISSIONS`.

frontend/views/containers/group-settings/roles-and-permissions/PermissionActionMenu.vue
```diff
@@ -0,0 +1,84 @@
...
+        },
+        {
+          id: 'remove',
+          label: L('Remove'),
+          icon: 'trash-alt',
+          enabledFor: [GROUP_ROLES.ADMIN, GROUP_ROLES.MODERATOR_DELEGATOR]
+        }
+      ]
+
+      return list.filter(option => option.enabledFor.includes(this.role))
```
taoeffect: Instead of checking the roles here, we should be checking the permissions.

To be able to edit or remove permissions, you need both `DELEGATE_PERMISSIONS`, and then you can only edit permissions that you have.

By combining `DELEGATE_PERMISSIONS` and `REMOVE_PERMISSIONS`, we only need to check for `DELEGATE_PERMISSIONS`.

frontend/views/containers/group-settings/roles-and-permissions/RolesAndPermissions.vue
```diff
@@ -0,0 +1,163 @@
...
+    role: GROUP_ROLES.MODERATOR_DELEGATOR,
+    permissions: GROUP_PERMISSIONS_PRESET.MODERATOR_DELEGATOR
+  },
+  {
+    id: 'user-3',
+    username: 'Fake user 3',
+    role: GROUP_ROLES.MODERATOR,
+    permissions: GROUP_PERMISSIONS_PRESET.MODERATOR
+  }
+]
```
taoeffect: Can we see one for Custom too just to see what it looks like?

taoeffect (changes requested):
Great work @SebinSong!

Left a few comments and tested it, seems to work well on both desktop and mobile!

Can you also merge `master` since it's been a while?

frontend/views/containers/group-settings/roles-and-permissions/PermissionActionMenu.vue
SebinSong: Okay, will remove`REMOVE_PERMISSIONS` and use `DELEGATE_PERMISSIONS` everywhere.

SebinSong:
@taoeffect 

Updated the PR per your feedbacks.

### 1.
I removed `REMOVE_PERMISSIONS` as advised. (I think it's more clear to me now what `DELEGATE_PERMISSIONS` means). Also, as per below comment you left in the issue:

> However, it would not mean that the delegator could delegate the delegate-permission to others. Only the admin would be able to do this. In other words, if the admin granted me moderator (delegator) role, I would only be able to grant moderator permissions to others, but not moderator (delegator).

I created a `ASSIGN_DELEGATOR` permission entry and added it to `ADMIN` preset only.

### 2.
Added a dummy user for a custom role here:

taoeffect (approved):
Great work @SebinSong!

PR #2508, opened 2025-01-14
Offline cache

### LINKED ISSUE #2503: Offline-first PWA caching
### Problem

The PWA takes longer to open than it needs to because we're not taking advantage of offline-caching of pages.

### Solution

Get the site to work well offline and use the service worker's ability to cache requests to cache pages.

- Take extra care to make sure that the site still reloads when a new version is released and the old cache is invalidated
- Don't cache file attachments from chat

corrideat:


frontend/controller/app/identity.js
```diff
@@ -388,7 +388,9 @@ export default (sbp('sbp/selectors/register', {
             })
           } else {
             try {
-              await sbp('chelonia/contract/sync', identityContractID)
+              if (navigator.onLine !== false) {
+                await sbp('chelonia/contract/sync', identityContractID)
+              }
```
taoeffect: So... do we sync this once we do come back online? If so, where?

corrideat: No, we don't explicitly. It'd be handled like any other contract by that point.

taoeffect (commented):
Here's what I tried on Android:

1. I created the PWA and logged in with it (via the ngrok tunnel)
2. In the app launcher, I swiped up on the PWA app to kill it
3. Disabled phone's internet
4. Opened app again

Greeted with:

b

Followed by:

a

Is this something that's preventable? I.e. is it possible to still have the app load?

frontend/controller/serviceworkers/cache.js
```diff
@@ -0,0 +1,131 @@
+import sbp from '@sbp/sbp'
+import { NOTIFICATION_TYPE } from '~/shared/pubsub.js'
+
+const CACHE_VERSION = '1.0.0'
+const CURRENT_CACHES = {
+  assets: `assets-cache_v${CACHE_VERSION}`
+}
```
taoeffect: Should this version number be tied to some environment variable, e.g. `GI_VERSION` or something else?

corrideat: I was thinking about that. It could, but I don't think it needs to, so long as it's updated if breaking changes to the cache are made.

backend/routes.js
```diff
@@ -716,6 +716,7 @@ route.GET('/assets/{subpath*}', {
   // Files like `main.js` or `main.css` should be revalidated before use. Se we use the default headers.
   // This should also be suitable for serving unversioned fonts and images.
   return h.file(subpath)
+    .header('Cache-Control', 'public,max-age=604800,stale-while-revalidate=86400')
```
taoeffect, line 980: Hmm... why is this being added? I see that we have a special section for cached files above, so why should these be cached too when it seems like these are files that might change?

And if this is really necessary, could you add a comment here explaining why?

I want to avoid the app failing to update when files are updated...

taoeffect (changes requested):
Great work @corrideat! Review is ready.

One issue is that this seems to not work well still when offline. I found an app that seems to work when the PWA is force quit and restarted while offline, so it should be possible to improve the behavior:

- https://github.com/gokulkrishh/demo-progressive-web-app
- https://github.com/GoogleChromeLabs/squoosh

backend/routes.js
corrideat: Because if we don't store these in the cache, the app can't possibly work offline. Adding `stale-while-revalidate` means that it's fine to serve stale files while they're being revalidated (normally, all cached requests need to be revalidated).

corrideat: In any case, the comment is outdated, so I agree that this should be updated.

frontend/controller/serviceworkers/cache.js
```diff
@@ -103,23 +104,14 @@ if (
...
-                }
                 return cachedResponse
               }
             }
 
-            return fetch(request.clone?.() || request).then(async (response) => {
-              if (request instanceof URL) {
-                console.error('@@@@XX', 120, request, response.status)
-              }
+            return fetch(event.request.clone?.() || request).then(async (response) => {
```
corrideat: This is what seems to have fixed the issue. Note that I'm still using the original request here instead of the 'rewritten'  `request`.

frontend/controller/serviceworkers/cache.js
```diff
@@ -0,0 +1,149 @@
...
+          .match(request, { ignoreSearch: true, ignoreVary: true })
+          .then((cachedResponse) => {
+            if (cachedResponse) {
+              // If we're offline, return the cached response, if it exists
+              if (navigator.onLine === false) {
+                return cachedResponse
+              }
+            }
+
+            return fetch(event.request.clone?.() || request).then(async (response) => {
```
taoeffect: Can you add a comment here explaining what problem this `event.request.clone?.() || request` is fixing? If you're not sure why it works it's fine to say that, but it still needs a comment explaining why this was added in the first place since it falls into the category of "weirdness".

taoeffect (changes requested):
It's still getting stuck on something (stuck on the same white screen...)

taoeffect:
The error iOS Safari-based PWA gives:

IMG_8754

Doesn't happen with https://github.com/GoogleChromeLabs/squoosh

PR #2509, opened 2025-01-15
#2506 - Fix the redirection bug on `Send Message` button.

### LINKED ISSUE #2506: Broken `Send message` button on sidebar
### Problem

Steps to reproduce:

1. Create a group
2. Join the group as a different user
3. Go back to the user used to create a group and visit the group dashboard
4. Find the user from step 2 on the sidebar, hover over the profile and click on 'Send message'
5. We stay in the dashboard
6. Click on send message a second time
7. We are sent to the DM page as expected

#### What should happen

Instead of staying on the dashboard (step 5), we should be redirected to the DM page

#### What's actually happening

The DM chatroom is created (can verified by visiting the `Chat` page after step 5 but before step 6), but we stay on the dashboard. This could be due to either events not firing properly or due to missing or broken event handlers.

### Solution

Identify the issue and fix it so that we're redirected to the DM page without needing a second interaction.

### LINKED ISSUE #2510: Vue tooltip error when clicking "Send message" in profile button
### Problem

To reproduce:

1. In the `#general` chatroom, u2 clicks on u1's icon and clicks "Send Message":

Screenshot 2025-01-14 at 7 23 06 PM
Screenshot 2025-01-14 at 7 22 34 PM
Screenshot 2025-01-14 at 7 22 26 PM

### Solution

Fix.

SebinSong:
closes #2506 
closes #2510 

The logic to handle redirection to the DM once it's successfully created was missing there. So added it as the fix.

### DIFF
--- a/frontend/views/components/ProfileCard.vue
+++ b/frontend/views/components/ProfileCard.vue
@@ -181,9 +181,9 @@ export default ({
     async sendMessage () {
       const chatRoomID = this.ourGroupDirectMessageFromUserIds(this.contractID)
       if (!chatRoomID) {
-        const freshDMChatRoomID = await this.createDirectMessage(this.contractID)
-        if (freshDMChatRoomID) {
-          this.redirect(freshDMChatRoomID)
+        const freshChatRoomID = await this.createDirectMessage(this.contractID)
+        if (freshChatRoomID) {
+          this.redirect(freshChatRoomID)
         }
       } else {
         if (!this.ourGroupDirectMessages[chatRoomID].visible) {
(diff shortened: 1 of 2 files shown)


frontend/views/components/ProfileCard.vue
```diff
@@ -176,12 +176,15 @@ export default ({
       )
     },
     toggleTooltip () {
-      this.$refs.tooltip.toggle()
+      this.$refs.tooltip?.toggle()
```
SebinSong, line 179: @taoeffect I actually was not able to reproduce this error you attached:

image

But looking at the error stack trace which points to `sendMessage()`, it appears that some actions there like `this.redirect(...)` causes the `Tooltip.vue` to be destroyed earlier than it gets to `this.toggleTooltip()`. That makes `this.$refs.tooltip` become `undefined`. So I made a fix for it here.

(But let me keep testing to see if I can actually reproduce it)

taoeffect (approved):
Nice!

PR #2519, opened 2025-01-17
Closes #2517

### LINKED ISSUE #2517: Firefox Service Worker issue
### Problem

Image

While testing `master` with `GI_PERSIST=sqlite grunt dev` on http://localhost:3000, and then deleting all the IndexedDBs and unregistering the service worker, I now have two container tabs that have this issue. It seems they've become permanently broken as refreshing doesn't fix. I have no idea how to fix.

### Solution

Track down the bug.

corrideat:


frontend/model/database.js
```diff
@@ -26,26 +26,35 @@ const localforage = {
...
             }
-            const request = self.indexedDB.open(name + '--' + storeName)
 
-            // Create the object store if it doesn't exist
-            request.onupgradeneeded = (event) => {
-              const db = event.target.result
-              db.createObjectStore(storeName)
-            }
+            const openDB = (version?: number) => {
+              const request = self.indexedDB.open(name + '--' + storeName, version)
```
taoeffect: Could you please add a comment here explaining:

1. What this second parameter to `.open` is
2. Why this is needed
3. A reminder to users that if they get the `'DB is blocked'` error, to try closing all open tabs except one (and why that is necessary)

taoeffect (changes requested):
Excellent work @corrideat! Surprisingly this seemed to fix it, as long as there is only one tab open. One RFC:

frontend/model/database.js
```diff
@@ -26,26 +26,43 @@ const localforage = {
...
+              // Create the object store if it doesn't exist
+              request.onupgradeneeded = (event) => {
+                const db = event.target.result
+                db.createObjectStore(storeName)
+              }
+
+              request.onsuccess = (event) => {
+                const db = event.target.result
+                if (!db.objectStoreNames.contains(storeName)) {
+                  return openDB(db.version + 1)
```
taoeffect, line 48: `version` is initially `undefined`... and `undefined + 1` = `NaN`... so, is this right?

corrideat: `db.version` isn't the same as `version`.

taoeffect: The reference says:

> The version property of the [IDBDatabase](https://devdocs.io/dom/idbdatabase) interface is a [64-bit integer](https://firefox-source-docs.mozilla.org/nspr/index.html) that contains the version of the connected database. When a database is first created, this attribute is an empty string.

So... 

```
> '' + 1
'1'
> '1' + 1
'11'
```

Is this what you intended?

corrideat: No, that's not what I intended, but:

1. That's not an authoritative source. <https://w3c.github.io/IndexedDB/#database-construct>, which is, mentions no empty string.
2. Even then, it wouldn't matter because when the DB is first created, we can assume that `onupgradeneeded` will run, so this check wouldn't be needed to begin with. We only need this to trigger a version upgrade.
3. Even then, going from `1` to `11` to `111` &c. should be fine.

corrideat: I've also added a logging line to `db.version` and I only ever see a number, not `''`. I'm guessing an empty string isn't used (which makes sense, since it's supposed to be an integer type).

taoeffect (approved):
LGTM

PR #2524, opened 2025-01-19
Better error handling. Closes #2522.

### LINKED ISSUE #2522: Regression introduced in PendingApproval.vue
### Problem

When creating a group and joining it for the first time (while testing Sebin's PR #2521), I saw this:

Image

This is caused by recently merged code here:

Image

Note that this did not prevent the group from being joined eventually, but it is a bug that needs to be fixed.

### Solution

Fix the bug.

corrideat:


shared/domains/chelonia/localSelectors.js
```diff
@@ -90,5 +91,23 @@ export default (sbp('sbp/selectors/register', {
         sbp('okTurtles.events/emit', CONTRACTS_MODIFIED_READY, subscriptionSet, { added, removed })
       })
     })
+  },
+  'chelonia/externalStateWait': async function (contractID) {
```
taoeffect, line 98: Above line 95 could you please give a detailed description of:

1. What this function is doing
2. Why it exists, e.g. the sorts of scenarios it should be used in

shared/domains/chelonia/localSelectors.js
```diff
@@ -90,5 +91,23 @@ export default (sbp('sbp/selectors/register', {
...
+    await sbp('chelonia/contract/wait', contractID)
+    const { cheloniaState } = await sbp('chelonia/contract/fullState', contractID)
+    const localState = sbp(this.stateSelector)
+    if (cheloniaState.height >= localState.contracts[contractID]?.height) return
```
taoeffect: A comment above this line saying what it's checking would be helpful too

shared/domains/chelonia/localSelectors.js
```diff
@@ -90,5 +91,23 @@ export default (sbp('sbp/selectors/register', {
...
+
+    return new Promise((resolve) => {
+      const removeListener = sbp('okTurtles.events/on', EVENT_HANDLED_READY, (cID) => {
+        if (cID !== contractID) return
+
+        const localState = sbp(this.stateSelector)
+        if (cheloniaState.height >= localState.contracts[contractID]?.height) {
+          resolve()
+          removeListener()
```
taoeffect, line 116: This seems somewhat odd, I rarely see any code after `resolve()`... would it make sense to flip these two lines so that `removeListener()` is called first?

taoeffect (changes requested):
Review ready!

shared/domains/chelonia/localSelectors.js
corrideat: I don't think so. The idea here (and in general, I think) is to resolve when the promise should resolve. The cleanup code comes after `resolve()` because it doesn't affect the settlement of the promise. For example, if `removeListener()` should throw, it makes sense for the promise to still resolve. Investing the order would have a different effect.

taoeffect (approved):
Nice work @corrideat!

PR #2525, opened 2025-01-19
Remove unnecessary code after re-creating contracts

### LINKED ISSUE #2515: Delete old code that's not necessary
### Problem

There's some code we don't need after we wipe everything.

E.g. from #2514:

```js
  // For compatibility for old contracts.
  // TODO: This is to be deleted in a later release, once contracts are recreated.
  'state/vuex/commit': (key, data) => {
    switch (key) {
      case 'deleteChatRoomScrollPosition':
      case 'setChatRoomScrollPosition':
        return sbp('okTurtles.events/emit', NEW_CHATROOM_UNREAD_POSITION, data)
      // These are handled by events
      case 'removeNotification':
        return
      case 'setCurrentChatRoomId':
        return
    }

    throw new Error('Invalid selector `state/vuex/commit`: Not allowed in SW.')
  }
```

There might be some other code in the project we won't need anymore for the "Final breaking changes" milestone.

### Solution

Remove this code. For some of the old code, e.g. upgrade code, it would be better to comment it out instead of removing it, as it will remain as a useful example for when we need to perform upgrades again.

corrideat:
Closes #2515

EDIT (by @taoeffect): This PR contains may different useful examples of how to upgrade contracts and local state with Chelonia.


taoeffect:
This can be removed:

https://github.com/okTurtles/group-income/pull/2693/files#diff-12b5b0fa8291d2a411f51da50e9bf60ab4d619531faaa40784007f367c047d81R71-R75

corrideat:
Not sure it can, it says:

```
        // For backward- and forward-compatibility (`currentSyncs` may not be
        // defined in a different SW version)
```

taoeffect:
2.0 will be a blank slate for everyone... and everyone will have been running the latest version of the SW by the time we release it.

corrideat:
But that ignores the 'forward-compatibility' part of the comment. In the future, this `currentSyncs` thing could be done in a different way, making removing it a more difficult process.

taoeffect:
Can leave as-is then 👍

frontend/model/state.js
```diff
@@ -21,6 +19,7 @@ import chatroomModule from '~/frontend/model/chatroom/vuexModule.js'
 
 // Wrapper function for performing contract upgrades and migrations
 // TODO: Consider moving this function into a different file
+// eslint-disable-next-line no-unused-vars
```
taoeffect, line 24: Could you add to the comment here a link to this PR that says something like, "to see examples how to use this function, see this PR"?

corrideat: Added comment.

taoeffect (approved):
Nice 👍

Issue #2249, opened 2024-07-23
[UI Bug] Long username in Tooltip looks cropped.


SebinSong:
### Problem

<img src='https://github.com/user-attachments/assets/4c418c3f-b4e6-4e00-b8c4-b4517bcc4273' width='520'>

### Solution
Fix it.

PR #2531, opened 2025-01-21
#2520 - Sentence is cut off in the payment-streak tooltip 

### LINKED ISSUE #2520: Regression: tooltip showing streaks is cut off
### Problem

I checked on both Brave and Firefox:

Image

### Solution

While the starting position is correct now, the tooltip bubble doesn't show the full list. Fix it show that it does:

Image

SebinSong:
closes #2520 

**[ Fix screenshot ]**

---

This regression was introduced by the fix for #2249 which is something we must keep too. So as a fix, I created a boolean prop named `noEllipsis` that allows to opt out of this fix.

### DIFF
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -173,8 +173,8 @@ export default ({
             break
           case 'bottom-left':
             x = 0
-            y = '100%'
-            absPosition = { bottom: `-${spacing}px` }
+            y = `${spacing}px`
+            absPosition = { top: '100%' }
             break
           case 'bottom-right':
             x = '-100%'
@@ -391,6 +391,7 @@ export default ({
 
 .c-anchored-tooltip {
   width: max-content;
+  height: max-content;
 }
 
 .c-background {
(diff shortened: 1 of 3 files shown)


taoeffect (changes requested):
Nice work @SebinSong!

Can you please check all of the other places where this new flag is needed and add it?

For example, it's clearly needed here too:

Screenshot 2025-01-20 at 7 49 17 PM

SebinSong:
@taoeffect 
I can't reproduce that issue. As you can see from below screenshots below, multiple items are not cut off.



By using real-db:



I believe this issue has been fixed by below rule in [this previous commit](https://github.com/okTurtles/group-income/pull/2531/commits/84476a10bbc391983024ca6d0e2414380c1d64cc).

```
.c-anchored-tooltip {
  ...
  height: max-content;
}
```

Once it's ensured you are testing in `sebin/task/#2520-tooltip-cut-off` branch, It believe the bug is gone.

taoeffect (approved):
Tested again and it seems to work! Nice job!

PR #2532, opened 2025-01-21
#2518 - HTML rendering bug in `BannerGeneral.vue`

### LINKED ISSUE #2518: HTML in BannerGeneral
### Problem

Prior to @corrideat's fix in PR #2514 for issue #2513, in commit 4e19087340881c99ca9aeb20759ce9ba2b74fa6d, logging in from scratch as user `greg` results in a "serious error banner" that has HTML inlined to it:

Image

@corrideat fixed the bug causing the banner to appear, but not the bug causing HTML to appear in the banner instead of being rendered.

### Solution

Fix it so that HTML is rendered in the BannerGeneral (e.g. using `v-safe-html` or whatever).

SebinSong:
closes #2518 

**[ Fix screenshot ]**
Can see the `<a />` tag is rendered correctly in the DOM.

image

---

BTW, [This comment](https://github.com/okTurtles/group-income/issues/2518#issuecomment-2599149664) mentions that https://github.com/okTurtles/group-income/pull/2139 introduced this bug but that is not correct...

That PR is a fix for chatroom-specific bugs and `RenderMessageWithMarkdown.vue`(I built it) is not designed to be used inside a banner. So I think it's somewhere further up in the past commit history.

### DIFF
--- a/frontend/views/components/banners/BannerGeneral.vue
+++ b/frontend/views/components/banners/BannerGeneral.vue
@@ -7,19 +7,17 @@
       aria-live='polite'
     )
       i(:class='`icon-${ephemeral.icon} is-prefix`')
-      render-message-with-markdown(:text='ephemeral.message')
+      .c-banner-content(v-safe-html:a='ephemeral.message')
 </template>
 
 <script>
 import TransitionExpand from '@components/TransitionExpand.vue'
 import { debounce } from '@model/contracts/shared/giLodash.js'
-import RenderMessageWithMarkdown from '@containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js'
 
 export default ({
   name: 'BannerGeneral',
   components: {
-    TransitionExpand,
-    RenderMessageWithMarkdown
+    TransitionExpand
   },
   data: () => ({
     ephemeral: {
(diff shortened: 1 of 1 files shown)


frontend/views/components/banners/BannerGeneral.vue
```diff
@@ -7,19 +7,17 @@
...
     )
       i(:class='`icon-${ephemeral.icon} is-prefix`')
-      render-message-with-markdown(:text='ephemeral.message')
+      .c-banner-content(v-safe-html:a='ephemeral.message')
 </template>
 
 <script>
 import TransitionExpand from '@components/TransitionExpand.vue'
 import { debounce } from '@model/contracts/shared/giLodash.js'
-import RenderMessageWithMarkdown from '@containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js'
```
SebinSong, line 16: Using `RenderMessageWithMarkdown.vue` here is wrong (As it's a component built for chatroom) but regardless, it internally used `v-safe-html` custom directive which is the thing we need here. So I just replaced this component with a `<div />` that uses this directive.

taoeffect (approved):
Nice!

PR #2533, opened 2025-01-22
#2529 - Fix the chat regression

### LINKED ISSUE #2529: One more chat regression
### Problem

The spacing between blockquotes and lists is off now (it's too large). Example:

How it's supposed to look:

Image

How it looks now (incorrect):

Image

### Solution

Restore the proper spacing without breaking anything that was fixed before.

SebinSong:
closes #2529 

**[Fix screenshot - debugged using real DB data]**


SebinSong:
This fix does not break the previous fix:

image

taoeffect (approved):
Although it seems like a bit of a hack, it seems to work well. 🧐

Congrats on figuring it out!

PR #2544, opened 2025-01-24
Address infinite loop when fetching events

### LINKED ISSUE #2542: Deadlock on syncing logged in identity contract
### Problem

Can't load the site because my identity contract is stuck syncing in the SW:

Image

### Solution

It seems that `eventsAfter` is calling `fetchEventsStreamReader` on a wrong value that's resulting in an infinite loop.

It seems that the server's identity contract is at height 92 while the local state is at height 96 for some reason.

If this happens, freeze the state somehow, and offer the user the ability to rebuild it (#2080).

Explain to the user that something really bad has happened. It's a possible server attack to hide history, but more likely the server is missing data due to some data loss the server.

To test the fix you'll need to create a copy of the DB, write some data to the identity contract, then revert the DB to the copy.

EDIT: apparently there was a fork in the chain somehow! 😮

It seems the database on the server somehow got restored to an older version.

Note, there are two situations that this leads to and we need to handle them both:

1. On my end, I can't even log in because of the deadlock. It prevents loading the identity contract.
2. On the person I was DMing with's end, they are able to log in, but they aren't able to load the DM. It sits there spinning.

As they are different, both of these situations need to be detectable and recoverable.

We need to:

1. Tell the user there's been a fork.
2. Give them the option to rebuild the state. (In the case of my situation, they really have no choice...)

corrideat:


frontend/main.js
```diff
@@ -104,8 +104,19 @@ async function startApp () {
...
+    if (error?.name === 'ChelErrorForkedChain') {
+      const rootState = sbp('state/vuex/state')
+      const type = rootState.contracts[contractID].type || '(unknown)'
+      const retry = confirm(L('There was a serious issue when trying to sync the contract with ID {contractID} of type {type}. The data we received doesn\'t match what we have on file, which could mean there has been tampering or possibly a loss of data on the server. Unfortunately, we can\'t recover the original data. However, if you think this might be an error, you can try deleting your local records and syncing the contract again.\n\nDo you want to attempt this?', { contractID, type }))
+
+      if (retry) {
+        sbp('chelonia/contract/sync', contractID, { resync: true }).catch((e) => {
+          console.error('Error during re-sync', contractID, e)
+        })
+      }
```
taoeffect: A few concerns:

If this were to happen, it's very possible that not just one contract but multiple contracts will need to be re-sync'd. From a UI perspective, what would happen? They would see this message and have to answer it multiple times? What if they say "no" to resyncing one contract and "yes" to another? How can they change their mind and resync the contract they said "no" to originally?

My recommendation is to instead do the the following:

1. Instead of an `confirm`, make it an `alert`
2. Resync the contract (no asking users to decide this)
3. `console.error` here the details (i.e. which `contractID` it was; don't show this to users)
4. Then change the message to this:

```
The server's data for '{type}' has diverged from ours.
This can happen in extremely rare circumstances because of a problem on the server.
Group Income will attempt to automatically recover, but some recent messages or events may be missing.

If problems persist, please open the Troubleshooting page under the User Settings and resync all contracts.
```

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "group-income",
-  "version": "1.1.0",
+  "version": "1.2.1",
```
taoeffect: Please revert this file and merge `master`

frontend/main.js
corrideat: > If this were to happen, it's very possible that not just one contract but multiple contracts will need to be re-sync'd.

The thing is that isn't supposed to happen at all.

> From a UI perspective, what would happen? They would see this message and have to answer it multiple times?

With this implementation, yes.

> What if they say "no" to resyncing one contract and "yes" to another?

Then they'd be honoured as such

> How can they change their mind and resync the contract they said "no" to originally?

If they interact with the contract again, the same prompt will be shown.

----

1. Instead of an confirm, make it an alert
2. Resync the contract (no asking users to decide this)
3. console.error here the details (i.e. which contractID it was; don't show this to users)
4. Then change the message to this:

I'm not sure that this is a good idea, because it indicates a very serious issue that history has been rewritten. At best, it'd cause data loss. At worst, it could be malicious changes.

taoeffect: OK, we can leave it as an option. Still, please add the `console.error` to log the `contractID` (instead of showing it to the user), and here's another attempt at a message to show to the user:

```
The server's history for '{type}' has diverged from ours. This can happen in extremely rare circumstances due to either malicious activity or a bug.

To fix this, the contract needs to be resynced, and some recent events may be missing. Would you like to do so now?

(If problems persist, please open the Troubleshooting page under the User Settings and resync all contracts.)
```

frontend/main.js
```diff
@@ -104,8 +104,19 @@ async function startApp () {
...
+  sbp('okTurtles.events/on', SERIOUS_ERROR, (error, { contractID }) => {
     sbp('gi.ui/seriousErrorBanner', error)
+    if (error?.name === 'ChelErrorForkedChain') {
+      const rootState = sbp('state/vuex/state')
+      const type = rootState.contracts[contractID].type || '(unknown)'
+      const retry = confirm(L('There was a serious issue when trying to sync the contract with ID {contractID} of type {type}. The data we received doesn\'t match what we have on file, which could mean there has been tampering or possibly a loss of data on the server. Unfortunately, we can\'t recover the original data. However, if you think this might be an error, you can try deleting your local records and syncing the contract again.\n\nDo you want to attempt this?', { contractID, type }))
+
+      if (retry) {
+        sbp('chelonia/contract/sync', contractID, { resync: true }).catch((e) => {
+          console.error('Error during re-sync', contractID, e)
```
taoeffect, line 118: It would behoove us to let the user know about this. So in addition to the this `console.error`, can you please add an alert that says:

```
There was a problem resyncing the contract: {errMsg}

Please see the Application Logs under User Settings for more details. The Troubleshooting page in User Settings may be another way to fix the problem.
```

shared/domains/chelonia/internals.js
```diff
@@ -1286,7 +1286,7 @@ export default (sbp('sbp/selectors/register', {
           const { done, value: event } = await eventReader.read()
           if (done) {
             if (!latestHashFound) {
-              throw new ChelErrorUnrecoverable(`expected hash ${latestHEAD} in list of events for contract ${contractID}`)
+              throw new ChelErrorForkedChain(`expected hash ${latestHEAD} in list of events for contract ${contractID}`)
```
taoeffect, line 1289: (Commenting here because it's the closest place I can comment)

This code has been radically changed ... and it seems like I didn't review it properly when it was.

For example, there's a comment here from the old code that says:

```js
// remove the first element in cases where we are not getting the contract for the first time
```

But that's not what the code below it does.

That comment was talking about something important, which is now gone... what happened?

The comment needs to be reconciled with the code.

Next, there's what's actually there:

```js
for (let skip = has(state.contracts, contractID) && has(state.contracts[contractID], 'HEAD'); ; skip = false) {
```

What is this doing? This needs to be commented and clearly explained. I don't get it. What is this `skip` thing and why is it here?

shared/domains/chelonia/utils.js
```diff
@@ -686,7 +688,11 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
                   const hash = GIMessage.deserializeHEAD(currentEvent).hash
                   const height = GIMessage.deserializeHEAD(currentEvent).head.height
                   if (height !== sinceHeight || (sinceHash && sinceHash !== hash)) {
-                    controller.error(new Error('hash() !== since'))
+                    if (height === sinceHeight && sinceHash && sinceHash !== hash) {
+                      controller.error(new ChelErrorForkedChain('Forked chain: hash() !== since'))
+                    } else {
+                      controller.error(new Error('hash() !== since'))
+                    }
```
taoeffect: Why are we just logging the error and not throwing it?

Also, it's possible `hash() !== since` could be logged in other places in the future too so to avoid ambiguity, could the `console.error` messages please be updated to make it clear where this is happening?

Also, could you put the values into the parenthesis? E.g.

```
Forked chain: hash(${hash}) !== since(${sinceHash})
```

Also, above on line 681 we have a comment that's inconsistent with the code:

```js
// Check if the event limit is reached; if so, throw an error
```

No error is thrown.

shared/domains/chelonia/utils.js
```diff
@@ -731,6 +743,14 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
             }
             break
           }
+          case 'eod': {
+            if (remainingEvents === 0 || sinceHeight >= latestHeight) {
+              controller.close()
+            } else {
+              controller.error(new Error('Unexpected end of data'))
+            }
```
taoeffect: Similar question, is a `throw` needed here? If not, why not? Could a comment be added at the first instance of this pattern that explains why the error isn't thrown?

taoeffect (changes requested):
Great work @corrideat!

Review ready!

shared/domains/chelonia/internals.js
corrideat: The skip is doing this:

```
 remove the first element in cases where we are not getting the contract for the first time
```

shared/domains/chelonia/utils.js
corrideat: This isn't logging anything, it's in fact throwing an error.

See <https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultController/error>

shared/domains/chelonia/internals.js
taoeffect: Oh.

shared/domains/chelonia/utils.js
taoeffect: Oh, I misread `controller` as `console` 🤦 

What about this comment from that link you shared?

```
This method is rarely used, since usually it suffices to return a rejected promise
from one of the underlying sink's methods. However, it can be useful for suddenly
shutting down a stream in response to an event outside the normal lifecycle of
interactions with the underlying sink.
```

What's the reason you chose to do it this way as opposed to `return a rejected promise`? EDIT: and why is a `return` still necessary?

shared/domains/chelonia/internals.js
corrideat: See below:

```
          if (skip) continue
          await sbp('chelonia/private/in/handleEvent', contractID, event)
```

If `skip` is true (which only happens for the first event, if there's existing state), then `handleEvent` is not called.

shared/domains/chelonia/utils.js
corrideat: ETA: Maybe you misread `console` instead of `controller`?

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "group-income",
-  "version": "1.2.1",
+  "version": "1.1.0",
```
taoeffect: Can you revert this lockfile?

shared/domains/chelonia/utils.js
```diff
@@ -707,17 +707,22 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
                   const deserialized = GIMessage.deserializeHEAD(currentEvent)
                   sinceHeight = deserialized.head.height
                   sinceHash = deserialized.hash
+                  state = 'read-eos'
+                } else {
+                  // If the response came empty, assume there are no more events
+                  // after. Mostly this prevents infinite loops if a server is
+                  // claiming there are more events than it's willing to return
+                  // data for.
+                  state = 'eod'
```
taoeffect: I've requested two LLMs to assist with reviewing this file. Please check if they found anything meaningful:

### DeepSeek

The code has several critical bugs, particularly in the `eventsAfter` function. Here are the key issues and their fixes:

### 1. **Use of Node.js `Buffer` in Browser Environment**
**Issue:** The code uses `Buffer.from(value).toString()` which is Node.js-specific and not available in browsers, leading to runtime errors.

**Fix:** Replace `Buffer` with the browser's `TextDecoder`:

```javascript
const decoder = new TextDecoder();
buffer += decoder.decode(value, { stream: true }).trim();
```

### 2. **Incorrect Event Height and Hash Validation**
**Issue:** The first event's height is checked against `sinceHeight` instead of `sinceHeight + 1`, and the hash check is incorrect.

**Fix:** Update the checks to validate the event's `previousHEAD` and correct height:

```javascript
if (count === 0) {
  const deserialized = GIMessage.deserializeHEAD(currentEvent);
  const { hash, head: { height, previousHEAD } } = deserialized;
  if (height !== sinceHeight + 1) {
    throw new Error(`Expected event height ${sinceHeight + 1}, got ${height}`);
  }
  if (sinceHash && previousHEAD !== sinceHash) {
    throw new ChelErrorForkedChain(`Forked chain: expected ${sinceHash}, got ${previousHEAD}`);
  }
  sinceHash = hash;
  sinceHeight = height;
}
```

### 3. **Handling Empty Responses Gracefully**
**Issue:** Empty responses with higher `latestHeight` caused errors instead of advancing `sinceHeight`.

**Fix:** Update the 'eod' state to advance `sinceHeight` and refetch:

```javascript
case 'eod': {
  if (remainingEvents === 0 || sinceHeight >= latestHeight) {
    controller.close();
  } else {
    sinceHeight = latestHeight;
    state = 'fetch';
  }
  return;
}
```

### 4. **Proper Error Propagation**
**Issue:** Network errors in `fetchEventsStreamReader` weren't properly handled.

**Fix:** Ensure the stream controller propagates errors by rejecting the promise in `pull`.

---

**Corrected Code for `eventsAfter` Function:**

```javascript
export function eventsAfter (contractID: string, sinceHeight: number, limit?: number, sinceHash?: string, { stream }: { stream: boolean } = { stream: true }): ReadableStream | Promise<any[]> {
  if (!contractID) {
    throw new Error('Missing contract ID');
  }

  const fetchEventsStreamReader = async () => {
    requestLimit = Math.min(limit ?? MAX_EVENTS_AFTER, remainingEvents);
    const eventsResponse = await fetch(`${this.config.connectionURL}/eventsAfter/${contractID}/${sinceHeight}${Number.isInteger(requestLimit) ? `/${requestLimit}` : ''}`, { signal });
    if (!eventsResponse.ok) throw new Error('Unexpected status code');
    if (!eventsResponse.body) throw new Error('Missing body');
    latestHeight = parseInt(eventsResponse.headers.get('shelter-headinfo-height'), 10);
    if (!Number.isSafeInteger(latestHeight)) throw new Error('Invalid latest height');
    requestCount++;
    return eventsResponse.body.getReader();
  };

  const signal = this.abortController.signal;
  let requestCount = 0;
  let remainingEvents = limit ?? Infinity;
  let eventsStreamReader;
  let latestHeight;
  let state = 'fetch';
  let requestLimit, count;
  let buffer = '';
  let currentEvent;
  const decoder = new TextDecoder(); // Initialize TextDecoder once

  const s = new ReadableStream({
    async pull(controller) {
      for (;;) {
        try {
          switch (state) {
            case 'fetch': {
              eventsStreamReader = await fetchEventsStreamReader();
              state = 'read-new-response';
              count = 0;
              break;
            }
            case 'read-eos':
            case 'read-new-response':
            case 'read': {
              const { done, value } = await eventsStreamReader.read();
              if (done) {
                if (remainingEvents === 0 || sinceHeight >= latestHeight) {
                  controller.close();
                  return;
                } else if (state === 'read-new-response' || buffer) {
                  throw new Error('Invalid response: done too early');
                } else {
                  state = 'fetch';
                  break;
                }
              }
              buffer += decoder.decode(value, { stream: true }).trim();
              if (state === 'read-new-response') {
                if (buffer[0] !== '[') throw new Error('Invalid response: no array start');
                buffer = buffer.slice(1);
              } else if (state === 'read-eos') {
                throw new Error('Invalid data after end of stream');
              }
              state = 'events';
              break;
            }
            case 'events': {
              const nextIdx = buffer.search(/(,|\])/);
              if (nextIdx < 0) {
                state = 'read';
                break;
              }
              let enqueued = false;
              const eventValue = buffer.slice(0, nextIdx).trim();
              if (eventValue) {
                if (count === requestLimit) throw new Error('Too many events');
                currentEvent = b64ToStr(JSON.parse(eventValue));
                if (count === 0) {
                  const deserialized = GIMessage.deserializeHEAD(currentEvent);
                  const { hash, head: { height, previousHEAD } } = deserialized;
                  if (height !== sinceHeight + 1) {
                    throw new Error(`Expected height ${sinceHeight + 1}, got ${height}`);
                  }
                  if (sinceHash && previousHEAD !== sinceHash) {
                    throw new ChelErrorForkedChain(`Forked chain: expected ${sinceHash}, got ${previousHEAD}`);
                  }
                  sinceHash = hash;
                  sinceHeight = height;
                }
                if (count++ !== 0 || requestCount !== 0) {
                  controller.enqueue(currentEvent);
                  enqueued = true;
                  remainingEvents--;
                }
              }
              if (buffer[nextIdx] === ']') {
                state = 'read-eos';
                buffer = buffer.slice(nextIdx + 1).trim();
              } else {
                buffer = buffer.slice(nextIdx + 1).trimStart();
              }
              if (enqueued) return;
              break;
            }
            case 'eod': {
              if (remainingEvents === 0 || sinceHeight >= latestHeight) {
                controller.close();
              } else {
                sinceHeight = latestHeight;
                state = 'fetch';
              }
              return;
            }
          }
        } catch (e) {
          controller.error(e);
          break;
        }
      }
    }
  });

  return stream ? s : collectEventStream(s);
}
```

**Summary of Fixes:**
- Replaced `Buffer` with `TextDecoder` for browser compatibility.
- Corrected height and previous hash validation to ensure chain continuity.
- Handled empty responses by advancing `sinceHeight` to `latestHeight`.
- Improved error propagation to handle network issues and invalid responses.

### Claude 3.5 Sonnet

I've reviewed the code with particular focus on `eventsAfter` and `fetchEventsStreamReader`. Here are a few potential issues and suggestions:


1. In `eventsAfter`, there's a potential issue with error handling in the stream processing:

```javascript:shared/domains/chelonia/utils.js
// ... existing code ...
case 'read-eos': // End of stream case
case 'read-new-response': // Just started reading a new response
case 'read': { // Reading from the response stream
  const { done, value } = await eventsStreamReader.read()
  // If done, determine if the stream should close or fetch more
  // data by making a new request
  if (done) {
    // No more events to process or reached the latest event
    // Using `>=` instead of `===` to avoid an infinite loop in the
    // event of data loss on the server.
    if (remainingEvents === 0 || sinceHeight >= latestHeight) {
      controller.close()
      return
    } else if (state === 'read-new-response' || buffer) {
      // If done prematurely, throw an error
      throw new Error('Invalid response: done too early')
    } else {
      // If there are still events to fetch, switch state to fetch
      state = 'fetch'
      break
    }
  }
// ... existing code ...
```


The error handling here could be improved by:

- Adding cleanup of the eventsStreamReader when errors occur
- Adding proper error handling for network failures during stream reading
- Adding timeout handling for long-running stream reads



3. There's a potential memory leak in the event processing loop:

```javascript:shared/domains/chelonia/utils.js
case 'events': {
  const nextIdx = buffer.search(/(?<=\s*)[,\]]/)
  if (nextIdx < 0) {
    state = 'read'
    break
  }
  // ... existing code ...
  buffer = buffer.slice(nextIdx + 1).trimStart()
  // ... existing code ...
}
```


The `buffer` could potentially grow very large if the stream contains malformed data. Consider adding a maximum buffer size check:

```js
case 'events': {
  const nextIdx = buffer.search(/(?<=\s*)[,\]]/)
  if (nextIdx < 0) {
    state = 'read'
    break
  }
  // ... existing code ...
  buffer = buffer.slice(nextIdx + 1).trimStart()
  // ... existing code ...
}
```


The `buffer` could potentially grow very large if the stream contains malformed data. Consider adding a maximum buffer size check:

```js
const MAX_BUFFER_SIZE = 1024 * 1024 // 1MB

case 'events': {
  if (buffer.length > MAX_BUFFER_SIZE) {
    throw new Error('Buffer size exceeded maximum allowed size')
  }
  const nextIdx = buffer.search(/(?<=\s*)[,\]]/)
  // ... existing code ...
}
```


4. The stream error handling could be improved by adding a try/catch around the main processing loop:

```js
async pull (controller) {
  try {
    for (;;) {
      // ... existing code ...
    }
  } catch (e) {
    controller.error(e)
    throw e
  } finally {
    if (eventsStreamReader) {
      await eventsStreamReader.cancel()
    }
  }
}
```

taoeffect (approved):
Nice!

PR #2546, opened 2025-01-25
More general API for periodic notification interval

### LINKED ISSUE #2527: Create generalized API for requesting periodic push notifications
### Problem

This code needs to only run for requested push subscriptions:

```js
// Recurring task to send messages to push clients (for periodic notifications)
setInterval(() => {
  const pubsub = sbp('okTurtles.data/get', PUBSUB_INSTANCE)
  // Notification text
  const notification = JSON.stringify({ type: 'recurring' })
  // Find push subscriptions that do _not_ have a WS open. This means clients
  // that are 'asleep' and that might be woken up by the push event
  Object.values(pubsub.pushSubscriptions || {})
    .filter((pushSubscription: Object) => pushSubscription.sockets.size === 0)
    .forEach((pushSubscription: Object) => {
      postEvent(pushSubscription, notification).catch((e) => {
        console.warn(e, 'Error sending recurring message to web push client', pushSubscription.id)
      })
    })
// Repeat every 12 hours
}, 12 * 60 * 60 * 1000)
```

### Solution

Implement it in a generic way so that apps can specify when registering push notifications that the server periodically send these to them.

Note: there should be a minimum request period that you can't go below, e.g. no app should be able to request periodic notifications more frequently than say every 5 minutes. In other words, the server should enforce this limit, and return an error to clients that want more frequent notifications. The error message should include the server's configured minimum time, so that if some server say wants it to be every 30 minutes, apps can figure out and adjust to this new minimum time as needed. There could also be a new RESTful GET API (e.g. `/config`) that returns the server's configuration values such as minimum notification time. Alternatively, the server can do `Math.max(requestedTime, SERVER_MIN_PERIODIC_TIME)`, with some sort of warning returned (this way is probably simpler and better..).

corrideat:
Closes #2527.


backend/push.js
```diff
@@ -126,11 +128,14 @@ export const subscriptionInfoWrapper = (subcriptionId: string, subscriptionInfo:
...
     },
+    'settings': {
+      value: extra.settings || {}
+    },
     'sockets': {
       value: new Set()
     },
     'subscriptions': {
-      value: new Set(channelIDs)
+      value: new Set(extra.channelIDs)
```
taoeffect, line 121: `o3-mini-high` suggests:

```js
  value: new Set(extra.channelIDs || [])
```

taoeffect (commented):
This is not related to these changes but surrounding code — both `o3-mini-high` and `Gemini 2.5 Pro` complained about the `deleteSubscriptionFromIndex` implementation:

## `o3-mini-high`

<details>
In deleteSubscriptionFromIndex the code manually slices the index string. This logic does not reliably remove the target ID and may leave a leading or trailing “\x00” (which in turn would create an empty string in the array after splitting). For instance, the original code reads:

```js
const deleteSubscriptionFromIndex = async (subcriptionId: string) => {
  await sbp('okTurtles.eventQueue/queueEvent', 'update-webpush-indices', async () => {
    const currentIndex = await sbp('chelonia.db/get', '_private_webpush_index')
    const index = currentIndex.indexOf(subcriptionId)
    if (index === -1) return
    const updatedIndex = currentIndex.slice(0, index > 1 ? index - 1 : 0) +
      currentIndex.slice(index + subcriptionId.length)
    await sbp('chelonia.db/set', '_private_webpush_index', updatedIndex)
  })
}
```

A more robust approach is to split the string on “\x00”, filter out the subscriptionId, and then join back. For example (also using the corrected parameter name):

```js
const deleteSubscriptionFromIndex = async (subscriptionId: string) => {
  await sbp('okTurtles.eventQueue/queueEvent', 'update-webpush-indices', async () => {
    const currentIndex = await sbp('chelonia.db/get', '_private_webpush_index')
    const ids = currentIndex.split('\x00')
    const filteredIds = ids.filter(id => id !== subscriptionId)
    await sbp('chelonia.db/set', '_private_webpush_index', filteredIds.join('\x00'))
  })
}
```
</details>

---

Also it noticed that in `addSubscriptionToIndex` the `subcriptionId` parameter is misspelled and should be `subscriptionId`.

## `Gemini 2.5 Pro`

<details>

**`backend/push.js` - Fragile Index Deletion Logic**

The logic for removing an entry from the null-byte separated index string could be error-prone, especially with edge cases (e.g., if `subscriptionId` somehow contained `\x00`, though unlikely). Using array methods after splitting might be more robust.

```diff
--- a/backend/push.js
+++ b/backend/push.js
@@ -24,8 +24,11 @@
const deleteSubscriptionFromIndex = async (subcriptionId: string) => {
  await sbp('okTurtles.eventQueue/queueEvent', 'update-webpush-indices', async () => {
    const currentIndex = await sbp('chelonia.db/get', '_private_webpush_index')
-    const index = currentIndex.indexOf(subcriptionId)
-    if (index === -1) return
-    const updatedIndex = currentIndex.slice(0, index > 1 ? index - 1 : 0) + currentIndex.slice(index + subcriptionId.length)
+    // More robust index update: split, filter, join
+    if (!currentIndex) return // Index doesn't exist or is empty
+    const ids = currentIndex.split('\x00')
+    const updatedIds = ids.filter(id => id !== subcriptionId)
+    // Avoid writing back if the ID wasn't found
+    if (updatedIds.length === ids.length) return
+    const updatedIndex = updatedIds.join('\x00')
    await sbp('chelonia.db/set', '_private_webpush_index', updatedIndex)
  })
}

```
</details>

taoeffect (changes requested):
Looks good! Minor changes requested by LLMs.

backend/push.js
corrideat:
```
[[], null, undefined, ''].map(v => [...new Set(v)])

> (4) [Array(0), Array(0), Array(0), Array(0)]
````

So, I don't think this change is needed. It would be if maybe `extra.channelIDs` could be `false`, but I don't think it can.

corrideat:
I'm dismissing the LLM reviews as the code is fine as is and there's no need to use arrays. They did however catch a misspelling that I'll correct (`subcriptionId` should be `subscriptionId`).

That said, we now have the index management factory functions, so I'll also use that.

taoeffect (approved):
Nicely done!

PR #2547, opened 2025-01-25
Add Grunt file option to disable slim contract builds

### LINKED ISSUE #2390: Disable slim build
### Problem

Recent changes by @corrideat have made it so that the full compiled contracts and their slim versions are about the same size.

If this is indeed so, there's no reason to build or load the slim versions, as it slows down the build time.

### Solution

1. Verify the difference in size between the contracts. If it's negligible, proceed to the next step.
2. Update the build process so that a constant is defined at the top that specifies whether or not to build and sign the slim version.
3. Use this constant in the build process. When it's true, build and sign the slim version, when it's false, don't.
4. Verify everything works still and the tests pass.

### Update 2025-01-26

It turns out that this is actually a really hard problem to solve correctly.

Please see [this conversation from Slack](https://okturtles.slack.com/archives/C0EH7P20Y/p1737916024993889) between @corrideat and myself regarding the changes in #PR 2547:

Image

In summary:

- We cannot disable the slim build right now because the contracts use the `L` function which in turn relies on selectors like `translations/init`, and these don't exist at the time the contracts are loaded, so that requires calling `sbp/selectors/register`, which isn't currently allowed in contracts
- Allowing `sbp/selectors/register` in contracts right now would introduce potential security issues to the system as it would allow contracts to potentially register selections for other contracts or other selectors in the app (by registering them first), and we want to create a system that is safe to load arbitrary, and even untrusted contracts.
- For translations to work, we cannot use the non-slim contract at all, as it requires zero dependencies on external dynamic data (which is what translations are). For the non-slim version of contracts to work, it would have to be English-only, but for the app we need the dynamic version.

snowteamer:
Closes #2390

### Summary of changes
- New `BUILD_SLIM_CONTRACTS` top-level constant in `Gruntfile.js`. Defaults to `true`.
- SBP selector `'sbp/selectors/register'` now allowed in contracts.
    This change was not intentional, but a way to allow non-slim contract bundles to load `translations.js` (which registers the `'translations/init'` selector).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,8 +19,8 @@ const chalk = require('chalk')
 const crypto = require('crypto')
 const { exec, fork } = require('child_process')
 const execP = util.promisify(exec)
-const { readdir, cp, mkdir, access, rm, copyFile, readFile } = require('fs/promises')
 const fs = require('fs')
+const { readdir, cp, mkdir, access, rm, copyFile, readFile } = require('fs/promises')
 const path = require('path')
 const { resolve } = path
 const packageJSON = require('./package.json')
@@ -37,6 +37,8 @@ const packageJSON = require('./package.json')
 // Not loading babel-register here since it is quite a heavy import and is not always used.
 // We will rather load it later, and only if necessary.
 // require('@babel/register')
+
+const BUILD_SLIM_CONTRACTS = true
 
 const {
   CI = '',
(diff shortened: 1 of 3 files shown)


taoeffect (commented):
@snowteamer Based on a conversation with @corrideat on Slack we're realizing this issue is way harder to address (perhaps impossible) than we originally realized.

I've updated the original issue with the following:

It turns out that this is actually a really hard problem to solve correctly.

Please see [this conversation from Slack](https://okturtles.slack.com/archives/C0EH7P20Y/p1737916024993889) between @corrideat and myself regarding the changes in #PR 2547:

Image

In summary:

- We cannot disable the slim build right now because the contracts use the `L` function which in turn relies on selectors like `translations/init`, and these don't exist at the time the contracts are loaded, so that requires calling `sbp/selectors/register`, which isn't currently allowed in contracts
- Allowing `sbp/selectors/register` in contracts right now would introduce potential security issues to the system as it would allow contracts to potentially register selections for other contracts or other selectors in the app (by registering them first), and we want to create a system that is safe to load arbitrary, and even untrusted contracts.
- For translations to work, we cannot use the non-slim contract at all, as it requires zero dependencies on external dynamic data (which is what translations are). For the non-slim version of contracts to work, it would have to be English-only, but for the app we need the dynamic version.

PR #2552, opened 2025-01-27
2080 implement troubleshooting page re sync contracts

### LINKED ISSUE #2080: Implement Troubleshooting page (Re-sync contracts)
### Problem

The troubleshooting page is currently unimplemented.

Screenshot 2024-06-17 at 9 56 43 AM

### Solution

- When the re-sync button is clicked, we need to resync and rebuild all of the contracts state. If the contracts need to be sync'd in a specific order, make sure to do that.
- Display status updates next to the "Status" text
  - Turn the square a yellow/orange color while is ongoing
  - Make it red in the case that the resync failed
  - Make it green again in the case it succeeded

--------------

### Update 2025-01-27

Actually, a simpler solution that would be less brittle and easier to implement would be to have this page delete the state and log the user out.

So:

1. Update the "Re-Sync" button to say, "Reset Group Income"
2. When they click it, let them know they'll be logged out and ask them to confirm if they're sure.

NOTE: only reset the app's state for the logged in user (not any other logged out users). Any forks are most likely to affect the logged in user only as the logged out users probably haven't synced the contracts recently.

corrideat:


frontend/views/containers/user-settings/Troubleshooting.vue
```diff
@@ -104,41 +74,15 @@ export default ({
...
-
-        // Change false to true to force a dummy error
-        if (false) { // eslint-disable-line
-          throw Error('Dummy error forced')
-        }
-
-        await this.dummy3secTask()
-
-        // All done!
+        this.ephemeral.status = 'recovering'
```
taoeffect, line 87: Per the issue, we need:

> When they click it, let them know they'll be logged out and ask them to confirm if they're sure.

In other words the text above on line 78 needs to be changed to something like:

```
This will reset Group Income and log you out.

You will need to log back in with your password and wait for a bit while the app's state is being rebuilt from scratch.

Are you sure?
```

frontend/views/containers/user-settings/Troubleshooting.vue
```diff
@@ -7,43 +7,19 @@
         | &nbsp;
         i18n.link(tag='button' @click='openAppLogs') See application logs
 
-      ul.c-legend
```
taoeffect, line 10: Need to update the text above too.

Make it say something like:

```
If you're encountering problems, you can try resetting Group Income.

This will delete any problematic state that's been built-up and log you out. Logging back in will take a bit of time to rebuild this state, but it should fix most problems.
```

taoeffect (changes requested):
Two minor text requests

frontend/controller/app/identity.js
```diff
@@ -462,24 +463,35 @@ export default (sbp('sbp/selectors/register', {
...
 
-        await sbp('state/vuex/save', true, state)
+          await sbp('state/vuex/save', true, state)
+        }
         await sbp('gi.db/settings/deleteStateEncryptionKey', encryptionParams)
         await sbp('appLogs/pauseCapture', { wipeOut: true }) // clear stored logs to prevent someone else accessing sensitve data
       }
+      // These should already be deleted, but we attempt to delete them again,
+      // just in case.
+      if (wipeOut) {
```
taoeffect, line 488: What does this comment mean, that it should "already be deleted"?

We don't delete the state permanently unless `wipeOut` is set... that least we're not supposed to, otherwise it would cause a long login each time.

frontend/views/containers/user-settings/Troubleshooting.vue
dotmacro: The suggested text is clear but not layman-friendly. How about something more like:

```
If you're having trouble with the app, you can try resetting Group Income.

This will delete the current app state, and log you out. After you log back in, it may take a few minutes for the app to reset, but that should fix most problems.

THIS WILL LOG YOU OUT.
```

taoeffect: Sounds fine 👍

frontend/controller/app/identity.js
corrideat: Yes, we do delete all of those keys when logging out. It doesn't cause a long login each time because the state is loaded from the encrypted store.

taoeffect (approved):
Great work @corrideat!

To help us quickly release the next update, I ended up pushing some commits to update those strings. I tested this PR and it works well!

PR #2554, opened 2025-01-28
#2528 Missed-payments streak issue

### LINKED ISSUE #2528: "0 members have missed payments" when it's not true
### Problem

Last month we clearly had missed payments:

Image

And yet:

Image

Note: the "on-time payment streaks" was correctly updated to remove a payment streak from the member that had one (and now missed payments).

Also, as a separate issue, this text could be improved to be clearer, as it's not clear if it's referring to the sender or receiver.

### Solution

Fix this stat so that it displays members who did not pay in the previous payment period.

Also, update the text to add "their" to make it clearer:

```
{span_}{membercount} members{_span} have {strong_}missed their payments{_strong}
```

SebinSong:
closes #2528 

I can see it now shows 1 missed-payment items too.

### DIFF
--- a/frontend/views/containers/dashboard/GroupMembersActivity.vue
+++ b/frontend/views/containers/dashboard/GroupMembersActivity.vue
@@ -164,8 +164,8 @@ export default ({
           ? L('{span_}1 member{_span} hasn\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails'])
           : L('{span_}{membercount} members{_span} haven\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails']),
         'missedPayments': this.missedPayments.length === 1
-          ? L('{span_}1 member{_span} has {strong_} missed payments {_strong}', argsMap['missedPayments'])
-          : L('{span_}{membercount} members{_span} have {strong_} missed payments {_strong}', argsMap['missedPayments']),
+          ? L('{span_}1 member{_span} has {strong_} missed their payments {_strong}', argsMap['missedPayments'])
+          : L('{span_}{membercount} members{_span} have {strong_} missed their payments {_strong}', argsMap['missedPayments']),
         'noVotes': this.noVotes.length === 1
           ? L('{span_}1 member{_span} hasn\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
           : L('{span_}{membercount} members{_span} haven\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
--- a/frontend/views/containers/dashboard/GroupMembersActivity.vue
+++ b/frontend/views/containers/dashboard/GroupMembersActivity.vue
@@ -164,8 +164,8 @@ export default ({
           ? L('{span_}1 member{_span} hasn\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails'])
           : L('{span_}{membercount} members{_span} haven\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails']),
         'missedPayments': this.missedPayments.length === 1
-          ? L('{span_}1 member{_span} has {strong_} missed their payments {_strong}', argsMap['missedPayments'])
-          : L('{span_}{membercount} members{_span} have {strong_} missed their payments {_strong}', argsMap['missedPayments']),
+          ? L('{span_}1 member{_span} has {strong_} missed payments {_strong}', argsMap['missedPayments'])
+          : L('{span_}{membercount} members{_span} have {strong_} missed payments {_strong}', argsMap['missedPayments']),
         'noVotes': this.noVotes.length === 1
           ? L('{span_}1 member{_span} hasn\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
           : L('{span_}{membercount} members{_span} haven\'t {strong_} voted in the last {proposalcount} proposals {_strong}', argsMap['noVotes'])
--- a/frontend/model/contracts/shared/constants.js
+++ b/frontend/model/contracts/shared/constants.js
@@ -52,6 +52,8 @@ export const STATUS_EXPIRING = 'expiring' // Only useful to notify users that th
 export const STATUS_EXPIRED = 'expired'
 export const STATUS_CANCELLED = 'cancelled'
 
+// group-streaks related
+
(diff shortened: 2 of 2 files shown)


SebinSong:
@taoeffect The PR is ready besides the heisen-bug in `group-chat-direct-message.spec.js`
(The spec passes fine when running the test locally)

taoeffect:
@SebinSong It seems this isn't a heisenbug, but a bug that we need to fix. It might be as a result of recent merges to `master`. My guess is it might have to do with recent changes to `ChatMain`. Let me know if you have any ideas!

frontend/views/containers/dashboard/GroupMembersActivity.vue
```diff
@@ -164,8 +164,8 @@ export default ({
           ? L('{span_}1 member{_span} hasn\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails'])
           : L('{span_}{membercount} members{_span} haven\'t {strong_} entered income details{_strong}', argsMap['noIncomeDetails']),
         'missedPayments': this.missedPayments.length === 1
-          ? L('{span_}1 member{_span} has {strong_} missed payments {_strong}', argsMap['missedPayments'])
-          : L('{span_}{membercount} members{_span} have {strong_} missed payments {_strong}', argsMap['missedPayments']),
+          ? L('{span_}1 member{_span} has {strong_} missed their payments {_strong}', argsMap['missedPayments'])
+          : L('{span_}{membercount} members{_span} have {strong_} missed their payments {_strong}', argsMap['missedPayments']),
```
taoeffect: Please revert the string change 🙏

taoeffect (changes requested):
Nice work @SebinSong. The tests are failing because of a bug in ChatMain that was introduced in a recent PR, so none of these failing builds are related to this PR.

Please revert the changes to the strings, and this PR should be good to merge.

taoeffect (changes requested):
So I tested this locally with `GI_PERSIST=sqlite grunt dev` and this didn't seem to fix the problem when testing with our group's database.

One other note, if you run into issues running `grunt dev` like this:

```
Running "esbuild:watch" (esbuild) task
✘ [ERROR] Invalid define value (must be an entity name or valid JSON syntax): 'v1.2.1-13-ga04fafdf1
'

✘ [ERROR] Invalid define value (must be an entity name or valid JSON syntax): 'v1.2.1-13-ga04fafdf1
'
```

The solution is to modify the Gruntfile to replace:

```js
const GI_GIT_VERSION = process.env.CI ? process.env.GI_VERSION : execSync('git describe --dirty').toString('ascii')
```

With:

```js
const GI_GIT_VERSION = process.env.CI ? process.env.GI_VERSION : execSync('git describe --dirty').toString('utf8').trim()
```

SebinSong:
@taoeffect 
Just tested with the GIG DB shared to me and it works as expected:

taoeffect:
I don't get this at all with the database... I also didn't get the extra empty month:

Screenshot 2025-01-28 at 6 08 25 PM

So I thought maybe forcing it to show the extra month by running `sbp('gi.actions/group/forceDistributionDate', { contractID: sbp('state/vuex/state').currentGroupId })` would fix it, but no, that didn't cause the missed payments to show either.

Very odd. I've sent you my copy of the group state in case it helps (as a DM on GIG).

SebinSong:
@taoeffect Right. That's actually something I observed too. If you run the app with the DB and go to the payments page and then check the distribution period in the current group settings, you will be able to observe this discrepancy:

image

Do you remember we worked on adding a logic to automatically detect/resolve this discrepancy in `state.js` a couple of years ago? (below) 

https://github.com/okTurtles/group-income/blob/22a497f01aa286255af7b2f2aa5aa692831e6ca2/frontend/model/state.js#L391-L415

This Vuex watcher is currently wrapped within a `CONTRACT_REGISTERED` event listener (with some extra `if` blocks) but it appears that this listener is never called, meaning the app is currently not benefiting from this auto-handling logic.

---

Anyways, this is a separate issue to this PR (which is why I did not mention it in the previous comment) and regarding how I got that empty last column in the graph was just exactly doing what the comments in the above `state.js` code says:

```
// Before bringing this logic in, how the app updates the state related to group distribution period was, 
// 'currentPaymentPeriod': gets auto-updated(t1) in response to the change of 'reactiveDate.date' when it passes into the new period. 
```

`updateCurrentDistribution` function here is where `updateGroupStreaks` is executed, which updates all sorts of group streak fields.

In order to call this `updateCurrentDistribution` function, I just updated the income details of my account. (It also can be triggered by any member making a payment, or kicking out a member, as you might be well aware.)

---
Long story short,
1) The reason why you don't see the last column is due to the discrepancy between `getters.currentPaymentPriod` and the actual `groupSettings.distributionDate`.
2) The app does have the logic to automatically resolve this discrepancy in `state.js` but it's currently wrapped within `CONTRACT_REGISTERED` event listener but it **appears** that this is never called. (I might be wrong and I think we need Ricardo's help for correctly investigating it)
3) To get `groupSettings.distributionDate` updated, `updateCurrentDistribution()` function needs to be called and that's either by any group member updating their income details, or any group member leaving, or any pledging member recording a payment.
4) Once you take one of the actions in 3) and `groupSettings.distributionDate` is correctly set to current period, you will see the fix in this PR work.

Hope these things make sense. Thanks.

taoeffect:
Had a chat with @corrideat about this [on Slack](https://okturtles.slack.com/archives/C02KARWJR/p1738166590387689?thread_ts=1738147631.436249&cid=C02KARWJR):

greg Yesterday at 8:03 AM
Ricardo is CONTRACT_REGISTERED a missing propagated event?

Ricardo Yesterday at 11:51 PM
That's correct. It'd seem like CONTRACT_REGISTERED is a missing propagated event. Fixing that part would be pretty simple, it needs to be added to sw-primary.js (there's a list of events, ll. 106-113).
However, even if it's 'fixed' that way, it might still not be enough. I took a look at the event handler in question, and it seems like this would benefit from being in the SW (possibly as a periodic notification):
Events are not guaranteed to be forwarded (there may be no windows to forward events to)
Even when the event is forwarded, if there are many open windows each of them will be executing that logic and trying to submit an action.
The way it's currently written, it only works for the current group.

Ricardo Today at 5:22 AM
CONTRACT_REGISTERED is forwarded in PR #2568

taoeffect (approved):
I'm gonna assume this is fixed based on @SebinSong's comments and with the PR #2568 merged.

PR #2555, opened 2025-01-28
Close #2553

### LINKED ISSUE #2553: Followup to infinite loop in eventsAfter (#2544)
### Problem

Followup to #2544 (we didn't test it).

So I tested this by using our GIG group DB and a DM with myself.

- I made a copy of the DB, sent myself a few DMs and switched to `#general`, then shut down the server, replaced the newer DB with the older copy, and re-ran `GI_PERSIST=sqlite grunt dev`
- Then I visited my DM with myself again to load it. I was greeted with the alert* and clicked OK.
- The chat fixed itself (I think; I might've had to first switch to `#general` and back, please test this), however the red banner remained visible when it should be removed since the problem is fixed
- Furthermore, another problem appeared, which was the backend being hammered with these requests:

Image

*When I tried this a second time, I oddly enough wasn't greeted with the alert, but did see the constant `streamEntriesAfter` on the backend... odd.

### Solution

Test this scenario thoroughly locally and fix all remaining bugs, specifically:

- The red critical banner should be dismissed if the resync was a success
- The infinite loop should be fixed for good

EDIT: [it seems the infinite loop was fixed](https://github.com/okTurtles/group-income/issues/2553#issuecomment-2619635109); issue is now about:

- Log BOTH the `GI_VERSION` and the `GI_GIT_VERSION` = `git describe --dirty` in the service worker when it starts up, so that we can be 100% sure that we're running the latest version. 
- Immediately dismiss the danger banner after the user clicks OK to resync the contract.
- Log the URL for `eventsAfter` when an exception is thrown (in production, the parameters are not logged)

corrideat:
EDIT (by @taoeffect): Closes #2553

### DIFF
--- a/shared/domains/chelonia/utils.js
+++ b/shared/domains/chelonia/utils.js
@@ -566,9 +566,11 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
     throw new Error('Missing contract ID')
   }
 
+  let lastUrl
   const fetchEventsStreamReader = async () => {
     requestLimit = Math.min(limit ?? MAX_EVENTS_AFTER, remainingEvents)
-    const eventsResponse = await fetch(`${this.config.connectionURL}/eventsAfter/${contractID}/${sinceHeight}${Number.isInteger(requestLimit) ? `/${requestLimit}` : ''}`, { signal })
+    lastUrl = `${this.config.connectionURL}/eventsAfter/${contractID}/${sinceHeight}${Number.isInteger(requestLimit) ? `/${requestLimit}` : ''}`
+    const eventsResponse = await fetch(lastUrl, { signal })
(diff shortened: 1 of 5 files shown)


shared/domains/chelonia/utils.js
```diff
@@ -747,6 +749,7 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
           }
         }
       } catch (e) {
+        console.error('[eventsAfter] Error', { lastUrl }, e)
```
taoeffect, line 752: Why log `{ lastURL }` instead of `lastURL`?

corrideat: `lastURL` isn't defined

taoeffect (approved):
🎉🎉🎉

PR #2568, opened 2025-01-30
Improve GI_VERSION update logic

### LINKED ISSUE #2562: Neither browser nor service worker seems to use new GI_VERSION when it's released
### Problem

This is what's seen in the FF SW console after an update:

Image

Note that the page is not refreshed ... and so the old SW is still used, as well as the old frontend code.

### Solution

I noticed that after a manual page refresh it seems like both the SW and the browser frontend finally got the new code.

So I'm guessing that the SW needs to tell the browser to refresh ...

Not sure what to do if there are no open tabs tho. Maybe open a tab with the site?

corrideat:
Closes #2562.

### DIFF
--- a/frontend/controller/backend.js
+++ b/frontend/controller/backend.js
@@ -14,14 +14,40 @@ const languageFileMap = new Map([
 ])
 
 sbp('okTurtles.events/on', NOTIFICATION_TYPE.VERSION_INFO, (versionInfo) => {
+  if (versionInfo.GI_VERSION === process.env.GI_VERSION) {
+    // No refresh necessary, we're already at the latest version
+    sessionStorage.removeItem(NOTIFICATION_TYPE.VERSION_INFO)
+    return
+  }
+
   console.info('New Group Income version available:', versionInfo)
-  // Prevent the client from trying to reconnect when the page starts unloading.
-  sbp('okTurtles.data/get', PUBSUB_INSTANCE).destroy()
   // TODO: allow the user to manually reload the page later.
+  try {
+    // Store the current VERSION_INFO in session storage to prevent infinite
+    // reload loops
+    const existingSerialized = sessionStorage.getItem(NOTIFICATION_TYPE.VERSION_INFO)
+    if (existingSerialized) {
+      const existingVersionInfo = JSON.parse(existingSerialized)
+      if (
+        Array.isArray(existingVersionInfo) &&
+        !(Date.now() - existingVersionInfo[0] >= 1e7) &&
+        versionInfo.GI_VERSION === existingVersionInfo[1].GI_VERSION
+      ) {
+        console.warn('[NOTIFICATION_TYPE.VERSION_INFO] A different Group Income version is available, but reloading has failed to address it', { existingVersionInfo, versionInfo })
+        return
+      }
+    }
+
+    sessionStorage.setItem(NOTIFICATION_TYPE.VERSION_INFO, JSON.stringify([Date.now(), versionInfo]))
+  } catch (e) {
+    console.error('[NOTIFICATION_TYPE.VERSION_INFO] Error in handler', e)
+  }
+  // Prevent the client from trying to reconnect when the page starts unloading.
+  sbp('okTurtles.data/get', PUBSUB_INSTANCE)?.destroy()
   window.location.reload()
 })
 
-sbp('sbp/selectors/register', {
+export default (sbp('sbp/selectors/register', {
   /**
    * Fetches a JSON object containing translation strings for a given language.
    *
@@ -39,4 +65,4 @@ sbp('sbp/selectors/register', {
         .then(handleFetchResult('json'))
     }
   }
-})
+}): string[])
--- a/frontend/controller/backend.js
+++ b/frontend/controller/backend.js
@@ -3,6 +3,7 @@
 import type { JSONObject } from '~/shared/types.js'
 
 import sbp from '@sbp/sbp'
+import { HOURS_MILLIS } from '~/frontend/model/contracts/shared/time.js'
 import { NOTIFICATION_TYPE } from '~/shared/pubsub.js'
 import { handleFetchResult } from './utils/misc.js'
 import { PUBSUB_INSTANCE } from './instance-keys.js'
@@ -30,7 +31,7 @@ sbp('okTurtles.events/on', NOTIFICATION_TYPE.VERSION_INFO, (versionInfo) => {
       const existingVersionInfo = JSON.parse(existingSerialized)
       if (
         Array.isArray(existingVersionInfo) &&
-        !(Date.now() - existingVersionInfo[0] >= 1e7) &&
+        !(Date.now() - existingVersionInfo[0] >= 2.5 * HOURS_MILLIS) &&
         versionInfo.GI_VERSION === existingVersionInfo[1].GI_VERSION
       ) {
         console.warn('[NOTIFICATION_TYPE.VERSION_INFO] A different Group Income version is available, but reloading has failed to address it', { existingVersionInfo, versionInfo })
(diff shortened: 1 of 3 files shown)


frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -359,6 +366,20 @@ self.addEventListener('message', function (event) {
...
           port.close()
         })
+
+        // If the window is outdated (different GI_VERSION), trigger an event
+        // of type 'NOTIFICATION_TYPE.VERSION_INFO'.
+        if (
+          currentVersionInfo &&
+          event.source &&
+          event.data.GI_VERSION !== currentVersionInfo.GI_VERSION
+        ) {
```
taoeffect: This logic doesn't make sense to me for a few reasons:
 
- The `currentVersionInfo` received is based on the value set by the `VERSION_INFO` handler above, and yet that value is based on `process.env.GI_VERSION`, which should be available here in the SW too; there's no reason to send it over, we should be able to directly access it there.
- The web socket, which exists in the SW, is what sends over the new version info to the SW. It's not supposed to have anything to do with the browser.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -265,6 +267,11 @@ sbp('okTurtles.events/on', CHELONIA_RESET, () => {
   sbp('gi.periodicNotifications/init')
 })
 
+let currentVersionInfo
+sbp('okTurtles.events/on', NOTIFICATION_TYPE.VERSION_INFO, (versionInfo) => {
+  currentVersionInfo = versionInfo
+})
+
```
taoeffect, line 274: Where is this event being emitted from? The web socket?

How do we know this will happen before the `ready` event?


(If this really needs to be here and be done this way, then a comment is definitely needed explaining all of this. Otherwise should be modified + probably a comment anyway)

frontend/controller/backend.js
```diff
@@ -14,14 +14,40 @@ const languageFileMap = new Map([
...
   // TODO: allow the user to manually reload the page later.
+  try {
+    // Store the current VERSION_INFO in session storage to prevent infinite
+    // reload loops
+    const existingSerialized = sessionStorage.getItem(NOTIFICATION_TYPE.VERSION_INFO)
+    if (existingSerialized) {
+      const existingVersionInfo = JSON.parse(existingSerialized)
+      if (
+        Array.isArray(existingVersionInfo) &&
+        !(Date.now() - existingVersionInfo[0] >= 1e7) &&
```
taoeffect: What is this 1e7 value? 2.8 hours? Why not use `HOURS_MILLIS` instead?

frontend/controller/backend.js
```diff
@@ -14,14 +14,40 @@ const languageFileMap = new Map([
...
+        versionInfo.GI_VERSION === existingVersionInfo[1].GI_VERSION
+      ) {
+        console.warn('[NOTIFICATION_TYPE.VERSION_INFO] A different Group Income version is available, but reloading has failed to address it', { existingVersionInfo, versionInfo })
```
taoeffect, line 37: Should probably be an `error`

taoeffect (changes requested):
Review ready!

frontend/controller/serviceworkers/sw-primary.js
corrideat: > The currentVersionInfo received is based on the value set by the VERSION_INFO handler above, and yet that value is based on process.env.GI_VERSION, which should be available here in the SW too; there's no reason to send it over, we should be able to directly access it there.

The entire point of this logic is to detect version discrepancies. `process.env.GI_VERSION` is indeed available in the SW, but by itself it'd not detect an outdated version. The rest of the comment doesn't itself make much sense to me. `currentVersionInfo` is _not_ based on `process.env.GI_VERSION`.

> The web socket, which exists in the SW, is what sends over the new version info to the SW. It's not supposed to have anything to do with the browser.

It has something to do with the browser in that the browser triggers a refresh.

frontend/controller/serviceworkers/sw-primary.js
corrideat: The web socket cannot emit events, but, yes, the WS receives a message that results in an event being emitted if there's a version discrepancy. This event is emitted soon after connection.

> How do we know this will happen before the ready event?

We don't.

frontend/controller/backend.js
corrideat: Yes, that's approximately 2.8 hours. It sounded like a reasonable value.

frontend/controller/backend.js
corrideat: It's not necessarily an error condition.

frontend/controller/serviceworkers/sw-primary.js
taoeffect: > We don't.

Then what is the point?

We need to immediately do the version check when the pubsub emits the event and then tell the browser to do the page refresh.

If this only happens on the `ready` event it defeats the purpose of this PR. Unless I'm missing something.

corrideat: > We need to immediately do the version check when the pubsub emits the event and then tell the browser to do the page refresh.

No, we don't. We need to send the event when it's available, which is what this PR does. This doesn't only happen on the `ready` event.

With this PR it's sent:

1. When a client connects, if appropriate. That's the purpose of this handler.
2. On already connected clients, when the event is received.

taoeffect: > On already connected clients, when the event is received.

Can you point to where this happens?

If I'm using GI right now (client open to sw), and an update happens while I'm using it, and the pubsub sends the event to the sw, where is the logic that triggers the page refresh immediately?

corrideat: > If I'm using GI right now (client open to sw), and an update happens while I'm using it, and the pubsub sends the event to the sw, where is the logic that triggers the page refresh immediately?

There isn't. That's why I said, 'with this PR'. Expanding on this: `frontend/controller/backend.js` handles this, but the event isn't currently forwarded to from the SW to the browser. This PR fixes that.

> Can you point to where this happens?

As mentioned, the event is emitted on (re)connection to the WS. This event is forwarded from the SW to the browser. In addition, the event is sent to new clients when they connect to the SW.

taoeffect: > There isn't. That's why I said, 'with this PR'.

And I'm asking, where in this PR, does the websocket trigger the page refresh immediately in connected clients?

corrideat: In `frontend/controller/backend.js`

frontend/controller/serviceworkers/sw-primary.js
taoeffect: > The entire point of this logic is to detect version discrepancies. process.env.GI_VERSION is indeed available in the SW, but by itself it'd not detect an outdated version

The problem is that if I load the page in a browser, my `process.env.GI_VERSION` may well be the latest one, right? But the SW might be running on the old one. Its `process.env.GI_VERSION` would be out of date.

So this check:

```js
event.data.GI_VERSION !== currentVersionInfo.GI_VERSION
```

Would not get triggered, because they'd be the same value, but the check should be:

```js
currentVersionInfo.GI_VERSION !== process.env.GI_VERSION
```

taoeffect: Ok... so this code is really useful only if the page that's loaded is loaded from cache, and therefore has an outdated `process.env.GI_VERSION` that it sends.

OK, could you add that as a comment to explain that?

frontend/controller/backend.js
taoeffect: Could you use `HOURS_MILLIS` for readability?

taoeffect (approved):
Great work @corrideat !

PR #2569, opened 2025-01-30
Handle identity contract fork on login.

### LINKED ISSUE #2564: Failure to login/recover in Brave browser tab that had forked
### Problem

After the update to 1.2.2, I clicked the OK button but it did not recover.

I don't remember exactly what happened, but somehow I ended up logged out and with the red error banner.

Image

Attempts to log back in all fail, even after doing doing <kbd>command+shift+r</kbd>:

Image

SW logs meantime:

Image

### Solution

Find the bug and fix.

corrideat:
Closes #2564.

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -110,10 +110,15 @@ async function startApp () {
     if (error?.name === 'ChelErrorForkedChain') {
       const rootState = sbp('state/vuex/state')
       if (!rootState.contracts[contractID]) {
+        // If `rootState.contracts[contractID]` doesn't exist, it means we're no
+        // longer subscribed to the contract. This could happen, e.g., if the
+        // contract has since been released. In any case, the absence of
+        // `rootState.contracts[contractID]` means that there's nothing to
+        // left to recover.
         console.error('Forked chain detected. However, there is no contract entry.', { contractID }, error)
         return
       }
-      const type = rootState.contracts[contractID] || '(unknown)'
+      const type = rootState.contracts[contractID].type || '(unknown)'
       console.error('Forked chain detected', { contractID, type }, error)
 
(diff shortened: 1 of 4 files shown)


frontend/main.js
```diff
@@ -109,17 +109,26 @@ async function startApp () {
     sbp('gi.ui/seriousErrorBanner', error)
     if (error?.name === 'ChelErrorForkedChain') {
       const rootState = sbp('state/vuex/state')
-      const type = rootState.contracts[contractID].type || '(unknown)'
+      if (!rootState.contracts[contractID]) {
+        console.error('Forked chain detected. However, there is no contract entry.', { contractID }, error)
+        return
+      }
+      const type = rootState.contracts[contractID] || '(unknown)'
```
taoeffect: Think this line should remain as it was..? `.type` missing...

frontend/main.js
```diff
@@ -109,17 +109,26 @@ async function startApp () {
     sbp('gi.ui/seriousErrorBanner', error)
     if (error?.name === 'ChelErrorForkedChain') {
       const rootState = sbp('state/vuex/state')
-      const type = rootState.contracts[contractID].type || '(unknown)'
+      if (!rootState.contracts[contractID]) {
+        console.error('Forked chain detected. However, there is no contract entry.', { contractID }, error)
+        return
```
taoeffect: Can you add a comment explaining here why we don't proceed with recovery? Is it because in this situation we're no longer subscribed...? But then how did we get the error if we weren't subscribed...?

taoeffect (changes requested):
Review ready!

taoeffect (approved):
Great work @corrideat !

Issue #2572, opened 2025-01-31
"French classes" heisenbug


taoeffect:
### Problem

https://cloud.cypress.io/projects/q6whky/runs/3854/test-results

### Solution

Fix the test.

PR #2576, opened 2025-01-31
#2575 - Make login modal undismissible while logging-in

### LINKED ISSUE #2575: Login Modal should not be dismissible while log-in is still being processed
### Problem

This is an issue noticed while running `GI_PERSIST=sqlite grunt dev` command(where log-in takes at least several seconds) but, When you click on the log-in button and then dismisses the modal by clicking on the modal backdrop while the async operations are still not done, it is dismissed.
But as log-in process is still ongoing in the background, a few seconds later the app all of a sudden navigates to the dashboard.

https://github.com/user-attachments/assets/46fc04d8-ad1b-46c0-a3dc-a3c6bc84afe7

This behavior could be misleading and make the user think the login attempt is completely discarded and they have to re-open the modal/log in again. (This is actually what I did when it first happened to me)

### Solution

Make sure the modal cannot be dismissed until the login process is completed either with success/error.

SebinSong:
closes #2575 

https://github.com/user-attachments/assets/24c36cd5-80f0-496d-8749-29c6b0d95613

`ModalTemplate.vue` already had a prop `loading` that allows us to achieve exactly what we want, so used it for the fix.

### DIFF
--- a/frontend/views/containers/access/LoginForm.vue
+++ b/frontend/views/containers/access/LoginForm.vue
@@ -86,17 +86,19 @@ export default ({
 
         const username = this.form.username
 
+        this.$emit('login-status', 'submitting')
         await sbp('gi.app/identity/login', {
           username,
           password: new Secret(this.form.password)
         })
         await this.postSubmit()
-        this.$emit('submit-succeeded')
+        this.$emit('login-status', 'success')
 
         requestNotificationPermission().catch(e => console.error('[SignupForm.vue] Error requesting notification permission', e))
       } catch (e) {
         console.error('FormLogin.vue login() error:', e)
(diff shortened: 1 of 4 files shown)


SebinSong:
The same thing actually happens in Sign-up modal too but I noticed sign-up process takes much shorter than log-in. Please let me know if we want this fix to sign-up modal too, though.

taoeffect:
> The same thing actually happens in Sign-up modal too but I noticed sign-up process takes much shorter than log-in. Please let me know if we want this fix to sign-up modal too, though.

Yeah it would make sense to fix this on both.

BTW that build error was #2572 again.

taoeffect (approved):
Fantastic work @SebinSong! 👏

PR #2578, opened 2025-02-01
Remove loding animation on prompt

### LINKED ISSUE #2574: Fix fork login logic
### Problem

The changes in #2569 were not a complete fix.

This code is not helping:

```js
      if (!rootState.contracts[contractID]) {
        // If `rootState.contracts[contractID]` doesn't exist, it means we're no
        // longer subscribed to the contract. This could happen, e.g., if the
        // contract has since been released. In any case, the absence of
        // `rootState.contracts[contractID]` means that there's nothing to
        // left to recover.
        console.error('Forked chain detected. However, there is no contract entry.', { contractID }, error)
        return
      }
```

I logged in on staging, and then afterward there was a copy of the DB from the live server. This caused a fork. However, now I'm stuck here:

Image

This is where i remain even when refreshing the page.

### Solution

Clearly this `if (!rootState.contracts[contractID])` is breaking things and preventing us from proceeding forward. Needs fixing.

corrideat:
Closes #2574

### DIFF
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -23,7 +23,7 @@ import router from './controller/router.js'
 import './controller/service-worker.js'
 import { SETTING_CURRENT_USER } from './model/database.js'
 import store from './model/state.js'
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
I don't know if this will fix it but let's find out!

PR #2579, opened 2025-02-01
Send transferables for event messages.

### LINKED ISSUE #2573: DataCloneError: Failed to execute 'postMessage' on 'Client': A MessagePort could not be cloned because it was not transferred.
### Problem

In v1.2.3, while logged out and first page load:

Image

Image

And I get more of these when I log in.

### Solution

Fix.

corrideat:
Closes #2573.

### DIFF
--- a/frontend/controller/serviceworkers/sw-primary.js
+++ b/frontend/controller/serviceworkers/sw-primary.js
@@ -104,7 +104,7 @@ sbp('okTurtles.events/on', CHELONIA_RESET, setupRootState)
 
 // These are all of the events that will be forwarded to all open tabs and windows
 ;[
-  CHELONIA_RESET, CONTRACTS_MODIFIED, CONTRACT_IS_SYNCING, CONTRACT_REGISTERED,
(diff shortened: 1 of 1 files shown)


frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -104,7 +104,7 @@ sbp('okTurtles.events/on', CHELONIA_RESET, setupRootState)
 
 // These are all of the events that will be forwarded to all open tabs and windows
 ;[
-  CHELONIA_RESET, CONTRACTS_MODIFIED, CONTRACT_IS_SYNCING, CONTRACT_REGISTERED,
+  CHELONIA_RESET, CONTRACTS_MODIFIED, CONTRACT_IS_SYNCING,
```
taoeffect, line 107: Why is this removing CONTRACT_REGISTERED? Didn't we just add that at Sebin's request?

taoeffect (approved):
🤞🤞 this does the trick!

PR #2584, opened 2025-02-01
PEK rotation improvements.

### LINKED ISSUE #2570: Failure to show username for contract upon join
### Problem

On the testing server, there's an account (`pwa2`) with a group and another user that joined (username is `greg`) that isn't displaying:

Image
Image

Two problems:

1. Username isn't shown
2. Clicking user's profile icon in the user list doesn't show the ProfileCard **and** doesn't show any errors in the console

### Solution

This is a repeatable problem. Using Troubleshooting to reset the data doesn't fix it.

1. Find out why username isn't shown and fix.
2. Find out why the ProfileCard isn't shown and why no error is shown in console either when clicked.

DM'd you credentials on GIG.

EDIT: according to @corrideat the problem is that the user had left some other group, triggering a rotation on their PEK, which means that the new group can't decrypt their username from their identity contract's attributes. So as a solution we can re-encrypt the data upon key rotation.

corrideat:
Closes #2570.

Testing instructions:

1. Create three users (`alice`, `bob`, `carol`)
2. `alice` creates a group `G1`
3. `bob` creates a group `G2`
4. `carol` joins `G1`
5. `carol` leaves `G1`. There are no warning or errors for `/rotateKeys`. The `pek` for `carol` (in `_vm.authorizedKeys`) has been rotated.
6. `carol` joins `G2`
7. `bob` observes that the username for `carol` appears in `G2`.


backend/push.js
```diff
@@ -239,6 +239,7 @@ export const postEvent = async (subscription: Object, event: ?string): Promise<v
   })
 
   if (!req.ok) {
+    console.debug('Error sending push notification', subscription.id, req.status)
```
taoeffect: Should this be console.debug or something higher like log or info?

corrideat: Maybe. I thought `debug` is appropriate since it's by definition a user-defined endpoint. Nothing stops a user from reporting an endpoint like `https://www.bankofamerica.com/`, and since it's a third-party endpoint, we also can't control the SLA.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -947,6 +947,8 @@ export default ({
         }
         sbp('gi.actions/identity/kv/setChatRoomReadUntil', {
           contractID: chatRoomID, messageHash, createdHeight
+        }).catch(e => {
```
corrideat, line 950: Bugfix

shared/domains/chelonia/chelonia.js
```diff
@@ -537,6 +538,20 @@ export default (sbp('sbp/selectors/register', {
     const keyId = findSuitableSecretKeyId(contractIDOrState, permissions, purposes, ringLevel, allowedActions)
     return keyId
   },
+  'chelonia/contract/setPendingKeyRevocation': function (contractID: string, names: string[]) {
```
corrideat, line 541: Improvement as per contract TODOs.

shared/domains/chelonia/chelonia.js
```diff
@@ -1693,9 +1708,17 @@ async function outEncryptedOrUnencryptedAction (
   if (opType === GIMessage.OP_ACTION_ENCRYPTED && !params.encryptionKeyId) {
     throw new Error('OP_ACTION_ENCRYPTED requires an encryption key ID be given')
   }
+  if (params.encryptionKey) {
```
corrideat, line 1713: This new parameter is needed to designate a raw key to use for encryption in `OP_ATOMIC`. The current implementation of `GIMessage` / `encryptedData` / `signedData` / `chelonia` isn't advanced enough to automatically switch to the new key on outgoing messages. The workaround is designating the key as a raw key.

shared/domains/chelonia/encryptedData.js
```diff
@@ -108,7 +108,7 @@ const decryptData = function (height: number, data: any, additionalKeys: Object,
   const key = additionalKeys[eKeyId]
 
   if (!key) {
-    throw new ChelErrorDecryptionKeyNotFound(`Key ${eKeyId} not found`)
+    throw new ChelErrorDecryptionKeyNotFound(`Key ${eKeyId} not found`, { cause: eKeyId })
```
corrideat, line 112: Adding the `cause` here helps identify the key that failed.

test/cypress/support/commands.js
```diff
@@ -113,7 +113,7 @@ Cypress.Commands.add('getByDT', (element, otherSelector = '') => {
 cySbpCheckCommand('giNoPendingGroupKeyShares', (sbp) => {
   const state = sbp('state/vuex/state')
   const pending = Object.keys(state.contracts)
-    .filter(contractID => state[contractID]._vm.type === 'gi.contracts/group')
+    .filter(contractID => state[contractID]._vm?.type === 'gi.contracts/group')
```
corrideat: For some reason, sometimes `_vm` was undefined in the tests, causing this to fail.

shared/domains/chelonia/internals.js
```diff
@@ -2091,6 +2116,15 @@ const handleEvent = {
     }
     this.config.reactiveSet(state.contracts[contractID], 'HEAD', hash)
     this.config.reactiveSet(state.contracts[contractID], 'height', height)
+    if (missingDecryptionKeyId) {
```
corrideat: This adds a log of missed decryption keys, which can be used to more accurately determine when a contract needs to be re-synced.

shared/domains/chelonia/internals.js
```diff
@@ -2046,17 +2066,22 @@ const handleEvent = {
       return acc
     }
 
-    const actionsOpV = ((msg: any): GIOpAtomic).reduce(reducer, [])
+    const actionsOpV = (Number.isFinite(errorIndex)
+      // $FlowFixMe[incompatible-call]
```
corrideat: The side-effects sometimes get called if there was an error (i.e., `ChelErrorWarning` has this behaviour). This only executes side-effects up to the point where process succeeded in `OP_ATOMIC`.

shared/domains/chelonia/internals.js
```diff
@@ -682,12 +685,18 @@ export default (sbp('sbp/selectors/register', {
...
+        for (let i = 0; i < v.length; i++) {
+          const u = v[i]
+          try {
+            if (u[0] === GIMessage.OP_ATOMIC) throw new Error('Cannot nest OP_ATOMIC')
+            if (!validateKeyPermissions(config, state, signingKeyId, u[0], u[1], direction)) {
+              throw new Error('Inside OP_ATOMIC: no matching signing key was defined')
+            }
+            await opFns[u[0]](u[1])
+          } catch (e) {
+            e[errorIndexSymbol] = i
```
corrideat: Point where the error occurred.

frontend/model/getters.js
```diff
@@ -461,23 +461,23 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }) =>
     Object.keys(state.contracts)
       .filter(contractID => state.contracts[contractID].type === 'gi.contracts/identity')
       .forEach(contractID => {
-        if (!state[contractID]) {
-          console.warn('[ourContactProfilesById] Missing state', contractID)
-          return
-        }
-        const attributes = state[contractID].attributes
+        const attributes = state[contractID]?.attributes
```
corrideat, line 464: This fixes the ProfileCard not showing if the username isn't known.

frontend/main.js
```diff
@@ -49,6 +49,16 @@ console.info('CONTRACTS_VERSION:', process.env.CONTRACTS_VERSION)
 console.info('LIGHTWEIGHT_CLIENT:', process.env.LIGHTWEIGHT_CLIENT)
 console.info('NODE_ENV:', process.env.NODE_ENV)
 
+if (process.env.CI) {
```
corrideat, line 52: Add logging information on CI to get a better stack trace and identify missing error handling.

frontend/controller/actions/index.js
```diff
@@ -191,8 +187,9 @@ sbp('sbp/selectors/register', {
         contractID,
         contractName,
         data: [
-          ...keyShares.map((data) => ['chelonia/out/keyShare', { data: encryptedOutgoingData(contractID, CEKid, data) }]),
-          ['chelonia/out/keyUpdate', { data: updatedKeys }]
+          ...(keyShares[0] ?? []),
```
corrideat: Function updated to support arbitrary operations. This also makes this function generic and no longer GI-specific.

frontend/controller/actions/identity.js
```diff
@@ -653,7 +653,20 @@ export default (sbp('sbp/selectors/register', {
     // share along with OP_KEY_UPDATE. In this case, we're sharing all keys
     // to their respective contracts and there are no keys to include in
     // the same event as OP_KEY_UPDATE. Therefore, we return undefined
-    return undefined
+    if (!newKeys.pek) return undefined
+    return [
+      undefined, // Nothing before OP_KEY_UPDATE
+      [
+        // Re-encrypt attributes with the new PEK
```
corrideat, line 660: This is 'the fix' for this issue, i.e., re-encrypting attributes.

shared/domains/chelonia/internals.js
taoeffect: OK, so based on the call here's how we've agreed conceptually this code should work:

There are two types of "errors" that we need to consider:

1. "Ignoring" errors
2. "Failure" errors

### Example: `OP_KEY_ADD`

1. IGNORING: an error is thrown because we wanted to add a key but the key we wanted to add is already there. This is not a hard error, it's an *ignoring error*. We don't care that the operation failed in this case because the intent was accomplished.
2. FAILURE: an error is thrown because we wanted to add a key that doesn't exist.

### Example: `OP_ACTION_ENCRYPTED`

1. IGNORING: An error is thrown because we don't have the key to decrypt the action. We ignore it.
2. FAILURE: An error is thrown by the process function during processing.

### Handling these in `OP_ATOMIC`

- ALL errors of class "IGNORING" should be ignored. They should not impact our ability to process the rest of the operations in the `OP_ATOMIC`. No matter how many of these are thrown, it doesn't affect the rest of the operations.
- ANY error of class "FAILURE" will call the rest of the operations to fail and the state to be reverted to prior to the `OP_ATOMIC`. No side-effects should be run. Because an intention failed.

Feel free to copy this 👆 as a comment in the code.

taoeffect (changes requested):
Preliminary review based on call ready!

frontend/controller/actions/index.js
```diff
@@ -191,8 +187,9 @@ sbp('sbp/selectors/register', {
...
+          ['chelonia/out/keyUpdate', { data: updatedKeys }],
+          ...(keyShares[1] ?? [])
```
taoeffect: I'm a little confused as to the nature of the "arbitrary operations" you say are supported here, as this seems to support at most 2 operations (`keyShares[0]` and `keyShares[1]`). What if there are 3?

EDIT: if this API is designed to support a maximum of 2 entries, then can you add comments explaining that above line 96 to explain both the signature of `shareNewKeysSelector` and its return value?

frontend/controller/actions/chatroom.js
```diff
@@ -227,20 +228,25 @@ export default (sbp('sbp/selectors/register', {
...
-          id: newId,
-          meta: {
-            private: {
-              content: encryptedOutgoingData(pContractID, CEKid, serializeKey(newKey, true))
-            }
-          }
-        }))
-      }
-    }))
+      return [
```
taoeffect, line 231: Since above this returns `undefined` on line 229... should the resulting `.map` result `.filter` out these `undefined`s?

EDIT: also, there's an inconsistency here with `'gi.actions/group/shareNewKeys'`, which instead of returning `undefined` returns `Promise.resolve()`...

frontend/main.js
```diff
@@ -49,6 +49,16 @@ console.info('CONTRACTS_VERSION:', process.env.CONTRACTS_VERSION)
...
+  const originalFetch = self.fetch
+  self.fetch = (...args) => {
+    return originalFetch.apply(self, args).catch(e => {
+      console.error('FETCH FAILED', args, new Error().stack, e)
+      throw e
+    })
```
taoeffect, line 58: For both of these I don't think you need to do `new Error().stack`, you can just use `e.stack`.

frontend/model/contracts/group.js
```diff
@@ -936,22 +930,14 @@ sbp('chelonia/defineContract', {
...
               }
             } else {
-              // avoid blocking the main thread
-              // eslint-disable-next-line require-await
-              (async () => {
-                alert(L("Couldn't join the #{chatroomName} in the group. Doesn't exist.", { chatroomName: CHATROOM_GENERAL_NAME }))
-              })()
+              console.error('Couldn\'t join the chatroom in the group. Doesn\'t exist.', { chatroomName: CHATROOM_GENERAL_NAME })
+
+              // TODO: event
```
taoeffect: Two things:

1. Minor readability nit: could you use `"` instead of `'` for the outer quotes so that you don't have to write `\'`?
2. What are these `// TODO: event` comments? Could you please either do the TODO or edit it so that it elaborates in detail what needs to be done and why? (I certainly don't know what this means, and likely future-you will forget too)

frontend/model/contracts/group.js
```diff
@@ -194,17 +194,11 @@ function memberLeaves ({ memberID, dateLeft, heightLeft, ourselvesLeaving }, { c
...
-  if (!state._volatile) state['_volatile'] = Object.create(null)
-  if (!state._volatile.pendingKeyRevocations) state._volatile['pendingKeyRevocations'] = Object.create(null)
-
-  const CSKid = findKeyIdByName(state, 'csk')
-  const CEKid = findKeyIdByName(state, 'cek')
-
-  state._volatile.pendingKeyRevocations[CSKid] = true
-  state._volatile.pendingKeyRevocations[CEKid] = true
+  sbp('chelonia/queueInvocation', contractID, () => {
+    return sbp('chelonia/contract/setPendingKeyRevocation', contractID, ['cek', 'csk'])
```
taoeffect: RFC: The other versions of this are accompanied by a call to `'gi.actions/out/rotateKeys'`.. where does that happen for this, and if it doesn't happen, why not?

frontend/model/contracts/identity.js
```diff
@@ -142,6 +147,11 @@ sbp('chelonia/defineContract', {
         }
       },
       process ({ data }, { state }) {
+        if (!state.attributes) {
+          // If it's not our own identity contract, attributes may not be
+          // defined
+          return
```
taoeffect, line 153: Why is there a call to `return`...? Other users should be able to read the latest attributes, and for that to happen we need to be able to proceed...

And if you're sure for some reason that's not true, a comment is needed to explain why

frontend/model/contracts/identity.js
```diff
@@ -180,6 +190,10 @@ sbp('chelonia/defineContract', {
       process ({ data }, { state }) {
         // NOTE: this method is always created by another
         const { contractID } = data
+        if (!state.chatRooms) {
+          // When creating a DM, we may not have the `.chatRooms` property
+          state.chatRooms = Object.create(null)
+        }
```
taoeffect: Why would we not have the `.chatRooms` property? It's created by the constructor

shared/domains/chelonia/chelonia.js
```diff
@@ -537,6 +538,20 @@ export default (sbp('sbp/selectors/register', {
...
+
+    if (!state._volatile) state._volatile = Object.create(null)
+    if (!state._volatile.pendingKeyRevocations) state._volatile.pendingKeyRevocations = Object.create(null)
+
+    for (const name of names) {
+      const keyId = findKeyIdByName(state, name)
+      if (keyId) {
+        state._volatile.pendingKeyRevocations[keyId] = true
+      }
+    }
```
taoeffect: Does this updated state need to be immediately saved to disk?

shared/domains/chelonia/chelonia.js
```diff
@@ -537,6 +538,20 @@ export default (sbp('sbp/selectors/register', {
...
+    const state = rootState[contractID]
+
+    if (!state._volatile) state._volatile = Object.create(null)
+    if (!state._volatile.pendingKeyRevocations) state._volatile.pendingKeyRevocations = Object.create(null)
+
+    for (const name of names) {
+      const keyId = findKeyIdByName(state, name)
+      if (keyId) {
+        state._volatile.pendingKeyRevocations[keyId] = true
+      }
```
taoeffect: If the `keyId` can't be found, shouldn't at least a warning be logged?

shared/domains/chelonia/internals.js
```diff
@@ -681,13 +685,52 @@ export default (sbp('sbp/selectors/register', {
     }
     if (!state._vm) config.reactiveSet(state, '_vm', Object.create(null))
     const opFns: { [GIOpType]: (any) => void | Promise<void> } = {
+      /*
+        There are two types of "errors" that we need to consider:
+        1. "Ignoring" errors
+        2. "Failure" errors
+        Example: OP_KEY_ADD
+        1. IGNORING: an error is thrown because we wanted to add a key but the key we wanted to add is already there. This is not a hard error, it's an ignoring error. We don't care that the operation failed in this case because the intent was accomplished.
+        2. FAILURE: an error is thrown because we wanted to add a key that doesn't exist.
```
taoeffect: Text here doesn't make sense, should say something like:

```
2. FAILURE: an error is thrown while attempting to add a key that doesn't exist.
```

shared/domains/chelonia/internals.js
```diff
@@ -681,13 +685,52 @@ export default (sbp('sbp/selectors/register', {
...
+              throw new Error('Inside OP_ATOMIC: no matching signing key was defined')
+            }
+            await opFns[u[0]](u[1])
+          } catch (e) {
+            if (e) {
+              if (e.name === 'ChelErrorDecryptionKeyNotFound') {
+                console.warn(`[chelonia] [OP_ATOMIC] WARN '${e.name}' in processMutation for ${message.description()}: ${e.message}`, e, message.serialize())
+                const missingDecryptionKeyIds = missingDecryptionKeyIdsMap.get(message)
+                if (missingDecryptionKeyIds) {
+                  missingDecryptionKeyIds.add(e.cause)
```
taoeffect: Are we guaranteed `e.cause` will be defined here?

shared/domains/chelonia/internals.js
```diff
@@ -681,13 +685,52 @@ export default (sbp('sbp/selectors/register', {
...
+        for (let i = 0; i < v.length; i++) {
+          const u = v[i]
+          try {
+            if (u[0] === GIMessage.OP_ATOMIC) throw new Error('Cannot nest OP_ATOMIC')
+            if (!validateKeyPermissions(config, state, signingKeyId, u[0], u[1], direction)) {
+              throw new Error('Inside OP_ATOMIC: no matching signing key was defined')
+            }
+            await opFns[u[0]](u[1])
+          } catch (e) {
+            if (e) {
```
taoeffect: Why is there an `if (e)` thing here...? In what situation can this not be defined?

shared/domains/chelonia/internals.js
```diff
@@ -681,13 +685,52 @@ export default (sbp('sbp/selectors/register', {
...
+                console.warn(`[chelonia] [OP_ATOMIC] WARN '${e.name}' in processMutation for ${message.description()}: ${e.message}`, e, message.serialize())
+                const missingDecryptionKeyIds = missingDecryptionKeyIdsMap.get(message)
+                if (missingDecryptionKeyIds) {
+                  missingDecryptionKeyIds.add(e.cause)
+                } else {
+                  missingDecryptionKeyIdsMap.set(message, new Set([e.cause]))
+                }
+                continue
+              } else {
+                console.error(`[chelonia] [OP_ATOMIC] ERROR '${e.name}' in processMutation for ${message.description()}: ${e.message || e}`, e, message.serialize())
```
taoeffect: This error message says it was a `processMutation` ... but AFAICT that isn't true, it could be some other operation other than `OP_ACTION_(UN)ENCRYPTED`

shared/domains/chelonia/internals.js
```diff
@@ -801,7 +845,9 @@ export default (sbp('sbp/selectors/register', {
                     key: deserializeKey(decrypted),
                     transient
                   }]))
-                  if (
+                  if (missingDecryptionKeyIds?.includes(key.id)) {
```
taoeffect, line 855: Can you add a comment here explaining what this is doing?

shared/domains/chelonia/internals.js
```diff
@@ -2091,6 +2151,19 @@ const handleEvent = {
...
+      if (!missingDecryptionKeyIds) {
+        missingDecryptionKeyIds = []
+        this.config.reactiveSet(state.contracts[contractID], 'missingDecryptionKeyIds', missingDecryptionKeyIds)
+      }
+      missingDecryptionKeyIdsForMessage.forEach(keyId => {
+        if (missingDecryptionKeyIds.includes(keyId)) return
+        missingDecryptionKeyIds.push(keyId)
+      })
+      this.config.reactiveSet(state.contracts[contractID], 'height', height)
+    }
```
taoeffect: Could you please add a comment above this block here explaining:

1. What this is doing and why
2. Why it's being done here

shared/domains/chelonia/internals.js
```diff
@@ -2091,6 +2151,19 @@ const handleEvent = {
     }
     this.config.reactiveSet(state.contracts[contractID], 'HEAD', hash)
     this.config.reactiveSet(state.contracts[contractID], 'height', height)
+    const missingDecryptionKeyIdsForMessage = missingDecryptionKeyIdsMap.get(message)
+    if (Array.isArray(missingDecryptionKeyIdsForMessage)) {
```
taoeffect:
```
> Array.isArray(new Set())
false
```

shared/domains/chelonia/internals.js
```diff
@@ -2091,6 +2151,19 @@ const handleEvent = {
...
+      let missingDecryptionKeyIds = state.contracts[contractID].missingDecryptionKeyIds
+      if (!missingDecryptionKeyIds) {
+        missingDecryptionKeyIds = []
+        this.config.reactiveSet(state.contracts[contractID], 'missingDecryptionKeyIds', missingDecryptionKeyIds)
+      }
+      missingDecryptionKeyIdsForMessage.forEach(keyId => {
+        if (missingDecryptionKeyIds.includes(keyId)) return
+        missingDecryptionKeyIds.push(keyId)
+      })
+      this.config.reactiveSet(state.contracts[contractID], 'height', height)
```
taoeffect: Why is this being called here again? It's already done on line 2153

taoeffect (changes requested):
My review is ready!

## Claude's review of `internals.js`

After reviewing the diff and the full file, here are a few potential issues and observations:

1. In the new code for handling errors in `OP_ATOMIC`, there's a potential issue with the missingDecryptionKeyIds handling. The code checks if `missingDecryptionKeyIds` is a Set, but later in the file, it's treated as an array. This inconsistency could lead to runtime errors.

```javascript
// In OP_ATOMIC:
const missingDecryptionKeyIds = missingDecryptionKeyIdsMap.get(message)
if (missingDecryptionKeyIds) {
  missingDecryptionKeyIds.add(e.cause)
} else {
  missingDecryptionKeyIdsMap.set(message, new Set([e.cause]))
}

// Later in applyProcessResult:
if (Array.isArray(missingDecryptionKeyIdsForMessage)) { // This expects an array, not a Set
  // ...
}
```

2. The error handling in `OP_ATOMIC` might have an issue with the nested conditions:

```javascript
if (e) {
  if (e.name === 'ChelErrorDecryptionKeyNotFound') {
    // ...
  } else {
    // ...
  }
  // ...
  if (e.name === 'ChelErrorWarning') continue
}
throw e
```

The `throw e` is outside the first `if (e)` block, which means it will always throw the error even if `e` is falsy. It should probably be inside the first `if (e)` block.

3. There's a potential memory leak with the WeakMap usage. While WeakMaps are good for garbage collection, the code should ensure that all message objects are properly cleaned up when they're no longer needed. The code does include cleanup in the `finally` block, but it might be worth adding additional cleanup in error cases.

4. In the new code that handles missing decryption key IDs, there's a duplicate height update:

```javascript
if (Array.isArray(missingDecryptionKeyIdsForMessage)) {
  // ...
  missingDecryptionKeyIdsForMessage.forEach(keyId => {
    if (missingDecryptionKeyIds.includes(keyId)) return
    missingDecryptionKeyIds.push(keyId)
  })
  this.config.reactiveSet(state.contracts[contractID], 'height', height) // This is redundant
}
```

This height update is redundant because it's already being set earlier in the function:

```javascript
this.config.reactiveSet(state.contracts[contractID], 'height', height)
```

These issues should be addressed to improve the robustness of the code.

## Gemini 2.0 Flash Thinking's Review of `internals.js`

Screenshot 2025-02-05 at 2 10 57 PM

shared/domains/chelonia/internals.js
corrideat: Probably bad refactoring.

shared/domains/chelonia/internals.js
corrideat: Right

shared/domains/chelonia/internals.js
corrideat: `throw undefined`, `throw null`, etc.

frontend/controller/actions/index.js
corrideat: The `keyShares` name is perhaps confusing, and I'll update it accordingly, same with your suggestion for explaining the signature. That said, it really does support arbitrary operations, in any quantity (see the spread operator). `keyShares[0]` is what comes before `keyUpdate` and `keyShares[1]` is what comes after. A hypothetical `keyShares[2]` doesn't make sense.

frontend/controller/actions/chatroom.js
corrideat: That's got to do with `flow`. I checked now and it seems to be fine to just return, but sometimes it complains if `.map` doesn't return a value.

frontend/model/contracts/group.js
corrideat: That's done later on `gi.contracts/group/leaveGroup`

frontend/model/contracts/identity.js
corrideat: There's nothing to delete or to read if the attributes aren't defined.

frontend/model/contracts/identity.js
corrideat: That requires the constructor being called. It won't be if the decryption key isn't available (for example, because it's been rotated).

shared/domains/chelonia/chelonia.js
corrideat: Probably

shared/domains/chelonia/internals.js
corrideat: It's defined for that error, unless someone generates an error by the same name manually.

shared/domains/chelonia/internals.js
corrideat: Other operations are also called via `processMutation`, but I updated the comment to say `processMessage`, which is more accurate.

frontend/main.js
corrideat: No, they are distinct values. `e.stack` will not have information about the caller, which is needed to figure out the offending location. If `e.stack` were enough, we wouldn't even need this code, since the exception will appear in the console.

frontend/model/contracts/group.js
corrideat: If you see the code that's been removed, it was informing the user of an issue there. That's not really possible with the SW (and even then, it'd lead to convoluted code), so this is meant to say that an event should be fired. I'll add the events.

test/cypress/integration/group-member-removal.spec.js
```diff
@@ -291,4 +291,40 @@ describe('Group - Removing a member', () => {
...
+  it('userBot joins groupB', () => {
+    cy.giLogin(`userbot-${userId}`, { toGroupDashboardUponSuccess: false, bypassUI: true })
+
+    cy.giAcceptGroupInvite(invitationLinks.anyone_groupB, {
+      username: `userbot-${userId}`,
+      existingMemberUsername: `user2-${userId}`,
+      groupName: groupNameB,
+      isLoggedIn: true,
+      // TODO: Sometimes it seems to get stuck here when bypassing the UI
+      bypassUI: false
```
taoeffect: It doesn't seem like this fixed the failing build, so can this be returned to `true`?

- Prior commit failure: https://cloud.cypress.io/projects/q6whky/runs/3885/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D
- This commit: https://cloud.cypress.io/projects/q6whky/runs/3886/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D

shared/domains/chelonia/internals.js
taoeffect: Wouldn't it still make sense then to do something here, like log a `console.error` in an `else { .. }` branch?

frontend/model/contracts/identity.js
taoeffect: Could you add a comment here explaining this? I think most people reading this code will have no idea about this.

test/cypress/integration/group-member-removal.spec.js
```diff
@@ -291,4 +291,40 @@ describe('Group - Removing a member', () => {
...
+    assertMembersCount(3)
+
+    cy.getByDT('groupMembers').contains('userbot-')
+    cy.window().its('sbp').then(sbp => {
+      const rootState = sbp('state/vuex/state')
+      const userbotId = rootState.namespaceLookups[`userbot-${userId}`]
+      const PEKs = Object.values(rootState[userbotId]._vm.authorizedKeys).filter(key => key.name === 'pek')
+      if (PEKs.length < 2) throw new RangeError('Expected the PEK to have been rotated')
+    })
+  })
```
taoeffect: Nice tests 👍

shared/domains/chelonia/internals.js
corrideat: Not really, the error is re-thrown. The error being falsy is not particularly exceptional.

frontend/model/contracts/identity.js
corrideat: Sure, I can, but then this is a caveat that IMO should be well-documented in the Chelonia documentation. Because the constructor is a 'regular' action, it needs to be readable for it to run. Hence, it's not guaranteed that the constructor will run before other actins. Actually, because of the way things are implemented, the constructor could _never_ be run, if it's not issued (we issue `OP_CONTRACT` followed by `OP_ACTION_ENCRYPTED`; one could issue just `OP_CONTRACT`).

shared/domains/chelonia/internals.js
corrideat:
```
                  // If we've just received a known missing key (i.e., a key
                  // that previously resulted in a decryption error), we know
                  // our state is outdated and we need to re-sync the contract
```

test/cypress/integration/group-member-removal.spec.js
corrideat: Yeah, I was trying to test this as tests get stuck there sometimes. I've simplified the tests now without using `userbot`.

shared/domains/chelonia/internals.js
taoeffect: > The error being falsy is not particularly exceptional.

I disagree... we shouldn't be seeing any code throwing such errors, as we have none. And if suddenly some code does throw a falsy error that's something we should definitely want to know about with a `console.error` at the very least.

corrideat: Well, contracts could throw such errors (either intentionally, or some via some library). While I agree that it'd be helpful for (Group Income) QA purposes to know about such errors, I'm not convinced that the situation is especially 'error-worthy'. The only reason I'm doing `if (e)` is to avoid having to do several `e?.` inside.

As far as I know, there's nothing 'wrong' with throwing, e.g., `null`, other than that it doesn't provide much information about what happened.

taoeffect (approved):
Amazing work @corrideat !

PR #2585, opened 2025-02-01
IPv6 rate limiting

### LINKED ISSUE #2832: Rate limiting for IPv6
### Problem

https://github.com/okTurtles/group-income/pull/2280 implemented rate-limiting, which has a few limitations as noted in <https://github.com/okTurtles/group-income/pull/2280#pullrequestreview-2219347378>. One of such limitations is that the raw IP address (from `x-real-ip` or the remote address directly) is used as a limiter key.

With some caveats, the current approach is reasonable for IPv4 addresses because:

1. All possible IPv4 addresses can fit into (enough) RAM, so this isn't necessarily an avenue for DoS.
2. IPv4 addresses are scarce, and the assumption is that a would-be adversary would incur a certain cost for each address.
3. Because of the scarcity point, these days normally IPv4 addresses are allocated to users / customers one-by-one. Even if an AS provided customers with several IPv4 addresses, normally they won't be sequential or in the same subnet.

However, the approach of using the raw IP address isn't valid or reasonable for IPv4 because _none_ of the assumptions below hold:

1. It's impossible to fit all IPv6 addresses in memory (`2**128` of them). Even fitting all IPv6 subnets (2**64) would require immense resources. This means that it's possible to DoS the server by just connecting from different IPv6 sources (meaning the rate limiting DB will get bigger and bigger).
2. IPv6 addresses aren't exactly scarce. They can be acquired for free or very cheaply.
3. Since they're not scarce, it's not uncommon at all for a customer / user to be allocated a subnet (a /64) or even several subnets. This means in practice that the lower 64 bits of an IPv6 can and should be treated as noise for the purposes of rate limiting or source attribution.

That said, there are some situations where it does make sense to process IPv6 addresses differently:

1. IPv6 addresses may 'contain' or embed IPv4 addresses, as part of various transition mechanisms. Since these really are IPv4-in-IPv6 representations, it makes sense to extract the IPv4 address and use that as a rate limiter key.
2. Just like IPv4, IPv6 supports link-local addresses. These are addresses given to a network interface and only locally-accessible (i.e., they're not publicly routable). Although these addresses are rare, when and if they do occur, it might not make sense to discard the lower bits.

### Solution

Adjust the IP rate limiting code to disregard the least significant 64 bits of IPv6 addresses.

corrideat:


backend/routes.js
```diff
@@ -42,6 +42,61 @@ const limiterPerDay = new Bottleneck.Group({
   reservoirRefreshAmount: SIGNUP_LIMIT_DAY
 })
 
+const IPV4_ADDR_REGEX = /^(?:[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])(?:\.(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])){3}$/
```
taoeffect: Claude:

Let me break down this IP regex and test it for potential issues:

1. The regex attempts to match IPv4 addresses
2. It's anchored with `^` and `$` correctly
3. It has 4 groups (one for each octet) separated by dots

Let's test some edge cases:

Valid IPs that SHOULD match:
- 192.168.1.1 ✓
- 255.255.255.255 ✓
- 0.0.0.0 ✗ (Bug: doesn't match zero)
- 1.2.3.4 ✓

Invalid IPs that should NOT match:
- 256.1.2.3 ✓ (Correctly fails)
- 300.1.2.3 ✓ (Correctly fails)
- 1.2.3.256 ✓ (Correctly fails)
- 1.2.3 ✗ (Correctly fails)
- a.b.c.d ✓ (Correctly fails)

I found one bug: The regex doesn't allow `0` as a valid number in the first octet. Here's the corrected version:

```js
/^(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])(?:\.(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])){3}$/
```

The changes made:
- Added `0|` at the beginning of the first group

This corrected version will now properly match all valid IPv4 addresses including those starting with 0.

Also, you might want to consider simplifying it to this more readable version that does the same thing:

```js
/^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[3-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[3-9][0-9]?)$/
```

Both versions will work, but the second one is more commonly used and easier to maintain.

backend/routes.js
```diff
@@ -42,6 +42,61 @@ const limiterPerDay = new Bottleneck.Group({
...
+        // /64
+        return segments.slice(0, 4).join(':').toLowerCase() + '::'
+      }
+    } else {
+      throw new Error('Invalid IPv6 address')
+    }
+  }
+
+  throw new Error('Invalid address format')
+}
```
taoeffect: Claude:

Let me analyze the code and point out potential issues:

1. **IPv4 Regex Issue**: The current IPv4 regex doesn't allow for numbers 0-2 in the first octet. It starts with `[3-9]`, which means valid IPs like `1.1.1.1` or `2.2.2.2` would be rejected.

2. **IPv6 Compression Handling**: There are a few edge cases in IPv6 compression handling that might not be properly covered:
   - Multiple `::` in the address (invalid but not explicitly checked)
   - The code assumes there will always be 8 segments after decompression, which might not be true for all valid IPv6 addresses

3. **Array Index Access**: The code assumes `segments[7]` exists without checking the array length, which could cause undefined behavior.

4. **IPv4-Mapped IPv6 Addresses**: The handling of IPv4-mapped IPv6 addresses (like `::ffff:192.168.1.1`) might not be complete.

Here's a corrected version:

```js
const IPV4_ADDR_REGEX = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
const IPV6_SEGMENT_REGEX = /^[0-9a-fA-F]{1,4}$/

const limiterKey = (ip: string) => {
  if (IPV4_ADDR_REGEX.test(ip)) {
    return ip
  } else if (/^(?:[0-9a-fA-F]{0,4}:){0,7}[0-9a-fA-F]{0,4}(?:%[0-9a-zA-Z]{1,})?$/.test(ip)) {
    // Likely IPv6
    const [address, zoneIdx] = ip.split('%')
    const segments = address.split(':')

    // Check for multiple :: (invalid)
    if (address.split('::').length > 2) {
      throw new Error('Invalid IPv6 address: multiple :: found')
    }

    // Is this a compressed form IPv6 address?
    let isCompressed = false
    for (let i = 0; i < segments.length; i++) {
      if (!isCompressed && segments[i] === '') {
        const requiredSegments = 8 - (segments.length - 1)
        if (requiredSegments < 1) {
          throw new Error('Invalid IPv6 address: too many segments')
        }
        segments.splice(i, 1, ...new Array(requiredSegments).fill('0'))
        isCompressed = true
        continue
      }
      if (segments[i] && !IPV6_SEGMENT_REGEX.test(segments[i])) {
        throw new Error('Invalid IPv6 address: invalid segment')
      }
      // Remove leading zeroes
      if (segments[i]) {
        segments[i] = parseInt(segments[i], 16).toString(16)
      }
    }

    // Handle IPv4-mapped addresses
    if (segments.length === 8 && IPV4_ADDR_REGEX.test(segments[7])) {
      return segments[7]
    } else if (segments.length === 8 && IPV6_SEGMENT_REGEX.test(segments[7])) {
      if (zoneIdx) {
        // Use tagged (link-local) addresses in full
        return segments.join(':').toLowerCase() + '%' + zoneIdx
      } else {
        // Return first 64 bits for regular IPv6
        return segments.slice(0, 4).join(':').toLowerCase() + '::'
      }
    }

    throw new Error('Invalid IPv6 address: invalid format')
  }

  throw new Error('Invalid address format')
}
```

Key improvements:
1. Fixed IPv4 regex to allow all valid numbers
2. Added check for multiple `::` in IPv6 addresses
3. More robust handling of IPv6 compression
4. Added array length checks
5. Better error messages
6. More consistent handling of leading zeros
7. Improved validation of IPv6 segments

Remember to test this code thoroughly with various edge cases:
- Regular IPv4 addresses
- IPv4-mapped IPv6 addresses
- Compressed IPv6 addresses
- Link-local IPv6 addresses with zone indices
- Invalid addresses

taoeffect (commented):
Nice. See comments from Claude:

backend/routes.js
corrideat: > The regex doesn't allow 0 as a valid number in the first octet

That was intentional. If you see https://www.rfc-editor.org/rfc/rfc1122, section 3.2.1.3(a)-(b), addresses starting with 0 are not valid source addresses.

corrideat: > 1.2.3 ✗ (Correctly fails)

Not sure where that's coming from, but it fails

backend/routes.js
```diff
@@ -63,6 +63,65 @@ const cidLookupTable = {
   [multicodes.SHELTER_FILE_CHUNK]: 'application/vnd.shelter.filechunk+octet-stream'
 }
 
+const IPV4_ADDR_REGEX = /^(?:[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])(?:\.(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])){3}$/
+const IPV6_SEGMENT_REGEX = /^[0-9a-fA-F]{1,4}$/
+const limiterKey = (ip: string) => {
+  if (IPV4_ADDR_REGEX.test(ip)) {
```
taoeffect: Is this the fastest way to do this or would something like `ip.split('.').each(x => 0 < parseInt(x) <= 255` be faster?

taoeffect (commented):
The suggestion to use `'net'` seems reasonable to me:

### Claude Opus 4 (Thinking) feedback

Looking at the changes, I've identified several bugs and areas for improvement in the IP address handling logic:

## 1. IPv4 Regex Excludes Valid Addresses

```javascript
const IPV4_ADDR_REGEX = /^(?:[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])(?:\.(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])){3}$/
```

This regex excludes IPv4 addresses starting with 0, 1, or 2 in the first octet, which means valid addresses like `1.1.1.1`, `2.2.2.2`, `0.0.0.0` would fail. Additionally, it excludes octets 1-2 in subsequent positions.

**Fix:**
```javascript
const IPV4_ADDR_REGEX = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
```

## 2. Leading Zero Removal Can Result in Empty String

```javascript
segments[i] = segments[i].replace(/^0+/, '0')
```

This line appears twice (lines 98 and 104). If a segment is "0000", this will correctly return "0", but the same logic on line 107:

```javascript
segments[7] = segments[7].replace(/^0+/, '0')
```

should be consistent. However, the real issue is that you're not handling this consistently throughout.

**Fix:**
```javascript
segments[i] = segments[i].replace(/^0+/, '') || '0'
```

## 3. Missing Error Handling in Route

```javascript
const keyedIp = limiterKey(ip)
```

The `limiterKey` function can throw errors, but there's no try-catch block around this call. If an invalid IP format is passed, it will cause an unhandled error.

**Fix:**
```javascript
let keyedIp
try {
  keyedIp = limiterKey(ip)
} catch (error) {
  console.warn('Invalid IP address format:', ip, error.message)
  throw Boom.badRequest('Invalid IP address format')
}
```

## 4. IPv6 Compression Logic Has Edge Cases

The compression handling logic (lines 79-92) is complex and may not handle all edge cases correctly. For example, `::1` would result in `segments = ['', '', '1']`, and the current logic might not handle this properly.

**Improved approach:**
```javascript
// Handle :: compression more robustly
if (address.includes('::')) {
  const parts = address.split('::')
  if (parts.length > 2) {
    throw new Error('Invalid IPv6 address: multiple :: found')
  }
  const left = parts[0] ? parts[0].split(':') : []
  const right = parts[1] ? parts[1].split(':') : []
  const missing = 8 - left.length - right.length
  if (missing < 1) {
    throw new Error('Invalid IPv6 address: too many segments')
  }
  segments = [...left, ...Array(missing).fill('0'), ...right]
} else {
  segments = address.split(':')
  if (segments.length !== 8) {
    throw new Error('Invalid IPv6 address: wrong number of segments')
  }
}
```

## 5. IPv4-Mapped IPv6 Address Detection

The current logic checks if the 8th segment is an IPv4 address, but IPv4-mapped addresses like `::ffff:192.0.2.1` would have the IPv4 part in a different position after decompression.

**Fix:**
```javascript
// Check for IPv4-mapped or IPv4-compatible addresses
if (segments.length === 8) {
  // Check for ::ffff:x.x.x.x (IPv4-mapped)
  if (segments[0] === '0' && segments[1] === '0' && segments[2] === '0' && 
      segments[3] === '0' && segments[4] === '0' && segments[5] === 'ffff' && 
      segments[6] === '0') {
    // IPv4-mapped address
    if (IPV4_ADDR_REGEX.test(segments[7])) {
      return segments[7]
    }
  }
  // Your existing logic for other cases...
}
```

## Additional Suggestions:

Consider using a well-tested library like `ip-address` or Node's built-in `net.isIP()` for IP validation and parsing instead of implementing this logic from scratch. This would be more reliable and maintainable:

```javascript
const { isIPv4, isIPv6 } = require('net')
const { Address6 } = require('ip-address')

const limiterKey = (ip: string) => {
  if (isIPv4(ip)) {
    return ip
  } else if (isIPv6(ip)) {
    const addr = new Address6(ip)
    if (addr.v4) {
      // IPv4-mapped address
      return addr.v4.address
    }
    // Return /64 subnet
    return addr.subnet('/64').canonicalForm()
  }
  throw new Error('Invalid address format')
}
```

backend/routes.js
corrideat: Not sure. It might or might not be faster but a regex seems like the right tool for this. Also, your example will accept invalid addresses such as `1foo.2bar.3baz.4`, addresses of dubious validity such as `000000001.2.3.4`, and then also the length needs to be checked (there should be 4 segments). It'll also accept `0.x.x.x` which are _not_ valid addresses (although this is a minor issue)

taoeffect: Yeah ignore my suggestion, it's a bad one. The one from Opus might be better 😅

corrideat:
> 1. IPv4 Regex Excludes Valid Addresses

It does _not_ exclude valid addresses. It accepts addresses starting with `1`, `2`, `3`, `4`, etc. It doesn't accept addresses starting with `0` because those addresses are invalid.

The suggestion to use `node:net` (it'd be `isIP` that needs to be used) does sound reasonable. `isIP` seems to accept `0.x.x.x` addresses, which aren't valid source addresses, but, like I said in another comment, perhaps this is only really a minor issue.

> 2. Leading Zero Removal Can Result in Empty String

I don't see how this could happen.

> 3. Missing Error Handling in Route

Sure, but then leaving this as they are would also be handled by the server returning 500.

> 4. IPv6 Compression Logic Has Edge Cases

Not sure what this is referring to, but `::1` correctly gives `0:0:0:0::`. Actually, in this particular case the result likely should be `0:0:0:0:0:0:0:1` since `::1` is a special case, but probably not worth handling.

> 5. IPv4-Mapped IPv6 Address Detection

`limiterKey('::ffff:192.0.2.1')` correctly gives `192.0.2.1` (edit: I originally wrote `0:0:0:0::`, which is wrong but also not what `limiterKey` returns).


As with all of the other previous AI reviews, dismissing most of the points as there doesn't seem to be much substance. It's true that `net` could avoid the need for using regexps (and may or may not be faster) and that errors thrown by `limiterKey` could be handled especially. I will address these points if I have time after working on other issues.

corrideat:
It looks like I'll be using `net`'s `isIP`, but in case I don't, to settle the LLMs issue of 'excluding' certain addresses:

```js
(()=>{
    const r = /^(?:[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])(?:\.(?:0|[3-9]\d{0,1}|1\d{0,2}|2\d{0,1}|2[0-4]\d|25[0-5])){3}$/;
    for(let i=1;i<256;i++){console.error(i);for(let j=0;j<256;j++)for(let k=0;k<256;k++)for(let l=0;l<256;l++)if(!r.test(`${i}.${j}.${k}.${l}`))throw new Error(`Rejected valid address ${i}.${j}.${k}.${l}`)}
    console.error('Done')
})()
```

Results in no errors being thrown and `1`, `2`, `3`, ..., `254`, `255`, `Done` printed out. This means that _all_ valid addresses are accepted, except for the ones that start with `0`, which are valid in some cases but should never appear as a source address.

backend/routes.js
```diff
@@ -63,6 +64,61 @@ const cidLookupTable = {
...
+        // /64
+        return segments.slice(0, 4).join(':').toLowerCase() + '::'
+      }
+    } else {
+      throw new Error('Invalid IPv6 address')
+    }
+  }
+
+  throw new Error('Invalid address format')
+}
```
taoeffect, line 128: OK, I do not understand why this code exists at all. I read the comment linked to in `// See discussion: https://github.com/okTurtles/group-income/pull/2280#pullrequestreview-2219347378` but it didn't really address this AFAICT.

Why are you doing all of this processing for ipv6 IPs? Why are you creating a `limiterKey` function? Why does this PR exist at all? I notice that no issue is referenced and therefore no problem is referenced, and I genuinely don't understand why the code should be modified.

Could you please update the code with a comment that addresses these questions?

taoeffect (changes requested):
Hey @corrideat, I just realized that I don't understand what this PR is doing and why. I suppose this is why issues are important to be associated with PRs...

backend/routes.js
corrideat: Sure, I can create an issue, but I'm not sure what exactly isn't being addressed in the linked comment. (See also https://github.com/okTurtles/group-income/pull/2280#discussion_r1704357074 and your comment acknowledging this https://github.com/okTurtles/group-income/pull/2280#discussion_r1704389923)

The existing code will use whatever's given as IP as a limiter key. The linked comment says:

> d. At the very least, the key used should be the /64 block for IPv6 addresses. Not doing this makes this PR pointless and exposes the server to very easy DoS.

This PR implements using the /64 block for IPv6 addresses.

<details>
<summary>o3-mini (Why does this PR exist?)</summary>
This PR exists to improve the rate limiting mechanism by correctly handling IPv6 addresses. Previously, the rate limiter was keyed directly on the raw IP string, which worked fine for IPv4 addresses but was insufficient for IPv6. The comments in the diff point out that using IPv6 addresses as-is could allow spammers to bypass the rate limit more easily (for instance, by varying parts of the address within a /64).

Key changes include:

1. Importing the isIP function from Node’s net module to help determine if an address is IPv4 or IPv6.
2. Implementing a new function (limiterKey) that:
* Checks if the provided IP is IPv4 and uses it directly.
* For IPv6 addresses, it normalizes the address:
  * It handles compressed and zone-indexed IPv6 addresses.
  * For standard IPv6, it extracts a common prefix—specifically the first 64 bits (/64 subnet)—to ensure that rate limiting is applied at a subnet granularity. This prevents someone from generating multiple unique full IPv6 addresses as they would all map to the same key.
  * In cases of IPv4-embedded IPv6 addresses, it falls back to using the IPv4 form.
3. Updating the endpoints to use the normalized keyed IP (returned by limiterKey) rather than the raw IP when scheduling requests with the rate limiter.

Overall, this PR addresses a potential security and reliability concern by ensuring rate limiting effectiveness across both IPv4 and IPv6 address spaces.
</details>

taoeffect (approved):
Thanks for the explaination @corrideat! Makes sense!

PR #2586, opened 2025-02-03
#2581 - Improve login-error prompt style

### LINKED ISSUE #2581: Improve prompt that appears on login to forked chain
### Problem

This dialog:

- Has centered text
- Has a "Yes" button on the left and not default

Image

### Solution

- Make the text left-aligned
- Put the "Yes" button on the right and make it the default button.

SebinSong:
closes #2581 

[ screenshot ]

### DIFF
--- a/frontend/controller/app/identity.js
+++ b/frontend/controller/app/identity.js
@@ -432,12 +432,14 @@ export default (sbp('sbp/selectors/register', {
             question: wipeOut
               ? L('The server\'s history for your identity contract has diverged from ours. This can happen in extremely rare circumstances due to either malicious activity or a bug. {br_}To fix this, the contract needs to be resynced, and some recent events may be missing. {br_}Would you like to log out and resync data on your next login? {br_}Error details: {err}.', { err: errMessage, ...LTags() })
               : L('Do you want to log out? {br_}Error details: {err}.', { err: errMessage, ...LTags() }),
-            primaryButton: L('No'),
-            secondaryButton: L('Yes')
+            primaryButton: L('Yes'),
+            secondaryButton: L('No'),
+            primaryButtonStyle: 'primary', // make primary button 'filled' style
+            isContentCentered: !wipeOut
           }
 
           const result = await sbp('gi.ui/prompt', promptOptions)
-          if (!result) {
+          if (result) {
             sbp('gi.ui/clearBanner')
             return sbp('gi.app/identity/_private/logout', state, wipeOut)
           } else {
--- a/frontend/views/components/modal/Modal.vue
+++ b/frontend/views/components/modal/Modal.vue
@@ -58,13 +58,19 @@ export default ({
           if (this.content) this.replaceModal(toModal, omit(to.query, 'modal')) // if another modal is already open, replace it.
           else this.content = toModal
         }
-        const subcontent = to.query.subcontent ? to.query.subcontent.split('+').pop() : []
-        if (subcontent !== this.activeSubContent) {
-          // Try to find the new subcontent in the list of subcontent
-          const i = this.subcontent.indexOf(subcontent)
-          if (i !== -1) {
-            this.subcontent = this.subcontent.slice(0, i)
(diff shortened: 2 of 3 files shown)


frontend/controller/app/identity.js
```diff
@@ -432,12 +432,14 @@ export default (sbp('sbp/selectors/register', {
...
-            secondaryButton: L('Yes')
+            primaryButton: L('Yes'),
+            secondaryButton: L('No'),
+            primaryButtonStyle: 'primary', // make primary button 'filled' style
+            isContentCentered: !wipeOut
           }
 
           const result = await sbp('gi.ui/prompt', promptOptions)
-          if (!result) {
+          if (result) {
```
SebinSong, line 442: Now that 'Yes' is the primary button, this check must be negated.

frontend/views/components/modal/Modal.vue
```diff
@@ -58,13 +58,19 @@ export default ({
...
+          const subContentTo = to.query.subcontent.split('+').pop()
+          if (subContentTo !== this.activeSubContent) {
+            // Try to find the target subcontent in the list of subcontent
+            const i = this.subcontent.indexOf(subContentTo)
+            if (i !== -1) {
+              this.subcontent = this.subcontent.slice(0, i)
+            } else {
+              this.subcontent = [subContentTo]
+            }
+          }
```
SebinSong, line 73: This is irrelevant to the issue itself but a fix for a bug I encountered while working on the issue.
The bug was that, when the app is reloaded and it has 'sign-up' modal and also 'login-error' prompt on top of it, `this.subcontent` becomes a weird string value (it was alphabet `t` in my case) while the`Modal.vue` is assuming that `this.subcontent` is an array of strings at all times.
As a result, I was unable to close `SignupModal.vue`. So had to update this part with fixes like `L:71` above.

frontend/controller/app/identity.js
```diff
@@ -432,12 +432,14 @@ export default (sbp('sbp/selectors/register', {
             question: wipeOut
               ? L('The server\'s history for your identity contract has diverged from ours. This can happen in extremely rare circumstances due to either malicious activity or a bug. {br_}To fix this, the contract needs to be resynced, and some recent events may be missing. {br_}Would you like to log out and resync data on your next login? {br_}Error details: {err}.', { err: errMessage, ...LTags() })
               : L('Do you want to log out? {br_}Error details: {err}.', { err: errMessage, ...LTags() }),
-            primaryButton: L('No'),
-            secondaryButton: L('Yes')
+            primaryButton: L('Yes'),
+            secondaryButton: L('No'),
+            primaryButtonStyle: 'primary', // make primary button 'filled' style
```
taoeffect, line 437: Why is it that the default style is called `'outlined'`, and to make it filled, we call it `'primary'` instead of `'filled'`?

Wouldn't it make more sense to call it `'filled'`?

Note: if you change this, you'd need to look through the rest of the codebase for 'primary' and make sure everything is updated accordingly. Or you could add the style as an alias to `primary`, without changing the existing definition of `primary`, which would be safer.

taoeffect (commented):
Great work @SebinSong!

I could merge as-is, since everything works, or we could add the `'filled'` styled as part of this PR, or leave that as a separate issue. Let me know which you prefer!

frontend/controller/app/identity.js
SebinSong: > Why is it that the default style is called 'outlined', and to make it filled, we call it 'primary' instead of 'filled'?

Well, the `Prompt.vue` had both buttons as 'outlined' style before this PR and I'm only changing the style of one button for a particular prompt(Login Error) in this PR..

> Wouldn't it make more sense to call it 'filled'?

Nowhere in this project's CSS is below button style called 'filled' but instead it is called 'primary'.





(The term 'filled' is something people at my other company use for that kind of button style. I think I got it from there)

taoeffect (approved):
Nicely done!

PR #2590, opened 2025-02-06
#2548, #2430 - Fix chat scroll related bugs

### LINKED ISSUE #2430: Scrolling to a specific message using `mhash` key in query string does not work sometimes.
### Problem

NOTE: it did work for Greg in GIG group but did not work when I tried the same message link on my MacOS and also when tried with the latest `master` locally, hence the word 'sometimes' in the title.

**[Steps to reproduce]**

1. Create a group and go to the `#general` chatroom.
2. Add some messages so the scrollbar appears.
3. Copy one of the message link and try it.
6. The chatroom page does appear but it does not auto-scroll to the target message (but rather scroll to the top) and `mhash` key in the query string disappears too.

### Solution
Find out why this functionality has been lost and fix it.

### LINKED ISSUE #2548: Chatroom scrolling bug
### Problem

Switching between chatrooms when you've left scrolled to the bottom doesn't show the bottom when you switch back:

https://github.com/user-attachments/assets/f0a31c0d-df93-4717-bd3e-f183645a39b0

### Solution

Make sure that you're scrolled all the way down, while still preserving the ability to remember where you were.

If you're far enough down it should scroll you all the way to the bottom.

*Note: implement #2556 before implementing this issue.

SebinSong:
closes #2548 
closes #2430 

**[Fix for issue 2548]**

https://github.com/user-attachments/assets/9088d09a-d391-4b9b-8a39-91068d72ff21

**[Fix for issue 2430]**

https://github.com/user-attachments/assets/0c30bb87-b9bc-4226-b27f-0c874bf6e075

---

Even though I still have not demystified 100% of How message states in chatroom are managed / entire steps of how message states are initialized, I tried my best to ensure these updates don't end up breaking any existing behavior the chatroom has. Thanks.

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -191,7 +191,7 @@ const onChatScroll = function () {
     return
   }
 
-  if (curScrollTop > ignorableScrollDistanceInPixel) {
+  if (this.ephemeral.scrolledDistance > ignorableScrollDistanceInPixel) {
     // Save the current scroll position per each chatroom
     for (let i = 0; i <= this.messages.length - 1; i++) {
       const msg = this.messages[i]
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -166,10 +166,13 @@ const onChatScroll = function () {
 
   for (let i = this.messages.length - 1; i >= 0; i--) {
     const msg = this.messages[i]
+    const msgEl = this.$refs[msg.hash][0].$el
+
     if (msg.pending || msg.hasFailed) continue
-    const offsetTop = this.$refs[msg.hash][0].$el.offsetTop
+    const offsetTop = msgEl.offsetTop
     // const parentOffsetTop = this.$refs[msg.hash][0].$el.offsetParent.offsetTop
-    const height = this.$refs[msg.hash][0].$el.clientHeight
+    const height = msgEl.clientHeight
+
     if (offsetTop + height <= curScrollBottom) {
       const bottomMessageCreatedHeight = msg.height
       const latestMessageCreatedHeight = this.currentChatRoomReadUntil?.createdHeight
@@ -188,9 +191,9 @@ const onChatScroll = function () {
     return
   }
 
-  if (this.ephemeral.scrolledDistance > ignorableScrollDistanceInPixel) {
+  if (curScrollTop > ignorableScrollDistanceInPixel) {
     // Save the current scroll position per each chatroom
-    for (let i = 0; i < this.messages.length - 1; i++) {
+    for (let i = 0; i <= this.messages.length - 1; i++) {
       const msg = this.messages[i]
       if (msg.pending || msg.hasFailed) continue
       const offsetTop = this.$refs[msg.hash][0].$el.offsetTop
@@ -257,6 +260,7 @@ export default ({
         // NOTE: messagesInitiated describes if the messages are fully re-rendered
         //       according to this, we could display loading/skeleton component
         messagesInitiated: undefined,
+        scrollHashOnInitialLoad: null, // Message hash to scroll to on chatroom's initial load
         replyingMessage: null,
         replyingTo: null,
         unprocessedEvents: []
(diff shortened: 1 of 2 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -166,10 +166,13 @@ const onChatScroll = function () {
 
   for (let i = this.messages.length - 1; i >= 0; i--) {
     const msg = this.messages[i]
+    const msgEl = this.$refs[msg.hash][0].$el
```
SebinSong, line 169: DRYing reference to a particular message DOM element here.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -544,25 +548,27 @@ export default ({
...
         if (effect) {
-          eleTarget.scrollIntoView({ behavior: this.isReducedMotionMode ? 'instant' : 'smooth' })
+          eleMessage.scrollIntoView({ behavior: this.isReducedMotionMode ? 'instant' : 'smooth' })
           eleMessage.classList.add('c-focused')
           setTimeout(() => {
             eleMessage.classList.remove('c-focused')
           }, 1500)
         } else {
-          eleTarget.scrollIntoView()
+          eleMessage.scrollIntoView()
```
SebinSong, line 561: This part is the fix for #2548. (It's unclear what the original intention here was but) `eleTarget` here is a message placed above the **actual target** message and `eleTarget.scrollIntoView(...)` leads to not fully scrolling to the target message. (especially when that previous message has associated file-attachments and takes up large space in the DOM). So made a fix here.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -858,23 +864,7 @@ export default ({
...
-          // NOTE: we need to scroll to the message first in order to no more infiniteHandler is called
-          await this.updateScroll(messageHashToScroll, Boolean(mhash))
-
-          if (mhash) {
-            // NOTE: delete mhash in the query after scroll and highlight the message with mhash
-            const newQuery = { ...this.$route.query }
-            delete newQuery.mhash
-            this.$router.replace({ query: newQuery })
-          }
-        }
```
SebinSong, line 877: This logic here handles scrolling to a particular message after message states are initialized. And moving this logic to below `this.ephemeral.messagesInitiated = true` is the fix for issue #2430 .
I noticed `infiniteHandler()` and `renderMoreMessages()` methods are called multiple times throughout the entire chatroom initialization process (which honestly I have not been able to figure out yet exactly from where and why..) and **observed** that executing scrolling logic here looked premature most of the time.
So moved this logic and it fixed the `mhash` issue.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -188,14 +191,14 @@ const onChatScroll = function () {
     return
   }
 
-  if (this.ephemeral.scrolledDistance > ignorableScrollDistanceInPixel) {
+  if (curScrollTop > ignorableScrollDistanceInPixel) {
```
SebinSong, line 194: This is part of the fix for #2548 too.
Using `this.ephemeral.scrolledDistance` here is apparently wrong. If we see L165 below:
https://github.com/okTurtles/group-income/blob/5b89d7022aca2200db3d2fb01f94926696cda31f/frontend/views/containers/chatroom/ChatMain.vue#L164-L165

`ephemeral.scrolledDistance` isn't actually calculating scroll**ED** distance but rather measuring sroll**ABLE** distance. This value gets closer to `0` as the chatroom scrolls closer to the bottom(the latest message) and as a result the logic in below if block (which saves the current scroll position of the chatroom) 

```js
if (curScrollTop > ignorableScrollDistanceInPixel) { ... }
```

ends up not getting executed for a couple of most recent messages, in some cases.
So replaced it with `currScrollTop` (which is the correct variable that calculates currently scroll**ED** distance)

---
I didn't go ahead and update `this.ephemeral.scrolledDistance` itself in this PR because, this might cause unexpected bugs in other places where this is currently used.

taoeffect (changes requested):
Great work @SebinSong!

I can confirm that #2430 seems to be closed, however #2548 still seems to be an issue, and there are two problems:

### Problem 1: Scrolling to end still doesn't work at certain viewport sizes

When the browser window is a certain height/width, it still doesn't seem to work:


https://github.com/user-attachments/assets/5e0fa6b7-4ca1-4646-83eb-7263ac166f57

### Problem 2: Remembering scroll position now completely broken

The behavior we want is that if we're scrolled **near the end**, then when we switch back to that channel, it takes us all the way to the very end.

However, if we've scrolled kind of far up, then upon switching channels back, we should scroll to where we were.

With this PR now it seems that that behavior is completely broken, as it no longer remembers scroll position in my testing in FFDev.

SebinSong:
@taoeffect 
I'm actually closing this PR and just created a separate PR for #2430 in [here](https://github.com/okTurtles/group-income/pull/2598), As I think #2548 needs a closer investigation.

I will probably spend some more time on it in the future and send a separate PR Or, someone else could work on it(Tbh, I think whoever worked on implementing `NEW_CHATROOM_UNREAD_POSITION` related logic in `ChatMain.vue` in the first place is the best person to take this issue).

Issue #2588, opened 2025-02-03
LastLoggedIn still broken 😮‍💨


taoeffect:
### Problem

![Image](https://github.com/user-attachments/assets/1e3ba794-e6df-4969-b5ed-11b2694f9fde)

The `uCHELONIA_STATE`:

![Image](https://github.com/user-attachments/assets/1ad06328-bcf3-4d22-bdf5-f53c75fe716e)

The other state:

![Image](https://github.com/user-attachments/assets/0ecc20e7-661e-4e32-967c-3d59e88c78a8)

One other potential problem is that we use `new Date()` to get this value:

```js
    const now = new Date().toISOString()
```

The user's clock could be wrong.

### Solution

1. Figure out why it's not being updated correctly and fix it.
2. Use `new Date(sbp('chelonia/time'))` instead of naked `new Date()`

taoeffect:
![Image](https://github.com/user-attachments/assets/37255192-2e8c-4df7-bbfb-74f16172a24f)

Notice the Error updating lastLoggedIn `Error: kv/set invalid response status: 401`

And there are `Invalid signature time range` messages on the server, so maybe this is the cause?

taoeffect:
We've narrowed this down to likely being related to `?.data`:

```js
return (await sbp('chelonia/kv/get', contractID, KV_KEYS.LAST_LOGGED_IN))?.data || Object.create(null)
```

And related code, e.g.:

```js
    get data () {
      if (!encryptedData) return
      if (isSignedData(encryptedData.data)) {
        return encryptedData.data.valueOf()
      }
      return encryptedData.data
    },
```

Note there's the same issue in `identity-kv.js`:

```js
return (await sbp('chelonia/kv/get', identityContractID, KV_KEYS.UNREAD_MESSAGES))?.data || {}
```

And who knows, maybe that's related to fixing #2442 too?

corrideat:
I've been investigating this issue, and, while I can see on `groupincome.app` that it's apparently happening, the mechanism for it seems unclear.

The update happens by calling `'gi.actions/group/kv/updateLastLoggedIn'`, which has the following logic:

```js
      const getUpdatedLastLoggedIn = async (contractID) => {
        const current = await sbp('gi.actions/group/kv/fetchLastLoggedIn', { contractID })
        return { ...current, [identityContractID]: nowString }
      }

      const data = await getUpdatedLastLoggedIn(contractID)
[...]

PR #2592, opened 2025-02-07
Add `wsOnly` parameter to avoid broadcasting KV updates.

### LINKED ISSUE #2589: Frequent "site has been updated in the background" notifications for PWA
### Problem

On Android with a Brave PWA, I am getting these notifications entirely too frequently (many times an hour):

Image

What's more, is that these notifications often come in pairs for both the live site and testing, at the same time. This indicates that something else is going on than contract activity, especially the contracts in both don't have much activity.

### Solution

- First find out why this is happening, perhaps by using more verbose logging with `console.info` (will the logs be captured if the PWA isn't open..?)
- Then fix it so that they appear only to the minimum extent necessary.

EDIT: @corrideat has a theory that it may be that we're broadcasting KV updates over push. If so, we should disable this, as I see no reason to do so. Disabling will cause this message to appear less frequently. When we come back online (because the user opens the app) we should immediately query the KV stores for updates.

These KV updates are being broadcast multiple times a minute on the debug server, which almost nobody is using, so these definitely shouldn't be sent over push anyway:

Image

corrideat:
Changes to `kv/get` to propagate decryption errors (related to #2588).

Closes #2589.

### DIFF
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -30,12 +30,26 @@ export default (sbp('sbp/selectors/register', {
     console.info('group key-value store data loaded!')
   },
   'gi.actions/group/kv/fetchLastLoggedIn': async ({ contractID }: { contractID: string }) => {
-    return (await sbp('chelonia/kv/get', contractID, KV_KEYS.LAST_LOGGED_IN))?.data || Object.create(null)
+    const kvData = await sbp('chelonia/kv/get', contractID, KV_KEYS.LAST_LOGGED_IN)
+    // kvData could be falsy if the server returns 404
+    if (kvData) {
+      // Note: this could throw an exception if there's a signature or decryption
+      // issue
+      return kvData.data
+    }
+    return Object.create(null)
   },
   'gi.actions/group/kv/loadLastLoggedIn': ({ contractID }: { contractID: string }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const data = await sbp('gi.actions/group/kv/fetchLastLoggedIn', { contractID })
       sbp('okTurtles.events/emit', NEW_LAST_LOGGED_IN, [contractID, data])
+
+      // If running in a SW, update lastLoggedIn record. This is comparable to
+      // what the browser does when receiving the NEW_LAST_LOGGED_IN event
+      if (typeof WorkerGlobalScope === 'function') {
+        const rootState = sbp('state/vuex/state')
+        rootState.lastLoggedIn[contractID] = data
+      }
     }).catch(e => {
       console.error('[gi.actions/group/kv/loadLastLoggedIn] Error loading last logged in', e)
     })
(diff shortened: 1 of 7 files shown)


frontend/controller/actions/group-kv.js
```diff
@@ -30,12 +30,26 @@ export default (sbp('sbp/selectors/register', {
...
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const data = await sbp('gi.actions/group/kv/fetchLastLoggedIn', { contractID })
       sbp('okTurtles.events/emit', NEW_LAST_LOGGED_IN, [contractID, data])
+
+      // If running in a SW, update lastLoggedIn record. This is comparable to
+      // what the browser does when receiving the NEW_LAST_LOGGED_IN event
+      if (typeof WorkerGlobalScope === 'function') {
+        const rootState = sbp('state/vuex/state')
+        rootState.lastLoggedIn[contractID] = data
+      }
```
taoeffect: - Can you add a comment explaining why this code is being done?
- If it needs to be done shouldn't it be done as part of all the KV code? (e.g. identity.js and any other keys?)

frontend/controller/actions/group-kv.js
```diff
@@ -46,21 +60,22 @@ export default (sbp('sbp/selectors/register', {
...
       if (lastLoggedInRawValue) {
         const lastLoggedIn = new Date(lastLoggedInRawValue).getTime()
 
-        if ((Date.now() - lastLoggedIn) < LAST_LOGGED_IN_THROTTLE_WINDOW) return
+        if ((now - lastLoggedIn) < LAST_LOGGED_IN_THROTTLE_WINDOW) return
       }
     }
 
-    const now = new Date().toISOString()
+    const nowString = new Date(now).toISOString()
```
taoeffect, line 67: Same change needed for all other KV updates?

shared/domains/chelonia/chelonia.js
```diff
@@ -1638,30 +1638,81 @@ function parseEncryptedOrUnencryptedMessage ({
     return maybeEncryptedIncomingData(contractID, state, message, numericHeight, this.transientSecretKeys, aad, undefined)
   })
 
-  const encryptedData = unwrapMaybeEncryptedData(v.valueOf())
+  // Cached values
+  let encryptionKeyId
+  let innerSigningKeyId
+
+  // Lazy unwrap function
```
taoeffect, line 1662: Could you expand this comment a bit to explain why it's being done this way?

taoeffect (changes requested):
Great work @corrideat ! Prelim review ready!

frontend/controller/actions/group-kv.js
```diff
@@ -46,21 +60,22 @@ export default (sbp('sbp/selectors/register', {
       throw new Error('Unable to update lastLoggedIn without an active session')
     }
 
+    const now = sbp('chelonia/time') * 1000
```
taoeffect, line 56: ... what is the timescale returned by '`chelonia/time'`? Isn't it the same as `Date.now()`?

```
> new Date(Date.now())
2025-02-07T19:48:52.406Z
> new Date(Date.now()*1000)
+057075-05-18T15:42:05.000Z
> 
```

*EDIT: I see it's seconds... maybe we should use the same as `Date.now()` (and update existing code? Or is that a breaking change?)

corrideat: I disagree, since POSIX time is much more common than millisecond time outside of JS-land.

taoeffect: Alright, we can return to this later maybe, opened discussion for it: #2593

frontend/controller/actions/group-kv.js
corrideat: Are there other KV values that are date-based? I see only notifications, which have:

```
```

taoeffect: Yeah, that call to `Date.now()` should probably be turned into `'chelonia/time'` for consistency right?

EDIT: since `data.createdDate` is also based on `'chelonia/time'`

corrideat: Using `'chelonia/time'` here, which is mostly called from contracts, might need to be `async`, which would make the entire function async. I'd leave it as is an add a comment.

taoeffect: > which would make the entire function async

...why? `'chelonia/time'` is sync... not async

corrideat: I deleted the comment after checking where it's imported. But if `chelonia/time` is called from a contract, or from the browser it could be upgraded to async. However, we might need to upgrade contracts to await on calls to emit notifications, or basically anywhere external (i.e., defined outside the contract) selectors are used.

taoeffect (dismissed):
Fantastic work @corrideat !

taoeffect (approved):
Amazing work!

PR #2594, opened 2025-02-08
Rename `chelonia/db/` to `chelonia.db/`.

### LINKED ISSUE #2162: Rename `chelonia/db` to `chelonia.db` and lock `chelonia` domain
### Problem

The `chelonia` domain has access to various important state. It should not be possible to register a new selector in it after it is loaded.

### Solution

1. Lock the entire `chelonia` domain.
2. Since `chelonia/db/*` selectors are intended to be overwritable after Chelonia is loaded, rename them to be under the `chelonia.db` domain.

Consider also loading data [like `CHELONIA_STATE`](https://github.com/okTurtles/group-income/pull/2137#discussion_r1667427783) using `chelonia.db` instead of `gi.db`.

corrideat:
Some `chelonia/db` selectors are still defined, but those are not meant to be overwritten.

Closes #2162.


taoeffect (changes requested):
Great work @corrideat!

To close the issue completely the domain `chelonia` needs to be [locked using `'sbp/domains/lock'`](https://github.com/okTurtles/sbp/blob/master/docs/sbp-api.md#sbpdomainslock).

Could you please do that at the very end of the `chelonia.js` file?

taoeffect (approved):
Great work @corrideat!

PR #2595, opened 2025-02-09
User-local snapshots

corrideat:
This PR has the goal of speeding up login by:

1. Implementing a fast-path login, which doesn't require syncing all contracts before the state is persisted locally. This could allow for the login process to be interruptible.
2. Implementing an even faster path by saving the local state remotely on the sever (e.g., using the KV store), and restoring it when logging in, if possible. This would allow for an even faster login experience on new devices.

### DIFF
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -23,6 +23,31 @@ export default (sbp('sbp/selectors/register', {
     await sbp('gi.actions/identity/kv/loadNotificationStatus')
     console.info('identity key-value store data loaded!')
   },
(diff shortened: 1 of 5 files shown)


taoeffect:
Also, could this PR add support for sync progress? It would be great if we could display a progress bar along with a timer so that the user knows about how long it'll take for the sync to finish.

E.g. even though we don't know exactly the size of the full sync in advance, we do know how quickly we are syncing each message, and since on average most messages will have an average size, and since we know how many messages there are left to sync, we can estimate a completion time.

EDIT: here's an example of a progress bar that combines the progress of all currently syncing contracts. It shows which message we're on for one of the contracts (e.g. the identity.js contract), as well as which contract we're on:

a

PR #2598, opened 2025-02-10
#2430 - Broken chat `mhash`

### LINKED ISSUE #2430: Scrolling to a specific message using `mhash` key in query string does not work sometimes.

SebinSong:
closes #2430 

This is a follow-up PR of the PR https://github.com/okTurtles/group-income/pull/2590 (pls refer to [this comment](https://github.com/okTurtles/group-income/pull/2590#issuecomment-2646686031))

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -257,6 +257,7 @@ export default ({
         // NOTE: messagesInitiated describes if the messages are fully re-rendered
         //       according to this, we could display loading/skeleton component
         messagesInitiated: undefined,
+        scrollHashOnInitialLoad: null, // Message hash to scroll to on chatroom's initial load
         replyingMessage: null,
         replyingTo: null,
         unprocessedEvents: []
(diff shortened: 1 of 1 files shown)


SebinSong:
I've actually gone ahead and made another fix for the issue #2548 in [this commit](https://github.com/okTurtles/group-income/pull/2598/commits/cb219b13ab17f727c7bacc4febb847db1b42adbe) again.

It seems to work well most of times from my observation.
(One thing I've observed is that when the message chain of the chatroom has many attachments, sometimes old images take long time to load and rendered after the chatroom's message states are already initialized and it scrolls to the initial position. This results in a sudden jump to an arbitrary scroll position from that initial scroll position as the increased content height leads to incresed scrollable area too)

Other than this, if you still think that commit is not a fix for the #2548. I will revert that commit so that this PR can be merged as a fix for #2430.

Thanks,

taoeffect:
@SebinSong try making the window small and you'll see that even with no attachments it does not scroll all the way down:

Screenshot 2025-02-09 at 7 01 02 PM

I've noticed that in your attempts to fix #2548, you're relying on using `.scrollIntoView`.

Recall my comment here:

> The behavior we want is that if we're scrolled near the end, then when we switch back to that channel, it takes us all the way to the very end.

Have you tried for example, if `targetIsLatestMessage`, then _not_ using `.scrollIntoView` but some other method?

For example, AFAICT, all you need to do is the same thing that we already do when "Jump to latest message" banner is clicked, because that's literally what I do when I try to get it to go to the end (see screenshot above).

SebinSong:
**1)**
> The behavior we want is that if we're scrolled near the end, then when we switch back to that channel, it takes us all the way to the very end.

Apparently `ChatMain.vue` has an existing logic to handle this:

https://github.com/okTurtles/group-income/blob/592255670ec4e8dadf28a43a1ae37868e2813cba/frontend/views/containers/chatroom/ChatMain.vue#L191-L205

`ignorableScrollDistanceInPixel` here is `500px` and that seems to be a value that defines `near the end`. (The logic above is to only saves the current scroll-position if scroll is yet to hit `500px left from the bottom` but it goes beyond that point, save the latest message for the scroll position)
Thought I should leave it as is and that's why I discarded the previous PR that made updates to this logic.


**2)**
> all you need to do is the same thing that we already do when "Jump to latest message" banner 

Made this update in [this commit](https://github.com/okTurtles/group-income/pull/2598/commits/f7080d23934c06c69d69550e7224f4221331c13d).

It does seem to work well for me but try for yourself. (If this does not fix on your side, I will remove the updates for #2548 from this PR)

https://github.com/user-attachments/assets/7d4e6712-1730-4318-8c44-ede15924ad7d

taoeffect (approved):
🥳 🙌 

Excellent work @SebinSong!

Issue #2608, opened 2025-02-10
Cannot load the site at all now in one tab


taoeffect:
### Problem

On Brave, w/1.2.5, I can no longer load the site at all.

It started out with me not being able to mark notifications as read #2605:

![Image](https://github.com/user-attachments/assets/e0a6a52a-47dd-43f4-a6bf-6dd5d4597d62)

Then I noticed this odd error in the logs:

![Image](https://github.com/user-attachments/assets/d9d25a11-9935-470e-a0b7-8d867b13e569)

Apparently `date` was a function?!:

![Image](https://github.com/user-attachments/assets/63ee5f81-ddca-4072-bc71-fe9d8777e149)

Then I tried <kbd>command+shift+r</kbd> to force refresh the site, and now it's completely broken, even though I've cleared all the site data and unregistered the SW:

![Image](https://github.com/user-attachments/assets/dcd474ff-3c9c-4da0-bb47-14f807410437)

Note that the other sites (staging, testing, and debug) seem to be working...

This manifest does exist...

```
$ curl "https://groupincome.app/file/z9brRu3VN9mM6vEkuwmLRsbqLSLvtqQNtQhnJswWUFe21vjXXSXD"
{"head":"{\"manifestVersion\":\"1.0.0\"}","body":"{\"name\":\"gi.contracts/identity\",\"version\":\"1.2.5\",\"contract\":{\"hash\":\"z9brRu3VLkz9YyewJVfkfDjaZgZDQG39fNdMaoZVaCD3Ma94cK1m\",\"file\":\"identity.js\"},\"signingKeys\":[\"[\\\"edwards25519sha512batch\\\",\\\"IjAFp6gIzHW2HOIXXS9b4A3EKf8t9NNa5nndHcROiDk=\\\",null]\"],\"contractSlim\":{\"file\":\"identity-slim.js\",\"hash\":\"z9brRu3VMtVDbdvarC491cmf7ZyEN6UpjP8W2KFqx74ZxyBiU9sP\"}}","signature":{"keyId":"z2DrjgbCDg34SaBhFjNEF45AVodCUu7QwcFBksz3BDgN4BekrdN","value":"pCvt8UZnac8l7l8PiktUjQUlGpkXVfjEQZfhjAeRUTNO5KWqSURNeCmSp+CXQZAPnv9SwWCXG4bAwNulWSI4Cw=="}}
```

For some reason it is loading it from disk cache:

[...]

PR #2609, opened 2025-02-11
return no-cache on 404

taoeffect:
Partially addresses #2608.

### DIFF
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -71,6 +71,16 @@ const route = new Proxy({}, {
   }
 })
 
+// helper function that returns 404 and prevents client from caching the 404 response
+// which can sometimes break things: https://github.com/okTurtles/group-income/issues/2608
+function return404nocache (h) {
+  return h.response()
+    .code(404)
+    .header('Cache-Control', 'no-store, no-cache, must-revalidate, proxy-revalidate')
+    .header('Pragma', 'no-cache')
+    .header('Expires', '0')
+}
+
 // RESTful API routes
 
 // TODO: Update this regex once `chel` uses prefixed manifests
@@ -277,7 +287,7 @@ route.GET('/latestHEADinfo/{contractID}', {
     const HEADinfo = await sbp('chelonia/db/latestHEADinfo', contractID)
     if (!HEADinfo) {
       console.warn(`[backend] latestHEADinfo not found for ${contractID}`)
-      return Boom.notFound()
+      return return404nocache(h)
     }
     return HEADinfo
   } catch (err) {
@@ -473,7 +483,7 @@ route.GET('/file/{hash}', {
 
   const blobOrString = await sbp('chelonia/db/get', `any:${hash}`)
   if (!blobOrString) {
-    return Boom.notFound()
+    return return404nocache(h)
   }
   return h.response(blobOrString).etag(hash)
 })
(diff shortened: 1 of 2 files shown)


backend/routes.js
```diff
@@ -71,6 +71,16 @@ const route = new Proxy({}, {
   }
 })
 
+// helper function that returns 404 and prevents client from caching the 404 response
+// which can sometimes break things: https://github.com/okTurtles/group-income/issues/2608
+function notFoundNoCache (h) {
+  return h.response()
+    .code(404)
+    .header('Cache-Control', 'no-store, must-revalidate, proxy-revalidate')
```
corrideat: `*-revalidate` is redundant as well, since `no-store` already means that the response should _not_ be put in a cache to begin with.

corrideat: Also, I think the issue here is this, since by default 404 responses will not be cached.

> Most commonly, caches store the successful result of a retrieval request: i.e., a 200 (OK) response to a GET request, which contains a representation of the target resource (Section 9.3.1 of [HTTP]). **However, it is also possible to store redirects, negative results** (e.g., 404 (Not Found)), incomplete results (e.g., 206 (Partial Content)), and responses to methods other than GET if the method's definition allows such caching and defines something suitable for use as a cache key.

```js
  cache: {
    // Do not set other cache options here, to make sure the 'otherwise' option
    // will be used so that the 'immutable' directive gets included.
    otherwise: 'public,max-age=31536000,immutable'
  }
```

I'm not too familiar with HAPI, and not with the old version used in GI, but, instead of setting a cache policy and then 'undoing it' with this new `notFoundNoCache` wouldn't it be better to simply tell HAPI not to apply the `immutable` (or a custom) cache policy to errors?

backend/routes.js
```diff
@@ -71,6 +71,16 @@ const route = new Proxy({}, {
...
+    .header('Pragma', 'no-cache')
+    .header('Expires', '0')
```
corrideat: `Expires` is supposed to have a timestamp, so `0` isn't a valid value. However, it's not really relevant if `cache-control` is present, except for clients that don't support HTTP/1.1. I don't think any such clients exist (today) to any significant, and even if they did, they can be ignored.

backend/routes.js
```diff
@@ -71,6 +71,16 @@ const route = new Proxy({}, {
   }
 })
 
+// helper function that returns 404 and prevents client from caching the 404 response
+// which can sometimes break things: https://github.com/okTurtles/group-income/issues/2608
+function notFoundNoCache (h) {
+  return h.response()
+    .code(404)
+    .header('Cache-Control', 'no-store, must-revalidate, proxy-revalidate')
+    .header('Pragma', 'no-cache')
```
corrideat: `Pragma: no-cache` has never been standard for responses, and is also largely irrelevant for clients that understand `cache-control`.

corrideat:
I suggest either of these solutions:

1. Set `cache-control: immutable,...` _only_ for successful responses. From the HAPI docs, it seems like removing the `cache: otherwise` option would do this, plus adding the header to the response. This requires no additional changes because 404, etc. isn't supposed to be cached unless it says it should.
2. Leave things mostly are they currently are, but on 404 return only `cache-control: no-store` without `pragma` or `expires`.

corrideat (dismissed):

corrideat (approved):

PR #2610, opened 2025-02-12
#2602 Trigger "Leave channel" modal when user leaves channel

### LINKED ISSUE #2602: Trigger "Leave channel" modal when user leaves channel
### Problem

The user can accidentally remove themselves from a channel by clicking the "`x`" next to their name in the members list. This removes them instantly, and does not trigger the "Leave channel" modal.

Image

### Solution

"Leave channel" modal should also be triggered when the user removes themselves from the member list.

SebinSong:
closes #2602


taoeffect (approved):
Nice!

PR #2613, opened 2025-02-12
Fix #2606

### LINKED ISSUE #2606: Inside OP_ATOMIC: Non-object or null error thrown
### Problem

Version: 1.2.5

Fresh login:

Image

Seems to suggest `GIChatroomAlreadyMemberError` is null and not an object.

### Solution

Fix.

corrideat:


taoeffect (approved):
It's absurd that we both missed this.

PR #2618, opened 2025-02-12
Fix #2617

### LINKED ISSUE #2617: "Cannot read properties of undefined (reading 'strictProcessing')"
### Problem

While logging in fresh to GIG:

Image

### Solution

Think this shouldn't be happening... fix.

corrideat:


taoeffect (changes requested):
It doesn't make sense to me if we're going to ignore the error in the UI, but not ignore it in `OP_ATOMIC`.

If this is an error that should cause everything in `OP_ATOMIC` to revert, and we're reliably producing this error, then:

1. it's a bug and the error should be fixed and prevented from happening in the first place
2. if it does happen, the user should be notified of it, there should be a notification visible in other words (not ignored)

frontend/model/notifications/templates.js
```diff
@@ -40,7 +44,8 @@ export default ({
       errName: error.name,
       activity,
       action: action ?? opType,
-      contract: state.contracts[contractID]?.type ?? contractID
+      contract: state.contracts[contractID]?.type ?? contractID,
+      errMsg: error.message
```
taoeffect, line 48: I saw this was added in a recent commit, but I don't see it used. Did you intend to leave it unused?

corrideat: `errMsg` is indeed used, so this is a bugfix.

taoeffect (approved):
Fantastic work @corrideat !

PR #2622, opened 2025-02-13
#2603 - Character limit for poll option and question

### LINKED ISSUE #2603: Set a reasonable character limit for poll options :-D
### Problem

When creating a poll, the text boxes for options seem to accept an unlimited number of characters. (For reference, I gave up after 128,000 chars.)
Note: This did not seem to cause any issues with the UI; the app displayed the poll without delay, and the text of the long poll option was properly aligned, properly spaced, and wrapped.

### Solution

Limit the input to a reasonable number of characters.
Indicate the user has reached the char limit the same way we do elsewhere

SebinSong:
closes #2603 

Thought the poll question also needs this character limit so worked on adding this to both poll title / options. But pls let me know if it's not required, will remove it.

**[ Screenshot for Poll title ]**

**[ Screenshot for Poll option ]**

As you can see above, this 'Char-Length' UI is only displayed when the subject input element is focused. Please let me know if we want to change this behavior though.

### DIFF
--- a/frontend/model/contracts/shared/constants.js
+++ b/frontend/model/contracts/shared/constants.js
@@ -132,3 +132,4 @@ export const POLL_STATUS = {
 }
 
 export const POLL_MAX_OPTIONS = 20
+export const POLL_OPTION_MAX_CHARS = 10
--- a/frontend/model/contracts/shared/constants.js
+++ b/frontend/model/contracts/shared/constants.js
@@ -132,4 +132,5 @@ export const POLL_STATUS = {
 }
 
 export const POLL_MAX_OPTIONS = 20
-export const POLL_OPTION_MAX_CHARS = 10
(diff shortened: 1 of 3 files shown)


frontend/model/contracts/shared/constants.js
```diff
@@ -132,3 +132,5 @@ export const POLL_STATUS = {
 }
 
 export const POLL_MAX_OPTIONS = 20
+export const POLL_QUESTION_MAX_CHARS = 280
+export const POLL_OPTION_MAX_CHARS = 280
```
SebinSong, line 136: `280` here is a value I chose arbitrary from a existing value below:

```js
export const CHATROOM_DESCRIPTION_LIMITS_IN_CHARS = 280
```

Please let me know if we want to change this value.

taoeffect (approved):
Fantastic work @SebinSong!

PR #2633, opened 2025-02-16
Safeguards for SW being ready before activation and taking SBP calls

### LINKED ISSUE #2588: LastLoggedIn still broken 😮‍💨

### LINKED ISSUE #2616: Logged out after version update
### Problem

After the automatic refresh for the version update:

https://github.com/user-attachments/assets/bcc09064-9d32-471d-91ce-09522441a95f

📎  [groupincome.app-1739386554858.log](https://github.com/user-attachments/files/18772794/groupincome.app-1739386554858.log)

Look at how it's attempting to load `https://groupincome.app/assets/js/null/latestHEADinfo/z9brRu3VNPyi51PueSz6KnCKnZYZR897ueDzb4Sy1wA6RL6sLVXm` for some reason:

Image

### Solution

Fix.

corrideat:
Closes #2616 
Closes #2588 (EDIT by @taoeffect)


frontend/main.js
```diff
@@ -72,7 +72,7 @@ async function startApp () {
   // ?debug=true
   // force debug output even in production
   const debugParam = new URLSearchParams(window.location.search).get('debug')
-  if (process.env.NODE_ENV !== 'production' || debugParam === 'true') {
+  if (process.env.NODE_ENV !== 'production' || debugParam === 'true' || isNaN(NaN)) {
```
corrideat: Revert

frontend/controller/service-worker.js
```diff
@@ -341,6 +373,9 @@ sbp('sbp/selectors/register', {
 sbp('sbp/selectors/register', {
   'gi.notifications/*': swRpc
 })
+sbp('sbp/selectors/register', {
+  'sw/*': swRpc
+})
```
taoeffect, line 378: Is there a reason we're calling `'sbp/selectors/register'` 7 times here?

All of these selectors could be put into a single call to `'sbp/selectors/register'`...

corrideat: I thought the call to `sbp/selectors/register` were supposed to be for a single domain?

taoeffect: I think it can handle multiple domains, but I suppose yeah.. that is the convention... fair enough

corrideat: > I think it can handle multiple domains

Yeah, it can.

frontend/main.js
```diff
@@ -376,27 +363,14 @@ async function startApp () {
         const removeHandler = sbp('okTurtles.events/once', OPEN_MODAL, () => {
           this.removeLoadingAnimation()
         })
+        console.error('@@@WILL CLL LOGIN')
```
taoeffect: Debug logging...

frontend/setupChelonia.js
```diff
@@ -312,6 +313,7 @@ const setupChelonia = async (): Promise<*> => {
       console.error('[setupChelonia] Error syncing identity contract and groups', e)
     })
   })
+  */
```
taoeffect, line 316: This section being commented out... does it make sense to either:

- Leave it uncommented, but run it only when we're sure that we're no longer running in a SW?
- Or, if it's no longer necessary at all (even when not running in a SW), perhaps it should be deleted then?

corrideat: I doesn't seem to be needed now, and while it might be needed if not running in a SW, I'm not entirely sure. I commented it out in case it's needed when that's implemented.

frontend/main.js
```diff
@@ -354,20 +354,7 @@ async function startApp () {
...
-          navigator.serviceWorker?.ready,
-          new Promise((resolve, reject) => setTimeout(() => reject(new Error('SW ready timeout')), 10000))
-        ]).catch(e => {
-          console.error('[app] Service worker failed to become ready:', e)
-          // Fallback behavior
-          this.removeLoadingAnimation()
-          alert(L('Error while setting up service worker: {err}', { err: e.message }))
-          throw e
-        })
-      })()]).then(async ([identityContractID]) => {
```
taoeffect, line 370: Nice simplifications! 😄 🎉 

Was looking really hairy before.

frontend/main.js
```diff
@@ -354,20 +354,7 @@ async function startApp () {
...
+      sbp('gi.db/settings/load', SETTING_CURRENT_USER).then(async (identityContractID) => {
```
taoeffect: What do you think about replacing this whole `.then` thing with something like this instead while we're at it?

```js
;(async function () {
    try {
      const identityContractID = await sbp('gi.db/settings/load', SETTING_CURRENT_USER)
      oldIdentityContractID = identityContractID
      if (!identityContractID || this.ephemeral.finishedLogin === 'yes') return
      // Calling login could result in a prompt in case of an error; if the
      // loading animation is visible, it'll hide the prompt. We remove it,
      // so that it's possible to interact with the prompt.
      const removeHandler = sbp('okTurtles.events/once', OPEN_MODAL, () => {
        this.removeLoadingAnimation()
      })
      await sbp('gi.app/identity/login', { identityContractID })
      removeHandler()
      await sbp('chelonia/contract/wait', identityContractID)
      this.ephemeral.ready = true
      this.removeLoadingAnimation()
      setupNativeNotificationsListeners()
    } catch (e) {
      this.removeLoadingAnimation()
      oldIdentityContractID && sbp('appLogs/clearLogs', oldIdentityContractID).catch(e => {
        console.error('[main] Error clearing logs for old session', oldIdentityContractID, e)
      }) // https://github.com/okTurtles/group-income/issues/2194
      await sbp('gi.app/identity/logout')
      await sbp('gi.ui/prompt', {
        heading: L('Failed to login'),
        question: L('Error details: {reportError}', LError(e)),
        primaryButton: L('Close')
      })
    }
  })()
```

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -114,10 +114,22 @@ const setupRootState = () => {
 
 sbp('okTurtles.events/on', CHELONIA_RESET, setupRootState)
 
+const broadcastMessage = (...args) => {
+  self.clients.matchAll()
+    .then((clientList) => {
+      // eslint-disable-next-line require-await
+      return Promise.all(clientList.map(async (client) => {
+        return client.postMessage(...args)
```
taoeffect, line 122: Why is this an `async` function without an `await`? What's the reason for marking it `async` like this?

(Can you add a comment above line 120 explaining?)

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -135,33 +147,24 @@ sbp('okTurtles.events/on', CHELONIA_RESET, setupRootState)
...
 sbp('okTurtles.events/on', CONTRACT_REGISTERED, (contract) => {
   // Remove function types from contract data
   // This avoids unnecessary MessagePorts
-  const argsCopy = Object.fromEntries(Object.entries(contract).filter(([_, v]) => {
-    return typeof v !== 'function'
-  }))
+  const argsCopy = {
+    manifest: contract.manifest,
+    name: contract.name
+  }
```
taoeffect, line 160: This code got radically transformed... why?

And if it's safe to do it this way, then the comment on lines 155-156 no longer applies and should probably be either updated or removed.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -192,12 +190,7 @@ sbp('okTurtles.events/on', CAPTURED_LOGS, (...args) => {
     type: CAPTURED_LOGS,
     data
   }
-  self.clients.matchAll()
-    .then((clientList) => {
-      clientList.forEach((client) => {
-        client.postMessage(message, transferables)
-      })
-    })
+  broadcastMessage(message, transferables)
```
taoeffect, line 193: Nice DRYing!

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -114,10 +114,22 @@ const setupRootState = () => {
...
+      })).catch(e => {
+        console.error('[broadcastMessage] Error', args, e)
+      })
+    })
+}
+
 // These are all of the events that will be forwarded to all open tabs and windows
 ;[
 
-  CHELONIA_RESET, CONTRACTS_MODIFIED, CONTRACT_IS_SYNCING, CONTRACT_REGISTERED,
```
taoeffect, line 120: Oof, nice catch! Surprised we both missed this last time.

frontend/controller/serviceworkers/sw-primary.js
corrideat: Not really sure why that lint rule is there to begin with, as there are many reasons to do so. In this case, it's for `postMessage` to be run even if there's an error.

See `[1,2,3].map(async (v) => {throw v})` (1, 2, 3 are thrown) vs `[1,2,3].map((v) => {throw v})` (1 is thrown, then it stops).

frontend/controller/serviceworkers/sw-primary.js
corrideat: The comment on lines 155-156 very much applies and motivates this change. The previous code was (1) too complex and (2) wrong, because the functions are at the second level and not at the top level (`{a:{b:()=>{}}}` instead of `{a:()=>{}}`)

frontend/controller/serviceworkers/sw-primary.js
taoeffect: > See `[1,2,3].map(async (v) => {throw v})` (1, 2, 3 are thrown) vs `[1,2,3].map(async (v) => {throw v})` (1 is thrown, then it stops).

Did you mean something else for the second example? Those two code blocks are the same.

frontend/main.js
```diff
@@ -354,62 +354,37 @@ async function startApp () {
...
-        })
-        await sbp('gi.app/identity/login', { identityContractID })
-        removeHandler()
-        await sbp('chelonia/contract/wait', identityContractID)
-      }).then(() => {
-        // We know that `navigator.serviceWorker` is defined, since we've used
-        // it. This may need to be checked if a non-SW version is also supported.
-        const sw = ((navigator.serviceWorker: any): ServiceWorkerContainer)
-        const onready = () => {
+      ;(async function () {
```
taoeffect: Oops, I made a mistake here. This should be:

```js
;(async () => {
```

frontend/controller/service-worker.js
```diff
@@ -30,8 +30,53 @@ window.addEventListener('beforeinstallprompt', e => {
...
+      messageChannel.port1.close()
+    }
+    messageChannel.port1.onmessageerror = () => {
+      reject(new Error('Message error'))
+      messageChannel.port1.close()
+    }
+    const oncontrollerchange = () => {
+      navigator.serviceWorker.removeEventListener('controllerchange', oncontrollerchange, false)
+      reject(new Error('Service worker controller changed'))
+      waitUntilSwReady().catch(e => {
```
taoeffect: Could you add a comment above line 49 explaining what's going on here? E.g. why we're rejecting the promise and then calling `waitUntilSwReady` again? (i.e. for people like me who are unfamiliar with what `controllerchange` is and how this works)

corrideat: Simplified this a bit

frontend/controller/service-worker.js
```diff
@@ -306,15 +326,25 @@ const swRpc = (() => {
         reject(event.data)
         cleanup()
       }
+      const oncontrollerchange = (event: Event) => {
+        if (maxControllerChanges > 0) {
+          resolve(fn(maxControllerChanges - 1, ...args))
```
taoeffect, line 331: Does it make sense to call `resolve` here immediately like this?

Or should we first `await` on the return value of `fn` to make sure `waitUntilSwReady()` finishes, and then call `resolve` with the returned promise?

frontend/controller/service-worker.js
```diff
@@ -30,8 +30,53 @@ window.addEventListener('beforeinstallprompt', e => {
...
+    messageChannel.port1.onmessageerror = () => {
+      reject(new Error('Message error'))
+      messageChannel.port1.close()
+    }
+    const oncontrollerchange = () => {
+      navigator.serviceWorker.removeEventListener('controllerchange', oncontrollerchange, false)
+      // If the SW controller has changed (e.g., because of an update), we call
+      // `waitUntilSwReady` again and return the result of that call. That will
+      // also populate `serviceWorkerMap`
+      resolve(waitUntilSwReady())
```
taoeffect, line 54: Would it make more sense to do something like:

```js
waitUntilSwReady().then(resolve, reject)
```

?

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -289,6 +282,18 @@ sbp('sbp/selectors/register', {
...
+  'sw/version': () => {
+    return {
+      GI_VERSION: process.env.GI_VERSION,
+      GI_GIT_VERSION: process.env.GI_GIT_VERSION,
+      CONTRACTS_VERSION: process.env.CONTRACTS_VERSION,
+      LIGHTWEIGHT_CLIENT: process.env.LIGHTWEIGHT_CLIENT,
+      NODE_ENV: process.env.NODE_ENV
+    }
+  }
+})
```
taoeffect, line 295: Can you use this returned `GI_VERSION` value in the `TabWrapper.vue` to say:

```
Service worker: {swVer}
```

taoeffect (approved):
Fantastic work @corrideat !!!!

PR #2634, opened 2025-02-16
Handle subscription errors.

### LINKED ISSUE #2612: Registration failed - Push service error
### Problem

Image

A user was doing something (I think they were creating a group maybe on Android), and they got this error and the logs below:

📎 [gi_logs.json.txt.zip](https://github.com/user-attachments/files/18771257/gi_logs.json.txt.zip)

### Solution

1. The error message is useless and unhelpful, improve it both in the UI and in the logs with more details.
2. Then figure out why this happened and prevent it from happening again (if possible).
3. If it happens anyway, make sure that the app attempts to re-register

corrideat:
Closes #2612.

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -198,8 +198,13 @@ sbp('sbp/selectors/register', {
           return subscription
         }).catch(e => {
           if (!(retryCount > 3) && e?.message === 'WebSocket connection is not open') {
-            return new Promise((resolve) => {
-              sbp('okTurtles.events/once', ONLINE, () => {
+            return new Promise((resolve, reject) => {
+              const errorTimeoutId = setTimeout(() => {
+                reject(new Error('Timed out waiting to come back online'))
+                cancel()
+              }, 600e3)
+              const cancel = sbp('okTurtles.events/once', ONLINE, () => {
+                clearTimeout(errorTimeoutId)
                 setTimeout(() => resolve(sbp('service-worker/setup-push-subscription', (retryCount || 0) + 1)), 200)
               })
             })
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -1,5 +1,6 @@
 'use strict'
 
+import { L } from '@common/common.js'
 import sbp from '@sbp/sbp'
 import { CAPTURED_LOGS, LOGIN_COMPLETE, NEW_CHATROOM_UNREAD_POSITION, PWA_INSTALLABLE, SET_APP_LOGS_FILTER } from '@utils/events.js'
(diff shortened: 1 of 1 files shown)


frontend/controller/service-worker.js
```diff
@@ -194,13 +198,24 @@ sbp('sbp/selectors/register', {
           return subscription
         }).catch(e => {
           if (!(retryCount > 3) && e?.message === 'WebSocket connection is not open') {
-            sbp('okTurtles.events/once', ONLINE, () => {
-              setTimeout(() => sbp('service-worker/setup-push-subscription', (retryCount || 0) + 1), 200)
+            return new Promise((resolve) => {
+              sbp('okTurtles.events/once', ONLINE, () => {
+                setTimeout(() => resolve(sbp('service-worker/setup-push-subscription', (retryCount || 0) + 1)), 200)
+              })
```
taoeffect: The way this is written can cause the entire call to `'service-worker/setup-push-subscription'` to hang potentially indefinitely if this event never gets fired.

Can you add a timeout condition to call `reject` on the promise?

taoeffect (changes requested):
Nice work @corrideat! One change request to avoid a "stuck" app that just spins forever.

frontend/controller/service-worker.js
corrideat: I've added it as per your request, but the existing behaviour won't result in a ' "stuck" app that just spins forever', since no UI elements depend on this promise.

frontend/controller/service-worker.js
```diff
@@ -194,13 +198,29 @@ sbp('sbp/selectors/register', {
           return subscription
         }).catch(e => {
           if (!(retryCount > 3) && e?.message === 'WebSocket connection is not open') {
-            sbp('okTurtles.events/once', ONLINE, () => {
-              setTimeout(() => sbp('service-worker/setup-push-subscription', (retryCount || 0) + 1), 200)
+            return new Promise((resolve, reject) => {
+              const errorTimeoutId = setTimeout(() => {
+                reject(new Error('Timed out waiting to come back online'))
+                cancel()
+              }, 600e3)
```
taoeffect, line 205: I don't think users are going to wait 10 minutes to use our app.

Could you please reduce this to 7 seconds?

taoeffect (changes requested):
One last change request!

frontend/controller/service-worker.js
taoeffect: Ah, I see................. maybe that fact should be a comment then.

taoeffect (approved):
Let's hope this is sufficient 👍

PR #2635, opened 2025-02-16
Additional info

corrideat:
Add additional information to try to diagnose why #2588 happens.


frontend/controller/actions/group-kv.js
```diff
@@ -30,14 +30,24 @@ export default (sbp('sbp/selectors/register', {
...
+        return kvData.data
+      }
+      return {
+        _creator: identityContractID,
+        _reason: 'empty'
+      }
+    }
+    return {
+      _creator: identityContractID,
+      _reason: 'new'
```
taoeffect, line 49: You're sure this won't break the other KV stuff we have?

corrideat: It shouldn't, it's just adding new keys to it

taoeffect (approved):
Approved!

PR #2643, opened 2025-02-17
#2628 - Issues with `ChatMembersAllModal.vue`

### LINKED ISSUE #2628: Two `ChatMembersAllModal` issues for a public channel when user is not joinned.
There are currently two issues regarding `ChatMembersAllModal.vue` in a public-channel for a user who is not joined the channel yet.

### Problem 1.

In desktop screen, the user has access to the members-all modal like below:

But this same menu is hidden in non-desktop size viewport:

### Problem 2.

In `ChatMembersAllModal.vue`, the users who are the members of the chatroom are placed under `OTHERS (n)` section and it says `CHANNEL MEMBERS (0)` which is wrong information.

### Solution
- Problem 1: Either hide the 'members' button in both desktop/mobile or expose it in both device screens.
- Problem 2: Place the joinned members under `CHANNEL MEMBERS (n)` section.

SebinSong:
closes #2628 

**[Fix 1 - Arrow button is visible now and can access 'members' menu from it]**

not-joined-public-1

**[Fix 2 - Joinned members are placed in the correct place]**

not-joined-public-2


taoeffect (changes requested):
Great work @SebinSong!

I'm wondering if we can take this opportunity to hide the other members section if we are not joined to the chatroom?
 
Screenshot 2025-02-17 at 12 39 30 PM

To me it doesn't make much sense to show who's not in the channel if we're not in it. The only reason to show that section would be to add members to the channel if we are a part of it, but since we aren't, can we deconfuse and declutter the UI in this situation?

SebinSong:
@taoeffect 
> I'm wondering if we can take this opportunity to hide the other members section if we are not joined to the chatroom?

made that update:

taoeffect (approved):
Fantastic work @SebinSong!

Issue #2647, opened 2025-02-18
DMs not coming through


taoeffect:
### Problem

I asked @SebinSong to DM me after the 1.2.9 release, and he sent me a message but I got no push notification and not even a red number next to his name:

![Image](https://github.com/user-attachments/assets/7c15e654-3f02-4e42-b059-a6f9bcd506df)

The logs showed no errors:

![Image](https://github.com/user-attachments/assets/2f0381a0-3c92-4906-8aec-5b24dd3eca97)

The server shows:

![Image](https://github.com/user-attachments/assets/d6c94b66-94c6-4434-91a6-c0edeb0710e6)
![Image](https://github.com/user-attachments/assets/b4740ef7-fe8d-4ea3-893d-aa4af976c9f3)

Only after loading the DM with him did I see that there were new messages. Furthermore, it seems like these new messages are not getting cached because each time I switch to the DM with him I first see the old messages, and after a very brief delay do the new ones appear (EDIT: I'm not sure about this though, they seem to be cached now).

📎 See video: https://groupincome.app/app/group-chat/z9brRu3VHeajHf3iZMPsqsrL4KoeFhHAsFYuTSkxqotuNRv7oxwL?mhash=z9brRu3VFRogBWGyGuMFE2xL8vgitE49XZUPAuCTp2bxsb6UwQi9

### Solution

Figure out the problem and fix.

taoeffect:
Closed in #2649

PR #2649, opened 2025-02-18
misc push-related fixes and improvements

taoeffect:
Related to issues #2647 and #2653


frontend/controller/service-worker.js
```diff
@@ -204,20 +206,24 @@ sbp('sbp/selectors/register', {
...
-            )
+            console.info('[service-worker/setup-push-subscription] attempting to create a new subscription', subscription)
             return sbp('push/getSubscriptionOptions').then(function (options) {
               return registration.pushManager.subscribe(options)
             })
+          } else {
+            const subClone = cloneDeep(subscription)
+            if (subscription?.endpoint) {
+              // hide the full endpoint from the logs for privacy
+              subClone.endpoint = new URL(subClone.endpoint).host
```
corrideat: Not sure if this helps much, as the host will likely be the same for all browsers of the same vendor.

taoeffect: It leaves out their personal push endpoint

frontend/controller/serviceworkers/push.js
```diff
@@ -13,7 +13,7 @@ export default (sbp('sbp/selectors/register', {
         cachedVapidInformation &&
         // Cache the VAPID information for one hour. The server public
         // information should change very infrequently, if it changes at all.
-        (performance.now() - cachedVapidInformation[0]) < 3600e3
```
corrideat: Why did you change `performance` to `Date`? It seems wrong.

taoeffect: Why does it seem wrong? I see no reason for using `performance` ... changed it just in case `performance.now()` gets reset to `0` because of SW lifecycle stuff... can't happen with `Date.now()`, so changed it to be safe.

corrideat: But it's wrong now, because `Date` isn't monotonic.

> `performance.now()` gets reset to 0

That's not an issue, as these values aren't persistent. The entire thing will get reset 'because of SW lifecycle stuff'

taoeffect: > But it's wrong now, because Date isn't monotonic.

That doesn't matter because either way the condition will hold and the desired result will be achieved. The only place this would matter is if we were measuring something very small, but because it's comparing against something very large (`3600e3`) it doesn't matter.

corrideat: As I see it, there are ever two reasons to use `Date`, otherwise `performance` should be used:

1. A real-world 'date' is needed
2. A real-world 'date' isn't strictly needed, but a global timestamp is needed for synchronisation reasons that most devices can more or less agree on

taoeffect: > The entire thing will get reset 'because of SW lifecycle stuff'

If this statement isn't true due to some browser quirk then the difference between `0` and the value stored will be a big deal.

Whereas using `Date.now()` guarantees correct behavior.

corrideat: > That doesn't matter because either way the condition will hold and the desired result will be achieved. The only place this would matter is if we were measuring something very small, but because it's comparing against something very large (3600e3) it doesn't matter.

You're just making the code wrong where it wasn't for no good reason, since synchronisation isn't needed as the cache is local-only.

> The only place this would matter is if we were measuring something very small

The local clock can jump by arbitrary amounts, and 1 hour isn't so large in the grand scheme of things.

>  The only place this would matter is if we were measuring something very small

The only reason we need to start speculating on if it matters or not is because `Date` makes it wrong, which `performance`, as it was, works without any of these issues.

taoeffect: Alright, if you are confident that it won't get reset to `0` feel free to revert it

corrideat: It _will_ get reset to 0. I'm just saying it doesn't matter, because the cache isn't persisted, and will _also_ get reset to 0.

taoeffect: Yeah I mean that it won't get reset to `0` while the cache is still using the old value.

frontend/model/settings/vuexModule.js
```diff
@@ -70,14 +70,16 @@ const mutations = {
     state.increasedContrast = isChecked
   },
   setNotificationEnabled (state, enabled) {
-    if (state.notificationEnabled !== enabled) {
+    if (state.notificationEnabled !== enabled && enabled) {
```
taoeffect: Note: I might be both right and wrong about adding `&& enabled` here... investigating

- Right because it avoids infinite loop
- Wrong because it may be necessary to call `'service-worker/setup-push-subscription'` in both cases

frontend/controller/service-worker.js
corrideat: No, I get the privacy angle (I don't think there's something _stopping_ a push hostname to contain sensitive information, but in practice they don't). What I'm saying is that it makes logging it almost worthless, as you can predict with high confidence what it'll be based on the user agent.

frontend/views/containers/access/LoginForm.vue
```diff
@@ -93,12 +97,12 @@ export default ({
         })
         await this.postSubmit()
         this.$emit('login-status', 'success')
-
-        requestNotificationPermission().catch(e => console.error('[SignupForm.vue] Error requesting notification permission', e))
```
corrideat, line 97: This is here for a reason. In most browsers, the notification permission needs to be requested as a direct result of user interaction. Your change will make it more likely for the request to fail.

frontend/model/settings/vuexModule.js
```diff
@@ -70,19 +70,20 @@ const mutations = {
     state.increasedContrast = isChecked
   },
   setNotificationEnabled (state, enabled) {
-    if (state.notificationEnabled !== enabled && enabled) {
```
corrideat: By removing this check, as it originally was, `state.notificationEnabled !== enabled`, you're potentially introducing infinite loops, and also triggering logic that's unnecessary. If `notificationsEnabled` is true, it means that we're in the process of establishing a push subscription or have already got one. Calling `service-worker/setup-push-subscription` in either of these cases won't do anything (of value).

frontend/main.js
```diff
@@ -271,9 +264,14 @@ async function startApp () {
           this.initOrResetPeriodicNotifications()
           this.checkAndEmitOneTimeNotifications()
         }
-
         // NOTE: should set IdleVue plugin here because state could be replaced while logging in
         Vue.use(IdleVue, { store, idleTime: 2 * 60 * 1000 }) // 2 mins of idle config
+
+        // NOTE: the SignupForm and LoginForm call requestNotificationPermission() upon LOGIN_COMPLETE,
+        // which also results in a call to 'service-worker/setup-push-subscription'.
```
corrideat: I've said in a different comment why `LOGIN_COMPLETE` can't be used for this, but to reiterate it here: the notification permission needs to be requested from direct user interaction. If it's not, or if it takes too long between the user interaction and the permission request, the browser will make it fail.

frontend/controller/serviceworkers/push.js
```diff
@@ -80,13 +80,15 @@ export default (sbp('sbp/selectors/register', {
...
       map.set(socket, subscriptionInfo)
-      if (subscriptionInfo?.endpoint && reported !== subscriptionInfo.endpoint) {
-        // If the subscription has changed, report it to the server
-        pubsub.socket.send(createMessage(
-          REQUEST_TYPE.PUSH_ACTION,
-          { action: PUSH_SERVER_ACTION_TYPE.STORE_SUBSCRIPTION, payload: subscriptionInfo }
-        ))
-      } else if (!subscriptionInfo && reported) {
+      if (subscriptionInfo?.endpoint) {
+        if (!reported || subscriptionInfo.endpoint !== reported.endpoint) {
```
corrideat, line 93: It looks like you've caught a bug here (`reported.endpoint`). I think that still the nested `if` can be avoided by using, e.g., `subscriptionInfo?.endpoint && reported?.endpoint !== subscriptionInfo.endpoint`, or by replacing the line above by `map.set(socket, subscriptionInfo?.endpoint)`

frontend/controller/service-worker.js
```diff
@@ -199,61 +198,51 @@ sbp('sbp/selectors/register', {
...
+          subscription = await registration.pushManager.getSubscription()
+          let newSub = false
+          if (!subscription || (subscription.expirationTime != null && subscription.expirationTime <= Date.now())) {
+            subscription = await registration.pushManager.subscribe(await sbp('push/getSubscriptionOptions'))
+            newSub = true
+          }
+          subPrivVer = cloneDeep(subscription)
+          if (subscription?.endpoint) {
+            // hide the full endpoint from the logs for privacy
+            subPrivVer.endpoint = new URL(subPrivVer.endpoint).host
```
corrideat: There are other sensitive things in the subscription info besides the endpoint. Probably the entire `subPrivVer` thing needs to go.

frontend/controller/service-worker.js
```diff
@@ -199,61 +198,51 @@ sbp('sbp/selectors/register', {
...
-            const subClone = cloneDeep(subscription)
-            if (subscription?.endpoint) {
-              // hide the full endpoint from the logs for privacy
-              subClone.endpoint = new URL(subClone.endpoint).host
-            }
-            console.info('[service-worker/setup-push-subscription] got existing subscription:', subClone)
+      // Safari sometimes incorrectly reports 'prompt' when using `registration.pushManager.permissionState`
+      let subscription = null
+      let subPrivVer = subscription
+      if (notificationEnabled && granted) {
```
corrideat: Why add this check (`notificationEnabled`). That depends on the result of this function, so it seems like you're adding the risk that calling to set up the subscription will fail.

frontend/controller/service-worker.js
corrideat: Not exactly sure what these changes here are doing, other than some refactoring (I prefer things mostly like they were, because it was written to use immutable (`const`) variables, which is IMO easier to reason about) and removing the retry logic. In the case of removing the retry logic, I'm not entirely sure why you want to do that.

corrideat:
I've added comments where appropriate, mostly explaining why certain things are the way they were. I'm not too happy (meaning feeling neutral to negative) with the changes made to the frontend management logic (other than the extra logging and the bugfix for `reported`, using `reported.endpoint`, which I feel positive about), but it does look like the Vue components had some improvements.

However, I don't think these changes will help much with the issue at hand, namely notifications stopping working after an update. This is why:

1. On Slack, we've discussed these various 401 errors. I think getting to the bottom of this is what will fix the issue, or at least, it's a necessary condition.
2. You seem to be making an effort at increasing the reliability of reporting the subscription to the server. Without focusing on whether the changes accomplish this or not, doing so will likely have little effect (*) on the issue at hand. When a subscription is first reported, the server stores it, and, after that, subsequent failures to report it again (e.g., new WS connections) would result in unnecessary push events being sent by the server, rather than fewer. This is because reporting a subscription actually causes the server to _stop_ sending push events until that WS connection is closed. So, even if the reporting reliability is improved, the issue of notifications not being seen more than likely lies on the server.

(*) There's one exception to this, but it still requires an error on the server. That is, if the logic for storing subscriptions on the DB is faulty, and some or all subscriptions are lost.

frontend/controller/service-worker.js
taoeffect: Because if the user has disabled notifications we don't want push notifications.

frontend/controller/serviceworkers/push.js
taoeffect: I prefer the readability of this nested if, it's easier for me to follow what's going on as it reads more like English.

"If we have an endpoint, and either we did not have one before, or it's different, then report it".

Versus,

"If we have an endpoint and the one we had before or null is different from the one passed in, then report it."

It also then works better with the `else` clause below, as that can be read as:

"If we have an endpoint, or else if we have one from before"

Versus:

"If we have an endpoint and the one we had before or null is different from the one passed in, or else if we have one from before"

frontend/model/settings/vuexModule.js
taoeffect: It should not result in an infinite loop, that's why there's a condition `if (enabled)` around `setTimeout(() => sbp('state/vuex/commit', 'setNotificationEnabled', false), 100)`.

frontend/controller/service-worker.js
taoeffect: > In the case of removing the retry logic, I'm not entirely sure why you want to do that

Because at this point in time there shouldn't be a failure (because of the other refactoring changes make sure that `'service-worker/setup-push-subscription'` gets called after `LOGIN_COMPLETE`), and if there is now the `gi.ui/prompt` will be shown in more situations.

corrideat: >  there shouldn't be a failure

But there can be, and retrying seems like the reasonable thing to do here.

taoeffect: Retrying will automatically happen any time the user refreshes the page or goes offline and back online, or toggles the notifications switch.

frontend/controller/service-worker.js
corrideat: > Because if the user has disabled notifications we don't want push notifications.

I might need to review your changes to the UI, but I think this is something we'd discussed earlier, which motivated the original implementation. Currently, the 'real' permission is the browser permission, and the Vuex preference is meant as a convenience that eventually converges to the browser value. This is convenient because if you, say, grant notification permission, it will soon mean you'll start receiving notifications and that if you revoke the permission, you'll stop receiving notifications.

Note that the permission can't be revoked from the app, only from the UI.

If you want to support the use case 'notifications not wanted' but 'notification permission granted', you need to track both values separately, which complicates things for arguably little gain. This is further complicated by the fact that when the SW wakes up, it needs to check this preference before sending the subscription info to the server.

taoeffect: > If you want to support the use case 'notifications not wanted' but 'notification permission granted', you need to track both values separately, which complicates things for arguably little gain.

This PR already does that and it's fairly trivial to do.. the gain is that the user can disable push notifications from within the app if they want, and that the app will respect their preferences whether they've disabled notification in-app or in-browser, and the toggle respects both settings.

> This is further complicated by the fact that when the SW wakes up, it needs to check this preference before sending the subscription info to the server.

I don't think this is true. The SW doesn't need to do anything, it's already done what it needs to do when `'service-worker/setup-push-subscription'` is called.

frontend/model/settings/vuexModule.js
```diff
@@ -70,17 +70,20 @@ const mutations = {
...
+    // We do this call to `service-worker` here to avoid DRY violations.
+    // The intent is creating a subscription if none exists and letting the
+    // server know of the subscription.
+    // The parent `if` branch should prevent infinite loops
+    sbp('service-worker/setup-push-subscription').catch(e => {
+      console.error('[setNotificationEnabled] Error calling service-worker/setup-push-subscription', e)
+      // if we attempted to turn it on and failed, then turn it off (so the user sees that the toggle
+      // switch is off and attempts to turn it back on again)
+      if (enabled) {
+        setTimeout(() => sbp('state/vuex/commit', 'setNotificationEnabled', false), 100)
```
corrideat: But:

1. Why do we need a timeout at all?
2. If the notification is not changing, you're calling `service-worker/setup-push-subscription` unnecessarily

taoeffect: 1. So that `setNotificationEnabled` is called again after this invocation and is set to `false` after it has been set to `true` by this invocation. It's not a good idea to call a Vuex commit thing on itself inside of the commit that's being committed, so a `setTimeout` is used to let the current commit finish.
2. No it's necessary as this ensures that for example, if `state.notificationEnabled` is `true` because it was loaded from the Vuex state, that other calls to this in the code to set it to true (e.g. by `setupNativeNotificationsListeners()`) still trigger a call to `'service-worker/setup-push-subscription'`. I can add a comment though to explain this.

corrideat: If you switch the toggle in the notification settings, you'll call `state.notificationEnabled` twice. I think know what you're trying to do, but as mentioned in my other comment, you need have another check to track the preference separately.

Also re. 1, the 100 value isn't needed, it could be 0, or you could use `queueMicrotask`.

frontend/controller/service-worker.js
corrideat: It is true:

```
if (self.registration?.pushManager) {
  (() => {
    let inProgress = false
    sbp('okTurtles.events/on', PUBSUB_RECONNECTION_SUCCEEDED, () => {
      if (inProgress) return
      inProgress = true
      self.registration.pushManager.getSubscription().then((subscription) =>
        sbp('push/reportExistingSubscription', subscription?.toJSON())
      ).catch((e) => {
        console.error('Error reporting subscription on reconnection', e)
      }).finally(() => {
        inProgress = false
      })
    })
  })()
}
```

You need to check the user setting here too.

backend/push.js
```diff
@@ -242,21 +198,19 @@ export const postEvent = async (subscription: Object, event: ?string): Promise<v
...
     if ([401, 404, 410].includes(req.status)) {
-      console.warn(
-        new Date().toISOString(),
-        'Removing subscription',
-        subscription.id
-      )
-      deleteClient(subscription.id)
-      return
+      console.warn(new Date().toISOString(), 'Removing subscription', subscription.id)
+      await removeSubscription(subscription.id)
```
corrideat: By adding these various `await` statements, you're slowing things down, such as broadcasting messages.

taoeffect: I can remove this one, sure. The `await`s are important though, as without them the code is difficult to reason about and that leads to bugs.

taoeffect:
@corrideat I've addressed all remaining feedback, please check latest changes!

Note: I wasn't able to join groups when creating a link in FF and then joining in Safari.

Also in both Safari and Brave I noticed this upon creating a group:

Screenshot 2025-02-20 at 1 38 36 PM

However upon dismissing I saw that I was joined.

I don't think these issues have anything to do with this PR but might be related to some of that commented out code (#2653) or other issues. Opened an issue for this: #2654

backend/push.js
```diff
@@ -35,72 +36,47 @@ const saveSubscription = (server, subscriptionId) => {
...
+      }
+    } else {
+      // this can happen for example when a new subscription is added but then
+      // immediately removed because postEvent got a 401 Unauthorized when adding
+      // the new subscription. In this case removeSubscription could be later called
+      // again by the client but it's already been removed
+      console.warn(`removeSubscription: non-existent subscription '${subscriptionId}'`)
+    }
+    await deleteSubscriptionFromIndex(subscriptionId)
+    await sbp('chelonia.db/delete', `_private_webpush_${subscriptionId}`)
```
corrideat: I realise this was like this from before, but I'm wondering if maybe the order of these two operations should be reversed

backend/push.js
```diff
@@ -219,6 +174,7 @@ export const postEvent = async (subscription: Object, event: ?string): Promise<v
   // Note: web push notifications can be 'bodyless' or they can contain a body
   // If there's no body, there isn't anything to encrypt, so we skip both the
   // encryption and the encryption headers.
+  // console.debug(`postEvent to ${subscription.id}:`, event)
```
corrideat, line 177: Stray debug?

backend/push.js
taoeffect: I don't know, if there's a strong reason for it that can be given I'm certainly happy to do it.

backend/push.js
taoeffect: I think it's OK to leave here just in case someone is debugging this part of the code. I found it useful and that's why I decided to comment it out instead of removing it completely. If you really want me to remove it I can though.

backend/push.js
corrideat: The reason is that if the deletion fails, then it'll still be in the index.

taoeffect: How will reversing the order fix that? It first deletes from the index right now. If there's a problem with writing to the DB, it's the index deletion that will fail first, and then it won't get to deleting the subscriptionID.

Whereas, if it successfully deletes the subscription key, but then fails to delete it from the index, it will be in the index but won't exist.

frontend/controller/actions/utils.js
```diff
@@ -357,15 +357,21 @@ export async function syncContractsInOrder (groupedContractIDs: Object): Promise
...
-    }))
+    })
+    for (const [, contractIDs] of sortedContractTypes) {
+      // For each contract of this type, check if it still exists before syncing because
+      // e.g. syncing a group contract could have removed one of the chatroom contracts
+      // $FlowFixMe[incompatible-type]
+      for (const contractID of contractIDs) {
+        const { contracts } = sbp('chelonia/rootState')
+        if (contractID in contracts) {
+          await sbp('chelonia/contract/sync', contractID)
```
corrideat: Apart from 'check if it still exists before syncing because' (which IMO is fine), the code isn't equivalent (this is caused by removing `Promise.all`). With `Promise.all`, an exception will be thrown if there's an error, but processing will otherwise continue. With these changes, processing will halt upon the first error. I think this is a mistake, because we want to sync as many things as we possibly can.

Now, an issue is that `Promise.all` will return *before* all promises complete (if there was an error). Using, e.g., `Promise.allSettled` or a try-catch here would do.

frontend/controller/service-worker.js
```diff
@@ -183,72 +183,83 @@ sbp('sbp/selectors/register', {
   // In theory, the `PushManager` APIs used are available in the SW and we could
   // have this function there. However, most examples perform this outside of the
   // SW, and private testing showed that it's more reliable doing it here.
-  'service-worker/setup-push-subscription': async function (retryCount?: number) {
-    if (process.env.CI) {
+  'service-worker/setup-push-subscription': async function (attemptNumber = 1) {
+    if (window.Cypress || process.env.CI) {
```
corrideat: Not a strong feeling either way, but the reason I was _not_ checking for `window.Cypress` (and generally avoid doing so) is that these are runtime checks (code shipped) while `process.env.CI` is exclusively a buildtime check (no code shipped).

frontend/controller/actions/utils.js
taoeffect: Changes pushed!

frontend/controller/service-worker.js
taoeffect: Had to add it because without it the tests weren't passing. An alternative would be to create a Cypress envvar in `Gruntfile.js`...

frontend/model/notifications/nativeNotification.js
```diff
@@ -35,9 +45,13 @@ export const setupNativeNotificationsListeners = () => {
...
     typeof ServiceWorker !== 'function' ||
     typeof navigator.serviceWorker !== 'object'
-  ) return
+  ) {
+    console.warn("Notifications aren't available in this browser!")
+    return
+  }
 
   if (
+    !navigator.vendor === 'Apple Computer, Inc.' && // WebKit doesn't work
```
corrideat: 1. This check is always false, causing the `if` block to be ignored. The bug is that it should be `navigator.vendor !== 'Apple Computer, Inc.'` (or alternatively, `!(navigator.vendor === 'Apple Computer, Inc.')`).
2. Not a fan of UA sniffing like this, but in this scenario it might be the only option. However, have you been able to check if this `onchange` test works on macOS using different browsers? Apple cited OS limitations for this not working.

frontend/model/settings/vuexModule.js
```diff
@@ -70,17 +70,25 @@ const mutations = {
...
-        sbp('state/vuex/commit', 'setNotificationEnabled', false)
-        console.error('[setNotificationEnabled] Error calling service-worker/setup-push-subscription', e)
-      })
-    }
+    console.info('[setNotificationEnabled] set to:', enabled)
     state.notificationEnabled = enabled
+    // if necessary, prevents the service working from ignoring our notificationEnabled
+    // setting (which is not stored in the SW because it's Vuex-only) from requesting push
+    // notifications upon reconnecting to server
+    sbp('chelonia/configure', { disableNotifications: !enabled }).catch(e => {
```
corrideat: 1. I don't think this belong in Chelonia configuration, especially since Chelonia has nothing to do with handling these settings.
2. This isn't a persisted setting, so this will not work. This needs to be a persisted setting, so probably it needs to be saved in the identity contract preferences.

frontend/model/notifications/nativeNotification.js
taoeffect: Thanks, pushed a fix. And yes I tested 3 browsers on macOS (Brave, FF, and Safari). Both Brave and FF support the change listeners (although FF only notifies of changes to `'notifications'`).

frontend/model/settings/vuexModule.js
taoeffect: @corrideat I don't see why it matters that it's not persistent. It's set on startup. Are you saying the service worker loses all values in memory as it goes through its lifecycle? If so, that is a problem, but also tying Chelonia to a specific contract doesn't sound ideal either.

Either way, we can tackle this in a separate issue, not in this PR. Can you open that issue?

corrideat: Yes, the SW loses all values in memory.

> If so, that is a problem, but also tying Chelonia to a specific contract doesn't sound ideal either.

I don't think so, because none of this is done by Chelonia.

> we can tackle this in a separate issue, not in this PR

It's this PR that breaks it (to expand on this, this is because of your attempt at separating the in-app and browser notification preferences)

taoeffect: > I don't think so, because none of this is done by Chelonia.

Fair enough.

> It's this PR that breaks it

What breaks what?

corrideat: Users can 'opt out' of receiving notifications w/o resetting permissions, but this will not be honoured and will leave the app in an inconsistent state. I'm about to test exactly what this means for the app as a whole.

corrideat: I've now tested this and it's broken in this way:

1. Grant permission
2. Notifications are sent normally
3. Disable notifications using the app (permission intact)
4. Notifications stop
5. SW wakes up for some reason (which doesn't send the `LOGIN_COMPLETE`, i.e., without a browser client, such as in response to a periodic event, user manually waking it up, etc.)
6. Notifications start being sent again
7. Open app
8. Notification setting shows notifications off
9. Notifications are still being sent

backend/push.js
corrideat: It doesn't 'fix' it, but it being in the index but not existing is worse than the converse, based on the assumption that keys may not be enumerable.

corrideat:
I've been reviewing this and testing it, and it seems like most things work. I didn't yet test  this as comprehensively as the previous code (esp. re. the frontend parts), but it looks like it should work.

Can you as a last change request:

1. Fix the order for deleting subscriptions before updating the index
2. Revert the change to `shared/domains/chelonia/chelonia.js`
3. Save this setting to IndexedDB in `setNotificationEnabled` and use it in the relevant places. Actually, the Vuex setting may not be needed at all if using IndexedDB.

corrideat (approved):

PR #2655, opened 2025-02-21
#2651 - Fix the modal navigation issue

### LINKED ISSUE #2651: Visiting URL to User Settings modal doesn't work
### Problem

If I load any of these URLs directly they won't work, I'll be redirected to the dashboad:

- http://localhost:3000/app/dashboard?modal=UserSettingsModal
- http://localhost:3000/app/dashboard?modal=UserSettingsModal&tab=application-logs

### Solution

They should take me to the corresponding section in the UserSettings modal.

SebinSong:
closes #2651 

https://github.com/user-attachments/assets/13a21237-9daa-4a03-a8f8-6b7f9adf0463

### DIFF
--- a/frontend/controller/router.js
+++ b/frontend/controller/router.js
@@ -193,7 +193,7 @@ router.beforeEach((to, from, next) => {
     !['SignupModal', 'LoginModal'].includes(to.query.modal) &&
     !store.state.loggedIn) {
     // if modal is queried and,
-    // the requested modal is only meant to be used post authentication and,
+    // the requested modal is only meant to be used post authentication and
     // the user is not logged in now, navigate to home page and set the next query param
     // to redirect to the requested page after login.
     return next({ path: '/', query: { next: to.fullPath } })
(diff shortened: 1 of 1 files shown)


frontend/controller/router.js
```diff
@@ -194,8 +194,9 @@ router.beforeEach((to, from, next) => {
     !store.state.loggedIn) {
     // if modal is queried and,
     // the requested modal is only meant to be used post authentication and,
-    // the user is not logged in now, discard the navigation.
-    return next(from)
+    // the user is not logged in now, navigate to home page and set the next query param
+    // to redirect to the requested page after login.
+    return next({ path: '/', query: { next: to.fullPath } })
```
SebinSong: The guard logic here makes sure any modals that are part of Post-auth area aren't accessed if user has not logged in yet. (It was a bug-fix added about a couple of years ago, if I remember correctly)

Since then the app has grown in complexity and loading user-settings data during the initialization takes longer time than before. The cause of the issue was this logic gets executed earlier than user-setting is completely loaded. 
Taking advantage of the existing logic for `next` query param fixes this issue.

Reference:

https://github.com/okTurtles/group-income/blob/0fdb7ad2e03f0214720b967ad02115ff26efb0ea/frontend/views/pages/Home.vue#L136-L146

taoeffect (approved):
Beautifully done @SebinSong! Probably fixed some other bugs in the process! 😄

PR #2660, opened 2025-02-23
Avoid notifications while joining

### LINKED ISSUE #2629: Don't post chat notifications when joining a group
### Problem

When joining a new group, if a chatroom has an `@all` mention, then a notification and a sound will be triggered.

### Solution

When joining a group, don't post these old notifications, just ignore them.

corrideat:
Closes #2629

### DIFF
--- a/frontend/model/notifications/messageReceivePostEffect.js
+++ b/frontend/model/notifications/messageReceivePostEffect.js
@@ -73,6 +73,16 @@ async function messageReceivePostEffect ({
 
     const chatNotificationSettings = rootGetters.chatNotificationSettings[contractID] || rootGetters.chatNotificationSettings.default
     const { messageNotification, messageSound } = chatNotificationSettings
+
+    // If the contract is syncing (meaning we're loading the app, joining a
+    // chatroom, etc.), don't use a native notification or sound. Do this only
+    // for messages coming over the WS
+    const isSyncing = sbp('chelonia/contract/isSyncing', contractID)
+    // TODO: This could be an issue (false negative for emitting a native
+    // notification) for very large (over 4 KiB) push events, which are
+    // handled using a sync.
+    if (isSyncing) return
+
     const shouldNotifyMessage = messageNotification === MESSAGE_NOTIFY_SETTINGS.ALL_MESSAGES ||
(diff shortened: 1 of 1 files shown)


corrideat:
It turns out the solution was quite straightforward as you thought it'd be, @taoeffect.

I'm not using 'joining' as the criterion because it's ambiguous and likely wrong. For example, you get the notification sound when using a new device, even though you shouldn't.

frontend/model/notifications/messageReceivePostEffect.js
```diff
@@ -73,6 +73,16 @@ async function messageReceivePostEffect ({
...
+
+    // If the contract is syncing (meaning we're loading the app, joining a
+    // chatroom, etc.), don't use a native notification or sound. Do this only
+    // for messages coming over the WS
+    const isSyncing = sbp('chelonia/contract/isSyncing', contractID)
+    // TODO: This could be an issue (false negative for emitting a native
+    // notification) for very large (over 4 KiB) push events, which are
+    // handled using a sync.
+    if (isSyncing) return
+
```
taoeffect: So, I don't know what's going on, but we used to have code somewhere that would sync contracts when we came back online to get the events we missed while offline.

I don't remember where it is and it may have gotten removed during a refactoring. If so, stuff is definitely broken and we just haven't noticed it, perhaps because of how Chatrooms are rendered differently than groups.

In fact, I think it is broken. I just tried changing someone's display name on desktop while offline on the pwa and when i came back online the change didn't appear, so I'll open an issue for this.

Here's how this relates to this PR:

If we are offline and then come back online, we *do* want the sync to happen and for us to hear the notification. With the way this is implemented, that won't happen.

taoeffect (changes requested):
Review ready!

frontend/model/notifications/messageReceivePostEffect.js
corrideat: That code is there in `chelonia.js` and I've just tested it and it works (for changing a group name), but the banner while being offline doesn't appear.

You're right that the way this is currently implemented, notifications will be missed when coming back online. I'll mark this as a draft until it's resolved.

taoeffect (approved):
Awesome to see `firstSync` being used! 😄

PR #2668, opened 2025-02-25
#2641 - Display time-stamp on merged message

### LINKED ISSUE #2641: Merged messages should display timestap on hover
### Problem

Merged messages should display timestamp on hover, but currently don't show any additional time information.

Image

Figma: https://www.figma.com/design/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23766-264439&t=sgmDOdrueSKNzmUE-1

### Solution

Fix.

SebinSong:
closes #2641 

**[ Screenshot ]**

### DIFF
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -12,7 +12,7 @@
 
   .c-message-wrapper
     .c-sender-or-date
-      .c-time-stamp(v-if='isSameSender') {{ humanTimeString(datetime, { hour: "2-digit", minute: "2-digit", hour12: false }) }}
+      .c-time-stamp(v-if='isSameSender') {{ getTimeStamp(datetime) }}
       slot(v-else name='image')
         profile-card(:contractID='from' direction='top-left')
           avatar.c-avatar(:src='avatar' aria-hidden='true' size='md')
@@ -200,7 +200,6 @@ export default ({
   },
   methods: {
     humanDate,
-    humanTimeString,
     swapMentionIDForDisplayname,
     editMessage () {
       this.isEditing = true
@@ -248,6 +247,12 @@ export default ({
       minutes = minutes.toString().padStart(2, '0')
 
       return `${hours}:${minutes}`
+    },
+    getTimeStamp (datetime) {
+      const tString = humanTimeString(datetime, { hour: '2-digit', minute: '2-digit', hour12: true })
+      // Extract only numbers and colons, removing non-numeric characters like AM/PM
+      const cleaned = tString.replace(/[^\d:]/g, '')
+      return cleaned
     }
   }
 }: Object)
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -238,16 +238,6 @@ export default ({
         this.openMenu()
       }
     },
-    formatTimeString (datetime) {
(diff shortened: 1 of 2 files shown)


frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -235,6 +237,16 @@ export default ({
...
+    },
+    formatTimeString (datetime) {
+      const dateObj = new Date(datetime)
+      let hours = dateObj.getHours()
+      let minutes = dateObj.getMinutes()
+
+      hours = hours % 12 || 12 // Convert 0 to 12 for midnight
+      minutes = minutes.toString().padStart(2, '0')
+
+      return `${hours}:${minutes}`
```
taoeffect: AFAIK this type of manual formatting isn't necessary. Any time we want to display time information we should use the browser's `toLocaleDateString` or `toLocaleTimeString`.

We have a `humanDate` function that could be used for this purpose. If that doesn't work, then feel free to modify it add a third parameter to specify that it use `toLocaleTimeString`, e.g. something like this:

```js
export function humanDate (
  date: number | Date | string,
  options?: Intl$DateTimeFormatOptions = { month: 'short', day: 'numeric' },
  fnName?: 'toLocaleDateString' | 'toLocaleTimeString' = 'toLocaleDateString'
): string {
  const fallback = 'en-US-POSIX'
  const locale = typeof navigator === 'undefined'
    // Fallback for Mocha tests.
    ? fallback
    // Flow considers `navigator.languages` to be of type `$ReadOnlyArray<string>`,
    // which is not compatible with the `string[]` expected by `.toLocaleDateString()`.
    // Casting to `string[]` through `any` as a workaround.
    : ((navigator.languages: any): string[]) ?? navigator.language ?? fallback
  const dateObj = new Date(date)

  // Avoid returning 'Invalid Date'
  if (!isNaN(dateObj.valueOf())) return dateObj[fnName](locale, options)

  return ''
}
```

taoeffect (changes requested):
Nice work @SebinSong! One change request:

frontend/views/containers/chatroom/MessageBase.vue
SebinSong: @taoeffect 
Okay, let me update the logic to use `toLocaleTimeString`.

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -235,6 +237,12 @@ export default ({
       } else if (!this.isEditing) {
         this.openMenu()
       }
+    },
+    getTimeStamp (datetime) {
+      const tString = humanTimeString(datetime, { hour: '2-digit', minute: '2-digit', hour12: true })
+      // Extract only numbers and colons, removing non-numeric characters like AM/PM
+      const cleaned = tString.replace(/[^\d:]/g, '')
+      return cleaned
```
SebinSong, line 245: With 12-hour format, `date. toLocaleTimeString` return string that says whether AM or PM. eg)
```
console.log(date.toLocaleTimeString("en-US"));
// "7:00:00 PM"

console.log(date.toLocaleTimeString("ko-KR"));
// "오후 12:00:00"
```
reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#using_locales

This info does not fit the available space and is better to be removed.

taoeffect (approved):
Great work @SebinSong!

PR #2670, opened 2025-02-25
Sync contracts using online event

### LINKED ISSUE #2665: Banner that shows we're offline is missing
### Problem

It used to be the case that a white banner would appear telling us we're offline whenever we went offline.

Now in the PWA I just tested this and didn't see the banner.

### Solution

Restore this.

### LINKED ISSUE #2666: Contracts aren't synced when we come back online
### Problem

Related to this Review: https://github.com/okTurtles/group-income/pull/2660#discussion_r1966871609

Contracts are not sync'd when we come back online. It was thought that this happened here in `chelonia.js`:

```js
        'subscription-succeeded': (event) => {
          const { channelID } = event.detail
          if (this.subscriptionSet.has(channelID)) {
            // For new subscriptions, some messages could have been lost
            // between the time the subscription was requested and it was
            // actually set up. In these cases, force sync contracts to get them
            // updated.
            sbp('chelonia/private/out/sync', channelID, { force: true }).catch(err => {
              console.warn(`[chelonia] Syncing contract ${channelID} failed: ${err.message}`)
            })
          }
        }
```

But this `'subscription-succeeded'` is the wrong event.

So for example, as mentioned in the review, being offline and someone changing their display name, then coming back online, doesn't show the change.

### Solution

Listen to the correct event, the one for when we come back online.

corrideat:
Closes #2666 
Closes #2665


taoeffect (approved):
Great work @corrideat!

PR #2674, opened 2025-02-26
#2577 - Paste issue on iOS Safari

### LINKED ISSUE #2577: Cannot paste invite on iOS
### Problem

On `testing`, the invite link does not paste into the chat, even though it's on the clipboard.

1. Go to group settings
2. Copy invite link
3. Go to chat (channel or DM)
4. Attempt to paste invite link
5. Invite link is not posted 

Link not posted, "paste" clearly an option. NPS link (most recent message) was properly posted during same session.
Image

"[Your Invite]()" does post properly to the Notes app
Image

### Solution

Invite link (and other clipboard contents) should be able to be pasted into chat text box.

SebinSong:
closes #2577 

Fortunately I could observe this issue from the iOS emulator too.
Had to improve the previous fix for #2369 to make it work for both cases.

**[ Fix video recording ]**

https://github.com/user-attachments/assets/00ad1281-d0af-4068-8420-a677afb2ad96

### DIFF
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -620,7 +620,7 @@ export default ({
       return true
     },
     updateTextArea () {
-      if (!this.updateTextWithLines()) {
+      if (!this.$refs.textarea || !this.updateTextWithLines()) {
         // dont calculate again when the value is the same (ex: happens on shift+enter)
         return false
       }
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -359,7 +359,7 @@ export default ({
       },
       config: {
         messageMaxChar: CHATROOM_MAX_MESSAGE_LEN,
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -357,7 +358,8 @@ export default ({
         typingUsers: []
       },
       config: {
-        messageMaxChar: CHATROOM_MAX_MESSAGE_LEN
+        messageMaxChar: CHATROOM_MAX_MESSAGE_LEN,
+        debouncedHandleInput: debounce(this.updateTextArea, 250) // NOTE: This is a fix for the issue #2369 and #2577.
```
taoeffect: Hey Sebin, I see that this fixes the issue, but I don't understand how or why it works.

Could you add a comment above this line explaining why, for example, this change doesn't break the app? E.g. I thought this `debounce` might mean that keystrokes might get lost but it doesn't. Can you add a comment explaining how this fixes the problem and why no keystrokes are lost?

taoeffect (changes requested):
Great job @SebinSong! It works great!

Just one minor RFC (request-for-comment)

taoeffect (approved):
Nice work!

PR #2675, opened 2025-02-26
improve backend push related logging

taoeffect:


backend/push.js
```diff
@@ -199,11 +199,19 @@ export const postEvent = async (subscription: Object, event: ?string): Promise<v
 
   if (!req.ok) {
     const endpointHost = new URL(subscription.endpoint).host
-    console.warn(`Error ${req.status} sending push notification to '${subscription.id}' via ${endpointHost}`)
+    console.warn(
+      await req.text().then(text => {
+        try {
+          return (text && JSON.parse(text)) || '(empty response)'
```
corrideat: Why not just log the text as is? I don't think we should be parsing this as JSON.

backend/push.js
```diff
@@ -199,11 +199,19 @@ export const postEvent = async (subscription: Object, event: ?string): Promise<v
 
   if (!req.ok) {
     const endpointHost = new URL(subscription.endpoint).host
-    console.warn(`Error ${req.status} sending push notification to '${subscription.id}' via ${endpointHost}`)
+    console.warn(
+      await req.text().then(text => {
```
corrideat: This is missing the second argument to `then` or a call to `catch`.

taoeffect: what exactly would the `.catch` be for? `.text()` failing?

backend/push.js
taoeffect: Because it sends back JSON and the logs will look nice this way:

Screenshot 2025-02-26 at 10 32 02 AM

corrideat: But it slows things down and the response is potentially adversarial-controlled. I don't think we need the prettiness.

backend/push.js
corrideat: For example. If it fails, the app will currently crash.

corrideat: I'd replace the whole thing with `await req.text().catch((e) => '(error parsing body)')`

backend/push.js
taoeffect: It only slows things down imperceptibly on exception... I think it'll be OK.

The prettiness only happens if the sysadmin wants it (the actual logs are raw JSON).

corrideat: Also, if you're making it a PR, probably the payload should be logged at a different level (e.g., debug or info) to avoid filling up logs.

corrideat: I'm aware the prettiness only happens if the sysadmin wants it. But, you're not guaranteed that the response will be JSON to begin with, and it can be controlled by an adversary (and because of this, it can happen pretty often, as often as the adversary chooses, actually). There's no need to parse the payload, and the sysadmin can use `jq` if they want  pretty output.

corrideat (approved):
Nice

PR #2682, opened 2025-02-26
2677 error during sign up

### LINKED ISSUE #2656: Two joining group bugs
#### Bug 1

While working on #2649 and using the invite link to join a group, I saw this in the console (note: I don't remember if these happened immediately upon loading the link or after signing up the new user):

Image

Image

#### Bug 2

Similar situation, but definitely after clicking Sign Up:

Image

Image

Image

After clicking "No" and refreshing the page everything seem to work OK, but this shouldn't have happened to begin with. Very odd.

### Solution

Track down what's going on and fix these bugs.

### LINKED ISSUE #2677: Error during Sign Up
### Problem

Image

In Brave, which probably had a previous debug session signed in, I tried signing up and was greeted with this error.

📎  [logs.log](https://github.com/user-attachments/files/18992691/logs.log)

### Solution

Fix. I don't know why I'm seeing this error banner to begin with. If I was logged out then signup should work fine.

corrideat:
Closes #2656 (edit by @taoeffect)

### DIFF
--- a/frontend/controller/app/identity.js
+++ b/frontend/controller/app/identity.js
@@ -311,10 +311,19 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', 'APP-LOGIN', async () => {
       console.debug('[gi.app/identity/login] Scheduled call starting', identityContractID, username)
       if (username) {
+        // We expect that in development mode the same browser may be used and
+        // server data cleared often, so we skip the cache lookup for dev
+        // convenience.
         const nsIdentityContractID = await sbp('namespace/lookup', username, { skipCache: process.env.CI || process.env.NODE_ENV !== 'production' })
+        // If we've only been given a username, set `identityContractID` to the
+        // contract ID we've just looked up
         if (!identityContractID) {
           identityContractID = nsIdentityContractID
         } else if (nsIdentityContractID !== identityContractID) {
+          // However, if we _know_ what the contract ID should be (e.g., right
+          // after signing up, when we've ourselves created the contract), we
+          // check that the contractID we're signing in into is what we expect
+          // it to be.
           console.error(new Error(`Identity contract ID mismatch during login: ${identityContractID} != ${nsIdentityContractID}`))
           throw new GIErrorUIRuntimeError(L('Identity contract ID mismatch during login'))
         }
--- a/frontend/controller/app/identity.js
+++ b/frontend/controller/app/identity.js
@@ -311,7 +311,13 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', 'APP-LOGIN', async () => {
       console.debug('[gi.app/identity/login] Scheduled call starting', identityContractID, username)
       if (username) {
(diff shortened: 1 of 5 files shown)


frontend/controller/app/identity.js
```diff
@@ -311,7 +311,13 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', 'APP-LOGIN', async () => {
       console.debug('[gi.app/identity/login] Scheduled call starting', identityContractID, username)
       if (username) {
-        identityContractID = await sbp('namespace/lookup', username)
+        const nsIdentityContractID = await sbp('namespace/lookup', username, { skipCache: process.env.CI || process.env.NODE_ENV !== 'production' })
```
taoeffect, line 317: Why are we skipping the cache now if in CI or in development mode?

corrideat: Because in production we probably want an error / warning to happen if the username now refers to a different contract ID. But in development, this is a pretty common situation if you re-use a browser profile.

frontend/controller/app/identity.js
```diff
@@ -311,7 +311,13 @@ export default (sbp('sbp/selectors/register', {
...
+        if (!identityContractID) {
+          identityContractID = nsIdentityContractID
+        } else if (nsIdentityContractID !== identityContractID) {
+          console.error(new Error(`Identity contract ID mismatch during login: ${identityContractID} != ${nsIdentityContractID}`))
+          throw new GIErrorUIRuntimeError(L('Identity contract ID mismatch during login'))
+        }
```
taoeffect: Why is this new code here? What is it doing? Are you sure it's necessary? If so, the answer to these questions needs to be a comment above line 315

> L('Identity contract ID mismatch during login')

`contract ID` => `contractID`

taoeffect (commented):
Nice, just have some questions

frontend/controller/app/identity.js
taoeffect: OK could you explain that in a comment above line π?

frontend/controller/app/identity.js
corrideat: This code is checking that we're logging in to the contract we've just created, which is what was failing before due to a cache issue, but also generally an error from the perspective of the client

> contract ID => contractID

`contract ID` (with a space) is used elsewhere (actually, the exact string). I can change it, but for non-technical users probably contract ID is more meaningful than a made up word like `contractID`

taoeffect: > contract ID (with a space) is used elsewhere (actually, the exact string). I can change it, but for non-technical users probably contract ID is more meaningful than a made up word like contractID

Fair enough... although I feel like non-technical users are perfectly capable of picking up such "vocab".

Just a RFC then

corrideat: >  are perfectly capable of picking up such "vocab"

Sure they can, but it's a bit extra cognitive load. Although maybe it's less confusing than thinking about it in terms of a _contract_.

taoeffect (approved):
Nice!

PR #2683, opened 2025-02-26
Recreate push subscription on new VAPID key

### LINKED ISSUE #2681: "VAPID public key mismatch" error on server results in 401 push error
### Problem

This is from the server:

Image

### Solution

🤷‍♂

corrideat:
Closes #2681 

With this PR, it should now be possible to use the non-persistent in-memory storage with existing subscriptions.

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -221,7 +221,10 @@ sbp('sbp/selectors/register', {
           subscription = await registration.pushManager.getSubscription()
           if (subscription && !bufferEq(subscription.options.applicationServerKey, subscriptionOptions?.applicationServerKey)) {
             // This is a public key that belongs to the server
-            console.warn('VAPID server key changed; removing existing subscription and setting up a new one', { oldKey: subscription.options.applicationServerKey, newKey: subscriptionOptions.applicationServerKey })
+            console.warn('VAPID server key changed; removing existing subscription and setting up a new one', {
+              oldPubKey: subscription.options.applicationServerKey,
+              newPubKey: subscriptionOptions.applicationServerKey
+            })
             await subscription.unsubscribe()
             subscription = null
           }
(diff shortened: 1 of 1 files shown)


frontend/controller/service-worker.js
```diff
@@ -204,11 +216,17 @@ sbp('sbp/selectors/register', {
         let subID = null
         // get a real push subscription only if both browser permissions allow and user wants us to
         if (notificationEnabled && granted) {
+          const subscriptionOptions = await sbp('push/getSubscriptionOptions')
           subscription = await registration.pushManager.getSubscription()
+          if (subscription && !bufferEq(subscription.options.applicationServerKey, subscriptionOptions?.applicationServerKey)) {
```
taoeffect, line 222: With the way this is written, is it possible for `subscription.options.applicationServerKey` to be an array and for `subscriptionOptions?.applicationServerKey` to be undefined/null?

If so, the `bufferEq` implementation won't work on this.

frontend/controller/service-worker.js
```diff
@@ -204,11 +216,17 @@ sbp('sbp/selectors/register', {
...
+            console.warn('VAPID server key changed; removing existing subscription and setting up a new one', { oldKey: subscription.options.applicationServerKey, newKey: subscriptionOptions.applicationServerKey })
```
taoeffect: Should you be logging these keys (especially the new key)? If so, can you add a comment explaining why it's safe to do so? Remember users could be sending us these logs and we don't want to expose any secrets.

taoeffect (changes requested):
Great work @corrideat! Review ready!

frontend/controller/service-worker.js
corrideat: Well, these keys are public information (or at least, public keys) and they belong to the server, so I don't see this being an issue.

frontend/controller/service-worker.js
corrideat: I see what you mean. They way things work now, it shouldn't really be possible, but technically it's possible. It makes the check uglier, but I'll fix it.

taoeffect (approved):
Awesome!

PR #2688, opened 2025-02-27
hotfix-cmp-appserverkey

corrideat:
Testing:

1. Create a push subscription. Refresh the page. No warning about VAPID appears.
2. Restart the server (changing the VAPID key). Re-register. A warning about VAPID appears.

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -10,16 +10,35 @@ import { Secret } from '~/shared/domains/chelonia/Secret.js'
 import { deserializer, serializer } from '~/shared/serdes/index.js'
 import { getSubscriptionId } from '~/shared/functions.js'
 
+// The application server (public) key could be either an ArrayBuffer (which is
+// what we get from fetching the current subscription), or it could be a
+// base64url-encoded string (which is what we usually get from the server)
+// This string coerces both of these into an `Uint8Array` instance.
+const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer) => {
+  if (typeof v === 'string') {
+    v = Buffer.from(
+      // Convert from base64url to base64
+      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - v.length % 4) % 4),
+      'base64'
+    )
+    return v
+  }
+  return new Uint8Array(v)
+}
+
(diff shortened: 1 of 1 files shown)


frontend/controller/service-worker.js
```diff
@@ -10,16 +10,30 @@ import { Secret } from '~/shared/domains/chelonia/Secret.js'
 import { deserializer, serializer } from '~/shared/serdes/index.js'
 import { getSubscriptionId } from '~/shared/functions.js'
 
+const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer) => {
+  if (typeof v === 'string') {
+    v = Buffer.from(
+      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - v.length % 4) % 4),
+      'base64'
+    )
```
taoeffect: Could you add a comment here explaining what is going on here and where you found this code?

taoeffect (changes requested):
Review ready!

frontend/controller/service-worker.js
```diff
@@ -10,16 +10,35 @@ import { Secret } from '~/shared/domains/chelonia/Secret.js'
...
+const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer) => {
+  if (typeof v === 'string') {
+    v = Buffer.from(
+      // Convert from base64url to base64
+      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - v.length % 4) % 4),
+      'base64'
+    )
+    return v
+  }
+  return new Uint8Array(v)
```
taoeffect, line 26: Some return issues:

1. For strings, it returns a Buffer, but the comment says it's supposed to return a `Uint8Array`
2. For non-strings, if it's a `Uint8Array`, no need to coerce it to a `Uint8Array` again.
3. However, the comment says it can be either an `ArrayBuffer` or a `string`, but the function signature suggests a third type, a `Uint8Array`. So either the comment needs to be updated or `Uint8Array` should be removed from the function signature

corrideat: 1. `Buffer` extends `Uint8Array`
2. True, but that requires an extra if, and creating an Uint8Array is quite cheap as there's no data copying. Also, in practice this will never be `Uint8Array`. It'll be an `ArrayBuffer`.
3. Correct. The `Uint8Array` type is allowed in the types annotations mostly for flexibility.

taoeffect (approved):
Nice!

Issue #2690, opened 2025-02-27
Stuck "Loading events from server..." after upgrade.


taoeffect:
### Problem

After autorefresh on Brave to v1.2.15, this banner doesn't disappear:

![Image](https://github.com/user-attachments/assets/dd65ea16-cabb-448e-aea8-cfafa31a3910)

### Solution

Fix.

taoeffect:
Probably closed by #2693

Issue #2680, opened 2025-02-26
SW doesn't always update after an update


taoeffect:
### Problem

![Image](https://github.com/user-attachments/assets/b3ec6f8c-f477-4210-890d-b3ee0e3ca33a)

Sometimes, but seemingly not always, the SW doesn't actually update.. and in the screenshot above it almost seems like it updated and then somehow un-updated 🤯 

See the logs saying that it went from v1.2.13 to v1.2.12.

📎 Console Save As... logs: [groupincome.app-1740600201396.log](https://github.com/user-attachments/files/18993568/groupincome.app-1740600201396.log)

📎 [gi_logs(Brave_unupdate).json.txt](https://github.com/user-attachments/files/18993579/gi_logs.Brave_unupdate.json.txt)

Note: manually refreshing again fixed the problem.

### Solution

Fix.

Seems related to #2646.

EDIT: As a band-aid solution: have a periodic check every 30 minutes on the frontend, and when certain actions are taken (e.g. when user settings are opened), that checks the version of the SW and if it's different from the browser, display a banner with a link to refresh the page.

The good news is that when this happens on my PWA, it manages to update the SW after a short period of time when I've placed the app in the background and re-opened it. This seems to happen within about 10 minutes, so maybe there is no need to do anything...

taoeffect:
![Image](https://github.com/user-attachments/assets/8b280784-3dca-45e6-82c8-de95f1b09b74)

Seeing this in my PWA. Had to force quit it to get it to update.

PR #2693, opened 2025-03-02
Various fixes & improvements

### LINKED ISSUE #2687: Event handler warning for online and offline events in Brave
### Problem

After a recent update to 1.2.14, the testing site was redirected to a grey page:

Image

Also in the console are these warnings:

Image

### Solution

Maybe we can figure out why it was redirected to a blank page.

But also for this issue would be to fix where this is called:

Image

corrideat:
Definitely closes #2687.
May 'close' #2680, #2690.

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -2,7 +2,7 @@
 
 import { L } from '@common/common.js'
 import sbp from '@sbp/sbp'
-import { CAPTURED_LOGS, LOGIN_COMPLETE, NEW_CHATROOM_UNREAD_POSITION, NEW_CURRENT_SYNCS, PWA_INSTALLABLE, SET_APP_LOGS_FILTER } from '@utils/events.js'
+import { CAPTURED_LOGS, LOGIN_COMPLETE, NEW_CHATROOM_UNREAD_POSITION, CONTRACT_SYNCS_RESET, PWA_INSTALLABLE, SET_APP_LOGS_FILTER } from '@utils/events.js'
 import isPwa from '@utils/isPwa.js'
 import { HOURS_MILLIS } from '~/frontend/model/contracts/shared/time.js'
 import { GIMessage } from '~/shared/domains/chelonia/GIMessage.js'
@@ -68,8 +68,11 @@ const waitUntilSwReady = () => {
     const messageChannel = new MessageChannel()
     messageChannel.port1.onmessage = (event) => {
       if (event.data.type === 'ready') {
-        if (!event.data.currentSyncs) return
-        sbp('okTurtles.events/emit', NEW_CURRENT_SYNCS, event.data.currentSyncs)
+        // For backward- and forward-compatibility (`currentSyncs` may not be
+        // defined in a different SW version)
+        if (event.data.currentSyncs) {
+          sbp('okTurtles.events/emit', CONTRACT_SYNCS_RESET, event.data.currentSyncs)
+        }
         resolve()
       } else {
         reject(event.data.error)
@@ -193,6 +196,13 @@ sbp('sbp/selectors/register', {
         })
       }
 
+      // Set up an event listener to have SW call `skipWaiting` after
+      // installation is complete. The SW already does this, but we had some
+      // issues with the SW being installed but not activated after an update
+      // for yet-unknown reasons. This logic attemps to force a new call to
+      // `skipWaiting`, increasing the odds of the new SW replacing the old
+      // version. (When a SW updates, both versions exist simultaneously,
+      // although only one is connected to a page at a time.)
       swRegistration.addEventListener('updatefound', (e) => {
         const handler = (e) => {
           if (['activated', 'redundant'].includes(e.target.state)) {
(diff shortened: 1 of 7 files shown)


frontend/controller/service-worker.js
```diff
@@ -68,6 +68,8 @@ const waitUntilSwReady = () => {
     const messageChannel = new MessageChannel()
     messageChannel.port1.onmessage = (event) => {
       if (event.data.type === 'ready') {
+        if (!event.data.currentSyncs) return
```
taoeffect: This is a curious line.... could you add a comment about why it's here? Is this to handle old SW versions or something?

frontend/main.js
```diff
@@ -183,6 +183,9 @@ async function startApp () {
       this.ephemeral.syncs = this.ephemeral.syncs.filter(id => id !== contractID)
     }
   }
+  sbp('okTurtles.events/on', NEW_CURRENT_SYNCS, (currentSyncs) => {
+    this.ephemeral.syncs = currentSyncs
+  })
```
taoeffect: This seems like a handler that is supposed to update the frontend's understand of whatever contracts are currently being sync'd, but it doesn't seem like it's doing that because the only time the `NEW_CURRENT_SYNCS` event seems to be emitted is once at page load.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -325,7 +325,7 @@ const setupPromise = setupChelonia()
 
 self.addEventListener('install', function (event) {
   console.debug('[sw] install')
-  event.waitUntil(Promise.all([setupPromise, self.skipWaiting()]))
+  event.waitUntil(setupPromise.then(() => self.skipWaiting()))
```
taoeffect, line 331: Are you sure it shouldn't be `skipWaiting.then(() => self.setupPromise())` ?

(I'm not saying it should be... I just don't know myself)

frontend/controller/service-worker.js
```diff
@@ -191,6 +193,19 @@ sbp('sbp/selectors/register', {
...
+          if (['activated', 'redundant'].includes(e.target.state)) {
+            e.target.removeEventListener('statechange', handler)
+            return
+          }
+          if (e.target.state === 'installed') {
+            e.target.postMessage({ type: 'skip-waiting' })
+          }
+        }
+        e.target.installing.addEventListener('statechange', handler, false)
+      })
```
taoeffect, line 217: Could you add a brief comment above this explaining why this was added?

taoeffect (changes requested):
Nice work @corrideat!

Left a few comments!

Also, could you please edit the PR description to reference:

- Which issues this definitely closes
- Which issues this might close

frontend/main.js
corrideat: `NEW_CURRENT_SYNCS` is emitted on a SW update.

taoeffect: I see that, but shouldn't it be emitted whenever there are "new syncs"?

frontend/controller/serviceworkers/sw-primary.js
corrideat: Not sure myself, and the documentation isn't very helpful here. However, I think this is correct. `skipWaiting` tells the browser that the SW should immediately move from `installed` to `activated` (or from `waiting` to `activated`). We only want to do this after setup is complete, so this seems reasonable. With the previous code, I think it worked the same way too, since it'd wait until `setupPromise` had resolved, but now it's more explicit.

TL;DR: I think it doesn't matter when `skipWaiting` is called, but, if it made a difference, what's now there seems more correct.

frontend/controller/service-worker.js
corrideat: Yes, this should reasonanly handle old or new SW versions that don't have this field.

frontend/controller/serviceworkers/sw-primary.js
taoeffect: Can you copy/paste that as a comment into the code?

frontend/main.js
taoeffect: Can you rename `NEW_CURRENT_SYNCS` to `CONTRACT_SYNCS_RESET`?

taoeffect:
Could you also add this `no-store` to the `GET /time` to this PR? https://github.com/okTurtles/group-income/pull/2494#discussion_r1978033066

frontend/controller/service-worker.js
```diff
@@ -68,6 +68,10 @@ const waitUntilSwReady = () => {
     const messageChannel = new MessageChannel()
     messageChannel.port1.onmessage = (event) => {
       if (event.data.type === 'ready') {
+        // For backward- and forward-compatibility (`currentSyncs` may not be
+        // defined in a different SW version)
+        if (!event.data.currentSyncs) return
```
taoeffect: Actually, one more question: should this call `reject` before calling `return`?

corrideat: No, but `resolve` should be called. Fixed.

taoeffect (approved):
Nice! 🤞 that this does it!

PR #2696, opened 2025-03-03
#2556 - New Queue System for `ChatMain.vue`

### LINKED ISSUE #2556: New Queue System for ChatMain
### Problem

The design of `ChatMain.vue` is very problematic. It is currently causing tests to fail because of its very async natural and the way it handles switching chatrooms.

Because of its poor design in handling switching chatrooms, all over the code are lines like this after almost every `await`:

```js
if (!this.checkEventSourceConsistency(chatRoomID)) return
```

And this from bd0d0d1a49d76d8c53b694ce156fbdf116e08ab0:

```js
if (this.renderingChatRoomId !== this.summary.chatroomID) // NOTE: this is actually a bug, it should be this.ephemeral.renderingChatRoomId
```

We are checking `this.summary.chatroomID`, `this.ephemeral.renderingChatRoomId`, `this.currentChatRoomId` all over the place. This is wrong. We should be using only a single `renderingChatRoomId` for the component, and nothing else.

### Solution

Redesign how ChatMain handles switching chatrooms.

Make it so that ChatMain only cares about a single "chatroomID" variable, e.g. `renderingChatRoomId` or whatever.

Every time a switch in chatroomID is detected, do not mess with any of ChatMain's variables. Do not update anything. Instead, create a request to ChatMain for ChatMain to process in its own time, on a queue. This request will say, "Please update your `renderingChatRoomId` to this new value", and ChatMain will in the meantime ignore all new messages that come in on other chatroomID contracts.

In other words, ChatMain will process requests from a queue to update the current chatroomID in its `ephemeral` data. And these requests can stack up. So for example, Cypress can rapidly switch from chatroom to chatroom but ChatMain will only render and show chatroom for its own internal `ephemeral.chatroomID` (`renderingChatRoomId`, whatever). It will process any requests to update this variable in sequence, one at a time.

*Note: when pulling events off of this internal queue, ChatMain can also check this size of the queue. For example, it can decide that if there are more than 1 request on the queue, it can `pop()` and ignore the old ones and just process the most recent one, to speed things up.

With these changes made, you should be able to greatly clean up ChatMain and remove all paranoid and DRY-violating calls to `checkEventSourceConsistency` etc.

One last note: PR https://github.com/okTurtles/group-income/pull/2555 should probably be merged first as it makes a few changes to ChatMain in an attempt to fix the Cypress tests.

SebinSong:
closes #2556 

---
@taoeffect 
Firstly, the AI generated code snippet you attached [here](https://github.com/okTurtles/group-income/issues/2556#issuecomment-2683819280) has been a great help. I actually just took that snippet and enhanced the solution to borrow the major idea it suggests. Some parts did need some adjustments/simplification/fixes though, For examples:

**[1]** in `this.processSwitchQueue` below,

```js
    async processSwitchQueue () {
      if (this.isProcessingSwitch || this.chatroomSwitchQueue.length === 0) return

      this.isProcessingSwitch = true
      const target = this.chatroomSwitchQueue[this.chatroomSwitchQueue.length - 1]
      this.chatroomSwitchQueue = []
      this.renderingChatRoomId = target.chatRoomId
      this.renderingSummary = target.summary
      await this.initializeState(true)
      this.ephemeral.messagesInitiated = false
      if (this.$refs['infinite-loading']) {
        this.$refs['infinite-loading'].stateChanger.reset()
        this.$refs['infinite-loading'].attemptLoad()
      }
      this.isProcessingSwitch = false
      if (this.chatroomSwitchQueue.length > 0) this.processSwitchQueue()
    },
```
we do not really need `this.isProcessingSwitch` and `this.renderingSummary` here. also, the version of `vue-infinite-loading` package we are using does not have `.stateChanger.reset()` and `.attemptLoad()` APIs. So I've made some adjustments here.

**[2]**
It simplified the watcher for `this.summary` prop and removed below existing logics:
https://github.com/okTurtles/group-income/blob/3d5a29ceed5e0cc690f676c4d260a2426d18829f/frontend/views/containers/chatroom/ChatMain.vue#L1223-L1234 

Don't know the exact context but `sbp('chelonia/contract/isSyncing', toChatRoomId)` here (and its related logics) very much looks like a bug-fix added before so was concerned this removal might bring a bug in. So I actually reverted this change back but updated it to work with the new Queue system instead.

---
Cypress passes and a few rounds of my local tests with `GI_PERIST=sqlite grunt dev` did work fine. So hope it all looks good.


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1201,17 +1146,20 @@ export default ({
       const fromIsJoined = from.isJoined
 
       const initAfterSynced = (toChatRoomId) => {
-        if (toChatRoomId !== this.summary.chatRoomID || this.ephemeral.messagesInitiated) return
-        return this.setInitMessages()
+        if (toChatRoomId !== this.summary.chatRoomID || // If the user has switched to another chatroom during syncing, no need to process the chatroom that has been swithed away.
+          this.ephemeral.messagesInitiated) return
+        this.processSwitchQueue()
```
corrideat, line 1177: Missing `catch`

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1221,16 +1169,14 @@ export default ({
...
         // in race conditions as this is a watcher.
         Promise.resolve(sbp('chelonia/contract/isSyncing', toChatRoomId)).then((isSyncing) => {
-          // If the chatroom has changed since, return
-          if (this.summary.chatRoomID !== toChatRoomId) return
           if (isSyncing) {
-            toIsJoined && sbp('chelonia/queueInvocation', toChatRoomId, () => initAfterSynced(toChatRoomId))
+            sbp('chelonia/queueInvocation', initAfterSynced)
           } else {
-            this.refreshContent()
+            this.processSwitchQueue()
```
corrideat: Missing catch

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1153,24 +1091,12 @@ export default ({
...
     // Handlers for file-upload via drag & drop action
     dragStartHandler (e) {
       e.preventDefault()
       // handler function for 'dragstart', 'dragover' events.
 
-      if (!this.dndState.isActive) {
-        this.dndState.isActive = true
-      }
-
+      if (!this.dndState.isActive) this.dndState.isActive = true
```
taoeffect: This can be simply:

```js
this.dndState.isActive = true
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1221,16 +1169,14 @@ export default ({
         // that'd require this entire function to be `async`, which could result
         // in race conditions as this is a watcher.
         Promise.resolve(sbp('chelonia/contract/isSyncing', toChatRoomId)).then((isSyncing) => {
-          // If the chatroom has changed since, return
-          if (this.summary.chatRoomID !== toChatRoomId) return
           if (isSyncing) {
-            toIsJoined && sbp('chelonia/queueInvocation', toChatRoomId, () => initAfterSynced(toChatRoomId))
+            sbp('chelonia/queueInvocation', initAfterSynced)
```
taoeffect: `'chelonia/queueInvocation'` requires at least two arguments... I'm guessing this is missing the first argument (the `contractID` that represents the queue you're adding to, i.e. `toChatRoomId`).

FYI, Chelonia maintains a queue for each `contractID`. Adding SBP invocations or functions to this queue means that those functions will be run after any currently-processing events or syncs on that `contractID`.

Additionally, the code doesn't need this `Promise.resolve(sbp('chelonia/contract/isSyncing', toChatRoomId))` check that @corrideat added AFAICT. It can simply immediately add `initAfterSynced` to the queue, and whether or not we're currently syncing, the function will be called afterward. If there's nothing in the queue it'll be called immediately.

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: FYI, you can avoid adding `catch` to all these calls to `processSwitchQueue()` by surrounding the entire contents of the `processSwitchQueue` implementation in a `try/catch` block.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1201,17 +1146,20 @@ export default ({
       const fromIsJoined = from.isJoined
 
       const initAfterSynced = (toChatRoomId) => {
-        if (toChatRoomId !== this.summary.chatRoomID || this.ephemeral.messagesInitiated) return
-        return this.setInitMessages()
+        if (toChatRoomId !== this.summary.chatRoomID || // If the user has switched to another chatroom during syncing, no need to process the chatroom that has been swithed away.
```
taoeffect: Could you please put this comment above if `if` on its own line?

frontend/views/containers/chatroom/ChatMain.vue
corrideat: Adding to what @taoeffect said in the last paragraph, unless Chelonia is involved (i.e., using a `contractID` and requiring the behaviour above), `okTurtles.eventQueue/queueEvent` should be used.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -857,8 +833,6 @@ export default ({
         events = await sbp('chelonia/out/eventsAfter', chatRoomID, sinceHeight, latestHeight - sinceHeight + 1, undefined, { stream: false })
       }
 
-      if (!this.checkEventSourceConsistency(chatRoomID)) return
```
taoeffect, line 860: I'm realizing that in `async` methods specifically, we still need these `checkEventSourceConsistency` type checks in some places (i.e. similar to how we had it before, either before or after a call to `await`) because these `await` calls represent a time-split where the chatroom could have been changed.

So your function `chatroomHasSwitched` could be moved from being defined inside of `rerenderEvents` and placed back as a method like `checkEventSourceConsistency`.

Also, Claude noticed that similar checks need to be done inside of `scrollToMessage` .. e.g. maybe after the call to `'chelonia/out/eventsBetween'` and `await this.rerenderEvents(events)`..

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,6 +1110,25 @@ export default ({
...
+    },
+    async processSwitchQueue () {
+      if (this.ephemeral.chatroomSwitchQueue.length === 0) return
+
+      // Take the most recent chatroom entry on the queue and discard everything else. This way we can speed up the process of switching chatrooms.
+      const targetChatroomId = this.ephemeral.chatroomSwitchQueue.pop()
+      this.ephemeral.chatroomSwitchQueue = []
+      this.ephemeral.renderingChatRoomId = targetChatroomId
+
+      await this.initializeState(true)
```
taoeffect: Claude says:

> If `initializeState` fails, it could leave the component in an inconsistent state

Could you please double-check this (both here, inside of `initializeState`, and in all other places that call `initializeState`)?

taoeffect (changes requested):
Great work @SebinSong! Review ready!

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @taoeffect @corrideat These are helpful comments

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: It sounds like, by `inconsistent state`, it meant some scenarios like below:
**1]** User enters a chatroom. `initializeState` is executed.
**2]** Before `initializeState` in **1]** is completed, the user switches to another chatroom.
**3**] For some reason `initializeState` in **1]** fails and it crashes the component, which leads to being unable to completing chatroom switch operation in **2]**

If it does mean above,I think adding `try .. catch` in `processSwitchQueue` (which is pointed out in another comment) would resolve this concern too.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -842,7 +826,8 @@ export default ({
         const shouldLoadMoreEvents = messageHashToScroll && this.messages.findIndex(msg => msg.hash === messageHashToScroll) < 0
         if (shouldLoadMoreEvents) {
           const { height: latestHeight } = await sbp('chelonia/out/latestHEADInfo', chatRoomID)
-          if (!this.checkEventSourceConsistency(chatRoomID)) return
+
+          if (this.chatroomHasSwitchedFrom(chatRoomID)) return false
```
SebinSong: Re-adding this check here fixes below error that occurs from time to time:

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: have created `this.chatroomHasSwitchedFrom(chatroomID)` method and added it to a few places as well as `scrollToMessage`

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: done

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: According to the comment:

```js
    /**
     * Load/render events for one or more pages
     * @return {boolean | undefined} The return value is used to change the state inside the function 'infiniteHandler'
     * true: Loaded all the messages, and no more messages exists
     * false: Messages are loaded, but more messages exist
     * undefined: Loaded the wrong events from the server so those events should be ignoreed
     *            and no state changes are needed (like messagesInitiated). Since the renderingChatRoomId could be changed
     *            while processing this function, this function could do the redundant process. This normally happens
     *            when user switches channels very fast.
    */
```

This should return `undefined` as before, not `false` (same for the check below)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1201,36 +1169,30 @@ export default ({
       const fromIsJoined = from.isJoined
 
       const initAfterSynced = (toChatRoomId) => {
-        if (toChatRoomId !== this.summary.chatRoomID || this.ephemeral.messagesInitiated) return
-        return this.setInitMessages()
+        // If the user has switched to another chatroom during syncing, no need to process the chatroom that has been swithed away.
```
taoeffect, line 1175: @SebinSong I'd like to see if we can simplify this further. This entire function is still rather too complicated, IMO.

The idea of the queue'd switching system is that we want to act on the most recently switched-to chatroom. Let's say a super-human was able to switch channels very quickly. Let's say they switched between 20 channels in 500msec. We don't really want to react to each channel switch. Just the very last one.

So this suggests a queue and the use of `throttle` call a method on the Vue component to process the queue periodically. Let's say we will process the queue within 100msec of switching.

I think this watcher needs to mainly do 2 things:

1. It needs to push the channel on the queue.
2. It needs to call `this.processSwitchQueue`, and that function should be defined using `debounce` from `giLodash.js`. Similar to how `refreshContent` was debounced before.

Some other problems I noticed with the code:

`initAfterSynced` has a few issues:

1. it takes a parameter `toChatRoomId` that overshadows the outside `toChatRoomId` parameter, which is never actually filled in with any value. This is a major bug, as `'chelonia/queueInvocation'` is certainly not going to fill it in with a value.
2. It returns if `this.ephemeral.messagesInitiated` is truthy... I'm not sure why it does this. It probably shouldn't do this.

Also, this chunk:

```js
        sbp('chelonia/contract/isSyncing', toChatRoomId).then(isSyncing => {
          if (isSyncing) {
            sbp('chelonia/queueInvocation', toChatRoomId, initAfterSynced)
          } else {
            this.processSwitchQueue()
          }
        })
```

Can be simplified to:

```js
sbp('chelonia/queueInvocation', toChatRoomId, initAfterSynced)
```

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -783,9 +767,11 @@ export default ({
     async initializeState (forceClearMessages = false) {
       // NOTE: this state is rendered using the chatroom contract functions
       //       so should be CAREFUL of updating the fields
-      const chatroomID = this.nonReactive.renderingChatRoomId
       const messageState = await this.generateNewChatRoomState(forceClearMessages)
-      if (!this.checkEventSourceConsistency(chatroomID)) return
+
+      // If user has since switched to another chatroom, no need to update 'this.messageState'
+      if (this.summary.chatRoomID !== this.ephemeral.renderingChatRoomId) return
```
taoeffect: For consistency's sake, should this use `this.chatroomHasSwitchedFrom` too? (check the rest of the code for similar inconsistencies)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -570,19 +556,18 @@ export default ({
...
             .catch((e) => {
               console.debug(`Error fetching events or message ${messageHash} doesn't belong to ${contractID}`, e)
             })
-        if (!this.checkEventSourceConsistency(contractID)) return
+
+        if (this.chatroomHasSwitchedFrom(contractID)) return
         if (events && events.length) {
           await this.rerenderEvents(events)
-
-          if (!this.checkEventSourceConsistency(contractID)) return
```
taoeffect, line 585: We should probably add this back...

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -595,13 +580,12 @@ export default ({
...
     },
     updateScroll (scrollTargetMessage = null, effect = false) {
-      const contractID = this.summary.chatRoomID
+      const contractID = this.ephemeral.renderingChatRoomId
       if (contractID) {
         return new Promise((resolve) => {
           // force conversation viewport to be at the bottom (most recent messages)
           setTimeout(async () => {
             if (scrollTargetMessage) {
-              if (!this.checkEventSourceConsistency(contractID)) return
```
taoeffect, line 604: And restore this one too

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -854,11 +839,11 @@ export default ({
         if (this.messages.length) {
           sinceHeight = Math.max(0, this.messages[0].height - limit)
         }
+
+        if (this.chatroomHasSwitchedFrom(chatRoomID)) return false
```
taoeffect: I think this should be moved below line 844 as it was before, and also just `return` as before (to return `undefined`)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1082,32 +1036,26 @@ export default ({
...
+      const targetChatroomID = this.ephemeral.renderingChatRoomId
       sbp('okTurtles.eventQueue/queueEvent', CHATROOM_EVENTS, async () => {
-        // NOTE: invocations in CHATROOM_EVENTS queue should run in synchronous
-        if (!this.checkEventSourceConsistency(chatRoomID)) return
+        if (this.chatroomHasSwitchedFrom(targetChatroomID)) return
 
+        // NOTE: invocations in CHATROOM_EVENTS queue should run in synchronous
         try {
           const completed = await this.renderMoreMessages()
-          if (!this.checkEventSourceConsistency(chatRoomID)) return
```
taoeffect, line 1102: This should probably be restored as well

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,6 +1121,37 @@ export default ({
...
+      try {
+        await this.initializeState(true)
+        if (this.ephemeral.chatroomSwitchQueue.length > 0) {
+          // If the user has since switched to another chatroom while initializing this chatroom, stop here
+          // and care about the switched chatroom.
+          return this.processSwitchQueue()
+        } else {
+          this.ephemeral.messagesInitiated = false
+          this.ephemeral.unprocessedEvents = []
+          this.ephemeral.infiniteLoading?.reset()
```
taoeffect, line 1147: Grok says:

> this.ephemeral.infiniteLoading?.reset() in processSwitchQueue assumes the infinite-loading component is mounted, but it might not be if the component hasn’t rendered yet.

And suggests:

```js
this.$nextTick(() => this.$refs['infinite-loading']?.reset())
```

I don't know if this is a good idea or not, will leave it up to you. /cc @pieer  if you have thoughts

taoeffect (changes requested):
@SebinSong Some more feedback left!

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: Will do. be noted that the large number of these were already removed in [this code](https://github.com/okTurtles/group-income/issues/2556#issuecomment-2683819280) btw..

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: This line used to be something like
```js
if (this..ephemeral.infiniteLoading) {
  this.ephemeral.infiniteLoading.reset()
}
```
Let me leave it as is, as this has been working fine like this.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @taoeffect In addition to all these, I believe we can also remove below block too:
https://github.com/okTurtles/group-income/blob/3d5a29ceed5e0cc690f676c4d260a2426d18829f/frontend/views/containers/chatroom/ChatMain.vue#L1232-L1234

Apparently this `if` block is for detecting when the user is viewing a channel that they haven't joined yet and then clicking this `join channel`button:



But then I noticed `this.setInitMessages()` triggered by this action does nothing because it's blocked by this check (As user is already in the chatroom):
https://github.com/okTurtles/group-income/blob/3d5a29ceed5e0cc690f676c4d260a2426d18829f/frontend/views/containers/chatroom/ChatMain.vue#L895-L897 

So, let me remove this block too.

SebinSong:
PR ready for review again.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,7 +1126,41 @@ export default ({
...
+            heading: L('Failed to initialize chatroom'),
+            question: L('Error details: {reportError}', LError(e)),
+            primaryButton: L('Close')
+          })
+        }
+      }
+    },
+    processSwitchQueue: debounce(function () {
+      this._processSwitchQueue()
+    }, 250)
```
taoeffect: There's no reason to create a separate `_processSwitchQueue` function, you can put its contents inside here.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1007,7 +967,7 @@ export default ({
         // This ensures that `this.latestEvents.push(serializedMessage)` below
         // happens in order
         sbp('okTurtles.eventQueue/queueEvent', CHATROOM_EVENTS, async () => {
-          if (!this.checkEventSourceConsistency(contractID)) return
+          if (contractID !== this.ephemeral.renderingChatRoomId) return
```
taoeffect: Unless I'm mistaken, I think this could use `chatroomHasSwitchedFrom` for consistency too?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -944,23 +916,11 @@ export default ({
...
-      // for the current contract, which is needed for signature verification
-      // when calling processMessage.
-      // The watch is setup for this.summary and not for this.currentChatRoomId,
-      // which is why this check must also check for this.summary.chatRoomID
-      if (!this.checkEventSourceConsistency(contractID)) return
-
-      if (message) {
-        this.ephemeral.unprocessedEvents.push(message)
-      }
+      if (contractID !== this.ephemeral.renderingChatRoomId) return
```
taoeffect: And maybe here as well

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: Right. updated.

taoeffect (approved):
Great work @SebinSong! I think we've managed to improve the logic of ChatMain to make it a bit easier to think about how it works and especially how chatrooms are switched. Approved! 😄

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,7 +1126,38 @@ export default ({
...
           this.$refs.sendArea.fileAttachmentHandler(e?.dataTransfer.files, true)
       }
-    }
+    },
+    processSwitchQueue: debounce(async function () {
+      if (this.ephemeral.chatroomSwitchQueue.length === 0) return
+
+      // Take the most recent chatroom entry on the queue and discard everything else. This way we can speed up the process of switching chatrooms.
+      const targetChatroomId = this.ephemeral.chatroomSwitchQueue.pop()
+      this.ephemeral.chatroomSwitchQueue = []
```
corrideat, line 1135: I'm not sure why an array is needed here, when the array-property isn't really used.

taoeffect: Its theoretically possible for there to be multiple elements in the list if the chatroom switched really quickly

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,7 +1126,38 @@ export default ({
         e?.dataTransfer.files?.length &&
           this.$refs.sendArea.fileAttachmentHandler(e?.dataTransfer.files, true)
       }
-    }
+    },
+    processSwitchQueue: debounce(async function () {
+      if (this.ephemeral.chatroomSwitchQueue.length === 0) return
+
+      // Take the most recent chatroom entry on the queue and discard everything else. This way we can speed up the process of switching chatrooms.
+      const targetChatroomId = this.ephemeral.chatroomSwitchQueue.pop()
```
corrideat, line 1134: I wonder if maybe this couldn't be:

```js
const targetChatroomId = new String(this.ephemeral.chatroomSwitchQueue.pop())
```

This way, we can tell apart switching back to a chatroom from never having switched at all. At this point, this just aspirational.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -414,10 +403,7 @@ export default ({
       const sendMessage = (beforePrePublish) => {
         let pendingMessageHash = null
         const beforeRequest = (message, oldMessage) => {
-          if (!this.checkEventSourceConsistency(contractID)) return
           sbp('okTurtles.eventQueue/queueEvent', CHATROOM_EVENTS, async () => {
-            if (!this.checkEventSourceConsistency(contractID)) return
-
```
corrideat, line 420: Are you sure that these two checks should be removed? If we've switched, we'll try to do things that are unnecessary.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1184,7 +1126,38 @@ export default ({
...
+    processSwitchQueue: debounce(async function () {
+      if (this.ephemeral.chatroomSwitchQueue.length === 0) return
+
+      // Take the most recent chatroom entry on the queue and discard everything else. This way we can speed up the process of switching chatrooms.
+      const targetChatroomId = this.ephemeral.chatroomSwitchQueue.pop()
+      this.ephemeral.chatroomSwitchQueue = []
+      this.ephemeral.renderingChatRoomId = targetChatroomId
+
+      try {
+        await this.initializeState(true)
```
taoeffect, line 1139: After releasing v1.2.16 with these changes, I'm noticing that chatrooms are loading a bit more slowly than before upon switching between them.

I was trying to figure out what change could be causing this, and noticed that previously this call was called using `this.initializeState()` in `setInitMessages`.

I'm wondering if this could be the cause of the slowness. @SebinSong, do you know why `true` is now being passed in? Does it need to be passed in? Or would not passing in fix the slowness I'm seeing?

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: I don't see what this accomplishes... I think it's OK because we have checks elsewhere to ensure that `processSwitchQueue` is only called when the channel switches

PR #2702, opened 2025-03-06
Follow-up PR for #2556

SebinSong:


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1136,11 +1139,11 @@ export default ({
       this.ephemeral.renderingChatRoomId = targetChatroomId
 
       try {
-        await this.initializeState(true)
+        await this.initializeState()
         if (this.ephemeral.chatroomSwitchQueue.length > 0) {
           // If the user has since switched to another chatroom while initializing this chatroom, stop here
           // and care about the switched chatroom.
-          return this.processSwitchQueue()
```
taoeffect, line 1143: Why was this `return` removed? I think it makes sense to have it here personally, as it makes clear that we should not proceed with anything else.

SebinSong: @taoeffect While running cypress, occasionally this line creates this error saying `TypeError: Chaining cycle detected for promise`. removing return fixes this.

taoeffect: Ah I think I saw that error too. Makes sense. Could you add a comment explaining that above? I feel like it's something we should try to remember in case we ever revisit this code and someone thinks to themselves, "Hmm, I should probably add a `return` here, derp derp"

taoeffect (approved):
Perfect!

PR #2704, opened 2025-03-07
Error handling improvements

### LINKED ISSUE #2663: Handle "Missing pubsub instance" push setup error
### Problem

While starting up the app on mobile PWA, I encountered this:

Image

📎 [gi_logs(missing pubsub).json.txt.zip](https://github.com/user-attachments/files/18932241/gi_logs.missing.pubsub.json.txt.zip)

### Solution

Analyze the logs for what happened and fix this.

It's very strange that this seems to have happened as part of the `ONLINE` handler:

```
    {
      "timestamp": "2025-02-23T05:55:43.427Z",
      "source": "browser",
      "type": "error",
      "msg": [
        "came back online, tried to report push subscription, but got:",
        {
          "name": "Error",
          "message": "Missing pubsub instance",
          "stack": "Error: Missing pubsub instance\n    at Object.<anonymous> (https://groupincome.app/assets/js/sw-primary.js?routerBase=%2Fapp&standalone=1:34:70596)\n    at ds (https://groupincome.app/assets/js/sw-primary.js?routerBase=%2Fapp&standalone=1:3:76745)\n    at https://groupincome.app/assets/js/sw-primary.js?routerBase=%2Fapp&standalone=1:34:78703\n    at https://groupincome.app/assets/js/sw-primary.js?routerBase=%2Fapp&standalone=1:34:78725"
        }
      ]
    },
```

Note to self: #2662

corrideat:
May close #2663 (makes the error less likely by avoiding requesting VAPID keys and by adding more retries)

### DIFF
--- a/backend/push.js
+++ b/backend/push.js
@@ -228,6 +228,16 @@ export const pushServerActionhandlers: any = {
     if (applicationServerKey) {
       const ourVapidPublicKey = getVapidPublicKey()
       const theirVapidPublicKey = Buffer.from(applicationServerKey, 'base64').toString('base64url')
+      // If the applicationServerKey (VAPID) key doesn't match ours, something
+      // went wrong, possibly in the client. We don't proceed because we know in
+      // advance that we won't be able to use such a subscription (since the
+      // push endpoint will fail authentication and we'll receive a 401). So, we
+      // let the client know our key so that they can retry.
+      // NOTE: The VAPID key is different from the subscription information,
+      // which is different for each client. This is exclusively a check on the
+      // VAPID public key, which is a static value derived from
+      // `_private_immutable_vapid_key`, shared by all clients and used by the
+      // server to authenticate itself when connecting to push endpoints.
       if (ourVapidPublicKey !== theirVapidPublicKey) {
         socket.send(createMessage(REQUEST_TYPE.PUSH_ACTION, { type: PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY, data: getVapidPublicKey() }))
         console.warn({ ourVapidPublicKey, theirVapidPublicKey }, 'Refusing to store subscription because the associated public VAPID key does not match ours')
--- a/backend/push.js
+++ b/backend/push.js
@@ -224,18 +224,27 @@ export const pushServerActionhandlers: any = {
   async [PUSH_SERVER_ACTION_TYPE.STORE_SUBSCRIPTION] (payload) {
     const socket = this
     const { server } = socket
-    const subscription = payload
+    const { applicationServerKey, subscriptionInfo } = payload
+    if (applicationServerKey) {
+      const ourVapidPublicKey = getVapidPublicKey()
+      const theirVapidPublicKey = Buffer.from(applicationServerKey, 'base64').toString('base64url')
+      if (ourVapidPublicKey !== theirVapidPublicKey) {
+        socket.send(createMessage(REQUEST_TYPE.PUSH_ACTION, { type: PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY, data: getVapidPublicKey() }))
+        console.warn({ ourVapidPublicKey, theirVapidPublicKey }, 'Refusing to store subscription because the associated public VAPID key does not match ours')
+        return
+      }
+    }
     let subscriptionId = null
     let host = ''
     let subscriptionWrapper = null
     try {
-      subscriptionId = await getSubscriptionId(subscription)
+      subscriptionId = await getSubscriptionId(subscriptionInfo)
       subscriptionWrapper = server.pushSubscriptions[subscriptionId]
 
       if (!subscriptionWrapper) {
-        console.debug(`saving new push subscription '${subscriptionId}':`, subscription)
+        console.debug(`saving new push subscription '${subscriptionId}':`, subscriptionInfo)
         // If this is a new subscription, we call `subscriptionInfoWrapper` and store it in memory.
-        server.pushSubscriptions[subscriptionId] = subscriptionInfoWrapper(subscriptionId, subscription)
+        server.pushSubscriptions[subscriptionId] = subscriptionInfoWrapper(subscriptionId, subscriptionInfo)
         subscriptionWrapper = server.pushSubscriptions[subscriptionId]
         host = subscriptionWrapper.endpoint.host
         await addSubscriptionToIndex(subscriptionId)
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -10,38 +10,6 @@ import { SPMessage } from '~/shared/domains/chelonia/SPMessage.js'
 import { Secret } from '~/shared/domains/chelonia/Secret.js'
 import { getSubscriptionId } from '~/shared/functions.js'
 
-// The application server (public) key could be either an ArrayBuffer (which is
-// what we get from fetching the current subscription), or it could be a
-// base64url-encoded string (which is what we usually get from the server)
-// This string coerces both of these into an `Uint8Array` instance.
-const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer) => {
-  if (typeof v === 'string') {
-    v = Buffer.from(
-      // Convert from base64url to base64
-      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - v.length % 4) % 4),
-      'base64'
-    )
-    return v
-  }
-  return new Uint8Array(v)
-}
-
-const bufferEq = (a?: ArrayBuffer | Uint8Array, b?: ArrayBuffer | Uint8Array) => {
-  // eslint-disable-next-line eqeqeq
-  if (a == null || b == null) return a == b
-
-  const ab = strOrBufToBuf(a)
-  const bb = strOrBufToBuf(b)
-
-  if (ab.byteLength !== bb.byteLength) return false
-
-  for (let i = ab.byteLength - 1; i >= 0; i--) {
-    if (ab[i] !== bb[i]) return false
-  }
-
-  return true
-}
-
 const pwa = {
   deferredInstallPrompt: null,
   installed: false
@@ -261,20 +229,11 @@ sbp('sbp/selectors/register', {
         let subID = null
         // get a real push subscription only if both browser permissions allow and user wants us to
         if (notificationEnabled && granted) {
-          const subscriptionOptions = await sbp('push/getSubscriptionOptions')
           subscription = await registration.pushManager.getSubscription()
-          if (subscription && !bufferEq(subscription.options.applicationServerKey, subscriptionOptions?.applicationServerKey)) {
-            // This is a public key that belongs to the server
-            console.warn('VAPID server key changed; removing existing subscription and setting up a new one', {
-              oldApplicationServerPublicKey: subscription.options.applicationServerKey,
-              newApplicationServerPublicKey: subscriptionOptions.applicationServerKey
-            })
-            await subscription.unsubscribe()
-            subscription = null
-          }
           let newSub = false
           let endpoint = null
           if (!subscription || (subscription.expirationTime != null && subscription.expirationTime <= Date.now())) {
+            const subscriptionOptions = await sbp('push/getSubscriptionOptions')
             subscription = await registration.pushManager.subscribe(subscriptionOptions)
             newSub = true
           }
@@ -285,24 +244,12 @@ sbp('sbp/selectors/register', {
           console.info(`[service-worker/setup-push-subscription] got ${newSub ? 'new' : 'existing'} subscription '${subID}':`, endpoint)
         }
         console.info('[service-worker/setup-push-subscription] calling push/reportExistingSubscription on:', subID)
-        await sbp('push/reportExistingSubscription', subscription?.toJSON())
+        await sbp('push/reportExistingSubscription', subscription?.toJSON(), subscription?.options.applicationServerKey)
       } catch (e) {
         console.error('[service-worker/setup-push-subscription] error getting a subscription:', e)
-        if (e?.message === 'WebSocket connection is not open') {
-          if (attemptNumber >= 3) {
-            console.error('[service-worker/setup-push-subscription] maxAttempts reached, giving up')
-            // NOTE: we do not need to setup an ONLINE event listener here because one is already setup in main.js
-            throw e // give up
-          }
-          // this outer promise is a way to wait on this sub-call to finish without getting the eventQueue stuck
-          retryAttemptsPromise = new Promise((resolve, reject) => {
-            // try again in 1 second
-            setTimeout(() => {
-              sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
-            }, 1e3)
-          })
-          return // exit the event queue immediately and do not rethrow
-        } else {
+
+        if (attemptNumber >= 12) {
+          console.error('[service-worker/setup-push-subscription] maxAttempts reached, giving up')
           sbp('gi.ui/prompt', {
             heading: L('Error setting up push notifications'),
             question: L('Error setting up push notifications: {errMsg}{br_}{br_}Please make sure {a_}push services are enabled{_a} in your Browser settings, and then try reloading the app and toggling the push notifications toggle in the Notifications user settings.', {
@@ -313,8 +260,16 @@ sbp('sbp/selectors/register', {
             }),
             primaryButton: L('Close')
           })
+          // NOTE: we do not need to setup an ONLINE event listener here because one is already setup in main.js
+          throw e // give up
         }
-        throw e
+        // this outer promise is a way to wait on this sub-call to finish without getting the eventQueue stuck
+        retryAttemptsPromise = new Promise((resolve, reject) => {
+          // try again in 1 second
+          setTimeout(() => {
+            sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
+          }, 12e3)
+        })
       }
     })
     if (retryAttemptsPromise) {
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -265,10 +265,10 @@ sbp('sbp/selectors/register', {
         }
         // this outer promise is a way to wait on this sub-call to finish without getting the eventQueue stuck
         retryAttemptsPromise = new Promise((resolve, reject) => {
-          // try again in 1 second
+          // try again in 5 seconds
           setTimeout(() => {
             sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
-          }, 12e3)
+          }, 5e3)
         })
       }
     })
--- a/frontend/controller/serviceworkers/push.js
+++ b/frontend/controller/serviceworkers/push.js
@@ -120,7 +120,7 @@ export default (sbp('sbp/selectors/register', {
   //      to update the existing push subscription and replace it with a new
   //      one.
   'push/reportExistingSubscription': (() => {
-    const map = new WeakMap()
+    const reportedSubscriptionBySocket = new WeakMap()
     async function getSubID (subscription) {
       try {
         return await getSubscriptionId(subscription)
@@ -139,8 +139,8 @@ export default (sbp('sbp/selectors/register', {
       }
 
       const socket = pubsub.socket
-      const reported = map.get(socket)
-      map.set(socket, subscriptionInfo)
+      const reported = reportedSubscriptionBySocket.get(socket)
+      reportedSubscriptionBySocket.set(socket, subscriptionInfo)
       if (subscriptionInfo?.endpoint) {
         if (!reported || subscriptionInfo.endpoint !== reported.endpoint) {
           const subID = await getSubID(subscriptionInfo)
--- a/frontend/controller/serviceworkers/push.js
+++ b/frontend/controller/serviceworkers/push.js
@@ -59,6 +59,8 @@ export default (sbp('sbp/selectors/register', {
           oldApplicationServerPublicKey: subscription.options.applicationServerKey && Buffer.from(subscription.options.applicationServerKey).toString('base64'),
           newApplicationServerPublicKey: data.data
         })
+        // If unsubscribe fails, it doesn't make much sense to proceed, as we
+        // can't really create new subscription
         await subscription.unsubscribe()
         // return b/c device settings not loaded
         if (!sbp('chelonia/rootState').loggedIn || sbp('sw/deviceSettings/get', DEVICE_SETTINGS.DISABLE_NOTIFICATIONS)) return
@@ -77,7 +79,7 @@ export default (sbp('sbp/selectors/register', {
         return Promise.resolve(cachedVapidInformation[1])
       }
 
-      const result = new Promise((resolve, reject) => {
+      return new Promise((resolve, reject) => {
         const handler = ({ data }) => {
           if (data.type !== PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY) return
           sbp('okTurtles.events/off', REQUEST_TYPE.PUSH_ACTION, handler)
@@ -104,7 +106,6 @@ export default (sbp('sbp/selectors/register', {
           { action: PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY }
         ))
       })
-      return result
     }
   })(),
   // This function reports the existing push subscription to the server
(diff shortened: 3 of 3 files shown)


frontend/controller/serviceworkers/push.js
```diff
@@ -97,7 +149,18 @@ export default (sbp('sbp/selectors/register', {
...
-            { action: PUSH_SERVER_ACTION_TYPE.STORE_SUBSCRIPTION, payload: subscriptionInfo }
+            {
+              action: PUSH_SERVER_ACTION_TYPE.STORE_SUBSCRIPTION,
+              payload: {
+                applicationServerKey: applicationServerKey
+                  ? Buffer.from(applicationServerKey).toString('base64')
+                  : null,
+                settings: {
+                  heartbeatInterval: 12 * 60 * 60 * 1000
+                },
```
taoeffect, line 161: If I'm not mistaken, this requires #2546 to be merged first for it to work, right?

corrideat: This is here to reduce future merge conflicts. It won't do anything in this PR.

taoeffect (commented):
# `o3-mini-high` review

<details>
Below are some detailed findings and suggestions. In several cases I’m quoting the specific lines (or excerpts) that may be problematic or confusing, and then I’m providing a complete code snippet for how you might improve them.

---

## 1. VAPID Key Encoding Consistency (backend/push.js)

In the new STORE_SUBSCRIPTION handler you convert the incoming key like this:

```js
const theirVapidPublicKey = Buffer.from(applicationServerKey, 'base64').toString('base64url')
if (ourVapidPublicKey !== theirVapidPublicKey) {
  socket.send(createMessage(REQUEST_TYPE.PUSH_ACTION, { type: PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY, data: getVapidPublicKey() }))
  console.warn({ ourVapidPublicKey, theirVapidPublicKey }, 'Refusing to store subscription because the associated public VAPID key does not match ours')
  return
}
```

A couple of points to double‑check:

• Make sure that on the client side the public VAPID key is being sent (or stored) in a format that, when re‑encoded to base64url here, is consistent with what `getVapidPublicKey()` returns.

• If your client code sends the key as base64 (as in some later code in push/reportExistingSubscription), you may end up with a mismatch (base64 versus base64url). In that case it could be clearer to convert both the server’s expected key and the provided key using the same normalization. For example, you might define a helper:

```js
const normalizeVapidKey = (keyString) => {
  // Converts a base64 or base64url string into base64url format
  return Buffer.from(keyString, 'base64').toString('base64url')
}
```

Then your check becomes:

```js
const theirVapidPublicKey = normalizeVapidKey(applicationServerKey)
if (getVapidPublicKey() !== theirVapidPublicKey) {
  // …
}
```

This makes it explicit that the same conversion is used on both sides. (Of course, be sure that your client-side code uses the same convention.)

---

## 2. Re‑subscription in pushsubscriptionchange (frontend/controller/service-worker.js)

In your pushsubscriptionchange event you currently re‑subscribe using the old subscription’s options:

```js
if (event.oldSubscription) {
  subscription = await self.registration.pushManager.subscribe(event.oldSubscription.options)
}
await sbp('push/reportExistingSubscription', subscription?.toJSON(), subscription?.options.applicationServerKey)
```

Because the old subscription’s options may be outdated, it is more robust to obtain fresh options—just like in your `setup-push-subscription` code. For example, you might change it to:

```diff
self.addEventListener('pushsubscriptionchange', function (event) {
  // NOTE: Currently there is no specific way to validate if a push-subscription is valid. So it has to be handled in the front-end.
  event.waitUntil((async () => {
    try {
      let subscription = null
      if (event.oldSubscription) {
-        subscription = await self.registration.pushManager.subscribe(event.oldSubscription.options)
+        const subscriptionOptions = await sbp('push/getSubscriptionOptions')
+        subscription = await self.registration.pushManager.subscribe(subscriptionOptions)
      }
      await sbp('push/reportExistingSubscription', subscription?.toJSON(), subscription?.options.applicationServerKey)
    } catch (e) {
      console.error('[pushsubscriptionchange] Error resubscribing:', e)
    }
  })())
}, false)
```

This ensures that you always subscribe with the up‑to‑date options.

---

## 3. Magic Numbers for Retry Delays (frontend/controller/service-worker.js)

Within your setup‑push‑subscription handler you use a hard‑coded delay value when retrying:

```js
setTimeout(() => {
  sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
}, 12e3)
```

For clarity and to make future adjustments easier, consider defining a constant for the retry delay, for example:

```js
const RETRY_DELAY_MS = 12000
```

Then update the code:

```diff
-          setTimeout(() => {
-            sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
-          }, 12e3)
+          setTimeout(() => {
+            sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
+          }, RETRY_DELAY_MS)
```

This small refactoring improves maintainability.

---

## 4. Utility Function Duplication in Service Worker Push (frontend/controller/serviceworkers/push.js)

You’ve re‑introduced two functions (strOrBufToBuf and bufferEq) that convert strings/buffers and compare them:

```js
const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer) => {
  if (typeof v === 'string') {
    v = Buffer.from(
      // Convert from base64url to base64
      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - v.length % 4) % 4),
      'base64'
    )
    return v
  }
  return new Uint8Array(v)
}

const bufferEq = (a?: ArrayBuffer | Uint8Array, b?: ArrayBuffer | Uint8Array) => {
  // eslint-disable-next-line eqeqeq
  if (a == null || b == null) return a == b

  const ab = strOrBufToBuf(a)
  const bb = strOrBufToBuf(b)

  if (ab.byteLength !== bb.byteLength) return false

  for (let i = ab.byteLength - 1; i >= 0; i--) {
    if (ab[i] !== bb[i]) return false
  }

  return true
}
```

Because similar functions were recently removed from frontend/controller/service-worker.js, consider moving these helpers into a shared utility (e.g. in a file like utils/cryptoUtils.js). This way you reduce code duplication and avoid potential divergence later.

For example, you might create a file:

```js
// utils/cryptoUtils.js
// @flow

export const strOrBufToBuf = (v: string | Uint8Array | ArrayBuffer): Uint8Array => {
  if (typeof v === 'string') {
    return Buffer.from(
      // Convert from base64url to base64
      v.replace(/_/g, '/').replace(/-/g, '+') + '='.repeat((4 - (v.length % 4)) % 4),
      'base64'
    )
  }
  return new Uint8Array(v)
}

export const bufferEq = (
  a?: ArrayBuffer | Uint8Array,
  b?: ArrayBuffer | Uint8Array
): boolean => {
  if (a == null || b == null) return a == b

  const ab = strOrBufToBuf(a)
  const bb = strOrBufToBuf(b)

  if (ab.byteLength !== bb.byteLength) return false

  for (let i = ab.byteLength - 1; i >= 0; i--) {
    if (ab[i] !== bb[i]) return false
  }

  return true
}
```

Then import these functions where needed.

---

## 5. Clarify Variable Naming for the WeakMap (push/reportExistingSubscription)

Inside your push/reportExistingSubscription handler you use:

```js
const map = new WeakMap()
```

For clarity, you might rename the variable to something more descriptive like “reportedSubscriptionBySocket” or “subscriptionsCache” so that its purpose is immediately clear. For example:

```diff
-const map = new WeakMap()
+const reportedSubscriptionBySocket = new WeakMap()
```

...and then update its usage accordingly. This makes the code easier to read and maintain.
</details>

# `Gemini Pro 2.5` Review

<details>
Okay, I have reviewed the provided code changes. Here are my findings and suggestions:

### General Observations

The changes introduce a mechanism to handle VAPID public key mismatches between the client and the server. The client now requests the server's key, compares it with the key associated with its current push subscription, and updates the subscription if necessary. The backend also validates the key provided by the client when storing a subscription. This is a good enhancement for robustness. The retry logic in `service-worker/setup-push-subscription` has also been adjusted.

### Suggestions for Improvement

#### 1. `frontend/controller/service-worker.js`: Improve Retry Logic in `setup-push-subscription`

The retry logic in `service-worker/setup-push-subscription` has been changed from 3 attempts with a 1-second delay to 12 attempts with a 12-second delay. This means the function could potentially wait up to 144 seconds (2.4 minutes) before giving up. This seems quite long and uses a fixed delay. An exponential backoff strategy is generally preferred for retries as it increases the delay between attempts, reducing load during transient issues while still retrying persistently for longer outages.

```diff
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -104,25 +104,26 @@
       } catch (e) {
         console.error('[service-worker/setup-push-subscription] error getting a subscription:', e)
 
-        if (attemptNumber >= 12) {
+        // Exponential backoff: 2, 4, 8, 16, 32, 64 seconds (total ~2 mins)
+        const maxAttempts = 6
+        const delaySeconds = Math.pow(2, attemptNumber)
+        if (attemptNumber >= maxAttempts) {
           console.error('[service-worker/setup-push-subscription] maxAttempts reached, giving up')
           sbp('gi.ui/prompt', {
             heading: L('Error setting up push notifications'),
             question: L('Error setting up push notifications: {errMsg}{br_}{br_}Please make sure {a_}push services are enabled{_a} in your Browser settings, and then try reloading the app and toggling the push notifications toggle in the Notifications user settings.', {
               errMsg: e?.message,
               a_: '<a class="link" target="_blank" href="https://stackoverflow.com/a/69624651">',
               _a: '</a>',
-              br_: '<br/>'
+              br_: '<br />' // Use valid HTML for self-closing tag
             }),
             primaryButton: L('Close')
           })
-          // NOTE: we do not need to setup an ONLINE event listener here because one is already setup in main.js
           throw e // give up
         }
         // this outer promise is a way to wait on this sub-call to finish without getting the eventQueue stuck
         retryAttemptsPromise = new Promise((resolve, reject) => {
-          // try again in 1 second
+          console.warn(`[service-worker/setup-push-subscription] Retrying attempt ${attemptNumber + 1} in ${delaySeconds}s...`)
           setTimeout(() => {
             sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
-          }, 12e3)
+          }, delaySeconds * 1000)
         })
       }
     })

```
*Rationale:* This uses exponential backoff (2, 4, 8, 16, 32, 64 seconds), capping at 6 attempts (~2 minutes total wait time). It also slightly improves the HTML in the error message (`<br/>` to `<br />`).

#### 2. `frontend/controller/serviceworkers/push.js`: Potential Race Condition on VAPID Key Update

In the event handler added to `push/getSubscriptionOptions`, when a new VAPID key is received and differs from the current subscription's key, the code unsubscribes, checks login/notification state, and then potentially re-subscribes.

```javascript
// frontend/controller/serviceworkers/push.js Lines 48-59
       (async () => {
         const subscription = await self.registration.pushManager.getSubscription()
         if (!subscription) return
         if (bufferEq(subscription.options.applicationServerKey, data.data)) return
         console.warn('VAPID server key changed; removing existing subscription and setting up a new one', {
           oldApplicationServerPublicKey: subscription.options.applicationServerKey && Buffer.from(subscription.options.applicationServerKey).toString('base64'),
           newApplicationServerPublicKey: data.data
         })
         await subscription.unsubscribe()
         // return b/c device settings not loaded
         if (!sbp('chelonia/rootState').loggedIn || sbp('sw/deviceSettings/get', DEVICE_SETTINGS.DISABLE_NOTIFICATIONS)) return
         const newSubscription = await self.registration.pushManager.subscribe(cachedVapidInformation[1])
         await sbp('push/reportExistingSubscription', newSubscription?.toJSON(), newSubscription?.options.applicationServerKey)
       })()
```

While discussed during internal thought, and deemed likely acceptable, there's a small window where the browser might have no subscription if the user is logged out or has disabled notifications precisely when the key update arrives. If ensuring the browser *always* has the latest valid subscription (even if not immediately reported) is critical, you could modify the logic to always re-subscribe after unsubscribing due to a key mismatch, and only gate the *reporting* step based on login/settings state.

However, the current approach (only re-subscribing if logged in and enabled) is arguably safer as it avoids creating subscriptions that cannot be immediately reported to the backend. Stick with the current logic unless the potential brief lack of subscription is found to be problematic. No code change suggested here unless requirements dictate otherwise.

</details>

frontend/controller/serviceworkers/push.js
```diff
@@ -7,28 +7,83 @@ import { NOTIFICATION_TYPE, PUBSUB_RECONNECTION_SUCCEEDED, PUSH_SERVER_ACTION_TY
...
+  }
+
+  return true
+}
+
 export default (sbp('sbp/selectors/register', {
   'push/getSubscriptionOptions': (() => {
     let cachedVapidInformation
+
+    sbp('okTurtles.events/on', REQUEST_TYPE.PUSH_ACTION, ({ data }) => {
```
taoeffect, line 46: `'push/getSubscriptionOptions'` gets called multiple times, and each time it will set this event listener. This event listener is never removed, so there will be multiple event listeners running.

corrideat: I don't think so, note that this is an IIFE and the actual `push/getSubscriptionOptions` is defined later.

corrideat:
@taoeffect Ready to review. I didn't see many actionable items in the review.

frontend/controller/serviceworkers/push.js
```diff
@@ -7,28 +7,83 @@ import { NOTIFICATION_TYPE, PUBSUB_RECONNECTION_SUCCEEDED, PUSH_SERVER_ACTION_TY
...
+      if (oldKey === data.data) return
+      (async () => {
+        const subscription = await self.registration.pushManager.getSubscription()
+        if (!subscription) return
+        if (bufferEq(subscription.options.applicationServerKey, data.data)) return
+        console.warn('VAPID server key changed; removing existing subscription and setting up a new one', {
+          oldApplicationServerPublicKey: subscription.options.applicationServerKey && Buffer.from(subscription.options.applicationServerKey).toString('base64'),
+          newApplicationServerPublicKey: data.data
+        })
+        await subscription.unsubscribe()
```
taoeffect, line 64: Should we not proceed even if this throws? 🤔 

Seems like we should still try to proceed if this throws...

frontend/controller/serviceworkers/push.js
```diff
@@ -7,28 +7,83 @@ import { NOTIFICATION_TYPE, PUBSUB_RECONNECTION_SUCCEEDED, PUSH_SERVER_ACTION_TY
...
+        // return b/c device settings not loaded
+        if (!sbp('chelonia/rootState').loggedIn || sbp('sw/deviceSettings/get', DEVICE_SETTINGS.DISABLE_NOTIFICATIONS)) return
```
taoeffect, line 66: Will we get another opportunity to re-subscribe? Can the comment at least be updated to comment on this?

frontend/controller/serviceworkers/push.js
```diff
@@ -7,28 +7,83 @@ import { NOTIFICATION_TYPE, PUBSUB_RECONNECTION_SUCCEEDED, PUSH_SERVER_ACTION_TY
...
+
     return () => {
       if (
         cachedVapidInformation &&
         // Cache the VAPID information for one hour. The server public
         // information should change very infrequently, if it changes at all.
         (performance.now() - cachedVapidInformation[0]) < 3600e3
       ) {
-        return cachedVapidInformation[1]
+        return Promise.resolve(cachedVapidInformation[1])
```
taoeffect, line 79: On line 80 below, you cache `result` as if you're going to do something with it but nothing is done. Code can be simplified to immediately `return new Promise`

frontend/controller/serviceworkers/push.js
```diff
@@ -7,28 +7,83 @@ import { NOTIFICATION_TYPE, PUBSUB_RECONNECTION_SUCCEEDED, PUSH_SERVER_ACTION_TY
...
         const handler = ({ data }) => {
           if (data.type !== PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY) return
           sbp('okTurtles.events/off', REQUEST_TYPE.PUSH_ACTION, handler)
           clearTimeout(timeoutId)
-          resolve({
-            userVisibleOnly: true,
-            applicationServerKey: data.data
-          })
+          // Set in the handler above
+          resolve(cachedVapidInformation[1])
```
taoeffect, line 88: This code seems somewhat brittle to me. It says `Set in the handler above`, and that is technically true... and it is technically true that that code will also run first because that handler was registered first.... but if the order were different this could resolve with `undefined` or even throw an exception (`undefined[1]`)

backend/push.js
```diff
@@ -224,18 +224,27 @@ export const pushServerActionhandlers: any = {
...
-    const subscription = payload
+    const { applicationServerKey, subscriptionInfo } = payload
+    if (applicationServerKey) {
+      const ourVapidPublicKey = getVapidPublicKey()
+      const theirVapidPublicKey = Buffer.from(applicationServerKey, 'base64').toString('base64url')
+      if (ourVapidPublicKey !== theirVapidPublicKey) {
+        socket.send(createMessage(REQUEST_TYPE.PUSH_ACTION, { type: PUSH_SERVER_ACTION_TYPE.SEND_PUBLIC_KEY, data: getVapidPublicKey() }))
+        console.warn({ ourVapidPublicKey, theirVapidPublicKey }, 'Refusing to store subscription because the associated public VAPID key does not match ours')
+        return
+      }
```
taoeffect: Could there be some sort of infinite or semi-infinite loop here on initial subscribe?

1. `'push/reportExistingSubscription'` is called by `'service-worker/setup-push-subscription'`
2. Server sees the vapid keys don't match here (because they're being initially setup... ?)
3. Sends PUSH_ACTION+SEND_PUBLIC_KEY
4. Handler in `'push/getSubscriptionOptions'` gets called
5. Calls `'push/reportExistingSubscription'`... repeat

If not, could you add a comment here explaining why no loop will occur?

In general I'm not a big fan of this approach because of this... it seems rather weird and confusing an unnatural that attempting to store the subscription won't store the subscription in some circumstances and will instead trigger this message back and who knows that will happen from there...

frontend/controller/service-worker.js
```diff
@@ -313,8 +260,16 @@ sbp('sbp/selectors/register', {
...
+          // NOTE: we do not need to setup an ONLINE event listener here because one is already setup in main.js
+          throw e // give up
         }
-        throw e
+        // this outer promise is a way to wait on this sub-call to finish without getting the eventQueue stuck
+        retryAttemptsPromise = new Promise((resolve, reject) => {
+          // try again in 1 second
+          setTimeout(() => {
+            sbp('service-worker/setup-push-subscription', attemptNumber + 1).then(resolve).catch(reject)
+          }, 12e3)
```
taoeffect: The comment says 1 second yet the wait is 12 seconds.

frontend/controller/service-worker.js
```diff
@@ -285,24 +244,12 @@ sbp('sbp/selectors/register', {
           console.info(`[service-worker/setup-push-subscription] got ${newSub ? 'new' : 'existing'} subscription '${subID}':`, endpoint)
         }
         console.info('[service-worker/setup-push-subscription] calling push/reportExistingSubscription on:', subID)
-        await sbp('push/reportExistingSubscription', subscription?.toJSON())
+        await sbp('push/reportExistingSubscription', subscription?.toJSON(), subscription?.options.applicationServerKey)
       } catch (e) {
         console.error('[service-worker/setup-push-subscription] error getting a subscription:', e)
-        if (e?.message === 'WebSocket connection is not open') {
```
taoeffect, line 291: Do you remember the reasoning for choosing to retry on all errors?

taoeffect (changes requested):
Review ready!

taoeffect:
General comment: I think this PR closes #2663 more so by adding the greater number of retries than it does with the VAPID related changes... I'm not a huge an of the VAPID related changes because I think they confuse and complicate the codebase. Are these changes really needed?

frontend/controller/serviceworkers/push.js
corrideat: I don't think so here. Proceed and do what, exactly?

We call unsubscribe to remove the existing subscription and then the rest of the logic is to create a new subscription. If unsubscribing fails, we'll get an error like `DOMException: A subscription with a different application server key already exists.`

Since the rest of the code only makes sense after unsubscribing, then I'd say this is fine as is.

frontend/controller/serviceworkers/push.js
corrideat: Yes, that'll happen when the selector this is part of is called. I can add a comment.

frontend/controller/serviceworkers/push.js
corrideat: Are you saying to change `const result = new Promise((resolve, reject) => {...} return result` to `return new Promise((resolve, reject) => { ... }`? Sure, if you feel that's simpler.

frontend/controller/serviceworkers/push.js
corrideat: But the order is not different..? We need to wait on the server response being returned, and we already have the handler above to handle that message, and so we don't need to repeat the same logic. I'm not sure how to keep the same semantics in a simpler way.

frontend/controller/serviceworkers/push.js
taoeffect: Yes 👍

frontend/controller/serviceworkers/push.js
taoeffect: Maybe a comment acknowledging this reality would be helpful to readers who aren't familiar with how `okTurtles.events` works, to let them know that this code only works because of the order in which handlers were registered as `okTurtles.events` loops through event handlers in that order when emitting events.

backend/push.js
corrideat: > it seems rather weird and confusing an unnatural that attempting to store the subscription won't store the subscription

How is this any different than attempting to register a username with invalid characters, for example? Then it'd fail and no username will be registered.

In this case, we don't proceed because storing such a subscription is a waste of server resources. We cannot use a subscription with a mismatched VAPID key.

I don't really understand your infinite loop scenario or why they wouldn't match on step 2 or what being initially setup has to do with this, but maybe there's some confusion, so let's go step-by-step.

`reportExistingSubsciption` will send this:

```js
            {
              action: PUSH_SERVER_ACTION_TYPE.STORE_SUBSCRIPTION,
              payload: {
                applicationServerKey: applicationServerKey
                  ? Buffer.from(applicationServerKey).toString('base64')
                  : null,
                settings: {
                  heartbeatInterval: 12 * 60 * 60 * 1000
                },
                subscriptionInfo
              }
            }
```

We can ignore `settings`, which does nothing in this PR, and we can also ignore `subscriptionInfo`, which is irrelevant for this check. We are left with `applicationServerKey`, which is a _server-provided value_, and isn't supposed to change in the first place (it's an immutable value). But if it did change (new server, admin manually reset the database, admin deleted the key, etc.), then the check will (correctly) fail, but the new key will take its place and then there wouldn't be an infinite loop, unless the server admin replaces this code to always say the `applicationServerKey` has changed.

If you're worried about this kind of behaviour by the server, we could debounce the handler on the client.

As I said, initial (or existing, for that matter) subscriptions, which would be in the `subscriptionInfo` parameter, have nothing to do with this check.

frontend/controller/service-worker.js
corrideat: https://okturtles.slack.com/archives/C0EH7P20Y/p1741305585842639?thread_ts=1741276573.013899&cid=C0EH7P20Y

frontend/controller/service-worker.js
corrideat: Yes, good catch. And also, I think this is supposed to be 5 seconds.

corrideat:
>  so by adding the greater number of retries than it does with the VAPID related changes...

Well, I wasn't able to reproduce this issue, so I tried to improve upon things to make the issue less likely to occur (I don't say fixed, because I'm not sure due to the inability to reproduce).

It's true that the 'less likely' part is covered with the larger number of attempts combined with re-attempting for all errors.

However, the VAPID part is motivated by the fact that it saves a roundtrip (on _every_ reconnection, so it's a lot of roundtrips saved), which is more efficient and also reduces the likelihood of errors happening (since each round trip presents an opportunity for something to fail). Also, since the VAPID key is supposed to rarely change, if it changes at all, an 'optimistic' approach sounds better.

In terms of complexity, I guess this is slightly more complex. Going by lines of code, this PR adds ~27 lines. Going by logic, the main difference is that now, instead of having to manually make a new request to get the VAPID key, we get it immediately upon failure, which is in line with the changes in other PRs, like #2742 (motivation https://github.com/okTurtles/group-income/issues/2739#issuecomment-2777147518).

BTW, this also addresses another but probably unlikely bug: If for some reason we have the wrong VAPID key and create a subscription with it, now (in `master`), we'll send it to the server, which will store it, etc., only to later find out that it can't be used (using it will result in a 401). This instead 'corrects' such accidents.

frontend/controller/serviceworkers/push.js
corrideat: Do you have a suggestion for improving this otherwise to make it less brittle while keeping it simple too? I noticed, for example, that it would break if `debounce` is used as suggested in https://github.com/okTurtles/group-income/pull/2704#discussion_r2040149550

taoeffect (approved):
Nice!

PR #2708, opened 2025-03-09
Fix #2705

### LINKED ISSUE #2705: `ChatMain.vue`: Fix mismatched contract ID
### Problem

When switching from one chatroom to another, the following error happens:

```
ChatMain infiniteHandler() error: Error: chelonia/out/eventsBetween: Mismatched contract ID
    at Object.start (sw-primary.js?routerBase=%2Fapp&standalone=0:19022:28)

(anonymous)	@	logger.js:13
(anonymous)	@	ChatMain.vue:1107
await in (anonymous) (async)		
infiniteHandler	@	ChatMain.vue:1055
Show 10 more frames
```

This may or may not be related to some of the changes made in #2696. With the changes from that PR, the issue seems to reproduce easily. Nothing appears to be broken other than this entry being logged.

This is an issue because:

1. It means that the contractID logic is incorrect (at least when `eventsBetween` is called, possibly in other places), and it could result in other errors
2. Making this call results in a completely unnecessary network request.

### Solution

Determine what causes this issue and fix it so that either this call isn't made (if it's not needed) or it is made with the correct contract ID (i.e., the ID consistent with the event IDs used).

corrideat:


SebinSong:
Just tested this PR locally with `GI_PERSIST=sqlite grunt dev`, it appears that this PR does fix the issue.

But since it replaces the chatroom ID queue, it pretty much reverts the work for the queue system Greg proposed in #2556. So probably discuss with him first on whether he thinks this change is okay too.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1131,16 +1131,17 @@ export default ({
       }
     },
     processSwitchQueue: debounce(async function () {
```
SebinSong: (If it's decided that the queue is not necessary)
The name of this function should be changed too. (probably something like `processChatroomSwitch`)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1131,16 +1131,17 @@ export default ({
...
-      if (this.ephemeral.chatroomSwitchQueue.length === 0) return
+      if (!this.ephemeral.targetChatroomId) return
 
-      // Take the most recent chatroom entry on the queue and discard everything else. This way we can speed up the process of switching chatrooms.
-      const targetChatroomId = this.ephemeral.chatroomSwitchQueue.pop()
-      this.ephemeral.chatroomSwitchQueue = []
+      const targetChatroomId = this.ephemeral.targetChatroomId
```
SebinSong: This can be DRYed. (no need to assign it to a variable with the same name)

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -253,7 +253,7 @@ export default ({
         unprocessedEvents: [],
 
         // Related to switching chatrooms
-        chatroomSwitchQueue: [],
+        targetChatroomId: null,
```
SebinSong: how about naming it something else like `chatroomIdToSwitch` to make the purpose more clear?

taoeffect:
> But since it replaces the chatroom ID queue, it pretty much reverts the work for the queue system Greg proposed in https://github.com/okTurtles/group-income/issues/2556. So probably discuss with him first on whether he thinks this change is okay too.

Thanks @SebinSong for pointing this out.

Looking at the changes here it seems that it is functionally equivalent to what we did with the queue idea. It still keeps most of your changes to the way that the `summary` watcher worked, and still keeps the debounced `processSwitchQueue` function, which I agree with you should be renamed.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1130,30 +1130,31 @@ export default ({
...
+      const targetChatroomId = this.ephemeral.chatroomIdToSwitchTo
+      this.ephemeral.chatroomIdToSwitchTo = null
+
+      if (targetChatroomId === this.ephemeral.renderingChatRoomId) return
       this.ephemeral.renderingChatRoomId = targetChatroomId
 
       try {
         await this.initializeState()
-        if (this.ephemeral.chatroomSwitchQueue.length > 0) {
+        if (this.ephemeral.chatroomIdToSwitchTo) {
```
taoeffect, line 1144: This condition will always be false because on line 1137 you set it to `null`

taoeffect (changes requested):
Great work @corrideat! Found what I think is one bug in this PR:

taoeffect (commented):
Oops, I was mistaken, it could change because of the `await`.

Nicely done @corrideat!

taoeffect (approved):

PR #2709, opened 2025-03-09
Fast-path login

### LINKED ISSUE #2700: Improve logging in process
### Problem

Currently, we require a full contract sync (for all contracts) when logging in from a new device. Depending on network conditions and contract size, this can take a long time, which may make the login process feel slow, and it can also make it unreliable (e.g., if syncing fails, logging in will fail). On mobile devices, users may need to manually keep their devices awoke for login to succeed.

There's no particular reason why we need to sync all contracts before deciding whether logging in should succeed or not. In fact, we don't block the entire application in other circumstances.

### Solution

A longer term solution (*) to this is implementing #354 (or possibly #2595). PR #2595 (user-local snapshots) is currently on hold, but some of the work done (in particular, 93eafd56b50f45702ca22599e8bf08fdf852065e) would fix this issue, by relaxing the requirement on hard-sync.

That commit should work as is, but additional testing is needed to ensure that syncing really is resumable (e.g., in the event of network errors, suspended devices, etc.), and optionally to add some UI element (see #2595) indicating that the app is loading.

(*) Longer term in the sense of making logging in faster, but, even then, having contract syncs 'in the background' (whether a full sync or a snapshot sync) can still be useful.

corrideat:
Closes #2700


frontend/controller/actions/identity.js
```diff
@@ -395,26 +395,26 @@ export default (sbp('sbp/selectors/register', {
       }
 
       try {
-        if (!state) {
+        await sbp('gi.db/settings/save', SETTING_CURRENT_USER, identityContractID)
+        sbp('okTurtles.events/emit', LOGIN, { identityContractID, encryptionParams, state })
```
taoeffect, line 410: Why is this being moved here?

The point of `LOGIN`, as far as I recall, was to indicate that we've logged in (at least partially).

This event is being emitted before that's happened, as the identity contract has not been loaded. It should at least be emitted after the identity contract has been loaded.

taoeffect (changes requested):
I'm not sure about this PR...

frontend/controller/actions/identity.js
corrideat: But we have partially logged in at this point, in the sense of verifying credentials. This change is necessary:

1. To allow for resumable login (as the issue says). Waiting for the entire identity contract means that logging in is an 'all or nothing' approach
2. For offline use. If you see #2508, you'll notice a similar change there. This is because in the case of an offline session, we obviously can't wait for `sync` to complete.

The `LOGIN` event also isn't to indicate that we've logged in, that's `LOGIN_COMPLETE`. See the `LOGIN_FLOW.md` file:

>   * `LOGIN`: The `LOGIN` event represents an active login intent, which is to
>   say, it represents that **_there is an active login process_**, which may either
>   fail or succeed. This event can originate in the service worker (when
>   establishing a new session) or in a local browsing context (when there is
>   an active sesion).

>    * `LOGIN_COMPLETE`: The `LOGIN_COMPLETE` is emitted following a `LOGIN` event
>  and signifies that **_the login process completed successfully_**. This event is
>  only local, which means that it doesn't originate on the service worker. Its
>  purpose is to let the app know that an active user session exists and has been
>  initialised.

corrideat: This is the change I was talking about, which is uglier than this because it's testing for a condition that's difficult to detect (come to think of it, that might be why that PR is working during dev testing but not under real conditions).

https://github.com/okTurtles/group-income/pull/2508/files#diff-f9e60360184dfab0063900ef547eb2c2ec1f436744964bd7f532e6c437187c0aL422-R427

frontend/controller/app/identity.js
```diff
@@ -119,15 +119,19 @@ sbp('okTurtles.events/on', LOGIN, async ({ identityContractID, encryptionParams,
...
+            .find(cID => currentState[identityContractID].groups[cID] && !currentState[identityContractID].groups[cID].hasLeft)
+
+          if (gId) {
+            sbp('gi.app/group/switch', gId)
+          }
         }
-      }
+      }).catch(e => {
+        console.error('[LOGIN] Error setting current group ID', e)
+      })
```
taoeffect, line 135: To me it seems like this code is really cutting things close. I don't see how this code is guaranteed to run after the identity contract has been sync'd. It seems possible to me that after `sbp('okTurtles.eventQueue/queueEvent', EVENT_HANDLED` and all the awaits in between, that this code could get run before the calls to `await sbp('chelonia/contract/retain', identityContractID)` / `await sbp('chelonia/contract/sync', identityContractID)`... in which case this could break because it would do nothing when it's supposed to do stuff.

frontend/controller/actions/identity.js
taoeffect: The `Join.vue` page listens for `LOGIN` as well. If there's an issue syncing the identity contract, it might think everything worked out well. When you return to this PR, can you please double-check that as well?

taoeffect: Also, worth testing this to see what happens if syncing throws, and how recoverable the situation is.

snowteamer: What about renaming `LOGIN` to `LOGIN_START`, `LOGIN_INIT`, or `LOGIN_BEGIN` to mitigate the confusion?

corrideat: > What about renaming LOGIN to LOGIN_START, LOGIN_INIT, or LOGIN_BEGIN to mitigate the confusion?

That sounds pretty reasonable

frontend/controller/app/identity.js
corrideat: I think you may be right here... This should run _after_ syncing most of the time, if syncing fails, for example because one is offline, this could end up doing nothing.

Now, maybe this isn't really much of a concern. `/switch` does this:

```
    sbp('okTurtles.events/emit', SWITCH_GROUP, { contractID: groupId, isNewlyCreated })
    sbp('state/vuex/commit', 'setCurrentGroupId', { contractID: groupId, isNewlyCreated
````

The `SWITCH_GROUP` event seems to be unused, and we have an event listener on `JOINED_GROUP` that does that this block here is trying to do, i.e., set the current group ID if none is set. So, it'd look like if syncing isn't complete, then the event related to joining (switching) would be emitted later as more contracts are synced.

PR #2711, opened 2025-03-09
New DB backend

### LINKED ISSUE #2644: Have ability for files to be stored outside of DB
### Problem

It is difficult to scale the storage of large files with the way we are doing things right now with the sqlite database.

OTOH, the sqlite database is rather convenient for some situations, for example, to verify locally that changes work.

### Solution

So one potential solution is a new kind of backend hybrid storage type, where contract events and KV values are stored in the sqlite file, but files posted via `POST /file` are stored someplace else, e.g. the filesystem or elsewhere.

You could even configure this storage to via environment variables that "files should go {here}, while everything else is stored in the database {here}".

See also this discussion on Slack for more ideas:

https://okturtles.slack.com/archives/C0EH7P20Y/p1740334633436749?thread_ts=1739843766.532749&cid=C0EH7P20Y

> [!IMPORTANT]
>  This issue should take advantage of the special hash types introduced in #2494 to distinguish between different types of stored types.

EDIT: copying this idea from Slack to emphasize that this could be more about being a "router API" than about being a different db backend:

> perhaps the 2644 issue could be more about implementing an API that can be used to do this with any backend, more so than building out a unique backend in of itself. so for example, this way we'd be able to say: "route non-files to this sqlite db, and files to this sqlitedb", or "route non-files to this sqlitedb, and files to this filesystem path", or "route non-files to this sqlitedb, and files to this postgres server"

corrideat:
Closes #2644

### DIFF
--- a/backend/database-fs.js
+++ b/backend/database-fs.js
@@ -7,9 +7,39 @@ import { checkKey } from '~/shared/domains/chelonia/db.js'
 // Initialized in `initStorage()`.
 let dataFolder = ''
 
+// Some operating systems (such as macOS and Windows) use case-insensitive
+// filesystems by default. This can be problematic for Chelonia / Group Income,
+// as we rely on keys being case-sensitive. This is especially relevant for CIDs,
+// where collissions could lead to DoS or data corruption.
+async function testCaseSensitivity () {
+  const date = new Date()
+  const dateString = date.toISOString()
+  const originalKey = `_private_testCaseSensitivity_${date.getTime()}_${(0, Math.random)().toFixed(8).slice(2)}`
+  // We only replace the first `p` with `P`. We don't use, e.g., `.toUpperCase()`
+  // because case conversion depends on the locale, and `i` may sometimes be
+  // `I` and sometimes be `İ`.
+  const differentlyCasedKey = '_P' + originalKey.slice(2)
+  await writeData(originalKey, dateString)
+  try {
+    const valueOriginalCase = await readData(originalKey)
+    const valueDifferentCase = await readData(differentlyCasedKey)
+    if (valueOriginalCase?.toString() !== dateString) {
+      console.error(`Unexpected value on case-sensitivity test; expected ${dateString}`)
+      throw new Error('Unexpected value: original key does not have the correct value')
+    }
+    if (valueDifferentCase?.toString() === dateString) {
+      console.error(`Unexpected value on case-sensitivity test; expected a differently-cased key to have a value other than ${dateString}`)
+      throw new Error('Unexpected value: key with different case has the same value as the origical key')
+    }
+  } finally {
+    await deleteData(originalKey)
+  }
+}
+
 export async function initStorage (options: Object = {}): Promise<void> {
   dataFolder = resolve(options.dirname)
   await mkdir(dataFolder, { mode: 0o750, recursive: true })
+  await testCaseSensitivity()
 }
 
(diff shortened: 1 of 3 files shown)


taoeffect (changes requested):
Amazing work @corrideat - very clean and elegant!

One minor change request: could you rename this from `meta` to `router` to emphasize more the "routing" nature of it?

backend/database-router.js
```diff
@@ -0,0 +1,49 @@
...
+
+  return config
+}
+let config
+const backends = Object.create(null)
+
+const lookupBackend = (key) => {
+  const keys = Object.keys(config)
+  for (let i = 0; i < keys.length; i++) {
+    if (key.slice(0, keys[i].length) === keys[i]) {
```
snowteamer: Why are we slicing the key here, instead of doing `key.startsWith(keys[i])` ? Even then, I fail to understand why a plain `key === keys[i]` isn't enough.

backend/database-router.js
```diff
@@ -0,0 +1,49 @@
...
+
+// eslint-disable-next-line require-await
+export async function writeData (key: string, value: Buffer | string): Promise<void> {
+  return lookupBackend(key).writeData(key)
+}
+
+// eslint-disable-next-line require-await
+export async function deleteData (key: string): Promise<void> {
+  return lookupBackend(key).deleteData(key)
+}
```
snowteamer, line 50: Is there a way to avoid repeating the lookup call every time?

backend/database-fs.js
```diff
@@ -7,9 +7,39 @@ import { checkKey } from '~/shared/domains/chelonia/db.js'
...
+  try {
+    const valueOriginalCase = await readData(originalKey)
+    const valueDifferentCase = await readData(differentlyCasedKey)
+    if (valueOriginalCase?.toString() !== dateString) {
+      console.error(`Unexpected value on case-sensitivity test; expected ${dateString}`)
+      throw new Error('Unexpected value: original key does not have the correct value')
+    }
+    if (valueDifferentCase?.toString() === dateString) {
+      console.error(`Unexpected value on case-sensitivity test; expected a differently-cased key to have a value other than ${dateString}`)
+      throw new Error('Unexpected value: key with different case has the same value as the origical key')
```
taoeffect: Could you please update these error messagse (for both the console.error and the exception thrown) to:

> Filesystem database backend only works on case-sensitive filesystems. This appears to be a case insensitive file system.

backend/database-router.js
corrideat: 1. `key.startsWith(keys[i])` would also work
2. `key === keys[i]` isn't enough because the key in config is a prefix. Actually, maybe `keyPrefixes` and `keyPrefix` would be better names.

backend/database-router.js
corrideat: I don't think so, as we need to find out which backend to use.

backend/database-router.js
taoeffect: > key.startsWith(keys[i]) would also work

Can you check which is faster?

corrideat: I think either is about as fast. Perhaps `startsWith` has the potential to be slightly faster (as the equivalent slicing would be done by native code) and is perhaps more readable.

taoeffect (dismissed):
Great job @corrideat !

taoeffect (approved):

PR #2715, opened 2025-03-11
#2476 - Weird iOS bug related to `position:fixed`

### LINKED ISSUE #2476: "Delete message" not visible on iPhone 12
### Problem

I can't delete chat messages on iPhone 12 because the menu item doesn't show the menu item:

a

### Solution

Show the menu, also consider making each row in the menu shorter.

SebinSong:
closes #2476 

**[Video recording for the fix]**

https://github.com/user-attachments/assets/3dba8071-d825-4a7b-b5aa-d063a62f4592

---
As I mentioned in [this comment](https://github.com/okTurtles/group-income/issues/2476#issuecomment-2691783489), I decided that we need something similar to [Teleport](https://vuejs.org/guide/built-ins/teleport) in Vue 3 to properly fix this issue. So have used [portal-vue](https://www.npmjs.com/package/portal-vue) package that provides the same behaviour for Vue 2 which is a 0 depenency package.
(I believe it is a good addition because this can be used elsewhere in the project too where teleporting behavior is needed)


frontend/views/containers/chatroom/MessageNotification.vue
```diff
@@ -17,7 +17,6 @@ export default ({
   components: {
     MessageBase
   },
-  inject: ['chatMessageUtils'],
```
SebinSong, line 20: Noticed this isn't used anywhere. So removed it.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1168,9 +1174,7 @@ export default ({
   },
   provide () {
     return {
-      chatMessageUtils: {
-        scrollToMessage: this.scrollToMessage
-      }
```
SebinSong, line 1173: This isn't used anywhere. so removed it.

taoeffect (approved):
Seems to work well. Tested and verified! Great job @SebinSong !

PR #2722, opened 2025-03-13
#2716 - Ability to adjust notification volume

### LINKED ISSUE #2716: Ability to adjust notification alert volume
### Problem

There's no way to adjust the volume of the GI notification sound.

### Solution

In the User Settings > Notifications Settings, add a new section with a slider you can drag to adjust the sound volume.

When the user lets go of either the mouse, or lifts their finger (if on touch), play the sound so that the user can hear how loud it is set to.

Then update the code that plays the sound to use that setting.

SebinSong:
closes #2716 

https://github.com/user-attachments/assets/1cf708b3-fc8c-49d6-acfd-a6feb54bb77e

@taoeffect Pls let me know if you would like to update English copies anywhere.

### DIFF
--- a/frontend/model/settings/vuexModule.js
+++ b/frontend/model/settings/vuexModule.js
@@ -32,7 +32,7 @@ export const defaultSettings = {
   fontSize: 16,
   increasedContrast: false,
   notificationEnabled: null, // 3 values: null (unset), true (user-enabled), false (user-disabled)
-  notificationVolume: 1,
+  notificationVolume: null,
   reducedMotion: false,
   theme: defaultTheme,
   themeColor: defaultColor
--- a/frontend/model/settings/vuexModule.js
+++ b/frontend/model/settings/vuexModule.js
@@ -32,7 +32,6 @@ export const defaultSettings = {
   fontSize: 16,
   increasedContrast: false,
   notificationEnabled: null, // 3 values: null (unset), true (user-enabled), false (user-disabled)
-  notificationVolume: 1,
   reducedMotion: false,
   theme: defaultTheme,
   themeColor: defaultColor
@@ -50,9 +49,6 @@ const getters = {
   },
   isReducedMotionMode (state) {
     return state.reducedMotion === true
-  },
-  notificationVolume (state) {
-    return state.notificationVolume
   },
   theme (state) {
     return state.theme
@@ -92,9 +88,6 @@ const mutations = {
     sbp('service-worker/setup-push-subscription').catch(e => {
       console.error(`[setNotificationEnabled] Error calling setup-push-subscription (enabled=${enabled})`, e)
     })
-  },
-  setNotificationVolume (state, value) {
-    state.notificationVolume = value
   },
   setReducedMotion (state, isChecked) {
     state.reducedMotion = isChecked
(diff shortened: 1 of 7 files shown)


SebinSong:
@taoeffect This PR is ready for review.

frontend/views/containers/chatroom/MessageActionsMobile.vue
```diff
@@ -52,11 +52,6 @@ export default ({
     isActive () {
       return this.Menu.isActive
     }
-  },
-  watch: {
-    isActive (newVal) {
-      console.log('!@# isActive: ', newVal)
-    }
```
SebinSong, line 59: Found out this was not noticed in this merged PR https://github.com/okTurtles/group-income/pull/2715. Removing it here.

taoeffect (changes requested):
Nice work @SebinSong! It seems like I made a mistake in not specifying where the setting should be stored. Could you please update this PR to use the local Vuex per-device settings instead of the KV store? This is a setting that should be unique on all devices.

taoeffect (changes requested):
Awesome! Great work @SebinSong it works really well in my tests!

I realize that we didn't have any designs for this, so I'd like to request a few design related changes to the UI, let me know if you have any questions about them:

Screenshot 2025-03-13 at 5 55 32 PM

SebinSong:
@taoeffect Updated the PR with the requested design changes.

frontend/views/containers/user-settings/NotificationVolume.vue
```diff
@@ -0,0 +1,116 @@
...
+      },
+      ephemeral: {
+        volume: 100
+      }
+    }
+  },
+  computed: {
+    volumeFromStore () {
+      return this.$store.getters.notificationVolume || 1
+    }
```
taoeffect: `o3-mini-high` caught an issue here: since `notificationVolume` can be `0` now, that would be false, and this would then return `1` - not what we intended.

So the suggestion would be:

```js
// check done this way because `0` is falsy
return this.$store.getters.notificationVolume != null ? this.$store.getters.notificationVolume : 1
```

Note that the same change must be done in `Background.vue` as well, as that also has this same computed property.

taoeffect (changes requested):
Great work @SebinSong !

Almost missed a very subtle bug!

frontend/views/containers/user-settings/NotificationVolume.vue
SebinSong: The suggested code can be simplified to:
```js
return this.$store.getters.notificationVolume ?? 1
```

frontend/model/settings/vuexModule.js
```diff
@@ -32,6 +32,7 @@ export const defaultSettings = {
   fontSize: 16,
   increasedContrast: false,
   notificationEnabled: null, // 3 values: null (unset), true (user-enabled), false (user-disabled)
+  notificationVolume: null,
```
taoeffect: I think it would be safer / more robust to have this set back to `1` as you had it before, just in case there's any other code that we add in the future that relies on this value from outside of a computed property.

taoeffect (approved):
Great work @SebinSong!

PR #2724, opened 2025-03-13
chelonia/time scale adjusted to ms (from s)

### LINKED ISSUE #2593: Update timescale for 'chelonia/time' to be milliseconds (same as Date.now())
### Problem

It seems weird to have to first divide by `1e3` and [then multiply by `1e3`](https://github.com/okTurtles/group-income/pull/2592#discussion_r1947103325) all over the place.

### Solution

Why not just stick to `Date.now()`'s time scale for consistency?

corrideat:
Fix #2593

### DIFF
--- a/shared/domains/chelonia/time-sync.js
+++ b/shared/domains/chelonia/time-sync.js
@@ -112,10 +112,10 @@ export default (sbp('sbp/selectors/register', {
   // Get an estimate of the server's current time based on the time elapsed as
   // measured locally (using a monotonic clock), which is used as an offset, and
   // a previously retrieved server time. The time value is returned as a UNIX
-  // timestamp (seconds since 1 Jan 1970 00:00:00 UTC)
+  // _millisecond_ timestamp (milliseconds since 1 Jan 1970 00:00:00 UTC)
   'chelonia/time': function (): number {
     const monotonicNow = performance.now()
     const wallNow = wallBase - monotonicBase + monotonicNow
-    return (wallNow / 1e3 | 0)
+    return Math.round(wallNow)
   }
 }): string[])
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -53,7 +53,7 @@ export default (sbp('sbp/selectors/register', {
       throw new Error('Unable to update lastLoggedIn without an active session')
     }
 
-    const now = sbp('chelonia/time') * 1000
+    const now = sbp('chelonia/time')
     if (throttle) {
       const state = sbp('state/vuex/state')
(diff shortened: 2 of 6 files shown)


shared/domains/chelonia/time-sync.js
```diff
@@ -112,10 +112,10 @@ export default (sbp('sbp/selectors/register', {
   // Get an estimate of the server's current time based on the time elapsed as
   // measured locally (using a monotonic clock), which is used as an offset, and
   // a previously retrieved server time. The time value is returned as a UNIX
-  // timestamp (seconds since 1 Jan 1970 00:00:00 UTC)
+  // _millisecond_ timestamp (milliseconds since 1 Jan 1970 00:00:00 UTC)
   'chelonia/time': function (): number {
     const monotonicNow = performance.now()
     const wallNow = wallBase - monotonicBase + monotonicNow
-    return (wallNow / 1e3 | 0)
+    return Math.round(wallNow)
```
taoeffect, line 119: Why is `Math.round` being called here?

corrideat: Because this value should be an integer, the extra precision isn't significant. Also, the shelter auth header uses dots as a special character.

taoeffect: Ah I see. Well, in that case, it might be the case that `chelonia/time` is used in situations where the decimal place is useful, so could you please move this `.round` to the call-site that calls this selector for generating the header?

corrideat: The decimal place is never useful because it's noise. None of the values used as a source have sub-millisecond precision, and the fact that this does is an artifact of how it's calculated.

taoeffect: So what you're saying is that you are personally guaranteeing me that no Chelonia application, whether made by us or anyone else in the future, will ever have a need for sub-second timestamps? Can you give me some sort of proof of this?

corrideat: ~~Whether they do or don't is irrelevant. JavaScript (*) doesn't offer sub-millisecond timestamps. Note sub-**milli**second.~~

~~(*) As in the browser~~

Turns out I was wrong, and `performance.now()` isn't rounded to milliseconds like I thought it was for privacy, so technically this precision exists. However, this extra precision mostly _is_ noise because it tries to approximate the server time in ways that aren't strictly precise, so even the milliseconds is already pushing it.

taoeffect: Oh............................................................................. 🤦‍♂️ 

I'm having a day.

Pretend you didn't see any of that.

taoeffect (approved):
Great work @corrideat !

PR #2725, opened 2025-03-13
Add support for multiple DB backend instances

### LINKED ISSUE #2720: Ability to configure db backends in config file for database-router.js
### Problem

The `database-router.js` config file doesn't support the ability to pass in options to the different db backends.

Also, it doesn't support the ability to use the same backend twice.

### Solution

Change the config format such that the values are an object where `name` specifies the db backend, and `options` are the config passed in to it.

Instead of having the singleton design, make it possible to create instances of these backends. It's OK to use classes if it makes the code more elegant, but also if the code is equivalently elegant without them, then there's no need.

snowteamer:
Close #2720


backend/database-router.js
```diff
@@ -7,8 +7,8 @@ const readConfig = () => {
   if (!config['*']) {
     throw new Error('Fallback storage (*) is required')
   }
-
-  return config
+  // Return a sorted copy where entries with longer keys come first.
+  return Object.fromEntries(Object.entries(config).sort((a, b) => a[0].length - b[0].length))
```
taoeffect: Should be `b[0].length - a[0].length`

backend/database-router.js
```diff
@@ -17,20 +17,20 @@ const lookupBackend = (key) => {
...
 export async function initStorage (options: Object = {}): Promise<void> {
   config = readConfig()
-  const persistences = [...new Set(Object.values(config))]
-  await Promise.all(persistences.map(async (persistence) => {
-    const { initStorage, readData, writeData, deleteData } = await import(`./database-${persistence}.js`)
-    // $FlowFixMe[invalid-computed-prop]
-    backends[persistence] = { readData, writeData, deleteData }
-    await initStorage()
+  const entries = Object.values(config)
+  await Promise.all(entries.map(async ([keyPrefix, { name, options }]) => {
```
taoeffect: Pretty sure `keyPrefix` won't be a thing here, as the values of the object will just look like `{ name: 'fs' }`, `{ name: 'sqlite', options: { .. }` etc. You might have been thinking of `Object.entries`..

test/database.test.js
```diff
@@ -22,16 +22,12 @@ const range = (n) => [...new Array(n).keys()]
...
 names.forEach((name) => {
   const lowerCaseName = name.toLowerCase()
-  const {
-    clear,
-    initStorage,
-    readData,
-    writeData
-  } = require(`~/backend/database-${lowerCaseName}.js`)
+  const Ctor = require(`~/backend/database-${lowerCaseName}.js`).default
+  const { init, clear, readData, writeData } = new Ctor(options[lowerCaseName])
```
taoeffect, line 26: Could you add a test case for testing the `database-router.js`? That way we can catch bugs related to parsing the JSON config for that.

I think you will also need to add a class for the `database-router.js`

backend/database-router.js
```diff
@@ -7,8 +7,8 @@ const readConfig = () => {
   if (!config['*']) {
     throw new Error('Fallback storage (*) is required')
   }
-
-  return config
+  // Return a sorted copy where entries with longer keys come first.
```
taoeffect: Above this line, I didn't notice in @corrideat's PR that he's reading the JSON config from an envvar (as opposed to from a file pointed to by the envvar).

Could you update this code so that it reads from a JSON file? By default, if `GI_PERSIST=router` is used, it will read from `data/db-router-config.json` if that file exists.

backend/db-utils.js
```diff
@@ -0,0 +1,6 @@
+// Rebind public methods of the given DB wrapper to make them usable with destructuring.
+export function rebindMethods (object: Object, methodNames: string[]): void {
+  for (const name of methodNames) {
+    object[name] = object[name].bind(object)
+  }
+}
```
taoeffect: Since all of the databases are supposed to support the same API, could you maybe not pass in those names to `methodNames`, but instead have a const in this file at the top with those names?

That way we can keep things DRY and also be able to enforce that all subsequent implementations implement all of these methods or throw if they don't.

Also, another option to explore is to make a shared class from which all of these other classes inherit from, and then do this action there in the parent class constructor.

backend/database-fs.js
```diff
@@ -36,38 +35,56 @@ async function testCaseSensitivity () {
...
+        .then(names => names.map(name => join(this.dataFolder, name)))
+        .then(paths => paths.map(path => rm(path, { recursive: true })))
+    )
+  }
+
+  async readData (key: string): Promise<Buffer | string | void> {
+    // Necessary here to thwart path traversal attacks.
+    checkKey(key)
+    return await readFile(this.mapKey(key))
+      .catch(err => undefined) // eslint-disable-line node/handle-callback-err
```
taoeffect: Do you remember why we have this empty `.catch` to swallow errors?

We don't have one on `writeData`...

Are you sure it should be here? Does it have something to do with 404 handling?

backend/database-fs.js
```diff
@@ -36,38 +35,56 @@ async function testCaseSensitivity () {
...
-  checkKey(key)
-  return readFile(join(dataFolder, key))
-    .catch(err => undefined) // eslint-disable-line node/handle-callback-err
-}
+  // Maps a given key to a real path on the filesystem.
+  mapKey (key: string): string {
+    if (!this.depth) return join(this.dataFolder, key)
+    const keyChunks = key.match(/[A-Za-z]{1,3}/g) ?? []
+    return join(this.dataFolder, ...keyChunks.slice(0, this.depth), keyChunks.slice(this.depth).join(''))
+  }
```
taoeffect: Could you please expand the comment above to explain in more detail what this is supposed to be doing and also show some examples?

Also, is the regex correct? Should it use a `^` at the start?

And finally, could you add some tests that test this feature?

taoeffect (changes requested):
Great work @snowteamer!

Review ready!

Please also keep the `data/.gitkeep` file, that's important to ensure that directory is created when cloning this project.

backend/database-fs.js
snowteamer: Updating this function would allow to customize how the backend maps keys to real paths on the filesystem,
making it easier to implement new behaviors for this backend, without touching its other methods. That's the main point of it.

Examples for a key `abcdefgh`:

- If we simply want to use the key itself as filename, then we don't need the `depth` option at all. The path should be `data/abcdefgh`.
- If we want to split the keys in chunks of at most three chars in order to use a nested subfolder structure under `/data`, then we'd want `depth: 3`. The returned path should be `data/abc/def/gh`. Since the last chunk is simply a file name rather than a subfolder name, it can have fewer chars.

Of course the returned path should always be a valid path under `this.dataFolder`. I'm going to have a try at making this more clear.

snowteamer: Adding an initial `^` to the regex would only match one chunk at most because of the `g` flag.

backend/database-fs.js
snowteamer: It's there to return `undefined` in case the file does not exist, i.e. if the given key is not found in the DB. It would be a bit more precise to do this only for `ENOENT` errors. Should we do this?

OTOH, `writeData` doesn't care whether a file is already there, as long as it can overwrite that file.
Other kinds of errors are possible, but we let them bubble up for now.

backend/db-utils.js
snowteamer: > instead have a const in this file at the top with those names?

Sure

> shared class from which all of these other classes inherit from

That's an option, but I'm a bit wary about it. OO inheritance usually means headaches

backend/database-router.js
snowteamer: Not `./database-router-config.json` in the root folder? Or in some new `./config` folder?

backend/database-router.js
snowteamer: Good catch!

backend/database-fs.js
corrideat: I'm not so sure about the workings of this `depth` thing. Normally, for a key `abcdefgh` and depth 3, I'd have seen `a/b/c/abcdefgh`. I know that keys have a common prefix that ends up requiring pretty large depths, and don't have a good solution to this, other than maybe using the terminal part, e.g., `h/g/f/abcdefgh`.

Now, I see two concrete issues with storing `abcdefgh` as `abc/def/gh`:

1. Using `gh` as the name makes it difficult to know which key it is without knowing the exact location in the file system, and potentially makes it difficult to switch to a different path or to do regular sysadmin tasks (e.g., if you see a `gh` file that is too large, or an anti-malware tool warns you about `gh`, you need to further investigate which `gh` file it might be). Using `abcdefgh` as the file name doesn't have these issues (i.e., `abc/def/abcdefgh` instead of `abc/def/gh`).
2. The reason for using paths has to do with overcoming filesystem limitations and making certain things, like indexing and searching faster. If we exclude prefixes and assume CIDs are mostly random, a three-character path results in 58^3 ≈ 195000 combinations. Granted, this is a worst-case scenario, but I'm not sure how well file-systems would perform with such a large number of files or directories.

backend/database-router.js
corrideat: AFAIK, it's pretty standard for JSON config to be read this way, which also lends itself well to containers. If adding a file, I'd rename the env var to end in `_FILE` (ETA: see this is done) and also I wouldn't have a default location, or at least most definitely not have it be under `data` (this makes it conflict with actual data)

backend/db-utils.js
```diff
@@ -0,0 +1,8 @@
+const methodNames = ['init', 'clear', 'readData', 'writeData', 'deleteData']
+
+// Rebind public methods of the given DB wrapper to make them usable with destructuring.
+export function rebindMethods (object: Object): void {
+  for (const name of methodNames) {
+    object[name] = object[name].bind(object)
+  }
+}
```
taoeffect: Since we want to make it easy for people to be able to implement their own backends, I think we can make it more straightforward than "call this weird function". Let's go ahead and create a superclass that all of these db backends must inherit from. JS does actually make it possible for us to enforce various things through inheritance, as shown in this example:

```js
class A {
  constructor() {
    if (new.target === A) {
      throw new Error("Class A cannot be instantiated directly")
    }
    // Check required methods
    const requiredMethods = ['bar']
    for (const method of requiredMethods) {
      if (typeof this[method] !== 'function') {
        throw new Error(`Class ${this.constructor.name} must implement method '${method}'`)
      }
    }
  }
}

class B extends A {
  constructor () {
  }
}

class C extends A {
  constructor () {
    super()
  }
}

class D extends A {
  constructor () {
    super ()
  }
  bar () {}
}
```

Running this in node, we see that only subclass `D` can be instantiated:

```
> new B()
Uncaught:
ReferenceError: Must call super constructor in derived class before accessing 'this' or returning from derived constructor
    at new B (evalmachine.<anonymous>:17:15)
> new C()
Uncaught Error: Class C must implement method 'bar'
    at new A (evalmachine.<anonymous>:10:15)
    at new C (evalmachine.<anonymous>:23:5)
> new D()
D {}
```

class `A` will ensure that:

1. if subclasses define a constructor, that they must call `super()` inside of it
2. that subclasses implement the `requiredMethods`

This makes it easy for us to say, "Your db backend must inherit from the `DBBackend` class".

That's a good name for it too (`DBBackend`).

backend/db-utils.js
taoeffect: I think in this instance it would be really helpful; I replied here with why: https://github.com/okTurtles/group-income/pull/2725#discussion_r1996177904

backend/database-router.js
taoeffect: Yeah I think we should have a global config for configuring this and other things too... but that would be a separate issue. So for now `./database-router-config.json` sounds fine to me. Just remember to add that to `.gitignore`.

backend/database-fs.js
snowteamer: > I'd have seen a/b/c/abcdefgh

Agreed I've gotten the naming wrong with this `depth` option. It gets actually interpreted as "subfolder name length", so I should either rename it or make it behave like in your example.

>1. & 2.

Yea maybe this way of using paths is not optimal. In any case, I don't mean to make it mandatory; rather, for now I mean it as a simple implementation of the idea outlined in Slack here: https://okturtles.slack.com/archives/C0EH7P20Y/p1740338446462349?thread_ts=1739843766.532749&cid=C0EH7P20Y.
In a more complex version, we could have more options to better tune its behavior.

For example, if 1) We know that the random part of our keys only starts after the first six chars; and 2) We don't want more than say three "intermediate" subfolders with names at most two chars long to keep the system responsive, then we could return paths like `data/zL7mM9/d4/Xb/4T/F3G5Pgnw6gk2aUxQiv8H75h6AkG6hZAzoGZRJER8fchC`  (where `d4/Xb/4T` are the intermediate subfolders). This also leaves a quite long file name at the end of the path.

backend/db-utils.js
snowteamer: Okay! 👍

backend/database-fs.js
corrideat: > We know that the random part of our keys only starts after the first six chars

Yeah and no.

We have a bunch of keys that aren't CIDs, for example those starting with `_`, like `_private_`.

In addition, for those which are CIDs, I controlled the numbers used so that all keys have a fixed length prefix. But CID content types are of variable length, so we can't really make that assumption in general (at least, not without making additional assumptions, such as that the only keys supported will be of the 4 Shelter types).

corrideat: I think that maybe two different parameters would be needed here, the depth (number of segments) and the segment length. A segment length of 1 or 2 sounds reasonable, 3 maybe as well but is pushing it. A length of 4+ sounds pretty long (in terms of possible combinations).

In the end, the correct length and depth (whether these are separate configuration options or not) depends on how many files there'll be. It'd be nice to research different filesystems and their performance on different HW to make a decision here. That's probably too much work, so maybe an alternate solution is to make some upper bound assumptions, like this:

Say we want no more than ~3000 files per directory (a number I just made up), and we expect a server to have 300M contracts, and each contract has 100k events. If my math is right, a prefix length of 2 and a depth of 4 should work.

.gitignore
```diff
@@ -13,6 +13,9 @@ key.pub.json
 .eslintcache
 .stylelintcache
 
+# Configs
+/backend/database-router-config.json
```
taoeffect: Think this should just be `/database-router-config.json` or `database-router-config.json`

backend/database-fs.js
```diff
@@ -36,38 +35,59 @@ async function testCaseSensitivity () {
...
-    }
-    throw e
-  })
+  async clear () {
+    await Promise.all(
+      await readdir(this.dataFolder)
+        .then(names => names.map(name => join(this.dataFolder, name)))
+        .then(paths => paths.map(path => rm(path, { recursive: true })))
+    )
+  }
```
taoeffect: `o3-mini-high` suggests making this code simpler/clearer, and I like its suggestion:

```js
async clear () {
  const names = await readdir(this.dataFolder)
  const paths = names.map(name => join(this.dataFolder, name))
  await Promise.all(paths.map(path => rm(path, { recursive: true })))
}
```

backend/database-fs.js
```diff
@@ -36,38 +35,59 @@ async function testCaseSensitivity () {
...
+
+  async deleteData (key: string) {
+    await unlink(this.mapKey(key)).catch(e => {
+      // Ignore 'not found' errors
+      if (e?.code === 'ENOENT') {
+        return
+      }
+      throw e
+    })
+  }
```
taoeffect, line 98: For both `writeData` and `deleteData`, Gemini 2.5 Pro suggests adding `checkKey(key)` like in `readData`. Seems reasonable to me.

But perhaps these checks should be done in `backend/database.js` under the `'chelonia.db/*` selectors instead of here? That way they can be performed regardless of what database backend is used, ensuring future compatibility if there is a migration from one DB to a different one (e.g. sqlite -> fs).

backend/database-router.js
```diff
@@ -1,50 +1,94 @@
...
-    throw new Error('Fallback storage (*) is required')
+type Config = {
+  [string]: { name: string, options: Object }
+}
+type ConfigEntry = { name: string, options: Object }
+
+const { GI_PERSIST_ROUTER_CONFIG_PATH = './database-router-config.json' } = process.env
+
+export default class RouterBackend extends DatabaseBackend {
+  backends: Object
```
taoeffect: Gemini 2.5. Pro writes:

In `backend/database-router.js`, the `this.backends` property is implicitly typed as `Object`. While functional, providing a more specific FlowType annotation could improve type safety and potentially allow removing the `$FlowFixMe[incompatible-use]` comment in the `clear` method, as Flow would have better information about the methods available on the backend instances.

```diff
--- a/backend/database-router.js
+++ b/backend/database-router.js
@@ -15,7 +15,7 @@
 const { GI_PERSIST_ROUTER_CONFIG_PATH = './database-router-config.json' } = process.env
 
 export default class RouterBackend extends DatabaseBackend {
-  backends: Object
+  backends: { [string]: DatabaseBackend }
   config: Config
 
   lookupBackend (key: string): Object {
@@ -67,7 +67,7 @@
       throw new Error(`[${this.constructor.name}] ${errors.length} error(s) found in your config.`, { cause: errors })
     }
     // Init backends
-    this.backends = Object.create(null)
+    this.backends = (Object.create(null): { [string]: DatabaseBackend })
     const entries = ((Object.entries(this.config): any): ConfigEntry[])
     await Promise.all(entries.map(async entry => {
       const [keyPrefix, { name, options }] = entry
```

*(Note: The cast `(Object.create(null): { [string]: DatabaseBackend })` might be necessary during initialization depending on Flow's inference capabilities.)*

test/database.test.js
taoeffect: To implement this, one option would be for the test coding a way for the router backend to ignore the `GI_PERSIST_ROUTER_CONFIG_PATH`.

I notice for example in the `init` method for the router it takes an `options` object that is not used:

```js
async init (options: Object = {}): Promise<void> {
```

It could be coded so that if that's passed in it uses those instead of calling `this.readConfig()`

OTOH, I realize that doing it this way would mean that we're not testing the code for properly parsing the config file.

So an alternative is maybe to create a separate environment variable as @corrideat had it originally, so that for example if `GI_PERSIST_ROUTER_CONFIG` is defined, it is used instead of `GI_PERSIST_ROUTER_CONFIG_PATH`, and it would contain the entire text of the config file. It would then be defined here in this test file.

backend/database-router.js
```diff
@@ -1,50 +1,94 @@
...
-  config = readConfig()
-  const persistences = [...new Set(Object.values(config))]
-  await Promise.all(persistences.map(async (persistence) => {
-    const { initStorage, readData, writeData, deleteData } = await import(`./database-${persistence}.js`)
-    // $FlowFixMe[invalid-computed-prop]
-    backends[persistence] = { readData, writeData, deleteData }
-    await initStorage()
-  }))
-}
+  async init (options: Object = {}): Promise<void> {
```
taoeffect: Note that these `options` don't appear to be used here... see related comment in `database.test.js`

taoeffect (changes requested):
Fantastic work @snowteamer!

Think this PR is almost ready to be merged. See comments above that I just left!

backend/database-fs.js
snowteamer: The check is already there in the overriden versions of `'chelonia.db/set'` and `'chelonia.db/delete'` from `database.js`. It is not included in `'chelonia.db/get'`, but `undefined` will be returned if the key is invalid.
I think it's not included in the original selectors from `~/shared/domains/chelonia/db.js` because they use `okTurtles.data` as DB backend which is not susceptible to path traversal issues since it's an in-memory KV store.

corrideat: > I think it's not included in the original selectors from ~/shared/domains/chelonia/db.js because they use okTurtles.data as DB backend which is not susceptible to path traversal issues since it's an in-memory KV store.

I think it's a good idea to enforce name restrictions everywhere, because that can catch potential issues early. For example, imagine someone adding an entry called `_private/foo`. If they only test using the in-memory DB they may inadvertently introduce a change that'll break in production (which does enforce this restriction).

backend/DatabaseBackend.js
```diff
@@ -0,0 +1,20 @@
+const requiredMethodNames = ['init', 'clear', 'readData', 'writeData', 'deleteData']
+
+export default class DatabaseBackend {
+  constructor () {
+    // $FlowFixMe[unsupported-syntax]
+    if (new.target === DatabaseBackend) {
+      throw new Error('Class DatabaseBackend cannot be instantiated directly.')
+    }
+    // Check required methods.
```
corrideat: I like the idea of defining a common interface, but I do have a few suggestion / comments. I tend to prefer static build checks over runtime checks.

1. Do we _really_ need the runtime check here? They're not wrong, but feel a bit like unnecessary overhead and complexity without a clear use case (that use case could be, for example, third party libraries that provide backends that also may be of a different version)
2. Regardless of the answer to the previous question, can you please add types to this? The way you'd do it is by defining an `interface`. I've just checked and it works, like so:

```ts
interface Foo {
  bar(baz: string): number;
}

class Qux implements Foo {
  bar(baz) {
     return baz ? 42 : -1;
  }
}
```

backend/DatabaseBackend.js
```diff
@@ -0,0 +1,20 @@
...
+    // Also rebind them to the instance so as to make them usable with destructuring.
+    for (const name of requiredMethodNames) {
+      // $FlowFixMe[prop-missing]
+      if (typeof this[name] !== 'function') {
+        throw new Error(`Class ${this.constructor.name} must implement method '${name}'`)
+      }
+      // $FlowFixMe[prop-missing]
+      this[name] = this[name].bind(this)
```
corrideat, line 20: I understand _why_ this is being done, but it feels a bit weird considering how objects are generally used.

backend/database-fs.js
```diff
@@ -1,17 +1,16 @@
...
-// Initialized in `initStorage()`.
-let dataFolder = ''
+import DatabaseBackend from './DatabaseBackend.js'
 
 // Some operating systems (such as macOS and Windows) use case-insensitive
 // filesystems by default. This can be problematic for Chelonia / Group Income,
 // as we rely on keys being case-sensitive. This is especially relevant for CIDs,
 // where collisions could lead to DoS or data corruption.
-async function testCaseSensitivity () {
+async function testCaseSensitivity (backend: Object) {
```
corrideat, line 13: Going with the OOP approach, this _perhaps_ makes more sense as a class method, private if you'd like.

backend/database-fs.js
```diff
@@ -1,17 +1,16 @@
...
+  const { readData, writeData, deleteData } = backend
```
corrideat, line 14: This will be my last comment about the binding. It does feel a bit weird, and I think one issue with that could be continually having to double-check (in the future) that this is doing what it should do.

ETA: The reason this feels a bit weird is because `backend` is a class instance, where doing this unbinding generally isn't expected to work. If `backend` weren't a class instance, I'd say this would be fine.

backend/database-sqlite.js
```diff
@@ -3,88 +3,99 @@
 import { mkdir } from 'node:fs/promises'
 import { join, resolve } from 'node:path'
 import sqlite3 from 'sqlite3'
```
corrideat: In my PR #2737 I ended up having to replace `sqlite3` with `better-sqlite3` because `sqlite3` crashed when using workers. See 0635655eb9a6f16ce3749d98e99008138772ae82 for details.

My PR will have a merge conflict with yours, and a nice side-effect is that `better-sqlite3` has a simpler API. Would you considering making the switch here too?

corrideat (changes requested):
Your PR looks fine to me! Although I'm selecting 'Request changes', my approval isn't contingent on most of my comments, except that I think at least an `interface` is needed for static type checking.

backend/database-fs.js
corrideat: > It would be a bit more precise to do this only for ENOENT errors. Should we do this?

I think it'd be more precise, although maybe `ENOENT` isn't the only error that should return undefined. 

Nitpick: If you'll swallow the error maybe add a comment explaining why, and maybe log something if the error isn't `ENOENT`.

backend/DatabaseBackend.js
taoeffect: I think the runtime check is fine personally. @corrideat could you give more details about what sort of interface you'd like? Are you referring to an interface for implementing `requiredMethodNames`?

corrideat: I think a runtime check is most of the time the wrong answer, except in some cases when implementing a plug in system or something similar with third-party developers / components. I say this since:

1. A runtime check will only result in runtime errors, so issues can only be caught after execution.
2. The checks available at runtime are usually less comprehensive than those done statically (for example, here we're only checking for presence, while a static check can verify the entire signature)
3. It wastes opcodes on every check / run for something that could have been checked once during development.

1 & 2 is why I think that static types should be added, and 3 is why I think _maybe_ the runtime check can be eliminated entirely.

> Are you referring to an interface for implementing `requiredMethodNames`?

Correct. I think we should have a flow / TypeScript error if the methods don't exist or don't conform to the expected signature.

taoeffect: > It wastes opcodes on every check / run for something that could have been checked once during development.

Agreed with you in general, though in this case don't think (3) is an issue as it's only done once on startup.

snowteamer: Interfaces cannot have constructors, so I think an interface makes it difficult to define a shared `super` constructor.
Outside runtime checks, the main point of that constructor is binding instance methods.

Sure this convenience can be dropped, then we'll just have to prepend DB method calls with the usual object reference.

backend/database-fs.js
snowteamer: I didn't make it a class method to minimize diff changes. Another reason is I think that function could turn out to be useful somewhere else.

backend/database-sqlite.js
snowteamer: Sure, I'd rather avoid ending up with two separate SQLite dependencies

backend/database-fs.js
corrideat: > Another reason is I think that function could turn out to be useful somewhere else.

Sure, but in that case the `backend` argument seems odd. Since it's only used internally, I think it's fine as is though.

backend/DatabaseBackend.js
corrideat: > an issue as it's only done once on startup

And adding a short delay to _every startup_, but yeah, a small delay.

> Interfaces cannot have constructors, so I think an interface makes it difficult to define a shared super constructor.

With TypeScript an abstract class would cover this, but I don't think Flow (or the version we use) supports it. The current solution is verbose but looks fine.

corrideat (dismissed):
Approved, ~~although the intent of my comment about `better-sqlite3` was to actually use it in your `database-sqlite.js` file.~~ (sorry, I was looking at the wrong diff)

Still approved, but look at my new comments for some minor improvements.

backend/database-router.test.js
```diff
@@ -0,0 +1,42 @@
...
+  },
+  '*': {
+    name: 'fs',
+    options: {
+      dirname: './data'
+    }
+  }
+}
+
+describe('DatabaseRouter::validateConfig', () => {
```
corrideat: I think we could use `describe` and `it` from `node:test`, which will reduce the need to have extra dependencies.

backend/database-router.test.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-env mocha */
```
corrideat, line 1: This is no longer needed if using `node:test`

backend/database-router.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+import assert from 'node:assert'
+import RouterBackend from './database-router.js'
+import { cloneDeep, omit } from 'turtledash'
+
+const validConfig = {
```
corrideat, line 12: The config is valid, but my general comment about this is that the config isn't tested as far as I can see. It'd be useful to perhaps add a test verifying that keys go to the correct storage.

backend/database-router.test.js
```diff
@@ -0,0 +1,42 @@
...
+  'abc': {
```
corrideat: Maybe also a comment explaining that 'abc' is a prefix, for context to the reader?

backend/database-router.js
```diff
@@ -1,50 +1,96 @@
...
+import { readFile } from 'node:fs/promises'
+import DatabaseBackend from './DatabaseBackend.js'
+import type { IDatabaseBackend } from './DatabaseBackend.js'
 
-  if (!config['*']) {
-    throw new Error('Fallback storage (*) is required')
+type Config = {
+  [string]: { name: string, options: Object }
+}
+type ConfigEntry = { name: string, options: Object }
```
corrideat: If the order of these type declarations is reversed, it can be simplified and DRYed:

```ts
type ConfigEntry = { name: string, options: Object }
type Config = { [string]: ConfigEntry }
```

backend/database-router.test.js
taoeffect: Yeah and I would use a real key prefix personally

backend/database-router.js
```diff
@@ -1,50 +1,96 @@
...
+
+const { GI_PERSIST_ROUTER_CONFIG_PATH = './database-router-config.json' } = process.env
```
corrideat: @taoeffect I know this was a requested change, but I'm not happy with this being the sole option. The way it was before, which could read the config straight from an environment variable, is easier in certain situations, such as with containers.

taoeffect: OK either you or @snowteamer are free to later add such an envvar

backend/database-sqlite.js
```diff
@@ -2,89 +2,63 @@
...
-  })
-}
+  constructor (options: Object = {}) {
+    super()
+    const { dirname, filename } = options
+    this.dataFolder = resolve(dirname)
+    this.filename = filename
+  }
+
+  run (sql: string) {
```
corrideat, line 24: Why not just use `this.db.exec(sql)` (i.e., why `prepare` + `run`?)

backend/database-router.js
snowteamer: I'm ok to implement this, but what to do if both `GI_PERSIST_ROUTER_CONFIG`  and `GI_PERSIST_ROUTER_CONFIG_PATH` are defined?
- Make one of them have priority over the other
- Raise an error because of the ambiguity

taoeffect: @snowteamer I would make the envvar take precedence and add a comment to that effect

backend/database-sqlite.js
snowteamer: IIUC, the [docs](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#execstring---this) recommend to avoid using `exec` if possible.

test/database.test.js
snowteamer: In other backend classes, `init` has no `options` parameter, only the constructor has it.
So I think I'm going to remove it here, but still make the constructor accept a `config` option

backend/database-router.test.js
snowteamer: So I've added various tests for `lookupBackend()`, which should implicitly test if keys go to the correct storage, while delegating the actual `read-`, `write-` and `deleteData()` tests to underlying backend tests (which currently go in `database.test.js`)

taoeffect (dismissed):
Approved! Amazing work @snowteamer!

backend/database-router.js
```diff
@@ -5,17 +5,29 @@ import { readFile } from 'node:fs/promises'
...
+  GI_PERSIST_ROUTER_CONFIG_PATH = './database-router-config.json'
+} = process.env
 
 export default class RouterBackend extends DatabaseBackend implements IDatabaseBackend {
   backends: { [string]: DatabaseBackend }
   config: Config
 
+  constructor (options: Object = {}) {
+    super()
+    this.config = options.config ?? null
```
corrideat: `this.config` is being assigned `null` but its type is `Config`.

backend/database-router.test.js
```diff
@@ -1,27 +1,32 @@
-/* eslint-env mocha */
```
corrideat: I think these need to be added back if not using `node:test`.

corrideat (changes requested):
A few minor type-related changes.

backend/database-router.js
```diff
@@ -23,9 +23,9 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
   backends: { [string]: DatabaseBackend }
   config: Config
 
-  constructor (options: Object = {}) {
+  constructor (options: { config?: Config } = {}) {
     super()
-    this.config = options.config ?? null
+    if (options.config) this.config = options.config
```
corrideat, line 28: Minor comment: I think this _still_ doesn't address the type issue. You're saying that `config` has type `Config` (in line 24), but instead it has type `?Config` (or `Config | typeof undefined`, before it was `Config`).

I see that if you call `init` as you should, then `config` will then have the correct type. I think the types as they're currently defined are inaccurate, but I'm not sure if this is worth pursuing further.

corrideat (approved):
As my other comment indicates, I don't think the type issue has been entirely addressed, but I also think it's probably a minor semantic issue.

backend/database-router.js
taoeffect: @corrideat when this PR is merged, can you fix this when you merge `master` into your PR that I'm currently reviewing?

taoeffect (approved):

backend/database-router.js
snowteamer: @corrideat Yeah instance properties don't have their correct types yet when exiting the constructor call. They have to wait until `init()` has returned successfully. For example, the `backends` property isn't set in the constructor either which means `this.backends` is initially `undefined`.

PR #2726, opened 2025-03-17
#2721 - Fix the scroll-position bug

### LINKED ISSUE #2721: ChatMain no longer remember scroll position
### Problem

Image

I'm not sure what happened. Perhaps some recent changes to ChatMain broke its ability to remember scroll positions.

Switching on GIG between `#random` and `#general` doesn't remember scroll position, and seems stuck for me on an old message. In other words, it doesn't scroll to the end as it shou.d

### Solution

Fix.

SebinSong:
closes #2721 

https://github.com/user-attachments/assets/f22f285e-3c8c-402f-81b1-88a1b13862ab

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1128,7 +1128,7 @@ export default ({
         }
 
         if (this.ephemeral.messagesInitiated) {
-          // Sometimes event after 'messagesInitiated' is set to 'true', infiniteHandler() is called again and loads more messages.
+          // Sometimes even after 'messagesInitiated' is set to 'true', infiniteHandler() is called again and loads more messages.
           // In that case, we should defer 'auto-scrolling to the initial-position' until those additional messages are rendered.
           // This can be achieved by calling 'scrollToInitialPosition' here with setTimeout(),
           // and calling clearTimeout() at the start of infiniteHandler().
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -281,7 +281,7 @@ export default ({
   },
   mounted () {
     // setup various event listeners.
-    this.ephemeral.onChatScroll = debounce(onChatScroll.bind(this), 500)
+    this.ephemeral.onChatScroll = debounce(onChatScroll.bind(this), 300)
     sbp('okTurtles.events/on', EVENT_HANDLED, this.listenChatRoomActions)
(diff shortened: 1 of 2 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -887,6 +887,28 @@ export default ({
...
+            if (scrollingToSpecificMessage) {
+              const newQuery = { ...this.$route.query }
+              delete newQuery.mhash
+              this.$router.replace({ query: newQuery })
+            }
+
+            // This ensures that scroll-to-specific-message (or to the latest message) happens one time after the initial chatroom load.
+            this.ephemeral.scrollHashOnInitialLoad = null
+          })
+        }
```
SebinSong, line 911: From more thorough investigation, this appears to be a better place to execute the `initial-scroll` logic. From my observation, executing this logic from the original place (L1092 below) seemed premature especially when the chatroom has many image attachments. This leads to a situation like below for example,

[1] Some messages are rendered.
[2] initial-scroll logic is executed.
[3] Images that are rendered later causes the scroll-height of the chatroom to increase.
[4] This leads to a sudden jump of the scroll position.

After moving this logic here, I have not observed this happening again.

taoeffect (commented):
Interesting! @SebinSong - I've verified that this seems to fix the problem on my end, nice work!

Does this count though as a logic fix? It seems to leave the logic the same and just mess with the timings a bit. Are you sure this will be sufficient under all circmstances?

SebinSong:
@taoeffect 
Yes, the logics for **1.** remembering the scroll-position in `onChatScroll` method **2.** determining the initial scroll-position during the initial message-chain render and auto-scroll to it are actually not broken and are in the same places as they were.

I've actually spent a while yesterday trying to fix the issue by moving 2. above to somewhere else, but these little adjustments to the delay values were the best fix I could find so far..

Let me maybe spend some more time exploring/trying the APIs `vue-infinite-loading` provides, just to see if there is a better fix and then give you a heads up if any changes. (Or you could merge this PR as is).

SebinSong:

There isn't anything useful in `vue-infinite-loading`'s API (it only provides 1 event which is`@infinite`), so I've actually had to come up with a logic using setTimeout/clearTimeout pair. From my local testing with relatively large size chatroom(80+ messages with lots of attachments), This fix works most consistently so far, so updated the PR with it.

Thanks.

taoeffect (approved):
Seems to work well, great job @SebinSong!

Issue #1900, opened 2024-03-28
Implement disappearing / ephemeral messages


taoeffect:
### Problem

Once messages are sent in a chatroom contract, they're kinda there forever, even if they're "deleted".

It would be better if there was at least an option to send a message that could be really deleted.

### Solution

Create a new type of message under `MESSAGE_TYPES`, to allow for ephemeral messages, where the contents are stored as a separate file attachment that can be deleted (EDT: @corrideat also suggests the idea of using the KV store instead, which could also work). See related issue #1827 for storing auto-deleting attachments (that needs to be closed first).

Then update the UI to support these new message types and implement them. If designs for ephemeral messages aren't ready, then just implement the feature, but don't hook it up to the UI - or make it a simple thing like an icon (💣) next to the input field that gets toggled, and when messages are sent with this icon enabled, they are sent as ephemeral message types. And when ephemeral message types are deleted, the underlying message (in its separate file form) gets deleted.

EDIT: The kind of UI I see for this is basically a dropdown with the 💣 icon that shows these options:

- 💣 (on delete)
- 💣 1hr
- 💣 6hr
- 💣 1d
- 💣 3d
- 💣 1w
- 💣 1m
- 💣 4m

And the UI would look something like this:

[ 💣 ][ (on delete) v ]

Where the [ 💣 ] is toggleble on/off (blue background highlight when on), and the dropdown is next to it which sets the setting.

corrideat:
I've been looking into this issue and think that going with files (as originally proposed) instead of KV (as I later suggested) is probably the simplest and most flexible solution (ETA: *).

However, something that this issue doesn't take into account is that merely deleting files may not be sufficient for a 'disappearing' feature, if copies of the file still exist. Similarly, because CIDs are used, even a 'deleted' message has some on-chain information about the file that was stored.

One way to address these issues is implementing something like the [`X3DH`](https://signal.org/docs/specifications/x3dh/) protocol for disappearing messages. Signal also has the 'double ratchet' protocol after a key is established, but we probably don't need this if we run the `X3DH` protocol for each message.

`X3DH` would require some additional keys (the `3` in the name alludes to a triple DH agreement), which could be stored in the KV store. The main advantages of using something like `X3DH` is that it's a solid and industry-standard protocol and that it provides the kind of assurances one would expect about disappearing messages, such as forward-secrecy. I still need to look up how `X3DH` works for group chats.

[...]

PR #2730, opened 2025-03-20
#2718 - Replying to a message shouldn't quote the entire message

### LINKED ISSUE #2718: Replying to a message shouldn't quote the entire message
### Problem

Image

When using the reply-to-message feature, currently it quotes all of the text of the original message. It shouldn't.

### Solution

Limit the quote to maybe the first 100 to 150 characters, whatever looks good to you, and also test it with all sorts of different inputs, as it seems to be using markdown rendering to display the text, and I think it shouldn't... (just the raw text instead).

Test with replying to messages containing URLs and code blocks etc.

EDIT: @dotmacro points out that this is how it's supposed to look and behave based on Figma: https://www.figma.com/design/mxGadAHfkWH6qApebQvcdN/Group-Income-2.0?node-id=23766-264680&t=FFLywG0TmltfWoYr-1

Image

SebinSong:
closes #2718 

As requested the replied message UI:

**1.** strips out all the markdown syntaxes. (eg  `"is **bold**"` ->  `"is bold"`, `"is ~strike-through~"` -> `"is strike-through"`)
**2.** truncate the text to 180 characters (Pls let me know if this limit should be changed)

truncated-replying-msg

---
Please note that the transformations **1.** and **2.** above are done **before** the message is sent and the intention is to reduce the message size in the DB. What I've noticed is that this `replyingMessage` field(below) of the chat message contains the whole text currently (The comment says `(if too long, truncate)` but it has not been implemented yet)

https://github.com/okTurtles/group-income/blob/22c56d5807d7875e00a245d855b23ed90d75f34e/frontend/model/contracts/shared/types.js#L82-L85

 If the replied message content is huge, this is clearly a waste of data (Imagine multiple users reply to this huge content message). 
But this also means that the fix by this PR will only be applied to the messages that are sent **after** the fix is deployed. Hope it sounds good.

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -155,7 +155,7 @@ import { proximityDate, MINS_MILLIS } from '@model/contracts/shared/time.js'
 import { cloneDeep, debounce, throttle, delay } from 'turtledash'
 import { EVENT_HANDLED } from '~/shared/domains/chelonia/events.js'
 import { compressImage } from '@utils/image.js'
-import { stripMarkdownSyntax } from '@view-utils/markdown-utils.js'
+import { swapMentionIDForDisplayname } from '@model/chatroom/utils.js'
 
 const ignorableScrollDistanceInPixel = 500
 
@@ -640,9 +640,14 @@ export default ({
       } else if (!text && attachments.length) {
         this.ephemeral.replyingMessage = { hash, text: attachments[0].name }
       } else {
-        this.ephemeral.replyingMessage = {
-          hash, text: stripMarkdownSyntax(text, CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS)
+        const sanitizeAndTruncate = text => {
+          return swapMentionIDForDisplayname(text)
+            .replace(/\s+/g, ' ') // Normalize spaces
+            .trim()
+            .slice(0, CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS)
         }
+
+        this.ephemeral.replyingMessage = { hash, text: sanitizeAndTruncate(text) }
       }
 
       this.ephemeral.replyingTo = isTypeInteractive ? L('Proposal notification') : this.who(message)
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -150,6 +150,7 @@ import { proximityDate, MINS_MILLIS } from '@model/contracts/shared/time.js'
 import { cloneDeep, debounce, throttle, delay } from 'turtledash'
 import { EVENT_HANDLED } from '~/shared/domains/chelonia/events.js'
 import { compressImage } from '@utils/image.js'
+import { stripMarkdownSyntax } from '@view-utils/markdown-utils.js'
 
 const ignorableScrollDistanceInPixel = 500
 
@@ -641,7 +642,7 @@ export default ({
         }
         this.ephemeral.replyingTo = L('Proposal notification')
       } else {
-        this.ephemeral.replyingMessage = { text, hash }
+        this.ephemeral.replyingMessage = { hash, text: stripMarkdownSyntax(text, 180) } // !@#
         this.ephemeral.replyingTo = this.who(message)
       }
     },
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -86,7 +86,7 @@
           :isGroupCreator='isGroupCreator'
           :class='{removed: message.delete}'
           @retry='retryMessage(index)'
-          @reply='replyMessage(message)'
+          @reply='replyToMessage(message)'
           @scroll-to-replying-message='scrollToMessage(message.replyingMessage.hash)'
           @edit-message='(newMessage) => editMessage(message, newMessage)'
           @pin-to-channel='pinToChannel(message)'
@@ -143,7 +143,12 @@ import ViewArea from './ViewArea.vue'
 import Emoticons from './Emoticons.vue'
 import TouchLinkHelper from './TouchLinkHelper.vue'
 import DragActiveOverlay from './file-attachment/DragActiveOverlay.vue'
-import { MESSAGE_TYPES, MESSAGE_VARIANTS, CHATROOM_ACTIONS_PER_PAGE, CHATROOM_MEMBER_MENTION_SPECIAL_CHAR } from '@model/contracts/shared/constants.js'
+import {
+  MESSAGE_TYPES, MESSAGE_VARIANTS,
+  CHATROOM_ACTIONS_PER_PAGE,
+  CHATROOM_MEMBER_MENTION_SPECIAL_CHAR,
+  CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS
+} from '@model/contracts/shared/constants.js'
 import { CHATROOM_EVENTS, NEW_CHATROOM_UNREAD_POSITION, DELETE_ATTACHMENT_FEEDBACK } from '@utils/events.js'
 import { findMessageIdx } from '@model/contracts/shared/functions.js'
 import { proximityDate, MINS_MILLIS } from '@model/contracts/shared/time.js'
@@ -401,15 +406,7 @@ export default ({
 
       const data = { type: MESSAGE_TYPES.TEXT, text }
       if (replyingMessage) {
-        // NOTE: If not replying to a message, use original data; otherwise, append replyingMessage to data.
         data.replyingMessage = replyingMessage
-        // NOTE: for the messages with only images, the text should be updated with file name
-        if (!replyingMessage.text) {
-          const msg = this.messages.find(m => (m.hash === replyingMessage.hash))
-          if (msg) {
-            data.replyingMessage.text = msg.attachments[0].name
-          }
-        }
       }
 
       const sendMessage = (beforePrePublish) => {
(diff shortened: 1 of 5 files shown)


frontend/views/utils/markdown-utils.js
```diff
@@ -193,3 +194,20 @@ export function combineMarkdownSegmentListIntoString (
...
+
+export function stripMarkdownSyntax (markdownString: string, truncateTo: number = -1): string {
+  markdownString = swapMentionIDForDisplayname(markdownString) // eg. '@identityContractID' -> '@user1'
+
+  const sanitized = markdownString
+    .replace(/\*\*(.*?)\*\*/g, '$1') // 'bold'
+    .replace(/_(.*?)_/g, '$1') // 'italic'
+    .replace(/~(.*?)~/g, '$1') // 'strike-through'
+    .replace(/`(.*?)`/g, '$1') // 'inline code'
+    .replace(/```/g, '') // 'code block'
```
taoeffect: In this order, it produces this result (not fully stripped):

Screenshot 2025-03-19 at 6 50 56 PM

Swapping these two (doing ``` first) produces:

Screenshot 2025-03-19 at 6 51 56 PM

Which seems more correct to me since it actually removes the formatting, which is what we want a function called `stripMarkdownSyntax` to do.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -629,21 +627,25 @@ export default ({
...
         }
-        this.ephemeral.replyingTo = L('Proposal notification')
+      } else if (!text && attachments.length) {
+        this.ephemeral.replyingMessage = { hash, text: attachments[0].name }
       } else {
-        this.ephemeral.replyingMessage = { text, hash }
-        this.ephemeral.replyingTo = this.who(message)
+        this.ephemeral.replyingMessage = {
+          hash, text: stripMarkdownSyntax(text, CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS)
+        }
```
taoeffect: Another thought is: you could also display the markdown, just not render it. What do you think?

SebinSong: That will simplify things. I can make this update.

frontend/views/utils/markdown-utils.js
```diff
@@ -193,3 +194,20 @@ export function combineMarkdownSegmentListIntoString (
...
+    .replace(/~(.*?)~/g, '$1') // 'strike-through'
+    .replace(/```/g, '') // 'code block'
+    .replace(/`(.*?)`/g, '$1') // 'inline code'
+    .replace(/\[(.*?)\]\((.*?)\)/g, '$1') // links ([text](url) -> text)
+    .replace(/^>\s*/gm, '') // block-quote
+    .replace(/\s+/g, ' ') // Normalize spaces
+    .trim()
+
+  return truncateTo > 0 ? sanitized.slice(0, truncateTo) : sanitized
+}
```
SebinSong, line 213: Let me not remove this function though, in case this can be useful in the future.

frontend/views/utils/markdown-utils.js
SebinSong: done it

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -629,21 +627,30 @@ export default ({
...
+      if (isTypeInteractive) {
         const proposal = message.proposal
 
         this.ephemeral.replyingMessage = {
-          text: interactiveMessage(proposal, { from: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${proposal.creatorID}` }),
-          hash
+          hash, text: interactiveMessage(proposal, { from: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${proposal.creatorID}` })
         }
-        this.ephemeral.replyingTo = L('Proposal notification')
+      } else if (!text && attachments.length) {
```
taoeffect: The message contracts define `attachments` as an `optional`, so could accessing `attachments.length` could result in a bug?

If so, then either here we need `attachments?.length` or above a default value:

```js
const { text, hash, type, attachments = [] } = message
```

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
```diff
@@ -74,6 +75,9 @@ export default ({
         ...Object.keys(this.ourContactProfilesById).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
         makeChannelMention('[^\\s]+', true) // chat-mention as contractID has a format of `#:chatID:...`. So target them as a pattern instead of the exact strings.
       ]
+    },
+    showTrailingEllipsis () {
+      return this.isReplyingMessage && this.text.length === CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS
```
taoeffect, line 80: Probably want `>` instead of `===` right?

Interestingly enough, this doesn't seem to prevent the `...` from being shown for long messages, and I'm not sure why.

I did notice though that the ellipses is sometimes placed on a newline when it probably shouldn't be:

Screenshot 2025-03-20 at 10 26 13 AM

taoeffect (commented):
Great work @SebinSong!

It seems to work fairly well. I left a few comments with other potential issues that I noticed with the help of LLMs, please have a look.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: It's impossible that a message has no `text` content but does not have `attachments` either. (The `send` button won't be enabled in this case) But let me still go ahead and  do that `attachments?.length` here though.

corrideat: > It's impossible that a message has no text content but does not have attachments either

If you want to make this assumption, I strongly suggest that you either add this to the chatroom contract validation, or else that you don't make the assumption.

SebinSong: @corrideat Sounds a bit rude man... what's your intention of randomly popping up and saying that..?

corrideat: Sorry, I missed the second part of the message where you said you'll make the change anyhow, so I can see why it could come across like that. I was looking through open PRs and saw the part about the send button not being enabled and somehow missed the second part. Anyhow, the motivation was that the chatroom contract will be used by other applications possibly, and for robustness it should not assume things that aren't part of the contract.

SebinSong: @corrideat No worries.

> the motivation was that the chatroom contract will be used by other applications possibly, and for robustness it should not assume things that aren't part of the contract.

This sounds like an important point too.

corrideat: That'll teach me to make sure I read everything in full. BTW, another situation where this situation (attachments but no message body) might arise is https://github.com/okTurtles/group-income/issues/1900. However, we might need to treat that especially to work with this PR anyhow.

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
SebinSong: @taoeffect As I mentioned in [this PR description](https://github.com/okTurtles/group-income/pull/2730#issue-2933584308), truncating the replied message to the max 180 characters are done before the message is sent. This means the replied message being displayed in the message chain won't be longer than 180 characters and that's why. So I don't think this part needs a further change. Let me know otherwise though.

taoeffect: Well, one thought I had is that what if some time in the future we decide to change `CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS` to make it shorter?

Then `...` won't be displayed after older messages that we're replying to that are longer than that. So that might be a situation where we'd still want to use `>=` just to be safe.

Also, what about the image above that shows the ellipses on a newline? Is there a way to make sure that it's not on a newline by itself like that? It's not a big deal TBH, we could leave it as-is, but as a kind of attention-to-detail polish I think it would look better if it remained on the same line as the text.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @corrideat What I would love to mention here though is how things can sound. I've actually felt this a few times before that when you wrote `Hey I think this is wrong and you should do this` kinds of comments, they felt kind of disrespecful for some reason(especially when apparently there were no request for you to review the PR, even though I usually appreciate other dev's feedbacks). Maybe it's just that you just tend to say things bluntly and I've been more used to co-worker's style trying to be more polite than usual when they want to say someone else is wrong and they are right. But I can see you have no intention to harm other person's feeling so, no worries bro!

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
SebinSong: >  I think it would look better if it remained on the same line as the text.

This was my first approach actually and then I decided to make it `display: block`. I can revert it back.

SebinSong: updated so that the ellipsis is inline rather than a block.

image

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -376,7 +376,9 @@ export default ({
       }
     },
     replyingMessageText (message) {
-      return message.replyingMessage?.text || ''
+      return message.replyingMessage?.text
+        ? message.replyingMessage.text.slice(0, CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS)
+        : ''
```
SebinSong, line 381: @taoeffect 
 Truncating the text once more here will ensure that below concern of yours won't happen.

> Well, one thought I had is that what if some time in the future we decide to change
> CHATROOM_REPLYING_MESSAGE_LIMITS_IN_CHARS to make it shorter?
> Then ... won't be displayed after older messages that we're replying to that are longer than that.

taoeffect: I changed the value from 180 to 10, saved, and then to 200 (saved). Here you can see the result:

Screenshot 2025-03-21 at 9 58 09 AM

It's not a big deal though, not worth spending more time on so I'll go ahead and merge!

taoeffect (approved):
Nicely done @SebinSong!

PR #2731, opened 2025-03-20
#2729 - update `currentChatRoomReadUntil` when the latest message is deleted

### LINKED ISSUE #2729: `currentChatRoomReadUntil` getter doesn't get updated when the most recent message is deleted
### Problem
in the scroll event listener of `ChatMain.vue`, there is a logic that remembers to which point in the message chain the user has read:

https://github.com/okTurtles/group-income/blob/22c56d5807d7875e00a245d855b23ed90d75f34e/frontend/views/containers/chatroom/ChatMain.vue#L180-L188

This functions well in most cases but if the most recent message is deleted, this stops updating because `latestMessageCreatedHeight <= bottomMessageCreatedHeight` condition can never be `true` here.

(`latestMessageCreatedHeight` here is the `createdHeight` value of the deleted message and `bottomMessageCreatedHeight` is the value of  **the second most** recent message. So `latestMessageCreatedHeight` is always `> bottomMessageCreatedHeight` here temporarily until a new message is sent or received)

This can lead to temporarily breaking `auto-scroll-to-the-bottom when chatroom is initialised` functionality. (Once the user sends or receives a new message, this is gone. Hence the term 'temporarily' here.)

### Solution
Update the logic to take this case into account. (eg. `currentChatRoomReadUntil ` should be updated to the second most recent message if the latest one is deleted)

SebinSong:
closes #2729 

**[Before fix]**

Delete the latest message -> visit another channel -> come back to `#general` -> **does not** scroll to the bottom. 

https://github.com/user-attachments/assets/63aacabe-e458-49c7-b7a0-7589a8aaae6b

**[After fix]**

Delete the latest message -> visit another channel -> come back to #general -> **does** scroll to the bottom as expected.

https://github.com/user-attachments/assets/28b914e6-1fde-42fc-aa14-1cbb959c775c

### DIFF
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -82,14 +82,17 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
-  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight }: {
-    contractID: string, messageHash: string, createdHeight: number
+  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight, forceUpdate = false }: {
+    contractID: string,
+    messageHash: string,
+    createdHeight: number,
+    forceUpdate: boolean // In a rare case, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'. (reference: https://github.com/okTurtles/group-income/issues/2729)
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const getUpdatedUnreadMessages = async () => {
         const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
 
-        if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
+        if (forceUpdate || currentData[contractID]?.readUntil.createdHeight < createdHeight) {
           const { unreadMessages } = currentData[contractID]
           return {
             ...currentData,
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -86,7 +86,11 @@ export default (sbp('sbp/selectors/register', {
     contractID: string,
     messageHash: string,
     createdHeight: number,
-    forceUpdate: boolean // In a rare case, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'. (reference: https://github.com/okTurtles/group-income/issues/2729)
+    // In a rare case, such as when the latest message is deleted,
+    // the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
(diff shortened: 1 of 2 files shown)


frontend/controller/actions/identity-kv.js
```diff
@@ -82,14 +82,17 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
-  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight }: {
-    contractID: string, messageHash: string, createdHeight: number
+  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight, forceUpdate = false }: {
+    contractID: string,
+    messageHash: string,
+    createdHeight: number,
+    forceUpdate: boolean // In a rare case, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'. (reference: https://github.com/okTurtles/group-income/issues/2729)
```
corrideat: This doesn't really say much other than that it's a 'rare case', which isn't very useful or informative. (Yes, I know the issue is referenced, but that requires extra steps and the issue also doesn't explain anything about `forceUpdate`).

If you'll add a comment here, can you say what the 'rare case' is, or if you want to be more general, say, 'in some (rare) cases, such as when the latest message is deleted, ...'? That gives immediate context to someone reading the code.

Also, the line is pretty long and the comment would likely be better placed above 89, possibly split across multiple lines.

(The motivation for this is that in this case it's possible to provide a concise and self-contained explanation of the entire context, including the what and the why for `forceUpdate`).

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -948,15 +974,18 @@ export default ({
...
     },
-    updateReadUntilMessageHash ({ messageHash, createdHeight }) {
+    updateReadUntilMessageHash ({ messageHash, createdHeight, forceUpdate = false }) {
       const chatRoomID = this.ephemeral.renderingChatRoomId
       if (chatRoomID && this.isJoinedChatRoom(chatRoomID)) {
-        if (this.currentChatRoomReadUntil?.createdHeight >= createdHeight) {
-          // NOTE: skip adding useless invocations in KV_QUEUE queue
+        if (!forceUpdate && this.currentChatRoomReadUntil?.createdHeight >= createdHeight) {
+          // NOTE: skip adding useless invocations in KV_QUEUE queue.
+          //       'forceUpdate' flag here is for the rare case where the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
```
corrideat: Also here, adding at the end ', such as when the latest message is deleted' would clarify.

frontend/controller/actions/identity-kv.js
SebinSong: @corrideat valid point.

taoeffect (approved):
Fantastic work @SebinSong!

PR #2732, opened 2025-03-20
Timestamp meta on server

### LINKED ISSUE #2695: Add timestamp to GIMessage/SPMessage
### Problem

In some future use cases, e.g. especially search, it may be very useful to have GIMessages be timestamped.

For example, because search could be very slow compared to apps where the data is unencrypted on the server, one way to speed up searching for old messages would be to narrow the search range by timestamps. If the person knows roughly when the thing they're searching for was said, they could query the server for messages only in that time range.

### Solution

Add an ISOString timestamp to the GIMessage header.

Also update contracts to use this new value instead of `createdDate`.

corrideat:
Closes #2695.

**NB! The `date` header isn't implemented for `GET /file` due to limitations on doing this using the `_private_hidx=` key, which is based on the contract ID.** The `date` header for `/file` could be implemented, but using a key based on the CID (hash) of that particular message.

> [!CAUTION]
> This PR requires re-creating the database due to incompatible use some index keys.

### DIFF
--- a/backend/database.js
+++ b/backend/database.js
@@ -52,7 +52,7 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let { currentHash, ...serverMeta } = (
+    let { hash: currentHash, ...serverMeta } = (
       await sbp('chelonia/db/getEntryMeta', contractID, height)
     ) || {}
     let prefix = ''
@@ -69,13 +69,13 @@ export default ((sbp('sbp/selectors/register', {
           this.destroy()
           return
         }
-        if (!!currentHash && counter < limit) {
+        if (currentHash && counter < limit) {
           sbp('chelonia/db/getEntry', currentHash).then(async entry => {
             if (entry) {
               const currentPrefix = prefix
               prefix = ','
               counter++
-              const { currentHash: newCurrentHash, ...newServerMeta } = (
+              const { hash: newCurrentHash, ...newServerMeta } = (
                 await sbp('chelonia/db/getEntryMeta', contractID, entry.height() + 1)
               ) || {}
               this.push(
--- a/backend/database.js
+++ b/backend/database.js
@@ -41,7 +41,7 @@ if (!fs.existsSync(dataFolder)) {
 
 // Streams stored contract log entries since the given entry hash (inclusive!).
 export default ((sbp('sbp/selectors/register', {
-  'backend/db/streamEntriesAfter': async function (contractID: string, height: string, requestedLimit: ?number, options: { keyOps?: boolean } = {}): Promise<*> {
+  'backend/db/streamEntriesAfter': async function (contractID: string, height: string, requestedLimit: ?number): Promise<*> {
     const limit = Math.min(requestedLimit ?? Number.POSITIVE_INFINITY, process.env.MAX_EVENTS_BATCH_SIZE ?? 500)
     const latestHEADinfo = await sbp('chelonia/db/latestHEADinfo', contractID)
     if (latestHEADinfo === '') {
@@ -52,9 +52,8 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let currentHeight = parseInt(height)
     let { currentHash, ...serverMeta } = (
-      await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
+      await sbp('chelonia/db/getEntryMeta', contractID, height)
     ) || {}
     let prefix = ''
     let ended = false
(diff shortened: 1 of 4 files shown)


corrideat:
The last commit also implements the 'parallel keychain' that'll be used in snapshots in the following way (added to this PR as it uses the ability to add metadata to messages):

When a key operation is added to the chain, the `isKeyOp` flag is added to that message. When requesting an event stream from the server, one could add `?keyOps=true` to receive only key ops.

This approach is simple (requires minimal changes, doesn't store more information than is needed), general (i.e., it can be used with other filters and combined with other future filters) and effective (does what it should do and is indistinguishable from other approaches, from a results PoV) but has the drawback that it still requires that the server loops over many messages (however, this is relatively fast, as only index keys are used in this loop and irrelevant events are skipped over). The big-O complexity is still the same (linear).

I think that this solution offers good tradeoffs in terms of efficiency and speed, but there are other alternative solutions that I considered and dismissed.

1. New index of `keyOps` (using height). Instead of using an `isKeyOp` flag on messages, we could add these messages to an index, containing the height. For example, `_private_keyops_${cID}`, containing, e.g., `1,7,25,654,789`. This approach allows for skipping over entire blocks, but has these downsides: requires a new DB key, requires an additional lookup for fetching metadata, requires (even more) special logic for event streaming and the index may require too much memory.
2. New index of `keyOps` (using CIDs). Similar tradeoffs to using height, except that it requires even more space and requires more lookups for event streaming (for example, to get events after a certain height).
3. Entirely parallel chain. This could potentially work but requires duplicating data, which wastes space and introduces the possibility of mismatches. It also requires more work when writing data (to write to both chains) and requires special logic for event streaming (to know which chain to read from).
4. Indirect parallel chain. Similar to 3 but using references to avoid duplicating data. Other than saving space, it has similar tradeoffs than the 'entirely parallel chain', and requires additional lookups for each key.

shared/domains/chelonia/db.js
```diff
@@ -101,6 +101,17 @@ const dbPrimitiveSelectors = process.env.LIGHTWEIGHT_CLIENT === 'true'
 
 export default ((sbp('sbp/selectors/register', {
   ...dbPrimitiveSelectors,
+  'chelonia/db/getEntryMeta': async (contractID: string, height: number) => {
+    const entryMetaJson = await sbp('chelonia.db/get', `_private_hidx=${contractID}#${height}`)
+    if (!entryMetaJson) return
+
+    const entryMeta = JSON.parse(entryMetaJson)
+    return entryMeta
```
taoeffect: Can be just:

```js
 return JSON.parse(entryMetaJson)
```

shared/domains/chelonia/db.js
```diff
@@ -157,7 +168,10 @@ export default ((sbp('sbp/selectors/register', {
       await sbp('chelonia.db/set', entry.hash(), entry.serialize())
       await sbp('chelonia.db/set', getLogHead(contractID), JSON.stringify({ HEAD: entry.hash(), height: entry.height() }))
       console.debug(`[chelonia.db] HEAD for ${contractID} updated to:`, entry.hash())
-      await sbp('chelonia.db/set', `_private_hidx=${contractID}#${entryHeight}`, entry.hash())
+      await sbp('chelonia/db/setEntryMeta', contractID, entryHeight, {
+        currentHash: entry.hash(),
```
taoeffect: Could we name this simply `hash`? To me that makes more sense than `currentHash`, as it's the hash for this particular entry, and it's also nicely the same length as `date` too.

Above `currentHash` would still make sense though, e.g.:

```js
let { hash: currentHash, ...serverMeta } = (
```

backend/database.js
```diff
@@ -52,38 +52,52 @@ export default ((sbp('sbp/selectors/register', {
...
+        this.push('[')
+        cb()
+      },
       read (): void {
         if (ended) {
           this.destroy()
           return
         }
-        if (currentHash && counter < limit) {
+        if (!!currentHash && counter < limit) {
```
taoeffect: Is this `!!` really needed? (Why?)

taoeffect (changes requested):
Excellent work @corrideat!

Some minor change requests and some clarifying questions for you.

backend/database.js
corrideat: No, it's not really needed (as in it behaves exactly the same with or without), more of a style choice (trying to be explicit where possible, this makes it explicit it's coerced into a boolean).

Let me know if I should remove it.

shared/domains/chelonia/db.js
corrideat: Agreed that `hash` is a better name.

backend/database.js
taoeffect: I mean, if you're asking for my preference, I prefer it the way it was before, because if it works just the same, we save a couple opcodes (I'm neurotic like that) and it seems more readable to me personally

corrideat: > we save a couple opcodes (I'm neurotic like that) and it seems more readable to me personally

I'll remove it then if it's your preference, but, personally, I disagree that it saves opcodes or that it's more readable (if you know what `!!` does).

1. When implementing a JS interpreter, you'd need to check the type and see if it's truthy or falsy here. Whether you end up saving opcodes or not is highly speculative.
2. Readability is highly subjective, but I'd argue that `!!x` requires less cognitive context than `x` for a non-boolean type.

taoeffect: > When implementing a JS interpreter, you'd need to check the type and see if it's truthy or falsy here. Whether you end up saving opcodes or not is highly speculative.

Interesting. Didn't know that!

taoeffect (approved):
Nice work!

PR #2734, opened 2025-03-22
Ability to specify which KV events to subscribe to

### LINKED ISSUE #2727: Add ability to specify what types of events we want to subscribe for when subscribing to contracts
### Problem

Image

Currently when subscribing to contracts (for example, identity contracts), we request to receive all events, including events we might not be interested in, for example, KV updates that we can't read (see screenshot above).

This can generate unnecessary console errors and unnecessary traffic.

### Solution

Add ability to pass in options to tell the server what specific types of events we're interested in listening for.

Perhaps we're only interested in listening for specific KV keys, or no keys at all.

corrideat:
Closes #2727

### DIFF
--- a/backend/pubsub.js
+++ b/backend/pubsub.js
@@ -305,6 +305,7 @@ const defaultMessageHandlers = {
     } else {
       log.debug('Already subscribed to', channelID)
     }
+    // $FlowFixMe[incompatible-call]
     socket.send(createOkResponse({ type: SUB, channelID, kvFilter }))
   },
 
@@ -327,6 +328,7 @@ const defaultMessageHandlers = {
     } else {
       log.debug('Not subscribed to', channelID)
     }
+    // $FlowFixMe[incompatible-call]
     socket.send(createOkResponse({ type: KV_FILTER, channelID, kvFilter }))
   },
 
--- a/backend/pubsub.js
+++ b/backend/pubsub.js
@@ -26,7 +26,7 @@ import { postEvent } from './push.js'
 const { bold } = require('chalk')
 const WebSocket = require('ws')
 
-const { PING, PONG, PUB, SUB, UNSUB } = NOTIFICATION_TYPE
+const { PING, PONG, PUB, SUB, UNSUB, KV_FILTER } = NOTIFICATION_TYPE
 const { ERROR, OK } = RESPONSE_TYPE
 
 const defaultOptions = {
@@ -170,6 +170,7 @@ const defaultServerHandlers = {
     socket.pinged = false
     socket.server = server
     socket.subscriptions = new Set()
+    socket.kvFilter = new Map()
     socket.ip = request.headers['x-real-ip'] ||
       request.headers['x-forwarded-for']?.split(',')[0].trim() ||
       request.socket.remoteAddress
@@ -280,7 +281,7 @@ const defaultMessageHandlers = {
     server.broadcast(msg, { to: subscribers ?? [] })
   },
 
-  [SUB] ({ channelID }: SubMessage) {
+  [SUB] ({ channelID, kvFilter }: SubMessage) {
     const socket = this
     const { server } = this
 
@@ -293,6 +294,9 @@ const defaultMessageHandlers = {
     if (!socket.subscriptions.has(channelID)) {
       // Add the given channel ID to our subscriptions.
       socket.subscriptions.add(channelID)
+      if (Array.isArray(kvFilter)) {
+        socket.kvFilter.set(channelID, new Set(kvFilter))
+      }
       if (!server.subscribersByChannelID[channelID]) {
         server.subscribersByChannelID[channelID] = new Set()
       }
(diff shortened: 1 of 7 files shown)


backend/pubsub.js
```diff
@@ -395,11 +424,19 @@ const publicMethods = {
...
     if (channelID in server.subscribersByChannelID) {
-      yield * server.subscribersByChannelID[channelID]
+      const subscribers = server.subscribersByChannelID[channelID]
+      if (!kvKey) {
+        yield * subscribers
+      } else {
+        for (const subscriber of subscribers) {
+          const kvFilter = subscriber.kvFilter.get(kvKey)
+          if (!kvFilter || kvFilter.has(kvKey)) yield subscriber
+        }
```
taoeffect: This should be `channelID` instead of `kvKey`, right?

```js
for (const subscriber of subscribers) {
  const filterSet = subscriber.kvFilter.get(channelID)
  if (!filterSet || filterSet.has(kvKey)) yield subscriber
}
```

corrideat: You're right. Fixed.

frontend/setupChelonia.js
```diff
@@ -273,6 +275,25 @@ const setupChelonia = async (): Promise<*> => {
...
+    // Wait for the added contracts to be ready, then call the update function
+    if (!added.length) return
+    const rootState = sbp('chelonia/rootState')
+    added.forEach((cID) => {
+      switch (rootState.contracts[cID]?.type) {
+        case 'gi.contracts/identity':
+          if (cID === rootState.loggedIn?.identityContractID) {
+            sbp('chelonia/contract/setKvFilter', cID, [KV_KEYS.UNREAD_MESSAGES, KV_KEYS.PREFERENCES, KV_KEYS.NOTIFICATIONS])
+          }
+          break
```
taoeffect: This used to say `return`, which I think is correct, as now we will first set the filter for `cID` to `[KV_KEYS.UNREAD_MESSAGES, KV_KEYS.PREFERENCES, KV_KEYS.NOTIFICATIONS]` and then overwrite that with `[]` down below.

taoeffect: I think what you want is a `return` inside of the `if`, and a `break` here as you have it.

corrideat: Yes, that sounds about right. The way it was before it was setting no filter on other identity contracts

shared/domains/chelonia/chelonia.js
```diff
@@ -427,6 +427,9 @@ export default (sbp('sbp/selectors/register', {
     const rootState = sbp(this.config.stateSelector)
     return !!rootState?.secretKeys && has(rootState.secretKeys, keyId)
   },
+  'chelonia/contract/setKvFilter': function (contractID: string, filter?: string[]) {
```
taoeffect: Actually, I think it would make more sense to group this with the other KV selectors.

Could you rename this to `'chelonia/kv/setFilter'`?

backend/pubsub.js
```diff
@@ -301,7 +305,31 @@ const defaultMessageHandlers = {
     } else {
       log.debug('Already subscribed to', channelID)
     }
-    socket.send(createOkResponse({ type: SUB, channelID }))
+    // $FlowFixMe[incompatible-call]
+    socket.send(createOkResponse({ type: SUB, channelID, kvFilter }))
```
taoeffect, line 312: `o3-mini-high` isn't happy with these `// $FlowFixMe[incompatible-call]` comments and asks if casts can be used, e.g. `(createOkResponse({ … }): string)`

shared/domains/chelonia/chelonia.js
taoeffect: Oh, also could you add a comment above this showing both the syntax for setting a filter and removing it?

I think the syntax for removing all filters is to call `setFilter` with a single argument (the `contractID`, right?), but that's not documented here and isn't clear just from looking at this.

backend/pubsub.js
```diff
@@ -301,7 +305,31 @@ const defaultMessageHandlers = {
...
+      return
+    }
+    if (socket.subscriptions.has(channelID)) {
+      if (Array.isArray(kvFilter)) {
+        socket.kvFilter.set(channelID, new Set(kvFilter))
+      } else {
+        socket.kvFilter.delete(channelID)
+      }
+    } else {
+      log.debug('Not subscribed to', channelID)
```
taoeffect: Could you prefix this string with something like `[KV_FILTER]` to give it more context?

taoeffect (changes requested):
Great work @corrideat!

Review ready! Also, if you could please investigate why that test is failing... 🤔

backend/pubsub.js
corrideat: No, this isn't a cast issue. The reason it fails is because `kvFilter` can be undefined, and JSONType doesn't allow undefined. Trying to add `undefined` to JSONType also results in errors, so I figured this is fine until switching to TypeScript..

corrideat: (serialising `undefined` as JSON is fine, it just omits the field, unlike `null`, which keeps the field with  a `null` value)

taoeffect: Ah OK 👍

corrideat:
> if you could please investigate why that test is failing... 🤔

I think the failure is either unrelated, or, more likely, related to the bug introduced by setting the `[]` filter on the identity contract (the `return` -> `break` issue). This would prevent getting updates for `unreadMessages`, which is the error in the tests.

taoeffect (approved):
Fantastic work on this @corrideat!

PR #2741, opened 2025-04-03
#2728 - Truncate long message in chat

### LINKED ISSUE #2728: Truncate long messages in chat
### Problem

Messages can be very long (vertically). This can make it difficult to skim a large number of messages if a single message takes up the entire screen (or more).

### Solution

Detect whether a message is taking more than a certain amount of vertical space. Check how for example, Slack and/or Discord handle long messages at different window heights. Do they do it by a fixed pixel amount, or a percentage of the viewport area?

Then if it does exceed that length, encapsulate it inside of a container with a "show more" button, that when clicked expands the viewport to display the full message, and then turns into a "collapse" button to shrink it back down again.

SebinSong:
closes #2728 

**[ screenshot1 - very large text only message ]**

<br><br>

**[ screenshot2 - large message due to big images ]**

truncated_attachment

<br><br>

**[ screenshot3 - large markdown ]**

long-content-by-markdown

<br><br>

**[ screenshot4 - large chat-poll ]**

truncated_large-poll

---
@taoeffect 
regarding the 'show more/less' button design, my initial attempt was to make it as a button on a linear-gradient background, which is similar to the example video you posted [here](https://github.com/okTurtles/group-income/issues/2728#issuecomment-2762200769).
This looked fine for a dark background just like in the video, but could not really find gradient style that looks presentable enough for the light-theme of the app(No matter what I tried, they all looked ugly). So I changed to the current one.
This is the best I could come up with but if you would like another design, maybe we could discuss it with a new designer the company will have. Then I can come back to update this in the future.

### DIFF
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -22,8 +22,8 @@ export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 2 // in px
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 3 // The value of mobile is more tolerant considering smaller screen size.
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 1.5 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 2 // The value of mobile is more tolerant considering smaller screen size.
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -22,6 +22,8 @@ export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 2 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 3 // The value of mobile is more tolerant considering smaller screen size.
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -22,8 +22,8 @@ export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 1.25 // in px
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 1.5 // The value of mobile is more tolerant considering smaller screen size.
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 500 * 1.25 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 500 * 1.5 // The value of mobile is more tolerant considering smaller screen size.
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -22,8 +22,8 @@ export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 1.5 // in px
-export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 2 // The value of mobile is more tolerant considering smaller screen size.
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 1.25 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 1.5 // The value of mobile is more tolerant considering smaller screen size.
 
 export const TextObjectType = {
   Text: 'TEXT',
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -105,14 +105,17 @@
     @unpinFromChannel='$emit("unpin-from-channel")'
   )
 
-  .c-truncate-toggle-container(v-if='ephemeral.truncateToggle.enabled')
+  .c-truncate-toggle-container(
+    v-if='ephemeral.truncateToggle.enabled'
+    :class='{ "should-indent": isTypeText }'
+  )
     button.is-unstyled.c-truncate-toggle(
       :class='{ "is-showing": ephemeral.truncateToggle.isShowingAll }'
       type='button'
       @click.stop='toggleMessageTruncation'
     )
-      i18n(v-if='ephemeral.truncateToggle.isShowingAll') Show less
-      i18n(v-else) Show more
+      i18n.c-btn-text(v-if='ephemeral.truncateToggle.isShowingAll') Show less
+      i18n.c-btn-text(v-else) Show more
       i.icon-angle-down
 </template>
 
@@ -300,15 +303,19 @@ export default ({
 
       if (msgBodyEl) {
         const height = msgBodyEl.clientHeight
-        console.log('!@# msgHeight: ', height)
 
         if (height > threshold) {
           this.ephemeral.truncateToggle.enabled = true
         }
       }
     },
-    toggleMessageTruncation () {
-      this.ephemeral.truncateToggle.isShowingAll = !this.ephemeral.truncateToggle.isShowingAll
+    toggleMessageTruncation (e) {
+      const currentValue = this.ephemeral.truncateToggle.isShowingAll
+      this.ephemeral.truncateToggle.isShowingAll = !currentValue
+
+      if (currentValue) {
+        this.$el.scrollIntoView({ behavior: 'instant' })
+      }
     }
   },
   watch: {
@@ -503,18 +510,11 @@ export default ({
 
 .c-message.has-truncate-toggle {
   padding-bottom: 2.5rem;
-  --c-container-bg: linear-gradient(to bottom, rgba(0, 0, 0, 0) 45%, var(--c-bg-color) 95%);
-  --c-container-bg-opacity: 0.57;
-
-  .is-dark-theme & {
-    --c-container-bg: linear-gradient(to bottom, rgba(0, 0, 0, 0) 27.5%, var(--c-bg-color) 87.5%);
-    --c-container-bg-opacity: 0.825;
-  }
 
   .c-body,
   .c-full-width-body {
     &.is-truncated {
-      max-height: 36rem;
+      max-height: 28rem;
       overflow-y: hidden;
     }
   }
@@ -531,25 +531,36 @@ export default ({
     padding-bottom: 0.5rem;
     pointer-events: none;
 
+    &.should-indent {
+      padding-left: 4.25rem;
+    }
+
     button.c-truncate-toggle {
       z-index: 1;
       pointer-events: initial;
       color: $primary_0;
       font-size: $size_5;
-      padding: 0 1rem;
       display: inline-flex;
       align-items: center;
       justify-content: center;
       column-gap: 6px;
 
-      i {
-        font-size: 1.15em;
+      &:hover,
+      &:focus,
+      &:focus-within {
+        .c-btn-text {
+          text-decoration: underline;
+        }
       }
 
       &.is-showing {
         i {
           transform: rotate(180deg);
         }
+      }
+
+      i {
+        font-size: 1.15em;
       }
     }
   }
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -370,7 +370,6 @@ export default ({
 @import "@assets/style/_variables.scss";
 
 .c-message {
-  --c-bg-color: var(--background_0);
   padding: 0.5rem 1rem;
   scroll-margin: 20px;
 
@@ -411,7 +410,6 @@ export default ({
   }
 
   &:hover {
-    --c-bg-color: var(--general_2);
     background-color: $general_2;
 
     &:not(.pending, .failed) {
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -61,6 +61,7 @@
           :createdAt='datetime'
           :isGroupCreator='isGroupCreator'
           @delete-attachment='deleteAttachment'
+          @image-attachments-render-complete='checkMessageBodyHeight'
         )
 
       .c-failure-message-wrapper
@@ -120,6 +121,7 @@ import {
 } from '~/frontend/utils/constants.js'
 import { OPEN_TOUCH_LINK_HELPER } from '@utils/events.js'
 import { L, LTags } from '@common/common.js'
+import { getFileType } from '@view-utils/filters.js'
 
 export default ({
   name: 'MessageBase',
@@ -180,6 +182,10 @@ export default ({
     ...mapGetters(['userDisplayNameFromID']),
     hasAttachments () {
       return Boolean(this.attachments?.length)
+    },
+    hasImageAttachment () {
+      return Array.isArray(this.attachments) &&
+        this.attachments.some(attachment => getFileType(attachment.mimeType) === 'image')
     },
     isAlreadyPinned () {
       return !!this.pinnedBy
@@ -257,7 +263,7 @@ export default ({
     },
     checkMessageBodyHeight () {
       const msgBodyEl = this.isTypePoll ? this.$refs.msgFullWidthBody : this.$refs.msgBody
-      const threshold = this.chatMainConfig.isMobile
+      const threshold = this.chatMainConfig.isPhone
         ? CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE
         : CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP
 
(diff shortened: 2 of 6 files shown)


frontend/utils/constants.js
```diff
@@ -22,6 +22,8 @@ export const KILOBYTE = 1 << 10
 export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 750 * 1.25 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 750 * 1.5 // The value of mobile is more tolerant considering smaller screen size.
```
SebinSong: Pls let me know if we would love to change these values. As I said in [this comment](https://github.com/okTurtles/group-income/issues/2728#issuecomment-2760700835) too, the physical height of a message is checked against these thresholds when,

 1. the message is initially rendered
 3. below isPhone state in ChatMain.vue changes:

https://github.com/okTurtles/group-income/blob/a11b226e68e98b1a415ed61a31bb54f544a6686c/frontend/views/containers/chatroom/ChatMain.vue#L250-L252

I haven't used `resize` event here for a performance reason (Running the computation for all messages in the chat every time `resize` event is fired would be really bad. Especially if it's a large chatroom)

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -236,6 +298,47 @@ export default ({
...
+    }
+  },
+  mounted () {
+    if (
+      this.shouldCheckToTruncate &&
+      // NOTE: If the message has any image attached, defer this check until the <img /> DOMs are rendered.
+      //       (which is detected via 'image-attachments-render-complete' custom event in ChatAttachmentPreview.vue)
+      !this.hasImageAttachment
+    ) {
+      this.determineToEnableTruncationToggle()
```
taoeffect: The "Show more" toggle is determined once when the message is loaded. However, if the message is edited, `determineToEnableTruncationToggle` isn't calculated again.

So if the message is short initially, and then edited to be super long, `Show more` won't appear.

Likewise, if the message is long initially, and then made short, `Show more` won't disappear.

So maybe another watcher or something needs to be added for updating on message edit.

frontend/utils/constants.js
taoeffect: I feel like both of these values are slightly too long. Could you reduce them from `750` to `500`?

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
```diff
@@ -25,6 +25,7 @@
           :src='objectURLList[entryIndex]'
           :alt='entry.name'
           @click='openImageViewer(objectURLList[entryIndex])'
+          @load='onImageLoad(objectURLList[entryIndex])'
```
taoeffect: Above this line is:

```
template(v-if='isForDownload')
  .c-attachment-preview(
    v-for='(entry, entryIndex) in attachmentList'
    :key='entryIndex'
    class='is-download-item'
    tabindex='0'
  )
```

`o3-mini-high` says:

> Using the array index as the key works but may cause issues if the attachment list changes frequently (for example, when items are added or removed).
• If each attachment has a unique identifier (for instance a file ID or manifestCid), consider using that as the key.

I don't know if this needs to be changed, but if it's easy to grab the manifestCid from `entry` feel free to use that as `:key`.

taoeffect (changes requested):
Great work @SebinSong! I respect your editorial decision regarding how to render the "Show more" link. We can keep it the way you've designed for now.

Left some feedback!

frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
SebinSong: Good catch. updated.

frontend/views/containers/chatroom/MessageBase.vue
SebinSong: @taoeffect Have implemented this requirement by adding a watcher for `updatedDate` property of the message. By doing so, we can guarantee that the height calculation/comparison is run 'after' the updated message is already reflected(or rendered) to the DOM.

taoeffect:
@SebinSong this is fantastic work! I've tested it and it works great!

Now, since you are more skilled at Vue and CSS than I am, I will leave it up to you to decide whether or not this PR is ready to be merged. Below are two reviews of the latest changes from `Gemini Pro 2.5`, and `o3-mini-high`. If you see anything in their commentary that you feel is a good suggestion and want to implement it, go right ahead, otherwise let me know that the PR is ready to be merged and I'll merge it.

# `Gemini Pro 2.5` Review

<details>
Okay, I've reviewed the provided changes. Here's my feedback:

### General Observations

The changes introduce a message truncation feature for long messages, which seems useful. The implementation involves calculating message height, handling different message types (text/poll), accounting for image loading delays, and providing a "Show more/less" toggle. The necessary props (`updatedDate`) are passed down correctly. The use of `inject` for configuration and the new constants are appropriate.

### Specific Feedback

#### 1. Constants: Mobile Height Threshold

**File:** `frontend/utils/constants.js`

```diff
--- a/frontend/utils/constants.js
+++ b/frontend/utils/constants.js
@@ -12,8 +12,8 @@
  // TODO: fetch this value from a server API
  export const KILOBYTE = 1 << 10
  export const MEGABYTE = 1 << 20
 export const CHAT_ATTACHMENT_SIZE_LIMIT = 30 * MEGABYTE // in byte.
 export const IMAGE_ATTACHMENT_MAX_SIZE = 400 * KILOBYTE // 400KB
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 500 * 1.25 // in px
+export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 500 * 1.5 // The value of mobile is more tolerant considering smaller screen size.

```

**Issue:** The mobile threshold (`750px`) is significantly higher than the desktop threshold (`625px`). The comment mentions this makes mobile "more tolerant considering smaller screen size". However, this means a message needs to be *taller* on a typically *smaller* screen before it gets truncated. This seems counter-intuitive from a UX perspective, as it might result in very long messages taking up most of the mobile viewport before the "Show more" option appears.

**Suggestion:** Double-check the intended UX. If the goal is to truncate *sooner* on mobile to preserve screen real estate, the mobile threshold should likely be *lower* than the desktop one. For example:

```javascript
// frontend/utils/constants.js

// Example: Lower mobile threshold
export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_DESKTOP = 600 // in px (Example value)
export const CHAT_LONG_MESSAGE_HEIGHT_THRESHOLD_MOBILE = 450 // in px (Example value, lower than desktop)
```

Adjust the values based on the desired behaviour after re-evaluating the UX.

#### 2. CSS Variable Unused

**File:** `frontend/views/containers/chatroom/MessageBase.vue`

```scss
// frontend/views/containers/chatroom/MessageBase.vue

 .c-message {
+  --c-bg-color: var(--background_0); // Defined here...
   padding: 0.5rem 1rem;
   scroll-margin: 20px;

@@ -404,6 +531,7 @@
   }

   &:hover {
+    // --c-bg-color: var(--general_2); // But seemingly not used here...
     background-color: $general_2; // ... direct color is used instead.

     &:not(.pending, .failed) {

```

**Issue:** The CSS variable `--c-bg-color` is defined but doesn't seem to be used effectively. The `:hover` rule directly sets `background-color: $general_2;` instead of potentially overriding the variable.

**Suggestion:** Remove the `--c-bg-color` variable definition if it's not being used to simplify the CSS.

```diff
 // frontend/views/containers/chatroom/MessageBase.vue
 .c-message {
-  --c-bg-color: var(--background_0);
   padding: 0.5rem 1rem;
   scroll-margin: 20px;

```

#### 3. Attachment List Watcher Complexity and Potential Bug

**File:** `frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue`

```javascript
// frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue

   watch: {
     attachmentList (to, from) {
       if (from.length > to.length) {
         // NOTE: this will be caught when user tries to delete attachments
         const oldObjectURLMapping = {}
         if (from.length === this.objectURLList.length) {
           from.forEach((attachment, index) => {
             // Relies on manifestCid being present
             oldObjectURLMapping[attachment.downloadData.manifestCid] = this.objectURLList[index]
           })
           // Rebuilds list based on mapping
           this.objectURLList = to.map(attachment => oldObjectURLMapping[attachment.downloadData.manifestCid])
         } else {
           // NOTE: this should not be caught, but considered for the error handler
           // Fallback: Regenerate all URLs
           Promise.all(to.map(attachment => this.getAttachmentObjectURL(attachment))).then(urls => {
             this.objectURLList = urls
           })
         }
       }
     }
   }

```

**Issue:**
1.  **Complexity/Brittleness:** The logic to rebuild `objectURLList` when items are deleted relies heavily on mapping via `attachment.downloadData.manifestCid`. If an attachment being deleted (or remaining) lacks this property (e.g., during upload or if `downloadData` failed to populate), this logic will break or produce incorrect results. The fallback `else` block recalculates everything, suggesting the mapping approach isn't fully robust.
2.  **Potential Memory Leak:** This logic doesn't appear to call `URL.revokeObjectURL()` for the URLs corresponding to the deleted attachments. While `getAttachmentObjectURL` *might* handle revocation internally, it's not guaranteed by this code snippet. Relying solely on garbage collection for ObjectURLs is often insufficient.

**Suggestion:** Simplify the logic and ensure cleanup. Regenerating the list on change might be simpler and more robust, especially if `getAttachmentObjectURL` can efficiently retrieve cached URLs. Ensure ObjectURLs are revoked.

```javascript
// frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
import { shallowEqual } from '@common/common.js' // Or similar utility

export default {
  // ... (other parts)
  data () {
    return {
      objectURLList: [],
      loadedURLList: [],
      loadingBoxStyles: [],
      // Keep track of generated URLs for cleanup
      generatedObjectURLs: []
    }
  },
  // ... (computed, etc.)
  beforeDestroy () {
    // Clean up all generated ObjectURLs
    this.generatedObjectURLs.forEach(url => URL.revokeObjectURL(url))
    this.generatedObjectURLs = []
    if (this.shouldPreviewImages) {
      sbp('okTurtles.events/off', DELETE_ATTACHMENT, this.deleteAttachment)
    }
  },
  methods: {
    // ... (other methods)
    async updateObjectURLs () {
      if (!this.shouldPreviewImages) {
        this.objectURLList = []
        return
      }

      // Revoke previous URLs before generating new ones
      this.generatedObjectURLs.forEach(url => URL.revokeObjectURL(url))
      this.generatedObjectURLs = []
      this.loadedURLList = [] // Reset loaded list as well

      try {
        const urls = await Promise.all(
          this.attachmentList.map(attachment => this.getAttachmentObjectURL(attachment))
        )
        // Filter out potential nulls/undefined if getAttachmentObjectURL can fail gracefully
        this.objectURLList = urls.filter(url => !!url)
        // Store successfully generated URLs for future cleanup
        this.generatedObjectURLs = [...this.objectURLList]

        // Update loading box styles if needed (assuming it depends on attachmentList)
        if (this.isForDownload) {
          this.loadingBoxStyles = this.attachmentList.map(attachment => {
            return this.getStretchedDimension(attachment.dimension)
          })
        }
      } catch (error) {
        console.error('Failed to update object URLs:', error)
        this.objectURLList = []
        this.loadingBoxStyles = []
      }
    },
    // ... (rest of methods)
  },
  watch: {
    // Watch attachmentList with deep: false, compare references or a stable key
    attachmentList (newVal, oldVal) {
      // Optional: Check if lists are actually different if newVal/oldVal might be mutated
      if (!shallowEqual(newVal, oldVal)) { // Use a suitable comparison
        this.updateObjectURLs()
      }
    }
  },
  mounted () {
    // Initial generation
    this.updateObjectURLs()

    if (this.shouldPreviewImages) {
      sbp('okTurtles.events/on', DELETE_ATTACHMENT, this.deleteAttachment)
    }
  }
}

```
*Note: This simplified approach assumes `getAttachmentObjectURL` is reasonably fast (e.g., uses caching). If performance is critical and the mapping approach is necessary, it needs explicit handling for cases where `manifestCid` is missing.*

#### 4. Image Load Completion Event Reliability

**File:** `frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue`

```javascript
// frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue

     onImageLoad (url) {
       if (this.isForDownload) {
         this.loadedURLList = uniq([...this.loadedURLList, url])

         if (this.allImageAttachments.length === this.loadedURLList.length) {
           // Check if all image attachments are loaded in the DOM, notify the parent component.
           // (This can be enhanced to something like sbp('okTurtles.events/emit', IMAGE_ATTACHMENTS_RENDER_COMPLETE, messageHash) in the future,
           //  if this becomes useful in more places.)
           this.$nextTick(() => this.$emit('image-attachments-render-complete'))
         }
       }
     },

```

**Issue:** If `getAttachmentObjectURL` fails for an image, or the browser fails to load an image for other reasons, the `@load` event (`onImageLoad`) will not fire for that specific image. Consequently, `this.loadedURLList.length` might never reach `this.allImageAttachments.length`. This means the `image-attachments-render-complete` event would never be emitted. `MessageBase.vue` relies on this event to trigger its height check for messages with images, so the check might never run, potentially leaving a long message without the "Show more" toggle if it contains images that failed to load.

**Suggestion:** Make the completion signal more robust. Consider emitting the event even if some images fail, possibly after all attempts (successful or failed) have settled, or after a timeout.

**Option 1 (Emit when all settled):** Track pending loads.

```javascript
// frontend/views/containers/chatroom/file-attachment/ChatAttachmentPreview.vue
// (Conceptual - requires changes in updateObjectURLs/getAttachmentObjectURL potentially)

methods: {
  async updateObjectURLs() {
    // ... (revoke old URLs, reset lists)
    let pendingLoads = 0
    const loadPromises = this.attachmentList.map(async (attachment) => {
       if (this.fileType(attachment) === 'image') {
         pendingLoads++
         try {
           const url = await this.getAttachmentObjectURL(attachment)
           if (url) {
             // Add url to a temporary list to be assigned later
             return { success: true, url }
           } else {
             return { success: false } // Handle graceful failure from getAttachmentObjectURL
           }
         } catch (err) {
           return { success: false } // Handle error from getAttachmentObjectURL
         }
       } else {
         return null // Not an image
       }
    })

    const results = await Promise.all(loadPromises)
    this.objectURLList = results.filter(r => r?.success).map(r => r.url)
    this.generatedObjectURLs = [...this.objectURLList]
    // Now objectURLList is populated, images will start loading in the template

    // We still need to wait for @load events... this approach is complex.
    // Maybe tracking @load and @error events is better.
  }

  // --- Alternative: Track @load/@error ---
  data() {
    return {
      // ...
      imageLoadStatus: {} // Map: url -> 'pending' | 'loaded' | 'error'
    }
  },
  methods: {
    checkAllImagesSettled() {
      if (!this.isForDownload) return

      const imageURLs = this.objectURLList // URLs currently attempted
      const allSettled = imageURLs.every(url =>
          this.imageLoadStatus[url] === 'loaded' || this.imageLoadStatus[url] === 'error'
      )

      if (allSettled && imageURLs.length > 0) {
         // Check if we haven't emitted completion for this set of URLs yet
         // (Requires additional state to prevent multiple emits for the same list)
         this.$nextTick(() => this.$emit('image-attachments-render-complete'))
      }
    },
    onImageLoad(url) {
      this.imageLoadStatus[url] = 'loaded'
      this.checkAllImagesSettled()
    },
    onImageError(url) {
      this.imageLoadStatus[url] = 'error'
      this.checkAllImagesSettled()
    },
    async updateObjectURLs() {
       // ... (generate URLs as before) ...
       this.objectURLList = urls.filter(url => !!url)
       this.generatedObjectURLs = [...this.objectURLList]

       // Reset status for the new URLs
       this.imageLoadStatus = {}
       this.objectURLList.forEach(url => { this.imageLoadStatus[url] = 'pending' })

       // If no images, emit immediately? Or let checkAllImagesSettled handle it (length 0 case)
       if (this.objectURLList.length === 0 && this.allImageAttachments.length > 0) {
         // Edge case: all image URL generations failed
          this.$nextTick(() => this.$emit('image-attachments-render-complete'))
       } else {
          this.checkAllImagesSettled() // Check if somehow all settled immediately (e.g. 0 images)
       }
    }
  }
}
// Template requires adding @error handler:
// <img ... @load='onImageLoad(objectURLList[entryIndex])' @error='onImageError(objectURLList[entryIndex])'>

```

**Option 2 (Timeout in `MessageBase.vue`):** Add a fallback timer in `MessageBase.vue` that runs the height check if the event hasn't arrived within a reasonable time (e.g., 5 seconds after mount/update). This is simpler but less precise.

Choose the approach that best fits the project's needs for robustness vs complexity. The `@load`/`@error` tracking seems most accurate.
</details>

# `o3-mini-high` Review

<details>
Below are some specific issues and improvement suggestions I discovered while reviewing the changes. Let me know if you’d like additional context or clarifications on any of these points.

---

## 1. Issue: Undefined “fileType” Method in ChatAttachmentPreview.vue

In your computed properties you’re using “this.fileType” even though no such method or property is defined. For example, you have:

```js
computed: {
  shouldPreviewImages () {
    return !this.attachmentList.some(attachment => {
      return this.fileType(attachment) === 'non-image'
    })
  },
  allImageAttachments () {
    return this.attachmentList.filter(entry => this.fileType(entry) === 'image')
  },
  …
}
```

Since you already import the helper `getFileType` from '@view-utils/filters.js' (which is used in other parts of the code) you can update these computed properties to use that helper. For example, if the file’s MIME type is stored under `attachment.mimeType` you might do:

```js
computed: {
  shouldPreviewImages () {
    return !this.attachmentList.some(attachment => {
      return getFileType(attachment.mimeType) === 'non-image'
    })
  },
  allImageAttachments () {
    return this.attachmentList.filter(entry => getFileType(entry.mimeType) === 'image')
  },
  …
}
```

This change ensures that the helper is used consistently and avoids referencing an undefined function.

---

## 2. Issue: Use of replaceAll in getAttachmentId (ChatAttachmentPreview.vue)

In your new helper method you use the string method `replaceAll`:

```js
getAttachmentId (attachment) {
  return attachment.downloadData?.manifestCid || attachment.name.replaceAll(' ', '_')
}
```

While `replaceAll` is available in modern environments, it might cause issues if the project needs to support older browsers. To be safe (and to have similar behavior across browsers) you could use a global regular expression with `replace`:

```js
getAttachmentId (attachment) {
  return attachment.downloadData?.manifestCid || attachment.name.replace(/ /g, '_')
}
```

This yields the same result while having wider compatibility.

---

## 3. Issue: Nonstandard scrollIntoView Behavior Option

In the `toggleMessageTruncation` method (in MessageBase.vue) you call:

```js
this.$el.scrollIntoView({ behavior: 'instant' })
```

The option value `'instant'` isn’t specified in the standard (the two supported values are usually `'auto'` and `'smooth'`). If you need the scroll to happen immediately you could either simply omit the option or pass `{ behavior: 'auto' }`. For example:

```js
toggleMessageTruncation (e) {
  const currentValue = this.ephemeral.truncateToggle.isShowingAll
  this.ephemeral.truncateToggle.isShowingAll = !currentValue

  if (currentValue) {
    // When folding the expanded message, scroll to the message instantly.
    this.$el.scrollIntoView({ behavior: 'auto' })
  }
}
```

This ensures compatibility with browsers that follow the standard.

---

## 4. Suggestion: Update Watcher for attachmentList on Additions

In ChatAttachmentPreview.vue you watch the attachment list and only handle the case where attachments are removed:

```js
watch: {
  attachmentList (to, from) {
    if (from.length > to.length) {
      // deletion logic...
    }
  }
}
```

If attachments can also be added dynamically, you might want to recalculate the object URL list when the list grows. One way to handle _any_ change is to always re-map (or handle additions and deletions separately). For instance:

```js
watch: {
  attachmentList (to, from) {
    // Always update the URL list if the attachments array length changes
    if (to.length !== from.length) {
      Promise.all(to.map(attachment => this.getAttachmentObjectURL(attachment)))
        .then(urls => {
          this.objectURLList = urls
        })
    }
  }
}
```

Even if this approach is less optimized than trying to patch only on deletion, it may be acceptable if the attachment lists are short or if dynamic updates are infrequent.

---

## 5. Suggestion: Provide a Default for Injected chatMainConfig

In MessageBase.vue you inject `chatMainConfig` like this:

```js
inject: ['chatMainConfig'],
```

If there’s any risk that `chatMainConfig` might be undefined (for example, when MessageBase is used outside of the context that provides it), you can supply a default value. For example:

```js
inject: {
  chatMainConfig: {
    default: () => ({ isPhone: false })
  }
},
```

This makes your code more robust in case the injection is missing.
</details>

SebinSong:
@taoeffect The PR is ready for review again. I did make a few suggested updates by those LLMs. (Especially,`Gemini Pro 2.5` sounds pretty accurate and smart). But `1.` and `3.` of `o3-mini-high` feedbacks are actually incorrect informations. (Thought you would want to know this too)

taoeffect (approved):
Fantastic job @SebinSong! 🎉

PR #2742, opened 2025-04-07
Implement conflict resolution

### LINKED ISSUE #2739: Eliminate missing data due to poor KV conflict resolution
### Problem

If I'm not mistaken, it is still possible for two simultaneous KV updates to happen and for both of them to "go through" but only one of them to be applied, leading to missing data, as conflict check appears to only check that the hight is current:

```js
  try {
    const serializedData = JSON.parse(request.payload.toString())
    const { contracts } = sbp('chelonia/rootState')
    // Check that the height is the latest value. Not only should the height be
    // the latest, but also enforcing this lets us check that signatures are
    // using the latest (cryptograhpic) keys. Since the KV is detached from the
    // contract, in isolation it's impossible to know if an old signature is
    // just because it was created in the past, or if it's because someone
    // is reusing a previously good key that has since been revoked.
    if (contracts[contractID].height !== Number(serializedData.height)) {
      return Boom.conflict()
    }
```

There is also this that might protect against it:

```js
  // Some protection against accidental overwriting by implementing the if-match
  // header
  // If-Match contains a list of ETags or '*'
  // If `If-Match` contains a known ETag, allow the request through, otherwise
  // return 412 Precondition Failed.
  // This is useful to clients to avoid accidentally overwriting existing data
  // For example, client A and client B want to write to key 'K', which contains
  // an array. Let's say that the array is originally empty (`[]`) and A and B
  // want to append `A` and `B` to it, respectively. If both write at the same
  // time, the following could happen:
  // t = 0: A reads `K`, gets `[]`
  // t = 1: B reads `K`, gets `[]`
  // t = 2: A writes `['A']` to `K`
  // t = 3: B writes `['B']` to `K` <-- ERROR: B should have written `['A', 'B']`
  // To avoid this situation, A and B could use `If-Match`, which would have
  // given B a 412 response
  if (request.headers['if-match']) {
    if (!existing) {
      return Boom.preconditionFailed()
    }
    const expectedEtag = request.headers['if-match']
    if (expectedEtag === '*') {
      // pass through
    } else {
      // "Quote" string (to match ETag format)
      const cid = JSON.stringify(createCID(existing, multicodes.RAW))
      if (!expectedEtag.split(',').map(v => v.trim()).includes(cid)) {
        return Boom.preconditionFailed()
      }
    }
  }
```

However this returns 412, and that doesn't seem checked for on the frontend:

```js
      if (!response.ok) {
        if (response.status === 409) {
          if (--maxAttempts <= 0) {
            throw new Error('kv/set conflict setting KV value')
          }
          await delay(randomIntFromRange(0, 1500))
          if (typeof onconflict === 'function') {
            data = await onconflict(contractID, key, data)
            if (!data) { break }
          }
          continue
        }
        throw new Error('kv/set invalid response status: ' + response.status)
      }
```

### Solution

Either fix it with some of the existing code, or using a counter.

Here's how the "counter based approach" works:

Instead of storing key/values in the KV db like this:

```
<key> => <value>
```

We'd store them using a tuple like this:

```
<key> => (<counter>, <value>)
```

When the client retrieves the KV pair, it is given the current `<counter>` value too. It then increments it by `<increment>` when writing the updated value back. The server only accepts the write if `<currentCounter> = <previousCounter> + <increment>`.

This method is simple and doesn't require `If-Match` or even the `height` values, and guaranteed to ensure that each write either succeeds because it's building on the previous value or fails and then tries again.

As JS uses doubles to store values, we want the smallest value for `<increment>` that will produce a valid value for `<counter> + <increment>`, and then use a safe range for this, possibly with a modulus `%` to address overflow / wraparound.

EDIT: Actually, if we use modulus arithmetic I think it's safe to pick some smaller range and just use `1` for the `<increment>`. I.e. a range of like `0 to 100000` is perfectly representable by doubles and is more than sufficient to handle simultaneous writes from thousands of clients.

corrideat:
Closes #2739

Currently a draft as tests seem to be failing.

### DIFF
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -66,15 +66,15 @@ export default (sbp('sbp/selectors/register', {
 
     const nowString = new Date(now).toISOString()
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
+      const constructUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
         return [{ ...currentData, [identityContractID]: nowString }, etag]
       }
 
-      const data = getUpdatedLastLoggedIn()?.[0]
+      const data = constructUpdatedLastLoggedIn()[0]
       await sbp('chelonia/kv/set', contractID, KV_KEYS.LAST_LOGGED_IN, data, {
         encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
         signingKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk'),
-        onconflict: getUpdatedLastLoggedIn
+        onconflict: constructUpdatedLastLoggedIn
       })
     })
   }
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -66,12 +66,11 @@ export default (sbp('sbp/selectors/register', {
 
     const nowString = new Date(now).toISOString()
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedLastLoggedIn = async (contractID) => {
-        const current = await sbp('gi.actions/group/kv/fetchLastLoggedIn', { contractID })
-        return { ...current, [identityContractID]: nowString }
+      const getUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
+        return [{ ...currentData, [identityContractID]: nowString }, etag]
       }
 
-      const data = await getUpdatedLastLoggedIn(contractID)
+      const data = getUpdatedLastLoggedIn()?.[0]
       await sbp('chelonia/kv/set', contractID, KV_KEYS.LAST_LOGGED_IN, data, {
         encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
         signingKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk'),
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -98,10 +98,8 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = getUpdatedUnreadMessages()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      const data = {}
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/addChatRoomUnreadMessage': ({ contractID, messageHash, createdHeight }: {
@@ -119,10 +117,8 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = getUpdatedUnreadMessages()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      const data = {}
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/removeChatRoomUnreadMessage': ({ contractID, messageHash }: {
@@ -139,10 +135,8 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = getUpdatedUnreadMessages()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      const data = {}
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/deleteChatRoomUnreadMessages': ({ contractID }: { contractID: string }) => {
@@ -226,17 +220,18 @@ export default (sbp('sbp/selectors/register', {
     }
 
     const updatedOnConflict = async (...args) => {
-      if (typeof onconflict === 'function') {
-        return applyStorageRules(await onconflict(...args))
-      }
-      return null
+      const result = await onconflict(...args)
+      if (!result) return null
+
+      const [data, etag] = result
+      return [applyStorageRules(data), etag]
     }
 
     return sbp('chelonia/kv/queuedSet', {
       contractID: identityContractID,
       key: KV_KEYS.NOTIFICATIONS,
       data: applyStorageRules(data),
-      onconflict: updatedOnConflict
+      onconflict: onconflict ? updatedOnConflict : null
     })
   },
   'gi.actions/identity/kv/loadNotificationStatus': () => {
@@ -291,10 +286,8 @@ export default (sbp('sbp/selectors/register', {
         return isUpdated ? [currentData, etag] : null
       }
 
-      const data = getUpdatedNotificationStatus()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
-      }
+      const data = {}
+      await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
     })
   }
 }): string[])
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -75,9 +75,7 @@ export default (sbp('sbp/selectors/register', {
       }
 
       const data = getUpdatedUnreadMessages()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight }: {
@@ -98,15 +96,14 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = {}
-      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/addChatRoomUnreadMessage': ({ contractID, messageHash, createdHeight }: {
     contractID: string, messageHash: string, createdHeight: number
   }) => {
-    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async ({ currentData = {}, etag } = {}) => {
-      const getUpdatedUnreadMessages = () => {
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
+      const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {
         if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
           const index = currentData[contractID].unreadMessages.findIndex(msg => msg.messageHash === messageHash)
           if (index === -1) {
@@ -117,8 +114,7 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = {}
-      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/removeChatRoomUnreadMessage': ({ contractID, messageHash }: {
@@ -135,8 +131,7 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = {}
-      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { onconflict: getUpdatedUnreadMessages })
     })
   },
   'gi.actions/identity/kv/deleteChatRoomUnreadMessages': ({ contractID }: { contractID: string }) => {
@@ -149,8 +144,7 @@ export default (sbp('sbp/selectors/register', {
         return null
       }
 
-      const data = {}
-      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { onconflict: getUpdatedUnreadMessages })
     })
   },
   // Preferences
@@ -191,9 +185,7 @@ export default (sbp('sbp/selectors/register', {
       }
 
       const data = getUpdatedPreferences()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/savePreferences', { data, onconflict: getUpdatedPreferences })
-      }
+      await sbp('gi.actions/identity/kv/savePreferences', { data, onconflict: getUpdatedPreferences })
     })
   },
   // Notifications
@@ -254,9 +246,7 @@ export default (sbp('sbp/selectors/register', {
       }
 
       const data = getUpdatedNotificationStatus()?.[0]
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
-      }
+      await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
     })
   },
   'gi.actions/identity/kv/markNotificationStatusRead': (hashes: string | string[]) => {
@@ -286,8 +276,7 @@ export default (sbp('sbp/selectors/register', {
         return isUpdated ? [currentData, etag] : null
       }
 
-      const data = {}
-      await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
+      await sbp('gi.actions/identity/kv/saveNotificationStatus', { onconflict: getUpdatedNotificationStatus })
     })
   }
 }): string[])
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -218,7 +218,7 @@ export default (sbp('sbp/selectors/register', {
         return [{ ...currentPreferences, hideDistributionBanner }, etag]
       }
 
-      const data = getUpdatedPreferences()?.[0]
+      const data = getUpdatedPreferences()[0]
       await sbp('gi.actions/identity/kv/savePreferences', { data, onconflict: getUpdatedPreferences })
     })
   },
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -212,7 +212,7 @@ export default (sbp('sbp/selectors/register', {
     }
 
     const updatedOnConflict = async (...args) => {
-      const result = await onconflict(...args)
+      const result = await (onconflict: Function)(...args)
       if (!result) return null
 
       const [data, etag] = result
@@ -223,7 +223,7 @@ export default (sbp('sbp/selectors/register', {
       contractID: identityContractID,
       key: KV_KEYS.NOTIFICATIONS,
       data: applyStorageRules(data),
-      onconflict: onconflict ? updatedOnConflict : null
+      onconflict: typeof onconflict === 'function' ? updatedOnConflict : null
     })
   },
   'gi.actions/identity/kv/loadNotificationStatus': () => {
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -61,22 +61,20 @@ export default (sbp('sbp/selectors/register', {
     contractID: string, messageHash: string, createdHeight: number
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedUnreadMessages = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
-
+      const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {
         if (!currentData[contractID]) {
-          return {
+          return [{
             ...currentData,
             [contractID]: {
               readUntil: { messageHash, createdHeight },
               unreadMessages: []
             }
-          }
+          }, etag]
         }
         return null
       }
 
-      const data = await getUpdatedUnreadMessages()
+      const data = getUpdatedUnreadMessages()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
       }
@@ -86,23 +84,21 @@ export default (sbp('sbp/selectors/register', {
     contractID: string, messageHash: string, createdHeight: number
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedUnreadMessages = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
-
+      const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {
         if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
           const { unreadMessages } = currentData[contractID]
-          return {
+          return [{
             ...currentData,
             [contractID]: {
               readUntil: { messageHash, createdHeight },
               unreadMessages: unreadMessages.filter(msg => msg.createdHeight > createdHeight)
             }
-          }
+          }, etag]
         }
         return null
       }
 
-      const data = await getUpdatedUnreadMessages()
+      const data = getUpdatedUnreadMessages()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
       }
@@ -111,21 +107,19 @@ export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/kv/addChatRoomUnreadMessage': ({ contractID, messageHash, createdHeight }: {
     contractID: string, messageHash: string, createdHeight: number
   }) => {
-    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedUnreadMessages = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
-
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async ({ currentData = {}, etag } = {}) => {
+      const getUpdatedUnreadMessages = () => {
         if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
           const index = currentData[contractID].unreadMessages.findIndex(msg => msg.messageHash === messageHash)
           if (index === -1) {
             currentData[contractID].unreadMessages.push({ messageHash, createdHeight })
-            return currentData
+            return [currentData, etag]
           }
         }
         return null
       }
 
-      const data = await getUpdatedUnreadMessages()
+      const data = getUpdatedUnreadMessages()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
       }
@@ -135,19 +129,17 @@ export default (sbp('sbp/selectors/register', {
     contractID: string, messageHash: string
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedUnreadMessages = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
-
+      const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {
         const index = currentData[contractID]?.unreadMessages.findIndex(msg => msg.messageHash === messageHash)
         // NOTE: index could be undefined if unreadMessages is not initialized
         if (Number.isInteger(index) && index >= 0) {
           currentData[contractID].unreadMessages.splice(index, 1)
-          return currentData
+          return [currentData, etag]
         }
         return null
       }
 
-      const data = await getUpdatedUnreadMessages()
+      const data = getUpdatedUnreadMessages()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
       }
@@ -155,19 +147,16 @@ export default (sbp('sbp/selectors/register', {
   },
   'gi.actions/identity/kv/deleteChatRoomUnreadMessages': ({ contractID }: { contractID: string }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedUnreadMessages = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
+      const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {
         if (currentData[contractID]) {
           delete currentData[contractID]
-          return currentData
+          return [currentData, etag]
         }
         return null
       }
 
-      const data = await getUpdatedUnreadMessages()
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      const data = {}
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
     })
   },
   // Preferences
@@ -199,17 +188,18 @@ export default (sbp('sbp/selectors/register', {
   },
   'gi.actions/identity/kv/updateDistributionBannerVisibility': ({ contractID, hidden }: { contractID: string, hidden: boolean }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedPreferences = async () => {
-        const currentPreferences = await sbp('gi.actions/identity/kv/fetchPreferences')
+      const getUpdatedPreferences = ({ etag, currentData: currentPreferences = {} } = {}) => {
         const hideDistributionBanner = {
           ...(currentPreferences.hideDistributionBanner || {}),
           [contractID]: hidden
         }
-        return { ...currentPreferences, hideDistributionBanner }
+        return [{ ...currentPreferences, hideDistributionBanner }, etag]
       }
 
-      const data = await getUpdatedPreferences()
-      await sbp('gi.actions/identity/kv/savePreferences', { data, onconflict: getUpdatedPreferences })
+      const data = getUpdatedPreferences()?.[0]
+      if (data) {
+        await sbp('gi.actions/identity/kv/savePreferences', { data, onconflict: getUpdatedPreferences })
+      }
     })
   },
   // Notifications
@@ -258,18 +248,17 @@ export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/kv/addNotificationStatus': (notification: Object) => {
     const { hash, timestamp } = notification
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedNotificationStatus = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchNotificationStatus')
-        if (!currentData[hash]) {
-          return {
+      const getUpdatedNotificationStatus = ({ currentData = {}, etag } = {}) => {
+        if (!currentData?.[hash]) {
+          return [{
             ...currentData,
             [hash]: initNotificationStatus({ timestamp })
-          }
+          }, etag]
         }
         return null
       }
 
-      const data = await getUpdatedNotificationStatus()
+      const data = getUpdatedNotificationStatus()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
       }
@@ -281,8 +270,7 @@ export default (sbp('sbp/selectors/register', {
     }
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const notifications = sbp('chelonia/rootState').notifications.items
-      const getUpdatedNotificationStatus = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchNotificationStatus')
+      const getUpdatedNotificationStatus = ({ currentData = {}, etag } = {}) => {
         let isUpdated = false
         for (const hash of hashes) {
           const existing = notifications.find(n => n.hash === hash)
@@ -300,10 +288,10 @@ export default (sbp('sbp/selectors/register', {
             isUpdated = true
           }
         }
-        return isUpdated ? currentData : null
+        return isUpdated ? [currentData, etag] : null
       }
 
-      const data = await getUpdatedNotificationStatus()
+      const data = getUpdatedNotificationStatus()?.[0]
       if (data) {
         await sbp('gi.actions/identity/kv/saveNotificationStatus', { data, onconflict: getUpdatedNotificationStatus })
       }
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -222,7 +222,7 @@ export default (sbp('sbp/selectors/register', {
     return sbp('chelonia/kv/queuedSet', {
       contractID: identityContractID,
       key: KV_KEYS.NOTIFICATIONS,
-      data: applyStorageRules(data),
+      data: !!data && applyStorageRules(data),
       onconflict: typeof onconflict === 'function' ? updatedOnConflict : null
     })
   },
--- a/shared/domains/chelonia/chelonia-utils.js
+++ b/shared/domains/chelonia/chelonia-utils.js
@@ -8,9 +8,10 @@ export default (sbp('sbp/selectors/register', {
   // `chelonia/kv/set`. However, the `chelonia/kv/set` primitive is needed if
   // the queueing logic needs to be more advanced, the key to use requires
   // custom logic or _if the `onconflict` callback also needs to be queued_.
-  'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+  'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, ifMatch, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
     return sbp('chelonia/queueInvocation', contractID, () => {
       return sbp('chelonia/kv/set', contractID, key, data, {
+        ifMatch,
         encryptionKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, encryptionKeyName),
         signingKeyId: sbp('chelonia/contract/currentKeyIdByName', contractID, signingKeyName),
         onconflict
--- a/shared/domains/chelonia/chelonia-utils.js
+++ b/shared/domains/chelonia/chelonia-utils.js
@@ -1,5 +1,15 @@
 import sbp from '@sbp/sbp'
 
+// This file contains non-core parts of Chelonia, i.e., functionality that is
+// useful but optional. The threshold for something being 'optional' generally
+// is something that can be implemented externally using only public Chelonia
+// selectors.
+// Optional functionality can make certain assumptions about contracts or
+// actions to make things simpler or easier to implement.
+// Currently, a single selector is defined: 'chelonia/kv/queuedSet'.
+// TODO: Other things should be moved to this file, such as `encryptedAction`
+// (the wrapper) and 'gi.actions/out/rotateKeys'.
+
 export default (sbp('sbp/selectors/register', {
   // This selector is a wrapper for the `chelonia/kv/set` selector that uses
   // the contract queue and allows referring to keys by name, with default key
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1619,6 +1619,8 @@ export default (sbp('sbp/selectors/register', {
   // a general rule, you shouldn't be calling this selector directly unless
   // you're building a utility library or if you have very specific needs. In
   // this case, see if `chelonia/kv/queuedSet` covers your needs.
+  // `data` is allowed to be falsy, in which case a fetch will occur first and
+  // the `onconflict` handler will be called.
   'chelonia/kv/set': async function (contractID: string, key: string, data: Object, {
     ifMatch,
     innerSigningKeyId,
@@ -1632,7 +1634,7 @@ export default (sbp('sbp/selectors/register', {
     encryptionKeyId: ?string,
     signingKeyId: string,
     maxAttempts: ?number,
-    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }) => Promise<Object>,
+    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }) => Promise<[Object, string]>,
   }) {
     maxAttempts = maxAttempts ?? 3
     const url = `${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`
@@ -1707,6 +1709,7 @@ export default (sbp('sbp/selectors/register', {
           if (--maxAttempts <= 0) {
             throw new Error('kv/set conflict setting KV value')
           }
+          // Only retry if an onconflict handler exists to potentially resolve it
           await delay(randomIntFromRange(0, 1500))
           if (typeof onconflict === 'function') {
             if (await resolveData()) {
@@ -1714,8 +1717,10 @@ export default (sbp('sbp/selectors/register', {
             } else {
               break
             }
+          } else {
+            // Can't resolve automatically if there's no conflict handler
+            throw new Error(`kv/set failed with status ${response.status} and no onconflict handler was provided`)
           }
-          continue
         }
         throw new ChelErrorUnexpectedHttpResponseCode('kv/set invalid response status: ' + response.status)
       }
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -9,7 +9,7 @@ import { NOTIFICATION_TYPE, createClient } from '~/shared/pubsub.js'
 import type { SPKey, SPOpActionUnencrypted, SPOpContract, SPOpKeyAdd, SPOpKeyDel, SPOpKeyRequest, SPOpKeyRequestSeen, SPOpKeyShare, SPOpKeyUpdate } from './SPMessage.js'
 import type { Key } from '@chelonia/crypto'
 import { EDWARDS25519SHA512BATCH, deserializeKey, keyId, keygen, serializeKey } from '@chelonia/crypto'
-import { ChelErrorResourceGone, ChelErrorUnexpected, ChelErrorUnrecoverable } from './errors.js'
+import { ChelErrorResourceGone, ChelErrorUnexpected, ChelErrorUnexpectedHttpResponseCode, ChelErrorUnrecoverable } from './errors.js'
 import { CHELONIA_RESET, CONTRACTS_MODIFIED, CONTRACT_REGISTERED } from './events.js'
 // TODO: rename this to ChelMessage
 import { SPMessage } from './SPMessage.js'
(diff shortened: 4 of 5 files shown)


taoeffect:
# `o3-mini-high`

<details>
Below are some concrete issues and suggestions for improvement. (Let me know if you’d like more details or additional context.) One major change (and difference in behavior) is in the way the POST route’s “if‐match” header is handled. There’s also a slight mismatch between the (now tuple‐based) onconflict callbacks and their Flow type signatures. I’ve included specific code excerpts from your diffs along with improved versions.

---

## 1. If‐Match Header Handling in the POST /kv Route

In your updated version (backend/routes.js) you now require the header and “quote” the ETag value. However, note that in the original code you rejected requests if no KV existed before when an if‐match header was provided. In your new code the “if-match === '*'” branch does not check that an existing value is present. That may allow a client to create a new KV entry (when none exists) simply by sending “*” even though the original intent was to protect against accidental overwriting.

**Problematic lines:**

```js
const expectedEtag = request.headers['if-match']
if (!expectedEtag) {
  return Boom.badRequest('if-match is required')
}
// "Quote" string (to match ETag format)
const cid = existing ? createCID(existing, multicodes.RAW) : ''

if (expectedEtag === '*') {
  // pass through
} else {
  if (!expectedEtag.split(',').map(v => v.trim()).includes(`"${cid}"`)) {
    // We need this `x-cid` because HAPI modifies Etags
    return h.response(existing).etag(cid).header('x-cid', `"${cid}"`).code(412)
  }
}
```

**Suggestion:**  
Add an explicit check so that if a client sends if‐match "*" but no resource exists, your code returns a precondition error. For example:

```js
const expectedEtag = request.headers['if-match']
if (!expectedEtag) {
  return Boom.badRequest('if-match header is required')
}
if (expectedEtag === '*' && !existing) {
  return Boom.preconditionFailed()
}
const cid = existing ? createCID(existing, multicodes.RAW) : ''

if (expectedEtag !== '*') {
  const etags = expectedEtag.split(',').map(v => v.trim())
  if (!etags.includes(`"${cid}"`)) {
    // We need this "x-cid" because HAPI modifies Etags
    return h.response(existing).etag(cid).header('x-cid', `"${cid}"`).code(412)
  }
}
```

This way, you enforce that “*” is only allowed when there is an existing value—preserving the original intent.

---

## 2. Onconflict Callback Type Versus Return Value

In several places (including changes in chelonia.js and the identity–kv actions) you have refactored your onconflict callbacks so that they now return a tuple like `[updatedData, newIfMatch]` instead of a plain object. In chelonia.js the Flow type signature has not been updated to reflect this new contract.

**Problematic line (in chelonia.js):**

```diff
-    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }) => Promise<Object>,
+    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }) => Promise<Object>,
```

**Suggestion:**  
Update the Flow type signature to indicate that onconflict now returns a tuple (or, if you prefer, make it clear in documentation that the resolved value is in the form of [data, ifMatch]). For example, you might change it to something like:

```js
'onconflict: (args: {
  contractID: string,
  key: string,
  failedData: Object,
  status: number,
  etag: ?string,
  currentData: Object
}) => Promise<?[Object, string]>',
```

This makes the intended tuple return (or a nullish value when no update is needed) explicit. Also be sure to update any documentation/comments that explain the expected signature of these callbacks.

---

## 3. Optional Updates in the Identity–KV Callbacks

In some of your identity–kv actions you removed the “if (data)” check and always call the save function with:

```diff
-      const data = await getUpdatedUnreadMessages()
-      if (data) {
-        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
-      }
+      const data = getUpdatedUnreadMessages()?.[0]
+      await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
```

**Observation & Suggestion:**  
If the original intent was to perform a save only when there’s an update (i.e. when getUpdatedUnreadMessages returns a truthy value), you might want to keep that conditional. For example:

```js
const updateResult = getUpdatedUnreadMessages()
if (updateResult) {
  const data = updateResult[0]
  await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
}
```

That way you avoid triggering a save with an undefined value if there isn’t any update to make. (If your updated save functions are designed to handle an undefined data value, then this change may be optional—but it’s worth confirming the intended behavior.)
</details>

# `Gemini 2.5 Pro`

<details>
Okay, I've reviewed the provided diff and context. Here are my findings:

### Review Summary

The changes introduce optimistic concurrency control for the KV store using the `If-Match` header and ETag (based on the content's CID). This is implemented on both the backend (`routes.js`) and the frontend (`chelonia.js`, `identity-kv.js`, `group-kv.js`). The frontend `chelonia/kv/set` function is refactored to handle conflicts by fetching the current value and invoking a new `onconflict` handler, which receives more context (status, etag, current value) and returns the desired new value along with the ETag to use for the retry. The frontend action files (`identity-kv.js`, `group-kv.js`) are updated to use this new `onconflict` pattern.

Overall, the changes seem well-intentioned and implement a robust way to handle concurrent updates to KV store entries. However, there are a few issues, primarily related to type definitions and error handling in edge cases.

### Issues and Suggestions

**1. `shared/domains/chelonia/chelonia.js`: Incorrect FlowType for `onconflict`**

The FlowType definition for the `onconflict` parameter in `chelonia/kv/set` does not match its actual usage within the function or how it's called from `identity-kv.js`. The implementation passes an object with `status`, `etag`, `failedData`, a getter `currentData`, and `currentValue`, and expects a Promise resolving to `[newData, newEtag]` or `null`.

```diff
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1612,7 +1612,11 @@
     signingKeyId: string,
     maxAttempts: ?number,
     // NOTE: The return type is `[newData, newEtag]` or `null` if the conflict resolution decides not to proceed.
-    onconflict: (contractID: string, key: string, data: Object) => Promise<Object>,
+    // `currentData` is a convenience getter for `currentValue?.data`. `currentValue` is the full parsed result from the server.
+    // `failedData` is the data that failed to be written due to the conflict.
+    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, get currentData (): ?Object, currentValue: ?Object }) => Promise<[Object, ?string] | null>,
   }) {
     maxAttempts = maxAttempts ?? 3
     const url = `${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`

```

**2. `shared/domains/chelonia/chelonia.js`: Missing Error Handling for 409/412 Without `onconflict` Handler**

When a POST request in `chelonia/kv/set` receives a 409 (Conflict) or 412 (Precondition Failed) status, the code checks if an `onconflict` handler exists. If it *doesn't* exist, the code currently falls through and will eventually throw a generic `ChelErrorUnexpectedHttpResponseCode`. This is not ideal. It should likely throw a more specific error indicating a conflict or precondition failure occurred that couldn't be automatically resolved.

```diff
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1685,12 +1685,18 @@
           if (--maxAttempts <= 0) {
             throw new Error('kv/set conflict setting KV value')
           }
+          // Only retry if an onconflict handler exists to potentially resolve it
           await delay(randomIntFromRange(0, 1500))
           if (typeof onconflict === 'function') {
             if (await resolveData()) {
               continue
             } else {
               break
+            }
+          } else {
+            // If there's no conflict handler, we cannot resolve this automatically. Throw a specific error.
+            // TODO: Consider introducing specific ChelErrorConflict / ChelErrorPreconditionFailed errors
+            throw new Error(`kv/set failed with status ${response.status} and no onconflict handler was provided.`)
           }
           continue
         }

```
*Self-correction:* Added a specific error throw in the `else` block. Consider defining custom error classes like `ChelErrorConflict` or `ChelErrorPreconditionFailed` for better error handling downstream.

**3. `shared/domains/chelonia/chelonia.js`: Implicit GET Request Clarity**

When `chelonia/kv/set` is called without initial `data` but with an `onconflict` handler, it performs an initial `fetch` to get the current state. This `fetch` relies on the default HTTP method being GET. While correct, explicitly stating the method improves readability and reduces potential confusion.

```diff
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1644,6 +1644,7 @@
       } else {
         response = await fetch(url, {
           headers: new Headers([[
+            // GET request doesn't need If-Match, but needs auth
             'authorization', buildShelterAuthorizationHeader.call(this, contractID)
           ]]),
           signal: this.abortController.signal

```
*Self-correction:* Added `method: 'GET'` for clarity. Removed the previous comment as the Authorization header *is* needed for GET too.

**4. `frontend/controller/actions/group-kv.js`: Superfluous Optional Chaining**

The `getUpdatedLastLoggedIn` conflict handler always returns an array `[data, etag]`. Therefore, the optional chaining `?.[0]` used when calling it initially is unnecessary.

```diff
--- a/frontend/controller/actions/group-kv.js
+++ b/frontend/controller/actions/group-kv.js
@@ -62,7 +62,7 @@
         return [{ ...currentData, [identityContractID]: nowString }, etag]
       }
 
-      const data = getUpdatedLastLoggedIn()?.[0]
+      const data = getUpdatedLastLoggedIn()[0] // Handler always returns an array
       await sbp('chelonia/kv/set', contractID, KV_KEYS.LAST_LOGGED_IN, data, {
         encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
         signingKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk'),

```

**5. `backend/routes.js`: Mandatory `If-Match` Header Consideration**

The change makes the `If-Match` header mandatory for all POST requests to `/kv/...`.

```javascript
// backend/routes.js Line 751-753
  const expectedEtag = request.headers['if-match']
  if (!expectedEtag) {
    return Boom.badRequest('if-match is required')
  }
```

While this enforces optimistic locking for updates, it also requires clients to explicitly send `If-Match: ""` (or `If-Match: "*"`) even when they intend to *create* a new key-value pair (where no ETag exists yet). This might be slightly less convenient than allowing creation requests without `If-Match`. This is a design choice, but worth noting. The current implementation *does* handle `If-Match: ""` correctly for creation by comparing it against `""` when `existing` is null. If this strictness is intended, it's acceptable, otherwise, you might consider only requiring `If-Match` if the resource already exists.

No code change suggested here, just noting the design implication.

</details>

shared/domains/chelonia/chelonia.js
```diff
@@ -1643,50 +1643,110 @@ export default (sbp('sbp/selectors/register', {
...
     maxAttempts,
     onconflict
   }: {
+    ifMatch?: string,
     innerSigningKeyId: ?string,
     encryptionKeyId: ?string,
     signingKeyId: string,
     maxAttempts: ?number,
-    onconflict: (contractID: string, key: string, data: Object) => Promise<Object>,
+    onconflict: (args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }) => Promise<[Object, string]>,
```
taoeffect: This seems long and complicated enough to be a named type

frontend/controller/actions/group-kv.js
```diff
@@ -66,12 +66,11 @@ export default (sbp('sbp/selectors/register', {
...
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
-      const getUpdatedLastLoggedIn = async (contractID) => {
-        const current = await sbp('gi.actions/group/kv/fetchLastLoggedIn', { contractID })
-        return { ...current, [identityContractID]: nowString }
+      const getUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
+        return [{ ...currentData, [identityContractID]: nowString }, etag]
       }
 
-      const data = await getUpdatedLastLoggedIn(contractID)
+      const data = getUpdatedLastLoggedIn()?.[0]
```
taoeffect: A few things aren't making sense to me here:

1. The name of the function is `getUpdatedLastLoggedIn`, but it's no longer doing that here as `fetchLastLoggedIn` isn't being called. So at the very least a comment is needed to explain why, but ideally something else should be done.
2. Why is it doing `?.` ? It's guaranteed to return an array.

`o4-mini-high` had this to say (I haven't yet reviewed the rest of the code so can't comment on it yet):

<details>

## 2) **frontend/controller/actions/group‑kv.js** – onconflict logic drops existing state

You changed your merge functions to this form:

```js
const getUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
  return [{ ...currentData, [identityContractID]: nowString }, etag]
}
const data = getUpdatedLastLoggedIn()?.[0]
await sbp('chelonia/kv/set',
  contractID,
  KV_KEYS.LAST_LOGGED_IN,
  data,
  { onconflict: getUpdatedLastLoggedIn }
)
```

**Why this breaks:**  by passing `data` immediately you skip the _fetch+merge_ step on the very first write.  You only ever merge against an empty `{}`, so **all** other members’ timestamps get clobbered.  The whole point of `onconflict` being a function is that if you pass `data===undefined`, the core `chelonia/kv/set` will:

1. `GET /kv/...`  
2. call your `onconflict({ currentData, etag })`  
3. POST the merged result  

By supplying `data` up front you wind up in the “`if (data) POST directly`” branch.  

### Suggested fix
Remove the initial `data` argument so that you always go through the fetch & merge path:

```js
await sbp('chelonia/kv/set',
  contractID,
  KV_KEYS.LAST_LOGGED_IN,
  /* no data – let onconflict driver it */ undefined,
  {
    encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
    signingKeyId:   await sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk'),
    onconflict: getUpdatedLastLoggedIn
  }
)
```

That way your `getUpdatedLastLoggedIn({ currentData, etag })` will always see the _real_ current data before merging.

---

## 3) **frontend/controller/actions/identity‑kv.js** – same pattern in many places

Throughout `identity-kv.js` you unconditionally compute a `{currentData}` merge-up front and then call `saveChatRoomUnreadMessages` / `savePreferences` etc with an initial payload.  For all of these you should instead:

- Pass `undefined` (or `null`) as the `data` argument to your `queuedSet`/`set` call  
- Supply only an `onconflict` callback that does _fetch + merge_.  

For example, this block:

```js
// before
const data = getUpdatedUnreadMessages()?.[0]
if (data) {
  await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages',
    { data, onconflict: getUpdatedUnreadMessages }
  )
}

// after
await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages',
  {
    data: undefined,
    onconflict: getUpdatedUnreadMessages
  }
)
```

That guarantees that you never accidentally blow away remote updates.
</details>

shared/domains/chelonia/chelonia-utils.js
```diff
@@ -8,9 +8,10 @@ export default (sbp('sbp/selectors/register', {
   // `chelonia/kv/set`. However, the `chelonia/kv/set` primitive is needed if
   // the queueing logic needs to be more advanced, the key to use requires
   // custom logic or _if the `onconflict` callback also needs to be queued_.
-  'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+  'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, ifMatch, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
     return sbp('chelonia/queueInvocation', contractID, () => {
       return sbp('chelonia/kv/set', contractID, key, data, {
+        ifMatch,
```
taoeffect, line 24: Myself and `o4-mini-high` notice that this `ifMatch` parameter doesn't seem to be used anywhere...

<details>

## 4) **shared/domains/chelonia/chelonia-utils.js**  
You added an `ifMatch` parameter in the destructuring:

```diff
- 'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+ 'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, ifMatch, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
```

but you did **not** update your front‐end callers to ever pass `ifMatch`.  If you don’t actually need to expose `ifMatch` here then you should remove it.  Alternatively, if you do plan to pass it, you’ll need to thread it through from your HTTP‐layer into this wrapper.
</details>

taoeffect (changes requested):
Great work @corrideat! Preliminary review ready (stopped short of reviewing the last most complicated file due to running out of time).

Also including 1 AI review:

## o4-mini-high

<details>

Here are a few concrete problems I found in the diff, with suggestions for how to fix each of them. I’ve only called out lines that need changing; everything else in those files looks fine.

---

## 1) **backend/routes.js** – breaking `If‑Match` semantics

### Problem 1.1
You now **require** an `If‑Match` header on every write (and return 400 if it’s missing):

```js
+  const expectedEtag = request.headers['if-match']
+  if (!expectedEtag) {
+    return Boom.badRequest('if-match is required')
+  }
```

That will break any client that _doesn’t_ send `If‑Match`.  You almost certainly only want to apply the ETag check **if** the header is present, and otherwise do the old “no precondition” behavior.

**Old code** (only checked it if supplied):
```js
if (request.headers['if-match']) {
  … do precondition logic …
}
```

**Suggested fix**  
Restore that pattern.  Only return 412 or inject the ETag header when the header is actually there:

```js
// a bit further down in your handler
const expectedEtag = request.headers['if-match']
if (expectedEtag) {
  if (!existing) {
    // there is no prior value, so you can't match anything
    throw Boom.preconditionFailed('No existing ETag for this key')
  }
  // strip/quote whitespace, split on commas, find match
  const cid = createCID(existing, multicodes.RAW)
  const quotedCid = `"${cid}"`
  const allowed = expectedEtag
    .split(',')
    .map(v => v.trim())
    .includes(quotedCid)

  if (!allowed && expectedEtag !== '*') {
    // return a proper BOOM error (with ETag in the response headers)
    const err = Boom.preconditionFailed('ETag mismatch')
    err.output.headers = {
      'etag': quotedCid,
      'x-cid': quotedCid
    }
    throw err
  }
}
// else: no If-Match header → just continue
```

### Problem 1.2  
In your 409 branch you replaced `Boom.conflict()` with a raw `h.response(existing)…code(409)`:

```diff
-      return Boom.conflict()
+      return h.response(existing)
+        .etag(cid)
+        .header('x-cid', `"${cid}"`)
+        .code(409)
```

That now returns the _raw_ stored bytes as the body, instead of the usual JSON‐error.  If you really want to signal a conflict with the current payload, it is clearer to still use Boom:

```js
if (heightMismatch) {
  const err = Boom.conflict('Bad height')
  // Keep the client supplied ETag so they can re‑try
  err.output.headers = {
    'etag': `"${cid}"`,
    'x-cid': `"${cid}"`
  }
  throw err
}
```

---

## 2) **frontend/controller/actions/group‑kv.js** – onconflict logic drops existing state

You changed your merge functions to this form:

```js
const getUpdatedLastLoggedIn = ({ etag, currentData = {} } = {}) => {
  return [{ ...currentData, [identityContractID]: nowString }, etag]
}
const data = getUpdatedLastLoggedIn()?.[0]
await sbp('chelonia/kv/set',
  contractID,
  KV_KEYS.LAST_LOGGED_IN,
  data,
  { onconflict: getUpdatedLastLoggedIn }
)
```

**Why this breaks:**  by passing `data` immediately you skip the _fetch+merge_ step on the very first write.  You only ever merge against an empty `{}`, so **all** other members’ timestamps get clobbered.  The whole point of `onconflict` being a function is that if you pass `data===undefined`, the core `chelonia/kv/set` will:

1. `GET /kv/...`  
2. call your `onconflict({ currentData, etag })`  
3. POST the merged result  

By supplying `data` up front you wind up in the “`if (data) POST directly`” branch.  

### Suggested fix
Remove the initial `data` argument so that you always go through the fetch & merge path:

```js
await sbp('chelonia/kv/set',
  contractID,
  KV_KEYS.LAST_LOGGED_IN,
  /* no data – let onconflict driver it */ undefined,
  {
    encryptionKeyId: await sbp('chelonia/contract/currentKeyIdByName', contractID, 'cek'),
    signingKeyId:   await sbp('chelonia/contract/currentKeyIdByName', contractID, 'csk'),
    onconflict: getUpdatedLastLoggedIn
  }
)
```

That way your `getUpdatedLastLoggedIn({ currentData, etag })` will always see the _real_ current data before merging.

---

## 3) **frontend/controller/actions/identity‑kv.js** – same pattern in many places

Throughout `identity-kv.js` you unconditionally compute a `{currentData}` merge-up front and then call `saveChatRoomUnreadMessages` / `savePreferences` etc with an initial payload.  For all of these you should instead:

- Pass `undefined` (or `null`) as the `data` argument to your `queuedSet`/`set` call  
- Supply only an `onconflict` callback that does _fetch + merge_.  

For example, this block:

```js
// before
const data = getUpdatedUnreadMessages()?.[0]
if (data) {
  await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages',
    { data, onconflict: getUpdatedUnreadMessages }
  )
}

// after
await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages',
  {
    data: undefined,
    onconflict: getUpdatedUnreadMessages
  }
)
```

That guarantees that you never accidentally blow away remote updates.

---

## 4) **shared/domains/chelonia/chelonia-utils.js**  
You added an `ifMatch` parameter in the destructuring:

```diff
- 'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
+ 'chelonia/kv/queuedSet': ({ contractID, key, data, onconflict, ifMatch, encryptionKeyName = 'cek', signingKeyName = 'csk' }) => {
```

but you did **not** update your front‐end callers to ever pass `ifMatch`.  If you don’t actually need to expose `ifMatch` here then you should remove it.  Alternatively, if you do plan to pass it, you’ll need to thread it through from your HTTP‐layer into this wrapper.

---

## 5) Minor: missing import in **shared/domains/chelonia/chelonia.js**

You added `ChelErrorUnexpectedHttpResponseCode` to the import list but did not remove the old `ChelErrorUnexpected`.  If you’re not still using `ChelErrorUnexpected`, please delete it:

```diff
- import { ChelErrorResourceGone, ChelErrorUnexpected, ChelErrorUnrecoverable } from './errors.js'
+ import { ChelErrorResourceGone, ChelErrorUnexpectedHttpResponseCode, ChelErrorUnrecoverable } from './errors.js'
```

---

### Summary

1. **Do not require** `If‑Match` on every write; only apply ETag checks when the header is actually present.  
2. Use **Boom** to return 412 / 409 rather than a raw `h.response(existing)`, so clients get a consistent JSON error body.  
3. In all your front‐end KV‐set actions remove the “initial data” parameter and let `onconflict` fetch+merge for you—this avoids clobbering remote state.  
4. If you add `ifMatch` to your util, actually wire it through your callers or drop it.  
5. Clean up imports you no longer need.
</details>

shared/domains/chelonia/chelonia-utils.js
corrideat: That's because the default value is used. That will do a `get` first and populate it. (Actually, if `data` is provided, it'll to a `post` first and if there's a conflict, populate it; if `data` isn't provided a `get` is used first)

frontend/controller/actions/group-kv.js
corrideat: 1. 'Get' doesn't necessarily mean 'fetch', it can also refer to constructing or returning an object. I can try to find an alternative name. (ETA: Examples of this meaning are 'getters'; I've renamed it to constructUpdatedLastLoggedIn). ETA2: Actually, the original function which did call fetch still used get in the meaning of 'obtaining something' and not in the 'fetch' sense, since the point of this function is to modify the existing data.
2. In this case, yes. The `?` is redundant.
3. The AI isn't right. Passing `data` will _not_ clobber anything if there's data already stored on the server.

corrideat:
## 1) backend/routes.js – breaking If‑Match semantics

### Problem 1.1

Requiring `if-match` is _the_ point of this PR, as per your instructions to make
conflict resolution mandatory.

### Problem 1.2

Returning the payload instead of an error on 409 was also part of your instructions
(which I agree with) to avoid needing extra round trips

## frontend/controller/actions/group‑kv.js – onconflict logic drops existing state

LLMs seem really confused with these changes, but the logic is correct because
of conflict resultion. An initial `data` parameter will be used to set the KV
data if there's nothing there, else `onconflict` will be used. If this PR works
as intended, there's nothing to fix, and if things break, something else (likely
in the conflict resultion logic) is broken.

## frontend/controller/actions/identity‑kv.js – same pattern in many places

Ditto.

## shared/domains/chelonia/chelonia-utils.js: You added an ifMatch

Already discussed. `ifMatch` isn't necessary and will be automagically handled.
The parameter was added for completeness in case it's ever needed.

##  Minor: missing import in shared/domains/chelonia/chelonia.js

The LLM is wrong. (Noticed too this is the wrong error category, as it'd be [_superfluous / extra / unnecessary_] import, not _missing_ import)

frontend/controller/actions/identity-kv.js
```diff
@@ -319,8 +289,7 @@ export default (sbp('sbp/selectors/register', {
     }
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const notifications = sbp('chelonia/rootState').notifications.items
-      const getUpdatedNotificationStatus = async () => {
-        const currentData = await sbp('gi.actions/identity/kv/fetchNotificationStatus')
+      const getUpdatedNotificationStatus = ({ currentData = {}, etag } = {}) => {
```
taoeffect, line 292: Down below, unmodified is `getUpdatedPreferences()?.[0]`, which should at least be modified to `getUpdatedPreferences()[0]`, and any other updates as needed.

frontend/controller/actions/identity-kv.js
```diff
@@ -237,16 +210,15 @@ export default (sbp('sbp/selectors/register', {
...
         const hideDistributionBanner = {
           ...(currentPreferences.hideDistributionBanner || {}),
           [contractID]: hidden
         }
-        return { ...currentPreferences, hideDistributionBanner }
+        return [{ ...currentPreferences, hideDistributionBanner }, etag]
       }
 
-      const data = await getUpdatedPreferences()
+      const data = getUpdatedPreferences()?.[0]
```
taoeffect: This should be `getUpdatedPreferences()[0]` since it always returns something.

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
   // a general rule, you shouldn't be calling this selector directly unless
   // you're building a utility library or if you have very specific needs. In
   // this case, see if `chelonia/kv/queuedSet` covers your needs.
+  // `data` is allowed to be falsy, in which case a fetch will occur first and
+  // the `onconflict` handler will be called.
```
taoeffect, line 1650: ~~Since this comment recommends most users use `chelonia/kv/queuedSet`, and since `chelonia-utils.js` only contains that single selector, could you please move that selector into this file and ❌  `chelonia-utils.js`? (Just for readability's sake, it will also organize all relevant selectors together).~~

EDIT: ok, based on call, could you update `chelonia-utils.js` to add a comment to name the selectors that you've planning to move there?

corrideat: chelonia-utils.js is supposed to contain non-core parts of Chelonia. For example, the wrappers for encrypted action and key rotation could go in here.

(One of the reasons all of these things are non-core is that there are multiple ways of achieving the same result, and that certain conventions are imposed such as non-`#` key names)

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
   // a general rule, you shouldn't be calling this selector directly unless
   // you're building a utility library or if you have very specific needs. In
   // this case, see if `chelonia/kv/queuedSet` covers your needs.
+  // `data` is allowed to be falsy, in which case a fetch will occur first and
```
taoeffect, line 1649: If `data` is allowed to be falsey, then the Flowtype info should be updated below

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
-      const response = await this.config.fetch(`${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`, {
-        headers: new Headers([[
-          'authorization', buildShelterAuthorizationHeader.call(this, contractID)
-        ]]),
-        method: 'POST',
-        body: JSON.stringify(serializedData),
-        signal: this.abortController.signal
-      })
+      let response: Response
+      if (data || typeof onconflict !== 'function') {
```
taoeffect: Flowtype mismatch here for `onconflict`: `onconflict` is defined as being a function, yet this condition says it might not be.

Also, is there a situation where both `data` is falsey and `onconflict` is falsey? What happens then?

corrideat: > is there a situation where both data is falsey and onconflict is falsey? What happens then?

In that case a POST will be attempted.

taoeffect: > In that case a POST will be attempted.

And what will that POST do? Will it set the data to nothing? Is that the same as deleting data? I feel like we discussed the idea of adding a separate selector for deleting data.... Maybe a comment here about this?

corrideat: Hm, you're right that perhaps we need to specify this better.

No, it will _not_ delete data, but it will allow you to store values such as: `0`, `null`, `''`, `'0'`, `false`.

corrideat: Maybe the check could instead be `if (data === undefined)`?

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
+          signal: this.abortController.signal
+        })
+      } else {
+        response = await this.config.fetch(url, {
+          headers: new Headers([[
+            'authorization', buildShelterAuthorizationHeader.call(this, contractID)
+          ]]),
+          signal: this.abortController.signal
+        })
+      }
```
taoeffect: This is the case where `data` is falsey and `onconflict` is a defined — could you add a comment here about what's going on in this branch? It's unclear to me, as one would think that if `onconflict` is defined it would be used here but it's not

shared/domains/chelonia/chelonia.js
taoeffect: > Maybe the check could instead be if (data === undefined)?

I'm not sure what you mean...

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
+      const resolveData = async () => {
```
taoeffect: Could you please add some comments above this line to say what this function does / is for?

shared/domains/chelonia/chelonia.js
corrideat: what do you mean it's not? it is used below in `resolveData`. I can add an additional comment, but this branch is already commented before the function:

>  // `data` is allowed to be falsy, in which case a fetch will occur first and
>  // the `onconflict` handler will be called.

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
+          },
+          currentValue
+        })
+        if (!result) return false
+
+        data = result[0]
+        ifMatch = result[1]
+        return true
+      }
+      if (!data && typeof onconflict === 'function') {
```
taoeffect: > Maybe the check could instead be if (data === undefined)?

Whatever you decide above regarding this question, it should probably be applied here too?

corrideat: Correct

shared/domains/chelonia/chelonia.js
taoeffect: I mean it's not immediately used in this block... maybe the comment here can say that `onconflict` will be used in this case down below..

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
+        if (response.ok || response.status === 409 || response.status === 412) {
+          const serializedData = await response.json()
+          currentValue = parseEncryptedOrUnencryptedMessage.call(this, {
+            contractID,
+            serializedData,
+            meta: key
+          })
+        } else if (response.status !== 404 && response.status !== 410) {
+          throw new ChelErrorUnexpectedHttpResponseCode('[kv/set] Invalid response code: ' + response.status)
+        }
```
taoeffect: This whole section is a bit confusing to me as it's not clear to me what these response codes signify in terms of what was sent to the server.

E.g. why is 404 being paired with 410 here? And in what situation would 404 be returned?

Are 409 and 412 the same situation, if not, how are they different, and if they are, why do we use different codes?

Can you add a comment above these cases explaining all of this? Would make it a lot easier to read and reason about without having to jump to the server code.

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,110 @@ export default (sbp('sbp/selectors/register', {
...
+        const result = await onconflict({
```
taoeffect: Why do we call `onconflict` even if `response.ok` was returned (indicating there was no conflict)? Can you add a comment here about that too?

shared/domains/chelonia/chelonia.js
corrideat: I'll add comments, but this isn't sending anything to the server though, it's parsing a response.

404 and 410 are paired because they're very similar and represent there's no data (so new data can be set).

409 and 412 represent conflicts, although of different sorts. 409 is a height conflict while 412 is an if-match (CID / ETag) conflict.

This is parsing the response obtained, if it was 2xx or 409 or 412 (which will include a response that can be parsed). Otherwise, it'll throw unless the error is 404 / 410, which means that new data can be set.

shared/domains/chelonia/chelonia.js
corrideat: Sure. Because the initial request could've been a `GET`. `resolveData` should only be called after a conflict has occurred though (or after an initial fetch).

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,116 @@ export default (sbp('sbp/selectors/register', {
...
+        })
+        if (!result) return false
+
+        data = result[0]
+        ifMatch = result[1]
+        return true
+      }
+      if (!data && typeof onconflict === 'function') {
+        if (await resolveData()) {
+          continue
```
taoeffect, line 1756: This code is extremely confusing to me.

Above we check `if (!data` and we enter into here. In most cases `resolveData()` will result in `data` being set. Then we loop again as a result of the `continue`, but from that point on this code will never be run again because `!data` will be `false`.... Is that right? If so, why is it being skipped from that point on? (A comment is needed). If not, a bug.

corrideat: That'd be for the initial iteration. Later on, there's a similar code in the `if` that follows

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,116 @@ export default (sbp('sbp/selectors/register', {
...
+        if (response.status === 409 || response.status === 412) {
           if (--maxAttempts <= 0) {
             throw new Error('kv/set conflict setting KV value')
           }
+          // Only retry if an onconflict handler exists to potentially resolve it
           await delay(randomIntFromRange(0, 1500))
           if (typeof onconflict === 'function') {
-            data = await onconflict(contractID, key, data)
-            if (!data) { break }
+            if (await resolveData()) {
```
taoeffect, line 1770: The check for `typeof onconflict === 'function'` (or the inverse) is done 3 times... maybe this condition could be cached and DRY'd?

shared/domains/chelonia/chelonia.js
corrideat: I meant that the check could be made stricter and simpler (skipping the `onconflict` part).

corrideat: (I.e., if you want to save `false`, you shouldn't need to set `onconflict` to undefined)

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,135 @@ export default (sbp('sbp/selectors/register', {
...
+        if (response.status === 409 || response.status === 412) {
           if (--maxAttempts <= 0) {
             throw new Error('kv/set conflict setting KV value')
           }
+          // Only retry if an onconflict handler exists to potentially resolve it
           await delay(randomIntFromRange(0, 1500))
-          if (typeof onconflict === 'function') {
-            data = await onconflict(contractID, key, data)
-            if (!data) { break }
+          if (hasOnconflict) {
```
corrideat: This is Gemini's take on refactoring this function. I have reviewed it lightly and it also seems to work (as in, tests are passing). I have not thoroughly reviewed it for all cases.

Let me know if you prefer the refactored version or the initial version.

```js
  'chelonia/kv/set': async function cheloniaKvSet (
    contractID: string,
    key: string,
    // Initial data to set. If null/undefined, a fetch is performed first.
    initialData: ?Object,
    {
      // Target ETag/CID for conditional write (compare-and-swap).
      ifMatch: initialIfMatch,
      // Signing key for the inner payload (if different from outer).
      innerSigningKeyId,
      // Key ID for encrypting the data.
      encryptionKeyId,
      // Key ID for signing the request envelope.
      signingKeyId,
      // Maximum number of attempts including the initial one.
      maxAttempts = 3,
      // Callback to resolve conflicts or determine data after initial fetch.
      // Must return `false` to abort, or `[newData, newIfMatch]` to retry.
      onconflict
    }: {
      ifMatch?: ?string, // Allow null/undefined explicitly
      innerSigningKeyId: ?string,
      encryptionKeyId: ?string,
      signingKeyId: string,
      maxAttempts?: ?number, // Allow null/undefined for default
      onconflict?: ?ChelKvOnConflictCallback,
    }
  ) {
    // --- Input Validation & Setup ---
    if (initialData === undefined && typeof onconflict !== 'function') {
      throw new TypeError('chelonia/kv/set: `onconflict` callback is required when no initial `data` is provided.')
    }
    if (!signingKeyId) {
      throw new TypeError('chelonia/kv/set: `signingKeyId` is required.')
    }
    if (maxAttempts <= 0) {
      throw new RangeError('chelonia/kv/set: `maxAttempts` must be positive.')
    }

    const url = `${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`
    const buildAuthHeader = () => buildShelterAuthorizationHeader.call(this, contractID)
    const hasOnConflictHandler = typeof onconflict === 'function'
    const needsInitialFetch = initialData === undefined

    // State for the current attempt
    let dataForThisAttempt: ?Object = initialData
    // Use explicit null/undefined if passed
    let ifMatchForThisAttempt: ?string = initialIfMatch
    let attempt = 0
    let lastError = null

    // --- Retry Loop ---
    while (attempt < maxAttempts) {
      const isInitialFetch = attempt === 0 && needsInitialFetch
      attempt++
      let response: Response | undefined

      try {
        // --- Phase 1: Determine Data and Perform Action (GET or POST) ---
        if (!isInitialFetch) {
          // We have data, attempt to POST it
          const serializedData = outputEncryptedOrUnencryptedMessage.call(this, {
            contractID,
            innerSigningKeyId,
            encryptionKeyId,
            signingKeyId,
            data: dataForThisAttempt, // Use data for the current attempt
            meta: key
          })

          const headers = new Headers([
            ['authorization', buildAuthHeader()],
            // Use '""' for explicit empty ETag condition if ifMatch is nullish
            ['if-match', ifMatchForThisAttempt ?? '""']
          ])

          response = await this.config.fetch(url, {
            method: 'POST',
            headers: headers,
            body: JSON.stringify(serializedData),
            signal: this.abortController.signal
          })
        } else {
          // No data provided yet, need to perform a GET first
          // This only happens on the very first iteration if initialData was undefined.
          response = await this.config.fetch(url, {
            method: 'GET', // Explicitly GET
            headers: new Headers([['authorization', buildAuthHeader()]]),
            signal: this.abortController.signal
          })
          // The conflict handler below will process this GET response
        }

        // --- Phase 2: Handle Response ---

        // Success Case (POST succeeded)
        if (response.ok && !isInitialFetch) {
          // Successfully wrote the data. We are done.
          return // EXIT Success
        }

        // Conflict or Need-Data Cases (requires onconflict handler)
        // Statuses needing resolution:
        // - 200 OK (after GET): Need to determine initial data via onconflict.
        // - 409 Conflict (POST): Resource version conflict (height mismatch).
        // - 412 Precondition Failed (POST): ETag mismatch (`if-match` failed).
        // - 404 Not Found (after GET): Key doesn't exist, need initial data.
        // - 410 Gone (after GET): Key doesn't exist, need initial data.
        const needsConflictResolution =
          isInitialFetch
            ? response.ok // Successful GET needs data resolution
            : response.status === 409 || // Conflict
              response.status === 412 // Precondition Failed

        if (needsConflictResolution) {
          if (!hasOnConflictHandler) {
            // Should have been caught earlier if initialData was undefined,
            // but this handles the case where a POST fails without a handler.
            throw new Error(`chelonia/kv/set: Status ${response.status} requires an \`onconflict\` handler to proceed, but none was provided.`)
          }

          // Prepare context for onconflict callback
          let currentValue // Default to undefined if parsing fails or not applicable
          const etag = response.headers.get('x-cid') || response.headers.get('etag') || null // Standard ETag header

          // Try to parse current data from response body if applicable
          if (response.ok || response.status === 409 || response.status === 412) {
            try {
              const serializedData = await response.json()
              currentValue = parseEncryptedOrUnencryptedMessage.call(this, {
                contractID,
                serializedData,
                meta: key
              })
            } catch (parseError) {
              // Log parsing error? Or let onconflict handle potentially missing currentValue?
              console.warn(`chelonia/kv/set: Failed to parse response body for conflict resolution: ${parseError.message}`)
              // currentValue remains undefined
            }
          }
          // For 404/410 on GET, currentValue remains undefined, which is correct.

          // Invoke the conflict handler
          const conflictResult = await (onconflict: Function)({
            contractID,
            key,
            // The data we *tried* to set (or undefined if initial GET)
            failedData: dataForThisAttempt,
            status: response.status,
            etag,
            get currentData () {
              // Provide access via getter to avoid unnecessary parsing
              return currentValue?.data
            },
            // Pass the full parsed value (includes metadata if any)
            currentValue,
            // Pass current attempt number
            attempt: attempt
          })

          if (conflictResult === false) {
            // Handler decided to abort.
            return // EXIT Abort by Handler
          }

          if (!Array.isArray(conflictResult) || conflictResult.length !== 2) {
            throw new TypeError('chelonia/kv/set: `onconflict` handler must return `false` or an array `[newData, newIfMatch]`.')
          }

          // Update state for the *next* attempt
          dataForThisAttempt = conflictResult[0]
          ifMatchForThisAttempt = conflictResult[1]

          // If conflict resolution was triggered, apply a delay before retrying
          // (unless it was the initial GET which doesn't need a delay)
          if (response.status === 409 || response.status === 412) {
            // Only delay if we are going to retry
            if (attempt < maxAttempts) {
              // Exponential backoff basic
              await delay(randomIntFromRange(50, 500 + (attempt * 100)))
            }
          }
          // Loop continues for the next attempt with updated data/ifMatch
          continue // Go to next iteration
        } // end if (needsConflictResolution)

        // --- Phase 3: Handle Unexpected Errors ---
        // If we reach here, the response was not OK and not a handled conflict/fetch case.
        // Note: 404/410 after a POST are considered unexpected here.
        throw new ChelErrorUnexpectedHttpResponseCode(
           `[chelonia/kv/set] Unexpected HTTP status code on attempt ${attempt}: ${response.status}`
        )
      } catch (error) {
        lastError = error // Store the error from this attempt
        // If the error was due to the network request itself (not response
        // handling) or an unexpected status code, and we have retries left,
        // apply backoff delay.
        if (attempt < maxAttempts) {
          // Avoid delaying if error is immediate (e.g., TypeError before fetch)
          if (response || error instanceof ChelErrorUnexpectedHttpResponseCode) {
            await delay(randomIntFromRange(50, 500 + (attempt * 100)))
            continue // Retry
          } else {
            // Programming error (e.g. TypeError), don't retry, rethrow immediately
            throw error
          }
        } else {
          // No retries left, rethrow the last encountered error
          throw error
        }
      }
    } // end while (attempt < maxAttempts)

    // --- Phase 4: Handle Exhausted Retries ---
    // If the loop finished without returning or throwing, it means maxAttempts
    // were reached.
    throw lastError ?? new Error(`chelonia/kv/set: Failed to set KV value for key "${key}" after ${maxAttempts} attempts.`)
  },

```

taoeffect: @corrideat Could you compare that refactoring to the one below form `o4-mini (high)`? I asked it for both a review and a refactor.

The refactoring here seems (just from a quick visual inspection) nicer / shorter. However, it's missing the comments you added (which would need to be added back), assuming the code is correct that is.

## `o4-mini (high)` feedback

Below are the specific issues I spotted in your changeset around the `chelonia/kv/set` selector, followed by a proposed, much-simpler refactoring of that function.

---

## 1) `data?: ?Object` vs distinguishing `null` vs `undefined`

You wrote:

```js
// shared/domains/chelonia/chelonia.js
'chelonia/kv/set': async function (contractID: string, key: string, data?: ?Object, {
  ifMatch,
  … 
}) {
  …
  if (data !== undefined) {
    // POST path
  } else {
    // GET path
  }
  …
}
```

**Problem:** the signature `data?: ?Object` allows both `undefined` and `null`. But your branch test only checks for `undefined`. If a caller accidentally passes `null`, you will enter the “POST” path with `data === null`, and then blow up inside `outputEncryptedOrUnencryptedMessage`. You probably meant “no data at all” to always be `undefined`.

**Suggestion:**  
– Change the signature to `data?: Object` (no nullable).  
– Or in your code treat `data == null` as “no data”:

```diff
- if (data !== undefined) {
+ if (data != null) {
```

---

## 2) `onconflict` naming + type does not match actual callback arguments

You declared:

```js
export type ChelKvOnConflictCallback = (
  args: { contractID: string, key: string, failedData: Object, status: number, etag: ?string, currentData: Object }
) => Promise<[Object, string]>
```

but in your code you actually pass **two** additional fields:

```js
const result = await (onconflict: Function)({
  contractID,
  key,
  failedData: data,
  status: response.status,
  etag: response.headers.get('x-cid') || response.headers.get('etag'),
  get currentData () { return currentValue?.data },
  currentValue
})
```

– You pass `currentValue` and a getter `currentData`, but your Flow‐type only describes `currentData`.  
– Also the parameter is called `onconflict` in code but your type is `ChelKvOnConflictCallback`.  

**Suggestion:**  
1. Rename the param to `onConflict` (camelCase).  
2. Update the type to include everything you actually pass:

```js
export type ChelKvOnConflictCallback = (args: {
  contractID: string,
  key: string,
  failedData: Object,
  status: number,
  etag?: string,
  currentValue?: any,       // the full parsed object
  currentData?: any         // just the .data field
}) => Promise<[Object, string]>
```

---

## 3) duplicate “resolveData” logic

You have two almost‐identical blocks:

```js
// initial GET
if (data === undefined) {
  … 
  if (await resolveData()) {
    continue
  } else {
    break
  }
}

// conflict‐retry
if (!response.ok && (response.status === 409 || response.status === 412)) {
  …
  if (hasOnconflict) {
    if (await resolveData()) {
      continue
    } else {
      break
    }
  }
}
```

**Problem:** you’re copy-pasting the same two lines. This makes the function hard to follow.

**Suggestion:** pull the “fetch‐then‐merge” into a single helper, and simply loop. See the refactored version below.

---

## 4) building headers is noisy

```js
headers: new Headers([[
  'authorization', buildShelterAuthorizationHeader.call(this, contractID)
], [
  'if-match', ifMatch || '""'
]])
```

It’s clearer to do:

```js
const headers = new Headers()
headers.append('authorization', buildShelterAuthorizationHeader.call(this, contractID))
if (ifMatch != null) {
  headers.append('if-match', ifMatch)
}
…
```

---

## 5) avoid inline Flow casts in executable code

You wrote:

```js
const result = await (onconflict: Function)({ … })
```

Better to just call:

```js
const result = await onConflict({ … })
```

and trust your Flow annotation that `onConflict` is a function.

---

## Full refactored `chelonia/kv/set`

Below is a leaner, single‐loop implementation. It

1. separates **GET** vs **POST**,  
2. centralizes conflict resolution, and  
3. uses clearer variable names (`onConflict`, `maxTries`, etc.).  

```js
// shared/domains/chelonia/chelonia.js
import {
  // … other imports …
  ChelErrorUnexpectedHttpResponseCode
} from './errors.js'

// update the type to match what you pass in:
export type ChelKvOnConflictCallback = (args: {
  contractID: string,
  key: string,
  failedData: any,
  status: number,
  etag?: string,
  currentValue?: any,
  currentData?: any
}) => Promise<[any, string]>

export default sbp('sbp/selectors/register', {
  // …
  'chelonia/kv/set': async function (
    contractID: string,
    key: string,
    data: any,    // always require a data object
    {
      ifMatch,
      innerSigningKeyId,
      encryptionKeyId,
      signingKeyId,
      maxAttempts,
      onConflict
    }: {
      ifMatch?: string,
      innerSigningKeyId?: string,
      encryptionKeyId?: string,
      signingKeyId: string,
      maxAttempts?: number,
      onConflict?: ChelKvOnConflictCallback
    }
  ) {
    const url = [
      this.config.connectionURL,
      'kv',
      encodeURIComponent(contractID),
      encodeURIComponent(key)
    ].join('/')
    const tries = maxAttempts || 3
    let attempt = 0

    // helper to do a GET and let onConflict produce fresh data + ifMatch
    async function fetchAndMerge(this: any): Promise<boolean> {
      if (typeof onConflict !== 'function') {
        throw new TypeError('chelonia/kv/set: onConflict callback is required for initial fetch or conflict')
      }
      const resp = await this.config.fetch(url, {
        headers: new Headers({
          authorization: buildShelterAuthorizationHeader.call(this, contractID)
        }),
        signal: this.abortController.signal
      })
      if (![200, 409, 412, 404, 410].includes(resp.status)) {
        throw new ChelErrorUnexpectedHttpResponseCode(
          `chelonia/kv/set unexpected GET status ${resp.status}`
        )
      }

      // parse current value if available
      let currentValue
      if (resp.ok || resp.status === 409 || resp.status === 412) {
        const serialized = await resp.json()
        currentValue = parseEncryptedOrUnencryptedMessage.call(this, {
          contractID,
          serializedData: serialized,
          meta: key
        })
      }

      const newEtag = resp.headers.get('x-cid') || resp.headers.get('etag') || undefined
      const [mergedData, mergedIfMatch] = await onConflict.call(this, {
        contractID,
        key,
        failedData: data,
        status: resp.status,
        etag: newEtag,
        currentValue,
        currentData: currentValue && currentValue.data
      })

      // if consumer returns null/undefined, bail out
      if (mergedData == null) {
        return false
      }

      data = mergedData
      ifMatch = mergedIfMatch
      return true
    }

    // if client invoked `kv/set` without data, do an initial fetch+merge:
    if (data == null) {
      const shouldRetry = await fetchAndMerge.call(this)
      if (!shouldRetry) {
        return
      }
    }

    // now we have `data` and `ifMatch`, try POST up to N times
    while (true) {
      attempt++
      // build headers
      const headers = new Headers()
      headers.append(
        'authorization',
        buildShelterAuthorizationHeader.call(this, contractID)
      )
      headers.append('if-match', ifMatch || '""')

      // encrypt + sign
      const serialized = outputEncryptedOrUnencryptedMessage.call(this, {
        contractID,
        innerSigningKeyId,
        encryptionKeyId,
        signingKeyId,
        data,
        meta: key
      })

      const resp = await this.config.fetch(url, {
        method: 'POST',
        headers,
        body: JSON.stringify(serialized),
        signal: this.abortController.signal
      })

      // success!
      if (resp.ok) {
        return
      }

      // conflict, retry?
      if ((resp.status === 409 || resp.status === 412) && attempt < tries) {
        // optional back-off
        await delay(randomIntFromRange(0, 1500))
        const merged = await fetchAndMerge.call(this)
        if (!merged) {
          return
        }
        continue
      }

      // out of tries or unexpected code
      if (resp.status === 409 || resp.status === 412) {
        throw new Error(`chelonia/kv/set: conflict after ${attempt} attempts`)
      }
      throw new ChelErrorUnexpectedHttpResponseCode(
        `chelonia/kv/set invalid response ${resp.status}`
      )
    }
  },

  // … other selectors …
})
```

### Why it’s simpler

1. **One loop**, not two special cases.  
2. `fetchAndMerge` handles both the initial GET‐merge and subsequent conflict‐merge.  
3. Clear variable names (`onConflict`, `attempt`, `tries`).  
4. No inline Flow casts, no duplicate `resolveData` blocks.  
5. Better flow: if client doesn’t pass `data`, we do exactly one GET+merge, then fall through to the POST loop.

Feel free to adapt the exact error classes or logging hooks to your style, but this pattern minimizes duplication and is much easier to follow.

corrideat: > One loop, not two special cases.

The original code also has a single loop, although it does have two special cases for the initial fetch.

> fetchAndMerge handles both the initial GET‐merge and subsequent conflict‐merge.

This is wrong, because this does a `GET` on conflict which isn't needed since the conflict response returns a payload. So, this solution results in additional requests.

> Clear variable names (onConflict, attempt, tries).

Sure, but then I think that variable names were clear before too.

> No inline Flow casts, no duplicate resolveData blocks.

The inline Flow cast is needed because `onconflict` is optional. The 'duplicate' block is here too:

```js
      const shouldRetry = await fetchAndMerge.call(this)
      if (!shouldRetry) {
        return
      }
      // ...
        const merged = await fetchAndMerge.call(this)
        if (!merged) {
          return
        }
        continue
```

Arguably the duplicated logic is slightly shorter, but it's still pretty much duplicated, as in the original code

> Better flow: if client doesn’t pass data, we do exactly one GET+merge, then fall through to the POST loop.

This is wrong, as the original code already does this AFAICT, except that this will also do an additional GET each time there's a conflict.

taoeffect: What about the `2) onconflict naming + type does not match actual callback arguments` and `data?: ?Object vs distinguishing null vs undefined` thing it mentioned? Should that type stuff be updated?

Similar question for `onconflict?: ?ChelKvOnConflictCallback`

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,135 @@ export default (sbp('sbp/selectors/register', {
...
+          throw TypeError('onconflict required with empty data')
+        }
+        // This is only for the initial case; the logic is replicated below
+        // for subsequent iterations that require conflic resolution.
+        if (await resolveData()) {
+          continue
+        } else {
+          break
+        }
+      }
```
taoeffect: Could you please move the contents of this `if` block inside of the `else` branch on line `1691`?

shared/domains/chelonia/chelonia.js
```diff
@@ -1642,50 +1646,135 @@ export default (sbp('sbp/selectors/register', {
...
+        // These are not treated as errors since we could still set the value.
+        } else if (response.status !== 404 && response.status !== 410) {
+          throw new ChelErrorUnexpectedHttpResponseCode('[kv/set] Invalid response code: ' + response.status)
+        }
+        const result = await (onconflict: Function)({
+          contractID,
+          key,
+          failedData: data,
+          status: response.status,
+          etag: response.headers.get('x-cid') || response.headers.get('etag'),
```
taoeffect: Will the `etag` be properly set if `response.status` equals `404` or `410`?

If so, could you add a reassuring comment here to that effect above this line?

shared/domains/chelonia/chelonia.js
corrideat: `onconflict` has always been called `onconflict`, calling it `onConflict` is a valid choice too, but an aesthetic one.

> type does not match actual callback arguments
> You pass currentValue and a getter currentData, but your Flow‐type only describes currentData

That's correct. `currentValue` should be added.

> onconflict?: ?ChelKvOnConflictCallback

It's allowed to be optional (probably not very useful, as conflicts may happen, but developers might want to handle it some other way).

shared/domains/chelonia/chelonia.js
corrideat: The `etag` value is `''` by default, signalling writing to where there's no data. If the response was 404, this default value will be used, unless some other value was explicitly given (weird use case I guess)

taoeffect (approved):
Amazing work here @corrideat!

Issue #2630, opened 2025-02-14
Copy updates to Modal & other components to backend dashboard


taoeffect:
### Problem

Some updates were made recently to the behavior of `Modal.vue` and related components, as well as possibly other important components.

But these updates haven't been copied over to the copies that are under `backend/dashboard`.

### Solution

Copy them over and verify they work properly.

SebinSong:
@taoeffect 

### 1.
> Some updates were made recently to the behavior of Modal.vue

The recent updates I made to `Modal.vue` were bug-fixes for when two modals being displayed at the same time.
(Eg. A failure in group-creation modal leads to displaying an error prompt -> Then clicking the close button on the prompt does nothing. -> user is forced to refresh the app. This is one bug I fixed)

When I built the Modal system in the back-end dashboard, I did not implement this double modals concept because it brings extra complexity in the logic vs. there did not seem to be much benefit from it yet.

But I do think this Prompt dialog below would be useful for the project for various purposes.

<img src='https://github.com/user-attachments/assets/5b62a8fb-63a2-4a51-9019-0a1feef049d1' width='460'>

So, let me build this for Chelonia dashboard and also and work on ensuring it can display a prompt and a modal at the same time(The logic might not be exactly the same as that in GI)


### 2.
>  as well as possibly other important components.

Just took some time to search for some potentially useful UIs and for now I've found these two:

(1) Pills

<img src='https://github.com/user-attachments/assets/52bcf01d-82d8-49e5-b86b-edfd017cd280' width='420'>

(2) Character limit indicator UI

<img src='https://github.com/user-attachments/assets/055ab8df-642e-4908-9bd9-aa084b93e6d3' width='420'>

I will build these two too and add them to existing pages for demonstration.

SebinSong:
## DEV Progress Checklist

- [x] **1.** Build `Prompt.vue` / ensure it works as a modal on top of modal situation as well. (**status**: done ✅)

<img src='https://github.com/user-attachments/assets/4fe0a8aa-950c-4e78-ad68-c86c32c84b27' width='430'>

- [x] **2.** Build `CharLimitIndicator.vue` (**status**: done ✅)

<img src='https://github.com/user-attachments/assets/2dc79daf-b71e-4eb7-ae32-c785a3d6297c' width='570'>


- [x] **3.** Implement design for pills UI element. (**status**: done ✅)

[...]

PR #2743, opened 2025-04-07
#2492 - Tooltip issue on iOS-Safari

### LINKED ISSUE #2492: Cannot close tooltip on iOS
### Problem

On iOS, there is no way for a user to close a tooltip on a page except by opening a different tooltip.
The tooltips do disappear if the user opens the nav menu or switches away from the tab/page/screen showing the tooltip.

Note: This is the remaining issue from #2450 , which seems to be iOS only, as it has not been reproducible on Android.

### Solution

Tooltip should be closable. Clicking on other links/buttons or clicking on a random position in the background should close the tooltip.

SebinSong:
closes #2492 

**[ video recording for bug-fix ]**

https://github.com/user-attachments/assets/54f8c25d-df93-48a0-b2f4-f977a5436b5d

The issue happens because `blur` event does not fire on iOS Safari when user touches outside a target element. I've tried various other events suggested on the internet(eg. `mouseout`, `focusout` etc..) but nothing worked. So had to implement a workaround logic that uses `touchstart` event attached to `document.body` (details are commented inline).

### DIFF
--- a/frontend/views/components/Tooltip.vue
+++ b/frontend/views/components/Tooltip.vue
@@ -144,6 +144,19 @@ export default ({
         this.isActive = false
       }
     },
+    isIOSSafari () {
+      const ua = window.navigator.userAgent
+      return /iP(hone|od|ad)/.test(ua) && /Safari/.test(ua)
+    },
+    hideTooltipOnTouchOutside (e) {
+      // NOTE: This is a fix for the iOS-Safari-only issue https://github.com/okTurtles/group-income/issues/2492
+      //       where the tooltip does not close when the user touches outside the tooltip.
+      //       The issue happens because 'blur' event in iOS-Safari does not fire and as a workaround,
+      //       we listen for 'touchstart' event attached to the document.body and manually check if user touches outside the tooltip wrapper.
+      if (this.isActive && !this.triggerDOM.contains(e.target)) {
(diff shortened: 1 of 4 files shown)


frontend/views/components/Tooltip.vue
```diff
@@ -333,6 +352,10 @@ export default ({
     this.triggerDOM.removeEventListener('mouseleave', this.hide)
     this.triggerDOM.removeEventListener('focus', this.show)
     this.triggerDOM.removeEventListener('blur', this.hide)
+
```
taoeffect, line 355: Above this the LLMs found a significant typo:

`beforeDestory` => `beforeDestroy`

taoeffect (changes requested):
Great work @SebinSong!

Especially great comments! 👍 

Only one minor (but important) issue found, commented below.

Also, could you please update the `Tooltip.vue` file that's on the backend dashboard to have all these changes too?

SebinSong:
@taoeffect 
> Also, could you please update the Tooltip.vue file that's on the backend dashboard to have all these changes too?

I can take care of that as part of https://github.com/okTurtles/group-income/issues/2630 (Probably next week).

taoeffect (approved):
Awesome!

PR #2747, opened 2025-04-11
Running tests sequentially

### LINKED ISSUE #2740: Cypress: `group-chat-message.spec.js` fails with "Received login event but there already is an active session" error.
### Problem
Since merging the latest `master`, running cypress test locally fails with below error in `group-chat-message.spec.js`.

Image

### Solution
Identify the cause and fix it.

corrideat:
Fix #2740


SebinSong (approved):
The fix works well. Thanks @corrideat for quickly resolving it.

PR #2749, opened 2025-04-14
#2719 - Auto-scrolling does not work when the browser tab is inactive.

### LINKED ISSUE #2719: New chats do not autoscroll down when tab is in background on Firefox
### Problem

When the Group Income tab is in the background, the chat hides new messages behind the text input box. 

Image

Firefox desktop.

### Solution

Fix.

SebinSong:
closes #2719 

**[Video for the fix]**

https://github.com/user-attachments/assets/c0bc89ce-978e-4847-aa23-ba05a4837fe4

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -175,7 +175,7 @@ const onChatScroll = function () {
   const curScrollTop = this.$refs.conversation.scrollTop
   const curScrollBottom = curScrollTop + this.$refs.conversation.clientHeight
   const scrollTopMax = this.$refs.conversation.scrollHeight - this.$refs.conversation.clientHeight
-  this.ephemeral.scrolledDistance = scrollTopMax - curScrollTop
+  this.ephemeral.scrollableDistance = scrollTopMax - curScrollTop
 
   for (let i = this.messages.length - 1; i >= 0; i--) {
     const msg = this.messages[i]
@@ -201,7 +201,7 @@ const onChatScroll = function () {
     return
   }
 
-  if (this.ephemeral.scrolledDistance > ignorableScrollDistanceInPixel) {
+  if (this.ephemeral.scrollableDistance > ignorableScrollDistanceInPixel) {
     // Save the current scroll position per each chatroom
     for (let i = 0; i < this.messages.length - 1; i++) {
       const msg = this.messages[i]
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -616,11 +616,14 @@ export default ({
...
-        this.$refs.conversation.scroll({
+          this.$refs.conversation.scroll({
           left: 0,
           top: this.$refs.conversation.scrollHeight,
-          behavior: this.isReducedMotionMode
-            ? 'instant' // force 'instant' behaviour in reduced-motion mode regardless of the passed param.
+          // NOTE-1: Force 'instant' behaviour in reduced-motion mode regardless of the passed param.
+          // NOTE-2: Browsers suspend DOM animation when the tab is inactive. Passing 'smooth' option for an inactive browser window
+          //         leads to the scroll() action being ignored. So we need to explicitly pass 'instant' option in this case.
+          behavior: this.isReducedMotionMode || document.hidden
```
SebinSong, line 625: The cause of the issue was `Element.scroll()` with the `behavior: 'smooth'` option does not work when the browser tab is inactive. Below is an answer from ChatGPT regarding this behavior:

> This is a subtle (and frustrating) behavior in browsers—many of them throttle or suspend DOM updates, animations, and even some JavaScript operations when the tab is inactive to preserve resources.

Forcing the `behavior` option to be `instant` when `document.hidden` fixes the bug.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1049,12 +1053,13 @@ export default ({
 
           this.latestEvents.push(serializedMessage)
 
-          if (this.ephemeral.scrolledDistance < 50) {
+          // When the current scroll position is nearly at the bottom and a new message is added, auto-scroll to the bottom.
+          if (this.ephemeral.scrollableDistance < 50) {
```
SebinSong, line 1056: `this.ephemeral.scrollableDistance < 50` here is an existing logic, it means the auto-scrolling to the new message is **only** executed when the current scroll position is **near the bottom**. My guess for the intention is that we prevent this behavior when the user is reading old messages in the chat, which feels like something we shouldn't remove/interrupt. Pls let me know if we need any adjustment to this though.

taoeffect (approved):
Great work @SebinSong! 👏

PR #2751, opened 2025-04-14
Add previousKeyOp

### LINKED ISSUE #2750: Implement previousKeyOp
### Problem

For implementing #354 we are going to need a way to securely request just the key-affecting operations in a fast way.

### Solution

Image

Add a `previousKeyOp` field to the header of each message that points to the CID of the previous key-affecting operation. This could be a message using `OP_KEY_ADD` and related KEY operations, or it could be an `OP_ATOMIC` message that contains a key affecting operation.

corrideat:
Closes #2750 

> [!CAUTION]
> This is a breaking change that requires re-creating contracts

### DIFF
--- a/backend/database.js
+++ b/backend/database.js
@@ -79,6 +79,21 @@ export default ((sbp('sbp/selectors/register', {
         }
       }
     })()
+    const fetchMeta = async () => {
+      if (currentHeight > latestHEADinfo.height) {
+        return false
+      }
+      const meta = await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
+      if (!meta) {
+        return false
+      }
+
+      const { hash: newCurrentHash, ...newServerMeta } = meta
+      currentHash = newCurrentHash
+      serverMeta = newServerMeta
+
+      return true
+    }
     // NOTE: if this ever stops working you can also try Readable.from():
     // https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options
     const stream = new Readable({
@@ -96,33 +111,16 @@ export default ((sbp('sbp/selectors/register', {
           this.push(null)
           ended = true
         }
-        const fetchMeta = async () => {
-          if (currentHeight > latestHEADinfo.height) {
-            end()
-            return false
-          }
-          const { hash: newCurrentHash, ...newServerMeta } = (
-            await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
-          ) || {}
-          if (!newCurrentHash) {
-            end()
-            return false
-          }
-          currentHash = newCurrentHash
-          serverMeta = newServerMeta
-
-          return true
-        }
         if (counter < limit) {
           (async () => {
             try {
               if (options.keyOps) {
                 while (!serverMeta?.isKeyOp) {
                   await nextKeyOp()
-                  if (!fetchMeta()) return
+                  if (!await fetchMeta()) return end()
                 }
               } else {
-                if (!fetchMeta()) return
+                if (!await fetchMeta()) return end()
               }
               const entry = await sbp('chelonia/db/getEntry', currentHash)
               if (entry) {
--- a/backend/database.js
+++ b/backend/database.js
@@ -52,7 +52,7 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let currentHeight = parseInt(height)
+    let currentHeight = height
     let currentHash, serverMeta
     let prefix = ''
     let ended = false
--- a/backend/database.js
+++ b/backend/database.js
@@ -55,7 +55,6 @@ export default ((sbp('sbp/selectors/register', {
     let currentHeight = height
     let currentHash, serverMeta
     let prefix = ''
-    let ended = false
     const nextKeyOp = (() => {
       let index
 
@@ -65,6 +64,7 @@ export default ((sbp('sbp/selectors/register', {
         }
         const value = index?.find((h, i) => {
           if (Number(h) >= currentHeight) {
+            // Remove values that no longer are relevant from the index
             index = ((index: any): any[]).splice(i + 1)
             return true
           } else {
@@ -72,11 +72,18 @@ export default ((sbp('sbp/selectors/register', {
           }
         })
         if (value != null) {
-          currentHeight = Number(value)
+          const newHeight = Number(value)
+          currentHeight = newHeight
         } else {
           currentHeight = currentHeight - currentHeight % 10000 + 10000
           index = undefined
+          if (currentHeight > latestHEADinfo.height) {
+            return false
+          } else {
+            return null
+          }
         }
+        return true
       }
     })()
     const fetchMeta = async () => {
@@ -96,58 +103,40 @@ export default ((sbp('sbp/selectors/register', {
     }
     // NOTE: if this ever stops working you can also try Readable.from():
     // https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options
-    const stream = new Readable({
-      construct (cb) {
-        this.push('[')
-        cb()
-      },
-      read (): void {
-        if (ended) {
-          this.destroy()
-          return
-        }
-        const end = () => {
-          this.push(']')
-          this.push(null)
-          ended = true
-        }
-        if (counter < limit) {
-          (async () => {
-            try {
-              if (options.keyOps) {
-                while (!serverMeta?.isKeyOp) {
-                  await nextKeyOp()
-                  if (!await fetchMeta()) return end()
-                }
-              } else {
-                if (!await fetchMeta()) return end()
-              }
-              const entry = await sbp('chelonia/db/getEntry', currentHash)
-              if (entry) {
-                const currentPrefix = prefix
-                prefix = ','
-                counter++
-                this.push(
-                `${currentPrefix}"${strToB64(
-                  JSON.stringify({ serverMeta, message: entry.serialize() })
-                )}"`
-                )
-                currentHeight++
-                currentHash = undefined
-                serverMeta = undefined
-              } else {
-                end()
-              }
-            } catch (e) {
-              console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
-              end()
-            }
-          })()
-        } else {
-          end()
+    const stream = Readable.from((async function * () {
+      yield '['
+      await fetchMeta()
+      while (serverMeta && counter < limit) {
+        try {
+          const entry = await sbp('chelonia/db/getEntry', currentHash)
+          if (entry) {
+            const currentPrefix = prefix
+            prefix = ','
+            counter++
+            yield `${currentPrefix}"${strToB64(
+              JSON.stringify({ serverMeta, message: entry.serialize() })
+            )}"`
+            currentHeight++
+            currentHash = undefined
+            serverMeta = undefined
+          } else {
+            break
+          }
+          // queries with 'keyOps' always return the requested height, whether
+          // or not a keyOp.
+          if (options.keyOps) {
+            // Advance until the next key op
+            while ((await nextKeyOp()) === null);
+          }
+          await fetchMeta()
+        } catch (e) {
+          console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
+          break
         }
       }
-    })
+      yield ']'
+    })(), { encoding: 'UTF-8', objectMode: false })
+
     // $FlowFixMe[prop-missing]
     stream.headers = {
       'shelter-headinfo-head': latestHEADinfo.HEAD,
--- a/backend/database.js
+++ b/backend/database.js
@@ -30,6 +30,10 @@ const options = {
   }
 }
 
+// Segment length for keyop index. Changing this value will require rebuilding
+// this index. The value should be a power of 10 (e.g., 10, 100, 1000, 10000)
+export const KEYOP_SEGMENT_LENGTH = 10_000
+
 // Used by `throwIfFileOutsideDataDir()`.
 const dataFolder = path.resolve(options.fs.dirname)
 
@@ -55,17 +59,28 @@ export default ((sbp('sbp/selectors/register', {
     let currentHeight = height
     let currentHash, serverMeta
     let prefix = ''
+    // `nextKeyOp` is used to advance `currentHeight` when fetching keyOps, and
+    // more complex behaviour than simple 'increment by 1' is needed.
+    // It returns three possible values: `true`, indicating `currentHeight` has
+    // been set to the next `currentHeight`; `false`, indicating the end of
+    // the stream; and `null`, indicating the result is indeterminate and
+    // `nextKeyOp` should be called again (this is because key ops are indexed
+    // based on their height in segments of 10k height entries).
+    // `nextKeyOp` will do the following: if `currentHeight` points to a key op,
+    // it will leave `currentHeight` unchanged; otherwise, if `currentHeight`
+    // does _not_ point to a key op, `currentHeight` will be set to the smallest
+    // height larger than the current `currentHeight` value which is a key op.
     const nextKeyOp = (() => {
-      let index
+      let index: ?string[]
 
       return async () => {
         if (!index) {
-          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00')
+          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % KEYOP_SEGMENT_LENGTH}`))?.split('\x00')
         }
         const value = index?.find((h, i) => {
           if (Number(h) >= currentHeight) {
             // Remove values that no longer are relevant from the index
-            index = ((index: any): any[]).splice(i + 1)
+            index = ((index: any): string[]).slice(i + 1)
             return true
           } else {
             return false
@@ -75,7 +90,11 @@ export default ((sbp('sbp/selectors/register', {
           const newHeight = Number(value)
           currentHeight = newHeight
         } else {
-          currentHeight = currentHeight - currentHeight % 10000 + 10000
+          // We've exhausted the current index; we'll return `null` and advance
+          // height by KEYOP_SEGMENT_LENGTH so that we can read the next index
+          // upon the next invocation (or, if we've reached the end of the
+          // contract, we return `false`).
+          currentHeight = currentHeight - currentHeight % KEYOP_SEGMENT_LENGTH + KEYOP_SEGMENT_LENGTH
           index = undefined
           if (currentHeight > latestHEADinfo.height) {
             return false
@@ -86,6 +105,16 @@ export default ((sbp('sbp/selectors/register', {
         return true
       }
     })()
+    // `fetchMeta` fetches metadata information for entries based on height.
+    // Crucially, it fetches the entry hash (i.e., height -> hash lookup), which
+    // is needed for returning the correct data.
+    // The return value isn't currently used but is left as it may be useful in
+    // the future if this code is refactored. `false` indicates that no metadata
+    // could be found (indicating the end of a contract --or some kind of data
+    // corruption) and that the stream has ended. `true` indicates that the
+    // function executed successfully and the stream continues.
+    // Currently, the return value is not used because we're relying on the
+    // truthiness of `serverMeta` as a subtitute.
     const fetchMeta = async () => {
       if (currentHeight > latestHEADinfo.height) {
         return false
@@ -109,23 +138,29 @@ export default ((sbp('sbp/selectors/register', {
       while (serverMeta && counter < limit) {
         try {
           const entry = await sbp('chelonia/db/getEntry', currentHash)
-          if (entry) {
-            const currentPrefix = prefix
-            prefix = ','
-            counter++
-            yield `${currentPrefix}"${strToB64(
-              JSON.stringify({ serverMeta, message: entry.serialize() })
-            )}"`
-            currentHeight++
-            currentHash = undefined
-            serverMeta = undefined
-          } else {
-            break
-          }
+          // If the entry doesn't exist, we may have reached the end
+          if (!entry) break
+          const currentPrefix = prefix
+          prefix = ','
+          counter++
+          yield `${currentPrefix}"${strToB64(
+            JSON.stringify({ serverMeta, message: entry.serialize() })
+          )}"`
+          // Note for future improvement: implement a generator function for
+          // advancing height, which would make this logic more general by
+          // having just something like `await height.next()` instead of this
+          // avancement here and separate logic for each special case (like the
+          // `if` below for key ops).
+          currentHeight++
+          currentHash = undefined
+          serverMeta = undefined
           // queries with 'keyOps' always return the requested height, whether
           // or not a keyOp.
           if (options.keyOps) {
-            // Advance until the next key op
+            // Advance `currentHeight` until the next key op
+            // `nextKeyOp` relies on the height advancing after an ideration and
+            // the previous currentHeight++ will not result in incorrect
+            // behaviour.
             while ((await nextKeyOp()) === null);
           }
           await fetchMeta()
--- a/backend/database.js
+++ b/backend/database.js
@@ -41,7 +41,7 @@ if (!fs.existsSync(dataFolder)) {
 
 // Streams stored contract log entries since the given entry hash (inclusive!).
 export default ((sbp('sbp/selectors/register', {
-  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number): Promise<*> {
+  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number, options: { keyOps?: boolean } = {}): Promise<*> {
     const limit = Math.min(requestedLimit ?? Number.POSITIVE_INFINITY, process.env.MAX_EVENTS_BATCH_SIZE ?? 500)
     const latestHEADinfo = await sbp('chelonia/db/latestHEADinfo', contractID)
     if (latestHEADinfo === '') {
@@ -52,11 +52,33 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let { hash: currentHash, ...serverMeta } = (
-      await sbp('chelonia/db/getEntryMeta', contractID, height)
-    ) || {}
+    let currentHeight = parseInt(height)
+    let currentHash, serverMeta
     let prefix = ''
     let ended = false
+    const nextKeyOp = (() => {
+      let index
+
+      return async () => {
+        if (!index) {
+          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00')
+        }
+        const value = index?.find((h, i) => {
+          if (Number(h) >= currentHeight) {
+            index = ((index: any): any[]).splice(i + 1)
+            return true
+          } else {
+            return false
+          }
+        })
+        if (value != null) {
+          currentHeight = Number(value)
+        } else {
+          currentHeight = currentHeight - currentHeight % 10000 + 10000
+          index = undefined
+        }
+      }
+    })()
     // NOTE: if this ever stops working you can also try Readable.from():
     // https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options
     const stream = new Readable({
@@ -69,37 +91,62 @@ export default ((sbp('sbp/selectors/register', {
           this.destroy()
           return
         }
-        if (currentHash && counter < limit) {
-          sbp('chelonia/db/getEntry', currentHash).then(async entry => {
-            if (entry) {
-              const currentPrefix = prefix
-              prefix = ','
-              counter++
-              const { hash: newCurrentHash, ...newServerMeta } = (
-                await sbp('chelonia/db/getEntryMeta', contractID, entry.height() + 1)
-              ) || {}
-              this.push(
-                `${currentPrefix}"${strToB64(
-                  JSON.stringify({ serverMeta, message: entry.serialize() })
-                )}"`
-              )
-              currentHash = newCurrentHash
-              serverMeta = newServerMeta
-            } else {
-              this.push(']')
-              this.push(null)
-              ended = true
-            }
-          }).catch(e => {
-            console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
-            this.push(']')
-            this.push(null)
-            ended = true
-          })
-        } else {
+        const end = () => {
           this.push(']')
           this.push(null)
           ended = true
+        }
+        const fetchMeta = async () => {
+          if (currentHeight > latestHEADinfo.height) {
+            end()
+            return false
+          }
+          const { hash: newCurrentHash, ...newServerMeta } = (
+            await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
+          ) || {}
+          if (!newCurrentHash) {
+            end()
+            return false
+          }
+          currentHash = newCurrentHash
+          serverMeta = newServerMeta
+
+          return true
+        }
+        if (counter < limit) {
+          (async () => {
+            try {
+              if (options.keyOps) {
+                while (!serverMeta?.isKeyOp) {
+                  await nextKeyOp()
+                  if (!fetchMeta()) return
+                }
+              } else {
+                if (!fetchMeta()) return
+              }
+              const entry = await sbp('chelonia/db/getEntry', currentHash)
+              if (entry) {
+                const currentPrefix = prefix
+                prefix = ','
+                counter++
+                this.push(
+                `${currentPrefix}"${strToB64(
+                  JSON.stringify({ serverMeta, message: entry.serialize() })
+                )}"`
+                )
+                currentHeight++
+                currentHash = undefined
+                serverMeta = undefined
+              } else {
+                end()
+              }
+            } catch (e) {
+              console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
+              end()
+            }
+          })()
+        } else {
+          end()
         }
       }
     })
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -247,7 +247,7 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {
       limit: Joi.string().regex(POSITIVE_INTEGER_REGEX)
     }),
     query: Joi.object({
-      keyOps: Joi.string()
+      keyOps: Joi.boolean()
     })
   }
 }, async function (request, h) {
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -245,6 +245,9 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {
       contractID: Joi.string().regex(CID_REGEX).required(),
       since: Joi.string().regex(POSITIVE_INTEGER_REGEX).required(),
       limit: Joi.string().regex(POSITIVE_INTEGER_REGEX)
+    }),
+    query: Joi.object({
+      keyOps: Joi.string()
     })
   }
 }, async function (request, h) {
@@ -256,7 +259,7 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {
       return Boom.badRequest()
     }
 
-    const stream = await sbp('backend/db/streamEntriesAfter', contractID, Number(since), limit == null ? undefined : Number(limit))
+    const stream = await sbp('backend/db/streamEntriesAfter', contractID, Number(since), limit == null ? undefined : Number(limit), { keyOps: !!request.query['keyOps'] })
     // "On an HTTP server, make sure to manually close your streams if a request is aborted."
     // From: http://knexjs.org/#Interfaces-Streams
     //       https://github.com/tgriesser/knex/wiki/Manually-Closing-Streams
--- a/backend/server.js
+++ b/backend/server.js
@@ -139,6 +139,11 @@ sbp('sbp/selectors/register', {
       // is needed for the load & store operation.
       await sbp('backend/server/appendToContractIndex', contractID)
     }
+    // If this was a key op, add it to a keyop index. To prevent the index from
+    // growing too large, the index is segmented for every 10000 height values
+    if (cheloniaState.contracts[contractID].previousKeyOp === deserializedHEAD.hash) {
+      await appendToIndexFactory(`_private_keyop_idx_${contractID}_${deserializedHEAD.head.height - deserializedHEAD.head.height % 10000}`)(String(deserializedHEAD.head.height))
+    }
   },
   'backend/server/appendToContractIndex': appendToIndexFactory('_private_cheloniaState_index'),
   'backend/server/broadcastKV': async function (contractID: string, key: string, entry: string) {
@@ -299,6 +304,9 @@ sbp('sbp/selectors/register', {
           if (event) {
             await sbp('chelonia.db/delete', event)
             await sbp('chelonia.db/delete', eventKey)
+          }
+          if (i % 10000 === 0) {
+            await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
           }
         }
         await sbp('chelonia/db/deleteLatestHEADinfo', cid)
--- a/backend/server.js
+++ b/backend/server.js
@@ -9,7 +9,7 @@ import '~/shared/domains/chelonia/persistent-actions.js'
 import { SERVER } from '~/shared/domains/chelonia/presets.js'
 import { multicodes, parseCID } from '~/shared/functions.js'
 import type { SubMessage, UnsubMessage } from '~/shared/pubsub.js'
-import { appendToIndexFactory, initDB, removeFromIndexFactory } from './database.js'
+import { KEYOP_SEGMENT_LENGTH, appendToIndexFactory, initDB, removeFromIndexFactory } from './database.js'
 import { BackendErrorBadData, BackendErrorGone, BackendErrorNotFound } from './errors.js'
 import { SERVER_RUNNING } from './events.js'
 import { PUBSUB_INSTANCE, SERVER_INSTANCE } from './instance-keys.js'
@@ -140,9 +140,10 @@ sbp('sbp/selectors/register', {
       await sbp('backend/server/appendToContractIndex', contractID)
     }
     // If this was a key op, add it to a keyop index. To prevent the index from
-    // growing too large, the index is segmented for every 10000 height values
+    // growing too large, the index is segmented for every KEYOP_SEGMENT_LENGTH
+    // height values
     if (cheloniaState.contracts[contractID].previousKeyOp === deserializedHEAD.hash) {
-      await appendToIndexFactory(`_private_keyop_idx_${contractID}_${deserializedHEAD.head.height - deserializedHEAD.head.height % 10000}`)(String(deserializedHEAD.head.height))
+      await appendToIndexFactory(`_private_keyop_idx_${contractID}_${deserializedHEAD.head.height - deserializedHEAD.head.height % KEYOP_SEGMENT_LENGTH}`)(String(deserializedHEAD.head.height))
     }
   },
   'backend/server/appendToContractIndex': appendToIndexFactory('_private_cheloniaState_index'),
@@ -305,7 +306,7 @@ sbp('sbp/selectors/register', {
             await sbp('chelonia.db/delete', JSON.parse(event).hash)
             await sbp('chelonia.db/delete', eventKey)
           }
-          if (i % 10000 === 0) {
+          if (i % KEYOP_SEGMENT_LENGTH === 0) {
             await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
           }
         }
--- a/backend/server.js
+++ b/backend/server.js
@@ -291,8 +291,8 @@ sbp('sbp/selectors/register', {
       }
       await sbp('chelonia.db/delete', resourcesKey)
 
-      // Next, loop through all the events in the contract and delete them,
-      // starting with the most recent ones.
+      // Next, loop through all the events, except the very first one,
+      // in the contract and delete them, starting with the most recent ones.
       // If the deletion process is interrupted, parts of the contract will
       // still be able to be synced, but won't be to write to it (due to
       // latestHEADinfo not being deleted).
@@ -335,6 +335,8 @@ sbp('sbp/selectors/register', {
       await sbp('chelonia.db/delete', `_private_deletionTokenDgst_${cid}`)
       await removeFromIndexFactory(`_private_resources_${owner}`)(cid)
 
+      // Delete the first event and its associated keys. These were not deleted
+      // in the loop above that deletes events one by one.
       await sbp('chelonia.db/delete', `_private_hidx=${cid}#0`)
       await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_0`)
       await sbp('chelonia.db/set', cid, '')
--- a/backend/server.js
+++ b/backend/server.js
@@ -336,6 +336,7 @@ sbp('sbp/selectors/register', {
       await removeFromIndexFactory(`_private_resources_${owner}`)(cid)
 
       await sbp('chelonia.db/delete', `_private_hidx=${cid}#0`)
+      await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_0`)
       await sbp('chelonia.db/set', cid, '')
 
       await sbp('chelonia.db/delete', `_private_cheloniaState_${cid}`)
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -476,7 +476,7 @@ export class SPMessage {
 
   // `isKeyOp` is used to filter out non-key operations for providing an
   // abbreviated chain fo snapshot validation
-  isKeyOp (): string {
+  isKeyOp (): boolean {
     return !!(
       keyOps.includes(this.opType()) ||
       // $FlowFixMe[prop-missing]
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -325,6 +325,8 @@ export class SPMessage {
       get contractID () {
         return result.head?.contractID ?? result.hash
       },
+      // `description` is not a getter to prevent the value from being copied
+      // if the object is cloned or serialized
       description (): string {
         const type = this.head.op
         return `<op_${type}|${this.hash} of ${this.contractID}>`
@@ -461,7 +463,7 @@ export class SPMessage {
 
   hash (): string { return this._mapping.key }
 
-  previousKeyOp (): string { return this._head.previousKeyOp }
+  previousKeyOp (): ?string { return this._head.previousKeyOp }
 
   height (): number { return this._head.height }
 
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -235,6 +235,7 @@ export class SPMessage {
     {
       contractID,
       previousHEAD = null,
+      previousKeyOp = null,
       // Height will be automatically set to the correct value when sending
       // The reason to set it to Number.MAX_SAFE_INTEGER is so that we can
       // temporarily process outgoing messages with signature validation
@@ -245,6 +246,7 @@ export class SPMessage {
     }: {
       contractID: string | null,
       previousHEAD?: ?string,
+      previousKeyOp?: ?string,
       height?: ?number,
       op: SPOpRaw,
       manifest: string,
@@ -253,6 +255,7 @@ export class SPMessage {
     const head = {
       version: '1.0.0',
       previousHEAD,
+      previousKeyOp,
       height,
       contractID,
       op: op[0],
@@ -458,6 +461,8 @@ export class SPMessage {
 
   hash (): string { return this._mapping.key }
 
+  previousKeyOp (): string { return this._head.previousKeyOp }
+
   height (): number { return this._head.height }
 
   id (): string {
@@ -467,6 +472,18 @@ export class SPMessage {
 
   direction (): 'incoming' | 'outgoing' {
     return this._direction
+  }
+
+  // `isKeyOp` is used to filter out non-key operations for providing an
+  // abbreviated chain fo snapshot validation
+  isKeyOp (): string {
+    return !!(
+      keyOps.includes(this.opType()) ||
+      // $FlowFixMe[prop-missing]
+      (this.opType() === SPMessage.OP_ATOMIC && Array.isArray(this.opValue()) && this.opValue().some(([opT]) => {
+        return keyOps.includes(opT)
+      }))
+    )
   }
 
   // $FlowFixMe[unsupported-syntax]
@@ -521,3 +538,6 @@ function messageToParams (head: Object, message: SignedData<SPOpValue>): SPMsgPa
     signedMessageData: message
   }
 }
+
+// Operations that affect valid keys
+const keyOps = [SPMessage.OP_CONTRACT, SPMessage.OP_KEY_ADD, SPMessage.OP_KEY_DEL, SPMessage.OP_KEY_UPDATE]
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1318,7 +1318,6 @@ export default (sbp('sbp/selectors/register', {
 
     let msg = SPMessage.createV1_0({
       contractID,
-      originatingContractID,
       op: [
         SPMessage.OP_KEY_SHARE,
         params.signingKeyId
--- a/shared/domains/chelonia/db.js
+++ b/shared/domains/chelonia/db.js
@@ -59,6 +59,12 @@ export const prefixHandlers: Object = {
 sbp('sbp/selectors/unsafe', ['chelonia.db/get', 'chelonia.db/set', 'chelonia.db/delete'])
 // NOTE: MAKE SURE TO CALL 'sbp/selectors/lock' after overwriting them!
 
+// When using a lightweight client, the client doesn't keep a copy of messages
+// in the DB. Therefore, `chelonia.db/*` selectors are mostly turned into no-ops.
+// The `chelonia.db/get` selector is slightly more complex than a no-op, because
+// Chelonia relies on being able to find the current contract head. To overcome
+// this, if a head is requested, 'chelonia.db/get' returns information from
+// the Chelonia contract state.
 const dbPrimitiveSelectors = process.env.LIGHTWEIGHT_CLIENT === 'true'
   ? {
       'chelonia.db/get': function (key: string): Promise<HEADInfo | Buffer | string | void> {
@@ -153,7 +159,7 @@ export default ((sbp('sbp/selectors/register', {
           console.warn(`[chelonia.db] bad previousHEAD: ${entryPreviousHEAD}! Expected: ${contractHEAD} for contractID: ${contractID}`)
           throw new ChelErrorDBBadPreviousHEAD(`bad previousHEAD: ${entryPreviousHEAD}. Expected ${contractHEAD} for contractID: ${contractID}`)
         } else if (entryPreviousKeyOp !== contractPreviousKeyOp) {
-          console.warn(`[chelonia.db] bad previousKeyOp: ${entryPreviousKeyOp}! Expected: ${contractPreviousKeyOp} for contractID: ${contractID}`)
+          console.error(`[chelonia.db] bad previousKeyOp: ${entryPreviousKeyOp}! Expected: ${contractPreviousKeyOp} for contractID: ${contractID}`)
           throw new ChelErrorDBBadPreviousHEAD(`bad previousKeyOp: ${entryPreviousKeyOp}. Expected ${contractPreviousKeyOp} for contractID: ${contractID}`)
         } else if (!Number.isSafeInteger(entryHeight) || entryHeight !== (contractHeight + 1)) {
           console.error(`[chelonia.db] bad height: ${entryHeight}! Expected: ${contractHeight + 1} for contractID: ${contractID}`)
@@ -176,8 +182,14 @@ export default ((sbp('sbp/selectors/register', {
       }))
       console.debug(`[chelonia.db] HEAD for ${contractID} updated to:`, entry.hash())
       await sbp('chelonia/db/setEntryMeta', contractID, entryHeight, {
+        // The hash is used for reverse lookups (height to CID)
         hash: entry.hash(),
+        // The date isn't currently used, but will be used for filtering messages
         date: new Date().toISOString(),
+        // isKeyOp is used for filtering messages (the actual filtering is
+        // done more efficiently a separate index key, but `isKeyOp` allows
+        // us to bootstrap this process without having to load the full message)
+        // The separate index key bears the prefix `_private_keyop_idx_`.
         ...(entry.isKeyOp() && { isKeyOp: true })
       })
       return entry.hash()
--- a/shared/domains/chelonia/db.js
+++ b/shared/domains/chelonia/db.js
@@ -177,7 +177,8 @@ export default ((sbp('sbp/selectors/register', {
       console.debug(`[chelonia.db] HEAD for ${contractID} updated to:`, entry.hash())
       await sbp('chelonia/db/setEntryMeta', contractID, entryHeight, {
         hash: entry.hash(),
-        date: new Date().toISOString()
+        date: new Date().toISOString(),
+        ...(entry.isKeyOp() && { isKeyOp: true })
       })
       return entry.hash()
     } catch (e) {
--- a/shared/domains/chelonia/db.js
+++ b/shared/domains/chelonia/db.js
@@ -68,7 +68,8 @@ const dbPrimitiveSelectors = process.env.LIGHTWEIGHT_CLIENT === 'true'
         const value = (state?.HEAD
           ? JSON.stringify({
             HEAD: state.HEAD,
-            height: state.height
+            height: state.height,
+            previousKeyOp: state.previousKeyOp
           })
           : undefined)
         return Promise.resolve(value)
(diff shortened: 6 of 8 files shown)


backend/database.js
```diff
@@ -52,11 +52,48 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let { hash: currentHash, ...serverMeta } = (
-      await sbp('chelonia/db/getEntryMeta', contractID, height)
-    ) || {}
+    let currentHeight = parseInt(height)
```
taoeffect: o4-mini-high:

### 1) We don’t need `parseInt` on a `number` and it hides bugs

In **backend/database.js** you wrote:

```js
// (we know `height` is declared as a number in the signature)
let currentHeight = parseInt(height)
```

If `height` ever isn’t a clean integer this will silently floor it. Since `height` is already a number, just do:

```js
let currentHeight = height
```

or, if it really might come in as a string:

```js
let currentHeight = Number(height)
if (!Number.isSafeInteger(currentHeight)) {
  throw Boom.badRequest(`since must be an integer, got ${height}`)
}
```

backend/routes.js
```diff
@@ -256,7 +259,7 @@ route.GET('/eventsAfter/{contractID}/{since}/{limit?}', {
       return Boom.badRequest()
     }
 
-    const stream = await sbp('backend/db/streamEntriesAfter', contractID, Number(since), limit == null ? undefined : Number(limit))
+    const stream = await sbp('backend/db/streamEntriesAfter', contractID, Number(since), limit == null ? undefined : Number(limit), { keyOps: !!request.query['keyOps'] })
```
taoeffect, line 262: o4-mini-high:


### 3) Query‐param parsing for `?keyOps` is too naive

In **backend/routes.js**:

```js
query: Joi.object({
  keyOps: Joi.string()
})
```

and then

```js
{ keyOps: !!request.query['keyOps'] }
```

That means `?keyOps=false` still comes through as `true`.  Instead I’d suggest:

```js
// 1) Validate it as a boolean and give it a default
query: Joi.object({
  keyOps: Joi.boolean().default(false)
}),

// 2) Later:
const { keyOps } = request.query
const stream = await sbp(
  'backend/db/streamEntriesAfter',
  contractID,
  sinceInt,
  limitInt,
  { keyOps }
)
```

That way you get a real boolean, and “false” stays false.

backend/server.js
```diff
@@ -139,6 +139,11 @@ sbp('sbp/selectors/register', {
       // is needed for the load & store operation.
       await sbp('backend/server/appendToContractIndex', contractID)
     }
+    // If this was a key op, add it to a keyop index. To prevent the index from
+    // growing too large, the index is segmented for every 10000 height values
+    if (cheloniaState.contracts[contractID].previousKeyOp === deserializedHEAD.hash) {
+      await appendToIndexFactory(`_private_keyop_idx_${contractID}_${deserializedHEAD.head.height - deserializedHEAD.head.height % 10000}`)(String(deserializedHEAD.head.height))
+    }
```
taoeffect: o4-mini-high makes an interesting comment here, insisting that `.head` is a method, not a property.

Although it's mistaken, it does bring up a good point regarding the implementation of `deserializeHEAD`, and why some of the values returned by it are properties/getters and others (like `description ()`) are methods.

Could you update the `SPMessage.js` file to add a comment explaining this? On the call you mentioned it's so that `deserializedHEAD` could be serialized using `JSON.stringify` while including most properties except `description ()`. I'd prefer they all be methods for consistency, but that's a good reason to keep them as-is.

Please consider what would be more developer friendly... it might be a more developer-friendly API to have them all be straightforward methods (since `get head` _is_ a method anyway), have another method like `toJSON()` for serialization. OTOH, the implementation of `toJSON()` might then bug ugly. I'm honestly not sure, just thinking aloud here. At the very least a comment would be useful.

shared/domains/chelonia/chelonia.js
```diff
@@ -1318,7 +1318,6 @@ export default (sbp('sbp/selectors/register', {
 
     let msg = SPMessage.createV1_0({
       contractID,
-      originatingContractID,
```
taoeffect, line 1345: Why is this line being deleted? Doesn't seem right...

shared/domains/chelonia/utils.js
```diff
@@ -547,7 +547,7 @@ export const recreateEvent = (entry: SPMessage, state: Object, contractsState: O
   const newOp = [opT, newRawOpV]
 
   entry = SPMessage.cloneWith(
-    head, newOp, { previousHEAD, height: previousHeight + 1 }
+    head, newOp, { previousKeyOp, previousHEAD, height: previousHeight + 1 }
```
taoeffect, line 550: I'm not seeing where in this PR you are setting the `SPMessage`'s `_head.previousKeyOp`... Above in `chelonia.js` you do not do it, and it needs to be done in `createV1_0` (which has been updated to receive it, but this is never used)

backend/server.js
```diff
@@ -300,6 +305,9 @@ sbp('sbp/selectors/register', {
             await sbp('chelonia.db/delete', JSON.parse(event).hash)
             await sbp('chelonia.db/delete', eventKey)
           }
+          if (i % 10000 === 0) {
+            await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
+          }
```
taoeffect: o4-mini-high:

### 6) Deleting your segmented keyOp indices never removes the “0” bucket

In **backend/server.js** in your contract‐deletion loop you do:

```js
for (let i = latestHEADinfo.height; i > 0; i--) {
  // …
  if (i % 10000 === 0) {
    await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
  }
}
```

That will delete buckets at `10000`, `20000`, etc., but never at `0` (because `i>0` never reaches `i===0`).  You end up leaving `_private_keyop_idx_<cid>_0` around.  Instead, once you know the final height, delete every segment in one go:

```js
const last = latestHEADinfo.height
for (let segment = 0; segment <= last; segment += 10000) {
  await sbp(
    'chelonia.db/delete',
    `_private_keyop_idx_${cid}_${segment}`
  )
}
```

shared/domains/chelonia/SPMessage.js
```diff
@@ -458,6 +461,8 @@ export class SPMessage {
 
   hash (): string { return this._mapping.key }
 
+  previousKeyOp (): string { return this._head.previousKeyOp }
```
taoeffect: o4-mini-high points out:

> In your new `previousKeyOp(): string { … }` you probably want to return `?string` since it may be `null` on very old messages.  

I think this is correct as `OP_CONTRACT` would have a `null` previousOp

shared/domains/chelonia/chelonia.js
corrideat: This is no longer used. Search the project for `originatingContractID`, and you'll notice that this isn't used as an SPMessage property.

shared/domains/chelonia/utils.js
corrideat: It's set in utils.js:

```js
  entry = SPMessage.cloneWith(
    head, newOp, { previousKeyOp, previousHEAD, height: previousHeight + 1 }
  )
```

It's not set in `chelonia.js` for the same reason the `height` isn't set: we need to fetch the latest head info to find out.

taoeffect (changes requested):
(Preliminary review: not done yet!)

Great work on this PR @corrideat! I already gave you `o4-mini-high`'s review, and as I gotta run here are the unfiltered reviews from the two Gemini 2.5 variants:

## Gemini 2.5 Pro Preview

<details>
Okay, I've reviewed the provided changes and the context code. Here are my findings:

### Bugs

#### 1. Incorrect Deletion of Key Operation Indices

**File:** `backend/server.js`

**Lines:**
```javascript
          if (i % 10000 === 0) {
            await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
          }
```

**Issue:**
The logic for deleting the key operation indices (`_private_keyop_idx_*`) during contract deletion is flawed. The index keys are segmented based on the *start* of a 10000-height block (e.g., `_private_keyop_idx_${cid}_0`, `_private_keyop_idx_${cid}_10000`, `_private_keyop_idx_${cid}_20000`, etc.). The current code attempts to delete an index key only when the loop variable `i` (representing the height being deleted) is an exact multiple of 10000. This will miss deleting most index segment keys. For example, if a contract has a height of 15000, this loop will only attempt to delete `_private_keyop_idx_${cid}_10000` and `_private_keyop_idx_${cid}_0`, potentially leaving `_private_keyop_idx_${cid}_10000` behind if the contract started exactly at height 0 or 10000. It should delete all index segments that the contract might have spanned.

**Suggestion:**
Calculate the index segment keys based on the contract's height range *before* the loop and delete them explicitly.

```javascript
// backend/server.js
// Inside 'backend/deleteContract' function

      // ... (previous deletion logic) ...

      // Next, loop through all the events in the contract and delete them,
      // starting with the most recent ones.
      // If the deletion process is interrupted, parts of the contract will
      // still be able to be synced, but won't be to write to it (due to
      // latestHEADinfo not being deleted).
      const latestHEADinfo = await sbp('chelonia/db/latestHEADinfo', cid)
      if (latestHEADinfo) {
+       // --- Start of suggested change ---
+       // Calculate and delete all possible keyop index segments for this contract
+       const maxHeight = latestHEADinfo.height
+       let currentSegmentBase = maxHeight - (maxHeight % 10000)
+       while (currentSegmentBase >= 0) {
+         await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${currentSegmentBase}`)
+         // Ensure we don't loop infinitely if somehow currentSegmentBase is 0
+         if (currentSegmentBase === 0) break
+         currentSegmentBase -= 10000
+       }
+       // --- End of suggested change ---
+
        for (let i = latestHEADinfo.height; i > 0; i--) {
          const eventKey = `_private_hidx=${cid}#${i}`
          const event = await sbp('chelonia.db/get', eventKey)
          if (event) {
            await sbp('chelonia.db/delete', JSON.parse(event).hash)
            await sbp('chelonia.db/delete', eventKey)
          }
-         if (i % 10000 === 0) {
-           await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
-         }
        }
        await sbp('chelonia/db/deleteLatestHEADinfo', cid)
      }
```

### Improvements

#### 1. Unnecessary Type Cast in `streamEntriesAfter`

**File:** `backend/database.js`

**Lines:**
```javascript
        const value = index?.find((h, i) => {
          if (Number(h) >= currentHeight) {
            index = ((index: any): any[]).splice(i + 1) // <--- This line
            return true
          } else {
            return false
          }
        })
```

**Issue:**
The type cast `((index: any): any[])` is used to suppress a potential Flow error when assigning the result of `splice` back to `index`. However, `index` should already be typed as `?string[]` (or can be declared as such). The `splice` method on an array returns an array of the same type (`string[]` in this case). Assigning `string[]` back to a variable typed as `?string[]` is valid in Flow. The cast hides the type information and is unnecessary if `index` is correctly typed.

**Suggestion:**
Declare `index` with the correct Flow type and remove the cast.

```javascript
// backend/database.js
// Inside 'backend/db/streamEntriesAfter' function

    const nextKeyOp = (() => {
-     let index
+     // Declare index with Flow type
+     let index: ?Array<string>

      return async () => {
        if (!index) {
          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00')
        }
        const value = index?.find((h, i) => {
          if (Number(h) >= currentHeight) {
-           index = ((index: any): any[]).splice(i + 1)
+           // Remove unnecessary cast - index is already ?Array<string>
+           index = index?.splice(i + 1)
            return true
          } else {
            return false
          }
        })
```
*Self-correction:* Using `index?.splice(i + 1)` is safer since `index` could be null/undefined in edge cases, although the surrounding logic makes it unlikely *inside* the `find` callback. However, directly assigning `index.splice(i + 1)` should also work if Flow infers `index` is non-null within the `find` after the `index?.find` check. Let's stick to the simpler removal of the cast assuming Flow handles `index` being non-null within the callback:
```javascript
// backend/database.js (Alternative refinement)
        const value = index?.find((h, i) => {
          if (Number(h) >= currentHeight) {
            // Remove unnecessary cast & optional chaining if Flow correctly infers non-null
            index = index.splice(i + 1) 
            return true
          } else {
            return false
          }
        })
```
Choose the alternative refinement if you are confident Flow correctly infers `index` is non-null within the `find` callback scope established by `index?.find`. If unsure, the first suggestion with `index?.splice(i + 1)` is slightly safer, although potentially assigning `undefined` back to `index` if `index` somehow became null between the `find` check and the splice execution (highly unlikely). The most robust explicit way without optional chaining inside would be:
```javascript
// backend/database.js (Most robust explicit refinement)
        const value = index?.find((h, i) => {
          if (Number(h) >= currentHeight) {
            if (index) { // Explicit non-null check
              index = index.splice(i + 1)
            }
            return true
          } else {
            return false
          }
        })
```
Given the context, the "Alternative refinement" (`index = index.splice(i + 1)`) is likely the intended and cleanest solution, relying on Flow's inference.

---

The rest of the changes related to introducing `previousKeyOp`, `isKeyOp`, indexing key ops, and streaming only key ops seem logically sound and consistently implemented across the different files (`SPMessage.js`, `db.js`, `internals.js`, `utils.js`, `server.js`, `routes.js`). The core concept of maintaining and checking the `previousKeyOp` hash alongside the `previousHEAD` hash in `chelonia/private/db/addEntry` is crucial for the integrity of this new feature, and it appears to be correctly handled. StandardJS and FlowType conventions seem to be followed otherwise.

</details>

## Gemini 2.5 Pro Exp

<details>
Okay, I've reviewed the provided diff and context. Here are my findings and suggestions:

### Overall Impression

The changes introduce a mechanism to track and efficiently query "Key Operations" (KeyOps) within a contract's event history. This involves adding a `previousKeyOp` pointer to message headers and contract state, creating a secondary chain linking only these critical operations. A new backend index (`_private_keyop_idx_...`) and a streaming option (`keyOps: true`) are added to retrieve this shorter chain. The implementation seems generally sound but requires a few adjustments for correctness, robustness, and handling existing data.

---

### Issues and Suggestions

1.  **Potential Bug in KeyOp Index Lookup (`backend/database.js`)**

    In `backend/database.js`, the `nextKeyOp` function retrieves an index string, splits it, and searches for the next relevant height. If the index string retrieved from the database is empty (`''`) or contains empty segments (e.g., `"10\x00\x0025"`), `split('\x00')` can produce empty strings (`['']` or `['10', '', '25']`). `Number('')` evaluates to `0`. If `currentHeight` is also 0, the `find` condition `Number(h) >= currentHeight` might incorrectly match the empty string. It's safer to filter empty strings and explicitly check for `NaN`.

    ```diff
    --- a/backend/database.js
    +++ b/backend/database.js
    @@ -62,9 +62,13 @@ export default ((sbp('sbp/selectors/register', {
 
       return async () => {
         if (!index) {
-          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00')
+          // Filter out potential empty strings from split
+          index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00').filter(h => h !== '')
+          if (!index || index.length === 0) { // Handle empty or non-existent index segment
+            currentHeight = currentHeight - currentHeight % 10000 + 10000
+            return // Skip find logic if index is empty
+          }
         }
         const value = index?.find((h, i) => {
-          if (Number(h) >= currentHeight) {
+          const heightNum = Number(h)
+          // Add NaN check and use the parsed number
+          if (!isNaN(heightNum) && heightNum >= currentHeight) {
             index = ((index: any): any[]).splice(i + 1)
             return true
           } else {

    ```

2.  **Incorrect Condition for KeyOp Indexing (`backend/server.js`)**

    In `backend/server.js`, inside `chelonia/private/processEvent`, the code adds an entry to the KeyOp index if `cheloniaState.contracts[contractID].previousKeyOp === deserializedHEAD.hash`. This seems incorrect. The state's `previousKeyOp` points to the *previous* key op, not the current one being processed. The intention is likely to index the current entry *if it is a key op*. The check should use the `isKeyOp()` method of the message/HEAD object.

    ```diff
    --- a/backend/server.js
    +++ b/backend/server.js
    @@ -140,8 +140,8 @@ sbp('sbp/selectors/register', {
        await sbp('backend/server/appendToContractIndex', contractID)
      }
      // If this was a key op, add it to a keyop index. To prevent the index from
      // growing too large, the index is segmented for every 10000 height values
    - if (cheloniaState.contracts[contractID].previousKeyOp === deserializedHEAD.hash) {
    + if (deserializedHEAD.isKeyOp()) { // Check if the current entry is a key op
        await appendToIndexFactory(`_private_keyop_idx_${contractID}_${deserializedHEAD.head.height - deserializedHEAD.head.height % 10000}`)(String(deserializedHEAD.head.height))
      }
    },
    'backend/server/appendToContractIndex': appendToIndexFactory('_private_cheloniaState_index'),

    ```

3.  **Incomplete Deletion of KeyOp Index (`backend/server.js`)**

    In `backend/deleteContract`, the KeyOp index segments (`_private_keyop_idx_...`) are deleted inside the loop iterating through individual events, using an `i % 10000 === 0` check. This is inefficient and might miss deleting segments, particularly the highest segment if the contract height isn't a multiple of 10000, and the segment starting at height 0 if the loop condition `i > 0` prevents `i=0` from being checked. It's better to calculate the range of segments based on the height and delete them directly.

    ```diff
    --- a/backend/server.js
    +++ b/backend/server.js
    @@ -292,20 +292,20 @@ sbp('sbp/selectors/register', {
       // still be able to be synced, but won't be to write to it (due to
       // latestHEADinfo not being deleted).
       const latestHEADinfo = await sbp('chelonia/db/latestHEADinfo', cid)
       if (latestHEADinfo) {
    +    // Delete all key op index segments for this contract
    +    const highestSegmentStart = latestHEADinfo.height - latestHEADinfo.height % 10000
    +    for (let segmentStart = highestSegmentStart; segmentStart >= 0; segmentStart -= 10000) {
    +      await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${segmentStart}`)
    +    }
    +
    +    // Loop through and delete individual events/metadata
         for (let i = latestHEADinfo.height; i > 0; i--) {
           const eventKey = `_private_hidx=${cid}#${i}`
           const event = await sbp('chelonia.db/get', eventKey)
           if (event) {
             await sbp('chelonia.db/delete', JSON.parse(event).hash)
             await sbp('chelonia.db/delete', eventKey)
           }
    -      if (i % 10000 === 0) {
    -        await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_${i}`)
    -      }
         }
         await sbp('chelonia/db/deleteLatestHEADinfo', cid)
       }

    ```

4.  **Handling Existing Contracts (`shared/domains/chelonia/db.js`)**

    When this code is deployed, existing contracts won't have a `previousKeyOp` value stored in their `HEADInfo` (it will be `undefined`). The validation logic `entryPreviousKeyOp !== contractPreviousKeyOp` in `chelonia/private/db/addEntry` will likely fail incorrectly when the first new entry is processed for such a contract. The validation should be skipped if the stored `contractPreviousKeyOp` is `undefined` to allow the chain to build correctly starting from the first KeyOp processed post-upgrade.

    ```diff
    --- a/shared/domains/chelonia/db.js
    +++ b/shared/domains/chelonia/db.js
    @@ -150,7 +151,8 @@ sbp('sbp/selectors/register', {
          const { HEAD: contractHEAD, previousKeyOp: contractPreviousKeyOp, height: contractHeight } = HEADinfo
           if (entryPreviousHEAD !== contractHEAD) {
             console.warn(`[chelonia.db] bad previousHEAD: ${entryPreviousHEAD}! Expected: ${contractHEAD} for contractID: ${contractID}`)
             throw new ChelErrorDBBadPreviousHEAD(`bad previousHEAD: ${entryPreviousHEAD}. Expected ${contractHEAD} for contractID: ${contractID}`)
    -    } else if (entryPreviousKeyOp !== contractPreviousKeyOp) {
    +    // Skip previousKeyOp check if the stored value is undefined (handles transition for existing contracts)
    +    } else if (contractPreviousKeyOp !== undefined && entryPreviousKeyOp !== contractPreviousKeyOp) {
             console.warn(`[chelonia.db] bad previousKeyOp: ${entryPreviousKeyOp}! Expected: ${contractPreviousKeyOp} for contractID: ${contractID}`)
             throw new ChelErrorDBBadPreviousHEAD(`bad previousKeyOp: ${entryPreviousKeyOp}. Expected ${contractPreviousKeyOp} for contractID: ${contractID}`)
           } else if (!Number.isSafeInteger(entryHeight) || entryHeight !== (contractHeight + 1)) {

    ```

5.  **Type Safety (`backend/database.js`)**

    The cast `(index: any): any[]` in `nextKeyOp` suppresses Flow errors but bypasses type checking. While potentially necessary due to the dynamic nature of database returns and `split`, it's a point of potential fragility. If `sbp('chelonia.db/get', ...)` unexpectedly returned something other than a string or null/undefined, or if `split` behaved unexpectedly, this could lead to runtime errors. Acknowledging this as a workaround is important. Ensure robust handling around the `split` and `find` calls (as suggested in point 1) to mitigate this.

6.  **Flow Type Definition for `HEADInfo`**

    The structure stored for the log head (e.g., `_private_loghead_${contractID}`) now includes `previousKeyOp`. If there's an explicit FlowType definition for this structure (often named `HEADInfo` or similar), it needs to be updated:

    ```flow
    // Example: in a hypothetical types.js
    declare type HEADInfo = {
      HEAD: string,
      height: number,
      previousKeyOp: ?string // Ensure this field is added and optional
    };
    ```
    Make sure this definition is updated wherever it's used.

---

These changes should improve the robustness and correctness of the KeyOp feature implementation. Remember to update the FlowType definition for `HEADInfo` if applicable. Let me know if you have any other sections you'd like me to review.
</details>

See rest of comments/replies as well.

shared/domains/chelonia/chelonia.js
taoeffect: Are you sure? It does seem to be used when create a `'chelonia/out/keyRequest'` as part of `'gi.actions/group/join'`as well as with `'chelonia/out/actionEncrypted'` in `utils.js` as part of `encryptedAction`.

It's unclear to me though as `'chelonia/out/actionEncrypted'` doesn't seem to use it... however if it doesn't use it then the other code should be cleaned up too.

shared/domains/chelonia/db.js
```diff
@@ -68,7 +68,8 @@ const dbPrimitiveSelectors = process.env.LIGHTWEIGHT_CLIENT === 'true'
         const value = (state?.HEAD
           ? JSON.stringify({
             HEAD: state.HEAD,
-            height: state.height
+            height: state.height,
+            previousKeyOp: state.previousKeyOp
```
taoeffect, line 78: I am clearly starting to forget some of the codebase because I do not understand what this `'chelonia.db/get'` selector is doing or why.

It's taking in a `key` and treating it as a "log head"... why? And if it's not a log head it immediately returns nothing.

I understand that this selector is also overwritten in both the frontend and the backend, but why is it written the way it's written?

If you remember, could you add a comment above it to explain?

shared/domains/chelonia/db.js
```diff
@@ -147,10 +148,13 @@ export default ((sbp('sbp/selectors/register', {
...
           throw new Error(`No latest HEAD for ${contractID} when attempting to process entry with previous HEAD ${entryPreviousHEAD} at height ${entryHeight}`)
         }
-        const { HEAD: contractHEAD, height: contractHeight } = HEADinfo
+        const { HEAD: contractHEAD, previousKeyOp: contractPreviousKeyOp, height: contractHeight } = HEADinfo
         if (entryPreviousHEAD !== contractHEAD) {
           console.warn(`[chelonia.db] bad previousHEAD: ${entryPreviousHEAD}! Expected: ${contractHEAD} for contractID: ${contractID}`)
           throw new ChelErrorDBBadPreviousHEAD(`bad previousHEAD: ${entryPreviousHEAD}. Expected ${contractHEAD} for contractID: ${contractID}`)
+        } else if (entryPreviousKeyOp !== contractPreviousKeyOp) {
+          console.warn(`[chelonia.db] bad previousKeyOp: ${entryPreviousKeyOp}! Expected: ${contractPreviousKeyOp} for contractID: ${contractID}`)
+          throw new ChelErrorDBBadPreviousHEAD(`bad previousKeyOp: ${entryPreviousKeyOp}. Expected ${contractPreviousKeyOp} for contractID: ${contractID}`)
```
taoeffect: I realize that some of these say `console.warn` while others say `console.error` but there's a method to the madness.

The first check for `bad previousHEAD` was the only `console.warn` because it would actually occur during development and testing fairly frequently when messages would arrive out of order. The rest are more serious errors that are less likely to happen and that's why they're marked as `console.error`.

If the first check passes (correct previousHEAD), then all subsequent checks should pass too, so could you make this a `console.error` instead?

shared/domains/chelonia/db.js
```diff
@@ -165,11 +169,16 @@ export default ((sbp('sbp/selectors/register', {
...
+        HEAD: entry.hash(),
+        previousKeyOp: entry.isKeyOp() ? entry.hash() : entry.previousKeyOp(),
+        height: entry.height()
+      }))
       console.debug(`[chelonia.db] HEAD for ${contractID} updated to:`, entry.hash())
       await sbp('chelonia/db/setEntryMeta', contractID, entryHeight, {
         hash: entry.hash(),
-        date: new Date().toISOString()
+        date: new Date().toISOString(),
+        ...(entry.isKeyOp() && { isKeyOp: true })
```
taoeffect, line 193: Could you add a comment above line 178 for what each of these fields is used for?

I.e. hash is used to map the height to the hash in order to make forward streaming possible, date is used to map the height to the date in order to make search easier, and `isKeyOp` is used for...?

backend/server.js
```diff
@@ -328,6 +336,7 @@ sbp('sbp/selectors/register', {
       await removeFromIndexFactory(`_private_resources_${owner}`)(cid)
 
       await sbp('chelonia.db/delete', `_private_hidx=${cid}#0`)
+      await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_0`)
```
taoeffect, line 342: Could you add a comment here to explain why this line is here when the code above supposedly deletes these? (i.e. the loop `i > 0` issue might not be obvious to the reader)

backend/server.js
taoeffect: Just a reminder 👆

backend/database.js
```diff
@@ -41,7 +41,7 @@ if (!fs.existsSync(dataFolder)) {
 
 // Streams stored contract log entries since the given entry hash (inclusive!).
 export default ((sbp('sbp/selectors/register', {
-  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number): Promise<*> {
+  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number, options: { keyOps?: boolean } = {}): Promise<*> {
```
taoeffect, line 48: The changes in this file are a bit more complicated and I'm running out of time for today so I enlisted the help of some LLMs. Could you please look over their feedback and also let me know your opinion of how each of them did, and which did the best?

## `o4-mini-high`

<details>
Here are a few spots in your new `streamEntriesAfter` that I would flag as either potentially buggy, confusing, or just ripe for simplification.  I’ve quoted the lines in question and then shown you concrete alternative code.

---

### 1) Shadowing the outer `options` and awkward default parameter

In your diff you have:

```js
// ❌ this “options” here hides the file‑level `const options = { fs, sqlite }`
'backend/db/streamEntriesAfter': async function (
  contractID: string,
  height: number,
  requestedLimit: ?number,
  options: { keyOps?: boolean } = {}
): Promise<*> {
  …
  if (options.keyOps) { … }
```

That can get pretty confusing.  I’d recommend

  1. renaming the param to `{ keyOps }`
  2. giving it a default
  3. tightening the return type from `Promise<*>` to `Promise<Readable>`

For example:

```js
'backend/db/streamEntriesAfter': async function (
  contractID: string,
  height: number,
  requestedLimit: ?number,
  { keyOps = false }: { keyOps?: boolean } = {}
): Promise<Readable> {
  const limit = Math.min(
    requestedLimit ?? Infinity,
    Number(process.env.MAX_EVENTS_BATCH_SIZE) || 500
  )
  …
  if (keyOps) {
    …
  }
}
```

---

### 2) You never do an initial `fetchMeta()` in the “keyOps” branch

Right now your code does:

```js
if (options.keyOps) {
  while (!serverMeta?.isKeyOp) {
    await nextKeyOp()
    if (!await fetchMeta()) return end()
  }
} else {
  if (!await fetchMeta()) return end()
}
```

That means if I call with `keyOps=true` and the very first `height` **is** a keyOp, you will skip straight to `nextKeyOp()` and then call `fetchMeta()` on the same height anyway.  You’d be better off fetching **once** and then looping:

```diff
- if (keyOps) {
-   while (!serverMeta?.isKeyOp) {
-     await nextKeyOp()
-     if (!await fetchMeta()) return end()
-   }
- } else {
-   if (!await fetchMeta()) return end()
- }
+ // do at least one fetchMeta
+ if (!await fetchMeta()) return end()
+ if (keyOps) {
+   // skip forward until we land on a keyOp
+   while (!serverMeta.isKeyOp) {
+     await nextKeyOp()
+     if (!await fetchMeta()) return end()
+   }
+ }
```

---

### 3) Prefer `slice` over mutation with `splice`

Inside your `nextKeyOp` you wrote:

```js
const value = index?.find((h, i) => {
  if (Number(h) >= currentHeight) {
    // splice both *mutates* and returns the removed elements
    index = ((index: any): any[]).splice(i + 1)
    return true
  }
})
```

That both mutates the old `index` array and reassigns your variable.  It’s more predictable to just:

```js
let index: ?string[] = null

…

const value = index.find((h, i) => Number(h) >= currentHeight)
if (value != null) {
  // get everything *after* the one we just used
  index = index.slice(i + 1)
  currentHeight = Number(value)
}
```

This way you never mutate an array that you’re still holding a reference to.

---

### 4) Cast headers to strings and tighten the Flow type

You do:

```js
// $FlowFixMe[prop-missing]
stream.headers = {
  'shelter-headinfo-head': latestHEADinfo.HEAD,
  'shelter-headinfo-height': latestHEADinfo.height
}
```

– HTTP headers always end up as strings, so you should

  • do `String(latestHEADinfo.height)`  
  • give your stream a more precise Flow‐cast instead of a `$FlowFixMe`

For instance:

```js
const stream: Readable & { headers: { [string]: string } } = new Readable({ … })

stream.headers = {
  'shelter-headinfo-head': String(latestHEADinfo.HEAD),
  'shelter-headinfo-height': String(latestHEADinfo.height)
}
```

---

### 5) Consider using an async generator + `Readable.from` for clarity

Your inner `read()` with nested `if/else`, self‑invoking async functions, `ended` flags, etc. is quite hard to follow.  Node.js ≥ 12 supports

```js
const stream = Readable.from(async function* () {
  yield '['
  let i = 0
  while (i < limit) {
    // do your fetchMeta/nextKeyOp logic here …
    yield (i > 0 ? ',' : '') +
      JSON.stringify({
        serverMeta,
        message: entry.serialize()
      })
    i++
  }
  yield ']'
}(), { encoding: 'utf8' })
```

That way you get rid of:

  • the `this.destroy()` dance  
  • the `ended` flag  
  • the awkward self‑invoked async in `read()`  

and your code becomes a straight‑line async loop.

---

Putting it all together here’s a sketch of a tightened‑up version of just the top of that selector:

```js
/* … */
export default ((sbp('sbp/selectors/register', {
  'backend/db/streamEntriesAfter': async function (
    contractID: string,
    height: number,
    requestedLimit: ?number,
    { keyOps = false }: { keyOps?: boolean } = {}
  ): Promise<Readable> {
    const limit = Math.min(
      requestedLimit ?? Infinity,
      Number(process.env.MAX_EVENTS_BATCH_SIZE) || 500
    )

    const latest = await sbp('chelonia/db/latestHEADinfo', contractID)
    if (latest === '') {
      throw Boom.resourceGone(`contract ${contractID} has been deleted!`)
    }
    if (!latest) {
      throw Boom.notFound(`contract ${contractID} doesn't exist`)
    }

    // helper to load the next HEAD‐info
    let currentHeight = height
    let serverMeta, currentHash

    const fetchMeta = async () => {
      if (currentHeight > latest.height) {
        return false
      }
      const m = await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
      if (!m) return false
      // pull out the hash and leave the rest
      ;({ hash: currentHash, ...serverMeta } = m)
      return true
    }

    // helper to jump in 10k‐height blocks for keyOps
    let index: ?string[] = null
    const nextKeyOp = async () => {
      if (!index) {
        const raw = await sbp('chelonia.db/get',
          `_private_keyop_idx_${contractID}_${Math.floor(currentHeight / 10000) * 10000}`
        )
        index = raw ? raw.split('\x00') : []
      }
      const i = index.findIndex(h => Number(h) >= currentHeight)
      if (i >= 0) {
        currentHeight = Number(index[i])
        index = index.slice(i + 1)
      } else {
        currentHeight = Math.floor(currentHeight / 10000 + 1) * 10000
        index = null
      }
    }

    // now build a Readable from an async generator
    const stream: Readable & { headers: { [string]: string } } =
      Readable.from(async function* () {
        yield '['
        // do one fetch
        if (!await fetchMeta()) {
          yield ']'
          return
        }
        // if keyOps, skip until isKeyOp
        if (keyOps) {
          while (!serverMeta.isKeyOp) {
            await nextKeyOp()
            if (!await fetchMeta()) {
              yield ']'
              return
            }
          }
        }
        let count = 0
        while (count < limit) {
          const entry = await sbp('chelonia/db/getEntry', currentHash)
          if (!entry) break

          const payload = JSON.stringify({
            serverMeta,
            message: entry.serialize()
          })

          yield (count === 0 ? '' : ',')
          yield `"${strToB64(payload)}"`

          count++
          currentHeight++
          if (!await fetchMeta()) break

          if (keyOps) {
            while (!serverMeta.isKeyOp) {
              await nextKeyOp()
              if (!await fetchMeta()) break
            }
          }
        }
        yield ']'
      }(), { encoding: 'utf8' })

    stream.headers = {
      'shelter-headinfo-head': String(latest.HEAD),
      'shelter-headinfo-height': String(latest.height)
    }
    return stream
  }
}): any): string[])
```

This rewrite

  • avoids shadowing/confusing variable names  
  • makes the control flow of “fetch one meta, maybe skip, yield, loop” very explicit  
  • gets rid of manual `.push()` and `ended` and `this.destroy()`  
  • uses a single `Readable.from` call for easier testing  

I hope this helps! Let me know if you need any other contexts or if anything here doesn’t line up with your runtime.
</details>

## `o4-mini`

<details>
Here are a few issues I noticed, plus some suggestions to both tighten up the logic and make it more readable/maintainable. Everything I quote below is from your new `streamEntriesAfter` implementation in **`backend/database.js`**.

---

### 1) Nested async IIFE inside `.read()`  
  
Right now you have this:

```js
read (): void {
  if (ended) {
    this.destroy()
    return
  }
  const end = () => {
    this.push(']')
    this.push(null)
    ended = true
  }
  if (counter < limit) {
    (async () => {
      try {
        // ... a bunch of logic ...
      } catch (e) {
        console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
        end()
      }
    })()
  } else {
    end()
  }
}
```

A few problems here:  
  1.  You’re spawning a detached async IIFE on each `.read()`, which makes back‑pressure semantics tricky  
  2.  It’s very dense to follow the height/fetchMeta/nextKeyOp dance inside that IIFE  

**Suggestion:** switch to `Readable.from` + an async generator.  That way you can write a straight‑line async loop, yield each chunk, then `Readable.from` will handle back‐pressure for you.  Something like:

```js

// … inside your sbp selector …
const stream = Readable.from((async function * () {
  yield '['
  let first = true
  let currentHeight = height
  let serverMeta
  let currentHash
  let counter = 0

  const nextKeyOp = createNextKeyOpFn(contractID, () => currentHeight)
  const fetchMeta = async () => {
    if (currentHeight > latestHEADinfo.height) return false
    const meta = await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
    if (!meta) return false
    currentHash = meta.hash
    // strip off the `hash` field, leave everything else
    serverMeta = (({ hash, ...rest }) => rest)(meta)
    return true
  }

  while (counter < limit) {
    if (options.keyOps) {
      let ok = false
      // scan forward until we find a keyOp or exhaust
      while (!ok) {
        await nextKeyOp()
        ok = await fetchMeta()
        if (!ok) break
        ok = serverMeta.isKeyOp
      }
      if (!ok) break
    } else {
      if (!await fetchMeta()) break
    }

    const entry = await sbp('chelonia/db/getEntry', currentHash)
    if (!entry) break

    const prefix = first ? '' : ','
    first = false

    const payload = JSON.stringify({
      serverMeta,
      message: entry.serialize()
    })
    const chunk = `${prefix}"${strToB64(payload)}"`
    yield chunk

    counter++
    currentHeight++
  }

  yield ']'
})())
```

This has a few advantages:

  - you get **real** async/await flow control, rather than a detached IIFE  
  - you don’t have to manage `ended`, `this.destroy()`, or manual `.push(null)`  
  - readability goes way up (and it’s much easier to see exactly when you break out of the loop)  

---

### 2) `nextKeyOp` mutates its `index` array with `.splice()`  

```js
const nextKeyOp = (() => {
  let index

  return async () => {
    if (!index) {
      index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))
        ?.split('\x00')
    }
    const value = index?.find((h, i) => {
      if (Number(h) >= currentHeight) {
        index = ((index: any): any[]).splice(i + 1)
        return true
      } else {
        return false
      }
    })
    …
  }
})()
```

Splicing inside a `.find()` predicate is clever but rather opaque.  I’d recommend something like:

```js
const nextKeyOp = (() => {
  let index: number[] = []

  return async () => {
    if (!index.length) {
      const raw = await sbp('chelonia.db/get', 
        `_private_keyop_idx_${contractID}_${Math.floor(currentHeight / 10000) * 10000}`
      )
      index = raw
        ? raw.split('\x00').map(n => Number(n)).filter(n => !Number.isNaN(n)).sort((a, b) => a - b)
        : []
    }

    const next = index.find(n => n >= currentHeight)
    if (next != null) {
      // drop everything ≤ next
      index = index.filter(n => n > next)
      currentHeight = next
    } else {
      // bump to next 10 000 block
      currentHeight = Math.floor(currentHeight / 10000) * 10000 + 10000
      index = []
    }
  }
})()
```

That way you avoid mutating the same array in the middle of a `.find()` and it’s clearer what’s happening.

---

### 3) You’ve removed the **initial** `getEntryMeta` before the very first `.read()`  

Originally you had:

```diff
-    let { hash: currentHash, ...serverMeta } = (
-      await sbp('chelonia/db/getEntryMeta', contractID, height)
-    ) || {}
+    let currentHeight = height
+    let currentHash, serverMeta
```

Now your **first** call to `fetchMeta()` happens inside `.read()`.  That works, but it means you never know if your starting `height` is out of range until after you push `'['`.  If you want to early‐fail (e.g. if `height > HEAD`), you could do the initial fetch once up‐front:

```js
// right after you check for latestHEADinfo...
const firstMeta = await sbp('chelonia/db/getEntryMeta', contractID, height)
if (!firstMeta) {
  // no entries at `height`, so just return an empty array
  const empty = Readable.from(['[', ']'])
  empty.headers = { … }
  return empty
}
```

That way you never send a `[` without eventually sending at least one chunk (or `]`) back.

---

### 4) Minor stylistic nit: respect StandardJS (no semicolons + 2‑space indent)

You’re already mostly in line, but if you introduce my above code you’ll want to be sure you drop **all** trailing semicolons, e.g.:

```js
const stream = Readable.from((async function * () {
  // …
})())
```

not

```js
const stream = Readable.from((async function * () {
  // …
})());   // ← semicolon
```

---

In summary:

1.  **No obvious correctness bugs** jumped out—the new `keyOps` logic and fetch/fetch‑meta dance look sound.  
2.  I do recommend **refactoring** to an async generator + `Readable.from` to make the code far easier to read and maintain.  
3.  Clean up the `nextKeyOp` mutation logic to avoid hidden side‑effects.  
4.  (Optionally) pull out the very first `getEntryMeta` so you can early‑exit before emitting `[` if there really are no entries at the requested height.  

Let me know if you want more detail or any further context around any of these points.
</details>

## `Gemini 2.5 Pro Exp`

<details>
Okay, let's review the changes.

### Review of `backend/database.js`

The main change is the refactoring of the `backend/db/streamEntriesAfter` selector to support a new `options.keyOps` flag, allowing streaming of only key operations. This involved introducing helper functions `nextKeyOp` and `fetchMeta` and significantly changing the logic within the `read` method of the `Readable` stream.

**Potential Issues & Suggestions:**

1.  **Bug in `keyOps` Loop Initialization:**
    In the `read()` method, when `options.keyOps` is true, the logic enters a `while (!serverMeta?.isKeyOp)` loop. However, `serverMeta` is initialized as `undefined` outside the `try` block and only populated *inside* the loop after calls to `nextKeyOp()` and `fetchMeta()`. This means on the very first execution of `read()` for a stream with `keyOps: true`, the `while` loop condition `!serverMeta?.isKeyOp` will initially be true (because `serverMeta` is `undefined`), causing `nextKeyOp()` to be called *before* checking if the *initial* `height` passed to the function corresponds to a key operation. This could lead to the first potential key operation at the starting `height` being skipped.

    *   **Lines causing potential issue:**
        ```diff
        --- a/backend/database.js
        +++ b/backend/database.js
        @@ -61,9 +106,9 @@
          }
          // Number of entries pushed.
          let counter = 0
        -  let { hash: currentHash, ...serverMeta } = (
        -    await sbp('chelonia/db/getEntryMeta', contractID, height)
        -  ) || {}
        +  let currentHeight = height
        +  let currentHash, serverMeta // <- serverMeta starts as undefined
          let prefix = ''
          let ended = false
        +  const nextKeyOp = (() => {
        [...]
        @@ -118,32 +163,36 @@
                end()
              }
              if (options.keyOps) {
        -        while (!serverMeta?.isKeyOp) {
        +        while (!serverMeta?.isKeyOp) { // <- serverMeta is undefined on first run
                    await nextKeyOp()
                    if (!await fetchMeta()) return end()
                  }
                } else {
                  if (!await fetchMeta()) return end()
                }
                const entry = await sbp('chelonia/db/getEntry', currentHash)

        ```

    *   **Suggestion:** Fetch the metadata for the initial `currentHeight` *before* entering the `while` loop if `options.keyOps` is true.

        ```javascript
          // Inside read():
          if (counter < limit) {
            (async () => {
              try {
        +       // Fetch meta first if we might need to check isKeyOp immediately
        +       if (!serverMeta && currentHash === undefined) {
        +         // Only fetch if we don't already have meta/hash from a previous iteration
        +         if (!await fetchMeta()) return end()
        +       }
        +
                if (options.keyOps) {
                   while (!serverMeta?.isKeyOp) {
                     await nextKeyOp()
                     if (!await fetchMeta()) return end()
                   }
                 } else {
        +          // If keyOps is false and we haven't fetched meta yet (e.g., first run)
        +          if (!serverMeta && currentHash === undefined) {
        +            // This redundant check ensures fetchMeta() runs at least once
        +            // if options.keyOps is false and we didn't fetch above.
        +            // Alternatively, restructure to always call fetchMeta once before conditional logic.
        +            // Simplified approach below.
        +          }
                   if (!await fetchMeta()) return end()
                 }
                 const entry = await sbp('chelonia/db/getEntry', currentHash)
          // ... rest of the logic
        ```

    *   **Alternative (Simpler) Suggestion:** Restructure the logic slightly to ensure `fetchMeta` is always called once at the start of the `try` block before the `keyOps` check.

        ```javascript
        // Inside read():
        if (counter < limit) {
          (async () => {
            try {
              // Always try to fetch meta for the current height first.
              // This handles the initial call and subsequent calls naturally.
              if (!serverMeta && currentHash === undefined) {
                if (!await fetchMeta()) return end()
              }
        
              // If keyOps is enabled and the fetched meta isn't a key op, find the next one.
              if (options.keyOps && !serverMeta?.isKeyOp) {
                // Loop until a key op is found or the end is reached
                while (!serverMeta?.isKeyOp) {
                  await nextKeyOp()
                  if (!await fetchMeta()) return end()
                }
              }
        
              // If, after potentially finding the next key op or if keyOps was false,
              // we still don't have valid metadata (e.g., fetchMeta failed inside loop), end.
              // Note: fetchMeta() already returns false if meta is missing, so this check might be redundant
              // if fetchMeta() failures always result in end() being called immediately.
              // Keeping it for clarity might be okay, or remove if fetchMeta() path guarantees ending.
              // if (!serverMeta) return end() // Likely redundant check
        
              // At this point, we should have the correct currentHash and serverMeta
              // for either the next sequential entry or the next key op entry.
              const entry = await sbp('chelonia/db/getEntry', currentHash)
              if (entry) {
                const currentPrefix = prefix
                prefix = ','
                counter++
                this.push(
                  `${currentPrefix}"${strToB64(
                    JSON.stringify({ serverMeta, message: entry.serialize() })
                  )}"`
                )
                // Prepare for the next iteration
                currentHeight++
                currentHash = undefined
                serverMeta = undefined
              } else {
                // If entry doesn't exist for the hash we found in meta, something is wrong.
                console.error(`[backend] streamEntriesAfter: Entry not found for hash ${currentHash} at height ${currentHeight - 1}`) // Height was already incremented
                end()
              }
            } catch (e) {
              console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
              // Consider destroying with error for better propagation
              // this.destroy(e)
              end()
            }
          })()
        } else {
          end()
        }
        ```
        *(Self-correction: The second alternative seems cleaner and handles both `keyOps: true` and `keyOps: false` more consistently regarding the initial fetch.)*

2.  **Error Handling in Stream:**
    The `catch` block currently logs the error to the console and gracefully ends the stream by pushing `]` and `null`. While this prevents the stream from hanging, the consumer of the stream might not be aware that an error occurred internally. It's generally better practice for streams to emit an `error` event when they encounter an unrecoverable error.

    *   **Line:**
        ```javascript
        } catch (e) {
          console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
          end() // <- Graceful end, hides error from consumer
        }
        ```

    *   **Suggestion:** Use `this.destroy(e)` to signal an error occurred.

        ```javascript
        } catch (e) {
          console.error(`[backend] streamEntriesAfter: read() error: ${e.message}:`, e.stack)
          // Signal an error on the stream
          this.destroy(e)
        }
        ```

3.  **Efficiency of `nextKeyOp` Indexing:**
    The `nextKeyOp` function fetches index keys like `_private_keyop_idx_${contractID}_${blockStart}` where `blockStart` increments by 10000. This seems like a reasonable optimization if key operations are relatively sparse. It avoids iterating height by height. The use of `splice` to consume the fetched index block is also efficient for subsequent calls within the same block. No specific issues found here, but it relies on the index keys (`_private_keyop_idx_*`) being correctly maintained elsewhere.

4.  **Flow Type Cast:**
    The cast `((index: any): any[])` in `nextKeyOp` is necessary because Flow can't guarantee `index` is still an array after the `?.` optional chaining, even though the `find` method implies it is. While slightly verbose, it's a standard way to handle this limitation in Flow.

    *   **Line:**
        ```javascript
        index = ((index: any): any[]).splice(i + 1)
        ```
    *   **Comment:** This is acceptable Flow syntax to work around type inference limitations.

Overall, the refactoring introduces valuable functionality. The main concerns are the potential bug in the initial handling of `keyOps: true` and the error propagation from the stream. The suggested changes address these points.
</details>

## `Gemini 2.5 Pro Preview`

<details>
Okay, I've reviewed the provided code changes. Here's my feedback:

The main change introduces an `options: { keyOps?: boolean }` parameter to `backend/db/streamEntriesAfter` and refactors the streaming logic significantly to support fetching only key operations when this option is enabled. The refactoring also improves the general logic for streaming all entries compared to the previous version.

Here are a few points for improvement:

1.  **Potential Null/Undefined Issue in `nextKeyOp`**

    In the `nextKeyOp` function, if the key `_private_keyop_idx_${contractID}_${...}` exists but holds an empty string (`''`), the `split('\x00')` call will result in `['']`. The `find` loop then tries `Number('')` which evaluates to `0`. If `currentHeight` is also `0`, the `Number(h) >= currentHeight` condition might unexpectedly pass for this empty string entry, potentially causing issues. While `splice` might handle the index correctly, it relies on potentially confusing behavior. It's safer to explicitly handle the case where the fetched index string might be empty or the split results in an empty/invalid array.

    ```diff
    diff --git a/backend/database.js b/backend/database.js
    --- a/backend/database.js
    +++ b/backend/database.js
    @@ -68,7 +68,11 @@
 
       return async () => {
         if (!index) {
    -      index = (await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`))?.split('\x00')
    +      // Fetch and filter out potential empty strings resulting from split
    +      const rawIndex = await sbp('chelonia.db/get', `_private_keyop_idx_${contractID}_${currentHeight - currentHeight % 10000}`)
    +      index = rawIndex ? rawIndex.split('\x00').filter(h => h !== '') : undefined
    +      // Ensure index is undefined if empty after filtering
    +      if (index && index.length === 0) index = undefined
         }
         const value = index?.find((h, i) => {
           if (Number(h) >= currentHeight) {

    ```

2.  **Type Safety for Environment Variable**

    The calculation for `limit` relies on `process.env.MAX_EVENTS_BATCH_SIZE`. If this environment variable is set but is an empty string or non-numeric, `Number()` could return `0` or `NaN`. `Math.min` with `NaN` results in `NaN`, which could break the `< limit` check. It's better to explicitly parse and validate the number.

    ```diff
    diff --git a/backend/database.js b/backend/database.js
    --- a/backend/database.js
    +++ b/backend/database.js
    @@ -44,7 +44,8 @@
 
      // Streams stored contract log entries since the given entry hash (inclusive!).
      export default ((sbp('sbp/selectors/register', {
    -  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number): Promise<*> {
    -    const limit = Math.min(requestedLimit ?? Number.POSITIVE_INFINITY, process.env.MAX_EVENTS_BATCH_SIZE ?? 500)
    +  'backend/db/streamEntriesAfter': async function (contractID: string, height: number, requestedLimit: ?number, options: { keyOps?: boolean } = {}): Promise<*> {
    +    const envLimit = parseInt(process.env.MAX_EVENTS_BATCH_SIZE, 10)
    +    const maxLimit = !isNaN(envLimit) && envLimit > 0 ? envLimit : 500
    +    const limit = Math.min(requestedLimit ?? Number.POSITIVE_INFINITY, maxLimit)
         const latestHEADinfo = await sbp('chelonia/db/latestHEADinfo', contractID)
         if (latestHEADinfo === '') {
           throw Boom.resourceGone(`contractID ${contractID} has been deleted!`)

    ```
    *Self-correction: The original code used `?? 500`, which handles `undefined` or `null`. However, it doesn't handle `''` correctly as `Number('')` is `0`. The proposed `parseInt` approach with `isNaN` check is more robust.*

3.  **Redundant `else` Block in `read()`**

    The `else` block corresponding to the `if (options.keyOps)` check is slightly redundant, as the following `if (!await fetchMeta()) return end()` line is identical to the one inside the `else` block. You can remove the `else` block and execute the `fetchMeta` check unconditionally after the `keyOps` block.

    ```diff
    diff --git a/backend/database.js b/backend/database.js
    --- a/backend/database.js
    +++ b/backend/database.js
    @@ -129,10 +133,9 @@
                   await nextKeyOp()
                   if (!await fetchMeta()) return end()
                 }
    -          } else {
    -            if (!await fetchMeta()) return end()
               }
    +          // Fetch meta if not already fetched (or if keyOps loop finished without fetching)
    +          if (!currentHash && !await fetchMeta()) return end()
    +
               const entry = await sbp('chelonia/db/getEntry', currentHash)
               if (entry) {
                 const currentPrefix = prefix

    ```
    *Self-correction: On second thought, the `keyOps` `while` loop ensures `fetchMeta` is called until a key op is found or the end is reached. If the loop exits because `!serverMeta?.isKeyOp` becomes false, `currentHash` will be set. If it exits because `fetchMeta` returned false, `end()` is already called. Therefore, we only need to call `fetchMeta` if *not* in `keyOps` mode. The original structure is actually slightly better/clearer.*

    Let's revert the previous suggestion and propose a minor cleanup instead:

    ```diff
    diff --git a/backend/database.js b/backend/database.js
    --- a/backend/database.js
    +++ b/backend/database.js
    @@ -126,13 +130,13 @@
           (async () => {
             try {
               if (options.keyOps) {
    +            // Find the next key op entry
                 while (!serverMeta?.isKeyOp) {
                   await nextKeyOp()
                   if (!await fetchMeta()) return end() // Reached end while searching
                 }
               } else {
    -            if (!await fetchMeta()) return end()
    +            // Fetch the next entry (non-keyOp specific)
    +            if (!await fetchMeta()) return end() // Reached end
               }
               const entry = await sbp('chelonia/db/getEntry', currentHash)
               if (entry) {

    ```
    *This version keeps the original logic flow but removes the apparently redundant `if` check inside the `else` block, which was slightly confusing.*

Overall, the refactoring is a significant improvement, especially with the addition of the `keyOps` functionality. The new streaming logic inside `read()` is more robust and correctly handles fetching metadata before processing the entry. The error handling is also better encapsulated. Just address the minor points above for increased robustness.
</details>

taoeffect (changes requested):
Review ready!

shared/domains/chelonia/chelonia.js
corrideat: Pretty sure, as `SPMessage` doesn't use this at all. `chelonia/out/keyRequest` has `originatingContractID` as a parameter, but it's _not_ part of the `SPMessage`. It used to be the case, but this was removed a while ago (I don't remember the specific PR or commit, but it's when the `'chelonia/contract/disconnect'` selector was added) as this information is now encrypted.

corrideat: > then the other code should be cleaned up too.

It should, but I think this is the only place that needs to be cleaned up. Other places may also need cleaning up, but just because `originatingContractID` is used it doesn't mean it needs to be cleaned up; it's just passing it to `SPMessage` that's no longer necessary nor used.

shared/domains/chelonia/db.js
corrideat: This is the first thing that I worked on, so I may not remember the full details either. 

Until then, a 'lightweight' client didn't exist, and clients stored full copies of the chain. The 'lightweight' client is lightweight by basically making the DB a no-op, except that we needed to fetch the current contract head from the DB, so that's especially treated.

backend/database.js
corrideat: I still need to review the LLM feedback in more detail. There seem to be some plausible improvement suggestions, but I need to see those in more detail, as this is a place where correctness is especially important.

backend/server.js
corrideat: This is done AFAICT.

backend/routes.js
corrideat: I didn't expect it, but `boolean()` works (I didn't expect it as query params are strings)

backend/database.js
corrideat: I've reviewed the generated reviews and I don't think there's a lot relevant there, other than switching to an async generator (which I agree would be better).

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
-    const stream = new Readable({
-      construct (cb) {
-        this.push('[')
-        cb()
-      },
-      read (): void {
-        if (ended) {
-          this.destroy()
-          return
+    const nextKeyOp = (() => {
```
taoeffect, line 73: Above this could you please add a comment explaining the function's behavior as well as all possible return values and their corresponding meaning?

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
-    })
+    })()
+    const fetchMeta = async () => {
+      if (currentHeight > latestHEADinfo.height) {
+        return false
+      }
+      const meta = await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
+      if (!meta) {
+        return false
+      }
```
taoeffect, line 125: Hmm... just wondering, would this be an unusual situation, perhaps deserving of a `console.warn`?

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
+            return false
+          } else {
+            return null
+          }
         }
+        return true
       }
-    })
+    })()
+    const fetchMeta = async () => {
```
taoeffect, line 118: This returns different return values, presumably with differing meaning (which isn't documented), but the return values also don't seem to be used anywhere... So what's the point of returning a value that isn't ever used?

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
-          }).catch(e => {
-            console.error(`[backend] streamEntriesAfter: read(): ${e.message}:`, e.stack)
-            this.push(']')
-            this.push(null)
-            ended = true
-          })
+        const value = index?.find((h, i) => {
+          if (Number(h) >= currentHeight) {
+            // Remove values that no longer are relevant from the index
+            index = ((index: any): any[]).splice(i + 1)
```
taoeffect: Since using `splice()` mutates the original array, making this a bit hard to reason about.

Could you use `slice` instead?

```diff
-          index = ((index: any): any[]).splice(i + 1)
+          index = (index: any[]).slice(i + 1)
```

(Suggestion by `o4-mini-high` I agree with)

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
+            currentHash = undefined
+            serverMeta = undefined
+          } else {
+            break
+          }
+          // queries with 'keyOps' always return the requested height, whether
+          // or not a keyOp.
+          if (options.keyOps) {
+            // Advance until the next key op
+            while ((await nextKeyOp()) === null);
```
taoeffect: The manipulation of `currentHeight` here is a bit confusing to me, as it's both manipulated above with `currentHeight++` and here with `currentHeight = currentHeight - currentHeight % 10000 + 10000`

Could you please:

1. Double-check it's being manipulated correctly
2. Add a comment to explain here how this loop manipulates `currentHeight` and its relationship with the way `currentHeight` is manipulated above by increments?

It almost seems like you're combining two things that should be separate, and so I'm confused. If I understand correctly, if `options.keyOps` is set, only the keyOps should be returned, so then why have it part of code that returns all ops incrementally? Would it be possible to separate these methods of incrementing `currentHeight` more explicitly somehow? (e.g. have two different `Readable` streams?)

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
     }
     // Number of entries pushed.
     let counter = 0
-    let { hash: currentHash, ...serverMeta } = (
-      await sbp('chelonia/db/getEntryMeta', contractID, height)
-    ) || {}
+    let currentHeight = height
```
taoeffect, line 59: Just above this `o4-mini-high` points out that `process.env.MAX_EVENTS_BATCH_SIZE` isn't coerced to a number:

5) You do not coerce `process.env.MAX_EVENTS_BATCH_SIZE` to a number. As written:

```js
const limit = Math.min(
  requestedLimit ?? Number.POSITIVE_INFINITY,
  process.env.MAX_EVENTS_BATCH_SIZE ?? 500
)
```

If someone exports `MAX_EVENTS_BATCH_SIZE='1000'` your Math.min will see the string `"1000"`.  Always do:

```diff
-  process.env.MAX_EVENTS_BATCH_SIZE ?? 500
+  Number(process.env.MAX_EVENTS_BATCH_SIZE) || 500
```

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
+      if (!meta) {
+        return false
+      }
+
+      const { hash: newCurrentHash, ...newServerMeta } = meta
+      currentHash = newCurrentHash
+      serverMeta = newServerMeta
+
+      return true
+    }
```
taoeffect, line 132: Neither `nextKeyOp` nor `fetchMeta` handle errors in any way. Claude complained about this and offered the try/catch example below. My question is: would it make sense to add some specific error logging messages in both of these functions and then propagate the error?

<details>
### 1. Error handling in fetchMeta function

The `fetchMeta` function doesn't handle errors - any exceptions would propagate up to the generator. While the generator has a try/catch block, it would be clearer to handle errors in `fetchMeta` directly.

```js
const fetchMeta = async () => {
  try {
    if (currentHeight > latestHEADinfo.height) {
      return false
    }
    const meta = await sbp('chelonia/db/getEntryMeta', contractID, currentHeight)
    if (!meta) {
      return false
    }

    const { hash: newCurrentHash, ...newServerMeta } = meta
    currentHash = newCurrentHash
    serverMeta = newServerMeta

    return true
  } catch (e) {
    console.error(`[backend] fetchMeta error for ${contractID} at height ${currentHeight}: ${e.message}`, e)
    return false
  }
}
```
</details>

backend/database.js
```diff
@@ -52,57 +52,91 @@ export default ((sbp('sbp/selectors/register', {
...
+    // NOTE: if this ever stops working you can also try Readable.from():
+    // https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options
+    const stream = Readable.from((async function * () {
+      yield '['
+      await fetchMeta()
+      while (serverMeta && counter < limit) {
+        try {
+          const entry = await sbp('chelonia/db/getEntry', currentHash)
+          if (entry) {
```
taoeffect: Claude suggests moving this code below outside of the if by doing:

```js
  const entry = await sbp('chelonia/db/getEntry', currentHash)
  if (!entry) break
```

I think it would make it slightly cleaner.

taoeffect (changes requested):
Manual review of `backend/database.js` finished! Great work @corrideat!!

backend/database.js
corrideat: Splice is significantly faster because it doesn't need to make a copy, and in this case the entire point is discarding data, and there's no danger of mutating an array reference.

https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJ3ajRPM1ZuTjIwLV9NUzRUR3pFODIiLCJjb2RlIjoiREFUQS5zcGxpY2UoNTAwKSIsIm5hbWUiOiJTcGxpY2UiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiMEFKa0QycFJ6WG8xSFdGc0dQeWhMIiwiY29kZSI6IkRBVEEuc2xpY2UoNTAwKSIsIm5hbWUiOiJTbGljZSIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuIG5ldyBBcnJheSgxMDI0KS5maWxsKDUpIn19

backend/database.js
corrideat: Not necessarily an unusual situation, it can just be the end of the contract, although I've added height guards now. It could also happen if the contract is being simultaneously deleted (granted, a strange situation, but then also nothing actionable to do)

backend/database.js
corrideat: If you look at previous iterations of `fetchMeta`, the return value was used, but then I refactored this to simplify the logic and a side effect is that the return value no longer needs to be used. I'd still leave it so that it's easier to start using it in the future if the need arises (but I can add a comment about what they signify).

backend/database.js
corrideat: It's being manipulated correctly. I can double check again using a smaller value than `10000` and see (since this function has been refactored since).

The purpose of `nextKeyOp` is to advance `currentHeight` until the next key op. Since the index keys are segments of 10000 heights, the height needs to be advanced by ~10000 until we reach the end.

> It almost seems like you're combining two things that should be separate, and so I'm confused. If I understand correctly, if options.keyOps is set, only the keyOps should be returned, so then why have it part of code that returns all ops incrementally?

Having these be separate would result in more complex code and very likely DRY violations. The current logic is general enough that it can handle both cases (as well as future cases, such as adding other types of filters). Now, something that I think could be made more 'generic', but which likely is overkill right now, is having a generator function for `currentHeight`. That way, instead of doing `currentHeight++` and then `nextKeyOp`, we'd do something like `await height.next()`. This would make things clearer, but I think it may be too much for the current situation with only two possible cases.

Would you like me to add a comment about this future improvement, or to do it right now?

backend/database.js
corrideat: True, but then `Math.min` will coerce it into a number automatically. I'm not sure what the advantage of this is.

backend/database.js
corrideat: Not sure, we don't really expect these things to fail, and if they do fail, the error will be caught by the parent error handler.

backend/database.js
corrideat: Sure. I think I agree.

backend/database.js
corrideat: Never mind, the above benchmark isn't entirely accurate because `splice` will mutate the array and `DATA` isn't reset on each run. See the more accurate benchmark below.

Since they seem to have similar performance (and `slice` may actually be faster), I'll make this change.


https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJ3ajRPM1ZuTjIwLV9NUzRUR3pFODIiLCJjb2RlIjoiREFUQVswXSA9IERBVEFbMF0uc3BsaWNlKDUwMCkiLCJuYW1lIjoiU3BsaWNlIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IjBBSmtEMnBSelhvMUhXRnNHUHloTCIsImNvZGUiOiJEQVRBWzBdID0gREFUQVswXS5zbGljZSg1MDApIiwibmFtZSI6IlNsaWNlIiwiZGVwZW5kZW5jaWVzIjpbXX1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOltdfSwiZGF0YUNvZGUiOiJyZXR1cm4gW25ldyBBcnJheSgxMDI0KS5maWxsKDUpXSJ9fQ

taoeffect (approved):
Amazing work @corrideat ! 👏

PR #2753, opened 2025-04-17
#2712 - Implement 'mark message as unread' feature in the chatroom

### LINKED ISSUE #2712: Mark messages as unread menu item
### Problem

Similar to how Slack and other programs let you mark messages as unread, it would be nice if we had that feature too.

### Solution

Add a menu item to chatrooms that lets you mark a message as unread, and every message from that point on will be unread.

SebinSong:
closes #2712 

**[Video demo for the feature]**
As you can see, once 'mark unread' operation is processed, it is immediately reflected in the UI(eg. 'is-new' blue divider and also notification badge **if** there are any message where the user is mentioned). This state does not disappear until the user leaves and re-enters the chatroom. Hope it sounds good.

https://github.com/user-attachments/assets/7d785a4d-a344-4894-9e7d-71650de26632

Please let me know if we would love to change the icon to something else BTW:

### DIFF
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -124,17 +124,20 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const getUpdatedUnreadMessages = async () => {
         const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
-        const existingMessageHash = currentData[contractID]?.readUntil?.messageHash
+        const existingReadUntil = currentData[contractID]?.readUntil
 
-        if (existingMessageHash !== messageHash) {
-          return {
-            ...currentData,
-            [contractID]: {
-              readUntil: { messageHash, createdHeight, isManuallyMarked: true },
-              unreadMessages
-            }
+        // If the requested mark-unread hash has already been set, ignore it.
+        if (existingReadUntil &&
+          existingReadUntil.isManuallyMarked &&
+          existingReadUntil?.messageHash === messageHash) { return null }
+
+        return {
+          ...currentData,
+          [contractID]: {
+            readUntil: { messageHash, createdHeight, isManuallyMarked: true },
+            unreadMessages
           }
-        } else { return null }
+        }
       }
 
       const data = await getUpdatedUnreadMessages()
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -124,13 +124,17 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const getUpdatedUnreadMessages = async () => {
         const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
+        const existingMessageHash = currentData[contractID]?.readUntil?.messageHash
 
-        return {
-          ...currentData,
-          [contractID]: {
-            readUntil: { messageHash, createdHeight }, unreadMessages
+        if (existingMessageHash !== messageHash) {
+          return {
+            ...currentData,
+            [contractID]: {
+              readUntil: { messageHash, createdHeight },
+              unreadMessages
+            }
           }
-        }
+        } else { return null }
       }
 
       const data = await getUpdatedUnreadMessages()
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -130,7 +130,7 @@ export default (sbp('sbp/selectors/register', {
           return {
             ...currentData,
             [contractID]: {
-              readUntil: { messageHash, createdHeight },
+              readUntil: { messageHash, createdHeight, isManuallyMarked: true },
               unreadMessages
             }
           }
@@ -173,7 +173,7 @@ export default (sbp('sbp/selectors/register', {
       const getUpdatedUnreadMessages = async () => {
         const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
 
-        const index = currentData[contractID]?.unreadMessages.findIndex(msg => msg.messageHash === messageHash)
+        const index = currentData[contractID]?.unreadMessages?.findIndex(msg => msg.messageHash === messageHash)
         // NOTE: index could be undefined if unreadMessages is not initialized
         if (Number.isInteger(index) && index >= 0) {
           currentData[contractID].unreadMessages.splice(index, 1)
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -82,14 +82,17 @@ export default (sbp('sbp/selectors/register', {
       }
     })
   },
-  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight }: {
-    contractID: string, messageHash: string, createdHeight: number
+  'gi.actions/identity/kv/setChatRoomReadUntil': ({ contractID, messageHash, createdHeight, forceUpdate = false }: {
+    contractID: string,
+    messageHash: string,
+    createdHeight: number,
+    forceUpdate: boolean // In a rare case, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'. (reference: https://github.com/okTurtles/group-income/issues/2729)
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const getUpdatedUnreadMessages = async () => {
         const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
 
-        if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
+        if (forceUpdate || currentData[contractID]?.readUntil.createdHeight < createdHeight) {
           const { unreadMessages } = currentData[contractID]
           return {
             ...currentData,
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -86,8 +86,7 @@ export default (sbp('sbp/selectors/register', {
     contractID: string,
     messageHash: string,
     createdHeight: number,
-    // In some cases, such as when the latest message is deleted or the user uses the 'mark-unread' feature,
-    // the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
+    // In some cases, such as when the latest message is deleted, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
     // 'forceUpdate' flag is used to override the 'createdHeight' check below to allow this kind of update.
     // (reference: https://github.com/okTurtles/group-income/issues/2729)
     forceUpdate: boolean
@@ -107,6 +106,30 @@ export default (sbp('sbp/selectors/register', {
           }
         }
         return null
+      }
+
+      const data = await getUpdatedUnreadMessages()
+      if (data) {
+        await sbp('gi.actions/identity/kv/saveChatRoomUnreadMessages', { data, onconflict: getUpdatedUnreadMessages })
+      }
+    })
+  },
+  'gi.actions/identity/kv/markAsUnread': ({ contractID, messageHash, createdHeight, unreadMessages }: {
+    contractID: string,
+    messageHash: string,
+    createdHeight: number,
+    unreadMessages: Array<{ messageHash: string, createdHeight: number }>
+  }) => {
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
+      const getUpdatedUnreadMessages = async () => {
+        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
+
+        return {
+          ...currentData,
+          [contractID]: {
+            readUntil: { messageHash, createdHeight }, unreadMessages
+          }
+        }
       }
 
       const data = await getUpdatedUnreadMessages()
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -86,7 +86,7 @@ export default (sbp('sbp/selectors/register', {
     contractID: string,
     messageHash: string,
     createdHeight: number,
-    // In a rare case, such as when the latest message is deleted,
+    // In some cases, such as when the latest message is deleted or the user uses the 'mark-unread' feature,
     // the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
     // 'forceUpdate' flag is used to override the 'createdHeight' check below to allow this kind of update.
     // (reference: https://github.com/okTurtles/group-income/issues/2729)
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -86,7 +86,11 @@ export default (sbp('sbp/selectors/register', {
     contractID: string,
     messageHash: string,
     createdHeight: number,
-    forceUpdate: boolean // In a rare case, the 'readUntil' value needs to be set to the msg with lower 'createdHeight'. (reference: https://github.com/okTurtles/group-income/issues/2729)
+    // In a rare case, such as when the latest message is deleted,
+    // the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
+    // 'forceUpdate' flag is used to override the 'createdHeight' check below to allow this kind of update.
+    // (reference: https://github.com/okTurtles/group-income/issues/2729)
+    forceUpdate: boolean
   }) => {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const getUpdatedUnreadMessages = async () => {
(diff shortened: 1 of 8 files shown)


frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -133,6 +133,11 @@ export default ({
         action: 'openEmoticon',
         icon: 'smile-beam',
         conditionToShow: !this.isDesktopScreen
+      }, {
+        name: L('Mark unread'),
+        action: 'markAsUnread',
+        icon: 'envelope',
+        conditionToShow: !this.isMsgSender
```
SebinSong: `Mark unread` menu is only enabled on messages that are sent by other users(I believe it does not make much sense to unread messages that I _myself_ wrote/sent).

FYI, Slack does show this menu for messages sent by myself. But what it actually does in this case is, it finds an older message that is sent by other user and then mark that as unread. I feel this seems a bit unnecessary and more importantly, overcomplicates the logic. (feels reluctant because chatroom is already one of the most complicated area in the app IMO) Hope it sounds okay.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -995,6 +1005,64 @@ export default ({
...
       }
     },
+    async markAsUnread ({ messageHash, createdHeight }) {
+      const chatRoomID = this.ephemeral.renderingChatRoomId
+      if (!chatRoomID || !this.isJoinedChatRoom(chatRoomID)) { return }
+
+      // helper functions
+      const getUpdatedUnreadMessages = () => {
+        // This method filters the messages to store in 'unreadMessages' property (eg. messages that mentions me or contains '@all'),
+        // which are reflected as the unread-count badge in the UI.
```
SebinSong, line 1008: This method basically re-determines the messages to display as this numbered-badge UI when user uses 'mark-unread' feature (eg. messages that mentions `@my-username` or `@all`, messages sent in DM, Group proposal notifcation messages etc.):

frontend/views/containers/chatroom/MessageActions.vue
taoeffect: I actually use this feature quite frequently, so would very much appreciate being able to mark messages to myself as unread. 😄 

It gives me the ability to remind myself that I have a TODO or something that I need to do soon and not to forget about it.

frontend/controller/actions/identity-kv.js
```diff
@@ -115,6 +115,30 @@ export default (sbp('sbp/selectors/register', {
...
+        return {
+          ...currentData,
+          [contractID]: {
+            readUntil: { messageHash, createdHeight }, unreadMessages
+          }
+        }
+      }
+
+      const data = await getUpdatedUnreadMessages()
+      if (data) {
```
taoeffect, line 144: The way the other unread actions are written check for `if (data)` because `getUpdatedUnreadMessages` could actually return `null`, but the way this version of it is written, it will never return `null`. So either this `if` should be removed, or `null` should be returned in some special case of `getUpdatedUnreadMessages`.  I have not looked closely enough to decide which should be the case (so please investigate), just noticing this inconsistency.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -978,15 +979,24 @@ export default ({
...
+      messageHash,
+      createdHeight,
+      // 'forceUpdate' flag here is for the rare case where the 'readUntil' value needs to be set to the msg with lower 'createdHeight'.
+      // eg. when the latest message is deleted. (reference: https://github.com/okTurtles/group-income/issues/2729)
+      forceUpdate = false,
+      // 'Mark unread' feature allows user to set 'currentChatRoomReadUntil' to the message they want.
+      // So if user has used this functionality at least once, the chatroom should stop auto-updating the 'readUntil' data in the store (eg. while scrolling),
+      // So that the message that has been marked unread is kept until user leaves the chatroom and re-enter in the future.
+      // Hence, setting below flag to 'true' by default.
+      noUpdateWhenMarkUnreadIsUsed = true
```
taoeffect: I checked the codebase and this `noUpdateWhenMarkUnreadIsUsed` flag isn't used anywhere... so why have it?

frontend/views/containers/chatroom/ChatMain.vue
taoeffect: I see... if you want to have the badge number only reflect the number of @mentions in a chatroom, that's fine, but if a user marks a message in a chatroom as unread then we need to show *something*. I noticed that when none of the messages contain @mentions, then this PR does nothing, which isn't right.

On Slack, in the #random channel, if I mark a message as unread it makes the chatroom bold white in color. So we can either do that (and show the badge reflecting the number of @mentions), or alternatively, we can make it bold white and also count all messages in the counter (up to you). But either way it should do something.

taoeffect (changes requested):
Great work @SebinSong!

I've left a few comments. Also, below I've included feedback from `o4-mini` — **note:** I haven't reviewed this feedback, so it might contain irrelevant/incorrect feedback, especially since I gave it a lot of extra context to review (please check to see if there's anything useful though, and I'd be curious to hear your thoughts as to how it did):

## o4-mini

<details>
Here are a few spots I think could lead to bugs or be improved. I’ve quoted the exact lines and given replacement suggestions.

---

1) **Dangerous mutation / missing guard in removeChatRoomUnreadMessage**  
   
Current code (can throw if `unreadMessages` is `undefined`, and mutates `currentData` in place):  
```js
// frontend/controller/actions/identity-kv.js
...
const index = currentData[contractID]?.unreadMessages.findIndex(msg => msg.messageHash === messageHash)
// NOTE: index could be undefined if unreadMessages is not initialized
if (Number.isInteger(index) && index >= 0) {
  currentData[contractID].unreadMessages.splice(index, 1)
  return currentData
}
return null
...
```

**Problems**  
- If `currentData[contractID]` exists but `unreadMessages` is `undefined`, `.findIndex` throws.  
- Mutating `currentData` in place can lead to subtle bugs if some handler is holding the old reference.

**Suggested fix** – guard against missing array and do an immutable update (using shallow copies):  
```js
// frontend/controller/actions/identity-kv.js
...
const getUpdatedUnreadMessages = async () => {
  const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
  const chatData = currentData[contractID]
  const arr = Array.isArray(chatData?.unreadMessages)
    ? chatData.unreadMessages
    : []
  const idx = arr.findIndex(msg => msg.messageHash === messageHash)
  if (idx < 0) {
    return null
  }
  // build a new object rather than mutating
  return {
    ...currentData,
    [contractID]: {
      ...chatData,
      unreadMessages: [
        ...arr.slice(0, idx),
        ...arr.slice(idx + 1)
      ]
    }
  }
}
...
```

---

2) **Similarly, immutable update in addChatRoomUnreadMessage**  
   
Current (mutates `currentData` in place):  
```js
// frontend/controller/actions/identity-kv.js
if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
  const index = currentData[contractID].unreadMessages.findIndex(msg => msg.messageHash === messageHash)
  if (index === -1) {
    currentData[contractID].unreadMessages.push({ messageHash, createdHeight })
    return currentData
  }
}
```

**Suggested fix** – clone unreadMessages before pushing:  
```js
// frontend/controller/actions/identity-kv.js
if (currentData[contractID]?.readUntil.createdHeight < createdHeight) {
  const chatData = currentData[contractID]
  const arr = Array.isArray(chatData.unreadMessages)
    ? chatData.unreadMessages
    : []
  if (!arr.find(msg => msg.messageHash === messageHash)) {
    return {
      ...currentData,
      [contractID]: {
        ...chatData,
        unreadMessages: arr.concat({ messageHash, createdHeight })
      }
    }
  }
}
```

---

3) **Restrict “Mark unread” to relevant message types**  

Right now in `MessageActions.vue` you have:  
```js
{
  name: L('Mark unread'),
  action: 'markAsUnread',
  icon: 'envelope',
  conditionToShow: !this.isMsgSender
},
```

That will show “Mark unread” on _every_ non‑sender message (including, e.g., notifications). You probably only want it for chat messages (TEXT, INTERACTIVE, POLL). For example:  
```js
// frontend/views/containers/chatroom/MessageActions.vue
{
  name: L('Mark unread'),
  action: 'markAsUnread',
  icon: 'envelope',
  conditionToShow:
    !this.isMsgSender &&
    [MESSAGE_TYPES.TEXT, MESSAGE_TYPES.INTERACTIVE, MESSAGE_TYPES.POLL].includes(this.type)
},
```

---

4) **Close the menu after clicking “Mark unread”**  

In `MessageBase.vue` you bound the event:  
```pug
<message-actions
  …
  @markAsUnread='markAsUnread'
/>
```

But you never call your menu‑closing logic, so the actions menu stays open. You probably want something like:  
```js
// frontend/views/containers/chatroom/MessageBase.vue
methods: {
  …,
  markAsUnread () {
    this.chatMainUtils.markAsUnread({
      messageHash: this.messageHash,
      createdHeight: this.height
    })
    // close the menu (assuming you have a ref or a method)
    this.$refs.messageAction.closeMenu && this.$refs.messageAction.closeMenu()
  }
}
```

---

5) **Optional: clear the “manual mark” flag after post‑action**  

You set  
```js
this.ephemeral.messageHashToMarkUnread = targetMsg.hash
```
and never clear it (other than when switching rooms). That means _all_ future scrolls will silently skip auto‑updating read‑until. If you intended to resume auto‑scroll/updating once you’ve applied your post‑mark logic, you should clear it in your watcher:

```js
// in ChatMain.vue watcher:
watch: {
  currentChatRoomReadUntil (newRU) {
    if (newRU?.messageHash === this.ephemeral.messageHashToMarkUnread) {
      this.$nextTick(() => {
        this.markUnreadPostActions()
        // now allow auto‑updates again
        this.ephemeral.messageHashToMarkUnread = null
      })
    }
  }
}
```

Otherwise, add a comment to clarify that this is deliberate.

---

With these changes you’ll avoid silent crashes, accidental state mutations, confusing menu states, and overly‑broad display of “Mark unread.” Let me know if you need any further context!
</details>

frontend/views/containers/chatroom/MessageActions.vue
SebinSong: @taoeffect 
That requires to remove this logic from L971 - L977 below:

https://github.com/okTurtles/group-income/blob/51937576a77a67cf30711cad5006dbdb75e78c0e/frontend/views/containers/chatroom/ChatMain.vue#L960-L979

`setStartNewMessageIndex` method here is currently used to determine the initial scroll position using `currentChatRoomReadUntil` vallue and also display below `--- new ---` divider UI when needed.

image

Instead of simply going directly to the message stored in `currentChatRoomReadUntil`, currently it finds **a newer message that is sent by other user** and scroll to it. In order to implement your requirement, we need to change this behaviour. Should we be good to go ahead with it?

taoeffect: I'm not sure I fully understand the question or implications.... but feel free to go ahead and make it work. If you're saying it will break some other functionality though, please help me understand better what it will break.

SebinSong: @taoeffect 
I meant the behavior explained by below comment will be overriden. I'm not the author of this logic so don't know the full context of it either..

```
// NOTE: When the user switches channel before the message is not fully processed, 
//       (in other words, until this.variant(msg) === 'sent') 
//       the chatroomReadUntil position would not be saved correctly 
//       because the pending messages could not be saved in state. 
//       Considering those such cases, we shoud set 'isNew' position for the messages 
//       only whose sender is not ourselves. 
```

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: >  if you want to have the badge number only reflect the number of @mentions in a chatroom, that's fine ...

FYI That is what the badge number shown means currently in the app (not a new behavior I created in this PR).

> I noticed that when none of the messages contain @mentions, then this PR does nothing, which isn't right.

Not true.. It currently does do something which is displaying/positioning this `--- New ---` UI element to the message marked unread:

image

I agree the chatroom title having bold-style text makes things more explicit though. So let me try making this change.

frontend/controller/actions/identity-kv.js
SebinSong: @taoeffect Right. In this case, we can add a logic to check if the existing message hash in `readUntil` property is the same as the one requested (eg. User trying to mark the same message as unread multiple times.) and then returning `null` if that's the case.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: Had an idea for this but it doesn't seem to be needed. will remove it.

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: @taoeffect 
Made updates accordingly:



Green-box here is for the case when user has used `mark-unread` feature and there is no mentions (or proposal notification etc..) and below it is for the case where the unread messages does contain mentions.

---

BTW just FYI, `getUpdatedUnreadMessages` logic(determining the messages to display as the number badge) here were created based on the existing logics which are:

https://github.com/okTurtles/group-income/blob/51937576a77a67cf30711cad5006dbdb75e78c0e/frontend/controller/actions/chatroom.js#L28-L29

https://github.com/okTurtles/group-income/blob/51937576a77a67cf30711cad5006dbdb75e78c0e/frontend/controller/actions/chatroom.js#L46

https://github.com/okTurtles/group-income/blob/51937576a77a67cf30711cad5006dbdb75e78c0e/frontend/model/notifications/messageReceivePostEffect.js#L33-L39

frontend/views/containers/chatroom/ChatMain.vue
SebinSong: Removed it.

SebinSong:

I also did take a look at the `o4-mini` suggestion and adopted **1)** from it. Below are my assessment for its review.

1): OK.
2), 3): Unnecessary
4): Incorrect information
5): The behavior it thinks needs to be fixed is actually intentional.    

It does feel like another developer reviewing my work though (but without thorough investigation). So I feel the tool can be helpful.

taoeffect (changes requested):
Nice work @SebinSong, cool how the chatrooms are made bold now. 😎 👍 

I noticed though that I cannot mark DMs to myself as unread though, the menu doesn't appear:

Screenshot 2025-04-24 at 9 54 44 AM

frontend/controller/actions/identity-kv.js
```diff
@@ -115,6 +115,37 @@ export default (sbp('sbp/selectors/register', {
...
   },
+  'gi.actions/identity/kv/markAsUnread': ({ contractID, messageHash, createdHeight, unreadMessages }: {
+    contractID: string,
+    messageHash: string,
+    createdHeight: number,
+    unreadMessages: Array<{ messageHash: string, createdHeight: number }>
+  }) => {
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
+      const getUpdatedUnreadMessages = async () => {
+        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
```
corrideat, line 126: This is OK, but will need to be updated after #2742 is merged (the signature will be updated and the fetch will no longer be needed)

frontend/controller/actions/identity-kv.js
```diff
@@ -115,6 +115,37 @@ export default (sbp('sbp/selectors/register', {
...
+          ...currentData,
+          [contractID]: {
+            readUntil: { messageHash, createdHeight, isManuallyMarked: true },
+            unreadMessages
+          }
+        }
+      }
+
+      const data = await getUpdatedUnreadMessages()
+      if (data) {
```
corrideat, line 144: This here too might need to be updated.

frontend/controller/actions/identity-kv.js
```diff
@@ -115,6 +115,37 @@ export default (sbp('sbp/selectors/register', {
...
+      const getUpdatedUnreadMessages = async () => {
+        const currentData = await sbp('gi.actions/identity/kv/fetchChatRoomUnreadMessages')
+        const existingReadUntil = currentData[contractID]?.readUntil
+
+        // If the requested mark-unread hash has already been set, ignore it.
+        if (existingReadUntil &&
+          existingReadUntil.isManuallyMarked &&
+          existingReadUntil?.messageHash === messageHash) { return null }
+
+        return {
```
corrideat, line 134: And this return will need to be `[..., etag]`

corrideat (commented):
Added comments to the parts that'll conflict with https://github.com/okTurtles/group-income/pull/2742 and will need to be updated.

SebinSong:
@taoeffect 
It does display the `mark unread` menu in chatroom to myself:





Did you tested it with the latest commit?

frontend/controller/actions/identity-kv.js
SebinSong: @corrideat Thanks for a heads up. It seems that PR has not been merged yet. Will make these updates if that goes earlier than this PR.

taoeffect:
Oops, you're right @SebinSong! I must've been working off of old changes, my mistake, sorry about that!

taoeffect (approved):
Amazing work @SebinSong! Great job, just tested it and verified everything works great!

PR #2756, opened 2025-04-20
fetch Chelonia option

corrideat:
This is based on PR 'Offline cache' #2508, addressing issue #2503. The reason for this separate PR is that it might take a while to have #2508 ready, but these changes are generally useful.

These changes add a `fetch` option to Chelonia to override the global `fetch`. This is useful in many contexts, such as for debugging, for adding custom logic and other things. In PR #2508, this option is used for caching some API calls (since the network activity happens inside of the SW, the `fetch` event isn't used)

### DIFF
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -1254,7 +1254,7 @@ export default (sbp('sbp/selectors/register', {
       throw new TypeError('A contract ID must be provided')
     }
 
-    const response = await fetch(`${this.config.connectionURL}/ownResources`, {
+    const response = await this.config.fetch(`${this.config.connectionURL}/ownResources`, {
       method: 'GET',
       signal: this.abortController.signal,
       headers: new Headers([
@@ -1289,7 +1289,7 @@ export default (sbp('sbp/selectors/register', {
         throw new TypeError(`Either a token or a billable contract ID must be provided for ${cid}`)
       }
 
-      const response = await fetch(`${this.config.connectionURL}/deleteContract/${cid}`, {
+      const response = await this.config.fetch(`${this.config.connectionURL}/deleteContract/${cid}`, {
         method: 'POST',
         signal: this.abortController.signal,
         headers: new Headers([
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -5,6 +5,7 @@ import '@sbp/okturtles.events'
 import sbp from '@sbp/sbp'
 import { handleFetchResult } from '~/frontend/controller/utils/misc.js'
 import { cloneDeep, delay, difference, has, intersection, merge, randomHexString, randomIntFromRange } from 'turtledash'
+import { createCID, parseCID } from '~/shared/functions.js'
 import { NOTIFICATION_TYPE, createClient } from '~/shared/pubsub.js'
 import type { SPKey, SPOpActionUnencrypted, SPOpContract, SPOpKeyAdd, SPOpKeyDel, SPOpKeyRequest, SPOpKeyRequestSeen, SPOpKeyShare, SPOpKeyUpdate } from './SPMessage.js'
 import type { Key } from '@chelonia/crypto'
@@ -210,6 +211,7 @@ export default (sbp('sbp/selectors/register', {
       },
       whitelisted: (action: string): boolean => !!this.whitelistedActions[action],
       reactiveSet: (obj, key, value) => { obj[key] = value; return value }, // example: set to Vue.set
+      fetch: (...args) => fetch(...args),
       reactiveDel: (obj, key) => { delete obj[key] },
       // acceptAllMessages disables checking whether we are expecting a message
       // or not for processing
@@ -317,6 +319,7 @@ export default (sbp('sbp/selectors/register', {
   'chelonia/config': function () {
     return {
       ...cloneDeep(this.config),
+      fetch: this.config.fetch,
       reactiveSet: this.config.reactiveSet,
       reactiveDel: this.config.reactiveDel
     }
@@ -801,7 +804,7 @@ export default (sbp('sbp/selectors/register', {
     //
     //        // The following could take a long time. We want Chelonia
     //        // to still work and process events as normal.
-    //        return fetch(profilePictureUrl).then(doSomeWorkWithTheFile)
+    //        return this.config.fetch(profilePictureUrl).then(doSomeWorkWithTheFile)
     //      })
     //    }
     return sbp('chelonia/private/queueEvent', contractID, ['chelonia/private/noop']).then(() => sbp('chelonia/private/queueEvent', 'public:' + contractID, sbpInvocation))
@@ -1066,8 +1069,26 @@ export default (sbp('sbp/selectors/register', {
       return state
     })
   },
+  'chelonia/out/entry': async function (cid: string, { code }: { code?: number } = {}) {
+    const parsedCID = parseCID(cid)
+    if (Number.isInteger(code)) {
+      if (parsedCID.code !== code) {
+        throw new Error('Invalid CID content type')
+      }
+    }
+    const local = await sbp('chelonia.db/get', cid)
+    if (local) return local
+    const url = `${this.config.connectionURL}/file/${cid}`
+    const data = await this.config.fetch(url, { signal: this.abortController.signal }).then(handleFetchResult('text'))
+    const ourHash = createCID(data, parsedCID.code)
+    if (ourHash !== cid) {
+      throw new Error(`expected hash ${cid}. Got: ${ourHash}`)
+    }
+    await sbp('chelonia.db/set', cid, data)
+    return data
+  },
   'chelonia/out/latestHEADInfo': function (contractID: string) {
-    return fetch(`${this.config.connectionURL}/latestHEADinfo/${contractID}`, {
+    return this.config.fetch(`${this.config.connectionURL}/latestHEADinfo/${contractID}`, {
       cache: 'no-store',
       signal: this.abortController.signal
     }).then(handleFetchResult('json'))
@@ -1089,7 +1110,7 @@ export default (sbp('sbp/selectors/register', {
     let reader: ReadableStreamReader
     const s = new ReadableStream({
       start: async (controller) => {
-        const first = await fetch(`${this.config.connectionURL}/file/${startHash}`, { signal: this.abortController.signal }).then(handleFetchResult('text'))
+        const first = await this.config.fetch(`${this.config.connectionURL}/file/${startHash}`, { signal: this.abortController.signal }).then(handleFetchResult('text'))
         const deserializedHEAD = SPMessage.deserializeHEAD(first)
         if (deserializedHEAD.contractID !== contractID) {
           controller.error(new Error('chelonia/out/eventsBetween: Mismatched contract ID'))
@@ -1642,7 +1663,7 @@ export default (sbp('sbp/selectors/register', {
         data,
         meta: key
       })
-      const response = await fetch(`${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`, {
+      const response = await this.config.fetch(`${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`, {
         headers: new Headers([[
           'authorization', buildShelterAuthorizationHeader.call(this, contractID)
         ]]),
@@ -1668,7 +1689,7 @@ export default (sbp('sbp/selectors/register', {
     }
   },
   'chelonia/kv/get': async function (contractID: string, key: string) {
-    const response = await fetch(`${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`, {
+    const response = await this.config.fetch(`${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`, {
       headers: new Headers([[
         'authorization', buildShelterAuthorizationHeader.call(this, contractID)
       ]]),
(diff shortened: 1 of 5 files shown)


shared/domains/chelonia/internals.js
```diff
@@ -251,25 +251,15 @@ export default (sbp('sbp/selectors/register', {
...
     }
     const contractInfo = (this.config.contracts.defaults.preferSlim && body.contractSlim) || body.contract
     console.info(`[chelonia] loading contract '${contractInfo.file}'@'${body.version}' from manifest: ${manifestHash}`)
-    const source = await fetch(`${this.config.connectionURL}/file/${contractInfo.hash}`, { signal: this.abortController.signal })
-      .then(handleFetchResult('text'))
-    const sourceHash = createCID(source, multicodes.SHELTER_CONTRACT_TEXT)
-    if (sourceHash !== contractInfo.hash) {
-      throw new Error(`bad hash ${sourceHash} for contract '${contractInfo.file}'! Should be: ${contractInfo.hash}`)
-    }
+    const source = await sbp('chelonia/out/entry', contractInfo.hash, { code: multicodes.SHELTER_CONTRACT_TEXT })
```
corrideat: This new addition also DRY'es some of the code.

shared/domains/chelonia/chelonia.js
```diff
@@ -1066,8 +1069,26 @@ export default (sbp('sbp/selectors/register', {
       return state
     })
   },
+  'chelonia/out/entry': async function (cid: string, { code }: { code?: number } = {}) {
```
taoeffect: This is a neat new addition, but could you come up with a better name for it?

Right now it reads like it's sending an `SPMessage`, and that's not what it's doing.

shared/domains/chelonia/chelonia.js
```diff
@@ -1066,8 +1069,26 @@ export default (sbp('sbp/selectors/register', {
...
+    const parsedCID = parseCID(cid)
+    if (Number.isInteger(code)) {
+      if (parsedCID.code !== code) {
+        throw new Error('Invalid CID content type')
+      }
+    }
+    const local = await sbp('chelonia.db/get', cid)
+    if (local) return local
```
taoeffect: o4-mini-high:

### 1) chelonia/out/entry’s “if (local)” drops empty‐string or zero‐length values

```js
// shared/domains/chelonia/chelonia.js
+  'chelonia/out/entry': async function (cid: string, { code }: { code?: number } = {}) {
+    const parsedCID = parseCID(cid)
+    …
+    const local = await sbp('chelonia.db/get', cid)
+    if (local) return local
+    …
+  }
```

**Problem:** if your DB ever stores an empty string (`""`) or some other legitimately _falsy_ value, you’ll always fall back to the network rather than returning your cache.

**Suggested fix:** check explicitly for `null`/`undefined`, not truthiness:

```js
'chelonia/out/entry': async function (cid: string, { code }: { code?: number } = {}) {
  // first check the cache
  const local = await sbp('chelonia.db/get', cid)
  if (local != null) {
    return local
  }

  // now validate the CID
  const parsedCID = parseCID(cid)
  if (Number.isInteger(code) && parsedCID.code !== code) {
    throw new Error('Invalid CID content type')
  }

  const url = `${this.config.connectionURL}/file/${cid}`
  const data = await this.config.fetch(url, { signal: this.abortController.signal })
    .then(handleFetchResult('text'))

  // verify integrity
  const ourHash = createCID(data, parsedCID.code)
  if (ourHash !== cid) {
    throw new Error(`expected hash ${cid}. Got: ${ourHash}`)
  }

  await sbp('chelonia.db/set', cid, data)
  return data
}
```

shared/domains/chelonia/internals.js
```diff
@@ -374,7 +364,7 @@ export default (sbp('sbp/selectors/register', {
         // with other client's clocks.
         // See: https://github.com/okTurtles/group-income/issues/531
         try {
-          const response = await fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
+          const response = await this.config.fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
```
taoeffect, line 367: Pretty sure o4-mini-high is wrong about this, right?

### 4) Embedded `this.config.fetch` inside `new Function` / `sesImportVM`

```diff
// shared/domains/chelonia/internals.js
-          const response = await fetch(`${this.config.connectionURL}/time`, { … })
+          const response = await this.config.fetch(`${this.config.connectionURL}/time`, { … })
```

This snippet lives inside a template‐string passed to `new Function`.  **At runtime** the `this` in that sandboxed code almost certainly will **not** point back to your SBP domain, so `this.config` will be `undefined`.  

You’ll either need to:

1. **Inject** your `fetch` into the sandbox closure explicitly (e.g. add a `const fetch = parentFetch` at the top), or  
2. **Bind** the generated function to the correct `this` at invocation time, e.g.  
   ```js
   const fnFactory = new Function(codeAsString)
   const sandboxFactory = fnFactory.call(this)       // `this` is your domain
   ```
   so that inside `fetchServerTime`, `this.config` is available.

Without that you’ll get `TypeError: Cannot read property 'config' of undefined`.

shared/domains/chelonia/chelonia.js
```diff
@@ -1066,8 +1069,26 @@ export default (sbp('sbp/selectors/register', {
       return state
     })
   },
+  'chelonia/out/entry': async function (cid: string, { code }: { code?: number } = {}) {
+    const parsedCID = parseCID(cid)
+    if (Number.isInteger(code)) {
+      if (parsedCID.code !== code) {
+        throw new Error('Invalid CID content type')
+      }
+    }
```
taoeffect: Should we throw if `code` isn't an integer?

Also, Gemini 2.5 Pro Preview has a decent suggestion to improve the error here, although I don't think the second error for the hash needs to be modified:

1.  **`chelonia/out/entry` Error Specificity**

    In `chelonia/out/entry`, the error thrown when the content type code doesn't match is a generic `Error`. It would be slightly more informative to include the expected and actual codes. Similarly, the hash mismatch error could be more specific or use a custom error type if data integrity failures need special handling elsewhere.

    ```diff
    --- a/shared/domains/chelonia/chelonia.js
    +++ b/shared/domains/chelonia/chelonia.js
    @@ -1053,12 +1056,12 @@ export default (sbp('sbp/selectors/register', {
       const parsedCID = parseCID(cid)
       if (Number.isInteger(code)) {
         if (parsedCID.code !== code) {
    -      throw new Error('Invalid CID content type')
    +      throw new Error(`Invalid CID content type. Expected ${code}, got ${parsedCID.code}`)
         }
       }
       const local = await sbp('chelonia.db/get', cid)
       if (local) return local
       const url = `${this.config.connectionURL}/file/${cid}`
       const data = await this.config.fetch(url, { signal: this.abortController.signal }).then(handleFetchResult('text'))
       const ourHash = createCID(data, parsedCID.code)
       if (ourHash !== cid) {
    -      throw new Error(`expected hash ${cid}. Got: ${ourHash}`)
    +      throw new Error(`CID hash mismatch for ${cid}. Calculated hash: ${ourHash}`) // Or potentially: new ChelErrorDataIntegrity(...)
       }
       await sbp('chelonia.db/set', cid, data)
       return data

    ```

shared/domains/chelonia/time-sync.js
```diff
@@ -21,7 +21,7 @@ const syncServerTime = async function () {
   // Get our current monotonic time
   const startTime = performance.now()
   // Now, ask the server for the time
-  const time = await fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
+  const time = await this.config.fetch(`${this.config.connectionURL}/time`, { signal: this.abortController.signal })
```
taoeffect, line 24: Just comment: I think this is safe because `syncServerTime.call(this)` is used elsewhere... double-check though.

shared/domains/chelonia/chelonia.js
```diff
@@ -1066,8 +1069,26 @@ export default (sbp('sbp/selectors/register', {
...
+    const local = await sbp('chelonia.db/get', cid)
+    if (local) return local
+    const url = `${this.config.connectionURL}/file/${cid}`
+    const data = await this.config.fetch(url, { signal: this.abortController.signal }).then(handleFetchResult('text'))
+    const ourHash = createCID(data, parsedCID.code)
+    if (ourHash !== cid) {
+      throw new Error(`expected hash ${cid}. Got: ${ourHash}`)
+    }
+    await sbp('chelonia.db/set', cid, data)
```
taoeffect, line 1090: The use of a cache is a new addition... I suppose it is safe to use because it is based on `cid`s. Just making a note though as this is new logic being added that wasn't there before, not just DRYing.

taoeffect (changes requested):
Great work @corrideat! Left some feedback!

shared/domains/chelonia/chelonia.js
corrideat: Good point. Well, this does come from the offline cache PR so that's why. I didn't notice it as I made this PR it added new functionality.

taoeffect: Per the call: in 99% of cases the lightweight client is true and this therefore does nothing.

Therefore at minimum this needs to be commented here, but also IMO it would be better to not do this type of caching here anyway and rely only on the browser cache (whether it works or not), because then you'd be able to compare the two values and not have to worry about the comment above regarding `if (local)` ... i.e. you'd always verify that the CID corresponds to `""` and therefore would have to have an `if (local != null)` check or any corresponding cache code here. IMO that would simplify this code and so we should do it. I don't think it would affect performance much.

In the offline cache PR I think you said also do the caching setup in either the `fetch` implementation or `chelonia.db/get` anyway, so therefore this caching code would be possibly redundant?

corrideat: I'm not sure this really is redundant, as `'chelonia/private/db/addEntry'` has `await sbp('chelonia.db/set', entry.hash(), entry.serialize())`, which similarly does nothing for the lightweight client. Adding contract manifests and source to the DB when events are added seemed consistent, since events can't be processed without the corresponding contract.

shared/domains/chelonia/chelonia.js
corrideat: chelonia/out/fetchEntry?

shared/domains/chelonia/internals.js
corrideat: I think this might be right, but we probably shouldn't spend too much time on this since this code is commented out anyhow. When and if this comes into use, we need to ensure that references work properly, and IMHO it's better to have `this.config.fetch` and have that fail (so that we can correct it) than have `fetch` and have it potentially work.

shared/domains/chelonia/time-sync.js
corrideat: What are you referring to as 'safe'? Are you saying whether `this.config` is defined? In that case, it's safe, and the previous code already had `this.config.connectionURL`.

shared/domains/chelonia/chelonia.js
taoeffect: The selector should be descriptive of what the code is doing — without confusing the reader. The word `entry` is verboten here as it refers to `SPMessage` in many other places, and this has nothing to do with `SPMessage`.

So something like `chelonia/getFile` or `chelonia/fetchVerifiedFile`.

I don't like `out` because the meaning of `chelonia/out` is _sending something to the server_, which this code isn't doing.

I think it make sense to always verify the CID, as that's what the code did before, so `chelonia/fetchVerifiedFile` or `chelonia/fetchFile` would make sense, especially with the other comment here:

https://github.com/okTurtles/group-income/pull/2756#discussion_r2052868895

shared/domains/chelonia/chelonia.js
taoeffect: Regardless of everything else, I think we need to code needs to get rid of these lines:

```js
    const local = await sbp('chelonia.db/get', cid)
    if (local) return local
```

The previous code didn't have this, and the previous code also always verified that the CID matched — we need to restore that functionality (of verifying the CID always matches).

shared/domains/chelonia/chelonia.js
taoeffect: My thinking has evolved a bit here, now I'm for removing these lines completely: https://github.com/okTurtles/group-income/pull/2756#discussion_r2052892233

shared/domains/chelonia/chelonia.js
corrideat: Fine about `entry`. I think `file` is also confusing for similar reasons, but we can go with that, or verifiedFile.

Re. `/out/`, it _is_ sending something to the server (just not storing something on the sever). We already have `'chelonia/out/latestHEADInfo'`.

shared/domains/chelonia/chelonia.js
corrideat: But the current logic does verify the CID matches, since they had to to get added to the DB.

shared/domains/chelonia/chelonia.js
taoeffect: > Re. /out/, it is sending something to the server (just not storing something on the sever). We already have 'chelonia/out/latestHEADInfo'.

That's a good point...

We should also consider then the existing selector `'chelonia/fileDownload'`. Having this selector be named `chelonia/fetchVerifiedFile` would be confusing as it would somewhat conflict with that selector.

Let me know if you have any additional ideas based on this...

shared/domains/chelonia/chelonia.js
corrideat: `if (Number.isInteger(code)) {` is meant to allow passing undefined or null to skip this check. Maybe I could check for that instead?

shared/domains/chelonia/chelonia.js
taoeffect: Ah... good point. OK feel free to disregard that, and/or perhaps add a comment to that effect.

shared/domains/chelonia/time-sync.js
taoeffect: Yeah 👍

taoeffect (approved):
Nice work @corrideat !

PR #2757, opened 2025-04-21
#2738 - Unable to upload extensionless files

### LINKED ISSUE #2738: Unable to upload extensionless files
### Problem

In the chat, it's not possible to upload files that don't have an extension. Nothing seems to happen when such files are selected (i.e., same as if the upload window had been cancelled0.

### Solution

Identify the issue and fix it, so that all files are allowed.

SebinSong:
closes #2738 

**[1. Extensionless files are detected]**

https://github.com/user-attachments/assets/3e043f92-58c9-4626-af7d-b6709ba863c6

**[2. Drag&dropping a folder is silently ignored]**

https://github.com/user-attachments/assets/34a16ff2-954b-474a-a9d5-000fc3452783

### DIFF
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1202,8 +1202,8 @@ export default ({
         const items = e.dataTransfer.items
 
         if (items?.length) {
-          // 'drag-end' event of the browsers works for a folder too and turns it into a File() instance.
-          // Filtering out directories here by using .webkitGetAsEntry() method provided via DataTransferItem API.
+          // 'drag-end' event of the browsers detects a folder too and turns it into a File() instance, which is an unwanted behaviour.
+          // So filtering out directories here by using .webkitGetAsEntry() method provided via DataTransferItem API.
           // (Reference: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry)
           const detectedFiles = []
 
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1202,21 +1202,23 @@ export default ({
         const items = e.dataTransfer.items
 
         if (items?.length) {
-          // 'drag-end' event of the browsers detects a folder too and turns it into a File() instance, which is an unwanted behaviour.
-          // So filtering out directories here by using .webkitGetAsEntry() method provided via DataTransferItem API.
+          // 'drag-end' event of the browsers detects files as well as folders, and we only want
+          // files. The kind field and the getAsFile() methods are not helpful at telling them
+          // apart, as both will work as a 'File'. The only quick and reliable way to detect files
+          // is using webkitGetAsEntry.
           // (Reference: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry)
           const detectedFiles = []
 
           for (let i = 0; i < items.length; i++) {
             const item = items[i]
-            const entry = item.webkitGetAsEntry?.()
+            const entry = item.getAsEntry?.() || item.webkitGetAsEntry?.()
 
             if (entry) {
               entry.isFile && detectedFiles.push(item.getAsFile())
             } else {
               const file = item.getAsFile()
 
-              if (file.type !== '') {
+              if (file && file.type !== '') {
                 detectedFiles.push(file)
               }
             }
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1216,6 +1216,7 @@ export default ({
             if (entry) {
               entry.isFile && detectedFiles.push(item.getAsFile())
             } else {
+              // NOTE: if an old browser that does not support webkitGetAsEntry (eg. Internet Explorers), this fallback might accept a directory.
               const file = item.getAsFile()
 
               if (file && file.type !== '') {
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -933,7 +933,7 @@ export default ({
         Vue.set(this.messageState, 'contract', state)
       }
     },
-    scrollToIntialPosition () {
+    scrollToInitialPosition () {
       const hashTo = this.ephemeral.initialScroll.hash
       if (hashTo) {
         const scrollingToSpecificMessage = this.$route.query?.mhash === hashTo
@@ -1175,7 +1175,7 @@ export default ({
           // In that case, we should defer 'auto-scrolling to the initial-position' until those additional messages are rendered.
           // This can be achieved by calling 'scrollToInitialPosition' here with setTimeout(),
           // and calling clearTimeout() at the start of infiniteHandler().
-          this.ephemeral.initialScroll.timeoutId = setTimeout(this.scrollToIntialPosition, 150)
+          this.ephemeral.initialScroll.timeoutId = setTimeout(this.scrollToInitialPosition, 150)
         }
       })
     },
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1199,8 +1199,32 @@ export default ({
       if (this.dndState.isActive) {
         this.dndState.isActive = false
 
-        e?.dataTransfer.files?.length &&
-          this.$refs.sendArea.fileAttachmentHandler(e?.dataTransfer.files, true)
+        const items = e.dataTransfer.items
+
+        if (items?.length) {
+          // 'drag-end' event of the browsers works for a folder too and turns it into a File() instance.
+          // Filtering out directories here by using .webkitGetAsEntry() method provided via DataTransferItem API.
+          // (Reference: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry)
+          const detectedFiles = []
+
+          for (let i = 0; i < items.length; i++) {
+            const item = items[i]
+            const entry = item.webkitGetAsEntry?.()
+
+            if (entry) {
+              entry.isFile && detectedFiles.push(item.getAsFile())
+            } else {
+              const file = item.getAsFile()
+
+              if (file.type !== '') {
+                detectedFiles.push(file)
+              }
+            }
+          }
+
+          detectedFiles.length &&
+            this.$refs.sendArea.fileAttachmentHandler(detectedFiles, true)
+        }
       }
     },
     processChatroomSwitch: debounce(async function () {
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -719,8 +719,6 @@ export default ({
       this.$refs.fileAttachmentInputEl.click()
     },
     fileAttachmentHandler (filesList, appendItems = false) {
-      // filter out all entries whose types are unrecognisable.
-      // (eg. when a folder or Mac application is drag&dropped)
       filesList = Array.from(filesList)
 
       // User clicked 'Cancel button'.
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -719,15 +719,12 @@ export default ({
       this.$refs.fileAttachmentInputEl.click()
     },
     fileAttachmentHandler (filesList, appendItems = false) {
-      const containsExtension = v => /.+\.\w+$/.test(v) // check if a entry name ends with a file extension.
-      const isDirectory = entry => entry.type === '' && !containsExtension(entry.name)
+      // filter out all entries whose types are unrecognisable.
+      // (eg. when a folder or Mac application is drag&dropped)
+      filesList = Array.from(filesList)
 
-      filesList = Array.from(filesList).filter(entry => !isDirectory(entry)) // filter all directories from attachments first.
-      if (!filesList.length) {
-        // NOTE: it's either that user clicked 'Cancel button' or  anything that is not a file (e.g. directory) has been attached.
-        //       no action is needed in these cases.
-        return
-      }
+      // User clicked 'Cancel button'.
+      if (!filesList.length) { return }
 
       const list = appendItems && this.hasAttachments
         ? [...this.ephemeral.attachments]


frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1199,8 +1199,32 @@ export default ({
...
         this.dndState.isActive = false
 
-        e?.dataTransfer.files?.length &&
-          this.$refs.sendArea.fileAttachmentHandler(e?.dataTransfer.files, true)
+        const items = e.dataTransfer.items
+
+        if (items?.length) {
+          // 'drag-end' event of the browsers detects a folder too and turns it into a File() instance, which is an unwanted behaviour.
+          // So filtering out directories here by using .webkitGetAsEntry() method provided via DataTransferItem API.
+          // (Reference: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry)
```
SebinSong, line 1209: `webkitGetAsEntry` method here seems to have a reliable level of supports for desktop browsers:
https://caniuse.com/mdn-api_datatransferitem_webkitgetasentry

(it worked well for Chrome/Firefox/Safari in my Mac)

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -719,15 +719,12 @@ export default ({
       this.$refs.fileAttachmentInputEl.click()
     },
     fileAttachmentHandler (filesList, appendItems = false) {
-      const containsExtension = v => /.+\.\w+$/.test(v) // check if a entry name ends with a file extension.
-      const isDirectory = entry => entry.type === '' && !containsExtension(entry.name)
```
SebinSong, line 723: Removed this here and the logic to filter out directories has been moved to `drag-end` event listener. (As it's only an issue with drag&drop feature)

SebinSong:
The PR is ready for review.

@corrideat FYI. if you would like to test the fix.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1199,8 +1199,32 @@ export default ({
...
+          const detectedFiles = []
+
+          for (let i = 0; i < items.length; i++) {
+            const item = items[i]
+            const entry = item.webkitGetAsEntry?.()
```
corrideat: I didn't realise how messy APIs browser APIs are regarding this. It looks like the code below does a good job.

I wonder if we could simplify the logic here and at the same time maybe future proof this code here. I was looking at `webkitGetAsEntry` at MDN and indeed it seems to have good support, despite it being prefixed and non-standard. It, however, has this note on MDN:

> Note: This function is implemented as webkitGetAsEntry() in non-WebKit browsers including Firefox at this time; it may be renamed to getAsEntry() in the future, so you should code defensively, looking for both.

I see that there's a fallback down below for the case when `webkitGetAsEntry` isn't defined, which is good. However, looking at `getAsFile`, I see the following the spec: 'If [the drag data item kind](https://html.spec.whatwg.org/multipage/dnd.html#the-drag-data-item-kind) is not File, then return `null`.'

Since we want to keep files and filter out eveything else, it looked like this could be simplified to the following:

```js
for (let i = 0; i < items.length; i++) {
  const item = items[i]
  const file = item.getAsFile()
  if (file) detectedFiles.push(file)
}

detectedFiles.length &&
  this.$refs.sendArea.fileAttachmentHandler(detectedFiles, true)
```

However, I tested this and `item.getAsFile()` does _not_ work as described, and directories are converted to `File` objects.

Testing the fallback (`else` part), I found the following areas for improvement with line 1219 (`if (file.type !== '')`):

  1. `file` could be `null`, so the check should account for this (e.g., `file?.type`)
  2. Adding to the seemingly wrong behaviour of directories being reported as 'files', this mechanism isn't reliable either because it has the same bug as the previous code, namely that extensionless files will have an empty type and that directories with an 'extension' (try a directory named `a.txt`) will (!) have a type. Really odd.

However, it seems like I found a way to reliably detect directories, but which unfortunately is `async`. `File.prototype.arrayBuffer` will throw a `DOMException` for directories. What a mess this all is. Based on all of this, these are my suggestions:

1. Can we try to use both `getAsEntry` and `webkitGetAsEntry` as MDN suggests? (I'd do it in that order, `getAsEntry` first, but either way is fine)? I'm not so happy with using non-standard features but it really seems like it's one of the best ways of doing this.
2. Can we make this function `async`? If the answer is yes, perhaps we can improve the fallback to try `file.arrayBuffer` and see if it throws (directory) or not (file)? Otherwise, maybe we can remove the fallback entirely, since `webkitGetAsEntry` seems to be well-supported and does what's needed.
3. Technically, it'd be possible to simplify the `for` loop to something like this, without needing a fallback and without `webkitGetAsEntry`, but it probably performs worse and requires the function to be async:

```js
for (let i = 0; i < items.length; i++) {
  const item = items[i]
  const file = item.getAsFile()
  if (file) {
    try {
      await file.arrayBuffer()
      detectedFiles.push(file)
    } catch {
      // ignore
    }
  }
}

```

corrideat (changes requested):
I thought this was going to be a simple PR to review, but then I was met with `DataTransferItem` and all of its complexity. Overall, the PR looks fine and I think the implementation is quite reasonable given the constraints of `DataTransferItem`.

See my comment on `ChatMain.vue` (which was meant to be shorter) for some changes I'd consider (items 1. & 2.) Alternatively, 3. would work instead of 1. & 2. but requires making the function async and may perform worse.

Other than those minor changes, the rest of the PR looks fine and well-done.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: I'll add another change request:

> // 'drag-end' event of the browsers detects a folder too and turns it into a File() instance, which is an unwanted behaviour.

That comment is a good summary, but I wasn't entirely sure what the last part meant before looking into the different APIs and how they work. For this reason, would you consider expanding the comment to explicitly state that `webkitGetAsEntry` is the only reliable way of doing this because `DataTransferItem` has a `kind` attribute, but that's `file` for both files and directories and that `.getAsFile()` has the same issue (I think that's what '_turns it into a File() instance_') refers to.

For example, consider this comment (feel free to reword as appropriate):

> // 'drag-end' event of the browsers detects files as well as folders, and we only want
> // files. The `kind` field and the `getAsFile()` methods are not helpful at telling them
> // apart, as both will work as a 'File'. The only quick and reliable way to detect files
> // is using `webkitGetAsEntry`.

SebinSong: Thanks for thorough review @corrideat . (Quite informative too actually) I've made below updates from your feedbacks:

- file could be null, so the check should account for this (e.g., file?.type)
- Can we try to use both getAsEntry and webkitGetAsEntry as MDN suggests?
- Updating the summary comments to the above.

Will stick to using `webkitGetAsEntry` here though (as opposed to using the `async` solution) as it reliably achieves what we want.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -1199,8 +1199,34 @@ export default ({
...
+          for (let i = 0; i < items.length; i++) {
+            const item = items[i]
+            const entry = item.getAsEntry?.() || item.webkitGetAsEntry?.()
+
+            if (entry) {
+              entry.isFile && detectedFiles.push(item.getAsFile())
+            } else {
+              const file = item.getAsFile()
+
+              if (file && file.type !== '') {
```
corrideat, line 1222: Thanks for addressing the feedback, it looks much better now. Still, I'm not sure about this part here. I guess you can leave since it should never be reached, but as I mentioned, this doesn't seem like a reliable way to detect directories (a directory named `dir.txt` will have `type` equal to `text/plain`).

If you decide to leave this fallback, would you mind adding a comment here too, saying that we do it as a fallback but that it's not 100% accurate as some files have empty `type` and some directories have non-empty `type`?

SebinSong: https://github.com/user-attachments/assets/7a8a7e4f-0306-4414-9a77-4bde097bbfeb

@corrideat Have you tested the fix locally by yourself though? it does detect/ignore `dir.txt` correctly too(by `entry.isFile && ...`. In this case `entry.isDirectory` is `true`), as you can see from the video.

corrideat: Yes, I did, sorry if that wasn't clear, and it works. I'm referring to the fallback (if you take the `else` branch). (That's also why I said '_it should never be reached_', meaning that since all browsers apparently support `webkitGetAsEntry`, the else shouldn't be reachable)

The screenshot below is of a directory called `a.txt` and another one called `adirectory.d`

image

image

You can see the fallback by replacing line 1216 with `if (isNaN(1) && entry) {`, or by commenting out the `if` and leaving out just the `else` (i.e., commenting out 1216-1218 and 1224 (or 1217-1218)).

SebinSong: @corrideat Thanks for clarifying in details.

> That's also why I said 'it should never be reached', meaning that since all browsers apparently support webkitGetAsEntry.

I agree with this too, so let me just leave it as is.

corrideat: Right, but then can you please add a comment saying that the fallback can provide incorrect results? That'd be useful information to someone working on this in the future.

SebinSong: @corrideat Did it.
(BTW, That's when Internet Explorer is used according to [this](https://caniuse.com/mdn-api_datatransferitem_webkitgetasentry) can i use report.

corrideat (dismissed):
Looks good to me. It fixes the original issue well in a generic and reliable way. I've tested the changes locally too and it works.

taoeffect (changes requested):
Great work @SebinSong, it works really well in my tests. 👍 

I ran it through `o4-mini-high` and it noticed some inconsistencies in the comments + a typo in the `scrollToIntialPosition` name. I don't know if it's other comments are worthwhile though, I'll leave that up to you to decide.

<details>
I focused on the two areas you changed – the new drag‑and‑drop logic in ChatMain.vue and the updated fileAttachmentHandler in SendArea.vue – and spotted a few places you’ll want to tighten up.

---

### 1) ChatMain.vue → dragEndHandler

**Problem A**  
You only use `e.dataTransfer.items` and fall back on checking `file.type !== ''`.  That will…  
- silently drop files with no reported MIME‐type (e.g. `.md`, some office formats)  
- never fall back to `dataTransfer.files` in browsers that don’t populate `items`  

Original snippet you added:

```js
const items = e.dataTransfer.items

if (items?.length) {
  // …
  for (let i = 0; i < items.length; i++) {
    const item = items[i]
    const entry = item.getAsEntry?.() || item.webkitGetAsEntry?.()

    if (entry) {
      entry.isFile && detectedFiles.push(item.getAsFile())
    } else {
      // NOTE: … this fallback might accept a directory.
      const file = item.getAsFile()

      if (file && file.type !== '') {
        detectedFiles.push(file)
      }
    }
  }

  detectedFiles.length &&
    this.$refs.sendArea.fileAttachmentHandler(detectedFiles, true)
}
```

**Fix & improvement**  
1) Fall back to `dataTransfer.files` if `items` is empty or unsupported.  
2) Rely on `item.kind === 'file'` to detect true files rather than `file.type !== ''`.  
3) early‑return if there are no files.  
4) simplify to a single `.filter/.map` pipeline  

```diff
 dragEndHandler (e /*: DragEvent */) {
   e.preventDefault()
   if (!this.dndState.isActive) return

   this.dndState.isActive = false

-  const items = e.dataTransfer.items
-
-  if (items?.length) {
-    const detectedFiles = []
-
-    for (let i = 0; i < items.length; i++) {
-      const item = items[i]
-      const entry = item.getAsEntry?.() || item.webkitGetAsEntry?.()
-
-      if (entry) {
-        entry.isFile && detectedFiles.push(item.getAsFile())
-      } else {
-        const file = item.getAsFile()
-
-        if (file && file.type !== '') {
-          detectedFiles.push(file)
-        }
-      }
-    }
-
-    detectedFiles.length &&
-      this.$refs.sendArea.fileAttachmentHandler(detectedFiles, true)
-  }
+  const dt = e.dataTransfer
+  if (!dt) return
+
+  // Try items first (more granular), otherwise fall back to files
+  const files = dt.items && dt.items.length
+    ? Array.from(dt.items)
+      .filter(item => item.kind === 'file')
+      .map(item => item.getAsFile())
+      // filter out any nulls or directories
+      .filter(Boolean)
+    : Array.from(dt.files || [])
+
+  if (!files.length) return
+  // hand off to SendArea
+  this.$refs.sendArea.fileAttachmentHandler(files, true)
 }
```

---

### 2) SendArea.vue → fileAttachmentHandler

**Problem B**  
You removed the old “directory” filter but then left a misleading comment and you’re still mutating the `filesList` parameter:

```js
// filter out all entries whose types are unrecognisable.
filesList = Array.from(filesList)

// User clicked 'Cancel button'.
if (!filesList.length) { return }
```

- The comment implies you *do* filter “unrecognisable” entries, but you don’t.  
- Reassigning the parameter `filesList` makes it harder to follow and prevents strong Flow typing.  

**Fix & improvement**  
1) Convert `filesList` → a new `const files` array.  
2) Filter strictly to `instanceof File` so you never try to treat anything else as a File.  
3) Give the handler a proper Flow signature.  

```diff
-fileAttachmentHandler (filesList, appendItems = false) {
-  // filter out all entries whose types are unrecognisable.
-  filesList = Array.from(filesList)
-
-  // User clicked 'Cancel button'.
-  if (!filesList.length) { return }
+fileAttachmentHandler (
+  filesList /*: FileList | Array<File> */,
+  appendItems /*: boolean */ = false
+) /*: void */ {
+  // normalize to a real array of File
+  const files = Array.from(filesList)
+    .filter(f => f instanceof File)
+
+  // if user clicked "cancel" or nothing is a File, bail out
+  if (!files.length) return
```

Everything else in your loop can now refer to `files` instead of the old `filesList`.  This makes it crystal clear:

```js
for (const file of files) {
  const fileUrl = URL.createObjectURL(file)
  const fileSize = file.size
  …
}
```

---

### Extra polish

1) In **ChatMain.vue** you have a typo in your method name  
   `scrollToIntialPosition` → it should be `scrollToInitialPosition`  
   (and update all uses of it).  
2) Add Flow annotations to all your drag/drop handlers:  
   ```js
   dragStartHandler (e: DragEvent) { … }
   dragEndHandler   (e: DragEvent) { … }
   ```
3) Consider moving your drag/drop file‑detection logic into a little utility so both mobile and desktop code can share it.

Let me know if you’d like any deeper context or if there’s another diff you want me to comb through!
</details>

SebinSong:
Thanks @taoeffect , Made updates for those minor issues you mentioned.

taoeffect (approved):
Great work @SebinSong !

Issue #360, opened 2018-03-19
Get rid of NPM


taoeffect:
### Problem

- Instead of getting dependencies directly from maintainers, NPM represents another trusted third-party.
- NPM uses non-sandboxed build and install scripts that can potentially result in dev machine compromise, or [user compromise](https://medium.com/hackernoon/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5)
- NPM/Yarn etc, aren't really doing anything useful beyond the features that git already offers, and represent additional complexity and potential failure points for the project.

### Solution

EDIT: Jump to "see also" section below.

We can use git to integrity check (#359) and do shallow clones into the `node_modules` folder. We can specify precise version numbers via the commit hashes.

The difficult part is handling dependencies of dependencies. That is the true value that package managers bring.

How to deal with that?

A few options:

- Get rid of all dependencies that have dependencies.
- Write our own "package manager" to recurse through the `package.json` files and create our own `package-lock.json` with the git URLs and commit hashes.
- Use a combination of both NPM/yarn and git submodules, and slowly convert over to git-only

Related issues:

- #359 - `Add fsckObjects = true to a project .gitconfig file so that everyone has it`
- #340 - `Prevent package-lock.json from breaking when deps-of-deps update`

See also:

- [Deno](https://github.com/denoland/deno) - next version of nodejs
- [Git as a package registry](https://github.com/LoopPerfect/buckaroo/wiki/Git-as-a-Package-Registry) / [buckaroo](https://github.com/LoopPerfect/buckaroo)
- Git [submodule vs subtree](https://andrey.nering.com.br/2016/git-submodules-vs-subtrees/) (another [article](https://opensource.com/article/20/5/git-submodules-subtrees))

taoeffect:
Git submodules doesn't seem to always work. Example:

```
Initializing git submodules...
Submodule 'openvpn' (https://github.com/kylemanna/docker-openvpn.git) registered for path 'services/openvpn'
Cloning into '/Users/greg/.Trash/forums/services/openvpn'...
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
error: Server does not allow request for unadvertised object 73f8b02172a786937028b207621d1ddae71cb633
Fetched in submodule path 'services/openvpn', but it did not contain 73f8b02172a786937028b207621d1ddae71cb633. Direct fetching of that commit failed.
```

[...]

PR #2737, opened 2025-03-30
2736 owner and contract totals accounting keys

corrideat:
Closes #2736

PR #2759, opened 2025-04-27
Multi-engine bundle

corrideat:
[20250428] Currently a draft as I'm still thinking of improvements and conducting testing. However, early feedback is appreciated.

Should ｃｌｏｓｅ #586 
May ｃｌｏｓｅ #360 (in the sense that running the bundle mostly no longer requires npm)

To test, run `grunt build` and then use `node` or `deno` on `dist/backend/index.mjs` (you may need to set `API_PORT`)

Examples:

> `API_PORT=3000 deno run --allow-env --allow-sys --allow-net -R dist/backend/index.mjs`
> `API_PORT=3000 node dist/backend/index.mjs`


corrideat:
Not really sure what the `opn` alert is about, or rather why it appeared on this PR, since the diff doesn't include the word `opn`.

However, it looks like a deprecated package and we should see if it can be removed.

backend/routes.js
```diff
@@ -781,69 +781,73 @@ route.POST('/kv/{contractID}/{key}', {
...
-  // given B a 412 response
-  const expectedEtag = request.headers['if-match']
-  if (!expectedEtag) {
-    return Boom.badRequest('if-match is required')
-  }
-  // "Quote" string (to match ETag format)
-  const cid = existing ? createCID(existing, multicodes.RAW) : ''
+  // Use a queue to prevent race conditions (for example, writing to a contract
+  // that's being deleeted or updated)
+  return sbp('chelonia/queueInvocation', contractID, async () => {
```
taoeffect: Should this change be ported to #2737?

backend/server.js
```diff
@@ -341,6 +325,7 @@ sbp('sbp/selectors/register', {
       await sbp('chelonia.db/delete', `_private_hidx=${cid}#0`)
       await sbp('chelonia.db/delete', `_private_keyop_idx_${cid}_0`)
       await sbp('chelonia.db/set', cid, '')
+      sbp('chelonia/private/removeImmediately', cid)
```
taoeffect: Should this change be ported to #2737?

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -569,3 +569,5 @@ sbp('okTurtles.events/on', NEW_UNREAD_MESSAGES, (currentChatRoomUnreadMessages)
   const rootState = sbp('state/vuex/state')
   rootState.chatroom.unreadMessages = currentChatRoomUnreadMessages
 })
+
+export {}
```
taoeffect: ..What's this? (Weirdness = comment or remove)

taoeffect (commented):
Some questions

frontend/controller/serviceworkers/sw-primary.js
corrideat: That's an empty export declaration and means that there are no exports. This is needed because esbuild will add `export default require_sw_primary()` to `sw-primary.js`, which is a syntax error as `sw-primary.js` is loaded as a classic worker (no ESM).

It wasn't needed before because this seems to happen after adding `"type": "commonjs"` to `package.json`, which is supposed to be the default but apparently it's not really a default, as it affects things like this (esbuild), and it had to be added because trying to use the 'default' value (i.e., `commonjs`) actually results in Node trying to guess whether a file is ESM or CommonJS (and it incorrectly guesses ESM due to imports).

backend/server.js
corrideat: No, this bugfix doesn't really affect #2737. This is to remove contract state from memory when they're deleted.

backend/routes.js
corrideat: This is more likely to affect #2737 but it's still a separate bug.

scripts/esbuild-shims/dirname.js
```diff
@@ -0,0 +1,7 @@
+import { dirname } from 'node:path'
```
taoeffect, line 1: Could you please create a `README.md` file in this directory explaining what all these shims are for?

Gruntfile.js
```diff
@@ -66,6 +66,7 @@ const distAssets = `${distDir}/assets`
 const distCSS = `${distDir}/assets/css`
 const distContracts = `${distDir}/contracts`
 const distJS = `${distDir}/assets/js`
+const distBackend = `${distDir}/backend`
```
taoeffect, line 69: ~~Is this building the Chelonia Dashboard? If so, could you add a comment either here or below stating that? If not, could you make sure that the Chelonia Dashboard is also built?~~

BTW, PR #2783 updating the Chelonia dashboard was just merged into `master` so please merge `master`

EDIT: Oh wait, no the Chelonia Dashboard is separate and managed by `Gruntfile.dashboard.js` - we should probably figure out how to handle that.

backend/index.js
```diff
@@ -70,13 +70,13 @@ function logSBP (domain, selector, data: Array<*>) {
...
 
-module.exports = (new Promise((resolve, reject) => {
+export default (new Promise((resolve, reject) => {
   sbp('okTurtles.events/on', SERVER_RUNNING, function () {
     console.info(chalk.bold('backend startup sequence complete.'))
     resolve()
   })
   // call this after we've registered listener for SERVER_RUNNING
-  require('./server.js')
+  import('./server.js').catch(e => reject(e))
```
taoeffect: could probably just write `.catch(reject)`

taoeffect (changes requested):
Great work @corrideat! Seems to work fairly well. Left some comments and change requests.

Also, does this PR break running the project with `grunt dev`? The instructions say to run `grunt dev` and then `node`/`deno`, but `grunt dev` is supposed to boot up the servers on its own. Can you restore that functionality?

corrideat:
Marking this as ready for review, although it shouldn't be merged before #2737 (because this multi-engine build doesn't support workers; some additional logic will be needed to build those)

There still is the issue that tests randomly fail with Deno due to some kind of network error. But, this seems to resolve putting nginx between Deno and Cypress, so I'm not really sure what the issue is. Since it works with nginx, likely the issue is not in our code.

corrideat:
> Also, does this PR break running the project with grunt dev? The instructions say to run grunt dev and then node/deno, but grunt dev is supposed to boot up the servers on its own. Can you restore that functionality?

This PR shouldn't affect that functionality. Running with node/deno as per the instructions is a different way to run the backend that skips `grunt`, `babel`, file watching, etc. The reason one needs `grunt dev` first is to build assets (actually, I think `grunt build` should work too).

corrideat:
I've merged `master` into this PR. ~However, it's not quite ready to be merged yet as now `Gruntfile.js` needs to be updated to also produce a bundle for the workers (which didn't exist previously).~

taoeffect:
NOTE: it's important that this PR properly set `CHEL_AVOID_DECRYPTION_ATTEMPTS` per: https://github.com/okTurtles/group-income/pull/2833#discussion_r2124449608

backend/database-sqlite.js
```diff
@@ -53,11 +53,17 @@ export default class SqliteBackend extends DatabaseBackend implements IDatabaseB
...
+      return Buffer.from(value)
+    }
+    return await value
   }
 
   async writeData (key: string, value: Buffer | string) {
-    await this.writeStatement.run(key, value)
+    // Explicit conversion to Buffer seems to be needed for Deno compatibility
+    // (otherwise, the key appears to be dropped)
+    await this.writeStatement.run(key, Buffer.isBuffer(value) ? value : Buffer.from(value))
```
taoeffect, line 66: If we can avoid these changes it would be awesome.

@corrideat says that maybe getting rid of Buffer and using UInt8Array would possibly fix this. We should also explore using something other than `node:sqlite`...

corrideat:
> [!CAUTION]
> DO NOT MERGE

Leaving this PR open (mostly for reference) until `chel serve` is closed. However, this PR should no longer be needed after that, since `chel serve` takes a slightly different approach yet still renders this PR redundant.

corrideat:
Closed in favour of #2961

PR #2775, opened 2025-05-05
#2770 - Display chatroom name in bold when it has unread messages

### LINKED ISSUE #2770: Format chatrooms in bold when they have new messages
### Problem

When a chatroom we aren't currently viewing (but are joined to) has a new message, the chatroom should be made bold in the UI, but it isn't.

### Solution

Automatically make the chatroom bold if it has new messages (in addition to if we manually marked them as unread).

SebinSong:
closes #2770 

**[Video]**
As you can see, when a group member(`user-1`) sends new messages in the chat, the chatroom names in another user's (`user-2`) browser is displayed as bold. This also works when the `user-2` above is offline.

https://github.com/user-attachments/assets/fa589243-84a3-4751-a9e2-07ad5630024a

### DIFF
--- a/frontend/setupChelonia.js
+++ b/frontend/setupChelonia.js
@@ -176,8 +176,7 @@ const setupChelonia = async (): Promise<*> => {
           'chelonia/contract/hasKeysToPerformOperation',
           'gi.actions/identity/kv/initChatRoomUnreadMessages', 'gi.actions/identity/kv/deleteChatRoomUnreadMessages',
           'gi.actions/identity/kv/setChatRoomReadUntil',
-          'gi.actions/identity/kv/addChatRoomUnreadMessage', 'gi.actions/identity/kv/removeChatRoomUnreadMessage',
-          'gi.actions/identity/kv/markAsUnread'
+          'gi.actions/identity/kv/addChatRoomUnreadMessage', 'gi.actions/identity/kv/removeChatRoomUnreadMessage'
         ],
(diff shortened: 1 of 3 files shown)


frontend/setupChelonia.js
```diff
@@ -176,7 +176,8 @@ const setupChelonia = async (): Promise<*> => {
           'chelonia/contract/hasKeysToPerformOperation',
           'gi.actions/identity/kv/initChatRoomUnreadMessages', 'gi.actions/identity/kv/deleteChatRoomUnreadMessages',
           'gi.actions/identity/kv/setChatRoomReadUntil',
-          'gi.actions/identity/kv/addChatRoomUnreadMessage', 'gi.actions/identity/kv/removeChatRoomUnreadMessage'
+          'gi.actions/identity/kv/addChatRoomUnreadMessage', 'gi.actions/identity/kv/removeChatRoomUnreadMessage',
+          'gi.actions/identity/kv/markAsUnread'
```
taoeffect: What was the reason this selector was added to `allowedSelectors`?

`allowedSelectors` is only for selectors inside the app that can be called from contracts. But AFAICT `'gi.actions/identity/kv/markAsUnread'` is only ever called from `ChatMain.vue`, not from a contract.

taoeffect (commented):
Great work @SebinSong! I like how you solved this 😎 👍 

One question...

frontend/setupChelonia.js
SebinSong: In that case, this is not needed here. Removed it. (I thought I missed this in my previous PR)

taoeffect (approved):
Great work @SebinSong! 👏

PR #2776, opened 2025-05-05
Re-sync on PUB event

### LINKED ISSUE #2769: Failure to add user to private channel on testing
### Problem

1. On testing, in FFDev, I created a group with `u1` and invited `u2`.
2. As u2, in a separate container tab, I created a private channel and added `u1`

However, `u1` doesn't see the chatroom that they were added to!

u2's view:

Image

u1's view:

Image

Logs:

- [gi_logs.json(u1).txt](https://github.com/user-attachments/files/19987089/gi_logs.json.u1.txt)
- [gi_logs.json(u2).txt](https://github.com/user-attachments/files/19987090/gi_logs.json.u2.txt)

### Solution

Find and fix the problem.

corrideat:
Fix #2769. Verify that #2666 isn't a regression.

To test (on master and then this):

Use Chrome or a Chromium-based browser. Create a group with two users. Create a new network throttling setting with high latency (e.g., 5s) and use this for both u1 (the one creating the chatroom and adding u2) and for u2. u1 only needs to be throttled for the part where u2 is added to the chatroom.

### DIFF
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -599,6 +599,9 @@ export default (sbp('sbp/selectors/register', {
         // 'new' subscriptions as well as every time the connection is reset
         'subscription-succeeded': (event) => {
           const { channelID } = event.detail
+          // The check below is needed because we could have unsubscribed since
+          // requesting a subscription from the server. In that case, we don't
+          // need to call `sync`.
           if (this.subscriptionSet.has(channelID)) {
             // For new subscriptions, some messages could have been lost
             // between the time the subscription was requested and it was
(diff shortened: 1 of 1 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -591,25 +591,24 @@ export default (sbp('sbp/selectors/register', {
...
-          options.handlers?.['online']?.()
-        },
-        'reconnection-succeeded' () {
-          resyncOnOnline()
-          options.handlers?.['reconnection-succeeded']?.()
+        // Every time we get a REQUEST_TYPE.SUB response, which happens for
+        // 'new' subscriptions as well as every time the connection is reset
+        'subscription-succeeded': (event) => {
+          const { channelID } = event.detail
+          if (this.subscriptionSet.has(channelID)) {
```
taoeffect, line 605: Why would we need this `if` check if we're inside of the `'subscription-succeeded'` event? Isn't it guaranteed to be in there? (If not, could you add a comment explaining why it's necessary?)

corrideat: It's not guaranteed to be in there because we could have unsubscribed since.

taoeffect (approved):
Great work @corrideat!

PR #2778, opened 2025-05-06
Log warn for outgoing messages

### LINKED ISSUE #2773: Make errors for outgoing messages warnings
### Problem

Sometimes when sending a message there can be errors such as these:

Image

These aren't real errors: they would be if they happened during processing on incoming messages, but this is happening during the processing step when sending a message, in which case it should be a warning.

### Solution

Figure out a way to make these errors into warnings only when sending outgoing SPMessages. This will improve the developer experience.

corrideat:
Closes #2773

### DIFF
--- a/shared/domains/chelonia/internals.js
+++ b/shared/domains/chelonia/internals.js
@@ -14,7 +14,7 @@ import { encryptedIncomingData, encryptedOutgoingData, unwrapMaybeEncryptedData 
 import type { EncryptedData } from './encryptedData.js'
 import { ChelErrorKeyAlreadyExists, ChelErrorResourceGone, ChelErrorUnrecoverable, ChelErrorWarning, ChelErrorDBBadPreviousHEAD, ChelErrorAlreadyProcessed, ChelErrorFetchServerTimeFailed, ChelErrorForkedChain } from './errors.js'
 import { CONTRACTS_MODIFIED, CONTRACT_HAS_RECEIVED_KEYS, CONTRACT_IS_SYNCING, EVENT_HANDLED, EVENT_PUBLISHED, EVENT_PUBLISHING_ERROR } from './events.js'
-import { buildShelterAuthorizationHeader, findKeyIdByName, findSuitablePublicKeyIds, findSuitableSecretKeyId, getContractIDfromKeyId, keyAdditionProcessor, recreateEvent, validateKeyPermissions, validateKeyAddPermissions, validateKeyDelPermissions, validateKeyUpdatePermissions } from './utils.js'
+import { buildShelterAuthorizationHeader, findKeyIdByName, findSuitablePublicKeyIds, findSuitableSecretKeyId, getContractIDfromKeyId, keyAdditionProcessor, logEvtError, recreateEvent, validateKeyPermissions, validateKeyAddPermissions, validateKeyDelPermissions, validateKeyUpdatePermissions } from './utils.js'
 import { isSignedData, signedIncomingData } from './signedData.js'
 // import 'ses'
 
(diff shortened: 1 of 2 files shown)


shared/domains/chelonia/utils.js
```diff
@@ -92,7 +92,7 @@ const validateActionPermissions = (signingKey: SPKey, state: Object, opT: string
     // available for us to see. In this case, we ignore the missing key.
     // For incoming messages, we must check permissions and a missing
     // key means no permissions.
-    if (!innerSigningKey && direction === 'outgoing') return true
+    if (!innerSigningKey && msg._direction === 'outgoing') return true
```
taoeffect, line 95: Shouldn't we use `msg.direction()` instead of `_direction` to access this?

shared/domains/chelonia/utils.js
```diff
@@ -852,3 +852,13 @@ export const collectEventStream = async (s: ReadableStream): Promise<any[]> => {
   }
   return r
 }
+
+// Used inside processing functions for displaying errors at the 'warn' level
+// for outgoing messages to increase the signal-to-noise error. See issue #2773.
+export const logEvtError = (msg: SPMessage, ...args: any[]) => {
+  if (msg._direction === 'outgoing') {
```
taoeffect, line 861: Same question re `.direction()` here

taoeffect (changes requested):
Looks great @corrideat! Just one minor question / change request.

shared/domains/chelonia/utils.js
corrideat: I figured it's fine since SPMessage is part of Chelonia, and this saves a function call.

shared/domains/chelonia/utils.js
corrideat: Same answer.

shared/domains/chelonia/internals.js
```diff
@@ -1149,12 +1149,12 @@ export default (sbp('sbp/selectors/register', {
                     // reason to remain subscribed to this contract. In this
                     // case, attempt to release it.
                     sbp('chelonia/contract/release', foreignContract, { try: true }).catch(e => {
-                      console.error(`[chelonia] Error at OP_KEY_DEL internalSideEffectStack while attempting to release foreign contract ${foreignContract}`, e)
+                      logEvtError(message, `[chelonia] Error at OP_KEY_DEL internalSideEffectStack while attempting to release foreign contract ${foreignContract}`, e)
```
taoeffect: Is this really something that can happen with an outgoing message that is OK to leave as a warning?

corrideat: Outgoing messages typically don't have side effects and the `internalSideEffectStack` isn't used. So, this is mostly a discussion about hypotheticals.

As for your question, if we ever did execute side effects for outgoing messages, I presume it'll be in some kind of sandboxed way and errors still wouldn't have the same weight as they do on incoming messages, same as with `process`. Since this isn't what happens and I can't think of good reasons to run side effects for outgoing messages, nor a way to do so cleanly, I'm just guessing here.

taoeffect: If this shouldn't ever happen for outgoing messages, then perhaps this should be left as `console.error`, no? As this happening should be an exceptional circumstance that we should take note of if it does happen (since you say it shouldn't). If my reasoning sounds OK to you, then please revert.

corrideat: > then perhaps this should be left as console.error, no?

Maybe, I see arguments both ways. Happy to revert though. However,

> that we should take note of if it does happen (since you say it shouldn't)

If this is the reason, I don't think we should rely on obscure errors happening to detect this. If we want truly defensive coding, maybe at the top of this side effect there could be `if (outgoing) throw Error('Something really bad just happened')`.

Reasons for leaving it as is: 
 * Consistency

Reasons for reverting:
 * Currently, if it happened it'd likely be an error

Should I revert?

taoeffect: > If this is the reason, I don't think we should rely on obscure errors happening to detect this. If we want truly defensive coding, maybe at the top of this side effect there could be if (outgoing) throw Error('Something really bad just happened').

Sure, feel free to do that, then I would also revert as well, since it doesn't make sense to have `logEvtError` there in that case either because `logEvtError` implies it's a piece of code that can run on outgoing messages (and needs to be logged as a warning in that case).

taoeffect (approved):
Great work @corrideat!

PR #2787, opened 2025-05-15
Bugfix

### LINKED ISSUE #2786: FS backend bugs
### Problem

The recent merging of #2725 for the `router` backend has broken the way the `fs` backend is supposed to work.

From @corrideat:

> I think I've noticed a bug in GI. After the DB PR, now the fs backend is unusable because preloaded entries aren't stored for keyChunkLength=2 (and this can't be overridden, I guess unless using the router). Also, I missed it during my review, but the path it looks at is e.g., `/zL/mM/d/Xb/TJehLLmGVRWYwJryXfoGmBJVWxXbgNcLkVKAwEhBh` instead of  `/zL/mM/d/Xb/zLmMdXbTJehLLmGVRWYwJryXfoGmBJVWxXbgNcLkVKAwEhBh` (i.e., the file name should not be trimmed)

It's not just that though, it shouldn't be using sub-folders by default at all.

### Solution

1. Make sure that by default the `fs` backend doesn't use sub-folders, and that it doesn't use them when `GI_PERSIST=router` is used either.
2. Make sure that if they are enabled, that the final path is not trimmed

corrideat:
Fix #2786

### DIFF
--- a/backend/database-fs.js
+++ b/backend/database-fs.js
@@ -54,7 +54,7 @@ export default class FsBackend extends DatabaseBackend implements IDatabaseBacke
     if (!this.depth) return join(this.dataFolder, key)
     // TODO: optimize if necessary.
     const keyChunks = key.match(new RegExp('[A-Za-z=]{1,' + this.keyChunkLength + '}', 'g')) ?? []
-    return join(this.dataFolder, ...keyChunks.slice(0, this.depth), keyChunks.slice(this.depth).join(''))
+    return join(this.dataFolder, ...keyChunks.slice(0, this.depth), key)
   }
 
   async init () {
--- a/backend/database-fs.js
+++ b/backend/database-fs.js
@@ -1,7 +1,7 @@
 'use strict'
 
 import { mkdir, readdir, readFile, rm, unlink, writeFile } from 'node:fs/promises'
-import { dirname, join, resolve } from 'node:path'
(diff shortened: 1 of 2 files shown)


taoeffect (changes requested):
Nicely done @corrideat!

I seem to have possibly found another bug though while testing...

Using this `database-router-config.json` file to place file manifests (`zLGQo2DimCH8`) and files (`zLKHweRGqjMM`) under their own sub-folders of `data/files`:

```json
{
  "zLGQo2DimCH8": {
    "name": "fs",
    "options": {
      "dirname": "./data/files",
      "depth": 1,
      "keyChunkLength": 12
    }
  },
  "zLKHweRGqjMM": {
    "name": "fs",
    "options": {
      "dirname": "./data/files",
      "depth": 1,
      "keyChunkLength": 12
    }
  },
  "*": {
    "name": "sqlite",
    "options": {
      "dirname": "./data",
      "filename": "groupincome.db"
    }
  }
}
```

I am getting this layout:

Screenshot 2025-05-15 at 7 17 16 PM

Notice that for some reason `zLGQo2DimCH8` is getting cut off for some reason...

Am running this with:

```
$ SKIP_DB_FS_CASE_SENSITIVITY_CHECK=true GI_PERSIST=router grunt dev
```

taoeffect (approved):
Job well done @corrideat! 😄 👍 

Screenshot 2025-05-16 at 11 22 53 AM

PR #2788, opened 2025-05-15
Sync chatroom notification changes

### LINKED ISSUE #2771: Chatroom notifications related fixes
### Problem

Right now we correctly display notifications every time someone sends us a DM, but we do not receive notifications from private chatrooms we are a part of.

### Solution

Treat private chatrooms as DMs by default. In other words, for private chatrooms, update these settings:

Image

And then, I noticed at least on https://groupincome.app that even with these settings enabled I didn't see push notifications on mobile. Push notifications should be working for all messages, but the actual display of notifications might be disabled for these chatrooms even if these settings are enabled. Please double-check this and test using https://testing.groupincome.org, which is now running latest `master`.

corrideat:
Fix #2771

### DIFF
--- a/frontend/controller/app/chatroom.js
+++ b/frontend/controller/app/chatroom.js
@@ -1,5 +1,5 @@
 import sbp from '@sbp/sbp'
-import { DELETED_CHATROOM, JOINED_CHATROOM, LEFT_CHATROOM, NEW_CHATROOM_UNREAD_POSITION } from '@utils/events.js'
+import { DELETED_CHATROOM, JOINED_CHATROOM, LEFT_CHATROOM, NEW_CHATROOM_NOTIFICATION_SETTINGS, NEW_CHATROOM_UNREAD_POSITION } from '@utils/events.js'
 
 const switchCurrentChatRoomHandler = ({ identityContractID, groupContractID, chatRoomID }) => {
   const rootState = sbp('state/vuex/state')
@@ -48,6 +48,12 @@ sbp('okTurtles.events/on', NEW_CHATROOM_UNREAD_POSITION, ({ chatRoomID, messageH
   } else {
     sbp('state/vuex/commit', 'deleteChatRoomScrollPosition', { chatRoomID })
   }
+})
+sbp('okTurtles.events/on', NEW_CHATROOM_NOTIFICATION_SETTINGS, ({ chatRoomID, settings }) => {
+  sbp('state/vuex/commit', 'setChatroomNotificationSettings', {
+    chatRoomID,
+    settings
+  })
 })
 
 export default ([]: string[])
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -3,7 +3,7 @@
 import { deserializer, serializer } from '@chelonia/serdes'
 import { L } from '@common/common.js'
 import sbp from '@sbp/sbp'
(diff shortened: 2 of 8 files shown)


SebinSong:
I've just tested this PR locally and it appears to fix the issue I mentioned in [this comment](https://github.com/okTurtles/group-income/issues/2771#issuecomment-2876432443). Nice work.

taoeffect (changes requested):
Great work @corrideat, this seems to fix the issue with setting not being saved.

To close the issue, however, it also needs to modify the defaults for private chatrooms (see the screenshot in the issue).

Also, while testing this with `ngrok` between FFDev and Brave, I noticed that the new message sound wouldn't play at all. Could you please investigate that as well?

corrideat:
> Also, while testing this with ngrok between FFDev and Brave, I noticed that the new message sound wouldn't play at all. Could you please investigate that as well?

Could it be this?

image

taoeffect (commented):
Great work @corrideat! This PR seems pretty much ready to merge.

I was hoping we could also fix an issue with this logic in `Background.vue`:

```js
    shouldPlay () {
      return !isPwa() && (document.hidden || this.isAppIdle)
    },
```

Could we change this logic to be (pseudocode):

```js
    shouldPlay () {
      return !isPwa() && !(this.weAreLookAtChatroom(newMessage.chatroom()) && !this.isAppIdle)
    },
```

corrideat:
@taoeffect With the new proposed logic, if the ignore the PWA part, a sound would play as follows:

```
    1 = looking         && !1 = idle     =   !(1 & 0) = !0 = 1 (sound)
    1 = looking         && !0 = active   =   !(1 & 1) = !1 = 0 (no sound)
    0 = not looking     && !1 = idle     =   !(0 & 1) = !0 = 1 (sound)
    0 = not looking     && !0 = active   =   !(0 & 0) = !0 = 1 (sound)
```

In other words, a sound would always play, except when the app is not idle and one is looking at the chatroom. Unlike the current check, `document.hidden`, a sound would play when the app is not idle and in the foreground (for example, if a DM comes in while doing something else in the app).

The `document.hidden` check was introduced in <https://github.com/okTurtles/group-income/pull/2025>. Looking at the issue that's meant to address, <https://github.com/okTurtles/group-income/issues/2022>, the request was:

> Don't play the sound if the DM is open and in the foreground AND the user has recently interacted with the UI.

It looks like the change request brings it closer to the original issue request, but IMHO it can still be pretty annoying to get a notification sound, especially when the app is in the foreground. Imagine, for example, being part of a private chatroom or DM that's quite active but having a conversation you can't or won't follow. Or using Group Income in a library, or just personal preference.

Shouldn't this be some kind of global setting (similar to what Slack has, for example), where users can specify when a sound should play, if at all?

taoeffect:
I agree it should be a global setting too, but until then I want it to behave like Slack and also make sure people don't miss messages.

Right now it's too quiet.

As for the logic, I might have messed up. I wanted it to not play a sound when there's a message in the current chat we're looking at and not idle. 

EDIT: Actually my logic seems sound, but I think it should also take into account the sound setting on each channel

corrideat:
> I think it should also take into account the sound setting on each channel

This is already done prior to emitting the event.

taoeffect (approved):
Beautiful! Works perfectly!

PR #2790, opened 2025-05-18
Fix prompts

### LINKED ISSUE #2768: Failure to recover from server wipe
### Problem

If a server database is wiped while we are logged in, in some cases there seems to be a failure to load the page entirely. This is currently happening on `testing` with an account that I have that was logged in on Brave (Desktop).

Refreshing the page doesn't fix it:

Image

On the backend I see:

Image

### Solution

Figure out what the deal is and fix it.

corrideat:
Closes #2768

### DIFF
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -406,6 +406,20 @@ export default (sbp('sbp/selectors/register', {
       }
 
       try {
+        // When syncing from scratch, we need to call `retain` on our own
+        // identity contract to ensure the reference count is positive and we
+        // don't accidentally remove it.
+        // When not syncing from scratch, we call `sync` instead, since we don't
+        // want to increment the reference count every time the app is loaded.
+        // To decide what to call, we don't simply check for `state` (which
+        // would be defined when restoring from a saved state) because in some
+        // cases the reference count could be 0, resulting in an error.
+        // Instead, we check if the reference count exists. This check is
+        // correct and also ensures that, if the state gets corrupted for some
+        // reason (for example, because the contract is deleted from the server
+        // and then restored) and the reference count lost, we call `/retain`
+        // again, which will succeed and also ensure our own identity contract
+        // won't get removed.
         if (!cheloniaState?.contracts[identityContractID]?.references) {
           // Make sure we don't unsubscribe from our own identity contract
           // Note that this should be done _after_ calling
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -406,7 +406,7 @@ export default (sbp('sbp/selectors/register', {
       }
 
       try {
-        if (!state) {
+        if (!cheloniaState?.contracts[identityContractID]?.references) {
           // Make sure we don't unsubscribe from our own identity contract
           // Note that this should be done _after_ calling
           // `chelonia/storeSecretKeys`: If the following line results in
--- a/frontend/controller/app/identity.js
+++ b/frontend/controller/app/identity.js
@@ -30,13 +30,13 @@ const loadState = async (identityContractID: string, password: ?string) => {
       return { encryptionParams, state, cheloniaState }
     } else {
       // There's no state to restore
-      return { encryptionParams, state, cheloniaState: null }
+      return { encryptionParams, state: null, cheloniaState: null }
     }
   } else {
     const state = await sbp('gi.db/settings/load', identityContractID)
 
     // cheloniaState is only stored in settings when logging in with a password
-    // If there's an active session, then chelnoiaState is stored and
+    // If there's an active session, then cheloniaState is stored and
     // managed separately
     return { encryptionParams: null, state, cheloniaState: null }
   }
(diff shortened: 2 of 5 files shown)


shared/domains/chelonia/utils.js
```diff
@@ -575,8 +575,10 @@ export function eventsAfter (contractID: string, sinceHeight: number, limit?: nu
     const eventsResponse = await this.config.fetch(lastUrl, { signal })
     if (!eventsResponse.ok) {
       const msg = `${eventsResponse.status}: ${eventsResponse.statusText}`
-      if (eventsResponse.status === 410) throw new ChelErrorResourceGone(msg)
-      throw new ChelErrorUnexpectedHttpResponseCode(msg)
+      // $FlowFixMe[extra-arg]
+      if (eventsResponse.status === 404 || eventsResponse.status === 410) throw new ChelErrorResourceGone(msg, { cause: eventsResponse.status })
+      // $FlowFixMe[extra-arg]
+      throw new ChelErrorUnexpectedHttpResponseCode(msg, { cause: eventsResponse.status })
```
taoeffect, line 581: Note: these changes need to be ported to `libcheloniajs`

taoeffect (changes requested):
Nice work @corrideat! I had a chance to test this and found some issues:

### Problem 1: supposed to be logged out?

So, I tested this on FFDev:

1. Start server with `NODE_ENV=production GI_PERSIST=sqlite grunt dev`
2. Create group, post some messages to chatroom
3. Stop server, rename `groupincome.db` to `groupincome-back.db`
4. Start server, refresh page

I got this prompt with a "Continue" button:

> Your account seems to have been deleted from the server.
> To fix this, you need to log out and create a new account.

I clicked "Continue" but was immediately logged out:

Screenshot 2025-05-19 at 6 51 29 PM

If this intentional, the message needs to be changed and instead of "Continue" it should say "Log out".

If it's unintentional, the bug should be fixed.

### Problem 2: 

I stopped the server and renamed `groupincome-back.db` back to `groupincome.db`. Then I attempted to log back in with the account `u1`, however I couldn't and was greeted with this error message:

Screenshot 2025-05-19 at 6 52 24 PM

Seems like I'm permanently stuck being logged out and can't log back in even though the data is back. This should be fixed too as we don't want users permanently stuck if the sysadmin had to wipe the db and restore from backup.

corrideat:
> Problem 1: supposed to be logged out?

Thanks for pointing this out. Yes, you're supposed to be logged out, and I've adjusted the text accordingly.

Before, you would get a prompt asking you if you'd like to log out, with 'Yes' and 'No'. (Except that it didn't work, because 'Prompt' wasn't allowed in the pre-login state, leading to the blank page in the issue). After our discussion, I thought that we could keep the prompt, but there was no point (for now at least) in the 'No' option, so removed it and changed 'Yes' to 'Continue'. I concur that 'Log out' is a much better label than 'Continue'.

> Problem 2

Interesting. Thanks for the steps. This was quite reproducible, and I also a bad issue. I've fixed it now by making a better check for deciding whether to call 'retain' or 'release'. This fixes this issue and in general is a more robust check should the state get corrupted.

Working on this issue showed me that logging in is in some cases rough around the edges (for example, after a deletion / restore, as you found out). Along with the improvements in #2700 / #2709, I think we should test edge cases better and improve the login process.

frontend/controller/actions/identity.js
```diff
@@ -406,7 +406,7 @@ export default (sbp('sbp/selectors/register', {
       }
 
       try {
-        if (!state) {
+        if (!cheloniaState?.contracts[identityContractID]?.references) {
```
taoeffect, line 423: It was at least somewhat followable before, but now this line seems to come out of nowhere.

Could you please add a detailed comment explaining why we're checking this instead of `state`?

frontend/controller/app/identity.js
```diff
@@ -518,7 +518,15 @@ export default (sbp('sbp/selectors/register', {
...
         if (!wipeOut) {
+          // Save current Chelonia state into the state to be saved
           state.cheloniaState = cheloniaState
+          // These keys (contract state) will be restored from the Chelonia
+          // state upon login (see `loadState`). Remove them to avoid storing
+          // unnecessary data.
+          Object.keys(cheloniaState.contracts || {}).forEach((contractID) => {
+            delete state[contractID]
+          })
+          delete state.contracts
```
taoeffect, line 529: Is there a bug this is preventing? If so, could you update the comment to mention it?

frontend/controller/app/identity.js
```diff
@@ -515,7 +518,15 @@ export default (sbp('sbp/selectors/register', {
...
```
taoeffect: Is there a bug this is preventing? If so, could you update the comment to mention it?

taoeffect (changes requested):
Fantastic! Works well in my tests.

Just two RFCs and it looks good to merge.

frontend/controller/app/identity.js
corrideat: Not necessarily preventing a bug, but this is unnecessary information (see in `loadState` that it's discarded) and because of that, it's also potentially confusing.

frontend/controller/actions/identity.js
corrideat: This is what fixes the second bug. I'll add a comment explaining.

corrideat: Would this comment look OK?

```
// When syncing from scratch, we need to call `retain` on our own identity
// contract to ensure the reference count is positive and we don't accidentally
// remove it.
// When not syncing from scratch, we call `sync` instead, since we don't want to
// increment the reference count every time the app is loaded.
// To decide what to call, we check if the reference count exists. This check
// is correct and also ensures that, if the state gets corrupted for some reason
// (for example, because the contract is deleted from the server and then
// restored) and the reference count lost, we call `/retain` again.
```

taoeffect: Yes, I would just add, in addition to that, an explanation for why checking `!state` is insufficient.

EDIT: I realize that the comment in your head might explain that, but what I'm looking for is something like, "We don't simply check `!state` because _____"

taoeffect (approved):
Fantastic!!

PR #2802, opened 2025-05-25
Bugfix

corrideat:


backend/pubsub.js
```diff
@@ -436,7 +436,8 @@ const publicMethods = {
         yield * subscribers
       } else {
         for (const subscriber of subscribers) {
-          const kvFilter = subscriber.kvFilter.get(channelID)
+          // kvFilter may be undefined for push subscriptions
+          const kvFilter = subscriber.kvFilter?.get(channelID)
```
taoeffect, line 440: Is this really all that needs to be done?

(There's nothing we need to do to remove `subscriber` from `subscribers` or ..?)

taoeffect (commented):
Thanks @corrideat! A question:

backend/pubsub.js
corrideat: Yes, this is what fixes the issue of `subscriber.kvFilter` being undefined, which, as the comment says, is undefined for push subscriptions.

Since we're not sending KV events to push subscriptions (the logic for that is elsewhere), I suppose we could rewrite this to skip over push subscriptions entirely if `kvKey` is provided, as an alternative solution.

taoeffect (approved):
Nice!

PR #2809, opened 2025-05-27
Reimplement logic for hasNotReadTheLatestMessage

### LINKED ISSUE #2797: Channels sometimes aren't made bold
### Problem

Over on the testing site and in the testing group running 1.9.2 I noticed that in the PWA on Android the `#random` channel isn't being made bold when new messages appear in it.

I'm sending messages on Desktop as `greg`, and watching to see whether `#random` becomes bold or not. So far it's been over 5 minutes and it hasn't been.

Image

- 📎 [gi_logs.json.txt](https://github.com/user-attachments/files/20398638/gi_logs.json.txt) (for `u2` the user on Android PWA that's looking in the screenshot above)

UPDATE: I checked again after an hour and it finally made the channel bold. So the question is why did it take so long?

### Solution

Find out what's going on, and see if there's a way to make it react immediately. Although right now it's not even reacting slowly.

For reference this feature was implemented in #2775

EDIT: Whether or not to display a channel as bold can, instead of using the height of the message, just have a boolean variable that's toggled on and off (depending on whatever condition, e.g. "looked at the latest message in the chat"), and set using the KV store of the chatroom.

corrideat:
Closes #2797

### DIFF
--- a/frontend/views/containers/chatroom/ConversationsList.vue
+++ b/frontend/views/containers/chatroom/ConversationsList.vue
@@ -76,18 +76,16 @@ export default ({
       const entries = Object.keys(this.list.channels).map(chatId => {
         const chatroomState = this.$store.state.contracts[chatId]
         if (!chatroomState || typeof chatroomState?.height !== 'number') { return null }
-
-        const latestMsgHeight = chatroomState.height
+        const messages = this.$store.state[chatId]?.messages
         const chatReadUntilInfo = this.ourUnreadMessages?.[chatId]?.readUntil
 
-        if (!chatReadUntilInfo) {
-          // NOTE: If a user creates a new channel but leaves before any message is sent/received in there,
-          // it's 'readUntil' data is undefined. In this case, we can determine whether the latest message is not seen yet by checking if
-          // its height value is >= 3. (3 is the height value of the second message in every chatroom)
-          return [chatId, latestMsgHeight >= 3]
-        } else {
-          return [chatId, latestMsgHeight > chatReadUntilInfo.createdHeight]
-        }
+        // If there are no messages, assume everything has been read
+        if (!messages?.length) return [chatId, false]
+        // OTOH, if `readUntil` isn't set, assume the channel is unread
+        if (!chatReadUntilInfo) return [chatId, true]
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/ConversationsList.vue
```diff
@@ -76,18 +76,16 @@ export default ({
...
-        } else {
-          return [chatId, latestMsgHeight > chatReadUntilInfo.createdHeight]
-        }
+        // If there are no messages, assume everything has been read
+        if (!messages?.length) return [chatId, false]
+        // OTOH, if `readUntil` isn't set, assume the channel is unread
+        if (!chatReadUntilInfo) return [chatId, true]
+        // Otherwise, see if some message has a height higher than that of the
+        // last read message. If so, makr the channel as having unread messsages
+        return [chatId, messages.some(({ height }) => height > chatReadUntilInfo.createdHeight)]
```
taoeffect: Nice, seems to work too, however, try the following:

1. Create a group and a couple channels
2. Join u2 to the group and each channel
3. Then have u1 remove u2 from the group

If you did this, you might notice that each time u2 joins a channel (step 2), the channel becomes unread for u1. This is a problem, because it's annoying, and in a busy group it would mean the chatrooms would be constantly unread even though there isn't a meaningful message there.

Then, in step 3, when u1 kicks u2 out of the group, suddenly every channel that u2 was in becomes unread. This is also a problem for the same reason. So I think these types of messages (joined/leaved) need to ignored in this consideration.

Also, there's a typo in the comment: `makr` => `make`

taoeffect (changes requested):
Great work @corrideat! I tested it and reported my findings:

frontend/views/containers/chatroom/ConversationsList.vue
corrideat: That's actually working as intended, but I see your point. Should we ignore all 'notification' messages?

taoeffect: I hesitate to say "yes" because I don't remember what other types of notifications messages there could be...

taoeffect (approved):
Fantastic work @corrideat !!

PR #2810, opened 2025-05-27
Replace unnecessary reactive calls with regular operations

### LINKED ISSUE #2807: Remove unnecessary uses of reactiveSet and reactiveDel
### Problem

In Chelonia, we use `reactiveSet` & `reactiveDel` to set properties. However, in many cases this is unnecessary, such as when processing op codes as a copy of the state is being used (`reactiveSet` & `reactiveDel` should still be used when applying final changes to the state).

### Solution

Remove unnecessary uses of `reactiveSet` & `reactiveDel`, while keeping it for actual Chelonia state mutations.

corrideat:
Rationale:

Processing messages is always done on copies of the state, and the results are then applied to the root state after all operations have run. Since, it's not really necessary to incur into the overhead of using `reactiveSet` / `reactiveDel`.


taoeffect (approved):
Great work @corrideat!

These changes also need to be ported to libcheloniajs if they haven't already been, so I'll also add this to the milestone.

PR #2811, opened 2025-05-27
Backport libchelonia changes

corrideat:
This PR includes several logic changes backported from the work on a standalone `libchelonia`. Logic changes are defined, for the purposes of this PR, as changes to the structure of the code after removing all type information.

Some of the changes are cosmetic, but a few are bugfixes.

1. Removed handling of `foreignContractID` for `OP_CONTRACT` keys. This isn't necessarily a bugfix (I don't think the previous code was wrong), but it's an unused feature that I can't think of a good use case for. Probably I added this to solve a specific problem that was later addressed in a different way. Since the code itself isn't wrong, just weird, I commented it out.
2. `OP_CONTRACT` needs some special bootstrapping for its `authorizedKeys` state (since the state is initially empty). `OP_CONTRACT` is defined as allowing for encrypted keys (same as `OP_KEY_ADD`), but the way things are implemented won't in fact accept nor handle encrypted keys. This hasn't broken anything for a few reasons: we're not currently using these encrypted keys, and the code is (currently) written in such a way that encrypted keys will end up being ignored.
3. Call to `.valueOf()` in `SPMessage.prototype.description`. This allows for more complete logs for unencrypted actions that have an inner signature. This isn't as much of a bugfix as it is an enhancement, and it hasn't broken anything because we aren't using unencrypted actions much.
4. Also a bugfix in `SPMessage.prototype.description`. `.type` should be `.action`. `.type` doesn't exist. This hasn't broken anything because of the same reasons as the previous bugfix.
5. `pendingKeyRequests?.includes(k.name)` to `pendingKeyRequests?.some(pkr => pkr.name === k.name)` in `'chelonia/out/keyRequest'`. This is a bugfix, since `pkr` is an object and not a string. This is used for `havePendingKeyRequest`, which is used to return early (avoiding unnecessary duplicate key requests). This likely hasn't resulted in observable issues since it'd just result in unnecessary, duplicate key requests.

### DIFF
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -291,9 +291,9 @@ export class SPMessage {
       const value = rawSignedIncomingData(parsedValue)
       const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
         const k = unwrapMaybeEncryptedData(wk.valueOf())
-        if (!k.data) return null
+        if (!k) return null
         return [k.data.id, k.data]
-      })).filter(Boolean)
+      }).filter(Boolean))
       state = {
         _vm: {
           type: head.type,
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -129,6 +129,9 @@ const decryptedAndVerifiedDeserializedMessage = (head: Object, headJSON: string,
     (message: any).keys = (message: any).keys?.map((key, eKeyId) => {
       return maybeEncryptedIncomingData(contractID, state, key, height, additionalKeys, headJSON, (key) => {
         if (!key.meta?.private?.content) return
+        // The following two lines are commented out because this feature
+        // (using a foreign decryption contract) doesn't seem to be in use and
+        // the use case seems unclear.
         // const decryptionFn = key.meta.private.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
         // const decryptionContract = key.meta.private.foreignContractID ? key.meta.private.foreignContractID : contractID
         const decryptionFn = encryptedIncomingData
@@ -287,9 +290,10 @@ export class SPMessage {
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
       const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
-        const k = wk.valueOf()
-        return [k.id, k]
-      }))
+        const k = unwrapMaybeEncryptedData(wk.valueOf())
+        if (!k.data) return null
+        return [k.data.id, k.data]
+      })).filter(Boolean)
       state = {
         _vm: {
           type: head.type,
@@ -452,10 +456,14 @@ export class SPMessage {
     const type = this.opType()
     let desc = `<op_${type}`
     if (type === SPMessage.OP_ACTION_UNENCRYPTED) {
-      const value = this.opValue().valueOf()
-      // $FlowFixMe
-      if (typeof value.type === 'string') {
-        desc += `|${value.type}`
+      try {
+        const value = this.opValue().valueOf()
+        // $FlowFixMe
+        if (typeof value.action === 'string') {
+          desc += `|${value.action}`
+        }
+      } catch (e) {
+        console.warn('Error on .description()', this.hash(), e)
       }
     }
     return `${desc}|${this.hash()} of ${this.contractID()}>`
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -129,9 +129,10 @@ const decryptedAndVerifiedDeserializedMessage = (head: Object, headJSON: string,
     (message: any).keys = (message: any).keys?.map((key, eKeyId) => {
       return maybeEncryptedIncomingData(contractID, state, key, height, additionalKeys, headJSON, (key) => {
         if (!key.meta?.private?.content) return
-        const decryptionFn = message.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
-        // $FlowFixMe
-        const decryptionContract = ((message.foreignContractID ? message.foreignContractID : contractID): string)
+        // const decryptionFn = key.meta.private.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
+        // const decryptionContract = key.meta.private.foreignContractID ? key.meta.private.foreignContractID : contractID
+        const decryptionFn = encryptedIncomingData
+        const decryptionContract = contractID
         key.meta.private.content = decryptionFn(decryptionContract, state, key.meta.private.content, height, additionalKeys, headJSON, (value) => {
           const computedKeyId = keyId(value)
           if (computedKeyId !== key.id) {
@@ -285,9 +286,13 @@ export class SPMessage {
     // state
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
-      const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(k => [k.id, k]))
+      const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
+        const k = wk.valueOf()
+        return [k.id, k]
+      }))
       state = {
         _vm: {
+          type: head.type,
           authorizedKeys
         }
       }
@@ -447,7 +452,8 @@ export class SPMessage {
     const type = this.opType()
     let desc = `<op_${type}`
     if (type === SPMessage.OP_ACTION_UNENCRYPTED) {
-      const value = this.opValue()
+      const value = this.opValue().valueOf()
+      // $FlowFixMe
       if (typeof value.type === 'string') {
         desc += `|${value.type}`
       }
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -630,7 +630,7 @@ export default (sbp('sbp/selectors/register', {
                     return
                   }
                   sbp('chelonia/queueInvocation', msg.channelID, () => {
-                    (v: Function)(parseEncryptedOrUnencryptedMessage.call(this, {
+                    (v: Function).call(this.pubsub, parseEncryptedOrUnencryptedMessage.call(this, {
                       contractID: msg.channelID,
                       serializedData: msg.data
                     }))
@@ -1480,8 +1480,8 @@ export default (sbp('sbp/selectors/register', {
       const originatingState = originatingContract.state(originatingContractID)
 
       const havePendingKeyRequest = Object.values(originatingState._vm.authorizedKeys).findIndex((k) => {
-      // $FlowFixMe
-        return k._notAfterHeight == null && k.meta?.keyRequest?.contractID === contractID && state?._volatile?.pendingKeyRequests?.includes(k.name)
+        // $FlowFixMe
+        return k._notAfterHeight == null && k.meta?.keyRequest?.contractID === contractID && state?._volatile?.pendingKeyRequests?.some(pkr => pkr.name === k.name)
       }) !== -1
 
       // If there's a pending key request for this contract, return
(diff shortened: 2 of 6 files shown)


shared/functions.js
```diff
@@ -53,15 +53,15 @@ if (typeof globalThis === 'object' && !has(globalThis, 'Buffer')) {
...
+  return CID.create(1, multicode, digest).toString(base58btc)
 }
 
 // TODO: implement a streaming hashing function for large files.
 // Note: in fact this returns a serialized CID, not a CID object.
 export function createCID (data: string | Uint8Array, multicode: number = multicodes.RAW): string {
   const uint8array = typeof data === 'string' ? new TextEncoder().encode(data) : data
   const digest = blake2b256.digest(uint8array)
-  return CID.create(1, multicode, digest).toString(base58btc.encoder)
+  return CID.create(1, multicode, digest).toString(base58btc)
```
taoeffect, line 64: This seems like a rather significant change... what's going on here?

Why is this being done? Will this change the hashes?

corrideat: Not sure why this would seem significant, but, no, it won't change the hashes. This is actually a type-related change.

taoeffect: > Not sure why this would seem significant

It seems significant because it's pointing to an entirely different object AFAICT, no? I'll have to look more deeply into this.

corrideat: I don't think this change is significant WRT the logic, but maybe is a result of types slightly diverging in https://github.com/multiformats/js-multiformats/blob/master/src/cid.ts and https://github.com/okTurtles/chel-multiformats/blob/main/src/cid.ts

It's pointing to an entirely different object, but the object (let's call it `base`) still works.

`CID.prototype.toString()` calls `format()`, which calls `toStringV1()` (or `toStringV0()`, which then calls `base.encode()`.

If `base` is of type `Codec` (this change), then `Codec.prototype.encoder = oldBaseWithoutThisChange` and, crucially, `base.encode()` is defined as `return this.encoder.encode(input)`, so the same thing ends up being called.

shared/domains/chelonia/SPMessage.js
```diff
@@ -129,9 +129,10 @@ const decryptedAndVerifiedDeserializedMessage = (head: Object, headJSON: string,
     (message: any).keys = (message: any).keys?.map((key, eKeyId) => {
       return maybeEncryptedIncomingData(contractID, state, key, height, additionalKeys, headJSON, (key) => {
         if (!key.meta?.private?.content) return
-        const decryptionFn = message.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
-        // $FlowFixMe
-        const decryptionContract = ((message.foreignContractID ? message.foreignContractID : contractID): string)
+        // const decryptionFn = key.meta.private.foreignContractID ? encryptedIncomingForeignData : encryptedIncomingData
+        // const decryptionContract = key.meta.private.foreignContractID ? key.meta.private.foreignContractID : contractID
+        const decryptionFn = encryptedIncomingData
+        const decryptionContract = contractID
```
taoeffect, line 138: If you are just commenting out the old code and leaving it, could you also add a comment above the commented out code explaining why it was commented out and also why it is being kept around still?

shared/domains/chelonia/SPMessage.js
```diff
@@ -285,9 +286,13 @@ export class SPMessage {
     // state
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
-      const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(k => [k.id, k]))
+      const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
+        const k = wk.valueOf()
+        return [k.id, k]
+      }))
```
taoeffect: This also seems like a significant change, is this a bugfix?

shared/domains/chelonia/SPMessage.js
```diff
@@ -447,7 +452,8 @@ export class SPMessage {
     const type = this.opType()
     let desc = `<op_${type}`
     if (type === SPMessage.OP_ACTION_UNENCRYPTED) {
-      const value = this.opValue()
+      const value = this.opValue().valueOf()
+      // $FlowFixMe
```
taoeffect: Same question here, is this a bugfix?

shared/domains/chelonia/chelonia.js
```diff
@@ -1480,8 +1480,8 @@ export default (sbp('sbp/selectors/register', {
       const originatingState = originatingContract.state(originatingContractID)
 
       const havePendingKeyRequest = Object.values(originatingState._vm.authorizedKeys).findIndex((k) => {
-      // $FlowFixMe
-        return k._notAfterHeight == null && k.meta?.keyRequest?.contractID === contractID && state?._volatile?.pendingKeyRequests?.includes(k.name)
+        // $FlowFixMe
+        return k._notAfterHeight == null && k.meta?.keyRequest?.contractID === contractID && state?._volatile?.pendingKeyRequests?.some(pkr => pkr.name === k.name)
```
taoeffect, line 1484: And is this a bugfix?

shared/domains/chelonia/SPMessage.js
corrideat: This is a bugfix, in case opValue() is `SignedData`

shared/domains/chelonia/chelonia.js
corrideat: Yes, this is a bugfix because `pendingKeyRequests` contains an object (see in `utils.js`, `targetState._volatile.pendingKeyRequests.push({ contractID, name: key.name, hash, reference: reference?.data })`)

shared/domains/chelonia/SPMessage.js
corrideat: I'm not sure if it's 'significant' if were wrong (`valueOf()` is a no-op in that case). However, this is indeed a bugfix.

```ts
export type SPOpContract = { type: string; keys: (SPKey | EncryptedData<SPKey>)[]; parentContract?: string }
```

As you see, `keys[number]` could be an `SPKey` or an `EncryptedData<SPKey>`. This change correctly unwraps the encrypted data. Looking at this change again, I think we might want to use `unwrapMaybeEncryptedData` instead to prevent this from throwing. Would you agree?

shared/domains/chelonia/SPMessage.js
corrideat:
```ts
export type ProtoSPOpActionUnencrypted = { action: string; data: JSONType; meta: JSONObject }
export type SPOpActionUnencrypted = ProtoSPOpActionUnencrypted | SignedData<ProtoSPOpActionUnencrypted>

```

shared/domains/chelonia/encryptedData.js
```diff
@@ -350,6 +350,7 @@ export const isRawEncryptedData = (data: any): boolean => {
 }
 
 export const unwrapMaybeEncryptedData = (data: any): { encryptionKeyId: string | null, data: any } | void => {
+  if (data == null) return
```
corrideat, line 353: This is a 'type' fix, for making `unwrapMaybeEncryptedData` safe to call on `null` | `undefined`.

shared/domains/chelonia/utils.js
```diff
@@ -13,7 +13,7 @@ import { CONTRACT_IS_PENDING_KEY_REQUESTS } from './events.js'
 import type { SignedData } from './signedData.js'
 import { isSignedData } from './signedData.js'
 
-const MAX_EVENTS_AFTER = Number.parseInt(process.env.MAX_EVENTS_AFTER, 10) || Infinity
+const MAX_EVENTS_AFTER = Number.parseInt(process.env.MAX_EVENTS_AFTER || '', 10) || Infinity
```
corrideat, line 16: This is an insignificant change, except that TS won't like `parseInt(undefined)`

taoeffect:
It seems this PR contains many surprise bug fixes, however, the PR itself doesn't close an issue.

So I am left wondering why we didn't notice all of these bugs.

Could you please therefore update this PR's description to modify it from nothing to:

1. Explain why we didn't catch these bugs before (i.e. it seems like these are bugs - and yet they had no effect?!)
2. Explain what bugs are being fixed

shared/domains/chelonia/SPMessage.js
taoeffect: > I think we might want to use unwrapMaybeEncryptedData instead to prevent this from throwing. Would you agree?

As this is your (rather difficult to parse) code, I'm not sure I'm qualified to answer this question unless you explain the situation in more detail. Is this not a straightforward answer? If so, why not? Are there pros/cons to be considered? If so, what are they?

shared/domains/chelonia/SPMessage.js
corrideat: I'm not really sure why this original code was there to begin with. It seems like it allowed for a `meta.private.foreignContractID` attribute to define, well, a `foreignContractID` that was used for decrypting the contents (see comparable logic for `OP_KEY_SHARE`, except that `foreignContractID` there covers the entire op, not a single key). However, this has a few issues:

1. I'm not sure what the semantics of that were (maybe allowing for the use of encrypted keys in some weird key)
2. This functionality isn't implemented for `OP_KEY_ADD` or `OP_KEY_UPDATE`, or anywhere else. It seems unique to keys added using `OP_CONTRACT`
3. It doesn't seem to be used anywhere.

taoeffect: OK, please copy that 👆 into the comment then 🙏

shared/domains/chelonia/SPMessage.js
corrideat: > Is this not a straightforward answer? If so, why not?

Not sure what you're referring to (are you referring to your actual comment above, that I'm replying to; or are you referring to whether the answer I was looking for was a straightforward one?)

If the former, no, your answer isn't straightforward because you didn't actually provide an answer. If the latter, yes, it should be kind of straightforward with some additional context.

> Are there pros/cons to be considered? If so, what are they?

Generally, we don't treat decryption errors as errors. To borrow from Rust (sorry, first language that comes to mind to explain this), imagine we have `x` of type `EncryptedData`.

If we try to unwrap `x`, i.e., `x.valueOf()`, failing to decrypt will result in an error. In Rust, this is similar to invoking the `unwrap` trait on an `Option`, i.e., it'll fail hard if the option is `None` (since you're not allowed to unwrap something that doesn't exist).

Since, reiterating the previous point, generally we don't want to treat decryption errors as errors at the Chelonia level, `unwrapMaybeEncryptedData` handles unwrapping gracefully so that the operation is attempted, but decryption errors are ignored. This is why (and how) the server can process all kinds of messages with data it can't all read.

With that context, I believe we probably also want to use `unwrapMaybeEncryptedData` here. If the `.valueOf` invocation were to fail here, the entire `OP_CONTRACT` will fail (as there's no `try`), and the entire contract won't work. Probably this isn't what we want. OTOH, we aren't using this feature yet, and since `OP_KEY_ADD` can be used later, it could be a (somewhat unreasonable) requirement that `OP_CONTRACT` must only encrypted keys that can be decrypted (doing this likely will make them unprocessable by the server, unless the server has the relevant decryption key).

taoeffect: > Not sure what you're referring to (are you referring to your actual comment above, that I'm replying to; or are you referring to whether the answer I was looking for was a straightforward one?)

The latter.

> With that context, I believe we probably also want to use unwrapMaybeEncryptedData here. If the .valueOf invocation were to fail here, the entire OP_CONTRACT will fail (as there's no try), and the entire contract won't work. Probably this isn't what we want. OTOH, we aren't using this feature yet, and since OP_KEY_ADD can be used later, it could be a (somewhat unreasonable) requirement that OP_CONTRACT must only encrypted keys that can be decrypted (doing this likely will make them unprocessable by the server, unless the server has the relevant decryption key).

It's not clear to me what the context of this change / code is.

But yes, we'd probably want to fail in the least-bad way.

Do I understand correctly, this is a situation where the `OP_CONTRACT` is used on the server and it's trying to read the keys that are in it but one of them is encrypted?

If so, then yes, let's have the server just "do its best" and move on, instead of failing to process everything. We do the same for `OP_KEY_ADD`, right? So we should be consistent in our handling of it (attempt, fail, but move on).

corrideat: > Do I understand correctly, this is a situation where the OP_CONTRACT is used on the server and it's trying to read the keys that are in it but one of them is encrypted?

Correct

> Instead of failing to process everything. We do the same for OP_KEY_ADD, right? So we should be consistent in our handling of it (attempt, fail, but move on).

Agreed.

taoeffect (approved):
Nice!

PR #2814, opened 2025-05-28
No await on broadcast

### LINKED ISSUE #2812: Speed up writing to the KV store
### Problem

Clicking "Mark all as read" in notifications takes a while to react in the UI. The reason is the way the KV store update route is written.

### Solution

At the very least, we don't need to wait to broadcast the update before returning:

```js
    await sbp('backend/server/broadcastKV', contractID, key, request.payload.toString())

    return h.response().code(204)
```

Removing the `await` and adding a `.catch` instead would be one way to improve client-side performance. There might be others too, but we should at least do that.

corrideat:
Closes #2812


taoeffect (approved):
Nice and simple, and good catch re `broadcastEntry` too!

It seems it's still a bit slow in my local testing, but maybe that's just the latency of the request... this should at least speed it up a bit.

PR #2815, opened 2025-05-29
Feature detect

### LINKED ISSUE #2733: Investigate loading issues on Google Chrome on iOS
### Problem

I noticed someone tried to load https://groupincome.app and it stayed there on the loading spinner, didn't even time out.

### Solution

Find out if there's a problem with loading the app on Google Chrome on iOS and if so, fix it.

corrideat:
Closes #2733

### DIFF
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -98,6 +98,10 @@ sbp('sbp/selectors/register', {
       ])
       const swRegistration = await navigator.serviceWorker.register(`/assets/js/sw-primary.js?${params}`, { scope: '/' })
 
+      if (swRegistration.active == null && swRegistration.installing == null && swRegistration.waiting == null) {
+        throw new Error('No valid service worker found')
+      }
+
       // if an active service-worker exists, checks for the updates immediately first and then repeats it every 1hr
       await swRegistration.update()
       setInterval(() => sbp('service-worker/update'), HOURS_MILLIS)
--- a/frontend/controller/service-worker.js
+++ b/frontend/controller/service-worker.js
@@ -98,6 +98,10 @@ sbp('sbp/selectors/register', {
       ])
       const swRegistration = await navigator.serviceWorker.register(`/assets/js/sw-primary.js?${params}`, { scope: '/' })
 
+      // This probably happens if the SW immediately crashed after registration
+      // (e.g., trying to access indexedDB if it's undefined). In any case, we
+      // can't proceed when this happens and .update() below will fail. The goal
+      // of this check is to provide a more useful error message.
       if (swRegistration.active == null && swRegistration.installing == null && swRegistration.waiting == null) {
         throw new Error('No valid service worker found')
       }
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -252,7 +252,10 @@ async function startApp () {
       sbp('okTurtles.data/set', 'BANNER', bannerGeneral) // make it globally accessible
       if (!hasAllRequiredFeatures) {
         this.removeLoadingAnimation()
-        sbp('gi.ui/dangerBanner', L("Your browser doesn't support all features required to use Group Income. Please try a different browser."))
+        sbp('gi.ui/prompt', {
+          heading: L('Unsupported browser'),
+          question: L("This browser doesn't support all features required to use Group Income. Please try a different browser.")
+        })
         return
       }
       // display a self-clearing banner that shows up after we've taken 2 or more seconds
--- a/frontend/main.js
+++ b/frontend/main.js
@@ -35,6 +35,7 @@ import './views/utils/ui.js'
 import './views/utils/vError.js'
 import './views/utils/vFocus.js'
 // import './views/utils/vSafeHtml.js' // this gets imported by translations, which is part of common.js
+import hasAllRequiredFeatures from '@model/featureCheck.js'
 import Vue from 'vue'
 import notificationsMixin from './model/notifications/mainNotificationsMixin.js'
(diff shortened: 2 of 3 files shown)


frontend/main.js
```diff
@@ -249,6 +250,11 @@ async function startApp () {
       }
       const { bannerGeneral } = this.$refs
       sbp('okTurtles.data/set', 'BANNER', bannerGeneral) // make it globally accessible
+      if (!hasAllRequiredFeatures) {
+        this.removeLoadingAnimation()
+        sbp('gi.ui/dangerBanner', L("Your browser doesn't support all features required to use Group Income. Please try a different browser."))
```
corrideat: Let me know if you wish to adjust the wording here. This will display an empty page with just this banner when it happens.

frontend/controller/service-worker.js
```diff
@@ -98,6 +98,10 @@ sbp('sbp/selectors/register', {
       ])
       const swRegistration = await navigator.serviceWorker.register(`/assets/js/sw-primary.js?${params}`, { scope: '/' })
 
+      if (swRegistration.active == null && swRegistration.installing == null && swRegistration.waiting == null) {
+        throw new Error('No valid service worker found')
```
corrideat, line 106: This probably happens if the SW immediately crashed after installation (e.g., trying to access `indexedDB` if it's undefined), or that's my theory (as I have little to none observability into what actually happens). In any case, we can't proceed when this happens and `.update()` below will fail. The goal of this check is to provide a more useful error message.

frontend/main.js
dotmacro: Change "Your browser..." to "This browser..."

frontend/controller/service-worker.js
taoeffect: Could you please copy/paste your Github comment inline to the code? (In general, these sorts of comments are usually more helpful as real code comments so that future-you knows why this was added)

frontend/main.js
taoeffect: Could you please change this to a `gi.ui/prompt` instead of a banner?

This way users are less likely to miss it and it's really right there in their face as a show stopper.

frontend/controller/service-worker.js
taoeffect, line 106: Nit: o4-mini (high) recommends adding the scope: `No valid service worker found (scope: ${swRegistration.scope})`

frontend/model/featureCheck.js
```diff
@@ -0,0 +1,15 @@
...
+const hasReflect = typeof Reflect === 'object'
+const hasNavigator = typeof Navigator === 'function' && typeof navigator === 'object' && navigator instanceof Navigator
+const hasServiceWorker = hasNavigator && typeof ServiceWorkerContainer === 'function' && navigator.serviceWorker instanceof ServiceWorkerContainer
+// $FlowFixMe[cannot-resolve-name]
+const hasIndexedDB = typeof IDBFactory === 'function' && typeof indexedDB === 'object' && indexedDB instanceof IDBFactory
+const hasSessionStorage = typeof Storage === 'function' && typeof sessionStorage === 'object' && sessionStorage instanceof Storage
+
+console.info('Feature check', { hasFetch, hasReadableStream, hasWebSocket, hasReflect, hasNavigator, hasServiceWorker, hasIndexedDB, hasSessionStorage })
+
+const hasAllRequiredFeatures: boolean = hasFetch && hasReadableStream && hasWebSocket && hasReflect && hasNavigator && hasServiceWorker && hasIndexedDB && hasSessionStorage
```
taoeffect: I like this suggestion by o4-mini (high), as it avoids risking missing a variable, looks more elegant, makes logging the console.info more straightforward and DRY, and also is easier to extend in the future with other feature checks:

```js
const requiredFeatures = {
  fetch: typeof fetch === 'function',
  streams: typeof ReadableStream === 'function',
  webSocket: typeof WebSocket === 'function',
  reflect: typeof Reflect === 'object',
  navigator: typeof navigator === 'object' && 'serviceWorker' in navigator,
  indexedDB: typeof indexedDB === 'object',
  sessionStorage: typeof sessionStorage === 'object'
}

const hasAllRequiredFeatures: boolean =
  Object.values(requiredFeatures).every(Boolean)

export default hasAllRequiredFeatures
```

taoeffect (changes requested):
Great job @corrideat! Review ready here too!

frontend/controller/service-worker.js
corrideat: Not sure how that's relevant here, since there's only one scope

frontend/model/featureCheck.js
corrideat: Sure, but this is perhaps a bit slower (due to the `Object.values.every`) and mine also checks for consistency. For example, `typeof Storage === 'function' && typeof sessionStorage === 'object' && sessionStorage instanceof Storage` checks for the presence of sessionStorage as well as whether the API is what one expects.

Should I take this mostly as is or add the extra checks?

taoeffect: > Should I take this mostly as is or add the extra checks?

Oh, as far as the checks themselves go feel free to make them to your liking (I was just copy/pasting what it gave me, not recommending these updated values, but rather just the approach).

>  this is perhaps a bit slower

Here the code being concise and readable (and future-proof) is preferable to shaving off maybe a microsecond of difference at most from the page's first-load startup.

taoeffect (approved):
Looks good to me! Tested the prompt and it looks nice, without a button, I guess that makes sense.

PR #2816, opened 2025-05-29
Worker error handling

corrideat:
I've noticed that we didn't do much error handling with workers. This had the following consequences:

1. We won't know about crashes or their cause
2. Wokers crashing resulted in the application hanging when trying to send a worker task

This logic implements error reporting and automatic relaunch.

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -32,27 +32,47 @@ const createWorker = (path: string): {
   rpcSbp: (...args: any) => Promise<any>,
   terminate: () => Promise<number>
 } => {
-  const worker = new Worker(path)
-  const ready = new Promise((resolve, reject) => {
-    worker.on('message', (msg) => {
-      if (msg === 'ready') resolve()
+  let worker
+  let ready
+
+  const launchWorker = () => {
+    worker = new Worker(path)
+    ready = new Promise((resolve, reject) => {
+      worker.on('message', (msg) => {
+        if (msg === 'ready') {
+          worker.off('error', reject)
+          worker.on('error', (e) => {
+            console.error(e, 'Error on worker', path)
+            launchWorker()
+            ready.catch(e => {
+              console.error(e, 'Error on worker relaunch', path)
+              process.exit(1)
+            })
+          })
+          resolve()
+        }
+      })
+      worker.on('error', reject)
     })
-    worker.on('error', reject)
-  })
+  }
+  launchWorker()
 
   const rpcSbp_ = (...args: any) => {
-    return new Promise((resolve, reject) => {
+    return ready.then(() => new Promise((resolve, reject) => {
       const mc = new MessageChannel()
       mc.port2.onmessage = (event) => {
+        worker.off('error', reject)
         const [success, result] = ((event.data: any): [boolean, any])
         if (success) return resolve(result)
         reject(result)
       }
       mc.port2.onmessageerror = () => {
+        worker.off('error', reject)
         reject(Error('Message error'))
       }
       worker.postMessage([mc.port1, ...args], [mc.port1])
-    })
+      worker.once('error', reject)
+    }))
   }
 
   let rpcSbp = (...args: any) => ready.then(() => rpcSbp_(...args))
--- a/backend/server.js
+++ b/backend/server.js
@@ -33,17 +33,17 @@ const createWorker = (path: string): {
   terminate: () => Promise<number>
 } => {
   let worker
-  let ready
+  let ready: Promise<void>
 
   const launchWorker = () => {
     worker = new Worker(path)
-    ready = new Promise((resolve, reject) => {
+    return new Promise((resolve, reject) => {
       worker.on('message', (msg) => {
         if (msg === 'ready') {
           worker.off('error', reject)
           worker.on('error', (e) => {
             console.error(e, 'Error on worker', path)
-            launchWorker()
+            ready = launchWorker()
             ready.catch(e => {
               console.error(e, 'Error on worker relaunch', path)
               process.exit(1)
@@ -55,9 +55,9 @@ const createWorker = (path: string): {
       worker.on('error', reject)
     })
   }
-  launchWorker()
+  ready = launchWorker()
 
-  const rpcSbp_ = (...args: any) => {
+  const rpcSbp = (...args: any) => {
     return ready.then(() => new Promise((resolve, reject) => {
       const mc = new MessageChannel()
       mc.port2.onmessage = (event) => {
@@ -74,12 +74,6 @@ const createWorker = (path: string): {
       worker.once('error', reject)
     }))
   }
-
-  let rpcSbp = (...args: any) => ready.then(() => rpcSbp_(...args))
-  // Avoid unncessary `ready.then` if we already know the promise has resolved
-  ready.then(() => {
-    rpcSbp = rpcSbp_
-  })
 
   return {
     ready,
--- a/backend/server.js
+++ b/backend/server.js
@@ -3,7 +3,7 @@
 import Hapi from '@hapi/hapi'
 import sbp from '@sbp/sbp'
 import chalk from 'chalk'
-import { join } from 'node:path'
+import { basename, join } from 'node:path'
 import { Worker } from 'node:worker_threads'
 import { SPMessage } from '~/shared/domains/chelonia/SPMessage.js'
 import '~/shared/domains/chelonia/chelonia.js'
@@ -38,21 +38,26 @@ const createWorker = (path: string): {
   const launchWorker = () => {
     worker = new Worker(path)
     return new Promise((resolve, reject) => {
-      worker.on('message', (msg) => {
+      const msgHandler = (msg) => {
         if (msg === 'ready') {
           worker.off('error', reject)
           worker.on('error', (e) => {
-            console.error(e, 'Error on worker', path)
-            ready = launchWorker()
-            ready.catch(e => {
-              console.error(e, 'Error on worker relaunch', path)
+            console.error(e, `Running worker ${basename(path)} terminated. Attempting relaunch...`)
+            worker.off('message', msgHandler)
+            // This won't result in an infinite loop because of exiting and
+            // because this handler is only executed after the 'ready' event
+            // Relaunch can happen multiple times, so long as the worker doesn't
+            // immediately fail.
+            ready = launchWorker().catch(e => {
+              console.error(e, `Error on worker ${basename(path)} relaunch`)
               process.exit(1)
             })
           })
           resolve()
         }
-      })
-      worker.on('error', reject)
+      }
+      worker.on('message', msgHandler)
+      worker.once('error', reject)
     })
   }
   ready = launchWorker()
@@ -60,17 +65,24 @@ const createWorker = (path: string): {
   const rpcSbp = (...args: any) => {
     return ready.then(() => new Promise((resolve, reject) => {
       const mc = new MessageChannel()
-      mc.port2.onmessage = (event) => {
+      const cleanup = ((worker) => () => {
         worker.off('error', reject)
+        mc.port2.onmessage = null
+        mc.port2.onmessageerror = null
+      })(worker)
+      mc.port2.onmessage = (event) => {
+        cleanup()
         const [success, result] = ((event.data: any): [boolean, any])
         if (success) return resolve(result)
         reject(result)
       }
       mc.port2.onmessageerror = () => {
-        worker.off('error', reject)
+        cleanup()
         reject(Error('Message error'))
       }
       worker.postMessage([mc.port1, ...args], [mc.port1])
+      // If the worker itself breaks during an SBP call, we want to make sure
+      // this promise immediately rejects
       worker.once('error', reject)
     }))
   }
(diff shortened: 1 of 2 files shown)


shared/domains/chelonia/SPMessage.js
```diff
@@ -290,7 +290,7 @@ export class SPMessage {
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
       const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
-        const k = unwrapMaybeEncryptedData(wk.valueOf())
+        const k = unwrapMaybeEncryptedData(wk)
```
corrideat, line 293: This is unrelated to workers, but a small change that's needed for the use of `unwrapMaybeEncryptedData` to make sense. `unwrapMaybeEncryptedData` is supposed to take `X | EncryptedData<X>`, and the previous use was only `X`, since `.valueOf()` already unwraps.

backend/server.js
```diff
@@ -32,35 +32,49 @@ const createWorker = (path: string): {
...
 
-  const rpcSbp_ = (...args: any) => {
+  const launchWorker = () => {
+    worker = new Worker(path)
     return new Promise((resolve, reject) => {
+      worker.on('message', (msg) => {
+        if (msg === 'ready') {
+          worker.off('error', reject)
+          worker.on('error', (e) => {
+            console.error(e, 'Error on worker', path)
```
taoeffect: Could you update this error message to say: `Running worker ${filename(path)} terminated. Attempting relaunch...`

backend/server.js
```diff
@@ -32,35 +32,49 @@ const createWorker = (path: string): {
...
+            ready = launchWorker()
+            ready.catch(e => {
+              console.error(e, 'Error on worker relaunch', path)
+              process.exit(1)
+            })
```
taoeffect: So this allows at most 1 worker relaunch per runtime instance?

Does that make sense to do? Can you add a comment here explaining rationale?

backend/server.js
```diff
@@ -32,35 +32,49 @@ const createWorker = (path: string): {
...
         if (success) return resolve(result)
         reject(result)
       }
       mc.port2.onmessageerror = () => {
+        worker.off('error', reject)
         reject(Error('Message error'))
       }
       worker.postMessage([mc.port1, ...args], [mc.port1])
-    })
+      worker.once('error', reject)
```
taoeffect, line 86: Can you add a comment above this line to the effect of:

```js
// If the worker itself breaks during an SBP call, we want to make sure this promise immediately rejects
```

backend/server.js
```diff
@@ -32,35 +32,49 @@ const createWorker = (path: string): {
...
-  const rpcSbp_ = (...args: any) => {
+  const launchWorker = () => {
+    worker = new Worker(path)
     return new Promise((resolve, reject) => {
+      worker.on('message', (msg) => {
+        if (msg === 'ready') {
+          worker.off('error', reject)
+          worker.on('error', (e) => {
+            console.error(e, 'Error on worker', path)
+            ready = launchWorker()
```
taoeffect: If we call `launchWorker()` here like this, it will immediately switch the value of `worker`, and based on a comment from Claude I'm just wondering if that's OK or if we need to first turn off any event listeners?

backend/server.js
taoeffect, line 86: A separate concern from Claude Opus 4, and I actually don't know whether it's right or not here about these other listeners needing to be cleaned up:

### 2. Memory Leak with Event Listeners

```js
mc.port2.onmessage = (event) => {
  worker.off('error', reject)
  const [success, result] = ((event.data: any): [boolean, any])
  if (success) return resolve(result)
  reject(result)
}
mc.port2.onmessageerror = () => {
  worker.off('error', reject)
  reject(Error('Message error'))
}
```

**Problem**: The `worker.off('error', reject)` is called multiple times but `worker.once('error', reject)` is used, which should automatically remove the listener. However, if neither `onmessage` nor `onmessageerror` fires (e.g., worker crashes), the error listener might not be cleaned up.

**Suggested improvement**:
```js
const rpcSbp = (...args: any) => {
  return ready.then(() => new Promise((resolve, reject) => {
    const mc = new MessageChannel()
    const cleanup = () => {
      worker.off('error', reject)
      mc.port2.onmessage = null
      mc.port2.onmessageerror = null
    }
    
    mc.port2.onmessage = (event) => {
      cleanup()
      const [success, result] = ((event.data: any): [boolean, any])
      if (success) return resolve(result)
      reject(result)
    }
    mc.port2.onmessageerror = () => {
      cleanup()
      reject(Error('Message error'))
    }
    worker.postMessage([mc.port1, ...args], [mc.port1])
    worker.once('error', (err) => {
      cleanup()
      reject(err)
    })
  }))
}
```

backend/server.js
taoeffect: I _think_ Claude Opus 4 is mistaken here, that it should exit, but .. not 100% sure:

### 3. Potential Infinite Restart Loop

```js
ready = launchWorker()
ready.catch(e => {
  console.error(e, 'Error on worker relaunch', path)
  process.exit(1)
})
```

**Problem**: If the worker fails during restart, it will try to restart again, potentially creating an infinite loop until `process.exit(1)` is called. Consider implementing a retry limit.

**Suggested improvement**:
```js
let retryCount = 0
const MAX_RETRIES = 3

const launchWorker = () => {
  // ... existing code ...
  worker.on('error', (e) => {
    console.error(e, 'Error on worker', path)
    if (++retryCount >= MAX_RETRIES) {
      console.error(`Worker ${path} failed ${MAX_RETRIES} times, exiting`)
      process.exit(1)
    }
    ready = launchWorker()
  })
  // Reset retry count on successful initialization
  retryCount = 0
  resolve()
  // ... rest of the code ...
}
```

shared/domains/chelonia/SPMessage.js
taoeffect, line 293: This change is hard for me to review, and you would know better, so I'm just copy/pasting what Opus and o4-mini (high) have to say here and letting you decide:

## Opus 4 (Thinking)

<details>
### Potentially Breaking Change

```diff
- const k = unwrapMaybeEncryptedData(wk.valueOf())
+ const k = unwrapMaybeEncryptedData(wk)
```

**Analysis**: Looking at the `unwrapMaybeEncryptedData` function in the context, it checks if the input `isEncryptedData(data)` and then calls `data.valueOf()`. The change suggests that `wk` might already be the correct type to pass directly. However, this could be a breaking change if `wk` has its own `valueOf()` method that returns something different than `wk` itself.

Since `unwrapMaybeEncryptedData` will call `valueOf()` internally if needed, this change should be safe, but it's worth verifying that `wk` doesn't have a custom `valueOf()` implementation that was being relied upon.

The change appears correct given that `unwrapMaybeEncryptedData` handles the `valueOf()` call internally when needed.
</details>

## o4-mini (high)

<details>
You changed:

```diff
-        const k = unwrapMaybeEncryptedData(wk.valueOf())
+        const k = unwrapMaybeEncryptedData(wk)
```

### Why this can bite you

1) In **node** (i.e. `process.env.BUILD !== 'web'`),  
   `unwrapMaybeEncryptedData(encryptedDataInstance)` returns `undefined`  
   because you explicitly bail out in non‐web builds.  
   But `unwrapMaybeEncryptedData(wk.valueOf())` on the **raw** object  
   would treat it as a _plain_ object and return `{ data: … }`.

2) If `wk` is a Flow‐level `EncryptedData<SPKey>`, calling `.valueOf()` first  
   means you decrypt it; passing the wrapper directly may never decrypt it  
   on node, so `authorizedKeys` ends up missing that key.

### Suggested fix

If your goal here is to build the *initial* `authorizedKeys` map from an OP_CONTRACT  
payload, you almost certainly want to support both plain and encrypted keys––and in node  
you *do* want to pull `.valueOf()` out of any `EncryptedData` you see.  For clarity, I’d write:

```js
import { isEncryptedData } from './encryptedData.js'

…  
  const signed = rawSignedIncomingData(parsedValue)
  const keys   = signed.valueOf().keys || []
  const authorizedKeys = Object.fromEntries(
    keys.map((wk /*: SPKey | EncryptedData<SPKey> */) => {
      // 1) if it’s an EncryptedData wrapper, decrypt it immediately
      const clear = isEncryptedData(wk) ? wk.valueOf() : wk
      // 2) now unwrapMaybeEncryptedData is only for “did we unwrap JSON‐side?”
      const maybe = unwrapMaybeEncryptedData(clear) || { data: clear }
      return [maybe.data.id, maybe.data]
    })
  )
  state = { _vm: { type: head.manifest, authorizedKeys } }
}
```

#### Benefits of this approach

- You explicitly decrypt any `EncryptedData<SPKey>` in node.  
- You still handle the case where some keys might be raw arrays (unlikely but safe).  
- You remove the murkiness of chaining `.valueOf()` into `unwrapMaybeEncryptedData`.

</details>

taoeffect (changes requested):
Great work @corrideat! These do seems like meaningful improvements to the error handling logic.

I've added my comments, and I've filtered some of Opus 4's comments and added them in-line as needed.

For o4 mini (high), I'm also including this commentary here in case it's helpful but feel free to decide whether or not it's worthwhile:

## o4-mini (high)

<details>

## 1) backend/server.js – worker restart logic is brittle

### A) Initial failures never restart

You only attach your “auto‐restart” logic **after** the first `ready` arrives.  That means if the worker crashes *before* it ever says `"ready"`, your `ready` promise will reject and nobody ever tries to spin it up again:

```diff
  const launchWorker = () => {
    worker = new Worker(path)
    return new Promise((resolve, reject) => {
-     worker.on('message', (msg) => {
+     worker.on('message', (msg) => {
        if (msg === 'ready') {
          worker.off('error', reject)
          // ← restart logic lives here, but only AFTER first “ready”
          worker.on('error', (e) => {
            console.error(e, 'Error on worker', path)
            ready = launchWorker()  // ← OK on 2nd+ failures
```

**Problem**  
– If the very first `Worker` instance throws an error prior to sending `"ready"`, `launchWorker()` will call `reject()` and nothing ever re‐launches your worker.

**Fix**  
Attach your “restart on error” handler **immediately** (not only after `ready`).  Also, terminate the old worker so you don’t leak threads.  Finally, handle an unexpected `'exit'` as well as `'error'`.

Here is a more robust refactor of `createWorker`:

```js
import { Worker, MessageChannel } from 'worker_threads'

function createWorker(path /*: string */) /*: {
  ready: Promise<void>,
  rpcSbp: (...args: any[]) => Promise<any>,
  terminate: () => Promise<number>
} */ {
  let worker /*: Worker */
  let ready /*: Promise<void> */
  let _resolve /*: () => void */
  let _reject  /*: (err: any) => void */

  function spawn() {
    worker = new Worker(path)

    // prepare a fresh “ready” promise
    ready = new Promise((resolve, reject) => {
      _resolve = resolve
      _reject  = reject
    })

    // 1) If worker signals “ready”, resolve
    worker.once('message', msg => {
      if (msg === 'ready') _resolve()
    })

    // 2) On any error, kill & respawn
    const onFailure = (err) => {
      console.error(`Worker '${path}' failed:`, err)
      // clean up
      worker.removeAllListeners()
      worker.terminate().catch(()=>{})
      // respawn
      spawn()
    }
    worker.on('error', onFailure)
    worker.on('exit', code => {
      if (code !== 0) {
        console.error(`Worker '${path}' exited with code ${code}, restarting...`)
        onFailure(new Error(`exit code ${code}`))
      }
    })
  }

  // kick off first spawn
  spawn()

  // rpc method that waits for “ready”
  function rpcSbp(...args /*: any[] */) /*: Promise<any> */ {
    return ready.then(() => {
        const mc = new MessageChannel()
        mc.port2.once('message', ([ok, result]) => {
          resolve(ok ? result : Promise.reject(result))
        })
        mc.port2.once('messageerror', () => reject(new Error('messageerror')))
        // if the worker itself errors during this RPC, reject
        worker.once('error', reject)

        worker.postMessage([mc.port1, ...args], [mc.port1])
      })
    })
  }

  return {
    ready,
    rpcSbp,
    terminate: () => worker.terminate()
  }
}
```

**Key improvements**  
- We call `spawn()` immediately on **any** `'error'` or `'exit'`, even before `"ready"`.  
- We call `worker.terminate()` on the old thread, so you don’t leak threads.  
- We use `.once(...)` instead of `.on(...)` for one‐time events, so handlers are automatically removed.
</details>

backend/server.js
corrideat: No, this allows for multiple relaunches, but it terminates if relaunching immediately fails.

shared/domains/chelonia/SPMessage.js
corrideat: You'd need to review what `unwrapMaybeEncryptedData` does to understand this change, but, like Opus 4 said, `unwrapMaybeEncryptedData` already calls `.valueOf()`

backend/server.js
corrideat: That's OK so long as cleaning up uses the old references

taoeffect (approved):
Excellent!

PR #2817, opened 2025-05-29
Address KV issues on unread messages

### LINKED ISSUE #2780: Heisenbug in group-chat-message.spec: missing notifications
### Problem

When running `grunt test` locally, I sometimes get this error on #2737 (and probably `master` as well):

https://github.com/user-attachments/assets/a6eec431-d51d-4932-b218-65b76ae0e096

"2 new notifications" is expected.

EDIT: here's an example test https://cloud.cypress.io/projects/q6whky/runs/4590/test-results

EDIT2: here's a really weird one, similar, but different, and also doesn't make sense because the UI shows 1 notif + the code expects 1: https://cloud.cypress.io/projects/q6whky/runs/4596/test-results

### Solution

Find out what the bug is and fix.

corrideat:
Fix #2780

### DIFF
--- a/frontend/model/contracts/chatroom.js
+++ b/frontend/model/contracts/chatroom.js
@@ -356,17 +356,17 @@ sbp('chelonia/defineContract', {
           return
         }
 
-        const newMessage = createMessage({ meta, data, hash, height, state, innerSigningContractID })
-
-        // Ensure that this happens after everything has been initialised
+        // Ensure that this happens after everything has been initialised, i.e.,
+        // after all side effects from joins and rejoins have been processed.
+        // (In particular, this should happen after `kv/initChatRoomUnreadMessages`.)
+        // See <https://github.com/okTurtles/group-income/issues/2780>
         sbp('chelonia/queueInvocation', contractID, async () => {
           const state = await sbp('chelonia/contract/state', contractID)
-          const identityContractID = sbp('state/vuex/state').loggedIn.identityContractID
-
-          if (!state?.members?.[identityContractID]) {
+          if (!state?.members?.[me]) {
             return
           }
 
+          const newMessage = createMessage({ meta, data, hash, height, state, innerSigningContractID })
           sbp('okTurtles.events/emit', MESSAGE_RECEIVE_RAW, {
             contractID,
             data,
@@ -409,10 +409,19 @@ sbp('chelonia/defineContract', {
           return
         }
 
-        sbp('okTurtles.events/emit', MESSAGE_RECEIVE_RAW, {
-          contractID,
-          data,
-          innerSigningContractID
+        // Ensure that this happens after everything has been initialised, i.e.,
+        // after all side effects from joins and rejoins have been processed.
+        sbp('chelonia/queueInvocation', contractID, async () => {
+          const state = await sbp('chelonia/contract/state', contractID)
+          if (!state?.members?.[me]) {
+            return
+          }
+
+          sbp('okTurtles.events/emit', MESSAGE_RECEIVE_RAW, {
+            contractID,
+            data,
+            innerSigningContractID
+          })
         })
       }
     },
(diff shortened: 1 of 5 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -1684,12 +1684,14 @@ export default (sbp('sbp/selectors/register', {
...
-          meta: key
-        })
+        const serializedDataText = await response.text()
+        currentValue = serializedDataText
+          ? parseEncryptedOrUnencryptedMessage.call(this, {
+            contractID,
+            serializedData: JSON.parse(serializedDataText),
+            meta: key
+          })
+          : undefined
```
taoeffect: 409 will return:

```js
return h.response(existing).etag(cid).header('x-cid', `"${cid}"`).code(409)
```

Which will still be truthy and therefore will be passed to `JSON.parse`.

Same for 412:

```js
      // We need this `x-cid` because HAPI modifies Etags
        return h.response(existing).etag(cid).header('x-cid', `"${cid}"`).code(412)
```

(boyscouting: that `// We` should be on the same line as `return`...)

frontend/model/contracts/chatroom.js
```diff
@@ -358,11 +358,21 @@ sbp('chelonia/defineContract', {
...
+          if (!state?.members?.[identityContractID]) {
+            return
+          }
+
+          sbp('okTurtles.events/emit', MESSAGE_RECEIVE_RAW, {
+            contractID,
+            data,
+            innerSigningContractID,
+            newMessage
+          })
```
taoeffect: There's some inconsistency going on here and also some DRY violations:

- Inconsistency: this queueInvocation is being done for `addMessage` but not for other chatroom events like `editMessage`, `deleteMessage`, and others
- DRY violation: if you're planning on bailing early (but later), then you might as well consolidate the earlier lines that grab the `me` variable into here as well, and move everything else, including the `newMessage` creation into the queueInvocation block

Finally, the comment says:

```js
// Ensure that this happens after everything has been initialised
```

But it doesn't say why. Could you please expand it to explain why this is being done? To that effect, where in this PR is the actual heisenbug fix? Wherever it is, could you please add a comment there and link to the corresponding issue?

taoeffect (changes requested):
Nice work @corrideat! It will be amazing if this PR fixes the heisenbug. Review is ready, note that I couldn't quite tell where the fix is though (and that's one of the RFCs)

frontend/model/contracts/chatroom.js
```diff
@@ -469,7 +488,18 @@ sbp('chelonia/defineContract', {
...
+        // after all side effects from joins and rejoins have been processed.
+        sbp('chelonia/queueInvocation', contractID, async () => {
+          const state = await sbp('chelonia/contract/state', contractID)
+          if (!state?.members?.[me]) {
+            return
+          }
+
+          sbp('gi.actions/identity/kv/removeChatRoomUnreadMessage', { contractID, messageHash: data.hash }).catch(e => {
+            console.error('[gi.contracts/chatroom/deleteMessage/sideEffect] Error calling removeChatRoomUnreadMessage', e)
+          })
```
taoeffect, line 501: Similar to this boyscoutting change, I'm noticing that inside of the listener for `MESSAGE_RECEIVE_RAW` there's a naked call with no `.catch`:

```js
sbp('gi.actions/identity/kv/removeChatRoomUnreadMessage', { contractID, messageHash: data.hash })
```

Does it need one too?

EDIT: and inside of `messageReceivePostEffect` there's:

```js
    if (shouldAddToUnreadMessages) {
      sbp('gi.actions/identity/kv/addChatRoomUnreadMessage', { contractID, messageHash, createdHeight: height })
    }
```

Although that would be caught by the handler on the caller:

```js
  messageReceivePostEffect({
    contractID,
    messageHash: msgData.hash,
    height: msgData.height,
    text: msgData.text,
    isDMOrMention: isMentionedMe || state.attributes?.type === CHATROOM_TYPES.DIRECT_MESSAGE,
    messageType: !newMessage ? MESSAGE_TYPES.TEXT : data.type,
    memberID: innerSigningContractID,
    chatRoomName: state.attributes?.name
  }).catch(e => {
    console.error('[action/chatroom.js] Error on messageReceivePostEffect', e)
  })
```

corrideat: Yes, likely it does.

taoeffect (approved):
Looks good! Let's hope this does it 🤞 

Great work @corrideat tracking this down!

PR #2819, opened 2025-05-29
Use username for notification templates

### LINKED ISSUE #2774: Notification for non-monetary contribution shows contractID instead of username/displayName
### Problem

Image

Image

While testing #2772 locally, after u2 was added to the group I had them marked as receiving income and added a non-monetary pledge. This showed up as the notification above.

### Solution

The identity contractID should never be shown to users, especially when members are existing members part of a group. Figure out why it's being shown and fix.

When you figure out what's causing the problem, please double-check all other relevant places for the same issue.

corrideat:
Fix #2774

### DIFF
--- a/frontend/model/notifications/templates.js
+++ b/frontend/model/notifications/templates.js
@@ -533,8 +533,8 @@ export default ({
       })
     }
     const plaintextBodyContentMap = {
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Nice!

PR #2820, opened 2025-05-29
ModalTemplate.vue: Invalid prop: type check failed for prop "a11yTitle"

### LINKED ISSUE #2766: Blank page on refresh
### Problem

I sometimes see a blank background (in dark-mode, just the dark grey background) when opening the app on my phone. Today I was able to see this on FF desktop too.

I happened after I was testing a newly created group and was on the pending join group screen. It wasn't joining because of the FF bug where the web socket isn't running. So I refreshed the page a few times and it still wasn't joining, so I then did a hard refresh (<kbd>cmd+shift+r</kbd>) and saw this:

Image

I then refreshed (normal <kbd>cmd+r</kbd>) and saw this:

Image

Refreshed one more time and it finally joined the group.

### Solution

See if it's possible to avoid this situation so that at least it tries again and loads the site without the user having to refresh manually.

### LINKED ISSUE #2792: Vue error in ModalTemplate.vue: Invalid prop: type check failed for prop "a11yTitle".
### Problem

Noticed this in a WebKit browser (e.g. like Safari):

Image

### Solution

Fix.

taoeffect:
Note: I searched and couldn't find the actual place the title was missing from, so added a default.

Note: this might close #2766 too

### DIFF
--- a/frontend/views/components/modal/ModalTemplate.vue
+++ b/frontend/views/components/modal/ModalTemplate.vue
@@ -3,7 +3,7 @@
     data-test='modal'
     role='dialog'
     tabindex='-1'
-    :aria-label='a11yTitle'
+    :aria-label='a11yTitle || L("Modal Title Missing")'
     v-focus=''
   )
     transition(name='fade' appear)
--- a/strings/english.json
+++ b/strings/english.json
@@ -513,6 +513,7 @@
   "Mincome goal streak of: {streak} months": "Mincome goal streak of: {streak} months",
   "Mincome must be greater than 0": "Mincome must be greater than 0",
   "Minimum Income": "Minimum Income",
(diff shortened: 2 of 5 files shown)


corrideat:
Why not add a `if (process.env.CI) throw Error('Missing a11yTitle')` to detect this on tests?

taoeffect:
> Why not add a if (process.env.CI) throw Error('Missing a11yTitle') to detect this on tests?

It already throws and the tests don't fail. It's unclear exactly which modal has this. It seems to be possibly a dynamic modal of some sort, maybe via `gi.ui/prompt`? I'm not sure. So adding it wouldn't prevent the tests from failing, and it seems to happen in extremely rare circumstances.

corrideat (approved):
Well done

PR #2821, opened 2025-05-29
Limits for generic proposal

### LINKED ISSUE #2818: Set a reasonable character limit for generic proposal
### Problem

When creating a generic proposal, the "Name your proposal" text box seems to accept an unlimited number of characters. (For reference, I gave up after hundreds of emoji chars.)
Note: The long proposal name did not seem to cause any issues with the UI.

### Solution

- Indicate the user has reached the char limit the same way we do elsewhere

corrideat:
Closes #2818

### DIFF
--- a/frontend/model/contracts/group.js
+++ b/frontend/model/contracts/group.js
@@ -31,6 +31,8 @@ import {
   PROPOSAL_GENERIC,
   PROPOSAL_GROUP_SETTING_CHANGE,
   PROPOSAL_INVITE_MEMBER,
+  PROPOSAL_NAME_MAX_CHAR,
+  PROPOSAL_REASON_MAX_CHAR,
   PROPOSAL_PROPOSAL_SETTING_CHANGE,
   PROPOSAL_REMOVE_MEMBER,
   STATUS_CANCELLED, STATUS_EXPIRED, STATUS_OPEN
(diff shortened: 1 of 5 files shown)


frontend/model/contracts/group.js
```diff
@@ -629,7 +631,10 @@ sbp('chelonia/defineContract', {
       validate: actionRequireActiveMember((data, { state }) => {
         objectOf({
           proposalType: proposalType,
-          proposalData: object, // data for Vue widgets
+          proposalData: objectOf({
+            name: stringMax(PROPOSAL_NAME_MAX_CHAR),
+            reason: stringMax(PROPOSAL_REASON_MAX_CHAR)
+          }), // data for Vue widgets
```
taoeffect: A simple fix for the failing build would be to change this from `objectOf` to `objectMaybeOf`. That should be sufficient, although be careful as `proposalData` is intended to support arbitrary data. For the most part, `name` and `reason` should be universal across proposals... but there's a chance they might not be. If the tests pass, I'd say it's safe to leave it as `objectMaybeOf`.

corrideat: Indeed, this was a quick change, but I wanted to make sure later that this works the way it should. Maybe I'll address this as suggested now, and on our next call we could go over how to validate in a general manner.

taoeffect: I would prefer to merge this before our next call, so if `objectMaybeOf` fixes the failing build, I'm OK leaving it that way, I think it should be fine.

taoeffect (approved):
Fantastic work @corrideat !

PR #2825, opened 2025-05-31
Libcheloniajs

corrideat:
Running tests I noticed that sometimes `grunt test` fails with `Failed to fetch`. Further investigation shows that this might be because of the server not being ready to receive requests at the time the tests are launched. This could be fixed in `Gruntfile.js` by only marking the `backend:relaunch` task as ready after confirmation that the server is up and running.


backend/auth.js
```diff
@@ -2,7 +2,7 @@
 // https://hapijs.com/tutorials/auth
 // https://hapijs.com/tutorials/plugins
 
-import { verifyShelterAuthorizationHeader } from '~/shared/domains/chelonia/utils.js'
+import { verifyShelterAuthorizationHeader } from 'libchelonia/utils'
```
taoeffect: `@chelonia/lib`

taoeffect (changes requested):
@corrideat `@chelonia/lib` is now on NPM! Can you please replace all instances of `libchelonia` with `@chelonia/lib`?

taoeffect (approved):
Everything seems to work!! 🎉🎉🎉🎉🎉

PR #2833, opened 2025-06-03
Use something other than BUILD=web

### LINKED ISSUE #2805: Rename process.env.BUILD environment variable
### Problem

While attempting to create a demo of using Chelonia in a nodejs app, I noticed that unencrypting data didn't work.

It turned out that I needed to set `BUILD=web` for that `unwrapMaybeEncryptedData` would work properly. This doesn't make sense to do for a terminal nodejs app.

### Solution

Rename this variable to something else, and set it in the server. The reason this variable was created was, in part, to prevent the server's logs from filling up with `Unable to decrypt` messages.

So the server should set a variable, something along the lines of `CHEL_AVOID_DECRYPTION_ATTEMPTS=true`

corrideat:
Closes #2805

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -224,6 +224,7 @@ module.exports = (grunt) => {
         'process.env.CONTRACTS_VERSION': `'${CONTRACTS_VERSION}'`,
         'process.env.GI_VERSION': `'${GI_VERSION}'`,
         'process.env.GI_GIT_VERSION': `'${GI_GIT_VERSION}'`,
+        'process.env.CHEL_AVOID_DECRYPTION_ATTEMPTS': 'undefined',
         'process.env.LIGHTWEIGHT_CLIENT': `'${LIGHTWEIGHT_CLIENT}'`,
         'process.env.MAX_EVENTS_AFTER': `'${MAX_EVENTS_AFTER}'`,
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -224,7 +224,6 @@ module.exports = (grunt) => {
         'process.env.CONTRACTS_VERSION': `'${CONTRACTS_VERSION}'`,
         'process.env.GI_VERSION': `'${GI_VERSION}'`,
         'process.env.GI_GIT_VERSION': `'${GI_GIT_VERSION}'`,
-        'process.env.CHEL_AVOID_DECRYPTION_ATTEMPTS': 'undefined',
         'process.env.LIGHTWEIGHT_CLIENT': `'${LIGHTWEIGHT_CLIENT}'`,
         'process.env.MAX_EVENTS_AFTER': `'${MAX_EVENTS_AFTER}'`,
         'process.env.NODE_ENV': `'${NODE_ENV}'`,
@@ -448,7 +447,7 @@ module.exports = (grunt) => {
     const fork2 = function () {
       grunt.log.writeln('backend: forking...')
       child = fork(backendIndex, process.argv, {
-        env: { CHEL_AVOID_DECRYPTION_ATTEMPTS: "'true'", NODE_ENV, ...process.env },
+        env: { NODE_ENV, ...process.env },
         execArgv: ['--require', '@babel/register']
       })
       child.on('error', (err) => {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -447,7 +447,7 @@ module.exports = (grunt) => {
     const fork2 = function () {
       grunt.log.writeln('backend: forking...')
       child = fork(backendIndex, process.argv, {
-        env: { NODE_ENV, ...process.env },
+        env: { CHEL_AVOID_DECRYPTION_ATTEMPTS: "'true'", NODE_ENV, ...process.env },
         execArgv: ['--require', '@babel/register']
       })
       child.on('error', (err) => {
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -279,7 +279,7 @@ export class SPMessage {
     return new this(messageToParams(head, targetOp[1]))
   }
 
-  static deserialize (value: string, additionalKeys?: Object, state?: Object, unwrapMaybeEncryptedDataOverride?: Function = unwrapMaybeEncryptedData): this {
+  static deserialize (value: string, additionalKeys?: Object, state?: Object, unwrapMaybeEncryptedDataFn?: Function = unwrapMaybeEncryptedData): this {
     if (!value) throw new Error(`deserialize bad value: ${value}`)
     const { head: headJSON, ...parsedValue } = JSON.parse(value)
     const head = JSON.parse(headJSON)
@@ -290,7 +290,7 @@ export class SPMessage {
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
       const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
-        const k = unwrapMaybeEncryptedDataOverride(wk)
+        const k = unwrapMaybeEncryptedDataFn(wk)
         if (!k) return null
         return [k.data.id, k.data]
       }).filter(Boolean))
@@ -407,6 +407,10 @@ export class SPMessage {
     if (this._decryptedValue) return this._decryptedValue
     try {
       const value = this.message()
+      // TODO: This uses `unwrapMaybeEncryptedData` instead of a configurable
+      // version based on `skipDecryptionAttempts`. This is fine based on current
+      // use, and also something else might be confusing based on the explicit
+      // name of this function, `decryptedValue`.
       const data = unwrapMaybeEncryptedData(value)
       // Did decryption succeed? (unwrapMaybeEncryptedData will return undefined
       // on failure)
--- a/shared/domains/chelonia/SPMessage.js
+++ b/shared/domains/chelonia/SPMessage.js
@@ -279,7 +279,7 @@ export class SPMessage {
     return new this(messageToParams(head, targetOp[1]))
   }
 
-  static deserialize (value: string, additionalKeys?: Object, state?: Object): this {
+  static deserialize (value: string, additionalKeys?: Object, state?: Object, unwrapMaybeEncryptedDataOverride?: Function = unwrapMaybeEncryptedData): this {
     if (!value) throw new Error(`deserialize bad value: ${value}`)
     const { head: headJSON, ...parsedValue } = JSON.parse(value)
     const head = JSON.parse(headJSON)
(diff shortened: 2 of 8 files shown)


Gruntfile.js
```diff
@@ -447,7 +448,7 @@ module.exports = (grunt) => {
     const fork2 = function () {
       grunt.log.writeln('backend: forking...')
       child = fork(backendIndex, process.argv, {
-        env: { NODE_ENV, ...process.env },
+        env: { CHEL_AVOID_DECRYPTION_ATTEMPTS: "'true'", NODE_ENV, ...process.env },
```
taoeffect: Would it be possible to somehow set this inside of the server code itself instead of here?

It's really the server that's saying, "I don't want Chelonia to try to decrypt messages", not Grunt.

Gruntfile.js
```diff
@@ -224,6 +224,7 @@ module.exports = (grunt) => {
         'process.env.CONTRACTS_VERSION': `'${CONTRACTS_VERSION}'`,
         'process.env.GI_VERSION': `'${GI_VERSION}'`,
         'process.env.GI_GIT_VERSION': `'${GI_GIT_VERSION}'`,
+        'process.env.CHEL_AVOID_DECRYPTION_ATTEMPTS': 'undefined',
```
taoeffect: Two questions:

1. Does it make sense to include this here at all if it's just being set to `undefined`?
2. Did you verify that `process.env.BUILD` is still needed? (It's still there on line 222)

taoeffect (changes requested):
Nice work @corrideat! Review ready!

Gruntfile.js
corrideat: > Does it make sense to include this here at all if it's just being set to undefined?

Yes, because `process` itself is undefined in the browser (see the first failed build)

> Did you verify that process.env.BUILD is still needed? (It's still there on line 222)

It's still used in `internals.js`:

```js
        if (config.skipActionProcessing) {
          if (process.env.BUILD === 'web') {
            console.log('OP_ACTION_ENCRYPTED: skipped action processing')
          }
          return
        }
```

The comment also mentions that it's needed for `vuelidate`

Gruntfile.js
corrideat: I looked into this, using `config`. But since encryptedData is outside of Chelonia, doing it this way (with an env variable) seemed reasonable, which is probably why it was done like so previously.

> It's really the server that's saying, "I don't want Chelonia to try to decrypt messages", not Grunt.

In a way, but then this is Grunt that's executing the dev server, and in this case `CHEL_AVOID_DECRYPTION_ATTEMPTS` should be set to true. With PR #2759 there could be more flexibility to set or not this.

Gruntfile.js
taoeffect: > It's still used in `internals.js`:

I see... but it seems like it's being used for the same purpose as the other location that was replaced with `CHEL_AVOID_DECRYPTION_ATTEMPTS` in this PR.

So I think that usage should also be updated to use `CHEL_AVOID_DECRYPTION_ATTEMPTS` instead? And if so, then there are no more instances of `process.env.BUILD` AFAICT and that envvar can be removed.

Gruntfile.js
taoeffect: > But since encryptedData is outside of Chelonia

What do you mean...?

`'~/shared/domains/chelonia/encryptedData.js'`

corrideat: Thinking about this, I think that using `config` is superior, but we'd need to make more changes than just renaming an env variable. The changes would need to be done _in Chelonia_ every time `unwrapMaybeEncryptedData`. For example, `const wv = this.config.someNameToBeDefined ? unwrapMaybeEncryptedData(v) : undefined`.

corrideat: > What do you mean...?

I mean it's a utility library that's not got (easy) access to the same `this` that Chelonia does. Even though, yes, technically, it shares a directory.

Gruntfile.js
corrideat: > And if so, then there are no more instances of process.env.BUILD AFAICT and that envvar can be removed.

Again, this is incorrect due to:

> (...)  it's needed for vuelidate

----

ETA:

>  like it's being used for the same purpose as the other location

Mostly for the same purpose (I'd say slightly different, but close enough). You're right that it could be replaced too (but not that `BUILD` can be removed).

taoeffect: > Mostly for the same purpose (I'd say slightly different, but close enough). You're right that it could be replaced too (but not that BUILD can be removed).

OK, then we can keep build for vuelidate (I'll take your word for it..), but then if we both agree it's being used for the same purpose here can this instance also be updated to use the new `CHEL_AVOID_DECRYPTION_ATTEMPTS` (or `config.avoidDecryptionAttempts` - whatever the conclusion there is)?

Gruntfile.js
taoeffect: > I mean it's a utility library that's not got (easy) access to the same this that Chelonia does. Even though, yes, technically, it shares a directory.

I know this is an ugly hack, but could you in that case set a global variable? It would avoid the need for uglifying the code with this outer-check. The check would be able to be left in-place as it is currently.

EDIT: to be clear, what I mean is, `'chelonia/configure'` would still be used to set the variable, but inside that, it would then set a global variable.

corrideat: I think that what we're doing currently is an ugly hack that happens to be convenient, but using a global variable for this just seems wrong. I think that this mostly could be handled by Chelonia, even without such a check on every call.

shared/domains/chelonia/SPMessage.js
```diff
@@ -279,7 +279,7 @@ export class SPMessage {
     return new this(messageToParams(head, targetOp[1]))
   }
 
-  static deserialize (value: string, additionalKeys?: Object, state?: Object): this {
+  static deserialize (value: string, additionalKeys?: Object, state?: Object, unwrapMaybeEncryptedDataOverride?: Function = unwrapMaybeEncryptedData): this {
```
taoeffect: Since this parameter `unwrapMaybeEncryptedDataOverride` is used as the default behavior, can you rename it `unwrapMaybeEncryptedDataFn`?

(Suggestion from Claude, I concur)

shared/domains/chelonia/SPMessage.js
```diff
@@ -290,7 +290,7 @@ export class SPMessage {
     if (!state?._vm?.authorizedKeys && head.op === SPMessage.OP_CONTRACT) {
       const value = rawSignedIncomingData(parsedValue)
       const authorizedKeys = Object.fromEntries(value.valueOf()?.keys.map(wk => {
-        const k = unwrapMaybeEncryptedData(wk)
+        const k = unwrapMaybeEncryptedDataOverride(wk)
```
taoeffect: o4-mini (high) points out that that the returned instance will not use the `unwrapMaybeEncryptedDataOverride` function if/when `decryptedValue()` is called on it, but will use `unwrapMaybeEncryptedData`. Is that a problem? If so, can you update the code to save the function, and if not, could you add a comment explaining why it's OK?

shared/domains/chelonia/internals.js
```diff
@@ -741,7 +742,7 @@ export default (sbp('sbp/selectors/register', {
       },
       [SPMessage.OP_CONTRACT] (v: SPOpContract) {
         state._vm.type = v.type
-        const keys = keysToMap(v.keys, height)
+        const keys = keysToMap.call(self, v.keys, height)
```
taoeffect, line 745: o4-mini (high) is unhappy that in some paces in this file `self` and `this`, though the same, are used inconsistently with `.call`. It suggests picking one and sticking with it, e.g. here:

```js
--- a/shared/domains/chelonia/internals.js
+++ b/shared/domains/chelonia/internals.js
@@ -250,7 +250,7 @@ const handleEvent = {
       const stateForValidation = opT === SPMessage.OP_CONTRACT &&
         !state?._vm?.authorizedKeys
         ? {
-            _vm: { authorizedKeys: keysToMap.call(this, ((opV): any).keys, height) }
+            _vm: { authorizedKeys: keysToMap.call(self, ((opV): any).keys, height) }
           }
         : state
```

shared/domains/chelonia/utils.js
```diff
@@ -158,13 +157,14 @@ export const validateKeyPermissions = (msg: SPMessage, config: Object, state: Ob
   return true
 }
 
-export const validateKeyAddPermissions = (contractID: string, signingKey: SPKey, state: Object, v: (SPKey | EncryptedData<SPKey>)[], skipPrivateCheck?: boolean) => {
+export const validateKeyAddPermissions = function (contractID: string, signingKey: SPKey, state: Object, v: (SPKey | EncryptedData<SPKey>)[], skipPrivateCheck?: boolean) {
```
taoeffect, line 160: Here's some critique from o4-mini (high), feel free to decide what to do about it (if anything):

In utils.js you turned:

```js
export const validateKeyAddPermissions = (… ) => {
  const data = unwrapMaybeEncryptedData(wk)
  // …
}
```

into

```js
export const validateKeyAddPermissions = function (… ) {
  const data = this.config.unwrapMaybeEncryptedData(wk)
  // …
}
```

That works only if every call is done via .call(self, …).  It’s brittle—if someone forgets to .call, this is undefined and you’ll throw.

Alternative pattern
Pass unwrap (or even config) explicitly:

```js
// utils.js
export function validateKeyAddPermissions (
  config: Object,
  contractID: string,
  signingKey: SPKey,
  state: Object,
  wrappedKeys: (SPKey|EncryptedData<SPKey>)[],
  skipPrivateCheck?: boolean
) {
  wrappedKeys.forEach(wk => {
    const data = config.unwrapMaybeEncryptedData(wk)
    if (!data) return
    // … same as before
  })
}
```

Then in internals.js simply:

```js
// internals.js
validateKeyAddPermissions(
  config,
  contractID,
  signingKey,
  state,
  v,
)
```

This is clearer, removes any subtle `.call(self)` dance, and makes the dependency on `config.unwrapMaybeEncryptedData` explicit.

taoeffect (changes requested):
Great work @corrideat! Review ready!

corrideat:
I'll be dismissing the comments about using `.call`. To some extent true, but:

1. Adding more parameters is also brittle
2. One could technically forget to use `.call` or `.apply`, but with TypeScript that'll be a type error.

shared/domains/chelonia/SPMessage.js
corrideat: That's not really a problem based on current use. I concur that technically this could be improved upon, but since it doesn't seem to be a real issue, I'd just add a comment explaining this.

shared/domains/chelonia/utils.js
corrideat: Dimissing as per https://github.com/okTurtles/group-income/pull/2833#issuecomment-2941002904

shared/domains/chelonia/SPMessage.js
```diff
@@ -407,6 +407,10 @@ export class SPMessage {
     if (this._decryptedValue) return this._decryptedValue
     try {
       const value = this.message()
+      // TODO: This uses `unwrapMaybeEncryptedData` instead of a configurable
```
taoeffect, line 410: Is this meant to be a `TODO` and not a `NOTE`?

corrideat: It's a TODO as in something that can be dealt with later if the need arises (the intent was to signal this is like it is out of convenience, not strict API design)

taoeffect (approved):
Fantastic work @corrideat!

PR #2835, opened 2025-06-04
Avoid displaying contract IDs for new members on native notifications

### LINKED ISSUE #2834: contractID displayed instead of username / displayName for MEMBER_ADDED notification
### Problem

Getting this on Android push notifications:

Image

### Solution

Something is wrong with this code:

```js
  MEMBER_ADDED (data: { groupID: string, memberID: string }) {
    return {
      avatarUserID: data.memberID,
      title: rootState[data.groupID]?.settings?.groupName || L('Member added'),
      body: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
        name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.memberID}`,
        ...LTags('strong')
      }),
      plaintextBody: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
        name: sbp('state/vuex/getters').userDisplayNameFromID(data.memberID)
      }),
      icon: 'user-plus',
      level: 'info',
      linkTo: `/group-chat/${rootState[data.groupID]?.generalChatRoomId}`,
      scope: 'group',
      groupID: data.groupID
    }
  },
```

Possibly similar to issue #2774 that was closd.

EDIT: If it's not possible to solve this in the same way #2774 was addressed, then as an alternative solution the second sentence beginning with "Say hi" can be removed if the displayName is unavailable.

> [!IMPORTANT]
> 
> If going with that last approach, it might be safe to translate the two sentences separately and then concatenate them together, but double-check this, as I'm not sure if in all languages such concatenations would make sense even at the sentence-level.

corrideat:
Fix #2834


frontend/model/notifications/templates.js
```diff
@@ -140,16 +140,22 @@ export default ({
...
       body: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
         name: `${CHATROOM_MEMBER_MENTION_SPECIAL_CHAR}${data.memberID}`,
         ...LTags('strong')
       }),
-      plaintextBody: L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
-        name: sbp('state/vuex/getters').userDisplayNameFromID(data.memberID)
-      }),
+      plaintextBody: hasPlaintextName
+        ? L('The group has a new member. Say hi to {strong_}{name}{_strong}!', {
+          name: sbp('state/vuex/getters').userDisplayNameFromID(data.memberID)
```
taoeffect: Since you've already called this and stored the output as `plaintextName`, could you please use that variable here instead of calling it again?

taoeffect (changes requested):
Nice! I like the string you chose too 👍 

One minor DRY fix request:

frontend/model/notifications/templates.js
taoeffect: Also, question: does the same thing need to be done for `MEMBER_LEFT` and `MEMBER_REMOVED` which also use this pattern?

corrideat: >  could you please use that variable here instead of calling it again?

That was the intention.

> does the same thing need to be done for MEMBER_LEFT and MEMBER_REMOVED which also use this pattern?

Those are more likely to have the name, but the same can be done.

taoeffect (approved):
Nicely done!

PR #2837, opened 2025-06-05
Libcheloniajs backport

corrideat:
Minor changes introduced in libcheloniajs, backported. Found by performing a line-by-line diff.

Some (most) changes are related to TypeScript (in order to correct or facilitate correct types). However, a few are small bugfixes.

### DIFF
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -701,7 +701,7 @@ export default (sbp('sbp/selectors/register', {
   // although it can be used as a general-purpose API to process events received
   // from other external sources that are not managed by Chelonia itself (i.e. sources
   // other than the Chelonia-managed websocket connection and RESTful API).
-  'chelonia/handleEvent': async (event: string) => {
+  'chelonia/handleEvent': async function (event: string) {
     const { contractID } = SPMessage.deserializeHEAD(event)
     return await sbp('chelonia/private/in/enqueueHandleEvent', contractID, event)
   },
@@ -784,7 +784,7 @@ export default (sbp('sbp/selectors/register', {
             try {
               await contract.sbp(...sideEffect)
             } catch (e_) {
-              const e = e_
+              const e = (e_: Object)
               console.error(`[chelonia] ERROR: '${e.name}' ${e.message}, for pushed sideEffect of ${mutation.description}:`, sideEffect)
               this.sideEffectStacks[mutation.contractID] = [] // clear the side effects
               throw e
--- a/shared/domains/chelonia/chelonia.js
+++ b/shared/domains/chelonia/chelonia.js
@@ -612,7 +612,7 @@ export default (sbp('sbp/selectors/register', {
         ...options.handlers,
         // Every time we get a REQUEST_TYPE.SUB response, which happens for
         // 'new' subscriptions as well as every time the connection is reset
-        'subscription-succeeded': (event) => {
+        'subscription-succeeded': function (event) {
           const { channelID } = event.detail
           // The check below is needed because we could have unsubscribed since
           // requesting a subscription from the server. In that case, we don't
@@ -626,7 +626,7 @@ export default (sbp('sbp/selectors/register', {
               console.warn(`[chelonia] Syncing contract ${channelID} failed: ${err.message}`)
             })
           }
-          options.handlers?.['subscription-succeeded']?.(event)
+          options.handlers?.['subscription-succeeded']?.call(this, event)
         }
       },
       // Map message handlers to transparently handle encryption and signatures
@@ -645,7 +645,7 @@ export default (sbp('sbp/selectors/register', {
                     return
                   }
                   sbp('chelonia/queueInvocation', msg.channelID, () => {
-                    (v: Function).call(this.pubsub, parseEncryptedOrUnencryptedMessage.call(this, {
+                    (v: Function).call(this.pubsub, parseEncryptedOrUnencryptedMessage(this, {
                       contractID: msg.channelID,
                       serializedData: msg.data
                     }))
@@ -664,7 +664,7 @@ export default (sbp('sbp/selectors/register', {
                     return
                   }
                   sbp('chelonia/queueInvocation', msg.channelID, () => {
-                    (v: Function)([msg.key, parseEncryptedOrUnencryptedMessage.call(this, {
+                    (v: Function).call(this.pubsub, [msg.key, parseEncryptedOrUnencryptedMessage(this, {
                       contractID: msg.channelID,
                       meta: msg.key,
                       serializedData: JSON.parse(Buffer.from(msg.data).toString())
@@ -674,7 +674,7 @@ export default (sbp('sbp/selectors/register', {
                   })
                 }]
               case NOTIFICATION_TYPE.DELETION:
-                return [k, (msg) => (v: Function)(msg.data)]
+                return [k, (msg) => (v: Function).call(this.pubsub, msg.data)]
               default:
                 return [k, v]
             }
@@ -701,7 +701,7 @@ export default (sbp('sbp/selectors/register', {
   // although it can be used as a general-purpose API to process events received
   // from other external sources that are not managed by Chelonia itself (i.e. sources
   // other than the Chelonia-managed websocket connection and RESTful API).
-  'chelonia/handleEvent': async function (event: string) {
+  'chelonia/handleEvent': async (event: string) => {
     const { contractID } = SPMessage.deserializeHEAD(event)
     return await sbp('chelonia/private/in/enqueueHandleEvent', contractID, event)
   },
@@ -783,7 +783,8 @@ export default (sbp('sbp/selectors/register', {
             const sideEffect = sideEffects.shift()
             try {
               await contract.sbp(...sideEffect)
-            } catch (e) {
+            } catch (e_) {
+              const e = e_
               console.error(`[chelonia] ERROR: '${e.name}' ${e.message}, for pushed sideEffect of ${mutation.description}:`, sideEffect)
               this.sideEffectStacks[mutation.contractID] = [] // clear the side effects
               throw e
@@ -1707,7 +1708,7 @@ export default (sbp('sbp/selectors/register', {
         // race conditions and data not being properly initialised.
         // See <https://github.com/okTurtles/group-income/issues/2780>
         currentValue = serializedDataText
-          ? parseEncryptedOrUnencryptedMessage.call(this, {
+          ? parseEncryptedOrUnencryptedMessage(this, {
             contractID,
             serializedData: JSON.parse(serializedDataText),
             meta: key
@@ -1821,7 +1822,7 @@ export default (sbp('sbp/selectors/register', {
       throw new Error('Invalid response status: ' + response.status)
     }
     const data = await response.json()
-    return parseEncryptedOrUnencryptedMessage.call(this, {
+    return parseEncryptedOrUnencryptedMessage(this, {
       contractID,
       serializedData: data,
       meta: key
@@ -1839,7 +1840,7 @@ export default (sbp('sbp/selectors/register', {
     this.pubsub.setKvFilter(contractID, filter)
   },
   'chelonia/parseEncryptedOrUnencryptedDetachedMessage': function ({ contractID, serializedData, meta }: { contractID: string, serializedData: Object, meta?: ?string }) {
-    return parseEncryptedOrUnencryptedMessage.call(this, {
+    return parseEncryptedOrUnencryptedMessage(this, {
       contractID,
       serializedData,
       meta
@@ -1873,7 +1874,7 @@ function outputEncryptedOrUnencryptedMessage ({
   const signedMessage = innerSigningKeyId
     ? (state._vm.authorizedKeys[innerSigningKeyId] && state._vm.authorizedKeys[innerSigningKeyId]?._notAfterHeight == null)
         ? signedOutgoingData(contractID, innerSigningKeyId, (data: any), this.transientSecretKeys)
-        : signedOutgoingDataWithRawKey(this.transientSecretKeys[innerSigningKeyId], (data: any), this.transientSecretKeys)
+        : signedOutgoingDataWithRawKey(this.transientSecretKeys[innerSigningKeyId], (data: any))
     : data
   const payload = !encryptionKeyId
     ? signedMessage
@@ -1885,7 +1886,7 @@ function outputEncryptedOrUnencryptedMessage ({
   return serializedData
 }
 
-function parseEncryptedOrUnencryptedMessage ({
+function parseEncryptedOrUnencryptedMessage (ctx, {
   contractID,
   serializedData,
   meta
@@ -1897,9 +1898,9 @@ function parseEncryptedOrUnencryptedMessage ({
   if (!serializedData) {
     throw new TypeError('[chelonia] parseEncryptedOrUnencryptedMessage: serializedData is required')
   }
-  const state = sbp(this.config.stateSelector)[contractID]
+  const state = sbp(ctx.config.stateSelector)[contractID]
   const numericHeight = parseInt(serializedData.height)
-  const rootState = sbp(this.config.stateSelector)
+  const rootState = sbp(ctx.config.stateSelector)
   const currentHeight = rootState.contracts[contractID].height
   if (!(numericHeight >= 0) || !(numericHeight <= currentHeight)) {
     throw new Error(`[chelonia] parseEncryptedOrUnencryptedMessage: Invalid height ${serializedData.height}; it must be between 0 and ${currentHeight}`)
@@ -1909,7 +1910,7 @@ function parseEncryptedOrUnencryptedMessage ({
   const aad = (meta ?? '') + serializedData.height
 
   const v = signedIncomingData(contractID, state, serializedData, numericHeight, aad, (message) => {
-    return maybeEncryptedIncomingData(contractID, state, message, numericHeight, this.transientSecretKeys, aad, undefined)
+    return maybeEncryptedIncomingData(contractID, state, message, numericHeight, ctx.transientSecretKeys, aad, undefined)
   })
 
   // Cached values
@@ -2016,7 +2017,7 @@ async function outEncryptedOrUnencryptedAction (
   const signedMessage = params.innerSigningKeyId
     ? (state._vm.authorizedKeys[params.innerSigningKeyId] && state._vm.authorizedKeys[params.innerSigningKeyId]?._notAfterHeight == null)
         ? signedOutgoingData(contractID, params.innerSigningKeyId, (unencMessage: any), this.transientSecretKeys)
-        : signedOutgoingDataWithRawKey(this.transientSecretKeys[params.innerSigningKeyId], (unencMessage: any), this.transientSecretKeys)
+        : signedOutgoingDataWithRawKey(this.transientSecretKeys[params.innerSigningKeyId], (unencMessage: any))
     : unencMessage
   if (opType === SPMessage.OP_ACTION_ENCRYPTED && !params.encryptionKeyId) {
     throw new Error('OP_ACTION_ENCRYPTED requires an encryption key ID be given')
@@ -2064,7 +2065,7 @@ async function outEncryptedOrUnencryptedAction (
 // ```
 function gettersProxy (state: Object, getters: Object) {
   const proxyGetters = new Proxy({}, {
-    get (target, prop) {
+    get (_target, prop) {
       return getters[prop](state, proxyGetters)
     }
   })
--- a/shared/domains/chelonia/internals.js
+++ b/shared/domains/chelonia/internals.js
@@ -53,11 +53,11 @@ const getMsgMeta = function (message: SPMessage, contractID: string, state: Obje
   return result
 }
 
-const keysToMap = function (keys: (SPKey | EncryptedData<SPKey>)[], height: number, authorizedKeys?: Object): Object {
+const keysToMap = function (keys_: (SPKey | EncryptedData<SPKey>)[], height: number, authorizedKeys?: Object): Object {
   // Using cloneDeep to ensure that the returned object is serializable
   // Keys in a SPMessage may not be serializable (i.e., supported by the
   // structured clone algorithm) when they contain encryptedIncomingData
-  keys = keys.map((key) => {
+  const keys = keys_.map((key) => {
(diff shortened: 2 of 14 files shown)


shared/domains/chelonia/chelonia.js
```diff
@@ -626,7 +626,7 @@ export default (sbp('sbp/selectors/register', {
               console.warn(`[chelonia] Syncing contract ${channelID} failed: ${err.message}`)
             })
           }
-          options.handlers?.['subscription-succeeded']?.(event)
+          options.handlers?.['subscription-succeeded']?.call(this, event)
```
taoeffect, line 629: Exposing Chelonia's internals to the handlers? Why? Is it completely safe to be doing this? 🤔

shared/domains/chelonia/chelonia.js
```diff
@@ -1873,7 +1874,7 @@ function outputEncryptedOrUnencryptedMessage ({
   const signedMessage = innerSigningKeyId
     ? (state._vm.authorizedKeys[innerSigningKeyId] && state._vm.authorizedKeys[innerSigningKeyId]?._notAfterHeight == null)
         ? signedOutgoingData(contractID, innerSigningKeyId, (data: any), this.transientSecretKeys)
-        : signedOutgoingDataWithRawKey(this.transientSecretKeys[innerSigningKeyId], (data: any), this.transientSecretKeys)
+        : signedOutgoingDataWithRawKey(this.transientSecretKeys[innerSigningKeyId], (data: any))
```
taoeffect, line 1877: I'm trying to figure out what the 3rd parameter, `height` was used for, and got stuck here:

```js
      return (additionalData: ?string) => JSON.stringify(this.serialize(additionalData, height))
```

What is `this.serialize`? Is it:

```js
  const serializefn = (additionalData: ?string) => boundStringValueFn(additionalData || '')
```

But that only takes 1 parameter, so is the 3rd parameter `height` completely unused? Why's it there then?

shared/domains/chelonia/internals.js
```diff
@@ -655,7 +656,7 @@ export default (sbp('sbp/selectors/register', {
           if (!targetState._volatile.watch.some((tWatch) => {
             return (tWatch[0] === pWatch[0]) && (tWatch[1] === pWatch[1])
           })) {
-            targetState._volatile.push(pWatch)
+            targetState._volatile.watch.push(pWatch)
```
taoeffect, line 659: Was this the bugfix? Looks like a bugfix..

shared/domains/chelonia/chelonia.js
corrideat: This is exposing the pubsub AFICT, not Chelonia internals.

shared/domains/chelonia/internals.js
corrideat: There were a few. This is definitely one of them

shared/domains/chelonia/internals.js
```diff
@@ -1167,7 +1176,7 @@ export default (sbp('sbp/selectors/register', {
               }).catch((e) => {
                 // Using console.error instead of logEvtError because this
                 // is a side-effect and not relevant for outgoing messages
-                console.error('Error stopping watching events after removing key', { contractID, foreignContract, foreignKeyName, fkUrl })
+                console.error('Error stopping watching events after removing key', { contractID, foreignContract, foreignKeyName, fkUrl }, e)
```
taoeffect, line 1179: Does this code get run on the server? If so, the `e` should be the first parameter. If not, it can be left as-is.

shared/domains/chelonia/internals.js
```diff
@@ -1298,7 +1307,7 @@ export default (sbp('sbp/selectors/register', {
       config[`postOp_${opT}`]?.(message, state) // hack to fix syntax highlighting `
     }
   },
-  'chelonia/private/in/enqueueHandleEvent': function (contractID: string, event: string) {
+  'chelonia/private/in/enqueueHandleEvent': (contractID: string, event: string) => {
```
taoeffect: Why is this change being made? These are supposed to be un-bound functions so that they have access to the chelonia namespace via `this`

So either this should be reverted, or as a special case, commented clearly

shared/domains/chelonia/internals.js
```diff
@@ -1574,7 +1583,9 @@ export default (sbp('sbp/selectors/register', {
       const fKeyId = findKeyIdByName(foreignState, foreignKeyName)
       if (!fKeyId) {
         // Key was deleted; mark it for deletion
-        pendingKeyRevocations.find(([id]) => id === keyId)[1] = 'del'
+        if (pendingKeyRevocations[keyId] === true) {
+          this.config.reactiveSet(pendingKeyRevocations, keyId, 'del')
+        }
```
taoeffect, line 1588: This re-introduces `reactiveSet`, which I thought we were getting rid of?

shared/domains/chelonia/internals.js
corrideat: This function doesn't use `this`. This change was only made for functions that don't access it (TypeScript ensures things are bound correctly via `this` type checking)

shared/domains/chelonia/internals.js
corrideat: This may or may not be run on the server

shared/domains/chelonia/internals.js
```diff
@@ -1682,7 +1693,7 @@ export default (sbp('sbp/selectors/register', {
     const [keyShareEncryption, height, , [originatingContractID, rv, originatingContractHeight, headJSON]] = ((entry: any): [boolean, number, string, [string, Object, number, string]])
     entry.pop()
 
-    const krsEncryption = !!state._vm?.invites?.[signingKeyId]?.private
+    const krsEncryption = !!contractState._vm.authorizedKeys?.[signingKeyId]?._private
```
taoeffect, line 1696: Another bugfix it seems, why didn't we notice this?

corrideat: Indeed. Not sure why why didn't catch this.

shared/domains/chelonia/internals.js
```diff
@@ -1961,13 +1973,15 @@ export default (sbp('sbp/selectors/register', {
       if (!processingErrored) {
         // Gets run get when skipSideEffects is false
         if (Array.isArray(internalSideEffectStack) && internalSideEffectStack.length > 0) {
-          await Promise.all(internalSideEffectStack.map(fn => Promise.resolve(fn({ state: contractStateCopy, message })).catch((e) => {
+          await Promise.all(internalSideEffectStack.map(fn => Promise.resolve(fn({ state: contractStateCopy, message })).catch((e_) => {
+            const e = e_
```
taoeffect: I don't understand why these changes were made.... could you explain?

shared/domains/chelonia/internals.js
```diff
@@ -2008,7 +2024,7 @@ export default (sbp('sbp/selectors/register', {
 }): string[])
 
 const eventsToReingest = []
-const reprocessDebounced = debounce((contractID) => sbp('chelonia/private/out/sync', contractID, { force: true }).catch((e) => {
+const reprocessDebounced = debounce((contractID) => sbp('chelonia/private/out/sync', contractID, { force: true }).catch(() => {
```
taoeffect: Why isn't this `e` (now gone) being logged?

shared/domains/chelonia/internals.js
corrideat: I'm on mobile right now, but this runs as a side effect and accesses another contract IIRC. That should use reactiveSet

shared/domains/chelonia/utils.js
```diff
@@ -36,7 +36,7 @@ export const findSuitableSecretKeyId = (state: Object, permissions: '*' | string
       purposes.reduce((acc, purpose) => acc && k.purpose.includes(purpose), true) &&
       (Array.isArray(allowedActions)
         ? allowedActions.reduce((acc, action) =>
-          acc && (k.allowedActions === '*' || k.allowedActions?.includes(action)), true
+          acc && (k.allowedActions === '*' || !!k.allowedActions?.includes(action)), true
```
taoeffect, line 39: This seems unnecessary, worked fine before afaict...

shared/domains/chelonia/internals.js
corrideat: Not sure, it wasn't logged originally. Can take a closer look once on desktop again.

shared/domains/chelonia/utils.js
```diff
@@ -48,7 +48,7 @@ export const findContractIDByForeignKeyId = (state: Object, keyId: string): ?str
   if (!keyId || !state?._vm?.authorizedKeys?.[keyId]?.foreignKey) return
 
   try {
-    const fkUrl = new URL(state._vm?.authorizedKeys?.[keyId]?.foreignKey)
+    const fkUrl = new URL(state._vm.authorizedKeys[keyId].foreignKey)
```
taoeffect: Now we're confident these will be here? Why?

shared/domains/chelonia/internals.js
corrideat: You need to look at libcheloniajs to make sense of this. The short explanation is that it's for type casting. E is of type unknown (mixed in flow) but used as Error, which is wrong. Doing this assignment avoids the need of having to force a type cast every time

shared/pubsub.js
```diff
@@ -340,7 +340,7 @@ const defaultClientEventHandlers = {
     }
   },
 
-  offline (event: Event) {
+  offline () {
```
taoeffect, line 343: Not a big fan of these changes... before we knew what parameter could be used thanks to it being listed, but now that info is gone. So why were these changes made if no linter complained before?

corrideat: The type system is supposed to handle this. Not making these changes requires disabling eslint rules

corrideat: If you look at libchelonia: the linter does complain and the type info is still there from the type definitions, and using the wrong type will result in an error

shared/domains/chelonia/internals.js
taoeffect: Then we should probably move the `e` to the front

shared/domains/chelonia/utils.js
corrideat: Because the type is defined so, but also because of the check in the line above

shared/domains/chelonia/internals.js
taoeffect: I really don't like making this change here if we don't need it. In `libchelonia` you can add the type info to the parameters, but here we shouldn't be making this `=>`, can you revert?

shared/domains/chelonia/utils.js
corrideat: This forces the type to be a Boolean, while it could be a Boolean or nullish previously

shared/pubsub.js
taoeffect: The type system doesn't tell us — via the code — what the parameters could be... And it shouldn't interfere with our ability to do that

shared/domains/chelonia/utils.js
taoeffect: Could you then cache the value instead of making this very long string of accesses again?

shared/pubsub.js
corrideat: The type system does tell you what the parameters are, and it doesn't interfere with anything, StandardJS does (or the linter rather)

taoeffect (changes requested):
Review done! Great work @corrideat! I'm surprised that some of the bugfixes were for bugs that we somehow didn't notice...

Left a few change requests

shared/pubsub.js
taoeffect: > StandardJS does (or the linter rather)

But it's never complained before... so why was this change made?

corrideat:
I seem unable to find one of your comments, but it was about reverting a change here while keeping it in libchelonia IIUC (function to arrow function)

I don't think we should do that. We should either revert everywhere or not at all since the point of this PR is eliminating or minimising code differences between the two

taoeffect:
Are you referring to this?

https://github.com/okTurtles/group-income/pull/2837#discussion_r2129593067

If so, I agree that should be reverted in libchelonia too, as all chelonia namespace selectors should be bound to the same context

shared/domains/chelonia/internals.js
corrideat: To clarify, we're not getting rid of `reactiveSet`, only getting rid of it when unnecessary (such as when processing messages and working on a temporary state copy). I suspect that that's the case here, but will double check later

shared/domains/chelonia/chelonia.js
corrideat: There never was a third parameter AFAICT (or, if there was, it has long since been removed). However, `signedOutgoingData` matches the signature for this call, so it likely was a mistake.

As for why there was a `height` to `signedOutgoingDataWithRawKey`, I'm not sure. It was likely left over from a time when I was deciding on the most ergonomic API design. The `height` is, indirectly, used in `serialize` (and by extension `serializefn`) as additional data (well, the entire head is, see `messageToParams`). The `height` is used, directly, for incoming data, for the purpose of determining whether the key is valid.

shared/domains/chelonia/internals.js
corrideat: That's not what we do anywhere else in this file and IMHO we should find some other way of logging instead of forcing some weird pattern on generic code. FWIW, this code is unlikely to run on the server since `skipSideEffects: true` is part of the server presets (which a server administrator may override, presumably, or we could change in the future to something else)

shared/domains/chelonia/internals.js
corrideat: Yes, this was the comment I was referring to earlier. I think (and it seems we concur) that it should be either left as is (i.e., change kept in both places) or reverted in both places, for the sake of keeping code modifications at 0 or close to 0.

Now, regarding this specific change: it was mostly cosmetic, since arrow functions are clearer to reason about and these cases do _not_ use `this`. Using `this` accidentally will result in TypeScript complaining (`Object is possibly 'undefined'. ts(2532)`); OTOH, using `function` and accidentally using `this` (without a `this: type` annotation) will result in `Property 'xxx' does not exist on type '{}'. ts(2339)`), so I guess they're similar in terms of safety.

Based on benchmarking [1], arrow functions seem to be slightly faster, but perhaps not significantly (or even consistently) so.

I'll be happy to revert this.

[1] https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiIxa0plTE00aWZFdVZYLUZOOUZQMTUiLCJjb2RlIjoiY29uc3QgeyBiIH0gPSBEQVRBO1xuYi5jYWxsKE9iamVjdC5jcmVhdGUobnVsbCkpOyIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJBcnJvdyJ9LHsiaWQiOiJsM0NUenNRZjlyM1hZUGRlbUhocVMiLCJjb2RlIjoiY29uc3QgeyBhIH0gPSBEQVRBO1xuYS5jYWxsKE9iamVjdC5jcmVhdGUobnVsbCkpOyIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJSZWd1bGFyIn1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOltdfSwiZGF0YUNvZGUiOiJyZXR1cm4ge1xuICBhOiBmdW5jdGlvbiAoKSB7IGZvciAobGV0IGkgPSAwOyBpIDwgMTAwOyBpKyspIE1hdGgucmFuZG9tKCkqcGFyc2VJbnQoaSk7IH0sXG4gIGI6ICgpID0-IHsgZm9yIChsZXQgaSA9IDA7IGkgPCAxMDA7IGkrKykgTWF0aC5yYW5kb20oKSpwYXJzZUludChpKTsgfSxcbn07In19

shared/domains/chelonia/internals.js
corrideat: Double-checked as promised. This change is correct because it's being invoked like this:

```
sbp('chelonia/private/queueEvent', externalContractID, ['chelonia/private/deleteOrRotateRevokedKeys', externalContractID]).catch((e: unknown) => {
        console.error(`Error at deleteOrRotateRevokedKeys for contractID ${contractID} and externalContractID ${externalContractID}`, e)
      })
```

Since it's an async event and working on the 'real' state, `reactiveSet` is appropriate.

shared/domains/chelonia/internals.js
corrideat: To give a self-contained answer including an example:

See:

```ts
      try {
        await handleEvent.processMutation.call(this, message, contractStateCopy, internalSideEffectStack)
      } catch (e_) {
        const e = e_ as Error
        if (e?.name === 'ChelErrorDecryptionKeyNotFound') {
```

While in Flow, `e_` is typed as `any`, in TypeScript `e_` is (correctly) typed as `unknown` (which corresponds to `mixed` in Flow).

Therefore (and correctly, based on JS semantics), `e.name` isn't allowed because `e` could be a `number`, an object of a different shape, etc. However, we normally assume that `e_` is an `Error`. Fixing this on TS-land required me to make one of these changes:

1. Check everywhere whether `e_` is an Error. This would work and is the most correct solution, but it uglifies the code and adds a lot of repetition, even if I create an utility function for this task.
2. Cast `e_` to `Error` every time it's used (e.g., `(e_ as Error)?.name === 'ChelErrorDecryptionKeyNotFound'`). This requires no source modifications but it also adds lots of unnecessary repetition.
3. This solution, which does the cast above once.

shared/domains/chelonia/internals.js
corrideat: Took a look. I'd be surprised if `e` isn't logged somewhere already. However, yes, I think `e` should have been logged here and that it was a bug in the original code.

shared/domains/chelonia/utils.js
corrideat: Reverting this change results in:

```
No overload matches this call.
  Overload 1 of 3, '(callbackfn: (previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string, initialValue: string): string', gave the following error.
    Type 'boolean | undefined' is not assignable to type 'string'.
```

It can be avoided by explicitly typing `(acc: boolean | undefined, ...)`, but I think this change makes it clearer and explicit that a boolean is needed (rather than having to keep the context in mind that `undefined` is similar to `false`)

shared/domains/chelonia/utils.js
corrideat: Do you have any suggestions? It doesn't seem possible in this particular case.

1. First, we check for `state?._vm?.authorizedKeys?.[keyId]?.foreignKey` (`if` in line 48)
2. Then, we access it here.

I guess this would work, but it's ugly:

```js
  let fk
  if (!keyId || !(fk = state?._vm?.authorizedKeys?.[keyId]?.foreignKey)) return

  try {
    const fkUrl = new URL(fk)
    return fkUrl.pathname
  } catch {}
```

shared/pubsub.js
corrideat: I've checked and the rules seem to come from `@typescript-eslint/recommended` instead of `standard`, which explains why it's not complained before. The specific rule is `@typescript-eslint/no-unused-vars`.

Personally, I agree with the rule here, as the parameter here serves no real purpose. Example: we call `ary.map((el) => { ... })` throughout instead of `ary.map((el, i, a) => { ... })` unless we specifically need _a_.

You still know what parameters are allowed thanks to the type information (if you haven't worked with TypeScript projects before, I encourage you to open `libcheloniajs` in VSCode, Zed might work too, and try hovering over `offline`).

Even if you're using something like Notepad, the parameters allowed are still known in three ways: by manually looking at the type information, by manually looking at the call site and by attempting to build the project and getting a type error. The shape of the parameters to these functions may be overridden by Chelonia as well, so it's even more unlikely that these parameters will be helpful to anyone other than those working on `pubsub` / `chelonia`.

And, last point is that, since the parameter isn't being used, having it could provide the wrong information (e.g., see the other change related to `signedOutgoingDataWithRawKey`)

shared/domains/chelonia/internals.js
corrideat: Maybe I could to the type casting on GI to, so that it doesn't look like a nonsensical change?

shared/domains/chelonia/utils.js
taoeffect: Yeah, your suggestion, or, if it works, something like:

```js
const fk = state?._vm?.authorizedKeys?.[keyId]?.foreignKey
if (!keyId || !fk) return

...
```

shared/domains/chelonia/chelonia.js
taoeffect: > As for why there was a height to signedOutgoingDataWithRawKey, I'm not sure. It was likely left over from a time when I was deciding on the most ergonomic API design.

Was? It still is there, but it's been renamed to `_height?: number`

It should be removed if it's not used, right? 

> The height is, indirectly, used in serialize (and by extension serializefn) as additional data (well, the entire head is, see messageToParams). The height is used, directly, for incoming data, for the purpose of determining whether the key is valid.

Don't know what you mean... I don't see it being used at all in `signedOutgoingDataWithRawKey`

corrideat: > Was? It still is there, but it's been renamed to _height?: number
>
> It should be removed if it's not used, right?

I meant in the returned object. Yes, probably it should be removed (`_height`).

> Don't know what you mean... I don't see it being used at all in signedOutgoingDataWithRawKey

Like I said, it's being used _indirectly_ as `additionalData`. This is done in `messageToParams`. `height`/`_height` are not used at all here.

shared/domains/chelonia/utils.js
corrideat: Your suggestion won't work because `keyId` needs to be checked first.

```js
if (!keyId) return
const fk = state?._vm?.authorizedKeys?.[keyId]?.foreignKey
if (!fk) return
```

This works though.

shared/domains/chelonia/internals.js
taoeffect: > Maybe I could to the type casting on GI to, so that it doesn't look like a nonsensical change?

Yeah that sounds like a good idea 👍

taoeffect (approved):
Awesome work @corrideat ! 👏

PR #2838, opened 2025-06-08
Remove uses of global Buffer

corrideat:
Using global `Buffer` isn't appropriate because it's not available outside of Node and setting `globalThis.Buffer` pollutes the global scope unnecessarily.


taoeffect (approved):
Great job!

PR #2839, opened 2025-06-08
#2823 - Fix the overflowing proposal issue

### LINKED ISSUE #2823: Overflown text in proposals with long lines
### Problem

Creating a proposal with long text without spaces breaks the UI. See below.

Long proposal

Besides the text overflowing, the 'Read more' button does not seem to do anything visible.

Looking at `ProposalItem.vue`, I noticed a few issues:

1. There's some custom logic for attempting to add ellipsis. Not all languages use spaces, and not all languages use `...` as ellipsis (also, U+2026, `…` would be more typographically correct). Because of all of this, IMHO such custom logic just shouldn't be there (and instead, rely on the browser to do all of this).
2. The current logic also doesn't work when writing long names and descriptions, as shown in the screenshot.

Edited:

Note an additional issue apart from the overflow text: the vote buttons don't appear or appear broken. On desktop, I had to zoom out considerably to see the vote buttons:

Vote buttons on desktop mode zooming out

Zooming in I could also trigger the mobile view with visible buttons, but note that the 'no' button isn't visible and the 'yes' button is too wide.

Vote buttons on mobile mode zooming in

Edited 2:

I noticed (just as an example) that arbitrary text can be put by using non-breaking spaces.

Image

And in this case, the breakage is even worse because you can't even see whether your vote is aye or nay.

Image

### Solution

- To address this, I'd be better to avoid rewriting the text on the client, and instead rely on  the browser's text handling capabilities, such as using https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow and similar CSS to achieve the desired goal of hiding and showing some text. In addition, the text show somehow still show even if it doesn't fit, but it shouldn't overflow, so the 'read more' button should also apply CSS to force breaks.

SebinSong:
closes #2823 

As you can see from below, once the issue is fixed the 'Read more' button seems to work as expected too.

https://github.com/user-attachments/assets/a262ea39-ce38-4c07-9918-0343a0165638

### DIFF
--- a/frontend/views/containers/proposals/ProposalItem.vue
+++ b/frontend/views/containers/proposals/ProposalItem.vue
@@ -14,7 +14,7 @@ li.c-item-wrapper(data-test='proposalItem' :data-proposal-hash='proposalHash')
         )
 
       .c-main-content
-        p.has-text-bold(data-test='typeDescription')
+        p.c-proposal-title.has-text-bold(data-test='typeDescription')
           | {{typeDescription}}
           tooltip.c-tip(
             v-if='isToRemoveMe && proposal.status === statuses.STATUS_OPEN'
@@ -54,6 +54,7 @@ li.c-item-wrapper(data-test='proposalItem' :data-proposal-hash='proposalHash')
           i18n.has-text-danger(
             v-if='isExpiredInvitationLink'
           ) Expired
+
       proposal-vote-options(
         v-if='proposal.status === statuses.STATUS_OPEN'
         :proposalHash='proposalHash'
@@ -450,5 +451,10 @@ export default ({
   &-text {
     display: inline;
   }
+}
+
+.c-proposal-title,
+.c-reason {
+  word-break: break-word;
 }
 </style>


frontend/views/containers/proposals/ProposalItem.vue
```diff
@@ -451,4 +452,9 @@ export default ({
     display: inline;
   }
 }
+
+.c-proposal-title,
+.c-reason {
+  word-break: break-word;
```
corrideat, line 458: I'd also add `hyphens: auto` to allow for more natural breaks (especially if there are no breaking spaces, the code as is will break words in the middle, like `example` could be broken up as `ex` (line 1) `ample` (line 2) instead of `ex-` `ample`)

corrideat (approved):
I've tested this and it works well. See my comment for a small change request.

There's one part of the original issue that's _not_ being addressed here (how text is shortened after / before the `Read more` button is used), but, after discussing this with @taoeffect, we agreed that they're related by separate issues.

I'm approving this PR as is, although adding `hyphens: auto` would IMHO make it better.

frontend/views/containers/proposals/ProposalItem.vue
SebinSong: `hyphens: auto` apparently does not work when `word-break: break-word;` is set on the same element. In this case `word-break: break-word;` is necessary here to fix the issue so `hyphens: auto` is not applicable for this one.

SebinSong: I see `text-overflow: ellipsis` is suggested too. Beware that the intended behavior by `text-overflow: ellipsis` does not take effect when this rule is used alone. To make CSS take care of showing ellipsis for long text, these three properties should go together (reference: [MDN doc](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)):

```
.has-ellipsis {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

And due to having `white-space: nowrap;` here, now it will only allow us to have a single line, not multiple lines like how the proposal UI currently handles this:

image

But I think having single line with ellipsis is okay because there is 'show more' button there.

Try workin on #2844 yourself.

corrideat: > hyphens: auto apparently does not work when word-break: break-word;

Interesting. While that seems to be true for Chrome, it works fine on Firefox.

With `hyphen: auto`

image

Without

image

SebinSong: Yeah, If it wasn't this bug, I would absolutely love this rule.

corrideat: > I see text-overflow: ellipsis is suggested too.

That is indeed unfortunate. In that case, the `L` function suggestion for #2844 might be the better way of handling this.

taoeffect (approved):
Thanks @SebinSong, nice job!

PR #2840, opened 2025-06-08
#2827 - Fix overflowing username in chat

### LINKED ISSUE #2827: Overflown text in chat with long usernames/displaynames
### Problem

When a user has a long username or displayname, it can overflow the chat card. This also means the timestamp will be outside the visible window. 

Image

### Solution

1. Ensure the timestamp is visible. 
2. Determine the best way to handle the long username and do that

SebinSong:
closes #2827 

**[Fix screenshot]**

<br><br>

But let me know if we want it to show the whole username in multiple lines instead like below.

another overflow fix

### DIFF
--- a/frontend/views/containers/chatroom/MessageBase.vue
+++ b/frontend/views/containers/chatroom/MessageBase.vue
@@ -24,8 +24,8 @@
       slot(name='header')
         .c-who(v-if='!isEditing' :class='{ "sr-only": isSameSender }')
           profile-card(:contractID='from' direction='top-left')
-            span.is-title-4 {{ who }}
-          span.has-text-1 {{ humanDate(datetime, { hour: 'numeric', minute: 'numeric' }) }}
+            span.c-username.is-title-4 {{ who }}
+          span.c-datetime.has-text-1 {{ humanDate(datetime, { hour: 'numeric', minute: 'numeric' }) }}
 
       slot(name='body')
         render-message-text.c-replying(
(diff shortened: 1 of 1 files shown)


dotmacro:
Great! The fix screenshot (with "...") is best because that matches how we already treat overflowing usernames elsewhere, like the profile card:
image

frontend/views/containers/chatroom/MessageBase.vue
```diff
@@ -24,8 +24,8 @@
       slot(name='header')
         .c-who(v-if='!isEditing' :class='{ "sr-only": isSameSender }')
           profile-card(:contractID='from' direction='top-left')
-            span.is-title-4 {{ who }}
-          span.has-text-1 {{ humanDate(datetime, { hour: 'numeric', minute: 'numeric' }) }}
+            span.c-username.is-title-4 {{ who }}
```
taoeffect: As an accessibility improvement for long usernames, could you add a [`title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/title) attribute here too with the contents being the username?

These allow the full username to be shown on hover, and I think they also help with screen readers.

I think this is how it would look:

```pug
span.c-username.is-title-4(:title='who') {{ who }}
```

taoeffect (changes requested):
Fantastic PR @SebinSong!

One minor request for an accessibility improvement (use of `title` attribute)

frontend/views/containers/chatroom/MessageBase.vue
SebinSong: @taoeffect Updated:

taoeffect (approved):
Great work @SebinSong!

PR #2841, opened 2025-06-08
ChatMain fixes

### LINKED ISSUE #2830: Chatroom message loading bug: ChatMain loads too many messages when scrolled to top
### Problem

When there's a chatroom with many messages, and we've scrolled to the very top, the scroll position is correctly remembered.

However, when switching to another channel and then back to that original chatroom, it loads all of the messages to the very end instead of loading them in scroll increments.

This is especially a problem when linking to a specific message that is very high up, as it delays the loading of that message significantly when the message is in a chatroom with many messages.

### Solution

Whether scrolling up or down, the same number of chatroom messages should be loaded in increments when there are more messages to load.

Currently this works correctly for scrolling up, but it seems it's not properly done for scrolling down.

When you get to the bottom of the current set of messages and there are more messages to load, it should auto load the next set of messages (again, in increments).

### LINKED ISSUE #2905: Chat issue: clicking on the reply text doesn't load message when it's far away
### Problem

When replying to a message in a DM that requires scrolling up very far (forcing the spinner to appear a couple times), then loading the page afresh (to the latest message in a DM), and clicking on the quoted reply text, the page doesn't scroll up to the original message that was replied to.

### Solution

Reply behavior should work regardless of how far back the original message being replied to is.

### LINKED ISSUE #2926: Virtual keyboard occludes input fields
### Problem

This is mostly an issue that we've seen in happening in the chat area, although it could be happening on other places.

Relevant links:

https://github.com/okTurtles/group-income/pull/2841#issuecomment-3408756525
https://github.com/okTurtles/group-income/pull/2841#issuecomment-3398718622
https://github.com/user-attachments/assets/d8423cc2-0bc4-444e-837d-a75a23c11bf2
https://codepen.io/corrideat/pen/zxrdQwQ

This is what we want (note that the text area is visible and above the virtual keyboard):

Image

This is what will sometimes happen (note that the keyboard is hiding the text area):

Image

When this happening, the text area seems to become slightly visible when the user starts typing something, see:

Image

This seems to _mostly_ affect Android, as I haven't been able to reproduce this issue on iOS (however, it's possible that it can happen there too).

When working on trying to fix this (see the codepen above), I immediately stumbled upon issues, since it seems like browser makers have made it deliberately difficult to detect the presence of a virtual keyboard. The user-observable effect of a virtual keyboard is that the viewport shrinks (e.g., it's possible to scroll to the bottom to the page, even with the keyboard open). However, checking for any kind of height whether it be in `screen`, `window`, `documentElement`, `body`, some other element in the DOM, etc. seems to report the same value than it would when a virtual keyboard is _not_ present.

Furthermore, although there's a (at the time of writing, Chrome-only) [Virtual Keyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API), I was unsuccessful to do anything useful with it.

As a workaround for this issue, I tried positioning a `fixed` element on the bottom right corner, hoping that _that_ would always be visible. Unfortunately, it doesn't seem to work, and even fixed elements can be covered by the keyboard.

As for what _does_ work:

1. From my own testing at least, manually (this means done by user interaction, not programmatically) clicking / tapping on the text area works as intended, i.e., the content is 'shifted up' so that it's not covered by the keyboard.
2. Programmatically calling `textareaElement.scrollIntoView()` also seems to work in achieving the desired positioning. The issue is that I don't know how to detect that this situation is happening, and randomly calling `textareaElement.scrollIntoView()` is a bad idea. However, if this proves reliable, maybe we can find good heuristics for when to do this.
3. Regarding iOS: Safari doesn't allow websites to programmatically focus on an input element in a way that will display a virtual keyboard. In other words, `textareaElement.focus()` will 'focus' on it, but will not display a virtual keyboard, unless Safari determines that some user interaction occurred.

See this bug that might be relevant: https://issuetracker.google.com/issues/450752874. However, in my tests, `window.innerHeight` is the same whether a virtual keyboard is being displayed.

### Solution

Unsure how to tackle this, since this mostly seems like a browser bug that's impossible or almost impossible to detect. I propose this:

1. Reporting this as a bug to Chrome and see what the response it. Maybe it gets fixed or maybe not.
2. Regarding GI, explore the `.scrollIntoView()` workaround, since it shows promise. It should be possible to determine conditions narrow enough that will not worsen the user experience (example of a bad solution: `setInterval(()=>t.scrollIntoView(), 100)`, as this will waste CPU and break scrolling functionality).
3. Perhaps some other solution that I haven't tried

corrideat:
Closes #2830
Closes #2905
Closes #2926


shared/domains/chelonia/chelonia.js
```diff
@@ -1127,7 +1127,7 @@ export default (sbp('sbp/selectors/register', {
     const eventsAfterLimit = Math.min(beforeHeight + 1, limit)
     return sbp('chelonia/out/eventsAfter', contractID, offset, eventsAfterLimit, undefined, options)
   },
-  'chelonia/out/eventsBetween': function (contractID: string, startHash: string, endHeight: number, offset: number = 0, { stream } = { stream: true }) {
+  'chelonia/out/eventsBetween': function (contractID: string, startHash: string, endHeight: number, offset: number = 0, limit: number = 0, { stream } = { stream: true }) {
```
taoeffect: As these last parameters are optional, what do you think about moving them (`offset`, `limit`) in with the object at the end to improve readability/usage of the API?

corrideat: Sure, that could work

package.json
```diff
@@ -154,6 +154,7 @@
     "vue-infinite-loading": "2.4.5",
     "vue-router": "3.5.3",
     "vue-slider-component": "3.2.11",
+    "vue-virtual-scroller": "1.1.2",
```
taoeffect: Should probably remove the no longer used one too right?

frontend/views/components/vue-virtual-scroller/DynamicScroller.vue
```diff
@@ -0,0 +1,109 @@
...
+        name="after"
+      />
+    </div>
+    <span ref="endSentinel"></span>
+  </div>
+</template>
+
+<script>
+// import RecycleScroller from './RecycleScroller.vue'
+
```
taoeffect, line 64: Could you add a comment about the history file and what it will be used for?

corrideat: Added a `README.md` file to explain this.

package.json
corrideat: Correct.

corrideat:
Testing this today there seem to be at least two, in this case, relatively minor issues:

1. The message context menu sometimes appears in a way that it's difficult to interact with (e.g., parts of it are hidden behind the scroll area). This likely requires fixing the way that the up/down direction is calculated, which I should do anyhow, using `.getBoundingClientRect`
2. Opening a channel that we're not part of renders blank. I had fixed this in this PR (after it similarly broke) but appears to be broken again (now, because the signatures can't be processed as `._vm` is undefined). I'll look into fixing this again.

On the bright side, <https://github.com/okTurtles/group-income/issues/2905> seems to be fixed now with this PR.

taoeffect:
Two very strange UI bugs with the menu:

https://github.com/user-attachments/assets/91295a59-c54f-4062-9469-c206fb50746a

After clicking away, now when I hover over the message the menu is still there:


https://github.com/user-attachments/assets/019e79f0-2d6e-4388-891c-04b687ac167d

corrideat:
FWIW, the first issue (of the menu being displayed down even though it should be displayed up) seems to affect production too (but you can scroll down, so it's usable).

image

corrideat:
It looks like the CI Cypress issue is fixed now. It was fixed with commit dcaade09af1daed615b08015bea7503a91136bca, which addresses closing the menu options on `mouseleave`. I don't know exactly why the builds were failing before that, or why they were fixed with this commit.

As for the other (first) issue, which is the menu showing down when it should look up, I'm still thinking of a good solution. This is what's happening (and also why it happens in production):

1. The chat container (`c-body-conversation`) has a height that is less than the body (`c-body`). This height is the height of the actual content, and this allows for the messages to be 'pushed down' in this situation.
2. This means that the logic to compute the message position thinks it doesn't have enough space, so the position is down (this part can be trivially fixed by using `c-body` as reference).
3. However, the menu is shown using `position: absolute` and is a descendant of the _message_. There are some rules to hide overflow down the tree, which would cut the menu off. I'd think getting rid of this `overflow: hidden` isn't a good idea.
4. As a result, placing the menu 'up' will make the menu appear cut off.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -845,95 +1257,92 @@ export default ({
...
+      const oldScrollTop = conversation.scrollTop
+      // `ref` is used to distinguish between different invocations
+      const ref = { r: Math.random() }
+      this.ephemeral.rerenderRef = ref
+      conversation.style.overflow = 'hidden'
+      const direction = ''
+      // const pos = conversation.scrollHeight - conversation.scrollTop
+
+      // if (!this.ephemeral.maintainCurrentPosition) {
+      this.$nextTick(() => {
```
corrideat: All of this code following `isNaN()` (and until the end of the function body) is currently unused and likely should be removed. It was a different approach at computing scroll position when new messages came in.

frontend/views/components/menu/MenuParent.vue
```diff
@@ -24,17 +24,20 @@ export default ({
...
+    handleTrigger () {
+      this.$refs.details.open = true
+    },
     handleSelect (itemId) {
       this.closeMenu()
       this.$emit('select', itemId)
     },
     closeMenu () {
-      this.config.Menu.isActive = false
-      this.$emit('menu-close')
```
SebinSong, line 37: Maybe `menu-close` custom event here should be kept? (because menu can also be closed via `handleSelect()` above this, not only via `@toggle` event)

frontend/views/components/menu/MenuTrigger.vue
```diff
@@ -1,13 +1,11 @@
 <template lang='pug'>
-button(
-  v-if='hideWhenActive ? !isActive : true'
-  :aria-pressed='isActive ? "true" : "false"'
+summary(
   data-test='menuTrigger'
-  :class='{ "is-active" : isActive }'
-  @click='handleClick'
+  :class='{ "button": true, "is-active" : isActive }'
```
SebinSong: `summary(:class='{ "button": true, "is-active": isActive}')` here can be just `summary.button(:class='{ "is-active": isActive }')`

frontend/views/components/vue-virtual-scroller/DynamicScrollerItem.vue
```diff
@@ -0,0 +1,49 @@
...
+      type: Boolean,
+      default: false
+    },
+
+    tag: {
+      type: String,
+      default: 'div'
+    }
+  }
+}
```
SebinSong, line 48: I see these various props are defined here but can't seem to find logics associated with these. It seems `DynamicScroller.vue` in `ChatMain.vue` would just work without wrapping the message UI content with this component? If so, I guess this is safe to remove.

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -276,23 +276,25 @@ export default ({
...
   height: 100%;
   width: auto;
+}
 
+.c-menu {
   .c-content {
     @include tablet {
       width: 100%;
       left: auto;
       right: 0.5rem;
```
SebinSong: DOM structure here seems to be re-organized here. Not a serious thing but I see a minor alignment bug introduced here:

image

changing this line to `right: 0;` here fixes it.

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -14,20 +14,22 @@
...
+      :min-item-size='54'
+      key-field='hash'
+      @scroll-start='onScrollStart'
+      @scroll-end='onScrollEnd'
+      @scroll.passive='onChatScroll'
       ref='conversation'
       data-test='conversationWrapper'
-      @scroll='onChatScroll'
-      :class='{"c-invisible": !ephemeral.messagesInitiated}'
+      :data-loaded='ephemeral.messagesInitiated && !ephemeral.loadingUp && !ephemeral.loadingDown && messageState.contract.messages === ephemeral.messages && ephemeral.messages === messageState.contract.messages'
```
SebinSong: We could create a computed prop for this long series of conditions instead of writing these inline?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -127,11 +135,11 @@
 <script>
 import sbp from '@sbp/sbp'
 import { mapGetters } from 'vuex'
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
```
SebinSong: It seems this should be removed?

frontend/views/containers/chatroom/ChatMain.vue
```diff
@@ -148,40 +156,65 @@ import {
...
+  } else if (curScrollTop + 5 > scrollTopMax) {
+    this.onScrollEnd()
+  }
 
-  for (let i = this.messages.length - 1; i >= 0; i--) {
-    const msg = this.messages[i]
-    if (msg.pending || msg.hasFailed) continue
-    const offsetTop = this.$refs[msg.hash][0].$el.offsetTop
+  // for (let i = this.ephemeral.messages.length - 1; i >= 0; i--) {
+  for (let i = (scroller?.$_endIndex ?? this.ephemeral.messages.length) - 1; i >= (scroller?.$_startIndex ?? 0); i--) {
```
SebinSong: I could be wrong but `$_endIndex` and `$_startIndex` here are the features  from `vue-virtual-scroller` package which apparently was later replaced by `DynamicScroller.vue` you wrote. If so, I think these should be removed.

corrideat: You're not wrong. For context, the idea was to use `vue-virtual-scroller`, but this introduced some issues (#2896 has a list of them). Since this PR is already huge as it is, the idea was to leave using `vue-virtual-scroller` for later, but to keep the work I'd done integrating with it, so that it wouldn't need to be done again.

That's why the `scroller?.$_endIndex` part is there. You've raised a good point that this is now unnecessary and maybe it'd be better for performance to comment that out.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: Hm, not sure, maybe some of those styles are being used (but in that case, you're right still that it should be removed and vendored). Thanks for noticing it.

frontend/views/containers/chatroom/ChatMain.vue
corrideat: Good idea.

frontend/views/containers/chatroom/MessageActions.vue
corrideat: Good catch

frontend/views/components/menu/MenuParent.vue
corrideat: The 'toggle' event should be fired every time the menu is closed or opened, though.

See https://codepen.io/corrideat/pen/jEWMXmg

frontend/views/components/vue-virtual-scroller/DynamicScrollerItem.vue
corrideat: I've now deleted all of the unused props

frontend/views/components/menu/MenuParent.vue
corrideat: Marking this as resolved as it looks like it's working fine (see the sample provided, with the toggle event firing every time there's a change, whether user- or programatically-triggeed). If you still think this is an issue, please unresolve.

taoeffect (changes requested):
@corrideat Excellent! This seems really solid!

I tested it fairly thoroughly (I think), and I was able to find only one issue. Let me know if you can reproduce:

On Firefox Dev Edition:

1. Join u1
2. Join u2 in separate container tab, do not visit the chatroom, remain on dashboard
3. Back as u1 start entering a bunch of messages (say 200 or so)
4. Then as u2 switch to the chatroom

In my testing u2 would be taken to the very top of the chat history, instead of the very bottom.

taoeffect:
Here's Sonnet 4.5's review:

<details>


## Critical Issues

### 1. Bug in `processEvents` - Invalid height arithmetic

**Location:** `frontend/views/containers/chatroom/ChatMain.vue` (lines ~1255-1260)

```javascript
if (heightOfFirstEvent < heightOfFirstExistingEvent) {
  if (heightOfFirstEvent + deserializedEvents.length >= heightOfFirstExistingEvent) {
    if (heightOfFirstEvent + deserializedEvents.length <= heightOfLastExistingEvent) {
      this.latestEvents.unshift(...deserializedEvents.slice(0, heightOfFirstExistingEvent - heightOfFirstEvent))
    } else {
      this.latestEvents = deserializedEvents
    }
```

**Issue:** You're adding `deserializedEvents.length` (an array length) to `heightOfFirstEvent` (a contract height). Contract heights are not sequential counters that increment by 1 - they're arbitrary numbers. This arithmetic is incorrect.

**Suggested fix:** You should be comparing the height of the last event in the array:
```javascript
const heightOfLastEvent = deserializedEvents[deserializedEvents.length - 1].height()
if (heightOfFirstEvent < heightOfFirstExistingEvent) {
  if (heightOfLastEvent >= heightOfFirstExistingEvent) {
    if (heightOfLastEvent <= heightOfLastExistingEvent) {
      this.latestEvents.unshift(...deserializedEvents.slice(0, heightOfFirstExistingEvent - heightOfFirstEvent))
    } else {
      this.latestEvents = deserializedEvents
    }
```

Similar issues exist in the `else if` branch around line 1266.

### 2. Incorrect null check in `loadMoreMessages`

**Location:** `frontend/views/containers/chatroom/ChatMain.vue` (line ~1194)

```javascript
const beforeHeight = this.ephemeral.currentLowestHeight
if (this.ephemeral.loadingUp || beforeHeight === 0) return
```

**Issue:** This check allows `undefined` to pass through. If `currentLowestHeight` hasn't been initialized, `beforeHeight` will be `undefined`, and the check won't catch it.

**Suggested fix:**
```javascript
if (this.ephemeral.loadingUp || beforeHeight == null || beforeHeight === 0) return
```

### 3. Off-by-one error in `scrollToItem`

**Location:** `frontend/views/components/vue-virtual-scroller/DynamicScroller.vue` (line ~113)

```javascript
scrollToItem (index: number) {
  this.$el.querySelector('.vue-recycle-scroller__item-view:nth-child(' + index + ')')?.scrollIntoView()
}
```

**Issue:** CSS `:nth-child()` is 1-indexed, but JavaScript array indices are 0-indexed. When you pass `index: 0`, the selector looks for the 0th child, which doesn't exist.

**Suggested fix:**
```javascript
scrollToItem (index: number) {
  this.$el.querySelector('.vue-recycle-scroller__item-view:nth-child(' + (index + 1) + ')')?.scrollIntoView()
}
```

### 4. Potential event target mismatch in `handleToggle`

**Location:** `frontend/views/components/menu/MenuParent.vue` (line ~28)

```javascript
handleToggle (e) {
  if (this.config.Menu.isActive === e.target.open) return
  this.config.Menu.isActive = e.target.open
```

**Issue:** If there are nested `<details>` elements, `e.target` might not be the details element you expect - it could be a nested one. The `toggle` event doesn't bubble in a useful way here.

**Suggested fix:**
```javascript
handleToggle (e) {
  // Use currentTarget instead of target to get the element the listener is attached to
  const isOpen = this.$refs.details.open
  if (this.config.Menu.isActive === isOpen) return
  this.config.Menu.isActive = isOpen
```

## Minor Issues

### 5. Inconsistent event handler naming

**Location:** `frontend/views/containers/chatroom/ChatMain.vue`

The file has both `onScrollStart`, `onScrollEnd`, and `onScrollEvt` which is confusing naming. Consider renaming `onScrollEvt` to `onScrollDirectionChange` or similar for clarity.

### 6. Empty `<menu>` in template

**Location:** `frontend/views/components/ButtonDropdownMenu.vue` (line ~18)

```pug
menu-content.c-menu-content(:class='"bound-to-" + boundEdge')
  menu
    template(v-if='options')
```

You changed from `<ul>` to `<menu>`, which is semantically correct, but make sure your CSS is updated to handle the `<menu>` element properly. The reset styles you added look good, but double-check browser compatibility.

## Questions

1. In `ChatMain.vue` around line 441, you have `enqueue.call(this, async () => { ... })` - is there a reason you're using `.call()` instead of just calling `enqueue(() => { ... })`? The function is already bound in context.

2. The `DynamicScroller` stub implementation doesn't actually do virtual scrolling - it renders all items. Have you tested this with very large message lists (1000+ messages) to ensure performance is acceptable?

3. In the `closeMenu` method of `MenuContent.vue`, what's the intended behavior when clicking inside a details element? The logic seems unclear:
   ```javascript
   if (e.target?.closest('details') === this.$el.closest('details')) {
     return
   }
   ```

</details>

taoeffect:
And here's Grok 4's:

<details>

### Bug: Potential race condition in `loadMoreMessages` when loading downwards for non-joined chatrooms

In `frontend/views/containers/chatroom/ChatMain.vue`, the `loadMoreMessages` method has a special case for loading all messages when not joined to the chatroom (setting `limit = Infinity` and `sinceHeight = 0`). However, after processing events, it sets `this.ephemeral.postSetMessageState` to scroll to the bottom, but this callback doesn't check if the chatroom has switched before executing the scroll. This could lead to scrolling in the wrong chatroom if the user switches quickly during loading.

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a4d983d5f..a4d983d5f 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1,742 +1,742 @@
-        } else {
-          this.ephemeral.postSetMessageState = () => {
-            if (this.chatroomHasSwitchedFrom(chatRoomID)) return
-            if (!this.isJoinedChatRoom(chatRoomID)) {
-              this.jumpToLatest('instant')
-            }
-          }
-        }
+        } else {
+          this.ephemeral.postSetMessageState = () => {
+            if (this.chatroomHasSwitchedFrom(chatRoomID)) return
+            if (!this.isJoinedChatRoom(chatRoomID)) {
+              this.jumpToLatest('instant')
+            }
+          }
+        }
```

**Suggested fix**: Add a chatroom switch check inside the `postSetMessageState` callback to prevent scrolling in a stale chatroom. Update the code as follows:

```javascript
// In loadMoreMessages (downwards branch)
await sbp('chelonia/out/eventsAfter', chatRoomID, { sinceHeight, limit, stream: false }).then((events) => {
  if (this.chatroomHasSwitchedFrom(chatRoomID)) return
  return this.processEvents(events, direction).then(() => {
    // Special case: if loading events for a chatroom we're not part
    // of, scroll to bottom
    if (limit !== Infinity || sinceHeight !== 0) return
    this.ephemeral.postSetMessageState = () => {
      if (this.chatroomHasSwitchedFrom(chatRoomID)) return // Add this check
      if (!this.isJoinedChatRoom(chatRoomID)) {
        this.jumpToLatest('instant')
      }
    }
  })
}).finally(() => {
  if (this.ephemeral.loadingDown === instance) {
    this.ephemeral.loadingDown = false
  }
})
```

### Bug: Missing error handling in `processEvents` when deserializing events

In `frontend/views/containers/chatroom/ChatMain.vue`, the `processEvents` method deserializes events without catching potential deserialization errors (e.g., malformed events). This could crash the chat loading if invalid data is received.

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a4d983d5f..a4d983d5f 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1,742 +1,742 @@
-        const deserializedEvents = events.map((event) => {
-          return SPMessage.deserialize(event)
-        })
+        const deserializedEvents = events.map((event) => {
+          return SPMessage.deserialize(event)
+        })
```

**Suggested fix**: Wrap the deserialization in a try-catch and log errors, skipping invalid events to prevent full failure. Update the code as follows:

```javascript
// In processEvents
await enqueue.call(this, async () => {
  // ...
  const deserializedEvents = events.map((event) => {
    try {
      return SPMessage.deserialize(event)
    } catch (e) {
      console.error('[ChatMain.vue/processEvents] Error deserializing event', e)
      return null // Skip invalid event
    }
  }).filter(Boolean); // Remove nulls

  // If no valid events, return early
  if (!deserializedEvents.length) return;

  // Continue with existing logic...
})
```

### Bug: Infinite loop risk in `onChatScroll` when saving scroll position

In `frontend/views/containers/chatroom/ChatMain.vue`, the `onChatScroll` method iterates over visible messages to save the scroll position. If no valid message is found (e.g., all pending/failed), it might not emit the event, but repeated scrolls could spam emissions without checks.

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a4d983d5f..a4d983d5f 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1,742 +1,742 @@
-    if (this.ephemeral.scrollableDistance > ignorableScrollDistanceInPixel) {
-      // Save the current scroll position per each chatroom
-      const iterator = this.visibleMessageIterator()
-      for (const msg of iterator) {
-        if (msg.pending || msg.hasFailed) continue
-
-        sbp('okTurtles.events/emit', NEW_CHATROOM_SCROLL_POSITION, {
-          chatRoomID: this.ephemeral.renderingChatRoomId,
-          messageHash: msg.hash
-        })
-        break
-      }
-    }
+    if (this.ephemeral.scrollableDistance > ignorableScrollDistanceInPixel) {
+      // Save the current scroll position per each chatroom
+      const iterator = this.visibleMessageIterator()
+      for (const msg of iterator) {
+        if (msg.pending || msg.hasFailed) continue
+
+        sbp('okTurtles.events/emit', NEW_CHATROOM_SCROLL_POSITION, {
+          chatRoomID: this.ephemeral.renderingChatRoomId,
+          messageHash: msg.hash
+        })
+        break
+      }
+    }
```

**Suggested fix**: Add a flag to prevent re-emitting if the position hasn't changed. Update the code as follows:

```javascript
// In onChatScroll (scroll position saving part)
if (this.ephemeral.scrollableDistance > ignorableScrollDistanceInPixel) {
  const iterator = this.visibleMessageIterator()
  for (const msg of iterator) {
    if (msg.pending || msg.hasFailed) continue

    // Add check: only emit if different from current scroll position
    if (msg.hash !== this.currentChatRoomScrollPosition) {
      sbp('okTurtles.events/emit', NEW_CHATROOM_SCROLL_POSITION, {
        chatRoomID: this.ephemeral.renderingChatRoomId,
        messageHash: msg.hash
      })
    }
    break
  }
}
```

### Improvement: Optimize `visibleMessageIterator` for performance

In `frontend/views/containers/chatroom/ChatMain.vue`, `visibleMessageIterator` iterates over all messages even though it's a generator. Since it's a stub without real virtual scrolling, this could be slow for large message lists. Suggest limiting to visible range manually.

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a4d983d5f..a4d983d5f 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1,742 +1,742 @@
-    * visibleMessageIterator () {
-      if (!this.$refs.conversation) return
-      const messages = this.ephemeral.messages
-
-      for (let i = 0; i < messages.length; i++) {
-        const msg = messages[i]
-        if (!this.isMessageVisible(msg.hash)) continue
-        yield msg
-      }
-    },
+    * visibleMessageIterator () {
+      if (!this.$refs.conversation) return
+      const messages = this.ephemeral.messages
+
+      for (let i = 0; i < messages.length; i++) {
+        const msg = messages[i]
+        if (!this.isMessageVisible(msg.hash)) continue
+        yield msg
+      }
+    },
```

**Suggested fix**: Manually estimate visible range based on scroll position and average item height (e.g., 54 as min-item-size). Update the code as follows:

```javascript
// In ChatMain.vue
* visibleMessageIterator () {
  if (!this.$refs.conversation) return
  const conversation = this.$refs.conversation.$el
  const scrollTop = conversation.scrollTop
  const clientHeight = conversation.clientHeight
  const avgHeight = 54 // From min-item-size prop

  const startIndex = Math.max(0, Math.floor(scrollTop / avgHeight) - 5) // Buffer
  const endIndex = Math.min(this.ephemeral.messages.length, Math.ceil((scrollTop + clientHeight) / avgHeight) + 5)

  const messages = this.ephemeral.messages.slice(startIndex, endIndex)

  for (let i = 0; i < messages.length; i++) {
    const msg = messages[i]
    if (!this.isMessageVisible(msg.hash)) continue
    yield msg
  }
},
```

### Improvement: Add debounce to `applyTopPadding` to reduce reflows

In `frontend/views/containers/chatroom/ChatMain.vue`, `applyTopPadding` is called on every resize/mutation, which could cause performance issues with frequent changes. Suggest debouncing it.

```diff
diff --git a/frontend/views/containers/chatroom/ChatMain.vue b/frontend/views/containers/chatroom/ChatMain.vue
index a4d983d5f..a4d983d5f 100644
--- a/frontend/views/containers/chatroom/ChatMain.vue
+++ b/frontend/views/containers/chatroom/ChatMain.vue
@@ -1,742 +1,742 @@
-    applyTopPadding () {
-      const conversation = this.$refs.conversation?.$el
-      const conversationTP = conversation?.querySelector('.c-top-padding')
-      if (!conversation || !conversationTP || !conversation.parentElement) return
-      const padding = Math.max(conversation.parentElement.clientHeight - conversation.scrollHeight + conversationTP.clientHeight - 1, 0)
-      conversationTP.style.height = `${padding}px`
-    },
+    applyTopPadding () {
+      const conversation = this.$refs.conversation?.$el
+      const conversationTP = conversation?.querySelector('.c-top-padding')
+      if (!conversation || !conversationTP || !conversation.parentElement) return
+      const padding = Math.max(conversation.parentElement.clientHeight - conversation.scrollHeight + conversationTP.clientHeight - 1, 0)
+      conversationTP.style.height = `${padding}px`
+    },
```

**Suggested fix**: Wrap in debounce (e.g., 100ms) to batch updates. Add at the top of methods:

```javascript
// In methods
applyTopPadding: debounce(function () {
  const conversation = this.$refs.conversation?.$el
  const conversationTP = conversation?.querySelector('.c-top-padding')
  if (!conversation || !conversationTP || !conversation.parentElement) return
  const padding = Math.max(conversation.parentElement.clientHeight - conversation.scrollHeight + conversationTP.clientHeight - 1, 0)
  conversationTP.style.height = `${padding}px`
}, 100),
```

</details>

I'm curious which did better.

corrideat:
> In my testing u2 would be taken to the very top of the chat history, instead of the very bottom.

I was able to reproduce.

SebinSong (dismissed):

corrideat:
@taoeffect I was doing some quick debugging about the issue you found, and I observed this:

1. Your situation is reproducible
2. If you have a new user join _after_ messages are sent, then they'll be taken to the very bottom
3. Depending on how you look at it, this is working as intended. The reason you get to the top instead of the bottom is that `this.currentChatRoomReadUntil?.messageHash` points to the first message instead of the last (which in itself is correct).

How do you suggest this should be addressed? (based on the description of _why_ this happens)

taoeffect:
Ah...... yeahhhh... ok. Let's leave it as-is then.

Did either of the bots find anything worthwhile?

corrideat:
> Did either of the bots find anything worthwhile?

As usual most of it seems to be noise, but I'm going to go over it once more. At least one issue, e.g., the off-by-one one on `nth-child` at least seems plausible.

The other one that could be useful is the suggestion to estimate the visible elements on `visibleMessageIterator` to avoid iterating over every message (if lots of messages are loaded, most of them won't be visible, unless the screen is extremely tall).

corrideat:
> 1. Bug in processEvents - Invalid height arithmetic

The height arithmetic seems fine and heights are sequential, not arbitrary numbers. If this weren't correct, things would break.

> Incorrect null check in loadMoreMessages

The check isn't a `null` check, it's deliberately a check for `0`.

> Off-by-one error in scrollToItem

This seems like a legitimate issue

> Potential event target mismatch in handleToggle

The toggle event doesn't bubble at all, so this issue is bogus.

> Inconsistent event handler naming

Maybe on `onScrollEvt` could have a better name, but it's related to the action of scrolling.

> Empty <menu> in template

I fail to see the issue here.

> Bug: Potential race condition in loadMoreMessages when loading downwards for non-joined chatrooms

The code seems to have this check.

> Bug: Missing error handling in processEvents when deserializing events

A valid issue, but I'm not sure how to handle this in a different way that's also useful. Aborting in this case seems fine (?), since invalid events shouldn't be something that happens.

> Bug: Infinite loop risk in onChatScroll when saving scroll position

This seems like a valid issue, as is the suggested fix.

> Improvement: Optimize visibleMessageIterator for performance

Seems like a valid issue.

> Improvement: Add debounce to applyTopPadding to reduce reflows

The point of the top padding is to react to height changes. Debouncing it would cause weird user-visible reflows.

taoeffect:
Note that when testing these changes on staging I noticed a blank grey chatroom when switching to it shortly after logging in afresh:
 
Screenshot 2025-10-06 at 1 30 52 PM

If I waited for all network activity to stop before switching from the dashboard to the chatroom, this wouldn't happen.

NOTE: if you know the fix we can do it now, or we can make this a separate issue.

SebinSong (changes requested):
Found a UI regression in this PR last week but forgot to mention.
The chat message menu acts differently from the PROD when clicked:

**[This PR]**

https://github.com/user-attachments/assets/fec3a6fb-f769-4863-8cab-5b2f58548a81

**[PROD]**

https://github.com/user-attachments/assets/c154a16b-defa-46c1-8b1a-69911d4854cf

corrideat:
Thanks @SebinSong

frontend/views/containers/chatroom/MessageActions.vue
```diff
@@ -44,32 +44,32 @@
...
 
-    menu-parent(ref='menu' @menu-open='moreOptionsTriggered')
+    menu-parent(ref='menu' @menu-open='moreOptionsTriggered' @menu-close='moreOptionsClosed')
       menu-trigger.is-icon-small(
         :aria-label='L("More options")'
       )
         i.icon-ellipsis-h
       message-actions-mobile(
-        v-if='chatMainConfig.isPhone'
+        v-if='chatMainConfig.isPhonexx'
```
taoeffect: iPhonexx? 🤔

corrideat: Right, this should be removed.

corrideat: (Now, we can have a single element for the phone and \>tablet displays, since both will use the new `dialog`, but with different styling)

package.json
taoeffect: ~~This is still here?~~ SRY broken Github UI

shared/domains/chelonia/chelonia.js
taoeffect: ~~What happened here?~~ NM github UI is broken

package.json
corrideat: It's not from what I can see. I see only one change, which is removing "vue-infinite-loading"

taoeffect (changes requested):
While testing this with `ngrok` on mobile Android, I discovered two issues:

1. For some reason the unread message marker stays up and doesn't disappear when I'm viewing the channel. This was me signed in as u3 on mobile, switched to the #general channel, looking at it, and there was still the unread message (see video below). To get rid of it I had to post a message as u3 to the channel (not shown in video).
2. Then when tapping on the input field the UI is supposed to shift upward so that it's above the keyboard, but instead the keyboard appears hiding the UI.

See this video for bugs 1 & 2:

https://github.com/user-attachments/assets/d8423cc2-0bc4-444e-837d-a75a23c11bf2

I then discovered 3rd bug where if we are not joined to a channel with a lot of messages, and someone posts a link to a message high up in the conversation, then after we click it and start scrolling down a bit, it loads the entire history. Note that this behavior works correctly if we're already joined to the channel. See attached video for example of this bug:

https://github.com/user-attachments/assets/d459849a-2fdd-421d-ad3c-1d7e89d81291

corrideat:
Note that for the last issue, that already happens on master (meaning the entire history is always loaded) and that linking to a message on a chatroom that hasn't been joined doesn't work (https://github.com/okTurtles/group-income/issues/2857).

corrideat:
> Then when tapping on the input field the UI is supposed to shift upward so that it's above the keyboard, but instead the keyboard appears hiding the UI.

Apparently this too happens on groupincome.app. For clarity, this is what happens:

{6EB5AF7A-B67A-4203-807A-5FFA0EC5A0E8}

The desired behaviour is more like the screenshot below (that's not perfect either because the keyboard is still covering some of the input field; instead, the entire input field should be visible)

{378D5A47-5D70-4DCD-B5BA-DDFBAE32D43A}

SebinSong:
@corrideat 

> Apparently this too happens on groupincome.app.

I just tried chatroom both in PROD and in this PR and it does seem like there is a regression regarding the keyboard hiding the `<textarea />`. 

**[In PROD]**

https://github.com/user-attachments/assets/8712e8a9-1689-4e85-bf8a-ab7c1c8ccbcd

**[In this PR]**

https://github.com/user-attachments/assets/d7d93f92-0071-451d-849d-59ab767d593e

---

It is likely caused by the DOM restructure introduced in this PR. I think that cause should be tracked down and fixed first instead of looking for a way to detect if virtual keyboard is visible/hidden.

SebinSong:
@corrideat 
While checking that using iOS Safari, I saw a few more UI/UX regressions BTW:

**[1. Arrows showing next to menu buttons]**

image

<br>

This is caused by these DOM structure update (`<summary>` and `<details>`. they used to be `<div>` and `<button>`):



<br><br>

**[2. Cannot dismiss the channel-menu on mobile]**

https://github.com/user-attachments/assets/64f7fdc7-dfd8-4ab8-8847-e18fd4b5ee02

In firefox mobile-emulator:

https://github.com/user-attachments/assets/fbb16789-32e2-4baa-a4fe-7d975da20bf3

corrideat:
Thank you for the additional issues you reported! I'll add a new comment once these issues are fixed.

1. Arrows showing next to menu button

This doesn't seem to happen on real devices, although I've applied a CSS fix for that (`&::-webkit-details-marker, &::marker {display: none}`, before it was just the `::marker` selector). <ins>Should be fixed</ins>

2. regression regarding the keyboard hiding the textarea

I tried this on a real iOS device and was able to partially reproduce:

   a. If you switch, e.g., from the dashboard to the chatroom, I was able to reproduce (i.e., the textarea gets focused but it's not visible)
   b. Switching from one chatroom to another doesn't seem to trigger this issue (no textarea shown). This should actually cause the keyboard to show, since the textarea is supposed to have focus (and confirmed via CSS that it does; ETA: Checked on `groupincome.app`, and switching between chatrooms, or even switching to the chat page doesn't result in the keyboard being opened)
   c. Manually tapping on the textarea works the way it should (i.e., it's not covered by the keyboard)
   
ETA: Now I believe I have an explanation for this regression and why it behaves the way it does. On iOS, apparently the keyboard will only be displayed in response to some kind of user interaction. Previously, the SendArea wasn't being rendered as soon as the chatroom loaded, which it now is

ETA 2: <ins>Should be fixed by forcing scrolling into view, need to test more.</ins>

3. Cannot dismiss the channel-menu on mobile

Was able to reproduce and will fix it.  <ins>Should be fixed</ins>

frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -476,6 +481,12 @@ export default ({
         this.$refs.textarea.focus()
       }
     },
+    textAreaFocus (event) {
+      // Sometimes, on mobile, the virtual hardware keyboard appears
+      // over the page. This doesn't seem to be detectable, but scrolling
+      // seems to work around it.
+      event.target.scrollIntoView()
+    },
```
taoeffect, line 497: Could you modify this so that:

1. It only runs on Android devices (if that's detectable)
2. This code is run after a `500` msec delay
3. It scrolls into view not the text area but the entire block surrounding it (so that you can see the buttons below too)

corrideat: Is there a reason for these specific changes?

1. Browser sniffing should be a last resort and also is unreliable. I don't think this is breaking anything as is.
2. Running it after an arbitrary delay is, well, arbitrary and should also be avoided (the concern is introducing weird scrolling behaviour). It does seem like there's some room for improvement here though, as running it immediately doesn't seem to fix the issue.
3. This is a good idea

taoeffect: 1. Because it seems to work fine in other environments so adding this code could potentially mess those up by being worse than native behaviour 
2. Yeah a delay is needed for the keyboard appearing animation to compete

taoeffect (approved):
Fantastic work @corrideat!!! This was a massive achievement on your part!

corrideat:
@SebinSong A heads up that this PR looks ready. If you're able to give it a test or have feedback, please do so. Thanks!

SebinSong:
The issues I reported earlier seem to have been fixed now. I saw a couple of more UI bugs so leaving here.
(I think these can be taken care of as a follow-up by either of us.)

**[Firefox desktop only - The first item in the message-actions menu is pre-focused when the it is opened]**

https://github.com/user-attachments/assets/2a6faa84-77de-484b-9000-e304ba90f130

**['Channel options' text is off-centered in chatroom menu on mobile]**

image

---

`grunt dev --tunnel` didn't work stable enough today for me to test on my Android Chrome/Brave. So after it is deployed to PROD, let me do another round of general check in GIG on my Android browsers.

SebinSong (approved):

SebinSong:
BTW I saw [this behaviour](https://github.com/okTurtles/group-income/pull/2841#issuecomment-3408756525) I reported still happening while testing on iOS emulator. will see if this happens on my Android browsers once deployed to PROD.

corrideat:
> @SebinSong BTW I saw https://github.com/okTurtles/group-income/pull/2841#issuecomment-3408756525 I reported still happening while testing on iOS emulator. will see if this happens on my Android browsers once deployed to PROD.

There's sort of a fix (more of a hack) for Android. iOS seems to work fine (on a device at least), so it's not running there. If we see it happen on iOS (not too worried about the emulator) we might need to do something.

SebinSong:
Yes, could be an issue only happening on emulator.
This iOS emulator has been quite useful in reproducing/fixing multiple iOS specific issues reported in the past FYI:



So been using it for checking how things would behave on real device after development.
If other people using iOS device for GIG see this in PROD, they will report I think.

PR #2852, opened 2025-06-21
Zero out size

### LINKED ISSUE #2846: Zero out size when identity contract deleted
### Problem

When an identity contract and all its data are deleted, for some reason the `_private_ownerTotalSize`, `_private_ownerBalanceHistoryGranular`, and maybe `_private_ownerBalanceHistoryCoarse` still show non-zero values for the size of that identity contract.

### Solution

Make them reflect the actual size of data associated with that contract, whether it's 0 or something else.

corrideat:
Closes #2846

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -376,7 +376,7 @@ sbp('sbp/selectors/register', {
       const owner = await sbp('chelonia.db/get', `_private_owner_${cid}`)
       if (owner && !ultimateOwnerID) ultimateOwnerID = await lookupUltimateOwner(owner)
       const rawManifest = await sbp('chelonia.db/get', cid)
-      const size = ultimateOwnerID && await sbp('chelonia.db/get', `_private_size_${cid}`)
+      const size = await sbp('chelonia.db/get', `_private_size_${cid}`)
       // If running in a persistent queue, already deleted contract should not
       // result in an error, because exceptions will result in the task being
       // re-attempted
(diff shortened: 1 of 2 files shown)


backend/server.js
```diff
@@ -376,7 +376,7 @@ sbp('sbp/selectors/register', {
       const owner = await sbp('chelonia.db/get', `_private_owner_${cid}`)
       if (owner && !ultimateOwnerID) ultimateOwnerID = await lookupUltimateOwner(owner)
       const rawManifest = await sbp('chelonia.db/get', cid)
-      const size = ultimateOwnerID && await sbp('chelonia.db/get', `_private_size_${cid}`)
+      const size = await sbp('chelonia.db/get', `_private_size_${cid}`)
```
taoeffect, line 379: Removing the check for `ultimateOwnerID` is ultimately the only thing this PR seems to do, could you please help me understand how it closes the issue?

- Why was the check there before
- Why is it safe to remove it?
- Why does removing it fix the issue?
- And does removing it also result in the `_private_ownerBalanceHistoryGranular` having a `0` for the most recent `sizeTotal` entry? (Note: I don't think that matters much anymore, as long as we make sure OTS is correct...)

Also, in light of the recent discovery that in fact the server did (eventually) zero out the OTS for the identity contract that was removed (possibly after a restart though), is this change still necessary?

taoeffect (commented):
A few questions:

backend/server.js
corrideat: > * Why was the check there before

I reckoned it'd result in additional processing that wouldn't be necessary. This check used to skip updating size information for contracts that don't have an `ultimateOwnerID`, i.e., contracts that are themselves billable entities.

> * Why is it safe to remove it?

Because this just calls into the worker to schedule a task. The addition or removal of additional conditions is 'safe', however having conditions that are too lax (e.g., calling this on all contracts every minute) is undesirable because it'd result in additional computations. (The flip side is that conditions that are too strict results in the size not being updated as often as it should, see this issue).
 
> * Why does removing it fix the issue?

Because then size update calculations are performed for all contracts upon deletion. Previously, this would only be done for deleted contracts that weren't billable entitites.

> * And does removing it also result in the _private_ownerBalanceHistoryGranular having a 0 for the most recent sizeTotal entry? (Note: I don't think that matters much anymore, as long as we make sure OTS is correct...)

It does not, at least not directly. Size calculations are used by but separate from billing. Billing is supposed to stop due to the contract being removed from the billable entities list, so the owner total size is largely irrelevant. However, in the event that for some reason billing does continue, the new amount used should be 0 (or whatever the computed size after deletion was, which is supposed to be 0).

taoeffect (approved):
Seems OK 👍

PR #2858, opened 2025-06-26
Convert docs to mdBook

sirodoht:
Hi there!

This PR introduces a couple of configuration files in your docs directory that enable an [mdBook](https://github.com/rust-lang/mdBook) build. mdBook is a simple lightweight gitbook clone of the early gitbook theme.

There is two main files that define an mdBook setup:

1. `book.toml`, which includes certain metadata
2. `src/SUMMARY.md` which defines the website structure and the sidebar content

Additionally, I included a GitHub Actions workflow that builds the docs site and deploys to GitHub Pages (if that's something of interest) and a `docs/src/Introduction.md` page, as a frontpage for the docs website.


docs/src/Docs-Contribute.md
```diff
@@ -0,0 +1,19 @@
+# Docs Contribute
+
+Docs are build using [mdBook](https://github.com/rust-lang/mdBook).
```
taoeffect: Typo: `Docs are build` => `Docs are built`

taoeffect (commented):
Nice! Thank you @sirodoht for sending this in!

I found one typo and have a question for you: do you know if there's anything special we need to do to have these docs deployed to a Github Pages site? Or will merging be sufficient?

I see there's this comment:

```
# Set GITHUB_TOKEN permissions to allow deployment to GitHub Pages
```

But not sure what it means.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.51
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install mdBook
+        run: |
+          curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
```
corrideat: Considering specifically the `id-token: write` permission above, I'm not too happy with this 'install an unverified package from the Internet' part.

Ideally, this would download a specific, well-known pre-built version of mdbook and verify the hash.

ETA: Maybe the OIDC token part isn't as scary as I originally thought, but still the point about installing a specific version stands.

taoeffect: Ah that's a good point that @corrideat noticed, but also installing the entire Rust toolchain in order to build and run `mdbook` is quite slow.

Could this be updated to just download the binary itself from here?

https://github.com/rust-lang/mdBook/releases/tag/v0.4.51

And then yeah a `sha256sum -c` can be used to verify the integrity of the binary

sirodoht:
Thanks both! I have:

- [x] Rebased with latest `master` commits
- [x] Fixed the "Docs are build" typo
- [x] Improved the comment for the GitHub Actions workflow permissions
- [x] Switched installation of mdbook to downloading the binary and verifying the hash
- [x] Switched CI container image to fixed `ubuntu-24.04` instead of `ubuntu-latest`

To your question, @taoeffect, I think merging should be enough although I am cautious that there might be the case that one needs to enable GitHub Pages for this repository from [here](https://github.com/okTurtles/group-income/settings/pages).

taoeffect:
Awesome, thanks @sirodoht!

> I am cautious that there might be the case that one needs to enable GitHub Pages for this repository from [here](https://github.com/okTurtles/group-income/settings/pages).

I've looked at these settings and here's what they are currently:

Screenshot 2025-07-01 at 10 22 11 AM

Questions:

Do I switch the "Source" to Github Actions?


Screenshot 2025-07-01 at 10 20 30 AM

For the "Branch", do I pick `master` and `/docs`?

Screenshot 2025-07-01 at 10 20 34 AM

sirodoht:
@taoeffect 

> Do I switch the "Source" to Github Actions?

Yes!

> For the "Branch", do I pick `master` and `/docs`?

And yes.

I think just merging the PR also changes the settings automatically, but if not, those are the options we want.

corrideat (approved):
Thank you @sirodoht for the PR and for implementing the changes requested. It looks fine from what I can see.

taoeffect:
Thanks @sirodoht, just merged your PR!

Unfortunately, for whatever reason, even though the tests passed for the PR, they [failed after the merge into `master`](https://github.com/okTurtles/group-income/actions/runs/16030520810/job/45229637282):

Screenshot 2025-07-02 at 9 25 57 AM

Would you be able to send us a followup PR to address this?

And also something else to check: some of the documentation links to other documentation, for example, the [Troubleshooting section of the main README](https://github.com/okTurtles/group-income?tab=readme-ov-file#troubleshooting) links to the Troublehsooting docs. But now that link is broken. Likewise, other docs most likely link to other docs too and those links will now be broken and need to be fixed.

corrideat:
Regarding the fix:

I've just noticed that echo "${MDBOOK_SHA256}  mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz" | sha256sum -c - is wrong in docs.yml, which is why CI failed. The correct line filename should be `mdbook.tar.gz` as used in the above curl line (alternatively, curl could use the longer file name as output).

Since it looks like an additional commit will be needed, I'd also change `tar xf` to `tar -xzf`. Both work but the latter is more explicit and thus clearer IMHO.

taoeffect:
Ah, OK, this seems like something I can fix. So I'll go ahead and send a PR with this fix + fixing docs links.

taoeffect:
Thanks @sirodoht, we got it working!

http://okturtles.github.io/group-income/

EDIT: LOL, at least it was working. Now it's showing a 404 for some reason.

sirodoht:
Haha oh no, did it just break suddenly? Maybe retry the CI job?

taoeffect:
We've had a few commits made since, and it's still not working. :-\

sirodoht:
Hmm ok that's weird. The workflow seems to run fine. Are there any alerts on https://github.com/okTurtles/group-income/settings/pages?

taoeffect:
Don't see any...

sirodoht:
Does it say "Your site is live at <url> Last deployed by user"?

taoeffect:
Yeah:

> Your site is live at http://okturtles.github.io/group-income/
Last [deployed](https://github.com/okTurtles/group-income/actions/runs/16056764165) by @[taoeffect](https://github.com/taoeffect) taoeffect 4 hours ago

PR #2861, opened 2025-07-02
attempt to fix docs

taoeffect:
Followup to @sirodoht 's #2858


corrideat (dismissed):
LGTM

CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@
 - Check the [open issues](https://github.com/okTurtles/group-income/issues).
 - View the [project board](https://github.com/okTurtles/group-income/projects).
 - Check especially what's [`up for grabs`](../../labels/Note%3AUp-for-grabs), [`high priority`](../../labels/Priority%3AHigh), [`level: starter`](../../labels/Level%3AStarter), or has a **[`Bounty`](https://github.com/okTurtles/group-income/issues?q=is%3Aopen+is%3Aissue+label%3ANote%3ABounty)**
```
corrideat: Is removal of these intentional. If not, I suggest hardcoding the absolute URL.

taoeffect: Removal of what...? These lines are unmodified...

.github/workflows/docs.yml
```diff
@@ -30,8 +30,8 @@ jobs:
       - name: Install mdBook
         run: |
           curl -L -o mdbook.tar.gz "https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
```
corrideat, line 43: Commenting here as GH will not let me comment above. To avoid issues with building the docs, I suggest making this run on pull requests too, but skipping the actual deployment part (or if possible, deploying to a per-commit or per-branch site)

taoeffect: I've no idea how to do that. If you have specific changes you'd like to suggest...

CONTRIBUTING.md
corrideat: Sorry, I saw the `-` and wrongly believed they were deleted (my setup where I say this doesn't use colours). However, are you sure that the links will work on the GH pages site? It might still be worth it to set them to absolute URLs.

taoeffect: OK, sure that's a good suggestion, just pushed a fix

.github/workflows/docs.yml
corrideat: To make specific actions conditionals this should work (see https://docs.github.com/en/actions/reference/accessing-contextual-information-about-workflow-runs#context-availability):

```
if: ${{ github.event_name == 'push' }}
```

The entire file would be something like this:

```yml
name: Deploy docs

on:
  push:
    branches: ["master"]
  pull_request:
    branches: ["master"]
  workflow_dispatch:

# Grant permissions:
# - contents:read to checkout repository code
# - pages:write to deploy to GitHub Pages
# - id-token:write to authenticate via OIDC
# See:  https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token
permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  build:
    runs-on: ubuntu-24.04
    env:
      MDBOOK_VERSION: 0.4.51
      MDBOOK_SHA256: d611bcde080f1ab9932ca1724197ac2c23bb8c5d64581bb5da238bfbb4d39184
    steps:
      - uses: actions/checkout@v4
      - name: Install mdBook
        run: |
          curl -L -o mdbook.tar.gz "https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
          echo "${MDBOOK_SHA256}  mdbook.tar.gz" | sha256sum -c -
          tar -xzf mdbook.tar.gz
          chmod +x mdbook
          sudo mv mdbook /usr/local/bin/
      - name: Build with mdBook
        run: cd docs && mdbook build
      - name: Setup GitHub Pages
        id: pages
        if: ${{ github.event_name == 'push' }}
        uses: actions/configure-pages@v5
      - name: Upload artifact
        if: ${{ github.event_name == 'push' }}
        uses: actions/upload-pages-artifact@v3
        with:
          path: ./docs/book

  deploy:
    if: ${{ github.event_name == 'push' }}
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-24.04
    needs: build
    steps:
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v4
```

taoeffect: @corrideat does Opus make any good suggestions here?

I've reviewed the workflow file and found several issues and suggestions for improvement:

## Issues:

1. **Redundant event triggers**: The workflow runs on both `push` to master and `pull_request` to master. This means PRs will trigger the workflow twice - once for the PR event and once for the push event when merged. Consider removing the `pull_request` trigger since you're only deploying on push anyway.

2. **Missing dependency caching**: The workflow downloads mdBook on every run. This could be cached to improve performance.

3. **Security consideration**: While verifying the SHA256 is good, consider using the official mdBook GitHub Action instead of manually downloading and installing.

4. **Inefficient job separation**: The build and deploy jobs are separated, but the deploy job doesn't actually use the artifact from the build job on PRs (since deploy only runs on push).

## Suggested improvements:

```yml
name: Deploy docs

on:
  push:
    branches: ["master"]
  workflow_dispatch:

permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  build-and-deploy:
    runs-on: ubuntu-24.04
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    steps:
      - uses: actions/checkout@v4
      
      - name: Setup mdBook
        uses: peaceiris/actions-mdbook@v2
        with:
          mdbook-version: '0.4.51'
      
      - name: Build with mdBook
        run: cd docs && mdbook build
      
      - name: Setup GitHub Pages
        id: pages
        uses: actions/configure-pages@v5
      
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v3
        with:
          path: ./docs/book
      
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v4
```

Or if you want to keep manual installation with caching:

```yml
name: Deploy docs

on:
  push:
    branches: ["master"]
  pull_request:
    branches: ["master"]
  workflow_dispatch:

permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  build:
    runs-on: ubuntu-24.04
    env:
      MDBOOK_VERSION: 0.4.51
      MDBOOK_SHA256: d611bcde080f1ab9932ca1724197ac2c23bb8c5d64581bb5da238bfbb4d39184
    steps:
      - uses: actions/checkout@v4
      
      - name: Cache mdBook
        id: cache-mdbook
        uses: actions/cache@v4
        with:
          path: /usr/local/bin/mdbook
          key: mdbook-${{ env.MDBOOK_VERSION }}-${{ env.MDBOOK_SHA256 }}
      
      - name: Install mdBook
        if: steps.cache-mdbook.outputs.cache-hit != 'true'
        run: |
          curl -L -o mdbook.tar.gz "https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
          echo "${MDBOOK_SHA256}  mdbook.tar.gz" | sha256sum -c -
          tar -xzf mdbook.tar.gz
          chmod +x mdbook
          sudo mv mdbook /usr/local/bin/
      
      - name: Build with mdBook
        run: cd docs && mdbook build
      
      - name: Upload build artifact
        uses: actions/upload-artifact@v4
        with:
          name: book
          path: ./docs/book
          retention-days: 1

  deploy:
    if: ${{ github.event_name == 'push' }}
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-24.04
    needs: build
    steps:
      - name: Download build artifact
        uses: actions/download-artifact@v4
        with:
          name: book
          path: ./book
      
      - name: Setup GitHub Pages
        id: pages
        uses: actions/configure-pages@v5
      
      - name: Upload Pages artifact
        uses: actions/upload-pages-artifact@v3
        with:
          path: ./book
      
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v4
```

The main improvements in the second version:
- Adds caching for mdBook binary
- Properly passes artifacts between jobs
- Keeps the separation of concerns between build and deploy
- Uses newer versions of actions where applicable

corrideat: Not sure about the caching thing. I mean, maybe, but without further analysis I can't say if the solution offered is valid or not.

The inefficient job separation part maybe so; I was keeping the existing structure. In any case, since the jobs are skipped this shouldn't matter much.

'Redundant event triggers' is nonsense, as is 'Security consideration' (ok, if there's an official mdbook action it might make sense to use it, but that's for reasons other than security AFAICT)

corrideat: Also, `peaceiris/actions-mdbook@v2` doesn't look like an _official_ mdBook action to me. It seems to be a third-party GHA (this isn't to say we shouldn't use it, which I don't know, but I'm commenting on the official part).

taoeffect: @corrideat I'm thinking of going with the 2nd CI script Opus suggested... which doesn't use `peaceiris/actions-mdbook@v2`. What do you think, do you see any issues with it?

corrideat: I'm not sure about the cache part, it uploads the build artifact for pull requests, it downloads said artifact for deployment (which I'm thinking could lead to race conditions)

corrideat (approved):

PR #2862, opened 2025-07-03
#2848 - Don't play chat notification sound if the message has already been seen

### LINKED ISSUE #2848: Don't play chat notification sound if the message has already been seen
### Problem

If you've seen a new message on Device A, and you wake up Device B, then you'll still get a new message notification on Device B. This can be annoying, especially if you're logged in on many devices.

### Solution

Before displaying the notification, check to see if the message has already been read. If so, don't display the notification.

SebinSong:
closes #2848


frontend/controller/actions/chatroom.js
```diff
@@ -13,16 +14,12 @@ import { encryptedAction, encryptedNotification } from './utils.js'
...
 
-sbp('okTurtles.events/on', MESSAGE_RECEIVE_RAW, ({
-  contractID,
-  data,
-  innerSigningContractID,
-  // If newMessage is undefined, it means that an existing message is being edited
-  newMessage
-}) => {
-  const state = sbp('chelonia/contract/state', contractID)
+const messageReceivedRawQueue = []
```
SebinSong, line 18: The reason for adding this queue here:

```
    // Without identity-kv store loaded, logics in messageReceivedRawHandler() would use wrong
    // getters.chatRoomUnreadMessages and getters.ourUnreadMessages which leads to
    // wrong computations and thus wrong behaviour.
    // (eg. 'message-received' sound for DM plays even when the user has already read them from another device)
    // So queueing the events here and then processing them after the kv-store is loaded.
```

frontend/model/state.js
```diff
@@ -88,6 +88,16 @@ const initialState = {
...
+  },
+  kvStoreStatus: {
+    // Context: Various parts of the app are closely related to kv-store. (eg. chatroom's 'readUntil' property is managed
+    //          via kv-store so it's shared across multiple devices of the same user.)
+    //          and thus sometimes need to know whether kv-store data has been loaded from the server and
+    //          'kvStoreStatus' attribute here can be used to check the loading statuses of them.
+    //
+    // { [name]: 'non-init' | 'loading' | 'loaded' }
+    identity: 'non-init',
+    group: 'non-init'
```
taoeffect: Don't forget that any time this object is updated, it will only work for new users and in development mode. It won't work for users who have previously setup accounts.

So it needs to also be created in `'state/vuex/postUpgradeVerification'`.

Sidenote: I wish there were a way to define an automated LLM rule that detected this on PRs...

SebinSong: Right. forgot about it. updated.

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -593,3 +593,17 @@ sbp('okTurtles.events/on', NEW_CHATROOM_NOTIFICATION_SETTINGS, ({ chatRoomID, se
     }
   }
 })
+
+sbp('okTurtles.events/on', KV_LOAD_STATUS, ({ name, status }) => {
```
taoeffect: Noticing the comment on line 571, would it make sense to rename this to something like `NEW_KV_LOAD_STATUS` for consistency? 🤔

frontend/controller/serviceworkers/sw-primary.js
```diff
@@ -33,7 +33,7 @@ import {
   NEW_UNREAD_MESSAGES, NOTIFICATION_EMITTED, NOTIFICATION_REMOVED,
   NOTIFICATION_STATUS_LOADED, OFFLINE, ONLINE, RECONNECTING,
   RECONNECTION_FAILED, SERIOUS_ERROR, SWITCH_GROUP
-} from '../../utils/events.js'
+} from '~/frontend/utils/events.js'
```
taoeffect, line 36: Nice catch!

frontend/model/state.js
```diff
@@ -212,6 +228,11 @@ const mutations = {
   setLastLoggedIn (state, [groupID, value]) {
     Vue.set(state.lastLoggedIn, groupID, value)
   },
+  setKvStoreStatus (state, { name, status }) {
```
taoeffect, line 232: Since we're not using any type checking here to enforce the value of this variable, can you add a runtime check to catch typos? E.g. something like:

```js
if (!['non-init', 'loading', 'loaded'].includes(status)) {
  return console.warn(`bad setKvStoreStatus for ${name}: ${status}`)
}
```

An alternative would be to define constant variables and use those instead of string literals.

frontend/model/state.js
taoeffect: @corrideat A question for you, as I don't quite remember and am wondering if you happen to remember: does it make sense to have this `kvStoreStatus` be in the Vuex state or should it be set using `'sw/deviceSettings/set'`?

taoeffect (changes requested):
Great work @SebinSong! Review ready + a question for @corrideat!

Also, here's a review from Opus 4 which seems like it may have found some issues too:

### Opus 4

Looking at these changes, I can see this is implementing a system to track KV store loading status and queue message events until the identity KV store is loaded. Here's my review:

## Bug Found

In `frontend/controller/actions/chatroom.js`, there's a problematic condition:

```javascript
if (userReadUntil?.createdHeight > 1 && userReadUntil.createdHeight >= msgData.height) {
```

The check `userReadUntil?.createdHeight > 1` seems arbitrary and could skip messages that have been read if the `createdHeight` is exactly 1. This should probably be:

```javascript
if (userReadUntil?.createdHeight && userReadUntil.createdHeight >= msgData.height) {
```

Or if there's a specific reason for `> 1`, it should be documented with a comment.

## Minor Improvement

In `frontend/controller/actions/chatroom.js`, the `msgHash` is added to queued items but never used:

```javascript
const msgHash = (newMessage || data).hash
messageReceivedRawQueue.push({ contractID, data, innerSigningContractID, newMessage, msgHash })
```

Since `msgHash` isn't used in `messageReceivedRawHandler`, you can remove it:

```javascript
messageReceivedRawQueue.push({ contractID, data, innerSigningContractID, newMessage })
```

## Potential Race Condition

The event handler for `KV_LOAD_STATUS` processes the queue when the identity store is loaded:

```javascript
sbp('okTurtles.events/on', KV_LOAD_STATUS, ({ name, status }) => {
  if (name === 'identity' && status === 'loaded') {
    while (messageReceivedRawQueue.length > 0) {
      messageReceivedRawHandler(messageReceivedRawQueue.shift())
    }
  }
})
```

If more `MESSAGE_RECEIVE_RAW` events arrive while the queue is being processed, they'll be handled immediately (since `kvStoreStatus.identity` is now 'loaded'). This should work correctly, but you might want to add a comment documenting this behavior for clarity.

Other than these issues, the implementation looks solid. The approach of queueing messages until dependencies are loaded is a good pattern for avoiding race conditions.

frontend/model/state.js
corrideat: From what I can tell, this is being done correctly regarding the state, since the state is being duplicated in Vuex (that the SW doesn't have access to) and in the Chelonia state, and the SW event that sets this is correctly being relayed to the app.

As far as using `'sw/deviceSettings/set'`, that's meant for settings that are per-device (i.e., not synced). Since this looks like it's using the KV store, `'sw/deviceSettings/set'` doesn't seem appropriate.

This is from a quick review of the changes. I'll take a more detailed look and get back to you if my answer changes.

taoeffect: OK, it sounds like @SebinSong is correctly propagating the changes to the Chelonia state to the Vuex state using the events. 👍

SebinSong: Thanks for confirmation @corrideat

SebinSong:
@taoeffect 
All the LLM suggestions above sound good (especially `potential race condition` one) so made changes accordingly.

taoeffect (approved):
Great job @SebinSong !

PR #2863, opened 2025-07-03
#2859 - Make sure 'is typing' indicator disappears when leaving the chatroom

### LINKED ISSUE #2859: 'Is typing' indicator should be cleared when switching to another chatroom
### Problem
[ steps to reproduce ]

- Create users `u1` and `u2` and create a DM between the two. Both users are online and looking at the DM.
- `u1` starts typing. 'u1 is typing' message is being displayed on `u2` side.
- But `u2` then decides to read it later and switches to another channel (eg. `#general`)
- 'u1 is typing' indicator is still there, giving a misleading information. This should be cleared as soon as `u2` leaves the DM.

https://github.com/user-attachments/assets/d103ae7d-0dea-4ad9-9e17-e15f57d5c31b

### Solution

Fix the related front-end logic.

SebinSong:
closes #2859 

https://github.com/user-attachments/assets/87343a65-3ff5-4212-849b-7625c87e72d6

### DIFF
--- a/frontend/views/containers/chatroom/SendArea.vue
+++ b/frontend/views/containers/chatroom/SendArea.vue
@@ -907,7 +907,8 @@ export default ({
     },
     emitUserTypingEvent () {
       sbp('gi.actions/chatroom/user-typing-event', {
-        contractID: this.currentChatRoomId
+        contractID: this.currentChatRoomId,
+        innerSigningContractID: this.ourIdentityContractId
       }).catch(e => {
         console.error('Error emitting user typing event', e)
       })
@@ -918,7 +919,8 @@ export default ({
     },
     emitUserStopTypingEvent () {
       sbp('gi.actions/chatroom/user-stop-typing-event', {
-        contractID: this.currentChatRoomId
+        contractID: this.currentChatRoomId,
+        innerSigningContractID: this.ourIdentityContractId
       }).catch(e => {
         console.error('Error emitting user stopped typing event', e)
       })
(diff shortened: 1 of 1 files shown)


frontend/views/containers/chatroom/SendArea.vue
```diff
@@ -902,7 +907,8 @@ export default ({
     },
     emitUserTypingEvent () {
       sbp('gi.actions/chatroom/user-typing-event', {
-        contractID: this.currentChatRoomId
+        contractID: this.currentChatRoomId,
+        innerSigningContractID: this.ourIdentityContractId
```
SebinSong, line 911: This is unrelated to the fix of the issue. (The fix works without this)
But thought it's better to explicitly specify `innerSigningContractID` field here than not:

The app has been largely enhanced since I implemented this 'is-typing' feature a couple of years ago and noticed the logics here has been updated too. Looked at this and it got me wondering "Where do we specify **who** is typing now then?". And it took me a little while to find out that it's taken care of from below part. (defaulting to the currently logged-in user)

https://github.com/okTurtles/group-income/blob/aad6b4536b6470205044bbceb8300d8eeb302180/frontend/controller/actions/utils.js#L99-L107

So I felt explicitly specifying this here would be helpful readability-wise. Especially for a future dev with fresh eyes trying to make sense of this part. So let me add this change here too.

taoeffect (approved):
Thanks @SebinSong! Appreciate the video of the fix too!

PR #2869, opened 2025-07-14
Fix size calculations

### LINKED ISSUE #2865: Fix "Invalid stored updated size" error (Heisenbug)
### Problem

After PR #2852 was merged, `master` failed: https://github.com/okTurtles/group-income/actions/runs/16056627698/job/45312550382

Image

You can see the [corresponding Cypress run here](https://cloud.cypress.io/projects/q6whky/runs/4718/test-results/48ce3038-754f-48a4-bfb6-f30026348cf3?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D).

### Solution

Figure out what caused this and fix.

corrideat:
Fix #2865 

The issue was a race condition that's best explained with an example.

Imagine there's an identity contract `alice` that owns a `group` group contract and then the `alice` contract gets deleted shortly after being created (which would also delete the `group` contract due to cascading).

Because delta updates are batched, there'll be size deltas for `alice` and `group` both for creation (positive size) and deletion (negative size).

Normally, the `group` events will be processed by adjusting the size of `alice`. However, since `group` has been deleted, the delta code will not find an owner and will set the size on `group` directly, mistaking it for an ultimate owner. Then, when `group` is deleted, the ultimate owner is set to `alice` (this information was passed on for correct attribution and was correct). Hence, there's a size discrepancy due to the initial confusion.

This PR addresses this by explicitly giving hints on contract creation to the worker about which contracts are ultimate owners, and by using this information to detect 'orphaned' contracts, i.e., deleted contracts for which we can't find an owner. In these cases, delta processing is deferred until a later execution time during which the actual ultimate owner is known (since this is given at contract deletion time).

This issue was introduced in #2852 precisely because of sending a delta to the ultimate owner, which due to the mismatch, would result in a negative size. However, I think #2852 only made this more likely to happen rather than introduce the issue, since the part where things went wrong (i.e., wrongly attributing the positive size to the wrong contract) was already present.

### DIFF
--- a/backend/ownerSizeTotalWorker.js
+++ b/backend/ownerSizeTotalWorker.js
@@ -149,10 +149,27 @@ sbp('sbp/selectors/register', {
 
     // Map to store: ultimateOwnerID -> [totalDelta, Set<originalResourceIDs>]
     const ultimateOwners = new Map()
+    const orphansSet = new Set()
     // Phase 1: Find the ultimate owner for each resource and aggregate deltas.
     await Promise.all(deltaEntries.map(async ([contractID, delta]) => {
       // --- Potentially Slow Owner Lookup Loop ---
-      const ownerID = cachedUltimateOwnerMap.get(contractID) || await lookupUltimateOwner(contractID)
+      const cachedOwnerID = cachedUltimateOwnerMap.get(contractID)
+      const ownerID = cachedOwnerID || await lookupUltimateOwner(contractID)
+      // If the delta update is for an orphaned resource, it likely was
+      // deleted and needs special handling
+      if (!cachedOwnerID && ownerID === contractID) {
+        // Verify that the resource no longer exists
+        if (!(await sbp('chelonia.db/get', contractID))) {
+          // If so, add it back to `updatedSizeMap`, so that it gets processed
+          // later, when resource deletion manually sets `cachedOwnerID`
+          const current = updatedSizeMap.get(contractID) ?? 0
+          updatedSizeMap.set(contractID, current + delta)
+          // Deleted resources can't be processed from scratch, so add them to
+          // a set to later remove them from the temporary index.
+          orphansSet.add(contractID)
+          return
+        }
+      }
       cachedUltimateOwnerMap.delete(contractID)
       // Aggregate delta for the ultimate owner.
       const [val, ownedResourcesSet] = ultimateOwners.get(ownerID) || [0, new Set(ownerID)]
@@ -168,6 +185,7 @@ sbp('sbp/selectors/register', {
       // Remove the processed resource IDs from the persistent temporary index
       await removeFromTempIndex(Array.from(contributingResources))
     }))
+    await removeFromTempIndex(Array.from(orphansSet))
 
     // Reschedule the task for the next interval
     setTimeout(sbp, TASK_TIME_INTERVAL, 'backend/server/computeSizeTaskDeltas')
--- a/backend/ownerSizeTotalWorker.js
+++ b/backend/ownerSizeTotalWorker.js
@@ -149,6 +149,12 @@ sbp('sbp/selectors/register', {
 
     // Map to store: ultimateOwnerID -> [totalDelta, Set<originalResourceIDs>]
     const ultimateOwners = new Map()
+    // 'Orphans' in this context are deleted contracts for which we can't find
+    // an owner (because the owner information was deleted when the contract was
+    // deleted). In these cases, delta processing is deferred until a later
+    // execution time during which the actual ultimate owner is known
+    // (since this is given at contract deletion time and stored in
+    // `cachedUltimateOwnerMap`). See issue #2865 and PR #2869 for details.
     const orphansSet = new Set()
     // Phase 1: Find the ultimate owner for each resource and aggregate deltas.
     await Promise.all(deltaEntries.map(async ([contractID, delta]) => {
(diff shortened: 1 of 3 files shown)


backend/ownerSizeTotalWorker.js
```diff
@@ -149,10 +149,27 @@ sbp('sbp/selectors/register', {
...
     const ultimateOwners = new Map()
+    const orphansSet = new Set()
     // Phase 1: Find the ultimate owner for each resource and aggregate deltas.
     await Promise.all(deltaEntries.map(async ([contractID, delta]) => {
       // --- Potentially Slow Owner Lookup Loop ---
-      const ownerID = cachedUltimateOwnerMap.get(contractID) || await lookupUltimateOwner(contractID)
+      const cachedOwnerID = cachedUltimateOwnerMap.get(contractID)
+      const ownerID = cachedOwnerID || await lookupUltimateOwner(contractID)
+      // If the delta update is for an orphaned resource, it likely was
+      // deleted and needs special handling
```
taoeffect, line 165: Could you explain in the comment exactly what an "orphaned resource" is and how it was generated?

taoeffect: You added a fantastic explanation on the PR description just now, could you please add a link here to the comments in the code to that description?

taoeffect (changes requested):
Nice work @corrideat!

I left one RFC for improving a comment, and here's some other feedback from LLMs. The Opus one seems legit, am not 100% sure about Grok 4 but maybe it's also reasonable? Your call:

## Opus 4

#### Potential Issue with Set Initialization

```js
const [val, ownedResourcesSet] = ultimateOwners.get(ownerID) || [0, new Set(ownerID)]
```

This line has a bug. When creating a new Set, you're passing `ownerID` as the constructor argument, but `Set` expects an iterable. Since `ownerID` is a string, this will create a Set containing each character of the string as separate elements.

**Fix:**
```js
const [val, ownedResourcesSet] = ultimateOwners.get(ownerID) || [0, new Set()]
```

The `ownerID` gets added to the set on the next line anyway via `ownedResourcesSet.add(contractID)`.

## Grok 4

<details>

#### Problematic Lines
The issue stems from clearing the in-memory `updatedSizeMap` on each task run, combined with removing orphans from the persistent temporary index, which can cause pending deltas to be lost on restart. Specifically:

```js
// In 'backend/server/computeSizeTaskDeltas':
updatedSizeMap.clear()

// Later, in the map function for deltaEntries:
if (!cachedOwnerID && ownerID === contractID) {
  if (!(await sbp('chelonia.db/get', contractID))) {
    const current = updatedSizeMap.get(contractID) ?? 0
    updatedSizeMap.set(contractID, current + delta)
    orphansSet.add(contractID)
    return
  }
}

// At the end:
await removeFromTempIndex(Array.from(orphansSet))
```

**Explanation of the bug:**
- When an orphan is detected, the delta is added *back* to `updatedSizeMap` (which was cleared earlier in the function) to defer it for the next cycle.
- However, the orphan (contractID) is then removed from the persistent temporary index via `orphansSet`.
- If the server restarts *after* this detection but *before* the next task cycle (where the deferred delta would be processed using the cache set by deletion), the in-memory `updatedSizeMap` is lost.
- On restart, the temporary index no longer contains the contractID (due to the removal), so nothing is loaded into `updatedSizeList` for full recalculation. The pending positive delta (e.g., from creation) is effectively dropped.
- Meanwhile, the negative delta (from deletion) may still be processed in a future cycle, leading to an unbalanced subtraction from the ultimate owner's total size.
- This violates the net-zero expectation for quick create-delete cycles, as described in the issue context.

The comment justifies the removal ("Deleted resources can't be processed from scratch"), which avoids problematic full recalcs on deleted contracts in `computeSizeTask`. However, it introduces this restart vulnerability.

#### Suggested Improvement
To make the system more robust to restarts without losing deltas, avoid removing orphans from the temporary index. Instead, keep them persisted, and add a check in `computeSizeTask` to skip full recalculation for deleted contracts (treating them as net-zero or ignoring them, since their data is gone). This way, on restart, the CID remains in `updatedSizeList` but is skipped, effectively dropping the positive delta—but pair it with a mechanism to also drop or balance the negative delta (e.g., by checking for deletion in `updateSizeSideEffects` and skipping if already deferred). However, to keep it simple, the skip alone reduces risk, assuming the negative delta addition happens post-restart and can be handled in deltas.

For a fuller fix, consider persisting the ultimate owner (set during deletion) in the DB for orphans, then use it in both delta processing and full recalc to attribute a 0-size correctly. But this adds complexity.

Here's the modified code for the simple improvement (don't remove orphans, skip deleted in full recalc):

```js
// In 'backend/server/computeSizeTaskDeltas':
// Remove this line at the end (don't remove orphans from temp index):
// await removeFromTempIndex(Array.from(orphansSet))

// In 'backend/server/computeSizeTask', add a check after looking up ownerID:
await Promise.all(resourcesToRecalculate.map(async (contractID) => {
  const ownerID = await lookupUltimateOwner(contractID)
  // Add this skip for deleted orphans to avoid bogus full recalc (e.g., setting to 0)
  if (ownerID === contractID && !(await sbp('chelonia.db/get', contractID))) {
    // Optionally log: console.warn(`[ownerSizeTotalWorker] Skipping full recalc for deleted orphan ${contractID}`)
    return
  }
  // Group resources by their ultimate owner.
  const resources = ultimateOwners.get(ownerID)
  if (resources) {
    resources.add(contractID)
  } else {
    ultimateOwners.set(ownerID, new Set([contractID]))
  }
}))
```

This change ensures the pending CID persists across restarts but avoids incorrect full recalcs on deleted items. Note that in rare restart cases, the net size might still be off (e.g., only negative delta processed), but it's better than losing persistence entirely. If the task interval is short (e.g., seconds), the window is minimal.

</details>

taoeffect (approved):
Approved, great work @corrideat !

PR #2873, opened 2025-07-16
Fix #2872 and another Vue validation error

### LINKED ISSUE #2872: Channel switching bug when there's multiple groups
### Problem

I'm part of 3 groups, and all of a sudden on Brave I cannot switch to regular chatrooms on the 2 new groups that I'm in. Only the main group works for switching between chatroom channels, the others only support switching to DMs.

Notice that even when I set the chatroom to be the `#general` chatroom and switch to the Dashboard and then click on Chat, it takes me to the last DM I had selected instead:

https://github.com/user-attachments/assets/b49c9810-4611-4e34-aceb-d54090afd9ab

After logging in in Firefox, changing chatrooms works fine there, but something might happen to make it break again.

I should also mention that one of the users in Group 3 for some reason can no longer log in to Group 3 with their username & password, and I had to send them the invite so that they could login again under a new user. I don't know if this has anything to do with it.

Also: when they created a new chatroom in Group 3, I still couldn't switch to it in Brave (only in the recently logged in Firefox).  It used to work but all of a sudden it doesn't anymore in Brave.

### Solution

No idea.

taoeffect:
Closes #2872

### DIFF
--- a/frontend/model/chatroom/getters.js
+++ b/frontend/model/chatroom/getters.js
@@ -163,7 +163,7 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
   },
   groupIdFromChatRoomId (state, getters, rootState) {
     return (chatRoomID: string) => Object.keys(rootState.contracts)
-      .find(cId => rootState.contracts[cId].type === 'gi.contracts/group' &&
+      .find(cId => rootState.contracts[cId]?.type === 'gi.contracts/group' &&
         Object.keys(rootState[cId].chatRooms).includes(chatRoomID))
   },
   chatRoomsInDetail (state, getters, rootState) {
--- a/frontend/views/containers/group-settings/GroupLeaveModal.vue
+++ b/frontend/views/containers/group-settings/GroupLeaveModal.vue
@@ -126,7 +126,7 @@ export default ({
       confirmation: {
         [L('This field is required')]: required,
         [L('Does not match')]: function (value) {
-          return normalizeString(value) === normalizeString(this.code)
+          return normalizeString(value || '') === normalizeString(this.code)
         }
       }
     }


frontend/views/containers/group-settings/GroupLeaveModal.vue
```diff
@@ -126,7 +126,7 @@ export default ({
       confirmation: {
         [L('This field is required')]: required,
         [L('Does not match')]: function (value) {
-          return normalizeString(value) === normalizeString(this.code)
+          return normalizeString(value || '') === normalizeString(this.code)
```
corrideat, line 129: Can this possibly not be a string? Even then, maybe `??` is better than `||`, since it avoids having a fallback for `''` (i.e., if value is `''`. This is a minor revision though.

frontend/model/chatroom/getters.js
```diff
@@ -163,7 +163,7 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
   },
   groupIdFromChatRoomId (state, getters, rootState) {
     return (chatRoomID: string) => Object.keys(rootState.contracts)
-      .find(cId => rootState.contracts[cId].type === 'gi.contracts/group' &&
+      .find(cId => rootState.contracts[cId]?.type === 'gi.contracts/group' &&
```
corrideat, line 166: This looks like a needed fix.

frontend/views/containers/group-settings/GroupLeaveModal.vue
taoeffect: Yeah I saw an error on blur once

corrideat:
Have you been able to verify that this fixes the issue (by testing it doesn't reproduce with the DB)?

corrideat (approved):
LGTM

taoeffect:
> Have you been able to verify that this fixes the issue (by testing it doesn't reproduce with the DB)?

Well I reproduced the issue locally and verified this fixes it. I can re-open the issue if it doesn't fix in next release.

PR #2874, opened 2025-07-19
Closes #2868

### LINKED ISSUE #2868: Two display issues related to user leaving group
### Problem

Problem 1:

Sometimes, when someone leaves the group, their contractID is displayed instead of their username:

Image

NOTE: this only happens sometimes. On one device I see their username, on another their contractID.

Problem 2:

If you had a DM with this person, you will see your username appear instead a blank image and their username. Notice here how you see my name twice, that is because one is a DM with myself, and one is a DM with the person that left:

Image

### Solution

Problem 1: Show their username always.

Problem 2: Since their profile picture might not be cached, display a blank profile picture along with their username.

corrideat:
Closes #2868

* Upgrade verification code to sync namespace lookups and reverse namespace lookups
* Fix mention code so that any known username works
* Show DMs with former group members
* Fix missing partner DM info with deleted users

### DIFF
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -1,10 +1,29 @@
 'use strict'
 import sbp from '@sbp/sbp'
 import { KV_KEYS, KV_LOAD_STATUS } from '~/frontend/utils/constants.js'
-import { KV_QUEUE, NEW_PREFERENCES, NEW_UNREAD_MESSAGES, ONLINE, NEW_KV_LOAD_STATUS } from '~/frontend/utils/events.js'
+import { debounce, difference, intersection, union } from 'turtledash'
+import { KV_QUEUE, NAMESPACE_REGISTRATION, NEW_PREFERENCES, NEW_UNREAD_MESSAGES, ONLINE, NEW_KV_LOAD_STATUS } from '~/frontend/utils/events.js'
 import { isExpired } from '@model/notifications/utils.js'
 
 const initNotificationStatus = (data = {}) => ({ ...data, read: false })
+// Name discrepancies between the KV store and `namespaceLookups` may occur
+// due to being unsubcribed from an identity contract (e.g., someone has left
+// a group) or due to the username being deleted. This function attempts to
+// determine which case it is, and determine all of the names that are currently
+// valid.
+const checkAndAugmentNames = async (currentNames: string[]) => {
+  const ourNames = Object.keys(sbp('state/vuex/state').namespaceLookups)
+  const unconflictedNames = intersection(currentNames, ourNames)
+  const recheckedNames = await Promise.all(difference(union(currentNames, ourNames), unconflictedNames).map(async (name) => {
+    const value = await sbp('namespace/lookup', name, { skipCache: true })
+
+    return value ? name : null
+  })).then((names) => {
+    return names.filter((v) => !!v)
+  })
+
+  return union(unconflictedNames, recheckedNames)
+}
 
 sbp('okTurtles.events/on', ONLINE, () => {
   if (!sbp('state/vuex/state').loggedIn?.identityContractID) {
@@ -23,6 +42,7 @@ export default (sbp('sbp/selectors/register', {
     await sbp('gi.actions/identity/kv/loadChatRoomUnreadMessages')
     await sbp('gi.actions/identity/kv/loadPreferences')
     await sbp('gi.actions/identity/kv/loadNotificationStatus')
+    await sbp('gi.actions/identity/kv/loadCachedNames')
 
     console.info('identity key-value store data loaded!')
     sbp('okTurtles.events/emit', NEW_KV_LOAD_STATUS, { name: 'identity', status: KV_LOAD_STATUS.LOADED })
@@ -316,5 +336,55 @@ export default (sbp('sbp/selectors/register', {
 
       await sbp('gi.actions/identity/kv/saveNotificationStatus', { onconflict: getUpdatedNotificationStatus })
     })
+  },
+  // Namespace lookups
+  'gi.actions/identity/kv/fetchCachedNames': async () => {
+    const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
+    if (!identityContractID) {
+      throw new Error('Unable to fetch cached names without an active session')
+    }
+    return (await sbp('chelonia/kv/get', identityContractID, KV_KEYS.NS_CACHE))?.data || []
+  },
+  'gi.actions/identity/kv/saveCachedNames': () => {
+    const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
+    if (!identityContractID) {
+      throw new Error('Unable to update notification status without an active session')
+    }
+
+    const onconflict = async ({ currentData = [], etag } = {}) => {
+      const data = await checkAndAugmentNames(currentData)
+
+      data.sort()
+      currentData.sort()
+
+      // If there's no difference, there's no point in sending an update
+      if (data.length === currentData.length) {
+        let i = 0
+        for (; i < data.length; i++) {
+          if (data[i] !== currentData[i]) break
+        }
+        if (i !== data.length) return
+      }
+
+      return [data, etag]
+    }
+
+    return sbp('chelonia/kv/queuedSet', {
+      contractID: identityContractID,
+      key: KV_KEYS.NS_CACHE,
+      data: null,
+      onconflict
+    })
+  },
+  'gi.actions/identity/kv/loadCachedNames': () => {
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
+      const currentData = await sbp('gi.actions/identity/kv/fetchCachedNames')
+
+      await checkAndAugmentNames(currentData)
+    })
   }
 }): string[])
+
+// Debounced so that `checkAndAugmentNames` (which may affect the names
+// being stored) doesn't result in too many calls to saveCachedNames.
+sbp('okTurtles.events/on', NAMESPACE_REGISTRATION, debounce(() => sbp('gi.actions/identity/kv/saveCachedNames'), 300))
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -12,7 +12,7 @@ const initNotificationStatus = (data = {}) => ({ ...data, read: false })
 // determine which case it is, and determine all of the names that are currently
 // valid.
 const checkAndAugmentNames = async (currentNames: string[]) => {
-  const ourNames = Object.keys(sbp('state/vuex/state').namespaceLookups)
+  const ourNames = Object.keys(sbp('state/vuex/state').namespaceLookups || {})
   const unconflictedNames = intersection(currentNames, ourNames)
   const recheckedNames = await Promise.all(difference(union(currentNames, ourNames), unconflictedNames).map(async (name) => {
     const value = await sbp('namespace/lookup', name, { skipCache: true })
@@ -352,6 +352,7 @@ export default (sbp('sbp/selectors/register', {
     }
 
     const onconflict = async ({ currentData = [], etag } = {}) => {
+      if (!currentData) currentData = []
       const data = await checkAndAugmentNames(currentData)
 
       data.sort()
@@ -372,7 +373,7 @@ export default (sbp('sbp/selectors/register', {
     return sbp('chelonia/kv/queuedSet', {
       contractID: identityContractID,
       key: KV_KEYS.NS_CACHE,
-      data: null,
+      data: Object.keys(sbp('state/vuex/state').namespaceLookups || {}).sort(),
       onconflict
     })
   },
@@ -380,7 +381,7 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const currentData = await sbp('gi.actions/identity/kv/fetchCachedNames')
 
-      await checkAndAugmentNames(currentData)
+      await checkAndAugmentNames(currentData || [])
     })
   }
 }): string[])
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -14,13 +14,19 @@ const initNotificationStatus = (data = {}) => ({ ...data, read: false })
 const checkAndAugmentNames = async (currentNames: string[]) => {
   const ourNames = Object.keys(sbp('state/vuex/state').namespaceLookups || {})
   const unconflictedNames = intersection(currentNames, ourNames)
-  const recheckedNames = await Promise.all(difference(union(currentNames, ourNames), unconflictedNames).map(async (name) => {
-    const value = await sbp('namespace/lookup', name, { skipCache: true })
+  // Batch the lookups to avoid too many concurrent requests
+  const BATCH_SIZE = 10
+  const namesToCheck = difference(union(currentNames, ourNames), unconflictedNames)
+  const recheckedNames = []
 
-    return value ? name : null
-  })).then((names) => {
-    return names.filter((v) => !!v)
-  })
+  for (let i = 0; i < namesToCheck.length; i += BATCH_SIZE) {
+    const batch = namesToCheck.slice(i, i + BATCH_SIZE)
+    const results = await Promise.all(batch.map(async (name) => {
+      const value = await sbp('namespace/lookup', name, { skipCache: true })
+      return value ? name : null
+    }))
+    recheckedNames.push(...results.filter(v => !!v))
+  }
 
   return union(unconflictedNames, recheckedNames)
 }
@@ -364,7 +370,8 @@ export default (sbp('sbp/selectors/register', {
         for (; i < data.length; i++) {
           if (data[i] !== currentData[i]) break
         }
-        if (i !== data.length) return
+        // If `i` equals `data.length`, the loop has ended and all items matched
+        if (i === data.length) return
       }
 
       return [data, etag]
@@ -381,6 +388,8 @@ export default (sbp('sbp/selectors/register', {
     return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
       const currentData = await sbp('gi.actions/identity/kv/fetchCachedNames')
 
+      // `checkAndAugmentNames` will handle updating the namespace cache as
+      // necessary. The return value isn't needed.
       await checkAndAugmentNames(currentData || [])
     })
   }
--- a/frontend/controller/actions/identity-kv.js
+++ b/frontend/controller/actions/identity-kv.js
@@ -22,7 +22,9 @@ const checkAndAugmentNames = async (currentNames: string[]) => {
   for (let i = 0; i < namesToCheck.length; i += BATCH_SIZE) {
     const batch = namesToCheck.slice(i, i + BATCH_SIZE)
     const results = await Promise.all(batch.map(async (name) => {
-      const value = await sbp('namespace/lookup', name, { skipCache: true })
+      const value = await sbp('namespace/lookup', name, { skipCache: true }).catch(e => {
+        console.warn(`[checkAndAugmentNames] Failed to lookup name ${name}:`, e)
+      })
       return value ? name : null
     }))
     recheckedNames.push(...results.filter(v => !!v))
@@ -354,7 +356,7 @@ export default (sbp('sbp/selectors/register', {
   'gi.actions/identity/kv/saveCachedNames': () => {
     const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
     if (!identityContractID) {
-      throw new Error('Unable to update notification status without an active session')
+      throw new Error('Unable to update cached names without an active session')
     }
 
     const onconflict = async ({ currentData = [], etag } = {}) => {
--- a/frontend/controller/app/chatroom.js
+++ b/frontend/controller/app/chatroom.js
@@ -26,8 +26,8 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
     // TODO: Figure out a better way of doing this that doesn't require a timeout
     const setCurrentChatRoomId = () => {
       // Re-grab the state as it could be a stale reference
-      const rootState = sbp('state/vuex/state')
-      if (!rootState[chatRoomID]?.members?.[identityContractID]) {
+      const contractState = sbp('state/vuex/state')[chatRoomID]
+      if (!contractState?.members?.[identityContractID] || contractState.members[identityContractID].hasLeft) {
         if (++attemptCount > 5) {
           console.warn('[JOINED_CHATROOM] Given up on setCurrentChatRoomId after 5 attempts', { identityContractID, groupContractID, chatRoomID })
           return
--- a/frontend/controller/app/chatroom.js
+++ b/frontend/controller/app/chatroom.js
@@ -26,8 +26,8 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
     // TODO: Figure out a better way of doing this that doesn't require a timeout
     const setCurrentChatRoomId = () => {
       // Re-grab the state as it could be a stale reference
-      const contractState = sbp('state/vuex/state')[chatRoomID]
-      if (!contractState?.members?.[identityContractID] || contractState.members[identityContractID].hasLeft) {
+      const rootGetters = sbp('state/vuex/getters')
+      if (!rootGetters.isJoinedChatRoom(chatRoomID, identityContractID)) {
         if (++attemptCount > 5) {
           console.warn('[JOINED_CHATROOM] Given up on setCurrentChatRoomId after 5 attempts', { identityContractID, groupContractID, chatRoomID })
           return
--- a/frontend/model/chatroom/getters.js
+++ b/frontend/model/chatroom/getters.js
@@ -46,6 +46,8 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
         // NOTE: direct messages should be filtered to the ones which are visible and of active group members
         const members = Object.keys(chatRoomState.members)
         const isDMToMyself = members.length === 1 && members[0] === myIdentityId
+        // Explicitly don't filter out on `hasLeft` attribute, so that DMs can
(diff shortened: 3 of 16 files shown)


taoeffect:
Copying form [here](https://github.com/okTurtles/group-income/issues/2868#issuecomment-3094716964):

> There are two situations when someone leaves a DM: either "leaving the DM" (no UI for this, unimplemented) or "deleting their identity contract" (currently implemented), and a third it seems (leaving a group — though this shouldn't cause the DM to be left, it should cause the DM to be hidden in the UI for the Group).
> 
> When someone "leaves a DM" for whatever reason, we should probably just save their username to the contract for display purposes, and then log an event to the DM itself that "\<person\> left the DM", especially as it could be a DM with multiple people.
> 
> For leaving the group, the DM should remain unmodified, as DMs persist, but since the DM is with someone no longer in the group, that DM — if and only if it contains no other group members other than yourself — should be hidden in the group DM UI, but the DM contract should remain (for future display purposes in the global DM dashboard.

EDIT: copying comment below:

> So, here's my updated thinking: when someone leaves a DM, yes, their icon should be removed from the group DM bubble, but, in order to distinguish it from a DM with yourself, there should be some other way of visualizing the fact that it is/was a group DM.
> 
> So for example, a group DM bubble can have extra grey padding around the profile icons, say 3px extra padding, with a thin 1px-wide slightly darker shade of the padding color stroke on the edge of the ring. This will indicate to the user that this is in fact a group DM, and this padding/ring should appear around all group DMs.

corrideat:
Just for clarification: leaving a group AFAICT doesn't result in any issues (except that DMs may end up being inaccessible in the UI; fixed in this PR). It's only when an account is deleted that the issue of the DM showing up as a DM to yourself happens.

This also makes me wonder of this other situation we haven't discussed:

Imagine you have a DM with alice and bob. alice removes their account. Should the DM still show up in the UI as a DM with alice & bob, or just with bob? (I'm guessing the answer is alice & bob)

taoeffect:
> It's only when an account is deleted that the issue of the DM showing up as a DM to yourself happens.

AFAICT that's not true. In the example in the issue, Sebin's `sebin` account is still alive and well.

corrideat:
Also for clarification:

The events of people joining / leaving a DM _are_ implemented in the contract, just hidden in the UI. However, we don't retain this information in the contract state.

taoeffect:
> Imagine you have a DM with alice and bob. alice removes their account. Should the DM still show up in the UI as a DM with alice & bob, or just with bob? (I'm guessing the answer is alice & bob)

~~It should appear as a DM with Alice & Bob, but alice's profile icon should be missing (probably a grey circle)~~ and there should be a system message in the DM itself saying `alice left the DM`

EDIT: Err, no I think that's wrong actually, let me think on this more.

corrideat:
I was only able to reproduce this deleting an account, and that contract ID returns 410 on the server when I checked.

taoeffect:
> I was only able to reproduce this deleting an account, and that contract ID returns 410 on the server when I checked.

Oh, it seems you're right! I think I was misreading the DB.

taoeffect:
Slack actually doesn't let you leave DMs.

So I'm OK with doing the same thing, keeping track of all of the people who were added to the DM, and making that appear in the list, and if any of them deletes their account, then we just have a historical entry for their username, and their profile picture is just a grey circle (maybe with the first letter of their username inside of it, same as we do it with groups). OTOH if it's easier to save their profile picture, we can also do that.

taoeffect:
@corrideat had a brilliant idea for how to handle identities being deleted resulting in contractIDs appearing in the UI on new devices that are logging in: move the `namespaceLookups` and `reverseNamespaceLookups` keys to the identity contract KV store!

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
```diff
@@ -72,7 +71,7 @@ export default ({
     },
     possibleMentions () {
       return [
-        ...Object.keys(this.ourContactProfilesById).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
+        ...Object.keys(this.$store.state.reverseNamespaceLookups).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
```
taoeffect, line 79: Could you add a comment here explaining the reasoning for the choice to use `reverseNamespaceLookups` instead of something like `ourContactProfilesById` just so that we don't forget why it was done?

corrideat:
It looks like the following additional changes are needed, based on what've discussed:

1. Modify chat rendering logic _not_ to ignore system notifications about someone leaving a DM --- this will address the issue of someone being unaware of chatroom activity
2. Modify chatroom contract to preserve historical / former members (instead of deleting them from `.members`. We've said that compatibility doesn't matter (\*), so the changes here will be breaking. ~~(The more backwards-compatible alternative is to keep the existing behaviour on `.members`, but add removed members to a different key, like `.formerMembers`)~~; then, modify (if needed) the display logic to include all members in the chat name --- this will help address the display issue of former DM partners being missing from the DM name.
3. Store the namespace lookup cache into the KV store --- this will address most situations that result in a contract ID being displayed instead of a username (\*\*).

(\*) Actually, it seems this was a misunderstanding, and changes should _not_ break existing chatrooms. I think that if we modified the contract so that _going forward_ it could contain former members in `.members`, this shouldn't be a breaking change for existing chatrooms since the only chatrooms with extra member entries would be those that are using a newer version.

(\*\*) I originally thought that conflicts could be handled by a simple union operation (i.e., adding any missing members). However, the namespace lookup cache isn't strictly append-only, since entries can be deleted in certain cases if the server admin also deletes the namespace lookup. I'm not entirely sure yet how exactly to address this, but it seems like most likely we'll need to verify mismatched entries to ensure they still exist on the server.

frontend/model/chatroom/getters.js
```diff
@@ -53,7 +53,12 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
             const p2JoinedDate = new Date(chatRoomState.members[p2].joinedDate).getTime()
             return p1JoinedDate - p2JoinedDate
           })
-        const hasActiveMember = partners.some(memberID => Object.keys(getters.profilesByGroup(groupID)).includes(memberID))
+        // The following line ensured that DMs for former members were hidden.
+        // Until the DM global dashboard is implemented, this check has been
+        // replaced with a check for all members (past and present), so that
+        // a conversation doesn't suddenly become inaccessible.
+        // // const hasActiveMember = partners.some(memberID => Object.keys(getters.profilesByGroup(groupID)).includes(memberID))
+        const hasActiveMember = partners.some(memberID => !!rootState[groupID]?.profiles[memberID])
```
taoeffect, line 63: Seems fine

frontend/model/contracts/chatroom.js
```diff
@@ -266,7 +262,7 @@ sbp('chelonia/defineContract', {
         sbp('gi.contracts/chatroom/referenceTally', contractID, memberID, 'release')
         sbp('chelonia/queueInvocation', contractID, async () => {
           const state = await sbp('chelonia/contract/state', contractID)
-          if (!state || !!state.members?.[data.memberID] || !state.attributes) {
+          if (!state || !!state.members?.[data.memberID] || state.members[data.memberID].hasLeft || !state.attributes) {
```
SebinSong: Would you think `!!state.members?.[data.memberID]` in `if (!state || !!state.members?.[data.memberID] || state.members[data.memberID].hasLeft || !state.attributes)` has a typo? : Should have one `!` instead of two?

Cuz, it looks kinda confusing in that it checks either if the leaving person is a member of the chatroom or this leaving person has already left the group. This who conditional has two conditions that are visibly conflicting.
(Never mind if this is correct as is though)

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
```diff
@@ -72,7 +71,10 @@ export default ({
     },
     possibleMentions () {
       return [
-        ...Object.keys(this.ourContactProfilesById).map(u => makeMentionFromUserID(u).me).filter(v => !!v),
+        // Use `reverseNamespaceLookups` instead of `ourContactProfilesById`
+        // so that all known usernames can be treated as a mention, even those
+        // that correspond to contracts we're no longer subscribed to.
```
SebinSong: Would you think replacing `even those that correspond to contracts we're no longer subscribed to` -> `even those that correspond to members that have left the chatroom` is a good idea?

(BTW, I remember `reverseNamespaceLookups` was useful in finding the user name of a person who no longer is a member of the group. Wish this was commented somewhere for any future devs to know too.)

SebinSong (commented):
This PR looks good @corrideat . Just left a couple of comments.

frontend/views/containers/chatroom/chat-mentions/RenderMessageText.vue
corrideat: I see your point, but I think it's more accurate as is.

`reverseNamespaceLookups` (or `namespaceLookups`) is generally append-only. If someone, say, leaves a group, their entry will not be necessarily removed from `reverseNamespaceLookups`.

However, if you then log in from a new device, that contract will never be synced (assuming no other groups, DMs, etc.), and thus `reverseNamespaceLookups` will not contain an entry for it.

Maybe I can improve the comment to be clearer about this difference and how it relates to leaving a group.

frontend/model/contracts/chatroom.js
corrideat: Actually, this line was wrong after my changes that you've commented out, but the `!!` is AFAICT correct.

We want not to execute what follows if the contract has been removed or if the member has rejoined.

The original condition was thus:

`!state || !!state.members?.[data.memberID] || !state.attributes`

!state: State could be undefined if the contract has already been removed
!!state.members?.[data.memberID]: The member has rejoined
!state.attributes: I don't remember the exact specific significance of `attributes` here, but it's quite likely related to the constructor not having been executed (e.g., for private group chatrooms, where members are visible to all group members but not the rest of the contract).

frontend/model/state.js
```diff
@@ -179,6 +179,23 @@ sbp('sbp/selectors/register', {
...
+        .filter(([, value]) => !state.reverseNamespaceLookups[value])
+        .forEach(([name, value]) => {
+          state.reverseNamespaceLookups[value] = name
+        })
+      Object.entries(state.reverseNamespaceLookups)
+        .filter(([, name]) => !state.namespaceLookups[name])
+        .forEach(([value, name]) => {
+          state.namespaceLookups[name] = value
+        })
+    })()
```
taoeffect, line 198: Is there a reason for wrapping this in `;(() => ...`? Seems like it doesn't need it

corrideat: Yes, the reason is because it's part of the same logic and this makes it IMHO easier to identify it and remove or refactor it later.

frontend/model/contracts/chatroom.js
```diff
@@ -224,18 +224,15 @@ sbp('chelonia/defineContract', {
...
             throw new Error('Missing members state')
-          } else if (!state.members[memberID]) {
+          } else if (!state.members[memberID] || state.members[memberID].hasLeft) {
             throw new GIChatroomNotMemberError(`Can not leave the chatroom ${contractID} which ${memberID} is not part of`)
           }
         }
 
-        delete state.members[memberID]
+        state.members[memberID].hasLeft = true
+        // delete state.members[memberID]
```
taoeffect, line 233: Hmm... Opus asked a good question here, which is whether any migration code is needed for members that were previously deleted? Or is it too late at this point?

frontend/controller/actions/identity-kv.js
```diff
@@ -1,10 +1,29 @@
...
+  const recheckedNames = await Promise.all(difference(union(currentNames, ourNames), unconflictedNames).map(async (name) => {
+    const value = await sbp('namespace/lookup', name, { skipCache: true })
+
+    return value ? name : null
+  })).then((names) => {
+    return names.filter((v) => !!v)
+  })
+
+  return union(unconflictedNames, recheckedNames)
+}
```
taoeffect: Opus has a concern:

Each lookup makes an HTTP request to the server. If there are many conflicting names, this could result in many concurrent requests. Consider:

```javascript
// Batch the lookups to avoid too many concurrent requests
const BATCH_SIZE = 10
const namesToCheck = difference(union(currentNames, ourNames), unconflictedNames)
const recheckedNames = []

for (let i = 0; i < namesToCheck.length; i += BATCH_SIZE) {
  const batch = namesToCheck.slice(i, i + BATCH_SIZE)
  const results = await Promise.all(batch.map(async (name) => {
    const value = await sbp('namespace/lookup', name, { skipCache: true })
    return value ? name : null
  }))
  recheckedNames.push(...results.filter(v => !!v))
}
```

frontend/model/contracts/chatroom.js
corrideat: It's too late at this point, unless we re-sync contracts. However, we do keep track of historical group members, so we could do that for groups (just not for DMs).

corrideat: An unrelated note about this code here: for compatibility, tests (and the app in general) should work both with the current change and with 233 (the old way) _not_ commented out. In other words, we want things to still work if members are removed from the state (as it was) without setting `hasLeft`.

frontend/controller/actions/identity-kv.js
corrideat: That's a valid concern, but not so sure about the batching. At a first glance, it seems like a good idea.

frontend/controller/app/chatroom.js
```diff
@@ -26,8 +26,8 @@ sbp('okTurtles.events/on', JOINED_CHATROOM, ({ identityContractID, groupContract
     // TODO: Figure out a better way of doing this that doesn't require a timeout
     const setCurrentChatRoomId = () => {
       // Re-grab the state as it could be a stale reference
-      const rootState = sbp('state/vuex/state')
-      if (!rootState[chatRoomID]?.members?.[identityContractID]) {
+      const contractState = sbp('state/vuex/state')[chatRoomID]
+      if (!contractState?.members?.[identityContractID] || contractState.members[identityContractID].hasLeft) {
```
taoeffect: It's unfortunate that we find ourselves in a position where this complex check must be repeated in a billion places... Is there any possible solution to this? I suppose we can live with it but am wondering if you have any thoughts on this.

corrideat: I share the concern. We could create a utility function and call that, maybe? Or define a getter.

taoeffect: Yeah, either of those 👍 

The getter seems like it would make more sense

frontend/controller/actions/identity-kv.js
```diff
@@ -316,5 +336,56 @@ export default (sbp('sbp/selectors/register', {
...
+      data.sort()
+      currentData.sort()
+
+      // If there's no difference, there's no point in sending an update
+      if (data.length === currentData.length) {
+        let i = 0
+        for (; i < data.length; i++) {
+          if (data[i] !== currentData[i]) break
+        }
+        if (i !== data.length) return
```
taoeffect: I'm not quite understanding the logic here. This is saying that if there's a difference between `data` and `currentData` that we return from this `onconflict` function? If so, that means no update is made because there _is_ a difference? Isn't that backwards?

Or am I misunderstanding? Either way, can you add a comment here explaining?

frontend/controller/actions/identity-kv.js
```diff
@@ -316,5 +336,56 @@ export default (sbp('sbp/selectors/register', {
...
+      key: KV_KEYS.NS_CACHE,
+      data: Object.keys(sbp('state/vuex/state').namespaceLookups || {}).sort(),
+      onconflict
+    })
+  },
+  'gi.actions/identity/kv/loadCachedNames': () => {
+    return sbp('okTurtles.eventQueue/queueEvent', KV_QUEUE, async () => {
+      const currentData = await sbp('gi.actions/identity/kv/fetchCachedNames')
+
+      await checkAndAugmentNames(currentData || [])
```
taoeffect: What exactly is this `checkAndAugmentNames` doing here? Its return value isn't being used, so why is it here?

taoeffect (changes requested):
Great work on this PR @corrideat! Looking really solid, just a few minor change requests + comments & questions.

frontend/controller/actions/identity-kv.js
corrideat: This is returning if there's no difference. For that to happen, the length of data and currentData must be the same and the `for` loop must not be exited early.

frontend/controller/actions/identity-kv.js
corrideat: Because `checkAndAugmentNames` will cause namespace lookups, which will update the cached name entries. The return value doesn't need to be used for loading.

frontend/controller/actions/identity-kv.js
taoeffect: Then shouldn't that be `===` instead of `!==`?

frontend/controller/actions/identity-kv.js
taoeffect: Could you add a comment to that effect?

frontend/controller/actions/identity-kv.js
corrideat: Oh, I see your point. Yeah, I think so.

frontend/model/contracts/chatroom.js
```diff
@@ -492,7 +490,7 @@ sbp('chelonia/defineContract', {
         // after all side effects from joins and rejoins have been processed.
         sbp('chelonia/queueInvocation', contractID, async () => {
           const state = await sbp('chelonia/contract/state', contractID)
-          if (!state?.members?.[me]) {
+          if (!getters.isJoinedChatRoomForChatRoom(state)) {
```
taoeffect: So here and on line 414 you don't pass in `me`, but on line 363 you do. Could you pass in `me` here and line 414 too for consistency's sake?

frontend/model/chatroom/getters.js
```diff
@@ -171,7 +185,7 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
     const myIdentityId = rootState.loggedIn.identityContractID
     for (const contractID in chatRoomsInDetail) {
       const chatRoom = rootState[contractID]
-      if (chatRoom && chatRoom.attributes && chatRoom.members[myIdentityId]) {
+      if (getters.isJoinedChatRoom(contractID, myIdentityId)) {
```
taoeffect: The logic has slightly changed here. Before it was checking for the existence of `chatRoom` and `chatRoom.attributes`, and if they didn't exist it would take the `else` branch. Now, the new logic might lead to different behavior, although I'm not sure. Can you double-check?

frontend/model/chatroom/getters.js
```diff
@@ -128,7 +135,14 @@ const getters: { [x: string]: (state: Object, getters: { [x: string]: any }, roo
...
   isJoinedChatRoom (state, getters, rootState) {
-    return (chatRoomID: string, memberID?: string) => !!rootState[chatRoomID]?.members?.[memberID || getters.ourIdentityContractId]
+    return (chatRoomID: string, memberID?: string) => {
+      return getters.isJoinedChatRoomForChatRoom(rootState[chatRoomID], memberID)
+    }
+  },
+  chatRoomActiveMemberIds (state, getters, rootState) {
+    return (chatRoomID: string) => {
+      return getters.chatRoomActiveMemberIdsForChatRoom(rootState[chatRoomID])
+    }
```
taoeffect: This new getter has been added but is never used anywhere. So are you sure it should be here?

And if you want it, maybe it should be renamed to `chatRoomActiveMemberIdsForChatRoomID`?

frontend/model/contracts/chatroom.js
```diff
@@ -695,7 +693,8 @@ sbp('chelonia/defineContract', {
   methods: {
     'gi.contracts/chatroom/_cleanup': ({ contractID, state }) => {
       if (state?.members) {
-        sbp('chelonia/contract/release', Object.keys(state.members)).catch(e => {
+        const activeMemberIds = Object.keys(state.members).filter(memberID => !state.members[memberID].hasLeft)
```
taoeffect, line 697: Is this another one that should be replaced with the new `chatRoomActiveMemberIdsForChatRoom` getter?

frontend/controller/actions/identity-kv.js
```diff
@@ -316,5 +342,59 @@ export default (sbp('sbp/selectors/register', {
...
+    const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
+    if (!identityContractID) {
+      throw new Error('Unable to fetch cached names without an active session')
+    }
+    return (await sbp('chelonia/kv/get', identityContractID, KV_KEYS.NS_CACHE))?.data || []
+  },
+  'gi.actions/identity/kv/saveCachedNames': () => {
+    const identityContractID = sbp('state/vuex/state').loggedIn?.identityContractID
+    if (!identityContractID) {
+      throw new Error('Unable to update notification status without an active session')
```
taoeffect: o4-mini high caught: wrong error message here, should be:

```js
throw new Error('Unable to update cached names without an active session')
```

corrideat: Yep, good catch.

frontend/model/contracts/chatroom.js
corrideat: Possibly. Currently, `_cleanup` doesn't get passed `getters`, which is why I'm not using it.

frontend/model/contracts/chatroom.js
taoeffect: On the latest round, Opus 4 said something quite interesting. I'm pretty sure it's wrong, but neverthless I wasn't expecting this sort of reasoning from it:

### 1. Backward Compatibility Problem

```javascript
// frontend/model/contracts/chatroom.js, line 170
state.members[memberID].hasLeft = true
```

This assumes `state.members[memberID]` already exists, but for contracts created before this change, members who left would have been deleted. This will throw an error.

**Fix:**
```javascript
  state.members[memberID].hasLeft = true
} else {
  // Handle the case where member was already deleted in older contracts
  state.members[memberID] = { hasLeft: true }
}
```

frontend/model/chatroom/getters.js
corrideat: Good question. I added this unused getter for symmetry.

The name follows this convention (although that doesn't prevent the ForChatRoomID` name):

```
`-ForGroup` pattern:

Some getters (see, e.g., `groupSettings` and `groupSettingsForGroup`) are
implemented in pairs, with a variant having the `ForGroup` suffix. This is
because the non-suffixed version is supposed to work for the current group
(meaning the result of `currentGroupState`), while the suffixed version takes
a group state as a parameter. To avoid redundancy, the non-suffixed version is
implemented using the suffixed version (i.e., calling it with
`getters.currentGroupState` as a parameter).
```

frontend/model/contracts/chatroom.js
corrideat: Yes, good catch.

frontend/controller/actions/identity-kv.js
```diff
@@ -1,10 +1,35 @@
...
+  const recheckedNames = []
+
+  for (let i = 0; i < namesToCheck.length; i += BATCH_SIZE) {
+    const batch = namesToCheck.slice(i, i + BATCH_SIZE)
+    const results = await Promise.all(batch.map(async (name) => {
+      const value = await sbp('namespace/lookup', name, { skipCache: true })
+      return value ? name : null
+    }))
+    recheckedNames.push(...results.filter(v => !!v))
+  }
```
taoeffect: Opus 4 points to the lack of error handling here. I don't know if this is how we should handle it specifically, but should there be more error handling here?

### 2. Missing Error Handling in Batch Operations

```javascript
// frontend/controller/actions/identity-kv.js, lines 23-28
const results = await Promise.all(batch.map(async (name) => {
  const value = await sbp('namespace/lookup', name, { skipCache: true })
  return value ? name : null
}))
```

If one lookup fails, the entire batch fails. This should handle individual failures gracefully.

**Fix:**
```javascript
const results = await Promise.all(batch.map(async (name) => {
  try {
    const value = await sbp('namespace/lookup', name, { skipCache: true })
    return value ? name : null
  } catch (e) {
    console.warn(`Failed to lookup name ${name}:`, e)
    return null
  }
}))
```

frontend/model/chatroom/getters.js
corrideat: Yes, I'll double check. `getters.isJoinedChatRoom(contractID, myIdentityId)` should AFAICT be equivalent to `chatRoom && chatRoom.members && chatRoom.members[myIdentityId]`. However, the `.attributes` check might need to be added back if it's needed.

taoeffect (changes requested):
Think this is the final review! Great work @corrideat! Looking really solid! 💪

frontend/controller/actions/identity-kv.js
corrideat: I copied and pasted the previous suggestion you made. I think the revised code looks reasonable, although I might be tempted to do it with a `.catch`, like so:

```
const value = await sbp('namespace/lookup', name, { skipCache: true }).catch(e => {
  console.warn(`Failed to lookup name ${name}:`, e)
})
```

frontend/model/chatroom/getters.js
corrideat: Would you rather I remove or rename this getter?

frontend/controller/actions/identity-kv.js
taoeffect: > I copied and pasted the previous suggestion you made

\**Ahem*\* \**clears throat*\*

Screenshot 2025-07-27 at 12 07 20 PM

😛

corrideat: Fair enough. Fixed: I copied and pasted the previous suggestion you made <ins>by proxy</ins>.

Either way, the suggestion about error handling is a valid one and the code it replaced also lacked error handling.

frontend/model/chatroom/getters.js
taoeffect: If it follows the convention, then I suppose let's leave it as is

corrideat: If it _strictly_ followed the convention, it'd use `currentChatRoomState`, which would actually mean we could use it in `ChatMixin`.

taoeffect (approved):
This works really well! Amazing job @corrideat!

✅ Reviewed!
✅ Tested and verified! (local dev data only)

PR #2879, opened 2025-07-25
Add registration disabled check to event route (Closes #2878)

### LINKED ISSUE #2878: Envvar to disable signups / registration / sign ups
### Problem

Sometimes you want users to be able to log in and use the site, but not create new user accounts.

### Solution

Implement `CHELONIA_REGISTRATION_DISABLED`.

taoeffect:
Implement `CHELONIA_REGISTRATION_DISABLED`.

### DIFF
--- a/backend/routes.js
+++ b/backend/routes.js
@@ -221,7 +221,7 @@ route.POST('/event', {
           return Boom.unauthorized('This contract type requires ownership information', 'shelter')
         }
         if (process.env.CHELONIA_REGISTRATION_DISABLED) {
-          return Boom.unauthorized('Registration disabled', 'shelter')
+          return Boom.forbidden('Registration disabled')
         }
         // rate limit signups in production
         if (!SIGNUP_LIMIT_DISABLED) {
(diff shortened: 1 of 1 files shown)


backend/routes.js
```diff
@@ -217,7 +217,12 @@ route.POST('/event', {
         const manifest = await sbp('chelonia.db/get', deserializedHEAD.head.manifest)
         const parsedManifest = JSON.parse(manifest)
         const { name } = JSON.parse(parsedManifest.body)
-        if (name !== 'gi.contracts/identity') return Boom.unauthorized('This contract type requires ownership information', 'shelter')
+        if (name !== 'gi.contracts/identity') {
+          return Boom.unauthorized('This contract type requires ownership information', 'shelter')
+        }
+        if (process.env.CHELONIA_REGISTRATION_DISABLED) {
+          return Boom.unauthorized('Registration disabled', 'shelter')
```
corrideat: 401 Unauthorized doesn't sound like the right response type, since it's used to indicate the need for credentials, and in this case credentials won't help. 403 Forbidden sounds more apt for this kind of situation, or else a generic 400 Bad Request.

Also, for future reference, this logic here might need to change a bit as the server is made more generic (the issue is `(name !== 'gi.contracts/identity')`, which is GI-specific). I think this is one of the only few places where logic needs to be rethought.

corrideat (changes requested):

corrideat (approved):

PR #2882, opened 2025-07-28
Move ZKPP files into libchelonia

### LINKED ISSUE #2864: Move ZKPP code into @chelonia/lib
### Problem

zkpp.js and zkppConstants.js are essential for logging in to a Chelonia app, but they are still in this repo:

Image

### Solution

Move the following into @chelonia/lib:

- zkpp.js
- zkppConstants.js
- functions.js (partially moved already, just needs to be deleted here and with references updated)

And maybe the circular list too as a Chelonia utility... need to evaluate that though. If it's only being used by Group Income (and not Chelonia) then it should be moved into Group Income's `utils` folder.

corrideat:
Closes #2864


taoeffect (approved):
Excellent!

PR #2883, opened 2025-07-30
add scripts/dist.sh

taoeffect:
Makes it faster for me to generate releases and handles `._` files on macOS.

### DIFF
--- a/scripts/dist.sh
+++ b/scripts/dist.sh
@@ -1,19 +1,22 @@
 #!/bin/sh
 
-# exist on error
+# exit on error
 set -e
-VERSION=$(grep '"version"' package.json | sed 's/.*"version": *"\([^"]*\)".*/\1/')
-
 # echo commands
 set -x
 
+VERSION=$(grep '"version"' package.json | sed 's/.*"version": *"\([^"]*\)".*/\1/')
+TAR_FLAGS="--no-acls"
+
+# prevent ._ files from being included in the archive on macOS
+[ "$(uname -s)" = "Darwin" ] && TAR_FLAGS="--no-mac-metadata --no-acls"
+
 grunt clean
 grunt deploy
-# prevent ._ files from being included in the archive on macOS
-tar -czv --no-mac-metadata --no-acls -f gi-v${VERSION}.debug.tgz dist
+tar -czv $TAR_FLAGS -f gi-v${VERSION}.debug.tgz dist
 
 grunt clean
 NODE_ENV=production grunt deploy
-tar -czv --no-mac-metadata --no-acls -f gi-v${VERSION}.tgz dist
+tar -czv $TAR_FLAGS -f gi-v${VERSION}.tgz dist
 
 sha256sum gi-*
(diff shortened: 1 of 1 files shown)


akhileshthite (dismissed):
Looks good!

scripts/dist.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+VERSION=$(grep '"version"' package.json | sed 's/.*"version": *"\([^"]*\)".*/\1/')
+
+# echo commands
+set -x
+
+grunt clean
+grunt deploy
+# prevent ._ files from being included in the archive on macOS
+tar -czv --no-mac-metadata --no-acls -f gi-v${VERSION}.debug.tgz dist
```
corrideat: These options aren't portable.

For example:

```
tar: unrecognized option '--no-mac-metadata'
Try 'tar --help' or 'tar --usage' for more information.
```

Maybe using `find` along with `cpio` would have the desired effect.

For example,

```
find dist -not -name '._*' | cpio -H ustar -o - | gzip -9 > gi-v${VERSION}.debug.tgz
```

scripts/dist.sh
```diff
@@ -0,0 +1,19 @@
...
+
+grunt clean
+NODE_ENV=production grunt deploy
+tar -czv --no-mac-metadata --no-acls -f gi-v${VERSION}.tgz dist
```
corrideat: Same as above

scripts/dist.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+# exist on error
```
corrideat: Maybe you meant `exit`. Aslo, there's no need for two separate `set` calls, I think usually you'd do `set -ex`

scripts/dist.sh
```diff
@@ -0,0 +1,22 @@
...
+# exit on error
+set -e
+# echo commands
+set -x
+
+VERSION=$(grep '"version"' package.json | sed 's/.*"version": *"\([^"]*\)".*/\1/')
+TAR_FLAGS="--no-acls"
+
+# prevent ._ files from being included in the archive on macOS
+[ "$(uname -s)" = "Darwin" ] && TAR_FLAGS+=" --no-mac-metadata"
```
corrideat: This is better but still not entirely portable. There was [GNU-Darwin](https://en.wikipedia.org/wiki/Darwin_(operating_system)) and others that will match this but may not support `--no-mac-metadata`, which is a non-standard option that only exists on `bsdtar`, but doesn't exist on GNU tar or OpenBSD tar.

The suggested cpio option, `find dist -not -name '._*' | cpio -H ustar -o | gzip -9 > gi-v${VERSION}.debug.tgz`, should work everywhere.

taoeffect: No, it doesn't work because `._` files don't exist on macOS so the `find` command won't work

corrideat: But have you checked if the files get added to the archive?

corrideat: That `find` command should still work whether the `._` files exist or not.

corrideat (approved):

PR #2888, opened 2025-08-02
#2887 - `<a>` tag markdown rendering issue

### LINKED ISSUE #2887: Link rendering issues related to markdown-utils.js
### Problem

I think some recent changes to `markdown-utils` have impacted rendering components with `RenderMessageWithMarkdown`.

Specifically, rendering URLs like this no longer work: `[foo](/bar)`.

For example, try this:

```bash
$ export CHELONIA_SERVER_MESSAGES='[{"type":"giHomeWarning","message":"This is [a link](/to-a-page)"}]'
$ grunt dev
```

It will print the markdown instead of rendering the link.

And similarly, in chat messages, `test [message](/test)` doesn't get turned into a link.

The solution below fixes this, but **there's another problem** I just discovered:

Fully-qualified links like do render, but get replaced with http://localhost:3000/app/

```
$ export CHELONIA_SERVER_MESSAGES='[{"type":"giHomeWarning","message":"Here is [a subpage](http://localhost:3000/subpage)"}]'
$ grunt dev
```

Image

And in chat, `[a subpage](http://localhost:3000/subpage)` gets replaced with a link like this:

Image

#### Problem 1

Here's what Opus 4 thinks is the fix for the first problem. I tried its first suggestion and it worked.

I don't know if this is the best solution or not, so please verify it, and also make sure that nothing else is broken.

<details>

Looking at the code, I can see the issue. The problem is in the `markdown-utils.js` file, specifically in the custom link extension for the `marked` library.

Here's the problematic code:

```javascript
{
  name: 'link',
  level: 'inline',
  renderer (token) {
    const { isValid } = validateURL(token.href)
    if (isValid) {
      const { href, text } = token
      return `<a class="link" href="${href}" target="_blank">${text}</a>`
    }
    return token.raw  // <-- This returns the raw markdown when validation fails
  }
}
```

The issue is that `validateURL('/blah')` is likely returning `{ isValid: false }` for relative URLs like `/blah`. When the validation fails, the renderer returns `token.raw`, which is the original markdown text `[this is a link](/blah)`.

The `validateURL` function (imported from `'./misc.js'`) is probably only validating absolute URLs (e.g., `https://example.com`) and treating relative URLs as invalid.

**Solution:**

You have a few options to fix this:

1. **Modify the link renderer to handle relative URLs:**
```javascript
{
  name: 'link',
  level: 'inline',
  renderer (token) {
    const { href, text } = token
    const { isValid } = validateURL(href)
    
    // Check if it's a relative URL starting with /
    const isRelativeURL = href.startsWith('/')
    
    if (isValid || isRelativeURL) {
      return `<a class="link" href="${href}" target="_blank">${text}</a>`
    }
    return token.raw
  }
}
```

2. **Update the `validateURL` function to accept relative URLs as valid**

3. **Remove the validation entirely if you trust the markdown content:**
```javascript
{
  name: 'link',
  level: 'inline',
  renderer (token) {
    const { href, text } = token
    return `<a class="link" href="${href}" target="_blank">${text}</a>`
  }
}
```

The first option is probably the safest as it maintains validation for external URLs while allowing relative URLs to work properly.

</details>

---------

#### Problem 2

For the second problem, Claude Opus 4 thinks the problem is in `RenderMessageWithMarkdown.js`, and suggests replacing some of the code in `recursiveCall`, specifically, replacing this:

```js
          if (isHttpValid && url.origin === document.location.origin) {
            const path = url.pathname.split(this.$router.options.base)[1]
            const query = {}
            for (const [key, value] of url.searchParams) {
              query[key] = value
            }
            routerOptions.route = { path, query, hash: url.hash }
            routerOptions.href = this.$router.resolve(routerOptions.route).href
            routerOptions.isInAppRouter = true
          }
```

With this:

```js
          if (isHttpValid && url.origin === document.location.origin) {
            const base = this.$router.options.base || '/'
            // Only treat as in-app route if the path starts with the router base
            if (url.pathname.startsWith(base)) {
              const path = url.pathname.slice(base.length) || '/'
              const query = {}
              for (const [key, value] of url.searchParams) {
                query[key] = value
              }
              routerOptions.route = { path, query, hash: url.hash }
              routerOptions.href = this.$router.resolve(routerOptions.route).href
              routerOptions.isInAppRouter = true
            }
            // Otherwise, leave it as a regular external link
          }
```

Claude claims:

> This way:
> - `http://localhost:3000/app/dashboard` → handled by Vue Router as `/dashboard`
> - `http://localhost:3000/subpage` → left as a regular link (not handled by Vue Router)

Again, I don't know if this is the right solution, but it seems to work. Please double-check and verify.

SebinSong:
closes #2887 

`[a subpage](http://localhost:3000/subpage)` works fine now.

image

### DIFF
--- a/frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
+++ b/frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
@@ -32,8 +32,11 @@ const RenderMessageWithMarkdown: any = {
         if (entry.tagName === 'A' && entry.attributes.href) {
           const { href } = entry.attributes
           const { url, isHttpValid } = validateURL(href)
-          if (isHttpValid && url.origin === document.location.origin) {
-            const path = url.pathname.split(this.$router.options.base)[1]
+          const appRouteBase = this.$router.options.base
+          const appOrigin = document.location.origin + appRouteBase
+
+          if (isHttpValid && url.origin === appOrigin) {
+            const path = url.pathname.split(appRouteBase)[1]
             const query = {}
             for (const [key, value] of url.searchParams) {
               query[key] = value
--- a/frontend/views/utils/misc.js
+++ b/frontend/views/utils/misc.js
@@ -25,7 +25,22 @@ export const showNavMixin = {
 }
 
 export function validateURL (url: string): Object {
-  const response = {}
+  const pathOnlyRegExp = /^\/[^\s]*$/
+
+  const response = {
+    isValid: false,
+    isHttpValid: false,
+    isMailtoValid: false,
+    url: null
+  }
+
+  if (pathOnlyRegExp.test(url)) {
+    // If the passed URL is a path only string such as '/app/chatroom/chatID', we consider it as a valid URL.
+    response.isValid = true
+    response.isHttpValid = true
+    response.url = url
+    return response
+  }
 
   try {
     const objURL = new URL(url)
--- a/frontend/views/utils/misc.js
+++ b/frontend/views/utils/misc.js
@@ -27,7 +27,7 @@ export const showNavMixin = {
 export function validateURL (url: string): Object {
   const pathOnlyRegExp = /^\/[^\s]*$/
 
-  const response = {
+  const response: any = {
     isValid: false,
     isHttpValid: false,
     isMailtoValid: false,


taoeffect (commented):
Nice, this seems to fix the two problems mentioned in the issue, however it might also introduce a regression.

Correct me if I'm wrong, but this logic that Opus identified is meant to handle in-app links in a special way for a reason:

```js
        if (entry.tagName === 'A' && entry.attributes.href) {
          const { href } = entry.attributes
          const { url, isHttpValid } = validateURL(href)
          if (isHttpValid && url.origin === document.location.origin) {
            const path = url.pathname.split(this.$router.options.base)[1]
            const query = {}
            for (const [key, value] of url.searchParams) {
              query[key] = value
            }
            routerOptions.route = { path, query, hash: url.hash }
            routerOptions.href = this.$router.resolve(routerOptions.route).href
            routerOptions.isInAppRouter = true
          }
        }

        const opts = routerOptions.isInAppRouter
          ? {
              class: 'link',
              attrs: { href: routerOptions.href },
              on: {
                click: (e) => {
                  routerOptions.route && this.$router.push(routerOptions.route).catch(logExceptNavigationDuplicated)
                  e?.preventDefault()
                },
                touchhold: (e) => {
                  routerOptions.href && sbp('okTurtles.events/emit', OPEN_TOUCH_LINK_HELPER, routerOptions.href)
                  e?.preventDefault()
                }
              
```

Now with these changes that code no longer seems to be used...

I'm noticing two issues:

1. Links like `test [a link](/app/dashboard)` will not use any of that `isInAppRouter` related code, they will be loaded in a new tab instead.
2. Links like `test [a link](dashboard)` and `test [a link](#dashboard)` do not get rendered at all. Should they be? 🤔 

Do any of the code changes suggested by Opus make sense to incorporate?

taoeffect (changes requested):
> 1. Links like `test [a link](/app/dashboard)` will not use any of that `isInAppRouter` related code, they will be loaded in a new tab instead.
> 2. Links like `test [a link](dashboard)` and `test [a link](#dashboard)` do not get rendered at all. Should they be? 🤔

Thinking on it more, and to answer my own question, I think we should address these two link issues in this PR too while we're at it, because it would make it easier for us to link to specific features in the app using Markdown. And the 2nd issue points to valid links that should be rendered too, so we should render them.

taoeffect (approved):
Per Slack, we've agreed to leave the two issues in the review for a separate issue and merge this now.

PR #2891, opened 2025-08-05
#2889 - Fix two link related issues

### LINKED ISSUE #2889: Two remaining link rendering issues
### Problem

This is a followup to PR #2888, please [see this comment there](https://github.com/okTurtles/group-income/pull/2888#pullrequestreview-3080895415).

In summary:

> 1. Links like `test [a link](/app/dashboard)` will not use any of that `isInAppRouter` related code, they will be loaded in a new tab instead.
> 2. Links like `test [a link](dashboard)` and `test [a link](#dashboard)` do not get rendered at all. Should they be? 🤔

### Solution

See [this discussion on Slack](https://okturtles.slack.com/archives/C0EH7P20Y/p1754107850897959?thread_ts=1754094004.123099&cid=C0EH7P20Y) and also Opus' [suggested fixes here](https://github.com/okTurtles/group-income/issues/2887).

1. Make sure in-app links like `test [a link](/app/dashboard)` use the router's `.push` instead of opening up a new tab
2. Make sure links like `test [a link](dashboard)` and `test [a link](#dashboard)` get rendered properly. In the case of `[a link](dashboard)`, it should probably use the in-app router `.push`, and in the case of `#dashboard`, it might use `.push` or just use the browser's native ability to visit anchors.

SebinSong:
closes #2889 

Below is a video record for the fix 1. which is to perceive `/app/...` as an in-app route and navigate to it. (Pls note that for any invalid page route, the navigation fallbacks to the `/app/dashboard` page which is an existing behavior)

https://github.com/user-attachments/assets/f4c01d85-4d3c-46b1-98af-597d5de9395e

### DIFF
--- a/frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
+++ b/frontend/views/containers/chatroom/chat-mentions/RenderMessageWithMarkdown.js
@@ -31,11 +31,11 @@ const RenderMessageWithMarkdown: any = {
         const routerOptions = { isInAppRouter: false, route: {}, href: '' }
         if (entry.tagName === 'A' && entry.attributes.href) {
           const { href } = entry.attributes
-          const { url, isHttpValid } = validateURL(href)
+          const { url, isHttpValid } = validateURL(href, true)
           const appRouteBase = this.$router.options.base
           const appOrigin = document.location.origin + appRouteBase
 
-          if (isHttpValid && url.origin === appOrigin) {
+          if (isHttpValid && url.href.startsWith(appOrigin)) {
             const path = url.pathname.split(appRouteBase)[1]
             const query = {}
             for (const [key, value] of url.searchParams) {
--- a/frontend/views/containers/global-dashboard/NewsAndUpdates.vue
+++ b/frontend/views/containers/global-dashboard/NewsAndUpdates.vue
@@ -13,7 +13,6 @@
       .c-post-content
         h3.is-title-4 {{ post.title }}
         render-message-with-markdown(:text='post.content')
-        //- div(v-safe-html:a='renderMarkdown(post.content)')
 </template>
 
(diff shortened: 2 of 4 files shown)


taoeffect (changes requested):
@SebinSong You did a really great job on this PR, everything seems to work correctly, but I made a mistake in answering [your question `2-1` here](https://github.com/okTurtles/group-income/issues/2889#issuecomment-3152614662).

For links like `[a link](string)`, I gave the wrong answer, my apologies. We shouldn't break how links conventionally work, and by having `[a link](string)` go to `/app/string` we're violating standard link rules.

If this link appears in the chatroom, it will appear under a URL like:

```
http://localhost:3000/app/group-chat/zLDXeQ2AgfCuUycAeYWYwvMdDavziSHjEpsDWGZBTsFi5VswBAYyTgQc
```

Normally, a link like `[string](string)` should be rendered underneath the last directory in the currently URL, look at how Github does it: [string](string)

So for a link like `[string](string)` that appears under the chatroom, it should go to:

```
http://localhost:3000/app/group-chat/string
```

My bad for not catching this earlier when you asked!

SebinSong:
@taoeffect Thanks Greg.
It seems like Github handles a plain string by just using it as a `href` value, and that leads to the behavior you described by the browser. 

image

So I updated the logic to do the same.
In our case, it would behave like below:

`[string](string)` becomes `<a href='string'></a>` -> When clicked within a chatroom, browser triggers navigating to `app/group-chat/string` -> There is no chatroom that corresponds to this, so it immediately fallbacks to the `General` chatroom.

taoeffect (approved):
Excellent work @SebinSong! I've tested this and every type of link works correctly now, with the one exception it seems of anchor links, so I'll create a followup issue for that.

PR #2892, opened 2025-08-06
#2886 - Show red dots for new message

### LINKED ISSUE #2886: Show red dot on groups that have a new message in chatrooms (private or not)
### Problem

Private chatrooms by default result in notifications (similar to DMs), and it can be a bit disorienting to hear a notification, return to the app, and then having no idea where the notification came from (because it came from some other group).

### Solution

The solution should be to show the red dot on the group in the group switcher whenever there is any new chatroom activity in a group for which you have notifications enabled.

SebinSong:
closes #2886 

These badge UIs will be displayed if the noticiation settings are enabled for a private-chatroom.

### DIFF
--- a/frontend/model/notifications/messageReceivePostEffect.js
+++ b/frontend/model/notifications/messageReceivePostEffect.js
@@ -44,7 +44,7 @@ async function messageReceivePostEffect ({
   const shouldSoundMessage = messageSound === MESSAGE_NOTIFY_SETTINGS.ALL_MESSAGES ||
     (messageSound === MESSAGE_NOTIFY_SETTINGS.DIRECT_MESSAGES && isDMOrMention)
   const shouldAddToUnreadMessages = isDMOrMention ||
-    (privacyLevelPrivate && shouldSoundMessage) ||
+    (privacyLevelPrivate && (shouldNotifyMessage || shouldSoundMessage)) ||
     [MESSAGE_TYPES.INTERACTIVE, MESSAGE_TYPES.POLL].includes(messageType)
 
   await sbp('chelonia/contract/retain', contractID, { ephemeral: true })
--- a/frontend/model/notifications/messageReceivePostEffect.js
+++ b/frontend/model/notifications/messageReceivePostEffect.js
@@ -31,7 +31,21 @@ async function messageReceivePostEffect ({
   const identityContractID = rootState.loggedIn?.identityContractID
   if (!identityContractID) return
   const isDM = rootState[identityContractID].chatRooms[contractID]
-  const shouldAddToUnreadMessages = isDMOrMention || [MESSAGE_TYPES.INTERACTIVE, MESSAGE_TYPES.POLL].includes(messageType)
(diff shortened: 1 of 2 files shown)


frontend/model/notifications/messageReceivePostEffect.js
```diff
@@ -31,7 +31,21 @@ async function messageReceivePostEffect ({
...
+    : rootGetters.chatNotificationSettings.publicDefault
+  )
+  const { messageNotification, messageSound } = chatNotificationSettings
+  const shouldNotifyMessage = messageNotification === MESSAGE_NOTIFY_SETTINGS.ALL_MESSAGES ||
+  (messageNotification === MESSAGE_NOTIFY_SETTINGS.DIRECT_MESSAGES && isDMOrMention)
+  const shouldSoundMessage = messageSound === MESSAGE_NOTIFY_SETTINGS.ALL_MESSAGES ||
+    (messageSound === MESSAGE_NOTIFY_SETTINGS.DIRECT_MESSAGES && isDMOrMention)
+  const shouldAddToUnreadMessages = isDMOrMention ||
+    (privacyLevelPrivate && (shouldNotifyMessage || shouldSoundMessage)) ||
+    [MESSAGE_TYPES.INTERACTIVE, MESSAGE_TYPES.POLL].includes(messageType)
```
taoeffect, line 48: The new code in this PR seems to be this line:

```js
(privacyLevelPrivate && (shouldNotifyMessage || shouldSoundMessage)) ||
```

What I'm not sure about is the `privacyLevelPrivate` condition though. Remember that although the issue uses a private chatroom as an example, the solution part says:

> The solution should be to show the red dot on the group in the group switcher whenever there is any new chatroom activity in a group for which you have notifications enabled.

In other words, I think if we have notifications enabled for any type of channel, whether private or not, we should probably show the red dot.

taoeffect (commented):
Great work @SebinSong!

Added a comment, let me know if you agree:

test/cypress/integration/group-chat.spec.js
```diff
@@ -366,6 +366,17 @@ describe('Group Chat Basic Features (Create & Join & Leave & Close)', () => {
...
+    switchUser(user2)
+    cy.giRedirectToGroupChat()
+    cy.giSwitchChannel('channel22')
+    cy.giSendMessage(user2, 'Hi, There!')
+
+    switchUser(user1)
+    cy.getByDT('groupChatLink').get('.c-badge.is-compact[aria-label="1 new notifications"]').contains('1')
+    cy.giRedirectToGroupChat()
+  })
+
```
SebinSong, line 379: This cypress test item is still worth keeping because a private channel by default receives notifications for all messages:

image

and this test checks if the relevant UIs work correctly for this setting.

frontend/model/notifications/messageReceivePostEffect.js
SebinSong: Right. Then checking `privacyLevelPrivate` is not needed here. Updated.

taoeffect (approved):
Great work, LGTM!

PR #2894, opened 2025-08-09
Close DB on exit

### LINKED ISSUE #2884: Flush sqlite database to disk on shutdown
### Problem

We're having contract divergence issues on server restart with the sqlite backend forcing users to re-login.

### Solution

Tell sqlite to flush the database to disk before the process quits.

If that doesn't work, try switching to postgres (#2698).

corrideat:
Closes #2884 

This closes the DB when the process is terminated. However, it likely isn't sufficient alone to solve the issue, as it seems like the process running in production is being forcefully terminated by Docker, due to `bash` not passing through signals to subprocesses.

To fully address this issue, the way the Docker container is built and used should be modified too. The most immediate course of action is using `exec` to invoke the final command in `entrypoint.sh`. This will unfortunately result in some duplication, since the `run` subroutine can't be called with `exec`. A more comprehensive approach is removing `entrypoint.sh` and the need for it entirely, but that's a separate issue.

This PR also adds logging when the process receives a signal that causes it to terminate, which should be helpful for debugging and also to verify if changes made to `entrypoint.sh` (as discussed) actually successfully pass down termination signals.

### DIFF
--- a/backend/database-router.js
+++ b/backend/database-router.js
@@ -107,14 +107,18 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
   async clear (): Promise<void> {
     for (const backend of new Set(Object.values(this.backends))) {
       // $FlowFixMe[incompatible-use]
-      await backend.clear()
+      await backend.clear().catch(e => {
+        console.error(e, 'Error clearing DB')
+      })
     }
   }
 
   async close () {
     for (const backend of new Set(Object.values(this.backends))) {
       // $FlowFixMe[incompatible-use]
-      await backend.close()
+      await backend.close().catch(e => {
+        console.error(e, 'Error closing DB')
+      })
     }
   }
 }
--- a/backend/database-router.js
+++ b/backend/database-router.js
@@ -113,6 +113,7 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
 
   async close () {
     for (const backend of new Set(Object.values(this.backends))) {
+      // $FlowFixMe[incompatible-use]
       await backend.close()
     }
   }
--- a/backend/database-router.js
+++ b/backend/database-router.js
@@ -106,19 +106,27 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
 
   async clear (): Promise<void> {
     for (const backend of new Set(Object.values(this.backends))) {
-      // $FlowFixMe[incompatible-use]
-      await backend.clear().catch(e => {
-        console.error(e, 'Error clearing DB')
-      })
+      try {
+        // $FlowFixMe[incompatible-use]
+        await backend.clear()
+      } catch (e) {
+        // $FlowFixMe[incompatible-use]
+        const prefix = Object.entries(this.backends).find(([, b]) => b === backend)[0]
+        console.error(e, `Error clearing DB for prefix ${prefix}`)
+      }
     }
   }
 
   async close () {
     for (const backend of new Set(Object.values(this.backends))) {
-      // $FlowFixMe[incompatible-use]
-      await backend.close().catch(e => {
-        console.error(e, 'Error closing DB')
-      })
+      try {
+        // $FlowFixMe[incompatible-use]
+        await backend.close()
+      } catch (e) {
+        // $FlowFixMe[incompatible-use]
+        const prefix = Object.entries(this.backends).find(([, b]) => b === backend)[0]
+        console.error(e, `Error closing DB for prefix ${prefix}`)
+      }
     }
   }
 }
--- a/backend/database-router.js
+++ b/backend/database-router.js
@@ -110,4 +110,10 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
       await backend.clear()
     }
   }
+
+  async close () {
+    for (const backend of new Set(Object.values(this.backends))) {
+      await backend.close()
+    }
+  }
 }
--- a/backend/database-sqlite.js
+++ b/backend/database-sqlite.js
@@ -36,7 +36,6 @@ export default class SqliteBackend extends DatabaseBackend implements IDatabaseB
       throw new Error(`The ${filename} SQLite database is already open.`)
     }
     this.db = new Database(join(dataFolder, filename))
-    process.on('exit', () => this.db.close())
     this.run('CREATE TABLE IF NOT EXISTS Data(key TEXT NOT NULL PRIMARY KEY, value TEXT NOT NULL)')
     console.info(`Connected to the ${filename} SQLite database.`)
     this.readStatement = this.db.prepare('SELECT value FROM Data WHERE key = ?')
@@ -63,5 +62,9 @@ export default class SqliteBackend extends DatabaseBackend implements IDatabaseB
 
   async deleteData (key: string) {
     await this.deleteStatement.run(key)
+  }
+
+  close () {
+    this.db.close()
   }
 }
--- a/backend/database-sqlite.js
+++ b/backend/database-sqlite.js
@@ -36,6 +36,7 @@ export default class SqliteBackend extends DatabaseBackend implements IDatabaseB
       throw new Error(`The ${filename} SQLite database is already open.`)
     }
     this.db = new Database(join(dataFolder, filename))
+    process.on('exit', () => this.db.close())
     this.run('CREATE TABLE IF NOT EXISTS Data(key TEXT NOT NULL PRIMARY KEY, value TEXT NOT NULL)')
     console.info(`Connected to the ${filename} SQLite database.`)
     this.readStatement = this.db.prepare('SELECT value FROM Data WHERE key = ?')
--- a/backend/database.js
+++ b/backend/database.js
@@ -7,10 +7,12 @@ import fs from 'fs'
 import { readdir, readFile } from 'node:fs/promises'
 import path from 'node:path'
 import '@sbp/okturtles.data'
+import '@sbp/okturtles.events'
 import { checkKey, parsePrefixableKey, prefixHandlers } from '@chelonia/lib/db'
 import LRU from 'lru-cache'
 import { initVapid } from './vapid.js'
 import { initZkpp } from './zkppSalt.js'
+import { SERVER_EXITING } from './events.js'
 
 const Boom = require('@hapi/boom')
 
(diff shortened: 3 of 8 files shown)


backend/server.js
```diff
@@ -684,3 +684,20 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
     // Repeat every 1 hour
   }, 1 * 60 * 60 * 1000)
 })()
+
+const handleSignal = (signal: string, code: number) => {
+  process.on(signal, () => {
+    console.error(`Exiting upon receiving ${signal} (${code})`)
+    process.exit(128 + code)
```
taoeffect: Consider making this more developer/sysadmin friendly, if you want to add a number to the code, maybe add something like `100` so that the `code` is immediately readable

corrideat: Follows this convention: https://tldp.org/LDP/abs/html/exitcodes.html (see Fatal error signal "n"), which you can test, e.g., by running `cat` (no arguments) and sending various signals to it.

It's also readable in hex

taoeffect: Cool! Didn't know that, could you add a comment explaining that with that link?

akhileshthite (dismissed):
Updated chel serve: https://github.com/okTurtles/chel/pull/59/commits/04af8a19eba5e813a6d13a80d0a0d77a514cd3ee

backend/database-sqlite.js
```diff
@@ -36,6 +36,7 @@ export default class SqliteBackend extends DatabaseBackend implements IDatabaseB
       throw new Error(`The ${filename} SQLite database is already open.`)
     }
     this.db = new Database(join(dataFolder, filename))
+    process.on('exit', () => this.db.close())
```
taoeffect: This relies on the runtime having an called `exit` in the runtime that we're using, and also only closing the DB when that event is issued.

What if theoretically there were some need to close the DB under other circumstances? With this PR we wouldn't be able to do it. However, if we added a mandatory API to the `DatabaseBackend` class called `close`, then we could close the database connection whenever we needed to. It would give us more flexibility for future-proofing things.

backend/database-router.js
```diff
@@ -110,4 +110,11 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
       await backend.clear()
     }
   }
+
+  async close () {
+    for (const backend of new Set(Object.values(this.backends))) {
+      // $FlowFixMe[incompatible-use]
+      await backend.close()
+    }
```
taoeffect: We should attempt to close all databases, even if one throws

backend/database.js
```diff
@@ -235,8 +235,11 @@ export const initDB = async ({ skipDbPreloading }: { skipDbPreloading?: boolean
   if (persistence) {
     const Ctor = (await import(`./database-${persistence}.js`)).default
     // Destructuring is safe because these methods have been bound using rebindMethods().
-    const { init, readData, writeData, deleteData } = new Ctor(options[persistence])
+    const { init, readData, writeData, deleteData, close } = new Ctor(options[persistence])
     await init()
+    process.on('exit', () => {
+      close().catch(e => console.error(e, 'Error closing DB'))
```
taoeffect: Could the error message please specify which DB has failed? So for example:

```js
close().catch(e => console.error(e, `Error closing DB ${persistence}`))
```

backend/database-router.js
```diff
@@ -107,7 +107,18 @@ export default class RouterBackend extends DatabaseBackend implements IDatabaseB
...
+        console.error(e, 'Error clearing DB')
+      })
+    }
+  }
+
+  async close () {
+    for (const backend of new Set(Object.values(this.backends))) {
+      // $FlowFixMe[incompatible-use]
+      await backend.close().catch(e => {
+        console.error(e, 'Error closing DB')
```
taoeffect: Can you also include the name of the DB here too in the these error messages?

backend/server.js
```diff
@@ -684,3 +668,22 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
     // Repeat every 1 hour
   }, 1 * 60 * 60 * 1000)
 })()
+
+const handleSignal = (signal: string, code: number) => {
+  process.on(signal, () => {
+    console.error(`Exiting upon receiving ${signal} (${code})`)
+    // Exit codes follow the 128 + signal code convention.
+    // See <https://tldp.org/LDP/abs/html/exitcodes.html>
+    process.exit(128 + code)
```
taoeffect: Note: as @corrideat found out `process.exit` doesn't allow async functions to be called, so we might not want to call `process.exit` here but instead directly call the close function on all DBs. But @corrideat says "that doesn't work"... so... not sure

corrideat: > "that doesn't work"

To expand on this, `process.exit` _needs_ to be called, but it needs to be done right after closing the DBs and doing everything else that's needed. I've resolved this by emitting an exit event (distinct from the `exit` Node.js event, which can only call synchronous functions)

backend/index.js
```diff
@@ -78,5 +72,43 @@ process.on('unhandledRejection', (reason, p) => {
...
+  process.on(signal, () => {
+    console.error(`Exiting upon receiving ${signal} (${code})`)
+    // Exit codes follow the 128 + signal code convention.
+    // See <https://tldp.org/LDP/abs/html/exitcodes.html>
+    exit(128 + code)
+  })
+}
+
+// Codes from <signal.h>
+[
```
taoeffect: Hmm... it's my understanding that `standard` should complain about no `;` in front of the `[` here... Why didn't it? (Can you add the `;`?)

corrideat: It should but it doesn't (nor does it for `()` expressions).

akhileshthite (approved):
Made changes in `chel serve` and tested: https://github.com/okTurtles/chel/pull/59/commits/baf72ab2507708f1ae5c889d2cb878aae8eb6a85

backend/index.js
```diff
@@ -78,5 +72,43 @@ process.on('unhandledRejection', (reason, p) => {
...
+
+// Codes from <signal.h>
+;[
+  ['SIGHUP', 1],
+  ['SIGINT', 2],
+  ['SIGQUIT', 3],
+  ['SIGTERM', 15],
+  ['SIGUSR1', 10],
+  ['SIGUSR2', 11]
+].forEach(([signal, code]) => handleSignal(signal, code))
```
taoeffect, line 106: So I'm trying to figure out under what circumstances the server shuts down with a `0` status code... and I don't see one except for receiving the `process.on('message'` below... this seems fine, but what about integration with the `chel serve` PR, will it work properly there too?

taoeffect: Looking into this a bit more deeply, the shutdown behavior has definitely changed with a simple `^C`, but the end result seems the same.

This is how this PR behaves:

```
[15:43:40.339] INFO (39343): [pubsub] Server listening
[15:43:40.339] INFO (39343): Backend server running at: http://host.local:8000
[15:43:40.339] INFO (39343): backend startup sequence complete.
^C
user@host [pr]
-> % [15:43:42.791] ERROR (39343): Exiting upon receiving SIGINT (2)
[15:43:42.793] INFO (39343): [pubsub] Server closed
[15:43:42.794] DEBUG (39343): [sbp] backend/server/stop []
[15:43:42.797] INFO (39343): Hapi server down
echo $?
130
```

Notice the continued messages after the prompt is returned, and 130 status code.

This is `master`:

```
[15:44:18.971] INFO (39577): [pubsub] Server listening
[15:44:18.971] INFO (39577): Backend server running at: http://host.local:8000
[15:44:18.971] INFO (39577): backend startup sequence complete.
^C
user@host [master]
-> % echo $?
130
```

Immediate exit, no more messages, same status code.

Well, I'm OK with extra messages, but is it really an `ERROR` for `^C`? That seems wrong to me. Perhaps `INFO` would make more sense?

Questions:

1. Under what circumstances will we see `0`? I've found none so far. I thought `^C` would do it (but then again `master` also gives `130` with `^C`). I think that only happens with `grunt dev` when a change is made to the backend code and grunt sends a message to relaunch. Of course this isn't a use-case with `chel serve` (may be one with `chel dev` though).
2. What happens when a container is stopped? People who run docker containers usually look for a clean exit code, as without it some settings will relaunch the container, I think. How do we handle this? Have you tested with a container?
3. So what about the Deno-based `chel serve` stuff? Are the changes here compatible with that, and will they also handle `chel serve` in a docker container? The good news is that I think @akhileshthite has already copied over some version of these changes into there, so you could test there too...

corrideat: > So I'm trying to figure out under what circumstances the server shuts down with a 0 status code... and I don't see one except for receiving the process.on('message' below... this seems fine, but what about integration with the chel serve PR, will it work properly there too?

It'll shut down with a 0 when receiving a signal from Node. The way `chel serve` works is different (there's no Grunt process there) so likely it'll never have a 0.

> 1. Under what circumstances will we see 0? I've found none so far. I thought ^C would do it (but then again master also gives 130 with ^C). I think that only happens with grunt dev when a change is made to the backend code and grunt sends a message to relaunch. Of course this isn't a use-case with chel serve (may be one with chel dev though).

You're right, usually `0` will not happen. `^C` will result in 130. Whether that's 'wrong' or not I guess is a matter of perspective. For example, `cat` will exit with `130` on `^C`, while `nginx` (when running on the foreground at least) will return 0, same for `ntpd`.

> 2. What happens when a container is stopped? People who run docker containers usually look for a clean exit code, as without it some settings will relaunch the container, I think. How do we handle this? Have you tested with a container?

The container will not be relaunched upon an explicit stop (under sane configurations anyhow), but I see the point about looking for a clean exit code.

> 3. So what about the Deno-based chel serve stuff? Are the changes here compatible with that, and will they also handle chel serve in a docker container? The good news is that I think @akhileshthite has already copied over some version of these changes into there, so you could test there too...

The changes here should be compatible with `chel`, although `0` will likely never be a valid exit code (rn).

On docker stop, note that by default that'll send SIGTERM (whereas `^C` is SIGINT and `^\` is SIGQUIT).

taoeffect (approved):
Thanks for the PR @corrideat! It seems a lot more stable. I note here that we've reached the conclusion that the server process will likely never exit with a `0` status code, which is something we may want to return to in the future. Approved!

PR #2895, opened 2025-08-18
#202 - Step 2, 3, 4: Implement `Add/Remove/Edit` permissions

SebinSong:
Work on **step 2** items in [this TODO list](https://github.com/okTurtles/group-income/issues/202#issuecomment-2567190831)

This PR is about building this 'Add Permissions' modal and implement various features & updates associated with it.

<br>
<br>

Permission details set up via this modal will be displayed in the **Roles and permissions** section in the **Group Settings** page below.
 **add permissions** button in orange box is only exposed to members who have `delegate-permissions` permission granted, which are `admin` role and `moderator(delegator)` role.
 
**Note**: Implementing hide/enable other functionalities depending on granted permissions will be worked in **Step 4**  [here](https://github.com/okTurtles/group-income/issues/202#issuecomment-2567190831), FYI

image

<br>
<br>

For those who don't have `view-permissions` permission granted, This table will only show their own permission details like below:

image

---

@taoeffect 
These two UIs below I built in this PR have no Figma reference actually. (built by combining existing ingredients in the app) Once the project has a designer back, they could be reviewed again and improved. Hope it looks good to you anyways.

---
EDIT: As part of **Step 3** I built below two modals and implement functionalities associated with them

[**'Remove role'** modal]

image

<br>
<br>

[**'Edit permissions'** modal]

image

<br>
<br>

These are accessed via below menu:

image

---

<a href="https://app.devin.ai/review/okturtles/group-income/pull/2895">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>


taoeffect (changes requested):
Great work @SebinSong! Sorry I haven't had a chance yet to do a manual review. I might be able to finally have time soon, but in the meantime I ran this through Opus 4.5 twice (in different ways):

### Review 1 (more comprehensive context)

<details>


### 1. Logic Error in `inviteRevoke` Validation (group.js)

```javascript
if (!invite.creatorID !== innerSigningContractID && !myPermissions.includes(GROUP_PERMISSIONS.REVOKE_INVITE)) {
```

The `!` operator has higher precedence than `!==`, so this evaluates as `(!invite.creatorID) !== innerSigningContractID`, which is incorrect. This will allow anyone to revoke any invite.

**Fix:**
```javascript
if (invite.creatorID !== innerSigningContractID && !myPermissions.includes(GROUP_PERMISSIONS.REVOKE_INVITE)) {
  throw new TypeError(L('You do not have permission to revoke this invite.'))
}
```

---

### 2. Missing Authorization Check in `updatePermissions` (group.js)

The `updatePermissions` action doesn't validate that the caller has `DELEGATE_PERMISSIONS`. Any active member can modify permissions:

```javascript
'gi.contracts/group/updatePermissions': {
  validate: actionRequireActiveMember(arrayOf(
    objectMaybeOf({
      memberID: stringMax(MAX_HASH_LEN, 'memberID'),
      // ... no permission check
    })
  )),
```

**Fix:**
```javascript
'gi.contracts/group/updatePermissions': {
  validate: actionRequireActiveMember((data, { getters, message: { innerSigningContractID } }) => {
    arrayOf(
      objectMaybeOf({
        memberID: stringMax(MAX_HASH_LEN, 'memberID'),
        action: validatorFrom(x => Object.values(GROUP_PERMISSION_UPDATE_ACTIONS).includes(x)),
        roleName: stringMax(GROUP_ROLE_NAME_MAX_CHAR, 'roleName'),
        permissions: arrayOf(stringMax(GROUP_PERMISSION_MAX_CHAR, 'permissions'))
      })
    )(data)

    const myProfile = getters.groupProfile(innerSigningContractID)
    const myPermissions = myProfile?.role?.permissions || []
    const isGroupCreator = innerSigningContractID === getters.currentGroupOwnerID

    if (!isGroupCreator && !myPermissions.includes(GROUP_PERMISSIONS.DELEGATE_PERMISSIONS)) {
      throw new TypeError(L('You do not have permission to delegate permissions.'))
    }
  }),
```

---

### 3. Null Reference Error in `edit` Action (group.js)

When editing a member who has no role, this will crash:

```javascript
case 'edit':
  groupProfile.role = {
    name: item.roleName || groupProfile.role.name,
    permissions: item.permissions || groupProfile.role.permissions
  }
  break
```

**Fix:**
```javascript
case 'edit':
  if (!groupProfile.role) {
    throw new Error('Cannot edit role for member without an existing role.')
  }
  groupProfile.role = {
    name: item.roleName || groupProfile.role.name,
    permissions: item.permissions || groupProfile.role.permissions
  }
  break
```

---

## Security Issues

### 4. No Validation of Permission Values (group.js)

The permissions array accepts any string, allowing attackers to set arbitrary permissions:

```javascript
permissions: arrayOf(stringMax(GROUP_PERMISSION_MAX_CHAR, 'permissions'))
```

**Fix:**
```javascript
permissions: arrayOf(validatorFrom(x => Object.values(GROUP_PERMISSIONS).includes(x)))
```

---

### 5. No Prevention of ADMIN Role Assignment (group.js)

Users with `DELEGATE_PERMISSIONS` can assign the ADMIN role, which would give them full control:

```javascript
case 'add':
  groupProfile.role = {
    name: item.roleName,
    permissions: item.permissions
  }
  break
```

**Fix (add to validation):**
```javascript
// In the validation function
for (const item of data) {
  if (item.roleName === GROUP_ROLES.ADMIN) {
    throw new TypeError(L('Cannot assign admin role.'))
  }
  if (item.permissions?.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR) && 
      !myPermissions.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR)) {
    throw new TypeError(L('Cannot grant permissions you do not have.'))
  }
}
```

---

### 6. Cross-Contract Getter Access in Chatroom (chatroom.js)

The chatroom `delete` validation accesses group getters which may not be available:

```javascript
validate: actionRequireInnerSignature((_, { state, getters, meta, message: { innerSigningContractID } }) => {
  const myProfile = getters.groupProfile(innerSigningContractID)
  const myPermissions = myProfile?.role?.permissions || []
```

The chatroom contract doesn't have direct access to `groupProfile` getter. This will likely fail or return undefined.

**Recommendation:** The permission check should happen in the group contract's `deleteChatRoom` action, not in the chatroom contract. Remove this check from `chatroom.js` or pass group permissions differently.

---

## Improvements

### 7. Use Proper Error Type in Process Function (group.js)

```javascript
if (!groupProfile) {
  throw new Error(L('Member does not exist.'))
}
```

Process functions should use `Errors.GIErrorIgnoreAndBan` for invalid data:

```javascript
if (!groupProfile) {
  console.error(`updatePermissions: member ${item.memberID} does not exist`)
  throw new Errors.GIErrorIgnoreAndBan('updatePermissions: member does not exist')
}
```

---

### 8. Initialize `selected` from prop in GroupMembersDropdown (GroupMembersDropdown.vue)

```javascript
data () {
  return {
    ephemeral: {
      selected: null,  // Not initialized from value prop
      isActive: false
    }
  }
},
```

**Fix:**
```javascript
created () {
  if (this.value) {
    this.ephemeral.selected = this.value
  }
}
```

---

### 9. Fragile Cypress Test Wait

```javascript
cy.wait(3 * 1000) // eslint-disable-line cypress/no-unnecessary-waiting
```

**Recommendation:** Replace with a proper condition wait:
```javascript
cy.getByDT('someElement').should('be.visible')
// or
cy.giWaitUntilMessagesLoaded()
```

---

### 10. Missing `memberID` Validation in `updatePermissions`

The action doesn't prevent users from modifying their own permissions or the group creator's permissions:

**Add to validation:**
```javascript
for (const item of data) {
  if (item.memberID === innerSigningContractID) {
    throw new TypeError(L('Cannot modify your own permissions.'))
  }
  if (item.memberID === getters.currentGroupOwnerID) {
    throw new TypeError(L('Cannot modify the group creator\'s permissions.'))
  }
}
```

---

## Summary

The most critical issues are:
1. **Logic error** in `inviteRevoke` allowing anyone to revoke invites
2. **Missing authorization** in `updatePermissions` allowing any member to modify permissions
3. **No validation** of permission values allowing arbitrary strings
4. **Cross-contract access** issue in chatroom delete validation

I recommend addressing issues #1-5 before merging as they represent security vulnerabilities.


</details>

### Review 2 (less comprehensive context)

<details>


## 1. Critical Bug: Wrong Operator in `inviteRevoke` Validation

```javascript
if (!invite.creatorID !== innerSigningContractID && !myPermissions.includes(GROUP_PERMISSIONS.REVOKE_INVITE)) {
```

**Problem**: The `!` operator has higher precedence than `!==`, so this evaluates as `(!invite.creatorID) !== innerSigningContractID`, which converts `invite.creatorID` to a boolean first, then negates it. This is not the intended logic.

**Fix**:
```javascript
if (invite.creatorID !== innerSigningContractID && !myPermissions.includes(GROUP_PERMISSIONS.REVOKE_INVITE)) {
  throw new TypeError(L('You do not have permission to revoke this invite.'))
}
```

---

## 2. Security Issue: Missing Permission Check in `updatePermissions`

```javascript
'gi.contracts/group/updatePermissions': {
  validate: actionRequireActiveMember(arrayOf(
    objectMaybeOf({
      memberID: stringMax(MAX_HASH_LEN, 'memberID'),
      action: validatorFrom(x => Object.values(GROUP_PERMISSION_UPDATE_ACTIONS).includes(x)),
      roleName: stringMax(GROUP_ROLE_NAME_MAX_CHAR, 'roleName'),
      permissions: arrayOf(stringMax(GROUP_PERMISSION_MAX_CHAR, 'permissions'))
    })
  )),
```

**Problem**: The validation only checks data structure, but doesn't verify if the caller has `DELEGATE_PERMISSIONS` permission. Any active member could modify permissions.

**Fix**:
```javascript
'gi.contracts/group/updatePermissions': {
  validate: actionRequireActiveMember((data, { getters, message: { innerSigningContractID } }) => {
    // Validate data structure
    arrayOf(
      objectMaybeOf({
        memberID: stringMax(MAX_HASH_LEN, 'memberID'),
        action: validatorFrom(x => Object.values(GROUP_PERMISSION_UPDATE_ACTIONS).includes(x)),
        roleName: stringMax(GROUP_ROLE_NAME_MAX_CHAR, 'roleName'),
        permissions: arrayOf(stringMax(GROUP_PERMISSION_MAX_CHAR, 'permissions'))
      })
    )(data)

    // Check caller has permission to delegate
    const myProfile = getters.groupProfile(innerSigningContractID)
    const myPermissions = myProfile?.role?.permissions || []
    if (!myPermissions.includes(GROUP_PERMISSIONS.DELEGATE_PERMISSIONS)) {
      throw new TypeError(L('You do not have permission to update permissions.'))
    }

    // Prevent modification of admin role
    for (const item of data) {
      const targetProfile = getters.groupProfile(item.memberID)
      if (targetProfile?.role?.name === GROUP_ROLES.ADMIN) {
        throw new TypeError(L('Cannot modify admin permissions.'))
      }
      // Prevent assigning MODERATOR_DELEGATOR without ASSIGN_DELEGATOR permission
      if (item.roleName === GROUP_ROLES.MODERATOR_DELEGATOR && 
          !myPermissions.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR)) {
        throw new TypeError(L('You do not have permission to assign the delegator role.'))
      }
    }
  }),
```

---

## 3. Bug: `throw` in `process` Function

```javascript
  for (const item of data) {
    const groupProfile = state.profiles[item.memberID]

    if (!groupProfile) {
      throw new Error(L('Member does not exist.'))
    }
```

**Problem**: Throwing an error inside a `process` function after contract state has potentially been modified can leave the contract in an inconsistent state. All validation should happen in the `validate` function.

**Fix**: Move the member existence check to the `validate` function:
```javascript
validate: actionRequireActiveMember((data, { state, getters, message: { innerSigningContractID } }) => {
  // ... other validations ...
  
  for (const item of data) {
    if (!state.profiles[item.memberID]) {
      throw new TypeError(L('Member does not exist.'))
    }
  }
}),
  for (const item of data) {
    const groupProfile = state.profiles[item.memberID]
    // No need to check existence here, validated above
    
    switch (item.action) {
      // ...
    }
  }
}
```

---

## 4. Bug: `getters.groupProfile` Doesn't Exist in Chatroom Contract

```javascript
// In chatroom.js
'gi.contracts/chatroom/delete': {
  validate: actionRequireInnerSignature((_, { state, getters, meta, message: { innerSigningContractID } }) => {
    const myProfile = getters.groupProfile(innerSigningContractID)
    const myPermissions = myProfile?.role?.permissions || []
```

**Problem**: The `groupProfile` getter is defined in `group.js`, not in `chatroom.js`. This will cause an error when trying to delete a channel.

**Fix**: You need to pass the permission information differently, perhaps through the action data, or access the group state through a different mechanism. One approach:

```javascript
'gi.contracts/chatroom/delete': {
  validate: actionRequireInnerSignature((data, { state, getters, meta, message: { innerSigningContractID } }) => {
    // If not the creator, the action data should include proof of permission
    // which should be validated through the group contract
    if (state.attributes.creatorID !== innerSigningContractID) {
      // Need to validate permission through group contract state
      // This requires architectural consideration
      if (!data?.hasDeletePermission) {
        throw new TypeError(L('You do not have permission to delete this channel.'))
      }
    }
  }),
```

---

## 5. Security Issue: Users Can Modify Their Own Permissions

```javascript
// In PermissionTableRow.vue
td.td-action(v-if='!isAdmin && !isMe && canDelegatePermissions')
```

**Problem**: The frontend hides the action menu for self-modification, but the contract doesn't enforce this. A malicious user could craft a direct contract call to elevate their own permissions.

**Fix**: Add to the `validate` function in `updatePermissions`:
```javascript
// Prevent users from modifying their own permissions
if (data.some(item => item.memberID === innerSigningContractID)) {
  throw new TypeError(L('You cannot modify your own permissions.'))
}
```

---

## 6. Potential Null Reference in `GroupMembersDropdown`

```javascript
allActiveMembers () {
  return Object.keys(this.profilesByGroup(this.groupID || this.currentGroupId))
    .map(memberId => this.globalProfile(memberId))
}
```

**Problem**: `globalProfile(memberId)` might return `null` or `undefined` for some members.

**Fix**:
```javascript
allActiveMembers () {
  return Object.keys(this.profilesByGroup(this.groupID || this.currentGroupId))
    .map(memberId => this.globalProfile(memberId))
    .filter(Boolean)  // Filter out null/undefined profiles
}
```

---

## 7. Minor Issue: Arbitrary Wait in Cypress Test

```javascript
cy.wait(3 * 1000) // eslint-disable-line cypress/no-unnecessary-waiting
```

**Problem**: Arbitrary waits make tests flaky and slow.

**Suggestion**: Replace with a proper assertion wait:
```javascript
cy.getByDT('some-element-that-indicates-ready').should('exist')
```

---

## 8. Missing Default Case in Switch Statement

```javascript
switch (item.action) {
  case 'add':
    // ...
    break
  case 'edit':
    // ...
    break
  case 'remove':
    // ...
    break
}
```

**Suggestion**: Add a default case for robustness:
```javascript
switch (item.action) {
  case 'add':
    // ...
    break
  case 'edit':
    // ...
    break
  case 'remove':
    // ...
    break
  default:
    console.warn(`Unknown permission update action: ${item.action}`)
}
```

---

## Summary

The most critical issues are:
1. **Line 970 in group.js**: Wrong operator causing incorrect permission check for invite revocation
2. **Missing permission validation** in `updatePermissions` - major security hole
3. **Error thrown in process function** - can corrupt contract state
4. **Missing getter** in chatroom contract for group permissions
</details>

SebinSong:
@taoeffect Updated this PR per the LLM's feedback.

Both reviews were helpful in finding some contract validations I missed but I felt 'Review 2 (less comprehensive context)' were a little better, especially 3. 4. there.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,66 @@ sbp('chelonia/defineContract', {
...
+
+        if (data.some(item => item.permissions?.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR) &&
+          myProfile?.role?.name !== GROUP_ROLES.ADMIN)) {
+          // Only the group admin can assign the delegator role
+          throw new TypeError(L('You do not have permission to assign delegator role.'))
+        }
+
+        if (!myPermissions.includes(GROUP_PERMISSIONS.DELETE_CHANNEL)) {
+          throw new TypeError(L('You do not have permission to delegate permissions.'))
+        }
```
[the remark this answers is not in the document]
SebinSong: Good catch.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,66 @@ sbp('chelonia/defineContract', {
...
+              groupProfile.role = {
+                name: item.roleName,
+                permissions: item.permissions
+              }
+              break
+            case 'edit':
+              groupProfile.role = {
+                name: item.roleName || groupProfile.role.name,
+                permissions: item.permissions || groupProfile.role.permissions
+              }
```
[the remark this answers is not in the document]
SebinSong: Have added a validation check to ensure `edit` action is only performed on the user who has a role.

SebinSong:
@taoeffect The PR is ready for review again

test/cypress/integration/group-chat-message.spec.js
```diff
@@ -448,7 +448,7 @@ describe('Send/edit/remove/reply/pin/unpin messages & add/remove reactions insid
       cy.getByDT('conversationWrapper').scrollTo('top')
       // Wait period is due to debouncing in onChatScroll
       // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(201)
+      cy.wait(1001)
```
corrideat: I've found leaving this as is but increasing the loop count to 4 helps. If that works, it's preferrable because it's faster.

corrideat:
As a preliminary review, this PR looks well thought out and implemented. I'll be testing it and reviewing the changes more carefully for a more thorough review. However, it looks quite solid so far. Well done!

<details>
<summary>Some miscellaneous comments about the approach used, which apply to the issue and _not_ to this PR</summary>
There's an issue that I'm leaving here for completeness / referencing purposes, but which is out of scope for this PR (but not necessarily out of scope for the issue it's addressing). That is that permissions here are implemented strictly within actions.

While that is fine here, some permissions that affect keys need to be done with key operations as well in order to properly enforce them (invite revocation comes to mind; the actual invite is a key added with `OP_KEY_ADD`) and invites can be used without involving the contract at all (OTOH, the new changes to `@chelonia/lib` offer a way of doing this with a method for more proper enforcement, but that's still imperfect because the invite could be revoked.)

How to integrate permissions with keys can be left as a follow up issue or PR.
</details>

test/cypress/integration/group-chat-message.spec.js
SebinSong: changing it to `4` seems to fix the heisenbug. Updated.

frontend/model/contracts/chatroom.js
```diff
@@ -278,9 +278,9 @@ sbp('chelonia/defineContract', {
       }
     },
     'gi.contracts/chatroom/delete': {
-      validate: actionRequireInnerSignature((_, { state, meta, message: { innerSigningContractID } }) => {
-        if (state.attributes.creatorID !== innerSigningContractID) {
-          throw new TypeError(L('Only the channel creator can delete channel.'))
+      validate: actionRequireInnerSignature((data, { state, getters, meta, message: { innerSigningContractID } }) => {
+        if (state.attributes.creatorID !== innerSigningContractID && !data?.byPermission) {
```
corrideat: This `.byPermission` attribute seems to come from `myPermissions.includes(GROUP_PERMISSIONS.DELETE_CHANNEL)`, which can be forged by anyone. Now, I see that keeping state in sync between contracts can be a bit tricky, but, ideally, we'd want to actually check that the permission exists.

Is this intentional, and meant to be enforced some other way in the future? If so, having a TODO comment explaining would be nice.

One way of doing this is using side effects to keep permissions in sync between groups and chatrooms, which would actually also need to be enforced with signing keys and foreign signing keys.

When it comes to deleting things, there's the `/delete` action and there's the actual deletion, which requires access to a deletion token or to the `#sak`. We also need to refactor the accounting logic in the server to allow for actual deletions based on permissions.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,72 @@ sbp('chelonia/defineContract', {
...
+        }
+
+        for (const item of data) {
+          if (!state.profiles[item.memberID]) {
+            throw new TypeError(L(`${item.memberID} is not a member of the group.`))
+          }
+
+          if (item.memberID === innerSigningContractID) {
+            throw new TypeError(L('You cannot modify your own permissions.'))
+          }
```
corrideat, line 1146: IMHO it should be fine to edit your own permissions, so long as you're dropping the permissions you have.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,72 @@ sbp('chelonia/defineContract', {
...
+          throw new TypeError(L('Cannot assign admin role.'))
+        }
+
+        if (data.some(item => item.permissions?.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR) &&
+          myProfile?.role?.name !== GROUP_ROLES.ADMIN)) {
+          // Only the group admin can assign the delegator role
+          throw new TypeError(L('You do not have permission to assign delegator role.'))
+        }
+
+        if (!myPermissions.includes(GROUP_PERMISSIONS.DELEGATE_PERMISSIONS)) {
```
corrideat: Two comments here.

## First issue

The first one, which is perhaps just terminology (I see this is how it was used in #202), is that this should be called `ASSIGN_PERMISSIONS`, `GRANT_PERMISSIONS` or `GIVE_PERMISSIONS` or something like that. `DELEGATE_PERMISSIONS`, to me at least, implies that the permissions given depend on whomever is doing the delegating having those permissions. Based off the [definition of 'delegate'](https://www.google.com/search?q=Dictionary#dobs=delegate), _entrust (a task or responsibility) to another person_, this seems like a reasonable interpretation.

For example, if `u1` gives `u2` the delegate permission, and then `u2` delegates the view permission to `u3`, I would expect that, if `u2`'s view permission was later revoked, then `u3` would no longer have the ability to view either, because `u3`'s view permission was delegated by `u2` (who 'owned' this permission, if you will). Now, the way that this _currently_ works is that `u3`'s permissions are independent from what happens to `u2` in the future, which probably is good design, but just confusing terminology.

So, I wouldn't change how it works, but I'd change the name to use something that more clearly reflects what it does.

## Second issue

The second issue is that, if I'm understanding this code correctly, the delegate permission in itself is enough to grant someone else a permission the delegator may not have. I realise that in the current implementation the 'moderator - delegator` role has all permissions, but nevertheless a check wouldn't hurt.

<ins>I've just tested this, and it's possible to just have 'delegate' permission and then delegate other permissions.</ins>

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,72 @@ sbp('chelonia/defineContract', {
...
+          throw new TypeError(L('You do not have permission to delegate permissions.'))
+        }
+
+        if (data.some(item => {
+          return item.action === GROUP_PERMISSION_UPDATE_ACTIONS.EDIT && !state.profiles[item.memberID]?.role
+        })) {
+          throw new TypeError(L('Cannot edit permissions for a member who does not have a role.'))
+        }
```
corrideat: Is there a good reason to have differentiated actions for _editing_ a permission and for _adding_ a permission? Since the two actions are very similar, as an outsider to the discussion that resulted in this PR, an [upsert](https://wiki.postgresql.org/wiki/UPSERT#%22UPSERT%22_definition) strategy would be better.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1131,72 @@ sbp('chelonia/defineContract', {
...
+      }),
+      process ({ data }, { state }) {
+        for (const item of data) {
+          const groupProfile = state.profiles[item.memberID]
+
+          switch (item.action) {
+            case GROUP_PERMISSION_UPDATE_ACTIONS.ADD:
+              groupProfile.role = {
+                name: item.roleName,
+                permissions: item.permissions
```
corrideat: If I'm understanding this correctly, the role name and the permissions are stored separately, and the list of permissions is what matters, regardless of the role name.

I suggest not doing this, even if it makes checking for permissions slightly harder. One of the advantages of having roles is that they can change definitions over time. So, for example, the following is what should happen IMHO:

1. If the custom role is used and permissions A, B and C are given, then one should have exactly those permissions.
2. If a preset is used, containing permissions A and B, the actual permissions should depend on the preset at the current time. For example, in the future the same preset could have permissions _A, B and C_, or it could have just _A_, or it could have permissions _C and D_. The point in all of this is that the role is more useful if it can be dynamically changed.

corrideat (changes requested):
I've made a more thorough review of the code (focusing on the contracts logic mostly, for now I haven't made a thorough review of the UI code), and I have some suggestions for how things should work.

As in my previous review, the implementation looks functional and thorough but nevertheless has some aspects that might be improved upon. In summary:

1. The chatroom delete permission is pretty easy to circumvent. This could be fine for now, but at least a TODO should be added.
2. I'd rename the 'delegate' permission (despite this term being used in #202) to reflect what the code is actually doing.
3. I'd add checks to see that the permission being 'delegated' is actually a permission one has.
4. I'd also make roles work with 'dynamic' permissions, so that the permissions aren't burnt into a profile with the contemporary definition of the role.
5. The distinction between 'adding' and 'editing' a permission _in the contract_ (i.e., not in the UI) seems redundant.
6. In the UI, the 'custom' role doesn't seem to allow for granting the 'delegate' permission. I'd add this option (which, with item 3 above, would make it harmless).

frontend/model/contracts/group.js
SebinSong: There was Greg's request that only admin and a person with a delegator role to add/edit permissions. If someone can change their own permissions, there is no point of having roles with these privileges..

(won't be changing logics based on these kind of opinions unless Greg confirms it.)

frontend/model/contracts/group.js
SebinSong: The Term `DELEGATE_PERMISSIONS` for the ability to grant/edit/remove permissions was suggested by Greg and also got confirmed in the first PR of this issue. (can't find the exact comment for this but) I recall having `ADD_PERMISSIONS` separately initially but then was asked to use the term `DELEGATE_PERMISSIONS` for grant/edit/remove actions. If he thinks differenly now and will change it.

---
Have mentioned this in another comment too but below was Greg's request re `delegate-permission`:

- `DELEGATE_PERMISSIONS` are exclusively given to `ADMIN` and `MODERATOR - DELEGATOR` roles.
- Admin is the only role that can assign `MODERATOR - DELEGATOR` role.
- `MODERATOR - DELEGATOR` can add/edit/remove permissions of other members(except for himself/admin) too but that doesn't include `DELEGATE_PERMISSIONS` permission item.
- By doing this, only admin and 'moderator - delegator' roles have the ability to  grant/edit/remove permissions.

Current logic works as expected according to this.

>  the delegate permission in itself is enough to grant someone else a permission the delegator may not have.  I realise that in the current implementation the 'moderator - delegator` role has all permissions, but nevertheless a check wouldn't hurt.

Will leave it as is for now, but once this check becomes needed/necessary in the future, will add the check.

frontend/model/contracts/chatroom.js
SebinSong: It's a fair concern. Will look into this part again or leave a TODO comment as suggested.

frontend/model/contracts/chatroom.js
```diff
@@ -278,9 +279,9 @@ sbp('chelonia/defineContract', {
       }
     },
     'gi.contracts/chatroom/delete': {
-      validate: actionRequireInnerSignature((_, { state, meta, message: { innerSigningContractID } }) => {
-        if (state.attributes.creatorID !== innerSigningContractID) {
-          throw new TypeError(L('Only the channel creator can delete channel.'))
+      validate: actionRequireInnerSignature((data, { state, getters, meta, message: { innerSigningContractID } }) => {
+        if (state.attributes.creatorID !== innerSigningContractID && !getters.ourGroupPermissionsHas(GROUP_PERMISSIONS.DELETE_CHANNEL)) {
```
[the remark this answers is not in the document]
SebinSong: Hi @taoeffect ,
I've got a question about this `gi.contracts/chatroom/delete` action in `chatroom.js` contract.

It currently has a validation that checks if this action is performed by a group creator:
`state.attributes.creatorID !== innerSigningContractID`

But now that we are implementing `DELETE_CHANNEL` permission, this validation needs to be extended with checking if a person has this permission item.

The problem here is:
- Permissions are stored in group member profile. So `group.js` contract state needs to be loaded in this block, such as using a getter like `getters.ourGroupPermissionsHas(GROUP_PERMISSIONS.DELETE_CHANNEL)`.
- But as Devin pointed out here too, apparently `getters` in `(data, { state, getters, meta, ... }) => ...` only contain chatroom contract getters.
- Things like `sbp('state/vuex/getter')` can be useful in this case but I remember you mentioned a while ago(my memory might not be accurate but) calling Vuex states in `validation` and `process` blocks of contract actions will crash the app in certain environments.

**Q.** Would you think ensuring these checks(whether the member is a group creator or has `DELETE_CHANNEL` permission) happen where this action is called would suffice?
Currently there are two places:
- `sideEffect` block of `gi.contracts/identity/deleteDirectMessage` action.
- `sideEffect` block of `gi.contracts/group/deleteChatRoom` action.

In both places, there is no problem calling permission-related getters.

frontend/model/contracts/group.js
taoeffect: Regarding @corrideat's first concern, let's leave it named as-is for now. I don't think the word delegate necessarily contains within it a strict interpretation of what should be done with delegated permissions if the delegator themselves loses them.

Regading the second concern, this is a legitimate issue:

> I've just tested this, and it's possible to just have 'delegate' permission and then delegate other permissions.

Delegators should only be able to delegate permissions that they have, so we should fix that.

frontend/model/contracts/chatroom.js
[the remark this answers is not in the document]
taoeffect: > Would you think ensuring these checks(whether the member is a group creator or has DELETE_CHANNEL permission) happen where this action is called would suffice?

Yes, that sounds reasonable, but what are you going to do about the existing permissions check here? It will still reject the message if it comes from anyone other than the channel creator?

Consider creating a copy of the permissions on the chatroom contract.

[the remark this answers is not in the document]
corrideat: `ourGroupPermissionsHas` or such cannot safely be used within contracts, for two reasons:

1. Information about the current group doesn't exist in the SW context.
2. The group contract might be out-of-date compared to the chatroom contract, depending on sync order.

Overall, @taoeffect's suggestion of copying permissions to the chatroom sounds better.

frontend/model/contracts/group.js
corrideat: I'm not sure why this was marked as resolved since the issue hasn't been resolved. Roles should apply current role permissions, not 'role permissions at the time the role was assigned'.

If 'role permissions at the time the role was assigned' is used, then there's no need to store the 'role' in the contract (since it's not relevant) and then the UI should be updated to dynamically display a role based on actual permissions. For example, if I'm assigned the 'moderator role' with permissions A, B and C, and later on the the 'moderator role' has permissions A, B, C and D, I shouldn't be shown as a 'moderator' in the UI because I no longer have the permissions this role entails.

frontend/model/contracts/group.js
corrideat: I'm also not sure why this is marked as resolved, but it still seems unclear why we need have _editing_ and _adding_ be different things? (Note that this applies to the contract only. Having the UI have different workflows in these cases is not what this comment is about.)

frontend/model/contracts/chatroom.js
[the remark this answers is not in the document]
SebinSong: @taoeffect 
> "creating a copy of the permissions on the chatroom contract"

Did you mean copying the user's permissions and sending as a payload to `gi.contracts/chatroom/delete`, so that they can be accessed via `data` in the validation block?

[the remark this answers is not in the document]
corrideat: > Did you mean copying the user's permissions and sending as a payload to gi.contracts/chatroom/delete, so that they can be accessed via data in the validation block?

What we do for other similar fields (e.g., members) is keep contracts in sync with `sideEffects` so that changes on one contract are copied over to the other contract. This keeps contracts mostly in sync with one another.

If you were to send a copy of group permissions in `gi.contracts/chatroom/delete`, it wouldn't be too much different to `byPermission: true` because there's no way to verify that the payload you're sending is what the group contract really has.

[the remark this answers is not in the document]
taoeffect: > Did you mean copying the user's permissions and sending as a payload to `gi.contracts/chatroom/delete`, so that they can be accessed via `data` in the validation block? 

I mean copying over the relevant part of the permissions state object from the group contract to the chatroom contract state so that the chatroom permissions are accessed and work in the same way.

For example, you can look at other apps and see that when you view the member list for a chatroom in those apps, they'll have a little tag next to member names (e.g. `admin`, `mod`, etc.). We can do the same.

So think deeply about what permissions you'd like to have access to in the chatroom, and either have those permissions entirely there (no need to copy in that case), or have a duplicate (in this case, any permissions changes must be synchronized between the group and the chatroom).

Consider, for example, that someone could create a private chatroom, and in that case they are essentially the "admin" of that chatroom and have special privileges, even though they aren't a group admin.

frontend/model/contracts/group.js
SebinSong: > Note that this applies to the contract only. Having the UI have different workflows in these cases is not what this comment is about.

Yes adding/editing happen in two different places in the UI. upsert is a good suggestion too but
let me still leave it as is though, becuase 'adding' something always happens first before 'editting` and there might be cases where knowing it's either adding or editing is an important/useful bit of information (especially in the debugging).

frontend/model/contracts/group.js
SebinSong: I see what the potential issue mentioned now. (Sorry I marked it resolved as I didn't see an issue in the current logic at the time)
So it seems the mentioned concern is something like below case:

user1 is given a moderator role -> later, we decide to define moderator role to have more/less permissions, so update `moderator` preset -> then user1 has stale definition of moderator

If correct, I agree this is a legit issue.

There seems to be no need to store `roleName` info in the user profile. The UI can determine which role the member has by comparing their permissions and the presets of the roles (If no matches, then `custom`).

will change from:
```
{
  role: ...
  permissions: [ ... ]
}
```
to just:
```
permissions: [ ... ]
```

frontend/model/contracts/chatroom.js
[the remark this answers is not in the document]
SebinSong: @taoeffect 

> No. I mean copying over the relevant part of the permissions state object from the group contract to the chatroom contract so that the chatroom permissions are accessed and work in the same way.

If that is the same as what Ricardo said in the above:

> What we do for other similar fields (e.g., members) is keep contracts in sync with sideEffects so that changes on one contract are copied over to the other contract. This keeps contracts mostly in sync with one another.

Would you know where in the source code this happens? 
I haven't had to deal with this case so far in the project, so any existing example will be helpful.

Or since `DELETE_CHANNEL` is only relevant permission to chatroom (at least for now), We could add a field named something like `membersWithDeletePermission` in the chatroom attributes and keep it up-to-date whenever `DELETE_CHANNEL` permission related change happens.:



---
> For example, you can look at other apps and see that when you view the member list for a chatroom in those apps, they'll have a little tag next to member names (e.g. admin, mod, etc.). We can do the same.

BTW, since this is an UI only implementation(where there is no problem accessing group-profile related getters), chatroom contract update isn't neccessary for this particular one.

This can be worked on as a follow-up issue of this PR once the issue for this is created with task details.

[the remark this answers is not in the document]
taoeffect: > Would you know where in the source code this happens?
> I haven't had to deal with this case so far in the project, so any existing example will be helpful.

I would use a tool like [crush](https://github.com/charmbracelet/crush/) to ask this question in this codebase. You'll need to learn how to explain things to it though (the better you phrase the question, the better it will be able to give you an answer).

The general idea is:

1. When setting permissions in the group via an action, add a sideEffect (if it doesn't exist) and use chelonia/queueInvocation on the contractID to cause something to happen after a the end of the sync (this is a common pattern in group.js sideEffects).
2. Have the "something" be a check that if (and only if) the relevant chatroom doesn't have the latest permissions copied over, then emit an action to that contract to copy them over.
3. Implement the action on the chatroom to copy over the actions. Make sure that it works fine if it's called multiple times with the same permissions.

frontend/model/contracts/group.js
corrideat: Correct, staleness is the main issue.

See with @taoeffect how this should be addressed (there are two ways).

I'd find the static permissions solution (with the UI deciding dynamically a role) acceptable but suboptimal. It's suboptimal in the sense that it's not how roles typically work (they tend to be a set and forget thing).

My preferred solution would be storing the role (as you are doing) but not permissions (except for the custom permissions of course), and checking based on whether the user has the permissions _or_ a role. I'd keep the ability to have both permissions and roles for future compat too, but can be mutually exclusive too. For example, something like: `permissions = [...userPermissions, ...rolePermissions]`.

frontend/model/contracts/group.js
corrideat: The point about debugging is well taken.

However, the concern here is that it results in mostly duplicated logic that so far looks unnecessary. If at some point we changed the UI, we'd have to add extra checks to accommodate for this design choice in the contract. Likewise, if we implemented a 'headless' version of the app, we need extra checks to see if we're adding or editing permissions. This check doesn't seem to be useful except for deciding which action / parameters to use.

This is by no means a critical issue, but I'd prefer a simpler approach over a more complicated approach if I could choose and the result was the same.

frontend/model/contracts/group.js
SebinSong: I've already taken some time working on this with :

```
permissions: [ ... ]
```

which is a large change as it requires changing and checking every parts that are related to this all over again...

Will go with this solution unless there is a reason to revert this back found.

corrideat: There are good reasons to go with the other approach:

1. It's how roles typically work
2. It's simpler to implement
3. It avoids having users regularly update permissions to sync them with roles or having to write migration code to do this automatically.

I'll give my approval regardless of which of the two approaches you go with, but please check with @taoeffect whether 'static permissions' is how this feature was envisioned.

SebinSong: Actually, Is this something that you are suggesting?

roles with preset:
```
role: { name: string }
```

Custom role:
```
role: { name: 'custom', permissions: string[] }
```

This is actually something I had in mind initally too because having role name makes certain things in the UI layer easier.

(EDIT: missed the last comment above, didn't see it getting updated while I'm writing this)

corrideat: Sort of, yes, that works.

For maximum flexibility, you'd have both 'role' and 'permissions' (kind of like how it was originally; the point then was mostly about how permissions were checked). However, we might not need that flexibility and it's something that can be added later.

Going back to your question then, yes. I'd have no objections to that approach, and as you say it'd simplify things on the UI side.

(There's still the possibility that this truly was intended as 'static permissions', but I doubt that, and even then we can always define new presets instead of changing existing ones)

SebinSong: Thanks for confirmation @corrideat . wanted to go with the structure that makes sense to you too. Will make updates accordingly.

frontend/model/contracts/chatroom.js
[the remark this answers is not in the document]
SebinSong: @taoeffect  I have worked on almost all change requests (haven't pushed commits yet) other than this one and spent some time on this one too.
Since the main purpose of this PR is about implementing 'add/remove/editing' permissions, is it alright for me to work on implementing `DELETE_CHANNEL` permission as a separate issue after this PR?
This actually requires broader updates/attentions in other parts too besides what you requested here:

Not only taking care of when `DELETE_CHANNEL` permission is either added/removed, We also need to take care of when a user with this permission enters/leaves a chatroom and gets kicked out from the group etc..

I feel this one is a topic that deserves a dedicated focus and it would be helpful to me in development as well that way..!

If that sounds okay, will leave a todo comment for `DELETE_CHANNEL` permission in the source code.

frontend/model/contracts/group.js
SebinSong: makes sense.
I actually ended up merging adding/editing into upsert while making [this change](https://github.com/okTurtles/group-income/pull/2895#discussion_r2790662950) as required logics for the two actions are almost identical inside the `process` block of the contract action.

But there is still a need in the UI layer where they do need to be distinguished from each other. For example the success feedback messages are different: 'add' action -> 'role added successfully!' , for 'edit' action -> 'permission details are updated!' etc..
So some logics that treat them separately still remain in the UI side.

FYI, my development workflow in GI tends to be: what should we build in the UI? -> what are the required logics for them? -> ...

So sometimes, the need in the UI layers flow into the contracts as well.

Anyways, it's a good stuff to note. Thanks for it.

SebinSong:
PR ready for review again.
As commented [here](https://github.com/okTurtles/group-income/pull/2895#discussion_r2791297327), have left todo comments `DELETE_CHANNEL` in various places. Will pick them up as a follow-up after this PR is merged.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1125,71 @@ sbp('chelonia/defineContract', {
...
+
+          if (myRoleName === GROUP_ROLES.MODERATOR_DELEGATOR &&
+            data.some(item => item.permissions.some(permission => !myPermissions.includes(permission)))) {
+            throw new TypeError(L("You(moderator-delegator) cannot assign permissions that you don't have to others."))
+          }
+
+          if (data.some(item => item.roleName === GROUP_ROLES.ADMIN)) {
+            throw new TypeError(L('Cannot assign admin role.'))
+          }
+        }
```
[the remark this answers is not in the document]
SebinSong: Legit mistake..
Will get back to this tomorrow.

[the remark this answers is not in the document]
SebinSong: These checks are now correctly done against each item. Resolving it.

SebinSong:
The PR is ready for review again.

taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/group-income/actions/runs/22207627149)._

taoeffect:
/review_fast

---

_AI review [started](https://github.com/okTurtles/group-income/actions/runs/22207739215)._

taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/group-income/actions/runs/22207923749)._

taoeffect:
@SebinSong Above are two AI reviews using the same model but different environments (opencode vs crush), please let me know which one did better, or if they were about the same.

taoeffect:
/crush_fast (Sorry @SebinSong, I remembered that there's a bug in crush's default max context size for GLM-5, it's about half of what it should be. That's why I'm re-running the crush review as I've added a fix for that.)

---

_AI review [started](https://github.com/okTurtles/group-income/actions/runs/22210161153)._

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1125,69 @@ sbp('chelonia/defineContract', {
...
+
+            if (item.permissions?.includes(GROUP_PERMISSIONS.ASSIGN_DELEGATOR) && myRoleName !== GROUP_ROLES.ADMIN) {
+              throw new TypeError(L('You do not have permission to assign delegator role.'))
+            }
+
+            if (myRoleName === GROUP_ROLES.MODERATOR_DELEGATOR &&
+              item.permissions?.some(permission => !myPermissions.includes(permission))) {
+              throw new TypeError(L("You(moderator-delegator) cannot assign permissions that you don't have to others."))
+            }
+          }
```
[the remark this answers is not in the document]
SebinSong: This has been resolved too if you haven't noticed already. Marking it resolved.

SebinSong:

Personally It actually didn't feel like one was better than the other. They performance seem equal and also have several overlapping items.

But it _feels_ like `glm-5` doesn't perform reviews with broad/deep investigation in comparison to Devin or other LLM models you've used. For example:

it had multiple items(in both `opencode`/`crush` reviews) where it says about possible UI runtime error caused by `profile.role` potentially being `null`.
But I've dismissed all these items because `profile.role` being `null` is properly guarded in the parent component of all `.vue` files it raised this concern about. (doesn't seem to check something with considering the context it is used as well)

Also, there was one item(`issue 8` in [this review](https://github.com/okTurtles/group-income/pull/2895#issuecomment-3931104636)) where it says something that is not true (there is no other Vue lifecycle hooks other than `created()` in both files), and the suggested fix has an anti Vue pattern(`return` is not supposed to be used in `created()` hook).

But the rest items were helpful in finding several unused codes and typos.

Thanks,

taoeffect:
/crush_fast {{openrouter, google/gemini-3.1-pro-preview, google/gemini-3-flash-preview}}

Thanks @SebinSong! I'm gonna try one more model that's supposed to be better than GLM-5 just in case it's able to find some things those missed. I'll also double-check with @corrideat to see if there's anything he thinks we should wait for before merging this.

EDIT: (I will also test these changes manually before merging as well)

---

_AI review [started](https://github.com/okTurtles/group-income/actions/runs/22336843698)._

frontend/model/contracts/shared/getters/group.js
```diff
@@ -124,6 +126,43 @@ export default ({
...
+    }
+  },
+  getGroupMemberPermissionsById (state, getters) {
+    // Pass the group memberID and get all the permissions that the member has if any.
+    return (memberID) => {
+      const profile = getters.groupProfiles[memberID]
+      if (!profile?.role) return []
+      return profile.role.name === GROUP_ROLES.CUSTOM
+        ? profile.role.permissions
+        : GROUP_PERMISSIONS_PRESET[profile.role.name]
```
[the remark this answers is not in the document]
SebinSong: > .. When updatePermissions is called with action: 'upsert' and roleName: 'custom' but without a permissions field ..

The submit button in add/update permissions UI is never enabled if no permission items are selected with 'custom' role selected.. Dismissing this comment.

frontend/views/containers/group-settings/roles-and-permissions/RolesAndPermissions.vue
```diff
@@ -16,107 +15,93 @@ page-section.c-section(
...
     },
-    canDelegatePermissions () {
-      return this.myPermissions.includes(GROUP_PERMISSIONS.DELEGATE_PERMISSIONS)
+    groupPermissionsToDisplay () {
+      const myEntry = this.allGroupMemberRolesAndPermissions.find(entry => entry.memberID === this.ourIdentityContractId)
+      const otherEntries = this.allGroupMemberRolesAndPermissions.filter(entry => entry.memberID !== this.ourIdentityContractId)
+      const sortedEntries = myEntry ? [myEntry, ...otherEntries] : this.allGroupMemberRolesAndPermissions
+      return this.canViewOtherMembersPermissions
+        ? sortedEntries
+        : [myEntry]
```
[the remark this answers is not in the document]
SebinSong: updated

frontend/model/contracts/shared/getters/group.js
```diff
@@ -124,6 +126,43 @@ export default ({
...
+  },
+  getAllMemberIdsWithPermission (state, getters) {
+    // Get all members with a specific permission as a list of { roleName: string, permissions: string[], memberID: string }.
+    // Maybe make this getter takes an array of permissions(instead of a single permission) in the future if there is a need for it.
+    return (permission) => {
+      return getters.allGroupMemberRolesAndPermissions.filter(({ permissions }) => permissions.includes(permission))
+        .map(({ memberID }) => memberID)
+    }
+  },
+  getGroupMemberPermissionsById (state, getters) {
```
corrideat: Nitpick, but the name here seems confusing (to me, at least).  `getGroupPermissionsByMemberId` would more clearly convey what it does. The `ById` part makes one wonder which `Id` it's referring to.

frontend/model/contracts/group.js
```diff
@@ -809,6 +815,7 @@ sbp('chelonia/defineContract', {
         const memberToRemove = data.memberID || innerSigningContractID
         const membersCount = getters.groupMembersCount
         const isGroupCreator = innerSigningContractID === getters.currentGroupOwnerID
+        const myPermissions = getters.getGroupMemberPermissionsById(innerSigningContractID)
```
corrideat: Name here too. The logic below seems fine, but I'd call this `permissions` or `actorPermissions` / `signerPermissions` / etc. Generally, we use _me_, _my_, etc. for things that match the current user, and, in this case, everyone is running the check to determine if the action is valid.

frontend/model/contracts/group.js
```diff
@@ -953,14 +966,24 @@ sbp('chelonia/defineContract', {
...
+        if (!invite) {
           throw new TypeError(L('The link does not exist.'))
         }
+
+        const myPermissions = getters.getGroupMemberPermissionsById(innerSigningContractID)
+        const inviteCreatorID = state.invites?.[data.inviteKeyId]?.creatorID
+
+        // Only the creator and members with the REVOKE_INVITE permission can revoke an invite
+        if (inviteCreatorID !== innerSigningContractID && !myPermissions.includes(GROUP_PERMISSIONS.REVOKE_INVITE)) {
+          throw new TypeError(L('You do not have permission to revoke this invite.'))
```
corrideat: Also here, this is a check everyone runs, so 'You' is misleading.

As an alternative, you can say something like: 'Insufficient permission to revoke invite' or 'Missing invite revoke permission', or some other phrasing that's impersonal. This way, anyone getting this error in the logs can know why, without suggesting that it applies to the person getting the error.

frontend/model/contracts/group.js
```diff
@@ -1106,6 +1129,91 @@ sbp('chelonia/defineContract', {
...
+          throw new TypeError(L('You do not have permission to delegate permissions.'))
+        }
+
+        for (const item of data) {
+          if (!state.profiles[item.memberID]) {
+            throw new TypeError(L(`${item.memberID} is not a member of the group.`))
+          }
+
+          if (item.memberID === innerSigningContractID) {
+            throw new TypeError(L('You cannot modify your own permissions.'))
```
corrideat: And here. An alternative here could be: 'Cannot modify own permissions'

corrideat (commented):
This looks much better. Thank you for addressing the feedback.

One thing I've noticed now is that both in naming and errors there's a lot of personal pronouns (you, my, etc.) This form of address doesn't seem very accurate for checks that everyone is running. Concrete example: someone builds a custom client and attempts to change permissions in a way that's not allowed (say, giving themselves the `admin` permission). While this should be rejected, it'd be confusing for other people to get a message having 'you' in it. Of course, it's fine to say 'you' in the UI, so long as only the user performing the action receives it.

corrideat (dismissed):
Looks great!

PR #2900, opened 2025-08-25
Initial fix. Closes #2898.

### LINKED ISSUE #2898: Issue with unauthorized/expired keys
### Problem

Another user sent a DM to me but I did not receive it. Instead, I received a "key unauthorized" notification. There was otherwise no indication that any messages were sent or missing. 

Image

### Solution

1. Find the actual problem.
2. Fix it :-)

corrideat:
Closes #2898

TODO:

- ~~Verify this actually works (it seems to from a quick test)~~
- ~~Verify that in case of a key rotation while sending a message, the message gets re-attempted with the correct, new key. (Example: message gets sent with `csk1`, but fails due to a conflict and in the meantime `csk1` was updated to `csk2`, the new attempt should use `csk2`)~~
- ~~Make this backwards compatible (meaning that old contracts should still work _and_ to add a post upgrade hook to encrypt CSKs with PEKs)~~
- ~~For some reason the `_private` key attribute (which contains a key ID for the (outer) key that should be used (i.e., is pinned) for encrypting that particular key) contains a key ID. Double that this key ID properly gets updated as encryption keys are rotated. Example: `k1` has `id(ek1)` as `_private`. Then, this attribute should get updated to `id(ek2)` if `ek1` is rotated and replaced with `ek2`.~~

### DIFF
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -1069,7 +1069,7 @@ export default (sbp('sbp/selectors/register', {
   // This function is called from 'state/vuex/postUpgradeVerification'
   // Group (foreign) CSKs that were encrypted with a CEK will be removed and
   // re-added being encrypted with the PEK. This allows other group members
-  // to see which foreign keys exist in our identity contract, which eliminates
+  // to see which foreign group CSK keys exist in our identity contract, which eliminates
   // the need for 'guessing' and resolves the issue of created DM chatrooms
   // not showing up
     const state = sbp('chelonia/rootState')
--- a/frontend/controller/actions/identity.js
+++ b/frontend/controller/actions/identity.js
@@ -757,14 +757,34 @@ export default (sbp('sbp/selectors/register', {
         // share a group
         signingKeyId: await sbp('chelonia/contract/suitableSigningKey', partnerIDs[index], [SPMessage.OP_ACTION_ENCRYPTED], ['sig'], undefined, ['gi.contracts/identity/joinDirectMessage']),
         // See issue #2898. This works as follows:
-        // If using a new version of the app which uses the PEK to encrypt the
-        // foreign group CSK so that it's visible to members, suitableSigningKey
-        // will return a key ID.
-        // If using an older version of the app, in which the foreign group CSK
-        // isn't visible, suitableSigningKey will return undefined.
-        // innerSigningKeyId takes precedence over innerSigningContractID, and
-        // an inner signing key will be used as a fallback
-        // (from innerSigningContractID) if innerSigningKeyId is undefined
+        //   1. If using a new version of the app which uses the PEK to encrypt
+        //      the foreign group CSK so that it's visible to members,
+        //      `suitableSigningKey` will return a key ID (which should be the
+        //      key corresponding to the CSK of a group we have in common with
+        //      this user). In this case, the key corresponding to this key ID
+        //      will be used (as set in `innerSigningKeyId`).
+        //      `innerSigningContractID`, which is set to a common group ID, is
+        //      ignored in this case.
+        //   2. If using an older version of the app, in which the foreign
+        //      group CSK isn't visible to us (*), `suitableSigningKey` will
+        //      return `undefined`.
+        //      In this case, the logic in `encryptedAction` will use
+        //      `innerSigningKeyName` (implicitly, `csk`) and `innerSigningContractID`
+        //      to determine the key to use in a foreign signature.
+        //      Note that this works because `encryptedAction` requires
+        //      `innerSigningKeyId` to be `null` to omit an inner signature if
+        //      `innerSigningContractID` is given.
+        //      Also note that this second case is exactly the same behaviour
+        //      the app had previous to this fix.
+        //
+        // This logic ensures maximum backward and forward compatibility, using
+        // the newer mechanism of using a key we know is valid when possible
+        // and falling back to guessing that a group CSK might be valid when
+        // that fails.
+        //
+        // (*) This situation should only happen if the app bundle for the other
+        // user hasn't been updated for some reason, or if the `postUpgradeVerification`
+        // hook failed to run for some reason.
         innerSigningKeyId: await sbp('chelonia/contract/suitableSigningKey', partnerIDs[index], [SPMessage.OP_ACTION_ENCRYPTED + '#inner'], ['sig'], undefined, ['gi.contracts/identity/joinDirectMessage#inner']),
         innerSigningContractID: currentGroupId,
         hooks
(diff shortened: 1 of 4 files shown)


corrideat:
To test this:

> Verify this actually works

Try the reproduction steps on https://github.com/okTurtles/group-income/issues/2898#issuecomment-3210813695 on master vs this PR. Other than the issue not happening, one can also manually verify that the foreign group CSKs are visible to others by having others check the state of the identity contract.

ETA: A slight variation on those steps is below.

> Verify that in case of a key rotation while sending a message, the message gets re-attempted with the correct, new key

This is a bit tricky to test as it involves a race condition. After giving it some thought, the following seemed to work. Two browsers are required (or two profiles / container tabs).

On one browser, while logged in, run the following, while keeping the console open:

```js
;(async () => {
    await sbp('gi.actions/identity/setAttributes', {
        data: {}, contractID: identityContractID,
        hooks: {
            beforeRequest(newEntry, entry) {
                debugger
            }
        }
    })
})()
```

Execution should pause. If desired, make a note of the signing key corresponding to `entry`.

Then, on the second browser, and logged in as the same user, run the following (this will rotate the PEK and the CSK):

```js
;(async () => {
    await sbp('chelonia/contract/setPendingKeyRevocation', identityContractID, ['pek', 'csk'])
    await sbp('gi.actions/out/rotateKeys', identityContractID, 'gi.contracts/identity', 'pending', 'gi.actions/identity/shareNewPEK')
})()
```

Now, back on the first browser, resume execution. The debugger should briefly pause again. Note that now `newEntry` is using a different signing key from `entry`. This means that the new CSK is being used.

> Make this backwards compatible

There are two different things to test here:

1. Start off by creating a group in `master` with three members `a`, `b` and `c`. Close `b`'s browser and wait for about 30s or manually stop the SW (note: don't grant notification permissions, we don't want the SW starting up again in the next step). With `b`'s SW stopped, have `c` leave the group. Stop the server. Save the database and make a backup.
2. Switch to this branch and comment out `innerSigningContractID: currentGroupId,` in `identity.js` (this is to test the post upgrade hook). Start the server.
3. Open up `b`'s session again. This will ensure the upgrade code runs.
4. Have `a` create a DM with `b`. It should work.

Now:

1. Stop the server
2. Restore the DB from backup
3. Uncomment `innerSigningContractID: currentGroupId,`
4. Comment out `sbp('gi.actions/identity/upgradeGroupForeignCSKs', contractID, groupFKTuple)` in `state.js`. This is to test whether the old behaviour is preserved if there's no upgrade happening.
5. Open up `b`'s session again. The upgrade code won't run this time.
6. Have `a` create a DM with `b`. It should result in this issue again, which is what we want.  Without `innerSigningContractID: currentGroupId,`, the error would happen on `a` but not on `b`.

> For some reason the _private key attribute (which contains a key ID for the (outer) key that should be used [...]

The best way to test this is to look at the messages that are sent in the chain and see which key ID is being used for encryption (should be the latest PEK in this case).

frontend/controller/actions/identity.js
```diff
@@ -753,9 +753,19 @@ export default (sbp('sbp/selectors/register', {
...
           contractID: message.contractID()
         },
-        // For now, we assume that we're messaging someone which whom we
+        // For now, we assume that we're messaging someone with whom we
         // share a group
         signingKeyId: await sbp('chelonia/contract/suitableSigningKey', partnerIDs[index], [SPMessage.OP_ACTION_ENCRYPTED], ['sig'], undefined, ['gi.contracts/identity/joinDirectMessage']),
+        // See issue #2898. This works as follows:
+        // If using a new version of the app which uses the PEK to encrypt the
+        // foreign group CSK so that it's visible to members, suitableSigningKey
+        // will return a key ID.
```
taoeffect: ...Which key ID? Can the comment be updated to make that a bit clearer?

frontend/controller/actions/identity.js
```diff
@@ -753,9 +753,19 @@ export default (sbp('sbp/selectors/register', {
...
+        // If using an older version of the app, in which the foreign group CSK
+        // isn't visible, suitableSigningKey will return undefined.
+        // innerSigningKeyId takes precedence over innerSigningContractID, and
+        // an inner signing key will be used as a fallback
+        // (from innerSigningContractID) if innerSigningKeyId is undefined
```
taoeffect: This whole paragraph is a bit confusing to me:

- Why would the foreign group CSK not be visible? Isn't the idea that when older versions launch the app the keys will be rotated, and so they will be upgraded to use the PEK? So under what circumstance would this not be the case? (Can this be clarified in the comment?)
- OK, let's say that the foreign group CSK is not visible, what does that have to do with this `innerSigningKeyId`..? not clear to me from the comment.

Finally, the comment itself is confusing, as it says essentially, `an inner signing key will be used [...] if innerSigningKeyId is undefined`... what? Isn't that a self-contradictory statement? Could all of this be cleared up?

corrideat: > Why would the foreign group CSK not be visible? Isn't the idea that when older versions launch the app the keys will be rotated, and so they will be upgraded to use the PEK? So under what circumstance would this not be the case? (Can this be clarified in the comment?)

As the comment indicates, when using an _older version of the app_ (i.e., and app that hasn't been upgraded). If, as you say, the app is launched and the upgrade code run... then, it's a new version of the app (which included the upgrade). This also works in any case the upgrade didn't run for whatever ereason.

> OK, let's say that the foreign group CSK is not visible, what does that have to do with this innerSigningKeyId..? not clear to me from the comment.

Apparently the comment wasn't clear enough, but setting both `innerSigningContractID` and `innerSigningKeyId` looks (or should look) redundant. The reason for doing this, as the comment attempts to explain, is that one (ì.e., `innerSigningKeyId`) takes precedence over the other (i.e., `innerSigningContractID`). So, if the group CSK isn't visible, which means `innerSigningKeyId` will be undefined, then `innerSigningContractID` will be used. This fallback behaviour is exactly the current app behaviour.

> an inner signing key will be used [...] if innerSigningKeyId is undefined... what? Isn't that a self-contradictory statement?

No, it's not a self-contradictory statement, even if it might seem that way. The first part (inner signing key) refers to the key that will be used. The second part (`innerSigningKeyId`) refers to a way to determine which key will be used. There's some logic involved for determining which key, if any, will be used as an inner signing key, and `innerSigningKeyId` can be undefined.

> Could all of this be cleared up?

I guess so. It looked clear to me when I wrote it, and it still does, but perhaps because I'm familiar with how keys are chosen. Anyhow, the comment meant to state three things:

1. That `innerSigningKeyId` and `innerSigningContractID` are both deliberately being set
2. The reason for doing this (fallback for old apps)
3. What setting both does (i.e., `innerSigningKeyId` used first and if undefined then `innerSigningContractID` is used)

frontend/controller/actions/identity.js
corrideat: It should be the group CSK ID, but `suitableSigningKey` just returns a suitable key. I.e., some other key may be used if it's found. This is not just theoretical, as, for example, people may have several groups in common, and in this case one of those CSKs will be used (but not necessarily the one from a specific group).

This is to say, the comment was vague on purpose, but I can clarify that we expect this to be the group CSK.

frontend/controller/actions/identity.js
```diff
@@ -1053,5 +1063,69 @@ export default (sbp('sbp/selectors/register', {
...
+  ...encryptedAction('gi.actions/identity/setGroupAttributes', L('Failed to set group attributes.')),
+  'gi.actions/identity/upgradeGroupForeignCSKs': (contractID: string, groupFKTuple: [string, string, string][]) => {
+  // See issue #2898
+  // This function is called from 'state/vuex/postUpgradeVerification'
+  // Group (foreign) CSKs that were encrypted with a CEK will be removed and
+  // re-added being encrypted with the PEK. This allows other group members
+  // to see which foreign group CSK keys exist in our identity contract, which eliminates
+  // the need for 'guessing' and resolves the issue of created DM chatrooms
+  // not showing up
+    const state = sbp('chelonia/rootState')
```
taoeffect: Nit: could you align the spacing on the comments and the code?

corrideat:
Testing this on staging I noticed that now I incorrectly have two valid group CSKs; one which seems to be an older group key (but still valid on my contract, the red / first key), encrypted with the CEK, and a newer key, encrypted with the PEK.

image

corrideat:
Closing this PR as the approach that we decided on makes it redundant. A new PR will be opened with the new implementation.

PR #2901, opened 2025-08-26
Chatroom member list shouldn't show members that have left

### LINKED ISSUE #2893: Removed member still shows up in a chatroom's member list
### Problem
[Steps to reproduce]

1. creates users `u1`, `u2` and `u3`. `u1` creates a group and invites `u2` and `u3`
2. `u1` creates a private-chatroom and invites `u2` and `u3`
3. `u1` kicks out `u2` from the chatroom. Chatroom title correctly says current there is currently 2 members.
4. But if you click it to open`ChatMembersAllModal` modal, the list still shows `u2`.

### Solution
Found out the cause and fix it.

SebinSong:
closes #2893 

Can see the removed member is properly moved to the section below.

image

### DIFF
--- a/frontend/views/containers/chatroom/ChatMembersAllModal.vue
+++ b/frontend/views/containers/chatroom/ChatMembersAllModal.vue
@@ -167,6 +167,10 @@ export default ({
     ...mapState([
       'currentGroupId'
     ]),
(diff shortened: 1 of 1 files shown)


taoeffect (approved):
Nice 👍

Issue #2735, opened 2025-03-24
Very delayed "new distribution period" notification


taoeffect:
### Problem

![Image](https://github.com/user-attachments/assets/ee5b0386-856f-43d7-97a9-a64330a3f913)

As you can see in the screenshot, the new distribution started on the 18th, but the notification came in on the 23rd.

### Solution

Not sure what the solution is, but a fix is needed!

dotmacro:
This is still occurring, and the delay is (potentially) increasing, as this period's notification was received 10 days after start.

![Image](https://github.com/user-attachments/assets/bc7bcbec-6509-4750-a73a-aa9d987ce312)

dotmacro:
<img width="788" height="214" alt="Image" src="https://github.com/user-attachments/assets/d4d36713-3a8b-4ccf-b2d7-b4de92f726b4" />

I just received a notification that's more than two months delayed. I've logged out and logged back in multiple times since the payment period in the notification. At this point, I do not remember whether I received a timely notification about the same payment period.

This happened after a resync. This was the only delayed notification I received during the resync.

taoeffect:
Context for ☝️ this appears to have happened after a full resync.

dotmacro:
Is it possible that this notification is being triggered when the distribution is locked (i.e. when the first payment is made) rather than when the period begins? That would not explain long-delayed notifications. It could explain missing notifications (if no one makes a payment) or delayed notifications received late but during the correct period. 

Will test and report back.

dotmacro:
After testing, it seems like the `A new distribution period has started` notification is triggered by the first payment being made, rather than the start of the period. 

To reproduce:
1. Create a group with 2 pledging members above mincome, one below mincome
2. Adjust system clock to be sometime mid-period
3. Wait to ensure group syncs 
4. Mark 1 payment as sent
5. See that "A new distribution period has started" notifications instantly appear for all three members

PR #2908, opened 2025-09-10
#2735 - Delayed distribution period notification

SebinSong:
Fix attempt for #2735

### DIFF
--- a/frontend/model/notifications/mainPeriodicNotificationEntries.js
+++ b/frontend/model/notifications/mainPeriodicNotificationEntries.js
@@ -91,8 +91,12 @@ const periodicNotificationEntries: {
           const currentPeriod = rootGetters.groupSettingsForGroup(rootState[gId]).distributionDate
           if (!currentPeriod) { return null }
 
+          const nextPeriod = rootGetters.periodAfterPeriodForGroup(rootState[gId], currentPeriod)
+          const now = dateToPeriodStamp(new Date())
+          const isPeriodRelevant = comparePeriodStamps(now, currentPeriod) > 0 && comparePeriodStamps(now, nextPeriod) < 0
+
           return (
-            comparePeriodStamps(dateToPeriodStamp(new Date()), currentPeriod) > 0 &&
+            isPeriodRelevant &&
               !myNotificationHas(item => item.type === 'NEW_DISTRIBUTION_PERIOD' && item.data.period === currentPeriod, gId)
           )
             ? [gId, currentPeriod, profile.incomeDetailsType]


frontend/model/notifications/mainPeriodicNotificationEntries.js
```diff
@@ -91,8 +91,12 @@ const periodicNotificationEntries: {
           const currentPeriod = rootGetters.groupSettingsForGroup(rootState[gId]).distributionDate
           if (!currentPeriod) { return null }
 
+          const nextPeriod = rootGetters.periodAfterPeriodForGroup(rootState[gId], currentPeriod)
+          const now = dateToPeriodStamp(new Date())
+          const isPeriodRelevant = comparePeriodStamps(now, currentPeriod) > 0 && comparePeriodStamps(now, nextPeriod) < 0
+
           return (
-            comparePeriodStamps(dateToPeriodStamp(new Date()), currentPeriod) > 0 &&
+            isPeriodRelevant &&
```
SebinSong, line 99: This is the logic that determines to emit `NEW_DISTRIBUTION_PERIOD` notification and the previous logic does check if the target period start date is `< today` but is missing a check whether it is no longer relevant. So have updated the logic to check this too.

---

I've been taking a while trying to reproduce this issue locally but had no luck so far, so have not verified the fix myself locally. Hopefully, if this looks good, we could deploy it to the PROD and see if the issue stops happening.

dotmacro: Just concurring: I also cannot reliably trigger/reproduce this in testing.

taoeffect: @SebinSong I don't quite understand the logic here, are you sure it's correct?

I see that the new condition being added is `&& comparePeriodStamps(now, nextPeriod) < 0`

But this doesn't quite make sense to me, because it's asking, "Is it currently before the next period?"

But if we are currently ahead of the current period (in the next period), then how can we also be before the next period?

In other words, if the previous period was 1, and `now` is in period `2`, then how we be before `2`...? That doesn't make sense.

Do you mean `>` instead of `<`? (ahead of the next period?)

And are you sure this will fix it? Because:

1. we want to make sure that the notification also doesn't appear multiple times in the current period.
2. And the issue claims that the notification came in on the 23rd when it should have come in shortly on/around the 18th... so the issue is that there was no notification at that time... not that there were multiple notifications

SebinSong: @taoeffect 
I actually had the same confusion too when I first saw the existing logic. 
The confusion came from the function name `dateToPeriodStamp`. It sounds like `dateToPeriodStamp(new Date())` will convert it to be exactly the same as current period stamp but what it actually does is calling `.toISOString()` method.
So `comparePeriodStamps(now, currentPeriod)` actually calculates **the number days apart** between Today and the **start-date** of the current period.

This was referenced from this existing logic here, FYI:
https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/mainPeriodicNotificationEntries.js#L38-L43

---

The idea of this PR is pretty much to makes sure the `now` falls within below range if we visualize it as a timeline.

image


<br>
<br>

I was not able to reproduce the issue locally but my **guess** on why the issue happens was the currentPeriod calculated here(which is below):

```
const currentPeriod = rootGetters.groupSettingsForGroup(rootState[gId]).distributionDate
```
can be **far-old** value for some reason sometimes(Am not an expert here but eg. `rootState` here could be stale data by the time this logic is executed).
To visualize again, I thought below is what might have happened:

image


<br>
<br>

And since the existing logic below

https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/mainPeriodicNotificationEntries.js#L95

only checks if `now > start-date of the period`, thought additional logic to rule out below case would fix the issue.

image

<br>
<br>

Hope this clarifies things. Cheers,

SebinSong: My bad, I just found out this PR is trying to solve the specific issue mentioned in this particular comment:
https://github.com/okTurtles/group-income/issues/2735#issuecomment-3272200564

Also,

> we want to make sure that the notification also doesn't appear multiple times in the current period.

It does make sure of it currently:
https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/mainPeriodicNotificationEntries.js#L96

taoeffect: OK, I see what you're saying now. I suppose this change could [fix this issue](https://github.com/okTurtles/group-income/issues/2735#issuecomment-3272200564), but it won't [fix this issue](https://github.com/okTurtles/group-income/issues/2735#issuecomment-2833987902).

Here's my guess as to what's happening in that latter example: @dotmacro probably hadn't opened the site shortly after entering the new distribution period, and the webpage had probably been unloaded by the browser. But when she opened the site later (~10 days after) the code loaded up and triggered the notification.

So the issue (I'm guessing) is that this code is probably running in the browser tab, and not the service worker. Can you double-check this? If it were running in the service worker, it would be better, as then there might be a greater chance that a push-notification style notification could be posted. In fact, that's why you see the "site updated in the background" notifications - because the service worker is being periodically woken up by the server. That would give the service worker the opportunity to run these notification checks and post a notification. You could then properly get a notification from e.g. the PWA. I think we might even have related notification for that.

Could you investigate this question? Feel free to ask @corrideat if you have questions about how that works.

SebinSong: @taoeffect Here are what I've checked and confirmed from the investigation:

- Does service-worker have this `periodic check -> emit notification` logic running? -> yes [here](https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/controller/serviceworkers/sw-primary.js#L304-L312)
- is there a code to send a service-worker specific notification if the logic is running in SW? -> yes [here](https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/controller/serviceworkers/sw-primary.js#L557-L573) and [here](https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/nativeNotification.js#L137-L148)

Since below SW logic you mentioned:

> the service worker is being periodically woken up by the server.

is currently working correctly too (I saw "site updated in the background" just a few mins ago while writing this comment), It does not seem like there is any missing logic in SW for checking&sending that new distribution period notification.

---
This section is about my **guess** as to why it is happening.

I feel like this particular notification might have been emitted **twice**. (Once in service-worker first and then once again when Andrea logged back in after a long time)

As you said too, this periodic-notification logic is run in the browser as well after the user logs in **[1]**:
https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/main.js#L278-L291

There is also part where it makes sure the notification is added to the Vuex store too when emitted, so it is correctly reflected to the in-app notification UI:

https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/vuexModule.js#L12-L18

But apparently this syncing process(notification emitted in SW -> reflected in Vuex -> appears in in-app notification UI) is only happening while the user is online using the browser. (Because these `NOTIFICATION_EMITTED` and `NOTIFICATION_REMOVED` event listeners are set up while Vuex store is initialized)

So what happened could be:
1. New period notification was sent and presented as a service-worker notification.
2. After a while, Andrea logged in the app via a browser
3. There is no existing in-app notification for this particular 'new-period starts' notification
4. The periodic-check logic in **[1]** runs, the check for any duplicated notification item passes as there is no existing item.
5. A delayed notification is sent.

So, I feel we need to see if the logic for:

`ensuring the notification sent from the SW is correctly reflected in the Vuex when the user logs back in after that`

is missing or is working correctly.

The `rootState.notifications.items.splice(Math.max(0, index), 0, notification)` below is where a new notification item is added to the state loaded via `sbp('chelonia/rootState')`. Maybe the notification item added this way might not be loaded/merged to Vuex store when the app loads in the browser afterwards.

https://github.com/okTurtles/group-income/blob/f7aa6cb88626b8388df304ccbaeb0257b14d2f57/frontend/model/notifications/selectors.js#L42-L51

---

These are all my guess but If you agree, since this is the area where Ricardo has the most expertise, I believe we need his help with investigation. (Hope my time & effort on this were worth it though..!)

Thanks,

taoeffect: Nice investigative work @SebinSong!

I'm specifically wondering about whether or not the recurring heartbeat properly triggers this notification. I've traced it this far:

`serviceworkers/push.js` sets up:

```js
heartbeatInterval: 12 * 60 * 60 * 1000
```

`backend/server.js` sends:

```js
const notification = JSON.stringify({ type: 'recurring' })
```

`serviceworkers/push.js` then:

```js
  } else if (data.type === 'recurring') {
    event.waitUntil(
      sbp('gi.periodicNotifications/init')
    )
  }
```

Which calls `runNotificationListRecursive()`.

So the question then becomes, does `NEW_DISTRIBUTION_PERIOD` properly get triggered from there?

Because you talk about Vuex, but remember that Vuex is only available when the app has a window/tab open to it. Otherwise the SW runs in the background and there is no Vuex. But we still want the push notification to display the notification to the user (e.g. on their phone).

dotmacro: > I feel like this particular notification might have been emitted **twice**. (Once in service-worker first and then once again when Andrea logged back in after a long time)

Do you mean "logged in" or "opened"? I've opened and used the app a minimum of once per period since the beginning, but log in much less frequently. The delayed notifications do not always appear immediately after using the app.

SebinSong: @dotmacro Yes, that would be a better wording in this context.

SebinSong: @corrideat When you have time, could you help Greg with answering his question [here](https://github.com/okTurtles/group-income/pull/2908#discussion_r2380679119)? Since those are the SW related logics you implemented, I think you are a better person for this issue.

corrideat: @SebinSong The short answer is the same as what @taoeffect said, that native notifications (especially when triggered by a server push) _must_ happen in the service worker. Vue isn't available in the service worker, and there's no way for the service worker to open the app (and if there were, it wouldn't be good UX unless a triggered by a user action).

As for some background here, the 'hearbeat' notification is a workaround to regularly send a push notification to the app, so that the SW is woken up and periodic notifications may be shown. This isn't necessarily the 'best' way, since it can result in forced notifications (e.g., 'This app has been updated in the background'), but something that seems to work relatively reliably.

SebinSong: @corrideat 
Thanks for the detailed explanation.
In [#2735 Very delayed "new distribution period" notification](https://github.com/okTurtles/group-income/issues/2735), One periodic-notification item is shown as a delayed notification in the app. As Greg said above, it possibly suggests that there are cases where this heartbeat notification logic does not work as intended. Could you investigate this when you have the capacity? FYI, This PR I opened is a potential fix for this particular behaviour described in [this comment](https://github.com/okTurtles/group-income/issues/2735#issuecomment-3272200564), So please feel free to open a separate PR and re-assign yourself to the issue.

corrideat: I can try to have a look. At the moment, I only took a quick look at the issue. However, since both the native notification (?) and the in-app notification seem to be affected, the issue likely is in the logic to trigger this notification (which would be in the SW, if there is to be a native notification)

SebinSong: Thanks @corrideat

corrideat: Taking a look, `'gi.periodicNotifications/importNotifications'` is called from the SW and will import all notifications from `mainPeriodicNotificationEntries.js`. The subsequent call to `'gi.periodicNotifications/init'` in the SW is meant to run all notifications, so, in principle, it should all work (I haven't verified that it does, but if it doesn't, then there's a bug on either `mainPeriodicNotificationEntries.js` (i.e., not triggering a notification if it should), `'gi.periodicNotifications/importNotifications'` (i.e., failing to import the logic correctly) or in `'gi.periodicNotifications/init'` (i.e., failing to run the logic that would result in a notification)).

corrideat: I see my previous comment may be confusing, so let me rephrase:

This PR changes `mainPeriodicNotificationEntries.js`. This is the correct place to make changes that run in the SW, since these are imported from the SW (in fact, this file is only imported from the SW).

If this notification still fails to run, then there could be other issues in `gi.periodicNotifications` that cause this logic not to run.

taoeffect (approved):
Let's hope this fixes it... I'm skeptical but let's see

